PR #1418, opened 2020-07-19
Fix WXKG01LM if no click up event is received

ArkadiuszNiemiec:
Added timeout equals to 5 seconds timeout after `click down`. If no `click up` event is received (e.g. it was sent before `click up` event) it will reset the `long` variable so the next click is not treated as `click up` of previous one.


Koenkk:
Thanks!

PR #1426, opened 2020-07-23
Corrected battery information from Develco Smoke detector

paddenstoeltje:
To get a correct battery reading 'fz.battery_cr2032' should be used iso 'fz.battery'.


Koenkk:
I've refactored the code a bit, behaviour should stay the same, can you verify?

paddenstoeltje:
> 
> 
> I've refactored the code a bit, behaviour should stay the same, can you verify?

Seems to be working.

Wouter

Koenkk:
Thanks!

PR #1430, opened 2020-07-23
Fix Wally U02I007C.01 configuration

presslab-us:
Attribute report config fails for onOff with UNSUPPORTED_ATTRIBUTE.
Device does not need this to be configured to operate properly.


Koenkk:
Thanks!

Issue #1274, opened 2020-06-02
Request to add device！


unpack8866:
Request to add device！

[link：](https://www.xiaomiyoupin.com/detail?gid=118897&spmref=YouPinPC.$SearchFilter$1.search_list.3.99223308)

Model ：QBKG24LM
Vendor： Xiaomi
Description：Aqara double key wired wall switch D1
Supports：on/off, power measurement, temperature

    {
        zigbeeModel: ['lumi.switch.b2nacn02'],
        model: 'QBKG24LM',
        vendor: 'Xiaomi',
        description: 'Aqara double key wired wall switch D1',
        supports: 'on/off, power measurement, temperature',
        fromZigbee: [
            fz.QBKG03LM_QBKG12LM_LLKZMK11LM_state, fz.QBKG12LM_LLKZMK11LM_power, fz.QBKG03LM_QBKG12LM_operation_mode,
            fz.QBKG12LM_click, fz.ignore_multistate_report, fz.xiaomi_power,
        ],
        toZigbee: [tz.on_off, tz.xiaomi_switch_operation_mode],
        endpoint: (device) => {
            return {'left': 1, 'right': 2, 'system': 1};
        },
        onEvent: xiaomi.prevent_reset,
    },

Koenkk:
- Does changing the operation mode work? (https://www.zigbee2mqtt.io/devices/QBKG03LM.html#decoupled-mode)
- Does it also work ok with the config below? :

```js
{
    zigbeeModel: ['lumi.switch.b2nacn02'],
    model: 'QBKG24LM',
    vendor: 'Xiaomi',
    description: 'Aqara double key wired wall switch D1',
    supports: 'on/off, power measurement, temperature',
    fromZigbee: [fz.on_off, fz.xiaomi_power, fz.QBKG03LM_QBKG12LM_operation_mode],
    toZigbee: [tz.on_off, tz.xiaomi_switch_operation_mode],
    meta: {multiEndpoint: true},
    endpoint: (device) => {
        return {'left': 1, 'right': 2, 'system': 1};
    },
    onEvent: xiaomi.prevent_reset,
},
```

unpack8866:
I try to add three key switch D1;

How to modify device.js

Koenkk:
Is this the QBKG25LM? Should already be supported: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L730

unpack8866:
[...]

PR #1435, opened 2020-07-26
Update configuration for Aqara D1 wired wall switches

zhongfu:
I've removed the temperature and power measurement converters from the configuration for the QBKG22LM/QBKG25LM; it seems that they're not providing any actual readings, which would be in line with the behavior with the last-gen Aqara wired, no-neutral wall switches.

I've also added these settings to the QBKG25LM:
- `operation_mode` (i.e. decoupled mode),
- `power_outage_memory` (as in the ZNCZ04LM), and
- `do_not_disturb` (turn off LEDs between 9pm and 9am)

`operation_mode` has to be set with the endpoint corresponding to the switch you'd like to enable decoupled mode for (left/center/right). The last two can only be set with endpoint 1 (left), but changes will apply for all switches.

I've yet to test `do_not_disturb` with Z2M (gotta wait for 9pm...), but the switch does report the new value of the attribute when it gets changed. The other two works though.

More details are available in the commit messages.

Probably related to #1274, but I don't have the QBKG26LM on hand to verify and/or check for any additional features.


devices.js
```diff
@@ -653,10 +653,10 @@ const devices = [
...
         vendor: 'Xiaomi',
         description: 'Aqara D1 2 gang smart wall switch (no neutral wire)',
-        supports: 'on/off, action, power measurement',
+        supports: 'on/off, action',
         fromZigbee: [
             fz.on_off_xiaomi_ignore_endpoint_4_5_6, fz.xiaomi_on_off_action, fz.legacy_QBKG03LM_QBKG12LM_click, fz.QBKG03LM_buttons,
-            fz.QBKG03LM_QBKG12LM_operation_mode, fz.generic_device_temperature,
+            fz.QBKG03LM_QBKG12LM_operation_mode,
         ],
         toZigbee: [tz.on_off, tz.xiaomi_switch_operation_mode],
```
Koenkk, line 661: I see that QBKG25LM_operation_mode differs from xiaomi_switch_operation_mode, since these are both the same switches I expect the commands to be similar, does the operation mode actually work for the QBKG22LM now?

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt/pull/3977/files and https://github.com/Koenkk/zigbee2mqtt.io/pull/354/files

devices.js
zhongfu: That's right -- the QBKG22LM seems more similar to my QBKG04LMs than the QBKG25LM (e.g. no double/triple click but has hold event, doesn't turn on then off if I double click, doesn't enter pairing process if it's connected to a coordinator), which leads me to believe that hardware-wise the 1- and 2-gang D1 switches are practically the same as the 1st gen switches

zhongfu: also interestingly, the QBKG25LM's depth is less than that of the QBKG22LM and QBKG04LM iirc

Koenkk:
I've merged the converters with the existing ones were possible, can you check if everything still works? Then this can be merged.

zhongfu:
I can confirm that setting `operation_mode` still works for my QBKG22LM and my QBKG25LM.

As for `power_outage_memory` and `do_not_disturb`, I do get an Report Attributes packet back from the QBKG25LM when setting them. I'm not able to verify if they actually have any effect though: I can't turn the power off now, and `do_not_disturb` doesn't seem to be taking effect even though it's currently between 9pm and 9am local time here.

I took a look through the dumps I made of the communications between the Aqara Hub and the 
QBKG25LM, and there seems to be some differences between the behaviors of the Aqara Hub and Z2M:  
- even though the QBKG25LM requests for Time (cluster 0x000a attr 0x0000) and Time Zone (attr 0x0002) in the Read Attributes request, the coordinator/Z2M doesn't respond with the time zone
- the coordinator/Z2M responds with the correct UTC time (without the tz, of course), but the Aqara Hub responds with the local time as UTC time (e.g. 22:00Z instead of 20:00Z if the current time is 22:00+0200) and tz (attr 0x0002) = 0

I'll take a look again later when it's between 9pm and 9am UTC -- hopefully it works even if it doesn't receive the TZ from the coordinator

Koenkk:
@zhongfu let me know otherwise we can probably add this.

zhongfu:
Doesn't seem like `do_not_disturb` is working, unfortunately. Maybe it really does need to know the timezone for it to work

Koenkk:
Latest zigbee2mqtt dev branch should respond to timezone reads now.

zhongfu:
I can confirm that `do_not_disturb` works now, thanks!

image

Koenkk:
Great, to double check, is the timeZone value send by zigbee2mqtt the same as the one send by the gateway? (want to double check my implementation).

zhongfu:
Nope, the Aqara hub sends the local time for the time attr, and 0 for the time zone. The indicator lights also came back on at 9am, so it seems like it's working as intended.

(as a side note, it's probably possible to make it keep the indicators off all the time by sending it an incorrect time/timezone, but it's probably too hacky to be worth doing)

Koenkk:
I've changed the api for QBKG25LM to `{"operation_mode": {"state": VALUE}}` instead of `{"operation_mode": VALUE}` to keep it the same as others.

PR #1437, opened 2020-07-27
Add support for BTicino F20T60A

EPMatt:
This PR adds support for BTicino F20T60A DIN power consumption module, which is actually a Legrand 412015 module.

The issue was that the BTicino module identifies itself as ` DIN power consumption module` without the trailing NULL characters, and because of that it was not recognized as a supported device.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -10713,7 +10713,7 @@ const devices = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1438, opened 2020-07-28
Device T21W1Z showing ConfigureReporting errors

azamat5991:
Because of that line `zigbee2mqtt `is showing error messages, like `Failed to configure 'device1', attempt 2 (Error: ConfigureReporting .. genOnOff([{"attribute":"onOff","minimumReportInterval":0 ...`.


Koenkk:
Thanks!

PR #1443, opened 2020-07-30
Add support for Enbrighten in-wall smart switch model 43076

Inshakoor:
Added Enbrighten in-wall smart switch model 43076.


Koenkk:
Thanks!

PR #1445, opened 2020-07-31
NLG Q-REMOTE improvements

arodus:
I wasn't sure about the naming convention of `nlg_remote_command_set_scene` so I'm open for suggestions. 

The command is fired whenever the `.` or `. .` button is pressed where `scene_id` is the "id" of the button. 
`.` => scene_id = 1
`. .` => scene_id = 2

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -6128,7 +6128,7 @@ const devices = [
             fz.command_on, fz.command_off, fz.command_toggle, fz.command_step,
             fz.command_move_to_color_temp, fz.command_move_to_color, fz.command_stop,
             fz.command_move, fz.command_color_loop_set, fz.command_ehanced_move_to_hue_and_saturation,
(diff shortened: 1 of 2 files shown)


devices.js
```diff
@@ -6128,7 +6128,7 @@ const devices = [
             fz.command_on, fz.command_off, fz.command_toggle, fz.command_step,
             fz.command_move_to_color_temp, fz.command_move_to_color, fz.command_stop,
             fz.command_move, fz.command_color_loop_set, fz.command_ehanced_move_to_hue_and_saturation,
-            fz.tint404011_scene,
+            fz.tint404011_scene, fz.nlg_remote_command_set_scene,
```
Koenkk, line 6131: Can you try using `fz.command_recall` instead of adding a custom converter?

arodus: Works 👍

Koenkk:
Thanks!

PR #1446, opened 2020-08-01
Change TuYa Led Controller TS0503A & Add Aqara 1 Gang Smart Switch (with N) QBKG23LM

wolverinevn:
+ Add converters for TuYa Led Strip Controller TS0503A
+ Add device TuYa Led Strip Controller TS0503A
+ Add devices Aqara 1 Gang D1 Smart Switch (with Neutral line) QBKG23LM


Koenkk:
Thanks, I've changed the api from `{"color":{"h": 360, "s": 100}}` to `{"color":{"hue": 360, "saturation": 100}}` as I feel it's more clear.

wolverinevn:
Home Assistant does't seem to support hue and saturation by default. It use **h** and **s** instead.

Koenkk:
Thanks, reverted that to h/s

devices.js
```diff
@@ -11400,6 +11400,16 @@ const devices = [
             await configureReporting.rmsVoltage(endpoint);
         },
     },
+    // Tuya TYZS1L Led Strip Controller
+    {
+        zigbeeModel: ['TS0503A'],
+        model: 'TYZS1L',
```
blakadder: TYZS1L is the model name of the Zigbee module used not the device itself

wolverinevn: Right. Can't find the model so I guess we can just put it there likes "Led Controller TYZS1L".

blakadder: what's wrong with TS0503A which is the recognisable "model"

Koenkk:
@wolverinevn can you share the data/database.db entry of the TS0503A?

PR #1454, opened 2020-08-04
Neo Zigbee Siren Alarm https://szneo.com/en/products/show.php?id=241

kirovilya:


converters/fromZigbee.js
```diff
@@ -5569,6 +5569,49 @@ const converters = {
...
+    neo_t_h_alarm: {
+        cluster: 'manuSpecificTuyaDimmer',
+        type: ['commandSetDataResponse', 'commandGetData'],
+        convert: (model, msg, publish, options, meta) => {
+            const dp = msg.data.dp;
+            const data = msg.data.data;
+            const dataAsDecNumber = utils.convertMultiByteNumberPayloadToSingleDecimalNumber(data);
+            switch (dp) {
+            case 360: // 0x0168 [0]/[1] Alarm!
+                return {alarm: (dataAsDecNumber==0) ? 'OFF' : 'ON'};
```
Koenkk, line 5581: Instead of `OFF`/`ON`, doesn't it make more sense to return `true`/`false`?

kirovilya: ok

Koenkk:
Thanks!

p.s. can you also write some doc on what value can be used for e.g. `melody` and what `duration` means? I think this will help others a lot.

PR #1463, opened 2020-08-10
multiple Paulmann models with same zigbeeid

blakadder:
per https://github.com/ioBroker/ioBroker.zigbee/issues/586#issuecomment-650319541

i would make and educated guess that all 3 models (https://en.paulmann.com/search?sSearch=amaris) share the same zigbee id


Koenkk:
Thanks!

PR #1461, opened 2020-08-10
add 6 devices (2 xiaomi , 4 TuYa)

zhouF96:
https://www.aqara.com/cn/productDetail/d23 (HLQDQ01LM)

https://www.aqara.com/cn/productDetail/d24  (ZNTGMK11LM)

https://go.tuya.com/productDetail?code=82qhk2e8matq   (HY0080)

TS0004 is the 4 gang switch model 
D3-DPWK-TY is oemd thermostat controller BY http://www.chinatender.com/ （using TUYA's 6dfgetq zigbee module）

and TS0502A is TUYA's zigbee module which designed for spot lights with color temperature support

PR #1464, opened 2020-08-11
tuya_thermostat (TS0601): modes changed to presets and two other fixes

mgrom:
This PR is related to: https://github.com/Koenkk/zigbee2mqtt/pull/4070

- window detection fix - before there was no way to change this value through mqtt command,
- changed preset to away preset params - something named preset was actually away preset mode params (temp, and duration in days),
- modes changed to presets - modes in home assistant are fixed, so the best to be flexible is to use presets (ha calls them hold_modes)

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -312,14 +312,20 @@ const tuyaThermostat = (model, msg, publish, options, meta) => {
         return {away_preset_temperature: dataAsDecNumber};
     case 629: // 0x7502 away preset number of days
         return {away_preset_days: dataAsDecNumber};
-    case 1028: // 0x0404 Preset changed
-        if (utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatPreset').hasOwnProperty(dataAsDecNumber)) {
-            return {preset: utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatPreset')[dataAsDecNumber],
-                system_mode: utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatSystemMode')[dataAsDecNumber]};
+    case 1028: {// 0x0404 Preset changed
+        const ret = {};
+        const presetOk = utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatPreset').hasOwnProperty(dataAsDecNumber);
+        const modeOk = utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatSystemMode').hasOwnProperty(dataAsDecNumber);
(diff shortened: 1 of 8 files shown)


Koenkk:
I've just merged https://github.com/Koenkk/zigbee-herdsman-converters/pull/1461 which conflicts with this pr, can you rebase this pr?

Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt/pull/4070

mgrom:
No problem, I'll do it today.

mgrom:
@Koenkk I'm only a bit worried about this change: 
https://github.com/Koenkk/zigbee-herdsman-converters/blob/470125e9b277eb4615b06cfc39dce83f76422391/converters/fromZigbee.js#L315
Because it can break currently working installations. As I mentioned in other PR, homeassistant is not allowing to add any custom system modes, so to be able to control mode we should use presets instead, which are more flexible. 
But this change can make some damage for user who currently use this integration without homeassistant.
Any ideas how to proceed?

Koenkk:
Returning both is fine!

mgrom:
I think it's ready @Koenkk

converters/fromZigbee.js
```diff
@@ -307,15 +308,16 @@ const tuyaThermostat = (model, msg, publish, options, meta) => {
...
-        return {preset: dataAsDecNumber};
-    case 1028: // 0x0404 Mode changed
-        if (utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatSystemMode').hasOwnProperty(dataAsDecNumber)) {
-            return {system_mode: utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatSystemMode')[dataAsDecNumber]};
+    case 626: // 0x7202 away preset temperature
+        return {away_preset_temperature: dataAsDecNumber};
+    case 629: // 0x7502 away preset number of days
+        return {away_preset_days: dataAsDecNumber};
+    case 1028: // 0x0404 Preset changed
+        if (utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatPreset').hasOwnProperty(dataAsDecNumber)) {
```
Koenkk, line 316: this changes breaks e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/eafa4a8dfc9298541827363b06c893297ccce9e2/devices.js#L1529

mgrom: I've forgot to check SYstemMode dictionary. Fixed

Koenkk:
Thanks!

PR #1466, opened 2020-08-11
Update devices.js - ROB_200-010-0

Chamondgl:
Small update removing an unneeded genPowerCfg bit and inverting cover status state reporting.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -5553,12 +5553,12 @@ const devices = [
         vendor: 'ROBB',
         description: 'Zigbee curtain motor controller',
         supports: 'open, close, stop, position',
-        meta: {configureKey: 1},
+        meta: {configureKey: 2},
         fromZigbee: [fz.cover_position_tilt],
         toZigbee: [tz.cover_state, tz.cover_position_tilt],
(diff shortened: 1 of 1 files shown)


Chamondgl:
I have an issue with action buttons being greyed out as well. https://github.com/Koenkk/zigbee-herdsman-converters/issues/1467

devices.js
```diff
@@ -5553,12 +5553,12 @@ const devices = [
         vendor: 'ROBB',
         description: 'Zigbee curtain motor controller',
         supports: 'open, close, stop, position',
-        meta: {configureKey: 1},
+        meta: {configureKey: 2},
```
Koenkk, line 5556: To fix https://github.com/Koenkk/zigbee-herdsman-converters/issues/1467 try changing this line to

```
meta: {configureKey: 2, coverInverted: true},
````

Chamondgl: Thanks. That doesn't seem to do anything. Even after a re-pair it doesn't invert the reported state.

Chamondgl: It seems to magically work now. What an adventure.

Koenkk:
Thanks!

PR #1471, opened 2020-08-13
add LTG001

typester:
Although I couldn't find any details on Philips sites, this bulb is still available on Amazon here: https://www.amazon.com/gp/product/B07WTJZCZL/


Koenkk:
Thanks!

PR #1472, opened 2020-08-14
Fix cover_position_tilt's returned position.

middelink:
When running with invert_cover unset or false, the zigbee position sent to device as 100-position. Working as intended. However it also reports back the same inverted value as the new state. This is wrong. When HA sets the position to 35, it gets back 65.
Even worse, if it want to open the cover by sending it 100 (=HA-open), it gets back a 0 (=HA-closed) which then disabled the down arrow, making it annoying to figure out what just happened.

Of course, when invert_cover=true, no translation happens and setting it to 35, returns 35.


Koenkk:
Nasty bug, thanks for the fix!

PR #1496, opened 2020-08-23
Add power metering to Innr SP 120

Appelg:
Add power consumption to Innr SP 120.
Will show up as `energy` after a couple of hours.

#588 https://github.com/Koenkk/zigbee2mqtt/issues/588#issuecomment-675684698


Koenkk:
Thanks!

PR #1500, opened 2020-08-24
Add support for new version of MKS-CM-W5 (2 gang)

joeldavidw:
New version of MKS-CM-W5 2 gang identified itself as TS0601.


Koenkk:
Thanks!

PR #1505, opened 2020-08-25
Device Support: TS110F Lonsonho 1-gang

ginkage:
Support for 2-gang dimmer was already there, and now both 1-gang and 2-gang dimmer modules will be supported (they share the same model ID, so the distinction is going to be made based on manufacturerName).


Koenkk:
Thanks!

PR #1508, opened 2020-08-26
Add ZLL-NonColorController (Trust ZYCT-202)

martin-g-it:
Add Trust ZYCT-202 support as alias ZLL-NonColorController


Koenkk:
Thanks!

PR #1509, opened 2020-08-26
Device support: Lonsonho 2-gang switch

ginkage:
This is a weird one (as usual with Lonsonho). Technically, it's the same model name as Zemismart ZM-L03E-Z (a 3-gang wall switch), but it's 2-gang, and without buttons. Oh well.


ginkage:
This is the device itself: https://www.aliexpress.com/item/4001178298316.html

Koenkk:
Thanks again! :smile:

PR #1510, opened 2020-08-27
ZB-CT01

arteck:
add colorxy


Koenkk:
The name and model suggest this is a color temperature only light, is this a variant of this device?

arteck:
not according to a user.. it has color and colortemp

Koenkk:
ok!

PR #1512, opened 2020-08-28
add support for TUYA zigbee module aabybja

zhouF96:
add support for TUYA's zigbee module aabybja  , used on Curtain motor


Koenkk:
Thanks!

PR #1513, opened 2020-08-28
Add parse power and temperature for QBKG23LM

hoangphison:
Add support for aqara d1 1 gang parse `power` and `temperature`


Koenkk:
Thanks!

PR #1514, opened 2020-08-28
Add SR-ZG9001T4-DIM-EU

hanswilw:
Add basic support for https://www.sunricher.com/single-color-4-groups-zigbee-touch-remote-controller-sr-zg9001t4-dim-eu.html

Not too familiar with all the different fz functions, but works fine with reporting and binding atleast.


Koenkk:
Thanks!

PR #1515, opened 2020-08-29
Update support for ROB_200-010-0

Chamondgl:
The vendor updated the module to identify it as 'ROB_200-010-0'. The version that identifies as 'Motor Controller' is still being sent out. Devices seem identical.
Added the extra zigbeeModel to devices.js. Tested locally and works perfectly.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -6001,7 +6001,7 @@ const devices = [
         ],
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1517, opened 2020-08-31
Add ignore temperature report to device Sinope TH1124ZB

embak:
Sinope Thermostat TH1124ZB send msTemperatureMeasurement reports. No converter is associated with this message and the temperature is already provided in the hvacThermostat report as local_temperature.


Koenkk:
Thanks!

PR #1519, opened 2020-09-01
Support for YSR-MINI-01 with different ZigBee ID

tbnobody:
My LED Controller shows YSR-MINI-01 on the case but uses a different ZigBee ID.


Koenkk:
Thanks!

PR #1525, opened 2020-09-03
Basic support on/off for Aqara 3 gang with neutral wire

hoangphison:
Add basic support for Aqara 3 gang switch with neutral wire

aqara 3 gang switch


Koenkk:
Thanks!

PR #1526, opened 2020-09-03
Add OWON WSP404 Smart Plug

embak:
Addition of the OWON WSP404 on/off wall plug switch control, with power and energy measurement reporting.


Koenkk:
Thanks!

PR #217, opened 2019-01-17
RS 125\u0000\ other name

arteck:

PR #1531, opened 2020-09-05
Scenes support

K-r-i-s-t-i-a-n:
Bare support for scenes: Added converters toZigbee and fromZigbee for handling scenes

Corresponding pull request for zigbee-herdsman: [#217](https://github.com/Koenkk/zigbee-herdsman/pull/217)

Known Issue: commands with responses generate a timeout error in the logs


K-r-i-s-t-i-a-n:
How should we add those toZigbee converters to the devices in devices.js? I think it would be pretty inefficient to add them to each device separately. They should be according to the specification and thus be supported by every device. 

My thought: add as default to every device, exclude if device is not compatible.

K-r-i-s-t-i-a-n:
Poking around a little today I came up with the plan to write an extension for z2m:
publish to zigbee2mqtt/bridge/scenes/foo/add|remove|recall where foo is either a group or device and payload is the input to the converters. This would not change each device definition and would expose a usable interface.

I am working on a very preliminary version, should have something to share tomorrow. Any feedback would be great, still very new to this project.

Koenkk:
I think it would be better to go the Zigbee2MQTT route, this should provide a more neat integration since we can also persist the scenes in which a devices is in. This can be done in the `meta` (example: https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/extension/report.js#L161).

The api should then consist of 2 parts:
- Scene management (add/remove) under: `zigbee2mqtt/bridge/request/scenes/#`
- Scene recall via `zigbee2mqtt/my_group/set` payload `{"scene_recall": 1}`

In that way all changes except `scene_recall` can be removed as the call can be directly done in `scenes.js`

Question: if I understand correctly now, scene commands are always send to groups?

K-r-i-s-t-i-a-n:
- Scenes are actually not only used with groups, but can also be used on single devices, in which case groupId must be 0. A device can store multiple scenes with identical sceneId but different groupId (again can be 0 if not part of a group). I belive this makes the interface a little more tricky.

- Scene workflow:
  - add scene definition (i.e. what attributes to store) to scene table of the devices (identified by groupId and sceneId)
  - set desired attributes (onOff, brightness, ...)
  - store scene, i.e. store current values to scene table
  - later recall scene with groupId and sceneId

- Please help me understand your proposed interface:
  - Creation of scene: "zigbee2mqtt/bridge/request/scenes/#/add|remove|view"
    - does the "#" address sceneId or groupId/deviceId?
      - if sceneId:
        - sceneId might not be unique, only combination with either groupId or deviceId is
        - scenes are not tied to any group or device yet, only "lives" inside zigbee2mqtt, hence need a mechanism to write (i.e. add and store) those scenes to the devices
      - if groupId or deviceId:
        - thats what I implemented in [PR#4300](https://github.com/Koenkk/zigbee2mqtt/pull/4300), also added the scene_store & scene_recall command here
        - easy to transfer all converters from toZigbee.js to scenes.js
      - I really do not see the big picture yet what the preferred version would be? Question: what is the benefit of persistent scenes inside zigbee2mqtt 
  - Recalling of scene on devices/groups:
    - topic "zigbee2mqtt/my_group/set" or "zigbee2mqtt/my_device/set", {"scene_recall": 1}
    - certainly the logical place to recall a scene. Is there a clean way to do this without having the scene_recall converter in every device? Not sure if I understand your meta idea right.

Koenkk:
Forget what I said about `zigbee2mqtt/bridge/request/scenes/#/add|remove|view`, I think it's indeed much easier to leave everything in `toZigbee.js` for now.

If I understand correctly, the only reason for https://github.com/Koenkk/zigbee2mqtt/pull/4300 is to avoid having to add the converters to all devices?

K-r-i-s-t-i-a-n:
Yes, my original intention for PR4300 was to not add the converters to each single device and being able to integrate all converters directly in scenes.js. I still do not see the clear use-case for persistent information about scenes, but we might get there, then this approach could be more future proof.

@Koenkk  and all others: Open minded, let me know what you prefer.

Koenkk:
I've made some changes:
- The scene converters will now be available for every device, this is done here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/1531/files#diff-6c9a6acf22f90d1c6e524d9f3c5c1745R12492 , this makes https://github.com/Koenkk/zigbee2mqtt/pull/4300 obsolete
- For now I've only enabled `scene_store` and `scene_recall`, lets start as simple as possible and extend if needed, I think this already provides most of the functionality
- I've wrote some docs: https://github.com/Koenkk/zigbee2mqtt.io/blob/scenes/docs/information/scenes.md
- When you recall a scene, Zigbee2MQTT will now update the state of that device/group accordingly and publish this to MQTT
- When you send a scene store/recall to a group, the `groupid` of the property will be taken from that group.

@K-r-i-s-t-i-a-n please test out these changes and provide feedback. To test the changes you need to be on the latest zigbee2mqtt dev branch first (as I made some changes in zigbee-herdsman). https://www.zigbee2mqtt.io/how_tos/how-to-switch-to-dev-branch.html

K-r-i-s-t-i-a-n:
I tested it and the basic functionality works. Thanks!

Here are some remarks:
- I enabled `scenes_add`, as it is absolutely required to uses scenes. With `scenes_add`, `scenes_store`, and `scenes_recall` we have the bare minimum to use scenes
- I do not think the logic of storing states for groups works well as not all members will have the same attributes. For example my main use-case: different scenes  have only a certain subset of lamps turned on, thus different devices have different onOff states. For this we would need to store the values per device, sceneId, and groupId
- Another note to storing of attributes: in `scene_add` we define what attributes will be stored in the scene_table (e.g. onOff, brightness, ....). The storing in `scene_store` is currently hard-coded. 
- The logic of pulling groupId from either the actual group or set it to 0 should be fine. Only found one counter-example: `scene_view` must use a different logic but is not enabled anyways right now.

Koenkk:
> I enabled scenes_add, as it is absolutely required to uses scenes. With scenes_add, scenes_store, and scenes_recall we have the bare minimum to use scenes

What is the added value of the scene add? In my case I can create scenes fine with just the store command.

> I do not think the logic of storing states for groups works well as not all members will have the same attributes. For example my main use-case: different scenes have only a certain subset of lamps turned on, thus different devices have different onOff states. For this we would need to store the values per device, sceneId, and groupId

Will look into this.

K-r-i-s-t-i-a-n:
Regarding `scenes_add`:
did not understand the specifications correctly. Tried to directly `scenes_store` and it does also create an entry in the scenes table + stores values. The benefit of using `scenes_add` is that you can control which clusters and attributes will be stored, whereas `scene_store` will store all (relevant) attributes.

Koenkk:
Did some more updates, please test (requires latest dev branch). Also fixed the issue with different states for devices in a group.

Updated docs: https://github.com/Koenkk/zigbee2mqtt.io/blob/scenes/docs/information/scenes.md

Please test, I think this is ready to be merged.

K-r-i-s-t-i-a-n:
I just tested the new version:
- `scenes_store` and `scenes_add` work as expected.
  - Observation: on `scenes_recall`the state of every device is reported individually (as expected). Afterwards the same states are also reported again with the group name. I can argue this is correct/expected behaviour. Thus not sure if this should be changed
- `scenes_add` mostly works. Only issue: Setting the color_temp for the group directly works, via `scenes_add` only for some of my bulbs. I will work on this later today, maybe just a problem with my old hue bulbs. (from my side not a blocker to merge this PR, I will try to fix this later)


@Koenkk : Thanks for your support here!

devices.js
```diff
@@ -12483,6 +12483,13 @@ const devices = [
...
-module.exports = devices.map((device) =>
-    device.extend ? Object.assign({}, device.extend, device) : device,
-);
+
+module.exports = devices.map((device) => {
+    if (device.extend) {
+        device = Object.assign({}, device.extend, device);
+    }
+
+    device.toZigbee.push(tz.scene_store, tz.scene_recall, tz.scene_add);
```
sjorge, line 12492: Do we need to add this to all devices? There are quiet a few sensors that have no toZigbee functions as they are read only so scenes for not make sense for them.

Koenkk:
@K-r-i-s-t-i-a-n for the groups this is indeed expected behaviour, lets merge this now and fix later if required as I think the base functionality works.

devices.js
Koenkk: You are right, will only be added now if a device has at least 1 toZigbee converter. This will still lead to some devices have the scenes converters where they don't support it however it won't cause problems (we do the same for groups).

PR #1535, opened 2020-09-06
Removed timeout from Immax Neo 4in1 sensor

Nithador:
After more testing found out, the device should not have timeout. Sorry about previous mistake.


Koenkk:
Thanks!

PR #1538, opened 2020-09-07
Add Sengled E1F-N5E

clcain:
Add Sengled E1F-N5E to `devices.js`. The Sengled E1F-N5E is similar to the Sengled E11-N1EA but is smaller and has an E12 screw base.


Koenkk:
Thanks!

PR #1539, opened 2020-09-08
Replaced parsing of temperature for QBKG03LM

MisterVSK:
Temperature data from QBKG03LM now uses generic_device_temperature converter, but it is incompatible. Temperature comes in format similar to QBKG12LM where it is parsed in xiaomi_power_from_basic converter.
So I had to replace generic_device_temperature with xiaomi_power_from_basic in devices.js to get temperature work.
 Also added conditions to xiaomi_power_from_basic_info to parse only existed data

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2540,20 +2540,26 @@ const converters = {
             }
         },
     },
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #1541, opened 2020-09-09
Support low battery information and display battery voltage for Aqara S2 lock

dzungpv:


converters/fromZigbee.js
```diff
@@ -4459,6 +4467,22 @@ const converters = {
             return result;
         },
     },
+    ZNMS12LM_low_battery: {
+        cluster: 'genPowerCfg',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const result = {};
+            if (typeof msg.data['batteryAlarmMask'] == 'number') {
```
Koenkk, line 4475: I'm surprised to see the `batteryAlarmMask` is used for reporting a low battery. According to the docs this enable configuration of when alarms are generated, not about the battery state itself:

image

I expect that the `batteryAlarmState` is used for this: https://github.com/Koenkk/zigbee-herdsman-converters/blob/a2d3c9e1ffaa2a1a61a988f01e5e9c906e5d1572/converters/fromZigbee.js#L495

Can you verify this?

dzungpv:
It is Xiaomi, they not follow the standard. For tue lock state  I have to blind decode it.
 I use it for nearly two years, now it say "low battery" with voice, and the data show in the log, only "batteryAlarmMask":1. Before i never see it. So i only rely on the real data, not zigbee standard.

converters/fromZigbee.js
dzungpv: There is no data "batteryAlarmState" when it say " low battery, please replace". So please see my other comment.

dzungpv: You can rename it from battery_low to battery_alarm, no problem at all, but from the log only:
```
Zigbee2MQTT:debug 2020-09-09 08:21:56: Received Zigbee message from '0x00158d000xxx', type 'attributeReport', cluster 'genPowerCfg', data '{"batteryAlarmMask":1}' from endpoint 1 with groupID 0
Zigbee2MQTT:debug 2020-09-09 08:21:56: No converter available for 'ZNMS12LM' with cluster 'genPowerCfg' and type 'attributeReport' and data '{"batteryAlarmMask":1}'
```
So we know only value 1, i don't replace the battery yet (for debug other value), and it still function normally

Koenkk: No problem, thanks for checking!

PR #1543, opened 2020-09-10
add support for TS0601 curtain module (Binthen CV01A curtain)

zhouF96:
add support for TS0601 curtain module 
CV01A

please also check the pic of TS0004  D3-DPWK-TY  and TS0502A   the pic of those modules are mixed, please fix them

I've sent an email about this , but haven't receive any reply.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,6 +2,7 @@ name: "Close stale issues/pull requests"
 on:
   schedule:
   - cron: "0 0 * * *"
+  workflow_dispatch:
 
 jobs:
   stale:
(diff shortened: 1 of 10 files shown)


Koenkk:
Since there are multiple devices using `TS0601` as model identifier ( https://github.com/Koenkk/zigbee-herdsman-converters/blob/bb422f0c834c4c38ec94d52d25e1bae1dae0a012/devices.js#L11306) please add it as a fingerprint instead like https://github.com/Koenkk/zigbee-herdsman-converters/blob/bb422f0c834c4c38ec94d52d25e1bae1dae0a012/devices.js#L1315 , you can find the `manufacturerName` in the `data/database.db`

Koenkk:
Thanks!

PR #1554, opened 2020-09-14
FIX#1553 color_mode not parsed when value is 0

sjorge:
We skipped parsing color data where the value was 0, switch to hasOwnProperty to avoid this.


Koenkk:
Thanks!

PR #1561, opened 2020-09-17
Added Kwikset 99100-045 SMARTCODE_DEADBOLT_5

zombielinux:
After about a month of testing with this configuration, the lock works perfectly.


Koenkk:
Thanks!

PR #1562, opened 2020-09-17
Added Innr FL 140 C (LED Strip) to devices

kairauer:
I've added support for the Innr FL 140 C LED strip. Works great so far. From my understanding this is basically the same light strip as the FL 130 C with 1200lm instead of 1000lm.


Koenkk:
Thanks!

PR #1564, opened 2020-09-18
Added set of HEIMAN devices

Hacker-CB:


devices.js
```diff
@@ -376,6 +388,35 @@ const livolo = {
     },
 };
 
+const heiman = {
```
Koenkk, line 391: Please inline this in the `configure()`, they are only used once so no need to put globally.

devices.js
```diff
@@ -56,6 +56,18 @@ const readMeteringPowerConverterAttributes = async (endpoint) => {
     await endpoint.read('seMetering', ['multiplier', 'divisor']);
 };
 
+const writeCurrentTime = async (endpoint) => {
```
Koenkk, line 59: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/1564/files#r491344969

converters/fromZigbee.js
```diff
@@ -2692,6 +2692,43 @@ const converters = {
...
+            const result = {};
+            if (msg.data['batteryState']) {
+                const lookup = {
+                    0: 'not_charging',
+                    1: 'charging',
+                    2: 'charged',
+                };
+                result['battery_state'] = lookup[msg.data['batteryState']];
+            }
+            if (msg.data['tvocMeasuredValue']) result['voc'] = msg.data['tvocMeasuredValue'];
```
Koenkk, line 2726: Should `result['voc']` be `result['tvoc']` ?

converters/fromZigbee.js
```diff
@@ -2692,6 +2692,43 @@ const converters = {
             }
         },
     },
+    heiman_pm5: {
```
Koenkk, line 2695: `heiman_pm5` -> `heiman_pm25`

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt/pull/4395

converters/fromZigbee.js
Hacker-CB: As i understood, `voc` and `tvoc` is ithe same mesaure.
It was `voc` sensor in `homeassistant.js` before my commits, so I reused it.

devices.js
Hacker-CB: I thing this can be useful for other devices. Heiman and may be other brands. It is standard cluster and standard writing to it. 
I'm sure that at least Heiman will release some new device and this will be needed for it.

converters/fromZigbee.js
Koenkk: good!

Koenkk:
Thanks!

PR #1570, opened 2020-09-20
adds support of Philips Hue LTD011

RossBille:
Adds support for [Philips Hue LTD011 ](https://www.philips-hue.com/en-au/p/hue-white-ambiance-garnea-downlight/5110131H5). Essentially just a copy of https://www.zigbee2mqtt.io/devices/5900131C5.html as their functionality is the same.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -2246,7 +2246,7 @@ const devices = [
     },
     {
         zigbeeModel: ['LTD011'],
-        model: '5110131H5',
+        model: 'LTD011',
         vendor: 'Philips',
         description: 'Garnea Downlight',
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -2244,6 +2244,15 @@ const devices = [
         extend: hue.light_onoff_brightness_colortemp,
         ota: ota.zigbeeOTA,
     },
+    {
+        zigbeeModel: ['LTD011'],
+        model: '5110131H5',
```
RossBille, line 2249:
```suggestion
        model: 'LTD011',
```

Koenkk:
Changed the model back to `5110131H5` because that is the one findable on the internet. Thanks!

PR #1571, opened 2020-09-20
Add Qmotion blinds.

willu:
This adds the device config for some QMotion blinds.

There is a lot of functionality that isn't covered.  e.g. I'd love to read the battery level. 
The blinds implement Zigbee clusters 0,1,3,4,5,6,8,32,256,258,2821,64512 but I'm really only using 258.

I don't have model numbers for the blinds.  I have two blinds, a long blind and a short blind.  They have
different zigbeeModels but I only saw one model number printed on the packaging.


willu:
This seems to be failing a test because I don't know the model numbers.  (well, I know *one* blind model number, but not which of the two devices it applies to (I think the short blind, but I'm not sure).  And I don't know the other model number, so that test would still fail.)  It works fine in practice without the model numbers defined.

Koenkk:
Since they carry the same model and the config is the same I've merged them.

Can you provide pictures of these devices? (for https://www.zigbee2mqtt.io/information/supported_devices.html)

willu:
Um.  There are two different devices, and I don't think they can be merged.

One device has a zigbeeModel of 'Honeycomb Internal Battery', and the other has a zigbeeModel of 'Rollershade Internal Battery'.

I didn't know which one the model number applies to, so I added one ? to one, and two ?? to the other.

willu:
What is the right fix here?  Revert the merge and re-merge, or a forward fix over the top?

willu:
Re images:   The roller blinds look just like roller blinds.  I'd just be showing my fabric choice.

The product page for the remote is here: http://www.qmotionshades.com/products/25-controls/271-zigbee-ha1-2-multi-channel-qdr2-remote  It has this jpeg: http://www.qmotionshades.com/system/product_images/images/000/000/582/preview/QdR2_White.jpeg

willu:
Ahh - I see what you've done with merging them.  Yeah, that should work.  I don't think they actually have the same model ID - but seeing as I don't know which is which, this is fine.  Sorry for the noise.

PR #1574, opened 2020-09-20
Update zigbeeModel for SONOFF SNZB-03

gurpz1:
I am unsure if the original zigbee model was a typo however the latest model shipped to me is 'MSO1' and not MS01'


Koenkk:
Thanks!

PR #1577, opened 2020-09-21
add configureReporting.onoff

nukusinji:
add configureReporting.onoff in All dawon dns devices.
I verified function in devices.


Koenkk:
Thanks!

PR #1580, opened 2020-09-21
add LoraTap Remote

### LINKED ISSUE #1579: Possibly duplicate modelID. Need advice
I have a TS0043 device [LoraTap remote](https://www.loratap.com/products/tuya-smart-life-zigbee-smart-home-wireless-switch-3-gangs-remote-tuya-zigbee-hub-required-no-limit-to-control-home-device) which has same functions as the existing TS0043 wireless switches but the actions are reported as left, middle and right which does not really apply in this case. 

image

I couldn't find the manufacturer for the Smart9 and Lonsonho wireless switch to compare. This device is `_TZ3000_bi6lpsew`

Would this warrant a new device with fingerprint to report different actions or is that stretching it too far?

blakadder:
Adding new device

Fixes #1579 

image


Koenkk:
Thanks!

PR #1590, opened 2020-09-25
Fixed Bitron 902010/24A alert activation

alexbrickwedde:
Did some research on the Bitron 902010/24A smoke detector.


Koenkk:
Thanks, please also update docs by clicking the pencil icon here: https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/devices/902010_24.md

PR #1593, opened 2020-09-26
Add support for AIRAM CTR.U BX

realPy:
Add the support missing remote CTR.U BX.
Additionnal add a hold parameters in CTR U functions to easy implement brightness down and up on Hold.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3417,6 +3417,7 @@ const converters = {
 
             return {
                 action: `brightness_${direction}_hold`,
+                hold: `${direction}`,
                 rate: msg.data.rate,
             };
         },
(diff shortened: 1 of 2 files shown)


Koenkk:
I see the `direction` is already present in the `action` property, why add it another time?

realPy:
When we want use the Hold brigthness, we track the brighness_Hold action.
When we have it we begin to send a payload with "on" and the brightness. The action is override with "on" , then "".
If we want to use the repeat until from homeassistant, we need an attribute that change only twice: One when its Hold, the second when it's release.

Koenkk:
The problem I'm having with this is that the `hold` is persisted, which means that on e.g. a battery report `hold` will also be published. 

Also can you switch to the generic converters? The currently used converters (except `command_on` and `command_off`) are considered legacy. For e.g. `CTR_U_brightness_updown_click` use `command_step`

Note to self after this merge https://github.com/Koenkk/zigbee2mqtt/pull/4467

realPy:
@Koenkk I use generic function and remove the hold parameter

Koenkk:
Thanks!

PR #1604, opened 2020-09-29
add Osram 4058075816596

simonses1:


Koenkk:
This device is already supported (https://github.com/Koenkk/zigbee-herdsman-converters/blob/42defcd7f84e4cbf03c59f9221605b1c7cfd929d/devices.js#L3775)

PR #1606, opened 2020-09-29
Add exposes for lights. https://github.com/Koenkk/zigbee2mqtt/issues/4466

Koenkk:


devices.js
```diff
@@ -26,6 +26,7 @@ const tz = require('./converters/toZigbee');
 const utils = require('./converters/utils');
 const globalStore = require('./converters/store');
 const ota = require('./ota');
+const exposes = require('./lib/exposes');
```
nurikk, line 29: Kindly export exposes from main package, so it could be reused in external converters

lib/exposes.js
```diff
@@ -0,0 +1,47 @@
+module.exports = {
```
nurikk, line 1: We can use something like builder pattern here, and with this it we don't have to make this endless combinations of possible features.
https://gist.github.com/nurikk/00b55d6c3dc38ad45878a6d21fbeac88

PR #1607, opened 2020-09-29
Added ETOP HT-08 Thermostat and HT-10 TRV

Hacker-CB:
References:
* https://github.com/Koenkk/zigbee2mqtt/issues/4185
* https://github.com/Koenkk/zigbee2mqtt/issues/1856
* https://github.com/Koenkk/zigbee2mqtt/issues/4486
* http://www.etopcontrols.com/thermostat/detail/HT-08.html

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -2543,6 +2543,94 @@ const converters = {
             sendTuyaCommand(entity, 528, 0, [4, 0, 0, ...payloadValue]);
         },
     },
+    etop_thermostat_system_mode: {
+        key: ['system_mode'],
+        convertSet: async (entity, key, value, meta) => {
+            switch (value) {
+            case 'off':
+                await sendTuyaCommand(entity, 257, 0, [1, 0/* off */]);
+                break;
+            case 'heat':
+                await sendTuyaCommand(entity, 257, 0, [1, 1/* on */]);
+                await utils.sleepMs(500);
+                await sendTuyaCommand(entity, 1028, 0, [1, 0/* manual */]);
+                break;
+            case 'auto':
+                await sendTuyaCommand(entity, 257, 0, [1, 1/* on */]);
+                await utils.sleepMs(500);
+                await sendTuyaCommand(entity, 1028, 0, [1, 2/* auto */]);
+                break;
+            }
+        },
+    },
+    etop_thermostat_away: {
+        key: ['away'],
+        convertSet: async (entity, key, value, meta) => {
+            switch (value) {
+            case 'ON':
+                await sendTuyaCommand(entity, 257, 0, [1, 1/* on */]);
+                await utils.sleepMs(500);
+                await sendTuyaCommand(entity, 1028, 0, [1, 1/* away */]);
+                break;
+            case 'OFF':
+                await sendTuyaCommand(entity, 1028, 0, [1, 0/* manual */]);
+                break;
+            }
+        },
+    },
+    tuya_thermostat_weekly_schedule: {
+        key: ['weekly_schedule'],
+        convertSet: async (entity, key, value, meta) => {
+            const maxTransitions = utils.getMetaValue(entity, meta.mapped, 'weeklyScheduleMaxTransitions');
+            const supportedModes = utils.getMetaValue(entity, meta.mapped, 'weeklyScheduleSupportedModes');
+            const firstDayDpId = utils.getMetaValue(entity, meta.mapped, 'weeklyScheduleFirstDayDpId');
+
+            function transitionToData(transition) {
+                // Later it is possible to move converter to meta or to other place outside if other type of converter
+                // will be needed for other device. Currently this converter is based on ETOP HT-08 thermostat.
+                // see also fromZigbee.tuya_thermostat_weekly_schedule()
+                const minutesSinceMidnight = transition.transitionTime;
+                const heatSetpoint = Math.floor(transition.heatSetpoint * 10);
+                return [
+                    (minutesSinceMidnight & 0xff00) >> 8,
+                    minutesSinceMidnight & 0xff,
+                    (heatSetpoint & 0xff00) >> 8,
+                    heatSetpoint & 0xff,
+                ];
+            }
+            for (const [, daySchedule] of Object.entries(value)) {
+                const dayofweek = parseInt(daySchedule.dayofweek);
+                const numoftrans = parseInt(daySchedule.numoftrans);
+                let transitions = [...daySchedule.transitions];
+                const mode = parseInt(daySchedule.mode);
+                if (!supportedModes.includes(mode)) {
+                    throw new Error(`Invalid mode: ${mode} for device ${meta.options.friendlyName}`);
+                }
+                if (numoftrans != transitions.length) {
+                    meta.logger.warn(`Invalid numoftrans provided. Real: ${transitions.length} ` +
+                        `provided ${numoftrans} for device ${meta.options.friendlyName}`);
+                }
+                if (transitions.length > maxTransitions) {
+                    meta.logger.warn(`Reducing number of transitions from ${transitions.length} ` +
+                        `to ${maxTransitions} for device ${meta.options.friendlyName}`);
+                    transitions = transitions.slice(4);
+                }
+                if (transitions.length < maxTransitions) {
+                    meta.logger.warn(`Padding transitions from ${transitions.length} ` +
+                        `to ${maxTransitions} with last item for device ${meta.options.friendlyName}`);
+                    const lastTransition = transitions[transitions.length-1];
+                    while (transitions.length != maxTransitions) {
+                        transitions = [...transitions, lastTransition];
+                    }
+                }
+                const payload = [];
+                transitions.forEach((transition) => {
+                    payload.push(...transitionToData(transition));
+                });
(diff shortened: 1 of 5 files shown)


converters/toZigbee.js
```diff
@@ -2543,6 +2543,94 @@ const converters = {
...
+            case 'auto':
+                await sendTuyaCommand(entity, 257, 0, [1, 1/* on */]);
+                await utils.sleepMs(500);
+                await sendTuyaCommand(entity, 1028, 0, [1, 2/* auto */]);
+                break;
+            }
+        },
+    },
+    etop_thermostat_away: {
+        key: ['away'],
```
Koenkk, line 2567: I think it's better to change this to `away_mode`, seems Home Assistant also uses this.

converters/toZigbee.js
```diff
@@ -2543,6 +2543,94 @@ const converters = {
...
+            for (const [, daySchedule] of Object.entries(value)) {
+                const dayofweek = parseInt(daySchedule.dayofweek);
+                const numoftrans = parseInt(daySchedule.numoftrans);
+                let transitions = [...daySchedule.transitions];
+                const mode = parseInt(daySchedule.mode);
+                if (!supportedModes.includes(mode)) {
+                    throw new Error(`Invalid mode: ${mode} for device ${meta.options.friendlyName}`);
+                }
+                if (numoftrans != transitions.length) {
+                    meta.logger.warn(`Invalid numoftrans provided. Real: ${transitions.length} ` +
```
Koenkk, line 2610: I think it would be better to throw an exception here instead of a warning

converters/toZigbee.js
```diff
@@ -2543,6 +2543,94 @@ const converters = {
...
+                        `provided ${numoftrans} for device ${meta.options.friendlyName}`);
+                }
+                if (transitions.length > maxTransitions) {
+                    meta.logger.warn(`Reducing number of transitions from ${transitions.length} ` +
```
Koenkk, line 2614: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/1607/files#r496827408

devices.js
```diff
@@ -1441,6 +1459,31 @@ const devices = [
...
+        toZigbee: [
+            tz.etop_thermostat_system_mode,
+            tz.etop_thermostat_away,
+            tz.tuya_thermostat_child_lock,
+            tz.tuya_thermostat_current_heating_setpoint,
+            tz.tuya_thermostat_weekly_schedule,
+        ],
+        onEvent: tuya.setTime,
+        meta: {
+            weeklyScheduleMaxTransitions: 4,
```
Koenkk, line 1482: Can you wrap this in `thermostat`?

`meta: {thermostat: {weeklyScheduleSupportedModes, ....}}`

Koenkk:
Note to self, after this merge https://github.com/Koenkk/zigbee2mqtt/pull/4491/files

converters/toZigbee.js
Hacker-CB: What is behavior in standard cluster? 
Number of transistions is fixed and users should fill all of them, even they need few one?
For example. 
There is 4 transitions supported by the controller.
Is it possible to set only 2? In my case missing transitions will be filled by data from the last one.

Koenkk: I expect that not all have to be filled, are you able to test this? (I don't have a Zigbee thermostat)

Hacker-CB: I dont have Zigbee thermostat with pure clusters :(
In any case, I think it is possible to add similar behaviour `toZigbee.thermostat_weekly_schedule()` to prevent user-side errors.
How do you think?

Hacker-CB: One problem, existing zigbee thermostats does not have `meta.thermostat.weeklyScheduleMaxTransitions`, but may be it is possible to take it from state...
Or may be be it's better to be more strict and throw error in such cases?

Hacker-CB: I found only two of the supported thermostats which are able to buy.

- https://www.amazon.com/Centralite-3-Pearl-Touch-Thermostat/dp/B01G99KR94
- https://www.amazon.com/Zen-Within-Zen-01-W-Thermostat-Edition/dp/B00XCVV8CS

Koenkk: - But for this thermostat, does it work when setting just 2 transitions? If yes, I think we can leave this out
- I don't see the added value for `meta.thermostat` yet https://github.com/Koenkk/zigbee-herdsman-converters/pull/1607/files#diff-6c9a6acf22f90d1c6e524d9f3c5c1745R1484 , this converter is not used by other devices and it's unclear when it can be if they would have different meta values. I think they can be embedded in the converter for now.

Hacker-CB: * For this thermostat currently it will works with any number of transitions. If less then 4 transitions - it will pad missing with value with last one. If more than 4 - it will delete extra transitions.
* For other devices - I currently working on [HT-10](http://www.etopcontrols.com/thermostat/detail/HT-10.html), manufacturer told that it has same API, so this is reason I moved it to top level to the common converters. Also seems that `fromZigbee.tuyaThermostat` dp `112` and `113` (see image below) which are responsible for the schedule should be refactored to keep same API for the schedule. Problem is that they has only weekday/weekend (instead of full week) schedule. Also there is only one-way schedule in exisiting thermostats (fromZigbee). I will fix this if I will find any generic TuYa thermostat.

image

Koenkk: But what happens if you just 2 transitions? 

>  If more than 4 - it will delete extra transitions.

I would throw an exception in this case

Hacker-CB: This thermostat requres always 4 transitions.
So, I'm adding missing transition with same time and temperature:

**input:** 
```json
{
  "dayofweek": 1,
  "mode": 1,
  "numoftrans": 4,
  "transitions": [
    {
      "heatSetpoint": 11.5,
      "transitionTime": 0
    },
    {
      "heatSetpoint": 9,
      "transitionTime": 420
    },
  ]
}
```

**output:**
```json
{
  "dayofweek": 1,
  "mode": 1,
  "numoftrans": 4,
  "transitions": [
    {
      "heatSetpoint": 11.5,
      "transitionTime": 0
    },
    {
      "heatSetpoint": 9,
      "transitionTime": 420
    },
    {
      "heatSetpoint": 9,
      "transitionTime": 420
    },
    {
      "heatSetpoint": 9,
      "transitionTime": 420
    },
  ]
}
```

Hacker-CB: If you still prefer to throw exception in both cases - I can do this.

Koenkk: Only an exception when providng more than 4 transitions is enough. 

I have no more comments, can I mere this?

Hacker-CB:
I also added HT-10 TRV

Koenkk:
Thanks!

drzony:
@Hacker-CB I'm trying to do Tuya devices rework, could you verify data for the weekly schedule, according to what I got from the manufacturer it's:
```typescript
[prog_mode] + 4 * [2 bytes time, 2 bytes temp]
Where prog mode = (1 - one for all days, 2 - weekdays/weekend, 3 - weekdays/saturday/sunday, 4 - 7 day schedule)
```
Your version is without prog_mode

Hacker-CB:
@drzony ETOP implementation has 7 days schedule mode only, it does not have prog_mode.

drzony:
@Hacker-CB My thermostat has 7 day mode only too, but the data has the field for programming mode, my question is whether the schedule data starts at offset 0 or 1

Hacker-CB:
@drzony I don't know about your thermostat. 
ETOP has proprietary schedule. 
Try to use sniffer to know how you thermostat works with original TuYa.

drzony:
@Hacker-CB 
I know how the schedule is represented in my thermostat, I'm wondering whether ETOP has data at offset 0 or at offset 1.
In case of ETOP the data should be [4, time1.1, time1.2, temp1.1, temp1.2, time2.1, time2.2, temp2.1, temp2.2, time3.1, time3.2, temp3.1, temp3.2, time4.1, time4.2, temp4.1, temp4.2]
The format seems to be common to all Tuya compatible thermostats, that is why I'm asking. I want to change the converter to include this offset

Hacker-CB:
I added `tuya_thermostat_weekly_schedule`: https://github.com/Koenkk/zigbee-herdsman-converters/blob/f244ab7e281612b26bd6bb0c9395cb9ba3cbc094/converters/fromZigbee.js#L5698

Here is data format (2 bytes time and 2 bytes temperature):
https://github.com/Koenkk/zigbee-herdsman-converters/blob/f244ab7e281612b26bd6bb0c9395cb9ba3cbc094/converters/fromZigbee.js#L5715

drzony:
Yes I'm looking at the code currently. If you are sure that there is no program_mode byte, then I will add offset to meta, to have common implementation.

Hacker-CB:
> Yes I'm looking at the code currently. If you are sure that there is no program_mode byte, then I will add offset to meta, to have common implementation.

In case of ETOP there is no `program_mode`. Time starts at offset 0.
So, you can try to add meta to handle case with offset.

drzony:
Okay, thanks for the info

PR #1609, opened 2020-09-29
Added support for Sunricher ZG2835

martinroehder:
Added support for Sunricher ZG2835.
It is just the dimmer knob to use as an input device. I bought it as ROBB ROB_200-018-0.


Koenkk:
Thanks!

PR #1613, opened 2020-09-30
Add option to publish periodic brightness updates during rotation

MarBra:
I like to add an option to send move events until stop event comes in.

This will address the following feature request: 

Resolves https://github.com/Koenkk/zigbee2mqtt/issues/4372

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -11,6 +11,7 @@
 
 const common = require('./common');
 const utils = require('./utils');
+const globalStore = require('./store');
 
 const occupancyTimeout = 90; // In seconds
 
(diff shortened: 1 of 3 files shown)


Koenkk:
What is your use case for this?

MarBra:
> What is your use case for this?

I have basically the same use case mentioned in this issue https://github.com/Koenkk/zigbee2mqtt/issues/4372

I like to use the Symfonisk controller as sound controller and like to have updates of the rotation without using in external tool like appdaemon. Currently I'm using the legacy integration with debounce to get this done, but I like to move away from the legacy implementation.

Do you think it is feasible to get this merge in z2m? I have added the extra option to make sure that the behavior only changes, if someone likes to have periodic updates.

Thanks :)

Koenkk:
Since it's not enabled by default I'm fine with adding such. But instead of publishing `repeat_brightness_move_event` I have more preference towards publishing a simulated `brightness` value (between 0-255). 

I would also like to have @fliphess  and @ale1800 opinions to make sure this also covers their use cases.

fliphess:
@Koenkk Why didn't I come up with that 😄 

Instead of a repeating event, a value between 0-255 makes much more sense :)

fliphess:
Btw, shall I close my request in favour of this one? Or should I adjust the request so it matches your suggestion?

Koenkk:
You can leave it open, https://github.com/Koenkk/zigbee2mqtt/issues/4372 is the feature request, this is the implementation of it.

MarBra:
> Since it's not enabled by default I'm fine with adding such. But instead of publishing `repeat_brightness_move_event` I have more preference towards publishing a simulated `brightness` value (between 0-255).
> 
> I would also like to have @fliphess and @ale1800 opinions to make sure this also covers their use cases.

I think I don't fully get the point 😄 
Is the idea to publish a mqtt message with simulated `brightness` in a periodic manner?
Or only adjust the `brightness` value in the stop message?

Koenkk:
> Is the idea to publish a mqtt message with simulated brightness in a periodic manner?

Yes, the value will be updated periodically, the period duration can be specified in `configuration.yaml`.

MarBra:
> > Is the idea to publish a mqtt message with simulated brightness in a periodic manner?
> 
> Yes, the value will be updated periodically, the period duration can be specified in `configuration.yaml`.

Sounds good :) I will try to implement it.

converters/fromZigbee.js
```diff
@@ -1120,10 +1120,36 @@ const converters = {
...
+
             const direction = msg.data.movemode === 1 ? 'down' : 'up';
             const action = postfixWithEndpointName(`brightness_move_${direction}`, msg, model);
             const payload = {action, action_rate: msg.data.rate};
             addActionGroup(payload, msg, model);
+
+            if (options.hasOwnProperty('periodic_brightness_updates') &&
+            options.periodic_brightness_updates) {
+                let brightness = direction === 'up' ? 0 : 255;
+                const brightnessDelta = direction === 'up' ? 0.5 : -0.5;
```
fliphess, line 1135: It might be nice to make this delta variable from the configuration file, that way the user can decide in which speed the brightness is adjusted when moving the controller :)

MarBra: Good idea :)
I have added an option for the delta variable. So now there are two ways to adjust the speed (update interval and delta value).

Koenkk:
@MarBra is this ready for merge?

MarBra:
> @MarBra is this ready for merge?

For me yes :)

Koenkk:
@MarBra thanks, I've tested and had an issue were brightness was reset to 255/0 every time. I've fixed this and did some other refactoring. Can you check if everything still ok. 

Options now have to be specified like this:

```yaml
devices:
  '0x12345678'
    friendly_name: 'my_dimmer'
    simulated_brightness: true 
    # OR
    simulated_brightness: 
      delta: 20 # = default
      interval: 200 # = default
```

MarBra:
> @MarBra thanks, I've tested and had an issue were brightness was reset to 255/0 every time. I've fixed this and did some other refactoring. Can you check if everything still ok.
> 
> Option now have to be specified like this:
> 
> ```yaml
> devices:
>   '0x12345678'
>     friendly_name: 'my_dimmer'
>     simulated_brightness: true 
>     # OR
>     simulated_brightness: 
>       delta: 20 # = default
>       interval: 200 # = default
> ```

Works great. Thanks a lot :)

fliphess:
@MarBra @Koenkk You are amazing! :) Thanks so much for this effort :)

Koenkk:
Thanks all, also wrote some docs: https://github.com/Koenkk/zigbee2mqtt.io/blob/develop/docs/devices/ICTC-G-1.md#device-type-specific-configuration

PR #1614, opened 2020-09-30
Add DIYRuZ_Flower support

nurikk:
Can be merged after https://github.com/Koenkk/zigbee-herdsman/pull/231

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -569,7 +569,8 @@ const converters = {
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
             const temperature = parseFloat(msg.data['measuredValue']) / 100.0;
-            return {temperature: calibrateAndPrecisionRoundOptions(temperature, options, 'temperature')};
+            const property = postfixWithEndpointName('temperature', msg, model);
+            return {[property]: calibrateAndPrecisionRoundOptions(temperature, options, 'temperature')};
         },
     },
     device_temperature: {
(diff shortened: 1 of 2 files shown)


converters/fromZigbee.js
```diff
@@ -610,7 +619,12 @@ const converters = {
         cluster: 'msPressureMeasurement',
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
-            const pressure = parseFloat(msg.data['measuredValue']);
+            let pressure = 0;
+            if (msg.data.hasOwnProperty('scaledValue') && msg.data.hasOwnProperty('scale')) {
```
nurikk, line 623: Screenshot 2020-10-01 at 13 17 58

Koenkk, line 623: This will only work if the `scaledValue` and `scale` value is always send with the `msPressureMeasurement`. I would expect we need to have something similar to: https://github.com/Koenkk/zigbee-herdsman-converters/blob/f346bd72f1689348e7bc9e2e51247672d3c5addb/converters/fromZigbee.js#L713

devices.js
```diff
@@ -133,6 +133,11 @@ const configureReporting = {
         const payload = configureReportingPayload('measuredValue', 10, repInterval.HOUR, 5, overrides);
         await endpoint.configureReporting('msPressureMeasurement', payload);
     },
+    pressureExtended: async (endpoint, overrides) => {
+        const scaledValuePayload = configureReportingPayload('scaledValue', 10, repInterval.HOUR, 5, overrides);
+        const scalePayload = configureReportingPayload('scale', 10, repInterval.HOUR, 5, overrides);
```
Koenkk, line 138: Instead of configuring reporting for it, can we also read it once instead? (also depends on https://github.com/Koenkk/zigbee-herdsman-converters/pull/1614/files#r498788313)

nurikk:
@Koenkk  fixed

converters/fromZigbee.js
```diff
@@ -610,7 +619,13 @@ const converters = {
         cluster: 'msPressureMeasurement',
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
-            const pressure = parseFloat(msg.data['measuredValue']);
+            let pressure = 0;
+            if (msg.data.hasOwnProperty('scaledValue')) {
```
Koenkk, line 623: I see you added it for `scale`, but why not do the same for `scaledValue`?

nurikk: In my understanding this is like measured value, and can be sent without any other attributes

Koenkk: Ah sorry, it's an alternative to `measureValue` indeed.

PR #1620, opened 2020-10-02
Expose color_hs for Innr RB 2xx C

sjorge:
As mentioned in koenkk/zigbee2mqtt#4466

- expose Hue/Saturation for Innr RB 2xx C bulbs
- enable Hue/Saturation for Innr RB 285 C bulbs

I suspect all inner color bulbs might support Hue/Saturation with the enhancedHue = false meta. But I only own the E27 and E14 ones to test with.


Koenkk:
Thanks!

PR #1626, opened 2020-10-04
Adds EBL data validation

PedroLamas:
Following up on https://github.com/Koenkk/zigbee2mqtt/issues/4361, I've added a simple validation of Silabs EBL (and GBL) data by calculating and verifying the CRC32 checksum of the data.

This working fine with a pure javascript CRC32 provider (buffer-crc32), but it would be great if we can replace it with some alternative hardware accelerated version.

I couldn't find a valid update file with GBL format, but after googling for a while I found one that at least allowed me to confirm that this is working as expected.

The whole thing is based on the following:
- https://github.com/zigpy/zigpy/blob/master/zigpy/ota/validators.py
- https://github.com/SiliconLabs/sdk_support/blob/master/platform/base/hal/micro/cortexm3/common/ebl.h
- https://docs.silabs.com/mcu-bootloader/latest/group-EblParserFormat


Koenkk:
Thanks!

PR #1627, opened 2020-10-04
Added Hive SLR2 and SLT3 to devices.js

aliijm:
Added Hive SLR2 and SLT3 to devices.js (see [zigbee2mqtt#4540](https://github.com/Koenkk/zigbee2mqtt/issues/4540)).

The converters for the thermostat clusters don't seem to currently support multiple endpoints. This support would need to be added to allow the the SLR2 to function correctly, but I don't know how to do this.


Koenkk:
I've made some updates to fromZigbee.js to support multiEndpoint, can you check if it works OK now?

converters/toZigbee.js
```diff
@@ -968,11 +968,17 @@ const converters = {
     thermostat_occupied_heating_setpoint: {
         key: ['occupied_heating_setpoint'],
         convertSet: async (entity, key, value, meta) => {
-            const occupiedHeatingSetpoint = (Math.round((value * 2).toFixed(1)) / 2).toFixed(1) * 100;
```
Koenkk, line 971: Changes to toZigbee shouldn't be needed, the correct endpoint (entity) will be passed when using `zigbee2mqtt/[THERMOSTAT]/heat/set`/`zigbee2mqtt/[THERMOSTAT]/cool/set` as topic

aliijm: The reports from the thermostat are working, thanks. When I try to publish a command using `zigbee2mqtt/[THERMOSTAT]/heat/set` I get `Entity '[THERMOSTAT]/heat' is unknown`

Koenkk: Ah sorry, the endpoint names should be one of: https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/util/utils.js#L31

aliijm: When I add the endpoint names to that file it all works perfectly (without my changes to toZigbee). Should I submit a pr to zigbee2mqtt to add the heat and water endpoints? Thanks so much for your help!

Koenkk: No, I will add it.

Koenkk:
Thanks!

PR #1629, opened 2020-10-05
Added Aurora SmartPlug51AU

nrandell:
Support for Aurora SmartPlug51AU

Slightly strange device I think, reports more clusters than it seems to support. No idea when it has 2 endpoints. But this appears to work.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -12946,15 +12946,11 @@ const devices = [
         vendor: 'Aurora Lighting',
         description: 'Aurora Smart plug',
         supports: 'on/off, power measurement, temperature',
-        fromZigbee: [fz.identify, fz.on_off, fz.electrical_measurement_power, fz.metering_power, fz.device_temperature, fz.occupancy],
+        fromZigbee: [fz.identify, fz.on_off, fz.electrical_measurement_power, fz.metering_power, fz.device_temperature],
         toZigbee: [tz.on_off],
-        meta: {configureKey: 6, multiEndpoint: true},
+        meta: {configureKey: 1},
         endpoint: (device) => {
-            return {
-                default: 2,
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -12940,6 +12940,40 @@ const devices = [
...
+        model: 'SmartPlug51AU',
+        vendor: 'Aurora Lighting',
+        description: 'Aurora Smart plug',
+        supports: 'on/off, power measurement, temperature',
+        fromZigbee: [fz.identify, fz.on_off, fz.electrical_measurement_power, fz.metering_power, fz.device_temperature, fz.occupancy],
+        toZigbee: [tz.on_off],
+        meta: {configureKey: 6, multiEndpoint: true},
+        endpoint: (device) => {
+            return {
+                default: 2,
```
Koenkk, line 12954: - Can you provide a link to this device?
- Does this device indeed support multiple endpoints? Because in your HA discovery you didn't support this (https://github.com/Koenkk/zigbee2mqtt/pull/4562/files#diff-0c056233b68fd3653afb296f98947c8fR1843). This only make sense to add when there are e.g. multiple gangs to control

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt/pull/4562

nrandell:
Good spot on the multiple endpoints. I've had this device for a while and only just figured out the multiple endpoints were for writing! (I think). I've updated the pull request with a newer version that I've used to reconfigure the 4 devices I have.

As for the link to the product. https://www.develcoproducts.com/products/smart-plugs/smart-plug-mini-type-g-british

They were supplied by Aurora, but appear to be develco products

Koenkk:
Thanks!

PR #1634, opened 2020-10-07
Emulate release event for Aqara Opple

kirovilya:
Aqara Opple does not generate a release event when pressed for more than 5 seconds.
After 5 seconds of not releasing we assume release.


Koenkk:
Thanks!

PR #1635, opened 2020-10-08
Fix for HEIMAN Smart socket HS2SK (NXP version) haElectricalMeasurement.

Hacker-CB:
Device does not support reporting. Need to poll.


Koenkk:
Thanks!

PR #1636, opened 2020-10-08
Added Ecosmart D1533 A19 Canadian model

lokiberra:


Koenkk:
I see a lot of unexpected changes here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/1636/files, can you make a clean pr with only changes to `devices.js`?

PR #1637, opened 2020-10-08
Expose ubisys s1 as switch + numeric for power

sjorge:
Not 100% sure on the unit, should it be W or watt ?

I think this can also be done for the s2 but I don't have one to test and it has multiple endpoints for the switch part.


sjorge:
Update, also done so for s2, found some examples of multi-endpoint switches.

Koenkk:
Thanks!

PR #1638, opened 2020-10-08
Expose temp/humidity for HEIMAN HS1HT and SNZB-02

sjorge:


sjorge:
The original issue about implementing exposes also mentions specifying a device type... but it's not clear how to set those. 
Filled in data for some more devices I own.

Koenkk:
> The original issue about implementing exposes also mentions specifying a device type... but it's not clear how to set those.
Filled in data for some more devices I own.

I still need to think how to do this, first goal is to remove all manual mapping in `homeassistant.js`.

> Also is there a way to global expose properties? e.g. linkquality?

Done (for linkquality)

sjorge:
> I still need to think how to do this, first goal is to remove all manual mapping in homeassistant.js.

OK so I didn’t miss something, one step at a time :)

PR #1639, opened 2020-10-09
Added Ecosmart D1533 A19 Canadian Model

lokiberra:
Hopefully this is better with just the Devices.js file changed


lokiberra:
When I run 'npm test' it fails because of the duplicate device model. This is the model listed on the bulbs is D1533. The other D1533 that was already in there was PAR20 and these are A19 bulbs. What is the suggested way to merge these in to Devices.js?

lokiberra:
Just added A19 to the model name

Koenkk:
Yes please merge with https://github.com/Koenkk/zigbee-herdsman-converters/blob/b54fa83af513b66700eeb6eae3f840d4efe6ed3b/devices.js#L8742 and change description to PAR20/A19

lokiberra:
Updated

Koenkk:
Thanks!

PR #1642, opened 2020-10-09
TS0043 different states for same device

arteck:
we need a standard state for ts0043 with or without a other manufacturerName

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3835,9 +3835,7 @@ const converters = {
             const clickMapping = {0: 'single', 1: 'double', 2: 'hold'};
             let buttonMapping = null;
             if (model.model === 'TS0042') {
-                buttonMapping = {1: 'left', 2: 'right'};
-            } else if (model.model === 'TS0043' && msg.device.manufacturerName === '_TZ3000_bi6lpsew') {
-                buttonMapping = {1: '1', 2: '2', 3: '3'};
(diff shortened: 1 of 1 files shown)


Koenkk:
Why is the current one not good? We should prevent to change this as it will lead to a breaking change.

arteck:
we have a problem in iobroker with..

Koenkk:
Can you describe more of this problem?

kirovilya:
TS0043 model for devices with three buttons.
but with this model several manufacturers make devices.
and it is surprising that in one case the {action: 1_click} event arrives for a device with the TS0043 model, and if you take the same device from another manufacturer - you get {action: right_click}.
I believe that here it is necessary either to unify events for one device model, or to divide them into different device models.
another problem concerns iobroker.zigbee - we describe events for a device with separate entities, so we need to know what value will come to action for a specific device model. in the case of TS0043, we have a problem - the values ​​may be different.

I believe that it is better to bring the action value to one type, and not to divide by manufacturer.

Koenkk:
Thanks I understand your point now. In that case I propose to get rid of the whole if/else button mapping and always set `buttonMapping` to `const buttonMapping = {1: '1', 2: '2', 3: '3', 4: '4'};`

kirovilya:
I agree, we can remove the comparison with ManufacturerName

arteck:
i make a new one

PR #1645, opened 2020-10-10
Adds proper test OTA image with Silabs GBL data

PedroLamas:
As I found a few IKEA images with GBL data, I think it best to add at least one to the test images.


Koenkk:
Cool, thanks again! 🎖️

PR #1648, opened 2020-10-11
Added Immax 07703L

lukasnebohy:
Created from ETOP HT-10 (it is same HW and funcionality, only differrent vendor)


Koenkk:
Thanks!

PR #1656, opened 2020-10-13
Add support for Consciot tunable white lamp

Tediore:
Add support for this light (rebranded Linkind): https://www.amazon.com/gp/product/B07V1GJZYD

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -13055,7 +13055,7 @@ const devices = [
 
     // Linkind
     {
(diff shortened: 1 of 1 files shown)


tomdee:
@Tediore Thanks for fixing this and pointing me at it. I will be able to add this kind of support myself in future :+1:

Tediore:
No problem, this was an easy one since it's the same hardware as an existing supported product 🙂

PR #1659, opened 2020-10-13
ubisys D1: Allow ballast configuration

felixstorm:
Primarily allows to control dimmer min and max levels (and could also be used for other devices if they implement the ZCL ballast configuration cluster).

Documentation will follow as well.

Thanks,  
Felix

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -1167,7 +1167,9 @@ const converters = {
     },
     ballast_config: {
         key: ['ballast_config'],
+        // zcl attribute names are camel case, but we want to use snake case in the outside communication
         convertSet: async (entity, key, value, meta) => {
+            value = utils.toCamelCase(value);
             for (const [attrName, attrValue] of Object.entries(value)) {
                 const attributes = {};
                 attributes[attrName] = attrValue;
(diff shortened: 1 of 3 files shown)


converters/toZigbee.js
```diff
@@ -1163,6 +1163,45 @@ const converters = {
             entity.commandResponse('ssIasAce', 'panelStatusChanged', payload);
         },
     },
+    ballast_config: {
+        key: ['ballast_config'],
+        convertSet: async (entity, key, value, meta) => {
+            for (const [attrName, attrValue] of Object.entries(value)) {
+                const attributes = {};
+                attributes[attrName] = attrValue;
+                await entity.write('lightingBallastCfg', attributes);
```
Koenkk, line 1172: Can you also convert this to snake_case? You can copy the following function to do the conversion: https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/util/utils.js#L157

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/421

felixstorm:
@Koenkk Thanks for the hint to the conversion function - is it ok now like this?

I also needed to convert JSON attribute names from snake case to camel case, so I also added a function `toCamelCase`. And since I then anyway had this function, I also changed the attribute names in code to become snake case as well - just in case somebody looks at the code hoping to get the attribute names to put into the JSON...

Koenkk:
Thanks!

PR #1664, opened 2020-10-14
fix support for adurosmart light bulbs

jronnen7:
in the latest version of the adurosmart firmware for bulbs 81809, firmware version `1000-0007`, firmware date `20181008` they changed the zigbee model to be `ZLL-ExtendedColor` opposed to `ZLL-ExtendedColo`


Koenkk:
Thanks!

PR #1666, opened 2020-10-15
feat: Add support for Nue/3A Door sensor

andykelk:
The Nue / 3A door and window sensor is relatively simple and just exposes a binary contact sensor.


Koenkk:
Thanks!

PR #1670, opened 2020-10-16
Added new zigbeeModel for Lupus 2 channel relay

ochstobi:
I have purchased a Lupus 2 channel relay because it is in the supported devices list. After pairing I found out, that the zigbeeModel was increased.
`Device '0x00124b001ebeaf35' with Zigbee model 'PRS3CH2_00.00.05.11TC' is NOT supported`. So I added the new zigbeeModel number to the array.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -8097,7 +8097,7 @@ const devices = [
         },
     },
     {
-        zigbeeModel: ['PRS3CH2_00.00.05.10TC'],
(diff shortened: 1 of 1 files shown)


ochstobi:
One Question, if this would be merged, do i need to re-pair the device or just restart zigbee2mqtt?

Koenkk:
Nope you don't need to repair, thanks!

PR #1671, opened 2020-10-16
move 'aabybja' and '_TZE200_nkoabg8w' to correct their behavior

zhouF96:
move 'aabybja' and '_TZE200_nkoabg8w' to correct their behavior


Koenkk:
Thanks!

PR #1680, opened 2020-10-18
Add Hornbach Flair viyu Smart LED E14 candle

EmbeddedSwGuy:
add support for Hornbach Flair viyu Smart LED E14 candle


Koenkk:
Thanks!

PR #1682, opened 2020-10-19
Add support for the aluminum Philips Hue Argenta series.

jap:
Model numbers were copied from beaumotica.nl, and for the 3-spot model is a confirmed match with my zigbee traffic.
The only difference between these models is the paint on the frame and a couple of bits in the firmware of the bulbs.


Koenkk:
Thanks!

PR #1683, opened 2020-10-19
Add support for iCasa ICZB-R12D

ratenfanger:
Config is copied from the ICZB-R11D because these models are (except wiring) the same.
Tested and working :nerd_face: 

Manufacturer device info: https://en.icasa.io/copy-of-dimmers-en-switches


Koenkk:
Thanks!

PR #1686, opened 2020-10-20
Update BE468 to have both BE468/GBAK working

orditeck:
Hi,

As mentioned in the following issue: https://github.com/Koenkk/zigbee2mqtt/issues/4464
The model BE468GBAK is not the exact same as the BE468 model.
BE468GBAK uses endpoint index 1, BE468 uses endpoint index 2

Replacing `device.getEndpoint(specificNumber)` by `device.endpoints[0]` make sure both models work.


Koenkk:
Thanks!

PR #1689, opened 2020-10-20
Adding support to devices.js for Xfinity XHS2-UE

kbowman-newedenconsultinggroup:
Almost identical to the XHS2-SE from Sercom.  Both are sold as Xfinity devices - this pull request is for a version made by Visonic.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -10915,10 +10915,11 @@ const devices = [
         },
         exposes: [e.contact(), e.battery_low(), e.tamper(), e.temperature(), e.battery()],
     },
+    // Comcast Xfinity branded XHS2-UE Door / Window / Temp sensor by Universal Electronics, Inc
     {
         zigbeeModel: ['URC4460BC0-X-R'],
         model: 'XHS2-UE',
-        vendor: 'Visonic',
+        vendor: 'Universal Electronics, Inc.',
         description: 'Magnetic door & window contact sensor',
         supports: 'contact, temperature',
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -10915,6 +10915,23 @@ const devices = [
         },
         exposes: [e.contact(), e.battery_low(), e.tamper(), e.temperature(), e.battery()],
     },
+    {
+        zigbeeModel: ['URC4460BC0-X-R'],
+        model: 'XHS2-UE',
+        vendor: 'Visonic',
```
Koenkk, line 10921: Googling shows me the sensor is made by Xfinity? (https://www.xfinity.com/support/articles/installing-battery-xhs2-ue-sensor)

kbowman-newedenconsultinggroup:
Xfinity is a brand name of Comcast, which is a telco in the US.  They don't make (to my knowledge) anything.  There are several different manufactures that provide them with their home-security systems, among them Sercom and Visonic and at least one other.

That said, looking closer I had the wrong manufacturer.  This particular device is not made by Visonic, but rather by Universal Electronics, Inc.  Here's the zigbee alliance page:  https://zigbeealliance.org/zigbee_products/comcast-xfinity-home-contact-sensor/

Of course, now the tests are failing with an error in index.test.js missing key 'supports' and I have no idea why.  Is the vendor referenced elsewhere and my 

`        vendor: 'Universal Electronics Inc',`

throwing the error?

Koenkk:
Looks good now, thanks!

PR #1690, opened 2020-10-20
Add ZB-CL03 to YSR-MINI-01

itavero:
Received a new "Smart Light Strip Controller".
The model listed on it is YSR-MINI-01, which was already a known device, but apparently this one uses a different `modelId`.

Note that on the particular controller I have there is a label that indicates what kind of LED strips are supported, mine is only labelled (and wired) for RGB.
I checked internally and it looks like all the other electronics (for CCT) are there as well, but there just aren't any wires.

To be honest, I'm not 100% sure if this `modelId` is maybe strictly for the RGB only model, or if the behavior is identical to the other YSR-MINI-01 that are already known to zigbee-herdsman.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -6715,7 +6715,7 @@ const devices = [
 
     // YSRSAI
     {
-        zigbeeModel: ['ZB-CL01', 'FB56-ZCW20FB1.2'],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1696, opened 2020-10-22
Add Q-LED E27 socket light

konqi:
adding support for Paul Neumann light, p/n 100.001.96. Apparently these are not for sale anymore but I still have a couple of them about, so I thought I might as well...


Koenkk:
Thanks!

PR #1704, opened 2020-10-25
Fixes GBL validation for padded data

pedrolamas:
Following up on [this specific issue](https://github.com/Koenkk/zigbee2mqtt/issues/2921#issuecomment-715270071), I came to realize that I introduced a bug on the GBL data validation that shows for images with GBL data padded (which apparently is a thing!)

This fixes the problem by allowing the GBL data to be padded with 0x0 bytes, while still validating the CRC-32 of the whole block.

I've added the specific image that was failing to the tests and it is now working as expected


Koenkk:
Thanks!

Issue #1321, opened 2020-06-16
MCT-340 SMA error


j4c3:
I have paired a Visonic MCT-340 SMA, and it appears to function:
```
zigbee2mqtt:info 2020-06-16 10:57:31: MQTT publish: topic 'zigbee2mqtt/MCT340_1', payload '{"contact":true,"tamper":false,"battery_low":false,"linkquality":42,"battery":27,"voltage":2.8,"te
mperature":29,"last_seen":"2020-06-16T10:57:31-07:00"}'   
```
But reports an error in the logs on each reboot of zigbee2mqtt: 
```
zigbee2mqtt:error 2020-06-16 10:32:24: Failed to configure 'MCT340_1', attempt 1 (Error: ConfigureReporting 0x000d6f000afe1a8c/1 gen
PowerCfg([{"attribute":"batteryPercentageRemaining","minimumReportInterval":3600,"maximumReportInterval":62000,"reportableChange":0}
], {"timeout":10000,"manufacturerCode":null,"disableDefaultResponse":true}) failed (Error: Status 'UNSUPPORTED_ATTRIBUTE')          
    at Endpoint.<anonymous> (/app/node_modules/zigbee-herdsman/dist/controller/model/endpoint.js:326:23)                            
    at Generator.next (<anonymous>)                                                                                                 
    at fulfilled (/app/node_modules/zigbee-herdsman/dist/controller/model/endpoint.js:5:58))
```
I have tried to remove and re-pair with the same results.  Is there anything else I should try?  Or, more logs that would be useful?

Thanks!

Koenkk:
Fixed in latest dev branch.

Changes will be available in the latest dev branch in a few hours (https://www.zigbee2mqtt.io/how_tos/how-to-switch-to-dev-branch.html)

PR #1707, opened 2020-10-27
Fix MCT-340 E configure failing

airdrummingfool:
It looks like I messed up battery reporting in the `MCT-340` devices back in 9f7dcb202097bf1858ae46d5c34fc66de784e76d - when trying to configure for `batteryPercentageRemaining`, an `UNSUPPORTED_ATTRIBUTE ` error is thrown. Turns out it should be using `batteryVoltage`. This was already fixed for the `MCT-340 SMA` in 8cac16b031c636316a16d48ea49950ba2e97d811.

This PR is identical to the fix for #1321 (`MCT-340 E` and `MCT-340 SMA` are virtually identical as far as I can tell).


Koenkk:
Thanks!

PR #1708, opened 2020-10-28
Remove the support of TS0003_curtain_switch because of conflict with TS0003 switch. Support battery report for JAVIS lock.

dinhchinh82:
1. Remove the support of TS0003_curtain_switch because of conflict with TS0003 switch
I just found out there is conflict point when adding the TS0003_curtain_switch profile to the devices.js. By adding this configuration, all the 3 gang smart switch are recognized as curtain switch. The fingerprint of both device types are exactly same as "_TYZB01_fr1lzwta".

So please ignore this device support for the current release. I still working on this issue to find out the better solution.
2. Support battery report for JAVIS lock.


dinhchinh82:
Please note that this PR requires the following change in homeassistant.js

``'JS-SLK2-ZB': [cfg.sensor_action, cfg.sensor_battery],

Koenkk:
Thanks!

PR #1709, opened 2020-10-28
Split Saswell thermostats into 801 and 802

drzony:
Split into two models (Separate PR for zigbee2mqtt.io for image)


Koenkk:
Thanks!

PR #1715, opened 2020-10-30
Ecolink 4655BC0-R contact - add temperature

walthowd:
Adds temperature monitoring to this ecolink contact sensor


Koenkk:
Thanks!

PR #1716, opened 2020-10-31
Add support to Tuya GTZ02 radiator valve support, using Siterwell device entry

chmurson:
This PR adds support to radiator valve device that looks very much same as [Siterwell GS361A-H04](https://www.zigbee2mqtt.io/devices/GS361A-H04.html)

Zibbe2mqtt log message:
```
Received message from unsupported device with Zigbee model 'TS0601'
```
"TS601" looks like very popular model name in the `devices.js`, but apparently there are "no conflicts" in resolving the unsupported device with the current changes. I'm worries thought that they may appear in the future. I noticed that other devices that have "TS0601" as their `modelId`, have `manufacturerName` name provider under `fingerprint` field. 

Question from me: How can I get `manufacturerName` of the device, to avoid possible future conflicts of device resolving ?

------
How `TS0601` modelId popular is:
```
cat devices.js | grep TS0601
            {modelID: 'TS0601', manufacturerName: '_TZE200_whpb9yts'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_ebwgzdqq'},
        model: 'TS0601_dimmer',
        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_sbordckq'}],
        model: 'TS0601_curtain_switch',
            {modelID: 'TS0601', manufacturerName: '_TZE200_wunufsil'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_vhy3iakz'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_oisqyl4o'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_aqnazj70'}, // 4 gang
        model: 'TS0601_switch',
            {modelID: 'TS0601', manufacturerName: '_TZE200_nkjintbl'},
        model: 'TS0601_switch_2_gang',
            {modelID: 'TS0601', manufacturerName: '_TZE200_kyfqmmyl'},
        model: 'TS0601_switch_3_gang',
            {modelID: 'TS0601', manufacturerName: '_TZE200_5zbp6j0u'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_nkoabg8w'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_xuzcvlku'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_4vobcgd3'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_nogaemzt'},
        model: 'TS0601_curtain',
            {modelID: 'TS0601', manufacturerName: '_TZE200_wmcdj3aq'},
        model: 'TS0601_roller_blind',
        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_ckud7u2l'}],
        model: 'TS0601_thermostat',
        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_znzs7yaw'}],
            {modelID: 'TS0601', manufacturerName: '_TZE200_8vxj8khv'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_7tdtqgwv'},
        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_dhdstcqc'}],
        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_fqytfymk'}],
        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_aoclfnxz'}],
        zigbeeModel: ['ivfvd7h', 'eaxp72v\u0000', 'TS0601'],
```

Vendor links: 
https://allegro.pl/oferta/glowica-termostatyczna-st-zigbee-3-0-tuya-smart-9168843228
https://www.houseiq.pl/pl/p/Glowica-termostatyczna-ST-ZigBee-3.0-TUYA-Smart/955

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14158,7 +14158,8 @@ const devices = [
 
     // Siterwell
     {
-        zigbeeModel: ['ivfvd7h', 'eaxp72v\u0000', 'TS0601'],
+        zigbeeModel: ['ivfvd7h', 'eaxp72v\u0000'],
+        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_zivfvd7h'}],
         model: 'GS361A-H04',
         vendor: 'Siterwell',
         description: 'Radiator valve with thermostat',
@@ -14187,7 +14188,7 @@ const devices = [
         ],
         whiteLabel: [
             {vendor: 'Essentials', description: 'Smart home heizkörperthermostat premium', model: '120112'},
-            {vendor: "Tuya", description: "Głowica termostatyczna", model:"GTZ02"},
+            {vendor: 'Tuya', description: 'Głowica termostatyczna', model: 'GTZ02'},
         ],
     },
 
--- a/devices.js
+++ b/devices.js
@@ -14158,7 +14158,7 @@ const devices = [
 
     // Siterwell
     {
-        zigbeeModel: ['ivfvd7h', 'eaxp72v\u0000'],
+        zigbeeModel: ['ivfvd7h', 'eaxp72v\u0000', 'TS0601'],
         model: 'GS361A-H04',
         vendor: 'Siterwell',
         description: 'Radiator valve with thermostat',
@@ -14187,6 +14187,7 @@ const devices = [
         ],
         whiteLabel: [
             {vendor: 'Essentials', description: 'Smart home heizkörperthermostat premium', model: '120112'},
+            {vendor: "Tuya", description: "Głowica termostatyczna", model:"GTZ02"},
         ],
     },
 


Koenkk:
Since the `TS0601` is indeed an often used modelID I prefer to add it using the fingerprint. Can you share the `data/database.db` entry of this device?

MattWestb:
Perhaps this [RC-290.zigbee](https://www.pearl.at/at-a-NX4917-3103.shtml?vid=953&curr=ATS) is the same. I hopefully knowing that in some days :-))

chmurson:
> Since the `TS0601` is indeed an often used modelID I prefer to add it using the fingerprint. Can you share the `data/database.db` entry of this device?

I've managed to retrieve the manufacturer name from the file. I've updated the PR.

chmurson:
> Perhaps this [RC-290.zigbee](https://www.pearl.at/at-a-NX4917-3103.shtml?vid=953&curr=ATS) is the same. I hopefully knowing that in some days :-))

Yes, looks exactly the same :)

MattWestb:
Thanks for the commit !!!

The 5 pack with bridge have arriving to the postcentral in vienna so very likely i getting then in the beginning of the week :-))

I was thinking looking if the bridge is updating the firmware and perhaps capture the URL or the complete file is its possible.

chmurson:
@MattWestb Great to read that. Let us know od you come up with anything interesting. 

The model of radiator valve is one of three similarly priced I bought. I think it's the best in terms of features, as is simple and as only gives info about battery charge. However, it looks slight disappointing in real.

MattWestb:
For my its was the sipple design and local changing mode / temperature also then being offline.
I have reading that is in real its looks much more plastic than on the photos but still better that real cheap ones with small gray LCD and small membrane buttons for little more than 20 € chipped from china. I wold being nice with real battery level (%) then only low batt but its looks like all tuya is doing the same. 
Is written 2 years on one battery set but i think its needed being changed every year before the winter is coming but its not one large problem for my and only some weeks / months is too short then not being active for more then 6 months on one year. 
One missing futures i automatic conditioning of the valve that the real expensive is doing but i think its can being done with automation in HA (force heating for one minute and then back to antifreeze/off mode) in the summer with condition that the battery is not low (if its stop working with valve open its bad in the summer).

I keep you informed of the progress and giving feedback ;-))

Koenkk:
@chmurson since the model is TS0601, can you check if it works with https://github.com/Koenkk/zigbee-herdsman-converters/blob/4d98da963bc9ab286f7f9be8d46766b4f3e37e81/devices.js#L1762 instead?

chmurson:
@Koenkk 
Interesting, both configuration are working and shares similar set of very functionality. After looking more closely into it none of them is perfect. The states trio of "preset",  "system_model" and "away mode" is not working perfectly with none of the devices. 
- yours device entry @Koenkk 
  - seems to support to many values, either:
    - not supported by the device
    - allows to enter "service mode" (back to pairing mode, or rearranging valve - F1/F2 modes + flashing LED indicator) 
  - don't know if consumes properly the battery level broadcast
 - the image is incorrect 
- mine device entry on the other hand looks simpler, but:
  - incorrectly names the states
  - does not allow to enter manual mode via HA
    - only off and schedule 
  - mine was tested for a longer time, and it supports occasional broadcasts of battery indicator (not low power, but actually percentage of the battery)

@Koenkk I think I would prefer to go with mine, if we were to go with one of them, even if is not perfect. Next step would be to improve it by adjusting HA configuration. 

We could also wait until I adjust the HA configuration so the "state dropdown selectors" works better. 

How should we proceed with this one ?

chmurson:
> I have reading that is in real its looks much more plastic than on the photos but still better that real cheap ones with small gray LCD and small membrane buttons for little more than 20 € chipped from china. 

true, it's not THAT bad after all :) 

> I wold being nice with real battery level (%) then only low batt but its looks like all tuya is doing the same.

It actually reports battery level for me, but it does it only occasionally. I went down to 81% within when using it for about 1 month, with some random brand batteries that were included.

Koenkk:
@chmurson thanks for your detailed explanation. This is good for now, thanks!

PR #1722, opened 2020-11-02
Adding support for new Namron devices.

ivargj:
- 4512701: Zigbee 1 channel switch K2
- 4512719: Zigbee 2 channel switch K4


Koenkk:
Thanks!

PR #1729, opened 2020-11-04
Add FireAngel Monoxide Sensor with W2-Module Zigbee module support

tgebarowski:
Device info:
https://www.fireangel.co.uk/products/w2-co-10x

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -991,18 +991,6 @@ const converters = {
             };
         },
     },
-    ias_carbon_monoxide_alarm_2: {
-        cluster: 'ssIasZone',
-        type: 'commandStatusChangeNotification',
-        convert: (model, msg, publish, options, meta) => {
-            const zoneStatus = msg.data.zonestatus;
-            return {
-                carbon_monoxide: (zoneStatus & 1<<8) > 8,
-                tamper: (zoneStatus & 1<<2) > 0,
-                battery_low: (zoneStatus & 1<<3) > 0,
-            };
-        },
-    },
     ias_sos_alarm_2: {
         cluster: 'ssIasZone',
         type: 'commandStatusChangeNotification',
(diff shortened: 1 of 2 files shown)


converters/fromZigbee.js
```diff
@@ -991,6 +991,16 @@ const converters = {
             };
         },
     },
+    ias_carbon_monoxide_alarm_2: {
+        cluster: 'ssIasZone',
+        type: 'commandStatusChangeNotification',
+        convert: (model, msg, publish, options, meta) => {
+            const zoneStatus = msg.data.zonestatus;
+            return {
+                carbon_monoxide: (zoneStatus & 1<<8) > 8,
```
Koenkk, line 1000: Please also add:

```
                tamper: (zoneStatus & 1<<2) > 0,
                battery_low: (zoneStatus & 1<<3) > 0,
``` 

here

tgebarowski: Thanks, I've just added it.

converters/fromZigbee.js
```diff
@@ -991,6 +991,18 @@ const converters = {
...
```
Koenkk, line 1000: I've checked the other `alarm_2` converters, they all use `(zoneStatus & 1<<1) > 0,` (using the second bit). Why is this different here? Does it also work with: `(zoneStatus & 1<<1) > 0,` ?

tgebarowski: hmm.... I am not sure if I used correct naming convention for this converter (I was inspired by other monoxide converters). 
I managed to decode data from this monoxide sensor and it sent:
0b100100000 (188) when detecting monoxide and 0b100000 (32) when state was cleared. I came up with the conclusion that 8th bit is sensor state. I have no idea what is the meaning of other bits.

Koenkk: Thanks, seems it's not standard so I've created a non-generic converter for this.

PR #1732, opened 2020-11-06
Added support for zigbee 2 gang switch by eWeLink

KubaDee:
Adding new device - eWeLink 2 gang switch.
Pictures are here https://www.chytrevypinace.cz/Tlacitkovy-ZigBee-vypinac-2CH-d210.htm

Expecting that accordingly would be implemented also 1 gang and 3 gang variant. But without owning those I can only provide config for 2 gang variant.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14933,7 +14933,7 @@ const devices = [
             await configureReporting.onOff(endpoint);
         },
     },
-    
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1738, opened 2020-11-09
add support for Aqara JWSP001A and JWDL001A

zhouF96:
add support for Aqara JWSP001A and JWDL001A
JWDL001A
JWSP001A


Koenkk:
Thanks!

PR #1705, opened 2020-10-26
LoraTap mini relay for socket

kirovilya:
https://aliexpress.ru/item/1005001423113439.html

PR #1741, opened 2020-11-09
Adding support for SilverCrest 3 gang with USB from LIDL (SPSZ 3 A1)

trokamp:


trokamp:
I can change the state of a single outlet with this command:
zigbee2mqtt/0xec1bbdfffea96bd4/l1/set OFF

But the states reported back are all in the form of */state_l* as seen below, i.e. "/state_l3 OFF"
I would really like this to be in the form of */l*/state, i.e. "/l3/state OFF", but I haven't had  any success with it.

Also, when changing an outlet, it seems as the other 2 outlets feels an urge to report the very same states as well, resulting in every state to be reported three times. While I can easily live with that, it feels a bit annoying.

zigbee2mqtt/0xec1bbdfffea96bd4/state ON
zigbee2mqtt/0xec1bbdfffea96bd4/state_l1 OFF
zigbee2mqtt/0xec1bbdfffea96bd4/state_l2 ON
zigbee2mqtt/0xec1bbdfffea96bd4/state_l3 ON
zigbee2mqtt/0xec1bbdfffea96bd4/linkquality 48
zigbee2mqtt/0xec1bbdfffea96bd4/state ON
zigbee2mqtt/0xec1bbdfffea96bd4/state_l1 OFF
zigbee2mqtt/0xec1bbdfffea96bd4/state_l2 ON
zigbee2mqtt/0xec1bbdfffea96bd4/state ON
zigbee2mqtt/0xec1bbdfffea96bd4/state_l3 ON
zigbee2mqtt/0xec1bbdfffea96bd4/state_l1 OFF
zigbee2mqtt/0xec1bbdfffea96bd4/linkquality 54
zigbee2mqtt/0xec1bbdfffea96bd4/state_l2 ON
zigbee2mqtt/0xec1bbdfffea96bd4/state_l3 ON
zigbee2mqtt/0xec1bbdfffea96bd4/linkquality 48

Koenkk:
- Can you provide a link to this product (I need a picture for the docs)
- I've changed the config a bit, can you check if it still works with this?

> But the states reported back are all in the form of /state_l as seen below, i.e. "/state_l3 OFF"
I would really like this to be in the form of /l/state, i.e. "/l3/state OFF", but I haven't had any success with it.

This is expected behaviour indeed and cannot be changed

> Also, when changing an outlet, it seems as the other 2 outlets feels an urge to report the very same states as well, resulting in every state to be reported three times. While I can easily live with that, it feels a bit annoying.

Looks like strange behaviour off the switch itself (firmware), we probably have to live with it.

trokamp:
I realized, that I did not have the most recent version of Zigbee2MQTT running, and after updating to that, I can not make the device connect with the correct profile.
The problem is, that there's another profile with a more simple set of attributes, that gets in the way and becomes a candidate before my addition does.
Before I updated, it was enough to put my profile before the simple one, but something has happened to the filtering after I updated. So now order does not seem to matter.

The conflicting profile is this:
`   {
        zigbeeModel: ['TS011F'],
        model: 'TS011F',
        vendor: 'TuYa',
        description: 'Socket module',
        extend: generic.switch,
        whiteLabel: [
            {vendor: 'LoraTap', model: 'RR400ZB'},
        ],
    },`

It has the zigbeeModel (modelId) in common, but it doesn't supply a manufacturerName, which makes the existing one a "catch all" to anything that identifies itself as "TS011F", which in turn may not be what is expected. 

So either the filtering should consider more specific options over less specific options
OR the existing profile should be modified to take care of my addition as well, but I don't know how that is done without breaking the existing one (LoraTap RR400ZB).

trokamp:
Like I also wrote in the commit message, I had to combine my addidtion with an existing profile, as they otherwise would share the same zigbeeModel, which is illegal.
As I have no way of testing that the other item still works, I really don't know what to do from here.

When the code is loaded, I can control the device both with and without setting the endpoints, i.e. the following two commands does the same, so I'm not that worried :-)
.../l1/set OFF
.../set OFF

I have attached an image of the product.


SPSZ_3_A1

trokamp:
Oh and also - looks like it doesn't spam the MQTT queue any more. Only one message is submitted per state now, instead of multiple.

Koenkk:
The existing TS110F also needs to be identified via a fingerprint, we cannot share the definitions since the current TS110F only has one endpoint (and this one 3). Asked for the fingerprint in https://github.com/Koenkk/zigbee-herdsman-converters/pull/1705 and will update your PR once I have a response.

kirovilya:
My LoraTap socket module record (only one endpoint):
`
{"id":6,"type":"Router","ieeeAddr":"0x680ae2fffeb303a6","nwkAddr":53852,"manufId":4098,"manufName":"_TZ3000_oiymh3qu","powerSource":"Mains (single phase)","modelId":"TS011F","epList":[1],"endpoints":{"1":{"profId":260,"epId":1,"devId":9,"inClusterList":[0,4,5,6],"outClusterList":[25,10],"clusters":{"genBasic":{"attributes":{"modelId":"TS011F","manufacturerName":"_TZ3000_oiymh3qu","powerSource":1,"zclVersion":3,"appVersion":65,"stackVersion":0,"hwVersion":1,"dateCode":""}},"genOnOff":{"attributes":{"32770":2,"onOff":0,"onTime":0,"offWaitTime":0,"tuyaBacklightMode":1}}},"binds":[],"meta":{}}},"appVersion":65,"stackVersion":0,"hwVersion":1,"dateCode":"","zclVersion":3,"interviewCompleted":true,"meta":{},"lastSeen":1605067731496}`

trokamp:
I will revert my change that combined the two and push an updated TS011F and SilverCrest with the above information applied.

trokamp:
I have now changed the definition of TS011F according to the info given by @kirovilya - Can you verify that your device still works?

kirovilya:
I think it will work, only I have to change the name of the model.
I'll check in the evening

trokamp:
The name of the model was changed, only to make it more specific. I can revert it back to TS011F if it is a problem regarding backwards compatibility, i.e. other users unable to use their devices due to a name change?

kirovilya:
I think no need to revet name - 'TS011F_socket_module' is normal for me :)

trokamp:
Then I think it would make sense, to also consolidate the SilverCrest definition, so it is altso a TuYa device with a whiteLabel of SilverCrest etc.

Koenkk:
Looks good, thanks!

segalion:
I want to ask for this features:
• Multiple timer programs
• Countdown function
• Automatic switch on/off (at sunset/sunrise)
• Automatic time changeover (summer/winter) 
supposed in this device:
https://zigbeealliance.org/zigbee_products/smart-connect-strip-4/

Do you know if can be possible to send an command to "shutdown one switch for 1 minute" so the switch could power up automatically without another command?

This is very very useful for i,e. a watchdog to shutdown raspberry pi, conected to this switch. Its a not very common functionality, that supposes the harware has some kind of timer inside.

Thanks for support.

trokamp:
As far as I can tell, there's no support for anything like that, directly on the device. If you want timers, automation, etc., you will have to program it with HomeAssistant, OpenHAB or similar. As long as you don't boot your zigbee2mqtt or HA/OH machine, you will also be able to use your "watchdog" feature. It's just a matter of timers.

The device is just a "dumb" power switch, where you can turn on/off 3 sockets individually.

PR #1742, opened 2020-11-09
Big Tuya rework

drzony:
This rework separates data point ids from data type, creates a common list of Tuya data points and adds some other fixes.

Also it adds some debugging tools for Tuya devices

According to https://github.com/Koenkk/zigbee-herdsman/issues/249

Should me merged together with https://github.com/Koenkk/zigbee-herdsman/pull/253

### DIFF
--- a/converters/common.js
+++ b/converters/common.js
@@ -103,6 +103,157 @@ const TuyaThermostatForceMode = {
     2: 'close',
 };
 
+const TuyaThermostatScheduleMode = {
+    1: 'single', // One schedule for all days
+    2: 'weekday/weekend', // Weekdays(2-5) and Holidays(6-1)
+    3: 'weekday/sat/sun', // Weekdays(2-6), Saturday(7), Sunday(1)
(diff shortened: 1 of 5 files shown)


Koenkk:
@Hacker-CB can you do a final check for these changes, does it still work with your ETOP thermostats after this? Also requires https://github.com/Koenkk/zigbee-herdsman/pull/253

Hacker-CB:
@Koenkk I checked with HT-08. All basic functions, including `away` mode works as expected.
For the `weekly_schedule` - we still need to make some exposes standard.

Koenkk:
@Hacker-CB thanks for testing, @drzony big thanks for doing this!

drzony:
I'm out of steam today, but I'll do (maybe I'll manage tomorrow) a PR for zigbee2mqtt.io with a documentation about "Adding new Tuya devices", I'll describe how to use the "debug" converters, my python script and how to add data points

PR #1743, opened 2020-11-10
Add support for RGB Genie ZB-5121

airdrummingfool:
Here's basic support for the RGB Genie ZB-5121 "Micro remote with single scene recall".

The remote generates some OTA request traffic, but at this time I don't have enough information to add OTA support. I have reached out to RGB Genie, and hopefully I'll be able to add it in a future PR.


Koenkk:
Thanks!

PR #1747, opened 2020-11-11
Expose position for compatible thermostats

ayufan:
This adds `position` as an HA exposed parameter.

Requires: https://github.com/Koenkk/zigbee2mqtt/pull/4936


Koenkk:
Thanks!

PR #1748, opened 2020-11-11
Added support for SilverCrest motion sensor

trokamp:


trokamp:
SMSZ_1_A1

devices.js
```diff
@@ -1790,6 +1790,24 @@ const devices = [
...
+            {vendor: 'SilverCrest', model: 'SMSZ 1 A1'},
+        ],
+        fromZigbee: [fz.battery, fz.ias_occupancy_alarm_1],
+        toZigbee: [],
+        meta: {configureKey: 1},
+        configure: async (device, coordinatorEndpoint) => {
+            const endpoint = device.getEndpoint(1);
+            await bind(endpoint, coordinatorEndpoint, ['genPowerCfg']);
+        },
+        exposes: [e.battery(), e.occupancy()],
```
Koenkk, line 1809: `fz.ias_occupancy_alarm_1` potentially also supports, `e.low_battery()` and `e.tamper()`, can you check if this works?

trokamp: It works... I get an update 'false' for the low_battery and tamper is true, when I unclick the device. There's a small push-button on the back, which is pressed when the device is clicked in place. Then the tamper is false.

Koenkk: I expect battery low to work, about the tamper usually it can be triggered by opening the device (e.g. to replace the battery). It should go to true after doing this.

trokamp: It certainly goes to false if tampered with. It's enough to unclick the device from its mount, then the button on the back is released, hence setting tamper to true and back to false when remounted.

Koenkk:
Thanks!

PR #1749, opened 2020-11-11
Added Philips Hue Bloom with BT (929002375901)

K-r-i-s-t-i-a-n:
The excisting 2 versions of Bloom in devices.js do not include colortemp, not sure if they support it. The new version does, so I added it separately.

https://www.philips-hue.com/de-de/p/hue-white---color-ambiance-bloom-tischleuchte/8718699770983


Koenkk:
Thanks!

PR #1363, opened 2020-06-30
Fix Iris 3320L Conversion to only examine Alarm1 bit

r3dey3:
Iris 3320L from zigbee conversion uses all bits of the zonestatus to determine if the sensor is open or closed; However only bit 0 actually indicates the open close status. (See https://docs.smartthings.com/en/latest/ref-docs/zigbee-ref.html#accessing-a-property-attribute)

Fixes #1362

PR #705, opened 2019-10-22
Improve support for Centralite 4257050-ZHAC dimmable plug

airdrummingfool:
This PR adds dimming and fixes power reporting for the Centralite 4257050-ZHAC dimmable plug.

There are a couple notes I want to make known before this is merged:
- The Centralite `4257050-ZHAC` does not automatically restore the previous dim level when powered off and back on. E.g. turn on, set dimmer to 50%, turn off, turn back on: dimmer is at 100% (I'd expect 50%). I added a workaround to support this using the existing z2m state mechanism. If this is not desired, I can certainly remove it; on the other hand, I believe my changes are generic enough that they might work for other dimmable devices with similar issues.
- I suspect that power reporting for `4256251-RZHAC` and `4257050-RZHAC` is broken (wrong powers of 10, doesn't take into account that different attributes info comes in different messages), but as I do not have these devices to test with, I don't want to mess with the code.

Please let me know if you have any concerns or want changes!

PR #1752, opened 2020-11-12
Fix Iris 3210-L and Centralite 4257050-ZHAC configure

airdrummingfool:
Hi Koen,
I noticed that configuration was failing on my Iris 3210-L smart plugs. The error was `UNSUPPORTED_ATTRIBUTE`, and it turns out that they (at least the two I have) don't support the `acVoltageMultiplier` or `acVoltageDivisor` attributes of the `haElectricalMeasurement` cluster. The voltage is reported in volts (so it needs no multiplier or divisor).

Apparently I broke this in #705 😞 , and Centralite 4257050-ZHAC dimming plugs are also affected. I'm pretty surprised this got past me (and stood for a year), but as I have never been able to update the firmware on these things, I'm not sure how anything could have changed behavior-wise... and it is definitely throwing errors now.

With these changes, my 3210-L plugs and 4257050-ZHAC dimming plugs now configure fine. I'm not sure if this needs a `configureKey` bump or not.

I also noticed some unused code (due to #1363) and removed it.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3578,13 +3578,6 @@ const converters = {
             return {};
         },
     },
-    iris_3320L_contact: {
-        cluster: 'ssIasZone',
-        type: 'commandStatusChangeNotification',
-        convert: (model, msg, publish, options, meta) => {
-            return {contact: msg.data.zonestatus === 36};
-        },
-    },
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks, this does not need a bump of the configureKey, if configure failed it will always re-execute it.

PR #1757, opened 2020-11-13
Adding other Philips Centura SKU

keesschollaart81:
Same device as this one: https://www.zigbee2mqtt.io/devices/5045148P7.html but with another name. Tested it locally, works!


Koenkk:
Thanks Kees! :smile:

PR #1758, opened 2020-11-14
Support for Develco SPLZB-131

timmikk:
Noticed that in the support for Develco SPLZB-132 power plug was added recently. That device looks very similar to the SPLZB-131 that I have, so I just copied it's entry and changed the device model. After quick testing the SPLZB-131 seems to be now working with the Zigbee2mqtt.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14352,6 +14352,31 @@ const devices = [
 
     // Develco
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1763, opened 2020-11-15
Add support for GL-H-001 Gledopto RF bridge

sfstar:
Add configuration that allows GLEDOPTO RF bridge GL-H-001 to be recognized and adopted.
Currently known issues:
The RGB color selection seems to happen in a shifted matter when using the homeassistant color wheel.
(pink translates to a yellowish color while light green is translated to a shade of blue).
The scene selection (breathe, blink etc.) isn't working.

Image of the device: https://images-na.ssl-images-amazon.com/images/I/41j2JWzGjPL._AC_SL1000_.jpg

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -6902,6 +6902,18 @@ const devices = [
         },
     },
     {
+        fingerprint: [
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1768, opened 2020-11-16
Added GLEDOPTO Pro Models Koenkk/zigbee2mqtt#4852

regnerus:
Added the new GLEDOPTO Pro models. Gledopto mentions that one of the new features is the ablility to set the min. brightness much darker than the regular/plus versions. When I added the Pro controllers I don't see any difference however, should something else be changed or is it just a false marketing promise?

See also issue: Koenkk/zigbee2mqtt#4852

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -6990,7 +6990,7 @@ const devices = [
         zigbeeModel: ['GL-C-008P'],
         model: 'GL-C-008P',
         vendor: 'Gledopto',
-        description: 'Zigbee LED controller RGB + CCT plus model',
+        description: 'Zigbee LED controller RGB + CCT pro model',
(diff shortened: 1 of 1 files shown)


Nebr88:
Hi @regnerus . Could you please tell if GL-D-003P was included

Koenkk:
@regnerus I don't expect the brightness can be set lower or they are using some non standard Zigbee stuff for this (which I don't expect since they claim Hue compatibility and don't have a own bridge (AFAIK))

regnerus:
@Koenkk thanks for merging the Pull request! I contacted Gledopto about this, if there is some different way for setting the brightness lower I will update.

PR #1769, opened 2020-11-16
Added converter for LeTV 8key FreePad mod

xyzroe:
[LeTV 8key FreePad mod](https://modkam.ru/?p=1791)

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -4440,6 +4440,55 @@ const devices = [
         },
     },
     {
+        zigbeeModel: ['FreePad_LeTV_8'],
+        model: 'FreePad_LeTV_8',
+        vendor: 'DIYRuZ',
+        description: '[LeTV 8key FreePad mod](https://modkam.ru/?p=1791)',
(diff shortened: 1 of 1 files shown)


Koenkk:
This looks very similar to https://github.com/Koenkk/zigbee-herdsman-converters/commit/4170fdcd93c40aed3d372935b8c07e0bc8cfd40e , @nurrik should that one be reverted?

xyzroe:
@Nurrik used incorrect model id so it won’t work.
In addition, when using one converter, the number of buttons is incorrect, the image is incorrect., picture and keys count.
It seems logical to me to use different converters for different devices, even if the firmware is almost the same.

nurikk:
> This looks very similar to [4170fdc](https://github.com/Koenkk/zigbee-herdsman-converters/commit/4170fdcd93c40aed3d372935b8c07e0bc8cfd40e) , @nurikk should that one be reverted?

It's the same firmware flashed into different devices

Koenkk:
good, thanks!

PR #1770, opened 2020-11-16
SNZB-02 reporting frequency increased

Pigotka:
* increased frequency to once per 15min or 0.25 degree (once per hour was not enough for temperature regulation)
* fixes Koenkk/zigbee2mqtt#4958

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -13441,7 +13441,7 @@ const devices = [
         exposes: [e.battery(), e.temperature(), e.humidity()],
         fromZigbee: [fz.temperature, fz.humidity, fz.battery],
         toZigbee: [],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

roflmao:
@Pigotka Wont this possibly drain the battery more? Would it be possible to make this configurable?

Pigotka:
> Wont this possibly drain the battery more? Would it be possible to make this configurable?
@roflmao please read the bug, I described exactly this there
In short, I'm missing knowledge about possibilities of z2m and whether this is device or sw thing. If former the it could affect battery life. But what I did is that now the sensor behaves like for example Aquara ones and they even have smaller battery.

PR #1775, opened 2020-11-17
Add temperature sensor plus limit network spamming to SPLZB-132 (mimicks jw123's SPLZB-134 patch)

NameLessJedi:


Koenkk:
Thanks!

PR #1778, opened 2020-11-18
Snzb freq

Pigotka:
* further adjusting frequency to match other temperature sensors
* 0.5 degree change should eliminate frequent reports for big temperature changes and max report once per 30min reduce updates for constant temperature environments
Note: not changing configureKey as previous was not released yet


Koenkk:
Thanks!

PR #1782, opened 2020-11-19
battery_low added to exposes for Tuya TRV (moes)

mgrom:
As long TRV doesn't report battery in percentage, but reports battery low status, this should be added to exposed.


Koenkk:
Thanks!

PR #1786, opened 2020-11-21
ZBT-CCTSwitch-D0001: dontDividePercentage should be false

qosmio:
This remote reports battery percentage from 200
so it should be divided by 2 to display properly.


Koenkk:
Thanks, the default is `false` so I've removed it.

PR #1788, opened 2020-11-21
Update devices.js Add support for Shutters central remote switch Legrand 067646

ozer59:
Add support for Shutters central remote switch Legrand 067646
work for me


Koenkk:
Thanks!

PR #1791, opened 2020-11-22
Ubisys S1/S2 default bindings should be stored

sjorge:
This PR fixes https://github.com/Koenkk/zigbee2mqtt/issues/4682 and stores the binding state of a fresh/factory reset S1 or S2 switch.

This allows for easy unbind/binding via the frontend. This PR required `addBindings` from zigbee-herdsman so it depends on this PR https://github.com/Koenkk/zigbee-herdsman/pull/256.

This was tested by removing the device from the network, allowing it to join and observing no binding.
Then do the same after this change, the default bindings show up and I was able to remove them via the frontend.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -12532,6 +12532,23 @@ const devices = [
             await readMeteringPowerConverterAttributes(endpoint);
             await configureReporting.instantaneousDemand(endpoint);
         },
+        onEvent: async (type, data, device) => {
+            /*
+             * As per technical doc page 18 section 7.3.4
+             * https://www.ubisys.de/wp-content/uploads/ubisys-s1-technical-reference.pdf
+             *
+             * This cluster uses the binding table for managing command targets.
+             * When factory fresh, this cluster is bound to endpoint #1 to
+             * enable local control.
+             *
+             * We use addBinding to 'record' this default binding.
+             */
+            if (type === 'deviceInterview') {
+                const ep1 = device.getEndpoint(1);
+                const ep2 = device.getEndpoint(2);
+                ep2.addBinding('genOnOff', ep1);
+            }
+        },
         ota: ota.ubisys,
     },
     {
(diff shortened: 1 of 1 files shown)


sjorge:
CC @felixstorm I don't have J1/D1/C4 to test, but once this is in, maybe you can do the same for those devices as I believe you at least have the C4?

devices.js
```diff
@@ -12532,6 +12532,23 @@ const devices = [
             await readMeteringPowerConverterAttributes(endpoint);
             await configureReporting.instantaneousDemand(endpoint);
         },
+        onEvent: async (type, data, device) => {
```
Koenkk, line 12535: Question: why not do this in `configure()`?

sjorge:
We don’t want to restore the bindings if someone runs configure on it via the api.
Well, add them to the _binds property again, as we’re just recording the devices defaults... which are only toe after joining until the user unbind them.

I did not find a way to detect if the device was in interviewing mode from configure.

~ sjorge

Koenkk:
Agree, thanks!

devices.js
sjorge: > We don’t want to restore the bindings if someone runs configure on it via the api.
> Well, add them to the _binds property again, as we’re just recording the devices defaults... which are only there after joining until the user unbind them.
> 
> I did not find a way to detect if the device was in interviewing mode from configure.
> 
> If we drop the configureKey and can ensure configure is only run once unless your remove/repair a device this can be moved. But we’re not there yet.
> 

From the e-mail, so we can resolve this cleanly for github

sjorge:
@felixstorm looks like this made it into the latest prd release of z2m now :)

felixstorm:
@sjorge Thanks! I am a bit busy now, but will hopefully find the time to implement it for the other ubisys devices after christmas.

PR #1793, opened 2020-11-22
update model for Lidl power strip

blakadder:
setting model number to whats on zigbeealliance.org which brings it in line with other Lidl devices


Koenkk:
Thanks!

PR #1797, opened 2020-11-23
Added Silvercrest (Lidl) Smart LED light strip 2M

PedroQ:


devices.js
```diff
@@ -15012,6 +15012,14 @@ const devices = [
             return {'l1': 1, 'l2': 2, 'l3': 3};
         },
     },
+    {
+        fingerprint: [{modelID: 'TS0505A', manufacturerName: '_TZ3000_riwp3k79'}],
+        model: 'HG06104A',
+        vendor: 'Silvercrest',
+        description: 'Smart LED light strip 2M',
+        extend: generic.light_onoff_brightness_colortemp_colorxy,
+        meta: {applyRedFix: true},
```
Koenkk, line 15021: Have you checked that this is really needed? (`applyRedFix`)

PedroQ: Yes, had do add it because the controller doesn't change color when setting it to HA's red (x: 0.701 y: 0.299)

Koenkk: Thanks

PR #1805, opened 2020-11-25
Use float types for tuya temperature readings

jekkos:
As I was visualizing some of the temperature reading, I found that they are being sent as strings instead of floats. I guess this was done because of the toFixed that truncates the number of decimals (and then turns it into a string). However most of the tooling for visualization expects this data to be numbers and not strings.


Koenkk:
Thanks!

PR #1809, opened 2020-11-27
HSB fix for HG06467

MichielMag:
Found out that HA sends decimal values for saturation, but the device doesn't like that.
Added rounding for that number.

Also: the hue expects 1-359 instead of 1-360.

And only use state.brightness, because that is what HA does.


Koenkk:
Thanks!

PR #1811, opened 2020-11-27
added new modelId for Feibit STH01ZB

incub77:
I've ordered this device as "EWelink Smart Temperature And Humidity Sensor". It identifies with manufacturerName "fei" and looks like the Feibit device. Long story short, adding the modelId for the Feibit makes this thing work just fine.


Koenkk:
Thanks!

PR #1815, opened 2020-11-28
Improve scene error messages

sjorge:
See koenkk/zigbee2mqtt#5102, 

- warn when we can't restore state for a scene_recall (but don't error, as we still send the zigbee command and the device could have still have scenes we don't know about)
- split generic into missing ID and wrong payload type for scene_add


Koenkk:
Thanks!

PR #1820, opened 2020-11-29
Add Fingerprint for Silvercrest Type J Switch

bfritscher:
In Switzerland we have Type J Sockets which has a different fingerprint.
https://zigbeealliance.org/classic-product-search/#zigbeecertifiedproducts/productdetails3/5ea15274d8a12e0017eeae01/


Koenkk:
Thanks!

PR #1821, opened 2020-11-29
Сhanged priority to use external converters

xyzroe:
findByZigbeeModel and findByDevice return last candidate of all possible

Please check is it possible to do like this?
I got everything working.

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,9 +36,7 @@ function getFromLookup(zigbeeModel) {
     return lookup.get(zigbeeModel);
 }
 
-function addDefinition(definition, addInFront) {
-    definition['addInFront'] = addInFront ? true : false;
-
+function addDefinition(definition) {
     definitions.push(definition);
 
     if (definition.hasOwnProperty('fingerprint')) {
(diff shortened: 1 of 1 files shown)


Koenkk:
Instead of doing this I prefer to add an additional parameter to https://github.com/Koenkk/zigbee-herdsman-converters/blob/3ec30751ac017104d7371195e45923be69930bc0/index.js#L39 called `addInFront` (default false) and set that to true here: https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/extension/externalConverters.js#L13

xyzroe:
And then, at https://github.com/Koenkk/zigbee-herdsman-converters/blob/3ec30751ac017104d7371195e45923be69930bc0/index.js#L59 and https://github.com/Koenkk/zigbee-herdsman-converters/blob/3ec30751ac017104d7371195e45923be69930bc0/index.js#L68 check if exists any item with addInFront = true, return than one? Right?
Ok I'l work on it later.

xyzroe:
I tried, but I'm not so good at tests writing. 
But I checked everything manually. 
I am my opinion everything works as expected.

nurikk:
> Instead of doing this I prefer to add an additional parameter to
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/3ec30751ac017104d7371195e45923be69930bc0/index.js#L39
> 
> called `addInFront` (default false) and set that to true here: https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/extension/externalConverters.js#L13

Why do we need to do this? Why not let to override internal converter by default? This behaviour is quite safe, since user need to do two steps:
1) put external converter in data folder
2) add external converter into configuration.json
This can't cause any unexpected behaviour.

xyzroe:
I started from made returning last candidate of all possible, it is the easiest way to implement this function.

Koenkk:
@xyzroe sorry to switch opinions, but I think the solution provided by @nurikk is better indeed.

So the only changed required here is:

```js
- return candidates ? candidates[0] : null; 
+ return candidates ? candidates[candidates.length - 1] : null;
```

xyzroe:
How about findByDevice() ?  
I think it is also necessary to change the order

nurikk:
> How about findByDevice() ?
> I think it is also necessary to change the order

Yes, apply this logic everywhere where it necessary. I just gave small example

xyzroe:
returned everything as it was originally in the commit e064dd9

Koenkk:
Thanks!

PR #1832, opened 2020-12-01
Fixes #1817

### LINKED ISSUE #1817: iCasa Keypad Pulse wrong battery value
iCasa Keypad model ICZB-KPD1* (* 2S or 4S or 8S) exposes a battery percentage from 0-100 only its converted with a scale of 0-200 which results in half of what it should be. Tested with a new battery which changed from 37.5 (old battery) to 50 (new battery). Old battery didn`t needed to be changed if right value was exposed. Is this a easy fix?

m4v1:
Fixed #1817 adding: 
```javascript
battery: {dontDividePercentage: true}
```
to selected devices.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -9673,6 +9673,7 @@ const devices = [
         model: 'ICZB-KPD14S',
         vendor: 'iCasa',
         description: 'Zigbee 3.0 Keypad Pulse 4S',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Issue #1573, opened 2020-09-20
Request to add Phillips Hue Centris fixture


TheGreyDiamond:
Please add the Phillips Hue Centris fixture (in the 4-spot variant). I would do it myself, but the docs on adding a light are not very good and I'm a noob in that area. I do not have the light just yet, but it will be delivered to me soon. Any help is appreciated.
(Product link: https://www.philips-hue.com/de-de/p/hue-white---color-ambiance-centris-4er-deckenspot/5060730P7 )

Koenkk:
Once you receive it, pair it to Zigbee2MQTT and provide the logging of this.

Also what does this bulb support: color? color_temperature? both?

TheGreyDiamond:
I will surely do! I think all of the parts are "Hue White Color Ambiance" so, both.

TheGreyDiamond:
They updated the shipping day and moved it to the middle of November. This will take some time.

TheGreyDiamond:
@Koenkk Please remove the stale lable, cause its still a thing. The shipping has been moved (again)

Koenkk:
@TheGreyDiamond stale label is automatically removed once you comment

TheGreyDiamond:
Okay, weird because its still visible for me.

Koenkk:
@TheGreyDiamond it will do it at 12:00

TheGreyDiamond:
Hmm still stale.... Sorry to annoy you. I finally got shipping information, so it should arrive soon.

copystring:
I have got this. Added support in devices.js and works just fine.

Not sure how to create a PR... my fork is not up to date and not sure how to update. Is it OK if I post my changes here?

copystring:
OK. Not sure if the way I did it is OK. Tried to merge upstream branch to mine. Did not work 100% but in the end only devices.js is modified. Hope this is ok like this. Not used to git/github. Let me know.

TheGreyDiamond:
I also got the lamp now and it just joined as a hue bulbish device, so....

PR #1834, opened 2020-12-01
Support Philips Hue Bluetooth White & Color Ambiance Spot Centris with 4 Spots

copystring:
See https://github.com/Koenkk/zigbee-herdsman-converters/issues/1573


Koenkk:
Can you provide the `data/database.db` entry of these devices?

copystring:
https://drive.google.com/file/d/17HQhlE-OqvrV3Dt3zatV1hPYdZ1Kp7aM/view?usp=drivesdk

Koenkk:
If I correctly understand, this is a single device (with 4 spots) that pairs as 4 separate devices? If so, I think it makes more sense to merge all the zigbeeModels (`zigbeeModel: ['5060730P7_01', '5060730P7_02', etc...],`) and change the model to `5060730P7` so that we only have 1 definition.

copystring:
Yes it is one device with spots but it also has one panel making it 5 in total.
Sure can make it one model instead.

Koenkk:
Thanks!

PR #1835, opened 2020-12-01
Add Philips Hue outdoor Impresss wall lamp (low-voltage)

avacore:
This pull request adds Zigbee device model 1745930P7 which is a low voltage version of the Hue Impress wall lamp. Copies the configuration of the other Hue Impress wall lamps, which seems to work fine.


Koenkk:
Thanks!

PR #1837, opened 2020-12-02
scene_add should handle color_temp for ZCL version <= 6

sjorge:
ColorTemperatureMireds was added in ZCLv7, not many (any) currently support his.

```
Since there is a direct relation between ColorTemperatureMireds and XY, color temperature, if supported, is stored as XY in the scenes table.
```

Seems to indicate it is OK to store it as XY instead, this change introduces a new optional property `ZCL` that specifies the ZCL version.
We default it to v6 for most compatibility, but when set to >= 7 we will use the new attribute. This will also allow us to do other things in the future
if the ZCL gets updates again.

In practice the default of ZCLv6 should be fine for most people, bulb seem to properly report colorTemperature after the change.
(Tested on Trådfri WS)

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3685,13 +3685,27 @@ const converters = {
             const sceneid = value.ID;
             const scenename = '';
             const transtime = value.hasOwnProperty('transition') ? value.transition : 0;
-            let zclversion = 6;
-            if (value.hasOwnProperty('ZCL')) {
-                if (typeof value.ZCL !== 'number') {
-                    throw new Error('ZCL version must be a number.');
-                }
-                zclversion = value.ZCL;
-            }
+            /*
+             * ZCL version 7 added support for ColorTemperatureMireds
+             *
+             * Currently no devices seem to support this, so always
+             * fallback to XY conversion. In the future if a device
+             * supports this, or other features get added this can
+             * be uncomment to add version specific code paths.
+             *
+             * Conversion to XY is allowed according to the ZCL:
+             * `Since there is a direct relation between ColorTemperatureMireds and XY,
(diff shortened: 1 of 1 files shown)


sjorge:
This should fix koenkk/zigbee2mqtt#4926 the ZCLv6 vs ZCLv7 also would explain the behavior of the trådfri 5 button remote just using XY also.

sjorge:
@Koenkk we could probably drop the whole version think and just always do the conversion, but I like the option to have it use the native attribute when the bulbs support it, this also allows us to expand in the future and gate behind the version if new things get added.

Koenkk:
> @Koenkk we could probably drop the whole version think and just always do the conversion, but I like the option to have it use the native attribute when the bulbs support it, this also allows us to expand in the future and gate behind the version if new things get added.

Since we did not found any bulbs which support this I propose to remove the version part and just always use color X/Y (currently we can't even verify if this works). The current code can be left commented with a description.

sjorge:
OK, still seems to work with it comment, let me clean it up a bit more.

MattWestb:
From where is Z2M getting the `zclversion` ?
Basic cluster attribute 0x0000 ?

sjorge:
Nowhere, you have to pass as a ZCL property in the payload. But I commented the code now as Koen asked.

MattWestb:
Good because the the ZCL ver can only being (0) 1, 2 and 3 for detecting if its Zigbee 3 or not or greenpower device stacks then pairing.
And the first 7 is coming then you is finding one and adding its in herdsman-converters :-))

Koenkk:
Big thanks! 🎖️

PR #1838, opened 2020-12-02
Adding support for LED Bulb by Lonhonso ZB-RGBCW

TSergeymsk:
Adding support for LED Bulb by Lonhonso

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -15387,7 +15387,7 @@ const devices = [
             await configureReporting.onOff(endpoint);
         },
     },
-    
+
     // Led RGBCW Bulb, Lonsonho, Tuya
     {
         zigbeeModel: ['ZB-RGBCW'],
--- a/devices.js
+++ b/devices.js
@@ -15387,6 +15387,15 @@ const devices = [
             await configureReporting.onOff(endpoint);
         },
     },
+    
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -2302,6 +2302,13 @@ const devices = [
             await configureReporting.onOff(endpoint);
         },
     },
+    {
+        zigbeeModel: ['ZB-RGBCW'],
+        model: 'Zigbee 3, RGW+W+C',
```
Koenkk, line 2307: Please fill in a meaningful model here, preferably something that can be Googled and leads to this bulb.

TSergeymsk: Unfortunately, there is no meaningul model in the paper manual nor in the Aliexpress lot. Just "ZigBee 3.0 Tuya Smart Led Lamp Bulb Light E27 220V RGB+W+C" in description.
I've seen several mentions of id ZB-RGBCW on git, but without any model number.

Koenkk: No problem, can you provide a picture of the device?

TSergeymsk: Sure.
ZB-RGBCW

Koenkk: Thanks!

PR #1842, opened 2020-12-03
TS0601_thermostat system_mode fix

drzony:
Possibly fixes https://github.com/Koenkk/zigbee2mqtt/issues/5185

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -435,15 +435,12 @@ const tuyaThermostat = (model, msg, publish, options, meta) => {
     case common.TuyaDataPoints.mode: {
         const ret = {};
         const presetOk = utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatPreset').hasOwnProperty(value);
-        const modeOk = utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatSystemMode').hasOwnProperty(value);
         if (presetOk) {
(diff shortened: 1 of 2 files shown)


converters/fromZigbee.js
```diff
@@ -435,15 +435,12 @@ const tuyaThermostat = (model, msg, publish, options, meta) => {
     case common.TuyaDataPoints.mode: {
         const ret = {};
         const presetOk = utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatPreset').hasOwnProperty(value);
-        const modeOk = utils.getMetaValue(msg.endpoint, model, 'tuyaThermostatSystemMode').hasOwnProperty(value);
```
Koenkk, line 438: If I understand correctly, this thermostat doesn't have modes but only presets?

drzony: I don't have the device, but it seems that the person who wrote this code confused system_mode and preset. In HA there are 'off', 'heat', 'cool', 'fan_only' and 'auto' system modes. So TRV can have at most 'off' and 'heat'. In this case I cannot see any data point that could turn off the thermostat, so only 'heat' is left.

drzony: The case here that this is not really a thermostat, but a thermostatic valve regulator

drzony: I think the same problem will be with Siterwell TRV, but nobody mentioned it yet

Koenkk: Thanks for explaining!

PR #1856, opened 2020-12-05
Added BYUN M415-6C smoke sensor

xyzroe:
https://item.taobao.com/item.htm?spm=a1z09.2.0.0.5c2e2e8d5AD1Ti&id=620637370020&_u=6209ep2lc6818f

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -15444,6 +15444,18 @@ const devices = [
             await configureReporting.onOff(endpoint);
         },
     },
+    
+    //BYUN
+    {
+        zigbeeModel: ['Windows switch  '],
+        model: 'M415-6C',
+        vendor: 'BYUN',
+        description: 'Smoke sensor',
+        supports: 'Smoke alarm',
+        fromZigbee: [fz.byun_smoke_on, fz.byun_smoke_off],
+        toZigbee: [],
+        exposes: [e.smoke()],
+    },
 ];
 
 
--- a/devices.js
+++ b/devices.js
@@ -15451,7 +15451,6 @@ const devices = [
         model: 'M415-6C',
         vendor: 'BYUN',
         description: 'Smoke sensor',
-        supports: 'Smoke alarm',
         fromZigbee: [fz.byun_smoke_on, fz.byun_smoke_off],
         toZigbee: [],
         exposes: [e.smoke()],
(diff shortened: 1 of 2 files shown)


devices.js
```diff
@@ -15444,6 +15444,18 @@ const devices = [
...
     },
+
+    // BYUN
+    {
+        zigbeeModel: ['Windows switch  '],
+        model: 'M415-6C',
+        vendor: 'BYUN',
+        description: 'Smoke sensor',
+        supports: 'Smoke alarm',
+        fromZigbee: [fz.byun_smoke_on, fz.byun_smoke_off],
```
Koenkk, line 15455: Can you try with `fromZigbee: [fz.ias_smoke_alarm_1]` instead here?

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/486

devices.js
xyzroe: I conducted a test for 48 hours, tried low supply voltage, case tamper, test and real triggers: the sensor sends only two types of clusters `ssIasZone` and `pHMeasurement`, when triggered and restored, respectively.I tried with `ias_smoke_alarm_1`, this allows you to get a smoke alarm, but then it returns a lot of arguments not related to this device (enrolled, tamper, battery_low,  supervision_reports, restore_reports, trouble, ac_status).

It seems to me that information that is deceiving is not worth showing (battery, tamper so on), so I made my own converter. 
also the proposed converter allows to process the smoke sensor restore message.

xyzroe: I conducted a test, tried low supply voltage, case tamper, test and real triggers: the sensor sends only two types of clusters `ssIasZone` and `pHMeasurement`, when triggered and restored, respectively. [pastebin](https://pastebin.com/7M6Lzxif)
I tried with `ias_smoke_alarm_1`, this allows you to get a smoke alarm, but then it returns a lot of arguments not related to this device (enrolled, tamper, battery_low, supervision_reports, restore_reports, trouble, ac_status).
It seems to me that information that is deceiving is not worth showing (battery, tamper so on), so I made my own `fz.byun_smoke_on`
Also to process the smoke sensor restore message I made `fz.byun_smoke_off`

Koenkk: Thanks for looking into this.

Koenkk:
Thanks!

xyzroe:
Thanks for your work!

PR #1857, opened 2020-12-05
Added support for Philips Hue 4" and 5/6" color downlights

artemkolotilkin:
Adding support for Philips Hue white and color ambient lights:
https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-downlight-4-inch/5996511U5
https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-downlight-5-6-inch/5996611U5

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -3422,6 +3422,15 @@ const devices = [
         ota: ota.zigbeeOTA,
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1864, opened 2020-12-06
Improved sysytem_mode/preset handling

drzony:


drzony:
@Koenkk A few words of explanation:
1. Home Assistant seems to display 'auto' system_mode as schedule: image
2. Saswell thermostats have schedule on a separate DP ID, so it's possible to make it work
3. TS0601_thermostat has schedule as one of the presets, so it will be a bit more difficult to do it as system_mode, I'll look into this later.
4. Any input about doing point 3 would be welcome (do we want to keep it as preset or do some weird logic to make it work HA style?)

Koenkk:
- Can this be merged?
- AFAIK schedule can never be a system mode right? This is not allowed according to https://www.home-assistant.io/integrations/climate.mqtt/#modes

drzony:
"Schedule" in terms of the thermostats == "auto" in terms of HA.
Yes this can be merged it works, I just translate "schedule on"/"schedule off" into "auto"/"heat".
The issue is with other TuYa based thermostats which have no "schedule on"/"schedule off", but they have "schedule" in their list of presets. So the logic there would be:
From zigbee:
```
if preset_in == 'schedule'
  return system_mode = 'auto' to HA
else
  return preset = preset_in to HA
```
toZigbee:
```
if system_mode == 'auto'
  send preset = 'schedule' to device
else
  send system_mode to device or ignore it (this logic needs some thinking)
```

Koenkk:
Thanks for the explanation!

PR #1870, opened 2020-12-07
Added support for xfinity uei xhk1 ue

dettrick:


dettrick:
Added support for XHK1-UE Zigbee Keypad

Issue #1587, opened 2020-09-24
ZB-ONOFFPlug-D0005 reports 0 for current and voltage


magnayn:
The ZB-ONOFFPlug-D0005 (SmartThings UK plug) seems to report 0 for both current and voltage:

```
zigbee2mqtt/SmartPlug_1/power 57
zigbee2mqtt/SmartPlug_1/current 0
zigbee2mqtt/SmartPlug_1/voltage 0
zigbee2mqtt/SmartPlug_1/linkquality 49
zigbee2mqtt/SmartPlug_1/state ON
zigbee2mqtt/SmartPlug_1/energy 288.48
```

@mike182uk fixed the power/energy readings in 1d5f6669 (thanks!); I don't know if this is just a defect in the plugs I have not reporting the right data, or in fz.electrical_measurement_power not being right or something else.

The configure seems to include these items though.

Some logs:
[ziglog.txt](https://github.com/Koenkk/zigbee-herdsman-converters/files/5276942/ziglog.txt)

Happy to try to fix but I might need some help in next steps..

magnayn:
Don't believe I can remove labels..

Koenkk:
Looks the device not support current and voltage but because I don't have the device I cannot confirm this. @mike182uk  is this supported?

PR #1872, opened 2020-12-07
Ignore voltage and current values for ZB-ONOFFPlug-D0005

pedrolamas:
The SmartThings power plug 2019 (GP-WOU019BBDWG) seems to only monitor the power and energy values, so we can ignore voltage and current as they never change from 0.

This mitigates https://github.com/Koenkk/zigbee2mqtt/issues/5198 and https://github.com/Koenkk/zigbee-herdsman-converters/issues/1587


Koenkk:
Thanks!

PR #1874, opened 2020-12-07
Lumi.plug.mmeu01.part two improvement

PeterVoronov:
Extending functionality for lumi.plug.mmeu01, i.e. ZNCZ04LM:
New states:
  auto_off - auto off state, to switch the plug off in case of low level of consumption (< 2.6 W);
  consumer_connected - show the status of connection of the consume device(s) to the plug;
  led_disabled - show the state of the led disable at night feature;
  power_outage_memory - show the state of power outage memory feature;
New switches:
  auto_off - auto off feature, to switch the plug off in case of low level of consumption (<2.6 W);
  led_disabled - feature to set led disabled at night;
Changes to be committed:
	modified:   converters/fromZigbee.js
	modified:   converters/toZigbee.js
	modified:   devices.js
	modified:   lib/exposes.js


PeterVoronov:
Hello,
there is my second improvement for this plugs ...
Most useful things:
Consumer connected state and possibility to check, is power outage memory is set on or off.

lib/exposes.js
```diff
@@ -303,12 +303,14 @@ module.exports = {
     presets: {
         action: (values) => new Enum('action', access.STATE, values).withDescription('Triggered action (e.g. a button click)'),
         aqi: () => new Numeric('aqi', access.STATE).withDescription('Air quality index'),
+        auto_off: () => new Binary('auto_off', access.STATE, true, false).withDescription('Indicates whether the device has auto off with low consumer load functionality enabled or not'),
```
Koenkk, line 306: access should be `access.STATE_SET` here (since you can set it via toZigbee), same goes for `power_outage_memory` and `led_disabled`.

lib/exposes.js
```diff
@@ -325,6 +327,7 @@ module.exports = {
         illuminance: () => new Numeric('illuminance', access.STATE).withDescription('Raw measured illuminance'),
         illuminance_lux: () => new Numeric('illuminance_lux', access.STATE).withUnit('lx').withDescription('Measured illuminance in lux'),
         keypad_lockout: () => new Lock().withState('keypad_lockout', '1', '0', 'Enables/disables physical input on the device'),
+        led_disabled: () => new Binary('led_disabled', access.STATE, true, false).withDescription('Indicates whether the device has LED disable at night state or not'),
```
Koenkk, line 330: I think we should rename this to `led_disabled_night` because `led_disabled` makes me think the led is always disabled.

converters/fromZigbee.js
```diff
@@ -1643,20 +1643,62 @@ const converters = {
...
+                        payload.current = precisionRound(value * 0.001, 4);
+                        break;
+                    case 152:
+                        // power
+                        payload.power = precisionRound(value, 2);
+                        break;
+                    default:
+                        // if (meta.logger) meta.logger.debug(`plug.mmeu01: unknown index $(index) with value ${value}`);
+                    }
+                    if (meta.logger) meta.logger.debug(`plug.mmeu01: recorded index ${index} with value ${value}`);
```
Koenkk, line 1674: I don't think we should always log here. maybe only in default case?

Koenkk: Same goes for all logging below e.g. ` if (meta.logger) meta.logger.debug(`plug.mmeu01: attr. 513 (pwr_out_mem) has value ${msg.data['513']}`);`

converters/fromZigbee.js
```diff
@@ -1643,20 +1643,62 @@ const converters = {
...
+                payload.led_disabled = msg.data['515'] === 1;
+            }
+            if (msg.data.hasOwnProperty('519')) {
+                // 0x0207
+                if (meta.logger) meta.logger.debug(`plug.mmeu01: attr. 519 (cons_conn) value ${msg.data['519']}`);
+                payload.consumer_connected = msg.data['519'] === 1;
+            }
+            if (msg.data.hasOwnProperty('523')) {
+                // 0x020B
+                if (meta.logger) meta.logger.debug(`plug.mmeu01: attribute 523 (max_load) has value ${msg.data['523']}`);
```
Koenkk, line 1699: Nothing is set in the payload here.

converters/fromZigbee.js
```diff
@@ -1643,20 +1643,62 @@ const converters = {
                     default:
                         // if (meta.logger) meta.logger.debug(`plug.mmeu01: unknown vtype=${data[i+1]}, pos=${i+1}`);
                     }
-                    payload[index] = value;
-                    // if (meta.logger) meta.logger.debug(`plug.mmeu01: recorded index ${index} with value ${value}`);
+                    switch (index) {
```
Koenkk, line 1646: I think we should compact this code with if else. e.g.
```js
if (index === 3) payload.temperature = calibrateAndPrecisionRoundOptions(value, options, 'temperature');
else if (index === 100) payload.state = value === 1 ? 'ON' : 'OFF';
etc...
```

Sames goes for all if's here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/1874/files#diff-b676ab0df4d49b8572967cb4c92a2dc6264091131f6b1828ca13d0f9656e367aR1677

Koenkk:
Put some comments, but would like to thank you a lot, personally I'm very interested in the auto_off function as I have my kettle connected to this device.

converters/fromZigbee.js
PeterVoronov: It's for detecting, but ok, will add

converters/fromZigbee.js
PeterVoronov: Ok

lib/exposes.js
PeterVoronov: Accepted

lib/exposes.js
PeterVoronov: Missed it, will fix

PeterVoronov: Correct me, if I'm wrong, but I understand this way:
STATE_SET - only for set
STATE_GET - only for get
STATE - for both

PeterVoronov:
> Put some comments, but would like to thank you a lot, personally I'm very interested in the auto_off function as I have my kettle connected to this device.

Thanks and You are welcome ;-)

lib/exposes.js
PeterVoronov: And I have more strategical question:
Should be such states not binary, but like a switch - literally "on/off' ?

converters/fromZigbee.js
PeterVoronov: Not see a big difference, but - as you wish ;-)

PeterVoronov:
Done

PeterVoronov:
finally done ;-)
At least it all working for me ...

Koenkk:
big thanks!

Koenkk:
@PeterVoronov how does the auto_off work? I've tested with a kettle but the plug does not automatically turn off once the water is cooked (checked that the power is 0). The power outage works correctly

PeterVoronov:
> @PeterVoronov how does the auto_off work? I've tested with a kettle but the plug does not automatically turn off once the water is cooked (checked that the power is 0). The power outage works correctly

In my case it switched off after 20 minutes after I set the auto_off on a system without load.
[There](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Xiaomi-manufacturer-specific-clusters,-attributes-and-attribute-reporting#cluster-fcc0-lumi) is listed a parameter for the power threshold , but i didn't saw a parameter for the timeout.

PR #1882, opened 2020-12-09
Support for Useelink SM-SO306E/K/M

nikosillo:
Add this device
https://es.aliexpress.com/item/4000981707898.html
It's the same model as the Useelink SM-SO306E but the usb enpoint changes from 7 to 5 (7 does not work) and the manufacturerName is added: _TZ3000_o005nuxx


nikosillo (commented):
Correct

nikosillo (commented):
3

Koenkk:
Thanks!

PR #1885, opened 2020-12-09
Added another Heiman door sensor clone

stasheq:
Works fine after this simple addition.
Sold under "Ferguson" brand:
http://ferguson-digital.eu/en/smart-home-ferguson/smart-home-multipurpose-sensor-fs1mp/


Koenkk:
Thanks!

PR #1889, opened 2020-12-11
Add support for LEDVANCE AC25697

t-bourne:
add support for "LEDVANCE SMART+ CLASSIC E27 RGBW" light.

model: AC25697
product code: 4058075208391 (LEDVANCE)

See: https://www.ledvance.com/consumer/products/smart-home/smart-home-products-with-zigbee-technology/smart-home-lamps/classic-lamps-with-zigbee-technology/smart-classic-multicolour/index.jsp

I need some feedback on the "model" tag for this manufacture: In Ledvance section only product codes are used as "model", in OSRAM section the label written on the lamp is used (e.g. "AC10787").  Which one should be used/ is preferred?

Both npm tests mentioned in Readme are passed. Light is working in my setup with homeassistant (on/off, brightness, TW, colors).

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -5143,7 +5143,7 @@ const devices = [
         zigbeeModel: ['A60 RGBW Value II'],
         model: 'AC25697',
         vendor: 'LEDVANCE',
-        description: 'SMART+ CLASSIC E27 RGBW',
+        description: 'SMART+ classic E27 RGBW',
         extend: ledvance.light_onoff_brightness_colortemp_colorhs,
         ota: ota.ledvance,
     },
--- a/devices.js
+++ b/devices.js
@@ -5139,6 +5139,14 @@ const devices = [
         extend: ledvance.light_onoff_brightness_colortemp_colorxy,
         ota: ota.ledvance,
     },
+    {
+        zigbeeModel: ['A60 RGBW Value II'],
+        model: 'AC25697',
+        vendor: 'LEDVANCE',
+        description: 'SMART+ CLASSIC E27 RGBW',
+        extend: ledvance.light_onoff_brightness_colortemp_colorhs,
+        ota: ota.ledvance,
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -5139,6 +5139,14 @@ const devices = [
         extend: ledvance.light_onoff_brightness_colortemp_colorxy,
         ota: ota.ledvance,
     },
+    {
+        zigbeeModel: ['A60 RGBW Value II'],
+        model: 'AC25697',
+        vendor: 'LEDVANCE',
+        description: 'SMART+ classic E27 RGBW',
+        extend: ledvance.light_onoff_brightness_colortemp_colorhs,
```
Koenkk, line 5147: Why did you use `colorhs` here and not `colorxy` (which is preferred)?

Koenkk:
Model looks good, most important thing is that you can find the device with google when searching on it.

devices.js
t-bourne: Simply due to ignorance... I'll change it and notify you (resolve conversation) if done.
Unfortunatelly I can (re-) test the correct functionality soonest mid of january... (If you're sure that it will no interfere with the current code, we could change it on the fly without testing).

Sorry for the delayed reply, stress at work and christmas preparations (merry xmas btw!)....

t-bourne:
> 
> 
> Model looks good, most important thing is that you can find the device with google when searching on it.

Hm... you can find the lamp either on the website or the official zigbee "supported devices" homepage.
The "model" (AC...) will lead you to Zigbee page (without a photograph) and the "product code" will lead you to Ledvance website (with photograph). The product code is not listed on the device, but the other hints on the device will lead to the LEDVANCE website.

I would go with the model "zigbee description" as the other information on the lamp will lead to you the LEDVANCE website.

PR #1891, opened 2020-12-11
Add Support für EcoDim ED-10015

volkerrichert:
and fix ED-10011 & ED-10012. Try are mixed up... 10012 is the 4 button switch


Koenkk:
Thanks!

PR #1898, opened 2020-12-13
Small improvements to scene_add/scene_remove

sjorge:
- scene_remove should wait for response before updating meta
- scene_add should try and remove scene with provided I before adding

See koenkk/zigbee2mqtt#5268 if we do not remove the scene first and we do 2 or more scene_add calls for the same ID, there are unexpected results where the previous payload is not discarded.

```
DEV="0xec1bbdfffeaf89de"
CP="color_temp"
CPV=("250" "454")
#CP="color"
#CPV=("{\"x\":0.3818,\"y\":0.3797}" "{\"x\":0.5018,\"y\":0.4153}")

set -x
mosquitto_pub -t zigbee2mqtt/${DEV}/set  -m "{\"scene_remove\": 22}"
sleep 3
mosquitto_pub -t zigbee2mqtt/${DEV}/set  -m "{\"scene_add\": {\"ID\": 22, \"state\": \"ON\", \"${CP}\": ${CPV[1]}, \"brightness\": 254}}"
sleep 3
mosquitto_pub -t zigbee2mqtt/${DEV}/set  -m "{\"scene_recall\": 22}"
sleep 3
mosquitto_pub -t zigbee2mqtt/${DEV}/set  -m "{\"${CP}\": ${CPV[0]}}"
sleep 3
mosquitto_pub -t zigbee2mqtt/${DEV}/set  -m "{\"scene_add\": {\"ID\": 22, \"state\": \"ON\", \"brightness\": 32}}"
sleep 3
mosquitto_pub -t zigbee2mqtt/${DEV}/set  -m "{\"scene_recall\": 22}"
```

Now works as expected, the final state of the bulb is on at ~12% brightness with a ct of 250, before this change the ct would have been 454, which was not the expected result.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3806,7 +3806,9 @@ const converters = {
         convertSet: async (entity, key, value, meta) => {
             const groupid = entity.constructor.name === 'Group' ? entity.groupID : 0;
             const sceneid = value;
-            await entity.command('genScenes', 'remove', {groupid, sceneid}, getOptions(meta.mapped));
+            const response = await entity.command(
+                'genScenes', 'remove', {groupid, sceneid}, getOptions(meta.mapped),
+            );
 
             const isGroup = entity.constructor.name === 'Group';
             const metaKey = `${sceneid}_${groupid}`;
@@ -3819,11 +3821,13 @@ const converters = {
                         }
                     }
                 }
-            } else {
+            } else if (response.status === 0) {
                 if (entity.meta.scenes && entity.meta.scenes.hasOwnProperty(metaKey)) {
                     delete entity.meta.scenes[metaKey];
                     entity.save();
                 }
+            } else {
+                throw new Error(`Scene remove not succesfull ('${common.zclStatus[response.status]}')`);
             }
         },
     },
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3782,20 +3782,37 @@ const converters = {
                 }
             }
 
-            const response = await entity.command(
-                'genScenes', 'add', {groupid, sceneid, scenename, transtime, extensionfieldsets}, getOptions(meta.mapped),
+            /*
+             * Remove scene first
+             *
+             * Multiple add scene calls will result in the current and previous
+             * payloads to be merged. Resulting in unexpected behavior when
+             * trying to replace a scene.
+             *
+             * We accept a SUCESS or NOT_FOUND as a result of the remove call.
+             */
+            const removeresp = await entity.command(
+                'genScenes', 'remove', {groupid, sceneid}, getOptions(meta.mapped),
             );
 
-            if (isGroup) {
-                if (meta.membersState) {
-                    for (const member of entity.members) {
-                        saveSceneState(member, sceneid, groupid, state);
+            if (isGroup || (removeresp.status === 0 || removeresp.status == 139)) {
+                const response = await entity.command(
+                    'genScenes', 'add', {groupid, sceneid, scenename, transtime, extensionfieldsets}, getOptions(meta.mapped),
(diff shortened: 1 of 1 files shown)


sjorge:
Further improvements are possible as mentioned in [here](https://github.com/Koenkk/zigbee2mqtt/issues/5268#issuecomment-743984522)

Koenkk:
Thanks looks good.

When a command is send to a group you will never get a response so this is expected (also the individual devices don't respond in this case).

sjorge:
@Koenkk found some odd hard to reproduce on demand but common issue when using groups :(

With group of 3+ bulbs 1 or 2 don’t always get the new group.

I think the remove and add happen to fast. Would adding a small delay when using groups be acceptable? It seems to help, I was testing with a 1000ms delay, 500ms also seems to improve it, but still had one this morning at 500ms... thoughts? (Only happens with groups, with a single bulb I found no issues)

sjorge:
Been doing more testing... this is annoying the delay required seems to depend on how busy the network is.
But it looks like they messages get properly queued up though!

The actual issues seems to do doing a recall to soon after, it happens when some bulbs have processed both remove and add, while others just the remove. The recall delay seems to be between 300ms-800ms depending on load :(

So not sure this is something we should handle in z2m or not... I'm leaning on no.

sjorge:
OK, spend another hour on so on this.

And I have some questions...

@koenkk How are mqtt set messages handler? Sequentially or parallel?

Doing the remove and add call within scene_add seems to work, even for groups where we don't get a reply.
However once your throw in recall... this is were things go bad occasionally. Depending on how busy it is, if i do the recall after (2 + (1 * # bulbs)) seconds it always seems fine. But this is a long time to wait. The faster I do, the more chances of it going wrong with big groups of bulbs. When it goes wrong it is either:

- one or more bulbs don't change (Assumption is  after remove but before new add)
- one or more bulbs hit the old scene (Assumption before remove)
- one or more bulbs don't do anything (Network too busy?)

As I believe the group scene commands use broadcast instead of unicast, which puts some more load on the mesh.

That only makes sense though the z2m does things in parallel though, if it does it purely sequential the recall should always happen after the scene add.

If we are doing things in parallel, we might need some sort of exception to scene stuff targeting the same group/device. To ensure they are done in sequence.

e.g. scene_add, scene_recall in the pipeline should be executed in that order.

Koenkk:
@sjorge messages are handled sequentially in Zigbee2MQTT but this is no guarantee that the message arrive at the devices in this order when using groups. This is because we don't know when a device received a message (so you never know when to send the next one so the device processes them in order).

What is the use case for calling recall so fast after the add? In my opinion it does not look like a problem when you have to wait a few seconds after this.

sjorge:
I'm (ab)using scene_add + recall to change bulbs while they are off, to do adaptive ligthing.

Also to the off -> on transition when there is more than just a state. e.g. homekit allows you to do stuff like:
turn living room red, which results in on + color=red, if you do scene_add + recall, the bulbs turns on as red. 
If you don't use a scene the bulbs either turns on to previous color, then flashes to red or just turns on at the previous color. Depending on the order you send stuff.

The unpredictability here is kind of annoying, For the adaptive lighting i now use a 5 second between the add and recall.
I also only use scene_add+recall when the bulb is off and the new state = on + a color. But here it's more tricky, I don't really want to wait 5 seconds or more. 1 is acceptable but that still fails 1/3 times.

Once we have scene_view I can probably switch to 3 fixed scenes for the adaptive lightning though. But still trying to work out who that would work in my head.

I have a rought idea that is basically the same as binding stuff, but again groups are a issue there. I need to think on that some more.

Basically scenes are awesome, unless groups are involved -_-

PR #1901, opened 2020-12-13
add PSMP5_00.00.03.15TC to devices

m1k1o:
I have Climax plug, that is identified as `PSMP5_00.00.03.15TC`. I have to modify devices everytime I want to use that plug, so it would be good to have it there by default.


Koenkk:
Thanks!

PR #1904, opened 2020-12-14
Add Yale YRD216-HA2-619 Assure Lock

rjbez17:
Adds support for the Yale YRD216-HA2-619 Push Button Assure Lock.

Tested using the 1.16.1 and 1.16.2 tagged containers.


Koenkk:
Thanks!

PR #1906, opened 2020-12-14
Fix TuYa TRV schedule reporting

drzony:
It turns out that high bits of schedule hours contain additional information.
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/5298


Koenkk:
Thanks again!

PR #1910, opened 2020-12-14
Adding support to Yale YRD216-HA2-619 lock

lsilva171:


Koenkk:
I've made some changes to device.js, can you rebase it?

PR #1912, opened 2020-12-15
On branch bht-002-gclzb-sensor-tuning

### LINKED ISSUE #1905: BHT-002-GCLZB sensor set and get uses different values ...
Hello, colleagues.

Looking in the code, i see, the FromZigbee the sensor values is returned as a string:
```
    case common.TuyaDataPoints.moesSensor:
        switch (value) {
        case 0:
            return {sensor: 'IN'};
        case 1:
            return {sensor: 'AL'};
        case 2:
            return {sensor: 'OU'};
        default:
            return {sensor: 'Not supported'};
        }
```
But in ToZigbee it uses a numerik values:
```
    // send an mqtt message to topic '/sensor' to change the temperature sensor setting - options [0=IN|1=AL|2=OU]
    moes_thermostat_sensor: {
        key: ['sensor'],
            await sendTuyaDataPointEnum(entity, common.TuyaDataPoints.moesSensor, value);
        },
    },
```
what is version is right?
I'm ready to prepare a PR, but what it better ? 
By my understanding - it have to be identical type of values for in and out

PeterVoronov:
Added possibility to process string values 'IN','AL' and 'OU' in parallel
with the numeric ones 0,1,2 in toZigbee to be comply with the data returned
by fromZigbee. (see the [issue](https://github.com/Koenkk/zigbee-herdsman-converters/issues/1905) discussion)

Changes to be committed:
	modified:   converters/toZigbee.js

Fixes #1905

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3123,13 +3123,13 @@ const converters = {
         key: ['sensor'],
         convertSet: async (entity, key, value, meta) => {
             if (typeof value === 'string') {
-                if (value === 'IN') value = 0;
-                else if (value === 'AL') value = 1;
-                else if (value === 'OU') value = 2;
-                else value = -1;
+                const lookup = {'IN': 0, 'AL': 1, 'OU': 2};
+                value = lookup.hasOwnProperty(value) ? lookup[value] : -1;
             }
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1916, opened 2020-12-15
add typescript definitions

### LINKED ISSUE #1915: Add typescript support
Hi there! I would like to build on top of zigbee2mqtt which is using internally objects retrieved by zigbee-herdsman (which has types) and zigbee-herdsman-converters (which does not have types).

I couldn't find types in the @types repository and also not in this repo.
Is it possible to add a module .d.ts file? Or would you prefer them in the @types repo?

I am happy to help if needed.

Thanks!

MartinMuzatko:
fixes #1915 

any input is welcome

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,78 @@
+export enum ZigbeeDeviceAccess {
+    STATE = 1,
+    SET = 2,
+    STATE_SET = 3,
+    STATE_GET = 5,
+    ALL = 7,
+}
+
+export interface ZigbeeDeviceExposeGeneric {
+    access: ZigbeeDeviceAccess
+    description: string
+    name: string
+    property: string
+    type: string
+}
+
+export interface ZigbeeDeviceExposeEnum {
+    type: 'enum'
+    values: string[]
+}
+
+export interface ZigbeeDeviceExposeNumeric {
+    type: 'numeric'
+    value_min: number
+    value_max: number
+}
+
+export interface ZigbeeDeviceExposeBinary {
+    type: 'binary'
+    value_off: boolean
+    value_on: boolean
+}
+
+export interface ZigbeeDeviceExposeText {
+    type: 'text'
+}
+
+export interface ZigbeeDeviceExposeLight {
+    type: 'light'
+    features: ZigbeeDeviceExpose[]
+}
+
+export interface ZigbeeDeviceExposeComposite {
+    type: 'composite'
+    features: ZigbeeDeviceExpose[]
+}
+
+export type ZigbeeDeviceExpose = ZigbeeDeviceExposeGeneric & (ZigbeeDeviceExposeEnum | ZigbeeDeviceExposeNumeric | ZigbeeDeviceExposeBinary | ZigbeeDeviceExposeText | ZigbeeDeviceExposeLight | ZigbeeDeviceExposeComposite)
+
+export interface ZigbeeDevice {
+    date_code: string
+    definition?: {
+        description: string
+        exposes: ZigbeeDeviceExpose[]
+        model: string
+        vendor: string
+    }
+    endpoints: Record<number, { bindings: string[], clusters: {} }>
+    friendly_name: string
+    ieee_address: string
+    interview_completed: boolean
+    interviewing: boolean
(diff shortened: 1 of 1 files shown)


Koenkk:
Looks good. @nurikk maybe you also want to check

index.d.ts
```diff
@@ -0,0 +1,78 @@
+export enum ZigbeeDeviceAccess {
+    STATE = 1,
```
nurikk, line 2: Since access property is just a bitmask it should be refactored into something like this (and this definition is explicit btw):

```typescript
export enum ZigbeeDeviceAccess {
    NONE = 0b000,
    STATE = 0b001,
    WRITE = 0b010,
    READ = 0b100,
}
```

for example in current proposal: ZigbeeDeviceAccess.STATE_SET = 3 which is just `0b001 | 0b010 ` but it should look like this:
`ZigbeeDeviceAccess.STATE | ZigbeeDeviceAccess.WRITE`

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+    SET = 2,
+    STATE_SET = 3,
+    STATE_GET = 5,
+    ALL = 7,
+}
+
+export interface ZigbeeDeviceExposeGeneric {
+    access: ZigbeeDeviceAccess
+    description: string
```
nurikk, line 11: description might be missing `description?: string`

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+    name: string
+    property: string
+    type: string
+}
+
+export interface ZigbeeDeviceExposeEnum {
```
nurikk, line 17: I have my own implementation, you might want to have a look
https://github.com/nurikk/z2m-frontend/blob/a341b4dd5704d645c11845deb2213915119a0e03/src/types.ts#L100

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export interface ZigbeeDeviceExposeComposite {
+    type: 'composite'
+    features: ZigbeeDeviceExpose[]
+}
+
+export type ZigbeeDeviceExpose = ZigbeeDeviceExposeGeneric & (ZigbeeDeviceExposeEnum | ZigbeeDeviceExposeNumeric | ZigbeeDeviceExposeBinary | ZigbeeDeviceExposeText | ZigbeeDeviceExposeLight | ZigbeeDeviceExposeComposite)
+
+export interface ZigbeeDevice {
```
nurikk, line 50: Device definition is slightly more complicated
https://github.com/nurikk/z2m-frontend/blob/a341b4dd5704d645c11845deb2213915119a0e03/src/types.ts#L195

nurikk (changes requested):

index.d.ts
MartinMuzatko: Makes sense. I'll add the change :)

MartinMuzatko: Is there any description we can give for what the individual access flags mean?
I can't really imagine what the difference is between `STATE` and `STAGE_GET`

nurikk: >I can't really imagine what the difference is between STATE and STAGE_GET
Yes, you've already noticed problem with this aliases :)

Check documentation about access property
https://github.com/Koenkk/zigbee2mqtt.io/blob/develop/docs/information/exposes.md#access

index.d.ts
MartinMuzatko: I thought about extending also. Makes it easier to grab the correct type from the definitions. 👍

index.d.ts
MartinMuzatko: Wouldn't it make sense to copy the entire definition then? Or move them to a shared library?

nurikk: Yes, having separate shared library will be nice and handy, I could start using it as well

itavero: For [my zigbee2mqtt homebridge plugin](https://github.com/itavero/homebridge-z2m), I've also declared a lot of these models in my own code (a bit less detailed though, as I don't need all the information in my plugin).
For me having a separate library that defines these interfaces (which, as far as I can tell, map to the data on MQTT as well), could also be very nice.

PR #1918, opened 2020-12-15
Fix Yale YRD216-HA2-619 entry

lsilva171:


devices.js
```diff
@@ -10722,6 +10722,23 @@ const devices = [
...
     },
+    {
+        zigbeeModel: ['YRD216 PBDB'],
+        model: 'YRD216-HA2-619',
+        vendor: 'Yale',
+        description: 'Real living keyless push button deadbolt lock',
+        supports: 'lock/unlock, battery',
+        fromZigbee: [fz.lock, fz.lock_operation_event, fz.battery],
+        toZigbee: [tz.generic_lock],
+        meta: {configureKey: 1, battery: {dontDividePercentage: true}},
```
Koenkk, line 10733: Don't decrease the configurekey (keep it 3)

lsilva171: This lock is basically the same as YRD210 which has a configurekey of 1.

So, the question now is should YRD210 have its configurekey change to 3 as well?

Koenkk: No: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L6

PR #1919, opened 2020-12-16
Ignore reserved bits in occupancy condition

vladimirnesterov:
The cluster attribute is described as follow:

4.8.2.2.1.1 Occupancy Attribute
The Occupancy attribute is a bitmap: map8 0b0000 000x RP - M
Bit 0 specifies the sensed occupancy as follows: 1 = occupied, 0 = unoccupied.
All other bits are reserved.

This change is proposed to ignore reserved bits that can be used by some vendors.


Koenkk:
Thanks!

PR #1922, opened 2020-12-16
added SSM-U02 Aqara T1 relay (no neutral)

kdeyko:
I don't know much about node.js and I'm not a programmer, I just have this relay and wanted to make it work with z2m :)
Well, it works just fine even with this short config.
Please let me know if I need to make more tests and/or add something to the code. Hope, with your assistance we can make this public.
Thanks!

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -1400,9 +1400,7 @@ const devices = [
         model: 'SSM-U02',
         vendor: 'Xiaomi',
         description: 'Aqara single switch module T1 (without neutral). Doesn\'t work as a router and doesn\'t support power meter',
-        fromZigbee: [fz.on_off],
-        exposes: [e.switch()],
-        toZigbee: [tz.on_off],
+        extend: generic.switch,
     },
 
     // TuYa
--- a/devices.js
+++ b/devices.js
@@ -1395,6 +1395,15 @@ const devices = [
             await configureReporting.deviceTemperature(endpoint);
         },
     },
+    {
+        zigbeeModel: ['lumi.switch.l0agl1'],
+        model: 'SSM-U02',
+        vendor: 'Xiaomi',
+        description: 'Aqara single switch module T1 (without neutral). Doesn\'t work as a router and doesn\'t support power meter',
+        fromZigbee: [fz.on_off],
+        exposes: [e.switch()],
+        toZigbee: [tz.on_off],
+    },
 
     // TuYa
     {


kdeyko:
Hmm, just noticed that the log is now full of such messages:
```
ERROR: Failed to call 'Receive' 'onZigbeeEvent' (TypeError: Cannot read property 'type' of undefined at /app/lib/extension/receive.js:117:42 at Array.filter (<anonymous>) at Receive.onZigbeeEvent (/app/lib/extension/receive.js:116:65) at Controller.callExtensionMethod (/app/lib/controller.js:384:44) at Controller.onZigbeeEvent (/app/lib/controller.js:267:14) at Zigbee.emit (events.js:314:20) at Controller.<anonymous> (/app/lib/zigbee.js:70:52) at Controller.emit (events.js:314:20) at Controller.<anonymous> (/app/node_modules/zigbee-herdsman/dist/controller/controller.js:548:22) at Generator.next (<anonymous>))
```
Though the device still works just fine.

@Koenkk
Could you please help me to debug this issue?

devices.js
```diff
@@ -1395,6 +1395,15 @@ const devices = [
             await configureReporting.deviceTemperature(endpoint);
         },
     },
+    {
+        zigbeeModel: ['lumi.switch.l0agl1'],
+        model: 'SSM-U02',
+        vendor: 'Xiaomi',
+        description: 'Aqara single switch module T1 (without neutral). Doesn\'t work as a router and doesn\'t support power meter',
+        fromZigbee: [fz.on_off],
```
Koenkk, line 1403: Try removing `fromZigbee`, `exposes` and `toZigbee` here and add `extend: generic.switch,`

kdeyko: Did it and absolutely nothing has changed: the same error in logs and the device is still working.

Koenkk: I'm quite sure the error is not in the config. Probably you are using some incompatible files (e.g. wrong version of zigbee-herdsman-converters with zigbee2mqtt). Try to fully sync with the dev branch and the error should be gone. (https://www.zigbee2mqtt.io/how_tos/how-to-switch-to-dev-branch.html)

kdeyko: Indeed, it works like a charm on dev branch. Thank you!

opa4h: Hi! I have such a relay. After your edits, everything worked except for passing the state. If I turn off the relay with a physical key, then the state in HA remains unchanged.

kdeyko: Checked it, and yes - my setup works the same way.
@Koenkk, do you have any idea on how to make it "see" the state changes when pressing the physical key?

Koenkk: Try changing the entry to:

```js
{
    zigbeeModel: ['lumi.switch.l0agl1'],
    model: 'SSM-U02',
    vendor: 'Xiaomi',
    description: 'Aqara single switch module T1 (without neutral). Doesn\'t work as a router and doesn\'t support power meter',
    extend: preset.switch,
    meta: {configureKey: 1},
        await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff']);
    },
},
```

kdeyko: I couldn't make it work with this config, though my test were not 100% correct probably, since I'm running z2m in docker on Synology. I tried:
- just altering the `devices.js` inside the container, then restart it --> container crashes
- doing as described [here](https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html#docker) --> Synology lacks the npm's dependencies
- the same as above, but building the stuff in official node docker container (tried buster and alpine), then moving the entire `zigbee-herdsman-converters` folder to the z2m container (as volume, of course) --> container crashes
- something similar I even can't remember but still - container crashes

@Koenkk what am I doing wrong?
@opa4h could you try the proposed solution?

opa4h: Checked it out! Now it works!

kdeyko: Great news! I’ll make a PR with this change.

Issue #1925, opened 2020-12-17
Aqara Motion Sensor action on illuminance


AlexxIT:
Zigbee network can skip messages. You can check this if you trace the sequence numbers of the messages in the raw zigbee data. With each message, the sequence number increases by one, but sometimes there are gaps.

Motion Sensor on each motion sends two messages: illuminance and motion. Always in that way.
https://www.zigbee2mqtt.io/devices/RTCGQ11LM.html

Sometimes only a message about illumination can reach, but a message about movement does not come.

I suggest activating motion when receiving illumination data.

It is also important to note that the illumination comes once every 55 minutes with the Basic cluster. There is no need to trigger the movement.

Koenkk:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/1926 is merged, closing this.

PR #1926, opened 2020-12-17
Add motion to illuminance data on Aqara Motion Sensor RTCGQ11LM #1925

AlexxIT:
Description in issue #1925


Koenkk:
Interesting solution, thanks!

Addresses: https://github.com/Koenkk/zigbee2mqtt/issues/2274

PR #1927, opened 2020-12-17
Fix IKEA Controller toggle button after refactor

drzony:
Toggle button was incorrectly placed in `genLevelCtrl' endpoint instead of 'genOnOff'


fnfid:
seem to solve the not working toggle button, so similar issue is closed in advance. Thanks a lot!!!!

Koenkk:
Thanks! :smile:

Issue #1908, opened 2020-12-14
Not a single lock actually exposing lock_state


itavero:
While working on itavero/homebridge-z2m#18, I noticed that none of the locks defined in `devices.js` is actually exposing the `lock_state` defined in `lib/exposes.js`.
Digging into the `fromZigbee` converters, it seems that the `lock_state` property _can_ be present if `fz.lock` is used.

Is there some way we can get the exposes for the locks that do actually provide this property?
(I mean, we can add it to all of them, but does someone actually now which locks do or do not provide this information.)

Koenkk:
It's safe to assume all locks using `fz.lock` expose a lock state. The `state` of these locks is actually based on the lock state, and only the `fz.lock` converter provides the `state`. Meaning if they wouldn't have a lock state they won't have a state, if this was the case I would expect a lot of complaints from users, which we didn't have :) 

Added it.

itavero:
@Koenkk Sorry to get back to you on this one. I didn't realize that `lock` is actually a type with features.
Shouldn't the `lock_state` just be a feature of the `lock` type, or what's the criteria for something to be a feature?

itavero:
I'll create a PR with the proposed changes.

PR #1930, opened 2020-12-17
Add lock_state to features of lock type in exposes information

itavero:
As mentioned in the comments of #1908, given that the `lock_state` is always present in a lock, I think it should be part of the lock type in the exposes information.


Koenkk:
I agree that this should be part of the features of a lock. Thanks!

PR #1931, opened 2020-12-18
Added support for Philips Hue Iris (Generation 4) Copper

SamiHuilla:
Added support for [Hue Iris gen 4 Copper limited edition](https://www.philips-hue.com/en-gb/p/hue-white-and-colour-ambiance-iris-copper-limited-edition/8719514264564). 'zigbeeModel' is the same as 'Material number (12NC)' in the product specifications: 929002376801. Each color of the 4th gen seem to have different third last digit in the specs, but I can only confirm this one.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -3680,6 +3680,15 @@ const devices = [
         ota: ota.zigbeeOTA,
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1933, opened 2020-12-18
Adjust Xiaomi device description and clearify silvercrest description

Skaronator:
The new Xiaomi Sensors which model id ends with 1**2**LM are actually called T1 sensors. The T1 series are Zigbee 3.0 Devices while the 1**1**LM devices are Zigbee 1.2.

Here are some sources:
* https://community.hubitat.com/t/new-aqara-t1-series-zigbee-3-0-devices-finally-available/49006
* https://zigbeealliance.org/zigbee_products/door-and-window-sensor-t1/
* https://zigbeealliance.org/zigbee_products/wireless-mini-switch-t1/
* https://zigbeealliance.org/zigbee_products/water-leak-sensor-t1/
* (And matching naming scheme here https://zigbee.blakadder.com/vendors.html)

The Silvercrest Power Strip description was a bit confusing for me. I though the _possible_ means that it MIGHT be possible to switch the USB ports. Now I think the possible means that the power strip is optional available with and without USB ports.

Both are not true. You cannot switch the USB Ports (I've checked the device internals :D) and you cannot get them without USB.

The naming is more similar with the blakadder.com website: https://zigbee.blakadder.com/Lidl_HG06338.html


Koenkk:
Thanks!

PR #1940, opened 2020-12-19
Fixed temperature exposes for DIYRuZ_Flower

xyzroe:
Added separate exposes for DS18B20 and BME280 temperatures. 
Sorted exposes by sensors.


Koenkk:
Thanks!

PR #1946, opened 2020-12-20
Support for Revolt Thermostatic Radiator NX-4911

Schuemi:
It seems to be a white labled GS361A-H04 from Siterwell


Koenkk:
Thanks!

PR #1950, opened 2020-12-20
Update devices.js

Vladimirov71:
Lonsonho TS130F copy converter from TS130F LoraTap SC400. Add new function. Its works!


Koenkk:
Thanks!

PR #1952, opened 2020-12-21
For bht-002-gclzb replacement the 'preset' key for toZigbee by 'preset_mode' to make it identical with fromZigbee

PeterVoronov:
For bht-002-gclzb replacement the 'preset' key for toZigbee by 'preset_mode' to make it identical with fromZigbee

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3129,7 +3129,7 @@ const converters = {
         },
     },
     moes_thermostat_mode: {
-        key: ['preset_mode'], // to make it identical with fromZigbee
+        key: ['preset', 'preset_mode'], // to make it identical with fromZigbee
         convertSet: async (entity, key, value, meta) => {
             const hold = value === 'hold' ? 0 : 1;
             const schedule = value === 'program' ? 0 : 1;
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3129,7 +3129,7 @@ const converters = {
         },
     },
     moes_thermostat_mode: {
-        key: ['preset'],
+        key: ['preset_mode'], // to make it identical with fromZigbee
         convertSet: async (entity, key, value, meta) => {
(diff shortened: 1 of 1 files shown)


converters/toZigbee.js
```diff
@@ -3129,7 +3129,7 @@ const converters = {
         },
     },
     moes_thermostat_mode: {
-        key: ['preset'],
+        key: ['preset_mode'], // to make it identical with fromZigbee
```
Koenkk, line 3132: please change this to `key: ['preset', 'preset_mode'],` to prevent a breaking change

PeterVoronov: > please change this to `key: ['preset', 'preset_mode'],` to prevent a breaking change

Make sense. Done

Koenkk:
Thanks!

Koenkk:
Thinking about this again, the preferred way is `preset` not `preset_mode`. So I've reverted this and add `preset`  to thefromZigbee converter.

sjorge:
Should we also do this for the eurotonic trv’s boost mode?

currently we hack around with system_mode, which... is not really great.

PeterVoronov:
> Thinking about this again, the preferred way is `preset` not `preset_mode`. So I've reverted this and add `preset` to the fromZigbee converter.

Looks better 


> Should we also do this for the eurotonic trv’s boost mode?
> 
> currently we hack around with system_mode, which... is not really great.

May be simple - "mode" ? Or status ?
`case tuya.dataPoints.state: // Thermostat on standby = OFF, running = ON
                return {system_mode: value ? 'heat' : 'off', mode: value ? 'heat' : 'off'};`

Koenkk:
`system_mode`  is heavily used already so I don't think we should change it. 

@sjorge 

> Should we also do this for the eurotonic trv’s boost mode?

Don't know the details about this but sounds good

PR #1979, opened 2020-12-26
Adds: Variant of YALE YRD226 Lock to converters

Jhonattan-Souza:
Here it is the print of the model. Tested and working.

image


Koenkk:
Thanks!

PR #1983, opened 2020-12-27
System modes based on "force" attribute

mgrom:
I've made System Modes based on Force attribute. So we have Auto, Heat and Off modes. Presets are working only with Auto mode. Only for old Moes devices this won't work because they don't have force attrribute.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2803,8 +2803,7 @@ const converters = {
             case tuya.dataPoints.fanMode:
                 return {fan_mode: tuya.fanModes[value]};
             case tuya.dataPoints.forceMode: // force mode 0 - normal, 1 - open, 2 - close
-                return {system_mode: {0: 'auto', 1: 'heat', 2: 'off'}[value]};
-                // return {force: tuya.thermostatForceMode[value]};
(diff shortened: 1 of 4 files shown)


Koenkk:
Can you provide some more context what problem is addressed here?

mgrom:
It's not a problem, but improvement. Moes TRV uaes

mgrom:
It's not a problem but improvement.
Moes TRV is using presets currently, and this is ok. New version of this valve provides attribute "force". This attribute says: normal => use presets (and use PID algorithm), open => just do full open of valve (not PID), close => full close (not PID). So basically it's auto, heat and off modes.
So I've made this mapping in thus PR

Koenkk:
I see, thanks!

Xites:
The documentation is still with 'force'.
https://www.zigbee2mqtt.io/devices/TS0601_thermostat.html#tuya-ts0601_thermostat

mgrom:
@Xites let's keep it this way. Force will stay with us (sounds like line from Star Wars :D ). There are going to be some changes in Modes and Presets, because current behaviour is not inline with HA.

Xites:
@mgrom Hmm, finding a solution for this issue costed me an hour. I checked the documentation and I thought I was wrong.
A note in the docs would have been nice.

PR #1984, opened 2020-12-27
Add dimmer configuration for ubisys D1(-R)

guydriesen:


guydriesen:
Relates to https://github.com/Koenkk/zigbee-herdsman/pull/277 & https://github.com/Koenkk/zigbee2mqtt.io/pull/528

converters/toZigbee.js
```diff
@@ -2528,6 +2528,32 @@ const converters = {
             ], manufacturerOptions.ubisys));
         },
     },
+    ubisys_dimmer_setup: {
+        key: ['ubisys_dimmer_setup'],
```
Koenkk, line 2532: I propose to chnage this to `ubisys_dimmer_mode` so we don't have a nested object. This allows it to add it to exposes, (as enum, see example here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/38752abc80c9af8459d2a96977fc20b92009450a/devices.js#L735)

converters/toZigbee.js
```diff
@@ -2528,6 +2528,32 @@ const converters = {
...
+            }
+            // re-read effective settings and dump them to the log
+            converters.ubisys_dimmer_setup.convertGet(entity, key, meta);
+        },
+        convertGet: async (entity, key, meta) => {
+            const log = (dataRead) => {
+                meta.logger.warn(
+                    `ubisys: Dimmer setup read for '${meta.options.friendlyName}': ${JSON.stringify(utils.toSnakeCase(dataRead))}`);
+            };
+            log(await entity.read('manuSpecificUbisysDimmerSetup', ['capabilities']));
```
Koenkk, line 2552: Instead of logging, just read here and add a fromZigbee converter to translate it to something user friendly. (https://github.com/Koenkk/zigbee2mqtt.io/pull/528/files#r549327509)

converters/toZigbee.js
guydriesen: Good idea, will try it (new to javascript ;-))

guydriesen:
Added convertor for the different attributes and removed the nested object to support the exposes. Works nicely in the ui.

guydriesen:
Not sure if this the best option to have the dimmer phase control mode available for configuration in exposes. This is basically a one time setup, so it might be better to have it in the devices.yaml file. That's also why it was initially coded like the `ballast_config`.
What do you think? And can you give me some clues how to implement?
Thanks

Koenkk:
>  so it might be better to have it in the devices.yaml file.

I don't agree on this. These options are stored on the device side, not on the Z2M side. Allowing these options to be put in devices.yaml would add a lot of code on the Z2M side to sync the current selected options (there is also currently no mechanism for this).

PR looks good to me, let me know if it can be merged.

guydriesen:
Ok, I see. I was wondering if there's some sort of device configuration available to place configuration options like this. On/Off, brightness,... are used on a daily basis, but things like the dimmer setup, ballast configuration are typically only configured during installation and it might not be bad to prevent accidental change (in case of the dimmer setup this could even destroy the device!).

If it's fine for you, it can be merged.
Thanks for the support!

Koenkk:
Thanks!

PR #1987, opened 2020-12-28
Innr RS 230 C supports hue/sat

sjorge:
Recently got an Innr RS 230 C to play with, the spot also support Hue/Sat.

I guess this might be the case for all color bulbs from Innr, but I don't own any strips.


Koenkk:
Thanks!

PR #1989, opened 2020-12-28
Add support for BTicino L441C/N4411C/NT4411C

MidnessX:
This device is basically identical to the Legrand 067771 (only changing the zigbeeModel value is sufficient for it to work).


Koenkk:
Thanks!

PR #1990, opened 2020-12-28
removed hardcoded heat mode

mgrom:
small fix to  #1983 
There was Heat mode hardcoded, and in some situations it was being sent to MQTT


Koenkk:
Thanks!

PR #1991, opened 2020-12-28
Use consistent naming of Centralite devices

Skaronator:
I'm not sure if this vendor property is important for the device to work properly but the different case types break the Table on the documenation page.

image

A quick google search suggest that Centralite is the correct case and not CentraLite.


Koenkk:
Thanks!

PR #1996, opened 2020-12-29
HEIMAN HS2CM-N-DC curtain new device id

alparslana:
HS2CM-N-DC curtain motor is now also announcing as CurtainMo-EF-3.0.


Koenkk:
Thanks!

PR #1997, opened 2020-12-29
Support action and setup default binds for Ubisys D1

sjorge:
As per koenkk/zigbee2mqtt#5428

The actions work, I also added the onEvent as per the D1's tech reference.


Koenkk:
Thanks!

PR #1999, opened 2020-12-30
Fix empty position value of the tuya curtain

d3m3vilurr:
after change the position, TS0601 curtain returns 4 bytes with dp3
some cases this data has extra value in the second byte like;

`00 64 00 1b`

a last byte is position value, it would be safe with `& 0xFF` operation


Koenkk:
Thanks!

PR #2003, opened 2020-12-31
Support for Sinope RM3250ZB 50 amps load controller

cyber289:
Add support for Sinope RM3250ZB 50 amps load controller.

This has been tested successfully with the device.


Koenkk:
Thanks!

PR #2005, opened 2021-01-01
Typo in spelling of triple in action enum values

itavero:
The correct spelling is "triple", not "tripple".
Unfortunately, it seems the following converters have a typo in them:
 - `ptvo_multistate_action`
 - `qlwz_letv8key_switch`

I decided to only update the exposes information and *not* fix the typo in the converters, as this might break existing automations.
The exposes information should now contain the same spelling as the converter uses.


Koenkk:
Thanks!

PR #2006, opened 2021-01-01
Ikea E1743 loses bind/reporting after powerloss

sjorge:
This should fix #5477

Unfortunately I had issues doing the remind manually/via the frontend so I ended up repairing the device instead. But it was confirmed by @MattWestb that a rebind was also needed in addition to a re-reporting setup.

I wonder if we should also do this for:
- E1812: shortcut button
- E1524/E1810: 5 button puck

### DIFF
--- a/lib/ikea.js
+++ b/lib/ikea.js
@@ -2,7 +2,7 @@ async function bulbOnEvent(type, data, device) {
     /**
      * IKEA bulbs lose their configured reportings when losing power.
      * A deviceAnnounce indicates they are powered on again.
-     * Reconfigure the configured reoprting here.
+     * Reconfigure the configured reporting here.
      * NOTE: binds are not lost so rebinding is not needed!
      */
     if (type === 'deviceAnnounce') {
@@ -17,6 +17,34 @@ async function bulbOnEvent(type, data, device) {
     }
 }
 
+async function batteryDeviceOnEvent(type, data, device) {
+    /**
+     * IKEA battery deivces sometimes lose the bind and
+     * reporting configuration after replacing the battery.
+     * A deviceAnnounce indicates they are powered on again.
+     * Reconfigure the bind and reporting here.
+     */
+    if (type === 'deviceAnnounce') {
+        for (const endpoint of device.endpoints) {
+            const b = endpoint.binds.find((b) => b.cluster.name === 'genPowerCfg');
+            if (b) {
+                await endpoint.bind('genPowerCfg', b.target);
+            }
+
+            const c = endpoint.configuredReportings.find(
+                (c) => c.cluster.name === 'genPowerCfg' && c.attribute.name === 'batteryPercentageRemaining',
+            );
+            if (c) {
+                await endpoint.configureReporting(c.cluster.name, [{
+                    attribute: c.attribute.name, minimumReportInterval: c.minimumReportInterval,
+                    maximumReportInterval: c.maximumReportInterval, reportableChange: c.reportableChange,
+                }]);
+            }
+        }
+    }
+}
+
 module.exports = {
     bulbOnEvent,
+    batteryDeviceOnEvent,
 };
--- a/devices.js
+++ b/devices.js
@@ -2093,6 +2093,7 @@ const devices = [
             await reporting.batteryPercentageRemaining(endpoint);
         },
         ota: ota.tradfri,
+        onEvent: ikea.batteryDeviceOnEvent,
     },
     {
         zigbeeModel: ['TRADFRI transformer 10W', 'TRADFRI Driver 10W'],
@@ -2187,6 +2188,7 @@ const devices = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg']);
             await reporting.batteryPercentageRemaining(endpoint);
         },
+        onEvent: ikea.batteryDeviceOnEvent,
     },
     {
         zigbeeModel: ['TRADFRI on/off switch'],
@@ -2209,6 +2211,7 @@ const devices = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg']);
             await reporting.batteryPercentageRemaining(endpoint);
         },
+        onEvent: ikea.batteryDeviceOnEvent,
     },
     {
         zigbeeModel: ['TRADFRI SHORTCUT Button'],
@@ -2228,6 +2231,7 @@ const devices = [
             await reporting.bind(endpoint, defaultBindGroup, ['genPowerCfg']);
             await reporting.batteryPercentageRemaining(endpoint);
         },
+        onEvent: ikea.batteryDeviceOnEvent,
     },
     {
         zigbeeModel: ['SYMFONISK Sound Controller'],
@@ -2245,6 +2249,7 @@ const devices = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genLevelCtrl', 'genPowerCfg']);
             await reporting.batteryPercentageRemaining(endpoint);
         },
+        onEvent: ikea.batteryDeviceOnEvent,
     },
     {
         zigbeeModel: ['TRADFRI motion sensor'],
@@ -2263,6 +2268,7 @@ const devices = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg']);
             await reporting.batteryPercentageRemaining(endpoint);
         },
+        onEvent: ikea.batteryDeviceOnEvent,
     },
     {
         zigbeeModel: ['TRADFRI signal repeater'],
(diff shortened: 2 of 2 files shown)


sjorge:
OK, this code is throwing errors because it can't find coordinatorEndpoint.

Which seems to be passed into configure() and not in the onEvent.

MattWestb:
All IKEA controlling devices is more or less the same only different number of buttons also the motion sensors is the same but the old one is still ZLL but i think its more or lest the same with older zigbee stack. 
If getting generic problem then applying to all devices that is the same design = all IKEA controlling devices.

Known IKEA controlling devices: Motion Old and New, 5Btn Remote Old and New, Old round dimmer, 2 Btn remotes On/Off, Open/Coles, Knyplan, Symfonisk and the new scenes buton.

MattWestb:
All known OTA firmware released plus some not released for IKEA devices under [IKEA](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/OTA-Image-Types---Firmware-versions#ikea) with device ID in hex and dec.

MattWestb:
If you taking out the battery and clicking sometimes on the button so the power in the capacitors is emptying and the putting the battery back the device should sending one deviceAnnounce (if not loosing all config) for testing your code :-))

sjorge:
> If you taking out the battery and clicking sometimes on the button so the power in the capacitors is emptying and the putting the battery back the device should sending one deviceAnnounce (if not loosing all config) for testing your code :-))

Yeah that is what I've been doing, but looks like a console.log() does not show up on the console... but the main issue is that I am not sure how to get `coordinatorEndpoint`, this is passed in from z2m... and there is no zigbee object available to fetch the coordinator device from.

MattWestb:
Then one zigbee 3 or ZLL device is coming back into the network its sending on` Device_annce ZDO command` and if its rejoining one ZB3 network its also sending on` Mgmt_Permit_Joining_req ZDO command` to the coordinator (I think with 380 as payload if i remember right) so its possible detecting if the device is "resetted" / clan / lost config parameters.

lib/ikea.js
```diff
@@ -17,6 +19,23 @@ async function bulbOnEvent(type, data, device) {
...
+     * IKEA battery deivces sometimes lose the bind and
+     * reporting configuration after replacing the battery.
+     * A deviceAnnounce indicates they are powered on again.
+     * Reconfigure the bind and reporting here.
+     */
+    if (type === 'deviceAnnounce') {
+        // XXX: how do we get coordinatorEndpoint ?
+        const coordinatorEndpoint = ?; // this.zigbee.getDevicesByType('Coordinator')[0].getEndpoint(1)
+        const endpoint = device.getEndpoint(1);
+        await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg']);
```
Koenkk, line 33: Something like this should do it:
```js
for (const endpoint of device.endpoints) {
  const b = endpoint.binds.find(b => b.cluster.name === 'genPowerCfg');
  if (b) {
     await endpoint.bind('genPowerCfg', b.target);
  }

 const r = endpoint.configuredReportings.find(r => r.cluster.name === 'genPowerCfg' && r.attribute.name === 'batteryPercentageRemaining')
 if (r) {
    await endpoint.configureReporting(c.cluster.name, [{
            attribute: r.attribute.name, minimumReportInterval: r.minimumReportInterval,
            maximumReportInterval: r.maximumReportInterval, reportableChange: r.reportableChange,
        }]);
  }

}
```

sjorge: That seems to do the trick after making lint happy and undoing a lint change that broke it again.

sjorge:
Took a few tries, but seems to work now. Lets see if the device reports new battery tomorrow.
Last device that had a <100% battery so *fingers crossed*

sjorge:
Looking good!

```
Zigbee2MQTT:debug 2021-01-02 13:38:08: Received Zigbee message from 'office/light/mood/remote', type 'attributeReport', cluster 'genPowerCfg', data '{"batteryPercentageRemaining":87}' from endpoint 1 with groupID 0
Zigbee2MQTT:info  2021-01-02 13:38:10: MQTT publish: topic 'zigbee2mqtt/office/light/mood/remote', payload '{"battery":87,"linkquality":78,"update":{"state":"idle"}}'
```

The device just checked in! It wasn't a new battery but before it was in the 40% range. (I reused the battery I pulled out of the other device I was using for testing)

MattWestb:
Looks great !!!

I hope the attribute reporting and OTA checking is not killing the battery with the standard times setting. Its have being more or less wild discussions on deCONZ and ZHA about draining batteries on IKEA devices and some can being connected to some bugs in the EZSP Zigbee stack that is pretty nasty for the bettys and some because of the HW is being bad or bad quality batteries (I have returning one E1743 after the first lockdown the it have easting 3 batteries in 2 months).
Alos Tasmota have disabled attribute reporting then many users having problems with batteries on IKEA devices but it can being related the coordinator firmware (EFR32 second gen and most users is using older chips  = different CPU architecture  = different bugs).

Good coding done !!!!

sjorge:
We're not using the default though!

image

We're only allow a report if
- the last one was more than 1 hour ago
- there was at least a 1% change
- we do one ~1-2x day (max = 17h)

Koenkk:
Thanks!

sjorge:
@Koenkk I think we might need to revert this one.
I had 3 remotes die while being at 0% battery today, all ones that had fresh batteries installed during testing.

There is definitely something fucky with these after they have there battery removed for a bit... but rebinding/reporting setup seems to do more harm than good. They did all report once or twice after they had the battery swapped... but it seems odd all three are dead now.

MattWestb:
Thts mysterium its being larger and larger for every day !!!

Sure is one or more bugs in the Silabs zigbee stack that can doing bad things but its like raining snow and cant see wath is behind the large mess of dropping things from above.

Was the one that was not losing the group working with new battery and reporting ok and not draining the new batteries ?

I was thinking that some reporting is not being setting up OK and the device is trying sending attribute reports and its not working and the devices is trying in infit until have draining the battery completely.

Sure is its happens on al large zigbee systems also on IKEA GW but not so oft is my experience.

Do you like diving in some sniffing how IKEA is setting up the remote in their system and comparing how Z2M is doing it ??

One user have making shiffing with the IKEA GW is pairing remotes and bulbs that can giving the correct way to configuring the IKEA controlling devices https://github.com/Koenkk/zigbee2mqtt/issues/4414#issuecomment-699495669.

sjorge:
I don't own an IKEA GW.

All of them were reporting OK after the initial remove battery, put in new battery with the code introduce by this PR.

I made them leave and completely rejoined them last night, so hopefully now they stay OK.

MattWestb:
I knowing you dont have one but the https://github.com/Koenkk/zigbee2mqtt/issues/4414#issuecomment-699495669 have network key and looks being compleet sniffs of the pairing.

I think resetting and repairing can being one way around the problem but still not knowing the falling part if its some configuration that is bad or missing or if its "only" the bugs in the silabs zigbee stack.

Good that you was measure the batteries for verifying its being drained and not false reported values.

Koenkk:
> @Koenkk I think we might need to revert this one.
> I had 3 remotes die while being at 0% battery today, all ones that had fresh batteries installed during testing.
> 
> There is definitely something fucky with these after they have there battery removed for a bit... but rebinding/reporting setup seems to do more harm than good. They did all report once or twice after they had the battery swapped... but it seems odd all three are dead now.
> 
> My two other remotes are still working, not sure the battery status has update but those happen to be the 2 that had not had there batteries swapped because they were still 100% and still are now...
> 
> Edit:
> 
> * 2/3 had forgotten there binding to the group and had to have the back button pushed to get the to show up again
> * 1/3 still had it's group bind and it showed up with a fresh battery.
>   I check the last 2 batteries I pulled, they read 2.02 and 1.93 volts.
> 
> I rechecked the 3 others ones of the same blister that are left and they read 3.34, 3.29, 3.32... so I think the pack is not bad.

reverted

sjorge:
> reverted

👍

How do we feel about adding a additional note to the device page? Stating they may need to be repaired after replacing a battery for battery reporting to work?

Koenkk:
Sounds good (fits under de notes) can you make a PR?

sjorge:
Done https://github.com/Koenkk/zigbee2mqtt.io/pull/547

MattWestb:
For info new finding with IKEA and Samsung controlling devices can having problems with pull control configuration not being setted up OK and they draining batteries.
https://github.com/dresden-elektronik/deconz-rest-plugin/pull/4163#issue-552760697

And Tasmota and ZHA is having problem with one bug in silabs stack (in the coordinator) that is making it worse on Sonoff zigbee bridge that can making the same also with Philips HUE dimmer switches.

The last work around is reconfiguration of pull control and attribute reporting so the device is working as intended and can going in sleep until updated NCP firmware is made for fixing it.

I think the repairing is needed for all IKEA and samsung controller devices for getting the attribute reporting / pull control working OK after battery is being completely drained also in Z2M.

Issue #1974, opened 2020-12-25
Exposes using hue/Saturation and hue_correction configuration


asgothian:
I have been trying to implement the hue_correction feature in when controlling a lamp using the "expose" feature. This was done on this bulb (https://www.zigbee2mqtt.io/devices/9290012573A.html).

Running the tests, i found that the expose did not in fact announce that the lamp could be controlled via hue/saturation, and the hue_correction feature listed on the device page is used only when a lamp is controlled via hue/saturation.

I did patch in the expose on my system and was able to complete my implementation. When checking the exposes in devices.js i found that only a select few lamps actually offer the 'hs' expose, even if their page lists the 'hue_correction' feature, and as such indicates that the light should be controllable via hue/saturation

I wanted to ask if this is "by design" or "an error", before i go through and make a large PR adding this feature to the lamps which (given the documentation) should support it.

A.

Koenkk:
I think we should indeed expose hue/saturation for Hue bulbs, can you make a pr?

asgothian:
Will do. It will take me a few days to crosscheck the device description and the devices to ensure that exposes and fz.tz fit as planned.

A

itavero:
On a related note: if a device supports both `color_hs` and `color_xy`, which one would be the preferred way of controlling the bulb?
If I read this I would say it would be `color_hs`, but I think elsewhere I got the impression that `color_xy` would be preferred as this is supported by most lights.

Asking for itavero/homebridge-z2m#18

asgothian:
This question cannot be answered globally. It depends a lot on what should be done with the light. For link to color pickers and web interfaces or direct color controls, `color_xy` Is often easier. For situations where only the color should be changed but the color intensity and brightness should remain, control via `color_hs` Is easier. This was my reason for looking into this topic to begin with.

I would suggest to always support both if the light itself supports both. Going for only one is a limiting factor in my eyes. I will even go as far as suggesting to always support `color_hs` as incoming format, even on bulbs which internally only allow use of `color_xy`, but this might require some additional magic in the converters - so this is for the future.

A.

sjorge:
Also some bulb respond differently so keep that in mind.

- Innr: xy needs a odd 'redfix', but via Hue/Sat it works without overal I find the colors to be closer to what I expect with hs.
- Osram: bulb seems to become pink when using XY for pure red at 100% brightness, with hs it is a tiny bit dimmer but stays red.

itavero:
[...]

PR #2009, opened 2021-01-02
Hue/Saturation expose for Hue Lights

asgothian:
- added a predefined light_brightness_colorxyhs  exposes
- defined presets with _colorxyhs which use the respective exposes.
- extended the hue and ledvance presets to use the defined presets with _colorxyhs
- switched all hue lights which used the .._colorxy preset to .._colorxyhs preset.

Note: I was only able to test this with a few hue lamps. It may be possible that some lamps will not react when controlled via hue/sat

Discussion see issue #1974


Koenkk:
What is the reason not all `preset.hue.light_onoff_brightness_colorxy` have been migrated to `preset.hue.light_onoff_brightness_colorxyhs` (e.g. `7299355PH`)

devices.js
```diff
@@ -156,8 +178,16 @@ const preset = {
             ...preset.light_onoff_brightness_colortemp_colorxy,
             toZigbee: preset.light_onoff_brightness_colortemp_colorxy.toZigbee.concat([tz.ledvance_commands]),
         },
+        light_onoff_brightness_colorxyhs: {
```
Koenkk, line 181: Doesn't seem to be used, why add it?

devices.js
```diff
@@ -156,8 +178,16 @@ const preset = {
...
+            ...preset.light_onoff_brightness_colorxyhs,
+            toZigbee: preset.light_onoff_brightness_colorxy.toZigbee.concat([tz.ledvance_commands]),
+        },
+        light_onoff_brightness_colortemp_colorxyhs: {
```
Koenkk, line 185: Doesn't seem to be used, why add it?

devices.js
asgothian: I added the preset for all types to make it easier to actually add the function later. After what @sjorge wrote about Osram/Ledvance bulbs i added the functionality to those too, so now they are actually being used, too.

devices.js
asgothian: see comment above

devices.js
asgothian: It could be an option to remove the .._colorxy if
- all ledvance bulbs support xy and hs
- we do not expect to have a bulb in the future which only accepts either hs or xy (rgb)

This holds true for hue and general color too.

A.

Koenkk: > we do not expect to have a bulb in the future which only accepts either hs or xy (rgb)

Please remove any unused configurations. It's easy to re-add in the future.

asgothian: Done. Since "colorxyhs" seems to become the new default, i also refactored so that the default expose and presets are now
..._color, rather than ..._colorxyhs
At the same time, i have removed the preset.light_onoff_brightness_colortemp_colorhs - it was no longer in use. 
I did not touch the preset.light_onoff_brightness_colortemp_colorxy - too many devices use that which i have no information on.

Koenkk:
Thanks!

PR #2016, opened 2021-01-03
add support for Hommyn WS-20-Z

NewFolk:
It look like the device is made by ORVIBO but sold under the brand Hommyn (same as MS-20-Z)


Koenkk:
Thanks!

PR #2023, opened 2021-01-04
Add support for StartUpColorTemperatureMireds

sjorge:
Depends on koenkk/zigbee-herdsman#281

NOTE: startUp here refers to a physical off -> on cycle, not via genOnOff cluster.

`previous` is mapped to 0xFFFF which switches to the last set color.

Tested with:
- Innr CWS: works as expected
- IKEA GU10 WS: can set/get... but has no effect
- IKEA GUNTARP WS: can set/get... but as no effect
- IKEA FLOALT (no firmware update): UNSUPPORTED_ATTRIBUTE
- Hue E14 candle: works as expected
- Müller Licht Tint E27: works as expected, but needs to be powered of for ~30-60sec.
- OSRAM SMART+ LED CLASSIC E27 RGBW: UNSUPPORTED_ATTRIBUTE

Default values (of interest)
- Innr: previous
- Hue: 366
- IKEA: 370
- Tint: previous

There is a mixed bag of support on the bulbs, but since it's part of the spec, seems like a nice thing to have.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -574,7 +574,7 @@ const converters = {
 
             value = Number(value);
 
-            // ensure value withing range
+            // ensure value within range
             const [colorTempMin, colorTempMax] = light.findColorTempRange(entity, meta.logger);
             value = light.clampColorTemp(value, colorTempMin, colorTempMax, meta.logger);
 
@@ -584,6 +584,27 @@ const converters = {
         },
         convertGet: async (entity, key, meta) => {
             await entity.read('lightingColorCtrl', ['colorTemperature']);
+        },
+    },
+    light_colortemp_startup: {
+        key: ['color_temp_startup'],
+        convertSet: async (entity, key, value, meta) => {
+            if (typeof value === 'string' && value.toLowerCase() == 'previous') {
+                // 0xffff = restore previous value
+                value = 65535;
+            }
+
+            value = Number(value);
+
+            // ensure value within range
+            const [colorTempMin, colorTempMax] = light.findColorTempRange(entity, meta.logger);
+            value = light.clampColorTemp(value, colorTempMin, colorTempMax, meta.logger);
+
+            await entity.write('lightingColorCtrl', {startUpColorTemperature: value}, utils.getOptions(meta.mapped, entity));
+            return {state: {color_temp_startup: value}};
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('lightingColorCtrl', ['startUpColorTemperature']);
         },
     },
     light_color: {
--- a/devices.js
+++ b/devices.js
@@ -66,7 +66,7 @@ const preset = {
         toZigbee: [
             tz.light_onoff_brightness, tz.light_colortemp, tz.ignore_transition, tz.ignore_rate, tz.effect,
             tz.light_brightness_move, tz.light_colortemp_move, tz.light_brightness_step,
-            tz.light_colortemp_step,
+            tz.light_colortemp_step, tz.light_colortemp_startup,
         ],
         meta: {configureKey: 1},
         configure: async (device, coordinatorEndpoint, logger) => {
@@ -109,6 +109,7 @@ const preset = {
             tz.light_onoff_brightness, tz.light_color_colortemp, tz.ignore_transition, tz.ignore_rate,
             tz.effect, tz.light_brightness_move, tz.light_colortemp_move, tz.light_brightness_step,
             tz.light_colortemp_step, tz.light_hue_saturation_move, tz.light_hue_saturation_step,
+            tz.light_colortemp_startup,
         ],
         meta: {configureKey: 1},
         configure: async (device, coordinatorEndpoint, logger) => {
@@ -126,6 +127,7 @@ const preset = {
             tz.light_onoff_brightness, tz.light_color_colortemp, tz.ignore_transition, tz.ignore_rate,
             tz.effect, tz.light_brightness_move, tz.light_colortemp_move, tz.light_brightness_step,
             tz.light_colortemp_step, tz.light_hue_saturation_move, tz.light_hue_saturation_step,
+            tz.light_colortemp_startup,
         ],
         meta: {configureKey: 1},
         configure: async (device, coordinatorEndpoint, logger) => {
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -440,6 +440,11 @@ const converters = {
                 result.color_temp = msg.data['colorTemperature'];
             }
 
+            if (msg.data.hasOwnProperty('startUpColorTemperature')) {
+                result.color_temp_startup = msg.data['startUpColorTemperature'];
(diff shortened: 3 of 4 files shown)


converters/toZigbee.js
```diff
@@ -586,6 +586,30 @@ const converters = {
...
+            }
+
+            value = Number(value);
+
+            // ensure value within range
+            const [colorTempMin, colorTempMax] = light.findColorTempRange(entity, meta.logger);
+            value = light.clampColorTemp(value, colorTempMin, colorTempMax, meta.logger);
+
+            await entity.write('lightingColorCtrl', {startUpColorTemperature: value}, utils.getOptions(meta.mapped, entity));
+            return {state: {color_temp_startup: value}};
```
sjorge, line 607: @Koenkk: Is it OK to not return a state? I think perhaps a read call would be better.

That way bulbs that return UNSUPPORTED_ATTRIBUTE don't have it in there cached state.

sjorge:
Once this gets merged, I'd like to create a page somewhere to document this...

Is there an easy way to add a section to all bulbs for manufacturer X? So we can add it for all Hue and Innr bulbs?

converters/toZigbee.js
Koenkk: In case of an unsupported_attribute an exception should be thrown at `entity.write` so the return statement is never hit.

devices.js
```diff
@@ -105,6 +105,7 @@ const preset = {
             tz.light_onoff_brightness, tz.light_color_colortemp, tz.ignore_transition, tz.ignore_rate,
             tz.effect, tz.light_brightness_move, tz.light_colortemp_move, tz.light_brightness_step,
             tz.light_colortemp_step, tz.light_hue_saturation_move, tz.light_hue_saturation_step,
+            tz.light_colortemp_startup,
```
Koenkk, line 108: Also has to be added to `light_onoff_brightness_colortemp_colorxy`

Koenkk:
> Once this gets merged, I'd like to create a page somewhere to document this...
> 
> Is there an easy way to add a section to all bulbs for manufacturer X? So we can add it for all Hue and Innr bulbs?

I feel this should be added to the exposes of light, then it will be automagically added to the device page and it's possible to set it via the frontend.

sjorge:
> > Once this gets merged, I'd like to create a page somewhere to document this...
> > Is there an easy way to add a section to all bulbs for manufacturer X? So we can add it for all Hue and Innr bulbs?
> 
> I feel this should be added to the exposes of light, then it will be automagically added to the device page and it's possible to set it via the frontend.

Oh can we specify a range for those? It's basically color_temp range + 'previous'

converters/toZigbee.js
```diff
@@ -586,6 +586,30 @@ const converters = {
             await entity.read('lightingColorCtrl', ['colorTemperature']);
         },
     },
+    light_colortemp_startup: {
+        key: ['color_temp_startup', 'color_temp_startup_percent'],
+        convertSet: async (entity, key, value, meta) => {
+            if (typeof value === 'string' && value.toLowerCase() == 'previous') {
+                // 0xffff = restore previous value
+                value = 65535;
+            } else if (key === 'color_temp_startup_percent') {
```
Koenkk, line 595: remove `else` here and make it a separate if branch, otherwise `color_temp_startup_percent: 'previous'` won't work.

Koenkk:
> Oh can we specify a range for those? It's basically color_temp range + 'previous'

The fact that it's both an expose.enum and expose.numeric is problematic indeed (at least doesn't fix in the current exposes). Don't have a solution for this yet. Maybe add two exposes for the same property? (one numeric and one enum with only value `previous`). @nurrik do you maybe have an idea here?

converters/toZigbee.js
sjorge: How so? We check value will still be a string and set to previous right? So the first branch should be taken?

sjorge: Yes, I’d prefer to remove the _percent!

sjorge:
I addressed the feedback, aside form the expose...

sjorge:
@Koenkk I think we need a 'new' type, of course it will need support form the frontend I guess...

Would something like in the last commit work?

Koenkk:
Not sure about it yet, @nurikk what do you think about this from the frontend perspective?

sjorge:
> Not sure about it yet, @nurikk what do you think about this from the frontend perspective?

We could just not expose the 'previous' value and just have it as Numeric?

Koenkk:
@sjorge that is acceptable for now indeed

sjorge:
> @sjorge that is acceptable for now indeed (but I still like the idea about something like the variant as I see this as a valid use case)

Shall I update to just the Numeric for now so this can be merged? And we can pick up the Variant discussion in an issue/discussion topic?

Koenkk:
> Shall I update to just the Numeric for now so this can be merged? And we can pick up the Variant discussion in an issue/discussion topic?

yes please

sjorge:
> > Shall I update to just the Numeric for now so this can be merged? And we can pick up the Variant discussion in an issue/discussion topic?
> 
> yes please. In the `withDescription()` put something like send `65535` to disable.

It's called 'previous' in the ZCL hence why I also called it that.

nurikk:
As an option you can use another exposes(enum)+another key in your converter, and name if for example options with one value "previous startup color" .something like that , so you would not mix color temperatures and options

nurikk:
I definitely not recommend introducing new varian exposes type for this

sjorge:
We already sort of do something like a Variant for color though, both xy and hs go to the same key.
I don't think exposing more keys is the way to go.

nurikk:
Or just expose enum with single value "previous" with appropriate description

sjorge:
> Or just expose enum with single value "previous" with appropriate description

That's pretty close to the original commit for the variant type (https://github.com/Koenkk/zigbee-herdsman-converters/issues/2030), but instead of using an intermediate `Variant` type, just push 2 entries to this.features with the same key.

Let me rebase this to that approach, as they only purpuse the type had is to group them, which can be done based on key anyway.

sjorge:
Done

sjorge:
@Koenkk some shower pondering... can we dynamically set min and max range based on the attributes we’re now reading?

Koenkk:
thanks again! 🎖️ 

> can we dynamically set min and max range based on the attributes we’re now reading?

Was also thinking about this, it's possible but won't be reflected in the docs. I was hoping bulbs from the same manufacturer share the same min/max values so we can easily hardcode them (in this way docs are also updated with this).

sjorge:
Not conclusive but the trend seems to he 250-450 for my WS stuff, although those are all ikea.

The RGB bulbs seem to range from 150 to 555, but it’s mostly blue and blood orange at those. So that would make sense why you need RGB.

~ sjorge

Koenkk:
@sjorge maybe we can already add it for those we know? Could you maybe make a PR for this, I can cross check with my devices.

sjorge:
I’ll double check my test bulbs next time I have a bit of time to grab the values.

~ sjorge

PR #2025, opened 2021-01-04
Ballast configuration

guydriesen:
Added convertors and exposes for lighting ballast level controls to make it easier to configure these parameters on the ubisys D1

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -247,7 +247,7 @@ const converters = {
             'ballast_maximum_level'],
         // zcl attribute names are camel case, but we want to use snake case in the outside communication
         convertSet: async (entity, key, value, meta) => {
-            if (key === 'mode_phase_control') {
+            if (key === 'ballast_config') {
                 value = utils.toCamelCase(value);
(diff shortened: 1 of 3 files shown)


converters/toZigbee.js
```diff
@@ -240,14 +240,26 @@ const converters = {
...
+            'ballast_minimum_level',
+            'ballast_maximum_level'],
         // zcl attribute names are camel case, but we want to use snake case in the outside communication
         convertSet: async (entity, key, value, meta) => {
-            value = utils.toCamelCase(value);
-            for (const [attrName, attrValue] of Object.entries(value)) {
-                const attributes = {};
-                attributes[attrName] = attrValue;
-                await entity.write('lightingBallastCfg', attributes);
+            if (key === 'mode_phase_control') {
```
sjorge, line 250: Shouldn't `mode_phase_control` also be in the key array above?

guydriesen: Good catch! Was mixing things up, that should have been `ballast_config`...

Koenkk:
Thanks!

PR #2027, opened 2021-01-05
Add automatic trigger to get Peanut Plug OTA check to work

airdrummingfool:
Hi Koen,
Here's a crack at fixing https://github.com/Koenkk/zigbee2mqtt/issues/4846.

Note that I did _not_ do what you suggested in that issue (add code to `toZigbee.js` and then manually trigger with an MQTT message); instead, I am automatically triggering the zigbee message whenever an OTA update is checked.

While I think this is easier for the end user, I did have to export `zigbeeOTA.getImageMeta` so I could use it. I'm also not very familiar with Zigbee endpoints (or the z2m implementation thereof), so I just kinda guessed at how to write to the `genScenes` endpoint and it ended up working out.

If you don't like this implementation, please let me know what you want changed. Alternatively, I can switch to your original suggestion if you'd still prefer that.

One more note: I haven't tested actually updating a Peanut Plug yet, as I only have one that needs an update. Once you approve this PR, I'll test the full update process and make sure everything works like we expect it to.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -10525,7 +10525,7 @@ const devices = [
         description: 'Peanut Smart Plug',
         fromZigbee: [fz.on_off, fz.electrical_measurement],
         toZigbee: [tz.on_off],
-        ota: ota.zigbeeOTA,
+        ota: ota.securifi,
         meta: {configureKey: 4},
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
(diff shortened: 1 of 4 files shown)


Koenkk:
Good that you didn't listen, I like this implementation much more :smile: 

Anything left before I can merge this?

airdrummingfool:
The OTA update went smoothly and all is now good to go!

I did have to make one change: if the Peanut Plug "forgets" that an OTA update is available (e.g. if it loses power) but z2m "remembers", then the trigger data wasn't sent when the actual update process started, so the Peanut Plug wasn't responding. Now the trigger is sent before starting the update process (in addition to being sent before the update check) to make sure the Peanut Plug responds.

Koenkk:
Thanks!

PR #2028, opened 2021-01-05
Add scene_remove_all / Allow hue/sat in scene_add

sjorge:
Although I'm wondering more and more if something like the following would be better:

`{"id": 123, "group":"my_group", "state": "ON", "color_temp: 360}` => zigbee2mqtt/bridge/request/scenes/add
`{"id": 123, "device":"my_device"}` => zigbee2mqtt/bridge/request/scenes/store
`{"id": 123, "device":"my_device"}` => zigbee2mqtt/bridge/request/scenes/recall
`{"id": 123, "device":"my_device"}` => zigbee2mqtt/bridge/request/scenes/remove
`{"group":"my_group"}` => zigbee2mqtt/bridge/request/scenes/remove_all

That opens the door to implement `view`, `get_membership`, ...

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3619,6 +3619,34 @@ const converters = {
             }
         },
     },
+    scene_remove_all: {
+        key: ['scene_remove_all'],
+        convertSet: async (entity, key, value, meta) => {
+            const groupid = entity.constructor.name === 'Group' ? entity.groupID : 0;
+            const response = await entity.command(
+                'genScenes', 'removeAll', {groupid}, utils.getOptions(meta.mapped),
+            );
+
+            const isGroup = entity.constructor.name === 'Group';
+            if (isGroup) {
+                if (meta.membersState) {
+                    for (const member of entity.members) {
+                        if (member.meta.scenes) {
+                            member.meta.scenes = {};
+                            member.save();
+                        }
+                    }
+                }
+            } else if (response.status === 0) {
+                if (entity.meta.scenes) {
+                    entity.meta.scenes = {};
+                    entity.save();
+                }
+            } else {
+                throw new Error(`Scene remove all not succesfull ('${herdsman.Zcl.Status[response.status]}')`);
+            }
+        },
+    },
     TS0003_curtain_switch: {
         key: ['state'],
         convertSet: async (entity, key, value, meta) => {
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3584,7 +3584,7 @@ const converters = {
                 'genScenes', 'remove', {groupid, sceneid}, utils.getOptions(meta.mapped),
             );
 
-            if (isGroup || (removeresp.status === 0 || removeresp.status == 139)) {
+            if (isGroup || (removeresp.status === 0 || removeresp.status == 133 || removeresp.status == 139)) {
                 const response = await entity.command(
                     'genScenes', 'add', {groupid, sceneid, scenename, transtime, extensionfieldsets}, utils.getOptions(meta.mapped),
                 );
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3547,9 +3547,27 @@ const converters = {
                     } catch (e) {
                         e;
                     }
-                    const xy = typeof val === 'string' ? utils.hexToXY(val) : val;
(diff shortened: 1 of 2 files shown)


sjorge:
I did a push with lint fix but github.com is not picking it up?

MattWestb:
One device can having more scenes with the the same ID but belongs different groups.
Example grope 0x0001 have scene ID 1, 2, 3  and 4. Groupe 0x0003 have scene 1, 4, and 6.
Recalling the scenes is done with grupe and scene ID so grope 0x0003 scene 1 is only recaling it and not the scene for the groupe 0x0002 scene ID 1.

That is giving the possibility to using overlapping groups with overlapping scenes thats can being very good to using. 

Groupe 0x0000 is in ZLL and ZB3 reserved as default scene group (but i dont knowing how its intended to being used / working).

sjorge:
@MattWestb that should already work, you can add the same scene ID to both a device (0x0000) or to a group

MattWestb:
Great and you is doing good move forward getting the complete package working as its looks :-)))

sjorge:
> Great and you is doing good move forward getting the complete package working as its looks :-)))

For other stuff like get members/view/... we probably need to implement this as something outline in the original comment.
Not sure what to do with these convertors at that point. As the other code will live in z2m we can't simply call one from the other...

I guess abstracting this to a `lib/scenes` and having the convertor and z2m use that something that could work.
But I'm not 100% sure that could work, and such a big refactor is not something I feel super comfortable with.

Koenkk:
> Not sure how to only write the enhanchedHue and saturation fields but this works for my enhancedHue supporting Innr/Tint bulbs. It does not work with trådfri (no Hue/Sat support in the colorCapabilities) or with the Innr RB 250 C (no enhancedHue support)

That's not possible, it's an array and all values that you dont' want to set have to be left 0 (your sniff looks good).


> That opens the door to implement view, get_membership, ...

Why is this not possible now?





Can you also open a PR for https://github.com/Koenkk/zigbee2mqtt.io/blob/develop/docs/information/scenes.md , then this can be merged (have no comments on the code).

sjorge:
> > That opens the door to implement view, get_membership, ...
> 
> Why is this not possible now?

You could probably do scene_view, but how would you do get_membership? 

> Can you also open a PR for https://github.com/Koenkk/zigbee2mqtt.io/blob/develop/docs/information/scenes.md , then this can be merged (have no comments on the code).

Will do

sjorge:
Done koenkk/zigbee2mqtt.io#546

MattWestb:
Getting membership its looks like requesting view scene.
Request:
```
ZigBee Application Support Layer Data, Dst Endpt: 1, Src Endpt: 1
    Frame Control Field: Data (0x40)
    Destination Endpoint: 1
    Cluster: Scenes (0x0005)
    Profile: Home Automation (0x0104)
    Source Endpoint: 1
    Counter: 133
ZigBee Cluster Library Frame
    Frame Control Field: Cluster-specific (0x11)
        .... ..01 = Frame Type: Cluster-specific (0x1)
        .... .0.. = Manufacturer Specific: False
        .... 0... = Direction: Client to Server
        ...1 .... = Disable Default Response: True
    Sequence Number: 34
    Command: Get Scene Membership (0x06)
    Payload
        Group ID: 0x002a
```
Respond:
```
ZigBee Network Layer Data, Dst: 0x0304, Src: 0xda58
ZigBee Application Support Layer Data, Dst Endpt: 1, Src Endpt: 1
    Frame Control Field: Data (0x40)
    Destination Endpoint: 1
    Cluster: Scenes (0x0005)
    Profile: Home Automation (0x0104)
    Source Endpoint: 1
    Counter: 74
ZigBee Cluster Library Frame
    Frame Control Field: Cluster-specific (0x19)
        .... ..01 = Frame Type: Cluster-specific (0x1)
        .... .0.. = Manufacturer Specific: False
        .... 1... = Direction: Server to Client
        ...1 .... = Disable Default Response: True
    Sequence Number: 34
    Command: Get Scene Membership Response (0x06)
    Payload
        Scenes Status: Success (0x00)
        Scene Capacity: 61
        Group ID: 0x002a
        Scene Count: 0
```
Its first necessary to knowing with groups the devices is having that can having scenes configured and then requesting of each group for getting all scenes stored in the device.

Koenkk:
Thanks!

PR #2029, opened 2021-01-05
Added Müller Licht remote control support (404022)

nmaupu:
Hi,

My device is the following:
https://zigbee.blakadder.com/Muller_Licht_404022.html

It should be supported but device's config is missing from `devices.js` file...

This configuration is taken from `404002`'s model which is similar but the `404022` remote also has two buttons for _color temperature_ which are not bound with this configuration. I am not sure how to implement that in `converters/fromZigbee.js`.

Cluster: `lightingColorCtrl`
Command: `commandMoveToColorTemp`

In the meantime, on/off and step up/down are working as expected with this PR.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -9652,6 +9652,16 @@ const devices = [
         },
     },
     {
+        zigbeeModel: ['tint-Remote-white'],
+        model: '404022',
+        description: 'Tint dim remote control',
+        vendor: 'Müller Licht',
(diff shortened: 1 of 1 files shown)


Koenkk:
>  I am not sure how to implement that in converters/fromZigbee.js.

to the fromZigbee array add `fz.command_move_to_color_temp`.

nmaupu:
Fixed, nothing to add in the `exposes` array ?

Koenkk:
Added: `color_temperature_move`. Thanks!

PR #2032, opened 2021-01-05
Added Support for Aurora Wall Remotes

ashp8i:
Added Support for Aurora AOne One Gang Wireless Battery Rotary Dimmer and Aurora AOne Two Gang Wireless Battery Rotary Dimmer

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14212,8 +14212,8 @@ const devices = [
         description: 'AOne One Gang Wireless Battery Rotary Dimmer',
         supports: 'action, action_step_size, action_transition_time, action_color, color, color_temp, linkquality, battery',
         fromZigbee: [fz.battery, fz.command_on, fz.command_off, fz.command_step, fz.command_step_color_temperature],
-        toZigbee: [tz.light_brightness_step],
-        exposes: [e.battery(), e.action(['on', 'off', 'brightness_step_up', 'brightness_step_down'])],
+        toZigbee: [],
+        exposes: [e.battery(), e.action(['on', 'off', 'brightness_step_up', 'brightness_step_down', 'color_temperature_step_up', 'color_temperature_step_down'])],
         meta: {configureKey: 1, battery: {voltageToPercentage: '3V_2100'}},
         endpoint: (device) => {
             return {'default': 1};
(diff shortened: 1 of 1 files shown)


ashp8i:
Not sure why lint is failing

devices.js
```diff
@@ -14205,6 +14205,46 @@ const devices = [
             await reporting.instantaneousDemand(endpoint);
         },
     },
+    {
+        zigbeeModel: ['1GBatteryDimmer50AU'],
+        model: 'AU-A1ZBR1GW',
+        vendor: 'Aurora Lighting',
+        description: 'AOne one gang wireless battery rotary dimmer',
+        fromZigbee: [fz.battery, fz.command_on, fz.command_off, fz.command_step, fz.command_step_color_temperature],
+        toZigbee: [tz.light_brightness_step],
```
Koenkk, line 14214: Why was `tz.light_brightness_step` added here? Would expect it to be empty. Same for https://github.com/Koenkk/zigbee-herdsman-converters/pull/2032/files#diff-5520dd49755e0a353d3857b8963a91cb4b8d87a01b6b8eb7bb6cce1c180eac83R14233

devices.js
```diff
@@ -14205,6 +14205,46 @@ const devices = [
...
+        exposes: [e.battery(), e.action([
+            'on', 'off', 'brightness_step_up', 'brightness_step_down', 'color_temperature_step_up', 'color_temperature_step_down'])],
+        meta: {configureKey: 1, battery: {voltageToPercentage: '3V_2100'}},
+        endpoint: (device) => {
```
Koenkk, line 14218: This shouldn't be needed.

devices.js
ashp8i: sorry not a pro got confused by the controller giving out the step value:

DEBUG: Received Zigbee message from 'Single Rotary Dimmer Remote', type 'commandStep', cluster 'genLevelCtrl', data '{"stepmode":0,"stepsize":26,"transtime":3}' from endpoint 1 with groupID 0
INFO: MQTT publish: topic 'zigbee2mqtt/Single Rotary Dimmer Remote', payload '{"action":"brightness_step_up","action_step_size":26,"action_transition_time":0.03,"linkquality":60,"update":{"state":"idle"},"update_available":false}'

ashp8i: thought passing on the value might be of some importance

devices.js
ashp8i: which line are you referring to

ashp8i: the device shows a greenpower: -

image

not sure if it is of any relevance

ashp8i: also 
image

ashp8i: when used with an aurora hub which I do have at hand dimmer ready with a zigbee sniffer also supports changing colour and color wheel possibly scenes

Koenkk: I'm refering to the endpiont line

devices.js
Koenkk: passing = fromZigbee, sending to device = toZigbee

devices.js
ashp8i: I realised it won't be needed as there are only two endpoints1 and 242 and the one being used is 1 any idea what 242 would do (greenpower?)

Koenkk: Probably this device can serve as a green power sink or proxy. (for more info search Zigbee green power).

PR #2036, opened 2021-01-06
Müller Licht Tint supports hue/saturation

sjorge:
On adventure in colorCapabilities, I found out these do support hue/saturation! (Tested both a GU10 and an E27)

They do need enhancedHue set to false though, which is annoying as that means we can't trust colorCapabilities! It is set to 31 for this both E27 and GU10 but neither change color when setting via enhancedHue, but they also don't reject it with UNSUPPORTED_ATTRIBUTE *sad all around*


Koenkk:
Thanks!

Issue #2035, opened 2021-01-06
ROB_200-007-0 issues an unsupported command.... 


noci2012:
The following is in the log for such a device:

Received Zigbee message from 'my4buttons', type 'commandQueryNextImageRequest', cluster 'genOta', data '{"fieldControl":1,"fileVersion":22,"imageType":1008,"manufacturerCode":4644}' from endpoint 1 with groupID null

What ota entries are needed, how to find out?

Koenkk:
Does the manufacturer publish OTA files for these devices? 
- If not you can ignore this
- If yes, add the firmware files here: https://github.com/koenkk/zigbee-ota and add `ota: ota.zigbeeOTA,` to the device.

noci2012:
As i can't find firmware with ROBB-shop or manuafacturer, i added zf.ignore_genOta to the fromZigbee.

noci2012:
PR #2040 updates this

Koenkk:
Closing this as https://github.com/Koenkk/zigbee-herdsman-converters/pull/2040 is merged

Issue #2034, opened 2021-01-06
ROB_200-007-0 is in devices.js and stil claimed to be unsupported


noci2012:
The log mentions this after interviewing the device:  After successful pairing....

Device '0xdddddddddddddd' with Zigbee model 'ROB_200-007-0' and manufacturer name 'ROBB smarrt' is NOT supported, please follow https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html

The device is recognized AFTER adding a fingerprint entry to the devices.js entry for ROB_200-007-0
       fingerprint: [ {modelID: 'ROB_200-007-0', manufacturerName: 'ROBB smarrt'}],

The device does have 8 buttons:
4 on and 4 off, not 8 sets of buttons.
so on_5.. on_8 etc. seem to be wrong.

Simanias:
Yup got exactly the same issue.
`Zigbee2MQTT:warn  2021-01-06 08:09:31: Received message from unsupported device with Zigbee model 'ROB_200-007-0' and manufacturer name 'ROBB smarrt'`

defcon84:
This is the changeset where the exposes are added:
https://github.com/Koenkk/zigbee-herdsman-converters/commit/6433de2c950f5b7b403238a8ab298034abad52cf#diff-5520dd49755e0a353d3857b8963a91cb4b8d87a01b6b8eb7bb6cce1c180eac83R7028
These are indeed too many.

While not correct, the device works properly here with zigbee2mqtt version 1.17.0 commit: 07cdc9d.

ROBB smarrt with double 'r' is correct.

Simanias:
Okey @defcon84 , thanks for the quick reply, however no joy here on 1.17.0.1 with home assistant.
Device shows up, but says unsupported and is not added to the MQTT integration.
<img width="1616" alt="Schermafbeelding 2021-01-06 om 10 38 16" src="https://user-images.githubusercontent.com/4976061/103753408-4ce6f980-500b-11eb-9406-55baf0eebb98.png">

defcon84:
Hmm weird. 
Hassio z2m addon v1.17.0.1 uses z2m v1.17.0, so it should be the same as mine.
Can you post the full debug log while pairing.

Simanias:
[...]

PR #2040, opened 2021-01-06
Update devices.js

noci2012:
Fix entry for ROBB-shop 4 way switch. 
- fingerprint is needed to get recognized
- only 4 buton pairs,
- added ignore ota request. (no firmware found).


noci2012:
Solution for Issue: #2034 and #2035

Koenkk:
Thanks!

PR #2041, opened 2021-01-06
feat: support timed on/off commands in _TYZB01 switches

majektom:
Switch modules representing themselves as `modelID: 'TS0003', manufacturerName: '_TYZB01_*'` are capable of receiving **onWithTimedOff** command in **onOff** cluster. However, they interpret the command on their own, non-standard way. This PR adds support for timed on and timed off commands for these devices. In order to issue timed on or off command, the following mqtt message would need to be sent: `{ "topic": "zigbee2mqtt/${friendly_name}/set", "payload": { "state(_${endpoint_name}": {"state": "(on|off)", "time_in_seconds": ${seconds}}}}`.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -148,19 +148,32 @@ const converters = {
         },
     },
     TYZB01_on_off: {
-        key: ['state'],
+        key: ['state', 'transition'],
         convertSet: async (entity, key, value, meta) => {
-            if (typeof value !== 'object') {
-                return converters.on_off.convertSet(entity, key, value, meta);
+            const result = await converters.on_off.convertSet(entity, key, value, meta);
+            const lowerCaseValue = value.toLowerCase();
+            if (!['on', 'off'].includes(lowerCaseValue)) {
+                return result;
             }
-            utils.validateValue(value.state, ['on', 'off']);
-            const timeInSeconds = Number(value.transition);
+            const messageKeys = Object.keys(meta.message);
+            const transitionValue = function() {
+                if (messageKeys.includes('state')) {
+                    return meta.message.transition;
+                }
+                if (meta.endpoint_name) {
+                    return meta.message[`transition_${meta.endpoint_name}`];
+                }
+                return null;
+            }();
+            if (!transitionValue) {
+                return result;
+            }
+            const timeInSeconds = Number(transitionValue);
             if (!Number.isInteger(timeInSeconds) || timeInSeconds < 0 || timeInSeconds > 0xfffe) {
-                throw Error('The JSON object must have \'transition\' property which is ' +
-                            'convertible to an integer in the range: <0x0000, 0xFFFE>');
+                throw Error('The transition value must be convertible to an integer in the '+
+                            'range: <0x0000, 0xFFFE>');
             }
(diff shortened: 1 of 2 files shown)


converters/toZigbee.js
```diff
@@ -147,6 +147,35 @@ const converters = {
...
         },
     },
+    TYZB01_on_off: {
+        key: ['state'],
+        convertSet: async (entity, key, value, meta) => {
+            if (typeof value !== 'object') {
+                return converters.on_off.convertSet(entity, key, value, meta);
+            }
+            utils.validateValue(value.state, ['on', 'off']);
+            const timeInSeconds = Number(value.time_in_seconds);
```
Koenkk, line 157: Instead of `time_in_seconds` can you rename this to `transition` (to be more in line with other devices).

converters/toZigbee.js
```diff
@@ -147,6 +147,35 @@ const converters = {
             await entity.read('genOnOff', ['onOff']);
         },
     },
+    TYZB01_on_off: {
+        key: ['state'],
+        convertSet: async (entity, key, value, meta) => {
+            if (typeof value !== 'object') {
```
Koenkk, line 153: If I understand correctly now this only work via `{"state": {"state": "ON", "transition":1}}`, however `{"state":"ON","transition":1}` is preferred (and in line with others.. Change `key: ['state'],` to `key: ['state', 'transition'],` and access the state through `const state = meta.message.state` and `const transition = meta.message.transition`.  Also make sure this converter works with `{"state":"TOGGLE"}` ( `converters.on_off.convertSet(entity, key, value, meta);` should be called in this case)

converters/toZigbee.js
```diff
@@ -147,6 +147,48 @@ const converters = {
             await entity.read('genOnOff', ['onOff']);
         },
     },
+    TYZB01_on_off: {
+        key: ['state', 'transition'],
+        convertSet: async (entity, key, value, meta) => {
+            const result = await converters.on_off.convertSet(entity, key, value, meta);
```
Koenkk, line 153: I would expect the `converters.on_off.convertSet` only be called on a toggle. Now always first an on/off command is send followed by the onWithTimedOff of this converter. Is this correct?

majektom: Yes. This is intended. And just to clarify because I might have not made it clear in the PR's description.  The command `set: { state: "ON", transition: 10 }` means - turn on and then turn off after 10 seconds. The modules misinterpret the Zigbee **onWithTimedOff** command in **onOff** cluster. They understand the command as follows:
* if the switch is on (onOff == 1) and received `onWithTimedOff: { ctrlbits: (no-matter), ontime: (no-matter), offwaittime: (x) }` command, turn the switch off after x seconds
* if the switch is off (onOff == 0) and received `onWithTimedOff: { ctrlbits: (no-matter), ontime: (x), offwaittime: (no-matter) }`, turn the switch on after x seconds

Koenkk: Ah sorry, now I understand this MR. The `transition` doesn't fit in this case, `time_in_seconds` like you initially did suits better.

Koenkk:
Thanks!

PR #2044, opened 2021-01-07
Added support for LEDVANCE Undercabinet light (4058075173989)

darookee:
Added support for LEDVANCE Undercabinet light (4058075173989)


Koenkk:
Thanks!

PR #2048, opened 2021-01-08
Müller Licht Tint remote (model 404002) "recall" messages

andreas-seiderer:
Hello,

I just added the "recall" command from "ZBT-Remote-EU-DIMV2A2" to "ZBT-DIMController-D0800" I didn't add and thus were not sent via MQTT.

The only difference now to "ZBT-Remote-EU-DIMV2A2" should be the missing "fz.ignore_basic_report". Is it required (I didn't get any errors until now)? If yes, it could also be added.

Best regards,

Andreas

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -9640,9 +9640,9 @@ const devices = [
         model: '404002',
         description: 'Tint dim remote control',
         vendor: 'Müller Licht',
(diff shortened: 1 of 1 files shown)


Koenkk:
Looks good, `fz.ignore_basic_report` does nothing and only supresses `no converter.. ` messages.

andreas-seiderer:
Ok, thanks!

Issue #2045, opened 2021-01-07
Uncaught exception crashes Zigbee-Herdsman


asgothian:
Hello.

i have been looking into [this issue](https://github.com/ioBroker/ioBroker.zigbee/issues/927) on the iobroker.Zigbee Adapter, which comes from deep within the Zigbee Herdsman code.

After a long session with someone who has the affected Thermostat i was able to confirm that
- The event tuya.onEventSetTIme triggers the crash
- The Underlying herdsman code (endpoint.command(...) seems to fail due to the thermostat not responding after the command was sent, and then throws an error
- Introducing a try/catch in the onEventSetTime function removes the crash, but also hides the error from the log.

I had planned to introduce a pull request adding the try/catch to the onEventSetTime function, but looking through the code it seemed that none of the "onEvent" functions do any kind of error handling. This leads me to believe that the error may be better caught within the onEvent handler in the Zigbee Herdsman.

Since this is a breaking error, i would wish for a quick solution, but due to lack of knowledge on typescript and the zigbee-herdsman, the only fix i could offer would be said try/catch in tuya.js.

I can offer to make said pull-request to the Zigbee-Herdsman-Converters (which is why i opened this issue here, rather than on the zigbee-herdsman). I can also offer to attempt a fix in the herdsman itself, but would need a few pointers where exactly to look.

A

Koenkk:
This issue should indeed be fixed in the tuya.js with a try/catch, can you make a pr?

asgothian:
Will do.

PR #2052, opened 2021-01-09
Error handling for Tuya onEvent functions

asgothian:
See issue #2045:

A try / catch with an empty catch added to prevent the Zigbee-Herdsman to crash due to an uncaught exception in an async/await scenario.

The catch path is empty as i see no where to go when the error occurs. It is logged in the herdsman as error and can be output using the herdsman debug functionality.

A.


Koenkk:
Thanks!

PR #2055, opened 2021-01-10
support Lonsonho Tuya Zigbee Smart Valve Controller

swex:
Here is device on aliexpress: https://www.aliexpress.com/item/4001150563365.html (upgraded version)
Don't know where to get 'real' model id, so used item name.


Koenkk:
Thanks!

PR #2061, opened 2021-01-11
Added support for Iluminize 511.000 (ZigBee 3.0 LED-controller)

jd1900:
Added support for Iluminize 511.000 (ZigBee 3.0 Universal LED-controller, 5 channel 4A, RGBCCT LED)

Basically same as '511.040' just changing model and description.
Tested and works great.
More info about the device:
https://www.iluminize.com/de/shop/led-steuerung/led-controller/product/596-511-000-zigbee-controller-4a.html

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -11180,6 +11180,19 @@ const devices = [
         extend: preset.light_onoff_brightness_colortemp_colorxy,
     },
     {
+        zigbeeModel: ['HK-ZD-RGBCCT-A', '511.000'],
+        model: '511.000',
+        vendor: 'Iluminize',
+        description: 'ZigBee 3.0 Universal LED-controller, 5 channel 4A, RGBCCT LED',
+        extend: preset.light_onoff_brightness_colortemp_colorxy,
+        meta: {configureKey: 1},
+        configure: async (device, coordinatorEndpoint, logger) => {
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -11179,6 +11179,19 @@ const devices = [
...
         extend: preset.light_onoff_brightness_colortemp_colorxy,
     },
+    {
+        zigbeeModel: ['HK-ZD-RGBCCT-A', '511.000'],
+        model: '511.000',
+        vendor: 'Iluminize',
+        description: 'ZigBee 3.0 Universal LED-controller, 5 channel 4A, RGBCCT LED',
+        extend: preset.light_onoff_brightness_colortemp_colorxy,
+        meta: {configureKey: 1},
+        configure: async (device, coordinatorEndpoint, logger) => {
```
Koenkk, line 11189: Why was the configure added here?

jd1900: To be honest I didn't even realize of that. As described, I just copied configuration from 511.040 which shares same capabilities and Zigbee 3.0 compatibility and AFAIK only changes the number of channels and amperage. But supports same things.

Koenkk: The `511.040` doesn't have this so I will remove it. Thanks!

PR #2069, opened 2021-01-12
Fix Tuya cover inversion for Zemismart ZM79E-DT

DomiStyle:
This pull request fixes the incorrect close/open commands for my Zemismart ZM79E-DT that got changed a while back when all the Tuya curtains and roller blinds were merged.

The issue (mostly) comes from this line: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/converters/toZigbee.js#L3370
Since my curtain is not a roller blind the incorrect enum was sent for my device. Also, an incorrect invert condition was used when receiving data from the curtain.

Initial version from me: https://github.com/DomiStyle/zigbee-herdsman-converters/commit/d3f2d249eb858e2ffd6718378643f52d3f76ac62
Last commit changing this: https://github.com/Koenkk/zigbee-herdsman-converters/commit/2a860950bf6b564135b3fdb96e90f7db6afa9176

I attached the condition for reversing the open/close commands to the invert attribute now, since that is set by default on the roller blinds anyway.

I'm not sure if this breaks anything for the roller blind folks, maybe someone with a Tuya roller blinds can chime in if this works for them.

@dinhchinh82 @chinhhut @Koenkk @TheEnd98

I can confirm that the Zemismart ZM79E-DT works perfectly with these changes. Reversing it properly reverses the position but also reverses the close/open commands, which seems to be needed for the roller blinds?


converters/toZigbee.js
```diff
@@ -3357,24 +3357,23 @@ const converters = {
...
                 if (value >= 0 && value <= 100) {
-                    const invert = !(meta.mapped.meta && meta.mapped.meta.coverInverted ?
-                        !meta.options.invert_cover : meta.options.invert_cover);
                     value = invert ? 100 - value : value;
                     await tuya.sendDataPointValue(entity, tuya.dataPoints.coverPosition, value);
                 } else {
                     throw new Error('TuYa_cover_control: Curtain motor position is out of range');
                 }
             } else if (key === 'state') {
-                const isRoller = meta.mapped.model === 'TS0601_roller_blind';
```
Koenkk, line 3370: Inverting should only invert the position, not the state. open should still be open, not closed. The only thing we want to reverse is position.

By default it should be open = 100, close = 0

DomiStyle: @Koenkk That's what I thought, but the previous version sent 0 for close and 2 for open only for roller blinds. All curtain motors had 2 for close and 0 for open. My curtain motor also needs 0 for close and 2 for open though.

So there are 3 options I can think of:
* Add the previous check if model is TS0601_roller_blind but also check if zigbeeModel contains my specific model
* Apply the roller blind behaviour to all curtains
* Split up my curtain (owvfni3) from the other curtain motors [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L1380)

I'm not sure which ones need 2 for close and 0 for open or if its not necessary at all but I assume it was added for a reason.

Koenkk: @DomiStyle i propose to merge `TS0601_curtain` and `TS0601_roller_blind` and check the `manufacturerName` if 2 or 0 needs to be send for open/close.

DomiStyle: @Koenkk Where would be the best place to store which manufacturerNames need to be inverted by default and which need to reverse open/close commands? In an array somewhere or hardcoded in the tuya_cover_control function?

DomiStyle: There's this issue which also seems to have trouble with the cover state: https://github.com/Koenkk/zigbee2mqtt/issues/5596

Not even stop seems to be mapped correctly in this case. Maybe we should have multiple presets and add the covers one by one once its known which direction they go?

Koenkk: > edit: In the tuya.js maybe?

Yes, add a function: `isTuyaCoverInverted`, and by default return false.

DomiStyle:
@Koenkk I added isCoverInverted and getCoverStateEnums to tuya.js to handle both scenarios.

Tested on my curtain once again and it works fine with the default values, I added all the overrides as they were, in addition to the one mentioned in https://github.com/Koenkk/zigbee2mqtt/issues/5596#issuecomment-759408189.

In case anyone complains in the future they can add their manufacturerName to the coverPositionInvert and/or coverStateOverride.

I also added some debug output to help finding out which enums are used for the cover.

Koenkk:
I've made some small changes, merged some more definitions into `TS0601_cover`.

Also changed the default enums to `{close: 2, open: 0, stop: 1}`. This was the default before this MR and is used by most devices. Changing the default would break devices which we don't know the manufacturerName of (all those identified via the `zigbeeModel`). 

Can you do a final check? Then I will merge this

DomiStyle:
@Koenkk Should I add my curtain to the fingerprint section instead of the zigbeeModel section too or does it not matter? Or should I list it in both?

Mine is currently listed as `owvfni3\u0000`

Koenkk:
@DomiStyle no that's not needed, only the ones with `TS0601` since that zigbeeModel is used a lot . Thanks!

RikImbens:
Is the ZM79E-DT curtain motor Zigbee version compatible with the DeConz software? Or is this discussion about another platform?

dinhchinh82:
@DomiStyle @Koenkk hello, today I just updated new version of z2m 1.17.1 and my curtains did not works anymore.
The status of the curtain is not correct. If I set the invert cover by the remote, it is not able to control the curtain.

PR #2073, opened 2021-01-13
Fix DIYRuZ_AirSense reporting

nurikk:
Configure reporting only for measuredValue attributes

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -4316,16 +4316,14 @@ const devices = [
         meta: {configureKey: 1},
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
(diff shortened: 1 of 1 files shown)


dmitrychepel:
please leave support scaled values as is.
in the forks it still used,
and I do not have plans to remove this support.
I see that this code works correct.

nurikk:
> please leave support scaled values as is.
> in the forks it still used,
> and I do not have plans to remove this support.
> I see that this code works correct even if current client is not support return scaled values.

Kindly use another zigbee model in forks or use external converter. Additionally you can manually configure reporting from z2m admin interface

Koenkk:
Thanks!

Issue #2076, opened 2021-01-13
Missing OTA for Sengled E12-N1E


busheezy:
Sengled E12-N1E does not actually have an OTA image.

https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L5853

It references imageType 1016 but the image does not exist.
https://github.com/Koenkk/zigbee-OTA/blob/master/index.json

```ERROR: Failed to call 'OTAUpdate' 'onZigbeeEvent' (AssertionError [ERR_ASSERTION]: No image available for imageType '1016' at getImageMeta (/app/node_modules/zigbee-herdsman-converters/lib/ota/zigbeeOTA.js:15:5) at runMicrotasks (<anonymous>) at processTicksAndRejections (internal/process/task_queues.js:97:5) at async isNewImageAvailable (/app/node_modules/zigbee-herdsman-converters/lib/ota/common.js:263:18) at async Object.isUpdateAvailable (/app/node_modules/zigbee-herdsman-converters/lib/ota/common.js:254:23) at async OTAUpdate.onZigbeeEvent (/app/lib/extension/otaUpdate.js:51:31) at async Controller.callExtensionMethod (/app/lib/controller.js:381:21))```

I have 10 of these bulbs so I am getting quite a bit of spam and it is hard to view the UI.

Koenkk:
Closing as https://github.com/Koenkk/zigbee-herdsman-converters/pull/2078 is merged

PR #2077, opened 2021-01-13
Remove OTA from Sengled E12-N1E

busheezy:
Issue: https://github.com/Koenkk/zigbee-herdsman-converters/issues/2076

I believe removing this line will prevent a lot of log spam for anyone using this bulb.

Thanks


busheezy:
I see my tabs are all over the place. Damn. Should I just make a new PR?

PR #2084, opened 2021-01-14
Update device exposes access with toZigbee values

pedrolamas:
The way the exposes works has hard-coded indication for get/set access on an property.

However, the device still needs to have a `toZigbee` entry with corresponding `convertGet` and/or `convertSet` for it work, so seems to me that those are actually the ones who should indicate the access to a property.

This code will take the whole `toZigbee` value of each device, create a map for each existing property and if one can get/set on it, and then update the exposes according to that information.

At this stage I'm not sure of the whole implications of this change(!), but I do know that the frontend now doesn't show the "refresh" button for properties that don't have a `toZigbee` entry with `convertGet`:

image

This should be a better fix for this frontend issue: https://github.com/nurikk/z2m-frontend/issues/115

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -46,6 +46,7 @@ const {repInterval, defaultBindGroup, OneJanuary2000} = require('./lib/constants
 const reporting = require('./lib/reporting');
 
 const e = exposes.presets;
+const ea = exposes.access;
 const preset = {
     switch: () => ({
         exposes: [e.switch()],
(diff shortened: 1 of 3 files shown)


Koenkk:
Correct me if I'm wrong but this feels a bit strange. I feel the correct value should be set in the `exposes:` off the definition itself (and not touch it later).

Instead of correcting here, what if we make a test out of this?

pedrolamas:
I do get your point but the problem right now is that we have devices with properties in exposes indicating that one can do a get/set, yet the converter is missing on the toZigbee.

A good example is the smart outlets with power monitoring: most of them have the `exposes.power` indicating to the frontend that "get" is possible (thus showing the refresh button), but when you check the toZigbee, there is no handler for "power" property.

My "fix" is basically making sure that exposes doesn't say one can do a get when thst property is missing from toZigbee or it doesn't have a convertGet, but I'm open to suggestions on better ways of fixing this!

pedrolamas:
Added a test to check that the device exposes access indication does conform with the existing `toZigbee` entries.

Koenkk:
Thanks, I really like the test part. But I don't know about the automatic adaption in devices.js yet. The problem I'm having with it that you can define a value that is later overridden again (which can cause confusion/feel like magic). I think either the `access` property should be completely generated (but I don't think it's possible now because we don't know about the `STATE` or completely manually defined (where the tests ensure that's its correctly defined).

Let me know what you think. I've added a small script to dump the exposes: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/scripts/dumpExposes.js which allows to do a textual compare. I did one with before and after your changes:  [before.txt](https://github.com/Koenkk/zigbee-herdsman-converters/files/5831373/before.txt) and  [after.txt](https://github.com/Koenkk/zigbee-herdsman-converters/files/5831375/after.txt). If you agree I will update this pr with the manual adaptions (as I think I can correct it quite fast)

pedrolamas:
Fully agree! Comparing the before and after, seems most of the differences are around the power/energy monitoring properties (so that would be just a couple of new `toZigbee` converters to fix).

 guess we can fix those cases (need to be careful as "power" attribute can be provided by "seMetering" or "haElectricalMeasurement") and then go fix the remaining ones!

Koenkk:
Big thanks!

pedrolamas:
Really like how this got in the end! Thank you!! 🙂

PR #2085, opened 2021-01-15
Support SMART+ LED CLASSIC E27 RGBW V2

sjorge:
As mentioned in this issue https://github.com/Koenkk/zigbee2mqtt/issues/5740

I did swap the vender to OSRAM to be inline with the other bulbs.


Koenkk:
Thanks!

PR #2097, opened 2021-01-16
Add support for Philips Hue spot "Fugato" (5063331P7)

Staacks:
This is just another Hue spot with a different id that reflects the design of the lamp instead of the bulbs. I just copied the definition and adjusted id and name. Now it works perfectly.

(Just set up my first zigbee2mqtt system a few days ago and this was the only device not working out of the box. Thanks a lot for this fantastic project and the freedom it offers!)


Koenkk:
Thanks!

PR #2098, opened 2021-01-16
Add additional zigbeeModel for for ORVIBO SM10ZW

erikman:
Related to https://github.com/Koenkk/zigbee2mqtt/issues/5212

I found out that the buggy firmware seems to automatically use the coordinatorEndpoint as `iasCieAddr`, so I just needed to add the `zigbeeModel` and use a recent version of zigbee-herdsman for the device to work (interview still fails, will update documentation for that).

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -12755,7 +12755,7 @@ const devices = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2100, opened 2021-01-17
Change default group of E1766 so it doesn't control group 0 devices

nikechristo:
I recently added an Aqara curtain motor to my setup and I noticed that my Ikea E1766 switches started controlling them. Upon further investigation, I stumbled upon https://github.com/Koenkk/zigbee2mqtt/issues/2772 which changes the default group for a couple of Ikea switches but not E1766. In this PR I am fixing it for Ikea E1766.


Koenkk:
Thanks!

PR #2101, opened 2021-01-17
E220-KR2N0Z0-HA

castorw:
Adds support for `E220-KR2N0Z0-HA`, same build as `ZB-SW02` only with different LED colours.


Koenkk:
Thanks!

PR #2103, opened 2021-01-17
Add support for iCasa Keypad Pulse 2 (ICZB-KPD12)

Derecho:
As the title says.

Support was easy to add as other models in the series were already supported, I've only removed the converters for functionalities not supported by this keypad compared to those with more buttons. Functionality tested in default legacy mode and with legacy mode off, all events recognised. (Battery percentage is sent by the device on a short press of both the I and O button). Linting and testing passed fine.

I haven't added a documentation page for this device as I've read that that is supposed to be generated automatically. Should I do anything about adding a picture for this device?


devices.js
```diff
@@ -9571,6 +9571,17 @@ const devices = [
             await reporting.onOff(endpoint);
         },
     },
+    {
+        zigbeeModel: ['ICZB-KPD12'],
+        model: 'ICZB-KPD12',
+        vendor: 'iCasa',
+        description: 'Zigbee 3.0 Keypad Pulse 2',
+        meta: {battery: {dontDividePercentage: true}},
+        fromZigbee: [fz.command_on, fz.legacy.genOnOff_cmdOn, fz.command_off, fz.legacy.genOnOff_cmdOff, fz.battery,
```
Koenkk, line 9580: Please don't use the `fz.legacy` here.
- `fz.legacy.genOnOff_cmdOn`, `fz.legacy.genOnOff_cmdOff` can be removed
- `fz.legacy.cmd_move_with_onoff` -> `fz.command_move`
- `fz.legacy.cmd_stop_with_onoff` -> `fz.command_stop`

Derecho: Hm, ok. But for sake of consistency, shouldn't those also be removed/replaced for the Pulse 4S and 8S then? Those products are very similar and this is where I copied these conversions from.

Koenkk: No because that will create a breaking change for existing users.

Koenkk:
Thanks!

PR #2104, opened 2021-01-17
Fix model numbers

blakadder:
reference: 

https://m1.lmcdn.fr/media/1/5f3175aeb7c50568561f82b9/.pdf for LXEK-2

https://m2.lmcdn.fr/media/1/5f317525b7c50568561f82b7/.pdf for LXEK-1


Koenkk:
Thanks!

PR #2110, opened 2021-01-18
TuYa TS0504A RGBW LED Controller support.

milo1000:
Bought that one recently over internet (https://allegro.pl/oferta/kontroler-led-zigbee-3-0-rgbw-5-24v-dc-tuya-9809731686). Cannot find much about that, it presents itself as TS0504A, and behaves similar to TS0505A (as far as I am able to test).

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -1302,7 +1302,6 @@ const devices = [
     },
     {
         fingerprint: [{modelID: 'TS0504A', manufacturerName: '_TZ3000_nzbm4ad4'}],
-        zigbeeModel: ['TS0504A'],
         model: 'TS0504A',
         vendor: 'TuYa',
         description: 'RGBW LED controller',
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -1300,6 +1300,14 @@ const devices = [
         description: 'Light controller',
         extend: preset.light_onoff_brightness_colortemp(),
     },
+    {
+        fingerprint: [{modelID: 'TS0504A', manufacturerName: '_TZ3000_nzbm4ad4'}],
+        zigbeeModel: ['TS0504A'],
```
Koenkk, line 1305: Please remove the zigbeeModel line and check if it still works.

milo1000: Yep, still working.

Koenkk:
Thanks!

PR #2112, opened 2021-01-19
Read batteryPercentageRemaining and batteryVoltage on reporting setup

sjorge:
This should result in us having battery available right after pairing!

See https://github.com/Koenkk/zigbee2mqtt/issues/5810#issuecomment-762694582

### DIFF
--- a/lib/reporting.js
+++ b/lib/reporting.js
@@ -48,10 +48,12 @@ module.exports = {
             'batteryPercentageRemaining', repInterval.HOUR, repInterval.MAX, 0, overrides,
         );
         await endpoint.configureReporting('genPowerCfg', p);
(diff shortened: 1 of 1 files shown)


sjorge:
I chose to modify lib/reporting as that seemed like the proper place to put it.
That way we also cover all current and future devices.

I also deliberately did not bump configureKey, having to wake up all devices, for what is essentially a noop configuration wise. Seems like a hassle to put people through, and in 99% of the case the battery value will have been reported if they paired more than 24h ago anyway.

sjorge:
I tested this by removing and pairing one of my ikea remotes, battery was properly ready during the configure step and available in the state output shortly after

Koenkk:
Thanks!

PR #2113, opened 2021-01-19
Limit color_temp range where known

sjorge:


sjorge:
This doesn't seem to work :( not sure why yet though.

lib/exposes.js
```diff
@@ -430,10 +430,10 @@ module.exports = {
         light_brightness: () => new Light().withBrightness(),
         light_brightness_color: () => new Light().withBrightness().withColor((['xy', 'hs'])),
         light_brightness_colorhs: () => new Light().withBrightness().withColor(['hs']),
-        light_brightness_colortemp: () => new Light().withBrightness().withColorTemp().withColorTempStartup(),
-        light_brightness_colortemp_color: () => new Light().withBrightness().withColorTemp().withColorTempStartup().withColor(['xy', 'hs']),
-        light_brightness_colortemp_colorhs: () => new Light().withBrightness().withColorTemp().withColorTempStartup().withColor(['hs']),
-        light_brightness_colortemp_colorxy: () => new Light().withBrightness().withColorTemp().withColorTempStartup().withColor(['xy']),
+        light_brightness_colortemp: (colorTempRange) => new Light().withBrightness().withColorTemp(colorTempRange).withColorTempStartup(colorTempRange),
```
sjorge, line 433: @Koenkk I assume it's failing here, because this is colorTempRange is not passed because this isn't a function as I first thought... any idea on how to cleanly pass these down ?

Koenkk: Why do you think this doesn't work? Looks ok here.

```
const devices = require('../devices');

const d = devices.find(d => d.model === '9290012573A');
console.log(d.exposes[0].features);
```

outputs

```
...
  Numeric {
    type: 'numeric',
    name: 'color_temp',
    property: 'color_temp',
    access: 7,
    unit: 'mired',
    value_min: 153,
    value_max: 500,
    description: 'Color temperature of this light'
  },
  Numeric {
    type: 'numeric',
    name: 'color_temp_startup',
    property: 'color_temp_startup',
    access: 7,
    unit: 'mired',
    value_min: 153,
    value_max: 500,
    description: 'Color temperature after cold power on of this light',
    presets: [ [Object] ]
  },
...
```

sjorge:
Hmm it was still at the old default in the frontend for me.

And when I added a console.log inside the expose function it returned undef.

~ sjorge

lib/exposes.js
sjorge: Tweaked my debug logs, it is indeed working.
Will do some more testing and mark this as Ready.

sjorge:
Added some Kelvin presets for the frontend too, I dropped 1000K... as that is way out of range for any bulb.

sjorge:
@Koenkk maybe we should switch to 3 presets for color_temp?
250, 370, 454 ? these are within range of all bulbs I've tested so far and seem to match ikea's cold, neutral, warm and the also align very close to Hue's neutral color which is 366 ?

Koenkk:
@sjorge I don't know about the preset yet. Currently this is not the same as the preset for `color_temp_startup` since the friendly name for this preset is accepted by the converter, `3000K` is not. I think this should be addressed in a different PR. For the color temp we can do something like `cool`, `warm` and `neutral` where cool is the min level, warm the max and neutral the average of min and max.

sjorge:
> @sjorge I don't know about the preset yet. Currently this is not the same as the preset for `color_temp_startup` since the friendly name for this preset is accepted by the converter (`previous`), `3000K` is not. I think this should be addressed in a different PR. For the color temp we can do something like `cool`, `warm` and `neutral` where cool is the min level, warm the max and neutral the average of min and max. But the converters should be updated to support this.

SGTM, dropped the commit.

Koenkk:
Thanks!

PR #2119, opened 2021-01-20
Add cool, neutral, and warm presets for color_temp(_startup)

sjorge:
I also removed the reverse mapping we had for color_temp_startup.
It didn't make sense for me for color_temp to have it report 248,249,'cool',251,252, ... so I brought color_temp_startup inline with this.

You can of course still set it by the preset name.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -691,7 +691,8 @@ const converters = {
     light_colortemp: {
         key: ['color_temp', 'color_temp_percent'],
         convertSet: async (entity, key, value, meta) => {
-            const preset = {'warm': 454, 'neutral': 370, 'cool': 250};
+            const [colorTempMin, colorTempMax] = light.findColorTempRange(entity, meta.logger);
+            const preset = {'warmest': colorTempMax, 'warm': 454, 'neutral': 370, 'cool': 250, 'coolest': colorTempMin};
 
             if (key === 'color_temp_percent') {
                 value = Number(value) * 3.46;
@@ -709,7 +710,6 @@ const converters = {
             value = Number(value);
 
             // ensure value within range
-            const [colorTempMin, colorTempMax] = light.findColorTempRange(entity, meta.logger);
             value = light.clampColorTemp(value, colorTempMin, colorTempMax, meta.logger);
 
             const payload = {colortemp: value, transtime: utils.getTransition(entity, key, meta).time};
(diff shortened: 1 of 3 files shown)


sjorge:
Works fine, tested with Trådfri FLOALT, Trådfri GU10, and Innr RB 250 C

converters/toZigbee.js
```diff
@@ -691,11 +691,21 @@ const converters = {
     light_colortemp: {
         key: ['color_temp', 'color_temp_percent'],
         convertSet: async (entity, key, value, meta) => {
+            const preset = {'warm': 454, 'neutral': 370, 'cool': 250};
```
Koenkk, line 694: Can we base the min/max on the actual color temp values provided by the bulb? I think the default should be according the default color temp range (150-500) `{'warm': 500, 'neutral': 325, 'cool': 150}`

Would be good to add an additional function in `light.js` which can do this.

Koenkk:
Pinging @nurikk here so he knows this is comming

converters/toZigbee.js
sjorge: I don't think that's a good idea, ideally they are the same across different bulbs.

If I set all mine to neutral, I'd expect them to be the same. (370 is ikea and Innr default)
cool and warm values I picked from ikea too and they are in range of all bulbs I have collected colorTempMin/Max data from.

Koenkk: I see, agree with that 

> However, maybe we can add 2 extra presets min and max ? (that would be bulb dependent?)

Good idea, but perhaps `warmest` and `coolest` is more self explanatory?

sjorge: OK, just did min/max, will rename them now... I like those names better.

sjorge:
Oops, I broke 'previous' let me fix that

sjorge:
All fixed and good to go, retested the Innr and Trådfri.

For Innr coolest and warmest map to min/max as expected. For Trådfri too, but they are the same as cool/warm for them.
I also added descriptions to make that clear (I hope).

lib/exposes.js
```diff
@@ -241,6 +247,11 @@ class Light extends Base {
         }
         this.features.push(new Numeric('color_temp_startup', access.ALL).withUnit('mired').withValueMin(range[0]).withValueMax(range[1])
             .withDescription('Color temperature after cold power on of this light')
+            .withPreset('coolest', range[0], 'Coolest temperature supported')
+            .withPreset('cool', 250, 'Cool temperature (250 mireds)')
```
nurikk, line 251: Not sure about mireds for description, entire my life I used Kelvins to describe light temperature

nurikk: Like on any light bulb box you'll see Kelvin not mired

sjorge: color_temp uses mireds though, and it's the exact value we pass.

nurikk: Yes, but in the title we can pass kelvin's 
Cool temperature (xxxxK)

sjorge: I find it kind of confusing if we list Kelvin here but the attribute required Mireds, but having both Kelvin and Mireds in the description seems good.

Koenkk:
Big thanks again @sjorge 🏅

PR #2122, opened 2021-01-21
Update devices.js for Trust devices

yiirus:
Hello. After joined Trust devices ZLL-ColorTempera and and ZLL-DimmableLigh i'm have unsupported devices. Analysis of the situation showed that in devices.js set vendor Trust, but devices get vendor Trust International B.V. I fix it and i'm have supported devices! Perhaps other devices Trust are returned to the manufacturer as well, but I do not have them and cannot verify them.


yiirus:
You're right

PR #2123, opened 2021-01-21
'operation_mode' control for Opple Switch 1/2 btn

uncle-fed:
Seems it was implemented for 3-button switch but got forgotten for 1 and 2 button version.


Koenkk:
Thanks!

PR #2124, opened 2021-01-21
Periodically update Tuya Thermostat with a time

ayufan:
It appears that Tuya thermostat pretty quickly
offsyncs it is internal clock resulting in a schedule
not executed properly.

This makes us to send a new updated time every hour
when a new notification from device is received (which do happen
pretty often).

### DIFF
--- a/lib/tuya.js
+++ b/lib/tuya.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const constants = require('./constants');
+const globalStore = require('./store');
 
 const dataTypes = {
     raw: 0, // [ bytes ]
@@ -84,8 +85,14 @@ async function onEventSetTime(type, data, device) {
 }
 
 // set UTC and Local Time as total number of seconds from 00: 00: 00 on January 01, 1970
+// force to update every device time every hour due to very poor clock
 async function onEventSetLocalTime(type, data, device) {
-    if (data.type === 'commandSetTimeRequest' && data.cluster === 'manuSpecificTuya') {
+    const nextLocalTimeUpdate = globalStore.getValue(device, 'nextLocalTimeUpdate');
+    const forceTimeUpdate = nextLocalTimeUpdate == null || nextLocalTimeUpdate < new Date().getTime();
+
+    if ((data.type === 'commandSetTimeRequest' && data.cluster === 'manuSpecificTuya') || forceTimeUpdate) {
+        globalStore.putValue(device, 'nextLocalTimeUpdate', new Date().getTime() + 3600 * 1000);
+
         try {
             const utcTime = Math.round(((new Date()).getTime()) / 1000);
             const localTime = utcTime - (new Date()).getTimezoneOffset() * 60;


ayufan:
This error message from CI is wrong, this identifier is an indicator of a `BigInt`:

```
/home/runner/work/zigbee-herdsman-converters/zigbee-herdsman-converters/lib/tuya.js
  91:71  error  Parsing error: Identifier directly after number
````

**Comment by Koenkk** [comment]
BigInt is not supported by the current eslint config, I also don't think it's needed and can be done easily with the Date.

Can you try with the following code:

```js
    const nextLocalTimeUpdate = globalStore.getValue(device, 'next_local_time_update');
    const forceTimeUpdate = nextLocalTimeUpdate == null || nextLocalTimeUpdate < new Date().getTime();

    if ((data.type === 'commandSetTimeRequest' && data.cluster === 'manuSpecificTuya') || forceTimeUpdate) {
        globalStore.putValue(device, 'next_local_time_update', new Date().getTime() + 3600000);

        try {
```

ayufan:
@Koenkk Maybe I'm wrong, but I decided to not use `Date` for a reason of time shift and this not being monotonic time :) Do you see this as a problem?

Koenkk:
The timeshift doesn't matter right? You just need to determine the relative time between two points so wether it's behind in timezones doesn't matter?

ayufan:
@Koenkk OK, I don't know JS too much, but this can be acceptable: https://stackoverflow.com/questions/8047616/get-a-utc-timestamp. It returns UTC time, but the timeshift can still happen if time correction is done because for ex. NTP. The only case where usage of `getTime()` would break if a date on a system would be going way too fast and be in the future, and later be corrected to a date in a past via NTP.

ayufan:
But, if we were to retain `process.hrtime.bigint()`, the `n` could be changed to `BigInt(value)` and achieve the same outcome. 

@Koenkk I will switch to `globalStore`, do you have preference of `hrtime` vs `Date()`?

Koenkk:
Both `n` and `BigInt` won't work (already tested) since these are not supported by the current eslint config so `Date()` is the only option here.

ayufan:
@Koenkk Thanks. Updated, and tested that it works on a smaller interval (60s).

Koenkk:
Thanks1

ayufan:
I looked at logs after night. It nicely send time update in a roughly one hour intervals.

ayufan:
I also look at the magnitude of timeshift:

After around 14h I see around 10m of timeshift on thermostat.

jvcsw:
Hi everyone, 

I think this problem is back with the latest version of z2m (1.22.2). In the last 2 days my thermostat (MOES) changes the time randomly. 

Is it possible? 

Thanks

Koenkk:
@jvcsw already fixed in the dev branch (#10489)

jvcsw:
Ok, thanks @Koenkk 

With the version 1.22.1 works fine, I'll wait for the new version 1.22.3.

PR #2126, opened 2021-01-21
update device local time on Moes room thermostat device

insipiens:
This PR will update the Moes Room thermostat device with the correct time. A function that was missing from the initial onboarding of the device to Z2m


insipiens:
seems a duplicate key, I will check again

PR #2127, opened 2021-01-22
Configure away preset temperature for Tuya

ayufan:
This allows to configure `away_preset_temperature` for `Tuya`.

This is already supported by `toZigbee: e.away_preset()`.


Koenkk:
Thanks!

PR #2132, opened 2021-01-23
Add multiEndpoint support to ias_contact_alarm_1 converter

nurikk:
Add multiEndpoint support to ias_contact_alarm_1 converter


Koenkk:
Thanks!

PR #2135, opened 2021-01-23
Add support for Tuya cover speed and Zemismart AM43

DomiStyle:
See https://github.com/Koenkk/zigbee2mqtt/issues/5596#issuecomment-765587234

Needs an update in the documentation too.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -1471,7 +1471,13 @@ const devices = [
         ],
         fromZigbee: [fz.tuya_cover, fz.ignore_basic_report],
         toZigbee: [tz.tuya_cover_control, tz.tuya_cover_options],
-        exposes: [e.cover_position().setAccess('position', ea.STATE_SET)],
+        exposes: [
+            e.cover_position().setAccess('position', ea.STATE_SET),
+            exposes.composite('options', 'options')
+                .withFeature(exposes.numeric('motor_speed', ea.STATE_SET)
+                    .withValueMin(0)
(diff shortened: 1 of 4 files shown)


devices.js
```diff
@@ -1466,6 +1467,7 @@ const devices = [
             {vendor: 'Moes', model: 'AM43-0.45/40-ES-EB'},
             {vendor: 'Larkkey', model: 'ZSTY-SM-1SRZG-EU'},
             {vendor: 'Zemismart', model: 'ZM25TQ', description: 'Tubular motor'},
+            {vendor: 'Zemismart', model: 'AM43', description: 'Roller blind motor'},
         ],
         fromZigbee: [fz.tuya_cover, fz.ignore_basic_report],
         toZigbee: [tz.tuya_cover_control, tz.tuya_cover_options],
```
Koenkk, line 1473: Can you add the following to `exposes` 
```
exposes.composite('options', 'options').withFeature(exposes.numeric('motor_speed', ea.STATE_SET).withValueMin(0).withValueMax(255).withDescription('Motor speed'))

```

After this you should be able to configure the motor speed from the frontend under the exposes tab (and docs wil be automatically updated with this on next release)

DomiStyle: @Koenkk Added it. I can't test on my curtain since it's not supported but it shows up in the UI.

Koenkk:
Thanks!

Issue #2129, opened 2021-01-22
Hue/sat gamma correction results in off primary colors


sjorge:
As #2017 got rather messy as I was figuring stuff out, one of the discovery will be documented here.

### Intro
When messing around with hue/sat, mostly when testing enhancedHue vs Hue I noticed colors when set where off, for the primary colors. They seem to be more washed out.

### Problem
Skipping https://github.com/Koenkk/zigbee-herdsman-converters/blob/cf864be3828ad542cbfcbee2c73d4687e4e43cad/converters/toZigbee.js#L897 seems to help for somewhat, but they still look a bit washed out. (Both with Innr and Tint, so I think it's a general thing and not a brand specific issue)

So I'm not convinced that is the issue, aside from that there are also some odd out of range values send from time to time.  
But the later is fixed by switching to `value.saturation = Math.min(255, hsv.s * (2.54));` as mentioned here https://github.com/Koenkk/zigbee-herdsman-converters/issues/2017#issuecomment-753666345

I wonder if we should only apply the gammeCorrection for enhancedHue and not Hue? I couldn't find much info on why we are doing this correction. Not that it will fix the washed out colors though.

It might be a brightness issue? as lowering the brightness to around 250 instead of 255 gives close to the same result with xy vs hs... which is interesting.

Koenkk:
The gamma correction was added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/1183 by @jasperro  . It seems this was specifically done for OSRAM and Hue. But if I understand @sjorge correctly this makes the color worse with Hue?

sjorge:
This can get confusing quickly... lets try to clarify from the get go.

**Hue** -> hue property
**Philips Hue** -> Ex. Philips, now Signify brand of bulbs.

It seems to make things worse for my Innr and Tint bulbs, my Philips Hue bulb which I have around for testing seems to be fine.

I'm seeing this when setting the colors by having Hue and Saturation in the color payload. Using XY I do not have the issue.

Koenkk:
Gamma correction is not applied for X/Y, only for Hue and Saturation. So I suggest to fix this by only doing the gamma correction when the bulb is OSRAM or Signify Hue

sjorge:
Sounds good, let me think on how to do this cleanly. 

As it gets passed deep into utils.rgbToXY too...

sjorge:
Oh no :( I was doing more testing as I also have an Osram bulb around.

[...]

PR #2139, opened 2021-01-23
[WIP] Split Müller Licht Tint 404005 from 404012/404000

sjorge:
The Müller Licht Tint 404005 (GU10) does not support enhancedHue. The MLI Tint 404000 (E27) does (No E14 to test)

Switching to fingreprints allows us to differentiate between them, we also expose Hue/Sat support now.

image
image

Both are now recognizes correctly, the trick was switching both to fingerprint and adding the 2nd endpoint for the GU10.
I also verified frontend could set XY and HS on both, it now works as expected.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -758,6 +758,7 @@ const converters = {
 
             // Set correct meta.mapped for groups
             meta = utils.updateGroupMetaMapped(entity, meta);
+            const metaMapped = utils.getOptions(meta.mapped, entity);
 
             if (value.hasOwnProperty('r') && value.hasOwnProperty('g') && value.hasOwnProperty('b')) {
                 const xy = utils.rgbToXY(value.r, value.g, value.b);
@@ -779,7 +780,7 @@ const converters = {
                 value.brightness = hsv.v * (2.54);
                 newState.brightness = value.brightness;
 
-                if (meta.mapped && meta.mapped.meta && meta.mapped.meta.enhancedHue === false) {
+                if (metaMapped && metaMapped.enhancedHue === false) {
                     value.hue = Math.round(hsv.h / 360 * 254);
                     command = 'moveToHueAndSaturationAndBrightness';
                 } else {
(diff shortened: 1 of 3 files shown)


Koenkk:
As discussed in https://github.com/Koenkk/zigbee-herdsman-converters/issues/2129#issuecomment-766102502 this is likely not gonna work well. There is a very high chance that future version of the 404012 or 404000 ship with this newer firmware, then the model doesn't match anymore. Since the only difference is the enhancedHue the solution from https://github.com/Koenkk/zigbee-herdsman-converters/issues/2129#issuecomment-766102502 is better.

sjorge:
I'm not sure how doing that would be better, as I don't think differentiating on firmware is going to work.
Can we do the full endpoint match for 
```
            {ID: 1, profileID: 260, deviceID: 269, inputClusters: [0, 3, 4, 5, 6, 8, 768, 2821, 4096, 65244], outputClusters: [10, 25]},
            {ID: 242, profileID: 41440, deviceID: 97, inputClusters: [], outputClusters: [33]},
```

In the function?

Koenkk:
Yes, the passed device provides the full device with all endpoints. Why will differentiating based on the firmware version won't work?

sjorge:
I think the difference is based on the hardware and all of those bulbs could have a firmware with the same version. 

My guess is that checking the clusters to see if haDiagnostics is present would be better. I think the GU10 (at least mine that doesn’t do ehs) might ship with a unstable development firmware. If it shipped with a final version I would expect the haDiagnostics cluster not to be there, like on the E27. (Older firmware verison)

~ sjorge

sjorge:
I hit an issue I think, looks like the function in meta gets ignored by merging the meta object later to apply the extend :(
getOptions does not see `enhancedHue` when it's a function.

sjorge:
I think I need some help after all, I can't seem to to get it to work.
I pushed what I have now, this is on top of #2140 as it looks like the getOptions need to be ordered after the new updateGroupMetaMapped  function, it's seems to always still be unset as it is always using the enhancedHue variant.

sjorge:
A weird side effect, it also seems to trigger a configure log message the first time getOptions is called.

sjorge:
Going to close this, I'm about to destroy the bulb with my current implementation that is not working at all. I will start completely fresh on this this weekend after #2140 is merged and my urge to destroy the bulb passes.

Issue #2128, opened 2021-01-22
Inconsistent colors with set/scene_add vs set/color


sjorge:
As #2017 got rather messy as I was figuring stuff out, one of the discovery will be documented here.

### Intro
I frequently use red colors for my lights, sometimes the colors were off. As I have a mix of Innr and Tint in active use (Hue and Osram for testing). Red is a tricky one for Innr but there is a work around, I initially thought it had to do something with that.

As I was testing with a mix of xy and hs at the time (mostly in relation to scene_add). But the in general it looks like colors are slightly different with set/color vs set/scene_add + set/scene_recall.

### Problem 1: applyRedFix
One of the differences is that scene_add does not do anything with applyRedFix, and Innr bulbs thus ignore `{x: 0.701, y: 0.299}` with using scene_add.

We should also do this check and adjust accordingly for in scene_add

### Problem 2: colors are different with set/scene_add vs set/color
Below is an example of my Tint GU10, it shows the individual leds so its easy to see.

Result of:
```
[root@amethyst /opt/zigbee2mqtt]# mosquitto_pub -t zigbee2mqtt/livingroom/light/annex/bulb/set/scene_add -m '{"ID": 123, "color": {"x": 0.701, "y": 0.299}}'
[root@amethyst /opt/zigbee2mqtt]# mosquitto_pub -t zigbee2mqtt/livingroom/light/annex/bulb/set/scene_recall -m 123
```
![image](https://user-images.githubusercontent.com/379665/105538293-797d5f80-5cf3-11eb-91b1-6633a01a263c.png)

This looks good, clean red as I expect it.

Result of:
```
[root@amethyst /opt/zigbee2mqtt]# mosquitto_pub -t zigbee2mqtt/livingroom/light/annex/bulb/set/state -m 'ON'
[root@amethyst /opt/zigbee2mqtt]# mosquitto_pub -t zigbee2mqtt/livingroom/light/annex/bulb/set/color -m '{"x": 0.701, "y": 0.299}'
```

![image](https://user-images.githubusercontent.com/379665/105538512-c6f9cc80-5cf3-11eb-9bba-984b0bda114a.png)

This is blue mixed in and is not correct. (Not 100% tracked this one down, I was initially also testing with hs which passes through utils.gammaCorrectHSV() which definitely does funky stuff as `{"hue": 0, "saturation": 100}` gets correct and when converting the correction back it is definitely not red, and doesn't look red either.)

[...]

Issue #2037, opened 2021-01-06
Adventures with ColorCapabilities


sjorge:
I'm creating this to track info/process in my advendtures with colorCapabilities...

- [x] read ColorCapabilities for all bulbs
- [x] collect data, check if they lie... sadly they do 👎 
- [ ] automatically do the right thing in the color convertor based on value

## read ColorCapabilities for all bulbs
This was easy, done as part of reading ColorTemperaturePhysicalMinMireds and ColorTemperaturePhysicalMaxMireds change. #2010

## collect data
Reported data

- IKEA Trådfri CWS: xy
- IKEA Tradfri WS: ct
- Innr RB 250 C: hs, **ehs**, xy, ct
- Innr RB 285 C: hs, ehs, xy, ct
- Muller Light Tint: hs, **ehs**, xy, ct
- IKEA FLOALT: ct
- IKEA GUNNARP: ct
- Hue E14: hs, ehs, xy, ct
- Osram E27: hs, ehs, xy, ct

This is bad news, as both the Innr RB 250 C, and Tint, report support for enhancedHue... which neither seem to accept.
The neither are also not rejecting it with UNSUPPORT_ATTRIBUTE, which means either: busted firmware or we maybe we're writing values that are out of range. I have not dug out wireshark yet to dig deep.

OK, they actually do respond!
<img width="699" alt="image" src="https://user-images.githubusercontent.com/379665/103805763-53e62a00-5054-11eb-89ee-c8f3aa6a537c.png">

<img width="666" alt="image" src="https://user-images.githubusercontent.com/379665/103805448-e5a16780-5053-11eb-8a5f-2858f8149e35.png">

But why would they announce support if they, in fact, do not support it... this is troublesome.

## automatically do the right thing in the color converter based on value
Looks like we can differentiate between xy and ct capable bulbs with this date...

Hue/Saturation is tricky... but something like this can work:

| bit | support  | converter |
| :-- | :--  | :-- |
| 0 | hs | enable hue/sat support in convertor |
| 1  | eh | enable enhancedHue if on manufacturer whitelist or enhancedHue == true |
| 3 | xy | enable xy |
| 4 | ct | enable ct |

manufacturer list seems to be:
- Hue
- Osram

Tint doesn't seem to work at all
Innr mixed bag of some bulbs working other not, so we should err on the side of not auto enabling.

### raw data dump
| device | lightingColorCtrl |
| --- | --- |
| **RGB(W)** ||
[...]

PR #2140, opened 2021-01-23
Fallback to XY when calling scene_add with hue/sat when bulb does not support enhancedHue

sjorge:
While retesting applyRexFix for #2128 which looks like it is not needed when using scene_add (scenes certainly different on a lot of fronts 😞 ).

But while testing I was comparing XY and HS I noticed the same odd bulb crashing behavior discovered here https://github.com/Koenkk/zigbee-herdsman-converters/issues/2037#issuecomment-755516717 with bulbs that lie about enhancedHue support (hello Innr RB 250 C and Tint GU10).

```
[root@amethyst /opt/zigbee2mqtt]# mosquitto_pub -t zigbee2mqtt/office/light/desk/set/scene_add -m '{"ID": 123, "color": {"x": 0.701, "y": 0.299}}'
[root@amethyst /opt/zigbee2mqtt]# mosquitto_pub -t zigbee2mqtt/office/light/desk/set/scene_recall -m 123
[root@amethyst /opt/zigbee2mqtt]# mosquitto_pub -t zigbee2mqtt/office/light/desk/set/scene_add -m '{"ID": 124, "color": {"hue": 0, "saturation": 100}}'
[root@amethyst /opt/zigbee2mqtt]# mosquitto_pub -t zigbee2mqtt/office/light/desk/set/scene_recall -m 124
```

Now that with #2139 we correctly set enhancedHue=false for tint, I added a fallback. Both hs or xy in scene_add now result in the same result* for Innr RB 250 C, Tint GU10, Innr RB 285 C, Tint E27, Osram E27.

*objectively the same, the bulbs turn red/green/blue respectively when setting to the prime colors, Tint GU10 in particular still produces inconsistent results between XY and HS when **not** using scenes but are now the same when using scenes (which makes sense as both are now XY for scenes).

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -760,18 +760,7 @@ const converters = {
             const newState = {};
 
             // Set correct meta.mapped for groups
-            // * all definition metas are the same -> copy meta.mapped[0]
-            // * mixed device models -> meta.mapped = null
-            if (entity.constructor.name === 'Group' && entity.members.length > 0) {
-                for (const memberMeta of Object.values(meta.mapped)) {
-                    // check all members are the same device
-                    if (meta.mapped[0] != memberMeta) {
-                        meta.mapped = [null];
-                        break;
-                    }
-                }
-                meta.mapped = meta.mapped[0];
-            }
+            utils.updateGroupMetaMapped(entity, meta);
 
             if (value.hasOwnProperty('r') && value.hasOwnProperty('g') && value.hasOwnProperty('b')) {
                 const xy = utils.rgbToXY(value.r, value.g, value.b);
@@ -3735,6 +3724,9 @@ const converters = {
             const scenename = '';
             const transtime = value.hasOwnProperty('transition') ? value.transition : 0;
 
+            // Set correct meta.mapped for groups
+            utils.updateGroupMetaMapped(entity, meta);
+
             const state = {};
             const extensionfieldsets = [];
             for (let [attribute, val] of Object.entries(value)) {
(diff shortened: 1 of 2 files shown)


converters/toZigbee.js
```diff
@@ -3769,14 +3783,28 @@ const converters = {
...
                     } else if (color.hasOwnProperty('hue') && color.hasOwnProperty('saturation')) {
-                        const hsv = utils.gammaCorrectHSV(utils.correctHue(color.hue, meta), color.saturation, 100);
-                        extensionfieldsets.push(
-                            {
-                                'clstId': 768,
-                                'len': 13,
-                                'extField': [0, 0, (hsv.h % 360 * (65535 / 360)), (hsv.s * (2.54)), 0, 0, 0, 0],
-                            },
-                        );
+                        if (meta.mapped && meta.mapped.meta && meta.mapped.meta.enhancedHue === false) {
```
Koenkk, line 3786: I think we should not allow this since this behaviour is very implicit. Throwing an error with a message like: "hue/saturation not supported, use XY" makes more sense I think.

sjorge:
Makes for more complex logic externally though.

If you can set it via set/color, I’d expect it to work via set/scene_add too. It’s just unfortunate scenes don’t support the non enhancedHue variant.

~ sjorge

sjorge:
Which similarly in set/color we also do different things for hue/sat depending on what is supported. We map it to either enhancedHue or normal hue and there corresponding command.

We’re already doing color_temp -> xy mapping in scene_add too. If we throw and error here, we should probably also not do the mapping for color_temp in scene_add either.

~ sjorge

converters/toZigbee.js
```diff
@@ -3723,6 +3723,20 @@ const converters = {
             const scenename = '';
             const transtime = value.hasOwnProperty('transition') ? value.transition : 0;
 
+            // Set correct meta.mapped for groups
+            // * all definition metas are the same -> copy meta.mapped[0]
+            // * mixed device models -> meta.mapped = null
+            if (isGroup && entity.members.length > 0) {
```
Koenkk, line 3729: This code is a dupe of https://github.com/Koenkk/zigbee-herdsman-converters/blob/f8226da029df72fa82c3514284bfcc553dec6741/converters/toZigbee.js#L762, can you create a function out of this in utils.js?

Koenkk:
> Which similarly in set/color we also do different things for hue/sat depending on what is supported. We map it to either enhancedHue or normal hue and there corresponding command. We’re already doing color_temp -> xy mapping in scene_add too. If we throw and error here, we should probably also not do the mapping for color_temp in scene_add either. ~ sjorge
> […](#)
> On 23 Jan 2021, at 21:53, Jorge Schrauwen ***@***.***> wrote: ﻿Makes for more complex logic externally though. If you can set it via set/color, I’d expect it to work via set/scene_add too. It’s just unfortunate scenes don’t support the non enhancedHue variant. ~ sjorge >> On 23 Jan 2021, at 21:41, Koen Kanters ***@***.***> wrote: >> > ﻿ > @Koenkk commented on this pull request. > > In converters/toZigbee.js: > > > @@ -3769,14 +3783,28 @@ const converters = { > ); > state['color'] = {x: color.x, y: color.y}; > } else if (color.hasOwnProperty('hue') && color.hasOwnProperty('saturation')) { > - const hsv = utils.gammaCorrectHSV(utils.correctHue(color.hue, meta), color.saturation, 100); > - extensionfieldsets.push( > - { > - 'clstId': 768, > - 'len': 13, > - 'extField': [0, 0, (hsv.h % 360 * (65535 / 360)), (hsv.s * (2.54)), 0, 0, 0, 0], > - }, > - ); > + if (meta.mapped && meta.mapped.meta && meta.mapped.meta.enhancedHue === false) { > I think we should not allow this since this behaviour is very implicit. Throwing an error with a message like: "hue/saturation not supported, use XY" makes more sense I think. > > — > You are receiving this because you authored the thread. > Reply to this email directly, view it on GitHub, or unsubscribe.

Agree, this is fine

converters/toZigbee.js
sjorge: Indeed it is, seems reasonable to abstract that. I'll try to do it tomorrow.

sjorge: Still needs testing, but can't move the PR back to draft it seems

sjorge:
Done, retested and still works (tested against both Innr RB 250 C and Tint GU10)

sjorge:
OK I think part of my struggler is that I am testing this in combination with the tint split. I'm going to just test with the Innr RB 250 C and take the tint split code out of the equation for now.

sjorge:
I added some console.logs and this seems to work just fine.

```
# Innr RB 250 C
XXX: scene_add -> hs2xy
# Osram bulb with enhancedHue support
XXX: scene_add -> hs
# Group with only Innr RB 250 C
XXX: scene_add -> hs2xy
# Group with only Osram bulb with enhancedHue support
XXX: scene_add -> hs
```

All my problems last night stemmed from my extra changes to split the tint bulbs, I was about to chuck these out the window. I will look at that again but probably not until the weekend because else i will destroy the bulb out of frustration.

However, this PR good to go, as far as I am concerned.

Koenkk:
Thanks!

PR #2144, opened 2021-01-24
Update exposes const documentation

nurikk:
Current implementation with 1/2/4/7 numbers is rather voodoo magic than easily understandable code :)


Koenkk:
Thanks!

PR #2146, opened 2021-01-24
Huesat colortemp

Koenkk:
https://github.com/Koenkk/zigbee2mqtt/issues/5875

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -215,6 +215,12 @@ function gammaCorrectRGB(r, g, b) {
     return {r: Math.round(r*255), g: Math.round(g*255), b: Math.round(b*255)};
 }
 
+function hueSaturationToMireds(hue, saturation) {
+    const rgb = hsvToRGB(hue, saturation, 100);
+    const xy = rgbToXY(rgb.r, rgb.g, rgb.b);
+    return xyToMireds(xy.x, xy.y);
+}
+
 function hsvToRGB(h, s, v) {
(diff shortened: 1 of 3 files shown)


lib/utils.js
```diff
@@ -203,6 +203,56 @@ function rgbToXY(red, green, blue) {
...
+        red = red / green;
+        blue = blue / green;
+        green = 1.0;
+    } else if (blue > red && blue > green && blue > 1.0) {
+        red = red / blue;
+        green = green / blue;
+        blue = 1.0;
+    }
+
+    // Reverse gamma correction
```
sjorge, line 237: Should this not also use the fancy hue variant done by gammaCorrectRGB ?

Koenkk: Which variant do you mean?

sjorge: https://github.com/Koenkk/zigbee-herdsman-converters/blob/d751f07919cc1122da3f0ac75053a7ba0cfe137b/lib/utils.js#L184 was based on the Signify Hue dev docs right... should we move this to a seperate reverse function too reverse this correction?

Koenkk: Ah didn't know it was hue, but seems so: https://github.com/usolved/cie-rgb-converter/blob/79a9eb81abfd8126a94ff7bee69fc8daaf79d6e1/cie_rgb_converter.js#L4

But this should be the reverse of it (both functions are taken from `cie_rgb_converter.js`)

PR #2147, opened 2021-01-24
Touch up to make temperature consistent with other TRV (Tuya)

insipiens:
touch up to make temperature values consistent with other TRV reporting - as float (not string)

This helps with reporting data via influxdb which requires values sent

Tested on my docker container


Koenkk:
Thanks!

PR #2149, opened 2021-01-25
Adding controls for the behavior when the device is powered on

alphazulu:
Adding controls for the behavior when the device is powered on
For Moes MS-104BZ

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -60,14 +60,14 @@ const converters = {
         },
     },
     moes_power_on_behavior: {
-        key: ['moes_power_on_behavior'],
+        key: ['power_on_behavior'],
         convertSet: async (entity, key, value, meta) => {
             value = value.toLowerCase();
             const lookup = {'off': 0, 'on': 1, 'previous': 2};
             utils.validateValue(value, Object.keys(lookup));
             const pState = lookup[value];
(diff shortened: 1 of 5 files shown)


converters/toZigbee.js
```diff
@@ -59,6 +59,20 @@ const converters = {
             await entity.read('genOnOff', ['startUpOnOff']);
         },
     },
+    moes_power_on_behavior: {
+        key: ['moes_power_on_behavior'],
```
Koenkk, line 63: Please name it `power_on_behavior` instead of `moes_power_on_behavior`

alphazulu: But there is already power_on_behavior, with  "const lookup = {'off': 0, 'on': 1, 'toggle': 2, 'previous': 255};"
And for Moes  MS-104BZ you needs"const lookup = {'off': 0, 'on': 1, 'previous': 2};"

Koenkk: I know, I didn't mean changing the name of the converter but they `key` and property in `fromZigbee`: https://github.com/Koenkk/zigbee-herdsman-converters/pull/2149/files#diff-b676ab0df4d49b8572967cb4c92a2dc6264091131f6b1828ca13d0f9656e367aR657

PR #2150, opened 2021-01-25
Add state to Tuya covers

DomiStyle:
See https://github.com/Koenkk/zigbee2mqtt/issues/5596#issuecomment-765632339

Tested and shows up in the UI properly now once it arrives at the set location.


Koenkk:
Thanks!

PR #2153, opened 2021-01-25
Added new device "LEDVANCE AC08560"

snueffel88:


snueffel88:
@Koenkk , just saw that's not possible to add this device to zigbee2mqtt.io since there is already an Osram device with the same model id ('AC08560') and a AC08560.md for this reason, too. However, Zigbee2Mqtt did not recognize it since it is announced with the vendor name "LEDVANCE". Should it be added nevertheless?

Koenkk:
Thanks, renamed the existing `AC08560` to `AC08560-DIM`, so `AC08560` is free to use.

PR #2165, opened 2021-01-27
Converter definition runtime validation

nurikk:
Validate converter, since we're exposing addDeviceDefinition for external converters, there are cases when users can add converter with invalid structure and crash frontend

example: https://github.com/nurikk/z2m-frontend/issues/279


Koenkk:
Thanks!

PR #2166, opened 2021-01-27
Workaround for broken illuminance implementation in RTCGQ11LM, see ht…

rikvb:
Workaround for broken illuminance implementation in RTCGQ11LM, see https://community.openhab.org/t/correctly-calculate-lux-values-at-xiaomi-aqara-motion-sensors-illuminance-channel/98603

I own several RTCGQ11LM and without this fix the lux value does not exceed 10, even during broad daylight.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3596,7 +3596,7 @@ const converters = {
             if (msg.data['65281']) {
                 const result = {};
                 if (msg.data['65281'].hasOwnProperty('11')) {
-                    const illuminance = msg.data['65281']['11'];
+                    const illuminance = 10000 * Math.log10(msg.data['65281']['11'] + 1);
                     // DEPRECATED: remove illuminance_lux here.
                     result.illuminance = calibrateAndPrecisionRoundOptions(illuminance, options, 'illuminance');
                     result.illuminance_lux = calibrateAndPrecisionRoundOptions(illuminance, options, 'illuminance_lux');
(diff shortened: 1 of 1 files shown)


Koenkk:
This does not seem to apply to all devices. E.g. mine reports correctly:

image

rikvb:
Interesting, as I just assumed it was an implementation error in the device. Wouldn't make sense either that they created a new batch (without changing the device ID) and made this kind of error.

Not sure how dark your hall is, but a top of ~110 is still very dark. I would expect, if it has any windows, to see 5-digit values there in the middle of the day. Even if it doesn't, with somewhat bright lights turned on there I would expect several hundreds. Do you own any others that are near windows?

Koenkk:
I can try by putting one close to a window but unfortunately it's rainy/dark today.

I think I understand why this conversion is needed for openhab. According to the spec the send illuminance value can be converted to lux with the following formula: `illuminanceLux = Math.pow(10, (illuminance - 1) / 10000)`.  For Zigbee2MQTT this is done here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/f69551636d1fc5d25fe4c9c34f356e7c9587614f/converters/fromZigbee.js#L329

The RTCGQ11LM does not follow the spec and the send illuminance is already in lux (so `Math.pow(10, (illuminance - 1) / 10000)` shouldn't be applied again. I expect openhab to do this so the conversion is done twice. This is not the case for Zigbee2MQTT since the value of the sensor is not converted again:  https://github.com/Koenkk/zigbee-herdsman-converters/blob/f69551636d1fc5d25fe4c9c34f356e7c9587614f/converters/fromZigbee.js#L3654

PR #2167, opened 2021-01-27
Cleanup GLEDOPTO products, and add OTA support for pro models

washcroft:
This PR cleans the naming of GLEDOPTO products - making them consistent and in some cases more useful.

Also adds OTA support for pro models (see https://github.com/Koenkk/zigbee-OTA/pull/19).


lib/ota/zigbeeOTA.js
```diff
@@ -8,10 +8,13 @@ const axios = common.getAxios();
  */
 
 async function getImageMeta(current, logger, device) {
+    const modelId = device.modelID;
     const imageType = current.imageType;
     const manufacturerCode = current.manufacturerCode;
     const images = (await axios.get(url)).data;
-    const image = images.find((i) => i.imageType === imageType && i.manufacturerCode === manufacturerCode);
+    const image = images.find((i) => i.imageType === imageType && i.manufacturerCode === manufacturerCode &&
+        (!i.modelId || i.modelId === modelId));
```
Koenkk, line 16: Why was the modelId added here, according to the Zigbee spec image can be chosen based on the `manufacturerCode`and `imageType`.

washcroft: To facilitate easier use of the generic OTA update mechanism based on the index.json meta file in the zigbee-OTA repo, `manufacturerCode` and `imageType` isn't enough (see my PR in that repo).

TheJulianJES: I also have some older GLEDOPTO devices which have ``manufacturerCode = 0`. I guess that could further complicate things.

washcroft: Only "V_1_0" onwards of their new Pro series products supports OTA, older stuff won't even bother checking if there is updated firmware.

TheJulianJES: Yeah, I have the entire series of Pro controllers. They check on their own very frequently. The are all "V_1_0_0" but have different "current_file_version" when they checkin if there's an update.
And some ask for an image with imageType=0 and manufacturerCode=0. Others only have imageType=0.

washcroft: Interesting, the only alternative I guess is to ignore `imageType` and `manufacturerCode` entirely, and just go off the known device `modelId`. I think once everything is running a newer firmware, then they'll all be sending the correct manufacturerCode. Just be thankful they finally added an OTA mechanism I guess :D

Koenkk:
Thanks!

jd1900:
Where are the Gledopto OTA files taken from?
I don't see the proper code for that.
Is there a Gledopto PRO OTA repository?

washcroft:
> Where are the Gledopto OTA files taken from?
> I don't see the proper code for that.
> Is there a Gledopto PRO OTA repository?

Supplied by GLEDOPTO

PR #2171, opened 2021-01-27
Konke smoke detector

senarvi:
This is the first time I'm using zigbee2mqtt. My other Zigbee devices were detected automatically, but there was no configuration for my smoke detector. This is the device:

http://www.ikonke.us/pro/smokeDetector/
https://www.aliexpress.com/item/4000478377409.html?spm=2114.12010615.8148356.2.54e97f42Ak8zH2

I created this entry based on the instructions and the existing examples. It detects smoke when I press the test button on the smoke detector. I wasn't able to get other information like battery status reading. I don't see any "No converter available" messages in the log. I don't know if there's anything else I could try to check whether the entry is complete.

The bottom of the device says:
YW-S1
Photoelectric Smoke Detector

The Aliexpress page says "Product Model: KK-SA-H03".

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -12000,7 +12000,7 @@ const devices = [
         vendor: 'Konke',
         fromZigbee: [fz.ias_smoke_alarm_1],
         toZigbee: [],
-        exposes: [e.smoke(), e.battery_low(), e.tamper()],
+        exposes: [e.smoke(), e.battery_low()],
     },
     {
         zigbeeModel: ['3AFE130104020015', '3AFE270104020015', '3AFE280104020015'],
--- a/devices.js
+++ b/devices.js
@@ -11994,6 +11994,15 @@ const devices = [
         exposes: [e.temperature(), e.humidity(), e.battery()],
     },
     {
+        zigbeeModel: ['3AFE010104020028'],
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -11993,6 +11993,15 @@ const devices = [
...
         exposes: [e.temperature(), e.humidity(), e.battery()],
     },
+    {
+        zigbeeModel: ['3AFE010104020028'],
+        model: 'TW-S1',
+        description: 'Photoelectric smoke detector',
+        vendor: 'Konke',
+        fromZigbee: [fz.ias_smoke_alarm_1],
+        toZigbee: [],
+        exposes: [e.smoke()],
```
Koenkk, line 12003: Also add `e.battery_low(), e.tamper()` here, if it works battery low should be false (assuming your battery is new), tamper can be triggered by removing the back cover probably.

senarvi: Done. batter_low is false, and the battery is new. I wasn't able to trigger the tamper signal by removing the back cover though. The device looks extremely simple and I don't know how it could detect that.

Koenkk: Thanks, removed the tamper stuff.

PR #2172, opened 2021-01-28
Added Iluminize 5120.1100

Sacred-Shadow:
Iluminize has reissued its dimming actuator.
Technically & the product name is the same as the model 511.201.

Tested and working. (on/off & brightness)


Koenkk:
Thanks!

PR #2173, opened 2021-01-28
Set enhancedHue=false for MLI Tint GU10

sjorge:
3rd attempt at fixing this bulb.
- [x] drop `updateGroupMetaMapped` as we have `getMetaValue` which is better
- [x] `getMetaValue` should expand functions in meta
- [x]  target haDiagnostic cluster to id GU10 bulb

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -759,8 +759,6 @@ const converters = {
             let command;
             const newState = {};
 
-            // Set correct meta.mapped for groups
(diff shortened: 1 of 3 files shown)


Koenkk:
Thanks!

sjorge:
I wonder if there are any other places were we should switch to getMetaValue... as it should now handle groups and function meta properties.

Koenkk:
Probably for all converters which can be used with groups.

PR #2175, opened 2021-01-28
Switch toZigbee to getMetaValue

sjorge:
As per https://github.com/Koenkk/zigbee-herdsman-converters/pull/2173#issuecomment-769302099 I had a look around, I think these are all the others. I don't own any device that uses those though, so I couldn't test these changes with hardware.


Koenkk:
Thanks 😄  🏅

PR #2177, opened 2021-01-28
Support Innr BF 265 (B22 version of RF 265)

intrinseca:
Clone the RF 265 as the BF 265

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -5448,7 +5448,7 @@ const devices = [
         vendor: 'Innr',
         description: 'B22 filament bulb dimmable',
         extend: preset.light_onoff_brightness(),
-        meta: {applyRedFix: true, turnsOffAtBrightness1: true},
+        meta: {turnsOffAtBrightness1: true},
     },
     {
         zigbeeModel: ['RB 185 C'],
@@ -5481,7 +5481,7 @@ const devices = [
         vendor: 'Innr',
         description: 'E27 bulb',
         extend: preset.light_onoff_brightness(),
-        meta: {applyRedFix: true, turnsOffAtBrightness1: true},
+        meta: {turnsOffAtBrightness1: true},
     },
     {
         zigbeeModel: ['RF 265'],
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -5491,6 +5491,14 @@ const devices = [
         extend: preset.light_onoff_brightness(),
         meta: {applyRedFix: true, turnsOffAtBrightness1: true},
     },
+    {
+        zigbeeModel: ['BF 265'],
+        model: 'BF 265',
+        vendor: 'Innr',
+        description: 'B22 bulb filament clear',
+        extend: preset.light_onoff_brightness(),
+        meta: {applyRedFix: true, turnsOffAtBrightness1: true},
```
sjorge, line 5500: You marked the bulb as onoff_brightness, so you don't need the applyRedFix here.

intrinseca: I just cloned the RF 265 - should it be removed from there (and similar bulbs?) as well?

sjorge: At least for the following, they all use `preset.light_onoff_brightness()` which does not add the light_color convertor which uses that meta option.
- BF 263
- RB 265
- RF 265
- RB 165
- RB 162
- RB 175 W
- RS 122
- RS 125
- RS 225
- RS 226
- RB 145
- RB 245
- RF 261
- RF 263
- RF 264
- BY 165
- BY 265
- PL 110
- PL 115
- ST 110
- UC 110
- DL 110 N
- DL 110 W
- SL 110 N
- SL 110 M
- SL 110 W
- AE 260
- BE 220

Not 100% certain, but I also think these:
- RB 278 T
- RB 178 T
- RS 128 T
- RS 228 T
- RS 229 T
- RB 248 T
- RB 148 T

As far as I can see only the 'C' modules that do color XY need the meta.applyRedFix.

intrinseca: Are you happy to remove this from (at least) the first set even though I don't have access to the devices to test? If so, I will update the PR.

sjorge: Sounds good to me,m

intrinseca: @sjorge ready for review

Koenkk:
Thanks!

PR #2185, opened 2021-01-31
support Bitron zigbeeModel 902010/28. 

ernstkl:
This device is similar to the ../25 but without the metering functionality. Tested by me.


Koenkk:
Thanks!

PR #2191, opened 2021-02-02
Livolo support: Adding dimmer device

whitepail:
Adding Livolo dimmer device support (according to Koenkk/zigbee2mqtt#5516)


natoriousbigg:
@whitepail Your code for adding the dimmer works, except for the ON/OFF state doesn't appear to be working. Can you help fix that?

PR #2192, opened 2021-02-02
FIX #2190: getMetaValue should allow passing default

### LINKED ISSUE #2190: currentHue and currentSaturation not querried anymore with /get
https://github.com/Koenkk/zigbee-herdsman-converters/blob/3ed4ad4430dd7ddc333ceb331aef3faeeea20a91/converters/toZigbee.js#L995

Looks like now we're no longer querying currentHue and currentSaturation for any bulbs?

As far as I can see only for one device do we have `meta.supportsHueAndSaturation` set to false, but the no code seems to also assume `undefined` as not false too so we never take the branch to push the extra attributes. Can we change this back to `if (utils.getMetaValue(entity, meta.mapped, 'supportsHueAndSaturation', 'allEqual') !== false) {` ?

@Koenkk you recently changed it after my first refactor to getMetaValue, I assume you had a reason to? But anyway, as far as I can tell we never hit the branch so we never query these properties now. So we need to either just drop the if, if that is the intented behavior or change it back I think. (ideally I think we do read these)

sjorge:
- getMetaValue now takes a 'defaultValue'
- added an explicit default for most `meta` properties found in devices.js

Exception being the TuYa stuff as I don't fully understand how it's used but `enhancedHue`, `supportsHueAndSaturation`, `coverInverted`, `turnsOffAtBrightness1`, and `applyRedFix` were pretty obvious. This also allowed us to get rid of harder to read non-default for cases like we had with enhanchedHue and supportsHueAndSaturation where the default is true.

Small bonus, looks like get/color now also works for mixed groups!

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -163,7 +163,7 @@ const converters = {
                 value = lookup[value];
             }
 
-            const invert = utils.getMetaValue(entity, meta.mapped, 'coverInverted') ?
+            const invert = utils.getMetaValue(entity, meta.mapped, 'coverInverted', 'allEqual', false) ?
                 !meta.options.invert_cover : meta.options.invert_cover;
             const position = invert ? 100 - value : value;
             await entity.command(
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #2178, opened 2021-01-29
Siterwell GS361A-H04, system_mode and Home Assistant #5968

piars777:
This PR fixes the issue described here https://github.com/Koenkk/zigbee2mqtt/issues/5968

PR #2194, opened 2021-02-02
Siterwell GS361A-H04, system_mode and Home Assistant #5968

creckord:
This is an update to PR #2178. It fixes the issue described here Koenkk/zigbee2mqtt#5968

In addition to the original PR, it introduces a new `thermostatSystemModes4` with the Siterwell-specific modes and removes the problematic `force` value from the received state.


creckord:
My Javascript foo is too weak apparently :(

I would have loved to change the `siterwell_thermostat` converter function to delegate to `tuya_thermostat` except for the `case tuya.dataPoints.mode` to avoid all that copy and paste code. But I couldn't figure out how to reference it from there.

Koenkk:
Since the only change is:

image

I suggest to add the following line to `tuya_thermostat`:

```
ret.system_mode = getMetaValue(msg.endpoint, model, 'tuyaThermostatSystemMode')[value];
```

creckord:
That makes a lot of sense :D Done.

creckord:
Although, come to think of it: Does that assumption hold for all Tuya TRVs? I.e. does the preset always correspond 1:1 to the system_mode?

Koenkk:
Thanks!

PR #2197, opened 2021-02-03
Adding Byun gas sensor

dmitrymoshalkov:
Gas sensor from taobao
https://item.taobao.com/item.htm?spm=a1z10.3-c.w4002-8117930726.13.41dd1106ckz1wA&id=620636502665
manufacturer Byun
Model id: GAS SENSOR


Koenkk:
Thanks!

PR #2200, opened 2021-02-04
Support for Sengled Multicolor Light Strip (E1G-G8E)

KenGrinder:
Used previous Sengled devices as template.
Everything seems to be working locally with no errors for converters with debugging turned on.


Koenkk:
Thanks!

PR #2203, opened 2021-02-05
SR-ZG9101EA-5C and SR-ZG1029-5C from Sunricher

jd1900:
Add support for SR-ZG9101EA-5C and SR-ZG1029-5C from Sunricher.
One of them is exactly the same as 511.00 from Iluminize. Sunricher is the actual manufacturer.
The other is the same with higher amperage. Both (SR-ZG9101EA-5C and SR-ZG1029-5C) are identified as HK-ZD-RGBCCT-A.


Koenkk:
Thanks!

PR #2204, opened 2021-02-06
Add support for RGBCCT and CCT Sunricher remotes

jd1900:
Add support for RGBCCT and CCT Sunricher remotes:
- Model: ZG2858A
- Model: ZG2819S-CCT (very similar to ZG2819S-RGBW from Iluminize, which is manufactured as well from Sunricher).

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -11607,7 +11607,43 @@ const devices = [
             await reporting.onOff(endpoint);
         },
     },
-
+    {
+        zigbeeModel: ['ZG2819S-CCT'],
+        model: 'ZG2819S-CCT',
+        vendor: 'Sunricher',
+        description: 'Zigbee handheld remote CCT 4 channels',
+        fromZigbee: [fz.battery, fz.command_move_to_color, fz.command_move_to_color_temp, fz.command_move_hue,
+            fz.command_step, fz.command_recall, fz.command_on, fz.command_off,
+            fz.command_on, fz.command_off, fz.command_toggle, fz.command_stop, fz.command_move, 
+            fz.command_color_loop_set, fz.command_ehanced_move_to_hue_and_saturation, fz.tint_scene],
+        exposes: [e.battery(), e.action([
+            'color_move', 'color_temperature_move', 'hue_move', 'brightness_step_up', 'brightness_step_down', 'recall_*', 'on', 'off', 'toggle',
+            'brightness_stop', 'brightness_move_up', 'brightness_move_down', 'color_loop_set', 'enhanced_move_to_hue_and_saturation', 'scene_*'])],
+        toZigbee: [],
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -11607,6 +11607,44 @@ const devices = [
...
     },
+    {
+        zigbeeModel: ['ZG2819S-CCT'],
+        model: 'ZG2819S-CCT',
+        vendor: 'Sunricher',
+        description: 'Zigbee handheld remote CCT 4 channels',
+        fromZigbee: [fz.battery, fz.command_move_to_color, fz.command_move_to_color_temp, fz.command_move_hue,
+            fz.command_step, fz.command_recall, fz.command_on, fz.command_off,
+            fz.command_on, fz.command_off, fz.command_toggle, fz.command_stop, fz.command_move,
+            fz.command_color_loop_set, fz.command_ehanced_move_to_hue_and_saturation, fz.tint_scene],
```
Koenkk, line 11618: Is `tint_scene` actually needed here? Can you test if the scene commands also work without it?

devices.js
```diff
@@ -11607,6 +11607,44 @@ const devices = [
...
+    },
+    {
+        zigbeeModel: ['ZG2858A'],
+        model: 'ZG2858A',
+        vendor: 'Sunricher',
+        description: 'Zigbee handheld remote RGBCCT 3 channels',
+        fromZigbee: [fz.battery, fz.command_move_to_color, fz.command_move_to_color_temp, fz.command_move_hue,
+            fz.command_step, fz.command_recall, fz.command_on, fz.command_off,
+            fz.command_on, fz.command_off, fz.command_toggle, fz.command_stop, fz.command_move,
+            fz.command_color_loop_set, fz.command_ehanced_move_to_hue_and_saturation, fz.tint_scene],
```
Koenkk, line 11637: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/2204/files#r571458384

devices.js
jd1900: I'm not really sure, as I bound the remote to use it.
The scenes allows to "retrieve" current values and store the scene in the remote.

How could I test it?

devices.js
jd1900: Same answer here :)

devices.js
Koenkk: Remove `fz.fz.tint_scene` from this list and check if still all `action` are send correctly.

jd1900: It seems so.

I didn't see any error or unsupported message.
So safe to remove.

devices.js
jd1900: Same here, safe to remove.

Koenkk:
Thanks!

Issue #1998, opened 2020-12-30
Checksum verification of OTA images on download


davidjb:
At the moment, the OTA update process checks various headers inside a downloaded image during parsing (version, size, type, manufacturer, etc) but of these, just checking the size could allow corruption during downloading to pass silently.  For example, if the headers and size remained the same but the content was corrupted or modified during transfer for whatever reason (e.g. network or memory corruption, bit flips, modification over insecure HTTP or other malicious intent, etc), then a downloaded file could still pass the current tests.

Adding a known checksum like a sha512 hash to each image in the list of known OTA images and implementing a check of the file just after downloading would catch this type of issue, ensuring the firmware you're downloading is exactly what's expected. 

If this sounds reasonable, I can look at creating a PR here and updating https://github.com/Koenkk/zigbee-OTA/blob/master/index.json for the existing firmware.

Koenkk:
Valid request IMHO. Can you make a PR?

davidjb:
Sure, will give it a go in a few days' time - I'll look at updating the 'add' scripting too in zigbee-OTA.

davidjb:
First PR is https://github.com/Koenkk/zigbee-OTA/pull/25 - adds a sha512 sum to all OTA firmware + the scripts for adding firmware.

davidjb:
Second PR is at https://github.com/Koenkk/zigbee-herdsman-converters/pull/2205 - adds verification of checksums if present + tests to verify validation.

Koenkk:
Since https://github.com/Koenkk/zigbee-herdsman-converters/pull/2205 is merged I assume this can be closed now, thanks!

davidjb:
Yep, all good @Koenkk.  I don't have devices from the other OTA providers but will have a look at checksum validation for those when I get a moment.

PR #2205, opened 2021-02-06
Verify checksums of Zigbee-OTA images

davidjb:
This implements #1998 - this PR verifies the checksum of OTA images after download, throwing an error if the validation fails and passes through the sha512 checksums for the Zigbee-OTA images.  Tests are included to test successful and non-successful validation use cases.

Other OTA providers can use this as a base and have hash validation added too -- a quick look into the API of Ledvance shows they return a sha256 hash.


Koenkk:
Thanks!

PR #2211, opened 2021-02-07
Update description on Hue white bulbs

davidjb:
Several Philips Hue white bulbs are the same zigbee model number but have different connectors (e.g. E27 and B22, such as https://www.philips-hue.com/etc/philips/b2c/compareproducts.html?locale=en_NZ&catalog=CONSUMER&ctns=8718696554951,8718696449691).

This PR adds the alternate connector options to the descriptions; I'd considered making the descriptions generic by removing the connector but then someone searching for the specific connector may not find it in the docs.


Koenkk:
Thanks!

PR #2212, opened 2021-02-07
Add sha256 validation for Ledvance OTA updates

davidjb:
This PR continues on from #1998 by adding vendor-supplied sha256 checksum validation for Ledvance API OTA updates (none of the other vendors appear to publish checksums at this point).

The hash algorithm selection in `common.js` is simplistic at present - working with sha512 and sha256 - but can be added to (and refactored) when other checksum algorithms are needed.


Koenkk:
Thanks again!

PR #2213, opened 2021-02-07
Update min/max ct range for ZL1000100-CCT-US-V1A02

sjorge:
This should address  Koenkk/zigbee2mqtt#6123

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14541,7 +14541,7 @@ const devices = [
         model: 'ZL1000100-CCT-US-V1A02',
         vendor: 'Linkind',
         description: 'Zigbee LED 9W A19 bulb, dimmable & tunable',
-        extend: preset.light_onoff_brightness_colortemp(),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

sjorge:
Maybe we should put out a call for database entries somehow to fill in the missing data?

Koenkk:
We can use the `configure()` to set this. Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/b3db136d9ad4213e8f2a2217480e85d290ec219d/devices.js#L1677

sjorge:
We already read and save those attributes. We just don't use the stored entries in the exposes info.

Koenkk:
Ah I see, the expose part is indeed a static thing and cannot be updated from the actual runtime device. Currently this does not fit well with how the whole expose thing has been setup.

sjorge:
Maybe we can start by making sure newly added devices at least have it filled in.

Koenkk:
Yes agree, I will check this.

PR #2222, opened 2021-02-09
Added LightSolutions 91-948

Duckle29:
Added LightSolutions zigbee switch
https://www.smart.dk/mini-zigbee-afbryder-200w.html

works with my device.


Koenkk:
Thanks!

PR #2225, opened 2021-02-09
ROB_200-008-0: Added zigbeeModel, added battery report

Derecho:
I recently purchased this device, and initially Zigbee2Mqtt reported it as unsupported. Turns out the device is already supported, just that the version I bought identifies under a different zigbee model, so I've added it. Now it shows up recognised and works as expected.

Since these keypads are whitelabels similar to the iCasa series that I've previously added to my setup, I also knew they can report battery state during regular button presses, so I copied over that behaviour too after testing it.

On another note; would it be useful if I were to make another PR for adding a `whiteLabel` attribute to the iCasa ICZB-KPD12 device that I own? These ROBB products have this attribute listed but the iCasa products do not, yet it's quite clear they are made by the same manufacturer judging from the products themselves, their packaging, and their instruction booklet.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -7632,7 +7632,7 @@ const devices = [
         whiteLabel: [{vendor: 'Sunricher', model: 'SR-ZG9001K8-DIM'}],
     },
     {
-        zigbeeModel: ['ZG2833K4_EU06'],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2226, opened 2021-02-09
ICZB-KPD12: Added whiteLabel attribute

Derecho:
I have a setup here with both products from iCasa and ROBB, and I noticed from the ROBB entries in `devices.js` that there is a an optional `whiteLabel` attribute defined for some of the products. The iCasa product series are also whitelabels from Sunricher, so I added the respective entry for the device that I own.


Koenkk:
Thanks!

PR #2228, opened 2021-02-10
ROB_200-018-0: Added support

Derecho:
As per the title. Basically a copy of the existing Sunricher ZG2835 support, tested and verified ok.


Koenkk:
Thanks!

PR #2230, opened 2021-02-10
Support Hama TRV (white-label Saswell SEA802-Zigbee)

rkallensee:
Add support for the [Hama TRV](https://www.hama.com/00176592/hama-smart-radiator-thermostat-for-hama-wlan-heating-control) which is a white-label Saswell SEA802-Zigbee.


Koenkk:
Thanks!

PR #2231, opened 2021-02-10
Don't read voltage multipliers for 3200-Sgb

crowbarz:
`reporting.readEletricalMeasurementMultiplierDivisors()` returns `UNSUPPORTED_ATTRIBUTE` for model `3200-Sgb` because it does not support reading of `acVoltageMultiplier` and `acVoltageDivisor`. This PR updates the model to read only the current and power multiplier/divisors.


Koenkk:
Thanks!

PR #2233, opened 2021-02-10
Fix battery reporting for model 3325-S

crowbarz:
Changed model for `3325-S` to read `batteryVoltage` instead of `batteryPercentageRemaining`, just like the [SmartThings DTH](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy) does.
Expose `battery` property (set by `fz.battery`), remove `battery_low` property as it is superseded by `battery` property.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -7885,16 +7885,16 @@ const devices = [
         model: '3325-S',
         vendor: 'SmartThings',
         description: 'Motion sensor (2015 model)',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2235, opened 2021-02-11
ROB_200-018-0: Added support for colour temperature setting

Derecho:
After reading the manual with the product a few more times and intepreting it in various ways, I discovered how to change operating modes in the product, and how they allow the user to perform different operations with the dimming knob. Mode 1 is default and only handles brightness regulation, mode 2 and 4 can also send colour temperature messages which I've now added support for. Mode 3 and 4 supposedly can also change hue, but I'm not seeing any zigbee messages for that alas.


Koenkk:
Thanks!

PR #2238, opened 2021-02-13
Support Iris 3450-L Smart Fob

presslab-us:
This requires pull request: https://github.com/Koenkk/zigbee-herdsman/pull/307

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1317,6 +1317,22 @@ const converters = {
             return result;
         },
     },
+    checkin_presence: {
+        cluster: 'genPollCtrl',
+        type: ['commandCheckin'],
+        convert: (model, msg, publish, options, meta) => {
+            const useOptionsTimeout = options && options.hasOwnProperty('presence_timeout');
+            const timeout = useOptionsTimeout ? options.presence_timeout : 100; // 100 seconds by default
+
+            // Stop existing timer because motion is detected and set a new one.
+            clearTimeout(globalStore.getValue(msg.endpoint, 'timer'));
+
+            const timer = setTimeout(() => publish({presence: false}), timeout * 1000);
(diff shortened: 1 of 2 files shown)


presslab-us:
There are a couple improvements I could think of to this pull request, but I couldn't figure out how to do them.  Maybe if they aren't too complicated I could implement them with some help.
1.  Support the `presence_timeout` setting.  I can't figure out how to get this setting from the device configure function, which is needed when the `checkinInterval` is set up.  As it is now, if it's set to something other than 100 seconds it won't work right.
2. Change the `presence` property to true if any message is received at all, not just the `commandCheckin`.  There is a device `onEvent` function but I don't know how to publish anything from there.  It would need to set up the timer again as it's currently being done in the `checkin_presence` converter.

Koenkk:
1. Do you mean in the `configure()`? This is currently not possible indeed. But I expect it to only not work correctly when the presence_timeout < 100, 100 >= should work
2. You can create custom converters for all other commands (e.g. command_on) from there you can call the command_on converter like this: https://github.com/Koenkk/zigbee-herdsman-converters/blob/c9158dfcb92ae52018cce0d259e2d7d0e32ca050/converters/fromZigbee.js#L3713

presslab-us:
1. Yeah, `configure()`.  I'd think the point of reducing the `checkinInterval` would be to reduce the battery consumption with an increase and improving the response time with a decrease.  I'm not sure that changing `presence_timeout` without changing `checkinInterval` would be very useful.
2. Ok, I can do that.  I just didn't want to create a lot of unnecessary converters.  I can reduce the size as you showed by reusing the other converters.

Koenkk:
Cool, let me know when you are ready

presslab-us:
Sure, it is good now, unless you can think of a good solution for #1.  Maybe just something for a future improvement.

Koenkk:
Thanks!

PR #2131, opened 2021-01-23
Add support for yookee cover

barlock:
Support for Yoolax smart blinds

https://smile.amazon.com/Yoolax-Motorized-Rechargeable-Restaurant-Customized/dp/B07QW65LTC?sa-no-redirect=1

They appear as "Yookee" for the manufacturer even though they sell as "Yoolax" 🤷‍♂️ 

The instruction manual actually shows Orvibo as the manufacturer for adding to alexa. so I used a nearly identical config.

PR #2239, opened 2021-02-13
invert yookee cover reporting

barlock:
Follow up to this pr: https://github.com/Koenkk/zigbee-herdsman-converters/pull/2131

After using it for a while, noticed that positions were being reported backwards.


Koenkk:
Thanks!

PR #2241, opened 2021-02-14
Update Sinope TH1123ZB and TH1124ZB

embak:
This update to Sinope TH1123ZB and TH1124ZB :
- Allow to set the thermostat display time format ( 24h / 12h )
- Report electrical measurements: Power, Current and Voltage

Some of those might not be available for your thermostat firmware version. (This is my case)


Koenkk:
Thanks!

Issue #2125, opened 2021-01-21
No converter found for power and energy on 


jekkos:
Great that I found the device below as supported in the documentation

https://www.zigbee2mqtt.io/devices/SP_120.html

I have tried to set it up and I'm able to connect it to the zigbee2mqtt. I'm seeing this message on startup
```
{"commit":"9674570","coordinator":{"meta":{"maintrel":3,"majorrel":2,"minorrel":6,"product":0,"revision":20190619,"transportrev":2},"type":"zStack12"},"log_level":"debug","network":{"channel":11,"extendedPanID":"0xdddddddddddddddd","panID":6754},"permit_join":true,"version":"1.16.2"}
```
When I publish this to the zigbee2mqtt topic of the device

```
 mosquitto_pub -h xxx -u xxx -P xxx -t zigbee2mqtt/power_meter/get -m "{\"power\": \"\"}"
```

I get no reply but see the following in the daemon logs

```
No converter available for 'power' ()
```

Any idea what could be wrong here? Thanks!

jekkos:
On the other hand, when I try to request the device state I see this message

```
{"message":"Publish 'get' 'state' to 'power_meter' failed: 'Error: Read 0x60a423fffe8cce79/1 genOnOff([\"onOff\"], {\"timeout\":10000,\"disableResponse\":false,\"disableRecovery\":false,\"disableDefaultResponse\":true,\"direction\":0,\"srcEndpoint\":null,\"reservedBits\":0,\"manufacturerCode\":null,\"transactionSequenceNumber\":null}) failed (Data request failed with error: 'MAC no ack' (233))'","meta":{"friendly_name":"power_meter"},"type":"zigbee_publish_error"}
```

travelography:
Hi Jekkos, maybe the following link will help you https://github.com/Koenkk/zigbee2mqtt/issues/3731

jekkos:
Thanks for the pointer @travelography. I'll look into it.

sjorge:
power doesn't have a toZigbee converter only a fromZigbee one.

It either comes from
- https://github.com/Koenkk/zigbee-herdsman-converters/blob/ac94b3ef958abe299bec2725b93959be087dc8be/converters/fromZigbee.js#L525
- https://github.com/Koenkk/zigbee-herdsman-converters/blob/ac94b3ef958abe299bec2725b93959be087dc8be/converters/fromZigbee.js#L565

To be able to use /get you need a something like

```
[...]

PR #2243, opened 2021-02-14
Allow /get/power for Niko sockets and plugs

sjorge:
As mentioned in #2125 most of our plugs/sockets don't allow getting of the power attribubte.

This PR adds it for Niko sockets and plugs, I also tested my Ubisys S1/S2 but they unfortunately lock up when trying :(
https://github.com/Koenkk/zigbee-herdsman-converters/issues/2125#issuecomment-778756661


Koenkk:
Thanks!

PR #2244, opened 2021-02-14
Always expose `previous` for color_temp_startup

sjorge:
Got filtered out in 3bcc1080b44749e6e0f7bf4351b88da22739e9d1, but 65535 is special as it lies outside the normal allowed range of color_temp so it will never be within min/max range!

It means use previous color_temperature.

image


Koenkk:
Well spotted, thanks!

PR #2251, opened 2021-02-16
Added Tradfri occupancy_timeout device config option

isorin:
Similar to Xiaomi Aqara motion sensor device option, added the ability to override the default detection cooldown timeout.
This allows custom motion detection loop control in HA automation by setting the value lower than the re-trigger interval of the device.

Related PR:
https://github.com/Koenkk/zigbee2mqtt.io/pull/629


Koenkk:
Thanks!

Issue #2253, opened 2021-02-16
Want a way to configure lightingColorCtrl options


sjorge:
## what ?
I've been playing with the `options` attribute of the `lightingColorCtrl` cluster a bit.

<img width="909" alt="image" src="https://user-images.githubusercontent.com/379665/108092918-77ea5180-707d-11eb-9b9e-73f6b2a3c09f.png">

## why
I am **abusing** scenes to change the color of  bulbs even when they are off, mostly for adaptive lighting. This works, but it's very very ugly. (But on the upside it made me improve scene support ¯\_(ツ)_/¯)

For my FLOALT panels I was not using scenes as they supported setting color_temp (or limited color_xy set) while off. It works a lot smoother (no scene_add, scene_recall dance which is probably not great for the bulb and unneeded traffic on the mesh)

I did some experimenting and the old FLOALT had the ExecuteIfOff bit set to 1, after the upgrade last week it is now set to 0 and it stopped working. 

## options
I already did the preliminary PR Koenkk/zigbee-herdsman#310 as we did not have the attribute listed.

I currently have a 'working' implementation that is rather rough and I am not happy with, I could polish it but there might be a better way?

toZigbee:
```
    light_execute_if_off: {
        key: ['execute_if_off'],
        convertSet: async (entity, key, value, meta) => {
            await entity.write('lightingColorCtrl', {options: value}, utils.getOptions(meta.mapped, entity));
            return {state: {execute_if_off: value}};
        },
        convertGet: async (entity, key, meta) => {
            await entity.read('lightingColorCtrl', ['options']);
        },
    },
```
fromZigbee:
```
    light_execute_if_off: {
        cluster: 'lightingColorCtrl',
        type: ['attributeReport', 'readResponse'],
        convert: (model, msg, publish, options, meta) => {
            const result = {};

            if (msg.data.hasOwnProperty('options')) {
                result.execute_if_off = msg.data['options'];
            }

            return result;
        },
    },
```

Polishing this would probably be something like:
- rename execute_if_off to `light_options`
- make the value a object with a key called `ExecuteIfOff`
- do a proper bitshift based on the key's true/false value

That leaves the door open for future options if ZCL is updated.

@Koenkk some feedback on this would be welcome, polishing this a bit is certainly doable... but I feel like there might be better alternatives? 

[...]

PR #2254, opened 2021-02-16
Setup lightingColorCtrl options

sjorge:
See #2253, we can still go the other way if you want. But I don't think having this as yet another payload attribute is useful.
No errors as we have this behind a try/catch, works OK for all my CWS bulbs with the exception of Hue.

It also works on my Trådfri panels! But not my Trådfri GU10 bulbs. Still nice to have though.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -78,12 +78,13 @@ const preset = {
             exposes,
             fromZigbee: [fz.color_colortemp, fz.on_off, fz.brightness, fz.level_config, fz.ignore_basic_report],
             toZigbee,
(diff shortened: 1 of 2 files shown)


Koenkk:
I would expect this to be a configurable option, not something that is always set to a certain option. Why not create a new toZigbee converter to set this?

sjorge:
See the linked issue, originally I did.

sjorge:
Some new thing popped up while re-reading the ZCL, I can't seem to move it back to draft PR.

PR #2255, opened 2021-02-16
Feature/support innr rc 110

mrsteakhouse:
Adds support for Innr RC 110.

The device must be manually added to a group for the scene buttons and main buttons to work properly.
If there is an API  to add the device during configuration to a new group, please let me know.
Adding the endpoints each to different groups does not work on my setup.


Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/630

devices.js
```diff
@@ -5977,6 +5977,42 @@ const devices = [
...
     },
+    {
+        zigbeeModel: ['RC 110'],
+        model: 'RC 110',
+        vendor: 'Innr',
+        description: 'Remote Control',
+        fromZigbee: [fz.ignore_basic_report,
+            fz.ignore_command_off,
+            fz.command_on_state, fz.command_off_state,
+            fz.innr_rc110_levelCtrl_move, fz.innr_rc110_levelCtrl_stop, fz.innr_rc110_levelCtrl_step,
```
Koenkk, line 5988: Please use the existing generic converters instead of adding custom ones, e.g. `fz.innr_rc110_levelCtrl_stop` -> `fz.command_stop`.

mrsteakhouse: i had some issues with the existing implementation. it was impossible to have an individual simulated brightness level per endpoint.
maybe the existing one can be refactored to accept options to shape the payload.

Koenkk: > maybe the existing one can be refactored to accept options to shape the payload.

 yes please, this will save a lot of code.

Could you update it?

mrsteakhouse: Yes, i will look into this.

mrsteakhouse: I have finished refactorings. Please let me know about any suggestions.

converters/fromZigbee.js
```diff
@@ -5182,7 +5207,14 @@ const converters = {
             return {occupancy: (zoneStatus & 1<<2) > 0};
         },
     },
-
+    innr_rc110_scene: {
```
Koenkk, line 5210: Why not use `command_move_to_level`?

lib/utils.js
```diff
@@ -103,6 +103,23 @@ function postfixWithEndpointName(value, msg, definition) {
     }
 }
 
+function getSimulatedBrightnessKeys(options, msg, definition) {
```
Koenkk, line 106: I don't think this function is necessary. The current keys are ok as the global store works per endpoint, not per device (so there won't be conflicts)

converters/fromZigbee.js
```diff
@@ -1000,35 +1012,48 @@ const converters = {
...
-                clearInterval(globalStore.getValue(msg.endpoint, 'simulated_brightness_timer'));
-                globalStore.putValue(msg.endpoint, 'simulated_brightness_timer', undefined);
+                const keys = getSimulatedBrightnessKeys(options, msg, model);
+                clearInterval(globalStore.getValue(msg.endpoint, keys.timer));
+                globalStore.putValue(msg.endpoint, keys.timer, undefined);
+
+                if (options.simulated_brightness.multi_endpoint) {
+                    const property = postfixWithEndpointName('brightness', msg, model);
+                    const brightness = globalStore.getValue(msg.endpoint, keys.brightness, 255);
+                    payload = {[property]: brightness};
```
Koenkk, line 1053: If I understand correctly, with this the `action` will be removed from the `payload`?

PR #2256, opened 2021-02-16
TERNCY-DC01: add battery percentage readout

o0v7r0o:
I was surprised that Terncy DC-01 door sensor does not expose battery status. Indeed it does, just have to add corresponding converters.
I have also tried to set up reporting, but it fails -> I think it is not supported by the sensor.


Koenkk:
Thanks!

PR #2260, opened 2021-02-18
Add support for accelerometer axes in Samsung Smartthings Multipurpose Sensor (F-MLT-US-2)

declanshanaghy:
See: https://github.com/Koenkk/zigbee2mqtt/issues/6021

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3406,7 +3406,23 @@ const converters = {
         cluster: 'manuSpecificSamsungAccelerometer',
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
-            return {moving: msg.data['acceleration'] === 1 ? true : false};
+            const payload = {};
+            if (msg.data.hasOwnProperty('acceleration')) payload.moving = msg.data['acceleration'] === 1;
+
+            // eslint-disable-next-line
+            // https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy#L222
+            /*
+                The axes reported by the sensor are mapped differently in the SmartThings DTH.
+                Preserving that functionality here.
+                xyzResults.x = z
+                xyzResults.y = y
+                xyzResults.z = -x
+            */
+            if (msg.data.hasOwnProperty('z_axis')) payload.x_axis = msg.data['z_axis'];
+            if (msg.data.hasOwnProperty('y_axis')) payload.y_axis = msg.data['y_axis'];
+            if (msg.data.hasOwnProperty('x_axis')) payload.z_axis = - msg.data['x_axis'];
+
+            return payload;
         },
     },
     byun_smoke_false: {
--- a/devices.js
+++ b/devices.js
@@ -8132,11 +8132,19 @@ const devices = [
             await endpoint.write('manuSpecificSamsungAccelerometer', {0x0002: {value: 0x0276, type: 0x21}}, options);
             await reporting.temperature(endpoint);
             await reporting.batteryVoltage(endpoint);
-            const payload = reporting.payload('acceleration', 10, repInterval.MINUTE, 1);
-            await endpoint.configureReporting('manuSpecificSamsungAccelerometer', payload, options);
+            const payloadA = reporting.payload('acceleration', 10, repInterval.MINUTE, 1);
+            await endpoint.configureReporting('manuSpecificSamsungAccelerometer', payloadA, options);
+            const payloadX = reporting.payload('x_axis', 10, repInterval.MINUTE, 1);
+            await endpoint.configureReporting('manuSpecificSamsungAccelerometer', payloadX, options);
+            const payloadY = reporting.payload('y_axis', 10, repInterval.MINUTE, 1);
+            await endpoint.configureReporting('manuSpecificSamsungAccelerometer', payloadY, options);
+            const payloadZ = reporting.payload('z_axis', 10, repInterval.MINUTE, 1);
+            await endpoint.configureReporting('manuSpecificSamsungAccelerometer', payloadZ, options);
         },
-        exposes: [e.temperature(), e.contact(), e.battery_low(), e.tamper(), e.battery(),
-            exposes.binary('moving', ea.STATE, true, false)],
+        exposes: [
+            e.temperature(), e.contact(), e.battery_low(), e.tamper(), e.battery(),
+            ...e.acceleration(),
+        ],
     },
     {
         zigbeeModel: ['multi'],
--- a/devices.js
+++ b/devices.js
@@ -8143,7 +8143,7 @@ const devices = [
         },
         exposes: [
             e.temperature(), e.contact(), e.battery_low(), e.tamper(), e.battery(),
-            e.acceleration(), e.x_axis(), e.y_axis(), e.z_axis(),
+            e.moving(), e.x_axis(), e.y_axis(), e.z_axis(),
         ],
     },
     {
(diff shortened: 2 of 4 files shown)


declanshanaghy:
This works fine but it has an odd quirk that I'd like to resolve before merging. If any of the axes values haven't changed then that attribute is not included in the next MQTT publish, this causes a value of 0 to show up in Home Assistant for that axis.

It would be preferable to publish the last value each time.

converters/fromZigbee.js
```diff
@@ -3409,6 +3409,17 @@ const converters = {
             return {moving: msg.data['acceleration'] === 1 ? true : false};
         },
     },
+    smartthings_accelerometer: {
+        cluster: 'manuSpecificSamsungAccelerometer',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            return {
+                x_axis: msg.data['x_axis'],
```
crowbarz, line 3417: I think you might need to check for the existence of the `x_axis`, `y_axis` and `z_axis` attributes in the Zigbee packet before you include it in the returned object. That could be causing the odd quirk that you mentioned. I'm not a JS guy, reference to a nonexistent attribute this would throw an exception in Python but in JS maybe it returns `null` which is converted somewhere to 0.

converters/fromZigbee.js
```diff
@@ -3409,6 +3409,17 @@ const converters = {
             return {moving: msg.data['acceleration'] === 1 ? true : false};
         },
     },
+    smartthings_accelerometer: {
```
crowbarz, line 3412: Given it's processing the same cluster, would it make more sense for this code to go into `smartthings_acceleration` instead?

devices.js
```diff
@@ -1180,7 +1180,8 @@ const devices = [
 
     // TuYa
     {
-        fingerprint: [{modelID: 'TS011F', manufacturerName: '_TZ3000_rk2yzt0u'}, {modelID: 'TS0001', manufacturerName: '_TZ3000_o4cjetlm'}, {manufacturerName: '_TZ3000_o4cjetlm'}],
+        fingerprint: [{modelID: 'TS011F', manufacturerName: '_TZ3000_rk2yzt0u'},
```
crowbarz, line 1183: This patch looks like it might have been generated by a formatter ...

converters/fromZigbee.js
declanshanaghy: Hmmm, ya sounds about right. In JS this will return `undefined` and maybe somewhere upstream keys with undefined values are being removed.

devices.js
declanshanaghy: I did this because the linter complained that the line was too long. Not sure how it got through before this PR.

converters/fromZigbee.js
declanshanaghy: It might. Since I'm not vary familiar with zigbee internals or the conventions of this repo, I separated it out. Could use a recommendation on this.

declanshanaghy: done

converters/fromZigbee.js
Koenkk: In case that e.g. `x_axis` is not defined it should not be returned. If e.g. first a message with `acceleration` and `x_axis` is received and next a message with only `acceleration` the `x_axis` property will be cleared (while it should be preserved.

Changing to:

```js
return {
    ...(msg.data.hasOwnProperty('acceleration') && {moving: msg.data['acceleration'] === 1}),
    ...(msg.data.hasOwnProperty('x_axis') && {x_axis: msg.data['x_axis']}),
    ...(msg.data.hasOwnProperty('y_axis') && {y_axis: msg.data['y_axis']}),
    ...(msg.data.hasOwnProperty('z_axis') && {z_axis: msg.data['z_axis']}),
}
```

should solve it.

devices.js
Koenkk: This change can be reverted, already fixed it on the master branch. (so rebase it afterwards)

converters/fromZigbee.js
crowbarz: You might also want to look here at the [SmartThings DTH code](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy#L222), it does this mangling with the received `x/y/z_axis` attributes when manufacturer = `SmartThings` (the only model supported by the DTH with this manufacturer is `multiv4`):
```
xyzResults.x = z
xyzResults.y = y
xyzResults.z = -x
```

declanshanaghy: Thx.
Do i need to worry about the manufacturer available to me in this location to implement the same check that we have or should I just go with the "SmartThings" case?

crowbarz: The SmartThings DTH seems to be written to cover a single product ("SmartThings multi-sensor") that have multiple hardware revisions that are sometimes made by completely different OEM manufacturers. In z2m, each hardware revision has a separate device profile that can be set up to trigger either the same or different converters. So I'd suggest that you just code the case specific for `multiv4`. If someone decides to add accelerator support to any of the other devices supported by that DTH, then they can either modify the converter you wrote in this PR to handle the other cases, or (more likely as it's relatively simple) write a separate converter that mangles `x/y/z-axis` as required for that device.

declanshanaghy: > xyzResults.x = z
> xyzResults.y = y
> xyzResults.z = -x

This explains why I had to change all my orientation logic when I switched this sensor over from SmartThings to ZHA in Home Assistant.

declanshanaghy:
Where can I augment the HA MQTT discovery data that's published for this?
I'd at least like to set the `unit_of_measurement` properly.
Probably a few others too.

See: https://photos.app.goo.gl/uSSoQkHyAHTDgZ8c6

lib/exposes.js
```diff
@@ -512,5 +512,13 @@ module.exports = {
             .withFeature(new Numeric('duration', access.SET).withUnit('s').withDescription('Duration in seconds of the alarm')),
         week: () => new Enum('week', access.STATE_SET, ['5+2', '6+1', '7']).withDescription('Week format user for schedule'),
         window_detection: () => new Switch().withState('window_detection', true, 'Enables/disables window detection on the device', access.STATE_SET),
+        acceleration: () => {
```
Koenkk, line 515: Can we keep these separate? I prefer to keep everything in `presets` the same.

Koenkk:
> unit_of_measurement

You can fill in the `withUnit()` part here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/2260/files#diff-d0e9d08b90d447f440404af9227490f81b63e0d3a02ebd8235023c8b1ec72d47R518

lib/exposes.js
declanshanaghy: Originally i had it in a section called `accelerometer`. 
I figured since it's the same category of "things" it makes sense together.

What do you mean by "keep everything in presets the same" though?

Does that mean after something is added to presets, it shouldn't change?
I suppose that makes sense because other devices that use `moving` via `acceleration` would pick up extra attrs that they might not support.

declanshanaghy:
> > unit_of_measurement
> 
> You can fill in the `withUnit()` part here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/2260/files#diff-d0e9d08b90d447f440404af9227490f81b63e0d3a02ebd8235023c8b1ec72d47R518

Ahh ya. I found that. But setting it to `''` caused the check in `zigbee2mqtt` to fallback on `'-'` anyway. So I opened another MR to handle that case.

lib/exposes.js
Koenkk: Instead of the `acceleration()` add the 4 presets (moving, x_axis, y_axis, z_axis`) as separate entries instead of grouping them (like all other presets)

PR #2264, opened 2021-02-18
updates SR-ZG9040A to add power measurements

orrpan:
Besides this it should (according to [documentation](https://www.m.nu/Common/File/Download?guid=7FD33DFB-BA0E-4987-A5A9-7B1620B0C211) support switching between `push switch`, `Normal on/off switch` and `3-Way switch`). Do you think it could be done or is this done at construction/initial flash?

And maybe I should create a sunricher.preset instead. 

From docs.
image

image

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -11705,7 +11705,7 @@ const devices = [
         fromZigbee: preset.light_onoff_brightness().fromZigbee.concat([fz.electrical_measurement, fz.metering, fz.ignore_genOta]),
         toZigbee: preset.light_onoff_brightness().toZigbee,
         meta: {configureKey: 2},
-        exposes: [e.light_brightness(), e.power(), e.energy()],
+        exposes: [e.light_brightness(), e.power(), e.voltage(), e.current(), e.energy()],
         whiteLabel: [{vendor: 'YPHIX', model: '50208695'}, {vendor: 'Samotech', model: 'SM311'}],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -11735,7 +11735,23 @@ const devices = [
...
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            const binds = ['genOnOff', 'genLevelCtrl', 'haElectricalMeasurement', 'seMetering'];
+            await reporting.bind(endpoint, coordinatorEndpoint, binds);
+            await reporting.onOff(endpoint);
+            await reporting.brightness(endpoint);
+            await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
+            await reporting.activePower(endpoint);
+            await reporting.rmsCurrent(endpoint, {min: 10, change: 10});
+            await reporting.rmsVoltage(endpoint, {min: 10});
```
Koenkk, line 11751: Reporting is setup for voltage and current here but it's missing from `exposes`. Does this device provide voltage and current measurements? If yes add it to exposes, if not remove the setup reporting lines.

Koenkk:
> Do you think it could be done or is this done at construction/initial flash?

AFAIK there is no standard way to do this in Zigbee so probably it's a manufacturer specific extension. You will need to sniff traffic with the original gateway to reverse engineer this: https://www.zigbee2mqtt.io/how_tos/how_to_sniff_zigbee_traffic.html

devices.js
orrpan: yes, `e.voltage(), e.current(),` should be added. I'm realizing that `SR-ZG9040A` and `ZG2835RAC` (also missing to expose voltage and current) is exactly the same config. Should they be two different setup with only different `zigbeeModel:`, `model:` and `description:`, or should this be refactored some way?

Koenkk: You can add the same code there

Koenkk:
Thanks!

PR #2266, opened 2021-02-19
Add support for Sengled E1G-G8E

clcain:
Add configuration for [Sengled E1G-G8E](https://us.sengled.com/products/sengled-smart-led-multicolor-light-strip) smart LED multicolor light strip to device list.


Koenkk:
Thanks but this device is already supported: https://github.com/Koenkk/zigbee-herdsman-converters/blob/405aca02f92c69039090c9df33f682a4cbf86867/devices.js#L6332

PR #2268, opened 2021-02-19
Fix moes thermostat negative local temp

blackalegator:
Local temperature comes as a 16bit signed integer with the last digit as a fraction. When chunks are saved into JS's integer sign bit is lost. As a result, for example, at -2.9 local temperature zigbee converter outputs 6550.6. This change fixes this problem.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2739,7 +2739,8 @@ const converters = {
             case tuya.dataPoints.moesMinTemp:
                 return {min_temperature: value};
             case tuya.dataPoints.moesLocalTemp:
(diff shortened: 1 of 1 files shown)


blackalegator:
There also stands a question about how is 
case tuya.dataPoints.moesTempCalibration work. It looks like it is doing somthing similar but then why does it compare to 4000 and not 2048 and substracts 4096 instead of 4095?

insipiens:
It was done because it worked  ;)

if there are more elegant code solutions great....

Koenkk:
Thanks!

PR #2273, opened 2021-02-21
Allow get/power for Ubisys S1/S2

sjorge:
Ubisys support verified that we should indeed be able to read this.

Did some more digging and found the issue, well there is no issue, you just have to specify the correct EP.

```
[root@amethyst ~]# mosquitto_pub -t zigbee2mqtt/office/light/main/switch/3/get/power -m ''
```

```
Zigbee2MQTT:debug 2021-02-21 17:59:52: Received MQTT message on 'zigbee2mqtt/office/light/main/switch/3/get/power' with data ''
Zigbee2MQTT:debug 2021-02-21 17:59:52: Publishing get 'get' 'power' to 'office/light/main/switch'
Zigbee2MQTT:debug 2021-02-21 17:59:52: Received Zigbee message from 'office/light/main/switch', type 'readResponse', cluster 'seMetering', data '{"instantaneousDemand":1}' from endpoint 3 with groupID 0
Zigbee2MQTT:info  2021-02-21 17:59:52: MQTT publish: topic 'zigbee2mqtt/office/light/main/switch', payload '{"linkquality":57,"power":1,"state":"ON","update":{"state":"idle"}}'
```

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -1301,6 +1301,12 @@ const converters = {
             await entity.read('haElectricalMeasurement', ['activePower']);
         },
     },
+    metering_power: {
+        key: ['power'],
+        convertGet: async (entity, key, meta) => {
+            await entity.read('seMetering', ['instantaneousDemand']);
+        },
+    },
     // #endregion
 
     // #region Non-generic converters
--- a/devices.js
+++ b/devices.js
@@ -13318,13 +13318,17 @@ const devices = [
         model: 'S1',
         vendor: 'Ubisys',
         description: 'Power switch S1',
-        exposes: [e.switch(), e.power(), e.action([
-            'toggle', 'on', 'off', 'recall_*',
-            'brightness_move_up', 'brightness_move_down', 'brightness_stop',
-        ])],
+        exposes: [e.switch(), e.power().withAccess(ea.STATE_GET).withEndpoint('meter').withProperty('power'),
+            e.action([
+                'toggle', 'on', 'off', 'recall_*',
+                'brightness_move_up', 'brightness_move_down', 'brightness_stop',
+            ])],
         fromZigbee: [fz.on_off, fz.metering, fz.command_toggle, fz.command_on, fz.command_off, fz.command_recall, fz.command_move,
             fz.command_stop],
-        toZigbee: [tz.on_off, tz.ubisys_device_setup],
+        toZigbee: [tz.on_off, tz.metering_power, tz.ubisys_device_setup],
+        endpoint: (device) => {
+            return {'l1': 1, 's1': 2, 'meter': 3};
+        },
         meta: {configureKey: 3},
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(3);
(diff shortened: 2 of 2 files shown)


sjorge:
@Koenkk I wonder if we can make it more user friendly?

Something like this perhaps?

```
diff --git a/devices.js b/devices.js
index 104aaaa..6cf1c06 100755
--- a/devices.js
+++ b/devices.js
@@ -13283,6 +13283,9 @@ const devices = [
         fromZigbee: [fz.on_off, fz.metering, fz.command_toggle, fz.command_on, fz.command_off, fz.command_recall, fz.command_move,
             fz.command_stop],
         toZigbee: [tz.on_off, tz.metering_power, tz.ubisys_device_setup],
+            return {'l1': 1, 's1': 2, 'meter': 3};
+        },
         meta: {configureKey: 3},
             const endpoint = device.getEndpoint(3);
@@ -13323,7 +13326,7 @@ const devices = [
             fz.command_stop],
         toZigbee: [tz.on_off, tz.metering_power, tz.ubisys_device_setup],
         endpoint: (device) => {
-            return {'l1': 1, 'l2': 2, 's1': 3, 's2': 4};
+            return {'l1': 1, 'l2': 2, 's1': 3, 's2': 4, 'meter': 5};
         },
         meta: {configureKey: 3, multiEndpoint: true},
```

?

sjorge:
Interesting too, the test error is because I did not update the expose, but doing so doesn't help as it will take the wrong EP.

Koenkk:
The endpoint names has to be one of: https://github.com/Koenkk/zigbee2mqtt/blob/6484feecb7c4727c0623a8a6264bac8b3403bbb3/lib/util/utils.js#L33

If you take e.g. `l3` it should work with `withEndpoint('l3')`.

sjorge:
l3 isn’t really a good name l1 (and l2) are already used for the L1 and L2 feeds on the devices

Koenkk:
I'm fine with adding `meter` to the endpoint names, does it work after this?

sjorge:
To test I just add ‘meter’ to the array and use it in the endpoint map?

I’ll do that next time I’m on the computer.

sjorge:
With `meter` appended I can now indeed query it using 
```
mosquitto_pub -t zigbee2mqtt/office/light/main/switch/meter/get/power -m ''
```

Koenkk:
It indeed expects `power_meter`. But since it has only 1 power attribute (looks like total of device? not per endpoint) you can try changing `return {'l1': 1, 's1': 2, 'meter': 3};` to `return {'l1': 1, 's1': 2, 'default': 3};`. Then it should work without `.withEndpoint('meter'),` and with `mosquitto_pub -t zigbee2mqtt/office/light/main/switch/get/power -m ''`

sjorge:
> It indeed expects `power_meter`. But since it has only 1 power attribute (looks like total of device? not per endpoint) you can try changing `return {'l1': 1, 's1': 2, 'meter': 3};` to `return {'l1': 1, 's1': 2, 'default': 3};`. Then it should work without `.withEndpoint('meter'),` and with `mosquitto_pub -t zigbee2mqtt/office/light/main/switch/get/power -m ''`

Then that breaks the 'state' though (which is on l1)

sjorge:
l1 -> genOnoff cluster with state
s1 -> client cluster for the switch
meter -> seMetering cluster (for l1) or on the dual version for l1 + l2, it has combined results.

sjorge:
I think doing `e.power().withEndpoint('meter').withProperty('power')` might work, the withEndpoint first overrides the property from power -> power_meter, then the withProperty sets it to power again. I don't think it's designed to be use like this though, perhaps an extra flag on withEndpoint() to indicate if it should update the property name is needed.

Koenkk:
> Then that breaks the 'state' though (which is on l1)

Thats indeed true, but since this device has multiple endpoints it's not meant to ben controlled with `state`, only with `state_l1` and `state_l2`. I'm fine with both solutions so let me know what you think is the best

sjorge:
The Ubisys S2 is using state_l1 and state_l2... I think having Ubisys S1 use state_l1 and marking the seMetering cluster as the default might be the cleanest, let me
play around with expose first though.

sjorge:
@Koenkk setting `.withProperty('power')` after `.withEndpoint('meter')` works fine!
This also leaves the whole attributes the same for both Ubisys S1 and Ubisys S2 so we don't break any existing setups.

When manually doing the /get for power we need to specify the meter endpoint, which is expected and I think that's a good solution.

I will do a PR against z2m to add the `meter` endpoint also.

PR #2274, opened 2021-02-21
added Hue Centris variant, clarified description

markus1540:
Hi, I've added the 2 spots variant for Philips Hue Centris.
Documentation will follow.


Koenkk:
Thanks!

PR #2278, opened 2021-02-22
Added BTicino K4027C support

blackwiz4rd:
The K4027C has the same pattern as the other ;). Tested, however as stated in the comment, the position is not reliable yet.


Koenkk:
Thanks!

PR #2281, opened 2021-02-23
Add Megos (Aldi) led panels

BMOD89:
Add Megos led panels which were offered by Aldi

@Koenkk I have set the vendor to Megos because Aldi sell them as Megos panels. 
But I am not sure if the Aldi bulbs should also be renamed to Megos. They are labled as "Lightway - powered by Megos", but are available on the Megos web page as well. What do you think?


Koenkk:
Thanks.

I've changed the vendor to Aldi, as they are more recognisable in this way. I've added MEGOS to the description.

PR #2283, opened 2021-02-23
Allow /get for color_mode

sjorge:
As mentioned in the discussion linked in koenkk/zigbee2mqtt#6402

Here is a local diff I've been carrying for a while, it allows manually getting color_mode and actually parses the random numbers into something more descriptive.

I use this to manually drop either `color_temp`, `color.x` + `color.y`, or `color.hue` + `color.saturation` from the json payload before processing, so I also have the correct color info.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3836,23 +3836,27 @@ const converters = {
             const sceneid = value;
             await entity.command('genScenes', 'recall', {groupid, sceneid}, utils.getOptions(meta.mapped));
 
+            const addColorMode = (newState) => {
+                if (newState.hasOwnProperty('color_temp')) {
+                    newState.color_mode = constants.colorMode[2];
+                } else if (newState.hasOwnProperty('color')) {
+                    if (newState.color.hasOwnProperty('x')) {
+                        newState.color_mode = constants.colorMode[1];
+                    } else {
+                        newState.color_mode = constants.colorMode[0];
+                    }
+                }
+
+                return newState;
+            };
+
             const isGroup = entity.constructor.name === 'Group';
             const metaKey = `${sceneid}_${groupid}`;
             if (isGroup) {
                 const membersState = {};
                 for (const member of entity.members) {
                     if (member.meta.hasOwnProperty('scenes') && member.meta.scenes.hasOwnProperty(metaKey)) {
-                        const newState = member.meta.scenes[metaKey].state;
-                        if (newState.hasOwnProperty('color_temp')) {
-                            newState.color_mode = constants.colorMode[2];
-                        } else if (newState.hasOwnProperty('color')) {
-                            if (newState.color.hasOwnProperty('x')) {
-                                newState.color_mode = constants.colorMode[1];
-                            } else {
-                                newState.color_mode = constants.colorMode[0];
-                            }
-                        }
-                        membersState[member.getDevice().ieeeAddr] = newState;
+                        membersState[member.getDevice().ieeeAddr] = addColorMode(member.meta.scenes[metaKey].state);
                     } else {
                         meta.logger.warn(`Unknown scene was recalled for ${member.getDevice().ieeeAddr}, can't restore state.`);
                         membersState[member.getDevice().ieeeAddr] = {};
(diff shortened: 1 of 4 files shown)


Koenkk:
I want to wait to merge this once there is more clarity in: https://github.com/home-assistant/architecture/discussions/519

Besides changes in the pr also all converters have to be updated to return the correct color_mode (e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/cc6c4d009e5a37bb20aa42b0d5ebaa6a71018793/converters/toZigbee.js#L1044)

sjorge:
Yeah, this is just what I had so far as a set of patches I carry (when not testing stuff).

I can do the tz color and color_temp converters, do you want it in this or a separate PR?

sjorge:
Seems https://github.com/home-assistant/core/pull/47720 got merged, sadly it does not really map to ZCL's color_mode cleanly (the on/off and dimmer stuff)

mapping ZCL color_mode updates to the following should work pretty well though and it probably a good start.
```
COLOR_MODE_COLOR_TEMP = "color_temp"
COLOR_MODE_HS = "hs"
COLOR_MODE_XY = "xy"
```

I'll rebase the PR later tonight and rename ct -> color_temp. I will also set color_mode when /set is called. The filtering of the inactive keys for color and color_temp I'll do in a follow up PR. Not sure how we can do this for now as it looks like just returned color with the active attributes have them merged with the old values too 🤔

sjorge:
@Koenkk that should cover setting color_mode correctly I think.

sjorge:
Found an issue with scene_recall not updating color_mode, retesting with extra change.

sjorge:
Thats... annoying 
```
Zigbee2MQTT:debug 2021-03-16 17:25:24: Received Zigbee message from 'office/light/desk/bulb', type 'readResponse', cluster 'lightingColorCtrl', data '{"colorMode":2}' from endpoint 1 with groupID 0
Zigbee2MQTT:debug 2021-03-16 17:25:46: Received Zigbee message from 'office/light/desk/bulb', type 'attributeReport', cluster 'lightingColorCtrl', data '{"colorMode":0}' from endpoint 1 with groupID 0
```

I have to wait ~1-2 sec before reading colorMode, I guess we'll need to just set it based on what is in state instead...

sjorge:
OK, 2nd way give the expected behavior, would still have liked to just read it. But I couldn't find a way to read it after X timeout. Oh well.

converters/toZigbee.js
```diff
@@ -514,12 +520,13 @@ const converters = {
 
                 await entity.command('lightingColorCtrl', 'moveColorTemp', payload, utils.getOptions(meta.mapped, entity));
 
-                // As we cannot determine the new brightness state, we read it from the device
+                // We cannot determine the color temperaturefrom the current state so we read it, because
+                // - Color mode could have been swithed (x/y or colortemp)
```
Koenkk, line 524: I don't understand this change, if colorMode could change we should also ready the currentX/currentY?

sjorge: It's not a requirement that currentHue/currentSatuaturation/currentX/currentY gets updates.
Only what is the active color_mode 

I also observe this, e.g. innr will update colorTemperature when within range (or not when out of range), the tint bulb doesn't seem to do it.

Given we set colorTemperature, reading both should be fine and get us the current colorMode and bulb reported colorTemperature value.

sjorge: That does pose an interesting question, I guess the proper thing to do would be In both the color_temp and color converters... would be to read colorMode, check the result and then read either colorTemperature, currentHue + currentSaturation, or currentX + currentY.

That would guarantee we have the correct information from the bulb for the mode the bulb is set too. colorMode is a required value so barring manufacturer ignoring the ZCL (*sigh*) we should always be able to read colorMode if we have the ligthingColorCtrl cluster.

converters/toZigbee.js
```diff
@@ -3836,23 +3836,27 @@ const converters = {
             const sceneid = value;
             await entity.command('genScenes', 'recall', {groupid, sceneid}, utils.getOptions(meta.mapped));
 
+            const addColorMode = (newState) => {
```
sjorge, line 3839: <3

converters/toZigbee.js
Koenkk: I'm wondering if this converter should just return `return {state: {color_mode: constants.colorMode[2]}}`. This should avoid the read. 

I assume that when a bulb is e.g. in xy mode and you do a `colortemp_move` it switches to color temp mode thus the color_mode is always 'color_temp`.

sjorge: That could work too yes, I added it to the read because we are already doing one anyway, and it seems to still be one read request according to wireshark with 2 attribute instead of 1 requested. So I don't think this adds extra load on the network.

Koenkk: Did you encounter bulbs which do not support the `colorMode` property?

sjorge: Not yet :) I tested my usual batch of Tint CWS E27, GU10, Trådfri WS GU10, CWS E27, WS FLOAT, WS GUNTARP, Innr RB 250 C, Osram A65 CWS E27.

Koenkk: Good, let's assume for now that every bulb which supports `colortemp_move` also support `colorMode` (otherwise the colorMode could be wrong).

sjorge: image

Until otherwise, I think this is a safe assumption given it is marked M without any additional notes.

PR #2289, opened 2021-02-24
Support for ROBB smarrt: ROB_200-018-0

mackshot:
They use now a different modelID: ROB_200-018-0


mackshot:
Ok, somebody else was faster ;)

PR #2290, opened 2021-02-24
remove erroneously added etop error status code

RubenKelevra:
looks like it originated from here:

https://github.com/drzony/zigbee-herdsman-converters/blob/f3709f9a3b5ed8d3607e921bbe8ebe32714bdac1/converters/fromZigbee.js#L315

### DIFF
--- a/lib/tuya.js
+++ b/lib/tuya.js
@@ -239,7 +239,6 @@ const dataPoints = {
     moesChildLock: 40,
     moesSensor: 43,
     moesSchedule: 101,
-    etopErrorStatus: 13,
     // Neo T&H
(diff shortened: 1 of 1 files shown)


Koenkk:
It's used here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/f5491356376189aa3c5d48997e2dd912613dba59/converters/fromZigbee.js#L2918 , why remove it?

RubenKelevra:
@Koenkk oh. Haven't found that one.

In this case this PR need to change. My point was that it isn't used for the TuYa valves.

I'll give the linked source code a look when I'm at home.

PR #2291, opened 2021-02-24
SmartThings F-MLT-US-2 (multiv4): fix battery readings

o0v7r0o:
I did a try to fix battery readings from F-MLT-US-2, instead of table lookup I introduced a function, here is a comparison with original implementation:

```
  V  | Ori |  Fun
2800 | 100 | 100
2700 | 100 |  98
2600 | 100 |  93
2500 |  90 |  87
2400 |  90 |  81
2300 |  70 |  74
2200 |  70 |  67
2100 |  50 |  59
2000 |  50 |  50
1900 |  30 |  40
1800 |  30 |  29
1700 |  15 |  17
1600 |   1 |   4
1500 |   0 |   0
```
I think It's not bad, at least one can see a trend even if the percentage is not changing in the original table.

Not sure that '3V_1500_2800' option name is good, any feedback is welcome.
PS 'Power source' is still not showing, have to figure out how to fix it.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -8187,7 +8187,7 @@ const devices = [
         description: 'Multipurpose sensor (2016 model)',
         fromZigbee: [fz.temperature, fz.battery, fz.ias_contact_alarm_1, fz.smartthings_acceleration],
         toZigbee: [],
-        meta: {configureKey: 2, battery: {voltageToPercentage: '3V_2500'}},
+        meta: {configureKey: 2, battery: {voltageToPercentage: '3V_1500_2800'}},
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

You can set the power source like this: https://github.com/Koenkk/zigbee-herdsman-converters/blob/cc3574666713f81aaae06937a530b70b4b891d79/devices.js#L6679 (`device.powerSource = Battery';`)

o0v7r0o:
Thanks! Probably a workaround with setting powerSource manually is indeed a good idea here, because my sensor is reporting
`No converter available for 'F-MLT-US-2' with cluster 'genBasic' and type 'readResponse' and data '{"powerSource":0}'`
but for 'battery' should be `{"powerSource":3}`.
I will ask others in the issue if they have the same behaviour.

PR #2292, opened 2021-02-25
Update constants.js

dettrick:
Updating the panelstatus constants to match 8-16 in the zigbee cluster library for PanelStatus field values. The existing implementation has an extra field which does not belong there (invalid_code)

image

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -103,14 +103,13 @@ const armMode = {
     1: 'arm_day_zones',
     2: 'arm_night_zones',
     3: 'arm_all_zones',
-    4: 'invalid_code',
-    5: 'exit_delay',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

DennisVonDerBey:
@Koenkk I think the device Centralite 3400-D relied on this field value. See [docs ](https://www.zigbee2mqtt.io/devices/3400-D.html#disarming-from-the-keypad). Is this a regression, because I am not able to send the `invalid_code` anymore, since this change happened, or has it been replaced by something else?

Koenkk:
@DennisVonDerBey you are right, it's called `exit_delay` now. Updated the docs

PR #2293, opened 2021-02-25
devices: all power_on_behavior to all bulbs

sjorge:
I just found out we are already including the level_config bits on all bulbs, also add power_on_behavior.
Seems to work fine for all tradfri stuff and tint bulbs. I decided to test as the new ikea gateway exposes this functionality for bulbs on newer firmware too.

Like level_config, I did not push it to expose.


Koenkk:
Thanks!

PR #2296, opened 2021-02-25
Added support for Tuya M515EGZT curtain motor

reneklootwijk:
Added support for another version of the curtain pull cord motor.


Koenkk:
Thanks!

PR #2297, opened 2021-02-26
Add a new Legrand device, Legrand 412170 (Drivia with Netatmo connected teleruptor)

lallspach:
Add a new Legrand device, Legrand 412170 (Drivia with Netatmo connected teleruptor).
This is a clone of the existing Legrand 412171 (Drivia with Netatmo connected contactor) with different zigbeeModel and model.
Same feature set, difference is lower power capacity (16A vs 20A).

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14660,6 +14660,27 @@ const devices = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2298, opened 2021-02-26
Fixed reporting for tuya_cover

reneklootwijk:
The M515EGZT curtain motor only reports dp: 2, the mode when changing its tilt position. However the previous code did not report the position for dp:2. I modified the code slightly to report the position also for dp:2.


Koenkk:
Thanks!

Koenkk:
Thanks!

PR #2299, opened 2021-02-27
Separate Ubisys S1 and S1-R to fix seMetering

jd1900:
As described in this issue (fixes https://github.com/Koenkk/zigbee2mqtt/issues/6472) we must separate S1 and S1-R, as S1-R uses endpoint 4 for seMetering.
I'm not sure if I should change something else, but this way it works fine the power measurement and on/off.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -13400,7 +13400,7 @@ const devices = [
 
     // Ubisys
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2310, opened 2021-03-01
devices: Xiaomi bulbs don't support color_temp_startup

sjorge:
Reported as koenkk/zigbee2mqtt#6500


sjorge:
Pushed a 2nd commit with a better fix I think, If not, we can drop the 2nd commit.

Koenkk:
Thanks!

PR #2311, opened 2021-03-01
Hama thermostatic radiator valve 00176592 with modelId w7cahqs

thomaszub:
Hi,

tried to pair with a Hama thermostatic radiator valve 00176592 and the device was not recognized with a modelId w7cahqs. Added the modelId in the fingerprint for the model SEA801-Zigbee/SEA802-Zigbee and now the device works flawlessly.

Kind regards

Thomas


Koenkk:
Thanks!

PR #2312, opened 2021-03-02
[BHT-002-GCLZB] Enable Local Temperature Calibration 

Minims:
Enable Local Temperature Calibration in zigbee2mqtt interface for Moes Thermostat BHT-002-GCLZB.

It was already working while pushing a payload via MQTT, but more user friendly via the interface.

Thanks @presslab-us for giving me the right direction 👍


Koenkk:
Thanks!

PR #2313, opened 2021-03-02
support Climax IR-9ZBS-SL motion sensor

nebman:
added support for "Climax IR-9ZBS-SL" motion detector.  
from looking at the pictures it became clear, this is the same product as Bitron AV2010/22 and adding the zigbee model number was the only thing needed to get the device working.  

I hope that this is the correct use of the property "whiteLabel", unfortunately i couldn't find any documentation on that.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -8625,13 +8625,14 @@ const devices = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2314, opened 2021-03-02
Cleanup/Optimizaiton for Viessmann ZK03840

sjorge:
While looking into #2275  I noticed there were a lot of manufacturer specific attributes, I have not gotten very far with those yet. 

However digging into the dumps I discovered:
- battery percent remaining reporting
- keypadLockout was read by the gateway and written, but not setup for reporting (makes sense, can't change it unless via the gateway)
- system_mode is limited to 'heat' and 'sleep'
- running_mode seems to return UNSUPPORTED_ATTRIBUTE on both read or reporting setup.

This PR does some general house keeping for the device configuration of the ZK03840, no functionality is gained or lost. (Just one broken expose)

Well technically we can now set the device to 'sleep' which seems to be a soft 'off'. Also battery reporting now works so I guess we gained some functionality :)

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -15934,12 +15934,13 @@ const devices = [
         meta: {configureKey: 1},
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
-            await reporting.bind(endpoint, coordinatorEndpoint, ['genBasic', 'genPowerCfg', 'genIdentify', 'genTime', 'genPollCtrl',
-                'hvacThermostat', 'hvacUserInterfaceCfg']);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genBasic', 'genPowerCfg', 'genIdentify', 'genTime', 'hvacThermostat']);
             await reporting.thermostatTemperature(endpoint);
             await reporting.thermostatOccupiedHeatingSetpoint(endpoint);
             await reporting.thermostatPIHeatingDemand(endpoint);
-            await reporting.thermostatKeypadLockMode(endpoint);
+
+            // read keypadLockout, we don't need reporting as it cannot be set physically on the device
+            await endpoint.read('hvacUserInterfaceCfg', ['keypadLockout']);
         },
     },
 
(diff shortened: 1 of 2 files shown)


devices.js
```diff
@@ -15930,16 +15930,18 @@ const devices = [
         toZigbee: [tz.thermostat_local_temperature, tz.thermostat_occupied_heating_setpoint, tz.thermostat_control_sequence_of_operation,
             tz.thermostat_system_mode, tz.thermostat_keypad_lockout],
         exposes: [exposes.climate().withSetpoint('occupied_heating_setpoint', 7, 30, 1).withLocalTemperature()
-            .withSystemMode(['off', 'auto', 'heat']).withRunningState(['idle', 'heat'], ea.STATE)],
-        meta: {configureKey: 1},
+            .withSystemMode(['heat', 'sleep'])],
```
Koenkk, line 15933: Can sleep be renamed to `off`? HA doesn't support this.

sjorge: I don't think we should: https://github.com/Koenkk/zigbee-herdsman-converters/blob/67fd494874dbc5e14998a2252b5942a3bf2b63c4/lib/constants.js#L33

It's set to 9, which is sleep according to the ZCL too. This seems like something to do in the hass plugin instead.

Koenkk: I see, thanks

sjorge: I'm also not sure if this has other effects yet. What I observed:
TRV as set to heat and actually heating the room from 20 -> 24C

Set it to sleep: 1. closed valve and stopped heating, 2. reporting seems to happen less frequent.

I guess it might be waiting for it to hit max reporting timer... or due to it not heating localTemp changed less often... that could also be the case. Although the room still got warmer due to the radiator still being hot.

sjorge: @Koenkk would something like

https://github.com/Koenkk/zigbee2mqtt/blob/52b80221730b7673e98a54385ea4c6ec332db850/lib/extension/homeassistant.js#L166

```
if (mode.values.includes('sleep')) {
 // remove 'sleep' from mode.values ?
}
```

Not sure there is a javascript/node version of pythons `mylist.remove('sleep')`, basically if hass.io doesn't do sleep, we can remove it there right? That should still allow things like the frontend or other expose consumers to set 'sleep'.

And hass.io in this case will only have 'heat'.

Koenkk: Yes, that's fine for me. Could you make a PR?

sjorge: I could, but I don't use hass.io so can't really test it. Will be tomorrow though.

PR #2316, opened 2021-03-02
Add support for Linkind contact sensor battery percent reporting

Tediore:
Also correct the battery reporting for the 1-key remote (expose battery sensor)


Koenkk:
Thanks!

PR #2318, opened 2021-03-03
Allow reading of keypad_lockout and cleanup

sjorge:
I wanted to expose keypad_lockout on the Viessman TRV, I noticed that we did not yet have a convertGet... then I noticed what we send to convertSet doesn't match what we return in fromZigbee 😭 

- add convertGet to keypad_lockout
- add return to convertSet for keypad_lockout
- use same constant mapping in fromZigbee for keypad_lockout
- update expose


lib/exposes.js
```diff
@@ -468,7 +468,7 @@ module.exports = {
         humidity: () => new Numeric('humidity', access.STATE).withUnit('%').withDescription('Measured relative humidity'),
         illuminance: () => new Numeric('illuminance', access.STATE).withDescription('Raw measured illuminance'),
         illuminance_lux: () => new Numeric('illuminance_lux', access.STATE).withUnit('lx').withDescription('Measured illuminance in lux'),
-        keypad_lockout: () => new Lock().withState('keypad_lockout', '1', '0', 'Enables/disables physical input on the device', a.STATE_SET),
+        keypad_lockout_lock1: () => new Lock().withState('keypad_lockout', 'lock1', 'unlock', 'Enables/disables physical input on the device', a.STATE_ALL),
```
Koenkk, line 471: Why rename this to `keypad_lockout_lock1`?

sjorge: In case we encounter  another combo say lock2 and unlock in the future. I can rename it back though.

Koenkk: I think it's good to rename it back, we can always change in the future

Koenkk:
Thanks! 🎖️

PR #2322, opened 2021-03-05
Add support for power on behavior to Linkind plugs

Tediore:
Also enable temp/humidity reporting for Iris 1117-S motion sensors
Also fix battery reporting for XHS2-UE contact sensor (doesn't support the `batteryPercentageRemaining` attribute of the `genPowerCfg` cluster, but it supports `batteryVoltage`)


Koenkk:
Thanks!

PR #2323, opened 2021-03-05
LEDVANCE AC10402

sjorge:
See https://github.com/Koenkk/zigbee2mqtt/issues/6581

@Koenkk we're still missing the colorTempMin/Max :( can we try to read them for unknow devices with the lighthingColorCtrl cluster? I looks others are read like currentHue,Saturation,X,Y,...


sjorge:
Got the min/max values. Should be complete now.

Koenkk:
Thanks!

PR #2325, opened 2021-03-06
Add support for VL-ZP07017/VL-XJ001 Livolo roller blind motor

aesedepece:
## Warnings

- This PR is still a work-in-progress.
- This works for Livolo roller blind motor model `VL-ZP07017` or `VL-XJ001` ([this on AliExpress][ali]).
- Livolo devices require an ad-hoc `pan_id` / `ext_pan_id` and only operate on channel `26`. [More info here][livolo-panid].
- I have tested this with a dedicated CC2531 coordinator.
- I don't own the original Livolo gateway, so I had to figure out some stuff by pure trial and error (e.g. had to bruteforce the `0x0401` payload).
- When I bought this model, the seller tried to sell me Livolo's gateway and warned me that this wouldn't work with 3rd party gateways. I stubbornly told them that I knew what I was doing. Obviously I didn't, so I'm trying hard here to make this work hahahaha, 

## What works

- Device can be discovered and interviewed
- Position (percentage of cover opening) and state (`OPEN` / `CLOSE`) events are received when the motor stops.
- Position can be changed through MQTT, so the main feature works! :D

## What does not work

- Everything stops working once the device goes out of pairing mode. 
- Reading position/state on demand or at pairing time.
- I have never used this with the original Livolo gateway and app, so I don't really know what the original features are. From the manual, I can imagine:
  - Stopping the motor while it's moving.
  - Change drive direction of the motor.
  - Set upper / lower limit for the roller blind.
  - Set moving speed.
  - Reading current value of all these settings.
  - Battery / power / solar panel status?

## Next steps

I know the issue about losing connection after a minute or so also affected others trying to implement the [Livolo dimmer][dimmer] and [Livolo switch][switch]. I don't know why in this case none of their solutions worked. Communication is immediately dropped once pairing is over. 

This device has a LED indicator on the front that blinks blue and red while pairing. After 1 minute, it stays red for 3 seconds. To me, it looks like the device is considering that it couldn't pair correctly because it was waiting for some special configuration command from the gateway and it never received it.

Do you have any clue on this? I know @whitepail had a similar "heartbeat" problem with the dimmer and it was [magically fixed by pairing once with the original gateway][fix].

What do you think I should do next? I'm trying at all costs not to buy the Livolo gateway just to pair it once and get it properly configured. If I buy it, I could try to sniff the traffic and look for the message that I'm missing. But I think @whitepail already tried and didn't find anything special, so I don't think I have better luck with that :man_shrugging: 

[ali]: https://es.aliexpress.com/item/1005001702490864.html
[livolo-panid]: https://www.zigbee2mqtt.io/devices/TI0001-dimmer.html#important
[dimmer]: https://github.com/Koenkk/zigbee2mqtt/issues/5516
[switch]: https://github.com/Koenkk/zigbee2mqtt/issues/592
[fix]: https://github.com/Koenkk/zigbee2mqtt/issues/5516#issuecomment-770883164

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2042,13 +2042,13 @@ const converters = {
         type: ['raw'],
         convert: (model, msg, publish, options, meta) => {
             const dp = msg.data[10];
-            if (msg.data.indexOf(Buffer.from([0x7a, 0xd1])) === 0) { // Normal operation messages
+            if (msg.data[0] === 0x7a & msg.data[1] === 0xd1) {
                 const reportType = msg.data[12];
                 switch (dp) {
                 case 0x0c:
                 case 0x0f:
                     if (reportType === 0x04) { // Position report
-                        const position = options.invert_cover ? 100 - msg.data[13] : msg.data[13];
+                        const position = meta.state.motor_direction === 'FORWARD' ? msg.data[13] : 100 - msg.data[13];
                         const state = position > 0 ? 'OPEN' : 'CLOSE';
                         return {position, state};
                     }
@@ -2058,9 +2058,9 @@ const converters = {
                     } else if (reportType === 0x13) { // Direction report
                         const direction = msg.data[13];
                         if (direction === 0x70) {
-                            options.invert_cover = false;
+                            return {motor_direction: 'REVERSE'};
                         } else if (direction === 0xf0) {
-                            options.invert_cover = true;
+                            return {motor_direction: 'FORWARD'};
                         }
                     }
                     break;
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2041,15 +2041,37 @@ const converters = {
         cluster: 'genPowerCfg',
         type: ['raw'],
         convert: (model, msg, publish, options, meta) => {
-            const stateHeader = Buffer.from([122, 209]);
-            if (msg.data.indexOf(stateHeader) === 0) {
-                const dp = msg.data[10];
-                if (dp === 12 || dp === 15) {
-                    const position = 100 - msg.data[13];
-                    const state = position > 0 ? 'OPEN' : 'CLOSE';
-                    return {position, state};
-                } else { // TODO: Unknown dps
-                    meta.logger.warn(`livolo_cover_state: Unhandled DP ${dp} for ${meta.device.manufacturerName}: ${msg.data.toString('hex')}`);
+            const dp = msg.data[10];
+            if (msg.data.indexOf(Buffer.from([0x7a, 0xd1])) === 0) { // Normal operation messages
+                const reportType = msg.data[12];
+                switch (dp) {
+                case 0x0c:
+                case 0x0f:
+                    if (reportType === 0x04) { // Position report
+                        const position = options.invert_cover ? 100 - msg.data[13] : msg.data[13];
+                        const state = position > 0 ? 'OPEN' : 'CLOSE';
+                        return {position, state};
+                    }
+                    if (reportType === 0x12) { // Speed report
+                        const motorSpeed = msg.data[13];
+                        return {motor_speed: motorSpeed};
+                    } else if (reportType === 0x13) { // Direction report
+                        const direction = msg.data[13];
+                        if (direction === 0x70) {
+                            options.invert_cover = false;
+                        } else if (direction === 0xf0) {
+                            options.invert_cover = true;
+                        }
+                    }
+                    break;
+                case 0x02:
+                case 0x03:
+                    // Ignore special commands used only when pairing, as these will rather be handled by `onEvent`
+                    return null;
+                default:
+                    // Unknown dps
+                    meta.logger.warn(`livolo_cover_state: Unhandled DP ${dp} for ${meta.device.manufacturerName}: \
+                     ${msg.data.toString('hex')}`);
                 }
             }
         },
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2037,6 +2037,23 @@ const converters = {
             }
         },
     },
+    livolo_cover_state: {
+        cluster: 'genPowerCfg',
+        type: ['raw'],
+        convert: (model, msg, publish, options, meta) => {
+            const stateHeader = Buffer.from([122, 209]);
+            if (msg.data.indexOf(stateHeader) === 0) {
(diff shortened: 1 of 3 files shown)


aesedepece:
I finally gave in and ordered a Livolo gateway so as to try to sniff the pairing messages for this particular device, which seems to be specially quirky.

As a side note, I just realized that there was no need to bruteforce the cluster (`0x0401`) as I was already getting those i the position notification messages, but I didn't noticed because of the endianness. E.g.: `7ad1________________0c0104044c0b` is a position report `0xc0`  for cluster `0x0401` stating that position is `0x44` (68%).

Following the same logic, I can guess that drive direction of the motor can be reversed using cluster `0x1301`, as every time I change this setting on the device (hold up and down buttons for 3 seconds), I'm getting a state report in the likes of `7ad1________________0c0113800c0b`. Haven't tried it yet though, because I'm giving more priority to solving the "heartbeat" issue.

During pairing, there is a message that comes from the device every second or so. It's always the same, regardless of the state of the device: `7ad1________________0200150000`. Only once I did get something different: `7ad1________________0200002002`.

Also during pairing, there's one message that is received only once: `7cd215d880________________131500`. This is the one I'm using for telling this specific device from other Livolo devices that use the same zigbee model. In this case, `0x131500` corresponds to `[19, 21, 00]` as used in `livolo_switch_state_raw`.

Another message that I've seen once in a while that remains unknown to me is `7ad1________________00200020000b`.

So in general, these are the datapoints I've been able to discover so far:
- `00` / `0x00`: unknown (see above)
- `02` / `0x02`: used during pairing.
- `12` / `0x0c`: position report (received every time the motor stops)
- `15` / `0x0f`: position ack (received as a response every time you send a position set command)

whitepail:
Whenever you will get the gateway, could you please not only sniff zigbee traffic, but also sniff debug log from gateway uart and publish it here (see my post regarding gateway uart earlier).

I wasn't able to write uart debug during first pairing so it's still unclear what messages has gateway sent to remove "heartbeat issue".

aesedepece:
Thanks for the tip @whitepail, I'll totally try to do that! :+1:

aesedepece:
@whitepail I'm trying to read from the STM32F412 @ 33600 bauds, and I'm just getting this upon powering up:

```
ed  17  29  23  d8  80  b3  a7  ff  31  6d  a9  88  38  ca  ca  21  a6  42  d8  fb  21  47  4c  7c  08  2b  3b  02  88  d8  08  d9  57  
```

But then nothing shows up later when trying to pair.

aesedepece:
Ok, I got it all to work.

Adding this in the `onEvent` handler totally did the trick for completing the pairing successfully and preventing the device from ever leaving the network:
```js
if (data.cluster === 'genPowerCfg' && data.type === 'raw') {
    const dp = data.data[10];
    if (data.data.indexOf(Buffer.from([0x7a, 0xd1])) === 0) {
        const endpoint = device.getEndpoint(6);
        if (dp === 0x02) {
            const options = {manufacturerCode: 0x1ad2, disableDefaultResponse: true, disableResponse: true,
                reservedBits: 3, direction: 1, writeUndiv: true};
            const payload = {0x0802: {value: [data.data[3], 0, 0, 0, 0, 0, 0], type: 0x8f}};
            await endpoint.readResponse('genPowerCfg', 0xe9, payload, options);
        }
    }
}
```

Namely, what was missing is the response to `7ad1XXYY____________00200020000b`, which needs to be `7cd21ae901020800XXYY000000000000`. Note that `XX` and `YY` are the first and second bytes of the IEEE address of the device.

aesedepece:
Some changes are still missing to correct the direction of the motor. I'll push them later.

aesedepece:
I'll be adding the `STOP` command next so the integration with HA feels more complete.

I've also figured out how to adjust motor direction, speed and upper/bottom limits, but I think I'll leave those for a future PR, as I need to think first what's the best way to expose them.

aesedepece:
Ok, after a deep dive into zigbee-herdsman, I noticed that the `STOP` option—among many other settings for this device—will not be possible at this time because support for the global `writeStruct` (`0x0f`) ZCL command is missing.

I think I'll leave this PR as it is for now, ask for merge, and start working on supporting some of the missing ZCL commands on zigbee-herdsman if @Koenkk deems it appropriate.

Koenkk:
Thanks!

whitepail:
I've tried to reuse "readResponse" login message that @aesedepece emulated for cover to bypass "heartbeat problem".
Looks like  it doesn't help for dimmer as message format differs between Z2M and Livolo GW.
In livolo gw log it has format 
7c d2 1a e9 01 02 08 **01** 22 **01** 00 00 00 00 00 00
Emulated by z2m
7c d2 1a e9 01 02 08 **00** 22 **00** 00 00 00 00

The difference is attr_status. If you set attr_status to "01", attribute value will be deleted according to foundation.ts
I was able to reproduce the heartbeat problem by unbinding device in Livolo app and after that binding in Z2M.

Will continue to work on this heartbeat problem.

alanieska:
livolo

Hello, could someone please explain to me where or what I should adjust for the entities up and down to work? because when you click on or off the device does nothing.
Thank you

grinco:
Folks, is there any way to get rid of the "heartbeat" problem except connecting the device to their original gateway? If not - would anyone be willing to "fix" mine? I'd really want to avoid buying a gateway for that - would rather pay for shipping to/from any of you.

captaindaveman:
Did anyone get this to work? I bought one on AliExpress thinking it was the zigbee2mqtt supported model Livolo TI0001, but wasn't. I'm using HA running with a CC2531, when I put it into pairing mode and permit join in zigbee2mqtt it doesn't even detect it.

grinco:
Any luck anyone? It's been collecting dust for over 2 years, I'm considering just throwing it away at that point...

whitepail:
Did you try with latest z2m version?
I think this problem has been resolved, at least all my recent devices
paired successfully without heartbeat problem.

вт, 21 февр. 2023 г., 21:39 grinco ***@***.***>:

grinco:
Yeah, tried running it on the latest z2m-edge. The device pairing completes in z2m, but the device keeps blinking the LED like it's still pairing. I can control it while it blinks the LED, however, once the pairing times out on the device is not controllable by z2m any more...

PR #2326, opened 2021-03-06
Add support for onWithTimedOff command to IKEA E1603/E1702

Jabe:
I've recently come across the _onWithTimedOff_ command which I wanted to add to the IKEA [wall plug](https://www.zigbee2mqtt.io/devices/E1603_E1702.html). In fact, the command is also supported by all my IKEA lamps.

Short summary: the command will on a device for the value (in 1/10 secs) specified by the parameter _on_time_ after which the device will turn off. There is also the param _off_wait_time_ which will discard any further _onWithTimedOff_  commands within the specified time. This could be very useful if you want to temporarily force off a device activated by a motion detector for example. Subsequent _onWithTimedOff_ commands update the respective timings.

As a starting point I've built this simple converter and changed the toZigbee spec of the wall plug. It is also my first time exploring the internals of z2m.

I've decided to stick to the original wording used by the zigbee spec:
* Turn on for 1 sec: `{"state":"onWithTimedOff","on_time":10}`
* Keep off for 1 min: `{"state":"onWithTimedOff","off_wait_time":600}`

It works but there are probably some shortcomings:
1. After sending the command the command doesn't return a new state. This is not trivial because of the whole _off_wait_time_ thing.
2. Maybe this should be part of the generic _on_off_ and _light_onoff_brightness_ converters? 😮 😄 

Cheers!

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -164,6 +164,40 @@ const converters = {
             await entity.read('genOnOff', ['onOff']);
         },
     },
+    on_off_timed: {
+        key: ['state', 'on_time', 'off_wait_time'],
+        convertSet: async (entity, key, value, meta) => {
+            const {message} = meta;
+            const state = message.hasOwnProperty('state') ? message.state : null;
+
+            if (state !== 'onWithTimedOff') {
+                return await converters.on_off.convertSet(entity, key, value, meta);
+            }
+
+            const onTime = message.hasOwnProperty('on_time') ? message.on_time : 0;
+            const offWaitTime = message.hasOwnProperty('off_wait_time') ? message.off_wait_time : 0;
+
+            if (!Number.isInteger(onTime)) {
+                throw Error('The on_time value must be convertible to an integer');
+            }
+            if (!Number.isInteger(offWaitTime)) {
+                throw Error('The off_wait_time value must be convertible to an integer');
+            }
+
+            await entity.command(
+                'genOnOff',
+                'onWithTimedOff',
+                {
+                    ctrlbits: 0,
+                    ontime: onTime,
+                    offwaittime: offWaitTime,
+                },
+                utils.getOptions(meta.mapped, entity));
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('genOnOff', ['onOff']);
+        },
+    },
     cover_via_brightness: {
         key: ['position', 'state'],
         convertSet: async (entity, key, value, meta) => {
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -152,47 +152,37 @@ const converters = {
         convertSet: async (entity, key, value, meta) => {
             value = value.toLowerCase();
             utils.validateValue(value, ['toggle', 'off', 'on']);
-            await entity.command('genOnOff', value, {}, utils.getOptions(meta.mapped, entity));
-            if (value === 'toggle') {
-                const currentState = meta.state[`state${meta.endpoint_name ? `_${meta.endpoint_name}` : ''}`];
-                return currentState ? {state: {state: currentState === 'OFF' ? 'ON' : 'OFF'}} : {};
-            } else {
-                return {state: {state: value.toUpperCase()}};
-            }
-        },
-        convertGet: async (entity, key, meta) => {
-            await entity.read('genOnOff', ['onOff']);
-        },
-    },
-    on_off_timed: {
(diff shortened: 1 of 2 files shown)


sjorge:
Those are part of ZCL so they should indeed probably be part of the onoff/onoff_brighness converters.

I think they should just be part of on_off and light_onoff_brightness and treat them like `transition` is currently treated, called it `state_timeout` or something (I'd personally use the same for both the on and off time as it's impossible to set both state on and off at the same time.

So it would look sometime like: `{"state": "ON", "timeout": 10}` and `{"state": "OFF", "timeout": 10}`

I'd probably make the unit 'seconds' and just do the multiplication. So value * 10. But tread the value like a float so `0.1` would still mean 1/10 of a second.

Koenkk:
Agree with @sjorge , then this code can be merged into the already existing `on_off` converter.

Jabe:
I though about using ON/OFF + timeout but one nifty feature of this command is setting _on_time_ and _off_wait_time_ at the same time.

Imaging a motion detector that issues this command:
`{"state":"onWithTimedOff","on_time":600,"off_wait_time":3000}`
So the lights turn on for 1 minute on motion. Now if you were to turn off the lights manually (_state_=off), the motion detector will not turn them back on for 5 minutes.

But I realize that's pretty advanced. Maybe we should go the divide and conquer route to cover probably 99.9% of cases:
`{"state":"on","timeout":60.0}`
That's very intuitive and aligns well with other commands. And then also allow an additional unique property like _off_wait_time_ or maybe simply _off_time_ for the sake of feature completeness.
Finally _state_=off + _timeout_ or _off_wait_time_ would throw an error because otherwise that would actually turn the device on.

sjorge:
> I though about using ON/OFF + timeout but one nifty feature of this command is setting _on_time_ and _off_wait_time_ at the same time.
> 

Ooh that is pretty neat! 

I pulled up the ZCL again:
```
4434 3.8.2.2.3 OnTime Attribute
4435 The OnTime attribute specifies the length of time (in 1/10ths second) that the “on” state SHALL be maintained before
4436 automatically transitioning to the “off” state when using the On with timed off command. If this attribute is set to
4437 0x0000 or 0xffff, the device SHALL remain in its current state.

4438 3.8.2.2.4 OffWaitTime Attribute
4439 The OffWaitTime attribute specifies the length of time (in 1/10ths second) that the “off” state SHALL be guarded to
4440 prevent an on command turning the device back to its “on” state (e.g., when leaving a room, the lights are turned off
4441 but an occupancy sensor detects the leaving person and attempts to turn the lights back on). If this attribute is set to
4442 0x0000, the device SHALL remain in its current state.
```

I don't feel very strong about using `timeout` as name.  Perhaps we can just use on_time and off_wait_time but use **seconds** as a more natural unit as described earlier.

And add some clarifying docs to the website on how on_time and off_wait_time work.

converters/toZigbee.js
```diff
@@ -148,16 +148,39 @@ const converters = {
         },
     },
     on_off: {
-        key: ['state'],
+        key: ['state', 'on_time', 'off_wait_time'],
         convertSet: async (entity, key, value, meta) => {
-            value = value.toLowerCase();
```
Koenkk, line 153: Please don't remove `value.toLowerCase();` here, now e.g. `ON` doesn't work anymore.

Koenkk:
Thanks! Could you also make a PR for https://www.zigbee2mqtt.io/devices/E1603_E1702.html#switch ? If you do I will make sure it also lands in the other pages (all other switches and bulbs)

Jabe:
Thanks for merging!! I'll have a look at the docs.

earwin:
A bit of an offtopic, if that's okay.
@Jabe, are you using this with Home Assistant? If so, how do you pass the state parameters?

Jabe:
> @Jabe, are you using this with Home Assistant? If so, how do you pass the state parameters?

Yes, wrap it in a script like this. It publishes to mqtt directly.
```yaml
schlafzimmer_heizdecke_60m:
  sequence:
    - service: mqtt.publish
      data:
        topic: "zigbee2mqtt/Steckdose 1/set"
        payload: >
          { "state": "onWithTimedOff", "on_time": 3600 }
```

Then simply call the script in lovelace. This turns on my wife's heating blanket for 1 hour btw. 😃😉

earwin:
> It publishes to mqtt directly.

I see. Thanks a bunch!

PR #2328, opened 2021-03-06
fromZigbee: use hasOwnProperty as constants.keypadLockoutMode is not an array

sjorge:
I messed up and forgot to include a commit, sorry about that.

constants.keypadLockoutMode is not an array but a Object so need hasOwnProperty() instead of includes()... or alternatively we could just do

```
result.keypad_lockout = constants.keypadLockoutMode[msg.data['keypadLockout']];
```

As there should really not be a value not in the constant... but given manufacturer like to play loosy goosy with the ZCL, I think the current check might be better.


Koenkk:
Thanks!

PR #2334, opened 2021-03-07
[BHT-002-GCLZB] Add more lisibility on program

Minims:
Add more lisibility on program values return by fromZigbee.js


Koenkk:
Thanks!

PR #2337, opened 2021-03-08
devices: Danfoss 014G2461 battery reporting and read keypad_lockout

sjorge:
My unit finally arrived (to compare it to the viessmann), a few optimizations can be carried over.
- read keypad_lockout during configure
- setup battery reporting

(We could probably tweak the other reporting values to be less aggressive as on the viessmann, but I opted to leave them as is for now.)


Koenkk:
Thanks!

PR #2338, opened 2021-03-08
Update 067776 to fix inverted working mode

jd1900:
coverInverted is not needed. It actually breaks device making unusable through HA, but works in Zigbee2MQTT.

I've observed a few things that don't work in this device properly. I'll open an issue to try to get some help.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14860,7 +14860,7 @@ const devices = [
             // support binary report on moving state (supposed)
             fz.legrand_binary_input_moving, fz.cover_position_tilt],
         toZigbee: [tz.cover_state, tz.cover_position_tilt, tz.legrand_identify, tz.legrand_settingAlwaysEnableLed],
-        meta: {configureKey: 1, coverInverted: true},
+        meta: {configureKey: 1},
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genBinaryInput', 'closuresWindowCovering', 'genIdentify']);


Koenkk:
When sending to `zigbee2mqtt/DEVICE_FRIENDLY_NAME/set` with payload `{"position": 100}`, is the cover fully open or closed?

jd1900:
{"position": 100} fully opens the cover.
{"position": 1-99} closes the cover but the it state to that value, being able to open or close the cover.
{"position": 0} closes the cover and set it value to closed, not being able to press "close" in the GUI.

jd1900:
I've added this issue which covers or the related issues with this device:
https://github.com/Koenkk/zigbee-herdsman-converters/issues/2339

With some help I could try to fix or the errors.

Koenkk:
> {"position": 100} fully opens the cover.

This is incorrect, 100 should fully close the cover. So coverInverted should stay.

Do you maybe have covert invert in your configuration.yaml?

In your post is see e.g. 

> For instance, I have this blind half open, however, I can't choose open or close, it only gets the last value when fully opened or fully closed. In this case, it was fully closed. After that, in HA only open is possible:

This is a HA issue, a curtain is either open or closed from the HA perspective.

jd1900:
> Do you maybe have covert invert in your configuration.yaml?
No. 

> This is a HA issue, a curtain is either open or closed from the HA perspective.
I don't get it.
For instance, when I add coverInverted back, and I set position to 100, this is what I get in Home Assistant:
image
The blinds are closed, but I can't open them. 

Without coverInverted when I close the blinds, I have the controls to open them in HA. Which is right.

So..., basically, when removing coverInverted it works somehow right, without it, it is totally messes.

Koenkk:
Sorry I was wrong, 100 should be open indeed, 0 closed. So this can be merged.

jd1900:
> Sorry I was wrong, 100 should be open indeed, 0 closed. So this can be merged.

It now makes sense. I was a bit clueless.

eRJe79:
I don't get it. I'm using the same Legrand 067776 cover switch with Zigbee2MQTT v1.21.1-1 and by default the shutter is behaving in reverse. When the shutter is closed, in the Zigbee2MQTT dashboard the position is 100, also in the HA dashboard the position is 100 but the shutter is reported open. Like in the post above.

Adding invert-cover to the configuration.yaml or devices.yaml doesn't make any change.

Is this issue suppose to be fixed with [667e42a](https://github.com/Koenkk/zigbee-herdsman-converters/commit/667e42a82d897b2bbb12ed5c559c247638c44944)?

eRJe79:
Sorry, disregard my message, I didn't notice I was in the wrong repository. Not using Herdsman.

PR #2343, opened 2021-03-09
SmartThings F-MLT-US-2 (multiv4): force power source

o0v7r0o:
Follow up for https://github.com/Koenkk/zigbee-herdsman-converters/pull/2291
Confirmed here https://github.com/Koenkk/zigbee2mqtt/issues/6021 that the sensor indeed reports an 'unknown' power source.


Koenkk:
Thanks!

PR #2347, opened 2021-03-11
devices: SNZB-02 and SNZB-04 should use batteryPercentageRemaining

sjorge:
Both SNZB-02 and SNZB-4 have batteryPrecentageRemaining in the database.db and seem to report it daily.
This would explain the battery % variang a bit as mentioned in koenkk/zigbee2mqtt#6657

The SNZB-02 was easy to configure most will do it on the first new message or when the button is hit.
The SNZB-04 was annoying, you need to either remove + join or remove the battery for it to  take the new reporting values. They also seem to apply when either batteryVoltage or batteryPercentageRemaining come in via attReport. But they don't apply when the ias_contact_alarm is triggered.

The configureKey for SNZB-04 was not bumpbed to not have to many errors in the logs. But even without reconfigure batteryPercentageRemaining is reported, usually shut before batteryVoltage, so it should be fine.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14230,6 +14230,7 @@ const devices = [
         vendor: 'SONOFF',
         whiteLabel: [{vendor: 'eWeLink', model: 'RHK06'}],
         description: 'Contact sensor',
+        exposes: [e.contact(), e.battery_low(), e.tamper(), e.battery(), e.battery_voltage()],
         fromZigbee: [fz.ias_contact_alarm_1, fz.battery],
         toZigbee: [],
         meta: {configureKey: 1},
@@ -14239,7 +14240,6 @@ const devices = [
             await reporting.batteryVoltage(endpoint);
             await reporting.batteryPercentageRemaining(endpoint);
         },
-        exposes: [e.contact(), e.battery_low(), e.tamper(), e.battery()],
     },
     {
         zigbeeModel: ['WB01', 'WB-01'],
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -14230,15 +14230,16 @@ const devices = [
         vendor: 'SONOFF',
         whiteLabel: [{vendor: 'eWeLink', model: 'RHK06'}],
         description: 'Contact sensor',
+        exposes: [e.contact(), e.battery_low(), e.tamper(), e.battery(), e.battery_voltage()],
         fromZigbee: [fz.ias_contact_alarm_1, fz.battery],
         toZigbee: [],
-        meta: {configureKey: 1, battery: {voltageToPercentage: '3V_2500'}},
+        meta: {configureKey: 1},
```
Koenkk, line 14236: Any reason not to bump `configureKey` here?

sjorge: Yes, see the PR comment? 
> The SNZB-04 was annoying, you need to either remove + join or remove the battery for it to take the new reporting values. They also seem to apply when either batteryVoltage or batteryPercentageRemaining come in via attReport. But they don't apply when the ias_contact_alarm is triggered.

It seems the device is reporting already, but forcing everyone to pry them from the location and remove/insert the battery or use a pin to reset them seems like a bad idea.

Koenkk: Sorry missed that. Thanks!

PR #2351, opened 2021-03-11
Enable window_open detection for Viessman TRV

sjorge:
Not sure what the other values for viessmannCustom0 are yet, so not renaming that at this time.

Reporting values taken from the pcap.


Koenkk:
Thanks!

PR #2354, opened 2021-03-11
Fix Tuya cover state being interpreted as position

DomiStyle:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/6601

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1542,9 +1542,7 @@ const converters = {
             const value = tuya.getDataValue(msg.data.datatype, msg.data.data);
 
             switch (dp) {
-            case tuya.dataPoints.state: // Confirm opening/closing/stopping (triggered from Zigbee)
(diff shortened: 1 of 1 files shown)


DomiStyle:
Since state and coverChange ware previously used to determine the running attribute this will break the running attribute on older models like mine. Since it doesn't really serve a purpose it shouldn't be an issue.

Newer models will report the running attribute via coverPosition.

Koenkk:
Thanks!

blowbuh:
Hi, my setup ist the following: Debian + Home Assistant (supervised) incl. AddOns (Z2MQTT + MQTT broker). 

I've used the auto-discovery function and was able tu successfully implement this device. However, I cannot see any position data at all, nor am I able to adjust this value. When I try to move the position slider in my Z2MQTT add-on nothing happens.
 
At least I can close and fully open it when I press the up and down arrows.
 
Is it this based on the bug described here or do I have to configure it manually to be be able to see and set position data?

PR #2355, opened 2021-03-11
Support for LK FUGA Wiser Wireless Double Relay

niklascp:
This PR adds support for LK FUGA Wiser Wireless Double Relay:
https://zigbeealliance.org/zigbee_products/lk-fuga-wiser-wireless-double-relay/

I have not been able to get other action than single press from the buttons. I hope in the future also to be able to detect long presses and double presses. From the factory the 4 buttons are bound to the two relays. If you want to decouple them you have to unbind the relation.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14857,7 +14857,7 @@ const devices = [
     },
     {
         zigbeeModel: ['LK Switch'],
-        model: 'LK Switch',
+        model: '545D6514',
         vendor: 'Schneider Electric',
         description: 'LK FUGA Wiser Wireless Double Relay',
         meta: {multiEndpoint: true, configureKey: 1},
--- a/devices.js
+++ b/devices.js
@@ -14855,6 +14855,29 @@ const devices = [
             await reporting.currentPositionLiftPercentage(endpoint);
         },
     },
+    {
+        zigbeeModel: ['LK Switch'],
+        model: 'LK Switch',
+        vendor: 'Schneider Electric',
+        description: 'LK FUGA® Wiser wireless double relay',
+        meta: {multiEndpoint: true, configureKey: 1},
+        fromZigbee: [fz.on_off, fzLocal.wiser_fuga_relay_command],
+        toZigbee: [tz.on_off],
+        endpoint: (device) => {
+            return {'l1': 1, 'l2': 2, 's1': 21, 's2': 22, 's3': 23, 's4': 24};
+        },
+        exposes: [e.switch().withEndpoint('l1'), e.switch().withEndpoint('l2'), e.action(['button_*_single'])],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            device.endpoints.forEach(async (ep) => {
+                if (ep.outputClusters.includes(6)) {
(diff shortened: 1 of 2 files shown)


devices.js
```diff
@@ -14855,6 +14855,29 @@ const devices = [
             await reporting.currentPositionLiftPercentage(endpoint);
         },
     },
+    {
+        zigbeeModel: ['LK Switch'],
+        model: '545D6514',
+        vendor: 'Schneider Electric',
+        description: 'LK FUGA Wiser Wireless Double Relay',
+        meta: {multiEndpoint: true, configureKey: 1},
+        fromZigbee: [fz.on_off, fz.wiser_fuga_relay_command],
```
Koenkk, line 14864: Instead of `fz.wiser_fuga_relay_command` please re-use `fz.command_on, fz.command_off`.

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/667

devices.js
niklascp: Hi @Koenkk, thanks for your review!

The reason I didn't use the existing `fz.command_on` and `fz.command_off` is that each button will only generate one of the types `commandOn` or `commandOff` on a separate endpoint. For example, pushing each button once generates:

```
Zigbee2MQTT:debug 2021-03-12 20:52:34: Received Zigbee message from '0x60a423fffe50109c', type 'commandOn', cluster 'genOnOff', data '{}' from endpoint 21 with groupID 0
Zigbee2MQTT:debug 2021-03-12 20:52:36: Received Zigbee message from '0x60a423fffe50109c', type 'commandOff', cluster 'genOnOff', data '{}' from endpoint 22 with groupID 0
Zigbee2MQTT:debug 2021-03-12 20:52:38: Received Zigbee message from '0x60a423fffe50109c', type 'commandOn', cluster 'genOnOff', data '{}' from endpoint 23 with groupID 0
Zigbee2MQTT:debug 2021-03-12 20:52:41: Received Zigbee message from '0x60a423fffe50109c', type 'commandOff', cluster 'genOnOff', data '{}' from endpoint 24 with groupID 0
```


Using `fz.command_on` and `fz.command_off` the resulting action will be `on_s1`, `off_s2`, `on_s3`, `off_s4` (if I understand the converters right). I thought that could be confusing, since it really is 4 separate push buttons. If you think this is fine, just let me know and I will correct the PR to use `fz.command_on` and `fz.command_off`.

Thanks! :)

Koenkk: I see, but with the current code you send a `single` for both `on` and `off`, doesn't it make sense to be able to use both events?

niklascp: Yes, ideally, the "problem" is e.g. `button 1` (i.e. top/left) always sends `on` and never `off`, while e.g. `button 2` (top/right) always sends `off` but never `on`. 

Anyhow, I just corrected the PR to re-use `fz.command_on` and `fz.command_off`, so there is no need for additions in `fromZigbee.js`

Koenkk:
Thanks!

PR #2356, opened 2021-03-11
Add angle and side properties the sensor cube

tim-hellhake:
This PR adds the `angle` and `side` properties for the Aqara cube.
Please excuse me if a added the properties to the wrong places.
This is my first PR in this repo.


Koenkk:
Thanks! and congratz with your first PR :)

PR #2359, opened 2021-03-12
Removing tamper exposure of SNZB-04

ai-xyz:
The SNZB-04 contact sensor just has a reed, LED and a push button (see. e.g [here](https://community.hubitat.com/t/sonoff-snzb-04-zigbee-door-window-contact-sensor-internal/47475)). There is no component inside that could trigger a tamper warning therefore i would suggest removing the exposure of tamper because it seems to be not existing.


Koenkk:
Thanks!

PR #2363, opened 2021-03-13
sg-v100-config-key

nukusinji:
configuration key 1->2


nukusinji:
Don’t need change key.

PR #2366, opened 2021-03-13
Added support for Heiman DoorSensor-EF-3.0

Multipage:
Added support for another version of Heiman HS1DS-E


Multipage:
Opend other pull request with edited name

PR #2367, opened 2021-03-13
Added support for Heiman HS1DS-EF

Multipage:
Added support for another version of a Heiman HS1DS-E  version . Version HS1DS-EF
Picture in Zigbee2MQTTneeds to be the same as the Heiman HS1DS-E


Koenkk:
Thanks!

PR #2369, opened 2021-03-13
Tuya wall switch module with power_on_behavior option

kirovilya:
Like Sonoff mini module https://aliexpress.ru/item/1005002294553548.html


Koenkk:
Thanks!

PR #2370, opened 2021-03-13
Add/fix support for occupancy with timeout for Centralite and Xfinity keypads

mariocao:
This PR adds/fixes the support for occupancy with timeout for the keypads Centralite 3400-D and Xfinity URC4450BC0-X-R.

Both keypads have an occupancy sensor that sends the following message when triggered:

```
Received Zigbee message from '<IEEE address>', type 'commandGetPanelStatus', cluster 'ssIasAce', data '{}' from endpoint 1 with groupID 0
```

The new converter just reuses the functoinality of `occupancy_with_timeout`.

Thanks @aesedepece for helping me with the PR! ;-)

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -849,14 +849,6 @@ const converters = {
             };
         },
     },
-    ias_ace_occupancy_with_timeout: {
-        cluster: 'ssIasAce',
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #2373, opened 2021-03-14
expose: allow setting the label for Lock()

sjorge:
For devices with multiple locks or for devices that have a lock function, allow setting of a label.
e.g. for TRV's with a keypad_lockout it is now called keypad_lockout instead of state which is confusing if there are multiple switches/locks all called 'state'


lib/exposes.js
```diff
@@ -82,15 +82,15 @@ class Lock extends Base {
         this.features = [];
     }
 
-    withState(property, valueOn, valueOff, description, access=a.ALL) {
+    withState(property, valueOn, valueOff, description, access=a.ALL, label='state') {
         assert(!this.endpoint, 'Cannot add feature after adding endpoint');
-        this.features.push(new Binary('state', access, valueOn, valueOff).withProperty(property).withDescription(description));
+        this.features.push(new Binary(label, access, valueOn, valueOff).withProperty(property).withDescription(description));
```
Koenkk, line 87: I don't think the extra argument is necessary. Looks more clean to just pass `property` to the first argument of `new Binary()`.

sjorge: That was what I first had, I'll change it back.

Koenkk:
I remember now why the name is always `state`. This allows to determine what feature indicates the state of the lock which breaks when we set the name (https://www.zigbee2mqtt.io/information/exposes#lock)

sjorge:
image

Before it was state instead of keypad lockout, having one is fine but the danfoss uses a few binary sensors so you end up with a few list labeled state, which is confusing.

Koenkk:
@nurikk what about using the `property` as the label in the frontend instead of `name`?

nurikk:
> @nurikk what about using the `property` as the label in the frontend instead of `name`?

It might be better to add a `label` property, instead of relying on `property` key

nurikk:
Currently we don't have a clear definition of what is
`feature.name`
`feature.type`

sjorge:
I think just not using e.Lock() here is the correct way to resolve this... it really isn't a 'Lock' Device but a extra feature on a Climate device. And that should work fine without any changes to the frontend.

Koenkk:
@sjorge what if it is exposed as a switch?

sjorge:
> @sjorge what if it is exposed as a switch?

Same issue https://github.com/Koenkk/zigbee-herdsman-converters/blob/fea5c35a1ba615df62f7c6400fa30a7449a796d2/lib/exposes.js#L68

Koenkk:
@sjorge what if we use:

```
keypad_lockout: () => new Binary('keypad_lockout', access.STATE_ALL, 'lock1', 'unlock').withDescription('Enables/disables physical input on the device'),
```

sjorge:
> ```
> keypad_lockout: () => new Binary('keypad_lockout', access.STATE_ALL, 'lock1', 'unlock').withDescription('Enables/disables physical input on the device'),
> ```

That was my suggestion when i wrote

> I think just not using e.Lock() here is the correct way to resolve this... it really isn't a 'Lock' Device but a extra feature on a Climate device. And that should work fine without any changes to the frontend.

That would indeed fix it.

sjorge:
🤔 
image

sjorge:
nvm :) I copied your line but it has an error in it, state.ALL is the right one, resting...

sjorge:
Looks good now and /get and /set still work

image

Koenkk:
Thanks!

PR #2379, opened 2021-03-15
Added support for Frient fire detector Develco HESZB-120

mamrai1:
Add support for Develco HESZB-120 Fire detector with siren. Exposes exaclty the same as the Develco Smoke detector


Koenkk:
Thanks!

Issue #2377, opened 2021-03-15
SAGE206612 only picks up primary bell


mrstegeman:
The device supports input from two separate doorbells, but the `action` value only changes when the primary doorbell is pushed.

mrstegeman:
@mattia480 I see that you added the initial device support in #1284. Any idea what's going on here?

mattia480:
> @mattia480 I see that you added the initial device support in #1284. Any idea what's going on here?

Can you set the Zigbee2mqtt Log Level to "debug", try to push the second door bell and post here the output of the log?

mrstegeman:
Here's the primary doorbell (working):
```
Zigbee2MQTT:debug 2021-03-16 10:02:28: Received Zigbee message from 'Doorbell', type 'commandOn', cluster 'genOnOff', data '{}' from endpoint 18 with groupID null
Zigbee2MQTT:info  2021-03-16 10:02:28: MQTT publish: topic 'zigbee2mqtt/Doorbell', payload '{"action":"on","linkquality":153}'
```

Here's the secondary doorbell (non-working):
```
Zigbee2MQTT:debug 2021-03-16 10:08:01: Received Zigbee message from 'Doorbell', type 'commandOff', cluster 'genOnOff', data '{}' from endpoint 18 with groupID null
Zigbee2MQTT:debug 2021-03-16 10:08:01: No converter available for 'SAGE206612' with cluster 'genOnOff' and type 'commandOff' and data '{}'
```

mattia480:
Fixed by #2383 
Just wait the merge. Thanks @mrstegeman for the issue 👍

mrstegeman:
Thank you for the quick fix!

Koenkk:
Closing this as https://github.com/Koenkk/zigbee-herdsman-converters/pull/2383 is merged

Changes will be available in the dev branch in a few hours from now. (https://www.zigbee2mqtt.io/how_tos/how-to-switch-to-dev-branch.html)

PR #2383, opened 2021-03-16
Fix EchoStar SAGE206612 Doorbell channels state

mattia480:
Fixed a bug that did not allow to view the correct channel (the one that correspond to the doorbell rang).
Now on HomeAssistant (or via mqtt) the states are sets to "bell1" and "bell2" (and not more to "on")


mattia480:
Reported by #2377

Koenkk:
Thanks!

PR #2384, opened 2021-03-16
Update device.js

haade-administrator:
add moes zts-eu 2 gang

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14834,6 +14834,21 @@ const devices = [
         },
     },
     {
+        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_g1ib5ldv'}],
+        model: 'ZTS-EU_2gang',
+        vendor: 'Moes',
+        description: 'Wall touch light switch (2 gang)',
+        exposes: [e.switch().withEndpoint('l1').setAccess('state', ea.STATE_SET),
+            e.switch().withEndpoint('l2').setAccess('state', ea.STATE_SET)],
+        fromZigbee: [fz.tuya_switch_2],
+        toZigbee: [tz.tuya_switch_state],
+        meta: {multiEndpoint: true},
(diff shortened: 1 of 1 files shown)


sjorge:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/58d0756414530bbcca38644bded7a2e74183b46c/devices.js#L1320

Isn't this one already supported?

Multipage:
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/58d0756414530bbcca38644bded7a2e74183b46c/devices.js#L1320
> 
> Isn't this one already supported?

Yes it is....and this is my point exactly on all the Heiman devices too.
(https://github.com/Koenkk/zigbee2mqtt/discussions/6731)
If you look at that part of the switches a lot is the same code..the only difference is the manufacturername or subversion.
I think the devices.js needs a cleanup before it's to late.
New devices must be checked if it already exists and if so AND it has seperate code it must have explanation why in the PR.

sjorge:
> New devices must be checked if it already exists and if so AND it has seperate code it must have explanation why in the PR.

That is how i noticed, CI fails on this because it detects the device as a duplicate.

Multipage:
> > New devices must be checked if it already exists and if so AND it has seperate code it must have explanation why in the PR.
> 
> That is how i noticed, CI fails on this because it detects the device as a duplicate.

Yes because this one is exactly duplicated:
 {"modelID":"TS0601","manufacturerName":"_TZE200_g1ib5ldv"}
 But there are other records for other TS0601's that imho could go with the existig TS0601 code
 If he had made a PR with seperate code like ( e.g.) {"modelID":"TS0601","manufacturerName":"_TZE200_g1ib5ldv2"}
 It wouldn't be detected by the script and would be added to devices.js without a question why it couldn't go under an existing TS0601

PR #2387, opened 2021-03-17
Fix STS-IRM-250 (motionv4) battery reporting

crowbarz:
Update STS-IRM-250 (motionv4) battery reporting to use battery reporting for F-MLT-US-2 (multiv4) as introduced in #2291
Force powerSource to battery for STS-IRM-250 (motionv4) as per #2343

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -8176,14 +8176,17 @@ const devices = [
         description: 'Motion sensor (2016 model)',
         fromZigbee: [fz.temperature, fz.ias_occupancy_alarm_2, fz.ias_occupancy_alarm_1, fz.battery],
         toZigbee: [],
-        meta: {configureKey: 1, battery: {voltageToPercentage: '3V_2500'}},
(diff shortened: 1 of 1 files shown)


sjorge:
Have you tried seeing if it support batteryPercentageRemaining, should be easy to check with the frontend now.
If it does, might be worth switching to that.

crowbarz:
> Have you tried seeing if it support batteryPercentageRemaining, should be easy to check with the frontend now.
> If it does, might be worth switching to that.

No, `UNSUPPORTED_ATTRIBUTE` when I tried to set up reporting on that attribute unfortunately.

sjorge:
> > Have you tried seeing if it support batteryPercentageRemaining, should be easy to check with the frontend now.
> > If it does, might be worth switching to that.
> 
> No, `UNSUPPORTED_ATTRIBUTE` when I tried to set up reporting on that attribute unfortunately.

That's a shame, but thanks for checking!

Koenkk:
Thanks!

PR #2391, opened 2021-03-18
Add support for Vizo VZ-222S 2-gang light switch without neutral line

chiahsing:
Tested locally with the device. While the device is a 2-gang switch, its model is TS0014, not TS0012


Koenkk:
Thanks!

PR #2392, opened 2021-03-18
Add support for EchoStar single gang light switch

Tediore:
Also correct battery reporting for Xfinity XHS2-SE contact sensor
Also add battery low sensor to Linkind one-key remote


Tediore:
Forgot to mention battery reporting doesn't seem to work properly on the Sage switch (it seems to always send 0) so I'll just keep an eye on it and see if anything changes.

PR #2398, opened 2021-03-20
Add support for Orvibo SP20-0 CO Sensor

popescunsergiu:
The Orvibo SP20-0 Carbon Monoxide Sensor is in fact a Heiman HS1CA-M sensor. This change adds support for this sensor with all its exposed functionalities.


Koenkk:
Thanks!

PR #2399, opened 2021-03-20
Centralite 3400-D add tamper button

DennisVonDerBey:
Pressing the tamper button triggers the following event:
```
type 'commandStatusChangeNotification', cluster 'ssIasZone', data '{"extendedstatus":0,"zonestatus":0}' from endpoint 1 with groupID 0
```
Releasing the tamper button triggers the following event:
```
type 'commandStatusChangeNotification', cluster 'ssIasZone', data '{"extendedstatus":0,"zonestatus":4}' from endpoint 1 with groupID 0
```

I am still new to this code base. Should I also explicitly check for the extendedstatus field? Should I coin the converter method name to the centralite 3400-D?

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1507,6 +1507,16 @@ const converters = {
             return converters.occupancy_with_timeout.convert(model, msg, publish, options, meta);
         },
     },
+    ias_zone_tamper: {
+        cluster: 'ssIasZone',
+        type: 'commandStatusChangeNotification',
+        convert: (model, msg, publish, options, meta) => {
+            const zoneStatus = msg.data.zonstatus;
+            return {
+                tamper: zoneStatus == 4 ? true : false   
+            };
+        },
+    },
     tuya_led_controller: {
         cluster: 'lightingColorCtrl',
         type: ['attributeReport', 'readResponse'],
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1513,7 +1513,7 @@ const converters = {
         convert: (model, msg, publish, options, meta) => {
             const zoneStatus = msg.data.zonstatus;
             return {
-                tamper: zoneStatus == 4 ? true : false
+                tamper: zoneStatus == 4 ? true : false,
             };
         },
     },
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1513,7 +1513,7 @@ const converters = {
         convert: (model, msg, publish, options, meta) => {
             const zoneStatus = msg.data.zonstatus;
             return {
-                tamper: zoneStatus == 4 ? true : false   
+                tamper: zoneStatus == 4 ? true : false
             };
         },
     },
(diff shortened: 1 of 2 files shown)


devices.js
```diff
@@ -9208,8 +9208,8 @@ const devices = [
         vendor: 'Centralite',
         description: '3-Series security keypad',
         meta: {configureKey: 1, battery: {voltageToPercentage: '3V_2100'}},
-        fromZigbee: [fz.command_arm_with_transaction, fz.temperature, fz.battery, fz.ias_ace_occupancy_with_timeout],
-        exposes: [e.battery(), e.temperature(), e.occupancy(), e.action([
+        fromZigbee: [fz.command_arm_with_transaction, fz.temperature, fz.battery, fz.ias_ace_occupancy_with_timeout, fz.ias_zone_tamper],
```
Koenkk, line 9211: Instead of `ias_zone_tamper` the existing `ias_no_alarm` should also work.

DennisVonDerBey: I tried it out but it gives me the following error:

>  No converter available for '3400-D' with cluster 'ssIasZone' and type 'commandStatusChangeNotification' and data '{"extendedstatus":0,"zonestatus":0}'

I guess it's because `ias_no_alarm` is of type `attributeReport`, but the data comes in as `commandStatusChangeNotification`?

Koenkk: For the `ias_no_alarm` change `type: 'attributeReport',` to `type: ['attributeReport', 'commandStatusChangeNotification'],`, then it should work

DennisVonDerBey: Ah makes sense.

I tried it out, but the state of the tamper field doesn't change using that method.

Koenkk: Can you try changing `const zoneStatus = msg.data.zoneStatus;` to `const zoneStatus = msg.type === 'attributeReport' ? msg.data.zoneStatus : msg.data.zonestatus`

aesedepece:
@DennisVonDerBey Model `URC4450BC0-X-R`from Xfinity is an almost identical supported device that is also missing the tamper converter. I'd take the chance to also add it in there.

devices.js
DennisVonDerBey: @Koenkk sorry it took so long. That also doesn't work.

converters/fromZigbee.js
```diff
@@ -1507,6 +1507,16 @@ const converters = {
             return converters.occupancy_with_timeout.convert(model, msg, publish, options, meta);
         },
     },
+    ias_zone_tamper: {
+        cluster: 'ssIasZone',
+        type: 'commandStatusChangeNotification',
+        convert: (model, msg, publish, options, meta) => {
+            const zoneStatus = msg.data.zonstatus;
+            return {
+                tamper: zoneStatus == 4 ? true : false,
```
Koenkk, line 1516: Can you try with: `tamper: (zoneStatus & 1<<2) > 0` instead here?

DennisVonDerBey: Works finally @Koenkk

DennisVonDerBey: Can we merge it now?

PR #2404, opened 2021-03-21
Update and Bugfix for Danfoss Thermostate

glad-wolf:
Setting the time (copied from HEIMAN devices)
Fixed bug in set trigger_time
More attributes exposed


Koenkk:
Thanks!

PR #2405, opened 2021-03-21
Zemismart ZM-CSW032-D Curtain/roller blind switch get and set position

ullrik:
I have update my previous code. This time if no more need to modify the cluster

It is necessary to add two elements in the "Configuration.yaml", time_close and time_open.
Here is my configuration in configuration,yaml ;
```
devices:
  '0xec1bbdfffeXXXXXX': 
    friendly_name: MyCover 
    time_close: 27 
    time_open: 28
...
```

I also planned that it could work without (as currently on the branch master).


converters/fromZigbee.js
```diff
@@ -4588,23 +4588,21 @@ const converters = {
...
-                    result.position = options.invert_cover ? 100 - result.position : result.position;
                 } else {
                     if (deltaTimeSec < timeCoverSetMiddle || deltaTimeSec > timeCoverSetMiddle) {
                         // postion cast float to int
                         result.position = currentPosition | 0;
-                        result.position = options.invert_cover ? 100 - result.position : result.position;
                     } else {
                         entry.CurrentPosition = lastPreviousAction;
                         result.position = lastPreviousAction;
-                        result.position = options.invert_cover ? 100 - result.position : result.position;
```
Koenkk, line 4600: Why was this removed? This prevents users from inverting their covers.

ullrik: Indeed, I cleaned up the code a little too much to match my installation. In my case with Home Assistant, I had the right information in my "cover-card".
I will put the lines back, thank you for your vigilance.

ullrik:
I don't understand why I have this 2 errors, because I don't changed this : 
```
/home/runner/work/zigbee-herdsman-converters/zigbee-herdsman-converters/devices.js
Error:   4077:79  error  A space is required after ','  comma-spacing
Error:   4077:90  error  A space is required after ','  comma-spacing
```

ullrik:
@Koenkk I apologize but I can't find where the problem is. 
Do you have any ideas ?

converters/fromZigbee.js
Koenkk: Can you re-add this?

Koenkk:
@ullrik don't worry about the errors, will fix them later.

ullrik:
Ok, can you merge now ? My update is finished.

Koenkk:
@ullrik I think the lines from https://github.com/Koenkk/zigbee-herdsman-converters/pull/2405/files#r602776447 still have to be re-added.

converters/fromZigbee.js
ullrik: Yes I do It, with one line after the condition. https://github.com/Koenkk/zigbee-herdsman-converters/blob/92fd707c41465277b540a7969d7f7a77a380c7af/converters/fromZigbee.js view the line 4600.
If you prefer, I can put back this line in each condition ?

ullrik:
@Koenkk I put the line back, but lower after the conditions.
And after a proofreading, I think it would be even better to put it right before the return.

converters/fromZigbee.js
Koenkk: Ah i see

ullrik:
@Koenkk  thanks for merge :)

Do you think that it is necessary to add the information "time_close" and "time_open" in page device  [https://www.zigbee2mqtt.io/devices/ZM-CSW032-D.html](url) ?
Also I see that after reboot I can't get old values back, is there a method to get old value after reboot (in state.json for example)?

PR #2406, opened 2021-03-22
Added support for Woox R7060 Smart Garden Irrigation Control

drzony:


Koenkk:
Thanks!

pichljan:
Hi, thank you very much for adding this device. I believe this device can report low battery status. Could it be modified to propagate the status from the device over zigbee? Thank you.

PR #2407, opened 2021-03-22
More complete support for Livolo cover

aesedepece:
Thanks to https://github.com/Koenkk/zigbee-herdsman/pull/33, this adds support for:
- STOP command
- adjusting motor speed
- adjusting motor direction
- exposing motor movement info

This is what the options look like now:
image

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -1457,46 +1457,62 @@ const converters = {
             await entity.command('genOnOff', 'toggle', {}, {transactionSequenceNumber: 0});
         },
     },
-    livolo_cover_position: {
-        key: ['state', 'position'],
+    livolo_cover_state: {
+        key: ['state'],
         convertSet: async (entity, key, value, meta) => {
-            if (key === 'position') {
-                const position = meta.state.motor_direction === 'FORWARD' ? value : 100 - value;
-                await entity.command('genOnOff', 'toggle', {}, {transactionSequenceNumber: 0});
-                const payload = {0x0401: {value: Buffer.from([position, 0, 0, 0, 0, 0, 0, 0]), type: 1}};
-                await entity.write('genPowerCfg', payload,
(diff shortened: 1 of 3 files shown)


aesedepece:
The linter complains about this PR but the offending line actually comes from the master branch. See #2408 :stuck_out_tongue_winking_eye:

aesedepece:
Added one more commit that improves handling of motor direction and initialization of some settings. I found these changes were necessary when playing with a second brand new unit of this device.

Koenkk:
Thanks!

converters/toZigbee.js
```diff
@@ -1457,37 +1457,90 @@ const converters = {
...
+            case 'CLOSE':
+                payload =
+                    {attrId: 0x0000, selector: null, elementData: [0x10, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]};
+                break;
+            case 'STOP':
+                payload =
+                    {attrId: 0x0000, selector: null, elementData: [0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]};
+                break;
+            default:
+                throw new Error(`Value '${value}' is not a valid cover position (must be one of 'OPEN' or 'CLOSE')`);
```
sjorge, line 1481: Should this also include 'STOP' as that seems to be a valid value as there is a case and payload for it above. (Although yes, STOP would not be a valid 'position' it is still a valid value)

aesedepece: That's true

lotohov:
Hello , tried to test new integration , but my covers detected as switch Zigbee2MQTT:info  2021-05-09 23:25:18: Device '0x00124b0021753df6' is supported, identified as: Livolo Zigbee switch (1 and 2 gang) (TI0001), how i can fix it ? I have 1.18.3 version

aesedepece:
Simply click "reconfigure" from the Zigbee2mqtt UI. Sometimes it takes a
couple of attempts, but it should eventually detect the device as a cover
instead of a switch.

Issue #2071, opened 2021-01-12
Fantem MultiSensor ZB003-X


kirovilya:
new strange device for Tuya.

https://aliexpress.ru/item/1005001878974427.html
![image](https://user-images.githubusercontent.com/8360230/104369677-e1cb8480-552e-11eb-8c9a-5cae62a5dc7d.png)
![image](https://user-images.githubusercontent.com/8360230/104369686-e5f7a200-552e-11eb-92b7-a2c0ae58f148.png)
![image](https://user-images.githubusercontent.com/8360230/104369775-09bae800-552f-11eb-911f-b9a473475ba2.png)

sniffer log
[multisensor.zip](https://github.com/Koenkk/zigbee-herdsman-converters/files/5804647/multisensor.zip)

issue for work with the device

dsantos21:
Hi! I've bought this device too, thanks for your PR! Just to let you know, there's a small mistake in this code. Tamper is not been shown, but it's reported as occupancy. And occupancy is not been reported. 

Effect: zigbee2mqtt cannot report this sensor motion, but if you shake it, is does report as occupancy :) believe me, I tried it with 2 different devices (I bought 4 of them!). 

As I could see in debug, zone status first bit is occupancy, and bit 3 is tamper, so:

Motion detect: zonestatus=1

Tamper and motion detect (impossible to cause just one, I think): zonestatus=5

No motion, nor tamper: zonestatus=0

I've no experience with git, but let me know with o can help somehow! 

Ive attached the debug pictures, but obviously they lake context :/
![Screenshot_20210201-214430_Home Assistant](https://user-images.githubusercontent.com/44759158/106536917-1603e680-64d8-11eb-9f03-03fa59c09a25.png)

dsantos21:
I think the bitwise shift was the problem (found it in from Zigbee2mqtt):

Original:
occupancy: (zoneStatus & 1<<2) > 0,

Solution (I think!):
occupancy: (zoneStatus & 1) > 0,

Course, I didn't try it. I have no idea how to test it... :( I just simulated the results using a site that allows you to run python expressions online.

dsantos21:
Just to let you know: I've just edited my "fromZigbee" file with that proposed solution and it's working now :DDD
All I have to do is not reboot my instance :D. On the weekend I'll look how to propose a PR.

drzony:
@dsantos21 @kirovilya 
[...]

PR #110, opened 2018-10-25
Sylvania Smart+A19 dimmable soft white bulb

wfenwick1:
{
        zigbeeModel: ['A19 W 10 year'],
        model: '74696',
        vendor: 'Sylvania',
        description: 'Smart+ A19 dimmable soft white bulb',
        supports: generic.light_onoff_brightness().supports,
        fromZigbee: generic.light_onoff_brightness().fromZigbee,
        toZigbee: generic.light_onoff_brightness().toZigbee,
    },

Issue #107, opened 2018-10-22
Rework of the converters


tb-killa:
Hi @Koenkk
Due to the large number of issues which are opened again and again with the same problems:

e.g. Do I have a new device, how can I integrate it ?
e.g. How can I edit the file xyz ?
e.g. Multiple Converter for other devices but somehow different.
etc.
etc.

I have considered to revise the current state of converter.

Therefore I thought about the following structure:

per device a separate file e.g. vendor_model_ownedefinition.js

content. the section for fromZ

the section for fromZigbee and toZigbee as well as the actual ``devices`` part
e.g.
```
{
        zigbeeModel: ['lumi.sensor_switch'],
        model: 'WXKG01LM',
        vendor: 'Xiaomi',
        description: 'MiJia wireless switch',
        supports: 'single, double, triple, quadruple, many, long, long_release click',
        fromZigbee: [fz.xiaomi_battery_3v, fz.WXKG01LM_click, fz.ignore_onoff_change, fz.ignore_basic_change],
        toZigbee: [],
    },
```

so we have an own exact assignment for the converters for each model.
The advantage: Dynamic editing per model as well as simple adding of own models (e.g. with HASSIO).

I can take over the revision, but I need your help regarding the exports and the associated processing.

Maybe you would be interested to do this with me soon ?

Koenkk:
Yes, but I doubt of this would solve the problems. Most people have trouble finding the `devices.js` file, understanding it isn't a problem most of the times. We get a lot of PR's supporting new devices which are most of the time good.

PR #2409, opened 2021-03-23
Tuya active reports

drzony:
Companion for https://github.com/Koenkk/zigbee-herdsman/pull/337

I've updated:
- Fantem ZB003X to use the new type in cluster, waiting for confirmation in: #2071 
- Tuya switches which used a combo of defined and 'raw' message types


Koenkk:
Thanks!

dsantos21:
Sorry @drzony, i didn't see your message before!

Heres an example of fantem debug:

zigbee-herdsman-converters:FantemZB003X: Unrecognized DP #110 with data {"status":0,"transid":0,"dp":110,"datatype":2,"fn":0,"data":{"type":"Buffer","data":[0,0,0,100]}}

zigbee-herdsman-converters:FantemZB003X: Unrecognized DP #107 with data {"status":0,"transid":0,"dp":107,"datatype":2,"fn":0,"data":{"type":"Buffer","data":[0,0,0,245]}}

Now I've a lot of these messages!

dsantos21:
Or would it be something like that?

Received Zigbee message from '0xxxxxxxxx', type 'commandActiveStatusReport', cluster 'manuSpecificTuya', data '{"data":{"data":[0,0,0,100],"type":"Buffer"},"datatype":2,"dp":110,"fn":0,"status":0,"transid":0}' from endpoint 1 with groupID 0

drzony:
@dsantos21 Are you on latest dev, do you have any custom changes? The unrecognized messages seem like a bug, since those DPs are defined in code.

drzony:
@dsantos21 Fauond a bug, after the merge everything should be OK.

dsantos21:
I'm using the latest release, 1.18.2. Please, let me know if i can help!

drzony:
@dsantos21 after #2442 is merged you can try it out by using dev branch: https://www.zigbee2mqtt.io/how_tos/how-to-switch-to-dev-branch.html
After this the `Unrecognized DP` messages should be gone and the device should fully work

dsantos21:
Ok, no problem! Motion, tamper and lux still works . As far as I can tell it seems to affect only temperature and humidity. 

What is this change about? :) a new way to read tuya data clusters? Just out of curiosity.

drzony:
The change we are commenting adds a new cluster type (probably Tuya changed something and now a different ID is used to send data), before this ZB003X used "raw" cluster type (which was in fact has the same format as old "getCommand" but due to changed ID it was not recognized). #2442 fixes a mistyping on my part when implementing above change.

dsantos21:
Thanks! Also, just to let you know: the change proposed in #2442 seems to be working fine!
So far I think that led enable/disable settings is inverted (on is off, off is on), but let me try a little while, could still be something here.

PS: Not sure if that's relevant, but right know there are two variants of Fantem ZB003X. I have both types, minor design change (plastic case). The new version doesn't bundle the USB cable anymore... :/

PR #2414, opened 2021-03-24
Add HG06335 battery voltage and percentage reporting

ai-xyz:
Hi, i've tested batteryPercentageRemaining and batteryVoltage reporting with [HG06335](https://www.zigbee2mqtt.io/devices/HG06335.html). Seems like both is working (i tested with reporting once per minute separately on both). Because other Motion Sensors are also configured with reporting one of those attributes (like the IKEA one) i've added this.


Koenkk:
Thanks!

PR #2415, opened 2021-03-25
Improve read and add write

sjorge:
I mostly have stuff similar to this added when working on a new device.
Might be useful to have them around in general, should also make it easier to ask people to try stuff when having problems to be able to write a raw payload to a cluster.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -26,8 +26,11 @@ const converters = {
     read: {
         key: ['read'],
         convertSet: async (entity, key, value, meta) => {
-            const result = await entity.read(value.cluster, value.attributes, (value.options === undefined ? {} : value.options));
+            const result = await entity.read(value.cluster, value.attributes, (value.hasOwnProperty('options') ? value.options : {}));
             meta.logger.info(`Read result of '${value.cluster}': ${JSON.stringify(result)}`);
+            if (value.hasOwnProperty('state_property')) {
+                return {state: {[value.state_property]: result}};
+            }
         },
     },
     factory_reset: {
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -26,7 +26,7 @@ const converters = {
     read: {
         key: ['read'],
         convertSet: async (entity, key, value, meta) => {
-            const result = await entity.read(value.cluster, value.attributes);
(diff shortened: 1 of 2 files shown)


converters/toZigbee.js
```diff
@@ -26,8 +26,20 @@ const converters = {
...
+                const state = {};
+                state[value.store] = result;
+                return {state: state};
+            }
+        },
+    },
+    write: {
+        key: ['write'],
+        convertSet: async (entity, key, value, meta) => {
+            await entity.write(value.cluster, value.payload, utils.getOptions(meta.mapped, entity));
```
Koenkk, line 41: Please also allow to pass options here similar to the `read`.

sjorge: Do we merge them with the getOptions or do we pass them in favor of getOptions?

converters/toZigbee.js
```diff
@@ -26,8 +26,20 @@ const converters = {
     read: {
         key: ['read'],
         convertSet: async (entity, key, value, meta) => {
-            const result = await entity.read(value.cluster, value.attributes);
+            const result = await entity.read(value.cluster, value.attributes, (value.hasOwnProperty('options') ? value.options : {}));
             meta.logger.info(`Read result of '${value.cluster}': ${JSON.stringify(result)}`);
+            if (value.hasOwnProperty('store')) {
+                const state = {};
+                state[value.store] = result;
+                return {state: state};
```
Koenkk, line 34: Can be simplified to

```js
return {state: {[value.store]: result}};
```

If I understand correctly, the goal is to expose the result on a specific property? Not sure if `store` is a good name.

sjorge: Yes, open to suggestions on the name. I’m not happy with store either but it was the best I could come up with to ‘store’ the result in the state.

sjorge: I called it 'state_property' instead, not sure it's a huge improvement. I did simplyfy the code.

converters/toZigbee.js
sjorge: Something like the new commit?

Koenkk: Looks good, thanks!

PR #2417, opened 2021-03-26
Added support for Heiman HS1TS-N v HT-EF-3.0

Multipage:
Added support for Heiman HS1TS-N Zigbee 3.0 version HT-EF-3.0


Koenkk:
Thanks!

PR #2423, opened 2021-03-27
Updated Datek APEX and added Datek HAN adapter

TheStigh:
Updated Datek APEX and added Datek HAN for firmware 0.2, next firmware for HAN will include temperature. New OTA file already mergen by Koen.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -17280,9 +17280,9 @@ const devices = [
 
     {
         zigbeeModel: ['HAN'],
-        model: 'Datek HAN',
+        model: 'Datek_HAN',
         vendor: 'Datek',
-        description: 'Datek Eva AMS HAN power-meter sensor',
+        description: 'Eva AMS HAN power-meter sensor',
         fromZigbee: [fz.metering, fz.electrical_measurement],
         toZigbee: [],
         ota: ota.zigbeeOTA,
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -17260,16 +17260,46 @@ const devices = [
...
+            await reporting.rmsCurrent(endpoint);
+            await reporting.activePower(endpoint);
+            await reporting.temperature(endpoint);
+        },
+        exposes: [e.power(), e.current(), e.voltage(), e.switch(), e.temperature()],
+    },
+
+    {
+        zigbeeModel: ['HAN'],
+        model: 'Datek HAN',
```
Koenkk, line 17283: Is there any model written on the device? (one where you can find the device with via Google?)

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee-herdsman-converters/pull/2423

devices.js
TheStigh: Nop, I did ask the manufacturer but the model name are DATEK HAN. That is why I in the device information have provided URL to the product page.

TheStigh: @Koenkk Can this be merged please (you said within hours) so I can get the OTA available to upgrade the HAN adapter, will be more available + fixing stability issues in this old fw.

Koenkk:
Thanks!

PR #2424, opened 2021-03-27
Support for PIR Motion Sensor Model XZ-CGV3 (Tuya)

esion-net:
[Ali Express Link](https://de.aliexpress.com/item/1005001763329111.html
)
I've not been able to test since I don't yet know the process for developing with hassio/docker (devices.js gets overwritten). 

Also, is there a wildcard-detection for 'fingerprint'? Maybe '_TYZB01_*' (or similar) would me more appropriate.


Koenkk:
Thanks!

A wildcard is not an option here since there are many device having a manufacturerName starting with `_TYZB01`

PR #2432, opened 2021-03-30
Updated Datek AMS Meter Reader, also with correct model

TheStigh:
@Koenkk Through the Zigbee Alliance, I found the actual model `HSE2905E `
Now all should be OK with this new adapter :)

https://zigbeealliance.org/zigbee_products/eva-meter-reader/

Updated doc's also, PR should be ready now.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -17327,17 +17327,17 @@ const devices = [
     },
 
     {
-        zigbeeModel: ['HAN'],
-        model: 'Datek_HAN',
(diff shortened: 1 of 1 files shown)


TheStigh:
Please have in mind, it is a brand new product released to the market from the manufacturers own brand (they are also an OEM manufacurer), EVA 

https://shop.evasmart.no/

I will start on the ID-LOCK integration hopefully this week.

Koenkk:
Thanks!

PR #2433, opened 2021-03-30
Update devices.js: Hampton Bay 99432

aofc:
Fixes Exposes for Hampton Bay 99432 and power source

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -12530,7 +12530,7 @@ const devices = [
         description: 'Universal wink enabled white ceiling fan premier remote control',
         fromZigbee: preset.light_onoff_brightness().fromZigbee.concat([fz.fan]),
         toZigbee: preset.light_onoff_brightness().toZigbee.concat([tz.fan_mode]),
-        exposes: [e.light_brightness(), e.fan()],
+        exposes: [e.light_brightness(), exposes.climate().withFanMode(["off", "low", "medium", "high", "on", "auto", "smart"])],
         meta: {disableDefaultResponse: true, configureKey: 1},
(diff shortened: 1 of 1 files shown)


devices.js
```diff
@@ -12530,14 +12530,16 @@ const devices = [
         description: 'Universal wink enabled white ceiling fan premier remote control',
         fromZigbee: preset.light_onoff_brightness().fromZigbee.concat([fz.fan]),
         toZigbee: preset.light_onoff_brightness().toZigbee.concat([tz.fan_mode]),
-        exposes: [e.light_brightness(), e.fan()],
+        exposes: [e.light_brightness(), exposes.climate().withFanMode(['off', 'low', 'medium', 'high', 'on', 'auto', 'smart'])],
```
Koenkk, line 12533: I don't understand this change, could you explain it?

Koenkk: The fact that it is not displayed correctly in the frontend is not a reason to change it here. Can you provide a screenshot of it here?

aofc: before the change
Capture d’écran 2021-04-01 145219

after the change
Capture d’écran 2021-04-01 145805

I know the frontend is a different project, but it does seem to work better this way, so I figured it was expected to be this way. as I said before, it's a ceiling fan with a light, it would be right to have the fan part be the first function that is exposed, if you think it's better

Koenkk: @nurikk can you look into this? Seems fan is not supported in the frontend

nurikk: Done

PR #2434, opened 2021-03-31
Add lib.utils.mapNumberRange function

blazewicz:
This utility function performs linear interpolation (or range conversion) of numeric values. It works exactly like Arduino's [map()](https://www.arduino.cc/reference/en/language/functions/math/map/) function.

In this PR I've also updated `converters/{to,from}Zigbee.js` in places where I found `mapNumberRange` suitable.

### Code using `utils.mapNumberRange` is much more readable.

```javascript
// wtf?
result.color_temp = Math.round(-1.36 * value + 500);

// ok, it maps value from range (0..255) to (500..153)
result.color_temp = mapNumberRange(value, 0, 255, 500, 153);
````

### No need for explicit `precisionRound` or `Math.round` calls.

```javascript
// magic arithmetic and explicit rounding
result.color.x = precisionRound(msg.data['currentX'] / 65535, 4);

// range conversion with defined result precision
result.color.x = mapNumberRange(msg.data['currentX'], 0, 65535, 0, 1, 4);
```

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -10,7 +10,7 @@
  */
 
 const {
-    precisionRound, isLegacyEnabled, toLocalISOString, numberWithinRange, hasAlreadyProcessedMessage,
+    precisionRound, mapNumberRange, isLegacyEnabled, toLocalISOString, numberWithinRange, hasAlreadyProcessedMessage,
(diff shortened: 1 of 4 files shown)


Koenkk:
Thanks!

PR #2436, opened 2021-03-31
Add model ID for metallic Hue Appear outdoor wall light.

Staacks:
Just one line changed: Added another model ID to the already existent Hue Appear outdoor wall light. Not sure if mine differs because of the color (I have the metallic one) or if this is a new revision, but it works perfectly.


Koenkk:
Thanks!

PR #2438, opened 2021-04-02
add support for Teekar/Tuya in-wall outlet

elmarx:
Hi,

I found this in-wall outlet at [german amazon](https://www.amazon.de/gp/product/B08VRGLKYF).

It's harder to find any model-id for this device than adding support for it… 

The whitelabel's website doesn't even list this device (or they did confuse zigbee with wlan, as the hole packaging and manual says "this is a WLAN device").

Whatever, with this additions to `devices.js` the switch works via zigbee2mqtt.


Koenkk:
Thanks!

PR #2440, opened 2021-04-02
Update regex to support new Ubisys firmware names

### LINKED ISSUE #2439: Ubisys firmware names changed
The regex for firmware filenames does not match the current firmware. The expected filename has to end with `.ota.zigbee` but the current filename ends with `ota1.zigbee`.

zivillian:
fixes #2439


sjorge:
LGTM, updated one S1 switch to test.

sjorge (approved):
LGTM

Koenkk:
Thanks!

Issue #2388, opened 2021-03-17
Strip color/color_temp properties for inactive modes


sjorge:
Ideally we strip the color.<attrib> and color_temp depending on the color_mode, this doesn't seem possible currently?
Returning a new color with with missing attributes just merged this with the old ones.

@Koenkk anyway to do this? Or do we need to add some extra code to `lib/extensions/publish.js` for this?

Also, perhaps we should read colorMode during light configuration to populate it initially?

Koenkk:
I think the first question is if this should be done (note that e.g. the x/y and color_temp are synced with each other).

sjorge:
Not sure, does ha expect this?
If we don’t, we should maybe also sync hs -> color_temp?

~ sjorge

Koenkk:
HA doesn't care about it (https://github.com/home-assistant/core/pull/47993#issuecomment-800862882)

sjorge:
Good, that makes it easier.
Should we still do the hs -> color_temp mapping?

~ sjorge

Koenkk:
> Should we still do the hs -> color_temp mapping?

Yes, I think the logic should be the following: if the current state contains any of the other color modes it should be synced.

e.g. if `meta.state` has `color: {h: XX, s: XX}` and color temp is set, h/s should be updated with the color temp but x/y shouldn't be added.

sjorge:
I've been looking at this a bit thinking about how best to tackle this, some observations.

Small improvements:
- everything needed for conversion seems to be in lib/utils.js, we can create a few wrappers for convenience though
- possible move some of those to lib/light.js as that is a better spot

Issues/questions
- doing the conversion on /set color, color_temp seems relatively easy (perhaps a shared function in lib/light.js that populate this)
- doing the conversion for /set saturation_move, hue_move, color_temp_move are trouble some.
  - we read colorTemperature, currentHue, currentSaturation as needed so we don't have a value to convert from
  - conversion can be done in fromZigbee too... but this has a few issues
    - bulbs that update other values themselves via report --> fight as to what is uses e.g. bulb report hue/sat + color_temp when configure -> innr seems to do a conversion and publish all of them, tint only the active ones BUT will republish the others once maxReportingInterval is reached with old values
    - I would prefer to keep the bulbs reported values as they should be closest to what the bulb is actually set too vs what we calculate from pre-made tables

I guess a complex solution would be to do both:
- pre-convert on set and publish
[...]

PR #2441, opened 2021-04-02
#2388 Implement color sync

sjorge:
- [x] rebase after #2453
- [x] ColorXY.toRGB()
  - [x] implement 
  - [x] unit test
- [x] syncColorState 
  - [x] implement
  - [x] call in toZigbee.light_colortemp
  - [x] call in toZigbee.light_color
  - [x] call in fromZigbee.color_colortemp 
  - [x] cleanup toZigbee. light_color_colortemp
- [ ] scenes
  - [x] set color_mode in `scene_add` 
  - [ ] call syncColorState in `scene_recall` ?
- [X] testing

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -17,6 +17,7 @@ const {
 const tuya = require('../lib/tuya');
 const globalStore = require('../lib/store');
 const constants = require('../lib/constants');
+const libColor = require('../lib/color');
 
 const converters = {
     // #region Generic/recommended converters
@@ -556,7 +557,10 @@ const converters = {
                 result.color_options = {execute_if_off: ((msg.data.options & 1<<0) > 0)};
             }
 
-            return result;
+            // handle color property sync
+            // NOTE: this should the last thing we do so we would have processed all values
+            //       the assign here is important as we do not want to wipe out other attributes
+            return Object.assign(result, libColor.syncColorState(result, meta.state, options));
         },
     },
     metering: {
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -980,38 +980,7 @@ const converters = {
             return {state: libColor.syncColorState(newState, meta.state, meta.options), readAfterWriteTime: zclData.transtime * 100};
         },
         convertGet: async (entity, key, meta) => {
-            /**
-              * Not all bulbs suport the same features, we need to take care we read what is supported.
-              * `supportsHueAndSaturation` indicates support for currentHue and currentSaturation
-              * `enhancedHue` indicates support for enhancedCurrentHue
-              *
-              * e.g. IKEA Tådfri LED1624G9 only supports XY (https://github.com/Koenkk/zigbee-herdsman-converters/issues/1340)
-              *
-              * Additionally when we get a get payload, only request the fields included.
-             */
-            const attributes = ['colorMode'];
-
-            if (!meta.message.color || (typeof meta.message.color === 'object' && meta.message.color.hasOwnProperty('x'))) {
-                attributes.push('currentX');
-            }
-            if (!meta.message.color || (typeof meta.message.color === 'object' && meta.message.color.hasOwnProperty('y'))) {
-                attributes.push('currentY');
-            }
-
-            if (utils.getMetaValue(entity, meta.mapped, 'supportsHueAndSaturation', 'allEqual', true)) {
-                if (!meta.message.color || (typeof meta.message.color === 'object' && meta.message.color.hasOwnProperty('hue'))) {
-                    if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
-                        attributes.push('enhancedCurrentHue');
-                    } else {
-                        attributes.push('currentHue');
-                    }
-                }
-                if (!meta.message.color || (typeof meta.message.color === 'object' && meta.message.color.hasOwnProperty('saturation'))) {
-                    attributes.push('currentSaturation');
-                }
-            }
-
-            await entity.read('lightingColorCtrl', attributes);
+            await entity.read('lightingColorCtrl', light.readColorAttributes(entity, meta));
         },
     },
     light_color_colortemp: {
@@ -1038,27 +1007,7 @@ const converters = {
             }
         },
         convertGet: async (entity, key, meta) => {
-            /**
-              * Not all bulbs suport the same features, we need to take care we read what is supported.
-              * `supportsHueAndSaturation` indicates support for currentHue and currentSaturation
-              * `enhancedHue` indicates support for enhancedCurrentHue
-              *
-              * e.g. IKEA Tådfri LED1624G9 only supports XY (https://github.com/Koenkk/zigbee-herdsman-converters/issues/1340)
-              *
-              * Additionally when we get a get payload, only request the fields included.
-             */
-            const attributes = ['colorMode', 'colorTemperature', 'currentX', 'currentY'];
-
-            if (utils.getMetaValue(entity, meta.mapped, 'supportsHueAndSaturation', 'allEqual', true)) {
-                if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
-                    attributes.push('enhancedCurrentHue');
-                } else {
-                    attributes.push('currentHue');
-                }
-                attributes.push('currentSaturation');
-            }
-
-            await entity.read('lightingColorCtrl', attributes);
+            await entity.read('lightingColorCtrl', light.readColorAttributes(entity, meta, ['colorTemperature']));
         },
     },
     effect: {
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -866,7 +866,7 @@ const converters = {
                 readAfterWriteTime: payload.transtime * 100};
         },
         convertGet: async (entity, key, meta) => {
-            await entity.read('lightingColorCtrl', ['colorTemperature']);
+            await entity.read('lightingColorCtrl', ['colorMode', 'colorTemperature']);
         },
     },
     light_colortemp_startup: {
@@ -989,7 +989,7 @@ const converters = {
               *
               * Additionally when we get a get payload, only request the fields included.
              */
-            const attributes = [];
+            const attributes = ['colorMode'];
 
             if (!meta.message.color || (typeof meta.message.color === 'object' && meta.message.color.hasOwnProperty('x'))) {
                 attributes.push('currentX');
@@ -1038,11 +1038,27 @@ const converters = {
             }
         },
         convertGet: async (entity, key, meta) => {
-            if (key == 'color') {
-                await converters.light_color.convertGet(entity, key, meta);
-            } else if (key == 'color_temp') {
-                await converters.light_colortemp.convertGet(entity, key, meta);
+            /**
+              * Not all bulbs suport the same features, we need to take care we read what is supported.
+              * `supportsHueAndSaturation` indicates support for currentHue and currentSaturation
+              * `enhancedHue` indicates support for enhancedCurrentHue
+              *
+              * e.g. IKEA Tådfri LED1624G9 only supports XY (https://github.com/Koenkk/zigbee-herdsman-converters/issues/1340)
+              *
+              * Additionally when we get a get payload, only request the fields included.
+             */
+            const attributes = ['colorMode', 'colorTemperature', 'currentX', 'currentY'];
+
+            if (utils.getMetaValue(entity, meta.mapped, 'supportsHueAndSaturation', 'allEqual', true)) {
+                if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
+                    attributes.push('enhancedCurrentHue');
+                } else {
+                    attributes.push('currentHue');
+                }
+                attributes.push('currentSaturation');
             }
+
+            await entity.read('lightingColorCtrl', attributes);
         },
     },
     effect: {
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -4175,7 +4175,7 @@ const converters = {
                             );
                         }
                         state['color_mode'] = constants.colorMode[0];
-                        state['color'] = newColor.hsv.toObject(true);
+                        state['color'] = newColor.hsv.toObject(false, false);
                     }
                 }
             }
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -4071,6 +4071,20 @@ const converters = {
                 for (const member of entity.members) {
                     if (member.meta.hasOwnProperty('scenes') && member.meta.scenes.hasOwnProperty(metaKey)) {
                         membersState[member.getDevice().ieeeAddr] = addColorMode(member.meta.scenes[metaKey].state);
+
+                        let recalledState = member.meta.scenes[metaKey].state;
+
+                        // add color_mode if saved state does not contain it
+                        if (!recalledState.hasOwnProperty('color_mode')) {
+                            recalledState = addColorMode(recalledState);
+                        }
+
+                        // XXX: meta.state is the groups state, we are leaking group state TO devices!
+                        //      e.g. 2 devices in the group, 1x state=ON, 2x state=OFF, we do a scene recall with just color data
+                        //           we now leaked either ON or OFF to both member devices
+                        Object.assign(recalledState, libColor.syncColorState(recalledState, meta.state, meta.options));
+
+                        membersState[member.getDevice().ieeeAddr] = recalledState;
                     } else {
                         meta.logger.warn(`Unknown scene was recalled for ${member.getDevice().ieeeAddr}, can't restore state.`);
                         membersState[member.getDevice().ieeeAddr] = {};
@@ -4079,7 +4093,16 @@ const converters = {
                 return {membersState};
             } else {
                 if (entity.meta.scenes.hasOwnProperty(metaKey)) {
-                    return {state: addColorMode(entity.meta.scenes[metaKey].state)};
+                    let recalledState = entity.meta.scenes[metaKey].state;
+
+                    // add color_mode if saved state does not contain it
+                    if (!recalledState.hasOwnProperty('color_mode')) {
+                        recalledState = addColorMode(recalledState);
+                    }
+
+                    Object.assign(recalledState, libColor.syncColorState(recalledState, meta.state, meta.options));
+
+                    return {state: recalledState};
                 } else {
                     meta.logger.warn(`Unknown scene was recalled for ${entity.deviceIeeeAddress}, can't restore state.`);
                     return {state: {}};
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -862,7 +862,8 @@ const converters = {
 
             const payload = {colortemp: value, transtime: utils.getTransition(entity, key, meta).time};
             await entity.command('lightingColorCtrl', 'moveToColorTemp', payload, utils.getOptions(meta.mapped, entity));
-            return {state: {color_temp: value, color_mode: constants.colorMode[2]}, readAfterWriteTime: payload.transtime * 100};
+            return {state: libColor.syncColorState({'color_mode': constants.colorMode[2], 'color_temp': value}, meta.state, meta.options),
+                readAfterWriteTime: payload.transtime * 100};
         },
         convertGet: async (entity, key, meta) => {
             await entity.read('lightingColorCtrl', ['colorTemperature']);
@@ -976,7 +977,7 @@ const converters = {
             }
 
             await entity.command('lightingColorCtrl', command, zclData, utils.getOptions(meta.mapped, entity));
-            return {state: newState, readAfterWriteTime: zclData.transtime * 100};
+            return {state: libColor.syncColorState(newState, meta.state, meta.options), readAfterWriteTime: zclData.transtime * 100};
         },
         convertGet: async (entity, key, meta) => {
             /**
@@ -1030,14 +1031,9 @@ const converters = {
         convertSet: async (entity, key, value, meta) => {
             if (key == 'color') {
                 const result = await converters.light_color.convertSet(entity, key, value, meta);
-                if (result.state && result.state.color.hasOwnProperty('x') && result.state.color.hasOwnProperty('y')) {
-                    result.state.color_temp = Math.round(libColor.ColorXY.fromObject(result.state.color).toMireds());
-                }
-
                 return result;
             } else if (key == 'color_temp' || key == 'color_temp_percent') {
                 const result = await converters.light_colortemp.convertSet(entity, key, value, meta);
-                result.state.color = libColor.ColorXY.fromMireds(result.state.color_temp).rounded(4).toObject();
                 return result;
             }
         },
@@ -4142,6 +4138,7 @@ const converters = {
                     const xScaled = utils.mapNumberRange(xy.x, 0, 1, 0, 65535);
                     const yScaled = utils.mapNumberRange(xy.y, 0, 1, 0, 65535);
                     extensionfieldsets.push({'clstId': 768, 'len': 4, 'extField': [xScaled, yScaled]});
+                    state['color_mode'] = constants.colorMode[2];
                     state['color_temp'] = val;
                 } else if (attribute === 'color') {
                     try {
@@ -4161,6 +4158,7 @@ const converters = {
                                 'extField': [xScaled, yScaled],
                             },
                         );
+                        state['color_mode'] = constants.colorMode[1];
                         state['color'] = newColor.xy.toObject();
                     } else if (newColor.isHSV()) {
                         const hsvCorrected = newColor.hsv.colorCorrected(meta);
@@ -4188,6 +4186,7 @@ const converters = {
                                 },
                             );
                         }
+                        state['color_mode'] = constants.colorMode[0];
                         state['color'] = newColor.hsv.toObject(true);
                     }
                 }
--- a/lib/color.js
+++ b/lib/color.js
@@ -253,26 +253,19 @@ class ColorXY {
         }
 
         // Reverse gamma correction
-        // XXX: results in failed tests (bad tests?)
-        // red = red <= 0.0031308 ? 12.92 * red : (1.0 + 0.055) * Math.pow(red, (1.0 / 2.4)) - 0.055;
-        // green = green <= 0.0031308 ? 12.92 * green : (1.0 + 0.055) * Math.pow(green, (1.0 / 2.4)) - 0.055;
-        // blue = blue <= 0.0031308 ? 12.92 * blue : (1.0 + 0.055) * Math.pow(blue, (1.0 / 2.4)) - 0.055;
+        red = red <= 0.0031308 ? 12.92 * red : (1.0 + 0.055) * Math.pow(red, (1.0 / 2.4)) - 0.055;
+        green = green <= 0.0031308 ? 12.92 * green : (1.0 + 0.055) * Math.pow(green, (1.0 / 2.4)) - 0.055;
+        blue = blue <= 0.0031308 ? 12.92 * blue : (1.0 + 0.055) * Math.pow(blue, (1.0 / 2.4)) - 0.055;
 
-
-        // Convert normalized decimal to decimal
-        red = Math.round(red * 255);
-        green = Math.round(green * 255);
-        blue = Math.round(blue * 255);
-
-        if (isNaN(red) || Object.is(red, -0)) {
+        if (isNaN(red)) {
             red = 0;
         }
 
-        if (isNaN(green) || Object.is(green, -0)) {
+        if (isNaN(green)) {
             green = 0;
         }
 
-        if (isNaN(blue) || Object.is(blue, -0)) {
+        if (isNaN(blue)) {
             blue = 0;
         }
 
--- a/lib/color.js
+++ b/lib/color.js
@@ -168,6 +168,17 @@ class ColorRGB {
         }
         return new ColorRGB(transform(this.red), transform(this.green), transform(this.blue));
     }
+
+    /**
+     * Returns color after reverse sRGB gamma correction
+     * @return {ColorRGB} raw RGB
+     */
+    gammaUncorrected() {
+        function transform(v) {
+            return v <= 0.0031308 ? 12.92 * v : (1.0 + 0.055) * Math.pow(v, (1.0 / 2.4)) - 0.055;
+        }
+        return new ColorRGB(transform(this.red), transform(this.green), transform(this.blue));
+    }
 }
 
 /**
@@ -252,22 +263,10 @@ class ColorXY {
             blue = 1.0;
         }
 
-        // Reverse gamma correction
-        red = red <= 0.0031308 ? 12.92 * red : (1.0 + 0.055) * Math.pow(red, (1.0 / 2.4)) - 0.055;
-        green = green <= 0.0031308 ? 12.92 * green : (1.0 + 0.055) * Math.pow(green, (1.0 / 2.4)) - 0.055;
-        blue = blue <= 0.0031308 ? 12.92 * blue : (1.0 + 0.055) * Math.pow(blue, (1.0 / 2.4)) - 0.055;
-
-        if (isNaN(red)) {
-            red = 0;
-        }
-
-        if (isNaN(green)) {
-            green = 0;
-        }
-
-        if (isNaN(blue)) {
-            blue = 0;
-        }
+        // This fixes situation when due to computational errors value get slightly below 0, or NaN in case of zero-division.
+        red = (isNaN(red) || red < 0) ? 0 : red;
+        green = (isNaN(green) || green < 0) ? 0 : green;
+        blue = (isNaN(blue) || blue < 0) ? 0 : blue;
 
         return new ColorRGB(red, green, blue);
     }
--- a/lib/color.js
+++ b/lib/color.js
@@ -753,10 +753,6 @@ function syncColorState(newState, oldState, options) {
         break;
     }
 
-    // always prefer newState.color.*
-    // e.g. when reading both xy and hs, we prefer the real values
-    Object.assign(result.color, newState.color);
-
     // drop empty result.color
     if (Object.keys(result.color).length === 0) delete result.color;
 
--- a/lib/color.js
+++ b/lib/color.js
@@ -665,8 +665,8 @@ function syncColorState(newState, oldState, options) {
     } else if (oldState.hasOwnProperty('color_mode')) {
         result.color_mode = oldState.color_mode;
     } else {
-        // fallback to xy unless color_temp found
-        result.color_mode = newState.hasOwnProperty('color_temp') ? 'color_temp' : 'xy';
+        result.color_mode = newState.hasOwnProperty('color_temp') ? 'color_temp' :
+            (newState.hasOwnProperty('color') && newState.color.hasOwnProperty('hue') ? 'hs' : 'xy');
     }
 
     // figure out target attributes
--- a/lib/color.js
+++ b/lib/color.js
@@ -643,7 +643,7 @@ class Color {
  */
 function syncColorState(newState, oldState, options) {
     const colorTargets = [];
-    const colorSync = (options && options.hasOwnProperty('color_sync')) ? (options.color_sync == true) : true;
+    const colorSync = (options && options.hasOwnProperty('color_sync')) ? options.color_sync : true;
     const result = {};
 
     // check if color sync is enabled
--- a/lib/color.js
+++ b/lib/color.js
@@ -730,7 +730,7 @@ function syncColorState(newState, oldState, options) {
                 result.color_temp = precisionRound(xy.toMireds(), 0);
             }
             if (colorTargets.includes('hs')) {
-                Object.assign(result.color, xy.toHSV().rounded(4).toObject(false, false));
+                Object.assign(result.color, xy.toHSV().rounded(0).toObject(false, false));
             }
         }
         break;
@@ -747,7 +747,7 @@ function syncColorState(newState, oldState, options) {
                 Object.assign(result.color, xy.rounded(4).toObject());
             }
             if (colorTargets.includes('hs')) {
-                Object.assign(result.color, xy.toHSV().rounded(4).toObject(false, false));
+                Object.assign(result.color, xy.toHSV().rounded(0).toObject(false, false));
             }
         }
         break;
--- a/lib/color.js
+++ b/lib/color.js
@@ -210,13 +210,81 @@ class ColorXY {
 
     /**
      * Converts color in XY space to temperature in mireds
-     * @param {ColorXYT} xy color in XY space
      * @return {number} color temp in mireds
      */
     toMireds() {
         const n = (this.x - 0.3320) / (0.1858 - this.y);
         const kelvin = Math.abs(437 * Math.pow(n, 3) + 3601 * Math.pow(n, 2) + 6861 * n + 5517);
         return kelvinToMireds(kelvin);
+    }
+
+    /**
+     * Converts CIE color space to RGB color space
+     * From: https://github.com/usolved/cie-rgb-converter/blob/master/cie_rgb_converter.js
+     * @return {ColorRGB}
+     */
+    toRGB() {
+        // use maximum brightness
+        const brightness = 254;
+
+        const z = 1.0 - this.x - this.y;
+        const Y = (brightness / 254).toFixed(2);
+        const X = (Y / this.y) * this.x;
+        const Z = (Y / this.y) * z;
+
+        // Convert to RGB using Wide RGB D65 conversion
+        let red = X * 1.656492 - Y * 0.354851 - Z * 0.255038;
+        let green = -X * 0.707196 + Y * 1.655397 + Z * 0.036152;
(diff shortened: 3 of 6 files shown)


converters/fromZigbee.js
```diff
@@ -531,7 +532,14 @@ const converters = {
                 result.color_options = {execute_if_off: ((msg.data.options & 1<<0) > 0)};
             }
 
-            return result;
+            // handle color property sync
+            // NOTE: this should the last thing we do so we would have processed all values
+            //       the assign here is important as we do not want to wipe out other attributes
+            if (colorSync) {
```
Koenkk, line 538: To avoid these repetitive if/else, we can always call the color sync function and pass meta.options to it. Then also `const colorSync = (meta.options && meta.options.hasOwnProperty('color_sync')) ? (meta.options.color_sync == true) : true;` doesn't have to be duplicated in every converter.

converters/fromZigbee.js
```diff
@@ -531,7 +532,14 @@ const converters = {
...
+                return Object.assign(result, syncColor(result, meta.state));
```
Koenkk, line 539: I would also move the `Object.assign` into the `syncColor` method. So here the if/else will be replaced by just `return syncColor(result, meta.state);`

sjorge: Not sure we can move this, as we only want to do this in the fromZigbee case. In the toZigbee we just return an empty state when not doing a color_sync as we just want the data from the read/reports. Inside syncColor() we have no idea where we’re called from.

sjorge: I actually have an idea how to do this, let’s see if it actually works.

sjorge: > I actually have an idea how to do this, let’s see if it actually works.
It does not :(

sjorge: I ended up with `return Object.assign(result, syncColor(result, meta.state, options));` in the fromZigbee converter.

It still improved readability. Unless we pass another var indicating if we calling from toZigbee or fromZigbee, but that becomes less readable IMHO.

Some more background why we can't just do the assign inside syncColor(). When color_sync is false for the fromZigbee path we should return an unmodified payload so all reported/read attribute get published untouched. When color_sync is true, we pass color_temp, and color modified based on color_mode and all others unmodified. In theory we could just return newState inside syncColor(), but for the toZigbee path we don't want to return {} in the color_sync == false path so we avoid the need for a debounce to be configured. Because otherwise we get 2 publish... or 1 + 2*group members for a group. One with the set value and nearly instantly one with the value reported by the bulb(s).

sjorge:
Oops de extra commit also dropped the XXX question about groups, is `meta.options` available for a group? I thought it was not.

blazewicz:
@sjorge you may want to check my PR #2453. Your changes could be done on top of mine, I already added few tests for `lib/light.js`

sjorge:
> @sjorge you may want to check my PR #2453. Your changes could be done on top of mine, I already added few tests for `lib/light.js`

Lets try and get your work in first I think, that would make most sense. The sync would need to happen (or not, depending on the device_option) at the end before returning a new state.

sjorge:
> Oops de extra commit also dropped the XXX question about groups, is `meta.options` available for a group? I thought it was not.
> 
> Edit: It looks like it is... that raises an interesting question, setting it on both the group and the device would work.
> But is that what users would expect?

OK so I hit this again in scene_add when saving the scene state to the database.... so is there a getMetaValue equalivanmt but for device options?

sjorge:
Rebased this on top of the new changes, didn't rewerite syncColorState yet so it's broken, hope to get around to that this weekend.

sjorge:
Started some testing and getting very weird results a blue color results in a mireds value of well above 500 and a red one of around 120... which doesn't make much sense the lower the mireds the cooler the light.

Also getting ridiculously large values like `{"color_temp":160.0908567318751}` (which was from a xy of red btw)

Koenkk:
@sjorge for xy -> mireds no lookup is used but a function: https://github.com/Koenkk/zigbee-herdsman-converters/blob/f67c255d1cfa4038358b72bace516c8bb6339820/lib/color.js#L217

blazewicz:
> Started some testing and getting very weird results a blue color results in a mireds value of well above 500 and a red one of around 120... which doesn't make much sense the lower the mireds the cooler the light. Could it be that the lookup table is backwards?

Great that you looking into that, while working on #2453 I tried to improve this, but got only slightly better precision in forth and back convertsion so I gave up. Below you can see my attempt on this:

```diff
diff --git a/lib/color.js b/lib/color.js
index 6e3e70d..d8dd8bd 100644
--- a/lib/color.js
+++ b/lib/color.js
@@ -1,4 +1,3 @@
-const kelvinToXyLookup = require('./kelvinToXy');
 const {precisionRound} = require('./utils');

 /**
@@ -186,8 +185,27 @@ class ColorXY {
      * @return {ColorXY} color in XY space
      */
     static fromMireds(mireds) {
-        const kelvin = miredsToKelvin(mireds);
-        return ColorXY.fromObject(kelvinToXyLookup[Math.round(kelvin)]);
+        // from https://en.wikipedia.org/wiki/Planckian_locus#Approximation
+        const T = miredsToKelvin(mireds);
+
+        let retX;
+        let retY;
+
+        if (T < 1667 || T > 25000) {
+            throw new Error('Color temperature out of range');
+        } else if (T < 4000) {
+            retX = -0.2661239e9 / Math.pow(T, 3) - 0.2343589e6 / Math.pow(T, 2) + 0.8776956e3 / T + 0.179910;
+            if (T < 2222) {
+                retY = -1.1063814 * Math.pow(retX, 3) - 1.34811020 * Math.pow(retX, 2) + 2.18555832 * retX - 0.20219683;
+            } else {
+                retY = -0.9549476 * Math.pow(retX, 3) - 1.37418593 * Math.pow(retX, 2) + 2.09137015 * retX - 0.16748867;
+            }
+        } else {
+            retX = -3.0258469e9 / Math.pow(T, 3) + 2.1070379e6 / Math.pow(T, 2) + 0.2226347e3 / T + 0.240390;
+            retY = 3.0817580 * Math.pow(retX, 3) - 5.87338670 * Math.pow(retX, 2) + 3.75112997 * retX - 0.37001483;
+        }
+
+        return new ColorXY(retX, retY);
     }

     /**
@@ -196,8 +214,9 @@ class ColorXY {
      * @return {number} color temp in mireds
      */
     toMireds() {
+        // from https://www.waveformlighting.com/tech/calculate-color-temperature-cct-from-cie-1931-xy-coordinates
         const n = (this.x - 0.3320) / (0.1858 - this.y);
-        const kelvin = Math.abs(437 * Math.pow(n, 3) + 3601 * Math.pow(n, 2) + 6861 * n + 5517);
+        const kelvin = 437 * Math.pow(n, 3) + 3601 * Math.pow(n, 2) + 6861 * n + 5517;
         return kelvinToMireds(kelvin);
     }

diff --git a/test/color.test.js b/test/color.test.js
index 510554b..658c15f 100644
--- a/test/color.test.js
+++ b/test/color.test.js
@@ -93,7 +93,7 @@ describe('lib/color.js', () => {
             const backConv = asXY.toMireds();
             const error = Math.abs(backConv - mireds);
             // better precision would require better lookup table or continuous conversion function
-            expect(error).toBeLessThan(2.3);
+            expect(error).toBeLessThan(1.8);
         });
     });
```

sjorge:
Initial  tests data, this has additional console.log() added to see what it was doing.

<details><summary>logging.patch</summary>

```patch
diff --git a/lib/color.js b/lib/color.js
index 13578ea..dfdfd3c 100644
--- a/lib/color.js
+++ b/lib/color.js
@@ -686,6 +686,7 @@ function syncColorState(newState, oldState, options) {
     ) {
         colorTargets.push('xy');
     }
+    console.log(`XXX: colorTargets=${JSON.stringify(colorTargets)}`);

     // sync color attributes
     result.color = {};
@@ -705,9 +706,11 @@ function syncColorState(newState, oldState, options) {
         if (result.color.hasOwnProperty('hue') && result.color.hasOwnProperty('saturation')) {
             const hsv = new ColorHSV(result.color.hue, result.color.saturation);
             if (colorTargets.includes('color_temp')) {
+                console.log("XXX: convert hs -> ct");
                 result.color_temp = precisionRound(hsv.toMireds(), 0);
             }
             if (colorTargets.includes('xy')) {
+                console.log("XXX: convert hs -> xy");
                 Object.assign(result.color, hsv.toXY().rounded(4).toObject());
             }
         }
@@ -727,9 +730,11 @@ function syncColorState(newState, oldState, options) {
         if (result.color.hasOwnProperty('x') && result.color.hasOwnProperty('y')) {
             const xy = new ColorXY(result.color.x, result.color.y);
             if (colorTargets.includes('color_temp')) {
+                console.log("XXX: convert xy -> ct");
                 result.color_temp = precisionRound(xy.toMireds(), 0);
             }
             if (colorTargets.includes('hs')) {
+                console.log("XXX: convert xy -> hs");
                 Object.assign(result.color, xy.toHSV().rounded(4).toObject(false, false));
             }
         }
@@ -744,9 +749,11 @@ function syncColorState(newState, oldState, options) {
         if (result.hasOwnProperty('color_temp')) {
             const xy = ColorXY.fromMireds(result.color_temp);
             if (colorTargets.includes('xy')) {
+                console.log("XXX: convert ct -> xy");
                 Object.assign(result.color, xy.rounded(4).toObject());
             }
             if (colorTargets.includes('hs')) {
+                console.log("XXX: convert ct -> hs");
                 Object.assign(result.color, xy.toHSV().rounded(4).toObject(false, false));
             }
         }
@@ -760,6 +767,9 @@ function syncColorState(newState, oldState, options) {
     // drop empty result.color
     if (Object.keys(result.color).length === 0) delete result.color;

+    console.log(`XXX: oldState=${JSON.stringify(oldState)}`);
+    console.log(`XXX: newState=${JSON.stringify(newState)}`);
+    console.log(`XXX: result=${JSON.stringify(result)}`);
     return result;
 }
```

</details>

Read color_temp (no cached state)
```
Zigbee2MQTT:debug 2021-05-08 15:25:35: Received MQTT message on 'zigbee2mqtt/0x00158d00038801e9/get' with data '{"color_temp":""}'
Zigbee2MQTT:debug 2021-05-08 15:25:35: Publishing get 'get' 'color_temp' to '0x00158d00038801e9'
Zigbee2MQTT:debug 2021-05-08 15:25:35: Received Zigbee message from '0x00158d00038801e9', type 'readResponse', cluster 'lightingColorCtrl', data '{"colorTemperature":370}' from endpoint 1 with groupID 0
XXX: colorTargets=["color_temp"]
XXX: oldState={}
XXX: newState={"color_temp":370}
XXX: result={"color_mode":"color_temp","color_temp":370}
Zigbee2MQTT:info  2021-05-08 15:25:35: MQTT publish: topic 'zigbee2mqtt/0x00158d00038801e9', payload '{"color_mode":"color_temp","color_temp":370,"linkquality":129}'
```

Read color (by default, this reads xy and hs (if supported))
```
Zigbee2MQTT:debug 2021-05-08 15:25:44: Received MQTT message on 'zigbee2mqtt/0x00158d00038801e9/get' with data '{"color":""}'
Zigbee2MQTT:debug 2021-05-08 15:25:44: Publishing get 'get' 'color' to '0x00158d00038801e9'
Zigbee2MQTT:debug 2021-05-08 15:25:44: Received Zigbee message from '0x00158d00038801e9', type 'readResponse', cluster 'lightingColorCtrl', data '{"currentSaturation":202,"currentX":29991,"currentY":26872,"enhancedCurrentHue":6388}' from endpoint 1 with groupID 0
XXX: colorTargets=["color_temp","hs","xy"]
XXX: convert ct -> xy
XXX: convert ct -> hs
XXX: oldState={"color_temp":370,"color_mode":"color_temp"}
XXX: newState={"color":{"x":0.4576,"y":0.41,"saturation":80,"hue":35.1}}
XXX: result={"color_mode":"color_temp","color":{"x":0.4576,"y":0.41,"hue":35.1,"saturation":80},"color_temp":370}
Zigbee2MQTT:info  2021-05-08 15:25:44: MQTT publish: topic 'zigbee2mqtt/0x00158d00038801e9', payload '{"color":{"hue":35.1,"saturation":80,"x":0.4576,"y":0.41},"color_mode":"color_temp","color_temp":370,"linkquality":132}'
```

Set xy (blue, via frontend)
```
Zigbee2MQTT:debug 2021-05-08 15:27:47: Received MQTT message on 'zigbee2mqtt/0x00158d00038801e9/set' with data '{"color":{"x":0.14912280701754385,"y":0.06140350877192982}}'
Zigbee2MQTT:debug 2021-05-08 15:27:47: Publishing 'set' 'color' to '0x00158d00038801e9'
XXX: colorTargets=["color_temp","hs","xy"]
XXX: convert xy -> ct
XXX: convert xy -> hs
XXX: oldState={"color_temp":370,"color_mode":"color_temp","color":{"x":0.4576,"y":0.41,"hue":35.1,"saturation":80}}
XXX: newState={"color_mode":"xy","color":{"x":0.14912280701754385,"y":0.06140350877192982}}
XXX: result={"color_mode":"xy","color":{"x":0.14912280701754385,"y":0.06140350877192982,"hue":240,"saturation":96.8627},"color_temp":548}
Zigbee2MQTT:info  2021-05-08 15:27:47: MQTT publish: topic 'zigbee2mqtt/0x00158d00038801e9', payload '{"color":{"hue":240,"saturation":96.8627,"x":0.14912280701754385,"y":0.06140350877192982},"color_mode":"xy","color_temp":548,"linkquality":132}'
```

Set hs (red, via frontend)
```
Zigbee2MQTT:debug 2021-05-08 15:28:32: Received MQTT message on 'zigbee2mqtt/0x00158d00038801e9/set' with data '{"color":{"hue":354,"saturation":100}}'
Zigbee2MQTT:debug 2021-05-08 15:28:32: Publishing 'set' 'color' to '0x00158d00038801e9'
XXX: colorTargets=["color_temp","hs","xy"]
XXX: convert hs -> ct
XXX: convert hs -> xy
XXX: oldState={"color_temp":548,"color_mode":"xy","color":{"x":0.14912280701754385,"y":0.06140350877192982,"hue":240,"saturation":96.8627}}
XXX: newState={"color_mode":"hs","color":{"hue":354,"saturation":100}}
XXX: result={"color_mode":"hs","color":{"hue":354,"saturation":100,"x":0.6373,"y":0.2703},"color_temp":140}
Zigbee2MQTT:info  2021-05-08 15:28:32: MQTT publish: topic 'zigbee2mqtt/0x00158d00038801e9', payload '{"color":{"hue":354,"saturation":100,"x":0.6373,"y":0.2703},"color_mode":"hs","color_temp":140,"linkquality":129}'
```

Set ct (neutral, via frontend)
```
Zigbee2MQTT:debug 2021-05-08 15:29:51: Received MQTT message on 'zigbee2mqtt/0x00158d00038801e9/set' with data '{"color_temp":370}'
Zigbee2MQTT:debug 2021-05-08 15:29:51: Publishing 'set' 'color_temp' to '0x00158d00038801e9'
XXX: colorTargets=["color_temp","hs","xy"]
XXX: convert ct -> xy
XXX: convert ct -> hs
XXX: oldState={"color_temp":140,"color_mode":"hs","color":{"x":0.6373,"y":0.2703,"hue":354,"saturation":100}}
XXX: newState={"color_mode":"color_temp","color_temp":370}
XXX: result={"color_mode":"color_temp","color":{"x":0.4599,"y":0.4106,"hue":31.866,"saturation":81.9608},"color_temp":370}
Zigbee2MQTT:info  2021-05-08 15:29:51: MQTT publish: topic 'zigbee2mqtt/0x00158d00038801e9', payload '{"color":{"hue":31.866,"saturation":81.9608,"x":0.4599,"y":0.4106},"color_mode":"color_temp","color_temp":370,"linkquality":132}'
```

sjorge:
Done for today, I've left out testing with reporting enabled as this will have the same behavior as doing a /get call as our fromZigbee follows the same login for attributeReport and readResponse, this makes testing easier and more consistent as we're not at the whims of the device's firmware.

Some questions that needs answers though, with the long weekend next week, I should have plenty of time to hack on it then.

sjorge:
> Great that you looking into that, while working on #2453 I tried to improve this, but got only slightly better precision in forth and back conversion so I gave up. Below you can see my attempt on this:

Results do indeed seem very close, but still seem... odd.

CT=250 => cool blueish light
CT=500 => warm orangish light

So it's weird that a red rbg color ends up closer to 250 and a blue rgb color to 500... Innr build do a conversion in firmware and what we're calculate is miles away, those do show 555 (max ct) for red and 150 (min ct) for blue RGB colors. But color conversion is not my strong side, so if both the old and your now code produce similar results, lets just go with that for now.

blazewicz:
I just checked my function and its pretty much spot-on with planckian locus produced by [python colour lib](https://colour.readthedocs.io/en/develop/generated/colour.plotting.plot_planckian_locus_in_chromaticity_diagram_CIE1931.html).

On the diagram below, thick red line are datapoints produced by my function for mireds in range (150, 450), higher value is warmer temperature.


Zrzut ekranu 2021-05-8 o 20 13 49

blazewicz:
Python code used:

```python
import numpy as np
import colour

def from_mireds(mireds):
    T = 1_000_000 / mireds

    def transform_point(T):
        if (T < 1667 or T > 25000):
            raise ValueError(f'Color temperature out of range {T:e}')
        elif (T < 4000):
            retX = -0.2661239e9 / T**3 - 0.2343589e6 / T**2 + 0.8776956e3 / T + 0.179910
            if (T < 2222):
                retY = -1.1063814 * retX**3 - 1.34811020 * retX**2 + 2.18555832 * retX - 0.20219683
            else:
                retY = -0.9549476 * retX**3 - 1.37418593 * retX**2 + 2.09137015 * retX - 0.16748867
        else:
            retX = -3.0258469e9 / T**3 + 2.1070379e6 / T**2 + 0.2226347e3 / T + 0.240390;
            retY = 3.0817580 * retX**3 - 5.87338670 * retX**2 + 3.75112997 * retX - 0.37001483
        return np.array([retX, retY])
    
    return np.array([transform_point(p) for p in T])

mireds = np.linspace(150, 450, 100)
xy = from_mireds(mireds)
fig, ax = colour.plotting.plot_planckian_locus_in_chromaticity_diagram_CIE1931([])
ax.plot(xy.T[0], xy.T[1], 'r', linewidth=3)
```

sjorge:
It's the other direction that produce weird ( to me at least) (xy -> mireds), as you shown above mireds -> xy looks pretty spot on.

e.g. 
- `{x: 0.7006, y: 0.2993}` results in color_temp of 555
- `{x: 0.1355, y: 0.0399}` results in color_temp of 153

When set on an Innr RB 250 C, which makes sense to be. Blue = cold, Red = Warm (hot event)

Yet when using lib/color they result in:
- 159.50030021441165
- 574.4479886962636

Which is opposite of what I would expect, not saying these conversions are wrong as math is not my strong suit, but the way the Innr bulbs does it in firmware seems more logical to me where blue=color_temp_min and red=color_temp_max.

blazewicz:
Actually I think everything is fine with both functions, but color you are providing cannot be mapped to Planckian Locus. I don't know how the device you are testing is doing it.

This lines crossing the Locus are isotherms, meaning that all colors along them have equal color temperature. This is why most of the spectrum can have a color temperature assigned to them.

Below you can see example of such convertions. Green dot `[0.437, 0.349]` is converted to mireds, and then back to XY, resulting in blue dot `[0.4792,  0.4141]`, a different color, but same temperature. Conversion from mireds to XY will always result with a dot on the Locus.

Zrzut ekranu 2021-05-9 o 09 39 47

This image from [Wikipedia](https://en.wikipedia.org/wiki/Planckian_locus) shows it in more detail, but in Kelvin instead of mireds.

Planckian Locus

Here you can see where on the graph is the blue color you mentioned (`{x: 0.1355, y: 0.0399}`) (green dot in bottom left corner).

Zrzut ekranu 2021-05-9 o 09 48 47

It should result in some extremely high color temperature when you look for matching isotherm.

Similarly, the red point (`{x: 0.7006, y: 0.2993}`) would result in some absurdly high mireds value too as the color temperature is reaching its asymptote around this point. Our approximation simply doesn't work in this region.

Zrzut ekranu 2021-05-9 o 09 54 30

Also here you can see a percentage error of mireds -> xy -> mireds convertion using approximations from Wikipedia:

Zrzut ekranu 2021-05-9 o 10 06 56

sjorge:
Let's go with the Innr bulb is doing funky wrong stuff (it sure as hell does a lot of other funky stuff too)

That still leaves other issues to solve before this can be merged though (https://github.com/Koenkk/zigbee-herdsman-converters/pull/2441#issuecomment-835367546)

*snip test bits*

> ❗ Our color_mode was color_temp at the time of the reads, so the initial read we end up with an updates xy and hs payload, as we ensure we preserve any read values. The read of color_temp we do after, corrects these again as we convert from ct -> xy + hs.
> 
> @Koenkk what should we be doing here?
> 
>   * keep things in sync, and discard the xy and hs values when we read color when our color_mode is set to color_temp
> 
>   * preserve the values from newState if present (current behavior), from a user perspective if I read color to get the xy and hs value... one might expect to get the actual values from the device? But this breaks the sync state, which some user might not expect?
> 
> 
> If we go with 'keep things in sync' should we log a warning?
> 
> This also applies to when reporting, as fromZigbee does not care if it is attributeReport or readResponse.
> 
> **This is not group vs device thing, it just because during this test I read color then color_temp instead of the other way around as before**

*snip test bits*

> @Koenkk there are also some other open issues/questions:
> 
>     * xy->rgb function https://github.com/usolved/cie-rgb-converter/blob/master/cie_rgb_converter.js seems to apply reverse gamma correction, not sure this should be applied or not, we get small deviations if we do so I skipped this in ColorXY.toRGB().
> 
>   * groups are fragile
>       
>     * group.device*.color_sync=true + group.color_sync=true ==> OK
>     * group.device*.color_sync=false + group.color_sync=false ==> OK
>     * group.device*.color_sync=false + group.color_sync=true ==> undefined behavior
>     * group.device*.color_sync=true + group.color_sync=false ==> undefined behavior
> 
>   * interactions with `scene_recall`, we should probably do a syncColorState there too right?
> 
> 
> Bit more info on the undefined group behavior, /set will check the group's color_sync property while a /get or a attReport for that matter will use the device's color_sync. This makes sense but results in weird behavior if the group and device(s) have different settings.
> 
> Do we leave this as is, and make it the users responsibility to ensure the group and device's option for color_sync match? That is certainly the easiest, but I think this will result in a lot of issues being opened.
> 
> I have a solution in mind, but I wasn't able to implement it because I can't figure out how to get the correct info.
> We only use the group's meta.options.color_sync if the group member's meta.options.color_sync are not the same for each member. If they are, we will use that value instead.
> 
> If the user creates a group with 4 bulbs that all have color_sync=false, the group will also use color_sync=false.
> If the devices color_sync does not match the group color_sync we should log a warning (unless the group doesn't have one and we're defaulting to true)
> 
> But I am not sure how to get at this information.

lib/color.js
```diff
@@ -219,6 +218,75 @@ class ColorXY {
...
+
+        if (isNaN(green) || Object.is(green, -0)) {
+            green = 0;
+        }
+
+        if (isNaN(blue) || Object.is(blue, -0)) {
+            blue = 0;
+        }
+
+        return new ColorRGB(red, green, blue);
```
blazewicz, line 279: `ColorRGB` class uses normalized values (0..1), with (0..255) `.toXY()` and `.toHSV()` methods won't work. Also I'd omit any rounding in processing to avoid computation errors.

lib/color.js
```diff
@@ -219,6 +218,75 @@ class ColorXY {
         return kelvinToMireds(kelvin);
     }
 
+    /**
+     * Converts CIE color space to RGB color space
+     * From: https://github.com/usolved/cie-rgb-converter/blob/master/cie_rgb_converter.js
+     * @return {ColorRGB}
+     */
+    toRGB() {
```
Koenkk, line 226: @sjorge continuing on:

> xy->rgb function https://github.com/usolved/cie-rgb-converter/blob/master/cie_rgb_converter.js seems to apply reverse gamma correction, not sure this should be applied or not, we get small deviations if we do so I skipped this in ColorXY.toRGB().


I think we should, we also do it when setting rgb via xy.

lib/color.js
```diff
@@ -219,6 +218,75 @@ class ColorXY {
         return kelvinToMireds(kelvin);
     }
 
+    /**
```
Koenkk, line 221: @sjorge Continuing on 

> interactions with scene_recall, we should probably do a syncColorState there too right?

Yes

sjorge: OK, I'll add the calls there too on the next pass.

lib/color.js
sjorge: I remove the comment so it gets applies the reverse correction..

lib/color.js
sjorge: Dropping the normalization to decimal probably will also get rid of the -0 I'm seeing 🤞

lib/color.js
```diff
@@ -563,9 +640,140 @@ class Color {
...
+/*
+ * Sync all color attributes
+ * NOTE: behavior can be disable by setting the 'color_sync' device/group option
+ * @param {Object} newState state with only the changed attributes set
+ * @param {Object} oldState state from the cache with all the old attributes set
+ * @param {Object} options meta.options for the device or group
+ * @return {Object} state with color, color_temp, and color_mode set and syncronized from newState's attributes
+ *                  (other attributes are not included make sure to merge yourself)
+ */
+function syncColorState(newState, oldState, options) {
```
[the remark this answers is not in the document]
sjorge: Yeah but currently we default to 'true' even for groups, so it should probably be
group option if set, if not, device*.option if the same, default to true ?

How would I get the option for all group members? I thought there was a function but I'm not finding it.

converters/fromZigbee.js
```diff
@@ -556,7 +557,10 @@ const converters = {
                 result.color_options = {execute_if_off: ((msg.data.options & 1<<0) > 0)};
             }
 
-            return result;
+            // handle color property sync
```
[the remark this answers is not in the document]
sjorge: I think it's still important to read xy+hs when doing a /get/color and color_temp when doing a /get/color_temp if color_sync=false. 

So something like:
color_sync=false -> same behavior as now + read color_mode so we at least color_mode will be updates with whatever the device is currently using
color_sync=true -> read color_mode, colorTemperature +  read for colorTemp, currentX, currentY, currentHue, currentSaturation, or do we go for one read with everything? I know xy+hs works fine in one read not sure we can push that to 6 attributes? I tested the Innr and it worked fine. I guess I could just test the tint, osram, ikea and hue too. If all of them support 6 attributes it's probably safe to do so.

lib/color.js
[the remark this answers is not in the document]
Koenkk: I've searched but couldn't find it. The behaviour should be same as for devices (color sync true) and we don't care about the options of members set for the group.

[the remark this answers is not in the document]
sjorge: There is still an issue of very underfined behavior though even if we always just look at the group option or device option depending on the entity... As that is just for the /set part, the /get part will always use whatever is set on the device.

So having a 3 bulbs in  a group with one of them with color_sync=false will give random results in the group mqtt topic depending on whose readResponse/attributeReport we process last. I personally don't see a use case for this. It's either going to be color_sync=false (my use case) or color_sync=true (probably everyone using HA, default) mixing false/true for different bulbs is gonna cause trouble regardless I think.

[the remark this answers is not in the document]
sjorge: Is there a 'global_options' type thing we can pass in from z2m ? To disallow having a mixed-setup ?

converters/fromZigbee.js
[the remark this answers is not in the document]
sjorge: Hmm... something else I just thought about. To avoid UNSUPPORTED_ATTRIBUTE we need a way to check if the bulb supports colorTemperature or not. (Like we do for currentHue/currentSaturation)

I guess having unique convertGet's the following is easiest:
light_color_colortemp -> read all of them (* in one or two reads)
light_color -> read call of them except colorTemperature
light_colortemp -> read colorTemperature (I guess technically we should also read colorMode and currentX/currentY, given that scenes pre ZCLv7 would use xy for colorTemperature even if the bulb does not support xy 😞 )

lib/color.js
sjorge: Will push this change once I've done the other change requests, but... it's not looking good.

```
 FAIL  test/color.test.js
  ● lib/color.js › ColorXY › .toRGB() - red

    expect(received).toStrictEqual(expected) // deep equality

    - Expected  - 3
    + Received  + 3

      Object {
    -   "blue": 0,
    -   "green": 0,
    -   "red": 1,
    +   "blue": 0.004251014680729584,
    +   "green": -0.005923094817025792,
    +   "red": 0.9999999999999999,
      }

      107 |             ['blue', {x: 0.1355, y: 0.0399}, {red: 0, green: 0, blue: 1.0}],
      108 |         ])('.toRGB() - %s', (_name, xy, rgb) => {
    > 109 |             expect(libColor.ColorXY.fromObject(xy).toRGB().toObject()).toStrictEqual(rgb);
          |                                                                        ^
      110 |         });
      111 |
      112 |         test.each([

      at test/color.test.js:109:72

  ● lib/color.js › ColorXY › .toRGB() - green

    expect(received).toStrictEqual(expected) // deep equality

    - Expected  - 3
    + Received  + 3

      Object {
    -   "blue": 0,
    -   "green": 1,
    -   "red": 0,
    +   "blue": 0.00014235404949858123,
    +   "green": 0.9999999999999999,
    +   "red": -0.00035358849383455696,
      }

      107 |             ['blue', {x: 0.1355, y: 0.0399}, {red: 0, green: 0, blue: 1.0}],
      108 |         ])('.toRGB() - %s', (_name, xy, rgb) => {
    > 109 |             expect(libColor.ColorXY.fromObject(xy).toRGB().toObject()).toStrictEqual(rgb);
          |                                                                        ^
      110 |         });
      111 |
      112 |         test.each([

      at test/color.test.js:109:72

  ● lib/color.js › ColorXY › .toRGB() - blue

    expect(received).toStrictEqual(expected) // deep equality

    - Expected  - 3
    + Received  + 3

      Object {
    -   "blue": 1,
    -   "green": 0,
    -   "red": 0,
    +   "blue": 0.9999999999999999,
    +   "green": 0.0005596045277429061,
    +   "red": -0.00012705215838029117,
      }

      107 |             ['blue', {x: 0.1355, y: 0.0399}, {red: 0, green: 0, blue: 1.0}],
      108 |         ])('.toRGB() - %s', (_name, xy, rgb) => {
    > 109 |             expect(libColor.ColorXY.fromObject(xy).toRGB().toObject()).toStrictEqual(rgb);
          |                                                                        ^
      110 |         });
      111 |
      112 |         test.each([

      at test/color.test.js:109:72
```

lib/color.js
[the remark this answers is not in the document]
sjorge: Do we need some sort of new option then to toggle this globally? Maybe not, as it seems 
```
device_options:
  color_sync: false
```

seems to apply to groups as wel, although this isn't very clear from the naming.

Koenkk:
@sjorge is this ready for merge?

sjorge:
> @sjorge is this ready for merge?

No, going over the feedback and resolving once I've fixed those things. 
Haven't pushed yet, also found issues with ColorXY.toRGB() that needs another look.

sjorge:
Will try to do testing after dinner, otherwise it will be tomorrow.
I need to get my hue bulb into my test network.

lib/color.js
```diff
@@ -219,6 +218,68 @@ class ColorXY {
         return kelvinToMireds(kelvin);
     }
 
+    /**
+     * Converts CIE color space to RGB color space
+     * From: https://github.com/usolved/cie-rgb-converter/blob/master/cie_rgb_converter.js
+     * @return {ColorRGB}
+     */
+    toRGB() {
```
sjorge, line 226: @blazewicz can you have another look at this, going from XY -> RGB I'm not getting the expected red/green/blues, they all seem off a bit and have negative values too which definitely doesn't seem right.

converters/toZigbee.js
```diff
@@ -988,7 +989,7 @@ const converters = {
               *
               * Additionally when we get a get payload, only request the fields included.
              */
-            const attributes = [];
+            const attributes = ['colorMode'];
 
             if (!meta.message.color || (typeof meta.message.color === 'object' && meta.message.color.hasOwnProperty('x'))) {
```
sjorge, line 994: @Koenkk do we still need this logic? As we're now always reading everything?

converters/toZigbee.js
```diff
@@ -1030,23 +1031,34 @@ const converters = {
...
-                result.state.color = libColor.ColorXY.fromMireds(result.state.color_temp).rounded(4).toObject();
                 return result;
             }
         },
         convertGet: async (entity, key, meta) => {
-            if (key == 'color') {
-                await converters.light_color.convertGet(entity, key, meta);
-            } else if (key == 'color_temp') {
-                await converters.light_colortemp.convertGet(entity, key, meta);
+            /**
```
sjorge, line 1041: @Koenkk this one is nearly the same as light_color, but it also reads colorTemperature.

Maybe we should love the building of `attributes` array to a separate function in lib/light.js, we can then call it in both light_color.convertGet and light_color_colortemp.convertGet. Something like

```function readLightColorAttributes(meta, additionalAttributes=[]) {}```

Then we can just set additionalAttributes to `['colorTemperature']` for light_color_colortemp.

File: converters/toZigbee.js (hunk 9 above)
Comment (Koenkk) [reply converters/toZigbee.js:994]: What do you mean with "always reading everything"? We only read what is requested?

File: converters/toZigbee.js (hunk 10 above)
Comment (Koenkk) [reply converters/toZigbee.js:1041]: Agree

File: converters/toZigbee.js (hunk 9 above)
Comment (sjorge) [reply converters/toZigbee.js:994]: Given the chance requested earlier to read colorMode so we could properly do the sync, …

only ready colorMode and say currentlX is not going to do what the user expected, unless our colorMode is xy.

**Comment by sjorge** [comment]
@Koenkk excellent! Even hue support reading all attributes at ones, I can't find my osram bulb though :(

```
Zigbee2MQTT:debug 2021-05-12 20:41:08: Received Zigbee message from '0x0017880104259333', type 'readResponse', cluster 'lightingColorCtrl', data '{"colorMode":2,"colorTemperature":370,"currentSaturation":143,"currentX":30107,"currentY":26897,"enhancedCurrentHue":8382}' from endpoint 11 with groupID 0
```

**Comment by sjorge** [comment]
Aside from the currently unresolved issues with ColorXY.toRGB() it seems to be working for a groups too, assuming that bulbs/groups have the same color_sync value of true.

image

File: converters/toZigbee.js [hunk 11]
```diff
@@ -4075,6 +4036,17 @@ const converters = {
...
                         membersState[member.getDevice().ieeeAddr] = addColorMode(member.meta.scenes[metaKey].state);
+
+                        let recalledState = member.meta.scenes[metaKey].state;
+
+                        // add color_mode if saved state does not contain it
+                        if (!recalledState.hasOwnProperty('color_mode')) {
+                            recalledState = addColorMode(recalledState);
+                        }
+
+                        recalledState = libColor.syncColorState(recalledState, member.meta.state, member.meta.options);
```
sjorge, line 4047: While testing scene_recall changes, I noticed `member.meta.state` and member.meta.options` are undefined, how would I go about getting this info?

sjorge: It's probably safe to use the group's meta.options, we can't use the groups meta state though... if we do we might leak state from the group to the device for not uniform groups!

converters/toZigbee.js
```diff
@@ -4075,6 +4036,20 @@ const converters = {
...
                         membersState[member.getDevice().ieeeAddr] = addColorMode(member.meta.scenes[metaKey].state);
+
+                        let recalledState = member.meta.scenes[metaKey].state;
+
+                        // add color_mode if saved state does not contain it
+                        if (!recalledState.hasOwnProperty('color_mode')) {
+                            recalledState = addColorMode(recalledState);
+                        }
+
+                        // XXX: meta.state is the groups state, we are leaking group state TO devices!
```
sjorge, line 4047: @Koenkk with this it works, but see the XXX comment.
There has to be a clean way of doing it. This is the last remaining issue together with the toRGB() issue

sjorge:
Phew, long evening, 2 items remaining before this is good to go:
- ColorXY.toRGB() seems to be incorrect
- scene_recall is leaking state from group -> member device which is not good at all.

lib/color.js
blazewicz: After removing reverse gamma correction it seem about right, expect of slight calculation errors.

For values below 0 I'd suggest clipping it back at 0:

```diff
diff --git a/lib/color.js b/lib/color.js
index cf252d2..a49675d 100644
--- a/lib/color.js
+++ b/lib/color.js
@@ -263,17 +263,10 @@ class ColorXY {
             blue = 1.0;
         }

-        if (isNaN(red)) {
-            red = 0;
-        }
-
-        if (isNaN(green)) {
-            green = 0;
-        }
-
-        if (isNaN(blue)) {
-            blue = 0;
-        }
+        // This fixes situation when due to computational errors value get slightly below 0, or NaN in case of zero-division.
+        red = (isNaN(red) || red < 0) ? 0 : red;
+        green = (isNaN(green) || green < 0) ? 0 : green;
+        blue = (isNaN(blue) || blue < 0) ? 0 : blue;

         return new ColorRGB(red, green, blue);
     }
```

To fix the tests you should use `.rounded()` method, like I used in test for `ColorRGB.toXY()`:

```diff
diff --git a/test/color.test.js b/test/color.test.js
index 36d4339..c9e5663 100644
--- a/test/color.test.js
+++ b/test/color.test.js
@@ -102,11 +102,11 @@ describe('lib/color.js', () => {
         });

         test.each([
-            ['red', {x: 0.7006, y: 0.299}, {red: 1.0, green: 0, blue: 0}],
+            ['red', {x: 0.7006, y: 0.2993}, {red: 1.0, green: 0, blue: 0}],
             ['green', {x: 0.1724, y: 0.7468}, {red: 0, green: 1.0, blue: 0}],
             ['blue', {x: 0.1355, y: 0.0399}, {red: 0, green: 0, blue: 1.0}],
         ])('.toRGB() - %s', (_name, xy, rgb) => {
-            expect(libColor.ColorXY.fromObject(xy).toRGB().toObject()).toStrictEqual(rgb);
+            expect(libColor.ColorXY.fromObject(xy).toRGB().rounded(4).toObject()).toStrictEqual(rgb, 4);
         });

         test.each([
````

File: lib/color.js (hunk 8 above)
Comment (blazewicz) [reply lib/color.js:226]: I responded in resolved thread, sorry.

https://github.com/Koenkk/zigbee-herdsman-converters/pull/2441/files#r632667617

File: lib/color.js [hunk 13]
```diff
@@ -219,6 +218,68 @@ class ColorXY {
...
+        } else if (blue > red && blue > green && blue > 1.0) {
+            red = red / blue;
+            green = green / blue;
+            blue = 1.0;
+        }
+
+        // Reverse gamma correction
+        red = red <= 0.0031308 ? 12.92 * red : (1.0 + 0.055) * Math.pow(red, (1.0 / 2.4)) - 0.055;
+        green = green <= 0.0031308 ? 12.92 * green : (1.0 + 0.055) * Math.pow(green, (1.0 / 2.4)) - 0.055;
+        blue = blue <= 0.0031308 ? 12.92 * blue : (1.0 + 0.055) * Math.pow(blue, (1.0 / 2.4)) - 0.055;
```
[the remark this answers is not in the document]
sjorge: I've tweaked it a bit, given that the math applied was the same, I use a transform function like the gammeCorrected function does.

sjorge:
@blazewicz Ah nice, yeah I did notice the numbers were super close without the reverse correction, so I guess it was indeed not needed like I first though.!

@Koenkk just one small issue remaining now: `scene_recall is leaking state from group -> member device which is not good at all.`

converters/toZigbee.js
Koenkk: @sjorge wasn't this the case already before? I don't understand how to `color_sync` causes this.

sjorge:
You might be right, with reporting this was already the case too.

I guess it’s good to go then?

~ sjorge

lib/color.js
```diff
@@ -563,9 +632,136 @@ class Color {
...
+ * NOTE: behavior can be disable by setting the 'color_sync' device/group option
+ * @param {Object} newState state with only the changed attributes set
+ * @param {Object} oldState state from the cache with all the old attributes set
+ * @param {Object} options meta.options for the device or group
+ * @return {Object} state with color, color_temp, and color_mode set and syncronized from newState's attributes
+ *                  (other attributes are not included make sure to merge yourself)
+ */
+function syncColorState(newState, oldState, options) {
+    const colorTargets = [];
+    const colorSync = (options && options.hasOwnProperty('color_sync')) ? (options.color_sync == true) : true;
```
Koenkk, line 646: @sjorge checking this now with zigbee2mqtt tests, but many fail due to color_sync being disabled by default. I don't have a strong opinion here, but what was the reason to disable it by default (since it was previously enabled by default)

sjorge: Does that line mean if we have color_sync and it’s true, we set it to true, otherwise false, when not set, it’s also true?

Koenkk: For default true it should be `const colorSync = (options && options.hasOwnProperty('color_sync')) ? options.color_sync : true;`

Koenkk:
@sjorge yes, found this with the z2m tests, ready to go!

PR #2445, opened 2021-04-03
Livarno Lux E27 bulb RGB does not support enhancedHue

sjorge:
As discovered in https://github.com/Koenkk/zigbee2mqtt/issues/6921 these bulbs do not support enhancedHue.

- [x] Update the saveClusterAttributeKeyValue (Wy are we even setting this if we can't read it?)
- [x] Mark enhacedHue as not support in meta


Koenkk:
Thanks!

PR #2449, opened 2021-04-04
Set enhancedHue to false for all Livarno RGB bulbs

sjorge:
As per https://github.com/Koenkk/zigbee2mqtt/issues/6921#issuecomment-812996045, apply the fix to all other bulbs too.


Koenkk:
Thanks!

PR #2450, opened 2021-04-04
[FIX #2444] color_temp_percent should honer colorTempMin/Max

### LINKED ISSUE #2444: color_temp_percent and colorTempRange
https://github.com/Koenkk/zigbee-herdsman-converters/blob/ff744d4573a68b8bb558266105986ade8318c218/converters/toZigbee.js#L763

I'm not 100% sure but I think we should use the colorTempPhysicalMin / Max as target mapping range instead of the old 153 / 500 here?

cc: @Koenkk

sjorge:
Fix for #2444


Koenkk:
Thanks!

PR #2437, opened 2021-04-01
Add support for Skydance 5-in-1 LED Controller WZ5

blazewicz:
Adds support for generic Tuya light source device as created in https://iot.tuya.com online creator.

closes https://github.com/Koenkk/zigbee2mqtt/issues/6867

PR #2453, opened 2021-04-04
Move common operations on colors to `lib/color.js`

blazewicz:
In this PR I've extracted color input normalization into z separate utility functions `light.extractColorFromValue` and `light.colorCorrectionsHSV`.

Main reason was to enable code reuse between `lightingColorCtrl` cluster and manufacturer specific Tuya API on which I'm working in #2437.

Also extracted piece of code can be tested individually and I found a bug with HSL->HSV conversion which I also fixed.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -929,7 +929,7 @@ const converters = {
                 const hsv = newColor.hsv;
                 const hsvCorrected = hsv.colorCorrected(meta);
                 newState.color_mode = constants.colorMode[0];
-                newState.color = hsv.toObject(true);
+                newState.color = hsv.toObject(false);
 
                 if (hsv.hue !== null) {
                     if (enhancedHue) {
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -1568,7 +1568,7 @@ const converters = {
                 return result;
             } else if (key == 'color_temp' || key == 'color_temp_percent') {
                 const result = await converters.gledopto_light_colortemp.convertSet(entity, key, value, meta);
-                result.state.color = libColor.miredsToXY(result.state.color_temp);
+                result.state.color = libColor.ColorXY.fromMireds(result.state.color_temp).rounded(4).toObject();
                 return result;
             }
         },
@@ -4142,9 +4142,9 @@ const converters = {
                     } catch (e) {
                         e;
                     }
-                    const newColor = libColor.Color.fromConverterArg(val);
 
-                    if (newColor.xy !== null) {
+                    const newColor = libColor.Color.fromConverterArg(val);
+                    if (newColor.isXY()) {
                         const xScaled = utils.mapNumberRange(newColor.xy.x, 0, 1, 0, 65535);
                         const yScaled = utils.mapNumberRange(newColor.xy.y, 0, 1, 0, 65535);
                         extensionfieldsets.push(
@@ -4155,7 +4155,7 @@ const converters = {
                             },
                         );
                         state['color'] = newColor.xy.toObject();
-                    } else if (newColor.hsv !== null) {
+                    } else if (newColor.isHSV()) {
                         const hsvCorrected = newColor.hsv.colorCorrected(meta);
                         if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
                             const hScaled = utils.mapNumberRange(hsvCorrected.hue, 0, 360, 0, 65535);
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -929,7 +929,7 @@ const converters = {
                 const hsv = newColor.hsv;
                 const hsvCorrected = hsv.colorCorrected(meta);
                 newState.color_mode = constants.colorMode[0];
-                newState.color = hsv.toObject();
+                newState.color = hsv.toObject(true);
 
                 if (hsv.hue !== null) {
                     if (enhancedHue) {
@@ -4181,7 +4181,7 @@ const converters = {
                                 },
                             );
                         }
-                        state['color'] = newColor.hsv.toObject();
+                        state['color'] = newColor.hsv.toObject(true);
                     }
                 }
             }
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -908,7 +908,7 @@ const converters = {
 
             if (newColor.isRGB() || newColor.isXY()) {
                 // Convert RGB to XY color mode because Zigbee doesn't support RGB (only x/y and hue/saturation)
-                const xy = newColor.isRGB() ? newColor.rgb.gammaCorrected().toXY() : newColor.xy;
+                const xy = newColor.isRGB() ? newColor.rgb.gammaCorrected().toXY().rounded(4) : newColor.xy;
 
                 // Some bulbs e.g. RB 185 C don't turn to red (they don't respond at all) when x: 0.701 and y: 0.299
                 // is send. These values are e.g. send by Home Assistant when clicking red in the color wheel.
@@ -920,7 +920,7 @@ const converters = {
                 }
 
                 newState.color_mode = constants.colorMode[1];
-                newState.color = xy.toObject(4);
+                newState.color = xy.toObject();
                 zclData.colorx = utils.mapNumberRange(xy.x, 0, 1, 0, 65535);
                 zclData.colory = utils.mapNumberRange(xy.y, 0, 1, 0, 65535);
                 command = 'moveToColor';
@@ -929,7 +929,7 @@ const converters = {
                 const hsv = newColor.hsv;
                 const hsvCorrected = hsv.colorCorrected(meta);
                 newState.color_mode = constants.colorMode[0];
-                newState.color = hsv.toObject(0);
+                newState.color = hsv.toObject();
 
                 if (hsv.hue !== null) {
                     if (enhancedHue) {
@@ -1037,7 +1037,7 @@ const converters = {
                 return result;
             } else if (key == 'color_temp' || key == 'color_temp_percent') {
                 const result = await converters.light_colortemp.convertSet(entity, key, value, meta);
-                result.state.color = libColor.ColorXY.fromMireds(result.state.color_temp).toObject(4);
+                result.state.color = libColor.ColorXY.fromMireds(result.state.color_temp).rounded(4).toObject();
                 return result;
             }
         },
@@ -4154,7 +4154,7 @@ const converters = {
                                 'extField': [xScaled, yScaled],
                             },
                         );
-                        state['color'] = newColor.xy.toObject(4);
+                        state['color'] = newColor.xy.toObject();
                     } else if (newColor.hsv !== null) {
                         const hsvCorrected = newColor.hsv.colorCorrected(meta);
                         if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
@@ -4181,7 +4181,7 @@ const converters = {
                                 },
                             );
                         }
-                        state['color'] = newColor.hsv.toObject(0);
+                        state['color'] = newColor.hsv.toObject();
                     }
                 }
             }
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -7,6 +7,7 @@ const herdsman = require('zigbee-herdsman');
 const legacy = require('../lib/legacy');
 const light = require('../lib/light');
 const constants = require('../lib/constants');
+const libColor = require('../lib/color');
 
 const manufacturerOptions = {
     xiaomi: {manufacturerCode: herdsman.Zcl.ManufacturerCode.LUMI_UNITED_TECH, disableDefaultResponse: true},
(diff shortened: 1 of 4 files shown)


lib/light.js
```diff
@@ -46,9 +48,68 @@ function clampColorTemp(colorTemp, colorTempMin, colorTempMax, logger) {
...
+        return {h: hsv[0], s: hsv[1], v: hsv[2]};
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('v')) {
+        return value;
+    } else if (value.hasOwnProperty('hsv')) {
+        const hsv = value.hsv.split(',').map((i) => parseInt(i));
+        return {h: hsv[0], s: hsv[1], v: hsv[2]};
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s')) {
+        return {h: value.h, s: value.s};
+    } else if (value.hasOwnProperty('h')) {
+        return {h: value.h};
```
sjorge, line 82: nit: return this as 'hue' as the is the term used in ZCL

lib/light.js
```diff
@@ -46,9 +48,68 @@ function clampColorTemp(colorTemp, colorTempMin, colorTempMax, logger) {
...
+    } else if (value.hasOwnProperty('s')) {
+        return {s: value.s};
```
sjorge, line 84: nit: return this as 'saturation' as the is the term used in ZCL

lib/light.js
```diff
@@ -46,9 +48,68 @@ function clampColorTemp(colorTemp, colorTempMin, colorTempMax, logger) {
...
+    } else if (value.hasOwnProperty('hsb')) {
+        const hsv = value.hsb.split(',').map((i) => parseInt(i));
+        return {h: hsv[0], s: hsv[1], v: hsv[2]};
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('v')) {
+        return value;
+    } else if (value.hasOwnProperty('hsv')) {
+        const hsv = value.hsv.split(',').map((i) => parseInt(i));
+        return {h: hsv[0], s: hsv[1], v: hsv[2]};
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s')) {
+        return {h: value.h, s: value.s};
```
sjorge, line 80: nit: return these as 'hue' and 'saturation' to stick close to ZCL

lib/light.js
```diff
@@ -46,9 +48,68 @@ function clampColorTemp(colorTemp, colorTempMin, colorTempMax, logger) {
...
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('b')) {
+        return {h: value.h, s: value.s, v: value.b};
+    } else if (value.hasOwnProperty('hsb')) {
+        const hsv = value.hsb.split(',').map((i) => parseInt(i));
+        return {h: hsv[0], s: hsv[1], v: hsv[2]};
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('v')) {
+        return value;
+    } else if (value.hasOwnProperty('hsv')) {
+        const hsv = value.hsv.split(',').map((i) => parseInt(i));
+        return {h: hsv[0], s: hsv[1], v: hsv[2]};
```
sjorge, line 78: nit: return these as 'hue', 'saturation', and 'brightness' to stick close to ZCL (if v is 0-100, we can map it to 0-254)

lib/light.js
```diff
@@ -46,9 +48,68 @@ function clampColorTemp(colorTemp, colorTempMin, colorTempMax, logger) {
...
+        return utils.hslToHSV(value.h, value.s, value.l);
+    } else if (value.hasOwnProperty('hsl')) {
+        return utils.hslToHSV(...value.hsl.split(',').map((i) => parseInt(i)));
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('b')) {
+        return {h: value.h, s: value.s, v: value.b};
+    } else if (value.hasOwnProperty('hsb')) {
+        const hsv = value.hsb.split(',').map((i) => parseInt(i));
+        return {h: hsv[0], s: hsv[1], v: hsv[2]};
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('v')) {
+        return value;
```
sjorge, line 75: nit: return these as 'hue', 'saturation', and 'brightness' to stick close to ZCL (if v is 0-100, we can map it to 0-254)

lib/light.js
```diff
@@ -46,9 +48,68 @@ function clampColorTemp(colorTemp, colorTempMin, colorTempMax, logger) {
...
+        return utils.hexToXY(typeof value === 'string' && value.startsWith('#') ? value : value.hex);
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('l')) {
+        return utils.hslToHSV(value.h, value.s, value.l);
+    } else if (value.hasOwnProperty('hsl')) {
+        return utils.hslToHSV(...value.hsl.split(',').map((i) => parseInt(i)));
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('b')) {
+        return {h: value.h, s: value.s, v: value.b};
+    } else if (value.hasOwnProperty('hsb')) {
+        const hsv = value.hsb.split(',').map((i) => parseInt(i));
+        return {h: hsv[0], s: hsv[1], v: hsv[2]};
```
sjorge, line 73: nit: return these as 'hue', 'saturation', and 'brightness' to stick close to ZCL (if v is 0-100, we can map it to 0-254)

lib/light.js
```diff
@@ -46,9 +48,68 @@ function clampColorTemp(colorTemp, colorTempMin, colorTempMax, logger) {
...
+    } else if (value.hasOwnProperty('rgb')) {
+        return utils.rgbToXY(...value.rgb.split(',').map((i) => parseInt(i)));
+    } else if (value.hasOwnProperty('hex') || (typeof value === 'string' && value.startsWith('#'))) {
+        return utils.hexToXY(typeof value === 'string' && value.startsWith('#') ? value : value.hex);
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('l')) {
+        return utils.hslToHSV(value.h, value.s, value.l);
+    } else if (value.hasOwnProperty('hsl')) {
+        return utils.hslToHSV(...value.hsl.split(',').map((i) => parseInt(i)));
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('b')) {
+        return {h: value.h, s: value.s, v: value.b};
```
sjorge, line 70: nit: return these as 'hue', 'saturation', and 'brightness' to stick close to ZCL (if v is 0-100, we can map it to 0-254)

lib/light.js
```diff
@@ -46,9 +48,68 @@ function clampColorTemp(colorTemp, colorTempMin, colorTempMax, logger) {
...
+    } else if (value.hasOwnProperty('r') && value.hasOwnProperty('g') && value.hasOwnProperty('b')) {
+        return utils.rgbToXY(value.r, value.g, value.b);
+    } else if (value.hasOwnProperty('rgb')) {
+        return utils.rgbToXY(...value.rgb.split(',').map((i) => parseInt(i)));
+    } else if (value.hasOwnProperty('hex') || (typeof value === 'string' && value.startsWith('#'))) {
+        return utils.hexToXY(typeof value === 'string' && value.startsWith('#') ? value : value.hex);
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('l')) {
+        return utils.hslToHSV(value.h, value.s, value.l);
+    } else if (value.hasOwnProperty('hsl')) {
+        return utils.hslToHSV(...value.hsl.split(',').map((i) => parseInt(i)));
```
sjorge, line 68: nit: return these as 'hue', 'saturation', and 'brightness' to stick close to ZCL (if v is 0-100, we can map it to 0-254)

lib/light.js
```diff
@@ -46,9 +48,68 @@ function clampColorTemp(colorTemp, colorTempMin, colorTempMax, logger) {
...
+    if (value.hasOwnProperty('x') && value.hasOwnProperty('y')) {
+        return value;
+    } else if (value.hasOwnProperty('r') && value.hasOwnProperty('g') && value.hasOwnProperty('b')) {
+        return utils.rgbToXY(value.r, value.g, value.b);
+    } else if (value.hasOwnProperty('rgb')) {
+        return utils.rgbToXY(...value.rgb.split(',').map((i) => parseInt(i)));
+    } else if (value.hasOwnProperty('hex') || (typeof value === 'string' && value.startsWith('#'))) {
+        return utils.hexToXY(typeof value === 'string' && value.startsWith('#') ? value : value.hex);
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('l')) {
+        return utils.hslToHSV(value.h, value.s, value.l);
```
sjorge, line 66: nit: return these as 'hue', 'saturation', and 'brightness' to stick close to ZCL (if v is 0-100, we can map it to 0-254)

lib/light.js
sjorge: technically I think we need to return this as `{"color": {"hue": x, "saturation": x}, "brightness": x}

converters/toZigbee.js
```diff
@@ -838,159 +838,46 @@ const converters = {
...
-                const hsv = utils.gammaCorrectHSV(utils.correctHue(value.hue, meta), 100, 100);
-                if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
-                    value.hue = utils.mapNumberRange(hsv.h % 360, 0, 360, 0, 65535);
-                    command = 'enhancedMoveToHue';
-                } else {
-                    value.hue = utils.mapNumberRange(hsv.h, 0, 360, 0, 254);
-                    command = 'moveToHue';
+                value.saturation = utils.mapNumberRange(hsvCorrected.s, 0, 100, 0, 254);
+                value.brightness = utils.mapNumberRange(hsvCorrected.v, 0, 100, 0, 254);
+                if (newColor.hasOwnProperty('h') && newColor.hasOwnProperty('s') && newColor.hasOwnProperty('v')) {
```
sjorge, line 860: Coupling back to https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453/files#r606783445,

This can then be simplified to just checking for hue/saturation ignoring the brightness component.

https://github.com/Koenkk/zigbee-herdsman-converters/blob/fa2f41c4514d4cd60e4cb2fe2a2cf3a239088c45/converters/toZigbee.js#L1001-L1006

This can then be move outside of the switch statement and placed behind a check for `brightness` key we extracted earlier.

lib/light.js
blazewicz: Yeah its a mess, must admit.

Some data format standardization is definitely needed here.

Currently color data is passed in many different formats with different value ranges.

Here I decided to go with `{h, s, v} (0..360, 0..100, 0..100)` because its most common format in `lib/util.js` and its basically target agnostic, representing hue in degrees, saturation and value in percent are universal, hence easy to understand when creating new integrations.

For example Tuya light source API uses string encoded HSV with ranges: `hue (0..360), saturation (0..1000), value (0..1000)` and separate call setting brightness in range `(0..1000)`. Brightness can be set both by `value` component of color and `brightness`.

Conversion to ZCL's `{hue, saturation} (0..254/0..65534, 0..254)` + `brightness (0..255)` is done only in ZCL specific part of the code (i.e. `toZigbee.light_color`.

IMHO its best to keep universal ranges in `lib/` and convert it always from base ranges in device (API) specific code. I can of course change the code to be closer to ZCL and then modify Tuya code to work with it if thats preffered.

blazewicz: Maybe I could move `extractColorFromValue` to `lib/utils.js` to avoid confusion that its related only to ZCL.

sjorge: I tend to have the ZCL open when working on things as reference. I’ve had issues a few times were internally name something that is also used in ZCL but in a different content or value range.

Given to the nature of the project I feel that sticking to ZCL defined names/structures would be preferred to avoid confusing and that makes referencing the ZCL easier. E.g we store xy without the Y component because that is what the ZCL does. So storing hue/saturation without either the brightness or level value seems consistent with that. We can store that info in the brightness property which (confusingly sometimes) we map to level later.

That being said, I have no strong preference to that over h,s,v + x,y either. Bu

sjorge: Been thinking about this a bit in my walk.

How about:
- move all color conversion functions to a new lib/color
- Have the extract function spit out 2 object: {x, y} and {h, s, l}, one of them or both can be null when there is no usable date found. (e.g. passing in only color.x and color.y in the input will have only the xy return object set and the hsl one is null)
- Discision making in the light_color converter can then be simplified to:
    - xy != null => prefer xy commands
    - xy == null && hsl != null => prefer enhancedHue commands, or hue commands if enhancedHue not supported
    - hsl != null => additional level command to update the brightness?
  
  All exported function in lib/color should then take either xy, hsl, or mired as input and return one of those as output

blazewicz: > move all color conversion functions to a new lib/color

That's a very good idea. All operations within this file should follow common, target agnostic conventions for params and returns.

I'd propose following data types:

```javascript
/**
 * @typedef {object} ColorHSV
 * @property {number} h hue component (0..360)
 * @property {number} s saturation component (0..100)
 * @property {[number]} v value component (0..100)
 */

/**
 * @typedef {object} ColorXY
 * @property {number} x X component (0..1)
 * @property {number} x Y component (0..1)
 */

/**
 * @typedef {object} ColorRGB
 * @property {number} r red component (0..1)
 * @property {number} g green component (0..1)
 * @property {number} b blue component (0..1)
 */

/**
 * @typedef {object} Color
 * @property {?ColorHSV} hsv
 * @property {?ColorXY} xy
 */

/**
 * Takes HSV color, returns HSV color.
 * @param {ColorHSV} color
 * @returns {ColorHSV}
 */
function hsvOperation(color) {
    /* ... */
}

/**
 * Takes HSV color, returns RGB color.
 * @param {ColorHSV} hsv
 * @returns {ColorRGB}
 */
function hsvToRgbOperation(hsv) {
    /* ... */
}

/**
 * Extract Color information from convertes input value.
 * @param {object} value
 * @returns {Color}
 */
function extractColorFromValue(value) {
    /* ... */
}
```

blazewicz: Or `hue, saturation, value` for HSV, but then I feel that we should also rename `r, g, b` to `red, green, blue`.

sjorge: SGTM, sorry for increasing the complexity of the PR a bit, but I think overall this should make make things a lot cleaner!

blazewicz: np, that's a good review

I already started the rewrite and added progress so far in new commit.

I'm marking this PR as a draft for now.

converters/toZigbee.js
```diff
@@ -838,177 +839,57 @@ const converters = {
...
-                    command = 'moveToHue';
+            const newColor = libColor.extractColorFromValue(value);
+            const newState = {color: newColor};
+
+            // TODO: define and implement convention what should be stored to state (ex. hsl vs {h,s,v})
+
+            if (newColor.hasOwnProperty('xy')) {
+                value.x = newColor.xy.x;
+                value.y = newColor.xy.y;
+            } else /* HSV */ {
```
sjorge, line 851: I think we should be safe here and use:

```
            } else if (newColor.hasOwnProperty('hsv')) {
                // ...
            } else {
                // throw exception about "color" not containing a valid xy or hsv payload.
            }
```

converters/toZigbee.js
```diff
@@ -838,177 +839,57 @@ const converters = {
...
-                newState.color = {hue: value.hue};
-                const hsv = utils.gammaCorrectHSV(utils.correctHue(value.hue, meta), 100, 100);
-                if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
-                    value.hue = utils.mapNumberRange(hsv.h % 360, 0, 360, 0, 65535);
-                    command = 'enhancedMoveToHue';
-                } else {
-                    value.hue = utils.mapNumberRange(hsv.h, 0, 360, 0, 254);
-                    command = 'moveToHue';
+                value.saturation = utils.mapNumberRange(hsvCorrected.saturation, 0, 100, 0, 254);
+                if (newColor.hasOwnProperty('hue') && newColor.hasOwnProperty('s')) {
```
sjorge, line 864: - newColor -> value
- check for `hue` and `saturation` as we are now using these names in value

converters/toZigbee.js
```diff
@@ -838,177 +839,57 @@ const converters = {
...
+                    if (enhancedHue) {
+                        command = 'enhancedMoveToHueAndSaturation';
+                    } else {
+                        command = 'moveToHueAndSaturation';
+                    }
+                } else if (newColor.hasOwnProperty('h')) {
```
sjorge, line 870: - newColor -> value
- check for `hue`

converters/toZigbee.js
```diff
@@ -838,177 +839,57 @@ const converters = {
...
+                    if (enhancedHue) {
+                        command = 'enhancedMoveToHue';
+                    } else {
+                        command = 'moveToHue';
+                    }
+                } else if (newColor.hasOwnProperty('s')) {
```
sjorge, line 876: - newColor -> value
- check for `saturation`

sjorge (commented):
Found a few logic flow errors

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee-herdsman-converters/pull/2437

blazewicz:
@sjorge while converting the code I realized it would be best to put into classes. Now code looks much cleaner IMO.

I'm not a JS person, so I'm open to any criticism.

converters/toZigbee.js
```diff
@@ -841,177 +842,69 @@ const converters = {
...
+
+            if (newColor.xy !== null) {
+                value.x = newColor.xy.x;
+                value.y = newColor.xy.y;
+                newState.color = newColor.xy.toObject(4);
+            } else if (newColor.hsv !== null) {
+                const enhancedHue = utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true);
+                const hsv = newColor.hsv;
+                const hsvCorrected = hsv.colorCorrected(meta);
+                newState.color = hsv.toObject();
```
sjorge, line 859: should we store corrected or uncorrect hsv here?

lib/color.js
```diff
@@ -0,0 +1,473 @@
...
+        return {hsv: new ColorHSV(h, s, b), xy: null};
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s') && value.hasOwnProperty('v')) {
+        return {hsv: new ColorHSV(value.h, value.s, value.v), xy: null};
+    } else if (value.hasOwnProperty('hsv')) {
+        const [h, s, v] = value.hsv.split(',').map((i) => parseInt(i));
+        return {hsv: new ColorHSV(h, s, v), xy: null};
+    } else if (value.hasOwnProperty('h') && value.hasOwnProperty('s')) {
+        return {hsv: new ColorHSV(value.h, value.s), xy: null};
+    } else if (value.hasOwnProperty('h')) {
+        // XXX: should this be allowed?
```
sjorge, line 452: I think this is fine to have partial data, as in ZCL we have commands to interact on partial data.

converters/toZigbee.js
blazewicz: I thought about this and I think we should store uncorrected one.

This is because end user shouldn't be aware that any correction happens at all. He selects preferred color on the color wheel and expects same color from the controlled light. Color correction should map color wheel in the app to virtual, malformed color wheel of the device and result of this correction shouldn't be of interest of end user.

If we would store state after correction, position on color when would magically jump to some other coordinates, this would be counter intuitive IMHO.

sjorge: True, but the question is then... should we reverse the correction (can we?) when the device returns the value via a read/report ?

lib/color.js
blazewicz: My original concern was about `state.color` updates. Lets say user sets `{"hue": 313, "saturation": 100}` and then changes to `{"saturation": 66}`. Resulting state would be `{..., "color": {"saturation": 66}}`

But this should be fixed in #2441, right?

sjorge:
Looking very good so far.

lib/color.js
sjorge: Perhaps the object can have a .isPartial() function to check if we have complete data or not.

converters/toZigbee.js
blazewicz: That is the tricky part. In Tuya I'm just ignoring color reported by device and just go with optimistic updates.

@Koenkk what do you think about this issue?

lib/color.js
sjorge: > My original concern was about `state.color` updates. Lets say user sets `{"hue": 313, "saturation": 100}` and then changes to `{"saturation": 66}`. Resulting state would be `{..., "color": {"saturation": 66}}`
> 

That currently not an issue as z2m will already merge the returned newState with the previous cached state.

> But this should be fixed in #2441, right?

It won't fix that specifically, as mentioned above there is nothing to fix. That PR is to fix the issue where xy, hs, and ct get out of sync.

So when we set XY, after that PR we will update hs and ct to contain values from the XY set. Like ways for setting ct we calculate xy and hs, ...

That PR will also make so we can disable this behavior and not update the state when setting the color, e.g. when using reporting for these attributes on the bulb. Some of them, .e.g Innr will do the conversion in the firmware and report all attributes. So we give the option to take the colors from the bulb too, this is useful when using a remote that can also set colors.

converters/toZigbee.js
sjorge: That should certainly be the default I think, see  https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453#discussion_r607667272

blazewicz: Reverse correction should be possible, but I think a better option would be to store the raw input in some hidden state.

lib/color.js
blazewicz: > > My original concern was about state.color updates. Lets say user sets {"hue": 313, "saturation": 100} and then changes to {"saturation": 66}. Resulting state would be {..., "color": {"saturation": 66}}
>
>That currently not an issue as z2m will already merge the returned newState with the previous cached state.

Ok, I thought it didn't merge nested objects. If it does then my concern is no longer valid.

> That PR is to fix the issue where xy, hs, and ct get out of sync. (...)

Got it, thanks

lib/utils.js
```diff
@@ -198,172 +198,6 @@ function getMetaValue(entity, definition, key, groupStrategy='first', defaultVal
...
- * Converts RGB color space to CIE color space
- * @param {Number} red
- * @param {Number} green
- * @param {Number} blue
- * @return {Array} Array that contains the CIE color values for x and y
- */
-function rgbToXY(red, green, blue) {
-    // Apply a gamma correction to the RGB values, which makes the color
-    // more vivid and more the like the color displayed on the screen of your device
-    const rgb = gammaCorrectRGB(red, green, blue);
```
blazewicz, line 212: Originally `rgbToXY` converter had this gamma correction built-in.

I've dropped it from `ColorRGB.toXY()` because I believe conversions shouldn't have any side-effects. Currently I don't have clear idea on where to put it back.

I'm thinking that maybe user input should have 3-rd option in `extractColorFromValue()` output. Then we could handle RGB input like

```javascript
if (newColor.rgb !== null) {
    const rgb = newColor.rgb;
    const xy = rgb.toXY();
    const xyCorrected = rgb.gammaCorrected().toXY();
    // this is related to problem with storing corrected vs raw input
    // what should be stored in case of RGB input?
    newState.color_mode = 'xy';
    newState.color = xy.toObject();  // TODO: conversion 0..1->0..255 is missing
    ...
}
```

I

sjorge: I think that is OK, I think this build in might also explain the slight shifts I see when doing hsv -> xy -> hsv again, which would make sense if it was applying corrections in the rgb step.

blazewicz: > That would also allow us to do rgb -> hs conversion properly if we have bubs that only support hs... not that I know of any atm.

Tuya light source supports only HSV AFAIK

sjorge: So yeah, definitely a good idea then :)

blazewicz: Ok, done

lib/light.js
blazewicz: @sjorge there is a problem with "hue" and "saturation" in state. Home Assistant doesn't seem to like it.

When I report "h" and "s" it works fine, but with "hue" and "saturation" color wheel feedback doesn't work.

With "h" and "s" dot is in the right place:

Zrzut ekranu 2021-04-7 o 09 06 05

With "hue" and "saturation" it doesn't move anywhere:

Zrzut ekranu 2021-04-7 o 09 07 09

I don't know if its a bug in HA or expected behavior though

sjorge: @Koenkk any idea? Maybe something to do with how the expose data is setup?

blazewicz: Expose setup sets "hue" and "saturation" for sure:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/22e7e9316a1356ae255b1688bc1eb8381e331ac5/lib/exposes.js#L288-L290

sjorge: https://github.com/Koenkk/zigbee2mqtt/blob/abd8a0926403d2b6d6c9a1a175f917262926c72f/lib/extension/homeassistant.js#L87-L101

You're not also exposing colorXY right? That seems to take priority if you do.

blazewicz: No, I'm definitely not, this is HA config:

```json
{
    "availability": [
        {
            "topic": "zigbee2mqtt/bridge/state"
        }
    ],
    "brightness": true,
    "brightness_scale": 254,
    "color_temp": true,
    "command_topic": "zigbee2mqtt/light_livingroom_ceiling/set",
    "device": {
        "identifiers": [
            "zigbee2mqtt_0x60a423fffe043cc3"
        ],
        "manufacturer": "TuYa",
        "model": "Zigbee & RF 5 in 1 LED Controller WZ5 (RGB+CCT mode) (TS0601_led_5ch_rgbcct)",
        "name": "light_livingroom_ceiling",
        "sw_version": "Zigbee2MQTT 1.18.1"
    },
    "hs": true,
    "json_attributes_topic": "zigbee2mqtt/light_livingroom_ceiling",
    "max_mireds": 454,
    "min_mireds": 250,
    "name": "light_livingroom_ceiling",
    "schema": "json",
    "state_topic": "zigbee2mqtt/light_livingroom_ceiling",
    "unique_id": "0x60a423fffe043cc3_light_zigbee2mqtt",
    "xy": false
}
```

blazewicz: Here, HA actually expects "h" and "s":

https://github.com/home-assistant/core/blob/6ec8e17e7b09449cd2a25f80c21732e8753b1e71/homeassistant/components/mqtt/light/schema_json.py#L231-L232

sjorge: Interesting.... https://www.home-assistant.io/integrations/light.mqtt/#hs-color

homeassistant seems to want h & s specifically and there seems to be no way of setting a custom value to grab as you can with some properties.

That probably explains issues of hs bulbs getting out of sync with HA, currently z2m copies the payload to the state.
So if you do a `../set/color` with `{"h": 120, "s": 100, "v": 100}` it would copy these fields ot the state. This would keep HA happy.

Of course when reporting is configured, the fromZigbee converter would get the report and set the `hue` and `saturation` fields for the color object. As long as the color was set by z2m this is not an issue. We just have both `hue` and `h` in the state. (Or when not using HA, most likely just `hue` as that is what I set for my homekit integration).

Same for doing a `.../get/color` as the fromZigbee convert would again set `hue`, `saturation` and/or `x`, and `y`.

But once a color changing remote comes into play (then you have to rely on a read/report) they colors would not be in sync, as the report could change it to hue=0, saturation=100 for red for example, but the h and s values HA is looking for are not updated.

The syncColor PR should fix that though, as it will copy whatever the active colorMode's values are to everything else it finds in the state.

@Koenkk that probably means we should in addition to the state, also check if homeassitant is set to true and always generate h + s values in the color payload if it is. As there could be situations where a bulb supports both and we only ever set XY via z2m but then use a remote that uses HS. In that case colorMode would change and we will sync hue/saturation -> x/y as that is in the state payload. But we won't generate h/s which is needed for HA.

Alternatively we just rename hue -> h and saturation -> s across the board ? although that would break non HA stuff.

sjorge: > Here, HA actually expects "h" and "s":
> 
> https://github.com/home-assistant/core/blob/6ec8e17e7b09449cd2a25f80c21732e8753b1e71/homeassistant/components/mqtt/light/schema_json.py#L249-L250

Yeah looks like we both came to the same conclusion at the same time, that as I mentioned also explained some weirdness when using color controling remotes I've seen reported.

blazewicz: BTW I found this: https://github.com/Koenkk/zigbee2mqtt/issues/6983

sjorge: We are already passing the color_mode properly in the replies, z2m should be good on that front

blazewicz: > @Koenkk that probably means we should in addition to the state, also check if homeassitant is set to true and always generate h + s values in the color payload if it is. As there could be situations where a bulb supports both and we only ever set XY via z2m but then use a remote that uses HS. In that case colorMode would change and we will sync hue/saturation -> x/y as that is in the state payload. But we won't generate h/s which is needed for HA.
>
> Alternatively we just rename hue -> h and saturation -> s across the board ? although that would break non HA stuff.

Or try to convince HA to support both "h"+"s" and "hue"+"saturation" in their code.

sjorge: That is also an option, but I’m not sure how open they would be to that.

They can’t cater to all possible software out there. The again they shouldn’t expect other software to cater to them either.

lib/utils.js
Koenkk: If I understand correctly, this has now been added back here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/03e0884d32a0e0dc568f474182588cdadd46f1b2/converters/toZigbee.js#L864  ?

converters/toZigbee.js
```diff
@@ -841,234 +842,98 @@ const converters = {
...
+                    // XXX: store raw RGB, raw XY or corrected XY?
+                    newState.color = {
+                        red: utils.mapNumberRange(rgb.red, 0, 1, 0, 255),
+                        green: utils.mapNumberRange(rgb.green, 0, 1, 0, 255),
+                        blue: utils.mapNumberRange(rgb.blue, 0, 1, 0, 255),
+                    };
+                    xy = rgb.gammaCorrected().toXY();
+                } else /* xy */ {
+                    newState.color = xy.toObject(4);
+                    // TODO: add gamma correction
```
Koenkk, line 867: I the color is already x/y I don't think we should do any gamma correction. So this todo can be removed.

converters/toZigbee.js
```diff
@@ -841,234 +842,98 @@ const converters = {
...
-                    value.hue = utils.mapNumberRange(hsv.h, 0, 360, 0, 254);
-                    command = 'moveToHueAndSaturationAndBrightness';
+            if (newColor.rgb !== null || newColor.xy !== null) {
+                // Convert RGB to XY color mode because at the moment all known devices support this mode.
+                newState.color_mode = constants.colorMode[1];
+                let xy;
+
+                if (newColor.rgb !== null) {
+                    const rgb = newColor.rgb;
+                    // XXX: store raw RGB, raw XY or corrected XY?
```
Koenkk, line 858: The current behaviour when setting via RGB is that the corrected XY is returned in the state. We only send the supported color modes in the ZCL: hue/saturation, x/y and color temperature (CC: @sjorge)

converters/toZigbee.js
```diff
@@ -841,234 +842,98 @@ const converters = {
...
+                    newState.color = {
+                        red: utils.mapNumberRange(rgb.red, 0, 1, 0, 255),
+                        green: utils.mapNumberRange(rgb.green, 0, 1, 0, 255),
+                        blue: utils.mapNumberRange(rgb.blue, 0, 1, 0, 255),
+                    };
+                    xy = rgb.gammaCorrected().toXY();
+                } else /* xy */ {
+                    newState.color = xy.toObject(4);
```
Koenkk, line 866: `xy.toObject(4);` always throws an exception (null pointer) since it's never initialised.

converters/toZigbee.js
sjorge: That is what I initially assumed yes, but this does not look to be true.

```
mosquitto_pub -t zigbee2mqtt/0xXXXXXXX/set/color -m '{"h": 120, "s": 100}'
```

Will currently add save/publish a color state entry with `h` and `s` set. While I expected a color state with `hue` and `saturation` set.

Base from the HA issue noted earlier, I think this was done because HA expect h and s to be present when using a hs color mode. It does not check for `hue` and `saturation`... which also explains why physical remotes that set hue and saturation, when reported or read back don't update HA.

lib/utils.js
blazewicz: you're correct

first apply gamma correction, then do transformation, but in two separate steps

converters/toZigbee.js
blazewicz: done

converters/toZigbee.js
blazewicz: ok, done

converters/toZigbee.js
blazewicz: you're right, I must have been tired when I wrote this

this piece of code is now simplified and fixed

converters/toZigbee.js
```diff
@@ -849,6 +849,11 @@ const converters = {
             const zclData = {transtime: utils.getTransition(entity, key, meta).time};
 
             if (newColor.rgb !== null || newColor.xy !== null) {
+                if (meta.state.hasOwnProperty('color')) {
+                    delete meta.state.color.hue;
+                    delete meta.state.color.saturation;
+                }
```
blazewicz: I'm not sure if this one is acceptable, but let me explain my reasoning.

I have a device that supports both `hs` and `xy`, I

1. set color to `rgb`/`xy`,
2. set color to `hs`.

I'd end up with `{state: color: {x: .., y: .., h: .., s: ..}, color_mode: "hs"}`, I know I can tell which set is correct by looking at `color_mode`, but I find it a bit confusing.

sjorge: Not sure I'd strip them though, as then it would be inconsistent with reporting.
Although if we do, that would remove the need for syncing the values...

Maybe it should have a `color_sync` and `color_strip` device option.

blazewicz:
## Important Note

For tests I use [IKEA LED1624G9](https://www.zigbee2mqtt.io/devices/LED1624G9.html#ikea-led1624g9) and Tuya LED driver from #2437.

I don't own any ZCL compatible light supporting `color_hs`.

So far both work fine, only big issue remaining is lack of compatibility with Home Assistant in `hs` mode as described in https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453#discussion_r608394765.

converters/toZigbee.js
```diff
@@ -841,234 +842,93 @@ const converters = {
...
-
-                if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
-                    value.hue = utils.mapNumberRange(hsv.h % 360, 0, 360, 0, 65535);
-                    command = 'enhancedMoveToHueAndSaturationAndBrightness';
-                } else {
-                    value.hue = utils.mapNumberRange(hsv.h, 0, 360, 0, 254);
-                    command = 'moveToHueAndSaturationAndBrightness';
+            if (newColor.rgb !== null || newColor.xy !== null) {
+                if (meta.state.hasOwnProperty('color')) {
+                    delete meta.state.color.hue;
```
sjorge, line 853: We should probably make this configurable

sjorge:
> So far both work fine, only big issue remaining is lack of compatibility with Home Assistant in `hs` mode as described in [#2453 (comment)](https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453#discussion_r608394765).

@Koenkk how about we start using `h` and `s` for these instead of `hue` and `saturation` we can copy h -> hue and s -> saturation when the **legacy** configuration property is set. Which IIRC defaults to true. That should fix the HA issue and give people time to migrate from the old properties.

Personally I like `hue` and `saturation` better though, but I am not sure how easy it is to convince HA to support both.

converters/toZigbee.js
blazewicz: Device option or application option?

converters/toZigbee.js
sjorge: For color_sync I used a device option but ran into some issues with groups... so on the application level might be better IMHO.

I'd expect you generally want all your bulbs to act the same.

converters/toZigbee.js
sjorge: For color_sync I used a device option but ran into some issues with groups... so on the application level might be better IMHO.

I'd expect you generally want all your bulbs to act the same. But either could work, I have no strong preference aside from groups + device_options is a bit trickier as I found out.

converters/toZigbee.js
blazewicz: > I'd expect you generally want all your bulbs to act the same.

I agree, and that's why I'd opt for application level config.

converters/toZigbee.js
sjorge: It would probably be sync or strip... so maybe a `color_processing` options that can take 'strip', 'sync', 'none' as value.

converters/toZigbee.js
Koenkk: I think this should be addressed in a different pr (the color_sync thing where @sjorge is working on)

converters/toZigbee.js
```diff
@@ -841,234 +842,93 @@ const converters = {
...
-                value.brightness = utils.mapNumberRange(hsv.v, 0, 100, 0, 254);
-                newState.brightness = value.brightness;
-
-                if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
-                    value.hue = utils.mapNumberRange(hsv.h % 360, 0, 360, 0, 65535);
-                    command = 'enhancedMoveToHueAndSaturationAndBrightness';
-                } else {
-                    value.hue = utils.mapNumberRange(hsv.h, 0, 360, 0, 254);
-                    command = 'moveToHueAndSaturationAndBrightness';
+            if (newColor.rgb !== null || newColor.xy !== null) {
```
Koenkk, line 851: Might be nicer to add a method to the color object so that you can do `newColor.isRGB()`, `newColor.isXY()`

converters/toZigbee.js
```diff
@@ -841,234 +842,93 @@ const converters = {
...
-                value.brightness = utils.mapNumberRange(hsv.v, 0, 100, 0, 254);
-                newState.brightness = value.brightness;
-                if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
-                    value.hue = utils.mapNumberRange(hsv.h % 360, 0, 360, 0, 65535);
-                    command = 'enhancedMoveToHueAndSaturationAndBrightness';
-                } else {
-                    value.hue = utils.mapNumberRange(hsv.h, 0, 360, 0, 254);
-                    command = 'moveToHueAndSaturationAndBrightness';
+
+                // Convert RGB to XY color mode because at the moment all known devices support this mode.
```
Koenkk, line 857: this should be: // Convert RGB to XY color mode because Zigbee doesn't support RGB (only x/y and hue/saturation)

lib/light.js
Koenkk: I think this is out of scope for this PR and it can be picked up with the color sync pr from @sjorge

converters/toZigbee.js
```diff
@@ -841,234 +842,93 @@ const converters = {
...
-                newState.brightness = value.brightness;
-                if (utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true)) {
-                    value.hue = utils.mapNumberRange(hsv.h % 360, 0, 360, 0, 65535);
-                    command = 'enhancedMoveToHueAndSaturationAndBrightness';
-                } else {
-                    value.hue = utils.mapNumberRange(hsv.h, 0, 360, 0, 254);
-                    command = 'moveToHueAndSaturationAndBrightness';
+
+                newState.color_mode = constants.colorMode[1];
+                newState.color = xy.toObject(4);
```
Koenkk, line 870: When set via XY no precision rounding should be done (only in case of RGB)

Koenkk:
> > So far both work fine, only big issue remaining is lack of compatibility with Home Assistant in `hs` mode as described in [#2453 (comment)](https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453#discussion_r608394765).
> 
> @Koenkk how about we start using `h` and `s` for these instead of `hue` and `saturation` we can copy h -> hue and s -> saturation when the **legacy** configuration property is set. Which IIRC defaults to true. That should fix the HA issue and give people time to migrate from the old properties.
> 
> Personally I like `hue` and `saturation` better though, but I am not sure how easy it is to convince HA to support both, although if we can convince them that would definitely be my preference. But I am not a HA user and have no contacts there.

This should be addressed in a different PR (color sync pr where we can pass an option if HA integration is enabled).

converters/toZigbee.js
blazewicz: I added this because original `rgbToXY` function had this 4 decimal places rounding built in.

HA sends x/y rounded to 3 decimal places so its not an issue there, but with computed values we could get full 16 decimal places in state which I don't think are ever necessary.

I don't have strong opinion here, I can remove it if you prefer so.

blazewicz:
> > > So far both work fine, only big issue remaining is lack of compatibility with Home Assistant in `hs` mode as described in [#2453 (comment)](https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453#discussion_r608394765).
> > 
> > 
> > @Koenkk how about we start using `h` and `s` for these instead of `hue` and `saturation` we can copy h -> hue and s -> saturation when the **legacy** configuration property is set. Which IIRC defaults to true. That should fix the HA issue and give people time to migrate from the old properties.
> > Personally I like `hue` and `saturation` better though, but I am not sure how easy it is to convince HA to support both, although if we can convince them that would definitely be my preference. But I am not a HA user and have no contacts there.
> 
> This should be addressed in a different PR (color sync pr where we can pass an option if HA integration is enabled).

Agreed that this should be discussed in separate PR/issue.

But for now we should choose one of available options to merge this PR and not break compatibility with HA.

1. rename `hue,saturation,value` -> `h,s,v` globally
2. update state with `{h: hsv.hue, s: hsv.saturation}`
3. keep both `h,s` and `hue,saturation` in state

converters/toZigbee.js
blazewicz: done

converters/toZigbee.js
blazewicz: done, `Color` is now a class with `is{HSV,RGB,XY}()` methods

converters/toZigbee.js
blazewicz: dropped the commit on @Koenkk request

converters/toZigbee.js
blazewicz: > I believe this was not done before so please remove it.

done

converters/toZigbee.js
```diff
@@ -841,234 +842,81 @@ const converters = {
...
+                zclData.colorx = utils.mapNumberRange(xy.x, 0, 1, 0, 65535);
+                zclData.colory = utils.mapNumberRange(xy.y, 0, 1, 0, 65535);
+                command = 'moveToColor';
+            } else if (newColor.isHSV()) {
+                const enhancedHue = utils.getMetaValue(entity, meta.mapped, 'enhancedHue', 'allEqual', true);
+                const hsv = newColor.hsv;
+                const hsvCorrected = hsv.colorCorrected(meta);
+                newState.color_mode = constants.colorMode[0];
+                newState.color = hsv.toObject(0);
```
blazewicz, line 874: Same as in https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453#discussion_r609997715

I added this because after color correction resulting values can have 18 character representation in state. HA shows full number and it breaks the color selection modal a bit.

converters/toZigbee.js
Koenkk: If with computed values you mean gamma corrected RGB I propose to only round in case of a gamma correction and don't do any rounding when x/y values were send.

Koenkk: After reading: https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453/files#r610569342

My opinion is that we should only round when computation has been done. If you directly set x/y or hue/saturation those exact values should be returned. Maybe the rounding can be done directly in the computation methods (e.g. .gammaCorrected())

sjorge:
So for the homeassistant expecting h + s, this can now be fixed by doing a move or copy here https://github.com/Koenkk/zigbee2mqtt/commit/5963f3b7f4b74dad8f026fed375c312f14de6900#diff-5cd90a075b18a077fab2f7f5a4066f63eade9d49e5ef0b023bc3878272786e9eR880, so it only effects things when HA is enabled.

sjorge:
What is still missing for this one? I'd like to pick up some of the colorsync work next week. As I will hopefully be a bit less busy then.

blazewicz:
Sorry for not responding earlier I've been pretty busy lately. I will look into finishing all remaining issues sometime this week.

converters/toZigbee.js
blazewicz: Addressed in https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453/commits/3d3933ef5696c00d3b491e53f36ddae811e0e65c, I added another intermediate method `rounded()` to apply rounding and included if after `ColorXY` object is created from color corrected RGB input.

lib/light.js
blazewicz: > I think this is out of scope for this PR and it can be picked up with the color sync pr from @sjorge . Let's keep the focus on refactoring and not behavioural changes (as this PR is already quite big)

I agree with this point. In https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453/commits/209d26d7086d1205c41777b2b310b8d349dd137b I changed default behaviour to return {h,s,v} as it was happening previously.

blazewicz:
> So for the homeassistant expecting h + s, this can now be fixed by doing a move or copy here [Koenkk/zigbee2mqtt@5963f3b#diff-5cd90a075b18a077fab2f7f5a4066f63eade9d49e5ef0b023bc3878272786e9eR880](https://github.com/Koenkk/zigbee2mqtt/commit/5963f3b7f4b74dad8f026fed375c312f14de6900#diff-5cd90a075b18a077fab2f7f5a4066f63eade9d49e5ef0b023bc3878272786e9eR880), so it only effects things when HA is enabled.

As mentioned in https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453#discussion_r624674693 I kept previous behavior. Any behavioural changes should be done in separate PR.

Koenkk:
Merging this, with the z2m tests I found the following changes (all are improvements in my opinion)
- The published colorXY is rounded now (4 decimals) when converted from the color temperature
- The old colorXY -> color temperature calculation was incorrect
- The RGB -> color XY is a little bit different now since no rounding is done anymore
- Due to rounding, slightly different value are published to the device when setting hue/saturation.

sjorge:
>     * Due to rounding, slightly different value are published to the device when setting hue/saturation.

I'm assuming we now publish the color corrected version as h + s rather than the old uncorrect value as hue + saturation?

Koenkk:
Yes, but only to the device, the same is published to MQTT

sjorge:
I'll try to clear some time to look at the sync stuff now that this is merged.

blazewicz:
Thanks @Koenkk for merging this, thank you and @sjorge for thorough review and many suggestions.

Sorry it took so long and maybe it isn't polished on every corner. This PR grew far beyond my initial assumptions and I had to learn JS while coding it, but I'm happy with the overall result. I believe any remaining issues will be fixed in the PR's to come.

Now I can get back to #2437 and hopefully I could open up a path for many new Tuya compatible lights.

sjorge:
I think it was a pretty polish PR :3

PR #2463, opened 2021-04-08
Enhance support for the Yookee D10110 and add position_min/position_max

night199uk:
I have a mix of IKEA Tradfri and Yookee (Yoolax/Graywind on Amazon) blinds. 
In:
https://github.com/Koenkk/zigbee2mqtt/issues/4611
There was a discussion of the max/min length problem with blinds. Setting max lengths for the IKEA blinds works great. But, for Yookee setting the max length does not translate to Zigbee. Sending a Zigbee 100% still unfurls the blind to maximum length. Sending a zigbee 0% goes past the minimum length set and completely rolls up the blind.

The original patch was a noted to be a little kludgy - e.g. not supporting open/close. However I primarily use HomeKit via homebridge-z2m to control my blinds which always sets {position: 100} or {position: 0} and never uses open/close. 

Therefore I resurrected the patch from here:
https://github.com/Koenkk/zigbee2mqtt/issues/4611
And polished it out a little. It's been tested with both IKEA and Yookee blinds that I have here.

While I was there I added a couple of other flags for quirks of these Yookee blinds (coverReceiveInverted and offByOne). Since otherwise these blinds are great and are about the cheapest Zigbee blinds you can buy; I hope this patch can be integrated!

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4836,6 +4836,17 @@ const converters = {
             return result;
         },
     },
+    D10110_cover_position_tilt: {
+        cluster: 'closuresWindowCovering',
+        type: ['attributeReport', 'readResponse'],
+        convert: async (model, msg, publish, options, meta) => {
+            if (msg.data.hasOwnProperty('currentPositionLiftPercentage') && msg.data['currentPositionLiftPercentage'] <= 100) {
+                // The Yookee D10110 SENDs it's position reversed, relative to the spec.
+                msg.data['currentPositionLiftPercentage'] = 100 - (msg.data['currentPositionLiftPercentage'] + 1);
+            }
+            return await converters.cover_position_tilt.convert(model, msg, publish, options, meta);
+        },
+    },
     PGC410EU_presence: {
         cluster: 'manuSpecificSmartThingsArrivalSensor',
         type: 'commandArrivalSensorNotify',
--- a/devices/yookee.js
+++ b/devices/yookee.js
@@ -10,9 +10,9 @@ module.exports = [
         model: 'D10110',
         vendor: 'Yookee',
         description: 'Smart blind controller',
-        fromZigbee: [fz.cover_position_tilt, fz.battery],
+        fromZigbee: [fz.D10110_cover_position_tilt, fz.battery],
         toZigbee: [tz.cover_state, tz.cover_position_tilt],
-        meta: {coverInverted: true},
+        meta: {coverInverted: false},
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg', 'closuresWindowCovering']);
--- a/devices/yookee.js
+++ b/devices/yookee.js
@@ -12,7 +12,6 @@ module.exports = [
         description: 'Smart blind controller',
         fromZigbee: [fz.D10110_cover_position_tilt, fz.battery],
         toZigbee: [tz.cover_state, tz.cover_position_tilt],
-        meta: {coverInverted: false},
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg', 'closuresWindowCovering']);


Koenkk:
The reason this was not integrated can be found here: https://github.com/Koenkk/zigbee2mqtt/issues/4611#issuecomment-751472585

There are several issues with this:
- OPEN/CLOSE will still fully open/close the cover
-  When a device is bound to it it will still fully open/close

In case you always send a position, why not just limit the max/min send position?

night199uk:
Thanks for reviewing!

On open/close: 
Actually, the Yookee D10110 does have the capability to set min/max positions, and open/close commands DO honour them already. The problem is those min/max positions don’t affect Zigbee position commands: sending position 0 *always* ignores the min stop and closes fully. Sending position 100 always opens to full length regardless of the max. 🤦‍♂️I guess this is a quirk of this particular manufacturer, but I hope it can be supported.

On the second question: basically “why not always send position: 2, or position: 98”? That’s basically how this was fixed for HomeAssistant. 

Because my client chain (HomeKit->Homebridge->homebridge-z2m) has no intimate knowledge of the blind amd its quirks; and all of those parts of the chain are designed to be transparent. There is no real infrastructure/logic in homebridge-z2m for per-blind configuration / quirks as far as I can see, and there is definitely none in HomeKit itself. Zigbee2mqtt is the part of the chain that knows what kind of end device this is and how to manage it’s unique quirks and allows for overriding specific device configuration. So in this case I do feel like zigbee2mqtt is the right place in the architecture to implement these quirks; I hope I can convince you :-)

night199uk:
Actually, I just read your 2nd point again and maybe I misunderstood you. 

Are you suggesting a simpler implementation in zigbee2mqtt, such as just doing:
position = numberInRange(position, min, max);
instead of scaling the number within the min max range? Or are you suggesting having homebridge/homebridge-z2m *send* a smaller position?

Koenkk:
The problem I'm having with this is that it gives user impression that they can calibrate their curtains with these options. This is in fact not the case (since it won't work for open/close commands). Calibration should be done at the curtain side, not at zigbee2mqtt.

night199uk:
I understand your concerns. Wondering if there's another way forward here?
Perhaps I could the changes out into a separate D10110_cover_position_tilt so these quirks are specific to this manufacturer and not applied to all blinds? I was trying to avoid doing that for various reasons, but wondering if a patch like that would be more acceptable?

night199uk:
Or perhaps your concerns are something that could be tackled by good documentation? I'd be quite happy to update the documentation for the position_min/position_max to explain the caveats behind using them.

Koenkk:
Looking at the ZCL spec, there seems to be a way to configure this at the device side.

Can you provide me the log when sending the following MQTT command? Send to `zigbee2mqtt/DEVICE_FRIENDLY_NAME/set` payload `{"read": {"cluster": "closuresWindowCovering", "attributes": ["installedOpenLimitLiftCm", "installedClosedLimitLiftCm"]}}`. The logged values should represent the max open/close position in centimetres.

night199uk:
Sorry; took a bit:
```
{"read": {"cluster": "closuresWindowCovering", "attributes": ["installedOpenLimitLiftCm", "installedClosedLimitLiftCm"]}}
{"level":"error","message":"No converter available for 'read' ([object Object])"}

```
I guess as it says there's no converter for read messages in fromZigbee for the Yookee D10110?

Koenkk:
Make sure you are on Zigbee2MQTT 1.18.2

night199uk:
So I finally got back to this and updated to the latest. Here's what I get with a recent zigbee2mqtt for the limitLiftCm:
{"read": {"cluster": "closuresWindowCovering", "attributes": ["installedOpenLimitLiftCm", "installedClosedLimitLiftCm"]}}
{"level":"info","message":"Read result of 'closuresWindowCovering': {\"installedOpenLimitLiftCm\":0,\"installedClosedLimitLiftCm\":65535}"}

How would I write these? Let me play about with them now.

night199uk:
Looks like that attribute is read-only perhaps for me?
mosquitto_pub -h x.x.x.x -t "zigbee2mqtt/0x60a423fffe0368a6/set" -m '{"write": {"cluster": "closuresWindowCovering", "payload": {"installedOpenLimitLiftCm": 2}}}'
{"write": {"cluster": "closuresWindowCovering", "payload": {"installedOpenLimitLiftCm": 2}}}
{"level":"error","message":"Publish 'set' 'write' to 'Office Blind 1' failed: 'Error: Write 0x60a423fffe0368a6/1 closuresWindowCovering({\"installedOpenLimitLiftCm\":2}, {\"timeout\":10000,\"disableResponse\":false,\"disableRecovery\":false,\"disableDefaultResponse\":true,\"direction\":0,\"srcEndpoint\":null,\"reservedBits\":0,\"manufacturerCode\":null,\"transactionSequenceNumber\":null,\"writeUndiv\":false}) failed (Status 'READ_ONLY')'"}

night199uk:
So I did some more testing and reading of the spec.
installedOpenLimitLiftCm/installedClosedLimitLiftCm are read-only by spec so it makes sense I couldn't write them.

I ran a bunch more experiments with the blind using the manual-recommended way of setting open/close limits; and no matter what I do the limits are not updated. It's interesting to note that the open/close limits on these blinds don't *really* act like open/close limits. e.g. with an open limit at ~2%; from fully closed, pressing open will stop at 2%. Pressing open again will then advance to 0%. A similar thing happens in reverse. So the open/close limits this blind supports aren't *real* open/close limits like the IKEA blinds.

There are other quirks of these motors I discovered while adding this patch: the position reporting is off-by-one; going from 0-99 and never *actually* reaching 100%. Also the reporting is reversed only a single direction: after setting position 0, the blind will report 99 when the operation is complete.

Koenkk:
Thanks for checking. It seems that this device doesn't behave very nicely. Therefore I suggest to create a special converter for this device instead. From that converter you can call `cover_position_tilt`. In this way this special behaviour in only available for this device. See e.g. how you can call one converter from another one: https://github.com/Koenkk/zigbee-herdsman-converters/blob/27107d048b111e3487b641ae38aac15cf7030beb/converters/toZigbee.js#L1135

night199uk:
@Koenkk So! After much back and forth with the vendor I was able to set up/down limits on my blinds correctly, and they do work exactly as for the Ikea blinds. I'm able to set the 0% and 100% accurately and have zigbee commands honour it.

However, this blind still has a 'quirk' in that the blind sends it's position reversed; i.e. for setPosition 100  the blind will return 0 on completion. So the blind is not fully inverted but half inverted. I have refreshed this PR just to accommodate that quirk now and taken your advice about how to implement it calling the underlying cover_position_tilt!

hopefully this is now good. Please take a look and let me know. I can't remove the stale tag, hopefully you can.

Koenkk:
Thanks, please also consider documenting the calibration procedure by create a PR for the docs. (click the pencil icon here: https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/devices/D10110.md)

PR #2464, opened 2021-04-09
Reduce the noise coming from small voltage changes on AU-A1ZBPIA

nrandell:
Change the voltage change threshold from 1 volt to 5 volts to reduce the amount of messages coming from the device


Koenkk:
Thanks!

PR #2467, opened 2021-04-10
Correct get/set position for bticino 4027C

PxPert:
4027C Is infact capable of Get and Set correct positions between 0 and 100. I've removed the specific bticino_4027C_cover_position and instead used the generic cover_position_tilt function to move the shutter, and removed the static "position: 50" on shutter move


Koenkk:
This was recently added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/2402 by @overclockato. @overclockato can you check this?

overclockato:
@PxPert , what your *4027C FW version?

PxPert:
Hi, Koenkk: Yes, I've checked that commit but as you can see the specific 4027C Function does only support full opening/closing of the shutters by sending a zigbee value of 0 (In case of MQTT position value < 50), or 100 (In case of MQTT position value > 50), while the 4027C I've tested fully support partial opening and partial position reporting.

This is the Zigbee dump reported by the OTA Request:
type 'commandQueryNextImageRequest', cluster 'genOta', data '{"fieldControl":0,"fileVersion":2048515,"imageType":15,"manufacturerCode":4129}' from endpoint 1 with groupID null

overclockato:
Hi,

@PxPert , I would like to know your *4027C FW version. You can get it from zigbee2mqtt WebGUI:
tapparella
Mine is 00b.
So, I have another question: did you test it on windows cover or in a balcony cover?

About code changes: I think they are not enough, if partial position reporting is working fine also bTicino_4027C_cover_state routine needs to be updated/removed. I'm not able to test this code because, I think, my FW is too old. I have to find out a way to update it.

Thanks, Bye!

PxPert:
Hi @overclockato,
It's curious because it seems I'm using an older version than yours, as you can see from the screenshot below:
Screenshot_20210411_112148

This switch is currently used as window cover.
Just one note: When the device is reset there seems to be **no** initial window reporting. The device need many full open / close rounds to automatically calibrate itself. I've seen this behaviour even when the switch is paired with standard bTicino Gateway

About the bTicino_4027C_cover_state: I've left it because it is correct to me: If you want to change the state to open/close it goes to 0 or 100. Maybe this function should only be aware of the inverted state (Just like the "standard" cover_position_tilt function does

overclockato:
Hi @PxPert,

as far I know the 1f is the newest available FW.
I planned to buy a bTicino gateway to upgrade my switches in order to support partial position values, I don't know an alternative way to do it.

About the pull request, if @Koenkk  agrees, you can merge it with a comment about the FW version.

Regards, Bye!

Koenkk:
@overclockato but does this mean it doesn't work correctly in your situation?

overclockato:
No, with the changes in this PR the reported cover position is not always reliable in my environment.

PxPert:
What about adding a specific setting to this device to disable partial position setting / reporting in case it is not supported by the firmware?

PxPert:
Closing PR, Will reopen another soon with additional option to disable partial position support

PxPert:
Opened new PR: https://github.com/Koenkk/zigbee-herdsman-converters/pull/2487

PR #2472, opened 2021-04-12
Fix Aqara D1 action expose data

tim-hellhake:


Koenkk:
Thanks!

MichelvanB (commented):
Ever sinds april my Aqara D1 double key does not work anymore. Before april it did work well.
I use Domoticz and the original settings for the 2 key worked.
Model WXKG07LM
D1
Is it something wrong in my setup? I am a beginner.

PR #2473, opened 2021-04-14
Modifying Lonsonho ZB-RGBCW with :

LucidityCrash:
proper temp range
    disabling unsupported startup color temp
    disabling effects as they only cause the Bulb to blink so no point

Adding 120 Degree version of Ajax Online AJ_ZB_GU10 with correct color temp range

Adding Aurora Lighting AOne 4W smart dimmable Vintage GLS lamp 1900K


Koenkk:
Thanks!

PR #2474, opened 2021-04-14
Add NEXENTRO Blind Actuator and reporting.currentPositionTiltPercentage

janikmu:
I aim to support a new flush-mounted blinds actuator with this PR (https://www.insta.de/files/Insta/media/product_docs/datenblaetter/DB_57008000_EN.pdf)

ZCL seems to be followed rather well, I could re-use most fromZigbee and toZigbee stuff, but additionally implemented a new reporting function, because I have use-cases where I only change the angle of the blinds but not the positions of the whole window cover and I want these (possibly external) changes reflected.

What I don't like is the modelId name "Generic UP Device". The manufacturer has promised to change this is the near future, and I will likely follow up with a new PR then. While the manufacturer name is "Insta GmbH", it indeed is "Insta" (whitelabel Gira and Jung products), as the manufacturer ID (4474) matches - is it best practice to name new products after an existing manufacturer, or use the supplied manufName even if it doesn't match in this case? 

What currently does not work, and I don't know why, is setting the tilt value, as `tz.cover_position_tilt` should support this.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -12751,13 +12751,11 @@ const devices = [
         toZigbee: [],
         ota: ota.zigbeeOTA,
     },
-
-    // NEXENTRO by Insta
     {
         zigbeeModel: ['Generic UP Device'],
         model: '57008000',
-        vendor: 'Insta GmbH',
-        description: 'Blinds actor with Lift/Tilt Calibration & with with inputs for wall switches',
+        vendor: 'Insta',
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

Domain4dmin:
The Nexentro Dimmor also identifies itself with "model":"57008000". The device joins successfully, but can't be used ("Device '...' has no endpoint 'undefined'"). 

What would be the solution here? Ask the manufacturer to use unique model identifiers?

PR #2476, opened 2021-04-15
Added 2 Matcall BV devices

sunsear:
Tested with externalConverters on my HassOS installation. Unfortunately I cannot test devices.js in this setup, I have thoroughly checked the syntax, but please check again.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -17523,7 +17523,7 @@ const devices = [
         description: 'Wireless Controlled LED retrofit downlight',
         extend: preset.light_onoff_brightness_colortemp({colorTempRange: [200, 370]}),
     },
-    
+
     // Matcall BV
     {
         zigbeeModel: ['ZG 401224'],
@@ -17533,7 +17533,7 @@ const devices = [
         exposes: [e.light_brightness()],
         fromZigbee: [fz.on_off, fz.brightness, fz.ignore_basic_report],
         toZigbee: [tz.light_onoff_brightness, tz.ignore_transition, tz.ignore_rate, tz.light_brightness_move,
-            tz.light_brightness_step, tz.level_config, tz.power_on_behavior]
+            tz.light_brightness_step, tz.level_config, tz.power_on_behavior],
     },
     {
         zigbeeModel: ['ZG 430700', 'ZG  430700'],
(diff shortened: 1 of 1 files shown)


sunsear:
OK, listing did that syntax check for me ;-)

I have seen of course that my 2 devices could extend from light_onoff_brightness, but I am unable to test that and prove that it works as I need to run them from External Converters in my HassOS setup.

devices.js
```diff
@@ -17524,6 +17524,28 @@ const devices = [
         extend: preset.light_onoff_brightness_colortemp({colorTempRange: [200, 370]}),
     },
 
+    // Matcall BV
+    {
+        zigbeeModel: ['ZG 401224'],
+        model: 'ZG 401224',
+        vendor: 'Matcall bv',
+        description: 'Matcall LED Dimmer driver',
+        exposes: [e.light_brightness()],
```
Koenkk, line 17533: For both `exposes` `fromZigbee` and `toZigbee` can be replaced with `extend: preset.light_onoff_brightness(),`

devices.js
```diff
@@ -17524,6 +17524,28 @@ const devices = [
...
+        fromZigbee: [fz.on_off, fz.brightness, fz.ignore_basic_report],
+        toZigbee: [tz.light_onoff_brightness, tz.ignore_transition, tz.ignore_rate, tz.light_brightness_move,
+            tz.light_brightness_step, tz.level_config, tz.power_on_behavior],
+    },
+    {
+        zigbeeModel: ['ZG 430700', 'ZG  430700'],
+        model: 'ZG 430700',
```
Koenkk, line 17540: Please remove space for both model numbers (see bottom of page 26: https://www.matcall.nl/wp-content/uploads/documentatie/MC-Easy-systeem.pdf)

sunsear: These model numbers are the actual reported model numbers in the Zigbee stack. I guess it does make more sense to make them list as the documentation does. On their website they are also referred to without spaces. Will make changes!

devices.js
sunsear: OK will do, although I cannot test that myself

sunsear:
Thanks for review Koen, I have added the requested changes.

Koenkk:
Thanks!

PR #2478, opened 2021-04-15
TuYa TS0503B LED Drv does not support enhancedHue

sjorge:
As observed in https://github.com/Koenkk/zigbee2mqtt/issues/7088, this device does not support enhancedHue


Koenkk:
Thanks!

PR #2479, opened 2021-04-15
Fix Bitron 902010/29 outdoor siren

FlyingDomotic:
Replace fz.ias_smoke_alarm_1 by fz.battery and remove e.smoke() as device is definitively an outdoor siren, not a smoke detector;-)


Koenkk:
Thanks!

PR #2480, opened 2021-04-15
Added another manufacturerName for RTX ZB-RT1 radiator valve

mictyd:
Hi,
I recently bought a ZB-RT1 by RTX radiator valve and it is recognized as `{modelId: 'TS0601', manufacturerName: '_TZE200_azqp6ssj'}`. It is already supported, but not by this model/manufacturer, so in this PR I'm adding it.


Koenkk:
Thanks!

PR #2487, opened 2021-04-17
Correct get/set position for bticino 4027C

PxPert:
Newer 4027C firmware Is infact capable of Get and Set correct positions between 0 and 100.
I've added support for partial position setting and reporting.
Optionally, in case of old firmware that doesn't support position reporting, it's possible to keep the traditional functionality by setting this specific device option:
  no_position_support: true
I've also fixed "invert_cover" support to be coherent with all the get / set functions


Koenkk:
@overclockato is this OK for you?

Please also make a pr for the docs: https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/devices/K4027C_L4027C_N4027C_NT4027C.md

PxPert:
Hi @Koenkk Sure.
Let's wait for feedback from @overclockato and then I'll update the docs

overclockato:
Hi @Koenkk , Hi @PxPert ,

the changes are okay for me too.

Thanks @Koenkk , thanks @PxPert, Regards.

Koenkk:
Thanks!

PR #2491, opened 2021-04-18
Add FK_V02

Koenkk:
https://github.com/Koenkk/zigbee2mqtt/issues/7019

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -2473,6 +2473,25 @@ const converters = {
             return {state: {state: value.toUpperCase()}};
         },
     },
+    frankever_threshold: {
+        key: ['threshold'],
+        convertSet: async (entity, key, value, meta) => {
+            // input to multiple of 10 with max value of 100
+            const thresh = Math.abs(Math.min(10*(Math.floor(value/10)), 100));
+            await tuya.sendDataPointValue(entity, tuya.dataPoints.frankEverTreshold, thresh, 'setData', 1);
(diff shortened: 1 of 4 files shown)


converters/toZigbee.js
```diff
@@ -2473,6 +2473,25 @@ const converters = {
...
+        key: ['threshold'],
+        convertSet: async (entity, key, value, meta) => {
+            // input to multiple of 10 with max value of 100
+            const thresh = Math.abs(Math.min(10*(Math.floor(value/10)), 100));
+            await tuya.sendDataPointValue(entity, tuya.dataPoints.frankEverTreshold, thresh, 'setData', 1);
+            return {state: {threshold: value}};
+        },
+    },
+    frankever_timer: {
+        key: ['timer'],
```
Koenkk, line 2486: @ArJay60  I noticed you created a timer for seconds minutes and hours, since it writes the same data point this makes no sense in my opinion. (it's the same things exposes 3 times). Therefore I only added the minute one.

ArJay60: Adding only minutes would be fine. The reason for the multiple exposes is the granularity of the slider to set a certain timer value. For instance the minute slider would have a reach from 0 to 3600. The ability to set a time for let's say 7 minutes will be very difficult to do with the slider ( fill it in numerical not of course). This was the reason the have multiple exposes.

Koenkk: Thanks!

PR #2492, opened 2021-04-18
add colorTempRange for Paul Neuhaus NLG-TW light

BMOD89:


BMOD89:
@Koenkk could you please explain the advantage of using colorTempRange? 

I have an issue with this lamp, that the color temperature is much warmer compared to other lamps/bulbs when grouping them with home assistant and assigning the same color temperature. Is there a way to add a correction map for the color temperature?

PR #2494, opened 2021-04-20
Add support for unbranded version of Orvibo RL804ZB

sdaicz:
I've got an unbranded version of [Orvibo RL804ZB](https://www.zigbee2mqtt.io/devices/RL804QZB.html)
It's identical to the pictures I've see of the branded version.  The only differences are:
* They removed the word Orvibo from the case and the manual
* The model number in the case reads SH-222ZB
* The zigbeeModel is different.

I think there's no need to create a separate entry in `devices.js`, but please let me know if you prefer me to do it.

### DIFF
--- a/devices.js
+++ b/devices.js
@@ -14404,7 +14404,7 @@ const devices = [
         extend: preset.light_onoff_brightness_colortemp_color(),
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2499, opened 2021-04-21
Add Hue Lily outdoor spot light base pack

Tapematch:
I have four Hue Lily outdoor spot lights, one base pack consisting of three lights and a power supply, and an extension pack with one light.
Except for the ZigBee model numbers the lights are the same.
The extension pack was already supported, the lights in the base pack strangely weren't.
I have added the model number of the lights in the base pack.


Koenkk:
Thanks!

Issue #2495, opened 2021-04-20
Github action to sort devices.js


svrooij:
Sometimes it is quite hard to find some device in the devices.js file.

Maybe it's a good idea to enforce alphabetical order with some kind of script. And off course a easy to use script to organize all the devices. What do you think off this idea?

nurikk:
Ideally we can split device definitions in sub-packages, like for xiaomi, philips, tuya
```

./devices.js

./devices/

./devices/xiaomi/devices.js

./devices/xiaomi/fromZigbee.js

./devices/xiaomi/toZigbee.js
```

Koenkk:
@nurikk I also had this idea already. Will try to do this soon.

Koenkk:
Created https://github.com/Koenkk/zigbee-herdsman-converters/pull/2500, please review

PR #2500, opened 2021-04-21
Split devices.js

Koenkk:
Fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/2495 

CC: @nurikk @svrooij  @sjorge 

- Created `scripts/splitDevices.js` to automate the split, splits vendors if they use more than 100 lines in devices.js
- TODO:
  - [x] Fix lint
  - [x] Move vendor specific `lib/presets.js` into vendor file


svrooij:
Splitting is nice, you just wrote a small program for it 😁

But is was actually talking about ordering devices based on vendor (not needed anymore) and model ID.

I'm not sure how to do that, parse the content of the device blocks and the reading the model id and reordering.

Maybe your split script can also be checking the need for the Xiaomi/tuya/hue import.

sjorge:
I like this, this certainly makes dealing with devices easier as the old huge devices.js was cumbersome at times.

Is there a performance impact at startup? If not, I wonder if we should go to a file-per-device type structure. That would certainly make it easier to have people try new device definitions or even different ones as they can just replace a single file completely.

svrooij:
> I wonder if we should go to a file-per-device type structure. That would certainly make it easier to have people try new device definitions or even different ones as they can just replace a single file completely.

Maybe even a separate npm package per vendor?

And how about just generating the file per vendor by creating a JSON file with a device definition. For example we create a JSON like:

```json
[
  {
    "vendor": "Innr",
    "modelId": "RSL115",
    "deviceType": "LightWithBrightness",
    ...
  }
]
```

And then create a script that will generate the device file per vendor.

svrooij:
@sjorge a file per device makes “check the device.js for existing device support” kind of hard. And browsing hunderds of files to find the correct model name also isn’t nice.

sjorge:
> @sjorge a file per device makes “check the device.js for existing device support” kind of hard. And browsing hunderds of files to find the correct model name also isn’t nice.

We can probably ship a small script that  could do the check?

nurikk:
Vendor specific toZigbee/fromZigbee converters might be also worth to move in vendor-specific files

nurikk:
> > @sjorge a file per device makes “check the device.js for existing device support” kind of hard. And browsing hunderds of files to find the correct model name also isn’t nice.
> 
> We can probably ship a small script that could do the check?

Good old `Cmd+F/Ctrl+F` will help :) Or github search bar

sjorge:
> @sjorge I think per vendor is good enough, per device will generate a lot of duplicate code + files.
> 
True, all the boiler plate would be duplicated the.

Koenkk:
I personally really liked the 1 devices.js file, but since it grew so big e.g. the Github editor couldn't deal with it anymore. This was for me the the only reason to split it. I still prefer to have as little as files as possible.

svrooij:
@Koenkk the "ordered by model id" is just me being a big fan off order. Ctrl + F should work just fine.

Koenkk:
Merged now, 2 follow up actions for me:
- Completely remove devices.js (but first https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html has to be updated).
- Allow `extend` for external converters and add more examples for it.

PR #2503, opened 2021-04-22
Incorrect type use

alparslana:
In the cover_position_via_brightness message, since the position variable is initialized as string, the produced json output sends position field with in string or numerical, depending on the value. The documentation states this field is numerical, therefore I suggest removal of toString() for correct data output.


Koenkk:
Thanks!

PR #2506, opened 2021-04-24
Remove output of enrolled due to missing zoneState

lnagel:
I propose to remove this attribute because it was highly confusing for me why the enrollment appears to be unsuccessful.

By looking at all the logs, it seems that this variable isn't even available, so this status can't be true at all.

image

image

Please let me know if I have misunderstood something.


Koenkk:
Agree, this is also the only `ias_*` converter that has ths.

PR #2514, opened 2021-04-26
added lock user status functions for YRD246 lock

jpg0:
This depends on https://github.com/Koenkk/zigbee-herdsman/pull/353

(Lock User Status allows switching user "slots" between available, enabled and disabled. This is part of the standard door lock zigbee cluster spec, but I have only enabled it for the lock mentioned in the title.)


Koenkk:
Thanks!

PR #2515, opened 2021-04-26
Update HESZB-120 to match changes in SMSZB-120

lnagel:
SMSZB-120 was updated previously to remove tamper, add test.


Koenkk:
Thanks!

PR #2517, opened 2021-04-26
Fix 2mn auto-Off on Tuya TS0013 and TS0014

H3buss:
Those switch auto-turn Off after 2mn. Applied same bind as TS0011 and TS0012 (tested working on TS0012 and TS0013)


Koenkk:
Thanks, after this change do you still get status updates when controlling the switch manually?

H3buss:
Yes I still get the updates on all switch endpoints.

Koenkk:
Great, thanks

PR #2526, opened 2021-04-27
Remove devices.js

Koenkk:
Follow up of https://github.com/Koenkk/zigbee-herdsman-converters/pull/2500, this pr removes the `devices.js` file. All files within the `devices` folder are loaded by `index.js` now.

[How to support new devices](https://github.com/Koenkk/zigbee2mqtt.io/blob/develop/docs/how_tos/how_to_support_new_devices.md) has been updated to use external converters instead. The external converters also support `extend` now; [example](https://github.com/Koenkk/zigbee2mqtt.io/blob/a1d369a65e5f280bac5426cee931aa0c62ffab6c/docs/externalConvertersExample/light.js#L22)

CC: @nurikk @sjorge @svrooij @kirovilya


README.md
```diff
@@ -3,12 +3,32 @@
...
+
 If any of those commands finish with an error your PR won't pass the tests and will likely be rejected.
+
+## Documentation of definition meta property
+- `configureKey`: required when a 'configure' is defined, this key is used by the application to determine if the content of the configure has been changed and thus needs to re-execute it. For a currently unsupported device you can set this to 1.
+- `multiEndpoint`: enables the multi endpoint functionallity in e.g. fromZigbee.on_off, example: normally this converter would return {"state": "OFF"}, when multiEndpoint is enabled the 'endpoint' method of the device definition will be called to determine the endpoint name which is then used as key e.g. {"state_left": "OFF"}. Only needed when device sends the same attribute from multiple endpoints.
+- `disableDefaultResponse`: used by toZigbee converters to disable the default response of some devices as they don't provide one.
+- `applyRedFix`: see toZigbee.light_color
+- `enhancedHue`: see toZigbee.light_color
+- `supportsHueAndSaturation`: see toZigbee.light_color
```
sjorge, line 24: perhaps mention defaults? e.g.

enhancedHue=>true
supportsHueAndSaturation=>true
applyRedFix=>false
...

README.md
```diff
@@ -3,12 +3,32 @@
...
 If you'd like to submit a pull request, you should run the following commands to ensure your changes will pass the tests:
 ```sh
 npm run lint
 npm test
 ```
+
 If any of those commands finish with an error your PR won't pass the tests and will likely be rejected.
+
+## Documentation of definition meta property
+- `configureKey`: required when a 'configure' is defined, this key is used by the application to determine if the content of the configure has been changed and thus needs to re-execute it. For a currently unsupported device you can set this to 1.
```
sjorge, line 19: I wonder, can we somehow hash the content of configure() and use that as dynamic key in the database?
If the hash changes, we need to reconfigure? No more messing with increasing ids. Out of scope of this though, just something that popped into my head.

sjorge (approved):
LGTM, nice work on this!

README.md
Koenkk: Done

README.md
Koenkk: Also thought about this (not sure if possible) but there are also situations where we want to change the configure() but do not want it to be reapplied (common for sleeping end devices).

README.md
```diff
@@ -3,12 +3,32 @@
...
+- `multiEndpoint`: enables the multi endpoint functionallity in e.g. fromZigbee.on_off, example: normally this converter would return {"state": "OFF"}, when multiEndpoint is enabled the 'endpoint' method of the device definition will be called to determine the endpoint name which is then used as key e.g. {"state_left": "OFF"}. Only needed when device sends the same attribute from multiple endpoints. (default: false)
+- `disableDefaultResponse`: used by toZigbee converters to disable the default response of some devices as they don't provide one. (default: false)
+- `applyRedFix`: see toZigbee.light_color (default: false)
+- `enhancedHue`: see toZigbee.light_color (default: false)
```
sjorge, line 23: The default of this is true

README.md
sjorge: I general doing a reconfigure with sleeping devices is tricky currently.

I think this wouldn’t be an issue if we just set a flag that it needs a reconfigure and execute it once the device checks in. That would also take the quick time event type situation away when trying to manually reconfigure a sleeping device that can’t easily be moved and is tricky to reach.

nurikk:
Preem work choom 👍RIP devices.js

README.md
Koenkk: Fixed

README.md
Koenkk: But what e.g. for the sonoff case were it can only be configured after rejoining?

sjorge: The bind is the problem there yeah, reporting can be updates just fine.
I suppose we can introduce a `configureOnce` meta or something. Where it will not try to be smart. As those, should be the exception.

Koenkk: Agree, its also the only device I know which has this problem. Note that reconfiguring is already retried when a Zigbee message is received from a device. Maybe we should just not do the `configure()` on startup for end devices (only when we receive a message from it). Found some resource for getting a method body: https://stackoverflow.com/questions/14885995/how-to-get-a-functionss-body-as-string . Anyway let's fix it in a different issue.

sjorge: Yeah as I said early, this is something for separate PR for sure.
I think for end devices we should indeed only do a reconfigure when we receive a message and (db saved hash != hash) && !meta.configureOnce, and after interviewing.

PR #2532, opened 2021-04-29
Remove configureKey from meta

Koenkk:
This removes the configureKey from meta, the configureKey can now be calculated instead. This saves us from manually having to bump this configureKey every-time. For many users it was also vague what this actually is, so this should avoid confusion.

This is a **breaking change**, previously the configureKey could be obtained with:

```js
const definition = myDevice;
const configureKey = definition.meta.configureKey
```

After this change it can be obtained with:

```js
const zhc = require('zigbee-herdsman-converters');
const definition = myDevice;
const configureKey =  zhc.getConfigureKey(definition);
```

Due to removal of all configureKey the PR is big, but the magic happens here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/2532/files#diff-1ab56e1a93b957ed970ce232988c20ed173a88d5cf8f6426849a51fcc981b73b

CC: @sjorge @nurikk @kirovilya


sjorge (approved):
LGTM

PR #2536, opened 2021-05-01
add support for another model number of the Econic light 1744030V7

gordlea:
This light appears to be the same as 1744030P7. I have them here and can confirm they are working with this.


Koenkk:
Thanks!

PR #2544, opened 2021-05-04
Add Wiser rotary dimmer.

rindlerblabla:
Would fix https://github.com/Koenkk/zigbee2mqtt/issues/7177. But think I need a bit of help with this PR, since it's adding ballast configuration. Am not really sure what I am doing, hehe.

### DIFF
--- a/devices/schneider_electric.js
+++ b/devices/schneider_electric.js
@@ -80,7 +80,7 @@ module.exports = [
     },
     {
         zigbeeModel: ['NHROTARY/DIMMER/1'],
-        model: 'Rotary dimmer',
+        model: 'WDE002334',
         vendor: 'Schneider Electric',
         description: 'Rotary dimmer',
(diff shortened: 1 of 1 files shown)


devices/schneider_electric.js
```diff
@@ -78,6 +78,24 @@ module.exports = [
             await reporting.onOff(endpoint);
         },
     },
+    {
+        zigbeeModel: ['NHROTARY/DIMMER/1'],
+        model: 'Rotary dimmer',
```
Koenkk, line 83: Please add the model written on the box or device (rule of thumb is that is should be googleable). Rest of of the code looks good, what part do you still question?

rindlerblabla: Sure! The model name differs depending on color but will add the name of the white standard one, WDE002334, which is what Schneider uses the most. Regarding the rest it was more a uncertainty from me, since I didnt find as many other examples using ballast config, but it works just fine locally when using my external converter.

Koenkk: Good, I will merge this.

PR #2546, opened 2021-05-04
Added Philips Hue Calla with zigbee model 1742030P7

wersim:
z2m already supports one of the small Philips Hue Calla models:
https://www.zigbee2mqtt.io/devices/8718696167991.html

Philips updated the zigbee model of this extension to 1742030P7:
https://www.philips-hue.com/de-de/p/hue-white---color-ambiance-calla-outdoor-sockelleuchte/1742030P7

So, this is just a copy of the already supported 8718696167991. I currently use this setup via an external converter in HASS.


test/colortemp_range_missing_allowed.txt
```diff
@@ -27,6 +27,7 @@ LED1923R5
 7602031P7
 8718696167991
 8718696170557
+1742030P7
```
Koenkk, line 30: Please don't add this but add the color temp range instead as described here: https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html#31-retrieving-color-temperature-range-only-required-for-lights-which-support-color-temperature

wersim: Thanks for advice! I've requested the range values via the lamp's topic and added them.

Debug Received Zigbee message from 'Terrasse_Calla_3', type 'readResponse', cluster 'lightingColorCtrl', 
data '{"colorTempPhysicalMax":500,"colorTempPhysicalMin":153}' from endpoint 11 with groupID 0

I'm not sure with regards to the zigbee2mqtt.io documentation. Did I understand it correctly that it will be autogenerated somehow or do I have to add at least the image for examle?

Koenkk: Yes it will be auto generated. Feel free to add an image, otherwise I will do it right before the next release (1 June)

PR #2548, opened 2021-05-05
add support for Heiman motion sensor HS1MS-M

accelle17:
I am migrating from deconz to z2m and got some HS1MS-M Heiman sensors. I have tested this using external converter and was able to made it work.


accelle17:
Hold on. Pulled too early. Tamper state works but it looks like the occupancy sensor doesn't go back to false. Is there a way to debug this or troubleshoot?

Koenkk:
Thanks!

PR #2550, opened 2021-05-05
Adds ballast config CCT5010-0001

rindlerblabla:
... and reporting brightness WDE002334.
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/7188.

### DIFF
--- a/devices/schneider_electric.js
+++ b/devices/schneider_electric.js
@@ -59,11 +59,16 @@ module.exports = [
         model: 'CCT5010-0001',
         vendor: 'Schneider Electric',
         description: 'Micro module dimmer',
-        extend: extend.light_onoff_brightness({noConfigure: true}),
+        fromZigbee: [fz.on_off, fz.brightness, fz.level_config, fz.lighting_ballast_configuration],
+        toZigbee: [tz.light_onoff_brightness, tz.level_config, tz.ballast_config],
+        exposes: [e.light_brightness().withLevelConfig(),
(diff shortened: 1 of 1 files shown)


devices/schneider_electric.js
```diff
@@ -59,11 +59,16 @@ module.exports = [
...
-        extend: extend.light_onoff_brightness({noConfigure: true}),
+        fromZigbee: [fz.on_off, fz.brightness, fz.level_config, fz.lighting_ballast_configuration],
+        toZigbee: [tz.light_onoff_brightness, tz.level_config, tz.ballast_config],
+        exposes: [e.light_brightness().withLevelConfig(),
+            exposes.numeric('ballast_minimum_level', ea.ALL).withValueMin(1).withValueMax(254)
+                .withDescription('Specifies the minimum light output of the ballast'),
+            exposes.numeric('ballast_maximum_level', ea.ALL).withValueMin(1).withValueMax(254)
+                .withDescription('Specifies the maximum light output of the ballast')],
         configure: async (device, coordinatorEndpoint, logger) => {
-            await extend.light_onoff_brightness().configure(device, coordinatorEndpoint, logger);
```
Koenkk, line 64: This line should not be removed.

rindlerblabla: Okay? I thought that when replacing "extend: extend.light_onoff_brightness" with fromZigbee, toZigbee and exposes that I should remove that line, but I should keep it still?

Koenkk: Yes please, otherwise the default configure method won't be called anymore.

Koenkk:
Thanks!

PR #2553, opened 2021-05-06
Add CCTFR6700 from Schneider electric

H3buss:
To work properly this device requires a temperature report from an external thermostat.

I've also made it work with a temperature report from the coordinator, but this requires adding a 'report' command into herdsman controller. It works with Z-stack based keys but I don't know if this could be supported by others.


converters/toZigbee.js
```diff
@@ -4828,6 +4828,20 @@ const converters = {
...
+        convertSet: async (entity, key, value, meta) => {
+            const lookup = {'relay': 1, 'pilot': 3};
+            value = value.toLowerCase();
+            utils.validateValue(value, Object.keys(lookup));
+            const mode = lookup[value];
+            await entity.write('schneiderSpecificPilotMode', {'pilotMode': mode}, {manufacturerCode: 0x105e});
+            return {state: {schneider_pilot_mode: mode}};
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('schneiderSpecificPilotMode', ['pilotMode'], {manufacturerCode: 0x105e});
```
Koenkk, line 4842: This get doesn't add any value since there is no corresponding fromZigbee converter. Please add one. (you can test it by doing `zigbee2mqtt/FRIENDLY_NAME/get` with payload `{"schneider_pilot_mode":""}`

H3buss:
Should be fixed, sorry for failed checks.

H3buss:
Note: the device requires a bind from another device providing msTemperatureMeasurement, maybe this should be added in description.

Koenkk:
Thanks!

You can also make a PR for the docs (zigbee2mqtt.io)

H3buss:
I still have to make the associated ambiant thermostat work (CCTFR6400). I have some questions on how to do that, how can I contact you for some insight?

Koenkk:
I prefer communication here, what is the issue?

H3buss:
1) The Thermostat CCTFR6700 (the one of this pull request) gets its measured temperature from an other device attribute report of a temperature cluster. I can send it from coordinator but this requires adding the 'report' capability to herdsman. It works, but is it the proper way to do things? 
The attribute cannot be written to: the device fails with 'unsupported attribute' when trying to do so.

2) The associated thermostat CCTFR6400 cannot bind to a device, it fails with 'bind table full'.  So binding is not a solution to report the temperature to CCTFR6700.
(Side note: this 'bind table full' failure response is not understood by herdsman, which reports the bind as successfull).

3) The thermostat is a battery device whith a UI that reads the thermostat status (setpoint, piHeatingDemand, etc..) from the coordinator when woken up, and expects a read response. 
I don't understand how to set this up in herdsman.

Here is what I currently have:
`{
        fingerprint: [{modelID: 'Thermostat', manufacturerName: 'Schneider Electric'}],
        model: 'CCTFR6400',
        vendor: 'Schneider Electric',
        description: 'Temperature/Humidity measurement with thermostat interface',
        fromZigbee: [fz.battery, fz.temperature, fz.humidity, fz.thermostat],
        toZigbee: [tz.thermostat_system_mode, tz.thermostat_running_state, tz.thermostat_local_temperature,
            tz.thermostat_occupied_heating_setpoint, tz.thermostat_control_sequence_of_operation],
        exposes: [
                    exposes.climate().withSetpoint('occupied_heating_setpoint', 4, 30, 0.5)
                                     .withLocalTemperature()
                                     .withSystemMode(['off', 'auto', 'heat'])
                                     .withRunningState(['idle', 'heat'])
                                     .withPiHeatingDemand(),
                    e.temperature(),
                    e.humidity(),
                ],
        meta: {configureKey: 1, battery: {dontDividePercentage: true}},
            const endpoint1 = device.getEndpoint(1);
            
            await reporting.bind(endpoint1, coordinatorEndpoint, ['genPowerCfg', 'msTemperatureMeasurement', 'msRelativeHumidity']);
            await reporting.temperature(endpoint1);
            await reporting.humidity(endpoint1);
            await reporting.batteryPercentageRemaining(endpoint1);
        },
            if (type === 'message' && data.type === 'read' && data.cluster === 'hvacThermostat') {
                const rsp = {
                    pIHeatingDemand: 20,
                    occupiedHeatingSetpoint: 2000,
                }
                await data.endpoint.readResponse(
                    'hvacThermostat', rsp, {}, data.meta.zclTransactionSequenceNumber,
                );
            }
        },`
However read response is never sent to the device. I'm not sur Herdsman sees the read request form the thermostat (I have no debug log listing the requests).

Koenkk:
1. I think it's fine to do this.
2. The bind table full is an error generated by the device. We cannot fix this from z2m.
3. That code should be correct, try to catch such read with the herdsman debug logging enabled.
See https://www.zigbee2mqtt.io/information/debug.html#zigbee-herdsman-debug-logging on how to enable the herdsman debug logging. Note that this is only logged to STDOUT and not to log files.

H3buss:
3. I tried debugging using 
`DEBUG=zigbee-herdsman:controller:* npm start`
I do get logs, but I don't see those reads. Its like the firmware is not forwarding them. (I'm using latest Zstack 3.0x for CC2531. I tried with ZStack Home 1.2 but got the same results

Koenkk:
Can you provide the sniff of that?

H3buss:
Here it is:
[CCTFR6400_Read_sniff.zip](https://github.com/Koenkk/zigbee-herdsman-converters/files/6445921/CCTFR6400_Read_sniff.zip)

Koenkk:
It may work after changing `https://github.com/Koenkk/zigbee-herdsman/blob/be70088a148f9cb7fea9dcc2d81fd4851d0dc8f3/src/adapter/z-stack/adapter/startZnp.ts#L38` to `{...EndpointDefaults, endpoint: 3, appprofid: 0x0104, appnuminclusters: 1, appinclusterlist: [Zcl.Utils.getCluster('hvacThermostat').ID]},` . But what if you just send a readReponse to the device occasionally, does that also work?

H3buss:
I still don't get the read request with the change to startZnp you provided.

Il tried to send the read response, but ils fails with "Write for 'BUFFER' not available".

H3buss:
I tried sending multiple read response on attribute reports from the device, no success so far.

H3buss:
Sorry, the suggested modification to startZnp does work, I needed to reset the coordinator so it could be reconfigured.

Issue #734, opened 2019-11-05
"local_temperature_calibration" new value change not updated SPZB0001 thermostat


vansoest:
The **local_temperature_calibration** attribute output is not updated to the new value, is always 0.

I set the **local_temperature_calibration** by
```zigbee2mqtt/thermostat/living_room/set {"local_temperature_calibration":1.4}``` and yes it changes the local temperature from 19.77°C to 21.17°C, but the update topic will not set my new offset value in the attributes json. 

Received topic:
```
zigbee2mqtt/thermostat/living_room 
{
  "local_temperature": 21.17,
  "linkquality": 94,
  "current_heating_setpoint": 21.5,
  "eurotronic_system_mode": 1,
  "system_mode": "heat",
  "occupied_heating_setpoint": 21,
  "unoccupied_heating_setpoint": 16,
  "eurotronic_error_status": 0,
  "pi_heating_demand": 32,
  "battery": 100,
  "local_temperature_calibration": 0
}
```

Eurotronic Spirit SPZB0001 thermostat

**Comment by Koenkk** [comment]
@mhelff @allofmex does this work for you?

**Comment by sti0** [comment]
You receive the actual calibration by publishing this payload
```
{
  "local_temperature_calibration": ""
}
```
to topic
`zigbee2mqtt/[DEVICE_ID]/get`

allofmex:
> @allofmex does this work for you?
Using ioBroker, not applicable there. Can't help here, sorry.

sti0:
This is still an issue. Even mentioned in https://github.com/Koenkk/zigbee-herdsman-converters/issues/788

sjorge:
No, I played with it a bit but it didn't seem to do anything (setting local_temperature_calibration)

sti0:
After some minutes the local_temperature should be increased / decreased by the value. Could you try to get the local_temperature?

sjorge:
I'm not seeing any change in local_temperatur or local_temperature_calibration afterwards, the later always returning 0

sti0:
did you publish 

`{"local_temperature_calibration":""}`

to

`zigbee2mqtt/DEVICE/get`

?

sti0:
did you publish 

`{"local_temperature_calibration":""}`

to

`zigbee2mqtt/DEVICE/get`

?

sjorge:
- publish a `{"local_temperature_calibration":"1"}` to `zigbee2mqtt/DEVICE/set`
[...]

PR #2556, opened 2021-05-07
Added support for Philips Hue Iris (Generation 4) Silver

gusi1994:
As requested by @thomasvoni in #734 I added the same data for the silver philips hue iris limited edition like for the gold version.

@thomasvoni Maybe you could test this before.


Koenkk:
Thanks!

PR #2558, opened 2021-05-08
 @H3buss Added measured temperature SET for CCTFR6700 

H3buss:
Requires report capability from herdsman


converters/toZigbee.js
```diff
@@ -4849,6 +4849,63 @@ const converters = {
...
+        },
+    },
+    schneider_thermostat_system_mode: {
+        key: ['system_mode'],
+        convertSet: async (entity, key, value, meta) => {
+            let systemMode = utils.getKey(constants.thermostatSystemModes, value, undefined, Number);
+            if (systemMode === undefined) {
+                systemMode = utils.getKey(legacy.thermostatSystemModes, value, value, Number);
+            }
+            globalStore.putValue(entity, 'systemMode', systemMode);
```
Koenkk, line 4865: Is the only reason for this to buffer the commands until the device awakes?

H3buss: These attributes are read by the device, they cannot be written.

Koenkk: What if we just send a readReponse here? (without the device requested for one). Does that also work?

H3buss: When the device is asleep, it doesn't acknowledge any frame.

Also, spamming readResponses to the device does not seem to work either, even when awake. The device seems to correlate the responses to its requests somehow (with sequence number maybe?)

Koenkk: I've updated herdsman to allow read response to all attributes. Can you check if it works?  (update to the latest dev branch first). After that:

- Remove the read response code from the `onEvent` handler
- Replace `globalStore.putValue(entity, 'systemMode', systemMode);` with `endpoint.saveClusterAttributeKeyValue('hvacThermostat', {systemMode: systemMode})`

This has the benefit that these attributes are persisted when z2m is restarted.

H3buss: 1) Do I still need the hacked firmware/modified startZnp of herdsman for this to work?

2) I'm not sure how to update herdsman to the latest dev version within my zigbee2mqtt npm environment.

Koenkk: 1. Firmware yes, modified startZnp no
2. Just follow https://www.zigbee2mqtt.io/how_tos/how-to-switch-to-dev-branch.html#bare-metal again

H3buss: Tested and it works, I've updated the code accordingly.
I still need to use globalStore for pending writes to keypad_lockout and for handling of schneider_ui_action where I need to know if screen is awake or not.

converters/fromZigbee.js
```diff
@@ -5556,6 +5556,53 @@ const converters = {
...
+                }
+            }
+        }
+    },
+    schneider_temperature: {
+        cluster: 'msTemperatureMeasurement',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const temperature = parseFloat(msg.data['measuredValue']) / 100.0;
+            const property = postfixWithEndpointName('temperature', msg, model);
```
Koenkk, line 5600: Why do we have a `temperature` when there is already a `local_temperature`? (we also do not do this for other thermostats)

converters/fromZigbee.js
```diff
@@ -5556,6 +5556,53 @@ const converters = {
...
+                screenAwake = screenAwake != undefined ? screenAwake : false;
+                let keypadLocked = msg.endpoint.getClusterAttributeValue('hvacUserInterfaceCfg', 'keypadLockout');
+                keypadLocked = keypadLocked != undefined ? keypadLocked != 0 : false;
+
+                // Emulate UI temperature update
+                if (data[1] === 'ScreenWake') {
+                    globalStore.putValue(msg.endpoint, 'screenAwake', true);
+                } else if (data[1] === 'ScreenSleep') {
+                    globalStore.putValue(msg.endpoint, 'screenAwake', false);
+                } else if (screenAwake && !keypadLocked) {
```
Koenkk, line 5579: Note that the `screenAwake` is not reliable (because when z2m restarts this value resets). I'm wondering why this is needed, is this to prevent the occupiedHeatingSetpoint from being decreased when the screen is sleeping?

converters/toZigbee.js
```diff
@@ -4849,6 +4849,63 @@ const converters = {
             await entity.read('schneiderSpecificPilotMode', ['pilotMode'], {manufacturerCode: 0x105e});
         },
     },
+    schneider_temperature_measured_value: {
+        key: ['temperature_measured_value'],
```
Koenkk, line 4853: This is the temperature that the thermostat will try to achieve right? I think `target_temperature` is  a better name.

converters/toZigbee.js
```diff
@@ -4849,6 +4849,63 @@ const converters = {
...
+        convertSet: async (entity, key, value, meta) => {
+            await entity.report('msTemperatureMeasurement', {'measuredValue':Math.round(value*100)});
+        },
+    },
+    schneider_thermostat_system_mode: {
+        key: ['system_mode'],
+        convertSet: async (entity, key, value, meta) => {
+            let systemMode = utils.getKey(constants.thermostatSystemModes, value, undefined, Number);
+            if (systemMode === undefined) {
+                systemMode = utils.getKey(legacy.thermostatSystemModes, value, value, Number);
```
Koenkk, line 4863: For new integration we don't support legacy, so please remove this line.

converters/toZigbee.js
H3buss: This is the measured temperature, provided by an external sensor. 
The target temp is occupied_heating_setpoint

converters/fromZigbee.js
H3buss: The temperature appear as a sensor on home assistant, which helps automation. But I concur that this is not really needed: the temperature could be reported on local temperature only.

converters/fromZigbee.js
H3buss: When the device is awoken by button press, the button pressed event is sent by the device but it should not change the setpoint.
Note: The device computes and  displays the updated setpoint locally, then reads the actual setpoint back from the coordinator after a few seconds. For good user experience, the behavior of the device should be matched as closely as possible.

converters/toZigbee.js
H3buss: Fixed

converters/fromZigbee.js
H3buss: Fixed

converters/toZigbee.js
```diff
@@ -4849,6 +4849,60 @@ const converters = {
...
+    schneider_thermostat_keypad_lockout: {
+        key: ['keypad_lockout'],
+        convertSet: async (entity, key, value, meta) => {
+            const keypadLockout = utils.getKey(constants.keypadLockoutMode, value, value, Number);
+
+            try {
+                await entity.write('hvacUserInterfaceCfg', {keypadLockout}, {timeout: 500});
+            } catch(e) {
+                // Will probably fail if device is not awake
+                globalStore.putValue(entity, 'pending_writes', {cluster: 'hvacUserInterfaceCfg', data: keypadLockout.keypadLockout},);
```
Koenkk, line 4901: Instead of the `pending_writes` thing I've added a new option: `sendWhenActive`. This will send the message when the coordinator receives a message from it. Can you test it? You can remove this whole try catch block and replace it with: `await entity.write('hvacUserInterfaceCfg', {keypadLockout}, {timeout: 500, sendWhenActive: true});` 

Changes will be available in the dev branch in a few hours from now. (https://www.zigbee2mqtt.io/how_tos/how-to-switch-to-dev-branch.html)

H3buss: I tested it and something seems not right: on the device the "lock" indication is spuriously toggling between on and off.
Also, the readback of "saved cluster attributes" seems unstable.

Do you have any idea of what could be wrong? 
I'll try to investigate a bit deeper tomorow.

Koenkk: You can look at the herdsman debug log to see what is happening.

See https://www.zigbee2mqtt.io/information/debug.html#zigbee-herdsman-debug-logging on how to enable the herdsman debug logging. Note that this is only logged to STDOUT and not to log files.

H3buss: Sorry I did not comment back, but this is working fine after the pull request on herdsman you just closed.

converters/fromZigbee.js
```diff
@@ -5556,6 +5556,51 @@ const converters = {
...
     },
+    schneider_ui_action: {
+        cluster: 'wiserDeviceInfo',
+        type: 'attributeReport',
+        convert: (model, msg, publish, options, meta) => {
+            if (hasAlreadyProcessedMessage(msg)) return;
+
+            const data = msg.data['deviceInfo'].split(',');
+            if(data[0] === 'UI' && data[1]) {
+                publish({action: data[1]});
```
Koenkk, line 5567: instead of calling publish here, do `const result = {action: data[1]};` and replace `publish({occupied_heating_setpoint: occupiedHeatingSetpoint/100});` with:

```js
result.occupied_heating_setpoint = occupiedHeatingSetpoint/100;
return result;
```

converters/toZigbee.js
```diff
@@ -4849,6 +4849,56 @@ const converters = {
...
+        convertSet: async (entity, key, value, meta) => {
+            const occupiedHeatingSetpoint = (Math.round((value * 2).toFixed(1)) / 2).toFixed(1) * 100;
+            entity.saveClusterAttributeKeyValue('hvacThermostat', {occupiedHeatingSetpoint: occupiedHeatingSetpoint});
+            return {state: {occupied_heating_setpoint: value}};
+        },
+    },
+    schneider_thermostat_control_sequence_of_operation: {
+        key: ['control_sequence_of_operation'],
+        convertSet: async (entity, key, value, meta) => {
+            let val = utils.getKey(constants.thermostatControlSequenceOfOperations, value, undefined, Number);
```
Koenkk, line 4877: these 3 lines can be simplified to:

```
const val = utils.getKey(constants.thermostatControlSequenceOfOperations, value, value, Number);
```

lib/store.js
```diff
@@ -44,6 +44,30 @@ function clearValue(entity, key) {
...
+function pushValue(entity, key, value) {
+    const entityKey = getEntityKey(entity);
+    if (!store.has(entityKey)) {
+        store.set(entityKey, new Array());
+    }
+
+    store.get(entityKey)[key].push(value);
+}
+
+function popValue(entity, key) {
```
Koenkk, line 56: These changes can be reverted now.

converters/fromZigbee.js
Koenkk: The temperature can also be retrieved from the thermostat, it's better not to have duplicated data.

converters/fromZigbee.js
H3buss: Fixed

lib/store.js
H3buss: Reverted

converters/fromZigbee.js
H3buss: I removed the temperature key and expose.

Koenkk:
Did some small updates, can you check if everything is OK? Action should be in snake case now instead of camel case. If ok this can be merged.

H3buss:
Tested, and it works.

Will you also publish the modified firmware?

converters/toZigbee.js
H3buss: I'm sorry to come back on that issue, but I installed my production environnement and the the thermostat no longer take the readResponses into account. I think I messed-up when I tested with the hacked firmware a while ago.

After some troubleshooting, the readResponses are sent from coordinator endpoint 3, and the sent attributes are correct, however the profile is incorrect: should be 0x0104 for Home Automation, and is 0x0105.

How can I set the profile for the read responses?

Mega-Touk:
Hello @H3buss 
Firstly thanks for your work (& Koenkk too)

Do you have any idea why energy is never reported by the CCTFR6700 ?
I always get payload that look like this: `'{"energy":null,"......'`
I also have some "No converter available" messages, maybe this is related ?
```
debug No converter available for 'CCTFR6400' with cluster 'haDiagnostic' and type 'attributeReport' and data '{"lastMessageLqi":216,"lastMessageRssi":-46}'
debug No converter available for 'CCTFR6400' with cluster 'haDiagnostic' and type 'attributeReport' and data '{"57345":724}'
```


Also the documentation is mentioning a `bind from a sensor,` [here ](https://www.zigbee2mqtt.io/devices/CCTFR6700.html#pairing-and-configuring) i don't really get how to implement this. I'm using an automation for now but it feel a bit heavy. Could we make the documentation easier to understand and apply ?

Sorry if I'm not using the right place to discuss about this but I'm not sure if i should open a new issue or not.

H3buss:
I don't know why the device do not report the energy. I'm pretty sure I got it working at some point during my tests but now it doesn't for some reason. I tried to bind the attribute manually  again, to no avail. The debug message you see are from haDiagnostic cluster and I don't see how they could be related to this issue.
On HA I use integration of power to compute energy as a fallback. This is not pretty but works.

As all my devices are now used in my production environment, I cannot work on this without braking part on my home automation. Not great for the WAF ;).

As for the reporting of ambiant temperature from a sensor, I agree that the documentation is unclear. The important missing part is the topic you need to publish the temperature to: `temperature_measured_value`. 
In the documentation, I should remove the "bind" part. I never tested it and doubt it would work. Using an automation is the way to get it working.
On HA, I use a custom component to create a virtual `Climate` entity that links one CCTFR6400 to one or more CCTFR6700. It performs the measure temperature synchonisation from the CCTFR6400 to the CCTFR6700, but also sends the `pi_heating_demand` back to display the heating state. 
My code is realy not pretty but I can share it if necessary.

PR #2559, opened 2021-05-08
Added support for Samotech SM323 retrofit dimmer

avee87:
Adding SM323 listed here: https://www.samotech.co.uk/products/zigbee-dimmer-switch-hue-compatible/
This one is very similar to SM309 so pretty much copypasted the config and changed model name.

Tested on my setup that device is identified and can be controlled properly.

### DIFF
--- a/devices/samotech.js
+++ b/devices/samotech.js
@@ -37,6 +37,7 @@ module.exports = [
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl']);
             await reporting.onOff(endpoint);
+            await reporting.brightness(endpoint);
         },
     },
 ];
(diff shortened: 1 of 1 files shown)


avee87:
Huh, looks like this one actually conflicts with Ecodim dimmer because it has the same model name: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/ecodim.js#L8-L22
Is there a way to disambiguate them? And do I even need to bother as that thing just seems to work despite being a different device?

Koenkk:
Probably this device is just a whitelabel of it, does it identify correctly with the `fingerprint` in place?

Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/744

Koenkk:
Fingerprint indeed has priority over zigbeeModel so this will work.

PR #2564, opened 2021-05-10
Fix EcoDim ED-10011

bartrikers6224:
The EcoDim ED-10011 actually sends the actions in the payload including the postfix "_1".

image


Koenkk:
@bartrikers6224 I don't know how the domitcz integration works, but if both have the same endpoint I propose to revert this change and remove `multiEndpoint: true` from this device.

bartrikers6224:
Ok, I've reverted this change and removed multiEndpoint:true and created a new [pull-request](https://github.com/Koenkk/zigbee-herdsman-converters/pull/2582)

Issue #2557, opened 2021-05-08
Add support for Tuya Smart Zigbee Metering Plug 16A Eu 


oetelaar24:
Hi,

I've recently bought a couple of these at ali: https://nl.aliexpress.com/item/1005002344798281.html

Apparently they are quite new, so support is not yet added. It pairs successfully and reports itself as:
`Zigbee model 'TS011F' and manufacturer name '_TZ3000_cphmq0q7'`

I'm not really sure how to go about adding support for this device, as I do not have a Tuya gateway to be able to sniff the traffic. I'm also kinda hoping it's not needed and we can just try and "copy" the converter-settings from a known device like the TS0121_plug or TS0601_din. 

I've tried using the known devices definition in an external converters file (TZ3000_cphmq0q7.js), however, only the TS0121_plug worked a little bit in that I was able to control the switch, but it also errors and turns it off:
`Zigbee2MQTT:error 2021-05-08 12:24:49: Failed to call 'OnEvent' 'onZigbeeEvent' (ReferenceError: globalStore is not defined
,    at Object.onEvent (/app/lib/util/externally-loaded.js:36:20)
,    at OnEvent.callOnEvent (/app/lib/extension/onEvent.js:33:39)
,    at OnEvent.onZigbeeEvent (/app/lib/extension/onEvent.js:17:18)
,    at Controller.callExtensionMethod (/app/lib/controller.js:384:44))
`
Can anyone help me with the best next step to take?

Kind regards,
-Bart

oetelaar24:
Okay, I've given it another shot. But still not getting much further...

Tried with this external converter to get tuya data dumps:

`const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
const tz = require('zigbee-herdsman-converters/converters/toZigbee');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const reporting = require('zigbee-herdsman-converters/lib/reporting');
const extend = require('zigbee-herdsman-converters/lib/extend');
const e = exposes.presets;
const ea = exposes.access;

const definition = {
    // Since a lot of Tuya devices use the same modelID, but use different data points
    // it's usually necessary to provide a fingerprint instead of a zigbeeModel
    fingerprint: [
        {
            // The model ID from: Device with modelID 'TS0601' is not supported
            // You may need to add \u0000 at the end of the name in some cases
            modelID: 'TS011F',
[...]

PR #2571, opened 2021-05-14
Add support for TS011F (_TZ3000_cphmq0q7) smart plug

yratanov:
Inspired by TS0121_plug, with slight change.

https://github.com/Koenkk/zigbee2mqtt/issues/7463
https://github.com/Koenkk/zigbee2mqtt/issues/7405
https://github.com/Koenkk/zigbee-herdsman-converters/issues/2557

image
image

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -627,6 +627,45 @@ module.exports = [
         },
     },
     {
+        fingerprint: [{modelID: 'TS011F', manufacturerName: '_TZ3000_cphmq0q7'},],
+        model: 'TS011F',
+        description: 'Smart EU Plug',
+        vendor: 'TuYa',
+        fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fz.ignore_basic_report, fz.tuya_switch_power_outage_memory],
+        toZigbee: [tz.on_off, tz.tuya_switch_power_outage_memory],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement']);
+            endpoint.saveClusterAttributeKeyValue('seMetering', {divisor: 100, multiplier: 1});
+            endpoint.saveClusterAttributeKeyValue('haElectricalMeasurement', {
+                acVoltageMultiplier: 1, acVoltageDivisor: 1, acCurrentMultiplier: 1, acCurrentDivisor: 1000, acPowerMultiplier: 1,
+                acPowerDivisor: 1,
+            });
+        },
+        // This device doesn't support reporting correctly.
+        // https://github.com/Koenkk/zigbee-herdsman-converters/pull/1270
+        exposes: [e.switch(), e.power(), e.current(), e.voltage().withAccess(ea.STATE),
+            e.energy(), exposes.enum('power_outage_memory', ea.STATE_SET, ['on', 'off', 'restore'])
+                .withDescription('Recover state after power outage')],
+        onEvent: (type, data, device, options) => {
+            const endpoint = device.getEndpoint(1);
+            
+            if (type === 'stop') {
+                clearInterval(globalStore.getValue(device, 'interval'));
+                globalStore.clearValue(device, 'interval');
+            } else if (!globalStore.hasValue(device, 'interval')) {
+                const seconds = options && options.measurement_poll_interval ? options.measurement_poll_interval : 60;
+                const interval = setInterval(async () => {
+                    try {
+                        await endpoint.read('haElectricalMeasurement', ['rmsVoltage', 'rmsCurrent', 'activePower']);
+                        await endpoint.read('seMetering', ['currentSummDelivered']);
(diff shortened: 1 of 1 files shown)


oetelaar24:
Hi, @yratanov ,

Thanks for this, you also mentioned my thread. As soon as I saw your post, I merged your change to my local tuya.js file.

However I seem to be getting an error when it tries to configure the plug. Otherwise, it does seem to work well! Which is great!

The error is:
```
Zigbee2MQTT:error 2021-05-14 06:02:13: Failed to configure '0x847127fffe8d9579', attempt 3 (TypeError: Cannot read property 'configureKey' of undefined
,    at Configure.configure (/app/lib/extension/configure.js:138:69))
```

yratanov:
@oetelaar24 yep, I've seen this error, and I added 
```js
    meta: {
        configureKey: 1,
    },
```
locally, but it seems it's an outdated method, so I'm not sure how to fix it properly.

oetelaar24:
Ah yes, I see this method being used by others as well. It does resolve the error. 
I see it mentioned here a while back: https://github.com/Koenkk/zigbee2mqtt/issues/1758

devices/tuya.js
```diff
@@ -626,6 +626,45 @@ module.exports = [
...
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement']);
+            endpoint.saveClusterAttributeKeyValue('seMetering', {divisor: 100, multiplier: 1});
+            endpoint.saveClusterAttributeKeyValue('haElectricalMeasurement', {
+                acVoltageMultiplier: 1, acVoltageDivisor: 1, acCurrentMultiplier: 1, acCurrentDivisor: 1000, acPowerMultiplier: 1,
+                acPowerDivisor: 1,
+            });
+        },
+        // This device doesn't support reporting correctly.
+        // https://github.com/Koenkk/zigbee-herdsman-converters/pull/1270
```
Koenkk, line 646: Are you sure this device does not support reporting? Can you check if power reporting also works when:
- Removing the `onEvent`
- Replacing configure with:

```js
    await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement']);
    await reporting.rmsVoltage(endpoint);
    await reporting.rmsCurrent(endpoint);
},
```

devices/tuya.js
```diff
@@ -626,6 +626,45 @@ module.exports = [
             }
         },
     },
+    {
+        fingerprint: [{modelID: 'TS011F', manufacturerName: '_TZ3000_cphmq0q7'},],
+        model: 'TS011F',
+        description: 'Smart EU Plug',
+        vendor: 'TuYa',
+        fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fz.ignore_basic_report, fz.tuya_switch_power_outage_memory],
```
Koenkk, line 634: Do you get an updated `state` when controlling the plug via the button on it?

Koenkk:
> @oetelaar24 yep, I've seen this error, and I added
> 
> ```js
>     meta: {
>         configureKey: 1,
>     },
> ```
> 
> locally, but it seems it's an outdated method, so I'm not sure how to fix it properly.

Configure key isn't needed anymore since Z2M 1.18.3**-dev**

devices/tuya.js
yratanov: Yes

devices/tuya.js
yratanov: Thank you! It does work like that. I've pushed the changes

yratanov:
@Koenkk 
> Configure key isn't needed anymore since Z2M 1.18.3-dev (it is in 1.18.3)

It's weird, because I have 1.18.3-1.  But I'm adding this code as external converter, as described [here](https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html), may be it's somehow different.

Koenkk:
@yratanov `1.18.3-1` != `1.18.3-dev`, for dev use the edge version of the addon.

devices/tuya.js
Koenkk: perfect

devices/tuya.js
yratanov: I was wrong. It doesn't update. I'll revert and create another PR

McGiverGim:
Only as information, I've this device, seems to work, but the image is totally different:
https://www.aliexpress.com/item/1005002495230651.html
image

I suppose it not matters...

Koenkk:
@McGiverGim I received exactly this one indeed. Picture is currently not correct but will be fixed with the next release.

oetelaar24:
It soes seem to work well. However, and I'm unsure why it's happening, some of the plugs have started to turn off at a seemingly random moment.. Which is kindoff a bummer when the device behind it needs to be on...
It could be the plug is faulty ofcourse, but I'm not sure how I can see that...

jusitnhu14:
Hi was wondering if someone can point me to the external converter MQTT for Zigbee Model
TS011F
Zigbee Manufacturer
_TZ3000_myruaaxg
Its a TUYA 1 gang plug and USB charging port

Thanks

PR #2572, opened 2021-05-14
Added support for Innr RC110 Remote Control

danieledwardgeorgehitchcock:
Added support for Innr RC110 Remote Control

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5777,6 +5777,14 @@ const converters = {
             return result;
         },
     },
+    rc_110_level_to_scene: {
+        cluster: 'genLevelCtrl',
+        type: ['commandMoveToLevel', 'commandMoveToLevelWithOnOff'],
+        convert: (model, msg, publish, options, meta) => {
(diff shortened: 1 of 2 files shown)


danieledwardgeorgehitchcock:
Not sure why the check is failing? This is working locally...

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/750

converters/fromZigbee.js
```diff
@@ -5777,6 +5777,14 @@ const converters = {
             return result;
         },
     },
+    rc_110_level_to_scene: {
```
Koenkk, line 5780: Why did you choose to add a custom converter instead of using `fz.command_move_to_level`?

danieledwardgeorgehitchcock: The scenes buttons are reported as specific values and transition times which didn’t really make much sense from a user perspective (a pressed button will always be this value) as the buttons on the device are labelled as scenes, the returned brightness level isn’t really useful

Koenkk: thanks!

danieledwardgeorgehitchcock:
Resolves issues with #2255

PR #2573, opened 2021-05-14
Added support for Philips Hue White ambiance Milliskin (Aluminum)

TomSabel:
Added support for Philips Hue White ambiance Milliskin (Aluminum)


Koenkk:
Thanks!

PR #2574, opened 2021-05-14
The first version of GL-S-004P had GL-S-005P

alekslyse:
Simple change. The GL-S-005P has changed name to GL-S-004P - GL-S-004P was in the database. same light, same socket, same wattage, same model (pro) - just a different model id, same bulb


Koenkk:
Thanks!

PR #2576, opened 2021-05-15
Add fingerprint for Tuya Smart Zigbee Plug 16A EU

ferbar:
As far as I can see zigbeeModel and fingerprint are OR-ed, so this should not break other TS0121. I think there is no need to copy&paste the whole block. Switch / voltage / current / power are working.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -590,6 +590,7 @@ module.exports = [
     },
     {
         zigbeeModel: ['TS0121'],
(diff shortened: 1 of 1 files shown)


ferbar:
Is there a way to add a second image to the documentation? (https://www.zigbee2mqtt.io/devices/TS0121_plug.html)

Thanks!!!

Koenkk:
Let's continue this one in https://github.com/Koenkk/zigbee-herdsman-converters/pull/2571

PR #2577, opened 2021-05-15
Updated reporting defaults for Develco devices

lnagel:
I've taken the default settings from Develco technical documentation and just specified them in the default reporting config here.

This helps device setup because I don't have to manually update all the values every time the device is rejoined.


Koenkk:
Thanks!

PR #2578, opened 2021-05-15
Add Woox R9077 RGB bulb

ofalvai:
Add support for the Woox R9077 RGB bulb, which is a TuYa TS0505A with small changes I had to make:

- add `applyRedFix` because the Home Assistant color wheel wasn't working properly
- remove the color temp startup feature because it seems to be unsupported (I got an error when I tried changing the setting via the UI)

I still get an error though on startup about an unsupported attribute:

```
Failed to setup reporting for '0x842e14fffe0e3604' - Error: Read 0x842e14fffe0e3604/1 lightingColorCtrl(["colorCapabilities"], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')
```

Is this expected? Can I remove this error somehow?


ofalvai:
I see that the tests are failing because of the missing color temp range. I followed the instructions to get the range, but this seems to be unsupported:

```
Publish 'set' 'read' to 'balcony' failed: 'Error: Read 0x842e14fffe0e3604/1 lightingColorCtrl(["colorTempPhysicalMin","colorTempPhysicalMax"], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')'
```

Should I add the model to the allowlist?

Koenkk:
> I still get an error though on startup about an unsupported attribute:

This is because of the reporting feature being enabled which is deprecated (see second point of https://github.com/Koenkk/zigbee2mqtt/releases/tag/1.17.1 how to migrate0

> Should I add the model to the allowlist?

Yes, will do it.

devices/tuya.js
```diff
@@ -358,6 +358,14 @@ module.exports = [
         fromZigbee: [fz.on_off, fz.tuya_led_controller, fz.brightness, fz.ignore_basic_report],
         exposes: [e.light_brightness_colortemp_colorhs().removeFeature('color_temp_startup')],
     },
+    {
+        fingerprint: [{modelID: 'TS0505A', manufacturerName: '_TZ3000_keabpigv'}],
+        model: 'R9077',
+        vendor: 'Woox',
```
Koenkk, line 364: Since this is a different vendor, please create a new file `woox.js` instead of adding it to `tuya.js`.

ofalvai: Done. I've also moved the other Woox device from `tuya.js` to the new file.

Koenkk: Thanks!

PR #2582, opened 2021-05-16
Fix EcoDim ED-10011 #2

bartrikers6224:
Revert previous fix and removed multiEndpoint:true for EcoDim ED-10011 (as both buttons on the device have the same endpoint)

### DIFF
--- a/devices/ecodim.js
+++ b/devices/ecodim.js
@@ -36,7 +36,7 @@ module.exports = [
         vendor: 'EcoDim',
         description: 'Zigbee 2 button wall switch - black',
         fromZigbee: [fz.command_on, fz.command_off, fz.command_move, fz.command_stop, fz.battery],
-        exposes: [e.battery(), e.action(['on', 'off', 'brightness_move_up', 'brightness_move_down', 'brightness_stop'])],
+        exposes: [e.battery(), e.action(['on_1', 'off_1', 'brightness_move_up_1', 'brightness_move_down_1', 'brightness_stop_1'])],
         toZigbee: [],
         meta: {multiEndpoint: true},
     },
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

frlp542:
Hello Guys

I see you changed the behaviour of this device but after updating to this release my ED-10011 don't respond.
the reason : I have ED-10011's with 4 buttons. (perhaps early batches) therefore the multi endpoint is a requirement for my switches in order to see each button individually.

bartrikers6224:
Are you sure they are ED-10011's? The ED-10011 only has 2 buttons (https://www.ecodim.nl/ed-10011.html), the ED-10013/ED-10012 seem to have 4 buttons (https://www.ecodim.nl/ed-10013.html)

Op 6 jun. 2021 09:50 schreef frlp542 ***@***.***>:

Hello Guys

I see you changed the behaviour of this device but after updating to this release my ED-10011 don't respond.
the reason : I have ED-10011's with 4 buttons. (perhaps early batches) therefore the multi endpoint is a requirement for my switches in order to see each button individually.

frlp542:
200% sure. I've got 3 ED1011 that are 4 buttons 
even more... the 2 ED10012 I have are a 2 buttons

I have them since last year and made the initial device integration in zigbee2mqtt :)
but this doesn't mean a thing because the manufacturar could have changed something too.
what I can tell you is that  I think I have a a version 1. if you check the serial number above the part number   

IMG_3048

bartrikers6224:
That's interesting... The serial nr on my 10011 also ends with 'V1', meaning it's also a version 1 I think?


Op 6 jun. 2021 14:38 schreef frlp542 ***@***.***>:

200% sure. I've got 3 ED1011 that are 4 buttons
even more... the 2 ED10012 I have are a 2 buttons

I have them since last year and made the initial device integration in zigbee2mqtt :)
but this doesn't mean a thing because the manufacturar could have changed something too.
what I can tell you is that I think I have a a version 1. if you check the serial number above the part number

[IMG_3048]<https://user-images.githubusercontent.com/62246396/120924588-b8c1b100-c6d4-11eb-8d58-67d5f5401dd2.jpg>

frlp542:
indeed could be the same. 
This means we have got 2 different hardware specs with the same hardware identifier. :(

PR #2589, opened 2021-05-17
Wiser Smart Europe Device Support

pklokke:
- Added Room and Radiator Thermostats, Heating and Load Actuators and 2nd Gen Smartplug
- All devices excluding Smartplug require holding set button for 5 seconds to join
- Smartplug will join with short press, but only on channels 11,15,20,25
- Relies on commands added to PR Koenkk/zigbee-herdsman#363

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5565,30 +5565,13 @@ const converters = {
         type: 'read',
         convert: async (model, msg, publish, options, meta) => {
             const response = {};
-            let setpoint = 20*100;
-            if (meta.state.hasOwnProperty('occupied_heating_setpoint')) {
-                setpoint = (Math.round((meta.state.occupied_heating_setpoint * 2).toFixed(1)) / 2).toFixed(1) * 100;
-            }
-
-            if (msg.data[0] == 'minHeatSetpointLimit') {
-                response['minHeatSetpointLimit'] = 7*100;
-            } else if (msg.data[0] == 'maxHeatSetpointLimit') {
-                response['maxHeatSetpointLimit'] = 30*100;
-            } else if (msg.data[0] == 'occupiedHeatingSetpoint') {
-                response['occupiedHeatingSetpoint'] = setpoint;
-            } else if (msg.data[0] == 'systemMode') {
-                response['systemMode'] = 4;
-            } else if (msg.data[0] == 0xe010) {
+            if (msg.data[0] == 0xe010) {
                 // Zone Mode
                 const lookup = {'manual': 1, 'schedule': 2, 'energy_saver': 3, 'holiday': 6};
                 const zonemodeNum = lookup[meta.state.zone_mode];
                 response[0xe010] = {value: zonemodeNum, type: 0x30};
-            } else {
-                meta.logger.warn(`'${meta.device.ieeeAddr}' read req from unsupported attribute from hvacThermostat '${msg.data[0]}'`);
+                await msg.endpoint.readResponse(msg.cluster, msg.meta.zclTransactionSequenceNumber, response, {srcEndpoint: 11});
             }
-
-            await msg.endpoint.readResponse(msg.cluster, msg.meta.zclTransactionSequenceNumber, response, {srcEndpoint: 11});
-
             converters.wiser_smart_vact_update_params.convert(model, msg, publish, options, meta);
         },
     },
@@ -5677,7 +5660,8 @@ const converters = {
             const attribute = {};
             const result = {};
 
-            // the UI client on the thermostat also updates the server, so no need to readback/send again on next sync
+            // The UI client on the thermostat also updates the server, so no need to readback/send again on next sync.
+            // This also ensures the next client read of setpoint is in sync with the latest commanded value.
             attribute['occupiedHeatingSetpoint'] = msg.data['setpoint'];
             msg.endpoint.saveClusterAttributeKeyValue('hvacThermostat', attribute);
 
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5578,16 +5578,11 @@ const converters = {
                 response['occupiedHeatingSetpoint'] = setpoint;
             } else if (msg.data[0] == 'systemMode') {
                 response['systemMode'] = 4;
-            } else if (msg.data[0] == 'wiserSmartPriorityLevel') {
-                response['wiserSmartPriorityLevel'] = 0xff;
-            } else if (msg.data[0] == 'wiserSmartMasterShortAddress') {
-                response['wiserSmartMasterShortAddress'] = 0x0000;
-            } else if (msg.data[0] == 'wiserSmartZoneMode') {
+            } else if (msg.data[0] == 0xe010) {
+                // Zone Mode
                 const lookup = {'manual': 1, 'schedule': 2, 'energy_saver': 3, 'holiday': 6};
                 const zonemodeNum = lookup[meta.state.zone_mode];
-                response['wiserSmartZoneMode'] = zonemodeNum;
-            } else if (msg.data[0] == 'wiserSmartHactConfig') {
-                response['wiserSmartHactConfig'] = 0x80;
+                response[0xe010] = {value: zonemodeNum, type: 0x30};
             } else {
                 meta.logger.warn(`'${meta.device.ieeeAddr}' read req from unsupported attribute from hvacThermostat '${msg.data[0]}'`);
             }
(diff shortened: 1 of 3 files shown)


pklokke:
This now only requires the commands added to the zigbee herdsman PR, the attributes are all handled numerically in the to/from zigbee converters

converters/toZigbee.js
```diff
@@ -4982,6 +4982,38 @@ const converters = {
...
+    wiser_sed_occupied_heating_setpoint: {
+        key: ['occupied_heating_setpoint'],
+        convertSet: async (entity, key, value, meta) => {
+            return {state: {'occupied_heating_setpoint': value}};
+        },
+    },
+    wiser_sed_thermostat_local_temperature_calibration: {
+        key: ['local_temperature_calibration'],
+        convertSet: async (entity, key, value, meta) => {
+            globalStore.putValue(entity, 'localTemperatureCalibrationUpdated', true);
```
Koenkk, line 5006: Instead of saving this in the globalStore and responding to the read request with a read response in fromZigbee there is now a better way to do this in the latest-dev branch.

You can replace this line with `endpoint.saveClusterAttributeKeyValue('hvacThermostat', {localTemperatureCalibration: value})` and zigbee-herdsman will automatically respond upon read.

pklokke: That's a nice feature. :-) It should allow me to refactor alot of these, in principle I can do away with all/most of `wiser_smart_thermostat_client`

However some of them need to be pushed out to the device (via an attribute write) if/when the value is updated from z2m (like the example you highlight above).  Is there a best practice for these type of attribute updates for sleepy end devices?

pklokke: I just checked, and it seems that 1a2097a8773cf521fb727aa40c194b63a8a1c1ac only supports named attributes, is this by design, or could it be extended to support number only attributes?

Koenkk: > However some of them need to be pushed out to the device (via an attribute write) if/when the value is updated from z2m (like the example you highlight above). Is there a best practice for these type of attribute updates for sleepy end devices?

Just implemented this, see https://github.com/Koenkk/zigbee-herdsman-converters/pull/2558/files#r635451761

Koenkk: > I just checked, and it seems that Koenkk/zigbee-herdsman@1a2097a8773cf521fb727aa40c194b63a8a1c1ac only supports named attributes, is this by design, or could it be extended to support number only attributes?

If needed I can extend it.

converters/toZigbee.js
```diff
@@ -4985,6 +4985,8 @@ const converters = {
     wiser_vact_calibrate_valve: {
         key: ['calibrate_valve'],
         convertSet: async (entity, key, value, meta) => {
+            await entity.command('hvacThermostat', 'wiserSmartCalibrateValve',
```
pklokke, line 4988: Using sendWhenActive with `entity.command` doesn't appear to work as intended, it seems to dispatch it immediately

converters/toZigbee.js
```diff
@@ -5004,15 +5006,18 @@ const converters = {
...
+                {srcEndpoint: 11, disableDefaultResponse: true, sendWhenActive: true});
             return {state: {local_temperature_calibration: value}};
         },
     },
     wiser_sed_thermostat_keypad_lockout: {
         key: ['keypad_lockout'],
         convertSet: async (entity, key, value, meta) => {
-            globalStore.putValue(entity, 'keypadLockoutUpdated', true);
+            const keypadLockout = utils.getKey(constants.keypadLockoutMode, value, value, Number);
+            await entity.write('hvacUserInterfaceCfg', {keypadLockout},
```
pklokke, line 5019: using `sendWhenActive` with entity.write appears to defer the dispatch, however when the device wakes up it gets stuck in a loop and is constantly sending the write attribute command over the air.

converters/toZigbee.js
pklokke: Hi @Koenkk thanks for the feedback, I have worked a bit on refactoring it (see the following commits), I may be holding one of the APIs wrong or else there may be some underlying issue, the behavior isn't quite as expected.

pklokke: > If needed I can extend it.

I have given the example if we don't below, as it may be an unnecessary complication of the core attribute handling, for my one edge case. The workaround below gets the job done, albeit in a slightly clunky way.

converters/toZigbee.js
Koenkk: Are you sure you tested on the latest dev branch?

pklokke: I pulled my zigbee-herdsman checkout in my node_modules up to master and built the javascript, but perhaps there are other dependencies i've missed.  I'll update to dev and recheck.

Koenkk: > using sendWhenActive with entity.write appears to defer the dispatch, however when the device wakes up it gets stuck in a loop and is constantly sending the write attribute command over the air.

this issue should be fixed now. (https://github.com/Koenkk/zigbee-herdsman/commit/81d8bb9416eb91662c58c70f49ccebdc7fe4efb7)

Change is the in latest-dev, pull the latest dev branch and execute `npm ci` will make sure you are on the correct version.

pklokke: I just double-checked, and I can reproduce this with the latest dev branch, only when sending a cluster command though (attribute read and writes for example work fine).

converters/toZigbee.js
pklokke: This looping issue seems to be resolved with the latest fix in zigbee-herdsman 👍

converters/toZigbee.js
pklokke: I'll keep hunting the root cause, there does seem to be something quite suspect in the log entry though:

`zigbee-herdsman:controller:endpoint Command 0x00124b000426a8a0/11 hvacThermostat.wiserSmartCalibrateValve({"srcEndpoint":11,"disableDefaultResponse":true,"sendWhenActive":true}, {"sendWhenActive":false,"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":false,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) +6s`

converters/toZigbee.js
pklokke: Apparently I should RTFL, the answer was staring me in the face... I had no payload, but options, so i need to insert an empty payload in the call.

converters/toZigbee.js
pklokke: > Hi @Koenkk thanks for the feedback, I have worked a bit on refactoring it (see the following commits), I may be holding one of the APIs wrong or else there may be some underlying issue, the behavior isn't quite as expected.

These `sendWhenAcive` issues are now resolved, either by fixes on your side, or by me not being foolish and forgetting parameters.  Just the item below.

converters/fromZigbee.js
```diff
@@ -5560,6 +5560,84 @@ const converters = {
             return result;
         },
     },
+    wiser_smart_thermostat_client: {
+        cluster: 'hvacThermostat',
+        type: 'read',
+        convert: async (model, msg, publish, options, meta) => {
+            const response = {};
+            if (msg.data[0] == 0xe010) {
```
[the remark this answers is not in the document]
Koenkk: I checked how to implement this but this would get a bit dirty. (since we don't know the dataType). I propose to add these attributes to `cluster.ts` with a `manufacturerCode` instead (like: https://github.com/Koenkk/zigbee-herdsman/commit/3eed20c2861a3a5024b751baaa841e310322f4b1)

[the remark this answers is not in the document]
pklokke: As mentioned in Koenkk/zigbee-herdsman#363, these devices don't set the MS bit, so they wouldn't be recognised by the handler with `manufacturerCode` set.  The product's behaviour is non-standard (they're older, ZHA 1.2-ish devices, with some internal Schneider cruft) so perhaps its better to keep the brain-dead stuff in the handler rather than messing up the rest of the code.

Another option would be extending the api to allow the datatype to be embedded for "anonymous" attributes, like this:
`endpoint.saveClusterAttributeKeyValue('hvacThermostat', {0xe010: {value: 1, type: 0x30}});` but at first glance the handling for this in typescript is a bit tricky.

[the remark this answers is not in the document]
Koenkk: Totally agree

Koenkk:
Thanks!

Koenkk:
@pklokke I've temporarily reverted this PR, with the z2m test I found that the HA integration for `EER51000` won't work since it has no `local_temperature`. Could you add this?

pklokke:
> @pklokke I've temporarily reverted this PR, with the z2m test I found that the HA integration for `EER51000` won't work since it has no `local_temperature`. Could you add this?

Sure, I'll add this and repost the PR

PR #2592, opened 2021-05-17
Danfoss ally hive trv

robertalexa:
Hi @Koenkk 

This is my PR to do with Danfoss Ally and Hive TRV. I do not have the specs for the Hive TRV but it exposes pretty much everything that I have seen in the Danfoss specs. With that in mind, I have done the changes to both devices.

Changes:
- fixed wrong descriptions
- code cleanup
- add support to show control mounting
- fix internal window
- fix external window
- fix algorithm scale

I things these brings these 2 devices as close to full support as possible. There were some other attributes exposed but as far as I could understand from documentation they would not provide any advantage, so I didn't support them.

As mentioned in my issue https://github.com/Koenkk/zigbee2mqtt/issues/7461 the above changes (plus hopefully the below) will close my issue plus both https://github.com/Koenkk/zigbee2mqtt/issues/4504 and https://github.com/Koenkk/zigbee2mqtt/issues/4783

This leaves me with one single issue for these 2 devices, and that is how the occupied_heating_setpoint is set. But I am struggling with this and I would appreciate your help. Basically it will need a custom command for it following this spec
image
0x40 -> payload {type: 1, setpoint: degrees}

Let me know what your think. And please give us a hand with that command (I have tried something and failed misserably)

Rob


robertalexa:
@Koenkk

As a continuation of the previous issue, I have made a few more changes and updated the PR.

The PR is passing npm lint and npm test.

Would you be kind to help me with the setpoint command? I am not sure how to approach this to be honest with you, I am fairly new to this project, as you can see this is my first contribution. Please see the PR description for details on the command.

Please review my code and let me know if you find anything weird. Bear in mind that I have used the skeleton of the current Danfoss Ally implementation and from there I have updated both the TRVs.

Here is a screenshot of the Frontend with the device
image

And here is the entry in the db for it, just in case you need it
```
{"id":2,"type":"EndDevice","ieeeAddr":"0x842e14fffe203c4c","nwkAddr":63898,"manufId":4678,"manufName":"Danfoss","powerSource":"Battery","modelId":"TRV001","epList":[1],"endpoints":{"1":{"profId":260,"epId":1,"devId":769,"inClusterList":[0,1,3,10,32,513,516,2821],"outClusterList":[0,25],"clusters":{"genBasic":{"attributes":{"modelId":"TRV001","manufacturerName":"Danfoss","powerSource":3,"zclVersion":3,"appVersion":33,"stackVersion":8,"hwVersion":69,"dateCode":"20200402","swBuildId":"02.33.0008 02.33"}},"genPowerCfg":{"attributes":{"batteryPercentageRemaining":102}},"hvacUserInterfaceCfg":{"attributes":{"keypadLockout":0}},"hvacThermostat":{"attributes":{"16384":3,"16387":1,"16400":4,"16401":660,"16402":0,"16403":0,"16416":1,"16432":1,"16433":0,"16458":-8000,"localTemp":2484,"pIHeatingDemand":0,"occupiedHeatingSetpoint":3000}}},"binds":[{"cluster":1,"type":"endpoint","deviceIeeeAddress":"0x00124b0018e32e48","endpointID":1},{"cluster":513,"type":"endpoint","deviceIeeeAddress":"0x00124b0018e32e48","endpointID":1}],"configuredReportings":[{"cluster":1,"attrId":33,"minRepIntval":60,"maxRepIntval":43200,"repChange":1},{"cluster":513,"attrId":0,"minRepIntval":0,"maxRepIntval":600,"repChange":25},{"cluster":513,"attrId":8,"minRepIntval":0,"maxRepIntval":600,"repChange":1},{"cluster":513,"attrId":18,"minRepIntval":0,"maxRepIntval":600,"repChange":25},{"cluster":513,"attrId":16402,"minRepIntval":0,"maxRepIntval":600,"repChange":1},{"cluster":513,"attrId":16384,"minRepIntval":0,"maxRepIntval":3600,"repChange":1}],"meta":{}}},"appVersion":33,"stackVersion":8,"hwVersion":69,"dateCode":"20200402","swBuildId":"02.33.0008 02.33","zclVersion":3,"interviewCompleted":true,"meta":{"configured":-1063983552},"lastSeen":1621365108064}
```

Thanks,
Rob

robertalexa:
@Koenkk

I have managed to sort out the custom heating command. I have done a PR on herdsman.

This PR requires that code before merging in. https://github.com/Koenkk/zigbee-herdsman/pull/365

I am currently looking at another attribute but in essence I am done :)

Let me know what you think
Rob

robertalexa:
I am not done. No more changes on this.

Ready for you to review it.

Rob

converters/fromZigbee.js
```diff
@@ -2585,13 +2585,13 @@ const converters = {
                 result[postfixWithEndpointName('trigger_time', msg, model)] = msg.data[0x4011];
             }
             if (typeof msg.data[0x4012] == 'number') {
-                result[postfixWithEndpointName('mounted_mode', msg, model)] = (msg.data[0x4012]==1);
+                result[postfixWithEndpointName('mounted_mode_active', msg, model)] = (msg.data[0x4012]==1);
```
sjorge, line 2588: Might be worth to add those cluster attributes here:
https://github.com/Koenkk/zigbee-herdsman/blob/ef3d96c9ed4b75c57af1fc71e618add0f1c3064e/src/zcl/definition/cluster.ts#L1899 it should improve readability a lot.

Line I linked was for the Viessmann which is a similar TRV but without any documentation.

robertalexa: @sjorge Initially I have considered this https://github.com/robertalexa/zigbee-herdsman-converters/commit/dfccce4774f6356560927363be55d537aabcd6cd

However after discussing this with @Koenkk as part of a different issue he has advised agaisnt (unless i am misunderstanding this fully)

See issue here https://github.com/Koenkk/zigbee2mqtt/issues/7065

The fact that those clusters are named "CustomX" isn't great either, but based on the Danfoss specs (i too added support for Hive kinda blindly) I can map what some of those customs actually are.

viessmannCustom0: eTRV Open Window Detection - window_open_internal
viessmannCustom1: not sure
viessmannCustom2: not sure
viessmannWindowOpenForce: External Open Window Detected - window_open_external
viessmannCustom4: not sure
viessmannCustom5: not sure
viessmannCustom6: not sure
viessmannCustom7: Exercise day of week
viessmannCustom8: Exercise trigger time 
viessmannAssemblyMode: Mounting mode active
viessmannCustom10: Control algorithm scale factor
viessmannCustom11: not sure
viessmannCustom12: Regulation SetPoint Offset
viessmannCustom13: Adaptation run control
viessmannCustom14: Adaptation run status
viessmannCustom15: Adaptation run settings

Maybe in future we get them renamed.

@Koenkk Do you think the Danfoss Ally and Hive TRV should use clusters or just stick with msg.data[address]?

Rob

robertalexa: Let's see what Koen says about it. If he suggests using clusters instead of attribute address, then I can go through them, remove unused ones, rename the remaining and fix the devices depending on them as part of this PR.

But i agree that some of them should go, especially since we don not know what they actually do.

Rob

sjorge: Yeah I labeled them all and went through days of logging but they not all the same as the danfoss ones. They use a diff manufacture too.

Viessman was not helpful ar all when contact and they told me to stop reverse engineering there device.

robertalexa: LOL :) Typical corporate greed I suppose.

I contacted Danfoss to ask them if and how to flash their firmware on the Hive TRV (i dont like Hive, learnt the hard way) - they replied with a generic message...

I contacted Hive to ask about their specs, completely oblivious. They confirmed the device is identical to Danfoss Ally but is using a custom firmware, so I have no idea if it supports more or less features, but so far everything matched.

I do have some issues with thermostat_orientation though. The command to get and set work based on Danfoss Ally specs, but the display is not turning. Meh. But I have left it in so Danfoss users will benefit from it.

Koenkk: It's fine to add these to cluster.ts file, as long as the attributes have a manufacturerCode like the Viesmann ones.

robertalexa: @sjorge Did you say you already have something started on this or should I handle it, including "remove unused ones, rename the remaining and fix the devices depending on them as part of this PR."

@Koenkk Are you ok with me handling all of this in one go? Do another PR on herdsman and when ready we re-review this one?

Rob

sjorge: > @sjorge Did you say you already have something started on this or should I handle it, including "remove unused ones, rename the remaining and fix the devices depending on them as part of this PR."

Working on the PR now, nearly done.
It's basically rename viessmannCustom0 -> viessmanWindowOpen and drop the other viessmannCustomX's.
And update the reference to viessmannCustom0 in zhc ofcourse.

robertalexa: Is it worth renaming that to viessmanWindowOpenInternal? Because some trvs have 2 one for internal and one for external, like the ones at hand.

I will wait for your PRs to be merged, then start with "good" code basically and deal with my clusters.

Thanks for confirming

sjorge: Maybe? They Viessmann seems to be the same as the Danfoss with a very buggy firmware.

sjorge: Seems plausible, don’t have the TRV paired atm to verify.

robertalexa: @sjorge As part of me adding the clusters for the Danfoss, do you mind if I rename (herdsman and converters) viessmannWindowOpenForce to viessmannWindowOpenExternal? Thanks

sjorge: @robertalexa  you probably want to rename the attribute names in the mqtt payload too then.
I did look at the existing danfoss stuff and it didn't quiet line up, so aside the naming, the code should probably stay the same to be sure.

I don't have access to the viessmann to retest.

robertalexa: @sjorge I shall leave it as is. I guess it won't harm anyone. Any realistically we don't really know if they actually call it external. Thanks for confirming anyway

robertalexa: @sjorge Please help. At this point I can't tell if I am tired, stupid or just missing something. As you know there is s PR for herdsman clusters.
toZigbee.js
This code (old) works:
```
danfoss_mounted_mode_control: {
        key: ['mounted_mode_control'],
            const payload = {0x4013: {value: (value ? 0x00: 0x01), type: 0x10}};
            await entity.write('hvacThermostat', payload, manufacturerOptions.danfoss);
        },
            await entity.read('hvacThermostat', [0x4013], manufacturerOptions.danfoss);
        },
    }
```
This code (new) does not:
```
danfoss_mounted_mode_control: {
        key: ['mounted_mode_control'],
            const payload = {'danfossMountedModeControl': {value: (value ? 0x00: 0x01), type: 0x10}};
            await entity.write('hvacThermostat', payload, manufacturerOptions.danfoss);
        },
            await entity.read('hvacThermostat', ['danfossMountedModeControl'], manufacturerOptions.danfoss);
        },
    }
```

What am I missing? Or what other info can i send you for debugging? Thanks in advance.

fromZigbee.js
```
if (msg.data.hasOwnProperty('danfossMountedModeControl')) {
                result[postfixWithEndpointName('mounted_mode_control', msg, model)] = (msg.data['danfossMountedModeControl'] === 0);
            }
```
Getting the info works fine, setting it doesn't

Also here is some more debug info for both, to compare
```
Zigbee2MQTT:debug 2021-05-20 22:25:55: Publishing 'set' 'mounted_mode_control' to '0x842e14fffe203c4c'
  zigbee-herdsman:controller:endpoint Write 0x842e14fffe203c4c/1 hvacThermostat({"danfossMountedModeControl":{"value":1,"type":16}}, {"sendWhenActive":false,"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":4678,"transactionSequenceNumber":null,"writeUndiv":false}) +5s


Zigbee2MQTT:debug 2021-05-20 22:29:09: Publishing 'set' 'mounted_mode_control' to '0x842e14fffe203c4c'
  zigbee-herdsman:controller:endpoint Write 0x842e14fffe203c4c/1 hvacThermostat({"16403":{"value":1,"type":16}}, {"sendWhenActive":false,"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":4678,"transactionSequenceNumber":null,"writeUndiv":false}) +0ms
```
This is happening for SET on all the custom commands
@Koenkk Any thoughts?

robertalexa: I have commited my code so you can have an easy before/after
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/ea4a84aaaeb11b5b7e0f30945cbb7601bbc6ca0e

Hope this makes it easier.

sjorge: I think I know what is wrong, as the info is now in herdsman, you do not need to specify the data type.
```
const payload = {'danfossMountedModeControl': {value: (value ? 0x00: 0x01), type: 0x10}};
```
Should be
```
const payload = {'danfossMountedModeControl': (value ? 0x00: 0x01)};
```

converters/toZigbee.js
```diff
@@ -2037,8 +2037,9 @@ const converters = {
     danfoss_mounted_mode_control: {
         key: ['mounted_mode_control'],
         convertSet: async (entity, key, value, meta) => {
-            const payload = {'danfossMountedModeControl': {value: (value ? 0x00: 0x01), type: 0x10}};
+            const payload = {'danfossMountedModeControl': (value ? 0 : 1)};
             await entity.write('hvacThermostat', payload, manufacturerOptions.danfoss);
+            return {readAfterWriteTime: 200, state: {'mounted_mode_control': value}};
```
sjorge, line 2042: Do we want to return 0 and 1 here or true and false?

I also just checked: `danfossMountedModeControl: {ID: 0x4013, type: DataType.boolean, manufacturerCode: ManufacturerCode.DANFOSS},`

We probably want to take a bool as imput and just pass true or false as value here.

converters/toZigbee.js
```diff
@@ -2047,8 +2048,9 @@ const converters = {
     danfoss_thermostat_orientation: {
         key: ['thermostat_orientation'],
         convertSet: async (entity, key, value, meta) => {
-            const payload = {'danfossThermostatOrientation': {value: (value ? 0x01: 0x00), type: 0x10}};
+            const payload = {'danfossThermostatOrientation': (value ? 1: 0)};
```
sjorge, line 2051: Same since danfossThermostatOrientation is marked a bool, we can pass true/false

converters/fromZigbee.js
robertalexa: Thanks for pointing me in the right direction. You are absolutely right, but I initially didn't find it obvious. Thanks. I have pushed another commit which I hope will "Resolve this conversation"

I would appreciate some review and feedback in hive/danfoss.js

```
            // danfoss attributes
            await endpoint.configureReporting('hvacThermostat', [{attribute: 'danfossMountedModeActive', minimumReportInterval: 0,
                maximumReportInterval: constants.repInterval.MINUTES_10, reportableChange: 1}], options);
            await endpoint.configureReporting('hvacThermostat', [{attribute: 'danfossWindowOpenInternal', minimumReportInterval: 0,
                maximumReportInterval: constants.repInterval.HOUR, reportableChange: 1}], options);
            await endpoint.configureReporting('hvacThermostat', [{attribute: 'danfossHeatRequired', minimumReportInterval: 0,
                maximumReportInterval: constants.repInterval.MINUTES_10, reportableChange: 1}], options);

            // read keypadLockout, we don't need reporting as it cannot be set physically on the device
            await endpoint.read('hvacUserInterfaceCfg', ['keypadLockout']);
            await endpoint.read('hvacThermostat', [
                'danfossWindowOpenExternal',
                'danfossDayOfWeek',
                'danfossTriggerTime',
                'danfossAlgorithmScaleFactor',
                'danfossHeatAvailable',
                'danfossMountedModeControl'
            ], options);
```

I am not sure I did the right thing. This is the technical specs https://assets.danfoss.com/documents/176987/AM375549618098en-000101.pdf

But I would be very happy to be "schooled"

Thanks a lot @sjorge for your involvement and commitment with this PR and all the work around it.

converters/toZigbee.js
```diff
@@ -2067,7 +2069,7 @@ const converters = {
     danfoss_heat_available: {
         key: ['heat_available'],
         convertSet: async (entity, key, value, meta) => {
-            const payload = {'danfossHeatAvailable': {value: (value ? 0x01: 0x00), type: 0x10}};
```
sjorge: also bool

converters/toZigbee.js
sjorge: Do we know when it is true and false?
e.g. true = vertical, false = horizontal? If we do, we might want to rename this thermostat_vertical_orientation instead... or horizontal... depending on which one results in a true value here.

converters/toZigbee.js
```diff
@@ -2057,7 +2059,7 @@ const converters = {
     danfoss_algorithm_scale_factor: {
         key: ['algorithm_scale_factor'],
         convertSet: async (entity, key, value, meta) => {
-            const payload = {'danfossAlgorithmScaleFactor': {value: value, type: 0x20}};
+            const payload = {'danfossAlgorithmScaleFactor': value};
             await entity.write('hvacThermostat', payload, manufacturerOptions.danfoss);
```
sjorge, line 2063: Add a return here to be consistent with the other ones?

converters/toZigbee.js
```diff
@@ -2067,7 +2069,7 @@ const converters = {
     danfoss_heat_available: {
         key: ['heat_available'],
         convertSet: async (entity, key, value, meta) => {
-            const payload = {'danfossHeatAvailable': {value: (value ? 0x01: 0x00), type: 0x10}};
+            const payload = {'danfossHeatAvailable': (value ? 1: 0)};
             await entity.write('hvacThermostat', payload, manufacturerOptions.danfoss);
```
sjorge, line 2073: Also add a return?

converters/fromZigbee.js
sjorge: Mostly looks good, left a few comments that should make the code more readable and more conssitant.

sjorge: That pdf has some interesting stuff in it, looks like if you have multiple they can somehow be bound and tell eachother to enter pre-heat if I am reading that right... anyway before I get to distracted.

We might want to tweak the default reporting a bit to save on battery
```
            await reporting.batteryPercentageRemaining(endpoint, {min: 60, max: 43200, change: 1});
            await reporting.thermostatTemperature(endpoint, {min: 0, max: constants.repInterval.MINUTES_10, change: 25});
            await reporting.thermostatPIHeatingDemand(endpoint, {min: 0, max: constants.repInterval.MINUTES_10, change: 1});
            await reporting.thermostatOccupiedHeatingSetpoint(endpoint, {min: 0, max: constants.repInterval.MINUTES_10, change: 25});
```

Seems rather aggressive, we can probably bumb batteryPercentageRemaining to min=1h or maybe even 3h, I doubt the TRV will drop by 1% every minute.


thermostatTemperature, thermostatPIHeatingDemand, thermostatOccupiedHeatingSetpoint dont have a minumum either. Perhaps pushing those to min=15sec or maybe even min=60sec

converters/fromZigbee.js
```diff
@@ -2572,35 +2572,38 @@ const converters = {
                 result[postfixWithEndpointName('pi_heating_demand', msg, model)] =
                     precisionRound(msg.data['pIHeatingDemand'], 0);
             }
-            if (typeof msg.data[0x4000] == 'number') {
-                result[postfixWithEndpointName('window_open_internal', msg, model)] = (msg.data[0x4000]);
+            if (msg.data.hasOwnProperty('danfossWindowOpenInternal')) {
+                result[postfixWithEndpointName('window_open_internal', msg, model)] = msg.data['danfossWindowOpenInternal'];
```
sjorge, line 2576: This maps to 
0x00: Quarantine
0x01: Windows are closed
0x02: Hold ,Windows are maybe about to open
0x03: Open window detected
0x04: In window open state from external

Right? We can probably use a nice enum here! (https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/lib/constants.js already has a bunch so adding a danfossWindowOpen or something.

Here is an example of how the mapping is done:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/50ad929821ff1a93dd3e826f2a4ed4fa078d8d6c/converters/fromZigbee.js#L144
https://github.com/Koenkk/zigbee-herdsman-converters/blob/50ad929821ff1a93dd3e826f2a4ed4fa078d8d6c/converters/toZigbee.js#L1118

converters/toZigbee.js
sjorge: Found it in the pdf:
0x00: Horizontal (Default)
0x01: VerticalDefault is 0, but overwritten to value from production configuration at Init. 

It might be cleaner to add a new enum as mentioned https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/files#r636517838

converters/fromZigbee.js
robertalexa: @sjorge I have changed the battery to 1 hours. With regards to the rest, correct me if I am wrong but if we change reporting of for example thermostatOccupiedHeatingSetpoint to 60 seconds, and then you change the temperature physically on the TRV, it can then take up to 60 seconds for it to be reported back to z2h?

Apart from the 3 you have mentioned there are these ones with a min of 0 too. Would you suggest we increase them too?
```
await endpoint.configureReporting('hvacThermostat', [{attribute: 'danfossMountedModeActive', minimumReportInterval: 0,
                maximumReportInterval: constants.repInterval.MINUTES_10, reportableChange: 1}], options);
            await endpoint.configureReporting('hvacThermostat', [{attribute: 'danfossWindowOpenInternal', minimumReportInterval: 0,
                maximumReportInterval: constants.repInterval.HOUR, reportableChange: 1}], options);
            await endpoint.configureReporting('hvacThermostat', [{attribute: 'danfossHeatRequired', minimumReportInterval: 0,
                maximumReportInterval: constants.repInterval.MINUTES_10, reportableChange: 1}], options);
```

Also would you be kind to explain to me what exactly does this read do? I must confess that I have filled it in but without full understanding of the behaviour
```
await endpoint.read('hvacThermostat', [
                    'danfossWindowOpenExternal',
                    'danfossDayOfWeek',
                    'danfossTriggerTime',
                    'danfossAlgorithmScaleFactor',
                    'danfossHeatAvailable',
                    'danfossMountedModeControl'
                ], options);
```

Thanks, Rob

converters/toZigbee.js
robertalexa: Done

https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/aa227ca98224cb5f5700c0755db6f481982b9b8f

converters/toZigbee.js
robertalexa: Done
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/aa227ca98224cb5f5700c0755db6f481982b9b8f

converters/fromZigbee.js
sjorge: > example thermostatOccupiedHeatingSetpoint to 60 seconds, and then you change the temperature physically on the TRV, it can then take up to 60 seconds for it to be reported back to z2h?

Sort of ...
- physically set the temp to 24 on the TRV
- report gets send
- physically set the temp to something else within minInterval
- report gets delayed until minInterval is passed

We can safely raise the minimum for anything that is not physically changeable on the device.

> Also would you be kind to explain to me what exactly does this read do? I must confess that I have filled it in but without full understanding of the behaviour

To get the initial state we read some attributes, usually for things that do not have reporting setup and are not user trigger-able on the device. e.g. for the viessmann we read the keypadLockout from hvacUserInterface, so we have a value available for the frontend. They only way to set it, is via mqtt messages so it makes no sense to setup reporting for that attribute as it cannot be changed without going through zigbee2mqtt.

robertalexa: Based on your reply I have done the following:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/cf0c6b41fa8272477b7c6493807523a44e3952c7

So the 1st reporting will be instant as a result of a change and the 2nd report will be delayed up to minInterval from 1st report.

I hope this change will "Resolve this conversation" chain. But I will leave this up to you to close? (i hope you can see the button too?)

converters/toZigbee.js
robertalexa: @sjorge Apologies if I might sound dumb but not sure how would you like me to approach this.

One thing to note is that danfossMountedModeControl specs is:
```
0x00 Go to mounting mode (the eTRV can now be
mounted on a valve)
0x01 Go to Mounted posittion (the eTRV now act as if it's
mounted on a valve)
```

I guess for heat_available this code:
```
const payload = {'danfossHeatAvailable': (value ? 1: 0)};
            await entity.write('hvacThermostat', payload, manufacturerOptions.danfoss);
```
Can be simplified to this:
```
await entity.write('hvacThermostat', {'danfossHeatAvailable': value}, manufacturerOptions.danfoss);
```

Is this what you are referring to?

converters/toZigbee.js
robertalexa: on my Hive thermostat this command is not doing anything, unless I am missing something. The command works fine and the new value is reported back correctly, but the display does not turn. I wonder if this is Hive's firmware being dodgy. Settings this from the Hive app works though, but not until you turn the heating physically, only after that the display updates. Unfortunately I do not have an extra sniffer nor have I ever sniffed traffic so not sure how to approach. I am hoping that this will work for Danfoss Ally users though...

robertalexa: saying that, I have just noticed a different attribute on a different cluster that might do just that. let me test
image

robertalexa: @sjorge Yep, that did the trick. https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/47be7ab768405dcec0fe8ceda151c372217331af

converters/toZigbee.js
sjorge: Correct, as this is a bool value we can further reduce it to just the simplified version.

converters/fromZigbee.js
sjorge: Left one more comment, but will resolve this one as that one shows up as a seperate thread.

converters/toZigbee.js
sjorge: Looks like they are indeed 2 different things, one probably indicated if the valve is horizontal or vertical, the other which way the display is. E.g. when it's vertical you might still want to flip it for example if the TRV is in a corner.

Renaming `thermostat_orientation` to `thermostat_vertical_orientation` would make sense I think, as true indicates it's vertical and false horizontal. Or either we map true -> 'vertical' and false -> 'horizontal' and leave the name.

So the mqtt publish would be `{'thermostat_orientation': 'vertical'}` (true) or `{'thermostat_orientation': 'horizontal'}` (false)... but renaming the attribute seems easier and accomplishes the same effect?

converters/toZigbee.js
```diff
@@ -2004,95 +2004,140 @@ const converters = {
...
+        convertGet: async (entity, key, meta) => {
+            await entity.read('hvacThermostat', ['danfossWindowOpenInternal'], manufacturerOptions.danfoss);
+        },
+    },
+    danfoss_window_open_external: {
         key: ['window_open_external'],
         convertSet: async (entity, key, value, meta) => {
-            const payload = {0x4003: {value: (value ? 0x01: 0x00), type: 0x10}};
+            const payload = {'danfossWindowOpenExternal': (value ? 1: 0)};
             await entity.write('hvacThermostat', payload, manufacturerOptions.danfoss);
```
sjorge, line 2130: `await entity.write('hvacThermostat', {'danfossWindowOpenExternal': value}, manufacturerOptions.danfoss);` should be fine too I think, given the type of bool for danfossWindowOpenExternal

devices/hive.js
```diff
@@ -113,22 +113,92 @@ module.exports = [
...
+            await reporting.batteryPercentageRemaining(endpoint, {min: constants.repInterval.HOUR, max: 43200, change: 1});
+            await reporting.thermostatTemperature(endpoint, {min: 0, max: constants.repInterval.MINUTES_10, change: 25});
+            await reporting.thermostatPIHeatingDemand(endpoint, {min: 0, max: constants.repInterval.MINUTES_10, change: 1});
+            await reporting.thermostatOccupiedHeatingSetpoint(endpoint, {min: 0, max: constants.repInterval.MINUTES_10, change: 25});
+
+            // danfoss attributes
+            await endpoint.configureReporting('hvacThermostat', [{
+                attribute: 'danfossMountedModeActive',
+                minimumReportInterval: constants.repInterval.MINUTE,
+                maximumReportInterval: constants.repInterval.MINUTES_10,
```
sjorge, line 169: Somehow github ate my comment, might have done it to an older commit...

Does having a max interval of 10 minutes makes sense? It will report on this at least 6 times per hour then... I doubt people will be unmounting the TRV from the valve very often. Probably having max the same as the batteryPercentage would make sense (43200). Especially since we also set reportableChange it will report as soon as the valve gets dismounted or mounted.

converters/fromZigbee.js
sjorge: Now that I am awake... it would be clean to add thes to constants.js I think.

```
const danfossWindowOpen = {
    0: 'quarantine',
    1: 'closing',
    2: 'hold',
    3: 'open',
    4: 'external_open',
};
```

Then in fromZigbee:
```
result[postfixWithEndpointName('window_open_internal', msg, model)]  = constants.danfossWindowOpen.hasOwnProperty(msg.data['danfossWindowOpenInternal']) ?
                    constants.danfossWindowOpen[msg.data['danfossWindowOpenInternal']] : msg.data['danfossWindowOpen'];
```

There is no toZigbee for danfossWindowOpenInternal so nothing needs to be done there.

That should make it so people don't have to look at the PDF/source to know that 3 and 4 are both open but the later because it was forced etc...

converters/toZigbee.js
robertalexa: Solved here
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/5338a8efcec58d1018d252f62bb519ace285fefd

converters/toZigbee.js
robertalexa: Solved here
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/5338a8efcec58d1018d252f62bb519ace285fefd

converters/toZigbee.js
robertalexa: Bool Solved here
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/5338a8efcec58d1018d252f62bb519ace285fefd

thermostat_vertical_orientation makes the most sense
Solved here
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/270708f40cb87a08b384d97d1a7cc2eb066187ed

devices/hive.js
robertalexa: @sjorge Your wild comment - you can delete it if you want since it is incomplete.
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592#commitcomment-51125580

I have changed that as per your suggestion. We didn't however have reportable change set up for danfossMountedModeActive, but I have added it to .read section.

Can you please have a look and let me know?
https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/0ed0e733d0046b0828d9c6225eac8c9271989dcf

sjorge: Reporting LGTM now.

converters/toZigbee.js
robertalexa: @sjorge Hi, had to disappear earlier so I couldn't leave a comment despite pushing a commit for this. Please see https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/293cef0be63c9ae7ba4bc853581dca14271c07c7

Now that works fine, but the false/true states are exact opposites of what the specification says. I have updated the description in the devices but I am not sure I should do that. What is your thoughts?

PS: I have tried to invert the values instead of description but somehow failed. I was however in a rush so maybe I did oversee sth

converters/fromZigbee.js
robertalexa: Thanks for the feedback and telling me the enum option. https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592/commits/1fc978fd5ba3e478039c1e98662ed9c2f39bd51a

constans.js - i have decided to add the attribute with the rest of the thermostat ones rather than at the bottom of the list, I have debated it for a while and chose this option
fromZigbee.js - your "else" key is not correct but it doesn't matter :). Also, I hate the formatting of the ternary I had to do in my code but Lint didn't like it my way (despite me thinking is much more readable). This is how I would have done it
image

I guess this resolves this too :)

robertalexa:
@sjorge @Koenkk Would you guys be kind to do a final review?

All Jorge's suggestions have been actioned and the device support should be good.

For the sake of leaving a "paper-trail", out of the leftover attributes these are the most notable ones, and my thoughts:
- I have decided not to try to implement the TRV on board scheduling on the basis of the schedule is lost on OTA (Danfoss only, Hive doesn't have OTA) and power cycle. It feels like in practice there will be better suited solutions for looking after schedules 
- There is no running_state or system_mode. They only support Heating/Heat. Will this cause an issue on HA?
- Load Balancing - The specs mention this but it is unclear how it works or how do multiple TRVs link together

Tomorrow I hope I will be successful in testing all this code in HA and see how everything looks like. Never looked at the code for it...

Thanking you both for your guidance and support
Rob

converters/toZigbee.js
sjorge: So when setting `danfossMountedModeControl` to true, `danfossMountedModeActive` will also turn to true because it can be mounted. Once you mount it both go to false?

Or if you don't mount it and send false to `danfossMountedModeControl`, then `danfossMountedModeActive` transitions to false?

Wait... or is it the other way around, the description is a bit weird for those. I'd say just pass true/false as is to both without inverting.

sjorge:
Just went over the PR again, LGTM.



>     * There is no running_state or system_mode. They only support Heating/Heat. Will this cause an issue on HA?
HA does tend to use these to control the device between off/heat/auto, the docs indeed state it does heat only. I think exposing this with just the 'heat' option expose should fine fine. The drop down in HA will just be fixed IIRC. (Not using HA myself)

I wonder how this device exposes the 'off' mode though, for the Viessmann it was via system_mode=sleep (9)
Based on the description of pIHeatingMode (`Level of heating demanded by the PI loop in percent 0: when the thermostat is in “off”`) it does seem to have an 'off' mode of sorts.

Stab in the dark is probably only via a schedule were you mark a timeslot as not needing heating and there won't be a real 'off' mode, just a don't heat now state.

>     * Load Balancing - The specs mention this but it is unclear how it works or how do multiple TRVs link together

Yeah I noticed that too in the PDF, it's very confusing but it indeed looks multiple TRV can be combined to be in the same 'room' and they will do some pre-heat stuff if one of them needs heating for example but how this all works is unclear. I'm assuming a binding of sorts is going to be involved but ???

It's probably safe to ignore this, if somebody really wants it they can  always figure it out and add the support.

Koenkk:
Ready for merge?

converters/toZigbee.js
robertalexa: Yes, they should always be the same value. Active gets reported but Control is not. You need to read it to get a value. Also Active is not writable, only Control is.

This is with the TRV mounted on the radiator
image

I agree that the description in the specs might be wrong, at least based on what I am getting from the TRV response, maybe it is an oversight on their side.

No further changes to this code

robertalexa:
@Koenkk Before you do, can you please share your thought on HA with regards to running_state and system_mode?

I do not have them exposed, any idea how it will behave in HA?

Apart from this it is all good, but if you can foresee any issues I would rather correct them before merging.

Thanks both!

Koenkk:
@robertalexa I've never used thermostats with HA but for z2m HA code and HA docs (https://www.home-assistant.io/integrations/climate.mqtt/) this is optional.

robertalexa:
Tbh danfoss didn't have it before and Hive was all wrong anyway. It can't be worse than it was.

Should i see any issues after i will go from there.

Thanks you @sjorge @Koenkk for your help

robertalexa:
Oh and by the way, do i need to do anything for the readme of .io ?

Koenkk:
@robertalexa I don't think, everything covered in the `exposes` will automatically be added with the next release.

robertalexa:
@sjorge @Koenkk I have just finished setting up HA on a virtual machine to test the dev branch using the Edge addon.

Everything works as expected on z2m side.

The only problem I have is indeed with the system_mode. Here is how the device looks like in HA
image

image

image

Could you please shed some light on what I need to do to only support "Heat" instead of all the possible commands? e.g. to change the template that gets exposed to HA?

Thanks in advance

sjorge:
Try adding it with only 'heat' in the list of supported modes: `.withSystemMode(['heat']),`

robertalexa:
Hey dude, this is exactly what I was about to suggest

`.withSystemMode(['heat']).withRunningState(['heat'])`

Apart from this line I do not think I need to add anything else to the devices code?

I can understand your angle about it, there seems to be an enum declared already, i should just be able to make use of it.

I don't actually understand what exercise day of week and exercise trigger actually is or does. But hey ho.

I will do some tests and then open a new PR.

I apologies @Koenkk for having to work on this again after you have already merged it.

sjorge:
> Apart from this line I do not think I need to add anything else to the devices code?

That should be enough, it's there for the viessmann too to avoid the same issue.

PR #2596, opened 2021-05-18
fixed wrong expose voltage.

nukusinji:
it is not support voltage.


devices/dawon_dns.js
```diff
@@ -250,6 +250,6 @@ module.exports = [
             await reporting.onOff(endpoint);
             await reporting.batteryPercentageRemaining(endpoint);
         },
-        exposes: [e.battery(), e.switch(), e.voltage()],
```
Koenkk, line 253: Does it support `e.battery()` instead?

Koenkk:
Thanks!

PR #2594, opened 2021-05-17
Add new model ID for Saswell TRVs

drzony:

PR #2601, opened 2021-05-19
Correct manufacturerName Saswell

hnykda:
I believe that @drzony 's https://github.com/Koenkk/zigbee-herdsman-converters/pull/2594 was a good direction, but based on the logs I see in my installation, I think it should be `_TZE200_zuhszj9s` instead. 

But note that I have zero understanding of this whole project.


drzony:
Yes, you are right

PR #2604, opened 2021-05-20
Add support for XDD11LM

gnep88:
Add support for XDD11LM: Aqara Opple MX960

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -53,6 +53,14 @@ module.exports = [
         ota: ota.zigbeeOTA,
     },
     {
+        zigbeeModel: ['lumi.light.cwopcn01'],
+        model: 'XDD11LM',
+        vendor: 'Xiaomi',
+        description: 'Aqara Opple MX960',
+        extend: xiaomiExtend.light_onoff_brightness_colortemp(),
+        ota: ota.zigbeeOTA,
+    },
+    {
         zigbeeModel: ['lumi.light.cwopcn02'],
         model: 'XDD12LM',
         vendor: 'Xiaomi',
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -57,7 +57,7 @@ module.exports = [
         model: 'XDD11LM',
         vendor: 'Xiaomi',
         description: 'Aqara Opple MX960',
-        extend: xiaomiExtend.light_onoff_brightness_colortemp(),
+        extend: xiaomiExtend.light_onoff_brightness_colortemp({colorTempRange: [175, 370]}),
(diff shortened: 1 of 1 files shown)


Koenkk:
Please add the color temp range as described in: https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html#31-retrieving-color-temperature-range-only-required-for-lights-which-support-color-temperature

gnep88:
Thanks Koenkk. I tried to send as what described but failed with following error. Is my Zigbee2mqtt 1.18.1 too old ? Shall I switch to the edge version ? 

Zigbee2MQTT:debug 2021-05-30 20:52:08: Received MQTT message on 'zigbee2mqtt/mylight/set' with data '{"read": {"cluster": "lightingColorCtrl", "attributes": ["colorTempPhysicalMin", "colorTempPhysicalMax"]}}'


Zigbee2MQTT:error 2021-05-30 20:52:08: Failed to call 'EntityPublish' 'onMQTTMessage' (TypeError: Cannot read property 'key' of undefined
    at /app/lib/extension/publish.js:185:56
    at Array.find (<anonymous>)
    at EntityPublish.onMQTTMessage (/app/lib/extension/publish.js:185:42)
    at Controller.callExtensionMethod (/app/lib/controller.js:380:44)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5))

Koenkk:
1.18.3 will do it but you probably need to switch to the new repo: https://github.com/zigbee2mqtt/hassio-zigbee2mqtt#installation

gnep88:
This is the response I received

`Debug Received Zigbee message from 'mylight', type 'readResponse', cluster 'lightingColorCtrl', data '{"colorTempPhysicalMax":370,"colorTempPhysicalMin":175}' from endpoint 1 with groupID 0
Info Read result of 'lightingColorCtrl': {"colorTempPhysicalMin":175,"colorTempPhysicalMax":370}`

I added the configuration in my external converter XDD11LM.js and there are two behaviors I would like to double check with you:

1) The scroll bar disappears for this device (in the dashboard, and also "expose" section under the device), instead, only the textbox is there (with up and down arrow and the label "mired")

2) If I enter 172 in the textbox (which is lower than the minimum value allowed), there is no red warning toast. But the debug log shows this clearly
`Debug Requested color_temp 172 is lower than minimum supported 175, using minimum!`

Not sure if these are correct behaviors. Thanks!

Koenkk:
1. Looks like a frontend issue, please create one here: https://github.com/nurikk/z2m-frontend/issues
2. Yes that is expected.

PR #2611, opened 2021-05-22
Update tuya.js support for M2805EGBZTN

ratawhisk:
Adding Support for the Tubular Motor from Zemismart M2805EGBZTN


Koenkk:
Thanks!

PR #2614, opened 2021-05-22
Danfoss & hive further improvements

robertalexa:
This is a continuation of the previous PR as a result of testing in HA with regards to system_mode and running_state.

On top, I have updated day of week to be a nice enum as per @sjorge suggestion.

Apologies for creating a new PR when you have just merged the other. I hope this will be all final.

LE: this is a continuation of https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592

Rob

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2582,7 +2582,10 @@ const converters = {
                 result[postfixWithEndpointName('window_open_external', msg, model)] = (msg.data['danfossWindowOpenExternal'] === 1);
             }
             if (msg.data.hasOwnProperty('danfossDayOfWeek')) {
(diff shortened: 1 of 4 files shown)


robertalexa:
@sjorge 
image

imported tz.thermostat_system_mode
`exposes.climate().withSetpoint('occupied_heating_setpoint', 5, 32, 0.5).withLocalTemperature().withPiHeatingDemand()
                .withSystemMode(['heat']).withRunningState(['heat']),`

System Mode would be fine with that, but running state would be unpopulated. So i didn't include it. I think HA will cover for that and display system_mode instead is running_state is not present.

I have also made changes for day of the week to satisfy your suggestion.

Please have a look

sjorge:
LGTM

PR #2615, opened 2021-05-22
Wiser Smart Europe Device Support Attempt#2

pklokke:
- Added Room and Radiator Thermostats, Heating and Load Actuators and 2nd Gen Smartplug
- All devices excluding Smartplug require holding set button for 5 seconds to join
- Smartplug will join with short press, but only on channels 11,15,20,25
- Consolidated all the previous fixes
- Changed HACT and RTS to use `local_temperature` (used handy new schneider converter just made :-) )
- Removed some dead references on the HACT
- Fixed issue where `zone_mode` started as zero, stopping RTS from showing view unless explicitly selected

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5670,6 +5670,84 @@ const converters = {
             return {[property]: calibrateAndPrecisionRoundOptions(temperature, options, 'temperature')};
         },
     },
+    wiser_smart_thermostat_client: {
+        cluster: 'hvacThermostat',
+        type: 'read',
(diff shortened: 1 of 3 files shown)


pklokke:
Fingers crossed this is inline with whats needed, I don't use HA myself so it isn't part of my normal testing, happy to get it to interop though

Koenkk:
Thanks again!

pklokke:
> Thanks again!

No worries! and thanks to you @Koenkk for all the support!

PR #2617, opened 2021-05-23
Allow skipping of publishing color changes on set

sjorge:
When using reporting publish on set will make the color sliders jump around, we can optionally skip the publishing of the color set state to avoid this. By default we publish the color property on set.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -868,8 +868,13 @@ const converters = {
 
             const payload = {colortemp: value, transtime: utils.getTransition(entity, key, meta).time};
             await entity.command('lightingColorCtrl', 'moveToColorTemp', payload, utils.getOptions(meta.mapped, entity));
-            return {state: libColor.syncColorState({'color_mode': constants.colorMode[2], 'color_temp': value}, meta.state, meta.options),
-                readAfterWriteTime: payload.transtime * 100};
+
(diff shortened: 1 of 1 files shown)


Koenkk:
There is already an option available which does this called `optimistic`, see https://www.zigbee2mqtt.io/information/configuration.html#all-devices

sjorge:
I original had optimistic set to false but that had all sort of weird side effects with groups and non color payloads.

Koenkk:
@sjorge can you open an issue for this to debug this further? I prefer to not add a new option for such a specific case, especially since I already would expect the `optimistic` option to handle this case.

sjorge:
> @sjorge can you open an issue for this to debug this further? I prefer to not add a new option for such a specific case, especially since I already would expect the `optimistic` option to handle this case.

See my edit above,

> Would extending optimisic with a list of attribute enable/disable it for for be a better solution? The issues I am seeing are specific to color stuff, as the reported values are not always the same as the set values so it jumps around, other values like brightness don't have that issue.

sjorge:
I think I found the issue https://github.com/Koenkk/zigbee2mqtt/blob/f2e39af12cfaf46166b053ee43b28fb862d46731/lib/extension/publish.js#L227 it always publishes for groups when doing /set which is why it wasn;t working with optimistic=true but it's working as expected with this change.

PR #2630, opened 2021-05-27
Addded E26 variant to IKEA LED1924G9

nin97:
Tested with my E26 variant purchased at IKEA Canada on May 2021. Working fine with Zigbee2mqtt 1.18.3.

Please note it is 800lm and not 806lm but it matches the same model name.


Koenkk:
Thanks!

PR #2631, opened 2021-05-28
TuYa TS0216 Sound and flash siren with linear mapped volume

iglu-sebastian:
For mine with `manufacturerName` `_TYZB01_0wcfvptl` this is needed as volume seems to be a linear function as described in the commits.


iglu-sebastian:
@Koenkk if this has no other side effects this makes more user friendly to read.

Koenkk:
@iglu-sebastian is the range now from 0 - 100?

iglu-sebastian:
In original code my alarm was on max. volume when set to `10` and lowest volume set to `100`. Anything below `10` the alarm stayed off.

Thought this makes it a bit user friendly.

Koenkk:
Cool!

PR #2633, opened 2021-05-28
Fixed description for TS0011F_plug

avee87:
This is a tiny change just removing EU from this plug description because I just bought a UK plug and it uses the same signature :)


Koenkk:
Thanks!

PR #2635, opened 2021-05-29
Add support for Develco AQSZB-110

sjorge:
I don't own the device, but it was simple enough to add. I also moved some converters to the correct region.

This fixes koenkk/zigbee2mqtt#7171 , there are some changes compared to what was tested in the issue, mostly lint + a custom battery converter to support the battery low as per the develco technical docs.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1437,6 +1437,48 @@ const converters = {
             return {...payload1, ...payload2};
         },
     },
+    develco_voc: {
+        cluster: 'develcoSpecificAirQuality',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const voc = parseFloat(msg.data['measuredValue']);
+            const vocProperty = postfixWithEndpointName('voc', msg, model);
+
+            let airQuality;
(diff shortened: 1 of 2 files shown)


converters/fromZigbee.js
```diff
@@ -1437,6 +1437,48 @@ const converters = {
...
+        cluster: 'genPowerCfg',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            /*
+             * Per the technical documentation for AQSZB-110:
+             * To detect low battery the system can monitor the "BatteryVoltage" by setting up a reporting interval of every 12 hour.
+             * When a voltage of 2.5V is measured the battery should be replaced.
+             * Low batt LED indication–RED LED will blink twice every 60 second.
+             */
+            const result = converters.thermostat.convert(model, msg, publish, options, meta);
```
Koenkk, line 1477: I don't understand why the thermostat converter is called here. (should be the battery one?)

sjorge: Oops, indeed that should be battery.

Koenkk:
Thanks! 🎖️

PR #2643, opened 2021-06-01
Support position & battery for Qmotion HDM40PV620

pabergman:
Expose shade position and battery percentage as well as reporting for Qmotion HDM40PV620 shades.

### DIFF
--- a/devices/qmotion.js
+++ b/devices/qmotion.js
@@ -10,8 +10,8 @@ module.exports = [
         model: 'QZR-ZIG2400',
         vendor: 'Qmotion',
         description: '5 channel remote',
-        fromZigbee: [fz.identify, fz.cover_position_tilt],
-        exposes: [e.action(['identify']), exposes.numeric('position', ea.STATE)],
+        fromZigbee: [fz.identify],
+        exposes: [e.action(['identify'])],
         toZigbee: [],
     },
     {
@@ -26,7 +26,7 @@ module.exports = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg', 'closuresWindowCovering']);
             await reporting.batteryPercentageRemaining(endpoint);
             await reporting.currentPositionLiftPercentage(endpoint);
-        }
+        },
         exposes: [e.cover_position(), e.battery()],
     },
 ];
--- a/devices/qmotion.js
+++ b/devices/qmotion.js
@@ -19,8 +19,14 @@ module.exports = [
         model: 'HDM40PV620',
         vendor: 'Qmotion',
         description: 'Motorized roller blind',
-        fromZigbee: [fz.identify],
+        fromZigbee: [fz.identify, fz.cover_position_tilt, fz.battery],
         toZigbee: [tz.cover_state, tz.cover_position_tilt],
-        exposes: [e.cover_position()],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg', 'closuresWindowCovering']);
+            await reporting.batteryPercentageRemaining(endpoint);
+            await reporting.currentPositionLiftPercentage(endpoint);
+        }
+        exposes: [e.cover_position(), e.battery()],
     },
 ];


pabergman:
The code for the remote exposes `fz.cover_position_tilt` which I am not sure is correct, but I am weary of removing it. 

The shades also support setting max/min positions (had their hub before I switched to z2m), couldn't find anything similar for other shades. Do you have any recommendations on where to look or should I be asking qmotion for specs?

Thanks!

Koenkk:
> The code for the remote exposes fz.cover_position_tilt which I am not sure is correct, but I am weary of removing it.

Why?

> The shades also support setting max/min positions (had their hub before I switched to z2m), couldn't find anything similar for other shades. Do you have any recommendations on where to look or should I be asking qmotion for specs?

Probably something vendor specific, you need to sniff the traffic with the original gateway to reverse engineer this: https://www.zigbee2mqtt.io/how_tos/how_to_sniff_zigbee_traffic.html

pabergman:
> Why?

Just worried about breaking something, but tested and from what I can tell it doesn't do anything so I removed it in #60895f3 (along with comma fix).

> you need to sniff the traffic with the original gateway to reverse engineer this

Will give that a try as well as asking for the specs.

devices/qmotion.js
```diff
@@ -10,17 +10,23 @@ module.exports = [
         model: 'QZR-ZIG2400',
         vendor: 'Qmotion',
         description: '5 channel remote',
-        fromZigbee: [fz.identify, fz.cover_position_tilt],
-        exposes: [e.action(['identify']), exposes.numeric('position', ea.STATE)],
```
Koenkk, line 14: I would expect the remote send a `position`, but it seems in your case it doesn't? (since you mentioned this doesn't break anything. @willu since you added this in https://github.com/Koenkk/zigbee-herdsman-converters/pull/1571 , could you comment on this?

pabergman: In my setup the remotes are tied directly to the blinds (connected using button presses and tug commands). When I look at the traffic in the logs when commands are being sent I don't see anything about position.

```
Debug Received Zigbee message from 'Kitchen Shade Remote', type 'raw', cluster 'manuSpecificPhilips', data '{"data":[5,121,17,106,0,0],"type":"Buffer"}' from endpoint 1 with groupID 62706
Debug No converter available for 'QZR-ZIG2400' with cluster 'manuSpecificPhilips' and type 'raw' and data '{"data":[5,121,17,106,0,0],"type":"Buffer"}'
Debug Received Zigbee message from 'Kitchen Shade Remote', type 'raw', cluster 'manuSpecificPhilips', data '{"data":[5,121,17,107,3,20,1],"type":"Buffer"}' from endpoint 1 with groupID 62706
Debug No converter available for 'QZR-ZIG2400' with cluster 'manuSpecificPhilips' and type 'raw' and data '{"data":[5,121,17,107,3,20,1],"type":"Buffer"}'

Debug Received Zigbee message from 'Living Room Shade Remote', type 'raw', cluster 'manuSpecificPhilips', data '{"data":[5,121,17,6,0,0],"type":"Buffer"}' from endpoint 1 with groupID 42183
Debug No converter available for 'QZR-ZIG2400' with cluster 'manuSpecificPhilips' and type 'raw' and data '{"data":[5,121,17,6,0,0],"type":"Buffer"}'
Debug Received Zigbee message from 'Living Room Shade Remote', type 'raw', cluster 'manuSpecificPhilips', data '{"data":[5,121,17,7,3,20,1],"type":"Buffer"}' from endpoint 1 with groupID 42183
Debug No converter available for 'QZR-ZIG2400' with cluster 'manuSpecificPhilips' and type 'raw' and data '{"data":[5,121,17,7,3,20,1],"type":"Buffer"}'
Debug Received Zigbee message from 'Living Room Shade Remote', type 'raw', cluster 'manuSpecificPhilips', data '{"data":[5,121,17,8,3,20,2],"type":"Buffer"}' from endpoint 1 with groupID 42183
Debug No converter available for 'QZR-ZIG2400' with cluster 'manuSpecificPhilips' and type 'raw' and data '{"data":[5,121,17,8,3,20,2],"type":"Buffer"}'
```

The data they send seems to be in the  format of `[5,121,17,8,3,20,2]`, when I try to read that

```
5, 121, 17 - Not sure what that is, thought it might be identifier but the two remotes I tested 
send the same values.  Might be for programming commands.
8 - message number (I am guessing)
20 - position of shade
2 - the shade group, five buttons/on remotes  - 1,2,4,8,16 and 0 for all
```

Koenkk: > In my setup the remotes are tied directly to the blinds

I would expect that in a setup where the remote is not directly bind to the blinds (and to the coordinator instead) `fz.cover_position_tilt` makes sense. So I don't think this can be removed.

willu: I haven't spent a huge amount of time reverse engineering these blinds.

I have never used the QMotion hub - I associated both my blinds  with Zigbee2MQTT when they were installed.  Once they were on the zigbee network, all the standard QMotion tug commands and remote konami codes worked to set things up.  e.g. you can use the tug commands to tell the blind to disconnect from the network, or to try to re-associate with a network.  You can use the remote and tug commands to bind "channels" on the remote to particular blinds, and then to set the minimum and maximum positions of those blinds.

I didn't try to interpret the 'manuSpecificPhilips' data.  I also see that on the network, but have just ignored it.  I did try sending more normal ZDO commands to the blind.  e.g.

    mosquitto_pub -t 'zigbee2mqtt/Qmotion door/set' -m '{ "position":10 }' -q 1

makes one of my blinds drop until it is 90% closed.  And then

    mosquitto_pub -t 'zigbee2mqtt/Qmotion door/set' -m '{ "state":"open" }' -q 1

will make it wind fully open again.

So, yes, the remote does bind directly to the blinds (although maybe not in the ZDO sense), but this doesn't stop something else from controlling the blinds as well. e.g. a cron job that makes sure all the blinds are up at midnight, or a different zigbee remote through the central controller.

willu: I just had a bit more of a quick play.  If I hit the "20%" button on the remote, then I get a command sent from the remote that looks like this: Received Zigbee message from 'Qmotion remote', type 'raw', cluster 'manuSpecificPhilips', data '{"data":[5,121,17,26,3,20,0],"type":"Buffer"}' from endpoint 1 with groupID 45153.  The blind then responds with: No converter available for 'QZR-ZIG2400' with cluster 'manuSpecificPhilips' and type 'raw' and data '{"data":[5,121,17,26,3,20,0],"type":"Buffer"}'  i.e. it responds with the same data.  If I hit a few buttons in a row, then patterns emerge.  The first three elements of the buffer are always 5, 121, 17.  The next number (the 26 above) seems to be some sequence counter that increments for each command sent - I haven't checked when it resets.  The number after that seems to be a command, 3 meaning "go to this %" and the number after that being the %.  So the example above was "go to 20%". Note that the % here is the opposite of that used in the "position" command above, so this would be position "80".

The up and down arrows are commands 0 and 1, but they seem to be stateful.  A command of 0 will start the blind going up, and then a command of 1 will stop it.  Then a second command of 1 will start it going down, and a command of 0 will stop it.

I haven't looked through logs to see if it is broadcasting its state with any regularity.  All I've seen are the responses to the remote.  They are immediate, and seem to be the current set-point of the blind rather than its current position.  e.g. if the blind is fully down and you hit the 20% button then you'll get an immediate response of 20% from the blind even though it will take 10 seconds to get to position.

willu:
I'm comfortable with merging this.  My original patch was very minimal - almost anything would be an improvement.

strobejb:
With a small tweak I was able to get this working. Battery level and position are being reported now. The patch was just missing the following line which needs to be added at the top:

`const reporting = require('zigbee-herdsman-converters/lib/reporting');`

Shall I open a new PR, or merge this change into the existing one?

Koenkk:
@strobejb please make a pr

strobejb:
> @strobejb please make a pr

ok, raised a PR: #3968

PR #2644, opened 2021-06-01
Switch AXIS Gear to use closureWindowCovering

sjorge:
After being stuck in transit for over 2 months, I finally got 2 of these last week.

However I was having some problems when trying to set a position other than 0 or 100.
Every so often (1/6 ish times) they would do nothing and flash leds RED on the unit, indicating an error, then I have to touch the power button on the unit to make it go again.

I didn't immediately see anything wrong from the z2m side of things, but I did notice the unit also supports closuresWindowCovering cluster. Out of ideas I tried using this instead of genLevelCtrl, so far I have not had any errors. (Granted I only played with it for about and hour raising and lowering it to different %'s or fully open and close.

We seem to lose the possibility to query state, but I'd say that's better than having to physically interact with the unit to get it going again. Also we gain 'stop' which is nice I guess.

### DIFF
--- a/devices/axis.js
+++ b/devices/axis.js
@@ -3,8 +3,6 @@ const fz = {...require('../converters/fromZigbee'), legacy: require('../lib/lega
 const tz = require('../converters/toZigbee');
 const reporting = require('../lib/reporting');
 const e = exposes.presets;
-const ea = exposes.access;
-
 module.exports = [
     {
         zigbeeModel: ['Gear'],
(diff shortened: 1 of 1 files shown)


sjorge:
As far as I can tell, this should not effect the way the user currently interacts with the blind from a mqtt payload/HA perspective.

sjorge:
Bunch of more testing, it seems the axis gear trips up if you send it multiple genLevelCtrl commands before the other one is finished. Still not able to replicate with closureWindowCovering so I'm pretty confident now this is the way to go.

Koenkk:
Thanks!

PR #2646, opened 2021-06-02
Add support for Current Products Group Hybrid E-Wand Gen 2

Leah-Anderson:
This adds support for the Gen. 2 Hybrid E-Wand from Current Products Corp. More information about this device can be found here (https://www.myewand.com/product-page/gen-2-hybrid-e-wand-single-unit). 

I had to add an additional preset in exposes.js to have a tilt slider without a position slider. If there is a more ideal way to do this let me know. 

I am still seeing this in the logs: 
```
No converter available for 'CP180335E-01' with cluster '64528' and type 'raw' and data '{"data":[13,99,18,77,128,192,213,224,128,254,2,0,0,0,0,88,63,0,0,207,65,0,255,0,0,107,0],"type":"Buffer"}'
```

I'm not quite sure what to do with that though but everything is working as expected.

This resolves this issue: Koenkk/zigbee2mqtt#7648

### DIFF
--- a/devices/current_products_corp.js
+++ b/devices/current_products_corp.js
@@ -1,0 +1,26 @@
+const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
+const tz = require('zigbee-herdsman-converters/converters/toZigbee');
+const exposes = require('zigbee-herdsman-converters/lib/exposes');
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #2648, opened 2021-06-03
Add support for Wiser 40/300-Series Module Dimmers

phindmarsh:
This PR adds support for the [Wiser 40/300-Series Dimmers](https://zigbeealliance.org/zigbee_products/wiser-40-300-series-module-dimmer-2/) `CH/DIMMER/1` devices.

It transpires they are very similar to the `NHROTARY/DIMMER/1`, but its a push button dimmer not a rotary one which is why I duplicated it rather than adding an additional model fingerprint to that stanza.

These are sold in New Zealand and Australia under the PDL/Clipsal brand, compatible with the Iconic range of sockets. They have a bunch of other zigbee-compatible devices on sale (switches, sensors, etc) but these are the only ones I have installed at the moment.

### DIFF
--- a/devices/schneider_electric.js
+++ b/devices/schneider_electric.js
@@ -108,6 +108,25 @@ module.exports = [
         },
     },
     {
+        zigbeeModel: ['CH/DIMMER/1'],
+        model: '41EPBDWCLMZ/354PBDMBTZ',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2651, opened 2021-06-03
Improve cover expose

sjorge:
Screen Shot 2021-06-03 at 18 11 24

Currently it is expose as Binary and looks like this, given that it never actually ends up in the state for the device as it just sends a command, using an Enum of the available commands makes more sense.

https://github.com/Koenkk/zigbee-herdsman-converters/blob/c60f3df7e942d698ea378d7d1813dd5f488a6e41/converters/toZigbee.js#L337-L345

It now looks like

image

### DIFF
--- a/lib/exposes.js
+++ b/lib/exposes.js
@@ -301,7 +301,7 @@ class Cover extends Base {
         super();
         this.type = 'cover';
         this.features = [];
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2667, opened 2021-06-07
Drop tamper expose for some HEIMAN devices

sjorge:
I think this can also be removed for there gas sensors and possible motion/door sensors too but I don't have access to any of those to verify.


Koenkk:
Thanks!

Issue #2664, opened 2021-06-06
Axis Gear has incorrect battery info


sjorge:
https://support.helloaxis.com/en/articles/5209549-charging-battery-warnings-on-the-app

Both of mine always show around 60-68% but given the device has the don’t divide set, that would fall in the 30% range mentioned if the gear is set to the wrong mode, solar when AC or AC when solar.

This leads me to believe that:
- we **do** need to device the value
- we are missing a way to indicate the power mode

I noticed there is a genPowerCfg on ep 169 and ep1 (current), I will experiment with the other ep and see if that one is correct. My current running theory is ep1 = solar, ep169 = ac power.

If this is correct, I think we will need a easy wat to toggle between both configurations. Not sure how to handle that but let’s figure out if my theory is correct.

sjorge:
@Koenkk this should work righht?

```
mosquitto_pub -t zigbee2mqtt/bridge/request/device/bind -m '{"clusters":["genPowerCfg"],"from":"cover/livingroom/side1/196","to":"Coordinator", "skip_disable_reporting": true}'
```

<img width="408" alt="image" src="https://user-images.githubusercontent.com/379665/121039340-496dbf00-c7b1-11eb-934f-9ad30410f700.png">


```
Zigbee2MQTT:info  2021-06-07 14:56:26: MQTT publish: topic 'zigbee2mqtt/bridge/response/device/bind', payload '{"data":{"clusters":[],"failed":[],"from":"cover/livingroom/side1/196","to":"Coordinator"},"error":"Nothing to bind","status":"error"}'
Zigbee2MQTT:error 2021-06-07 14:56:26: Nothing to bind
```

Koenkk:
That should work indeed, don't understand why it doesn't, can you do some debugging here: https://github.com/Koenkk/zigbee2mqtt/blob/9bd46932c365e886ed0b1f9b8740d8cce632230c/lib/extension/bind.js#L181 ?

sjorge:
[...]

PR #2668, opened 2021-06-07
Don't divide battery for Axis Gear

sjorge:
Part of #2664, now it shows ~30% as described in the issue when using AC power with the device in the wrong mode. (Not found a mode switch yet)


Koenkk:
Thanks!

PR #2669, opened 2021-06-07
ADD Woox R7049

mattleh:
i created a converter for r7049 with the basic functionalities there are still some dp wich are not set


Koenkk:
Thanks!

PR #2672, opened 2021-06-08
ADD Iluminize 5120.1110

mattleh:
https://www.amazon.de/gp/product/B08S43X273
same as 5120.1100 but with/without N


Koenkk:
Thanks!

PR #2674, opened 2021-06-09
TuYa TS0505A_led RGB+CCT LED overhaul

iglu-sebastian:
This fixes some behavior of the bulb and supports the `color_mode` attribute used by HA.

Since this bulb behaves somewhat different than what we else have, the code looks a bit ugly in my opinion, but it works now in all modes and remembers even the brightness between `color_temp` and `color` mode correctly.

I'm open for corrections and enhancements before merging of course.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -2266,114 +2266,118 @@ const converters = {
         convertSet: async (entity, key, value, meta) => {
             if (key === 'brightness' && meta.state.color_mode == constants.colorMode[2] &&
                 !meta.message.hasOwnProperty('color') && !meta.message.hasOwnProperty('color_temp')) {
-                const payload = {level: value, transtime: 0};
+                const zclData = {level: Number(value), transtime: 0};
 
-                await entity.command('genLevelCtrl', 'moveToLevel', payload, utils.getOptions(meta.mapped, entity));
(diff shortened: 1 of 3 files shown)


converters/toZigbee.js
```diff
@@ -2295,25 +2333,45 @@ const converters = {
...
             await entity.command('lightingColorCtrl', 'tuyaRgbMode', {enable: 1}, {}, {disableDefaultResponse: true});
-            await entity.command('lightingColorCtrl', 'tuyaMoveToHueAndSaturationBrightness', payload, {disableDefaultResponse: true});
-            // transtime cannot be set on these devices. They seem to have a default one of about 1500ms!
-            return {state: {color_temp: value, brightness: payload.brightness}, readAfterWriteTime: payload.transtime * 100};
+            await entity.command('lightingColorCtrl', 'tuyaMoveToHueAndSaturationBrightness',
+                payload, utils.getOptions(meta.mapped, entity));
+
+            globalStore.putValue(entity, 'brightness', payload.brightness);
+
+            return {
```
Koenkk, line 2358: Can you also add the colorsync here? Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/932669961aa1f8c6d0735722178350eb884a9a3a/converters/fromZigbee.js#L564

Info about this: https://www.zigbee2mqtt.io/devices/81809_81813.html#notes

iglu-sebastian:
@Koenkk Made some changes, how's that now?

PR #2676, opened 2021-06-10
add Schneider Electric NHPB/SWITCH/1

roumano:


devices/schneider_electric.js
```diff
@@ -152,6 +152,18 @@ module.exports = [
             await reporting.onOff(endpoint);
         },
     },
+    {
+        zigbeeModel: ['NHPB/SWITCH/1'],
+        model: 'NHPB/SWITCH/1',
```
Koenkk, line 157: Please add the model as written on the device, I expect something like `U201SRY2KWZB` (since other Schneider devices uses such model).

roumano:
Look like it's `SM2021W18_1`
Or do you think it's `NNZ13936-00` Or do you think it's `NNZ13936-0000163`

picture

Koenkk:
Both model numbers yield no results on google, could it be `S520530` ? (https://www.se.com/fr/fr/product/S520530/wiser-odace---interrupteur--10---bluetooth---blanc/). If so, make sure to also change the device picture name in https://github.com/Koenkk/zigbee2mqtt.io/pull/790

roumano:
Hi,
S520530 is a the bluethooth one,
So it's the `S520530W` (https://www.se.com/fr/fr/product/S520530W/wiser-odace---bouton-poussoir---10a---zigbee---blanc/) , i need to change my code and PR again ? or do you manage ?

Koenkk:
Thanks!

PR #2683, opened 2021-06-12
Fix for malfunctions 'boost_heating_countdown', 'eco_mode' und 'boost_heating' on the BRT-100-TRV

vladi1234:


vladi1234:
I have picture for thermostat. And I've already done OCR for instruction manuals. Is that still needed?

Koenkk:
Thanks, no further doc updates are needed.

PR #2689, opened 2021-06-14
Fix for malfunction 'window_detection' on Moes BRT-100-TRV.

vladi1234:


converters/fromZigbee.js
```diff
@@ -3075,7 +3075,7 @@ const converters = {
             case tuya.dataPoints.moesSwindowDetectionFunktion_A2:
                 return {window_detection: value ? 'ON' : 'OFF'};
             case tuya.dataPoints.moesSwindowDetection:
-                return {window_detection: value ? 'ON' : 'OFF'};
+                return {window: value ? 'CLOSED' : 'OPEN'};
```
Koenkk, line 3078: I don't understand why this has been renamed, could you explain it?

vladi1234: These are two different functions.
1. Window detection can be switched on and off.
2. Shows whether Window is recognized as opened or closed. For control.

Bildschirmfoto vom 2021-06-14 18-22-44
Bildschirmfoto vom 2021-06-14 18-18-25

Koenkk:
Thanks!

PR #2694, opened 2021-06-16
Add support for toggle command on Legrand 067773

couloum:
From my understanding, old firmwares supported actions `on` and `off`,
while new firmwares only support action `toggle`.
I suppose that adding action `toggle` would allow to support all
firmware versions.

More details here:
https://github.com/Koenkk/zigbee2mqtt/issues/2399#issuecomment-850793482
https://github.com/Koenkk/zigbee2mqtt/issues/2399#issuecomment-862707311


Koenkk:
Thanks!

PR #2699, opened 2021-06-18
Added support for Namron dimmer model 4512726 and added whitelabel info

owangen:
Added support for the 4512726 Namron dimmer. Product description https://www.namron.com/products/namron-zigbee-4-i-1-vridimmer/.  Added whitelabel information for different Namron products I could find matching whitelabel information. Namron products sold in Norway are whitelabel of Sunricher.


Koenkk:
Thanks!

PR #2704, opened 2021-06-21
Add support for Sengled E1C-NB7

Leah-Anderson:
This adds support for the Sengled E1C-NB7. Unfortunately I can't find any documentation on Sengled's site for this plug but here's the documentation I could find for it:
- https://www.amazon.com/gp/product/B092DBFFBY
- https://fccid.io/2AGN8-E1CNB7

### DIFF
--- a/devices/sengled.js
+++ b/devices/sengled.js
@@ -1,5 +1,6 @@
 const exposes = require('../lib/exposes');
 const fz = {...require('../converters/fromZigbee'), legacy: require('../lib/legacy').fromZigbee};
+const tz = require('zigbee-herdsman-converters/converters/toZigbee');
 const ota = require('../lib/ota');
 const reporting = require('../lib/reporting');
 const extend = require('../lib/extend');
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

angeloMendozaGit:
Hello, I just added the plug to my Home Assistant setup. The on/off works, but the energy monitoring isn't. Also, instead of showing 0 for kWh and W, it shows 'None'. Is it also possible to get other entities like Amperes and Voltage?

Leah-Anderson:
From the user manual it says "Select the smart plug from Devices page and check 
average power and cumulative power consumption" so I don't think it supports amperage and voltage. Additionally I let the plug run overnight and I didn't see anything in the logs to indicate it was missing anything. 

As for why it says none... I saw that at first but it eventually reported a value after a few minutes.

angeloMendozaGit:
Turns out, I think it is the plug that is broken. I tried my other 3 plugs and everything is working great. Thanks for adding this.

Leah-Anderson:
Glad it's working!

PR #2706, opened 2021-06-21
Update philips.js

mmelb:
Add 4090631P9 and 4090130P9, both had XX...P7 listed already, copied those.
New to git, hopefully I did this right.

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -472,7 +472,7 @@ module.exports = [
         vendor: 'Philips',
         description: 'Hue Flourish white and color ambiance pendant light',
         meta: {turnsOffAtBrightness1: true},
-        extend: hueExtend.light_onoff_brightness_colortemp_color(),
+        extend: hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}),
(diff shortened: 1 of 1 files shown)


devices/philips.js
```diff
@@ -852,6 +861,15 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp_color(),
         ota: ota.zigbeeOTA,
     },
+    {
+        zigbeeModel: ['4090130P9'],
+        model: '4090130P9',
+        vendor: 'Philips',
+        description: 'Hue Sana',
+        meta: {turnsOffAtBrightness1: true},
+        extend: hueExtend.light_onoff_brightness_colortemp_color(),
```
Koenkk, line 870: Please add the color temp range for both, see https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html#31-retrieving-color-temperature-range-only-required-for-lights-which-support-color-temperature

mmelb: Both have 153-500. I think it should be added now.

Koenkk:
Thanks!

PR #2713, opened 2021-06-24
Adds CCT711119

rindlerblabla:
Probably supports power metering as well. Didn't manage to add it though.

https://zigbeealliance.org/zigbee_products/wiser-smart-plug-16a-2/


Koenkk:
Thanks!

Issue #2720, opened 2021-06-25
zigbeeModel.toLowerCase is not a function


GregoryGost:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/index.js

function addToLookup(zigbeeModel, definition)
function getFromLookup(zigbeeModel)

Fix:
Replace from

`zigbeeModel = zigbeeModel ? zigbeeModel.toLowerCase() : null;`

to

`if (zigbeeModel) {
    zigbeeModel = String(zigbeeModel);
    zigbeeModel.toLowerCase();
} else {
    zigbeeModel = null;
}`

Koenkk:
How to reproduce this bug? Also please make a PR (click the edit icon in the right top of https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/index.js)

GregoryGost:
> How to reproduce this bug? Also please make a PR (click the edit icon in the right top of https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/index.js)

Thanks for the answer!

I am studying the ZigBee module CC2530 and use such a ZigBee model `gost.dbze18ms1.01`
I am installing zigbee2mqtt in python virtual environment on raspberry pi.
Try my temporary converter with the current mater assembly.

const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
const tz = require('zigbee-herdsman-converters/converters/toZigbee');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const extend = require('zigbee-herdsman-converters/lib/extend');
const ep = exposes.presets;
const ea = exposes.access;

const bind = async (endpoint, target, clusters) => {
    for (const cluster of clusters) {
        await endpoint.bind(cluster, target);
    }
};

const dbze18ms1 = {
	zigbeeModel: ['gost.dbze18ms1.01'],
	model: 'DBZE18MS1-01',
	vendor: 'ImAVendor',
	description: 'DBZE18MS1-01',
	fromZigbee: [ fz.battery ],
	toZigbee: [],
	exposes: [ ep.battery(), ep.battery_voltage(), ep.linkquality() ],
};

module.exports = dbze18ms1;

PR #2725, opened 2021-06-26
Update index.js

GregoryGost:
zigbeeModel.toLowerCase is not a function
https://github.com/Koenkk/zigbee-herdsman-converters/issues/2720

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,12 @@ function arrayEquals(as, bs) {
 }
 
 function addToLookup(zigbeeModel, definition) {
-    zigbeeModel = zigbeeModel ? zigbeeModel.toLowerCase() : null;
+    if (zigbeeModel) {
+        zigbeeModel = String(zigbeeModel);
+        zigbeeModel.toLowerCase();
+    } else {
+        zigbeeModel = null;
+    }
     if (!lookup.has(zigbeeModel)) {
         lookup.set(zigbeeModel, []);
     }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -20,7 +20,12 @@ function arrayEquals(as, bs) {
 }
 
 function addToLookup(zigbeeModel, definition) {
-    zigbeeModel = zigbeeModel ? zigbeeModel.toLowerCase() : null;
+    if (zigbeeModel) {
+        zigbeeModel = String(zigbeeModel);
```
Koenkk, line 24: I don't understand why this is needed, in the example from https://github.com/Koenkk/zigbee-herdsman-converters/issues/2720#issuecomment-869010706 `zigbeeModel` is already a string.

GregoryGost: Maybe.
But only this helped in my case.

I can try to reproduce the error, but only on Monday.

GregoryGost: I think this is my mistake in studying the device.
I have chosen the wrong encoding of some fields from the genBasic cluster. Because of what the functions did not work correctly.
Only after making edits to the code and subsequent firmware of the device, z2m started working correctly.

`{BASIC, {ATTRID_BASIC_DATE_CODE, ZCL_DATATYPE_CHAR_STR, R, (void *)zclGenericApp_DateCode}}`

`ZCL_UINT8` instead of `ZCL_DATATYPE_ENUM8`

Correct `ZCL_DATATYPE_ENUM8`

With the edits I made to the converter file index, I was able to spot this error.

Accordingly, whether to make these changes to the master branch or not is up to you.

Thank you for your responsiveness!

Koenkk: Good, since the `zigbeeModel` should always be a string this is not needed, I will close it.

PR #2727, opened 2021-06-26
Add power monitoring feature to SP2600ZB

charlesdagenais:
Sinope SP2600ZB have power monitoring feature. 
This enable it and add power in the reporting state


Koenkk:
Thanks!

PR #2732, opened 2021-06-27
Support 2-gang mode for LK Fuga wireless battery switch (550D6001)

pklokke:
Found the reason for the difference in operation, the wiser system guide outlines how to set 1-gang vs. 2-gang via button press.

Unfortunately, I don't have the gateway so I cannot reverse engineer which attribute they use to set it, but I could add the instructions to the pairing info at least.

### DIFF
--- a/devices/schneider_electric.js
+++ b/devices/schneider_electric.js
@@ -297,7 +297,7 @@ module.exports = [
         fromZigbee: [fz.command_on, fz.command_off, fz.command_move, fz.command_stop],
         toZigbee: [],
         endpoint: (device) => {
-            return {'s1': 21, 's2': 22};
+            return {'top': 21, 'bottom': 22};
         },
         meta: {multiEndpoint: true},
         exposes: [e.action(['on_*', 'off_*', 'brightness_*'])],
--- a/devices/schneider_electric.js
+++ b/devices/schneider_electric.js
@@ -296,11 +296,17 @@ module.exports = [
         description: 'LK FUGA wiser wireless battery 4 button switch',
         fromZigbee: [fz.command_on, fz.command_off, fz.command_move, fz.command_stop],
         toZigbee: [],
-        exposes: [e.action(['on', 'off', 'brightness_*'])],
+        endpoint: (device) => {
+            return {'s1': 21, 's2': 22};
+        },
+        meta: {multiEndpoint: true},
+        exposes: [e.action(['on_*', 'off_*', 'brightness_*'])],
         configure: async (device, coordinatorEndpoint, logger) => {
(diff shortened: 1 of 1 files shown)


devices/schneider_electric.js
```diff
@@ -296,11 +296,17 @@ module.exports = [
         description: 'LK FUGA wiser wireless battery 4 button switch',
         fromZigbee: [fz.command_on, fz.command_off, fz.command_move, fz.command_stop],
         toZigbee: [],
-        exposes: [e.action(['on', 'off', 'brightness_*'])],
+        endpoint: (device) => {
+            return {'s1': 21, 's2': 22};
```
Koenkk, line 300: `s1` and `s2` are non valid endpoint names, can you change this to `l1` and `l2`? These are allowed: https://github.com/Koenkk/zigbee2mqtt/blob/9bd46932c365e886ed0b1f9b8740d8cce632230c/lib/util/utils.js#L34

pklokke: I can do that, no problem.  Alternatively, would `top` and `bottom` be preferable?  This is how the button pairs physically relate to the remote unit.

Koenkk: Yes, `top` and `bottom` is better.

pklokke: Great, I've now used those.

Koenkk:
Thanks!

akafester:
I don't know if this is the right place to ask, but I have one og these, and it doesn't report any actions.
Is there a step I'm missing?

I'm on the latest Z2M version.

pklokke:
> I don't know if this is the right place to ask, but I have one of these, and it doesn't report any actions, just 'empty_string'.
> Is there a step I'm missing?
> 
> I'm on the latest Z2M version.

Probably best to log an issue, as we can't action it from this PR.  I'll update my Z2M and take a look

DennisDalgaard:
> Found the reason for the difference in operation, the wiser system guide outlines how to set 1-gang vs. 2-gang via button press.
> 
> Unfortunately, I don't have the gateway so I cannot reverse engineer which attribute they use to set it, but I could add the instructions to the pairing info at least.

Hi pklokke,
I've just added the 550D6001 to Z2M v. 1.22.0. My experience is that the top and buttom buttons are sending the same message (top on/top off). Have you found a solution for that? I have a Wiser gateway available for testing if needed.
BR. Dennis

akafester:
Just need to hold button 1 and 4 down at the same time for 10 seconds. The red light should blink and all 4 buttons should work independently.

DennisDalgaard:
> 

That's amazing! Works perfect. Thank you so much for fast reply.

Issue #1986, opened 2020-12-28
Add Xiaomi Aqara D1 switch QBKG26LM decoupled mode 


chankalson:
The Aqara D1 switch QBKG26LM is a 3-gang switch (with Neutral wire), however decoupled mode is not available even though the same 3-gang QBKG25LM (no neutral) has it.

ERROR: No converter available for 'operation_mode' ([object Object])

Koenkk:
First step would be to sniff the traffic with the original xiaomi gateway to figure out what to send. https://www.zigbee2mqtt.io/how_tos/how_to_sniff_zigbee_traffic.html .

Next add `tz.xiaomi_switch_operation_mode` to the `toZigbee` array of the QBKG26LM in devices.js and modify https://github.com/Koenkk/zigbee-herdsman-converters/blob/b441c7e3a2c040594e62c504c6466db0e904bd96/converters/toZigbee.js#L1436 accordingly (to what you sniffed)

chankalson:
Thanks for the response.

I'm using the Zigbee2mqtt add-on in Home Assistant. May I know where is the toZigbee.js file located?

Thanks again.

Koenkk:
For the addon it's a bit more complicated. SSH into the machine, figure out the container ID with `docker ps -a`, ssh into with `docker exec -it [CONTAINER_ID] sh`. The `toZigbee.js` should be under `node_modules/zigbee-herdsman-converters/converters/toZigbee.js`.

chankalson:
Hi @Koenkk , Happy New Year to you!

I've captured the traffic from the QBKG26LM to Aqara Hub. 
[QBKG26LM.zip](https://github.com/Koenkk/zigbee-herdsman-converters/files/5768381/QBKG26LM.zip)

I pressed each button on the QBKG26LM twice, thus turning it ON->OFF starting from left to right. After that I turned on the Wireless Switch function for all 3 buttons from the Mi Home app, and pressed each button once, from left to right.

May I know what do I need to modify in devices.js and toZigbee.js?

Thanks a lot for your help.

Koenkk:
@chankalson I need a sniff off setting the decoupled mode from the Xiaomi gateway (keep the sniff as short as possible, so start sniffing right before you change the decoupled mode). Also share your network key

chankalson:
@Koenkk, here's the new sniff file 
[...]

Issue #1616, opened 2020-10-01
aqara QBKG26LM power measurement support


lovelylain:
I tried the latest QBKG26LM configuration, it has support onOff action and report, can someone support the power measurement report?

Koenkk:
@sloopee @lovelylain do you see any `No converter available...` message from this device when debug logging is enabled and attaching a load to the switch?

To enable debug logging set in `configuration.yaml`:
```yaml
advanced:
  log_level: debug
```

sloopee:
debug log level already set up, no "no converter available...." seen. I guess it thinks it has the right "drivers" for it since it found a match but the code written for the switch currently doesn't have power measurement in it. I saw you write the following lines for this exact switch on another thread previously and the feedback was that it worked well. I'm trying to put the exact code into the "devices.js" file to see if it will work however i can't find the devices.js file.

{
    zigbeeModel: ['lumi.switch.n3acn3'],
    model: 'QBKG26LM',
    vendor: 'Xiaomi',
    description: 'Aqara D1 3 gang smart wall switch (with neutral wire)',
    supports: 'on/off, power measurement',
    fromZigbee: [fz.on_off, fz.xiaomi_power],
    toZigbee: [tz.on_off],
    meta: {multiEndpoint: true},
    endpoint: (device) => {
        return {'left': 1, 'center': 2, 'right': 3, 'system': 1};
    },
    onEvent: xiaomi.prevent_reset,
},

lovelylain:
@sloopee you should try the latest z2m version, which has merged PR #1525 , it has support on/off send and receive.
@Koenkk There is no zigbee msg received after pairing succeed. The on/off event of physical buttons are solved with `bind`, i don't know if power measurement can solve with bind too, I tried bind genAnalogInput but not working.

sloopee:
I'm already on the latest z2m version and like you said it only supports on.off without power.

Koenkk:
Does this device supports powers measurements when paired with the original Xiaomi gateway?

sloopee:
yep it does, all the aqara D1 with neutral switches can support power measurement as well.

Koenkk:
@zhongfu could you maybe help with this? (do you have any D1 with neutral switch?)

zhongfu:
> @zhongfu could you maybe help with this? (do you have any D1 with neutral switch?)

I don't have any Aqara neutral switches, unfortunately :/

[...]

PR #2734, opened 2021-06-28
Power reporting and decoupled mode for Aqara QBKG26LM

ruifung:
Apparently the genAnalogInput cluster isn't present to be interviewed. Though it's sending attribute reports for it and attempting to GET the attribute just gives a no response error. Hence the ea.STATE and no tz.xiaomi_power here. (https://github.com/Koenkk/zigbee-herdsman-converters/issues/1616)

Also the decoupled mode changes from https://github.com/Koenkk/zigbee-herdsman-converters/issues/1986#issuecomment-860604015 , disableResponse due to observed behavior where it wont send a response, though it works, causing it to repeat configuration - and turning the switch off everytime z2m restarts.

This has been tested on my local installation and power reporting works, as does decoupled mode actions. Though it will always send actions even when not decoupled.

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -416,7 +416,7 @@ module.exports = [
         vendor: 'Xiaomi',
         description: 'Aqara D1 3 gang smart wall switch (with neutral wire)',
         extend: extend.switch(),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #2735, opened 2021-06-28
Fix broken converter implementations for xiaomi-aqara decoupled control.

ruifung:
Fix the very obviously broken implementations of these converters.

fz.xiaomi_operation_mode_opple
- No longer tries to resolve an impossible mapping with a hardcoded string.
- Actually reads the value into an endpoint specific state value.
- Actually uses the endpoints() from the definition to load endpoint names.

tz.xiaomi_switch_operation_mode
- Added simpler syntax for setting, dropping the extra object / state wrapper. Old syntax still accepted to not break anything.
- No longer updates the operation_endpoint state attribute as it's not suitable for multi-endpoint devices.
- Uses endpointName from meta for actual endpoint name if it is a multiEndpoint definition.
- convertGet was failing due to not sending an array.

Fixed for and tested on QBKG26LM and QBKG24LM

This will require updating documentation.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4515,22 +4515,10 @@ const converters = {
                 0x01: 'control_relay',
                 0x00: 'decoupled',
             };
-            if (meta.multiEndpoint) {
-                const endpointNames = model.endpoint(meta.device);
-                for (const [name, id] of Object.entries(endpointNames)) {
-                    if (id === msg.endpoint.ID) {
-                        const mode = mappingMode[msg.data['512']];
-                        const payload = {};
-                        payload[`operation_mode_${name}`] = mode;
-                        return payload;
-                    }
-                }
-            } else {
-                const mode = mappingMode[msg.data['512']];
-                const payload = {};
-                payload[`operation_mode`] = mode;
-                return payload;
-            }
+            const mode = mappingMode[msg.data['512']];
+            const payload = {};
+            payload[postfixWithEndpointName('operation_mode', msg, model)] = mode;
+            return payload;
         },
     },
     qlwz_letv8key_switch: {
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4482,23 +4482,21 @@ const converters = {
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
             const payload = {};
-            if (['QBKG04LM', 'QBKG11LM', 'QBKG21LM'].includes(model.model)) {
+
+            if (!model.meta.multiEndpoint) {
                 const mappingMode = {0x12: 'control_relay', 0xFE: 'decoupled'};
-                const key = '65314';
+                const key = 0xFF22;
                 if (msg.data.hasOwnProperty(key)) {
                     payload.operation_mode = mappingMode[msg.data[key]];
                 }
-            } else if (['QBKG03LM', 'QBKG12LM', 'QBKG22LM'].includes(model.model)) {
-                const mappingButton = {'65314': 'left', '65315': 'right'};
+            } else {
+                const mappingButton = {0xFF22: 'left', 0xFF23: 'right'};
                 const mappingMode = {0x12: 'control_left_relay', 0x22: 'control_right_relay', 0xFE: 'decoupled'};
                 for (const key in mappingButton) {
                     if (msg.data.hasOwnProperty(key)) {
-                        const mode = mappingMode[msg.data[key]];
-                        payload[`operation_mode_${mappingButton[key]}`] = mode;
+                        payload[`operation_mode_${mappingButton[key]}`] = mappingMode[msg.data[key]];
                     }
                 }
-            } else {
-                throw new Error('Not supported');
             }
 
             return payload;
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4509,7 +4509,6 @@ const converters = {
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
             if (!msg.data.hasOwnProperty('512')) {
-                console.log('aqaraOpple missing 512 property.', msg.data);
                 return;
             }
             const mappingMode = {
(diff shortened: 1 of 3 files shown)


ruifung:
On a side note, this means that the fact that the QBKG26LM will emit events even in relay mode is less of a consequence since you can *actually* read the state now.

converters/toZigbee.js
```diff
@@ -1763,8 +1763,16 @@ const converters = {
                 return {state: {[`operation_mode${button !== 'single' ? `_${button}` : ''}`]: value.state}};
             } else if (['QBKG25LM', 'QBKG26LM', 'QBKG39LM'].includes(meta.mapped.model)) {
                 const lookupState = {control_relay: 0x01, decoupled: 0x00};
-                await entity.write('aqaraOpple', {0x0200: {value: lookupState[value.state], type: 0x20}}, manufacturerOptions.xiaomi);
-                return {state: {operation_mode: value.state}};
+                // Support existing syntax of a nested object just for the state field. Though it's quite silly IMO.
+                const targetValue = value.hasOwnProperty('state') ? value.state : value;
+                await entity.write('aqaraOpple', {0x0200: {value: lookupState[targetValue], type: 0x20}}, manufacturerOptions.xiaomi);
+                if (meta.mapped.meta.multiEndpoint) {
```
Koenkk, line 1769: I don't believe this `if` branch is needed. Zigbee2MQTT will automatically add the endpoint to all properties when setting via e.g. `zigbee2mqtt/my_ QBKG26LM/left/set`

converters/fromZigbee.js
```diff
@@ -4508,22 +4508,29 @@ const converters = {
         cluster: 'aqaraOpple',
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
-            const mappingButton = {
-                1: 'left',
-                2: 'center',
-                3: 'right',
-            };
+            if (!msg.data.hasOwnProperty('512')) {
+                console.log('aqaraOpple missing 512 property.', msg.data);
```
Koenkk, line 4512: Use `meta.logger` instead.

converters/fromZigbee.js
```diff
@@ -4508,22 +4508,29 @@ const converters = {
...
                 0x01: 'control_relay',
                 0x00: 'decoupled',
             };
-            for (const key in mappingButton) {
-                if (msg.endpoint.ID == key && msg.data.hasOwnProperty('512')) {
-                    const payload = {};
-                    const mode = mappingMode['512'];
-                    payload[`operation_mode_${mappingButton[key]}`] = mode;
-                    return payload;
+            if (meta.multiEndpoint) {
```
Koenkk, line 4519: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/2735/files#r660808325

converters/toZigbee.js
ruifung: Which is wierd, because that's not what I observed was happening with the original code. 
It resulted in a state of operation_mode instead of say, operation_mode_left

converters/fromZigbee.js
ruifung: Oops. Didn't mean to leave that in, it's a debug logging statement because I'm pretty sure logging the absence of a property there is just log spam under normal usage.

converters/toZigbee.js
ruifung: @Koenkk I double checked against the on_off converters, it does more or less the same thing there, so yeah. Changed the code to use the helper function from util to match.

Koenkk: That is indeed weird, I can't remembers why https://github.com/Koenkk/zigbee2mqtt/blob/9bd46932c365e886ed0b1f9b8740d8cce632230c/lib/extension/publish.js#L232 is needed, can you try removing it?

ruifung: Introduced here. https://github.com/Koenkk/zigbee2mqtt/commit/a77e8578688a394a3bd34fd179b983f42204a219

ruifung: Though personally, I don't think the inconsistency in handling that in z2m / herdsman-converters is within the scope of this PR though.

Koenkk: Made this consistent now in the latest dev, this should be handled in z2m and not in herdsman converters: https://github.com/Koenkk/zigbee2mqtt/commit/1985e2451797adbd91496012e869e151aa0fe7f6

ruifung: In which case, now you'd need to go through all the tz converters and change those that explicitly return a suffixed endpoint... or ignore properties that are already suffixed. tz.on_off for example, returns it suffixed already, that change will now result in... double suffixing?

Also, what about the fz path which is still suffix-ed manually? Because receive.js sure as heck doesn't handle that.

Koenkk: Let's keep the discussion of https://github.com/Koenkk/zigbee2mqtt/commit/1985e2451797adbd91496012e869e151aa0fe7f6#r52979633 here.

I'm wondering if we can make this converter more consistent:
- QBKG24LM: `zigbee2mqtt/FRIENDLY_NAME/system/set` -> `zigbee2mqtt/FRIENDLY_NAME/set` should also work (since it will use the first endpoint). Do you have this device to test with?
- QBKG25LM: we can select the correct endpoint in the converter based on the `button`.

ruifung: For QBKG25LM, QBKG26LM, QBKG39LM, these run the same code path where it sends to aqaraOpple cluster on the same endpoint as the onOff clusters. At least for what I can see in the converter (I only have the 26), so that's not a problem.

Maybe if it can detect what button from the *endpoint* and then just send to the 1st endpoint instead, that should make the converter's function overall more consistent between models, hiding that implementation detail from the user.

(Also, I have no idea how exposes will work for this wierd setup)

Koenkk: Thanks, reviewed the PR and looks good to go, can I merge it?

Regarding docs, I will remove all these sections: https://www.zigbee2mqtt.io/devices/QBKG04LM.html#decoupled-mode, since you've added them the `expose` now (great!) docs will be automatically generated under https://www.zigbee2mqtt.io/devices/QBKG04LM.html#exposes

ruifung: Yeah, go ahead. So far I've been running it on my local installation and it appears to be working fine.

Koenkk: Big thanks for cleaning this up :)

PR #2738, opened 2021-06-29
Add running_state reporting to Sinope thermostat

embak:
HA MQTT generates  warnings for Sinope thermostat, "running_state" not defined. This issue affect all Sinope thermostats.

This commit solve this by:
1. adding reporting of thermostatRunningState.
2. setting running_state according to piHeathingDemand update in legacy.sinope_thermostat_state(). This is required for thermostat firmware version that don't support running_state reporting...


Koenkk:
Thanks!

PR #2739, opened 2021-06-29
Updated Meter Reader to support Phase B and C incl new converter

TheStigh:
added `metering_datek` as per your change + updated code to datek.js for supporting Phase B & C.


Koenkk:
Thanks!

PR #2740, opened 2021-06-29
Parkside Smart Watering Timer 

mgrom:
https://github.com/Koenkk/zigbee2mqtt/issues/7695
Partial support for Parkside Smart Watering Timer 
It's partial because I wasn't able to force the device to start reporting correct state. For now there is possibility to Open/Close the valve and set number of minutes for auto close function.


devices/lidl.js
```diff
@@ -216,4 +217,16 @@ module.exports = [
             device.getEndpoint(1).saveClusterAttributeKeyValue('lightingColorCtrl', {colorCapabilities: 29});
         },
     },
+    {
+        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_htnnfasr'}],
+        model: 'PSBZS A1',
+        vendor: 'Parkside',
```
Koenkk, line 223: Please change this to Lidl and change the description to `'Parkside smart watering timer'`

devices/lidl.js
```diff
@@ -216,4 +217,16 @@ module.exports = [
...
+        description: 'Smart Watering Timer',
+        supports: 'switch, timer',
+        fromZigbee: [fz.ignore_basic_report],
+        toZigbee: [tz.on_off, tz.lidl_watering_timer],
+        onEvent: tuya.onEventSetTime,
+        configure: async (device, coordinatorEndpoint, logger) => {},
+        exposes: [],
```
Koenkk, line 230: I would expect `e.switch()` here.

mgrom: I haven't add it because the device is not reporting to coordinator, so after auto-off state in zigbee2mqtt is still ON. 
I'll add it anyway, just letting you know.

devices/lidl.js
mgrom: done

devices/lidl.js
mgrom: done

Koenkk:
Thanks!

PR #2744, opened 2021-06-29
Fix Ikea Styrbar Remote

asgothian:
Exposes and fz coding for Ikea Syrbar Remote (STYRBAR Remote N2, Model W2049) do not match.

fz generates action messages for 'on', 'off', 'brightness_move_up', 'brightness_move_down', 'brightness_stop' instead of the currently exposed messages, causing systems using the exposes to not properly receive messages.

A.


Koenkk:
Thanks!

PR #2745, opened 2021-06-29
Improves EBL detection

pedrolamas:
This will improve the EBL detection to better handle situations like the one reported here: https://github.com/Koenkk/zigbee-OTA/pull/46


Koenkk:
Thanks!

PR #2751, opened 2021-07-01
Added another model ID

rindlerblabla:
Obviously the second generation Wiser light switch has the same zigbee model ID as the first generation. 
https://www.se.com/ww/en/product/CCT5011-0002/wiser-micro-module-light-switch/


Koenkk:
Thanks!

PR #2755, opened 2021-07-02
sinope: add WL4200S sensor

andrew-d:
This adds the WL4200S sensor, which is the same as the WL4200 but has a remote sensor (i.e. "extension cord") that you can put somewhere without putting the whole device there.

Product page:
https://www.sinopetech.com/en/product/smart-water-leak-detector-with-sensor-zigbee/

I've been running this as an external converter for the past few weeks and it seems to work fine.


Koenkk:
Thanks!

andrew-d:
> Thanks!

And thank you as well!

PR #2756, opened 2021-07-02
Set {colorTempRange: [153, 370]}  for Ecosmart  model: 'A9A19A60WESDZ02' and model: 'A9BR3065WESDZ02'

tteck:
Set the correct color temperature range. Corrects Home Assistants slider.

### DIFF
--- a/devices/ecosmart.js
+++ b/devices/ecosmart.js
@@ -6,14 +6,14 @@ module.exports = [
         model: 'A9A19A60WESDZ02',
         vendor: 'EcoSmart',
         description: 'Tuneable white (A19)',
-        extend: extend.light_onoff_brightness_colortemp(),
+        extend: extend.light_onoff_brightness_colortemp({colorTempRange: [153, 370]}),
     },
     {
         zigbeeModel: ['Ecosmart-ZBT-BR30-CCT-Bulb'],
         model: 'A9BR3065WESDZ02',
         vendor: 'EcoSmart',
         description: 'Tuneable white (BR30)',
-        extend: extend.light_onoff_brightness_colortemp(),
+        extend: extend.light_onoff_brightness_colortemp({colorTempRange: [153, 370]}),
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

jwoodard80:
This most likely needs to be duplicated on the others in this as well. I just simply edited this one device.

jwoodard80:
@Koenkk  .... I am curious.... 

Based on what I read here https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html from below. Is it not possible to somehow poll the device on addition and set the min/max value? 

Perhaps it is too much work, just was wondering if I had missed something. 

Thanks


3.1 Retrieving color temperature range (only required for lights which support color temperature) #
If your device is a light and supports color temperature you need to define the color temperature range. This range indicates the minimum and maximum color temperature value the light supports. This can be retrieved from the light by sending to zigbee2mqtt/DEVICE_FRIENDLY_NAME/set with payload {"read": {"cluster": "lightingColorCtrl", "attributes": ["colorTempPhysicalMin", "colorTempPhysicalMax"]}}

The result will be logged to the Zigbee2MQTT log, e.g.

Zigbee2MQTT:info  2021-03-21 21:10:40: Read result of 'lightingColorCtrl': {"colorTempPhysicalMin":153,"colorTempPhysicalMax":500}
In the above example set colorTempRange to {colorTempRange: [153, 500]}, e.g.:

const definition = {
    zigbeeModel: ['myZigbeeModel'],
    model: 'myModel',
    vendor: 'myVendor',
    description: 'Super bulb',
    extend: extend.light_onoff_brightness_colortemp({colorTempRange: [153, 500]}), // <---
},
In case none of the existing converters fit you can add custom ones, external converter example for this can be found here.

Koenkk:
Yes this is possible, but we want to know it before runtime so it can also be in the docs.

PR #2758, opened 2021-07-03
Remove explicit post-fixing of return states with meta.endpoint_name in toZigbee converters.

ruifung:
See https://github.com/Koenkk/zigbee2mqtt/commit/1985e2451797adbd91496012e869e151aa0fe7f6


Koenkk:
Thanks!

PR #2760, opened 2021-07-03
Add battery voltage support for 3RSS009Z

pierlon:
This device reports the `genPowerCfg` cluster but I'm not sure if I've configured the device to expose it correctly.


Koenkk:
Thanks!

PR #2770, opened 2021-07-06
Add support Technicolor XHK1-TC TKA105

erwintenbrink:


devices/Technicolor
```diff
@@ -0,0 +1,43 @@
...
+const reporting = require('zigbee-herdsman-converters/lib/reporting');
+const extend = require('zigbee-herdsman-converters/lib/extend');
+const e = exposes.presets;
+const ea = exposes.access;
+
+module.exports = [
+    {
+        zigbeeModel: ['TKA105'],
+        model: 'XHK1-TC',
+        vendor: 'Technicolor',
```
Koenkk, line 13: This looks to be an xfinity product (https://www.xfinity.com/support/articles/installing-batteries-xhk1-ue-keypad), can you move it to the correct file and change the vendor here?

erwintenbrink: Shure if you want so.

At the first connection the pad sent technicolor as manufacturer in zigbee2mqtt. If it is no problem that the manufacturer in the .js files is different compared to what the message to zigbee2mqtt says please let me know and I'll change it.

Koenkk: Yes, please change to xfinity, it's preferred to have something recognizable.

erwintenbrink:
Added in the Xfinity file and created a new pull request 

https://github.com/Koenkk/zigbee-herdsman-converters/pull/2775

alekslyse:
I have exactly this version that report as Technicolor @erwintenbrink did you get yours to work - did you have to press something? Mine show up, the ID is correct, but vendor missing. I tried to create an external converter with Technicolor, but that didn't get picked up (probably cause of the same vendor id?)

PR #2771, opened 2021-07-07
Bayview252 patch2 - add Mercator ikuü 7W downlight to TuYa 0505B model.

bayview252:
Added fingerprint, whitelabel and appended some info to description as 0505B is described as an led strip controller.


Koenkk:
Thanks!

PR #2772, opened 2021-07-07
Makes ID-EU20FW09 whiteLabel of ID-UK21FW09

pedrolamas:
Following up on this: https://github.com/Koenkk/zigbee2mqtt.io/pull/716#issuecomment-875450341

As the manufacturer is HZC, that matches the Iolloi definition, so I'm moving the white-label association.


Koenkk:
Thanks!

PR #2775, opened 2021-07-08
Add XHK1-TC support Xfinity

erwintenbrink:
Add TKA105 support

### DIFF
--- a/devices/xfinity.js
+++ b/devices/xfinity.js
@@ -40,4 +40,37 @@ module.exports = [
             }
         },
     },
+    {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

robertalexa:
@erwintenbrink Can you please confirm if the device behaviour is the same as https://www.zigbee2mqtt.io/devices/URC4450BC0-X-R.html

As in, you need to send a response with the transaction to confirm the received arming command?

If so, I will go and update the documentation. Thanks a lot

PR #2776, opened 2021-07-08
Aqara E1 QBKG39LM power_outage_memory

kirovilya:
common for whole device

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -1202,8 +1202,7 @@ module.exports = [
                 .withFeature(exposes.enum('state', ea.STATE_SET, ['control_relay', 'decoupled']))
                 .withEndpoint('right'),
             e.action(['single_left', 'double_left', 'single_right', 'double_right', 'single_both', 'double_both']),
-            e.power_outage_memory().withEndpoint('left'),
-            e.power_outage_memory().withEndpoint('right'),
+            e.power_outage_memory(),
         ],
         onEvent: preventReset,
         configure: async (device, coordinatorEndpoint, logger) => {


devices/xiaomi.js
```diff
@@ -1202,8 +1202,7 @@ module.exports = [
                 .withFeature(exposes.enum('state', ea.STATE_SET, ['control_relay', 'decoupled']))
                 .withEndpoint('right'),
             e.action(['single_left', 'double_left', 'single_right', 'double_right', 'single_both', 'double_both']),
-            e.power_outage_memory().withEndpoint('left'),
-            e.power_outage_memory().withEndpoint('right'),
+            e.power_outage_memory(),
```
Koenkk, line 1205: Involving @ruifung  here.

I don't understand this, does this mean the power outage memory of the buttons cannot be configured individually?

kirovilya: yes, the setting is general for the entire device and does not apply to a specific button
image

kirovilya: changing this attribute affects both switch channels.

ruifung: > 
> 
> Involving @ruifung here.
> 
> I don't understand this, does this mean the power outage memory of the buttons cannot be configured individually?

I don't have this model of switch, so I can't say much but...

To my view the better question is, which endpoints is this cluster actually on though.

If it currently works for the entire switch regardless of if it's sending to endpoint 1 or 2, then exposing it without a specific endpoint would be an accurate representation.

If it *can* be controlled individually, and the xiaomi app/gateway just does magic behind the scenes then it's fine to leave it as is.

PR #2777, opened 2021-07-09
Change Aqara E1 expose for operation_mode to enum from composite

ruifung:
Missed this in the previous PR, the composite expose is no longer nessacary with the changes in the converters.


Koenkk:
Thanks!

PR #2780, opened 2021-07-09
Add color temp for philips hue 9290022943

sodoto:
Add color temp and correct range for Philips Hue White E12 model '9290022943'.


Koenkk:
Thanks!

PR #2789, opened 2021-07-12
Add the 1-spot version of the Philips Hue Adore spotlight

colmbuckley:
There's also a 1-spot version of the Adore spotlight (2-spot and 3-spot versions already supported), with model number 3417831P6. Works exactly as expected when adopted with an external converter.


Koenkk:
Thanks!

PR #2791, opened 2021-07-14
Added support for Vendor Onesti Products AS

eirikd:
Added vendor Onesti Product AS
zigbee module easyCodeTouch_v1


Koenkk:
Thanks!

PR #2794, opened 2021-07-14
Moes ZLD-RCW and Miboxer FUT039Z LED Controller

kirovilya:


converters/toZigbee.js
```diff
@@ -5368,6 +5368,23 @@ const converters = {
...
     },
+    tuya_do_not_disturb: {
+        key: ['do_not_disturb'],
+        convertSet: async (entity, key, value, meta) => {
+            await entity.command('lightingColorCtrl', 'tuyaDoNotDisturb', {enable: value ? 1 : 0});
+            return {state: {do_not_disturb: value}};
+        },
+    },
+    tuya_color_power_on_behavior: {
+        key: ['tuya_color_power_on_behavior'],
```
Koenkk, line 5379: Having `color_power_on_behavior` sounds better in my opinion.

converters/toZigbee.js
```diff
@@ -5368,6 +5368,23 @@ const converters = {
...
+        convertSet: async (entity, key, value, meta) => {
+            const lookup = {'initial': 0, 'previous': 1, 'cutomized': 2};
+            utils.validateValue(value, Object.keys(lookup));
+            await entity.command('lightingColorCtrl', 'tuyaOnStartUp', {
+                mode: lookup[value]*256, data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]});
+            return {state: {power_on_behavior: value}};
```
Koenkk, line 5385: The returned property here should match the key, (e.g. `return {state: {color_power_on_behavior: value}};`

converters/toZigbee.js
sjorge: Could we have this under `color_options` ? https://github.com/Koenkk/zigbee-herdsman-converters/blob/1c60448990c69d0737268381fcac6e8ce5253748/converters/toZigbee.js#L100 we use that for some of the generic lightingColorCtrl options, so having it use the same key would be more consistant?

kirovilya: @sjorge look at https://github.com/Koenkk/zigbee-herdsman/pull/390
it is not common case - it it tuya commands

converters/toZigbee.js
kirovilya: renamed

converters/toZigbee.js
Koenkk: Both are good, since I don't expect any other options this is fine.

PR #2798, opened 2021-07-16
Disable light effect on ID-UK21FW09

pedrolamas:
The Iolloi ID-UK21FW09 does not support any light effects, so disabling this now.


Koenkk:
Thanks!

PR #2804, opened 2021-07-17
Fix multiEndpoint check for empty meta xiaomi_operation_mode_basic

nurikk:
If `model.meta` is `undefined`, then bad things happens


Koenkk:
Thanks!

PR #2807, opened 2021-07-17
Added support for Hue Fugato 5063230P7, 5063330P7 and 5063430P7

owangen:
They are the same as *1P7, but has black finish and appears as different models.


owangen:
Fixed temperature range also

Koenkk:
Thanks!

PR #2811, opened 2021-07-18
Fix TS0601 (_TZE200_azqp6ssj)

tgorochowik:
The device is not supported correctly, it looks as though the issue is
an invalid key used when exporting the module (`Id` instead of `ID`)


Koenkk:
Thanks!

PR #2818, opened 2021-07-21
Livolo Dimmer and Switch bug fixes

Piscator74:
Already commented on these bugs quite a while back and finally found some time to commit the fixes.
See discussions:
* https://github.com/Koenkk/zigbee2mqtt/issues/5516#issuecomment-811482988
* https://github.com/Koenkk/zigbee2mqtt.io/pull/607#issuecomment-812860097

For the update in fromZigbee.js:
I left the original code (msg.data[10] === 7), but that does not work for me.

For the updates in devices.js:
That fixes exceptions related to endpoints (and actually makes the dimmer work in combination with the fix above).

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2246,6 +2246,9 @@ const converters = {
                 if (msg.data[10] === 7) {
                     const status = msg.data[14];
                     return {state: status & 1 ? 'ON' : 'OFF'};
+                } else if (msg.data[10] === 13) {
+                    const status = msg.data[13];
+                    return {state: status & 1 ? 'ON' : 'OFF'};
                 } else if (msg.data[10] === 5) { // TODO: Unknown dp, assumed value type
                     const value = msg.data[14] * 10;
                     return {
--- a/devices/livolo.js
+++ b/devices/livolo.js
@@ -52,6 +52,9 @@ module.exports = [
         toZigbee: [tz.livolo_socket_switch_on_off],
         extend: extend.switch(),
         configure: poll,
+        endpoint: (device) => {
+            return {'left': 6, 'right': 6};
+        },
         onEvent: async (type, data, device) => {
             if (type === 'stop') {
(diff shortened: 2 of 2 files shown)


Koenkk:
Thanks!

resh1309:
Can I ask a question regarding the Livolo switch. Will this fix be working only with zigbee2mqtt or do you need the Livolo gateway to connect to zigbee2mqtt?

Piscator74:
The livolo dimmer needs to pair one time with the original livolo gateway. After that, you can pair it with zigbee2mqtt and you do not need the livolo gateway anymore. This is an open issue and is discussed in the links above.

The livolo 2-gang switch should work flawlessly.

Note that you cannot connect a Livolo gateway to zigbee2mqtt. That works only with the Livolo app and is proprietary.

resh1309:
Thank you for your reply! 

I bought the following livolo switch
https://www.livolo.nl/a-64026688/modules-sr-touchschakelaar/livolo-module-sr-zwart-dimmer-universeel-zigbee-wifi-livolo-app/#description

Any experience on this one? 

When the livolo switch is paired with the livolo hub, you say that it is possible after that to pair with zigbee2mqtt. Is that only on channel 26 or is that then not necessary anymore to add in yml script? 

You also refer to the open issue 1 time registration on luvolo hub,  those are still in discussion? Do you mean you are still aiming to get livolo switches up and running without 1 time registration on livolo hub?

Piscator74:
I have a similar dimmer and it works great. 

For all the livolo devices you need to configure the 'advanced' settings it as is on the man pages. It will take you some time to set it up when you are trying it for the first time.

This is an open source project, so anyone who has time and the inclination, can fix it.

```
advanced:
  log_level: info
  ext_pan_id:
    - 33
    - 117
    - 141
    - 25
    - 0
    - 75
    - 18
    - 0
  channel: 26
```

PR #2823, opened 2021-07-23
rtx.js: fix require() to relative path

Hacker-CB:
rtx.js: fix require to relative path to allow work with npm link correctly while developing


Koenkk:
Thanks!

PR #2824, opened 2021-07-23
Add support for 5063430P7

SHxKM:
This adds support for the Fugato 4 spots

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1842,4 +1842,13 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp({colorTempRange: [153, 454]}),
         ota: ota.zigbeeOTA,
     },
+    {
+        zigbeeModel: ['5063430P7', '5063430P7', '5063430P7', '5063430P7'],
+        model: '5063430P7',
+        vendor: 'Philips',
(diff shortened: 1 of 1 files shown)


Koenkk:
This one is already supported: https://github.com/Koenkk/zigbee-herdsman-converters/blob/0577e2c2e55d364a2c33bb7e1cfc61992e6a8ea6/devices/philips.js#L1386

SHxKM:
@Koenkk it’s interesting however the way I’ve mapped it (with one string per bulb in the array) vs the way it was implemented. Any conventional wisdom on this difference?

Koenkk:
@SHxKM adding the same zigbeeModel multiple times doesn't make any difference.

SHxKM:
@Koenkk Interesting, so I’ve seen with some models we have to append `_n` to each one of the lights in a fixture, in fact this is what I tried to do first with one but it didn’t work. I wonder if I could have figured if I need or needn’t do that in any way? database.db and MQTT logs didn’t give much details. I just figured I will try to remove and just name each one of the devices the same (as the model name).

Koenkk:
You have to add all `modelId` from the `database.db` to the `zigbeeModel` array.

PR #2827, opened 2021-07-23
Fix Heiman HS2IRC no response.

Hacker-CB:
(fix https://github.com/Koenkk/zigbee2mqtt/issues/6897)

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -4241,25 +4241,28 @@ const converters = {
     heiman_ir_remote: {
         key: ['send_key', 'create', 'learn', 'delete', 'get_list'],
         convertSet: async (entity, key, value, meta) => {
+            const options = {
+                manufacturerCode: null,
+                ...utils.getOptions(meta.mapped, entity),
+            };
             switch (key) {
             case 'send_key':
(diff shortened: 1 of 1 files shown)


converters/toZigbee.js
```diff
@@ -4241,25 +4241,28 @@ const converters = {
     heiman_ir_remote: {
         key: ['send_key', 'create', 'learn', 'delete', 'get_list'],
         convertSet: async (entity, key, value, meta) => {
+            const options = {
+                manufacturerCode: null,
```
Koenkk, line 4245: I gues this forces the `manufacturerCode` to null instead of https://github.com/Koenkk/zigbee-herdsman/blob/a9819cb485fad01c21d75cdc05879fd8f5854f5e/src/zcl/definition/cluster.ts#L4154 ?

Hacker-CB:
Yes.
But its needed only for outgoing commands.

Hacker-CB:
@Koenkk I don't know why, but Heiman does not accept command if `manufacturerCode` is set.
Currently with `manufacturerCode = null` I tested and all works fine.
So it's workaround.

Koenkk:
Thanks!

PR #2828, opened 2021-07-23
Add support for Aqara roller shade motor

SerhiiSL:
Split Zigbee Model entry lumi.curtain & lumi.curtain.aq2 to define correct model name and description for Aqara roller shade motor (SRSC-M01)


Koenkk:
Thanks!

PR #2834, opened 2021-07-24
Add support for CentraLight 3328-G Micro Motion Sensor

rheidt:
On Windows, got lint to run fine after pulling back to replace EOL from CRLF to LF.


Koenkk:
Thanks!

PR #2835, opened 2021-07-25
Updated Develco SMSZB-120 to correctly support features

TheStigh:
- Now support IAS Zone
- Now support IAS WD
- Show correct enrolled status
- Show IASCIEAddress
- Removed `tz.warning` from exposes as this gave wrong impression of supported features
- Created correct duration and trigger

Device will only have supported attributes after OTA.

I've kept the `tz.warning` for backward compatibility for people using this to activate the siren, this is important, as we don't want to break something that has to do with an alarm. I have removed the exposes though.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1485,6 +1485,29 @@ const converters = {
             };
         },
     },
+    ias_enroll: {
+        cluster: 'ssIasZone',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const zoneState = msg.data.zoneState;
+            const iasCieAddr = msg.data.iasCieAddr;
+            const zoneId = msg.data.zoneId;
+            return {
+                enrolled: zoneState,
+                iascie_address: iasCieAddr,
+                zone_id: zoneId,
+            };
+        },
+    },
+    ias_wd: {
+        cluster: 'ssIasWd',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const result = {};
+            if (msg.data.hasOwnProperty('maxDuration')) result['max_duration'] = msg.data.maxDuration;
+            return result;
+        },
+    },
     command_on_presence: {
         cluster: 'genOnOff',
         type: 'commandOn',
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1493,8 +1493,8 @@ const converters = {
             const iasCieAddr = msg.data.iasCieAddr;
             const zoneId = msg.data.zoneId;
             return {
-                enrolled: zoneState,
-                iascie_address: iasCieAddr,
+                enrolled: zoneState !== 0,
+                ias_cie_address: iasCieAddr,
                 zone_id: zoneId,
             };
         },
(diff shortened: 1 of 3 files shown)


converters/fromZigbee.js
```diff
@@ -1485,6 +1485,29 @@ const converters = {
...
+    ias_enroll: {
+        cluster: 'ssIasZone',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const zoneState = msg.data.zoneState;
+            const iasCieAddr = msg.data.iasCieAddr;
+            const zoneId = msg.data.zoneId;
+            return {
+                enrolled: zoneState,
+                iascie_address: iasCieAddr,
```
Koenkk, line 1497: rename this to `ias_cie_address`.

converters/fromZigbee.js
```diff
@@ -1485,6 +1485,29 @@ const converters = {
...
     },
+    ias_enroll: {
+        cluster: 'ssIasZone',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const zoneState = msg.data.zoneState;
+            const iasCieAddr = msg.data.iasCieAddr;
+            const zoneId = msg.data.zoneId;
+            return {
+                enrolled: zoneState,
```
Koenkk, line 1496: This will be a number, but it makes more sense to expose a boolean instead (I think `enrolled: zoneState !== 0,`)

converters/toZigbee.js
```diff
@@ -1248,6 +1248,32 @@ const converters = {
     // #endregion
 
     // #region Non-generic converters
+    ias_wd: {
```
Koenkk, line 1251: This can be moved to the generic-converters region.

converters/toZigbee.js
```diff
@@ -1248,6 +1248,32 @@ const converters = {
...
+            return {state: {max_duration: value}};
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('ssIasWd', ['maxDuration']);
+        },
+    },
+    warning_simple: {
+        key: ['alarm'],
+        convertSet: async (entity, key, value, meta) => {
+            const alarmState = (value === 'alarm' || value === 'OFF' ? 0 : 1);
```
Koenkk, line 1264: If I understand correctly, this converter accepts `OFF` and `alarm`, I think `OFF` and `START` better explains what it does.

TheStigh: You want me to change this from ON / OFF to START/OFF ?

converters/toZigbee.js
TheStigh: Done

converters/fromZigbee.js
TheStigh: Tested, Done.

converters/fromZigbee.js
TheStigh: Done

converters/toZigbee.js
Koenkk: I don't understand why  you check for `value` `alarm`, `const alarmState = value === 'OFF' ? 0 : 1;` should be good.

TheStigh: Was leftover from earlier debugging :) Removed. PR already updated.

Koenkk:
Thanks!

PR #2838, opened 2021-07-25
Fix Yookee D10110 support

night199uk:
Some recent commits broke the D10110 support I think. This fixes it up. Also fix the off-by-one, which no longer seems to exhibit since resetting/reprogramming my blinds fully. Will provide the reset/update procedure for zigbee2mqtt docs soon.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5035,12 +5035,12 @@ const converters = {
     D10110_cover_position_tilt: {
         cluster: 'closuresWindowCovering',
         type: ['attributeReport', 'readResponse'],
-        convert: async (model, msg, publish, options, meta) => {
+        convert: (model, msg, publish, options, meta) => {
             if (msg.data.hasOwnProperty('currentPositionLiftPercentage') && msg.data['currentPositionLiftPercentage'] <= 100) {
                 // The Yookee D10110 SENDs it's position reversed, relative to the spec.
-                msg.data['currentPositionLiftPercentage'] = 100 - (msg.data['currentPositionLiftPercentage'] + 1);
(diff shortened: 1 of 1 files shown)


converters/fromZigbee.js
```diff
@@ -5035,12 +5035,12 @@ const converters = {
     D10110_cover_position_tilt: {
         cluster: 'closuresWindowCovering',
         type: ['attributeReport', 'readResponse'],
-        convert: async (model, msg, publish, options, meta) => {
+        convert: (model, msg, publish, options, meta) => {
             if (msg.data.hasOwnProperty('currentPositionLiftPercentage') && msg.data['currentPositionLiftPercentage'] <= 100) {
                 // The Yookee D10110 SENDs it's position reversed, relative to the spec.
-                msg.data['currentPositionLiftPercentage'] = 100 - (msg.data['currentPositionLiftPercentage'] + 1);
```
Koenkk, line 5041: This was added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/2463, @night199uk  why was the `+ 1` needed here?

night199uk: These blinds are "quirky" to say the least; I guess that's what you get for the cheapest Zigbee blinds I can find on the market.  If I remember correctly; I was seeing most of my blinds report 1-101 when I originally tested the report. That gets inverted elsewhere in the code to -1 - 99 because of invert: true instead of 0-100; so this +1 undid the offset. 

Since then one of my blind motors died and was replaced by the vendor and I figured out the correct calibration magic  for the motor to set top & bottom limits. Ever since I am no longer seeing the misreported values from the blinds and they are currently reliably reporting 0-100. Though; one direction is still inverted compared to the specs. 🤦

Issue #2841, opened 2021-07-26
Add support for toggle command on Legrand 067774


Ravexkgd:
With old firmwares supported actions on and off,
while new firmwares only support action toggle.
I suppose that adding action toggle would allow to support all
firmware versions.

The same, as for https://github.com/Koenkk/zigbee-herdsman-converters/pull/2694

Ravexkgd:
Info MQTT publish: topic 'zigbee2mqtt/0x00047400008ab95d', payload '{"action":"","battery":null,"last_seen":"2021-07-26T22:11:23+03:00","linkquality":27,"voltage":3000}'
Debug Received Zigbee message from '0x00047400008ab95d', type 'commandToggle', cluster 'genOnOff', data '{}' from endpoint 1 with groupID 0

Ravexkgd:
And working external converter with toggle and battery level:

const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
const tz = require('zigbee-herdsman-converters/converters/toZigbee');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const reporting = require('zigbee-herdsman-converters/lib/reporting');
const extend = require('zigbee-herdsman-converters/lib/extend');
const e = exposes.presets;
const ea = exposes.access;
const readInitialBatteryState = async (type, data, device) => {
    if (['deviceAnnounce'].includes(type)) {
        const endpoint = device.getEndpoint(1);
        const options = {manufacturerCode: 0x1021, disableDefaultResponse: true};
        await endpoint.read('genPowerCfg', ['batteryVoltage'], options);
    }
};

const device = {
        zigbeeModel: [' Double gangs remote switch\u0000\u0000\u0000\u0000'],
        model: '067774',
        vendor: '  Legrand',
        description: 'Wireless double remote switch',
        fromZigbee: [fz.identify, fz.command_on, fz.command_off, fz.command_toggle, fz.command_move, fz.command_stop, fz.battery],
        exposes: [e.battery(), e.action(['identify', 'on', 'off', 'toggle', 'brightness_move_up', 'brightness_move_down', 'brightness_stop'])],
        toZigbee: [],
        meta: {multiEndpoint: true, battery: {voltageToPercentage: '3V_2500'}},
        configure: async (device, coordinatorEndpoint, logger) => {
            const endpoint = device.getEndpoint(1);
[...]

PR #2845, opened 2021-07-27
Add support for toggle command on Legrand 067774

Ravexkgd:
With old firmwares supported actions on and off, while new firmwares only support action toggle.

Added support for toggle and battery level

As described in https://github.com/Koenkk/zigbee-herdsman-converters/issues/2841

### DIFF
--- a/devices/legrand.js
+++ b/devices/legrand.js
@@ -117,22 +117,22 @@ module.exports = [
         onEvent: readInitialBatteryState,
     },
     {
-        zigbeeModel: [' Double gangs remote switch', 'Double gangs remote switch'],
+        zigbeeModel: [' Double gangs remote switch\u0000\u0000\u0000\u0000'],
         model: '067774',
-        vendor: 'Legrand',
+        vendor: '  Legrand',
         description: 'Wireless double remote switch',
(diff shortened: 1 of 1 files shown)


devices/legrand.js
```diff
@@ -117,23 +117,23 @@ module.exports = [
         onEvent: readInitialBatteryState,
     },
     {
-        zigbeeModel: [' Double gangs remote switch', 'Double gangs remote switch'],
+        zigbeeModel: [' Double gangs remote switch\u0000\u0000\u0000\u0000'],
```
Koenkk, line 120: Why has `Double gangs remote switch` been removed here?

devices/legrand.js
```diff
@@ -117,23 +117,23 @@ module.exports = [
...
         model: '067774',
-        vendor: 'Legrand',
+        vendor: '  Legrand',
```
Koenkk, line 122: Please remove the trailing space.

devices/legrand.js
Ravexkgd: Because external converter doesn't work with it.
I think it's possible to use old variant, but i don't have device with old firmware version to check it.

devices/legrand.js
Ravexkgd: According to logs, that's the vendor name device send's.

PR #2850, opened 2021-07-30
Fix configure error

xfzha:
I'm still a novice, and I'm not familiar with many operations, so I'm sorry.


Koenkk:
No problem, thanks!

PR #2851, opened 2021-07-31
Fix `hue_power_on_color` action to use refactored color library

davidjb:
This applies the lib changes from https://github.com/Koenkk/zigbee-herdsman-converters/pull/2453 to Hue power on hex color conversion.

Previously, trying to set the power on color resulted in an error since the function no longer exists:

    TypeError: utils.hexToXY is not a function


Koenkk:
Thanks!

PR #2857, opened 2021-08-01
Add Aqara GZCGQ11LM

xfzha:
illuminance & illuminance_lux report correct.
set detection period time requires click the button of the device.


devices/xiaomi.js
```diff
@@ -1379,4 +1379,20 @@ module.exports = [
...
+    {
+        zigbeeModel: ['lumi.sen_ill.agl01'],
+        model: 'GZCGQ11LM',
+        vendor: 'Xiaomi',
+        description: 'Aqara T1 light intensity sensor',
+        fromZigbee: [fz.battery, fz.illuminance, fz.aqara_opple_report],
+        toZigbee: [tz.GZCGQ11LM_detection_period],
+        meta: {battery: {voltageToPercentage: '3V_2500'}},
+        exposes: [e.battery(), e.battery_voltage(), e.illuminance(), e.illuminance_lux(),
+            exposes.numeric('detection_period', exposes.access.ALL).withValueMin(1).withValueMax(59).withUnit('s')],
```
Koenkk, line 1391: What does this `detection_period` do? Can you add a description for it: `.withDescription('....')`

xfzha: May be the reporting interval.
There is no detailed description of <code>detection_period</code> in the aqara home. I think it means to report a change at an interval. In fact, if the light changes, it is reported according to this interval.

Koenkk: Then please add that as the description (its better than having nothing)

Koenkk:
Thanks!

PR #2860, opened 2021-08-02
Added Support For Vimar 03906 (GreenPower)

danieledwardgeorgehitchcock:
Tested working with generic GreenPower converter


Koenkk:
Thanks!

PR #2862, opened 2021-08-02
Added support for Aqara H1 EU with neutral (WS-EUK03 and WS-EUK04)

Mephistofeles:
I am not able to test WS-EUK03, however the differences should be similar to WS-EUK01 vs WS-EUK02
WS-EUK04 was tested.


Koenkk:
Thanks!

PR #2867, opened 2021-08-03
Remove running_state report from RC-2000WH thermostat

presslab-us:
This thermostat does not support `running_state` reporting. Home Assistant logs constant errors because this variable is never set in JSON.  This pull request fixes those repeated errors in the log.


Koenkk:
Thanks!

PR #2872, opened 2021-08-05
Added LTD012/5111531H5

luke-barnett:
Not sure if this should just be collapsed with the other Zigbee model for the Garnea range. Technically it has a different model number on the box so chose to create a new entry.

I have this running on my local instance using an external converter and all working nicely (I just duplicated the hueExtend section into my external converter).

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -96,7 +96,7 @@ module.exports = [
         vendor: 'Philips',
         description: 'Garnea downlight',
         meta: {turnsOffAtBrightness1: true},
-        extend: hueExtend.light_onoff_brightness_colortemp(),
+        extend: hueExtend.light_onoff_brightness_colortemp({colorTempRange: [153, 454]}),
         ota: ota.zigbeeOTA,
     },
     {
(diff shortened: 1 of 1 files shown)


devices/philips.js
```diff
@@ -90,6 +90,15 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp(),
         ota: ota.zigbeeOTA,
     },
+    {
+        zigbeeModel: ['LTD012'],
+        model: '5111531H5',
+        vendor: 'Philips',
+        description: 'Garnea downlight',
+        meta: {turnsOffAtBrightness1: true},
+        extend: hueExtend.light_onoff_brightness_colortemp(),
```
Koenkk, line 99: Please add the color temp as described on https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html#31-retrieving-color-temperature-range-only-required-for-lights-which-support-color-temperature

luke-barnett: Updated, I also have some LDT001 fixtures and so queries and set those as well. They happened to be the same, which may be the case for all of the downlights from Philips/Hue but I only have those two models to confirm.

PR #2876, opened 2021-08-06
PTVO firmware: the write command didn't work for endpoints 10-16

ptvoinfo:
PTVO firmware: the write command didn't work for endpoints 10-16


Koenkk:
Thanks!

PR #2881, opened 2021-08-06
Tint 404011 remote: Add direction of color temp changes

froogl:
The tint remote MLI-404011 by Müller Licht has an internal state which is emitted through the color temperature change buttons. In order to get the actual button which has been pressed we need to mirror that state internally.


Koenkk:
Thanks!

PR #2883, opened 2021-08-07
add new device

fritzschefelix:
Autodiscovery for my Hama 00176592 was not working, so I added a new entry to the list.


Koenkk:
Thanks!

PR #2886, opened 2021-08-08
add aqara zncz15lm overload protection

xfzha:
Has been tested.
I set the maximum allowable power to 100w, and then the water heater is plugged in. When the water heater heats up, the power is turned off


Koenkk:
Thanks!

PR #2890, opened 2021-08-09
Add support for external temp sensor

robertalexa:
This adds support for External Temp Sensor for Danfoss Ally and Hive Trv. This code is based on the Danfoss Ally specs sheet and tested on the Hive Trv.

This was added at the request of a user here: https://github.com/Koenkk/zigbee2mqtt/issues/4504#issuecomment-895470566

The PR needs https://github.com/Koenkk/zigbee-herdsman/pull/404 to be merged first in zigbee-herdsman.

### DIFF
--- a/devices/hive.js
+++ b/devices/hive.js
@@ -194,6 +194,12 @@ module.exports = [
                 maximumReportInterval: constants.repInterval.MINUTES_10,
                 reportableChange: 1,
             }], options);
+            await endpoint.configureReporting('hvacThermostat', [{
+                attribute: 'danfossExternalMeasuredRoomSensor',
+                minimumReportInterval: constants.repInterval.MINUTE,
+                maximumReportInterval: constants.repInterval.MAX,
+                reportableChange: 1,
+            }], options);
 
             await endpoint.read('hvacThermostat', [
                 'danfossWindowOpenExternal',
@@ -203,6 +209,7 @@ module.exports = [
                 'danfossHeatAvailable',
                 'danfossMountedModeControl',
                 'danfossMountedModeActive',
+                'danfossExternalMeasuredRoomSensor',
             ], options);
             // read keypadLockout, we don't need reporting as it cannot be set physically on the device
             await endpoint.read('hvacUserInterfaceCfg', ['keypadLockout']);
--- a/devices/hive.js
+++ b/devices/hive.js
@@ -126,7 +126,8 @@ module.exports = [
             tz.danfoss_mounted_mode_control, tz.danfoss_thermostat_vertical_orientation, tz.danfoss_algorithm_scale_factor,
             tz.danfoss_heat_available, tz.danfoss_heat_required, tz.danfoss_day_of_week, tz.danfoss_trigger_time,
             tz.danfoss_window_open_internal, tz.danfoss_window_open_external, tz.danfoss_load_estimate,
-            tz.danfoss_viewing_direction, tz.thermostat_keypad_lockout, tz.thermostat_system_mode],
+            tz.danfoss_viewing_direction, tz.danfoss_external_measured_room_sensor, tz.thermostat_keypad_lockout,
+            tz.thermostat_system_mode],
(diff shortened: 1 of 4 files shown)


devices/hive.js
```diff
@@ -144,6 +145,9 @@ module.exports = [
                 .withDescription('Whether or not the unit needs warm water. `false` No Heat Request or `true` Heat Request'),
             exposes.climate().withSetpoint('occupied_heating_setpoint', 5, 32, 0.5).withLocalTemperature().withPiHeatingDemand()
                 .withSystemMode(['heat']),
+            exposes.numeric('external_measured_room_sensor', ea.ALL)
+                .withDescription('Set at maximum 3 hours interval but not more often than every 30 minutes at every 100 ' +
+                    'value change. Resets every 3hours to standard (-8000=undefined).'),
```
sjorge, line 150: Is it reportable? If so maybe it will report when it is reset after 3 hours, that way the field can be automatically updated.

robertalexa: @sjorge The attribute is defined as Fixed reporting. 
image

The field is storing a value that you pass to it for temperature coming from an external sensor. There is not "time left" attribute reported back.

I think the way this is designed to work is for people that have inaccurate temp reading from the radiator compared to the rest of the room. Case in which they will be sending values constantly, since their problem will persist? The temp can only be set via gateway (z2m) so updating it will keep the latest value in z2m, and every change to the value will always be set by commands.

I do not know how beneficial reporting will be? I mean i guess it will be if someone does it once, and then they stop. But I guess people will either use it or not, with no in between. 

Here is what I added:
image

sjorge: I guess it depends, if you set it now and read it back in 3h 10m ... does it still return the value that was set? Or does it return -8000, if the later it would make sense to have it report the value. If reading the value after the time returns the set value, there is no need.

robertalexa: It would automatically change to -8000 (that is according to the specs document, TBH i have not tried it to see what happens) - and it does make sense for this behaviour otherwise the internal PID would never know to go back to standard operation.

Well I have done the commit for reporting, the reporting is set successfully so the device seems to accept the config.

For testing, with reporting configured, I have just set a value. I will report back in 3 hours.

robertalexa: @sjorge Reporting works as expected. Value changed from my set one of 2400 to -8000 when the 3 hours have elapsed. That said, i will mark this thread as resolved and I guess the PR is good for merging.

sjorge: Excellent!

sjorge (approved):

Koenkk:
Thanks!

Coffee0297:
The Popp 701721 is the same sensor as the danfoss 014G2461. is it possible to add this for the popp too?

robertalexa:
@Coffee0297 this already exists. Everything that the Danfoss has, Popp inherits :)

Coffee0297:
Ahh I see... Sorry... Just cant get the bind working from a aqara temp sensor to the popp. Then I assume the issue lie elsewhere.
Thanks

robertalexa:
@Coffee0297 You won't be binding the sensor to the popp. Instead you will use your automation flow (Home Assistant) to read the temperature and write it to the external temp sensor of the popp following the guidelines - as often as 5 minutes but not longer than 30 mins and only for a difference of 0.1 degrees. Hope this helps :)

OZ1SEJ:
This functionality is exactly what I am looking for! But I can't see that it's documented anywhere. Can anyone please point me in the direction of how to actually implement this?

Coffee0297:
You make it via automations
And if you don't wanna mean the internal sensor and room sensor you need to set "radiator covered"
Screenshot_20221119-171958
Screenshot_20221119-171922
Screenshot_20221119-171914

robertalexa:
This was both discussed above and documented on the .io website on the device documentation page.

OZ1SEJ:
Thanks - will definitely try this!

PR #2896, opened 2021-08-11
Adding configuration to Sinope DM2500ZB

onixx:
Adding some configuration options for the Sinope DM2500ZB dimmer

1. led_brightness_on (brightness of the status led when load is on)
2. led_brightness_off (brightness of the status led when load is off)
3. minimum_brightness (sets the minimum brightness of the dimmer to avoid flicker on some bulbs or make sure an inductive load has enough power to turn on at 1% brightness for example)

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -2977,6 +2977,33 @@ const converters = {
             }
         },
     },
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #2897, opened 2021-08-11
Update Ubisys J1 default bindings

sjorge:
I got my hands on a J1 and noticed we didn't add the default bindings for this one yet, I double checked the manual and verify that it matches with the hardware.


Koenkk:
Thanks!

PR #2899, opened 2021-08-11
add support for Miboxer FUT036Z

lustyffh:
[Single Color Led Controller](http://www.futlight.com/light/xqy/fut036z.html)


Koenkk:
Thanks!

PR #2900, opened 2021-08-11
Add support for Nedis ZBHTR10W

robertalexa:
This is as a result of https://github.com/Koenkk/zigbee2mqtt/issues/8333

The user requesting support mentioned this being a different TRV behind the hood, also provided the reported model and manufacturer name. Also tested at their end and works fine.

I do not own the device not I was able to test this myself, just pushing the PR for OP


Koenkk:
Thanks for helping out @robertalexa !

PR #2905, opened 2021-08-13
Added support for FUT035Z CCT Dimmer

fluffymadness:


devices/miboxer.js
```diff
@@ -27,4 +27,11 @@ module.exports = [
         vendor: 'Miboxer',
         extend: extend.light_onoff_brightness(),
     },
+    {
+        fingerprint: [{modelID: 'TS0502B', manufacturerName: '_TZ3210_frm6149r'}],
+        model: 'FUT035Z',
+        description: 'Dual White LED controller',
+        vendor: 'Miboxer',
+        extend: extend.light_onoff_brightness_colortemp({disableColorTempStartup: true})
```
Koenkk, line 35: Please add the color temp range as described on https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html#31-retrieving-color-temperature-range-only-required-for-lights-which-support-color-temperature

fluffymadness: Hi, I added an additional commit with the color temperature range.
I got no error or anything when testing with the modified file. I guess it should work ?

PR #2907, opened 2021-08-13
Add specific converter for Xiaomi JTYJGD01LMBW

### LINKED ISSUE #2889: Xiaomi/Honeywell smoke detector does not handle ALARM_2
When testing my Xiaomi smoke detector I have discovered that ALARM_2 (zonestatus 2) is currently not handled by any of the IAS-related conververs. For the Xiaomi smoke detector the following code in `fz.ias_smoke_alarm_1` is used:

```javascript
const zoneStatus = msg.data.zonestatus;
return {
    smoke: (zoneStatus & 1) > 0,
    tamper: (zoneStatus & 1<<2) > 0,
    battery_low: (zoneStatus & 1<<3) > 0,
    supervision_reports: (zoneStatus & 1<<4) > 0,
    restore_reports: (zoneStatus & 1<<5) > 0,
    trouble: (zoneStatus & 1<<6) > 0,
    ac_status: (zoneStatus & 1<<7) > 0,
    test: (zoneStatus & 1<<8) > 0,
};
```
As you can see a `zoneStatus` of ``1<<1`` is not handled at all. The big question is how to handle this. DeCONZ for example triggers the alarm for both ALARM_1 and ALARM_2 and does not distinguish between the two (see https://github.com/dresden-elektronik/deconz-rest-plugin/blob/16bd742c6fd67154cc8110fee4d9515c75c0508e/ias_zone.cpp#L465). I think we could just copy that behavior. This would mean:

```javascript
smoke: (zoneStatus & 1) > 0 || (zoneStatus & 1<<1) > 0,
```

@Koenkk Should we copy the deCONZ behavior or should we introduce an entirely new attribute for "ALARM_2"? I can send you a PR for the `fz.ias_smoke_alarm_1` converter, as this is the one I can test. But basically all converters for IAS zones are affected by this and maybe you would like to fix them all at the same time.

froogl:
Properly handle IAS ALARM_2 which is used for signalling test mode
(long-press of test button) on this particular device.

Closes #2889

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5409,18 +5409,10 @@ const converters = {
         cluster: 'ssIasZone',
         type: 'commandStatusChangeNotification',
         convert: (model, msg, publish, options, meta) => {
+            const result = converters.ias_smoke_alarm_1.convert(model, msg, publish, options, meta);
             const zoneStatus = msg.data.zonestatus;
-            return {
-                smoke: (zoneStatus & 1) > 0,
-                tamper: (zoneStatus & 1<<2) > 0,
-                battery_low: (zoneStatus & 1<<3) > 0,
-                supervision_reports: (zoneStatus & 1<<4) > 0,
-                restore_reports: (zoneStatus & 1<<5) > 0,
-                trouble: (zoneStatus & 1<<6) > 0,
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #2912, opened 2021-08-15
Add support for Benexmart wireless button switch

sapkra:
I tested it with the 1 button switch and it works fine. I'm not 100% sure about the model names and if it will still detect other devices which are not matching the fingerprint. It's my first contribution and it would be helpful to get some feedback on these topics.

The devices can be found gere:
https://de.aliexpress.com/item/4000793877542.html

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -445,10 +445,11 @@ module.exports = [
     },
     {
         zigbeeModel: ['TS0041'],
+        fingerprint: [{manufacturerName: '_TZ3000_tk3s5tyg'}],
         model: 'TS0041',
         vendor: 'TuYa',
         description: 'Wireless switch with 1 button',
-        whiteLabel: [{vendor: 'Smart9', model: 'S9TSZGB'}, {vendor: 'Lonsonho', model: 'TS0041'}],
+        whiteLabel: [{vendor: 'Smart9', model: 'S9TSZGB'}, {vendor: 'Lonsonho', model: 'TS0041'}, {vendor: 'Benexmart', model: 'ZM-sui1'}],
         exposes: [e.battery(), e.action(['single', 'double', 'hold'])],
         fromZigbee: [fz.tuya_on_off_action, fz.battery],
         toZigbee: [],


Koenkk:
The 2 button switch likely has a different manufactuerName, can you share the data/database.db entry of your device?

sapkra:
```json
{
  "id": 21,
  "type": "EndDevice",
  "ieeeAddr": "<ieeeAddr>",
  "nwkAddr": 38743,
  "manufId": 4098,
  "manufName": "_TZ3000_tk3s5tyg",
  "powerSource": "Battery",
  "epList": [
    1
  ],
  "endpoints": {
    "1": {
      "profId": 260,
      "epId": 1,
      "devId": 0,
      "inClusterList": [
        0,
        1,
        6
      ],
      "outClusterList": [
        25,
        10
      ],
      "clusters": {
        "genBasic": {
          "attributes": {
            "manufacturerName": "_TZ3000_tk3s5tyg",
            "powerSource": 3,
            "zclVersion": 3,
            "appVersion": 65,
            "stackVersion": 0,
            "hwVersion": 1,
            "dateCode": ""
          }
        }
      },
      "binds": [
        {
          "cluster": 1,
          "type": "endpoint",
          "deviceIeeeAddress": "<deviceIeeeAddress>",
          "endpointID": 1
        }
      ],
      "configuredReportings": [],
      "meta": {}
    }
  },
  "appVersion": 65,
  "stackVersion": 0,
  "hwVersion": 1,
  "dateCode": "",
  "zclVersion": 3,
  "interviewCompleted": true,
  "meta": {
    "configured": 1
  },
  "lastSeen": 1629050446334
}
```

sapkra:
btw. it's the same device with different firmwares. I have 3 capacitive sensors but 2 of them are disabled by firmware.

Koenkk:
Did you test this with all 3 devices? (1, 2 and 3 buttons)? If not, only the `TS0041` should be updated, the others likely have an other `manufacturerName`

sapkra:
I only tested the one with one button. I will remove the others soon.

sapkra:
But there is still the question if it would still work for other devices which are using the same `zigbeeModel` but don't match the fingerprint.

Koenkk:
> But there is still the question if it would still work for other devices which are using the same zigbeeModel but don't match the fingerprint.

It will, first a match against the fingerprint is tried, if that fails it is matched against the `zigbeeModel`.

PR #2914, opened 2021-08-15
Add support for Benexmart Tuya Bulb

sapkra:
https://de.aliexpress.com/item/1005002328614570.html

FYI: I have two bulbs and both are detected as different devices because they are using different tuya controllers.


Koenkk:
Thanks!

PR #2920, opened 2021-08-17
Add the `RGBW LED lamp GU10 6W dimbaar`

sjorge:
Seems to be the same as the E27 version but a GU10 spot, sadly no access to the hardware myself but my uncle send me logs so I managed to extract the model.


sjorge:
```Read result of 'lightingColorCtrl': {"colorTempPhysicalMin":153,"colorTempPhysicalMax":555}```

Managed to get teamvieuwer access to get the info

Koenkk:
Thanks!

PR #2921, opened 2021-08-17
Add power_on_behavior for MOES MS-104Z

nklebedev:
Added power_on_behavior for one gang version of MOES MS-104Z switch.


Koenkk:
Thanks!

PR #2923, opened 2021-08-18
feat: add new device TuYa Intelligent gas(water) valve controller PM02D-TYZ

va4ok:
Add new device PM02D-TYZ
Add new fromZigbee converter
Add new toZigbee converter

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -6292,6 +6292,21 @@ const converters = {
             }
         },
     },
+    tuya_intelligent_valve: {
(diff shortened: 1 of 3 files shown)


va4ok:
Not sure if i'm totally right with model and other device info so i've added the photo with this new device
TuYa_valve

Koenkk:
Already added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/2917

va4ok:
Yeah i see now, thanks.

PR #2924, opened 2021-08-18
Update centralite.js to add support for 3310-G temp & humidity sensor

nkasrawi:
This device seems very similar to the smart things 3310-S so I copied that config and adjusted the names.  It's working fine for me now.  The only issue is that the firmware version isn't being read correctly.


Koenkk:
Thanks!

PR #2928, opened 2021-08-20
Add E21-N1EA

ctalkington:


Koenkk:
Looks good, why is this marked as draft?

ctalkington:
I did have one question, the docs for Z2M suggest that 

> Repeat until your device does not produce any more log messages like: 2018-5-1 18:19:41 WARN No converter available for 'WSDCGQ01LM' with....

This device has several clusters that seem to suggest it may support more but I'm not seeing such messages in log.

```
"clusters":{"input":["genBasic","genIdentify","genGroups","genScenes","genOnOff","genLevelCtrl","seMetering","haDiagnostic","manuSpecificUbisysDimmerSetup","manuSpecificSamsungAccelerometer"] ...}}
```

Koenkk:
Clusters represent a set of functionallity, e.g. `genGroups` makes it possible to add this device to groups etc.

This config looks fine for a bulb.

PR #2930, opened 2021-08-20
Added support for Datek water sensor

TheStigh:
They have changed vendor to "Eva", but actual manufacturer are still "Datek" and same manufacturer id.

### DIFF
--- a/devices/datek.js
+++ b/devices/datek.js
@@ -154,8 +154,8 @@ module.exports = [
     {
         zigbeeModel: ['Water Sensor'],
         model: 'HSE2919E',
-        vendor: 'Eva',
-        description: 'Water leak sensor',
+        vendor: 'Datek',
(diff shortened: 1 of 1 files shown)


devices/datek.js
```diff
@@ -151,4 +151,26 @@ module.exports = [
             exposes.enum('service_mode', ea.ALL, ['deactivated', 'random_pin_1x_use',
                 'random_pin_24_hours']).withDescription('Service Mode of the Lock')],
     },
+    {
+        zigbeeModel: ['Water Sensor'],
+        model: 'HSE2919E',
+        vendor: 'Eva',
```
Koenkk, line 157: I suggest to just put Datatek here and change the description to `Eva water leak sensor`.

TheStigh: Done :)

PR #2933, opened 2021-08-21
Add support for Philips Bloom Gen4 - US Addition

jlmaners:
Hopefully this was done properly - has been tested and linted. 

Added an additional descriptor text to the previous Bloom entry as that model is defined as an EU/UK model. My guess it is about the included plug adaptor but they have not only different Zigbee Models but different models from Philips as well.

Let me know if any changes are needed.


Koenkk:
Thanks!

PR #2939, opened 2021-08-22
Explicitly specify the exposes for Schneider Wiser 4bth switch

henrik-olesen:
Explicitly specify the exposes for Schneider Electric LK Fuga Wiser 4bth switch (required for compatibility with homebridge-z2m)


Koenkk:
Thanks!

PR #2940, opened 2021-08-23
Added support for Datek Scene Selector

TheStigh:


devices/datek.js
```diff
@@ -173,4 +173,22 @@ module.exports = [
...
+        toZigbee: [tz.on_off],
+        meta: {battery: {voltageToPercentage: '3V_2500'}},
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg', 'genBasic', 'genOnOff',
+                'genLevelCtrl', 'msTemperatureMeasurement']);
+            await reporting.batteryVoltage(endpoint);
+            await reporting.temperature(endpoint, {min: constants.repInterval.MINUTES_10, max: constants.repInterval.HOUR, change: 100});
+        },
+        exposes: [e.battery(), e.temperature()],
```
Koenkk, line 192: Given that `fromZigbee` contains some `fz.command_*` I would also expect an `e.action(['on', OTHER_ACTIONS])` here.

TheStigh: I started with the e.action but removed it as I do only see the command issued for 2/10's of a second. I've currently resolved this with an automation in HA but later I would like to investigate how I can make the key pressed show for a longer time.

Though, the commands from the scene selector works according to the fz.'z.

Koenkk: @TheStigh that is expected behaviour when the homeassistant integration is enabled: https://www.zigbee2mqtt.io/information/FAQ.html#why-is-the-action-property-always-empty

TheStigh: @Koenkk Exactly how I resoved it but letting the `{"action":` trigger an automation.

Did remove the` fz.command_step` and added `e.action()`

PR #2943, opened 2021-08-23
Update develco.js

mamrai1:
new converter for Fire detector  HESZB-120 similar to SMSZB-120 except OTA

### DIFF
--- a/devices/develco.js
+++ b/devices/develco.js
@@ -172,13 +172,20 @@ module.exports = [
         model: 'HESZB-120',
         vendor: 'Develco',
         description: 'Fire detector with siren',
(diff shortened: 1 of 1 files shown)


Koenkk:
@TheStigh can you review this?

TheStigh:
> 
> 
> @TheStigh can you review this?

Should be OK, pretty much identical with the SMSZB-120. I wouldn't be surprised if they share firmware also - the only difference are fire vs smoke - still they both trip the `Bit0: Alarm 1` in `zoneStatus`

mamrai1:
I couldn't verify the firmware version.. I use conbee ii and seems it
doesn't read the custom x8000 address... So i removed the ota...

PR #2947, opened 2021-08-24
Support for Onesti Relax Smart Plug

OsteHovel:
Support for Onesti Relax Smart Plug
I made this a while ago but I figured out that it is time to publish it for everyone.
This device does not contain a relay, even tho it have the on/off state it will always be on.
https://onestiproducts.io/relax-smart-plug/
Smart_plug

image


Koenkk:
Thanks!

PR #2950, opened 2021-08-24
No converter available for 'AV2010/29A'

Rockman18:
No converter available for 'AV2010/29A' with cluster 'ssIasZone' and type 'commandStatusChangeNotification' and data '{"extendedstatus":0,"zonestatus":20}'

### DIFF
--- a/devices/bitron.js
+++ b/devices/bitron.js
@@ -79,7 +79,7 @@ module.exports = [
         model: 'AV2010/29A',
         vendor: 'Bitron',
         description: 'SMaBiT Zigbee outdoor siren',
-        fromZigbee: [fz.ias_no_alarm],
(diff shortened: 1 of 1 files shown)


devices/bitron.js
```diff
@@ -79,7 +79,7 @@ module.exports = [
         model: 'AV2010/29A',
         vendor: 'Bitron',
         description: 'SMaBiT Zigbee outdoor siren',
-        fromZigbee: [fz.ias_no_alarm],
+        fromZigbee: [fz.ias_siren],
```
Koenkk, line 82: The existing converter should not be removed, otherwise you will get a `No converter available..` for the `ssIasZone`.`attributeReport`

PR #2954, opened 2021-08-25
Add support for led/siren control and battery level for LS21001

Tediore:
Per https://github.com/Koenkk/zigbee2mqtt/discussions/8509

Thanks for creating the converter Koenkk!

Lint failed because of line length, which I also ignored when running my own checks. Test passes though. Should I tweak that at all?


Koenkk:
Thanks!

PR #2961, opened 2021-08-26
Add support for Iluminize 511.541

TheRealKillaruna:
Add support for new device Zigbee 3.0 wall dimmer RGBW 1 zone

### DIFF
--- a/devices/iluminize.js
+++ b/devices/iluminize.js
@@ -153,10 +153,10 @@ module.exports = [
         model: '511.541',
         vendor: 'Iluminize',
         description: 'Zigbee 3.0 wall dimmer RGBW 1 zone',
-        fromZigbee: [fz.command_recall, fz.command_on, fz.command_off, fz.command_move_to_color, fz.command_move_to_color_temp, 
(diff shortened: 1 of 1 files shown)


TheRealKillaruna:
@itavero I think I took over the wild card from the handheld remote (Iluminize 511.344), which has the same buttons (basically 4 scenes that can be stored and recalled). I do not remember what the actual recall messages were, but I could of course check it in debug mode. Would this be useful?

itavero:
For other systems using the `exposes` information (like the Homebridge plug-in I've made), this would help to support the device.

In my case, I have to map the possible values to HomeKit characteristics and events. If there's a wildcard, I can't do this, as I need the exact values.

TheRealKillaruna:
@itavero the possible values are `recall_1`, `recall_2`, `recall_3`, and `recall_4`. I verified this in the log files, where I have entries like `info  2021-09-08 21:02:37: MQTT publish: topic 'zigbee2mqtt/Wohnzimmer Schalter RGBW/action', payload 'recall_1'`
Is this the information you were looking for or do you need anything else?

PR #2966, opened 2021-08-28
Emulate running state for Danfoss Ally thermostat

henrik-olesen:
Emulate standard "running state" attribute of Danfoss Ally thermostats using manufacturer specific attribute "heat required".
In addition system mode is read during configuration (constant).
This allows better integration with systems relying on "running state" and "System mode" to visualise whether the thermostat is currently heating.


Koenkk:
Thanks!

PR #2970, opened 2021-08-29
Add support for Philips 3435731P7

xzwr:
Add support for Philips 3435731P7 Hue Adore white ambiance bathroom mirror.
Functionality identical to Phillips 3418631P6.
Color temperature range verified with two instances.


Koenkk:
Thanks!

PR #2971, opened 2021-08-29
Adding support for number of turns for TERNCY-SD01

JuliusSenkus:
Adding missing property to get how much rotary switch was turned.

### DIFF
--- a/devices/terncy.js
+++ b/devices/terncy.js
@@ -38,7 +38,7 @@ module.exports = [
         ota: ota.zigbeeOTA,
         meta: {battery: {dontDividePercentage: true}},
         exposes: [e.battery(), e.action(['single', 'double', 'triple', 'quadruple', 'rotate']),
-            exposes.text('direction', ea.STATE)],
+            exposes.text('direction', ea.STATE), e.number('number', 'number').withDescription('Number of turns')],
     },
     {
         zigbeeModel: ['TERNCY-LS01'],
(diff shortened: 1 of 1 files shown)


devices/terncy.js
```diff
@@ -38,7 +38,7 @@ module.exports = [
         ota: ota.zigbeeOTA,
         meta: {battery: {dontDividePercentage: true}},
         exposes: [e.battery(), e.action(['single', 'double', 'triple', 'quadruple', 'rotate']),
-            exposes.text('direction', ea.STATE)],
+            exposes.text('direction', ea.STATE), exposes.numeric('number', ea.STATE).withDescription('Number of turns')],
```
Koenkk, line 41: Shouldn't `number` be `action_number`?

JuliusSenkus: No, if it should, also instead `direction` there should be `action_direction`. Also tested it.
image

JuliusSenkus: Or am I unaware of some convention?

Koenkk: Ah sorry, you still have the legacy integration enabled, try disabling it: https://www.zigbee2mqtt.io/devices/TERNCY-SD01.html#device-type-specific-configuration and make sure to remove your `data/state.json` to purge the old attributes.

JuliusSenkus: Stopped service, set `legacy: false` for this device in `devices.yaml`, removed `state.json` and that seemed to break my scripts for this device, did not investigate much (maybe missing states), why was that needed?

PR #2974, opened 2021-08-30
Correct manufacturers for Xfinity security keypads

Rockman18:
Manufacturers for Xfinity devices are wrong.  XFinity is a reseller. 
The 2 keypads are the same device with 2 different manufacturers. The devices declare as manufacturer UEI (Universal Electronics Inc) for XHK1-UE or Technicolor for  XHK1-TC.

Capture d’écran du 2021-08-30 16-47-23

### DIFF
--- a/devices/technicolor.js
+++ b/devices/technicolor.js
@@ -8,43 +8,10 @@ const ea = exposes.access;
 
 module.exports = [
     {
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #2975, opened 2021-08-30
Add Ajax Online AJ_ZIGPROA60 12W A60 Bulb

KingJ:
This bulb does seem almost identical to the Gledopto GL-B-008P (i.e photos and firmware versions seem identical) but as I don't have the Gledopto bulb to hand I can't be 100% sure. Therefore i've added this as a new device definition for now rather than as a whitelabel.

[Colour temperature range confirmed by following the instructions here](https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html#31-retrieving-color-temperature-range-only-required-for-lights-which-support-color-temperature). When setting the brightness to 1, the bulb does turn off so i've set the `turnsOffAtBrightness1` meta property to `true`.


Koenkk:
Thanks!

PR #2976, opened 2021-08-30
Add Smart Round Ceiling Lamp Comfort RCL 240 T

Tri0L:
https://www.innr.com/en/product/smart-round-ceiling-lamp-comfort/


Koenkk:
Thanks!

PR #2980, opened 2021-08-31
Add support for various ZGP devices

fredericseiler:
Main issue Koenkk/zigbee-herdsman#398

This PR adds a `ieeeAddr` fingerprint filter to better identify ZGP devices.

Also adds support for:

- EnOcean PTM 215Z (FoH)
- EnOcean PTM 215ZE
- EnOcean PTM 216Z
- Philips Hue Tap
- Legrand ZLGP15
- Legrand ZLGP17
- Legrand ZLGP18


devices/green_power.js
```diff
@@ -28,6 +25,5 @@ module.exports = [
         fromZigbee: [fz.greenpower_7],
         toZigbee: [],
         exposes: [e.action(['*'])],
-        whiteLabel: [{vendor: 'EnOcean', description: 'Easyfit 1 or 2 gang switch', model: 'EWSxZG'}],
```
Koenkk, line 31: Just did some test with my Hue tap and PTM215Z and it works perfectly, I'm really happy with this solution.

Is there any reason to keep `green_power.js` and its converters? The only downside I see here that it might make an existing supported device unsupported, but I think chances are small. So if you agree please remove it.

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/888

Koenkk:
Thanks!

koalo:
I am not sure if is actually possible to assign the devices to different vendors according to the ieeeAddr. I have two devices falling in this category with the addresses 0x00000000017119eb and 0x0000000024227604. They are actually by the vendor "Led Trading", but now the first one is identified as EnOcean device (https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/enocean.js#L7) and the other one is not working anymore. The correct device was added with https://github.com/Koenkk/zigbee-herdsman-converters/commit/d6444cd701d03f23414fc8619401a402b95bfeda but the prefix does not match.

Are you sure that the address prefixes actually can be used in this case (then it would be the fault of Led Trading to assign wrong addresses - we still need an idea how to handle this...) or are the address prefixes just an assumption from the observation of a few devices?

fredericseiler:
The source address prefixes and device IDs used in this PR are [confirmed by EnOcean](https://github.com/Koenkk/zigbee-herdsman/issues/398#issuecomment-908500700).

koalo:
@fredericseiler Thanks a lot for the hint! Then apparently other vendors (at least Led Trading) do not really care about using always the same (or at least consistent) addresses. Any idea how we can handle this? Is there at least a way to have a fallback (i.e. handle all devices with addresses that do not match any other devices as generic)?

fredericseiler:
Well, the whole thing is clearly an issue for the CSA, but I'm pretty sure they're very busy with Matter and don't want to waste resources on ZGP open-source integration...

PR #2991, opened 2021-09-04
support ts0203 battery percentage

nukusinji:


devices/tuya.js
```diff
@@ -19,6 +19,11 @@ module.exports = [
         toZigbee: [],
         exposes: [e.contact(), e.battery_low(), e.tamper(), e.battery()],
         whiteLabel: [{vendor: 'CR Smart Home', model: 'TS0203'}],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
```
Koenkk, line 23: To be sure, please wrap this in a `try/catch`, there are many different versions of this device and it may not work for all

PR #2992, opened 2021-09-04
Hive and Danfoss Ally TRV - small change

robertalexa:
Hi,

1. Changed viewing_direction from a 0 to 1 numeric to a binary. I have no idea why i didn't do it like this from the start :|
2. fixed a typo from a recent PR
3. Added an example value to external temperature attribute.

Thanks


Koenkk:
Thanks!

PR #2995, opened 2021-09-05
Add support for Mextronic ZBZIGBEEW12365A LED Controller

jensweimann:
Add support for this device:
https://www.mextronic-shop.de/leuchten/zubehoer/led-steuerung-controller/dimmer-geraete/zigbee-controller-smarthome-einfarbige-dimmer-aktor-12-36v-dc-bis-5a/a-10910


Koenkk:
Shouldn't the brand be `iSTAR`? At least that is what the picture shows

jensweimann:
> Shouldn't the brand be `iSTAR`? At least that is what the picture shows

You're right, Mextronic is probably just a rebrand.

Koenkk:
Thanks!

PR #2998, opened 2021-09-06
Add battery support for Linkind ZS1100400-IN-V1A02

evilpie:
This motion sensor seems to support battery information. I haven't
directly tested this, but the code is just copied from the Door/Window
sensor.

debug Received Zigbee message from 'Montion Sensor', type 'attributeReport', cluster 'genPowerCfg', data '{"batteryPercentageRemaining":182}' from endpoint 1 with groupID 0
debug No converter available for 'ZS1100400-IN-V1A02' with cluster 'genPowerCfg' and type 'attributeReport' and data '{"batteryPercentageRemaining":182}'


Koenkk:
Thanks!

PR #2999, opened 2021-09-07
Detect additional TS0601 Tuya air sensor

CornerBit:
Ordered one of the Tuya indoor air quality sensors and it appears to be identical except it uses a new fingerprint. The model is often listed under AirBox02 on AliExpress.


Koenkk:
Thanks!

PR #3004, opened 2021-09-08
Fix support for SSWD01 device

Asleep-at-the-wheel:
I would like to preface this by stating that I have almost no clue what I am doing, I don't normally code and/or use git (I did most of this in vi and bash but it all seems to work). I essentially Benny Hill'd my way through everything here, so please double check.

I am attempting to fix this issue here:
https://github.com/Koenkk/zigbee2mqtt/issues/8556

The issue appears that the original lumping in of the device with a bunch of other tuya dimmers was causing issues with the unique data points this device uses. The set zigbee is fine but not the read. This device also has two new data points (minimum brightness and countdown off timers) that I have no idea how to code for but could identify. I plan to learn this over time and make the change. At this time I just want to get the devices working correctly.

I plan to submit a jpeg for this device too in the documentation.

Any feedback would be appreciated, I am a network security person in way over my head here...

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3653,11 +3653,11 @@ const converters = {
             } else if (meta.device.manufacturerName === '_TZE200_swaamsoy') {
                 // https://github.com/Koenkk/zigbee-herdsman-converters/pull/3004
                 if (msg.data.dp === 2) {
-                        return {brightness: mapNumberRange(value, 10, 1000, 0, 254)};
+                    return {brightness: mapNumberRange(value, 10, 1000, 0, 254)};
                 }
             } else { // TODO: Unknown dp, assumed value type
-                        return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
-                    }
+                    return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
+                }
         },
     },
     tuya_data_point_dump: {
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1506,25 +1506,6 @@ const converters = {
     // #endregion
 
     // #region Non-generic converters
-    ikuu_dimmer: {
-        cluster: 'manuSpecificTuya',
-        type: ['commandGetData', 'commandSetDataResponse'],
-        convert: (model, msg, publish, options, meta) => {
-            const dp = msg.data.dp; 
-            const value = tuya.getDataValue(msg.data.datatype, msg.data.data); 
-            switch (dp) {
-            case tuya.dataPoints.state:
-                    return {state: value ? 'ON': 'OFF'};
-            case tuya.dataPoints.ikuuDimmerLevel: 
-                    return {brightness: mapNumberRange(value, 10, 1000, 0, 254)};
-            // to do: 
-            // case tuya.dataPoints.ikuuMinBrightness:
-            //      return {min_brightness};
-            // case tuya.dataPoints.ikuuCountDown:
-            //      return  {seconds};
-            }
-        }
-       },
     elko_thermostat: {
         cluster: 'hvacThermostat',
         type: ['attributeReport', 'readResponse'],
@@ -3669,9 +3650,14 @@ const converters = {
             const value = tuya.getDataValue(msg.data.datatype, msg.data.data);
             if (msg.data.dp === tuya.dataPoints.state) {
                 return {state: value ? 'ON': 'OFF'};
+            } else if (meta.device.manufacturerName === '_TZE200_swaamsoy') {
+                // https://github.com/Koenkk/zigbee-herdsman-converters/pull/3004
+                if (msg.data.dp === 2) {
+                        return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
+                }
             } else { // TODO: Unknown dp, assumed value type
-                return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
-            }
+                        return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
+                    }
         },
     },
     tuya_data_point_dump: {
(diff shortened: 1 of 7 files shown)


converters/fromZigbee.js
```diff
@@ -1506,6 +1506,25 @@ const converters = {
...
+        cluster: 'manuSpecificTuya',
+        type: ['commandGetData', 'commandSetDataResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const dp = msg.data.dp; 
+            const value = tuya.getDataValue(msg.data.datatype, msg.data.data); 
+            switch (dp) {
+            case tuya.dataPoints.state:
+                    return {state: value ? 'ON': 'OFF'};
+            case tuya.dataPoints.ikuuDimmerLevel: 
+                    return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
```
Koenkk, line 1519: Could you remove the `level` value here? (since it has no added value since `brightness` is already there)

Asleep-at-the-wheel:
I believe I have done this. Sorry, I am still learning basically everything here.

I originally copied this line from the tuya dimmer section:
`return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};`

I wasn't really sure why it was there for similar devices.

A couple of questions.
1. Removing this value left the state in my Zigbee2MQTT front end. When this change eventually makes its way into my production system, how can I cleanly remove that state (in testing environment a repair fixed it, but that's not going to work easily in my real environment).
2. Where should I add the new image of the device? It looks nothing like the other Tuya dimmers it is currently associated with.

converters/fromZigbee.js
Asleep-at-the-wheel: I believe I have done this. Sorry, I am still learning basically everything here.

I originally copied this line from the tuya dimmer section:
`return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};`

I wasn't really sure why it was there for similar devices.

Koenkk:
1. remove it from `state.json`
2. Image has to be added here: https://github.com/Koenkk/zigbee2mqtt.io/tree/master/docs/images/devices
3. Do they work already? If so I can help adding the expose. If it doesn't work yet then you need to sniff traffic with the original gateway to reverse engineer how it works.

devices/tuya.js
```diff
@@ -230,8 +229,26 @@ module.exports = [
...
+    },
+            // Ikuu single gang no neutral dimmer device
+    {
+            fingerprint: [
+                {modelID: 'TS0601', manufacturerName: '_TZE200_swaamsoy'},
+            ],
+            model: 'SSWD01',
+            vendor: 'Mercator ikuü',
+            description: 'Zigbee smart dimmer',
+            fromZigbee: [fz.ikuu_dimmer, fz.ignore_basic_report],
```
Koenkk, line 242: I'm wondering if a new definition is really needed, the only difference is `fz. ikuu_dimmer ` vs `fz.tuya_dimmer`, I think it is better to just fix the `fz.tuya_dimmer` (add the datapoint for the brightness)

Asleep-at-the-wheel: I simply don't know how the other Tuya dimmer devices work, if they all have the same data points or not.

If they are all the same, then yeah that makes more sense. I didn't want to take the risk of breaking other people's stuff, but we can do that if it's best.

Koenkk: Can you try just changing:

```js
tuya_dimmer: {
    cluster: 'manuSpecificTuya',
    type: ['commandGetData', 'commandSetDataResponse'],
        const value = tuya.getDataValue(msg.data.datatype, msg.data.data);
        if (msg.data.dp === tuya.dataPoints.state) {
            return {state: value ? 'ON': 'OFF'};
        } else if (meta.device.manufacturerName === '_TZE200_swaamsoy' && msg.data.dp === 3) {
            return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
        } else { // TODO: Unknown dp, assumed value type
            return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
        }
    },
},
```

Asleep-at-the-wheel: Maybe I didn't explain the issue too well.

Data point 2 is the actual dimmer brightness level.
Data point 3 is a minimum brightness level that is set. This is because the dimming switch has no neutral wire, going lower than the minimum setting in brightness results in the load LED turning.

So when we see data point 2, we get brightness ok. But when we see data point 3, that else assumes it's brightness and sets incorrect values. So that proposed change will result in the same issue if I am not mistaken?

Would need a couple more ifs there to say:
if ikuu and dp is 2 then set brightness
elif ikuu and dp is 3 then set min-brightness
elif ikuu and dp is 6 set timer
else assume brightness

Want me to do that? But seems super hacky.

Koenkk: You are right, what about:

```js
tuya_dimmer: {
    cluster: 'manuSpecificTuya',
    type: ['commandGetData', 'commandSetDataResponse'],
        const value = tuya.getDataValue(msg.data.datatype, msg.data.data);
        if (msg.data.dp === tuya.dataPoints.state) {
            return {state: value ? 'ON': 'OFF'};
        } else {
            if (meta.device.manufacturerName === '_TZE200_swaamsoy') {
                // https://github.com/Koenkk/zigbee-herdsman-converters/pull/3004
                if (msg.data.dp === 3) {
                    return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
                }
            } else { // TODO: Unknown dp, assumed value type
                return {brightness: mapNumberRange(value, 10, 1000, 0, 254), level: value};
            }
        }
    },
},
```

Asleep-at-the-wheel: Ok, I have made that change with a minor adjustment. Does work in my testing.

Those other data points (3 and 6) I can't figure out yet, I will get to them when ever I can.

Koenkk:
Thanks!

PR #3005, opened 2021-09-08
Add TS0207_water_leak_detector

stefangries:
This device was previously recognized as CR Smart Home TS0207, which is wrong.

The now integrated Tuya TS0207 Water Leak Sensor is this device ([AliExpress](https://www.aliexpress.com/item/1005002941769280.html)):
TS0207 fw

Three questions:

- Is it correct, that I have to put the image to Koenkk/zigbee2mqtt.io/blob/master/docs/images/devices/TS0207_water_leak_detector.jpg ? Currently when using my code in an external converter Zigbee2MQTT shows [this image](https://slsys.github.io/Gateway/devices/png/TS0207.png).
- Is there anything more to do, that this new device will not be mixed up with the CR Smart Home TS0207?
- ~~The battery level is not indicated correctly at first, but only appears when you shortly press the pairing button again after successfully pairing. Is it possible that the sensor only sends the value periodically or when there is a change? Is it possible to instruct Zigbee2MQTT to actively poll the value?~~

### DIFF
--- a/devices/cr_smart_home.js
+++ b/devices/cr_smart_home.js
@@ -40,15 +40,6 @@ module.exports = [
         extend: extend.switch(),
     },
     {
-        zigbeeModel: ['TS0207', 'FNB54-WTS08ML1.0'],
-        model: 'TS0207',
-        vendor: 'CR Smart Home',
-        description: 'Water leak detector',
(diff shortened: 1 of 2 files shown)


Koenkk:
Picture should indeed be put there. I've removed the CR Smart Home definition completely because I know it is a whitelabel of TuYa.

PR #3009, opened 2021-09-10
Add Lockwood T-Lock (rebranded Yale Assure SL)

davidmonro:
The Lockwood T-Lock appears to be a Yale Assure SL with different stickers. It appears to work exactly the same, including the pin setting functionality.

Apparently the T-Lock was part of a deal with Telstra when they got into the smart home market about 5 years ago. They have since got out of that market and these locks are being sold off cheap as a result.

(Lockwood is an Australian subsidiary of Assa Abloy, which is also the parent company of Yale).

Thanks to http://leftfoot.com.au/blog/zigbee2mqtt-and-the-lockwood-t-lock for making it easy to find the external converter file to get this working in the first place.

### DIFF
--- a/devices/yale.js
+++ b/devices/yale.js
@@ -99,4 +99,14 @@ module.exports = [
         description: 'Assure lock SL',
         extend: lockExtend,
     },
+    {
+        // Appears to be a slightly rebranded Assure lock SL
+        // Just with Lockwood | Assa Abloy branding instead of Yale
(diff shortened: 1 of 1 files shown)


davidmonro:
Would you like this as a white-label version in the docs?

Koenkk:
Thanks! No need to add a whitelabel version to the docs (since it can be distinguished from the original device due to using a different `zigbeeModel`.

PR #3011, opened 2021-09-10
Add support for E26 version of LED1937T5

stanleylai:
E26 version is essentially the same as the E27 one, and tested to be fully functional. Just needed a slight change to zigbeeModel.


Koenkk:
Thanks!

PR #3014, opened 2021-09-10
Add support for SMaBiT devices + bugfixes

mhaid:
New devices:
AV2010/14
AV2010/16
AV2010/18
AV2010/21B
AV2010/22B
AV2010/33
AV2010/37


converters/toZigbee.js
```diff
@@ -1149,7 +1149,7 @@ const converters = {
         },
     },
     thermostat_keypad_lockout: {
-        key: ['keypad_lockout'],
+        key: ['keypad_lockout', 'keypad_lockout_2'],
```
Koenkk, line 1152: Why was this added?

Koenkk:
Note to self, after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/897

converters/toZigbee.js
mhaid: The device AV2010/32 uses the "Level 2 lockout" value for the KeypadLockout Attribute.
(For reference: Page 396 in the ZigBee cluster specifications.)

In the toZigbee Converter only level1 is currently implemented.

mhaid: And for why I added a new expose-entry:

keypad_lockout is implemented as a binary with the values 0 and 1.

keypad_lockout2 is newly implemented as a binary with the values 0 and 2.

Koenkk: A better solution is to remove the `keypad_lockout_1/2` and change the `keypad_lockout` to an enum instead (look in `exposes.js` for examples)

mhaid: You're right. I changed keypad_lockout from binary to enum and removed keypad_lockout_2.

PR #3023, opened 2021-09-12
Elko Super TR support for new attributes

owangen:
Added support for custom ELKO attributes: Regulatormode, regulatortime and night switching


Koenkk:
Thanks!

PR #3024, opened 2021-09-13
Add pin functionality to Schlage lock

protochron:
It turns out that the Schlage BE468 also supports changing the PIN codes
via Zigbee.

Add support for getting/setting PIN codes by adding the appropriate
`toZigbee` and `fromZigbee` calls along with the correctly exported
functions.


Koenkk:
Thanks!

PR #3028, opened 2021-09-15
Support for E3007 Ikea Starkvind air purifier

derfloDev:
Support for the new Ikea Starkvind air purifier.
At the moment only the fan control is working.
The device also has a air quality sensor (heimanSpecificPM25Measurement) but I wasn't able to get it work.

### DIFF
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -613,6 +613,6 @@ module.exports = [
         description: 'STARKVIND air purifier',
         exposes: [e.fan().withModes(['low', 'medium', 'high', 'on', 'auto'])],
         fromZigbee: [fz.fan],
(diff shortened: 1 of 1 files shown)


danieledwardgeorgehitchcock:
Have you got the database.db entry for this device so that we can see the clusters and may be able to help you getting the other features together?

Are you still getting debug warning messages about those features not being supported?

derfloDev:
@danieledwardgeorgehitchcock Thank for your help. I already opened a discussion https://github.com/Koenkk/zigbee2mqtt/discussions/8744. There you can see the database.db entry.

I never got any debug logs of not supported features. I don't know why. (Yes I turned debug level on xD)

PR #3029, opened 2021-09-15
E3007 Ikea Starkvind air purifier fanMode

derfloDev:
Support for fanMode on Ikea Starkvind air purifier


sjorge:
https://github.com/Koenkk/zigbee2mqtt/discussions/8744#discussioncomment-1333154 you probably also want to update the expose line as it's still the old one here.

PR #3033, opened 2021-09-16
Fix support for SSW01

Asleep-at-the-wheel:
As per here:
https://github.com/Koenkk/zigbee2mqtt/issues/8676

Again, please triple check this, don't normally code or use git; not to mention the janktastic arrangment of my testing environment and how I pushed this exact code.

If I put an image of the device here by model name: https://github.com/Koenkk/zigbee2mqtt.io/tree/master/docs/images/devices will that override the current image on the site that uses the chip set name?


Koenkk:
Thanks!

PR #3035, opened 2021-09-16
HS2DB Changes To Map Action

danieledwardgeorgehitchcock:
This is to resolve issues reported in this thread Koenkk/zigbee2mqtt#8746 whereby the Heiman doorbell reports a separate zoneStatus key (32768) as opposed to the standard TuYa key used in the Lidl version (1)


Koenkk:
Thanks!

PR #3039, opened 2021-09-17
Update Hue A19 1100 Lumen from extend to hueExtend

narc792:
Changing Hue A19 1100 lumen from extend to hueExtend for additional control of power on behavior.


Koenkk:
Thanks!

PR #3041, opened 2021-09-17
[Moes] ZTS-EU : Add converter for 'commandSetTimeRequest'

Rockman18:
ZTS_EU_1gang

Add onEvent : tuya.onEventSetLocalTime for Moes ZTS-EU switches to correct this log message


Koenkk:
Thanks!

PR #3042, opened 2021-09-18
Add initial support for Plaid Systems Spruce plant water sensor

nkasrawi:
This is working for me, though unfortunately the sensor is reported as a humidity sensor not a water moisture sensor, which would have been nice.  I also haven't yet been able to figure out how to get the battery data to report, but I will work on that in a further pull request.


Koenkk:
Thanks!

PR #3043, opened 2021-09-18
Fix battery reporting and address an issue with use of fahrenheit as the unit

nkasrawi:
I recently installed a Centralite Pearl thermostat and I'm in the US so of course we use Fahrenheit here.  The battery was always reporting 27% capacity even when the batteries were fresh, so I made a change which now reports 100% on fresh batteries - I'm still waiting to see what happens if/when the batteries drain (they are a battery backup so it may be a while).

I also ran into the issue as a result of using Fahrenheit.  Since the zigbee spec requires the use of Celsius for temperatures, the input values for the setpoints have to be set in Celsius.  In the case where the user supplied input values are fahrenherit, the automation software (in my case openhab) is responsible for converting them into Celsius.  However, the logic in the herdsman converters truncates input Celsius values to +/- 0.5 by doing:

```Math.round((value * 2).toFixed(1)) / 2```

This resulted in an unfortunate situation because the final effective Fahrenheit value that was being sent to the thermostat in Celsius was often different from the original provided integer value - sometimes by almost one full degree fahrenheit.

My revised solution based on @Koenkk suggestion is to introduce a configuration variable which can be set on the thermostat configuration and which controls conditional logic which will either perform the current logic (if using Celsius) or perform the following alternate logic (if using Fahrenheit):

1) Convert the input Celsius value back to Fahrenheit
2) Apply the same rounding logic to truncate the input Fahrenheit value to 0.0/0.5 as appropriate
3) Convert the rounded Fahrenheit value back to Celsius and round to 2 decimals

This will ensure that the final read back Celsius value, when converted to Fahrenheit will be aligned to 0.0/0.5.

To trigger this specific behavior, include a configuration value for the specific device:

``` 
 '0x000d6f000b4f5868':
    friendly_name: upstairs_thermostat
    thermostat_unit: 'fahrenheit'
```

I also removed the references to unoccupied setpoints since this thermostat doesn't understand that concept.

### CODE REVIEW

**Comment by Koenkk** [comment]
The Zigbee spec is in celsius, therefore all Zigbee devices are controlled through celsius.

Instead of adding a new converter, I suggest to take a more generic approach by modifying the existing converters (e.g. `thermostat_occupied_heating_setpoint`) by making the unit configurable. This can be done by:

- To your configuration.yaml

```yaml
devices:
  '0x12345678':
    friendly_name: 'my thermostat'
    thermostat_unit: 'fahrenheit'

- For the `toZigbee` converters you can access this option via: `meta.options.thermostat_unit`
- For the `fromZigee` converters it is: `options.thermostat_unit`
- If the option is set, convert from Fahrenheit -> celsius.

File: converters/toZigbee.js [hunk 1]
```diff
@@ -28,6 +28,13 @@ const manufacturerOptions = {
     viessmann: {manufacturerCode: herdsman.Zcl.ManufacturerCode.VIESSMAN_ELEKTRO},
 };
 
+function normalizeCelsiusVersionOfFahrenheit(value) {
```
Koenkk, line 31: Please move this function to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/lib/utils.js

converters/toZigbee.js
```diff
@@ -5683,6 +5714,26 @@ const converters = {
...
+        key: ['occupied_heating_setpoint'],
+        convertSet: async (entity, key, value, meta) => {
+            const occupiedHeatingSetpoint = (Math.round((value * 10).toFixed(1)) / 10).toFixed(1) * 100;
+            await entity.write('hvacThermostat', {occupiedHeatingSetpoint});
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('hvacThermostat', ['occupiedHeatingSetpoint']);
+        },
+    },
+    centralite_thermostat_occupied_cooling_setpoint: {
```
Koenkk, line 5727: These two seem to be unused now, please remove

devices/centralite.js
```diff
@@ -146,15 +146,16 @@ module.exports = [
         vendor: 'Centralite',
         description: '3-Series pearl touch thermostat,',
         fromZigbee: [fz.battery, fz.legacy.thermostat_att_report, fz.fan, fz.ignore_time_read],
-        toZigbee: [tz.factory_reset, tz.thermostat_local_temperature, tz.thermostat_local_temperature_calibration, tz.thermostat_occupancy,
-            tz.thermostat_occupied_heating_setpoint, tz.thermostat_occupied_cooling_setpoint, tz.thermostat_unoccupied_heating_setpoint,
-            tz.thermostat_unoccupied_cooling_setpoint, tz.thermostat_setpoint_raise_lower, tz.thermostat_remote_sensing,
-            tz.thermostat_control_sequence_of_operation, tz.thermostat_system_mode, tz.thermostat_weekly_schedule,
-            tz.thermostat_clear_weekly_schedule, tz.thermostat_relay_status_log, tz.fan_mode, tz.thermostat_running_state],
+        toZigbee: [tz.factory_reset, tz.thermostat_local_temperature, tz.thermostat_local_temperature_calibration,
```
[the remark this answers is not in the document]
nkasrawi: Yes - this device doesn't support thermostat_occupancy, thermostat_unoccupied_cooling_setpoint, thermostat_unoccupied_heating_setpoint, thermostat_weekly_schedule, thermostat_clear_weekly_schedule, thermostat_weekly_schedule - which I confirmed by trying to get/set those values through the dev console.

converters/toZigbee.js
nkasrawi: Done.

converters/toZigbee.js
nkasrawi: Done.

Koenkk:
Thanks!

PR #3051, opened 2021-09-19
Add a new fingerprint for the smart air box

nanassito:
Just got my hands on the smart air box and it has a fingerprint that isn't already in this list.


nanassito:
oh my bad it was already added

PR #3054, opened 2021-09-20
plaid.js: fix relative import

Hacker-CB:
@Koenkk - it will be better to add some checks for relative import in device converters ;)


Koenkk:
Thanks!

PR #3056, opened 2021-09-20
Philips LTC005

yhur:
I noticed this Philips Hue is not recognized by the Zigbee2MQTT, so did some comparison, came up with this modification on the philips.js. Please take a look and merge it if there is no problem.

[Philips Hue Within Ceiling Lamp](https://www.philips-hue.com/ko-kr/p/hue-white-ambiance-within-ceiling-light/4505548C5)
Screen Shot 2021-09-20 at 11 12 17 PM

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1067,7 +1067,7 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp(),
     },
     {
-        zigbeeModel: ['3216131P6', 'LTC005],
+        zigbeeModel: ['3216131P6', 'LTC005'],
         model: '3216131P6',
         vendor: 'Philips',
         description: 'Hue white ambiance Aurelle square panel light',
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1046,7 +1046,7 @@ module.exports = [
         vendor: 'Philips',
         description: 'Hue ambiance ceiling',                
         meta: {turnsOffAtBrightness1: true},                    
-        extend: hueExtend.light_onoff_brightness_colortemp(),                       
+        extend: hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 454]}),
     },  
     {
         zigbeeModel: ['LTC011'],
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1040,6 +1040,14 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp(),
         ota: ota.zigbeeOTA,
     },
+    {                                           
+        zigbeeModel: ['LTC005'],                      
+        model: '3216131P6',                                                         
+        vendor: 'Philips',
+        description: 'Hue ambiance ceiling',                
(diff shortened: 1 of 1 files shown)


devices/philips.js
```diff
@@ -1040,6 +1040,14 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp(),
         ota: ota.zigbeeOTA,
     },
+    {                                           
+        zigbeeModel: ['LTC005'],                      
+        model: '3216131P6',                                                         
+        vendor: 'Philips',
+        description: 'Hue ambiance ceiling',                
+        meta: {turnsOffAtBrightness1: true},                    
+        extend: hueExtend.light_onoff_brightness_colortemp(),
```
Koenkk, line 1049: Please add the color temperature range as described on https://www.zigbee2mqtt.io/how_tos/how_to_support_new_devices.html#31-retrieving-color-temperature-range-only-required-for-lights-which-support-color-temperature

yhur: Hi, @Koenkk,
Thank you for the review and the link to the information. I just updated and pushed to my repo after querying through the mqtt.

Would you take a look and let me know if you want me to reopen this or close this one and open a new PR?

devices/philips.js
```diff
@@ -1065,7 +1073,6 @@ module.exports = [
         description: 'Hue white ambiance Aurelle square panel light',
         meta: {turnsOffAtBrightness1: true},
         extend: hueExtend.light_onoff_brightness_colortemp(),
-        ota: ota.zigbeeOTA,
```
Koenkk, line 1068: Why was this removed?

devices/philips.js
```diff
@@ -1040,6 +1040,14 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp(),
         ota: ota.zigbeeOTA,
     },
+    {                                           
+        zigbeeModel: ['LTC005'],                      
+        model: '3216131P6',                                                         
+        vendor: 'Philips',
+        description: 'Hue ambiance ceiling',                
+        meta: {turnsOffAtBrightness1: true},                    
+        extend: hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 454]}),
```
Koenkk, line 1049: Please also add `ota: ota.zigbeeOTA,` here

devices/philips.js
yhur: The reason I removed is as follows.
1. To code, I talked with Sygnify(Philips Hue) but couldn’t get the proper information from them.
2. And they said this model, ie LTC005 is withdrawn from the market so no more development.
3. So I played with the definitions for other Synify models to come up with working set for LTC005
4. I selected 3216131P6 definition as the base and modified since the icon is similar to LTC005 
5. Since I couldn’t get no development/patch related information from Sygnify, I removed the OTA to prevent the wrong update.

devices/philips.js
Koenkk: Makes sense, thanks!

PR #3058, opened 2021-09-20
Added Eglo 33955

p0wertiger:
This is an iteration of AwoX product. Bought it in Polish hardware store (on sale for whatever reason since it's quite a good lamp). Previous implementation was giving me a generic bulb without color temperature so had to add a new configuration. Also includes color temperature range.

https://www.eglo.com/uk/ceiling-light-turcona-c-98565.html


Koenkk:
Thanks!

PR #3063, opened 2021-09-21
Adding Jasco Enbrighten ZB4102 plug and Philips LTA008 bulb

fifthecho:


devices/enbrighten.js
```diff
@@ -88,4 +88,16 @@ module.exports = [
             await reporting.onOff(endpoint);
         },
     },
+    {
+        zigbeeModel: ['43094'],
+        model: '43094',
+        vendor: 'Jasco Products',
```
Koenkk, line 94: Please use `Enbrighten` here or move it to a new file (`jasco.js`)

fifthecho: Dumb question (as it's my first Zigbee code) - the device advertises itself as "Jasco Products" - would this be a problem? Or is this just an identifier used by Zigbee2MQTT?

Koenkk: It is just a cosmetic thing, seems it should use `Enbrighten ` (https://www.youtube.com/watch?v=e4TMSGBuLLw&ab_channel=JascoProductsCompany)

fifthecho: Cool. Done and pushed.

Koenkk:
Thanks!

PR #3064, opened 2021-09-21
Rudimentary support for Neo NAS-PD07

plysdyret:
Just trying to get the plumbing in place so people with better knowledge than me can complete adding the device.

This bit works for humidity, temperature and occupancy.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2177,6 +2177,45 @@ const converters = {
             return result;
         },
     },
+    neo_nas_pd07: {
+                cluster: 'manuSpecificTuya',
+                type: ['commandSetDataResponse', 'commandGetData'],
+                convert: (model, msg, publish, options, meta) => {
+                const dp = msg.data.dp;
+
+                switch (dp) {
+
+                        case 101: {
+                                const pohyb = msg.data.data[0];
+                                if (pohyb > 0) {
+                                        return {occupancy: true};
+                                } else {
+                                        return {occupancy: false};
+                                }
+                        }
+
+                        case 102: {
+                                return {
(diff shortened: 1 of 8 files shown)


converters/fromZigbee.js
```diff
@@ -2177,6 +2177,27 @@ const converters = {
...
+    neo_nas_pd07: {
+        cluster: 'manuSpecificTuya',
+        type: ['commandSetDataResponse', 'commandGetData'],
+        convert: (model, msg, publish, options, meta) => {
+            const dp = msg.data.dp;
+            if (dp === 101) return {occupancy: msg.data.data[0] > 0};
+            else if (dp === 102) {
+                return {
+                    power_type: {0: 'battery_full', 1: 'battery_high', 2: 'battery_medium', 3: 'battery_low', 4: 'usb'}[value],
+                    battery_low: value === 3,
```
Koenkk, line 2189: `value` is undefined here (also line above), please fix.

plysdyret:
I hope it ended up correctly, I fumbled the merge a number of times.

mrTuomoK:
Hey @plysdyret could you also add handler for `dp === 103`, it's for the tamper... so add also` e.tamper()` to `exposes: []`. 

Like this for example `return {tamper: msg.data.data[0] > 0 ? true : false};`

I just tested on my device and seems to work.

Koenkk:
Thanks!

plysdyret:
> Thanks!

Thank you, great project =)

Leodjmao:
Hi !
The temperature is good until 25°C. From 25,1°C, it displays 0,1°C.

also, dp107 seems to be temperature min, 108 for max, 109 & 110 for humidity.
dp 111 or dp 112 I don't remember which one is for temperature alarm.
I think that the other is for humidity.

I don't see if the led can be deactivated.

Thanks for your job .

PR #3079, opened 2021-09-24
Apply red color fix to TS0503A

freszu:
When integrating this device it was impossible to set red color after a bit of research i've configured custom device with red color workaround present on other tuya devices which solved the issue.


Koenkk:
Thanks!

PR #3095, opened 2021-09-29
Renamed Starkvind from E3007 to E2007

derfloDev:
see https://github.com/Koenkk/zigbee2mqtt.io/pull/905#issuecomment-929508971


Koenkk:
Thanks!

PR #3104, opened 2021-10-01
Add support of Keypad KEPZB-110 from Develco/Frient

ioull:
Integration of the alarm panel / keypad KEPZB-110 from Develco/Frient :
https://www.develcoproducts.com/products/sensors-and-alarms/keypad/


Koenkk:
Thanks!

PR #3082, opened 2021-09-25
Move _TZ3000_dpo1ysak to TS011F_plug

IsThisUsernameFree:
According to seller, it supports power monitoring capabilities.

Screenshot is what I bought and I received `TS011F` with `_TZ3000_dpo1ysak` vendor.

PR #3076, opened 2021-09-24
Add support for TS011F/_TZ3000_dpo1ysak

dmaasland:
This PR adds the manufacturer "_TZ3000_dpo1ysak" for the TS011F plug. It's just another whitelabel product so nothing special. The [product page](https://aliexpress.com/item/1005002344798281.html) says it should have power metering but it doesn't :( .

PR #3107, opened 2021-10-01
Update tuya.js

kloknibor:
Probably fixing:

#3076 #3082


Koenkk:
This should only be done for the which do not support the configure reporting (since polling increases stress on the network and will provide less frequent updates). What are the `manufacturerName` of the devices that need this?

Z18Aplha:
I agree to Koenkk. 
This device supports reporting.

dmaasland:
The device reports itself as a TS011F with manufacturer _TZ3000_dpo1ysak. It does seem to behave differently for different people though. Perhaps a firmware difference or whitelabling issue. Mine does support power metering when queried manually via the dev console but haven't been able to get automatic reporting to work yet.

kloknibor:
I changed it to only poll for the '_TZ3000_dpo1ysak' device. @Koenkk can you re-check this PR? Thanks!

devices/tuya.js
```diff
@@ -788,6 +788,43 @@ module.exports = [
...
+        fingerprint: [{modelID: 'TS011F', manufacturerName: '_TZ3000_dpo1ysak'},
+        model: 'TS011F_plug',
+        description: 'Smart plug (with power monitoring by polling)',
+        vendor: 'TuYa',
+        fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fz.ignore_basic_report, fz.tuya_switch_power_outage_memory],
+        toZigbee: [tz.on_off, tz.tuya_switch_power_outage_memory],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement', 'seMetering']);
+            await reporting.rmsVoltage(endpoint, {change: 5});
```
Koenkk, line 801: I think all reporting setup can be removed here (it doesn't work anyway)

kloknibor: like this?

kloknibor:
Like this?

JohnRabbit1:
I think line 769 should be deleted (but probably you already did)

Koenkk:
Thanks!

KMDobrev:
Hallo,

I just did an update.  But still no energy measurement.

Screenshot_20211101-231748
Screenshot_20211101-231736

PR #3113, opened 2021-10-02
Add Philips Hue Still model 3261348P6

philipkocanda:
I noticed this model wasn't supported yet. Tested and working just like all other Hue white ambiance devices.


Koenkk:
Thanks!

PR #3121, opened 2021-10-04
Update neo.js - add battery_low to exposes for NAS-PD07

Masterz69:
Expose battery_low (reg. [#8996](https://github.com/Koenkk/zigbee2mqtt/issues/8996)).

As far I see - battery_low is calculated, published to MQTT, but not exposed as device property, thus not created in HA as entity.
If I understood correctly - e.battery_low() to be added to exposes array in same way as for NAS-AB02B0.


Koenkk:
Thanks!

PR #3129, opened 2021-10-06
Add support for `TS000x` Tuya switch modules

bogdal:
This PR extends the functionality of the latest Tuya switch modules (1CH, 2CH). Currently, z2m detects them as wall switches with limited options.

This PR requires [0xe001](https://github.com/Koenkk/zigbee-herdsman/pull/430) cluster.

2021-10-06 at 12 03

This is how it looks like in the dashboard:

2021-10-06 at 12 06
2021-10-06 at 15 11

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3728,6 +3728,16 @@ const converters = {
             });
         },
     },
+    tuya_switch_type = {
+        cluster: 'manuSpecificTuya_3',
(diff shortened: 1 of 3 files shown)


Koenkk:
Thanks!

protyposis:
Is there a possibility that `power_on_behavior` and `switch_type` could also be configured separately per endpoint, or did you rule that out?

PR #3134, opened 2021-10-07
add support for LWA017

fgather:
add support for new Hue white A60 bulb E27 1050lm with Bluetooth from Philips (LWA017)


fgather:
Thanks for this great library!

PR #3139, opened 2021-10-08
ADD New device Module _TZ3000_lnggrqqi - Update tuya.js

DavidFFerreira:
New module for same circuite breacker
Can u upload asap?

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -92,7 +92,7 @@ module.exports = [
         extend: extend.switch(),
     },
     {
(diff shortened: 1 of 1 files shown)


DavidFFerreira:
@Koenkk can u update the info on website too?

DavidFFerreira:
@Koenkk the manufacture name
_TZ3000_lnggrqqi

Koenkk:
We don't show `manufacturerName` on the website, I believe it doesn't have any added value. You cannot know the `manufacturerName` when buying the device, the only way is pairing it (and then you already found out if it is supported or not)

PR #3144, opened 2021-10-08
Add Schneider WDE002386 push button dimmer

tanordheim:
Adds support for the [Schneider Electric WDE002386 push button dimmer](https://www.se.com/ww/en/product/WDE002386/dimmer%2C-wiser%2C-exxact%2C-push-button%2C-universal%2C-led%2C-zigbee%2C-white/). This is essentially the same as the rotary dimmer introduced in #2544.

This is also sold in Norway under the name [Elko SmartDim Trykk Uni](https://elkosmart.elko.no/produkter/smartdim-trykk-uni-200w). I tested and verified this device integration using the Elko branded version, but as far as I can tell these should work identically.

### DIFF
--- a/devices/schneider_electric.js
+++ b/devices/schneider_electric.js
@@ -108,6 +108,25 @@ module.exports = [
         },
     },
     {
+        zigbeeModel: ['NHPB/DIMMER/1'],
+        model: 'WDE002386',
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

PR #3148, opened 2021-10-10
Add HK-DIM-A

kangaroomadman:
Add the HK-DIM-A Candeo device.

https://www.amazon.co.uk/Candeo-compatible-dimmable-incandescent-multiway/dp/B091HP42S8

### DIFF
--- a/devices/candeo.js
+++ b/devices/candeo.js
@@ -1,0 +1,19 @@
+const reporting = require('../lib/reporting');
+const extend = require('../lib/extend');
+
+module.exports = [
+    {
+        zigbeeModel: ['HK-DIM-A'],
+        model: 'HK-DIM-A',
+        vendor: 'Candeo',
+        description: 'ZigBee LED Dimmer Smart Switch',
+        extend: extend.light_onoff_brightness({noConfigure: true, disableEffect: true}),
+        configure: async (device, coordinatorEndpoint, logger) => {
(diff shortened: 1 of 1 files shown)


devices/candeo.js
```diff
@@ -0,0 +1,19 @@
...
+        model: 'HK-DIM-A',
+        vendor: 'Candeo',
+        description: 'ZigBee LED Dimmer Smart Switch',
+        extend: extend.light_onoff_brightness({noConfigure: true, disableEffect: true}),
+        configure: async (device, coordinatorEndpoint, logger) => {
+            await extend.light_onoff_brightness().configure(device, coordinatorEndpoint, logger);
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl']);
+            await reporting.onOff(endpoint);
+            await reporting.brightness(endpoint);
```
Koenkk, line 16: We cannot prevent reporting when controlled through z2m, but we can reduce the interval, change this line to: `await reporting.brightness(endpoint, {min: 3});` (device will report at least with intervals of 3 seconds now)

kangaroomadman: Thanks.

I've tested ```{min: 1}``` and it seems a reasonable compromise when you want to use the reported brightness for automations and don't want a 3 second delay.

PR #3159, opened 2021-10-12
Add a converter for the Alecto SMART-HEAT10 thermostatic radiator

jap:
This device is sold as Tuya compatible, and after some probing it
appears to look quite similar to the device in siterwell.js but lacks
some features.

One dpid is still a mystery; this device sends out the following:

2021-10-12 22:52:04: Received Zigbee message from 'zolder_thermostaat', type 'commandGetData', cluster 'manuSpecificTuya', data '{"data":{"data":[0],"type":"Buffer"},"datatype":4,"dp":17,"fn":0,"status":0,"transid":166}' from endpoint 1 with groupID 0
zigbee-herdsman-converters:tuyaThermostat: Unrecognized DP #17 with data {"status":0,"transid":166,"dp":17,"datatype":4,"fn":0,"data":{"type":"Buffer","data":[0]}}

It always seems to be zero.

Unfortunately the vendor was unable to offer more details.

### DIFF
--- a/devices/alecto.js
+++ b/devices/alecto.js
@@ -1,0 +1,26 @@
+const exposes = require('../lib/exposes');
+const fz = {...require('../converters/fromZigbee'), legacy: require('../lib/legacy').fromZigbee};
+const tz = require('../converters/toZigbee');
+const tuya = require('../lib/tuya');
+const e = exposes.presets;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #3170, opened 2021-10-15
Add support for Silvercrest TS0601 TRV variant

datalexum:
This PR adds support for the Silvercrest Zigbee thermostat sold by Lidl and others. It's simply a variant of the Tuya TS0601 thermostat so it was sufficient to only add the different manufacurerName (_TZE200_chyvmhay) to the tuya.js.
With this change the device gets detected and you can interact with it just like with any other TS0601 thermostat.


ciB89:
I would also like to help, but I really struggle to understand entirely what I need to do. I have a few of the devices here, so if any one could guide me on how to help I would love to.

ciB89:
I can now also confirm, this is not working at all.

PR #3174, opened 2021-10-15
Support for Schneider WDE002906

manfred-mp:
1-gang switch support. This switch is using dimming (move) with rate 255 which is not optimal. I will contact to Schneider Electric if they will change that or expose DefaultMoveRate for writing.


Koenkk:
Thanks!

PR #3175, opened 2021-10-15
Add Hue White & Color Ambiance Centris ceiling light (3 spots)

agross:
We already have a similar light, but in this configuration, the 3 spots are mounted linear. The other variant looks like a cross.


Koenkk:
Thanks!

PR #3176, opened 2021-10-15
Fix setpoint_change_source for danfoss and popp

kennylevinsen:
This entity was exposed as a binary even though it reports a string.
Convert the entity to an enum instead.


Koenkk:
Thanks!

PR #3185, opened 2021-10-18
Basic functions of TV01-ZB Moes TRV

kirovilya:
https://aliexpress.ru/item/1005003136135923.html

similar issue
https://github.com/Koenkk/zigbee2mqtt/issues/8522


lib/tuya.js
```diff
@@ -478,6 +505,12 @@ const msLookups = {
     },
 };
 
+const tvThermostatMode = {
+    0: 'auto',
+    1: 'manual',
+    3: 'holiday',
```
guandalf, line 511: Are you sure this shouldn't have been 2: 'holiday'? Is this maybe causing https://github.com/Koenkk/zigbee2mqtt/issues/16648 ?

PR #3189, opened 2021-10-19
Perenio PECLS01 water leak sensor converter

ShchirinAleks:
Added Perenio vendor-specific file with the converter for water leakage sensor PECLS01


Koenkk:
Thanks!

PR #3209, opened 2021-10-23
Add support for TS1101 (2CH) Tuya dimmer module

bogdal:
This PR adds support for one of the latest Tuya dimmer modules (2CH).

2021-10-23 at 02 12

2021-10-23 at 02 15

2021-10-23 at 10 22


bogdal:
Hey @D3M80L, I didn't observe the issue with min brightness that you mentioned on my side. Values shouldn't jump. Are you able to check the behavior of integration with different module? 

The minimum brightness tells the modules to which value the brightness can be reduced using a physical switch (momentary).
Some lights may flash when the brightness is too low.

PR #3217, opened 2021-10-24
Adding support for Danfoss Icon floor heating thermostat and regulator

nagydavid:
Adding support for Danfoss Icon floor heating thermostat and regulator.

### DIFF
--- a/devices/danfoss.js
+++ b/devices/danfoss.js
@@ -123,7 +123,6 @@ module.exports = [
         },
     },
     {
-        // Danfoss Icon Floor heating thermostats
         fingerprint: [
             {modelID: '0x8020', manufacturerName: 'Danfoss'}, // RT24V Display
             {modelID: '0x8021', manufacturerName: 'Danfoss'}, // RT24V Display  Floor sensor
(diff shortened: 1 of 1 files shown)


nagydavid:
Hi All, 

So after testing the devices for a week, I encountered the following issue.:

- It is a bit struggling to pair the device. It has to be done multiple times, because the device gets disconnected and leaving the network. 
- So multiple non exiting entities are exposed due to the 15 setting of multiple device. I wonder cannot be sent through the UI how much entity we would like to expose. 
- The biggest problem is, that I am unable to control the device through home assistant. Even though I have multiple climate entities, when ever I am trying to set the a new temperature temperature for one of the climate entities only the first climate entity will be changed.

nagydavid:
So this the output, if I am sending set_temperature from HA. Even though I am sending the request to endpoint 5, home assistant sends it to endpoint 1. As you can see the topic is not including the endpoint name.
```
Zigbee2MQTT:debug 2021-11-08 14:54:43: Received MQTT message on 'zigbee2mqtt/Heating System/set/occupied_heating_setpoint' with data '25.5'
Zigbee2MQTT:debug 2021-11-08 14:54:43: Publishing 'set' 'occupied_heating_setpoint' to 'Heating System'
Zigbee2MQTT:debug 2021-11-08 14:54:43: Received Zigbee message from 'Heating System', type 'attributeReport', cluster 'hvacThermostat', data '{"occupiedHeatingSetpoint":2550}' from endpoint 1 with groupID 0
```

robertalexa:
@Koenkk Could you please help with the above?

If this related to what you did for the Hive SLR2 recently? See commit here https://github.com/Koenkk/zigbee2mqtt/commit/a41c00faea66cb14cb7002ca30c66245c0b76218

Koenkk:
I think I've fixed the issue (it was not related to that commit).

@nagydavid could you check?

nagydavid:
Hi @Koenkk and @robertalexa 

After the changes in the edge, everything works as it should

PR #3218, opened 2021-10-24
Update LED1624G9 to support white spectrum

Zufallsfuchs:
Product should support both color and white spectrum controls.
Changed in ikea.js to reflect the change and remove code that became unnecessary, as it is included in tradfriExtend.
Unsure about if the meta -> supportsHueAndSaturation line 198 is still required, as it is not on any other lamp and I am unsure of its purpose.

### DIFF
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -194,10 +194,8 @@ module.exports = [
         model: 'LED1624G9',
         vendor: 'IKEA',
         description: 'TRADFRI LED bulb E14/E26/E27 600 lumen, dimmable, color, opal white',
-        extend: extend.light_onoff_brightness_color(),
-        ota: ota.tradfri,
+        extend: tradfriExtend.light_onoff_brightness_colortemp_color(),
         meta: {supportsHueAndSaturation: false},
-        onEvent: bulbOnEvent,
     },
     {
         zigbeeModel: ['TRADFRI bulb E26 CWS 800lm', 'TRADFRI bulb E27 CWS 806lm'],


ejjoman:
I have some (9) LED1624G9 deployed here at home and hyped af I installed this update in the hope to be able to control the white spectrum of the bulbs.
The new controls are visible now in the z2m UI, but they do nothing for me.
My bulbs are on the firmware version 1.3.013 from 2020-02-20. Is there anything else I have to do in order to be able to use the new controls?

Thanks!

Zufallsfuchs:
You‘re right, @ejjoman 
The update didn’t work for me either. The buttons are there but the light doesn’t respond to color temp changes.
@Koenkk there still seems to be something wrong with the configuration. I’m 100% sure the lights support white spectrum but yet they don’t respond. Any ideas?

Koenkk:
@Zufallsfuchs have you confirmed color temp works with the original gateway?

Zufallsfuchs:
@Koenkk No, sadly I don't have the original gateway to test, but I remember the packaging and website stating the light supports both color and white spectrum. Plus, when resetting the light, it turns to a non-RGB-mixed true-white color. (There's also a teardown online showing warm white, cold white and color LEDs: https://fccid.io/FHO-LED1624G9/Internal-Photos/Internal-Photos-3631433)

Maybe @ejjoman can confirm using the original gateway?

tomaszpieczykolan:
For me changing the color temp is also broken, which means that the white spectrum pickers in Home Assistant and HomeKit are not working. I think we might need to reverse this change?

marthubner:
I also think this change should be reverted because changing color temperature is not working.

I also found out that after recent update to firmware 2.3.086 the bulb now cycles through 9 built-in scenes when pressing left and right arrows on E1524/E1810 controller when bound to the bulb to Scenes cluster.

When cycling through these built-in scenes using the E1810 controller, the bulb was changing colors but these changes were not reported to zigbee2mqtt until I set Reporting of the bulb as seen here in the screenshot.

Screenshot 2021-12-19 16 03 37

(Originally there was only OnOff and LevelCtrl)

Now when pressing left or right button on the controller, the bulb changes color and also reports current x and y color and current scene to zigbee2mqtt but there is a converter missing for built-in scenes as seen in the log:

Screenshot 2021-12-19 16 24 31
`Debug No converter available for 'LED1624G9' with cluster 'genScenes' and type 'attributeReport' and data '{"currentScene":3}'`

{"currentScene":X} cycles from 0 to 8

The same probably applies to other IKEA bulbs which received the latest firmware update. For example my LED1903C5/LED1835C6 now reports currentScene from 0 to 2 and cycles between 3 colors when using the setting I described above. (It si only a white spectrum bulb)

@Koenkk Is it possible to make these bulbs report their color and scene automatically when paired?
Also I am probably not able to make the missing converter but I can provide the data if someone would want to do it.

tomaszpieczykolan:
@Koenkk It's been quite a while now with this annoying bug, can this MR be reversed?

Koenkk:
@tomaszpieczykolan reverted it now. 

PR #3221, opened 2021-10-25
Revert "Add support for Haozee TS0601 TRV variant"

Highvolt:
After some additional testing it's not working properly, some datapoints don't correctly map with the other TRV, including the battery information crashing the UI.
I am trying to work out the problems but for now it is best to remove this from the repo.

Sorry for the premature buggy commit.


Gofoodi:
@Highvolt did you have any success already? I also bought this thermostat and also for me the integration is not working properly yet.

Issue #3190, opened 2021-10-19
Support for Rti-Tek TRV 601 (_TZE200_a4bpgplm)


Oddiwahn:
Hi everyone,

I bought some TRVs [from Aliexpress](https://www.aliexpress.com/item/1005003172662104.html?spm=a2g0s.9042311.0.0.27424c4dM8naAg%5B%5D(url)) which I'd like to add support for.

So far I am planning to add the following data points, since they do no match with any already known ones (like Moes or Saswell). Just in case anybody is currently working on this as well, I'd like to share my progress here.

EDIT: final list of datapoints:

|DP|OUT|IN|EXP|data point name|label in iot.tuya|example as shown in iot.tuya|
|--|--|--|--|----------------|---------------------|---------------------|
|1|x|x|x|rtitekSystemMode|Modus|Manual, OFF, Auto, ON|
|2|x|x|x|rtitekHeatingSetpoint|Set temperature|15.00?|
|3||x|x|rtitekLocalTemp|Current temperature|18.90?|
|4||||rtitekBoostHeating|Rapid heating|Close|
|5||||rtitekBoostHeatingCountdown|Quick HEating Countdown||
|6||x|x|rtitekRunningState|Working status|Pause|Heating
|7||x|x|rtitekWindowState|Win state|Close|
|8|x|x|x|rtitekWindowDetection|Window check|ON OFF|
|12|x|x|x|rtitekChildLock|Child lock|ON OFF|
|13||x|x|rtitekBattery|Battery|100%|
|14||x|x|rtitekFaultAlarm|Fault Alarm||
|15|x|x|x|rtitekMinTemp|The lower limit of temperature|7.00?|
|16|x|x|x|rtitekMaxTemp|Set temperature ceiling|32.00?|
|17||x|x|rtitekScheduleMonday|Week program|AQgAANwPAABzEgAA3BYAAJY=|
|18||x|x|rtitekScheduleTuesday|Week program Tuesday|AgYAAMgMAACWEgAA3BYAAJY=|
|19||x|x|rtitekScheduleWednesday|Week program Wednesday|AwYAAMgMAACWEgAA3BYAAJY=|
|20||x|x|rtitekScheduleThursday|Week program Thu||
|21||x|x|rtitekScheduleFriday|Week program Fr||
|22||x|x|rtitekScheduleSaturday|Week program Sat||
[...]

PR #3238, opened 2021-10-27
Support for Haozee TS0601

Highvolt:
This PR adds support for another TRV from TUYA.
Now with the correct DP.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -821,15 +821,9 @@ module.exports = [
         ],
     },
     {
-        fingerprint: [{
-            // The model ID from: Device with modelID 'TS0601' is not supported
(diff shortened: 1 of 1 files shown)


Oddiwahn:
@Highvolt 
Yeah, thank you for just stealing my code and presenting it as your own here.
--> looks quite similar, eh? https://github.com/Oddiwahn/Rti-Tek-TRV601/blob/main/rti_tek.js

Also, just like last time, you committed again a non working versionand fucked up everything.
There is a reason why I did not create a pull request yet: Some stuff like temperature calibration is still not working.

https://github.com/Koenkk/zigbee-herdsman-converters/issues/3190

PR #3239, opened 2021-10-27
Add danfoss thermostat load balancing support

kennylevinsen:
Exposes the room-wide load balancing feature.

Depends on: https://github.com/Koenkk/zigbee-herdsman/pull/439

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3024,6 +3024,12 @@ const converters = {
                     result[postfixWithEndpointName('running_state', msg, model)] = 'idle';
                 }
             }
+            if (msg.data.hasOwnProperty('danfossLoadBalancingEnable')) {
+                result[postfixWithEndpointName('load_balancing_enable', msg, model)] = (msg.data['danfossLoadBalancingEnable'] === 1);
+            }
+            if (msg.data.hasOwnProperty('danfossLoadRoomMean')) {
+                result[postfixWithEndpointName('load_room_mean', msg, model)] = msg.data['danfossLoadRoomMean'];
+            }
             if (msg.data.hasOwnProperty('danfossLoadEstimate')) {
                 result[postfixWithEndpointName('load_estimate', msg, model)] = msg.data['danfossLoadEstimate'];
             }
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -2510,6 +2510,26 @@ const converters = {
             await entity.read('hvacThermostat', ['danfossWindowOpenExternal'], manufacturerOptions.danfoss);
         },
     },
+    danfoss_load_balancing_enable: {
+        key: ['load_balancing_enable'],
+        convertSet: async (entity, key, value, meta) => {
+            await entity.write('hvacThermostat', {'danfossLoadBalancingEnable': value}, manufacturerOptions.danfoss);
+            return {readAfterWriteTime: 200, state: {'load_balancing_enable': value}};
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('hvacThermostat', ['danfossLoadBalancingEnable'], manufacturerOptions.danfoss);
+        },
+    },
+    danfoss_load_room_mean: {
+        key: ['load_room_mean'],
+        convertSet: async (entity, key, value, meta) => {
+            await entity.write('hvacThermostat', {'danfossLoadRoomMean': value}, manufacturerOptions.danfoss);
+            return {readAfterWriteTime: 200, state: {'load_room_mean': value}};
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('hvacThermostat', ['danfossLoadRoomMean'], manufacturerOptions.danfoss);
+        },
+    },
     danfoss_load_estimate: {
         key: ['load_estimate'],
         convertGet: async (entity, key, meta) => {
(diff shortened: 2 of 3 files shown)


robertalexa:
@kennylevinsen Would you be kind to clarify how this operates? I have read the zigbee spec so I understand the values, but I fail to understand the logic. It feel almost like sth it missing.

The devices are not grouped by room (not to their knowledge at least), so how do they know who are they sharing the load with? I understand that the gateway is meant to calculate and send that value, but i still don't get it.

You have also not provided `withValueMin` and `withValueMax` for `load_room_mean` nor have you specified the values in the description, that might be something to improve.

I would appreciate your input on this :) Thanks

PS: there are also some folks here that are wondering the same thing, but I have directed them towards this PR.
https://community.home-assistant.io/t/danfoss-ally-trv-working-with-remote-temp-sensor/276686/46?u=horussjr

kennylevinsen:
The load balancing mechanism relies on the hub to calculate and distribute mean loads as appropriate.

For a given room, take the load estimates from all thermostats, calculate the mean value, and write it back to all the thermostats in the room. I'd imagine one should do this every 30 minutes or so, based on this being the recommendation for the external temperature feedback.

The idea would be that all thermostats report the how much they *think* they should be working individually, but then so base its action on how hard all the other radiators are working, giving a helping hand if needed. Don't have much experience with it in practice though...

Char-r:
@kennylevinsen When trying to push a load value to the Ally I get the following error:

Invalid value for number.load_room_mean: 1012 (range -8000.0 - 100.0)

robertalexa:
@Char-r This is something that has been worked on earlier this month
https://github.com/Koenkk/zigbee-herdsman-converters/pull/3461

Could you please confirm what is the highest value you have seen for load estimate (as that sort of value might be the max for load room mean). Based on that we can try to make an educated guess and change the limit. Currently, after the fix above, the min value is -8000 (not set), but the max value is 100. If you could provide some data I would be happy to help. I will enable the attributes at my end and see what values i get tomorrow, i currently have them disabled.

PS: the limit was imposed as a result of multiple issues and new changes to HA. This was an attempt to correct all entities across all devices.

Char-r:
> @Char-r This is something that has been worked on earlier this month #3461
> 
> Could you please confirm what is the highest value you have seen for load estimate (as that sort of value might be the max for load room mean). Based on that we can try to make an educated guess and change the limit. Currently, after the fix above, the min value is -8000 (not set), but the max value is 100. If you could provide some data I would be happy to help. I will enable the attributes at my end and see what values i get tomorrow, i currently have them disabled.
> 
> PS: the limit was imposed as a result of multiple issues and new changes to HA. This was an attempt to correct all entities across all devices.

Max value I saw was 1167 but honestly no idea how high it could go.

robertalexa:
I will keep an eye on my trvs tomorrow as they are being used. Once i think i have a good idea i will push a PR and get this resolved for the release that is due in the new year.

As to adding new attributes, the process would be to first do a PR on zigbee-herdsman where you defined the attributes. File name is clusters. Search for danfoss for inspiration. Make sure to define the attribute type as per danfoss specs document.

Then in zigbee-herdsman-converters you define the operations. This would be in fromZigbee and toZigbee files. Again, make sure you respect the standard from the specs document in terms of limits, reporting etc. And you would put those attributes to good use in the danfoss device file. If you do make changes make sure you replicate them on hive trv and popp trv as they are the same device.

You will need to test this in an environment where you manually build from your custom forks or just edit the files manually while replicating and commiting on your forks.

Feel free to tag me when you submit a pr and i would be happy to assist/review.

Ps, if you have something very specific in mind i might be able to find some time and get it added in for you

kennylevinsen:
> @kennylevinsen When trying to push a load value to the Ally I get the following error:
> 
> Invalid value for number.load_room_mean: 1012 (range -8000.0 - 100.0) since I am getting load value in the 1000s I don't think it is correct to set a max vaue here. On the other hand I don't know what the highest number set by danfoss could be.

100 is way too low a limit. Normal values for me during operation are up to 650 for increasing temperature in winter and around 100-ish for maintaining temperature.

I'm setting things straight on mqtt though, not through home assistant.

Char-r:
> As to adding new attributes, the process would be to first do a PR on zigbee-herdsman where you defined the attributes. File name is clusters. Search for danfoss for inspiration. Make sure to define the attribute type as per danfoss specs document.

Thanks, found the file, that's the bit I was missing. Though the two features I am interested in are already listed so will take a look at the converter.

> You will need to test this in an environment where you manually build from your custom forks or just edit the files manually while replicating and commiting on your forks.

last time I tried to load the Danfoss.JS converter as an external converter Z2M was not loading, let me give it another shot after I make the changes.

> Ps, if you have something very specific in mind i might be able to find some time and get it added in for you

thanks.Let me try to add them to the converter first, if I don't do too much of a hash hopefully it will save everyone some time.

robertalexa:
After checking my 8 TRVs this morning the highest values I found was 1648. This was from cold start, e.g. heating was off all night and then set to 22 in the morning. All TRVs worked as hard as possible to get in temp.

As a result, I have pushed a PR setting the load mean limit to 2000. This is to avoid a future change. Should this prove to still be too low (though i doubt it), we can amend it in future.

> I'm setting things straight on mqtt though, not through home assistant.

Yes, this was an issue that is stemming from HA, and in fact not even from that, it is from the MQTT number limits. So the above mentioned fix was an attempt to fix ALL affected attributes across ALL devices, some of them without knowledge, just common sense. We knew we might get it wrong, but we had to attempt to fix it.


> last time I tried to load the Danfoss.JS converter as an external converter Z2M was not loading, let me give it another shot after I make the changes.

Name the converter something else. Like `ally.js`. In your database file (z2m folder) change the zigbeeModel of one of your paired trvs from the official one to a custom one. e.g. add an extra character at the end. In your custom converter use the new model, this way there will be no overlapping between official device (from danfoss.js) and your converter (ally.js).

Hope this helps.

PR #3240, opened 2021-10-27
Danfoss thermostat is sleepy so use sendWhenActive

kennylevinsen:
As a Sleepy End Device, we should use sendWhenActive for all requests to the device.

However, rather than manually setting this flag whenever someone discovers issues with sleeping devices like done here, maybe we should enable sendWhenActive by default for sleepy end-devices, possibly using "battery-powered end-device" as heuristic. Open to suggestions.

Note: Built upon https://github.com/Koenkk/zigbee-herdsman-converters/pull/3239

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3024,6 +3024,12 @@ const converters = {
                     result[postfixWithEndpointName('running_state', msg, model)] = 'idle';
                 }
             }
+            if (msg.data.hasOwnProperty('danfossLoadBalancingEnable')) {
+                result[postfixWithEndpointName('load_balancing_enable', msg, model)] = (msg.data['danfossLoadBalancingEnable'] === 1);
+            }
+            if (msg.data.hasOwnProperty('danfossLoadRoomMean')) {
+                result[postfixWithEndpointName('load_room_mean', msg, model)] = msg.data['danfossLoadRoomMean'];
+            }
             if (msg.data.hasOwnProperty('danfossLoadEstimate')) {
                 result[postfixWithEndpointName('load_estimate', msg, model)] = msg.data['danfossLoadEstimate'];
             }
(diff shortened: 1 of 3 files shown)


kennylevinsen:
I have [a branch on zigbee-herdsman](https://github.com/kennylevinsen/zigbee-herdsman/tree/pollcontrol) where I respond to checkin commands and use them to control sendWhenActive behavior when present. It requests fast poll only if there are pending messages, marks the device as "active" for direct sending while we're emptying the queue, and stops fast poll when the queue has been emptied.

I unfortunately didn't seem to be getting any checkin commands on my network during my testing. I'll have to sniff some traffic, but I wonder if the commands will only be sent whenever the device has been sleeping for longer than LONG_POLL_INTERVAL. If that's the case (conclusion pending some sniffing), we may still need more than just checkin. I suspect the proper way to handle even zigbee 3.0 sleeping devices may then be to immediately send, and then only queue and rely on checkin when the request failed to be delivered.

This discussion is more about the general Zigbee stack handling than cluster converters at this point (apart from possibly enabling/disabling this behavior from here). @Koenkk do you want me to create an issue on zigbee-herdsman for this discussion?

MattWestb:
You can reading the attribute `checkin_interval (id: 0x0000)` on the SED (its 1/4 seconds) and if its being sett up OK with its parent is doing the checking request after the time have running out after it was active (being in sleep and radio is off but excluding pull request from its parent).

If not configuring IKEA controllers they is using 13200 = 55 minutes as default in the firmware.

For testing you can setting the checkin interval shorter so its easier doing the test but its then eating your batteries.

Then you is sniffing paring of one Zigbee 3 SED is the `End Device Timeout Request` then its telling its parent how long its can being sleeping without being deleted in the parents child table and need doing one rejoin if is extending this time, and the long pull interval must being shorter then it or you is getting rejoining all the time its trying sending somthing to its parent

Koenkk:
@kennylevinsen yes, sounds good please create an issue about this so we can properly track this.

kennylevinsen:
https://github.com/Koenkk/zigbee-herdsman/issues/445 opened with a summary of the information I have gathered so far.

kennylevinsen:
This will no longer be necessary once https://github.com/Koenkk/zigbee-herdsman/pull/453 is merged.

PR #3247, opened 2021-10-29
More Zonnsmart TRV01 refinements

geoffreylagaisse:


Koenkk:
I see you marked it as draft, is it ready for merge?

geoffreylagaisse:
@Koenkk almost, I'm still searching a way to poll the is_online automatically. The device only sends local temperature updates by triggering this Boolean. Maybe you know a way?

Koenkk:
What do you mean with poll `is_online`? We could also poll the local temperature e.g. every hour by using an onEvent handler: https://github.com/Koenkk/zigbee-herdsman-converters/blob/6f422d10d1d10e41cdc5dae22f584c5a4a9d2345/devices/tuya.js#L971

insipiens:
> @Koenkk almost, I'm still searching a way to poll the is_online automatically. The device only sends local temperature updates by triggering this Boolean. Maybe you know a way?

Geoffrey, Perhaps the behaviour is intentional, to preserve battery power as wifi consumes so much power.

You can get it to update local temp with set calibration. If you are going to add this should it be an optional flag as its not replicating original device behaviour?

PR #3251, opened 2021-10-30
Set color temp range for ledvance and osram

jpbede:
Set color temp. range for Ledvance 4058075208414 and Osram AC10787


Koenkk:
Thanks!

PR #3255, opened 2021-10-31
new white-label tuya curtain motor: Zemismart BCM500DS-TYZ-B

relaystar:
Zemismart BCM500DS-TYZ-B has a new manufacturerName but otherwise seems to work OK a white-label of all the other TS0601 curtain motors.


Koenkk:
Thanks!

PR #3267, opened 2021-11-02
Fixed PeCLS01 battery display (50% shown before)

schirin:
meta: {battery: {dontDividePercentage: true}} added in PECLS01 converter


Koenkk:
Thanks!

PR #3265, opened 2021-11-02
Sync the code on Danfoss Ally, Popp and Hive TRV

robertalexa:
Brought these 3 devices in line.

Included:
Setpoint_change_source
https://github.com/Koenkk/zigbee-herdsman-converters/pull/3176

Load Balancing
https://github.com/Koenkk/zigbee-herdsman-converters/pull/3239

Better reporting on Danfoss and Popp thinking about device battery life

Comment to remind people to handle all devices when doing PRs to one

No new implementations on my side, just house keeping.

PR #3268, opened 2021-11-02
danfoss: Use normal reporting intervals

kennylevinsen:
The Danfoss Ally thermostat had some odd reporting intervals configured. Use the ones from the reporting library. If we want to change thermostat reporting intervals we should do it in the reporting library instead.

### DIFF
--- a/devices/bitron.js
+++ b/devices/bitron.js
@@ -1,7 +1,6 @@
 const exposes = require('../lib/exposes');
 const fz = {...require('../converters/fromZigbee'), legacy: require('../lib/legacy').fromZigbee};
 const tz = require('../converters/toZigbee');
-const constants = require('../lib/constants');
 const reporting = require('../lib/reporting');
 const extend = require('../lib/extend');
 const e = exposes.presets;
--- a/devices/bitron.js
+++ b/devices/bitron.js
@@ -199,7 +199,7 @@ module.exports = [
                 'genBasic', 'genPowerCfg', 'genIdentify', 'genPollCtrl', 'hvacThermostat', 'hvacUserInterfaceCfg',
             ];
             await reporting.bind(endpoint, coordinatorEndpoint, binds);
-            await reporting.thermostatTemperature(endpoint, {min: 900, max: constants.repInterval.HOUR, change: 1});
+            await reporting.thermostatTemperature(endpoint);
             await reporting.thermostatTemperatureCalibration(endpoint);
             await reporting.thermostatOccupiedHeatingSetpoint(endpoint);
             await reporting.thermostatRunningState(endpoint);
--- a/devices/candeo.js
+++ b/devices/candeo.js
@@ -13,7 +13,7 @@ module.exports = [
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl']);
             await reporting.onOff(endpoint);
-            await reporting.brightness(endpoint, {min: 1});
+            await reporting.brightness(endpoint);
         },
     },
 ];
--- a/devices/danfoss.js
+++ b/devices/danfoss.js
@@ -73,15 +73,15 @@ module.exports = [
 
             // standard ZCL attributes
             await reporting.batteryPercentageRemaining(endpoint);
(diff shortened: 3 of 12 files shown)


robertalexa:
There was a reason for me setting the change value to 5 instead of 10. And to my knowledge, not all thermostats report their temperature with 2 decimals, some of them do it with just 1 decimal.

In the case of Danfoss Ally/Hive Trv/Popp, that would mean 2195(21.95C). Meaning a new report will be created currently on 21.92 and 22.00. Changing it to 10 would mean the reporting would happen at 21.85 and 22.05. Yes it is granular, but it makes measurement more accurate.

In all reality, last night I wanted to submit a PR for this too, but then stopped :)

Would you please be kind to submit the exact changes to Hive TRV and Popp TRV as they are exactly the same device? That would keep things clean.

Thanks

robertalexa:
PS: I have submitted a PR already for your 2 most recent changes to sync these 3 devices.

https://github.com/Koenkk/zigbee-herdsman-converters/pull/3265

Might be worth you pulling before anything :)

kennylevinsen:
> There was a reason for me setting the change value to 5 instead of 10. And to my knowledge, not all thermostats report their temperature with 2 decimals, some of them do it with just 1 decimal.

It doesn't in my opinion make sense to have different default reporting setting for Danfoss thermostats than say, Schneider, when they have the same capabilities. Keeping them to the library defaults and using same values for special attributes leave things nice and unopinionated - avoids having disagreements between users. :)

I don't see any harm in suggesting changes to the default reporting intervals in the reporting lib, but I don't see any reason for making Danfoss thermostats special unless we need to work around a quirk.

(We might not find the defaults perfect - I'd want MAX instead of HOUR - but they're just defaults, and special use cases like super granular or often reporting can be done by changing the reporting configuration at runtime.)

> Would you please be kind to submit the exact changes to Hive TRV and Popp TRV as they are exactly the same device? That would keep things clean.

Sure, completely forgot those this time around.

> Might be worth you pulling before anything :)

I'll check if there's something I missed. My original PR was bigger as you got in ahead of me for some of this. :)

robertalexa:
Agree with everything you have said :) I was not against you changing it, just trying to provide background of why it is the way it is.

As to MAX vs HOUR, if i understand the reporting mechanism correctly, I would also vote for MAX, with the conditions that "change" is set up correctly. So you actually get reporting as often as needed, but veeery rare if nothing changes.

Let me know if you need my help with anything, I own the Hive TRV so i can help with testing it should you ever have the need. :)

kennylevinsen:
Okay, the 3 identical TRV's have been synced up.

I considered doing a sweep over other thermostats in the same go (out of the 148 remaining custom reporting options, 81 are in the thermostat* group), but that's a bit more intrusive as I (obviously) don't own all those devices. @Koenkk, thoughts?

Koenkk:
@kennylevinsen I'm fine with a cleanup for all thermostats.

kennylevinsen:
WIth this I have reduced 187 custom reporting intervals to 108. Took thermostats, some temperature sensors and a few cases where only battery level was configured.

What remains seem to be predominantly power measurement devices, which I have none of so probably better to leave alone for now.

Koenkk:
> What remains seem to be predominantly power measurement devices, which I have none of so probably better to leave alone for now.

Changing this for power measurements is tricky, reason is that the value of the change depends on the power multiplier/divisor. E.g. `change: 5` means something different for a plug with different divisors/multipliers.

kangaroomadman:
How come the Candeo light dimmer switch was changed?

The reason that this was configured to `{ min: 1 }` was because otherwise it sends a lot of brightness reporting messages, which I have observed can arrive at the endpoint out of order. Therefore setting to 1s reporting interval meant the traffic was severely reduced and no out of order messages occurred.

kennylevinsen:
> How come the Candeo light dimmer switch was changed?

The majority of the custom reporting intervals have no explanation, seem mostly random or copy-pasted from other definitions, and are unnecessary considering we have good (heck, often *better*) defaults. This PR adjusted the 79 most trivial ones.

If one of the custom reporting intervals removed in this PR are necessary for device quirks, please do not hesitate to open a PR re-introducing it, but please add a comment describing why the default is suitable to help future contributors. :)

PR #755, opened 2019-11-17
Modify thermostats report time, add TH1124ZB model

ioio85:
* Modify report time: on my thermostat (TH1123ZB) if the `reportableChange` is 0 there is nothing reported. I hope this change will be ok with other type of thermostats.
* added TH1124ZB

PR #3269, opened 2021-11-02
reporting: Increase interval for pIHeatingDemand

kennylevinsen:
Reporting at least every 10 minutes is excessive for battery-powered
thermostats, and should never be necessary. Turn it back to at least
every hour like it used to be.

Note: The interval was previously lowered in https://github.com/Koenkk/zigbee-herdsman-converters/pull/755.


Koenkk:
FYI: @ioio85  

@kennylevinsen thanks!

PR #3272, opened 2021-11-03
support Evolveo Heat M30

danoh:
Hello, I have TRV from company Evolveo named Heat M30, in fact it seems be just TuYa TS0601_thermostat with another manufacturer name on the box. I tried this patch and it works like any other TuYa TRV.


Koenkk:
thanks!

PR #3277, opened 2021-11-03
IKEA STARKVIND should honer FanModeSequence

sjorge:
I noticed that FanModeSequence returns 0x02 for the STARKVIND, according to ZCL:
0x02 Low/Med/High/Auto


sjorge:
OK this might need a follow up, a side effect of the expose is that is also exposes fan_state which is ON or OFF... but that of course doesn't work because 'on' is not a supported fan mode for this device.

sjorge:
Setting `fan_state=ON` now matches what happens when you physically push the hardware button (which defaults to 'auto' as wel)

Koenkk:
Thanks!

PR #3281, opened 2021-11-04
Added IKEA E1812 double press command

io53:
The new update for E1812 makes it so that it sends "off" when you double press it.
I've only tested it on updated E1812.


Koenkk:
Thanks!

PR #3283, opened 2021-11-04
Sinopeadds

onixx:
adding running state to TH1400ZB


devices/sinope.js
```diff
@@ -237,6 +237,12 @@ module.exports = [
         vendor: 'Sinope',
         description: 'Zigbee smart dimmer',
         extend: extend.light_onoff_brightness({noConfigure: true}),
+        exposes: [e.light_brightness(), e.effect(), exposes.numeric('led_intensity_on', ea.SET).withValueMin(0).withValueMax(100)
```
Koenkk, line 240: Instead of adding all exposes here like `e.light_brightness()`, use `[...extend.light_onoff_brightness({noConfigure: true}).exposes, exposes.numeric('led_intensity_off', ea.SET).withValueMin(0).withValueMax(100).withDescription('Control status LED when load OFF')` instead

onixx:
Sorry, mistake here. I should have done a new branch. will try again

PR #3285, opened 2021-11-04
Fix WSDCGQ11LM pressure readings

Masterz69:
In short: Z2M converter using "measured" value for pressure, which is less precise, comparing with "scaled" value.

Here seen difference in graph between same model sensors (WSDCGQ11LM), 
- working through Xiaomi Gateway 3 -> Home Assistant (name: _Shed_) 
- and two of them working through Z2M -> same Home Assistant (_BedRoom_ & _Attic_).

image

As discovered by sniffing Zigbee traffic: packet contains two values, where _scaled_ one have more meaningful digits [(Telegram message):](https://t.me/zigbeer/235553)
image
Thanx Alex from Telegram.

By using another method/function to expose pressure data: ```pressure``` vs. ```WSDCGQ11LM_pressure``` we get more precise data and smooth graphs [(Telegram message):](https://t.me/zigbeer/235613)
image
Top line is data from BME sensor. You can see change for Xiaomi data for two another lines - with changed converter right from arrow added to screenshot.

Tested by Alex in docker environment.


Koenkk:
Thanks!

smartynov:
I can confirm this helps. Before (left half of the chart) and after (right half) applying this patch.

image

smartynov (commented):
Tried this and confirm it works

PR #3286, opened 2021-11-05
Added Sengled E1F-N9G

gleepwurp:


devices/sengled.js
```diff
@@ -216,4 +216,11 @@ module.exports = [
         description: 'Vintage LED edison bulb (ST19)',
         extend: extend.light_onoff_brightness(),
     },
+    {
+        zigbeeModel: ['E1F-N9G'],
+        model: 'E1F-N9G',
```
Koenkk, line 221: Does it not support dimming?

gleepwurp: Its supposed to, but giving off errors when I use the extend.light_onoff_brightness() function. Figured I would put it in with a working on/off function and work on the brightness thingy afterwards.

Koenkk: Can you provide the data/database.db of this device + a debug log when setting brightness fails?

gleepwurp: I'll try debugging over the weekend. Just though I would do this quick on/off to get running, and then debug the rest of it.

gleepwurp: I'll be, it worked this time around...  I'll go ahead an update the file appropriately. I've been having "weirdness" in my zigbee network lately, so it might have been just coincidence.

Koenkk:
Thanks!

PR #3288, opened 2021-11-05
Changed unit from mA to A for Hoch DIN

danieledwardgeorgehitchcock:
This is so that the device can be used in the energy statistics dashboard


Koenkk:
Thanks!

PR #3295, opened 2021-11-06
Add Philips Hue white ambiance E27 800lm

sjorge:
Based on info from https://github.com/Koenkk/zigbee2mqtt/issues/9527, this is the new 800lm version of the already supported 1100lm version.


Koenkk:
Thanks!

PR #3298, opened 2021-11-07
Add support for Tuya TRV HY367 TS0601 _TZE200_2atgpdho

michal-krasowski-red:


michal-krasowski-red:
Sorry, wrong account, will create a PR from personal account

PR #3301, opened 2021-11-07
Added thermostat manufacturerName

sc-ampersand-p:
The thermostat with the manufacturerName _TZE200_c88teujp shows up as supported by z2m and can be reached via HASS, but not all data points are exposed to HASS. Since it isn't listed here so far, I have added it.


MattWestb:
Sorry but i think this is one Saswell SEA802 TRV but cat saying 110% without picture of the device.

sc-ampersand-p:
Well damn, you're correct. Sorry about that. I'll close the PR.

I still have an issue of missing exposed attributes, but I'll address that in the right place.

Issue #3100, opened 2021-09-30
Porpose improvement to tuya_data_point_dump


Asleep-at-the-wheel:
Hello,
I find (currently) that following this procedure [here](https://www.zigbee2mqtt.io/how_tos/how_to_support_new_tuya_devices.html) causes Hassio containers of Z2MQTT to fail because the directory the converter is trying to use doesn't exist (also seen [here](https://github.com/Koenkk/zigbee2mqtt/issues/6292)).

https://github.com/Koenkk/zigbee-herdsman-converters/blob/d7b9aa4d039734bc840b1efa88d6594ea87fc7e5/converters/fromZigbee.js#L3728-L3759

The problem is that `fs.appendFile('data/tuya.dump.txt', dataStr, (err)` breaks the container the second a message comes in. It does work if you change it to `fs.appendFile('/data/tuya.dump.txt', dataStr, (err)` and then `docker exec` into the container to find and dig out the logs. 

Doing this how ever is serious amount of required knowledge for someone who might just be looking to provide some debug data for help or figure out how to add their own device while using Hassio.

So how about a change to the dump point from `data/` to the same directory that Zigbee2MQTT keeps its logs? That way the procedure will work for both Hassio and regular installations? Hassio users can easily copy paste the external converter and copy paste what ever they find to their request for new device post if they can't figure it out themselves.

Or maybe the other solution is to mount the /data/tuya.dump.txt file into some where accessible?

Koenkk:
> So how about a change to the dump point from data/ to the same directory that Zigbee2MQTT keeps its logs? That way the procedure will work for both Hassio and regular installations?

I'm fine with that, could you make a PR?

PR #3303, opened 2021-11-07
fix tuya.dum.txt file location

ciB89:
prior the location was 'data/tuya.dump.txt' which caused the addon to crash on Home Assistant (https://github.com/Koenkk/zigbee2mqtt/issues/6292). Adding a slash '/data/tuya.dump.txt' fixes the issue. In addition, changing the path to '/share/zigbee2mqtt/tuya.dump.txt' should make the file accessible from the host via samba.

proposed changes https://github.com/Koenkk/zigbee-herdsman-converters/issues/3100

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4002,7 +4002,7 @@ const converters = {
             });
             dataStr += '\n';
             const fs = require('fs');
-            fs.appendFile('data/tuya.dump.txt', dataStr, (err) => {
+            fs.appendFile('/share/zigbee2mqtt/tuya.dump.txt', dataStr, (err) => {
                 if (err) throw err;
             });
         },


converters/fromZigbee.js
```diff
@@ -4002,7 +4002,7 @@ const converters = {
             });
             dataStr += '\n';
             const fs = require('fs');
-            fs.appendFile('data/tuya.dump.txt', dataStr, (err) => {
+            fs.appendFile('/share/zigbee2mqtt/tuya.dump.txt', dataStr, (err) => {
```
Koenkk, line 4005: This may work for hassio, but will break other setups which don't have `/share`. With what exception does saving to `data/` fail?

ciB89: Zigbee2MQTT:debug 2021-11-10 11:18:29: Received Zigbee message from 'Büro', type 'commandGetData', cluster 'manuSpecificTuya', data '{"data":{"data":[0],"type":"Buffer"},"datatype":1,"dp":1,"fn":0,"status":0,"transid":116}' from endpoint 1 with groupID 0
Error: ENOENT: no such file or directory, open 'data/tuya.dump.txt'

As pointed out, at least adding a '/' in front of 'data/tuya.dump.txt' would fix this for Hassio.

Koenkk: What is logged when you add `console.log(__dirname);` just above `fs.appendFile`?

ciB89: nothing is logged; interestingly the error is also not logged in the log files, but shows only  when directly viewing the log from the addon

nurikk: What if use ./data/tuya.dump.txt?

flatsiedatsie:
I'm also running into this issue. I'm trying to add support for a new Tuya Device, but am stuck on this step.

`ENOENT: no such file or directory, open 'data/tuya.dump.txt'`

What's more, after this error Zigbee2MQTT doesn't seem to parse messages from other devices anymore, so I can't leave this debugging option enabled.

I've also been going through the code of my local instance to find where this command originates, but initially failed. I couldn't find any reference to `dump.txt` in `/lib` or `/dist` or `/scripts`.

For anyone finding this: it's in `zigbee2mqtt/node_modules/zigbee-herdsman-converters/converters/fromZigbee.js`

I've added the `console.log(__dirname);`. It gave:

`z2m stdout: /home/pi/.webthings/data/zigbee2mqtt-adapter/zigbee2mqtt/node_modules/zigbee-herdsman-converters/converters`

I then changed the filename to `./tuya.dump.txt` and did a search for where the file ended up. In my case it ended up here:

`/home/pi/mozilla-iot/gateway/tuya.dump.txt`

Which is weird, since that directory is deprecated by the WebThings Gateway project since it spun out of Mozilla :-D


In my opinion it would be nice if the `data/tuya.dump.txt` directory was relative to the location of `configuration.yaml`. And if it was relative to `configuration.yaml`, then perhaps the dump could be placed in the `/log` directory, which using the debugging option already creates? It's where I originally thought I would find it.

If that directory path is not available, then maybe `~/tuya.dump.txt` might not be a bad alternative?

flatsiedatsie:
I tried getting it to work, but didn't get very far.

External converter:
```
const e = exposes.presets;
const ea = exposes.access;
const tuya = require("zigbee-herdsman-converters/lib/tuya");

const definition = {
        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_kzm5w4iz'}],
        model: 'TS0601',
        vendor: 'TuYa',
        description: 'Vibration sensor',
        supports: 'vibration',
        fromZigbee: [
            //fz.ignore_basic_report, // Add this if you are getting no converter for 'genBasic'
            fz.tuya_data_point_dump,
            fz.tuya_smart_vibration_sensor, // 1
            //fz.tuyaVibration, // 10 
            //fz.thitBatteryPercentage, // 3
        ],
        toZigbee: [],
        //onEvent: tuya.onEventSetTime, // Add this if you are getting no converter for 'commandMcuSyncTime'
            await reporting.bind(endpoint, coordinatorEndpoint, ['genBasic']);
        },
        exposes: [e.vibration()], 
    };

module.exports = definition;
```


I added `tuyaVibration: 10,` to `tuya.js`

I added this to `from_zigbee.js`:

```
    tuya_smart_vibration_sensor: {
        cluster: 'manuSpecificTuya',
        //type: ['commandStatusChangeNotification'],
        type: ['commandGetData'],
            console.log("In converter!");
            const dp = msg.data.dp;
            console.log("- In converter, dp = ", dp);
            const value = tuyaGetDataValue(msg.data.datatype, msg.data.data); // This function will take care of converting the data to proper JS type
            console.log("- In converter, value = ", value);
            //if (utils.hasAlreadyProcessedMessage(msg)) return;
            switch (dp) {
            case tuya.dataPoints.tuyaVibration:
                return {vibration: {0: 'false', 1: 'true'}[value]}; // value is already converted to a number in JS
            default:
                meta.logger.warn(`zigbee-herdsman-converters:tuya_smart_vibration_sensor: NOT RECOGNIZED DP #${dp} with data ${JSON.stringify(msg.data)}`); // This will cause zigbee2mqtt to print similar data to what is dumped in tuya.dump.txt
            }
        },
    },
```

converters/fromZigbee.js
TweakIT: Wouldn't it be possible to catch the error and try the next possible location? Maybe it is also nicer to console.log the error, because the current method crashes the complete container. I tried implementing it but my JS skills are not sufficient.

ciB89:
This issue still persists and should not be closed.

PR #3304, opened 2021-11-08
devices: RGB genie: add support for ZB-1026

kruton:
This is a simple LED light controller that dims (no color or anything).

https://www.amazon.com/dp/B07N9LP87G


Koenkk:
Thanks!

PR #3308, opened 2021-11-08
Expose DJT11LM vibration as binary sensor

jpbede:
Expose vibration DJT11LM as a binary sensor. This makes it easier to integrate it with e.g. [alarmo](https://github.com/nielsfaber/alarmo/).

As the sensor only sends an vibration message and not an no_vibration message, a no_vibration message is sent by ourself after a configured timeout.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -6198,12 +6198,34 @@ const converters = {
     DJT11LM_vibration: {
         cluster: 'closuresDoorLock',
         type: ['attributeReport', 'readResponse'],
+        options: [exposes.options.vibration_timeout()],
         convert: (model, msg, publish, options, meta) => {
             const result = {};
 
             if (msg.data['85']) {
                 const vibrationLookup = {1: 'vibration', 2: 'tilt', 3: 'drop'};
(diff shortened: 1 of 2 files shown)


converters/fromZigbee.js
```diff
@@ -6198,12 +6198,34 @@ const converters = {
...
+                    globalStore.getValue(msg.endpoint, 'vibration_timers', []).forEach((t) => clearTimeout(t));
+                    globalStore.putValue(msg.endpoint, 'vibration_timers', []);
+
+                    // Set new timer to publish no_vibration message
+                    if (timeout !== 0) {
+                        const timer = setTimeout(() => {
+                            publish({vibration: false});
+                        }, timeout * 1000);
+
+                        globalStore.getValue(msg.endpoint, 'vibration_timers').push(timer);
```
Koenkk, line 6226: Why is there an array needed, I expect that one endpoint can only have 1 `vibration_timer`.

jpbede: Yeah you're indeed right. I've copy&pasted it from an other device 🤦‍♂️

Koenkk:
Thanks!

PR #3309, opened 2021-11-08
Flipped system mode array for TS0601_thermostat_1

Highvolt:
The map for the presets on this thermostat is flipped not allowing for the system modes to be changed.


Koenkk:
Thanks!

PR #3311, opened 2021-11-08
Restrict Candeo HK-DIM-A reporting interval to 1 second

kangaroomadman:
The original addition of this device had a minimum reporting interval of 1 second but no explaining comment so was cleaned up as part of #3268 

This has now been re-added with a suitable (hopefully?) comment.


kennylevinsen:
LGTM, and apologies for the accidental removal. :)

Koenkk:
Thnaks!

PR #3316, opened 2021-11-10
Updates to haozee_thermostat for TS0601_thermostat_1

wprinz:
added return values for haozeeBoostHeating and haozeeTempCalibration fromZigbee.js
added factor 10 for haozeeTempCalibration  in toZigbee.js


Koenkk:
Thanks!

PR #3318, opened 2021-11-10
Add Hive SLR1c

calumr:
It exactly the same as the SLR1b & the SLR1. Maybe we should just have 1 device with 3 entries in the zigbeeModel array?


Koenkk:
This is fine, thanks!

PR #3323, opened 2021-11-11
Add cool mode for Zen thermostat

gunnarsundberg:
Add `cool` to exposed system modes for zen thermostat in `devices/zen.js`


Koenkk:
Thanks!

PR #3328, opened 2021-11-11
Add support for V3 of the Hive dual channel receiver

robertalexa:
I have just noticed that Hive release a V3 of their thermostats (thanks to this recent PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/3318)

I added support for the Dual Channel SLR2c (the PR above only did Single Channel SLR1c)

Thanks Koen

LE: this is under the assumption that the devices behave identical as the previous generations, which they did so far. The only thing Hive changes is the "looks". Should there be any issues I will gladly help people in need

### DIFF
--- a/devices/hive.js
+++ b/devices/hive.js
@@ -466,6 +466,57 @@ module.exports = [
                     ' used. 0 to 360 to match the remote display').withEndpoint('water')],
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #3337, opened 2021-11-12
Develco energy meter

robertalexa:
This PR is finally adding support for the attributes added in: https://github.com/Koenkk/zigbee-herdsman/pull/413 and https://github.com/Koenkk/zigbee-herdsman/pull/419

The thread were the support has been provided is: https://github.com/Koenkk/zigbee2mqtt/issues/8638

I do not own the device but I've had people test the code and confirming everything behaves as expected.

Thanks

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -648,6 +648,26 @@ const converters = {
             return payload;
         },
     },
+    develco_metering: {
+        cluster: 'seMetering',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const result = {};
+            if (msg.data.hasOwnProperty('develcoPulseConfiguration')) {
+                result[postfixWithEndpointName('pulse_configuration', msg, model)] =
+                    msg.data['develcoPulseConfiguration'];
+            }
+
+            if (msg.data.hasOwnProperty('develcoInterfaceMode')) {
+                result[postfixWithEndpointName('interface_mode', msg, model)] =
+                    constants.develcoInterfaceMode.hasOwnProperty(msg.data['develcoInterfaceMode']) ?
+                        constants.develcoInterfaceMode[msg.data['develcoInterfaceMode']] :
+                        msg.data['develcoInterfaceMode'];
+            }
+
+            return result;
+        },
+    },
     electrical_measurement: {
         /**
          * When using this converter also add the following to the configure method of the device:
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -2596,7 +2596,7 @@ const converters = {
         convertSet: async (entity, key, value, meta) => {
             await entity.write('seMetering', {'develcoCurrentSummation': value}, manufacturerOptions.develco);
             return {state: {'current_summation': value}};
-        }
+        },
     },
     ZMCSW032D_cover_position: {
         key: ['position', 'tilt'],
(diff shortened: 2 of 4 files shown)


Koenkk:
Thanks again! 🎖️

RipperFox:
> interface mode as constants

Do I assume correctly that only the mode "electricity" with impulse counting is currently implemented?
As far I can see the the mode "IEC62056-21" would allow more data points (e.g. power per phase) with my MT681 meter as listed here:
https://wiki.volkszaehler.org/hardware/channels/meters/power/edl-ehz/iskraemeco_mt681

Problem is that "IEC62056-21" communicates via „Smart message language“, which would be another layer to parse. 

Maybe I'm totally in the wrong here - I'm a total newbie to this field and installed HomeAssistant, Zigbee2MQTT, a ConBee 2 and that Develco meter yesterday..

skrobul:
@RipperFox even when the 'electricity' mode is set, the extra information is being sent by the meter:

```
Zigbee2MQTT:debug 2021-11-13 16:14:26: Received Zigbee message from 'EnergyMonitor', 
type 'attributeReport', cluster 'seMetering', data: '{
  "currentSummDelivered": [
    0,
    2407
  ],
  "currentTier1SummDelivered": [
    65535,
    4294967295
  ],
  "currentTier2SummDelivered": [
    65535,
    4294967295
  ],
  "instantaneousDemand": 328,
  "status": 0
}' from endpoint 2 with groupID 0
```

Based on the [manual](https://github.com/Koenkk/zigbee-herdsman-converters/files/7612260/Technical.Manual.for.ZHEMI101.-.ZigBee.External.Meter.Interface.HA.pdf) I assume the `currentSummDelivered`, `currentTier1SummDelivered` and `currentTier2SummDelivered` would show the correct extra values already without a need to parse anything else. Those attributes are not exposed yet, but should be trivial to add - I just don't understand what 65535 and 4294967295 represent. Most likely they need to be interpreted using equations described in chapter 3.3.3.4.

Having said that, the `power`, `energy` and other attributes that are the same should already work as they do in the `electricity` mode.

RipperFox:
@skrobul  But in 'electricity' mode the Develco interface uses the simple impulse LED on the meter - which in my case just blinks 10,000 times per kWh.
To use the 'IEC62056-21' mode, I'm quite sure the optical serial interface would have to be used and that data would be SML formatted.
The manual you linked seems different from the [current one at the manufacturers site](https://www.develcoproducts.com/media/1782/zhemi101-technical-manual-external-meter-interface.pdf) as it mentions another InterFaceMode:
>  '0x0104 P1 dutch standard – DSMR 4.0 version'.

skrobul:
> That extra information would be sent by the interface, not the power meter, right?

Well yeah, afaik it's kinda impossible to derive that information from the blinking LED. 
Did you try switching the Develco to 'IEC62056-21' mode? Has the output of the received messages changed? If so, can you please paste the [debug](https://www.zigbee2mqtt.io/guide/usage/debug.html#enabling-logging) messages along with the values shown on the meter at the time? I guess that would be the first step to help someone implement support for it. It's also probably worth opening a new GH issue for that.

robertalexa:
I would suggest for the sake if "tidiness" to create a new issue thread specifically for this conversation, otherwise this PR will be bombarded with information unrelated to it.

As mentioned above, i do not own the device, just tried to help some people in the community with it. I do not mind trying to do the same for your situation @RipperFox but I would need you to test whatever we attempt. And that includes you connecting the meters correctly and whatever other implications.

PR #3338, opened 2021-11-12
Add pm25/air quality support to IKEA STARKVIND

sjorge:
image

Reporting for this cluster doesn't seem to work, it returns INVALID_DATA.

From what I gather from the various discussions over at the z2m repo is that the app will poll the value when viewing the device. The value also only updates when the fan is running, if the fan is off after a while the value will return 65535 indicating the current value is not known.

Edit: I assume we should update the notes for the device to indicate that the value needs to be manual refreshed and it's quirks?

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4245,6 +4245,32 @@ const converters = {
             }
         },
     },
+    ikea_pm25: {
+        cluster: 'manuSpecificIkeaPM25Measurement',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            if (msg.data['measuredValue']) {
+                const pm25 = parseFloat(msg.data['measuredValue']) / 100.0;
+                const pm25Property = postfixWithEndpointName('pm25', msg, model);
+
+                // Air Quality Scale (ikea app):
+                // 0-35=Good, 35-80=OK, 80+=Not Good
+                let airQuality;
+                const airQualityProperty = postfixWithEndpointName('air_quality', msg, model);
+                if (pm25 <= 35) {
+                    airQuality = 'good';
+                } else if (pm25 <= 80) {
+                    airQuality = 'ok';
+                } else if (pm25 < 65535) {
+                    airQuality = 'not_good';
+                } else {
+                    airQuality = 'unknown';
+                }
+
+                return {[pm25Property]: calibrateAndPrecisionRoundOptions(pm25, options, 'pm25'), [airQualityProperty]: airQuality};
+            }
+        },
+    },
     E1524_E1810_levelctrl: {
         cluster: 'genLevelCtrl',
         type: [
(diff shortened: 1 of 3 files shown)


converters/fromZigbee.js
```diff
@@ -4245,6 +4245,32 @@ const converters = {
             }
         },
     },
+    ikea_pm25: {
+        cluster: 'manuSpecificIkeaPM25Measurement',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
```
Koenkk, line 4251: Please add options like: https://github.com/Koenkk/zigbee-herdsman-converters/blob/04fc9d85c99f2488440d3b8d50e08ed0fc22a26c/converters/fromZigbee.js#L337

devices/ikea.js
```diff
@@ -625,10 +625,16 @@ module.exports = [
...
+            exposes.enum('air_quality', ea.STATE_GET, [
+                'good', 'ok', 'not_good', 'unknown',
+            ]).withDescription('Measured air quality'),
+        ],
         meta: {fanStateOn: 'auto'},
-        fromZigbee: [fz.fan],
-        toZigbee: [tz.fan_mode],
+        fromZigbee: [fz.fan, fz.ikea_pm25],
+        toZigbee: [tz.fan_mode, tz.ikea_pm25],
         configure: async (device, coordinatorEndpoint, logger) => {
```
Koenkk, line 638: Since the device does not report these values on its own, I think it is good to poll them (but not too often, maybe every 30 mins?). An example how to setup polling: https://github.com/Koenkk/zigbee-herdsman-converters/blob/04fc9d85c99f2488440d3b8d50e08ed0fc22a26c/devices/tuya.js#L1032 (make sure to also add an option for it https://github.com/Koenkk/zigbee-herdsman-converters/blob/04fc9d85c99f2488440d3b8d50e08ed0fc22a26c/devices/tuya.js#L1028 ). This option will be automagically generated in the docs so no need to further explain that this value is polled.

sjorge: I'm leaning against not doing that, the value will be unknown when the fan is not running anyway.

Koenkk: I see, that is fine

sjorge:
@Koenkk question about the the precision option, can I set the default to 2 for the device? Or is that always 0 ? I'm OK with either but the device supports up to 2

Koenkk:
If the default is 2, please set the default to 2.

sjorge:
> If the default is 2, please set the default to 2.

Yes, but how? It's not obvious to me where to set the default?

Koenkk:
Does it work after adding `pm25: 2` to https://github.com/Koenkk/zigbee-herdsman-converters/blob/1ad7e282849d804d2d3840c26ff7764e9be0c324/lib/utils.js#L75 ?

sjorge:
That seems to work, that will effect other pm25 sensors like the heiman ones too right?
Although those seem to get divided by 100.0, the ikea definitely needs that or otherwise I'd be living in a dusty saw mill.

PR #3344, opened 2021-11-12
Revert "Add pm25/air quality support to IKEA STARKVIND E2007 (#3338)"

sjorge:
This reverts commit 1ad7e282849d804d2d3840c26ff7764e9be0c324.

As per, https://github.com/Koenkk/zigbee2mqtt/discussions/8744#discussioncomment-1634437, Very likely this is reading junk that just happens to be within the range I was expecting.

Let's revert this for now while I try to map more of the attribute sniffed by hakaneriksson

Sorry about that, I guess I got a little to excited I was getting data seemed to match what I was expecting.

PR #3345, opened 2021-11-13
Add all features to IKEA STARKVIND

sjorge:
Merge after:
- https://github.com/Koenkk/zigbee-herdsman/pull/455

Super seeds:
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/3344

### DIFF
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -628,7 +628,11 @@ module.exports = [
         exposes: [e.fan().withModes(['off', 'low', 'medium', 'high', 'auto'])],
         meta: {fanStateOn: 'auto'},
         fromZigbee: [fz.fan, fz.ikea_air_purifier],
-        toZigbee: [tz.fan_mode],
+        toZigbee: [
+            tz.fan_mode, tz.ikea_air_purifier_fan_mode, tz.ikea_air_purifier_fan_speed,
+            tz.ikea_air_purifier_pm25, tz.ikea_air_purifier_child_lock, tz.ikea_air_purifier_led_enable,
+            tz.ikea_air_purifier_replace_filter,
+        ],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['hvacFanCtrl']);
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -4,6 +4,7 @@ const tz = require('../converters/toZigbee');
 const ota = require('../lib/ota');
 const constants = require('../lib/constants');
 const reporting = require('../lib/reporting');
+const {repInterval} = require('../lib/constants');
 const extend = require('../lib/extend');
 const e = exposes.presets;
 const ea = exposes.access;
@@ -650,8 +651,24 @@ module.exports = [
         ],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
-            await reporting.bind(endpoint, coordinatorEndpoint, ['hvacFanCtrl']);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['hvacFanCtrl', 'manuSpecificIkeaAirPurifier']);
             await reporting.fanMode(endpoint);
+
+            const options = {manufacturerCode: 0x117c};
+            await endpoint.configureReporting('manuSpecificIkeaAirPurifier', [{attribute: 'particulateMatter25Measurement',
+                minimumReportInterval: repInterval.MINUTE, maximumReportInterval: repInterval.HOUR, reportableChange: 1}],
+                options);
+            await endpoint.configureReporting('manuSpecificIkeaAirPurifier', [{attribute: 'filterOperationTime',
+                minimumReportInterval: repInterval.HOUR, maximumReportInterval: repInterval.MAX, reportableChange: 0}],
+                options);
+            await endpoint.configureReporting('manuSpecificIkeaAirPurifier', [{attribute: 'fanMode',
+                minimumReportInterval: 0, maximumReportInterval: repInterval.HOUR, reportableChange: 0}],
+                options);
+            await endpoint.configureReporting('manuSpecificIkeaAirPurifier', [{attribute: 'fanSpeed',
+                minimumReportInterval: 0, maximumReportInterval: repInterval.HOUR, reportableChange: 0}],
+                options);
+
+            await endpoint.read('manuSpecificIkeaAirPurifier', ['controlPanelLight', 'childLock']);
         },
     },
     {
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -627,13 +627,9 @@ module.exports = [
         vendor: 'IKEA',
         description: 'STARKVIND air purifier',
         exposes: [
-            e.fan().withModes(['off', 'low', 'medium', 'high', 'auto']),
-            exposes.enum('ikea_fan_mode', ea.ALL, [
-                'off', 'auto', '1', '1.5', '2', '2.5', '3', '3.5', '4', '4.5', '5',
-            ]).withDescription('Mode of this fan using finer fan speed control'),
-            exposes.enum('ikea_fan_speed', ea.STATE_GET, [
-                'off', 'auto', '1', '1.5', '2', '2.5', '3', '3.5', '4', '4.5', '5',
-            ]).withDescription('Current fan speed'),
+            e.fan().withModes(['off', 'auto', '1', '2', '3', '4', '5', '6', '7', '8', '9']),
+            exposes.numeric('fan_speed', exposes.access.STATE_GET).withValueMin(0).withValueMax(9)
+                .withDescription('Current fan speed'),
             e.pm25().withAccess(ea.STATE_GET),
             exposes.enum('air_quality', ea.STATE_GET, [
                 'good', 'ok', 'not_good', 'unknown',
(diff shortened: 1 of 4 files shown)


sjorge:
Thanks to @hakaneriksson for sniffing the traffic!

devices/ikea.js
```diff
@@ -627,18 +628,48 @@ module.exports = [
...
+            exposes.binary('child_lock', ea.ALL, 'LOCK', 'UNLOCK').withDescription('Enables/disables physical input on the device'),
+            exposes.binary('replace_filter', ea.STATE_GET, true, false)
+                .withDescription('Filter is older than 6 months and needs replacing'),
         ],
         meta: {fanStateOn: 'auto'},
-        fromZigbee: [fz.fan, fz.ikea_pm25],
-        toZigbee: [tz.fan_mode, tz.ikea_pm25],
+        fromZigbee: [fz.fan, fz.ikea_air_purifier],
+        toZigbee: [
+            tz.fan_mode, tz.ikea_air_purifier_fan_mode, tz.ikea_air_purifier_fan_speed,
```
Koenkk, line 649: Do I understand correctly that when e.g. setting a value trough `ikea_air_purifier_fan_mode` the device sends an update for the ZCL fan mode?

sjorge: No, ikea_air_purifier_fan_mode sends messages to the manufacturer specific cluster.

devices/ikea.js
```diff
@@ -627,18 +628,48 @@ module.exports = [
         description: 'STARKVIND air purifier',
         exposes: [
             e.fan().withModes(['off', 'low', 'medium', 'high', 'auto']),
+            exposes.enum('ikea_fan_mode', ea.ALL, [
+                'off', 'auto', '1', '1.5', '2', '2.5', '3', '3.5', '4', '4.5', '5',
+            ]).withDescription('Mode of this fan using finer fan speed control'),
+            exposes.enum('ikea_fan_speed', ea.STATE_GET, [
```
Koenkk, line 634: As you suggested in https://github.com/Koenkk/zigbee-herdsman/pull/455:

> @Koenkk input here would be welcome, I could just use a fan_mode and fan_speed state but then fan_mode will deviate from what is currently available in fan_mode and from the ZCL range, as this one would be based on the ikea cluster and not hvacFanCtrl, I kind of want... both ?

Deviating for the standard fan_mode is fine as long as it is in the `withModes([..])`, I cannot try it but https://github.com/Koenkk/zigbee2mqtt/blob/b3e9afaf3ee55e2d62f6775dc0c62e8550b8fef1/lib/extension/homeassistant.ts#L381 seems to be able to handle this.

sjorge: So if I understand correctly:
- drop fz.fan_mode, tz.fan-Mode
- update fz.ikea_air_purifier and tz.ikea_air_purifier_fan_mode and rename `ikea_fan_mode` to `fan_mode` and add `fan_state` to keep it consistant with tz.fan_mode's 2 properties
- for constancy also rename `ikea_fan_speed` property to just `fan_speed` ?

sjorge: The ikea app seems to use .5 increments but maybe we should just use 1-10 as a fan_speed/fan_mode level ? Those 'x.5' are a pain to send using mosquitto_pub depending on how they are quotes.

Koenkk: - With the original bridge, does the gateway use the ZCL fan cluster at all?


> The ikea app seems to use .5 increments but maybe we should just use 1-10 as a fan_speed/fan_mode level ? Those 'x.5' are a pain to send using mosquitto_pub depending on how they are quotes.

Good idea, so basically in toZigbee do value / 2, in fromZigbee * 2

sjorge: 0=off,
1=auto
10=1
15=2
20=3
25=4
30=5
35=6
40=7
45=8
50=9

Erm so 1-9, I thought there was a 55 too but there isn't.

Currently it's 10 * fan speed, so 10 * 1.5 => 15, ... although I'm using a constact that just maps it so we get off and auto in there too with just using getKey.

sjorge: >     * With the original bridge, does the gateway use the ZCL fan cluster at all?
AFAIK, know I think it's just there for compat with 3rd party gateways. It support off, auto, low, med, high according to the sequance cluster which gives access to about 1/3 of the speeds.

sjorge:
While testing I noticed sometimes fan_speed would not update, I don't know why as it is in the `state` var returned at the end of the fromZigbee converter, I added console log to dump msg.data and state just before it is returned:

```
XXX: data={"fanSpeed":30,"fanMode":30}, state={"fan_mode":"5","fan_state":"ON","fan_speed":5}
Zigbee2MQTT:info  2021-11-13 18:52:49: MQTT publish: topic 'zigbee2mqtt/airpurifier/bedroom', payload '{"air_quality":"unknown","child_lock":"UNLOCK","fan_mode":"5","fan_speed":4,"fan_state":"ON","led_enable":true,"linkquality":42,"pm25":0,"replace_filter":fals
```

sjorge:
All seems good, aside from the weird issue mention above but I don't think it's an issue with the converters, as when when I refresh fan_speed manually it does work properly, it just seems to ignore it when both fanMode and fanSpeed are update, yet both data and state that is returned seem correct. I wonder if something else inside z2m is eating that.

devices/ikea.js
Koenkk: > I think it's just there for compat with 3rd party gateways.

Makes sense, then this is good (just ikea specific converters)

Koenkk:
Looks good, thanks!

flatsiedatsie:
[This commit](https://github.com/Koenkk/zigbee-herdsman-converters/pull/3345/commits/d621fd3f7c9bd879789af4bcc033db2392c7a79e) set the correct exposes output to be levels between 0 and 5 (the same as indicated on the device). However, currently in the exposes data I'm getting I see 0 through 9 instead?

`ENUM values:  [ 'off', 'auto', '1', '2', '3', '4', '5', '6', '7', '8', '9' ]`

is this on purpose?

sjorge:
Yes, the IKEA app supports intermediate  levels too, using 0-9 makes math easier than using 2.5 for example. That is also why we moved away from the generic hvacCtrl and stated using IKEA’s custom cluster to her access to these additional speeds.

flatsiedatsie:
> using 0-9 makes math easier

What kind of math?

To me it seems the downside is that it may be confusing the end users? It confused me a little at first.

sjorge:
Why would we only want to support 5 level? The IKEA app supports all 9 levels, they just called them:
```
1
1.5
2
2.5
3
3.5
4
4.5
5
```

There are select-able in the Tradfri app, physically you cannot set the .5 speeds using the dial but, but they are supported by ikea.

> What kind of math?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/8f85478b55fbc66fd737e4b0121d118540ae995e/devices/ikea.js#L127-L136
https://github.com/Koenkk/zigbee-herdsman-converters/blob/8f85478b55fbc66fd737e4b0121d118540ae995e/devices/ikea.js#L138-L150

The values of the ikea specific cluster are (sorry it's been a while so I'm not 100% certain)
0 => off
1 => auto
10 => 1
15 => 1.5
20 => 2
...
50 => 5

This gives effectively off, auto and 9 levels of fan speed, now we just do some fancy math if the raw value is >= 10 to map that to 1-9, and if it's lower it's either off or auto.

There is also a read-only fanspeed property that skips 0 and 1 values and will always report the actual fan speed even when fanmode is set to auto.

flatsiedatsie:
ha, thanks for the detailed response :-)

As a designer I have a hard time agreeing that it's just cosmetic, although I understand where you're coming from. For me there is a value in knowing that if I set the 'virtual' value to 4, I will get the same amount of noise as when I use setting 4 on the device itself. It keeps things simple, mentally. As an interaction designer that simplicity is what I've been tought to strive for I guess. (Not meant as an argument to authority by the way, just explaining my.. desire.. for 5 levels.)

> given the device has well 9 speed levels.

Sure, but those could be represented in all kinds of ways. We could use 10, 20 .. 80, 90. The thing is that IKEA chose to squeeze it into 5 levels. Probably because it would keep things simple mentally, and also cheaper, as they would only need 5 leds to display the speed this way. You could say that IKEA was the first to express a preference.

Anyway, I accept that it will stay this way, don't get me wrong :-)


> I chose to stick closer to existing air quality ranges used by other sensors already implemented in zhc.

And I'm very glad you did!


> I did not find any other fan device that exposed a x.5 speed option

I guess we can both agree that IKEA should have just put 9 levels on the device and in the app ;-)

sjorge:
> I guess we can both agree that IKEA should have just put 9 levels on the device and in the app ;-)

It's even worse tbh, the device has 5, the ikea app has 9 but at x.5 intervals, AFAIK the leds seem to always reflect the lower end on the device. 

So selecting 1 or 1.5 on the ikea app has the LED at 1, 2 and 2.5 has the LED at 2, also when using auto mode it actually uses those x.5 levels too so you can be oh this is a nice sound level when it's on auto but you can never achief it with the physical knob. 

Child lock for example does have a physical LED on the device but no way, AFAIK to enable it without using the ikea app or z2m. It's certainly one of the more 'strange' devices they released.

hakaneriksson:
I agree that 1-9 is a bit confusing related to the physical knob. 0.5 steps would be preferable for UX I think or possibly 10, 15..50. In the Ikea app it's not either obvious that they have the half step. To toggle child lock on device you press the power button for 3 seconds. 
Screenshot_20211108-212700_Home smart

sjorge:
Switching it to 
```
1.0
1.5
2.0
2.5
3.0
3.5
4.0
4.5
5.0
```

Would not be to hard, not sure how that would play with things like home assistent though (cc: @Koenkk )

I have not super string preference for using 1-9, I do however have a strong preference for exposing all 9 levels as opposed to just 5.

Koenkk:
Maybe it would be more logical if percentage values are supported? This is also what HA does: https://www.home-assistant.io/integrations/fan.mqtt/#full-configuration

sjorge:
Are we using this anywhere at the moment?

I’m not sure how to cleanly map that though, there are only 9 speeds not 10 so we can’t just do 0-100 with a step of 10, we also have the auto value to somehow fit in this. (As this is fan_mode still)

~ sjorge

hakaneriksson:
I'm using the method set_percentage in HA Fan service currently in the custom card and it works as it is now.

sjorge:
https://github.com/Koenkk/zigbee2mqtt/blob/1a0a9a6eabf58855d3a6feda60bb79a1c34d17f8/lib/extension/homeassistant.ts#L433-L445

Looks like it auto detect this after filtering out `off` and `auto` from `fan_mode` leaving 1-9. I'm guessing this is also why all other devices with fan_mode also use whole numbers.

Probably another reason to just leave it as is to match other devices in z2m as we get this feature in homeassistent for free.

flatsiedatsie:
That does sound pretty useful. 

Ok, I'll implement a converter in the Webthings Z2M addon. Just wanted to make sure it wasn't better to do this upstream.

PR #3347, opened 2021-11-13
Update leedarson.js

jrosier:
Added devices:
E27 warm white
E27 tunable white
E14 tunable white
4-in-1 sensor
Motion sensor

### DIFF
--- a/devices/leedarson.js
+++ b/devices/leedarson.js
@@ -7,21 +7,21 @@ const e = exposes.presets;
 module.exports = [
     {
         zigbeeModel: ['ZBT-DIMLight-GLS0800'],
-        model: 'Leedarson E27 2700 K',
+        model: 'A60 806lm Dimmable E27',
         vendor: 'Leedarson',
         description: 'LED E27 warm white',
         extend: extend.light_onoff_brightness(),
     },
     {
         zigbeeModel: ['ZBT-CCTLight-GLS0904'],
-        model: 'Leedarson E27 tunable white',
+        model: 'A60 806lm Tunable White E27',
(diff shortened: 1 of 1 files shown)


devices/leedarson.js
```diff
@@ -5,6 +5,27 @@ const extend = require('../lib/extend');
...
+    {
+        zigbeeModel: ['ZBT-CCTLight-GLS0904'],
+        model: 'Leedarson E27 tunable white',
+        vendor: 'Leedarson',
+        description: 'LED E27 tunable white',
+        extend: extend.light_onoff_brightness_colortemp({colorTempRange: [153, 370]}),
+    },
+    {
+        zigbeeModel: ['ZBT-CCTLight-Candle0904'],
+        model: 'Leedarson E14 tunable white',
```
Koenkk, line 24: For this device and the 2 above please add the real model which is printed on the device/box (something that is searchable on google)

jrosier: These were samples I received (via another person) from the manufacturer. So unfortunately, I have no real model name, although I can use the names found at leedarson.com. They are sold under different brand names, e.g. Müller Licht.  But the zigbeeModel is published to zigbee2mqtt and therefore I had to add these.

jrosier: Found some model numbers and descriptions on the manufacturer's website, made some changes where applicable,

Koenkk: Thanks, then lets use the zigbee model instead. Could you make a pr to add pictures in https://github.com/Koenkk/zigbee2mqtt.io/tree/develop/public/images/devices ?

PR #3350, opened 2021-11-13
Update innr.js to support SP 234 Smart Plug 15A

albuquerquefx:
My first attempt at such a thing (Git and the rest!) so hopefully I got this right.  I've updated the INNR.JS to reflect the necessary configuration for the Innr SP 234 smart plug 15A, which supports both the basic on/off function and also various metering abilities. As you might guess, most of the functionality I cloned from the SP 120 entry, minus the power summary method which the Sp 234 does not appear to support.

I was able to validate this configuration using the code inside an External Converter .JS file within my own Zigbee2MQTT + Home Assistant Core instances, hosted within a BSD jail, connected to a ConBee 2. I'm using my two new SP 234 devices to monitor my dishwasher and my washing machine. All features appear to be available and report sensible data.

### DIFF
--- a/devices/innr.js
+++ b/devices/innr.js
@@ -478,6 +478,30 @@ module.exports = [
         },
     },
     {
+        zigbeeModel: ['SP 234'],
+        model: 'SP 234',
+        vendor: 'Innr',
+        description: 'Smart plug',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks

PR #3352, opened 2021-11-14
Rplace Zigbee manufacture to correct one

DJTerentjev:
Rplace Zigbee manufacture to correct one according to: https://github.com/Koenkk/zigbee2mqtt/issues/9465#issuecomment-968290949


Koenkk:
Thanks!

PR #3356, opened 2021-11-16
Added additional support for Philips Hue model 9290022268, Zigbee LWW002

beastie29a:
Initially the bulb joined as unsupported and came up with several options once the external converter file was configured. Also to note, the bulb came as part of the Philips Hue Inara Outdoor Lantern.


Koenkk:
Thanks!

PR #3362, opened 2021-11-16
Improve the mean_power calculation for the Elko thermostat

svhelge:
Correct the mean_power calculation for the Elko thermostat by exposing "load". 
In order to get a correct mean_power value back from the thermostat the power (load) when heating has to be configured for the device.


Koenkk:
Thanks!

PR #3364, opened 2021-11-17
Adding for WXKG01LM the possibility to enable a simple on-off mode

simonedimarzio:
Adding for WXKG01LM the possibility to enable a simple on-off mode:
- this option is useful when the device is connected to a common wall switch (when the click is never released in the <active> state)
- in this configuration only the 'single' and 'off' actions will alternate at each wall switch status change
- to enable this mode set 'mode_simple_on_off: true' in configuration.yaml device options

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5204,36 +5204,51 @@ const converters = {
             if (hasAlreadyProcessedMessage(msg)) return;
             const state = msg.data['onOff'];
 
-            // 0 = click down, 1 = click up, else = multiple clicks
-            if (state === 0) {
-                const timer = setTimeout(() => {
-                    publish({action: 'hold'});
-                    globalStore.putValue(msg.endpoint, 'timer', null);
-                    globalStore.putValue(msg.endpoint, 'hold', Date.now());
-                    const holdTimer = setTimeout(() => {
-                        globalStore.putValue(msg.endpoint, 'hold', false);
-                    }, options.hold_timeout_expire || 4000);
-                    globalStore.putValue(msg.endpoint, 'hold_timer', holdTimer);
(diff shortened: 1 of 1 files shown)


converters/fromZigbee.js
```diff
@@ -5204,36 +5204,51 @@ const converters = {
...
-                    publish({action: 'release', duration: duration});
-                    globalStore.putValue(msg.endpoint, 'hold', false);
+            if (options.mode_simple_on_off) {
+                // this option is useful when the device is connected to a common wall switch (when the click is never released in the <active> state)
+                // in this configuration only the 'single' and 'off' actions will alternate at each wall switch status change
+                // to enable this mode set 'mode_simple_on_off: true' in configuration.yaml device options
+                if (state === 0 || state === 1) {
+                    if (globalStore.getValue(msg.endpoint, 'click')) {
+                        globalStore.putValue(msg.endpoint, 'click', false);
+                        publish({action: 'single'});
```
Koenkk, line 5214: `action: 'on'` is better I think

Koenkk:
I'm don't think this logic should be on the z2m level, the state is e.g. not persisted across restarts of z2m

simonedimarzio:
> I'm don't think this logic should be on the z2m level, the state is e.g. not persisted across restarts of z2m

Any suggestions on where this logic shall be placed?
Anyway, I don’t think that loss of state after restart is a major issue, can be an assumption when using this mode.

Thanks
Simone

Koenkk:
In your controller software (e.g. Home Assistant)

simonedimarzio:
> In your controller software (e.g. Home Assistant)

Nope. The controller software does not receive any state change when the button is hold pressed.
The “simple mode” is done for this purpose, it triggers the controller software as soon as the switch state changes.

PR #3366, opened 2021-11-17
Add Chinese version of SSM-U02

DJTerentjev:


devices/xiaomi.js
```diff
@@ -1411,10 +1411,11 @@ module.exports = [
         ota: ota.zigbeeOTA,
     },
     {
-        zigbeeModel: ['lumi.switch.l0agl1'],
+        fingerprint: [{modelID: 'lumi.switch.l0acn1', manufacturerName: 'LUMI'}, {modelID: 'lumi.switch.l0agl1', manufacturerName: 'LUMI'}],
```
Koenkk, line 1414: Since this device uses a different model and zigbeeModel (meaning they can distinguished from each other) please add a new definition instead. `whiteLabel` should only be used when device cannot be distinguished from each other.

PR #3369, opened 2021-11-17
Switch to more verbose air_quality for STARKVIND

sjorge:
Originally this mapped to the limited set of 3 AQ values given in the ikea app.

Now it maps to the range given by EU for IAQ, using the same labels as the Develco Sensor for consistency.


devices/ikea.js
```diff
@@ -49,6 +53,161 @@ const tradfriExtend = {
...
+                        airQuality = 'poor';
+                    } else if (pm25 <= 75) {
+                        airQuality = 'unhealthy';
+                    } else if (pm25 <= 800) {
+                        airQuality = 'hazardous';
+                    } else if (pm25 < 65535) {
+                        airQuality = 'out_of_range';
+                    } else {
+                        airQuality = 'unknown';
+                    }
```
Koenkk, line 92: Not all of the above are in the `air_quality` expose, please add them.

sjorge: Good catch! I had a failing I forgot something but couldn't figure out what.

Koenkk:
Thanks!

PR #3370, opened 2021-11-17
Add definition for LCA009 (Hue White and color ambiance E26/A19 1600lm)

glward:
120V model of the new 1600lm Hue white and color bulbs


Koenkk:
Thanks!

PR #3383, opened 2021-11-20
Correct the color temperature range of 'JWSP001A'

airlychee:
Correct the color temperature range of the device


Koenkk:
Thanks!

PR #3385, opened 2021-11-20
Add version SML003 of Philips Hue Motion Detector

shypike:
Same as SML001, only has a different "zigbeeModel" and "model".

I tested this by adding it directly to the devices.js file in my zigbee2mqtt installation.
It doesn't look like the new model works any different than the previous one.
Although, the Hue app does make a distinction: when pairing it asks if the case says "Philips" or "Philips Hue".
The SML003 has "Philips Hue" on its case.

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1799,6 +1799,39 @@ module.exports = [
         ota: ota.zigbeeOTA,
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

shypike:
Wow, that's fast :)

PR #3386, opened 2021-11-20
Add support for Philips Aurelle 929003099101

filips:
I added support for a new model of the Philips Aurelle 120x30cm lamp.

For all I know it is feature identical with this one:
https://www.zigbee2mqtt.io/devices/3216331P6.html


Koenkk:
Thanks!

PR #3389, opened 2021-11-21
Correct the color temperature range of 'JWSP001A',fix QBKG40LM to get operation_mode states

airlychee:
Correct the color temperature range of the device


Koenkk:
Thanks!

PR #3390, opened 2021-11-21
Add Philips Hue 9290024689 bulb

aca02djr:
Bought this variant of the Hue white / color ambiance bulb today and was unsupported, so I have added it. Thanks!


Koenkk:
Thanks!

PR #3392, opened 2021-11-21
Support for 2021 Hue Cher Pendant

filips:
Here's another new Philips Hue product.

It is aesthetically and functionally similar to https://www.zigbee2mqtt.io/devices/4076131P6.html


Koenkk:
Thanks!

PR #3396, opened 2021-11-22
Add _TZ3000_ps3dmato plug to TS011F_plug_3

Danit2:
Add the LELLKI WK35 with power monitoring to  the TS011F_plug_3 because there also the applicationVersion: 69 and do not update the power automatic.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1018,11 +1018,12 @@ module.exports = [
             {modelID: 'TS011F', manufacturerName: '_TZ3000_dpo1ysak'},
             {modelID: 'TS011F', manufacturerName: '_TZ3000_cphmq0q7', applicationVersion: 69},
             {modelID: 'TS011F', manufacturerName: '_TZ3000_5f43h46b'},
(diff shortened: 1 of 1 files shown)


Koenkk:
There already exists a fingerprint with `_TZ3000_ps3dmato` without an application version in `TS011F_plug_1`, I would expect that one is picked first. Have you checked this?

If it picks `TS011F_plug_1` first you have to move the `TS011F_plug_3` above `TS011F_plug_1`

Danit2:
I have add my LELLKI WK35 to z2m. This was identified as `TS011F_plug_1`.

But i have the `appVersion: 69` so the energy reportig was not working. 

I don't know if there are plugs with a other `appVersion` and the fingerprint `_TZ3000_ps3dmato`

I don't know what is the best way so that nobody gets a problem.

Should i move the `plug_3` above the `plug_1`?

Koenkk:
Is it identified as `TS011F_plug_3` after this change?

Danit2:
Yes, it works. 

But i have only test it with the external converters.

I have add the plug_1 and the plug_3 in the same file. The plug_1 was first. 

But i have change it now.

It works also like this.

Koenkk:
Thanks!

PR #3397, opened 2021-11-22
Fix some problems of JWSP001A and JWSP001A

airlychee:
1.Correct the description and disable effect for JWSP001A.
2.fix QBKG40LM to get operation_mode states


Koenkk:
Thanks!

PR #3409, opened 2021-11-24
Add 8719514302235

LevitatingOrange:
This change adds the device with EAN 8719514302235, a philips hue filament bulb for e14 sockets, which is seemlingy only available in european markets. See also the pull request for the zigbee2mqtt.io device list [TODO, still have to do that].


Koenkk:
Thanks!

PR #3412, opened 2021-11-25
Valve position over reporting

Highvolt:
For all Haozee Thermostats the valve position is over reporting by a factor of 10, making 100% -> 1000%


Koenkk:
Thanks!

PR #3413, opened 2021-11-25
Update Lidl HG06463B description

tlvince:
This is a G95 globe bulb, not G30. Ref https://www.lidl.de/p/livarno-lux-leuchtmittel-spiralfilament-dimmbar-zigbee-smart-home/p100311870


Koenkk:
Thanks!

PR #3417, opened 2021-11-25
Added 929003055601 (4096730P6)

RonnyAL:
Different zigbeeModel than the existing entry. Seen in Norway.

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1221,7 +1221,7 @@ module.exports = [
         ota: ota.zigbeeOTA,
     },
     {
-        zigbeeModel: ['929003055601'], // The model ID from: Device with modelID 'lumi.sens' is not supported.
+        zigbeeModel: ['929003055601'],
         model: '4096730P6',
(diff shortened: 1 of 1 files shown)


devices/philips.js
```diff
@@ -1220,6 +1220,15 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp(),
         ota: ota.zigbeeOTA,
     },
+    {
+        zigbeeModel: ['929003055601'], // The model ID from: Device with modelID 'lumi.sens' is not supported.
```
sjorge, line 1224: Drop the comment on this line please.

RonnyAL: Done. I kept the original commit message though I'm not sure that's optimal. Again, I'm pretty new to this part of git.

sjorge: Koen squashes the commits when merging so it's not a problem.

Koenkk:
Thanks!

PR #3419, opened 2021-11-26
Deverses fix and open_window function

vladi1234:


converters/toZigbee.js
```diff
@@ -2838,7 +2837,7 @@ const converters = {
                     await tuya.sendDataPointEnum(entity, tuya.dataPoints.tvMode, 1 /* manual */);
                 }
                 break;
-            case 'window_detection':
+            case 'open_window':
```
Koenkk, line 2840: Why has this been renamed? For all other thermostats we use `window_detection`

vladi1234: Because it doesn't work as detection, I couldn't reproduce it. I have to use door / window contact to activate `open_window` so that `open_window_temperature` is activated.

PR #3421, opened 2021-11-27
Add support of CASA.IA  CCB432 Rail-Din relay and energy meter

simulot:
Here is the support for support of CASA.IA CCB432 rail-din relay and energy meter elaborated with the help of Burki24 on discord


Koenkk:
Thanks!

PR #3426, opened 2021-11-28
lib/color: syncColorState should take into account min/max of device

sjorge:
This fixes koenkk/zigbee2mqtt#9843


sjorge:
I played around with a Hue and Innr bulb and I did not get a color_temp value outside of the supported range when setting random colors using xy or hs.

Koenkk:
Thanks!

PR #3430, opened 2021-11-29
Added OWON THS317 temperature & humidity sensor

TimPox:
Configuration fails after inclusion, but if the unit's battery is removed and inserted back after inclusion the configuration succeeds and the unit works and reports normally.


Koenkk:
Thanks!

PR #3433, opened 2021-11-29
Support numeric value in Ikea tz air_purifier_fan_mode.

hakaneriksson:


Koenkk:
@sjorge could you review this?

devices/ikea.js
```diff
@@ -147,21 +147,24 @@ const ikea = {
...
+                switch (value) {
                 case 'off':
                     fanMode = 0;
                     break;
                 case 'auto':
                     fanMode = 1;
                     break;
                 default:
-                    fanMode = parseInt(((parseInt(value) / 2.0) * 10) + 5);
+                    fanMode = (+value / 0.2) + 5;
```
sjorge, line 163: The +value here is very confusing to read, can this go back to parseInt?

sjorge (changes requested):
The rest looks fine, it was originally string because the enum for ha was sending strings, accepting both is probably better

devices/ikea.js
hakaneriksson: Sure, reverted the unary plus operator. For me it got an error when setting fan mode from HA.

sjorge (approved):
LGTM

Koenkk:
Thanks!

PR #3439, opened 2021-12-01
Fix description

robertalexa:
Fixed misinterpretation of description. Thinking about it with a clear mind helped :) Also thanks @Mikk36 for raising this again and getting my brain in gear


Koenkk:
Thanks!

PR #3440, opened 2021-12-01
Updates to TS004F enable 12 states

fringefilmsoz:
Updates to make the TS004F work in 12 function mode. 4 buttons, each with press, double press and long press.

see https://github.com/Koenkk/zigbee2mqtt/discussions/7158#discussioncomment-1726532

Changes to two files only.
- Updates to TS004F model to run correct initialisation/paring sequence.
- Add to the tuya_on_off_action, alongside the TS0044.

Validated by a number of members in the above thread.


Koenkk:
Thanks!

PR #3442, opened 2021-12-01
Update tuya.js

StudioShemp:
Support for Mercator Ikuü SSW04 Quad switch no-neutral. Also corrected capitilisation on other Ikuü products (md files need update as Mercator Ikuü and Mercator ikuü are appearing in the supported devices list)


Koenkk:
Thanks!

PR #3445, opened 2021-12-01
Fix max boost_time for TRV TuYa TS0601_thermostat

Enzokot:
Max boost_time for TRV TuYa TS0601_thermostat is 900s
image


Koenkk:
Thanks!

PR #3454, opened 2021-12-03
Added manufacturer name for 'TuYa TS0601_cover'

DHM47:
I bought a Zemismart curtain motor model ZM79EDT which is supported according to the 'TuYa TS0601_cover' page.
image

However, it had a different manufacturer name than what is listed in the vendor file. So from my understanding, this should make it supported.
image


Koenkk:
thanks!

PR #3456, opened 2021-12-03
Added the support for Tuya Blind Motor _TZE200_r0jdjrvi

serhatVIRT:
Added the support for Tuya Blind Motor _TZE200_r0jdjrvi


Koenkk:
Thanks!

PR #3461, opened 2021-12-04
Estimate min/max value for numeric exposes with set access

Koenkk:
Attempt to provide a good estimate of the min/max value for numeric exposes which are settable. This should prevent HA complaining about the values being out of the default range (0-100) and improve the slider experience in the frontend/HA.

Also added a test case to ensure all settable numerics have a min/max value.

Closes https://github.com/Koenkk/zigbee2mqtt/issues/9972

CC: @sjorge @robertalexa @nurikk


devices/danfoss.js
```diff
@@ -45,7 +45,8 @@ module.exports = [
                 .withSystemMode(['heat']).withRunningState(['idle', 'heat'], ea.STATE),
             exposes.numeric('external_measured_room_sensor', ea.ALL)
                 .withDescription('Set at maximum 3 hours interval but not more often than every 30 minutes and 0.1 ' +
-                    'degrees difference. Resets every 3hours to standard. e.g. 21C = 2100 (-8000=undefined).'),
+                    'degrees difference. Resets every 3hours to standard. e.g. 21C = 2100 (-8000=undefined).')
+                .withValueMin(-8000).withValueMax(35),
```
sjorge, line 49: Max should be `3500` I think? Given that 21C = 2100 according to the description.

sjorge: Same for other thermostats based on this design

devices/immax.js
```diff
@@ -148,10 +148,14 @@ module.exports = [
         fromZigbee: [fz.battery, fz.ignore_basic_report, fz.illuminance, fz.ZB003X, fz.ZB003X_attr, fz.ZB003X_occupancy],
         toZigbee: [tz.ZB003X],
         exposes: [e.occupancy(), e.tamper(), e.battery(), e.illuminance(), e.illuminance_lux().withUnit('lx'), e.temperature(),
-            e.humidity(), exposes.numeric('reporting_time', ea.STATE_SET).withDescription('Reporting interval in minutes'),
-            exposes.numeric('temperature_calibration', ea.STATE_SET).withDescription('Temperature calibration'),
-            exposes.numeric('humidity_calibration', ea.STATE_SET).withDescription('Humidity calibration'),
-            exposes.numeric('illuminance_calibration', ea.STATE_SET).withDescription('Illuminance calibration'),
+            e.humidity(), exposes.numeric('reporting_time', ea.STATE_SET).withDescription('Reporting interval in minutes')
+                .withValueMin(0).withValueMax(1440),
+            exposes.numeric('temperature_calibration', ea.STATE_SET).withDescription('Temperature calibration')
```
sjorge, line 153: I'm confused, is this a internal device thing? Or should these really not be expose data but options ?

devices/danfoss.js
robertalexa: Should be -8000 to 3500.

Same for Hive and Popp

test/index.test.js
```diff
@@ -452,4 +452,19 @@ describe('index.js', () => {
         const definition = index.findByZigbeeModel('MCT-340 SMA');
         expect(index.getConfigureKey(definition)).toBe(1);
     });
+
+    it('Number exposes with set access should have a range', () => {
+        index.definitions.forEach((device) => {
+            if (device.exposes) {
+                const toCheck = [];
```
nurikk, line 459: Redundant variable

robertalexa:
I have left a few comments, apart from that, it looks good (as far as i can tell considering I am not familiar with some of the devices involved)

> This should prevent HA complaining

Currently because of those limits some of the attributes become unusable completely, so not just a warning. Based on that, it might be worth doing a hotfix release rather than wait for the usual 1st of the month?

Thanks for looking into this Koen

devices/immax.js
Koenkk: this device support calibration on the device side (instead of on the z2m side as usual)

Koenkk:
> Currently because of those limits some of the attributes become unusable completely, so not just a warning. Based on that, it might be worth doing a hotfix release rather than wait for the usual 1st of the month?

I only want to do this when many users are affected, I believe this is not the case. Those who are affected can use the dev branch for the time being.

robertalexa:
@Koenkk There was another review I have started which you have not answered

image

Koenkk:
@robertalexa I think you need to click the submit review button, comments that are Pending are not visible to others.

robertalexa:
Oh cr*p. Didn't realise that pending review are not visible. Please accept my apologies

devices/danfoss.js
```diff
@@ -64,7 +65,8 @@ module.exports = [
                 .withDescription('Whether or not the thermostat acts as standalone thermostat or shares load with other ' +
                     'thermostats in the room. The gateway must update load_room_mean if enabled.'),
             exposes.numeric('load_room_mean', ea.ALL)
-                .withDescription('Mean radiator load for room calculated by gateway for load balancing purposes'),
+                .withDescription('Mean radiator load for room calculated by gateway for load balancing purposes')
+                .withValueMin(-30).withValueMax(30),
```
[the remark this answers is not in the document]
Koenkk: Done, thanks!

Issue #3264, opened 2021-11-02
msg.device.getEndpoint is not a function


andrewostroumov:
After I've updated to the latest version of z2m I got this error from device TS0601 _TZE200_znzs7yaw
```
msg.device.getEndpoint is not a function
```

Logs
```
Zigbee2MQTT:debug 2021-11-02 09:58:01: Received Zigbee message from '0x842e14fffefdad21', type 'commandSetTimeRequest', cluster 'manuSpecificTuya', data '{"payloadSize":0}' from endpoint 1 with groupID 0
Zigbee2MQTT:error 2021-11-02 09:58:01: Exception while calling fromZigbee converter: msg.device.getEndpoint is not a function}
```

andrewostroumov:
Here the error https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/converters/fromZigbee.js#L2653

PR #3472, opened 2021-12-06
Fix hy_set_time_request converter getting endpoint

andrewostroumov:
I've got an issue with my Tuya TS0601 (_TZE200_znzs7yaw) thermostat and described here https://github.com/Koenkk/zigbee-herdsman-converters/issues/3264, so here the time of getting time from device


Koenkk:
Thanks!

PR #3474, opened 2021-12-06
Added support for new variant of IKEA LED1835C6

dsypniewski:
I've already tested this with the following external converter and it seems to work:
```
{
    zigbeeModel: ['TRADFRI bulb E17 WS 440lm'],
    model: 'LED1835C6',
    vendor: 'IKEA',
    description: 'TRADFRI bulb E17 WS 440 lumen, dimmable, white spectrum, opal white',
    extend: tradfriExtend.light_onoff_brightness_colortemp(),
}
```
One thing that could be possibly improved is to split those models so the description wouldn't contain three different screw sizes and 3 different lumen values in one description. I wonder how does IKEA release bulbs with the same model number but different specifications? :rofl: 

Info from log:
```
Warning Received message from unsupported device with Zigbee model 'TRADFRI bulb E17 WS 440lm' and manufacturer name 'IKEA of Sweden'
```

Product page: https://www.ikea.com/jp/en/p/tradfri-led-bulb-e17-440-lumen-wireless-dimmable-white-spectrum-chandelier-opal-white-00424320/

### DIFF
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -302,10 +302,10 @@ module.exports = [
         extend: tradfriExtend.light_onoff_brightness_colortemp(),
     },
     {
-        zigbeeModel: ['TRADFRI bulb E14 WS 470lm', 'TRADFRI bulb E12 WS 450lm'],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #3477, opened 2021-12-07
Add support for NUE-AUWZO2

danps1:
Adding support for NUE-AUWZO2 which is a new Double Power Point from Nue / 3A Smarthome.

These have a zigbeeModel of 'LXN-2S27LX1.0' which was detected as "Smart light switch - 2 gang v2.0" which is not correct.

I have removed that zigbeeModel from the Smart light switch, and created new entry for NUE-AUWZO2 as "Smart Zigbee Double Power Point" which is the name and model on the box and back of the device. Have confirmed that these settings work correctly for the 2x devices I just bought.

Cannot confirm that LXN-2S27LX1.0 is not also the zigbeeModel of a light switch, because I don't have any of these.

### DIFF
--- a/devices/nue_3a.js
+++ b/devices/nue_3a.js
@@ -286,7 +286,7 @@ module.exports = [
         zigbeeModel: ['LXN-2S27LX1.0'],
         model: 'NUE-AUWZO2',
         vendor: 'Nue / 3A',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #3478, opened 2021-12-08
Update philips.js

thatvalis:
Here is the matching device configuration to the image and .md file I uploaded to the zigbee2mqtt.io repository.
New at this so let me know if you need something else.


Koenkk:
Thanks!

PR #3481, opened 2021-12-08
Added new zigbee id for Philips Play Lightbar

Yunus1903:
I found a new zigbee model id for the [Philips Play Lightbar](https://www.zigbee2mqtt.io/devices/915005733701.html) after a recent purchase.
I have tested and confirmed that other than the different ID, it's completely the same.


Koenkk:
thanks!

PR #3483, opened 2021-12-08
Decrease minimum temperature change for SONOFF SNZB-02

jpbede:
Decrease the minimum temperature change the is reported, so the SONOFF SNZB-02 reports the temperature sooner.
This would fix the issue described at https://www.zigbee2mqtt.io/devices/SNZB-02.html.

Already tested this with my SNZB-02 sensor.


Koenkk:
Thanks!

PR #3490, opened 2021-12-10
Add 5128.10 illuminize shutter control

theimo1221:
Hello @Koenkk,

This is the first device which I add on my own, so please review.

I used the information from within ioBroker:
image

Additionally this ist stated in the device manual:
image

After adding device specific section and uploaded it to my iobroker instance I tried it in ioBroker object tree, through which I was able to control the shutter.

Additionally I tried to use the `closuresWindowCovering` which was working almost as wanted. But setting the position to 0% resulted in device going to 0 and then up to 7%

Best regards 

Thiemo


Koenkk:
Thanks!

Issue #3488, opened 2021-12-09
local temperature calibration are buggy after the last release and should accept float again


KartoffelToby:
Since the PR #3341 i have problems with my TRV calibration,

If i read the changes correctly the local_temperature_calibration have now a min and max and steps option,

by default min is set to -20 and max to 20 which should be good for most usecases, but my SPZB0001 and  SEA801-Zigbee/SEA802-Zigbee are limit to -6 and 6 thats not good, also see that the steps are 1 by default, as the most devices have 2 or more digs the calibration shoud accept float values too. It works well in older versions. 

Also have some bugs in rounding if a value outside of -6 6 is set and the device jumps to 4294967287

PR #3491, opened 2021-12-10
LocalTemperatureCalibration on TRVs to float and change max range

KartoffelToby:
For the issue #3488

This PR will change the calibration resolution to match the local temp resolution, it also make the max range wider to avoid issues that appeared since the last release


Padm59:
Please change change it so that we can use higher calibration precision!

Koenkk:
Thanks!

brunoleon:
Hello and thanks for your work.

I rebuilt zigbee2mqtt using tag 14.0.348 where this PR should be present, but the step is still 1 instead of 0.1.
Is there anything else to do to be able to use this code ?

Thanks

KartoffelToby:
@brunoleon I have Eurotronics Spirits, the 0.1 calibration will work with this patch for them and should work for others too, other TRVs have the problem of decimal to hex convert problems based on this PR. #3527 But i need help there

brunoleon:
OK @KartoffelToby I'll check how to test it next week probably

Se7enair:
@KartoffelToby I changed to the newest dev. The calibrationvalues are now float, but they where even with the latest stable. But the temperature from the thermostat is still only 0.5° steps.
I use also Eurotronic Thermostats

Zigbee2MQTT Version
1.22.1-dev commit: f9aa525

PR #3492, opened 2021-12-11
Add default attribute for visibility

vk496:
This attribute (activated by default) permits a dynamic exposition of entitities.

Coming from the issue https://github.com/Koenkk/zigbee2mqtt/issues/9929


vk496:
Since the exposes will be called as a function, this trick is no longer required.

More info: https://github.com/Koenkk/zigbee2mqtt/pull/10132#issuecomment-999105520

PR #3495, opened 2021-12-11
Add possibility to override OTA index with a local file

grafalex82:
This change adds a possibility to override and extend OTA index with a local index.json file. In turn this adds the following features:
- Adding OTA firmwares for own devices without having to modify Z2M/herdsman code
- Add new firmwares without uploading them to a global zigbee-OTA index
- At the same time changes in global zigbee-OTA repo will be still available

The code will first look at the overriden index file, then in global OTA index. This means there is a possibility to override a firmware for a specific device.

Note, that firmwares themselves have to be still downloaded from a http server. This part remains untouched.

Related zigbee2mqtt change is here: https://github.com/Koenkk/zigbee2mqtt/pull/10141
Discussion is here: https://github.com/Koenkk/zigbee2mqtt/discussions/10032


devbis (approved):
Nice functionality! Thanks!

grafalex82:
While a deeper testing I realized that override OTA index could be only a local file. With this change I am adding ability to use both local and http override index files.

Koenkk:
Thanks!

PR #3498, opened 2021-12-12
tuya MIR-HE200-TY - Changed motion from enum to binary for compatibility in Symcon

Burki24:


devices/tuya.js
```diff
@@ -1658,7 +1658,7 @@ module.exports = [
         toZigbee: [tz.tuya_radar_sensor],
         exposes: [
             e.illuminance_lux(), e.presence(),
-            exposes.enum('motion', ea.STATE, ['true', 'false']).withDescription('moving inside the range of the sensor'),
+            exposes.binary('motion', ea.STATE, ['true', 'false']).withDescription('moving inside the range of the sensor'),
```
Koenkk, line 1661: Please update the converter to expose the bool as a bool and not a string and change this to `exposes.binary('motion', ea.STATE, [true, false])`

Burki24: Hi Koenkk,

thanks for your help. The converter allready expose it as bool. So I've made only the mistake inside /devices/tuya.js.

Koenkk: I think the value is a string not a bool, please update https://github.com/Koenkk/zigbee-herdsman-converters/blob/d02a164ae94a0c51adca8ea31ff7aae141abe637/converters/fromZigbee.js#L7314 and to the same for `presence`

Burki24:
Hopefully, I finished that.

Koenkk:
Thanks!

Burki24:
Thanks back for bringing me to the right way.

PR #3499, opened 2021-12-12
danfoss: Add thermostat schedule support

kennylevinsen:
Note that I have not tested the popp/hive side of things.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -1154,6 +1154,20 @@ const converters = {
             await entity.read('hvacThermostat', ['ctrlSeqeOfOper']);
         },
     },
+    thermostat_programming_operation_mode: {
+        key: ['programming_operation_mode'],
+        convertSet: async (entity, key, value, meta) => {
+            const val = utils.getKey(constants.thermostatProgrammingOperationModes, value, undefined, Number);
+            if (val === undefined) {
+                throw new Error('Programming operation mode invalid, must be one of: ' +
+                    Object.values(constants.thermostatProgrammingOperationModes).join(', '));
+            }
+            await entity.write('hvacThermostat', {programingOperMode: val});
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('hvacThermostat', ['programingOperMode']);
+        },
+    },
     thermostat_temperature_display_mode: {
         key: ['temperature_display_mode'],
         convertSet: async (entity, key, value, meta) => {
--- a/devices/popp.js
+++ b/devices/popp.js
@@ -15,14 +15,15 @@ module.exports = [
         model: '701721',
         vendor: 'Popp',
         description: 'Smart thermostat',
-        fromZigbee: [fz.battery, fz.thermostat, fz.hvac_user_interface, fz.danfoss_thermostat],
+        fromZigbee: [fz.battery, fz.thermostat, fz.thermostat_weekly_schedule, fz.hvac_user_interface, fz.danfoss_thermostat],
         toZigbee: [tz.danfoss_thermostat_occupied_heating_setpoint, tz.thermostat_local_temperature, tz.danfoss_mounted_mode_active,
             tz.danfoss_mounted_mode_control, tz.danfoss_thermostat_vertical_orientation, tz.danfoss_algorithm_scale_factor,
             tz.danfoss_heat_available, tz.danfoss_heat_required, tz.danfoss_day_of_week, tz.danfoss_trigger_time,
             tz.danfoss_window_open_internal, tz.danfoss_window_open_external, tz.danfoss_load_estimate,
             tz.danfoss_viewing_direction, tz.danfoss_external_measured_room_sensor, tz.thermostat_keypad_lockout,
-            tz.thermostat_system_mode, tz.danfoss_load_balancing_enable, tz.danfoss_load_room_mean],
-        exposes: [e.battery(), e.keypad_lockout(),
+            tz.thermostat_system_mode, tz.danfoss_load_balancing_enable, tz.danfoss_load_room_mean,
+            tz.thermostat_weekly_schedule, tz.thermostat_clear_weekly_schedule, tz.thermostat_programming_operation_mode],
+        exposes: [e.battery(), e.keypad_lockout(), e.programming_operation_mode(),
             exposes.binary('mounted_mode_active', ea.STATE_GET, true, false)
                 .withDescription('Is the unit in mounting mode. This is set to `false` for mounted (already on ' +
                     'the radiator) or `true` for not mounted (after factory reset)'),
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -89,6 +89,10 @@ const converters = {
                 result[postfixWithEndpointName('control_sequence_of_operation', msg, model)] =
                     constants.thermostatControlSequenceOfOperations[msg.data['ctrlSeqeOfOper']];
             }
+            if (msg.data.hasOwnProperty('programingOperMode')) {
+                result[postfixWithEndpointName('programming_operation_mode', msg, model)] =
+                    constants.thermostatProgrammingOperationModes[msg.data['programingOperMode']];
+            }
             if (msg.data.hasOwnProperty('systemMode')) {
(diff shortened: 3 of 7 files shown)


converters/toZigbee.js
```diff
@@ -1154,6 +1154,19 @@ const converters = {
             await entity.read('hvacThermostat', ['ctrlSeqeOfOper']);
         },
     },
+    thermostat_programming_operation_mode: {
+        key: ['programming_operation_mode'],
+        convertSet: async (entity, key, value, meta) => {
+            let val = utils.getKey(constants.thermostatProgrammingOperationModes, value, undefined, Number);
+            if (val === undefined) {
```
sjorge, line 1161: Why not just
`let val = utils.getKey(constants.thermostatProgrammingOperationModes, value, value, Number);`

There result of that should be the same as far as I can see?

kennylevinsen: Hah, that was a terrible case of blind copy-pasta. It's taken from the "control sequence of operation" thing above it, and the snippet is as broken there as it is here...

I have changed it to throw an exception with the valid options instead.

sjorge: Throwing an exception does indeed make more sense!

devices/popp.js
```diff
@@ -66,7 +67,9 @@ module.exports = [
                 .withDescription('Mean radiator load for room calculated by gateway for load balancing purposes (-8000=undefined)')
                 .withValueMin(-8000).withValueMax(100),
             exposes.numeric('load_estimate', ea.STATE_GET)
-                .withDescription('Load estimate on this radiator')],
+                .withDescription('Load estimate on this radiator'),
+            exposes.enum('programming_operation_mode', ea.ALL, ['setpoint', 'schedule'])
```
Koenkk, line 71: Please move this expose to `exposes.js`, could you also improve the description?

kennylevinsen: > Please move this expose to exposes.js

As its own thing or as another `withXyz` on the climate expose?

> could you also improve the description

Sure, but it will be a little wordy.

Koenkk: > As its own thing or as another withXyz on the climate expose?

Own thing

> Sure, but it will be a little wordy.

No problem, no need to keep things short there.

Koenkk:
Looks good to me, if you can fix the merge conflict I can merge it.

kennylevinsen:
Sorry for the delay, rebased!

2Max2Pax:
Is this issue still open?

I 'm using Zigbee2MQTT version: 1.22.2 commit: 414c51f
In the "dev console" the attribute weekly_schedule (or thermostat_weekly_schedule )  is missing.

kennylevinsen:
It is merged and works. You set weekly_schedule over mqtt as an object that looks something like:

```javascript
{
   days: 0b1111111, // bitmask of days
   mode: 0x1, // heat mode
   transitions: [ // times are minutes since midnight
      {transitionTime: 660, heatSetpoint: 21.5},
      {transitionTime: 1200, heatSetpoint: 18.0},
   ]
}
```

Remember to change the programming operation mode to schedule for it to work.

2Max2Pax:
@kennylevinsen , thank you for the answer.
I published topic:
```
zigbee2mqtt/trv_studio/set/weekly_schedule
```
Payoload:
```
{"days": 0b1111111, "mode": 0x1, "transitions": [{"transitionTime": 660, "heatSetpoint": 21.5},{"transitionTime": 1200, "heatSetpoint": 18.0},]}
```
Tested also without the comma before `]`
I got the following error:
```
Publish 'set' 'weekly_schedule' to 'trv_studio' failed: 'TypeError: payload.transitions is not iterable'
```
I also tryed adding `"numoftrans":2` to the payload but I always get the same error

kennylevinsen:
For reference, this is the code snippet I used:

```javascript
const setSchedule = (schedule, days, radiator) => {
	console.log("Setting schedule" + (radiator ? " for " + radiator.key : ""), schedule);
	if (!days) {
		return;
	}

	for (let room_type in schedule) {
		let transitions = schedule[room_type];
		let payload = {
			dayofweek: days, // Already a bitmask
			mode: 0b01,      // Heating
			transitions: transitions.map(t => ({
				transitionTime: timeFromMidnight(t.time),
				heatSetpoint: t.temp,
			})),
		};
		payload.numoftrans = payload.transitions.length; // This should just be deduced by z2m...
		radiatorsForType(room_type)
			.filter(r => !radiator || radiator.key === r)
			.forEach(r => client.publish(r + "/set", JSON.stringify({
				weekly_schedule: payload,
				programming_operation_mode: "schedule",
			})));
	};
};
```

The error you are getting would be from [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/43266100bebf379d0ab8fe426f22ee2b41fd6016/converters/toZigbee.js#L1116), and seems to suggest that either there is no field named "transition" in what you sent, or it is not an array.

If you are sure that you are sending a well-formed JSON serialized message (remember to JSON.stringify), maybe add a log there to see if what the payload contains?

2Max2Pax:
I publish MQTT with mosquitto.
I made several tests, changing the value of `days:` , with or without comma before`]` in the topic
This is what i get subscribing the topic " zigbee2mqtt/#" :
```
zigbee2mqtt/trv_studio/set/weekly_schedule {"days": 0b1111111, "mode": 0x1, "transitions": [{"transitionTime": 660, "heatSetpoint": 21.5},{"transitionTime": 1200, "heatSetpoint": 18.0},]}
zigbee2mqtt/bridge/logging {"level":"error","message":"Publish 'set' 'weekly_schedule' to 'trv_studio' failed: 'TypeError: payload.transitions is not iterable'"}
```

I do not use java, so I do not use stringify.

in the whole log file, all the entries about weekly schedule look like:
```
[31mZigbee2MQTT:error[39m 2022-01-17 15:43:57: Publish 'set' 'weekly_schedule' to 'trv_studio' failed: 'TypeError: payload.transitions is not iterable'
```

kennylevinsen:
If that is what z2m is logging, then what you are sending is not valid JSON. What I quoted is JavaScript, and you need to serialize it as JSON first.

It should be enough to convert the binary and hex notated numbers to decimal and remove the trailing comma.

2Max2Pax:
Oooops, you are right. I just realized that in the test with decimal I made a typo ... corrected. No errors anymore.

Now, I still miss 2 points: 
1) How can I read back the weekly schedule?
2) If I want to make a different schedule for each day (or group of days) of the week, should I publish as many payloads as the dyas or group of days?

kennylevinsen:
1. You should be able to "get" the field as well.
2. Yes, write per schedule group you need. Note that any day included in a write is wiped of its previous schedule - there can only be one.

2Max2Pax:
I made some tests and **I could not find a way to read back the weekly schedule**. I suppose I do not use the correct payload or Z2M doesn't get correct data or doesn't expose it at all.

I noticed that I cannot set more than 6 transitions: Z2M doesn't return any error but the TRV simply doesn't apply any transition.
Changing numberOfWeeklyTrans and/or numberOfDailyTrans in the dev.console generates the following error:
```
Publish 'set' 'write' to 'trv_studio' failed: 'Error: Write 0x804b50fffe3da0a8/1 hvacThermostat({"numberOfDailyTrans":6,"numberOfWeeklyTrans":42}, {"sendWhenActive":true,"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'READ_ONLY')'
```
So there is no chance to use 7 or more transistions.

I'm going to test separate weekly schedule for today and tomorrow and see if it works.

kennylevinsen:
The numberOfXYZTrans fields tell you what the device limitation is - they are not something you can change. Doing more than 6 transitions in a day does seem aggressive for a thermostat considering the thermal mass of a room/home.

As for getting the weekly schedule, that may be a device bug. I recall someone saying that Danfoss thermostats only reply with their weekly schedule when asked *solely* for heat schedule, instead of all schedules as z2m asks. Not something I have tested. Would require a Danfoss quirk.

PR #3500, opened 2021-12-12
Osram AB3284 - add another manufacturer

lnxbil:
second batch of Osram AB32840 (Classic B40 TW - Lightify) and yet another manufacturer name.


Koenkk:
Thanks!

PR #3504, opened 2021-12-13
Update OSRAM DALI CONV LI

Danit2:
Update the Osram Dali Device.
Add four options for this device.
Add default binding of the pushbutton to endpoint 10 (first device) when the pushbuttoen is active.
Add default reporting when two devices ar connectet.

### DIFF
--- a/devices/osram.js
+++ b/devices/osram.js
@@ -401,7 +401,7 @@ module.exports = [
             {modelID: 'Zigbee 3.0 DALI CONV LI', endpoints: [
                 {ID: 10},
                 {ID: 25},
-                {ID: 242}
+                {ID: 242},
             ]},
         ],
         model: '4062172044776_1',
@@ -420,7 +420,7 @@ module.exports = [
             {modelID: 'Zigbee 3.0 DALI CONV LI', endpoints: [
                 {ID: 10},
                 {ID: 11},
-                {ID: 242}
+                {ID: 242},
             ]},
         ],
         model: '4062172044776_2',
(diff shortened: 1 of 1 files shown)


Koenkk:
If I understand correctly; the thing that is blocking to have it in one definition is that exposes is static, this is currently being worked on in https://github.com/Koenkk/zigbee2mqtt/pull/10132

Danit2:
Not only bequse of the static exposes.

When you add the device with one light it is good to have the MQTT Topic like `zigbee2mqtt/[FRIENDLY_NAME]/`
and when you add the device with two light like `zigbee2mqtt/[FRIENDLY_NAME]/l1` and `zigbee2mqtt/[FRIENDLY_NAME]/l2`

When you have every time two light in the definition and only one is connected, Z2M give everytime an error `[FRIENDLY_NAME] have no endpoint l2`. This is not nice.

It also make a different in the reporting if you have only one light on it or two.

Also the binding of the push button input cannot be done automatically when you have only one definition.

I think it makes sense to add all of the different definitions. Because this device is a different device with different settings and options in each mode.

Koenkk:
Could you explain me how to get the device in different modes?

Danit2:
Every time when you reset the device, the device will automatically check how much lights are connected.
After this you must press the push button input to activate this. (When you want) 
Then you can join to the network. 

So every time when you change something that also change the mode automatically, you have to reset and rejoin the device in the network.

Here you have the datasheet of the device: 
https://dammedia.osram.info/media/resource/hires/osram-dam-10310421/Zigbee%203.0%20DALI%20CONV%20LI.pdf

Koenkk:
Makes sense, thanks!

PR #3511, opened 2021-12-14
Add Elko/Schneider EKO09738 and EKO09716

Sabesto:
Add support for Elko/Schneider Electric EKO09738/EKO09716

[Elko Website EKO09716](https://elkosmart.elko.no/produkter/stikkontakt-enkel)
[Elko Website EKO09738](https://elkosmart.elko.no/produkter/stikkontakt-dobbel)

[Compliance Documentation EKO09716](https://zigbeealliance.org/zigbee_products/elko-smartsocket-single-16a/)
[Compliance Documentation EKO09738](https://zigbeealliance.org/no/zigbee_Produkter/elko-smartsocket-dobbel-16a/)

I have only tested EKO09738, but the only difference is that EKO09738 has an extra not-controlled/metered outlet. Looking at  Koenkk/zigbee2mqtt#7781 this should work.

The devices reports instantaneousDemand in mW, hence the new converter. I dont know if there is another way of doing it.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -652,6 +652,46 @@ const converters = {
             return payload;
         },
     },
+    EKO09738_metering: {
+        /**
+         * When using this converter also add the following to the configure method of the device:
+         * await readMeteringPowerConverterAttributes(endpoint);
+         */
+        cluster: 'seMetering',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const payload = {};
+            const multiplier = msg.endpoint.getClusterAttributeValue('seMetering', 'multiplier');
+            const divisor = msg.endpoint.getClusterAttributeValue('seMetering', 'divisor');
+            const factor = multiplier && divisor ? multiplier / divisor : null;
+
+            if (msg.data.hasOwnProperty('instantaneousDemand')) {
+                let power = msg.data['instantaneousDemand'];
+                if (factor != null) {
+                    power = (power * factor); // Unit reports in mW
+                }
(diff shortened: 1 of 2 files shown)


Koenkk:
You can control the used divisor in `configure()`, this should allow you to use the `fz.metering` converter (example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/2598565744a3474b23736d0b4b2be0f2c62a5170/devices/bitron.js#L140)

Sabesto:
> You can control the used divisor in `configure()`, this should allow you to use the `fz.metering` converter (example:
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/2598565744a3474b23736d0b4b2be0f2c62a5170/devices/bitron.js#L140
> 
> )
 seMetering divisor scales both instantaneousDemand and currentSummDelivered.  The  device reports currentSummDelivered in  Wh, while instantaneousDemand is reported in mW

Koenkk:
In that case it would be better to do something like https://github.com/Koenkk/zigbee-herdsman-converters/blob/a8fb06b7e94687c802c58497434db7dc8c34bd31/converters/fromZigbee.js#L602 and multiple either power or energy after calling the converter.

Sabesto:
> In that case it would be better to do something like
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/a8fb06b7e94687c802c58497434db7dc8c34bd31/converters/fromZigbee.js#L602
> 
> and multiple either power or energy after calling the converter.

Certainly! change done

Koenkk:
perfect, thanks!

PR #3515, opened 2021-12-14
Some modifications

z-x-f:
Modification of <code>aqara_opple</code>, syntax does not know the specification is not standardized, please check.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4957,15 +4957,15 @@ const converters = {
                     } else if (index === 3) payload.temperature = calibrateAndPrecisionRoundOptions(value, options, 'temperature'); // 0x03
                     else if (index === 100) {
                         if (['QBKG19LM', 'QBKG20LM', 'QBKG39LM', 'QBKG41LM', 'QBCZ15LM'].includes(model.model)) {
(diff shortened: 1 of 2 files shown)


converters/fromZigbee.js
```diff
@@ -4955,8 +4955,19 @@ const converters = {
                         payload.voltage = value;
                         payload.battery = batteryVoltageToPercentage(value, '3V_2100');
                     } else if (index === 3) payload.temperature = calibrateAndPrecisionRoundOptions(value, options, 'temperature'); // 0x03
-                    else if (index === 100) payload.state = value === 1 ? 'ON' : 'OFF'; // 0x64
-                    else if (index === 149) {
+                    else if (index === 100) {
+                        if (['QBKG19LM', 'QBKG20LM', 'QBKG39LM', 'QBKG41LM', 'QBCZ15LM'].includes(model.model)) {
+                            const mapping = model.model === 'QBCZ15LM' ? ['relay'] : ['left'];
```
Koenkk, line 4960: Why is mapping an array here? Can also be just `const mapping = model.model === 'QBCZ15LM' ? 'relay' : 'left';`

z-x-f: Because I'm not very familiar with programming languages :(

PR #3519, opened 2021-12-14
Add Schwaiger HAL500

nvschilleman:
Add support for Schwaiger HAL500, device reports manufacturer as LDS but I wasnt able to find any information about that brand name.

https://zigbee.blakadder.com/Schwaiger_HAL500.html shows this bulb as already supported by Z2M, zigbeeModel and supported features however don't match.


Koenkk:
Thanks!

PR #3520, opened 2021-12-15
Fix Sinope TH1124ZB and TH1300ZB piHeatingDemand

embak:
Like for the TH1123ZB, the TH1124ZB and TH1300ZB Sinope thermostats already report piHeatingDemand between 0 and 100.
This PR fix this.


Koenkk:
Thanks!

PR #3523, opened 2021-12-15
Add Support for zigfred uno by Siglis

dmugioiu:
Adapter file adds support for Smart In-Wall Switch zigfred uno.
You can find the product information here: https://www.zigfred.ch/

### DIFF
--- a/devices/siglis.js
+++ b/devices/siglis.js
@@ -8,9 +8,9 @@ const e = exposes.presets;
 module.exports = [
     {
         zigbeeModel: ['zigfred uno'],
-        model: 'zigfred uno',
+        model: 'ZFU-1D-CH',
         vendor: 'Siglis',
-        description: 'Smart In-Wall Switch',
+        description: 'zigfred uno Smart In-Wall Switch',
         exposes: [e.light_brightness_colorxy().withEndpoint('l1'), e.switch().withEndpoint('l2'), e.light_brightness().withEndpoint('l3')],
         fromZigbee: [fz.color_colortemp, fz.on_off, fz.brightness, fz.level_config, fz.power_on_behavior, fz.ignore_basic_report],
         toZigbee: [tz.light_onoff_brightness, tz.light_color, tz.ignore_transition, tz.ignore_rate, tz.light_brightness_move,
--- a/devices/siglis.js
+++ b/devices/siglis.js
@@ -10,7 +10,7 @@ module.exports = [
         zigbeeModel: ['zigfred uno'],
         model: 'ZFU-1D-CH',
         vendor: 'Siglis',
-        description: 'zigfred uno Smart In-Wall Switch',
+        description: 'zigfred uno smart in-wall switch',
         exposes: [e.light_brightness_colorxy().withEndpoint('l1'), e.switch().withEndpoint('l2'), e.light_brightness().withEndpoint('l3')],
         fromZigbee: [fz.color_colortemp, fz.on_off, fz.brightness, fz.level_config, fz.power_on_behavior, fz.ignore_basic_report],
         toZigbee: [tz.light_onoff_brightness, tz.light_color, tz.ignore_transition, tz.ignore_rate, tz.light_brightness_move,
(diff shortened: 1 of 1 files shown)


devices/siglis.js
```diff
@@ -0,0 +1,42 @@
...
+
+
+module.exports = [
+    {
+        zigbeeModel: ['zigfred uno'],
+        model: 'zigfred uno',
+        vendor: 'Siglis',
+        description: 'Smart In-Wall Switch',
+        exposes: [e.light_brightness_colorxy().withEndpoint('l1'), e.switch().withEndpoint('l2'), e.light_brightness().withEndpoint('l3')],
+        fromZigbee: [fz.color_colortemp, fz.on_off, fz.brightness, fz.level_config, fz.power_on_behavior, fz.ignore_basic_report],
```
Koenkk, line 15: Please use `extend:` instead of `fromZigbee`/`toZigbee`

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1075

devices/siglis.js
dmugioiu: Thanks for your review! I am struggling to see how I could use `extend: ` instead of `fromZigbee/toZigbee` cleanly, as this is a multi-endpoint composite device which has to expose three different device types `light_brightness_colorxy`, `switch`, `light_brightness` on three different endpoints.
Also, I can not find any similar examples in existing converters. Do you have a suggestion how to handle this situation?

dmugioiu: On second thought it might work this way, because `light_onoff_brightness_color` is a superset of the neccesary zigbee messages and `exposes ` will be overwritten. Do you agree?
```js
extend: extend.light_onoff_brightness_color(),
exposes: [e.light_brightness_colorxy().withEndpoint('l1'), e.switch().withEndpoint('l2'), e.light_brightness().withEndpoint('l3')],
```
I have no access to my test setup until monday, unfortunately.

Koenkk: > as this is a multi-endpoint composite device which has to expose three different device types

Missed that, sorry!

devices/siglis.js
```diff
@@ -0,0 +1,42 @@
...
+const tz = require('zigbee-herdsman-converters/converters/toZigbee');
+const exposes = require('zigbee-herdsman-converters/lib/exposes');
+const reporting = require('zigbee-herdsman-converters/lib/reporting');
+const e = exposes.presets;
+
+
+module.exports = [
+    {
+        zigbeeModel: ['zigfred uno'],
+        model: 'zigfred uno',
```
Koenkk, line 11: based on the picture on https://www.zigfred.ch/ this seems to should be `ZFU-1a-CH` (not sure if correct, hard to read)

dmugioiu: `zigbeeModel` is `['zigfred uno']` for sure, as reported by the firmware.
Regarding `model`, I propose to use the human readable model name zigfred uno for a better user experience, ZFU-1D-CH is the internal model name for certification. 
A possibility would be to use `zigfred uno (ZFU-1D-CH)` as `model`.

Koenkk: `ZFU-1D-CH` would be a good  `model` (since it yields results when searching for it on google)

devices/siglis.js
```diff
@@ -0,0 +1,42 @@
...
+const reporting = require('zigbee-herdsman-converters/lib/reporting');
+const e = exposes.presets;
+
+
+module.exports = [
+    {
+        zigbeeModel: ['zigfred uno'],
+        model: 'ZFU-1D-CH',
+        vendor: 'Siglis',
+        description: 'zigfred uno Smart In-Wall Switch',
```
Koenkk, line 13: Could you change this to `Zigfred uno smart in-wall switch` (brings it in line with the other descriptions), somehow I don't have edit rights

dmugioiu: Is `zigfred uno smart in-wall switch` also acceptable?
Reason: We avoid spelling "zigfred uno" with capitals in our Corporate Identity.
Of course I can change it like you proposed, if desired anyway.  :)

Koenkk: That is also fine!

Koenkk:
Thanks!

PR #3527, opened 2021-12-15
Prevent out of range error on tuya devices

KartoffelToby:
As the local calbiration should accept float values, the dec2hex conversion on tuya devices should handle it.

For now i have a simple int rounding fix, better way is to convert the hole float to hex.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -5373,6 +5373,7 @@ const converters = {
         key: ['local_temperature_calibration'],
         convertSet: async (entity, key, value, meta) => {
             if (value < 0) value = 0xFFFFFFFF + value + 1;
+            if (value % 1 !== 0) value = Math.round(Number(value));
             await tuya.sendDataPointValue(entity, tuya.dataPoints.saswellTempCalibration, value);
         },
     },
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -4973,6 +4973,27 @@ const converters = {
             }
         },
     },
+    nous_lcd_temperature_humidity_sensor: {
+        key: ['min_temperature', 'max_temperature', 'temperature_sensitivity', 'temperature_unit_convert'],
+        convertSet: async (entity, key, value, meta) => {
+            switch (key) {
+            case 'temperature_unit_convert':
+                await tuya.sendDataPointEnum(entity, tuya.dataPoints.nousTempUnitConvert, ['°C', '°F'].indexOf(value));
+                break;
+            case 'min_temperature':
+                await tuya.sendDataPointValue(entity, tuya.dataPoints.nousMinTemp, Math.round(value * 10));
+                break;
+            case 'max_temperature':
+                await tuya.sendDataPointValue(entity, tuya.dataPoints.nousMaxTemp, Math.round(value * 10));
+                break;
+            case 'temperature_sensitivity':
+                await tuya.sendDataPointValue(entity, tuya.dataPoints.nousTempSensitivity, Math.round(value * 10));
+                break;
+            default: // Unknown key
+                meta.logger.warn(`Unhandled key ${key}`);
+            }
+        },
+    },
     heiman_ir_remote: {
         key: ['send_key', 'create', 'learn', 'delete', 'get_list'],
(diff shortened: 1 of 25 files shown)


lib/tuya.js
```diff
@@ -47,6 +47,7 @@ function getDataValue(dataType, data) {
 }
 
 function convertDecimalValueTo4ByteHexArray(value) {
+    if(value % 1 !== 0) value = Math.round(Number(value))
```
KartoffelToby, line 50: Better were to convert the float value e.g. 2.4 to Hex, didn't know exacliy how is possible

lib/tuya.js
```diff
@@ -47,6 +47,7 @@ function getDataValue(dataType, data) {
 }
 
 function convertDecimalValueTo4ByteHexArray(value) {
+    if (value % 1 !== 0) value = Math.round(Number(value));
```
KartoffelToby, line 50: Better were to convert the float value e.g. 2.4 to Hex, didn't know exacliy how is possible

Koenkk:
@hacker-cb can you review this?

hacker-cb:
It's not good solution.

This case used only in `lidl_watering_timer` toZigbee converter.

@KartoffelToby seems you published not integer value to the `timer` key in MQTT and got this some error.

There are 2 solutions:

1. Don't put non-integer to the MQTT :)
2. Replace in `lidl_watering_timer` toZigbee converter:

```javascript
tuya.sendDataPointRaw(entity, tuya.dataPoints.lidlTimer, tuya.convertDecimalValueTo4ByteHexArray(value));
```

with

```javascript
tuya.sendDataPointValue(entity, tuya.dataPoints.lidlTimer, Math.round(value));
```

hacker-cb:
I'm not sure that Tuya supports float values in dp because both sides must know that there is float hex inside 4 hex bytes..

hacker-cb:
@KartoffelToby where are you using this case? I found only in `lidl_watering_timer`.

Are you writing some new converter?

converters/toZigbee.js
```diff
@@ -5373,6 +5373,7 @@ const converters = {
         key: ['local_temperature_calibration'],
```
KartoffelToby, line 5373: Its all about local_temperature_calibration, it should accept two digs resolution to match the local temp resolution, on devices like the Eurotronics Spirit it just works fine, but the tuya devices have this hex convertion thing. Its a little bit strange, i have saswell thermostats sometimes the value is rounded without any error, somtimes it throwes errors like "RangeError [ERR_OUT_OF_RANGE]: The value of "value" is out of range. It must be >= 0 and <= 255. Received 27_487_790_694" i think the hex converter should be able to convert 0.1 values as well or round it at least to prevent errors.

KartoffelToby:
@hacker-cb Think i go too deep, must be handled by the local_calibration functions, but i don't know every single tuya TRV

converters/toZigbee.js
hacker-cb: I guess it should be integer value with some precision.
For example, you can push `0.34` to the MQTT, but convert it to the integer with same precision as dp was configured on the device. For temperature I think it can be something like `100`, but need to sniff this value from the original app.

So, you should try something like `Math.round(value/100)` inside `local_temperature_calibration` function and pass integer value to the datapoint.

KartoffelToby: Sry i didn't understand your point, if you make a Math.round(value/100) on 0.34 or 2.3 it will be 0.

As an example, the saswell local_temperature is 18.8°C so if you want to calibrate it to 20°C the calibration must be 0.2

hacker-cb: Sorry, I mean `Math.round(value*100)` of cause.

hacker-cb: But I'm not sure that precision of this datapoint is exactly `100`. You should check it from the device docs or from the sniffer.

KartoffelToby: OK, yes i will investigate it, but i'm confused by the Hex 4 Bite stuff, maybe someone with more knowledge about tuya devices can write a proper converter, i tried the Math.round(value*10) this shoud be the right resolution, but get an range overflow to 4294967295 anyway

PR #3528, opened 2021-12-15
Add alternative Blitzwolf BW-SHP15 manufacturerName

wlcrs:
The new batch of Blitzwolf BW-SHP15 plugs that I received from Banggood have a new manufacturerName: '_TZ3000_u5u4cakc'

PR #3530, opened 2021-12-16
This batch of BW-SHP15 devices needs manual polling 😞

wlcrs:
After additional testing, this batch of plugs also doesn't seem to do any automatic reporting. This is also reported by other owners: https://github.com/Koenkk/zigbee-herdsman-converters/pull/3528#issuecomment-995623338

I've moved the fingerprint of my plug to the 'TS011F_plug_3' model, and added the exact applicationVersion for good measure.


Koenkk:
Thanks!

Schmetterfliege:
Hi, sorry for posting here but I have a question regardings this exact topic:

I have 10 of those Plugs, and all have that new firmware.
What would be your advice to me? Is there any solution for this polling thing that doesn't put a big load on my raspberry pi?
Is polling 10 of them e.g. every second with regards to ressources "neglectable"?
If so, how can I poll them?

wlcrs:
Hi @Schmetterfliege , hard to answer that question. Depends on what type of Raspberry Pi you have, how much load is already on that thing, ... You'll need to experiment to find out if it is manageable or not.

Schmetterfliege:
It's a Pi 4b with 4Gigs of memory. I'll try it to see if it's manageable - once I find out how to do that :D I'll clarify my question in the other Issue (#3528)

PR #3531, opened 2021-12-16
Fix toggle for _TZE200_amp6tsvy

sffetlio:
Fix for https://github.com/Koenkk/zigbee2mqtt/issues/8944 and https://github.com/Koenkk/zigbee2mqtt/issues/9280

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3213,8 +3213,8 @@ const converters = {
             const lookup = {l1: 1, l2: 2, l3: 3, l4: 4};
             const multiEndpoint = utils.getMetaValue(entity, meta.mapped, 'multiEndpoint', 'allEqual', false);
             const keyid = multiEndpoint ? lookup[meta.endpoint_name] : 1;
-            await tuya.sendDataPointBool(entity, keyid, value === 'ON');
-            return {state: {state: value.toUpperCase()}};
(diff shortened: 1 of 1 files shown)


converters/toZigbee.js
```diff
@@ -3213,8 +3213,8 @@ const converters = {
             const lookup = {l1: 1, l2: 2, l3: 3, l4: 4};
             const multiEndpoint = utils.getMetaValue(entity, meta.mapped, 'multiEndpoint', 'allEqual', false);
             const keyid = multiEndpoint ? lookup[meta.endpoint_name] : 1;
-            await tuya.sendDataPointBool(entity, keyid, value === 'ON');
-            return {state: {state: value.toUpperCase()}};
+            let newState = await converters.on_off.convertSet(entity, key, value, meta);
```
Koenkk, line 3216: You should not remove the `return` state here, otherwise the state will not be udpated.

sffetlio: I haven’t noticed such issue, the state seems to update correctly. How to test it? In fact with the return I get double log about state set.

Koenkk: It may work for your specific switch, but for TuYa switches that don't report the state themselves would break because of this.

I've updated the code, can you check if it works correctly with this?

PR #3532, opened 2021-12-16
Fix local_temp (diverse fix)

vladi1234:


vladi1234:
I cannot fix these error.


```

FAIL test/index.test.js (7.095 s)
  ● index.js › Exposes access matches toZigbee

    BRT-100-TRV - boost_heating_countdown, supports set: true

      393 |
      394 |                     if ((expose.access & exposes.access.SET) != (toZigbee && toZigbee.convertSet ? exposes.access.SET : 0)) {
    > 395 |                         throw new Error(`${device.model} - ${property}, supports set: ${!!(toZigbee && toZigbee.convertSet)}`);
          |                               ^
      396 |                     }
      397 |
      398 |                     if ((expose.access & exposes.access.GET) != (toZigbee && toZigbee.convertGet ? exposes.access.GET : 0)) {
```

devices/moes.js
This TVR function can only do ea.STATE

converters/fromZigbee.js
```diff
@@ -3598,7 +3598,7 @@ const converters = {
             case tuya.dataPoints.moesSheatingSetpoint:
                 return {current_heating_setpoint: value};
             case tuya.dataPoints.moesSlocalTemp:
-                return {local_temperature: (value / 10)};
+                return {local_temperature: (value / 10).toFixed(1)};
```
Koenkk, line 3601: `toFixed()` will return a string while this should be a number for sure.

converters/fromZigbee.js
```diff
@@ -3598,7 +3598,7 @@ const converters = {
             case tuya.dataPoints.moesSheatingSetpoint:
                 return {current_heating_setpoint: value};
             case tuya.dataPoints.moesSlocalTemp:
-                return {local_temperature: (value / 10)};
+                return {local_temperature: (value / 10).toFixed()};
```
Koenkk, line 3601: I see `toFixed()` is still used here (which will make it a string instead of a number)

vladi1234: It worked fine for me as before, but users really wanted to see .0 or .5 after the number, so I made .toFixed (1) that way.

What should I do that number looks like 20.0 or 20.5?

devices/moes.js
```diff
@@ -225,8 +225,7 @@ module.exports = [
                 'You can set up to 4 stages of temperature every for WEEKDAY ➀➁➂➃➄,  SATURDAY ➅ and SUNDAY ➆.'),
             exposes.binary('boost_heating', ea.STATE_SET, 'ON', 'OFF').withDescription('Boost Heating: press and hold "+" for 3 seconds, ' +
                 'the device will enter the boost heating mode, and the ▷╵◁ will flash. The countdown will be displayed in the APP'),
-            exposes.numeric('boost_heating_countdown', ea.STATE_SET).withUnit('min').withDescription('Countdown in minutes')
-                .withValueMin(0).withValueMax(15),
+            exposes.numeric('boost_heating_countdown', ea.STATE).withUnit('Min').withDescription('Countdown in minutes'),
```
Koenkk, line 228: There is a toZigbee converter for this so it seems to support `SET` (`tz.moesS_thermostat_boost_heating_countdown`)

vladi1234: I forgot, thank you!

Koenkk:
thanks!

PR #3537, opened 2021-12-18
Unconditionally add all generic toZigbee converters

### LINKED ISSUE #3261: Generic toZigbee converters are only added if a device has at least one converter defined
The generic `read` and `write` converters are useful for most devices, but they're not added unless a device has at least one toZigbee converter defined already, ref:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/1159051138d83a9aaef69d0e3f6cd5c46b901a45/index.js#L78-L80
They're useful on the dev console to probe which attributes a device supports, but it would be nice if they were added for all devices unconditionally.

normanr:
Generic toZigbee converters are only added if a device has at least one converter defined. They're useful for most devices, especially for using the dev console to probe which attributes are supported. Fixes #3261.

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,9 +75,7 @@ function addDefinition(definition) {
         };
     }
 
-    if (definition.toZigbee.length > 0) {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #3540, opened 2021-12-18
LELLKI WK34-EU Power Socket Wall EU with Power Measurment

DavidFFerreira:
Add Socket Wall EU with Power Measurment (WK34-EU)
It needs the const globalStore = require('../lib/store');

Picture of the device:
image

Aliexpress from Lellki
https://s.click.aliexpress.com/e/_9jXYiA

image
image


DavidFFerreira:
@Koenkk can you release in the next update? When?

PR #3545, opened 2021-12-18
Add Electrical Measurements to Sinope TH1300ZB

embak:
Add Zigbee haElectricalMeasurement with activePower, rmsCurrent and rmsVoltage reporting to Sinopé TH1300ZB.

Note: Sinopé has different firmware versions for each thermostats models.
Attribute activePower can be polled with meaningfull values but reporting was not supported on my TH1300ZB thermostats (FW 497 (2019-01-29) and 746 (2019-10-21)).


Koenkk:
Thanks!

PR #3551, opened 2021-12-20
Fix Sinope TH1123ZB configuration

embak:
Sinopé has different firmware versions which support different attributes reporting.

Code is refactored based on TH1124ZB to allow reporting of subsequent attributes when a previous one is not supported.


Koenkk:
Thanks!

PR #3557, opened 2021-12-21
Update philips.js

Burki24:
Added "duration" to exposes. Need this for using dimming-function in Symcon. Hopefully, it's okay for all.

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1739,7 +1739,7 @@ module.exports = [
         fromZigbee: [fz.ignore_command_on, fz.ignore_command_off, fz.ignore_command_step, fz.ignore_command_stop,
             fz.legacy.hue_dimmer_switch, fz.battery],
         exposes: [e.battery(), e.action(['on-press', 'on-hold', 'on-hold-release', 'up-press', 'up-hold', 'up-hold-release',
-            'down-press', 'down-hold', 'down-hold-release', 'off-press', 'off-hold', 'off-hold-release'])],
+            'down-press', 'down-hold', 'down-hold-release', 'off-press', 'off-hold', 'off-hold-release']), exposes.numeric('duration', ea.STATE).withUnit('ms')],
         toZigbee: [],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint1 = device.getEndpoint(1);
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1740,7 +1740,7 @@ module.exports = [
             fz.legacy.hue_dimmer_switch, fz.battery],
         exposes: [e.battery(), e.action(['on-press', 'on-hold', 'on-hold-release', 'up-press', 'up-hold', 'up-hold-release',
             'down-press', 'down-hold', 'down-hold-release', 'off-press', 'off-hold', 'off-hold-release']),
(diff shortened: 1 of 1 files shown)


devices/philips.js
```diff
@@ -1739,7 +1739,8 @@ module.exports = [
         fromZigbee: [fz.ignore_command_on, fz.ignore_command_off, fz.ignore_command_step, fz.ignore_command_stop,
             fz.legacy.hue_dimmer_switch, fz.battery],
         exposes: [e.battery(), e.action(['on-press', 'on-hold', 'on-hold-release', 'up-press', 'up-hold', 'up-hold-release',
-            'down-press', 'down-hold', 'down-hold-release', 'off-press', 'off-hold', 'off-hold-release'])],
+            'down-press', 'down-hold', 'down-hold-release', 'off-press', 'off-hold', 'off-hold-release']),
+        exposes.numeric('duration', ea.STATE).withUnit('s')],
```
Koenkk, line 1743: Duration is legacy and has been removed (https://www.zigbee2mqtt.io/devices/324131092621.html#options)

Burki24: What was the reason for that?
Did I missed something because it is send by the devices without having activate legacy-api.

Koenkk: > Did I missed something because it is send by the devices without having activate legacy-api.

The legacy api is enabled by default, you have to explicitly disable it as mentioned on https://www.zigbee2mqtt.io/devices/324131092621.html#options

> Is there an alternative, that I can use?

`action_duration`

Burki24: > The legacy api is enabled by default, you have to explicitly disable it as mentioned on

I thought, it's enough to disable legacy-api generally in settings->advance. Thanks. 
What is the diference between disabeling legacy-api in settings and disabeling legacy in devices?

> action_duration

Thanks, I'll try that.

Koenkk: The legacy api is the old MQTT api (has nothing to do with this)

Burki24: Thanks, I switched from duration to action_duration.

Koenkk:
I see you added duration to the action list now, but this is not correct, it should be a separate expose (like you had in https://github.com/Koenkk/zigbee-herdsman-converters/pull/3557/commits/c888fde0923804a70abfaa66b4bf60893df382bc but then called `action_duration` instead of `duration`)

devices/philips.js
Burki24: Hi Koenk,

I'm really thankfull for your support. That are mistakes, I do only once a time. But to my apologies, I am really a complete beginner at programming. 
Hopefully now it's all okay. On my local machine, it's running well.

Many, many thanks.

Burki24:
> I see you added duration to the action list now, but this is not correct, it should be a separate expose (like you had in [c888fde](https://github.com/Koenkk/zigbee-herdsman-converters/commit/c888fde0923804a70abfaa66b4bf60893df382bc) but then called `action_duration` instead of `duration`)

Hi Koenk,

I'm really thankfull for your support. That are mistakes, I do only once a time. But to my apologies, I am really a complete beginner at programming.
Hopefully now it's all okay. On my local machine, it's running well.

Many, many thanks.

Koenkk:
Looks good, thanks!

PR #3560, opened 2021-12-21
Update salus_controls.js add PS600 pipe temerature sensor

johnjamesmarsh:
Added PS600 Pipe Temperature Sensor (temperature, battery_low & tamper). The vendor shows as Computime and the zigbeeModel is SS909ZB. I have it reporting the temperature, battery low and tamper variables correctly once per minute. It also supports OTA.

I have 2 devices working via zigb3ee2mqtt & HomeAssistant. Happy to help anyone else who wants to test it

### DIFF
--- a/devices/salus_controls.js
+++ b/devices/salus_controls.js
@@ -6,6 +6,12 @@ const reporting = require('../lib/reporting');
 const extend = require('../lib/extend');
 const e = exposes.presets;
 
+const bind = async (endpoint, target, clusters) => {
+    for (const cluster of clusters) {
+        await endpoint.bind(cluster, target);
+    }
+};
+
 module.exports = [
     {
         zigbeeModel: ['SPE600'],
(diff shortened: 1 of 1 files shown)


devices/salus_controls.js
```diff
@@ -86,6 +86,20 @@ module.exports = [
...
         ota: ota.salus,
     },
+    {
+        zigbeeModel: ['SS909ZB'],
+        model: 'PS600',
+        vendor: 'Salus Controls',
+        description: 'Pipe temperature sensor',
+        fromZigbee: [fz.temperature],
+        toZigbee: [],
+        exposes: [e.temperature(), e.battery_low(), e.tamper()],
```
Koenkk, line 96: There are no fromZigbee converters which expose battery low and tamper so these should be removed or a converter should be added to fromzigbee

johnjamesmarsh:
Removed battery low and tamper "expose" statements. They are there so will work on the correct config for them later and resubmit

johnjamesmarsh:
Removed battery low and tamper "expose" statements. They are there so will work on the correct config for them later and resubmit

devices/salus_controls.js
```diff
@@ -83,7 +83,21 @@ module.exports = [
         description: 'Door or window contact sensor',
         fromZigbee: [fz.ias_contact_alarm_1],
         toZigbee: [],
-        exposes: [e.contact(), e.battery_low(), e.tamper()],
+        exposes: [e.contact()],
```
Koenkk, line 86: Why was `e.battery_low(), e.tamper()` removed here? `fz.ias_contact_alarm_1` supports these.

johnjamesmarsh: That was a fat fingered fumble. I was changing the PS600 line and somehow modified the OS600 line too. Sorry about that, I'll resubmit.

johnjamesmarsh:
I'm sorry, I'm making a real mess of this. I thought I knew what I was doing but the evidence does not support that. I'm just trying here to restore your version of the file

johnjamesmarsh:
I think this version has everything correct in it. I have tested it back in my system and the sensor is still reporting temperature.

Koenkk:
Thanks!

PR #3568, opened 2021-12-23
Add power on behavior for Ikea drivers

dchesterton:


Koenkk:
Thanks!

WasVerzachter:
Hi, I'm afraid something is not 100% OK; my drivers do not react on pressing on/off with IKEA remote E1743 or pressing the middle button on a IKEA E1524/E1810 remote (dimmer/brighter keys do work!).
In the debug screen of Zigbee2mqtt I can see the push action pass, however the driver does not react on it ..

PR #3569, opened 2021-12-23
Support Hue Gradient Signe table lamp (white)

agross:
https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-gradient-signe-table-lamp/046677803483


Koenkk:
Thanks!

PR #3574, opened 2021-12-25
fix MCCGQ12LM

nukusinji:
debug log:
No converter available for 'MCCGQ12LM' with cluster 'genOnOff' and type 'attributeReport' and data '{"onOff":1}'
No converter available for 'MCCGQ12LM' with cluster 'genOnOff' and type 'attributeReport' and data '{"onOff":0}'

Input claster show 'ssIasZone' not 'genonoff'
but real data is coming from 'genonoff'.


Koenkk:
Thanks!

PR #3576, opened 2021-12-26
Fix max limit for load mean

robertalexa:
This is a further change to your initial min/max estimation
https://github.com/Koenkk/zigbee-herdsman-converters/pull/3461

It turns out I was wrong about the max limit for load mean, where I have previously thought the limit is up to 100.

After some talking with users we have increased the limit. https://github.com/Koenkk/zigbee-herdsman-converters/pull/3239#issuecomment-1001056511

Thanks @Koenkk 
Hope you had a lovely Christmas and wishing you a Happy New Year!

### DIFF
--- a/devices/danfoss.js
+++ b/devices/danfoss.js
@@ -72,7 +72,7 @@ module.exports = [
                     'thermostats in the room. The gateway must update load_room_mean if enabled.'),
             exposes.numeric('load_room_mean', ea.ALL)
                 .withDescription('Mean radiator load for room calculated by gateway for load balancing purposes (-8000=undefined)')
-                .withValueMin(-8000).withValueMax(100),
+                .withValueMin(-8000).withValueMax(2000),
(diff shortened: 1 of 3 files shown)


kennylevinsen:
What's the upper limit in the Danfoss Ally spec? Maybe just use that?

Char-r:
> What's the upper limit in the Danfoss Ally spec? Maybe just use that?

Nothing in the specs, just says it is calculated by the gateway

kennylevinsen:
The Danfoss spec defines the attribute as having a max value of 0x7FFF (range max column). It's probably better to use that instead of guessing soft limits.

robertalexa:
0x7FFF  = 32767 decimal. That makes the HA slider unusable.

That limit is basically a fictitious one by the manufacturer, but that high value will never be used.

Hence 2000, since the highest i found in my testing and the data you 2 provided was in the region of 1.5k - 1.6k

Koenkk:
Thanks!

PR #3578, opened 2021-12-26
Expose additional features for Aqara Magic Cube

declanshanaghy:
* Exposes `side` to enable retrieval of the current side facing up
  * `side` remains set persistently, whereas `action_side` is blanked after the action is published
* Sets `action_side` in events where previously only `side` was being set

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5760,13 +5760,30 @@ const converters = {
             if (value === 0) result = {action: 'shake'};
             else if (value === 2) result = {action: 'wakeup'};
             else if (value === 3) result = {action: 'fall'};
-            else if (value >= 512) result = {action: 'tap', side: value-512};
-            else if (value >= 256) result = {action: 'slide', side: value-256};
-            else if (value >= 128) result = {action: 'flip180', side: value-128};
+            else if (value >= 512) result = {
+                action: 'tap',
+                action_side: value-512,
+                side: value-512
+            };
+            else if (value >= 256) result = {
+                action: 'slide',
+                action_side: value-256,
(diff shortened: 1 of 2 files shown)


converters/fromZigbee.js
```diff
@@ -5760,9 +5760,9 @@ const converters = {
             if (value === 0) result = {action: 'shake'};
             else if (value === 2) result = {action: 'wakeup'};
             else if (value === 3) result = {action: 'fall'};
-            else if (value >= 512) result = {action: 'tap', side: value-512};
-            else if (value >= 256) result = {action: 'slide', side: value-256};
-            else if (value >= 128) result = {action: 'flip180', side: value-128};
+            else if (value >= 512) result = {action: 'tap', action_side: value-512, side: value-512};
+            else if (value >= 256) result = {action: 'slide', action_side: value-256, side: value-256};
+            else if (value >= 128) result = {action: 'flip180', action_side: value-128, side: value-128};
```
Koenkk, line 5765: Why add `action_side` here? This is a duplicate of `side`.

flatsiedatsie: "side remains set persistently, whereas action_side is blanked after the action is published" ?

Then again, this 'blanking' is probably only experienced if the Home Assistant option is enabled. Disabling what might have the same intended effect?

Koenkk: I know, but I don't understand what value this adds, it is exactly the same as `side` only that it is cleared.

flatsiedatsie: Yeah I'm agreeing with you :-)

PR #3584, opened 2021-12-29
Support for 9290030674, clone of 9290019758

ketjap:
Looks like Philips released a newer version of 9290019758. I changed the values to 9290030674 and it looks like the rest is working exactly the same.


Koenkk:
Thanks!

PR #3593, opened 2021-12-31
Added LoraTap model SS6400ZB

metlinux:
Added LoraTap model SS6400ZB as equivalent for Tuya Model TS0044
It may looks different (from visual point of view) but it works 100% the same and is also already recognized as the Tuya TS0044


Koenkk:
Thanks!

PR #3595, opened 2021-12-31
_TZE200_a4bpgplm has while label GTZ06

mgwoj:
_TZE200_a4bpgplm has while label GTZ06 and it is sold by id3.pl / https://www.houseiq.pl/pl/p/Glowica-termostatyczna-ZigBee-3.0-GTZ06-TUYA/1620


Koenkk:
Thanks!

PR #3596, opened 2021-12-31
Support for _TZE200_zr9c0day to report heating state (is valve opened)

mgwoj:


Koenkk:
Thanks!

onpaj:
Do I have to update coordinator to make this actually work?

PR #3590, opened 2021-12-29
Update tuya.js

kfir-bar:
My first :) change for the item - https://neatho.me/en/catalog/smart-gadget/controller/tuya-zigbee-smart-knob-wireless-scene-switch-button/
* I didn't attach a picture.

PR #3598, opened 2022-01-01
Add another definition of ERS-10TZBVK-AA (#3590)

ginkage:
I seem to have a different revision of this device, with different fingerprint.


Koenkk:
Thanks!

PR #3599, opened 2022-01-01
Fix max/min temperature for TuYa TS0601_thermostat

Enzokot:
Max temperature for TRV TuYa TS0601_thermostat is 70 °C
Min temperature for TRV TuYa TS0601_thermostat is 1 °C
image


Kup9l777:
Hello! Please tell me how can I update and fix the above error?

PR #3600, opened 2022-01-01
Fix RTCGQ13LM `occupancy_timeout`

z-x-f:
Fix https://github.com/Koenkk/zigbee2mqtt/issues/10471
I am not sure what caused the original <code>occupancy_timeout</code> to not take effect, but I think it should be correct now.
The original <code>occupancy_timeout</code> should be <code>interval</code>.


Koenkk:
Is there a default occupancy timeout configured?

z-x-f:
I use aqara gateway to sniff and see that the device does not send occupied: false.

Koenkk:
Ah I see, `fz. occupancy_with_timeout` fix this indeed, thanks!

PR #3605, opened 2022-01-02
add namron 3802966

engvik:
Add support for device [Namron 3802966](https://www.namron.com/products/namron-zigbee-led-paere-4-8w-cct-gu10/).


Koenkk:
Thanks!

PR #3613, opened 2022-01-03
Use postfixWithEndpointName() for simluated_brightness

ivareri:
Add endpoint name for simulated brightness. Useful for multichannel switches like https://www.zigbee2mqtt.io/devices/4512703.html#namron-4512703

Values was already stored for each endpoint, so edits are only for payload / publish. Tested this with the Namron 4512703. 

Code is basically the same as the one used for the "normal" brightness control, so I think it should behave well with other multichannel or single channel switches, but I don't have any switches on hand for testing.


Koenkk:
Thanks!

PR #3614, opened 2022-01-03
Add support for iris 3450-L2

NPutting:
Tested with a copy of 3450-L and it looks to work for all buttons and presence detection.


Koenkk:
Thanks!

PR #3615, opened 2022-01-03
tuya_smoke: Accept commandDataReport in addition to commandDataResponse

gmbnomis:
Newer(?) versions of the TS0601 (_TZE200_uebojraa) smoke detector seem to use
commandDataReport instead of/in addition to commandDataResponse payloads.

Add commandDataReport to the accepted types.  While at it, change the
converter to use the new "multi-dp for loop" approach.  (A new DP 14 shows
up when doing this, but we currently don't know what it is)

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/10464

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4682,17 +4682,21 @@ const converters = {
     },
     tuya_smoke: {
         cluster: 'manuSpecificTuya',
-        type: ['commandDataResponse'],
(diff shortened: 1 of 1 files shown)


gmbnomis:
@Roelski tested this code in an external converter

Koenkk:
Thanks!

PR #3616, opened 2022-01-04
Add support for Schneider Electric 545D6115

kenni:
Add support for Schneider Electric LK Fuga wiser wireless socket outlet (545D6115)


Koenkk:
Thanks!

PR #3617, opened 2022-01-04
Fix exposes for BTicino switch and dimmer

Dreaming-Codes:
Light switch with neutral:
 \+ led_when_on
 \+ permanent_led

Dimmer switch with neutral:
 \+ dimmer_enabled
 \+ permanent_led
 \+ led_when_on


Koenkk:
thanks!

PR #3620, opened 2022-01-04
Skydance 5-in-1 LED Controller WZ5

kirovilya:
http://www.iskydance.com/index.php?c=product_show&a=index&id=1584

Based on previous developments from [this PR](https://github.com/Koenkk/zigbee-herdsman-converters/pull/2437) and [this issue](https://github.com/Koenkk/zigbee2mqtt/issues/6867).

No effects, but basic functions work.


devices/skydance.js
```diff
@@ -0,0 +1,74 @@
...
+        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_3thxjahu'}],
+        model: 'WZ5_rgbw',
+        vendor: 'TuYa',
+        description: 'Zigbee & RF 5 in 1 LED Controller WZ5 (RGBW mode)',
+        fromZigbee: [fz.tuya_light_wz5],
+        toZigbee: [tz.tuya_dimmer_state, tz.tuya_light_wz5],
+        exposes: [
+            exposes.light().withBrightness().setAccess('state', ea.STATE_SET).setAccess('brightness',
+                ea.STATE_SET).withColor('hs'),
+            exposes.numeric('white_brightness', ea.STATE_SET).withValueMin(0).withValueMax(254).withDescription(
```
Koenkk, line 53: Can this channel be controlled separately from the color channel?

kirovilya: only brightness.

Koenkk: So the white brightness can be 100% when the color brightness is also 100%? If yes then we should treat this as a separate light

kirovilya: yes. it is separate. I don't understand what's wrong ...

Koenkk: Currently it is exposes as a separate sensor, could it be exposed as a separate light? (so this device exposes 2 lights, 1 for color, 1 for white channel). With the current config only 1 light is exposed.

kirovilya: image
hmm ... if make another light, will have to remove the on / off switch, leave only the brightness.
so I decided to leave white brightness as an individual level. maybe I'm wrong

Koenkk: what happens if you turn the bulb off and set white_brightness to e.g. 200?

kirovilya: nothing

kirovilya: Снимок экрана от 2022-01-06 15-23-34

There are 5 work modes, each of which uses different outputs. In the last two modes, we can separately control the brightness of the white channel. To distinguish these last two modes, I added an attribute to the meta, which I check in the converter code and adjust the brightness accordingly.

Koenkk:
Thanks!

kirovilya:
Thank you, but https://github.com/Koenkk/zigbee-herdsman-converters/pull/3632

PR #3627, opened 2022-01-05
Update centralite.js

mwav3:
I'm submitting this PR to update the Centralite  3157100-E Thermostat to report the status of and control the temperature hold function.  I tested this on the thermostat I have and it works - it exposes a binary to control the value similar to other thermostats that support this function.


Koenkk:
Thanks!

PR #3633, opened 2022-01-06
Support changing per device scene states

RobertMe:
This is my first PR on Zigbee2mqtt to please bear with me :)

As per the discussion in the original ticket to add scenes support to Zigbee2mqtt scenes are stored in the lights based on a scene id and group id for which the group id defaults to 0 when setting the state on a light (using `zigbee2mqtt/<light name>/set`). This means one can currently only `store` a whole group of devices or `add` a whole group of devices with the same state. With this change one can call `scene_add` (or `scene_store`) on a light and by specifying the `group_id` the scene will be bound to that group.

So in short this allows:
1. Create a group with 5 lamps
2. `scene_store` or `scene_add` a scene for all lights in the group
3. Call `scene_add` on `zigbee2mqtt/<light name>/set` to override the state of a single light

Up for discussion is whether the same change should be applied to `scene_remove` as well. My guess (but didn't test) is that this would thus prevent such light to change it's state when a (group) scene is applied.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -5064,7 +5064,7 @@ const converters = {
         key: ['scene_store'],
         convertSet: async (entity, key, value, meta) => {
             const isGroup = entity.constructor.name === 'Group';
-            const groupid = isGroup ? entity.groupID : 0;
+            const groupid = isGroup ? entity.groupID : value.hasOwnProperty('group_id') ? value.group_id : 0;
             let sceneid = value;
             let scenename = null;
             if (typeof value === 'object') {
@@ -5164,7 +5164,7 @@ const converters = {
             }
 
             const isGroup = entity.constructor.name === 'Group';
-            const groupid = isGroup ? entity.groupID : 0;
+            const groupid = isGroup ? entity.groupID : value.hasOwnProperty('group_id') ? value.group_id : 0;
             const sceneid = value.ID;
             const scenename = value.name;
             const transtime = value.hasOwnProperty('transition') ? value.transition : 0;


Koenkk:
- Could you also make a pr to update the docs: https://www.zigbee2mqtt.io/guide/usage/scenes.html
- Have you checked that the correct state is published to MQTT when restoring the state?

RobertMe:
> * Could you also make a pr to update the docs: https://www.zigbee2mqtt.io/guide/usage/scenes.html

Sure, I will try to do so during the weekend

>     * Have you checked that the correct state is published to MQTT when restoring the state?

Nope, I didn't, but I just did :). The states as published for my Hue lights is correct and lights with a differing state also reflect this in the published messages.
One small sidenote though is that the state of the group is updated multiple times and reflects that of the light which was updated "last". So when having four lights in a group, 1, 2, 3 and 4, then when for example 2 has a different state then you might / will get the situation where the following messages are published:
1. 1 updated to [common state]
2. Group updated to [common state]
3. 2 updated to [custom state]
4. Group updated to [custom state]
5. 3 updated to [common state]
6. Group updated to [common state]
7. 4 updated to [common state]

(There doesn't seem to be a group update after light 4 changed, I presume because the state didn't change since the last message).

BTW: Personally I don't think this is part of this PR. This as previously one could already set multiple lights in the same group to a different state, after which one could store the group as a scene. So using those steps one could already create a scene without all lights having the same state.

RobertMe:
> * Could you also make a pr to update the docs: https://www.zigbee2mqtt.io/guide/usage/scenes.html

Done :) https://github.com/Koenkk/zigbee2mqtt.io/pull/1127

Koenkk:
thanks!

PR #3634, opened 2022-01-06
Add support for iL03_1 smart plug

NPutting:
Adding support for the iL03_1 smart plug from Iris. As far as I can tell it doesn't have power monitoring and is just on/off.

### DIFF
--- a/devices/iris.js
+++ b/devices/iris.js
@@ -144,4 +144,18 @@ module.exports = [
         },
         exposes: [e.switch(), e.battery()],
     },
+    {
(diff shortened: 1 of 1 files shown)


NPutting:
Power (battery vs AC) is showing as unknown but I'm unsure how to resolve that.

Koenkk:
Thanks,

> Power (battery vs AC) is showing as unknown but I'm unsure how to resolve that.


Added something to the `configure` to force the power source so this should be fixed

PR #3637, opened 2022-01-07
Fix IL06_1 battery

NPutting:
The Iris IL06_1 contact sensor wasn't being detected as battery mode for either me or my friend. Tested locally and it now shows both the % and `Battery OK` in the dashboard.


Koenkk:
Thanks!

PR #3619, opened 2022-01-04
Support for "Tuya Smart Vibration Sensor"

flatsiedatsie:
Adds support for "Tuya Smart Vibration Sensor".

This is my first attempt at adding a device, so perhaps I made mistakes. Feedback is very welcome.

See also Github issue: https://github.com/Koenkk/zigbee2mqtt/issues/10534

PR #3638, opened 2022-01-07
Fix `tuya_smart_vibration_sensor` converter to use new tuya DP interface

gmbnomis:
Commit 38cae96c76d95a0d974815878a4a4a7b0c601251 (Add TS0601_vibration_sensor
(#3619)) added the "from Zigbee" converter `tuya_smart_vibration_sensor`.
However, it uses the old lib/tuya.js interface which does not work anymore.

Adapt it to use the new interface (including a for loop) and add a couple of
tests (using the DP values in the logs from
https://github.com/Koenkk/zigbee2mqtt/issues/10534 as input).

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -7538,19 +7538,25 @@ const converters = {
         cluster: 'manuSpecificTuya',
         type: ['commandGetData', 'commandDataResponse', 'raw'],
         convert: (model, msg, publish, options, meta) => {
-            const dp = msg.data.dp;
-            const value = tuya.getDataValue(msg.data.datatype, msg.data.data);
-            switch (dp) {
-            case tuya.dataPoints.state:
-                return {contact: Boolean(value)};
-            case tuya.dataPoints.thitBatteryPercentage:
-                return {battery: value};
-            case tuya.dataPoints.tuyaVibration:
-                return {vibration: Boolean(value)};
-            default:
-                meta.logger.warn(`zigbee-herdsman-converters:tuya_smart_vibration_sensor: NOT RECOGNIZED ` +
-                    `DP #${dp} with data ${JSON.stringify(msg.data)}`);
+            const result = {};
+            for (const dpValue of msg.data.dpValues) {
+                const value = tuya.getDataValue(dpValue);
+                switch (dpValue.dp) {
+                case tuya.dataPoints.state:
+                    result.contact = Boolean(value);
+                    break;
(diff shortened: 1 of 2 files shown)


gmbnomis:
@flatsiedatsie fyi (when I had a look at #3619 and noticed its problems it was already merged)

flatsiedatsie:
@gmbnomis Ah, sorry about that. I thought I followed the [latest guide](https://github.com/Koenkk/zigbee2mqtt.io/blob/develop/docs/advanced/support-new-devices/02_support_new_tuya_devices.md) that you provided.

What exactly is this for-loop you mention?. Do you have a suggestion of where I can learn how to fix it? Do you have an example of a better converter perhaps?

Koenkk:
thanks again!

gmbnomis:
> @gmbnomis Ah, sorry about that. I thought I followed the [latest guide](https://github.com/Koenkk/zigbee2mqtt.io/blob/develop/docs/advanced/support-new-devices/02_support_new_tuya_devices.md) that you provided.

@flatsiedatsie Yikes, the changes to the converter code where undone when @Koenkk merged the master branch into the development branch. I think this was not intentional. @Koenkk Do you want me to repair this?
 
> What exactly is this for-loop you mention?. Do you have a suggestion of where I can learn how to fix it? Do you have an example of a better converter perhaps?

The current stable version of zibgee2mqtt assumes that only a single data point  can be transmitted in a packet. Last December I got a Tuya device that behaved differently. It sent two values in every message and  the second data point (battery state) was ignored.

That's why we changed the Tuya interface to receive and send multiple DP values per payload. This means that "fromZigbee" converters should use a loop to process all DP values that may be present in a message.

To reduce the risk of breaking things, we opted _not_ to change all existing converters to use such a loop. Instead they use a function `getDataValue`, which will log a warning when there are multiple DPs in a message. (see https://github.com/Koenkk/zigbee-herdsman-converters/blob/bdb25faf328a47c14d4755cdfd0d421f067ca60f/converters/fromZigbee.js#L7568 for an example.

New converters should use a for loop to iterate through all DP values and collect the result values in a loop, like your converter does now.

flatsiedatsie:
Thanks for the explanation and fix, appreciated.

Koenkk:
@gmbnomis whoos, thanks for noticing, fixed by https://github.com/Koenkk/zigbee2mqtt.io/commit/fdfb39ec7654140059795fa6dd0e5662ae3904f4

PR #3640, opened 2022-01-07
Extended support for BTicino L441C/N4411C/NT4411C and Legrand 067797

Yonn-Trimoreau:
Basically a copy/paste of https://github.com/Koenkk/zigbee-herdsman-converters/pull/3021

I think the following issue is also related: https://github.com/Koenkk/zigbee2mqtt/issues/10580

It seems that BTicino L441C/N4411C/NT4411C and Legrand 067797 use the exact same firmware or at least the same Zigbee API (and I suspect Legrand 067771 also, as I also have it at home and I use the same converter without any issue)

PS: a side note to future readers of this PR, you **must** update your switch's firmware using the Home+Control app and the Legrand gateway before associating it to your own Zigbee network (this can take a day or two, you cannot trigger the upgrade manually, and OTA update through zigbee2mqtt is not supported for the moment)

### DIFF
--- a/devices/bticino.js
+++ b/devices/bticino.js
@@ -30,19 +30,22 @@ module.exports = [
         vendor: 'BTicino',
         description: 'Dimmer switch with neutral',
         extend: extend.light_onoff_brightness({noConfigure: true}),
-        exposes: [
-            e.light_brightness(),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #3641, opened 2022-01-08
Adding LWA019

MatthewT1:
New type of bulb, 1100 Lumen with Bluetooth -  just copied an existing entry and confirmed it worked locally then updated this file w/ the same changes


Koenkk:
Thanks!

PR #3646, opened 2022-01-08
TuYa MIR-HE200-TY correction radar_scene

Burki24:
Some corrections for tuya_radar_sensor, could not be adjusted


Koenkk:
Thanks!

PR #3526, opened 2021-12-15
Add Danfoss TRV attribute: Radiator Covered

davidramiro:
As mentioned in Koenkk/zigbee2mqtt#10185 and relating to https://github.com/Koenkk/zigbee-herdsman/pull/476, firmware 1.18 introduces a new boolean at cluster 0x4016 called "Radiator Covered". This would be very helpful for users with external temperatur sensors to access.

Reference: https://assets.danfoss.com/documents/193613/AM375549618098en-000102.pdf

PR #3647, opened 2022-01-08
Use same device config for all Danfoss Ally clones

HenningCash:
#3526 introduced a new attribute for the Danfoss Ally TRV. The same attribute is supported on my Popp eT093WRO TRV.
Because these two models (and Hive's TRV001) are basically all the same only with different housings I added a utility method to extend other devices. This keeps the code DRY but I'm not sure if this has any side effects.


Koenkk:
looks good, thanks!

PR #3649, opened 2022-01-09
Expose setting alarm values on LCZ030/TS0201

HugeMissedSteak:
Allows the user to set the min/max alarm values for the humidity and temperature.
image

Includes a workaround for what I presume will be fixed after [PR 489](https://github.com/Koenkk/zigbee-herdsman/pull/489)


Koenkk:
thanks!

PR #3652, opened 2022-01-09
Add support for local firmware files

grafalex82:
[Previous change](https://github.com/Koenkk/zigbee-herdsman-converters/pull/3495) added possibility to have local OTA index file. This change adds possibility to use local firmware files. Having a web server is no longer a requirement for OTA update of DIY devices.

Firmware URL can have one of these options:
- an URL of the image stored on a web server (keep existing behavior)
- absolute path to an image file
- path relative to a data directory

For enabling the latter option, a new setDataDir() interface function is introduced. In case of Zigbee2MQTT, the data directory is the configuration directory. Zigbee2MQTT calls setDataDir() during startup. Other clients (non-Z2M) will have to call setDataDir() as well to enable non-absolute paths.

Example of a local index file:
```
[
    {
        "fileVersion": 2,
        "manufacturerCode": 4151,
        "imageType": 1,
        "url": "HelloZigbee.ota"
    }
]
```
Note that `hash` field (`sha256` or `sha512`), `fileSize` and `path` fields are not mandatory, and can be omitted in case of local firmware files.  Other fields still required, and have to correspond to the image file contents - `imageType` and `manufacturerCode` participate in searching available images for the particular device, while `fileVerision` is involved in decision making process whether the image is newer than flashed in the device.

### DIFF
--- a/lib/ota/zigbeeOTA.js
+++ b/lib/ota/zigbeeOTA.js
@@ -4,8 +4,10 @@ const common = require('./common');
 const axios = common.getAxios();
 const fs = require('fs');
 const URI = require('uri-js');
+const path = require('path');
 
-let overrideFileName = null;
+let overrideIndexFileName = null;
+let dataDir = null;
 
 /**
  * Helper functions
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

PR #3653, opened 2022-01-09
Add _TZ3210_bfwvfyx1 to TS0505B.

pidpawel:
First of all thanks for an awesome piece of software!

Secondly, this pull request adds `_TZ3210_bfwvfyx1` TuYa light bulb to a proper definition. I've tested it with the following external converter:

```

const definition = {
    fingerprint: [
        { modelID: 'TS0505B', manufacturerName: '_TZ3210_bfwvfyx1' },
    ],
    model: 'TS0505B',
    vendor: 'TuYa',
    description: 'Zigbee RGB+CCT light',

    extend: extend.light_onoff_brightness_colortemp_color({ colorTempRange: [153, 500] }),
    meta: { applyRedFix: true, enhancedHue: false },
};

module.exports = definition;
```

Everything seems to be working fine except the effects but I have no way of validating it with the manufacturer as it's a no-brand item.


Koenkk:
thanks!

PR #3655, opened 2022-01-10
Expose color, hue, and color temperature actions for RGBgenie ZB-5028

visualage:
Looks like the proper commands are defined in the `fromZigbee`, but those actions are not exposed in MQTT.


Koenkk:
Thanks!

PR #3656, opened 2022-01-10
Minor fix to color XY calculation

visualage:
The color XY uint16 representation in zigbee spec is in `s15Fixed16Number` format per ICC spec (https://www.color.org/specification/ICC1v43_2010-12.pdf). It is actually a 16-bit fixed-point number. If we don't consider negative numbers, it is exactly the integer divide by 65536 instead of 65535. Minor, but better to get it according to spec.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1380,8 +1380,8 @@ const converters = {
             const payload = {
                 action: postfixWithEndpointName(`color_move`, msg, model),
                 action_color: {
-                    x: precisionRound(msg.data.colorx / 65535, 3),
-                    y: precisionRound(msg.data.colory / 65535, 3),
+                    x: precisionRound(msg.data.colorx / 65536, 3),
+                    y: precisionRound(msg.data.colory / 65536, 3),
                 },
                 action_transition_time: msg.data.transtime,
             };


Koenkk:
Thanks!

TheJulianJES:
It looks like this change wasn't made in all places. Shouldn't cause any major issues, but it would be nice to have it correct everywhere.

If I'm right, it should also be changed in these places (and likely others):
https://github.com/Koenkk/zigbee-herdsman-converters/blob/aa44881e0ee01bcc8724f0f92c69ddf03899e518/converters/fromZigbee.js#L588-L591
https://github.com/Koenkk/zigbee-herdsman-converters/blob/aa44881e0ee01bcc8724f0f92c69ddf03899e518/converters/toZigbee.js#L1041-L1042
https://github.com/Koenkk/zigbee-herdsman-converters/blob/aa44881e0ee01bcc8724f0f92c69ddf03899e518/converters/toZigbee.js#L2023
https://github.com/Koenkk/zigbee-herdsman-converters/blob/aa44881e0ee01bcc8724f0f92c69ddf03899e518/converters/toZigbee.js#L5549-L5564
https://github.com/Koenkk/zigbee-herdsman-converters/blob/aa44881e0ee01bcc8724f0f92c69ddf03899e518/converters/toZigbee.js#L5590-L5591

PR #3657, opened 2022-01-10
Update TS0601_thermostat calibration according to documentation

brunoleon:
Having tried to make use of @KartoffelToby changes at https://github.com/Koenkk/zigbee-herdsman-converters/pull/3491 , I never managed to get a 0.1 step to work.

The documentation indicates +/- 9 degrees with a 1°C step only.

I'm not sure this is valid for all clones so I updated only for those I have
IMG_20220110_121057
.


Koenkk:
Thanks!

PR #3662, opened 2022-01-11
Add 'force' key to the OTA index files to force selection of a firmware image

grafalex82:
One of highly desired features is to force Zigbee2MQTT to use a specific version of a firmware, instead of one on the main OTA index. Unfortunately there is no way to downgrade the firmware, even with hacking the index or overriding information there. There is a special code in that does not allow flashing a firmware with a version lower than current.

With this change a new `force` key is added to work this around. If a firmware marked with `force` key it is considered as newer, regardless to the version burned in the file.

### DIFF
--- a/lib/ota/common.js
+++ b/lib/ota/common.js
@@ -265,6 +265,11 @@ async function isNewImageAvailable(current, logger, device, getImageMeta) {
     const meta = await getImageMeta(current, logger, device);
     const [currentS, metaS] = [JSON.stringify(current), JSON.stringify(meta)];
     logger.debug(`Is new image available for '${device.ieeeAddr}', current '${currentS}', latest meta '${metaS}'`);
+
+    if(meta.force) {
+        return -1; // Negative number means the new firmware is 'newer' than current one
+    }
+
     return Math.sign(current.fileVersion - meta.fileVersion);
 }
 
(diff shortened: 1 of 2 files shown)


sjorge:
The device itself will still reject this right? e.g. the ikea on/off, cover, valve remotes that got upgrade to a version without group support can't be downgraded as the device rejects the old firmware.

grafalex82:
I can't really answer this question, as different vendors may have different implementation. AFAIK Zigbee specification does not obligate devices to reject lower versions of the firmware.

Overall it would be nice if someone could test this. I tested only on my DIY device, and Zigbee2MQTT side seems working.

sjorge:
I know from messing around with zipyp that Hue and IKEA devices at least seem to not allow downgrades. Most of my other devices have no OTA available though.

sjorge:
Actually looks like ZCL even explicitly allows downgrades:
```
16016 It is the responsibility of the server to indicate to the clients when update images are available. The client MAY be
16017 upgraded or downgraded.
```

Koenkk:
@grafalex82 thanks again!

@sjorge thanks for checking

MattWestb:
All my IKEA light was possible doing downgrading with ZHA and deCONZ then i was testing before the last large update last months.
IKEA controllers is blocked in the firmware for downgrade and some bad test firmware is not accepting new updates and need SWD flashing for getting then "on track"

Also possible looking in the Zigbee certification docks is one point if downgrading is blocked or not (all new IKEA controllers i have looking at is blocked) but not all devices is certificated but its only trying if its working or not.

The "old" IKEA light firmware was recreating there default scenes then doing downgrade and then upgrading them and was working OK with there remotes. I have not have time testing all new firmware but its on the to do list.

PR #3664, opened 2022-01-12
Add colour temp range to Philips Hue white ambiance E14 (with Bluetooth) [9290022944]

JeffersonBledsoe:
`Read result of 'lightingColorCtrl': {"colorTempPhysicalMin":153,"colorTempPhysicalMax":454}`


Koenkk:
thanks!

sjorge:
Thanks for this! Interesting that this is also I think the first one that does not have a max of 500!

PR #3665, opened 2022-01-12
Update danalock.js

XLDom:
When Danapad installed (https://www.danalock.fr/danapad), we can create/update/remove the pincode


Koenkk:
thanks!

smulle48:
@Koenkk can't you revert this PR?

I have talked to Danalock, and they told me that this is not supported by the lock.

When using this "function" the network freeze until the message timeout.

Koenkk:
@smulle48 done

Issue #3591, opened 2021-12-29
No converter available for '6735/6736/6737' with key 'rm01_row_2'


Newan:
I use the Busch Jäger, and if i send true|false to the second row i get:

Errormessage:
`2021-10-14 14:53:11.048 | error | No converter available for '6735/6736/6737' with key 'rm01_row_2'`

Newan:
Debug log from ioBroker:

```
2021-12-29 22:44:50.687 - debug: zigbee.0 (25382) Publish {"brightness_relay":221} to "d85def11a10046b1"
--
2021-12-29 22:44:50.693 - debug: zigbee.0 (25382) Publish {"brightness_relay":254} to "d85def11a10049ce"
2021-12-29 22:44:50.876 - warn: admin.0 (3203) Read-only state "zigbee.0.d85def11a10046a8.rm01_row_2" has been written without ack-flag with value "false"
2021-12-29 22:44:50.882 - debug: zigbee.0 (25382) User stateChange zigbee.0.d85def11a10046a8.rm01_row_2 {"val":false,"ack":false,"ts":1640814290879,"q":0,"from":"system.adapter.admin.0","user":"system.user.admin","lc":1640814290879}
2021-12-29 22:44:50.887 - debug: zigbee.0 (25382) Change state 'rm01_row_2' at device 0xd85def11a10046a8 type '6735/6736/6737'
[...]

PR #3666, opened 2022-01-12
fix for #3591 and RM01 from zigbee fix

Newan:


converters/toZigbee.js
```diff
@@ -3331,6 +3331,12 @@ const converters = {
             }
         },
     },
+    RM01_row_button: {
+        key: ['rm01_row_2', 'rm01_row_3', 'rm01_row_4'],
+        convertSet: async (entity, key, value, meta) => {
+            return;
```
Koenkk, line 3337: This converter is empty, why is it needed?

Newan: i dont know a another solution.
My Idea, The fromZigbee function are:
```

RM01_on_click: {
        cluster: 'genOnOff',
        type: 'commandOn',
        options: [exposes.options.legacy()],
            if (isLegacyEnabled(options)) {
                const button = getKey(model.endpoint(msg.device), msg.endpoint.ID);
                return {action: `${button}_on`};
            } else {
                return fromZigbeeConverters.command_on.convert(model, msg, publish, options, meta);
            }
        },
    },


```
In legacy, the command send an 'rm01_row_2_on' or 'rm01_row_2_off'. I i send to zigbee (from ioBroker or somethink like that) 'rm01_row_2_on' or 'rm01_row_2_off' it come to an error like:

` '2021-10-14 14:53:11.048 | error | No converter available for '6735/6736/6737' with key 'rm01_row_2'
`
But the device work so, that always push the button it send always the on oder off command. The device don´t store the status from to the second row

PR #3667, opened 2022-01-12
Add support for Aurora AU-A1GSZ9RGBWE light bulbs with model code RGBBulb51AU

jackmurray:
I just received a bulb that lists RGBBulb51AU as the zigbee model number but says AU-A1GSZ9RGBWE on the box (last char E == E27 fitting, there's a B model too). It supports all the same features as the existing models that are supported, so this PR just adds the new model code to the existing definition.


Koenkk:
thanks!

PR #3668, opened 2022-01-12
_TZ3210_trm3l2aw added on TS0503B

markieo1:
I've added _TZ3210_trm3l2aw to the TS0503B since it's the same device.


Koenkk:
thanks!

PR #3670, opened 2022-01-12
Do not require fileVersion, manufacturerID, and imageType fields for local OTA images

grafalex82:
The only necessary field for local OTA images is a file name - everything else can be easily retrieved from the firmware image. No need to obligate users to fill these fields in the local OTA index. This change removes requirement for filling `fileVersion`, `manufacturerID`, and `imageType` fields.


grafalex82:
Supporting documentation https://github.com/Koenkk/zigbee2mqtt.io/pull/1136

Koenkk:
thanks!

PR #3676, opened 2022-01-14
Change Zemismart ZM_AM02 USB Zigbee dongle

Burki24:
Changed Zemismart ZM_AM02 from tuya Tubular Motor to Zemismart Roller Shade Converter. It's a special device to control  shadowing devices via RF.

https://zh_tw.zemismart.com/products/zmam02

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -7956,6 +7956,78 @@ const converters = {
             return {action};
         },
     },
+    ZMAM02: {
+        cluster: 'manuSpecificTuya',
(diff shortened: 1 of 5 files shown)


Burki24:
@Koenkk I think I need your help. I installed these modifications at home and all went fine. Making a PR, getting errors, I can't find. So it would be very helpfull, if you can give me a small hint, what`s wrong.

Many thanks in advance.

Burki24:
I cancel this. Coming back after installing all these test (jest etc.) on my windos VSC machine.

PR #3686, opened 2022-01-16
Add new device like MIR-HE200-TY with fall function

prokudin07:
MIR-HE200-TY has 2 modification:
1. [MIR-HE200-TY](https://www.zigbee2mqtt.io/devices/MIR-HE200-TY.html) without fall function 
    with  **manufacturerName: '_TZE200_vrfecyku'**
fall-no

2.  MIR-HE200-TY with fall function  **with manufacturerName: '_TZE200_lu01t0zl'**
fall-yes

Second modification has new exposes.
I added in z2m and result below

Снимок экрана 2022-01-16 в 02 50 14


Koenkk:
thanks!

PR #3696, opened 2022-01-18
add support for tuya XFY-CGQ-ZIGB iIlluminance sensor

regevbr:
I tested it locally and it works great!
This is the zigbee version of https://templates.blakadder.com/lightsensor_XFY-CGQ-GZ.html


Koenkk:
thanks!

PR #3698, opened 2022-01-18
Add 3 new zigbeeModel values for Hue white ambiance GU10 with Bluetooth

mdehaas:
First time doing this. I added these at the end of the file. These where part of the Milliskin White Ambiance - White multipack.


Koenkk:
bedankt 😄

PR #3702, opened 2022-01-19
Support 1746630V7

alp:
Added support for Philips Amarant linear outdoor light (1746630V7), see [its product page](https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-amarant-linear-outdoor-light/1746630V7)
Verified `colorTempRange` by reading attributes through the Dev console
Finally, tested that the device responds properly to all commands.


Koenkk:
thanks!

PR #3707, opened 2022-01-20
add support for Micro Matic ZB250 / SZ1000

geirra:
First try to add support for a device in zigbee2mqtt, so there might be errors.
The code is based on other devices.

### DIFF
--- a/devices/micromatic.js
+++ b/devices/micromatic.js
@@ -1,7 +1,7 @@
-const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
-const exposes = require('zigbee-herdsman-converters/lib/exposes');
-const reporting = require('zigbee-herdsman-converters/lib/reporting');
-const extend = require('zigbee-herdsman-converters/lib/extend');
+const fz = require('../converters/fromZigbee');
+const exposes = require('../lib/exposes');
+const reporting = require('../lib/reporting');
+const extend = require('../lib/extend');
 const e = exposes.presets;
 const ea = exposes.access;
 
--- a/devices/micromatic.js
+++ b/devices/micromatic.js
@@ -20,10 +20,10 @@ module.exports = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl', 'haElectricalMeasurement', 'seMetering']);
             await reporting.brightness(endpoint);
             await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
-            await reporting.rmsVoltage(endpoint);
-            await reporting.rmsCurrent(endpoint, {change: 50});
-            await reporting.activePower(endpoint);
-            await reporting.currentSummDelivered(endpoint);
+            await reporting.rmsVoltage(endpoint, {min: 10, change: 20}); // Voltage - Min change of 2v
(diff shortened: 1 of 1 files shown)


geirra:
Attached screenshot of the exposed dimmer.

Screenshot 2022-01-20 at 16 04 59

Koenkk:
> Is this based on SummDelivered? If so there is something strange going on, since this is reporting
While the reported value in the ui is listed as 1,2kwh

This looks good to me, the representation might looks strong but that is because this is an unsigned 48 bit integer.

Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1151

geirra:
@Koenkk thanks for the confirmation, i have tested it and with load it reports the correct numbers.

I fear that the value is going to spam the zigbee network tough, tried setting a min change to 0,001 or 0,01, but it doesn´t seem to have any affect.
Do you have any suggestions?

```
debug 2022-01-20 18:49:26: Received Zigbee message from 'Kjellerstue', type 'attributeReport', cluster 'seMetering', data '{"currentSummDelivered":[0,4554409]}' from endpoint 1 with groupID 0
debug 2022-01-20 18:49:31: Received Zigbee message from 'Kjellerstue', type 'attributeReport', cluster 'seMetering', data '{"currentSummDelivered":[0,4555014]}' from endpoint 1 with groupID 0
debug 2022-01-20 18:49:36: Received Zigbee message from 'Kjellerstue', type 'attributeReport', cluster 'seMetering', data '{"currentSummDelivered":[0,4555740]}' from endpoint 1 with groupID 0
debug 2022-01-20 18:49:42: Received Zigbee message from 'Kjellerstue', type 'attributeReport', cluster 'seMetering', data '{"currentSummDelivered":[0,4556345]}' from endpoint 1 with groupID 0
debug 2022-01-20 18:49:47: Received Zigbee message from 'Kjellerstue', type 'attributeReport', cluster 'seMetering', data '{"currentSummDelivered":[0,4556950]}' from endpoint 1 with groupID 0
debug 2022-01-20 18:49:52: Received Zigbee message from 'Kjellerstue', type 'attributeReport', cluster 'seMetering', data '{"currentSummDelivered":[0,4557676]}' from endpoint 1 with groupID 0
debug 2022-01-20 18:49:58: Received Zigbee message from 'Kjellerstue', type 'attributeReport', cluster 'seMetering', data '{"currentSummDelivered":[0,4558281]}' from endpoint 1 with groupID 0
```

Koenkk:
try with higher values, e.g. 10 or 100

geirra:
> try with higher values, e.g. 10 or 100

@Koenkk This what I don't find any information on, is the min change based on  the 48bit integer or the converted sum?

I have tried 10, 100, 1000 but it still hits the min interval. 

Noticed that there is one a few develco devices that does some tuning on this also. 
```
 await reporting.currentSummDelivered(endpoint, {change: [0, 20]}); // Limit reports to once every 5m, or 0.02kWh
```

But that doesn't seem to slow down the messages either in this case.

Koenkk:
Then instead of `change: [0, 20]` you can use `min` which is the minimum reporting interval in seconds

geirra:
@Koenkk This should be tuned pretty good now.
Added comments as to why the values are chosen.

So think it can be merged.

Koenkk:
thanks!

PR #3708, opened 2022-01-20
add colorhs to gardenpole definition

seidler2547:
The OSRAM Gardenpole Mini RGBW OSRAM (model AC0363900NJ) is visually identical to the 4058075208353, but only the OSRAM definition includes HSV color mode.

This commit brings them both to the same feature set. Without it, it's not possible to create all colors from Home Assistant, for example, a pure blue is not possible, it's always tinted purple.

I have both devices next to each other and the difference is very visible.

### DIFF
--- a/devices/ledvance.js
+++ b/devices/ledvance.js
@@ -1,6 +1,8 @@
 const ota = require('../lib/ota');
 const extend = require('../lib/extend');
+const exposes = require('../lib/exposes');
 const reporting = require('../lib/reporting');
+const e = exposes.presets;
 
 module.exports = [
     {
@@ -109,6 +111,7 @@ module.exports = [
         vendor: 'LEDVANCE',
         description: 'SMART+ gardenpole multicolour',
         extend: extend.ledvance.light_onoff_brightness_colortemp_color({colorTempRange: [153, 526]}),
+        exposes: [e.light_brightness_colortemp_colorhs([153, 526]), e.effect()],
         ota: ota.ledvance,
     },
     {


devices/ledvance.js
```diff
@@ -109,6 +111,7 @@ module.exports = [
         vendor: 'LEDVANCE',
         description: 'SMART+ gardenpole multicolour',
         extend: extend.ledvance.light_onoff_brightness_colortemp_color({colorTempRange: [153, 526]}),
+        exposes: [e.light_brightness_colortemp_colorhs([153, 526]), e.effect()],
```
Koenkk, line 114: Changing the above line to `extend: extend.ledvance.light_onoff_brightness_colortemp_color({colorTempRange: [153, 526], supportsHS: true}),` should have the same effect, can you check?

seidler2547: Unfortunately, no. HA still reports supported color modes as "xy" with the change you suggested. I even tried `supportsHS: true, supportsXY: false` and it wouldn't accept it. Only when I make it like in my PR, thus identical to the OSRAM variant, it works. The "state" tab in Z2M also reports the color mode in line with what Home Assistant reports.

Koenkk: That is because z2m prefers xy over hs in the discovery, this change actually disables xy and therefore hs comes up (which is not desired): https://github.com/Koenkk/zigbee2mqtt/blob/1a0a9a6eabf58855d3a6feda60bb79a1c34d17f8/lib/extension/homeassistant.ts#L147

seidler2547: Why is it not desired? What I observe is that with XY color, I cannot make the gardenpoles reproduce all possible colors, for example it's not possible to create a pure blue. With HS however, all colors are reproduced perfectly, as I would expect from RGB lights.

Koenkk: Because HA doesn't support this: https://www.home-assistant.io/integrations/light.mqtt/

> RGB, XY and HSV can not be used at the same time in state_topic messages. Make sure that only one of the color models is in the “color” section of the state MQTT payload.

If you prefer HS over XY you can modify the discovery topic by adding the following to your `configuration.yaml`:

```yaml
devices:
  "0x12345678":
    friendly_name: my_bulb
    homeassistant:
      light:
        supported_color_modes: ['color_temp', 'hs']
```

seidler2547: Okay, I can do that. But everyone having those lights will need to do that, because XY color mode doesn't produce the desired colors. Is that the best way then?

To be clear, when I run Z2M without this patch, I see HS and XY color controls in the Z2M web interface. When I tap on one of the colors in the HS controls, they come up perfectly. When I tap on one of the colors in the XY controls, they are not at all true to the desired color.

I also found https://github.com/Koenkk/zigbee2mqtt/issues/6402 which says that OSRAM leds have broken XY colors in the firmware. Since OSRAM and LEDVANCE are basically the same company with different names, I would guess their firmware is fairly similar, if not identical except for the IDs.

Koenkk: Good point, I've added a preferredColorMode option now. 

If you rebase this pr (it would be easiest to just recreate it from master again) and use the latest z2m dev (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) you can enable this with:

```js
extend: extend.ledvance.light_onoff_brightness_colortemp_color({colorTempRange: [153, 526], preferredColorMode: 'hs', supportsHS: true}),
```

sjorge: @Koenkk from my experience this is also true for Innr bulbs, to get deep red/blues I need to use hs to set them with xy they are either pinkish or purpleish.

I'm guessing adding this for the Innr bulbs (at least the ones I can test, but AFAIK all of them seem to be the same) would be OK?

Koenkk: @sjorge yes, however I'm wondering if this results in a breaking change.

@seidler2547 could you try to make a automation in HA which sets the color (in x/y), stop z2m, set preferHS: true, start z2m and trigger the automation?

sjorge: I guessing people who have a set of there preferred colors they are using HA will now get different colors that are closer to what they are actually setting.

So it defiantly going to be a noticeable change.

Koenkk: @sjorge not only noticeable, since the supported color modes will only include hs and colortemp, sending xy here might yield errors at the ha side

sjorge: > @sjorge not only noticeable, since the supported color modes will only include hs and colortemp, sending xy here might yield errors at the ha side

Would HA not see that device now wants HS instead of XY and send the correct one?

Koenkk: Yes, but existing automations could still send x/y values

seidler2547: > @sjorge yes, however I'm wondering if this results in a breaking change.
> 
> @seidler2547 could you try to make a automation in HA which sets the color (in x/y), stop z2m, set preferHS: true, start z2m and trigger the automation?

I will try that. I think it'll work fine, HA should convert the colour space.

seidler2547: Yes, it works fine. Upon further reflection, it was also always working fine when I set HS values in my existing automation but the light announced itself as XY, so HA just always does the conversion either way.

PR #3713, opened 2022-01-21
Configure Xiaomi SSM-U01 as router

McGiverGim:
It appears as EndDevice. With this change it appears as Router, as expected.

The device seems to work as router without problem:
image


Koenkk:
thanks!

PR #3716, opened 2022-01-21
Added support for Philips-branded Disney StoryLight (LLC013) - a re-badged LivingColors light

DJBenson:
Added LLC013 to philips.js - this is a variant of the LivingColors lights already supported but sold branded as the Disney StoryLight.


Koenkk:
thanks!

PR #3718, opened 2022-01-21
Adds support for TuYa clone TS0502B with manufacturer _TZ3210_rm0hthdo

nicoandra:
Bought 2 sets of lights from AliExpress, none of them are branded as TuYa but they work perfectly with the TuYa configuration as is; except for the manufacturer code.

This PR adds the manufacturer code to the list.


Koenkk:
thanks!

PR #3720, opened 2022-01-22
Add new model to 'Philips Hue white ambiance Adore 2 spots'

MarBra:
Model number tested and also checked that the color temprange is still the same.


Koenkk:
thanks!

PR #3721, opened 2022-01-22
updated e.action() for Matsee TuYa Garage door opener

octa22:
Now it defines the only value it can have ('trigger')

Signed-off-by: Ondrej Pecta <opecta@gmail.com>


Koenkk:
thanks!

PR #3724, opened 2022-01-22
Add Philips Hue Adore Recessed Bathroom Spotlight (3417511P9)

Broekman:
Also see: https://github.com/Koenkk/zigbee2mqtt.io/pull/1159


Broekman:
So a one and three packs box of these things ship with different IDs, confirmed on my z2mqtt (i.e. '3417611P6' = three, '3417511P9'=one). Is the model OK like this? Also used that model on the zigbee2mqtt.io PR for the image/markdown.

Koenkk:
looks good, thanks!

PR #3732, opened 2022-01-24
Adding Tuya Highbay Light TS0501A _TZ3000_yeg1e5eh

Jimmypl007:
I Apologise if i am doing anything wrong, I am really new to this, after fumbling through the tutorials I managed to get these lights supported my side.
and now i am struggling to figure out what i am doing here on GitHub, any help appreciated.

I think my next step is to upload a photo somehow over at the  zigbee2mqtt.io GitHub, and probably fail at adding an .md file..

The High-bay light is fairly simple it has some "scenes" option through tuya's app that i would think makes sense that they aren't available through the extension.
also (I guess) the extend exposes some "effect" option that no matter what you choose just does the same sort of "fading blink" that is the same as when the lights are in pairing mode.

I also apologise if this is not the area to say any of this..

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -204,7 +204,7 @@ module.exports = [
     {
         fingerprint: [{modelID: 'TS0501A', manufacturerName: '_TZ3000_yeg1e5eh'}],
         model: 'TS0501A',
-        description: 'Zigbee Highbay Light',
+        description: 'Zigbee light',
         vendor: 'TuYa',
         extend: extend.light_onoff_brightness(),
     },
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -202,6 +202,13 @@ module.exports = [
         meta: {applyRedFix: true},
     },
     {
+        fingerprint: [{modelID: 'TS0501A', manufacturerName: '_TZ3000_yeg1e5eh'}],
+        model: 'TS0501A',
+        description: 'Zigbee Highbay Light',
+        vendor: 'TuYa',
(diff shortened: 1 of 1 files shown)


Jimmypl007:
I also have no idea how i ended up in Patch:5 instead of master.. or if that was meant to happen or what that even really means..
just noticed its different from when i uploaded an image to zigbee2mqtt.io GitHub thing..

Koenkk:
Do you have a link to this device? (aliexpress?)

Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1165

Jimmypl007:
> Do you have a link to this device? (aliexpress?)
> 
> Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1165

I only have the reseller
https://greenhse.com/products/lighting-perth/industrial-lighting-perth/hb200p-p3.html

And the info provided is limited.
I was unable to find where they got them from.

Koenkk:
thanks!

Jimmypl007:
> Do you have a link to this device? (aliexpress?)
> 
> Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1165

Hi Koenk, just noticed the lights our now supported, but have the wrong image. It looks like they are using TS0501B image.

I also noticed the description says theyre zigbee lights rather than zigbee highbay lights although that is less important i guess..

Koenkk:
Lets fix that in https://github.com/Koenkk/zigbee2mqtt.io/pull/1165

PR #3738, opened 2022-01-25
Remove useless codes for some Aqara devices.

z-x-f:


Koenkk:
Is this not needed anymore due to the new firmware?

z-x-f:
Maybe because of the new firmware, these devices don't support `mode` anymore, so I re-tested and it does work. 
I don't have any devices with the old firmware anymore, so I can't be 100% sure it's because of the new firmware.
If there is a chance I will come across a device with an older firmware, I will retest it.

Koenkk:
thanks!

PR #3739, opened 2022-01-25
Fix UNSUPPORTED_ATTRIBUTE error message

r3pek:
Since the multiplier and dividor atributes are optional, this SmartPlugs don't use them and we need to not rely on them.

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/9561


Koenkk:
thanks!

PR #3741, opened 2022-01-25
 Add support for TS011F (_TZ3000_gjnozsaz) smart plug

ddequidt:
Add _TZ3000_gjnozsaz to TS011F plugs.
It works with polling workaround (TS011F_plug_3).


ddequidt:
Sorry; it is actually already supported. I mess up with versions.

PR #3742, opened 2022-01-25
Split off ZM-35H-Q from TS0202 to support additional functions

gmbnomis:
The PIR sensor supports setting/getting sensitivity and keep time (verified
using TuYa cloud and packet traces)

As the device is only able to receive when active (i.e.  typically when
motion is detected), use `{sendWhen: 'active'}` when sending.


gmbnomis:
I got a flavor of the TS0202 PIR sensor (`_TZ3000_msl6wxk9`) which supports additional attributes (sensitivity and keep_time). Judging from https://github.com/Koenkk/zigbee2mqtt/issues/9757, the device `_TZ3000_mcxw5ehu` behaves the same.

I was not sure whether to extract these as a new device (as in this PR), or whether to add the functionality to the existing TS0202 support (not all of these may have these functions). Of course, I am happy to change this PR to implement the latter if that's better.

Koenkk:
Looks good to me, thanks!

PR #3745, opened 2022-01-26
Added new model Tuya Air quality sensor TS0601 _TZE200_dwcarsat

1stsetup:


Koenkk:
thanks!

createcandle:
Wow, perfect timing. Just received this device and wanted to add support for it. You beat me to it :-D

createcandle:
I've been testing this device, and most sensors are working, but I also noticed some strange data. Perhaps it's a fluke of my device, but just in case: 

- My CO2 values jump up to 48100. Other values it shows are between 0 to 9 ppm. Normal outside air has about around 400 PPM. Indoor air normally varies between 400 and 700PPM. My other two CO2 sensors are showing more realistic values.

Screenshot 2022-02-02 at 21 05 10

- I also noticed one sudden spike across most outputs. Hopefully a fluke. One example:
Screenshot 2022-02-02 at 21 07 47

- The PM25 values remain 0, except during that one spike, where it reached 16ug. For comparison, my other two sensors indicate 1.4ug (MySensors DIY device) and 6ug (IKEA Air purifier).

Perhaps the device is still calibrating.


Something else: it seems to send data at a very high rate, around every 3-5 seconds, sometimes reaching once per second. 

This is not meant as criticism, just as feedback: this high rate might have some downsides.
- It's creating a huge amount of data points in my logs. This can lead to unnecessary wear and tear of my Raspberry Pi's SD card.
- It's more velocity than a smart home requires? What situation would benefit from updating CO2 every 2 seconds? 

- The high data rate is not "realistic" because the things that are being measured, such as CO2 levels in a room, change slowly.
- It also implies a level of accuracy that the hardware can't deliver. Measurements from these super cheap sensors aren't accurate, so a high datarate isn't increasing accuracy, it's just sending over sensor noise. 

Screenshot 2022-02-02 at 21 45 32

To "solve" this, I guess the smart home controller could do the job of calculating averages over time to smooth out the data and limit volume. But I just wanted to discuss the idea to do that "up stream" in the converter, with some sensible defaults, instead of having to implement it in each controller?

Some more arguments in favour of lowering the velocity:
- The high speed can be dangerous. Creating automations based on rapidly changing temperature could damage some devices. For example, imagine a gas heater that takes about 2 seconds to switch on (servo motor opens gas valve). If you create an automation where it switches on when the temperature drops below 19 degrees, and the input temperature then rapidly jumps back and forth between 18.9 and 19.1 a few times, this could damage such heaters. I understand that one easy solution here is no not rely on this sensor in such an automation, but I thought I'd share this risk, as not everyone might comprehend the risk beforehand.
- Knowing a bit about these sensors, I worry that a high poll rate could have negative effects on accuracy and lifespan. For example, from my work with Plantower and the SDS011 sensors, I know these sensor have elements that need to heat up, have some air flow over them a bit, and calculate an average. My [DIY devices](https://www.candlesmarthome.com/dust) measure fine dust once per minute, and this is partially done to increase the lifespan of the sensor. Perhaps the issue with the fine dust and CO2 sensor are caused by the high speed?
- I suspect this isn't a big issue, but it generates a lot of network traffic when compared with how often my other Zigbee devices communicate.

createcandle:
I opened up the device to have a look inside.

tuya_6-in-1-air-quality-sensor

TLDR: Don't expect much precision from this device.

This is my guesswork:

- The dust sensor seems to be a version of the Sharp GP2Y1010AU (without any branding, so maybe a copy). A classic, very old optical sensor. By now probably very cheap. Totally imprecise though. This sensor is designed to output to, say, 5 levels of dustyness. There is also no fan blowing air across it, it's all done with passive air flow. It's.. bad (but for the 25 euro price it makes sense). This type of sensor will get clogged with dust over time, leading to even wilder results.
- The VOC/Formaldehyde (and/or CO2..) sensor seems to be very old-school too, one of the [MQ-style models](https://www.mysensors.org/build/gas). It has 4 pins, so it seems like a 2-in-1 model? (GND, VCC, analog output 1, analog output 2). These have an expected lifespan of about 2 years, and are not very precise. But for the price of this device.. an understandable choice.
- Some good news: the temperature+humidty sensor looks more modern and is well placed, away from the other components. They even cut some slots in the PCB for extra airflow. This is good. I can't quite make out the part number. A12151, but I can't find any information on it.
- I don't immediately see a separate CO2 sensor. Not an infrared model, that's for sure. It might be 'virtual', calculated from the value of the VOC/Formaldehyde sensor.
- There seems to be a serial connection location on the board, likely used to flash the firmware. The chip is an STC model.
- The layout seems nice and clean.
- There is an unpopulated capacitor location? Maybe they left it out as a cost saving measure on later builds. This could be bad news for voltage stability though, and thus for sensor accuracy.

Hardware revision 1.0.5

1stsetup:
We can also see high reporting rates. Multiple times a second. I will ask my son, as it is his device, to look into the graphs and values. And I will let you know.

1stsetup:
I can confirm the same bouncy values for the different sensors. The only stable ones are the temprature and the humidity. 
Poor quality hardware.

createcandle:
Indeed.

Although at this point I'm assuming that the Tuya app will show more likely values for Co2 and fune dust? The current output seems to be in error.

Are converters able to / allowed to store values for a while? If so, then some smoothing could be implemented. Mayve something simple like:

- Fill a buffer with 10 values, then calculate the average
- Round that down to some realistic level (humidity could be INT, temperature could have one decimal, etc)
- If the value is not the same as the one that was previously sent onwards, then 'send' it.

Tscheffson:
I can also confirm the bouncy values especially for the CO2 Sensor (only jumping from 30-40 to 4600ppm). The 25pm is not recieving any values yet. Did it took some time for you to recieve values or could mine be already broken? I guess i have to look for a better Sensor. Any recommendation?

gavnoff:
hello i have the same problems with this sensor,
data from it is not realistic. At the same time, everything is correct in the original Tuya Smart application. Please let me know how to fix this.

e1z0:
any news?

Turton:
Hi, I also have the same issue @gavnoff.  The sensor values look like they are incorrectly being matched to the property.  The values out side the specification limits and incorrect units being displayed.

image

PR #3755, opened 2022-01-26
added Lonsonho TS110E Zigbee Smart Dimmer Modules, 1 & 2, Gang With N…

kiat-huang:
…eutral

Tested with bot 1 Gang and 2 Gang devices - 
https://lonsonho.com/products/lonsonho-tuya-zigbee-smart-dimmer-switch-module-with-neutral-1-2-gang-wireless-remote-control-compatible-alexa-google-home-1?variant=41250202517662 - when following these instructions -  https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_3-adding-converter-s-for-your-device


tvdbroeck:
I added a Lonsonho/Girier `TS110E` device to my network. I have the 1 gang version (manufacturer` _TZ3210_zxbtub8r`). I noticed that the convertor for the 1 gang version has a lot less features than the convertor for the 2 gang version. Even though these features are also supported by the 1 gang version. Specifically I was missing the `tuya_switch_power_outage_memory `and the `TS110E_switch_type` in the 1 gang version. The switch type was very important, because without it, the manual switch behaviour was unusable.
I made the following convertor, and this seems to work fine in my installation:

```
const definition = {
    fingerprint: [{modelID: 'TS110E', manufacturerName: '_TZ3210_zxbtub8r'}, {modelID: 'TS110E', manufacturerName: '_TZ3210_ngqk6jia'}],
    model: 'TS110E_1gang',
    vendor: 'Lonsonho',
    description: 'Zigbee smart dimmer module 1 gang with neutral',
    fromZigbee: extend.light_onoff_brightness({disablePowerOnBehavior: true}).fromZigbee.concat([
        fz.tuya_switch_power_outage_memory, fzLocal.TS110E_switch_type]),
    toZigbee: extend.light_onoff_brightness({disablePowerOnBehavior: true}).toZigbee.concat([
        tz.tuya_switch_power_outage_memory, tzLocal.TS110E_switch_type]),
    exposes: [
        e.light_brightness(),
        exposes.presets.switch_type_2(),
        exposes.enum('power_outage_memory', ea.ALL, ['on', 'off', 'restore']).withDescription('Recover state after power outage')
    ]
};
```
Is it possible to change this convertor on github, so other users can benefit as well?

Two additional remarks:

- the brand of my '1 gang dimmer' is Girier (https://nl.aliexpress.com/item/1005003693644741.html), but in the supported devices page (https://www.zigbee2mqtt.io/devices/TS110E_1gang.html), only the Lonsonho brand is mentioned. Could you add the Girier brand, so other users can more easily find it back?
- the 1 gang dimmer also seems to suffer from this bug: https://github.com/Koenkk/zigbee2mqtt/issues/12774 
Creating a dummy group with only this device, seems to solve the problem. Not sure what causes this, but could be useful to mention on the documentation page.

Thanks for all the hard work!

PR #3757, opened 2022-01-28
Better ZNJLBL01LM Support

invis-z:
Added battery info ( Koenkk/zigbee2mqtt#9447 ), fixed stop function and running state ( Koenkk/zigbee2mqtt#10971 ).

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5273,6 +5273,8 @@ const converters = {
                             payload.state_center = value === 1 ? 'ON' : 'OFF';
                         } else if (['RTCGQ12LM'].includes(model.model)) {
                             payload.illuminance = calibrateAndPrecisionRoundOptions(value, options, 'illuminance');
+                        } else if (['ZNJLBL01LM'].includes(model.model)) {
+                            payload.battery = value;
                         }
                     } else if (index ===102 ) {
                         if (['QBKG25LM', 'QBKG34LM'].includes(model.model)) {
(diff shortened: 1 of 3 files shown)


devices/xiaomi.js
```diff
@@ -1241,9 +1241,14 @@ module.exports = [
         model: 'ZNJLBL01LM',
         description: 'Aqara roller shade companion E1',
         vendor: 'Xiaomi',
-        fromZigbee: [fz.xiaomi_curtain_position, fz.battery, fz.cover_position_tilt, fz.ignore_basic_report, fz.xiaomi_curtain_options],
-        toZigbee: [tz.xiaomi_curtain_position_state, tz.xiaomi_curtain_options],
+        fromZigbee: [fz.xiaomi_curtain_acn002_position, fz.xiaomi_curtain_acn002_status, fz.cover_position_tilt, fz.ignore_basic_report,
+            fz.aqara_opple],
+        toZigbee: [tz.xiaomi_curtain_position_state],
         exposes: [e.cover_position().setAccess('state', ea.ALL), e.battery()],
```
Koenkk, line 1247: Please add `motor_state` and `running` to the exposes here (this will make sure it shows up in the z2m frontend + Home Assistant)

converters/fromZigbee.js
```diff
@@ -5805,6 +5805,160 @@ const converters = {
             }
         },
     },
+    xiaomi_curtain_acn002_aqara_opple: {
```
Koenkk, line 5808: This converter adds a lot of code duplication, can it be merged with: https://github.com/Koenkk/zigbee-herdsman-converters/blob/72b18217beb0e187061e692be7d6dd634c5493ec/converters/fromZigbee.js#L5139 ?

invis-z: The linked one is having some options that doesn't apply to the target device. I don't know if it is possible to set options for convertors based on model, so I copied it into a separate one.

invis-z: If it is possible to set the options based on the device model, it will be very helpful if you can point me to some examples if there is any. I'm not a professional developer so it is very possible that I missed something obvious. Thanks!

Koenkk: @invis-z if those options aren't supported by the device, the device will not send them so they won't appear in the published MQTT payload

invis-z: The problem is the data used here is sent by the device as a binary buffer and the convertor is decoding it. I'll try to dig a little bit further. Last time when I copied this whole part into the external convertor it is showing the unrelated options. It is also showing up in the settings(specific) page. The options are things like precision and only used internally within the convertor. I'll look again a bit later today.

Koenkk: I see your point about the options, don't worry about that, I can fix that.

invis-z: Just took a look at the original aqara_opple convertor again. it seems that only the RTCGQ12LM got the illumination data, yet the illumination related options is still in my settings page. According to the code, every/most device seems to support the temperature if location 3 is reported. My device did report it and it seems that the data is in the right area. Double checked the device's MCU (NXP JN5189) seems to include a temp probe, so it does make sense there. So, I think it should be okay to merge, but the issue is in the original aqara_opple convertor, it should not add the illumination related settings to unsupported devices. Also, after looking more closely into this one, I think it should be possible to further consolidate the code into this single convertor.

invis-z: The temperature seems is not exposed by most of the devices. So, it would probably be better if the options is only shown to the devices that is set to expose it. My temp is coming as a Int8, so I don't think the option is even useful in my case (and probably many others). Also, since the temperature seems to come from the MCU, it makes is less useful since it is not accruate enough to be used as room temp (also not updating frequent enough). So, if I am only setting it up for my device, I would remove both of the options, just my personal opinion and I understand it might break things removing stuff from a project at this scale.

Koenkk:
Thanks, I will fix the redundant options later

PR #3758, opened 2022-01-28
Add basic support (without regions) for RTCZCGQ11LM

Otnow:
Koenkk/zigbee2mqtt#11019

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -899,6 +899,28 @@ module.exports = [
         ota: ota.zigbeeOTA,
     },
     {
+        zigbeeModel: ['lumi.motion.ac01'],
+        model: 'RTCZCGQ11LM',
+        vendor: 'Xiaomi',
+        description: 'Aqara Presence Detector FP1 (regions not supported for now)',
+        fromZigbee: [fz.aqara_opple],
+        toZigbee: [tz.RTCZCGQ11LM_presence, tz.RTCZCGQ11LM_monitoring_mode, tz.RTCZCGQ11LM_approach_distance],
+        exposes: [e.presence().withAccess(ea.STATE_GET), exposes.text('presence_event', ea.STATE).withDescription('Presence ' +
+            'events: "enter", "leave", "left_enter", "right_leave", "right_enter", "left_leave", "approach", "away"'),
+            exposes.enum('monitoring_mode', ea.ALL, ['undirected', 'left_right']).withDescription('Monitoring mode with or ' +
+                'without considering right and left sides'),
+            exposes.enum('approach_distance', ea.ALL, ['far', 'medium', 'near']).withDescription('The distance at which the ' +
+                'sensor detects approaching'),
+            exposes.numeric('power_outage_count', ea.STATE).withDescription('Number of power outages (since last pairing)')],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            await endpoint.read('aqaraOpple', [0x0142], {manufacturerCode: 0x115f});
+            await endpoint.read('aqaraOpple', [0x0144], {manufacturerCode: 0x115f});
+            await endpoint.read('aqaraOpple', [0x0146], {manufacturerCode: 0x115f});
+        },
+        ota: ota.zigbeeOTA,
+    },
+    {
         zigbeeModel: ['lumi.sensor_magnet'],
         model: 'MCCGQ01LM',
         vendor: 'Xiaomi',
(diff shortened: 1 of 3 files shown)


devices/xiaomi.js
```diff
@@ -898,6 +898,28 @@ module.exports = [
...
         ota: ota.zigbeeOTA,
     },
+    {
+        zigbeeModel: ['lumi.motion.ac01'],
+        model: 'RTCZCGQ11LM',
+        vendor: 'Xiaomi',
+        description: 'Aqara Presence Detector FP1 (regions not supported for now)',
+        fromZigbee: [fz.aqara_opple],
+        toZigbee: [tz.RTCZCGQ11LM_presence, tz.RTCZCGQ11LM_monitoring_mode, tz.RTCZCGQ11LM_approach_distance],
+        exposes: [e.presence().withAccess(ea.STATE_GET), exposes.text('presence_event', ea.STATE).withDescription('Presence ' +
```
Koenkk, line 908: Does it make sense to use `occupancy` instead of `presence`?

devices/xiaomi.js
```diff
@@ -898,6 +898,28 @@ module.exports = [
...
+            'events: "enter", "leave", "left_enter", "right_leave", "right_enter", "left_leave", "approach", "away"'),
```
Koenkk, line 909: Since these are `presence_event` are events, I think it is better to expose them as `action` instead of  `presence_event`

devices/xiaomi.js
Otnow: I used `presence` because _RTCZCGQ11LM_ is functionally analogous to [TuYa MIR-HE200-TY](https://www.zigbee2mqtt.io/devices/MIR-HE200-TY.html), i.e. it defines the presence of motionless people.

devices/xiaomi.js
Otnow: I tested with `action`, but in this case it is not stored in _State_ and is cleared after the event. Is this how it should be with `action`? (using recent dev)

Koenkk: That is expected so it can be used to trigger upon (https://www.zigbee2mqtt.io/guide/faq/#why-is-the-action-property-always-empty).

Otnow: Ok, didn't know about this `action` feature.

It seems to me that if I use `action`, then in automation I will not be able to use the fact of transition from one event to another (for example, _left_enter -> right_leave_). Or am I wrong?

Koenkk: Is this an event or a state? e.g. given that `{"presence_event": "right_enter"}` is published, it can happen that if the device reports its battery after 50 mins the following is published again: `{"presence_event": "right_enter", "battery":100}`, this would not be correct right?

Otnow: > Is this an event or a state?

_This is an event, but it is also required to store its state_ so that the fact of transition from one state to another can be used in automation, i.e. to know what the previous state was.

This is how the state change in _Home Assistant_ looks like (in abbreviated form):

- `action` (_old_state_ is "" and so I can't create an automation considering the previous state):

  ```json
  {
      "event_type": "state_changed",
      "data": {
          "entity_id": "sensor.0x54ef44100042d6eb_action",
          "old_state": {
              "entity_id": "sensor.0x54ef44100042d6eb_action",
              "state": "",            
          },
          "new_state": {
              "entity_id": "sensor.0x54ef44100042d6eb_action",
              "state": "away",            
          }
      },    
  }
  ```
- `presence_event` (_old_state_ is "approach" and so I can create an automation considering the previous state):

  ```json
  {
      "event_type": "state_changed",
      "data": {
          "entity_id": "sensor.0x54ef44100042d6eb_presence_event",
          "old_state": {
              "entity_id": "sensor.0x54ef44100042d6eb_presence_event",
              "state": "approach",
          },
          "new_state": {
              "entity_id": "sensor.0x54ef44100042d6eb_presence_event",
              "state": "away",
          }
      },
  }
  ```

> e.g. given that `{"presence_event": "right_enter"}` is published, it can happen that if the device reports its battery after 50 mins the following is published again: `{"presence_event": "right_enter", "battery":100}`, this would not be correct right?

This will happen even more often (~ _every 5 - 5.5 min_) and is not a problem since the state does not change and there is no event (_"state_changed"_) in the _Home Assistant_ at that moment.

At this stage, according to my observations, _RTCZCGQ11LM_ does not generate consecutive identical events, and if an event occurs, then it differs from the previous one, which is the fact of its occurrence.

As an alternative, I suggest (fb4b445 + 0ddebc1) adding an `action`, and using `presence_event`, renamed `last_action`, to store the state.

Koenkk: I think understand what you mean now, the old implementation was correct, you can remove the action changes.

Koenkk:
Thanks, just in time to be included in tomorrows z2m release!

Otnow:
Thank you too for the prompt merging 🙂

Only the basic functionality of the sensor is presented here, and we still have to work to add a very interesting function for working with regions:

PR #3762, opened 2022-01-28
Add support `flip_indicator_light` for QBKG31LM

z-x-f:
It is also confirmed that the `mode` attribute is no longer supported due to the new firmware.


Koenkk:
thanks!

PR #3764, opened 2022-01-28
Update philips.js

nikolovnet:
Adding a block for Hue Amaze, which has a different model number for the european market. Currently the device is reported as 'Unsupported'.

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1433,6 +1433,15 @@ module.exports = [
         ota: ota.zigbeeOTA,
     },
     {
+        zigbeeModel: ['4023330P6'],
+        model: '4023330P6',
+        vendor: 'Philips',
+        description: 'Hue white ambiance suspension Amaze',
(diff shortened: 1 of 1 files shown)


devices/philips.js
```diff
@@ -1432,6 +1432,15 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp(),
         ota: ota.zigbeeOTA,
     },
+    {
+        zigbeeModel: ['4023330P6'],
+        model: '4023330P6',
+        vendor: 'Philips',
+        description: 'Hue white ambiance suspension Amaze',
+        meta: {turnsOffAtBrightness1: true},
+        extend: hueExtend.light_onoff_brightness_colortemp,
```
Koenkk, line 1441: Please add the color temperature range as described on https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_3-adding-converter-s-for-your-device

PR #3767, opened 2022-01-29
TP-Link MS100 Motion Sensor

Photo64:
This is a new device that supports the following exposes:
occupancy
battery_low
tamper (there does not appear to be hardware support for this item)
battery
illuminance
illuminance_lux
linkquality


Koenkk:
thanks!

PR #3774, opened 2022-01-29
Adding of RGB LED Controller FUT037Z

ulrikskjaerris:


ulrikskjaerris:
@Koenkk: WHy is this check failing, I can see the test reveals it should be duplicates of fingerprints, and I can not see the error, as the modelID and manufactor is different.
I have changed in my local Miboxer.js script, and verified this change works.

Koenkk:
Tests fail because this device is already supported: https://github.com/Koenkk/zigbee-herdsman-converters/blob/5d68e8dbb61af9a32c4b55e88da41648b35976a4/devices/tuya.js#L192

PR #3775, opened 2022-01-29
add generic 'command' to all devices

zivillian:
This is similar to the generic `read` and `write` and allows to execute arbitrary commands for debugging purposes.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -57,7 +57,8 @@ const converters = {
     command: {
         key: ['command'],
         convertSet: async (entity, key, value, meta) => {
-            await entity.command(value.cluster, value.command, (value.hasOwnProperty('payload') ? value.payload : {}), utils.getOptions(meta.mapped, entity));
+            const options = utils.getOptions(meta.mapped, entity);
+            await entity.command(value.cluster, value.command, (value.hasOwnProperty('payload') ? value.payload : {}), options);
             meta.logger.info(`Invoked '${value.cluster}.${value.command}' with payload '${JSON.stringify(value.payload)}'`);
         },
     },
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -54,6 +54,13 @@ const converters = {
             meta.logger.info(`Wrote '${JSON.stringify(value.payload)}' to '${value.cluster}'`);
         },
     },
+    command: {
+        key: ['command'],
+        convertSet: async (entity, key, value, meta) => {
+            await entity.command(value.cluster, value.command, (value.hasOwnProperty('payload') ? value.payload : {}), utils.getOptions(meta.mapped, entity));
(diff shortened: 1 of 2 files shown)


Koenkk:
Nice, @nurikk is this something we can also support from the frontend?

nurikk:
Sure, will add

nurikk:
> This is similar to the generic `read` and `write` and allows to execute arbitrary commands for debugging purposes.

Hi, @zivillian. I have added support for arbitrary command execution into z2m frontend, dev version, kindly help to test assap before the next release which is happening on 1st Feb

zivillian:
@nurikk I've tested the UI, but sadly there was a misunderstanding. The UI is expecting numerical input for command and cluster, but my PR is using strings.

Here's an example payload, which is working if sent via MQTT to `zigbee2mqtt/FRIENDLY_NAME/set`:

```json
{
  "command":{
    "cluster":"genLevelCtrl",
    "command":"moveToLevel",
    "payload":{
      "level":100,
      "transtime":0
    }
  }
}
```

nurikk:
> @nurikk I've tested the UI, but sadly there was a misunderstanding. The UI is expecting numerical input for command and cluster, but my PR is using strings.
> 
> Here's an example payload, which is working if sent via MQTT to `zigbee2mqtt/FRIENDLY_NAME/set`:
> 
> ```json
> {
>   "command":{
>     "cluster":"genLevelCtrl",
>     "command":"moveToLevel",
>     "payload":{
>       "level":100,
>       "transtime":0
>     }
>   }
> }
> ```

Yes, this is my intended behaviour, the only issue was that commands were sen't as  numeric string values, I've fixed and now they're sending as numbers. Using raw command numbers gives you grater freedom to hack, you can send any commands, there's no need in herdsman knowing them. Kindly update to the latest dev to get fixed version

Screenshot 2022-01-30 at 9 19 43 PM


For example:
```json
{
  "command":{
    "cluster":"genLevelCtrl",
    "command":"moveToLevel",
    "payload":{
      "level":100,
      "transtime":0
    }
  }
}
```

is the same as

````json
{
  "command":{
    "cluster": 8,
    "command": 0,
    "payload":{
      "level":100,
      "transtime":0
    }
  }
}
```

zivillian:
@nurikk I've updated and tested again, and it works like a charm. I would prefer to select the cluster and command with dropdowns, but I see the benefit of invoking unknown commands. Big thanks!

nurikk:
Yes, I understand that using dropdown picker is nicer user experience, and eventually we'll come to hybrid input mode, but for now I don't have much time. Feel free to raise a feature request in frontend repository, eventually I'll implement it

Koenkk:
Thanks both!

Koenkk:
CC @kirovilya ; might also be interesting for iobroker

PR #3781, opened 2022-01-31
Update Sinopé WL4200 and WL4200S devices

Maelstrom96:
The current water sensors definitions don't currently show battery information. I tried to validate if the low battery warning would work, and it didn't seem to be the case, so I changed the device definition. It now pulls the proper battery status and exposes them.

While checking the product page, I saw a reference that said that the devices could report "freeze risks", so I checked the available clusters and it does report temperatures, so that has been added too.

Also, I removed the useless Temper attribute that doesn't seem to do anything.

Lastly, I edited the WL4200S definition to properly differentiate it from the WL4200 model.

### DIFF
--- a/devices/sinope.js
+++ b/devices/sinope.js
@@ -304,17 +304,32 @@ module.exports = [
         model: 'WL4200',
         vendor: 'Sinopé',
         description: 'Zigbee smart water leak detector',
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

PR #3784, opened 2022-01-31
fix(tuya): extended TS1101_dimmer_module_1ch fingerprints

sjors98:
As discussed in: https://github.com/Koenkk/zigbee2mqtt/issues/11081
I bought the Tuya 1 on [AliExpress](https://nl.aliexpress.com/item/1005003012298844.html?gatewayAdapt=glo2nld&spm=a2g0o.9042311.0.0.dbe84c4dNKaMnY) but the zigbee2mqtt service didn't recognize it as supported. This should be the fix :)

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1235,7 +1235,8 @@ module.exports = [
         exposes: [e.switch().setAccess('state', ea.STATE_SET), e.voltage(), e.power(), e.current(), e.energy()],
     },
     {
-        fingerprint: [{modelID: 'TS1101', manufacturerName: '_TZ3000_xfs39dbf'}],
+        fingerprint: [{modelID: 'TS1101', manufacturerName: '_TZ3000_xfs39dbf'}, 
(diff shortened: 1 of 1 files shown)


sjors98:
Hmm I have a failing test on this one, but I don't really get why. Comparing it to the `TS0601_din` model I don't really see why my fingerprint is the same although it has different values?

Koenkk:
Test fails because this device is already supported (https://github.com/Koenkk/zigbee-herdsman-converters/blob/2c9c807d1fde5e8de085c84710e899b5fabdb3ed/devices/lonsonho.js#L201). Meaning it should work out-of-the-box in the latest-dev.  (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html)

sjors98:
Oh my lord I only now see that this device is indeed from the brand Lonsonho but with the Tuya protocol 🤦🏻 
Device works on the dev branch! Thank you verry much for the quick support! My ligths are nicely dimmable now :)

PR #3785, opened 2022-01-31
Update to add new model CM10ZW

CaptFennec:
The new CM10ZW model is a multi-functional 3 gang relay to replace the RL804QZB model.
It's the same device, with a different ID.
Issue : https://github.com/Koenkk/zigbee2mqtt/issues/11082


Koenkk:
Thanks!

PR #3787, opened 2022-01-31
Added support to Alcantara 2 (Acova) heater

alafon:
This PR adds support for the Alcantara 2 heater (Acova).

I've duplicated the original configuration entry made for the Percale 2 heater because they are not completly the same. The Percale 2 model has presence detection while the Alcantara model does not (even if it is not implemented).

Also, I removed `thermostat_occupied_cooling_setpoint` attribute from `toZigbee` since it does not make sense to me. This might also be removed for the Percale 2 model but I don't have it at home to test it before, and I'd prefer not to introduce a bug for existing users.

Side note: the `local_temperature` is actually (at least on the Alcantara model) the temperature setpoint that can be set on the device itself, and is not the temperature measured by the device, so I'm not sure if it should be reported as the `local_temperature` either...

### DIFF
--- a/devices/acova.js
+++ b/devices/acova.js
@@ -3,33 +3,64 @@ const fz = require('../converters/fromZigbee');
 const tz = require('../converters/toZigbee');
 const reporting = require('../lib/reporting');
 
-module.exports = [{
-    zigbeeModel: ['PERCALE2 D1.00P1.01Z1.00'],
-    model: 'PERCALE2',
-    vendor: 'Acova',
-    description: 'Percale 2 heater',
-    fromZigbee: [fz.thermostat, fz.hvac_user_interface],
-    toZigbee: [
-        tz.thermostat_local_temperature,
-        tz.thermostat_system_mode,
-        tz.thermostat_occupied_heating_setpoint,
-        tz.thermostat_unoccupied_heating_setpoint,
-        tz.thermostat_occupied_cooling_setpoint,
-        tz.thermostat_running_state,
-    ],
-    exposes: [
-        exposes.climate()
-            .withSetpoint('occupied_heating_setpoint', 7, 28, 0.5)
-            .withLocalTemperature()
-            .withSystemMode(['off', 'heat', 'auto'])
-            .withRunningState(['idle', 'heat']),
-    ],
-    configure: async (device, coordinatorEndpoint, logger) => {
-        const endpoint = device.getEndpoint(1);
-        await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg', 'hvacThermostat']);
-        await reporting.thermostatTemperature(endpoint);
-        await reporting.thermostatRunningState(endpoint);
-        await reporting.thermostatOccupiedHeatingSetpoint(endpoint);
-        await reporting.thermostatUnoccupiedHeatingSetpoint(endpoint);
+module.exports = [
+    {
+        zigbeeModel: ['PERCALE2 D1.00P1.01Z1.00'],
+        model: 'PERCALE2',
+        vendor: 'Acova',
+        description: 'Percale 2 heater',
+        fromZigbee: [fz.thermostat, fz.hvac_user_interface],
+        toZigbee: [
+            tz.thermostat_local_temperature,
+            tz.thermostat_system_mode,
+            tz.thermostat_occupied_heating_setpoint,
+            tz.thermostat_unoccupied_heating_setpoint,
+            tz.thermostat_occupied_cooling_setpoint,
+            tz.thermostat_running_state,
+        ],
+        exposes: [
+            exposes.climate()
+                .withSetpoint('occupied_heating_setpoint', 7, 28, 0.5)
+                .withLocalTemperature()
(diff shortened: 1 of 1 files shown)


alafon:
See Koenkk/zigbee2mqtt.io#1190 for the image.

I'll add this in the "Notes" section once the doc is auto-generated by docgen:

### Pairing

- Press and hold the `Mode` button to enter configuration mode
- If already paired with the Enki box, press and hold the `-` (minus) button until `Off` is displayed (you'll probably also have to first remove the heater from the Enki box using the dedicated app)
- Press and hold the `+` (plus) button until a code appears on the screen (this is a pairing code used with the Enki box app we don't use here)
- Turn on pairing mode in zigbee2mqtt
- Press (and sometimes hold) the `+` button until a "progress bar" appears on the screen 
- Wait for pairing. The device should show up in zigbee2mqtt in a few seconds and the screen should now display `On`
- Press and hold the `Mode` button to leave configuration mode (or wait for ~ 30 sec to 1 minute). The device should now display "Box" on the top right corner of the screen and the temperature setpoint

Koenkk:
Thanks!

philippemezzadri:
I have this heater (brand Acova, model Alcantara 2) but the reported vendor is not ACOVA but "ZEHNDER GROUP VAUX ANDIGNY". 
For now, the only way I have to get it work is to copy the proposed converter and change the vendor name.

alafon:
> I have this heater (brand Acova, model Alcantara 2) but the reported vendor is not ACOVA but "ZEHNDER GROUP VAUX ANDIGNY". For now, the only way I have to get it work is to copy the proposed converter and change the vendor name.

@Koenkk correct me if I'm wrong, but the `vendor` attribute is not taken into account to match a converter, only `zigbeeModel`, right ?

Koenkk:
> @Koenkk correct me if I'm wrong, but the vendor attribute is not taken into account to match a converter, only zigbeeModel, right ?

Yes only the zigbeeModel (can be found in the data/database.db entry of this device as `modelId`)

philippemezzadri:
@Koenkk and @alafon Thank you both for the clarification on `vendor`. I may have overlooked the documentation.

@alafon yes the difference I have with you is on `zigbeeModel` as I only have :
`zigbeeModel: ['ALCANTARA2 D1.00P1.00Z1.00']`
where your converters shows:
`zigbeeModel: ['ALCANTARA2 D1.00P1.02Z1.00\u0000\u0000\u0000\u0000\u0000\u0000']`

That may explain why it doesn't work for me.

alafon:
@philippemezzadri please have a look at your `/config/zigbee2mqtt/data/database.db` file to find out and make sure what zigbeeModel is reported when pairing to your system.

image

Then we can add it here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/acova.js#L38

```
        zigbeeModel: ['ALCANTARA2 D1.00P1.02Z1.00\u0000\u0000\u0000\u0000\u0000\u0000', 'YOUR ZIGBEEMODEL'],
```

philippemezzadri:
@alafon I looked at `/config/zigbee2mqtt/data/database.db` as suggested :

image

It seems I have the same `modelId` than you.

Strangely my external converter works well with ` zigbeeModel: ['ALCANTARA2 D1.00P1.00Z1.00']` but if I remove it, my heater is shown as "not supported". Should I unpair / re-pair ?

alafon:
The support for Alcantara 2 was merged in v14.0.406 which means having Koenkk/zigbee2mqtt @1.23.0+. You can check it in the Home Assistant addon admin:

image

philippemezzadri:
I am not using the HA add-on but a docker install on a deported raspberry pi.
Anyway, I did upgrade my image to the last release (1.23), removed my external converter and restarted z2m.

image

Conclusion : not working, the heater is still reported as unsupported.
I have to rely on my external converter.

alafon:
If you have it, other users would probably face the same issue so I'd suggest opening a PR to add the model with the extra chars:

```
        zigbeeModel: ['ALCANTARA2 D1.00P1.02Z1.00\u0000\u0000\u0000\u0000\u0000\u0000', 'ALCANTARA2 D1.00P1.02Z1.00'],
```

alexh3o:
> Side note: the `local_temperature` is actually (at least on the Alcantara model) the temperature setpoint that can be set on the device itself, and is not the temperature measured by the device, so I'm not sure if it should be reported as the `local_temperature` either...

On my 3 alcantara 2, it is indeed the same as the temperature attribute, i.e. the set temperature. It's bizarre and a shame the heater doesn't report the real local temperature. Have you or anyone made proggress on retrieving correct ambient temperature ?

alafon:
> > Side note: the `local_temperature` is actually (at least on the Alcantara model) the temperature setpoint that can be set on the device itself, and is not the temperature measured by the device, so I'm not sure if it should be reported as the `local_temperature` either...
> 
> On my 3 alcantara 2, it is indeed the same as the temperature attribute, i.e. the set temperature. It's bizarre and a shame the heater doesn't report the real local temperature. Have you or anyone made proggress on retrieving correct ambient temperature ?

I'm afraid not

PR #3788, opened 2022-01-31
WZ5: disable white channel control when changing color.

kirovilya:
Otherwise, the device does not always work correctly


Koenkk:
thanks!

PR #3790, opened 2022-01-31
Add new Zigbee Model for Kwikset SmartCode Deadbolt

carlos-sarmiento:
Installed a new SmarCode Deadbolt on my Zigbee network and saw that the model had changed. Added the model to the same object on herdsman and everything works as expected


Koenkk:
thanks!

PR #3802, opened 2022-02-02
Add support for Yale YMF40A RL

juslex:
Added the code for support to Yale YMF40A RL - https://github.com/Koenkk/zigbee2mqtt/issues/11199

Note: the batteries are at 100% but they appear at 50%, could you correct the code at this point?


Koenkk:
Added, fixed the battery percentage division.

PR #3803, opened 2022-02-02
Added support for TuYa S-LUX-ZB TS0601 Lux Sensor

cdemi:


Koenkk:
Please merge with https://github.com/Koenkk/zigbee-herdsman-converters/pull/3803/files#diff-7ae23d3c13c2d7f7e37106590684c791aba8093f8be813b7c74a4f1484017c3bR372

cdemi:
I think what you linked to isn't working correctly. If you meant to say to merge with the other "S-LUX-ZB" just a few lines before, the reason that I did not do so is that apparently this one doesn't support battery monitoring so it doesn't have `e.battery()` and `e.battery_low()`

image
image

What do you think, should I still merge them together?

Koenkk:
@cdemi do you maybe have a tuya gateway to confirm it indeed does not support battery monitoring? I expect it does but maybe uses a different data point ID for it: https://github.com/Koenkk/zigbee-herdsman-converters/blob/208cdbfee9a74682a5b8c0c26bfff9b01230dcac/converters/fromZigbee.js#L7261

cdemi:
No I don't have :(

Koenkk:
If you leave the device running for a day, do you get any `s_lux_zb_illuminance: NOT RECOGNIZED DP` in your logs that could potentially be the battery value? (https://github.com/Koenkk/zigbee-herdsman-converters/blob/208cdbfee9a74682a5b8c0c26bfff9b01230dcac/converters/fromZigbee.js#L7266)

Koenkk:
No that is not needed, this is part of the z2m logging not of the zigbee-herdsman one.

cdemi:
I haven't seen such logs so far and it's been running for quite a while now

devices/tuya.js
```diff
@@ -381,6 +381,19 @@ module.exports = [
         },
         exposes: [e.battery(), e.illuminance_lux(), e.battery_low()],
```
Koenkk, line 382: Given that the only difference is the battery and battery_low expose, try with removing the `S-LUX-ZB TS0601` and change this line to.

```js
    exposes: (device, options) => {
        if (device.manufacturerName === '_TZE200_pisltm67') {
            return [e.illuminance_lux()];
        } else {
            return [e.battery(), e.illuminance_lux(), e.battery_low()];
        }
    },
```

Note: make sure to run at least z2m 1.23.0

cdemi:
Aaa yes! This makes more sense!

Koenkk:
thanks!

leroyloren:
Hi, today this device has arrived and the battery is still detected with a decimal place (100.0%), the indicator looks like 0%.

` Received Zigbee message from '0xa4c138a8ac5ee486', type 'commandDataResponse', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0],"type":"Buffer"},"datatype":4,"dp":1}],"seq":19456}' from endpoint 1 with groupID 0
Info MQTT publish: topic 'zigbee2mqtt/0xa4c138a8ac5ee486', payload '{"battery":"100.0","battery_low":"0.0","illuminance_lux":"2","last_seen":"2022-02-18T15:39:48+01:00","linkquality":47}'`

`{
    "battery": "100.0",
    "battery_low": "0.0",
    "illuminance_lux": "2",
    "last_seen": "2022-02-18T15:39:48+01:00",
    "linkquality": 47
}`

obrazek

leroyloren:
I connected the device to the Tuya gate and the expressed value is 1-1000% luminosity. Low, medium, high (intensity level) are set at the same time

photo_2022-02-18_15-54-48
photo_2022-02-18_15-54-48 (3)
photo_2022-02-18_15-54-48 (2)

Koenkk:
Seems this device exposes `battery` and `battery_low` so will add that back (https://github.com/Koenkk/zigbee-herdsman-converters/commit/d214e49fb659a8d517af3bee9bc14aee6d7551da#commitcomment-75638794)

PR #3805, opened 2022-02-03
ZNJLBL01LM Getting the motor state result in wrong data.

invis-z:
Getting the motor state while running, the device will report the motor state as pause, dispite that the motor is actually running. So GET access should be removed. The device will publish the state attribute on state change, so it is not useful anyway.


Koenkk:
thanks!

PR #3808, opened 2022-02-03
danfoss: set min/max heatSetPointLimit to match the devices max range

sjorge:
Waiting for someone to test this, see https://github.com/Koenkk/zigbee2mqtt/discussions/11125

### DIFF
--- a/devices/danfoss.js
+++ b/devices/danfoss.js
@@ -143,6 +143,9 @@ module.exports = [
             // Time-master + synchronised
             const values = {timeStatus: 3, time: time, timeZone: ((new Date()).getTimezoneOffset() * -1) * 60};
             endpoint.write('genTime', values);
+
+            // Set min/max heat setpoint to support the maximum range
+            endpoint.write('hvacThermostat', {minHeatSetpointLimit: 500, maxHeatSetpointLimit: 3200});
         },
     },
     {


sjorge:
@robertalexa FYI, not sure you still have the device to possible test with though.

robertalexa:
@sjorge See my reply in the conversation thread.

TLDR - i don't think this PR is needed, but i am open to debate

sjorge:
Same, I'm not 100% convinced this is needed either. Also why it's a draft so it doesn't get merged too soon by accident.

robertalexa:
On my Hive TRV max 3200 works, 3500 -> INVALID VALUE error.

So 500 to 3200 can be configured correctly via Dev Console.

I would personally say not to push the PR, despite the command working correctly. OP was the only time we heard of sth like this, and we both know the Ally is super popular and a lot of people are engaged with the Ally thread. I am sure we would know from others if this was a true problem.

Ultimately the decision is yours, in my use case my minimum is 9C to match Google Home minimum. Pushing your commit will re-write my setup upon reconfigure, which happens with every z2m update (i think?)

robertalexa:
If anything, i wonder if maybe we should expose the min and max as attributes and then people can set them "easier", and also providing description in z2m rather than having to find documentation?

sjorge:
I'll close the PR, I guess we can indeed expose the min/max but I don't think we can update the expose data automatically to match. So that's probably going to cause more problems and confusion.

robertalexa:
We can read the values of those attributes though, so adding there here

image

Should automatically populate the Frontend and things would be nice and simple?

PR #3822, opened 2022-02-05
updated moes switch convertors

vladispavlov:
I have two moes switches with optional neutral
https://www.aliexpress.com/item/1005002631613176.html

The current convertor doesn't allow to change power on behavior and indicator light state, but this functionality is available in the tuya app. So I checked datapoints and wrote an updated converter. It works fine.

### DIFF
--- a/lib/tuya.js
+++ b/lib/tuya.js
@@ -578,6 +578,9 @@ const dataPoints = {
     garageDoorTrigger: 1,
     garageDoorContact: 3,
     garageDoorStatus: 12,
+    // Moes switch with optional neutral
+    moesSwitchPowerOnBehavior: 14,
+    moesSwitchIndicateLight: 15,
 };
 
 const thermostatWeekFormat = {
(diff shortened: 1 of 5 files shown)


Koenkk:
thanks!

lib/tuya.js
```diff
@@ -731,6 +734,19 @@ const ZMAM02 = {
...
+    powerOnBehavior: {
+        0: 'off',
+        1: 'on',
+        2: 'previous',
+    },
+    indicateLight: {
+        0: 'off',
+        1: 'switch',
+        2: 'position',
+    },
```
javicalle, line 747: Maybe not related, but I have a Moes switch that accept this posible values:
```
    off = 0x00
    light_when_on = 0x01
    light_when_off = 0x02
    freeze = 0x03
```
The freeze value keeps the current backlight state.

vladispavlov: I checked it in the tuya app and there are 3 values as it was defined.
<details><summary>screenshot</summary></details>

But I tested the "freeze" value and it works. I'll create a new pr. Thanks!

ciB89:
Should this device with neutral wire not also work as a repeater?

vladyspavlov:
> Should this device with neutral wire not also work as a repeater?

It depends on firmware. It wouldn't work as repeater with neutral line on current firmware.

PR #3824, opened 2022-02-05
updated moes switch convertors

vladispavlov:
Previous pr: https://github.com/Koenkk/zigbee-herdsman-converters/pull/3822
Not sure about https://github.com/Koenkk/zigbee-herdsman-converters/commit/6028a32ad8fbae02796e9611873104730337b397. Maybe there is a better way to fix this.

Sorry, I haven't fully tested the previous converter. Have fixed bugs and now it's ok.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -2420,6 +2420,12 @@ const converters = {
         convert: (model, msg, publish, options, meta) => {
             const dpValue = tuya.firstDpValue(msg, meta, 'moes_switch');
             const dp = dpValue.dp;
+
+            // tuya_switch datapoints
+            if (dp >= 1 && dp <= 4) {
+                return null;
+            }
+
             const value = tuya.getDataValue(dpValue);
 
             switch (dp) {
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -3218,7 +3218,7 @@ const converters = {
                 );
                 break;
             case 'indicate_light':
-                await tuya.sendDataPointValue(
+                await tuya.sendDataPointEnum(
                     entity,
                     tuya.dataPoints.moesSwitchIndicateLight,
                     utils.getKey(tuya.moesSwitch.indicateLight, value),


converters/fromZigbee.js
```diff
@@ -2420,6 +2420,12 @@ const converters = {
         convert: (model, msg, publish, options, meta) => {
             const dpValue = tuya.firstDpValue(msg, meta, 'moes_switch');
             const dp = dpValue.dp;
+
+            // tuya_switch datapoints
+            if (dp >= 1 && dp <= 4) {
```
Koenkk, line 2425: Why was this added, this prevents `fromZigbee:moes_switch: NOT RECOGNIZED DP` from being logged.

vladispavlov: If not to add this condition when the state changes it also displays warning `fromZigbee:moes_switch: NOT RECOGNIZED DP`. So `tuya_switch` and `moes_switch` converters work together at the same time.
I'm sure that it's solution not good, but it works fine. Maybe it makes sense to replace switch case to if conditions.

Koenkk: thanks for the explanation!

SimonMerrett:
Sorry if this is hijacking but the fact that this PR is fixing things makes me wonder if I am able to change something to enable binding of a Sonoff snzb-01 button to a Moes ZTS-EU_1gang light switch. I am able to bind the Sonoff buttons to Philips Hue bulbs but whenever I try to bind (command line and browser front end) I get 
```
Zigbee2MQTT:error 2022-02-07 18:33:12: Nothing to bind from 'sonoff_switch_13' to 'moes_switch_01'
Zigbee2MQTT:info  2022-02-07 18:33:12: MQTT publish: topic 'zigbee2mqtt/bridge/log', payload '{"message":{"from":"sonoff_switch_13","to":"moes_switch_01"},"type":"device_bind_failed"}'
Zigbee2MQTT:info  2022-02-07 18:33:12: MQTT publish: topic 'zigbee2mqtt/bridge/response/device/bind', payload '{"data":{"clusters":[],"failed":[],"from":"sonoff_switch_13/1","to":"moes_switch_01/1"},"error":"Nothing to bind","status":"error","transaction":"on6wv-9"}'
Zigbee2MQTT:error 2022-02-07 18:33:12: Nothing to bind
```
Is there something in the converters that should be changed to let the Sonoff button bind with the Moes light switch?

PR #3829, opened 2022-02-06
Adds Jasco 43095

GoldbergAlexander:
Adds Jasco 43095, aka Enbrighten Zigbee Switch, Dual Outlet Control Plug-In. Note this treats both outlets as one, but per docs they should be independently controllable.

Tested with the exact device and everything appears to function correctly (apart from the whole only being able to control both outlets together thing)


Koenkk:
thanks!

PR #3840, opened 2022-02-08
X5H electric thermostat support

vladyspavlov:
https://www.aliexpress.com/item/1005003264121802.html

```{"id":15,"type":"Router","ieeeAddr":"0xa4c13868925b0057","nwkAddr":23663,"manufId":4417,"manufName":"_TZE200_2ekuz3dz","powerSource":"Mains (single phase)","modelId":"TS0601","epList":[1,242],"endpoints":{"1":{"profId":260,"epId":1,"devId":81,"inClusterList":[4,5,61184,0],"outClusterList":[25,10],"clusters":{"genBasic":{"attributes":{"65503":"-��)i<��)iK��)i","65506":48,"65508":0,"stackVersion":0,"dateCode":"","appVersion":68}}},"binds":[],"configuredReportings":[],"meta":{}},"242":{"profId":41440,"epId":242,"devId":97,"inClusterList":[],"outClusterList":[33],"clusters":{},"binds":[],"configuredReportings":[],"meta":{}}},"appVersion":68,"stackVersion":0,"hwVersion":1,"dateCode":"","zclVersion":3,"interviewCompleted":true,"meta":{},"lastSeen":1644284062452,"defaultSendRequestWhen":"immediate"}```

Features:
- [x] State (system mode)
- [x] Mode
- [x] Working status (running state)
- [x] Sound (beeper)
- [x] Frost protection
- [x] Set heating temperature
- [x] Set temperature ceiling
- [x] Current temperature
- [x] Temperature correction
- [x] Weekly procedure (schedule)
- [x] Working day setting
- [x] Factory reset
- [x] Child lock
- [x] Sensor selection
- [x] Temperature difference
- [x] Protection temperature limit
- [x] Backplane screen brightness

### CODE REVIEW

File: lib/tuya.js [hunk 1]
```diff
@@ -581,6 +581,26 @@ const dataPoints = {
     // Moes switch with optional neutral
     moesSwitchPowerOnBehavior: 14,
     moesSwitchIndicateLight: 15,
+    // X5H thermostat
```
vladispavlov, line 584: Some of these datapoints are already defined for other devices.
Should I use predefined datapoints or create new ones for my device?
Is it possible some of those devices may receive a firmware update with updated datapoints and this will break my device?

lib/tuya.js
```diff
@@ -581,6 +581,26 @@ const dataPoints = {
...
+    x5hWeeklyProcedure: 30, // MAYBE equals to moesSchedule
+    x5hWorkingDaySetting: 31, // DONE // equals to tvWorkingDay // thermostatWeekFormat
+    x5hFactoryReset: 39, // DONE
+    x5hChildLock: 40, // DONE // equals to moesChildLock
+    x5hSensorSelection: 43, // DONE // equals to moesSensor
+    x5hFaultAlarm: 45, // DONE
+    x5hTempDiff: 101, // DONE
+    x5hProtectionTempLimit: 102, // DONE
+    x5hOutputReverse: 103, // DONE
+    x5hBackplaneBrightness: 104, // DONE
```
vladispavlov, line 603: Some of these datapoints are already defined for other devices.
Should I use predefined datapoints or create new ones for my device?
Is it possible some of those devices may receive a firmware update with updated datapoints and this will break my device?

converters/toZigbee.js
```diff
@@ -4952,6 +4952,195 @@ const converters = {
...
         },
     },
+    x5h_thermostat: {
+        key: ['system_mode', 'current_heating_setpoint', 'sensor', 'brightness_state', 'sound', 'frost_protection', 'week', 'factory_reset',
+            'local_temperature_calibration', 'protection_temp_limit', 'temp_diff', 'output_reverse', 'upper_temp', 'preset', 'child_lock'],
+        convertSet: async (entity, key, value, meta) => {
+            switch (key) {
+            case 'system_mode':
+                await tuya.sendDataPointBool(entity, tuya.dataPoints.x5hState, value === 'heat');
+                break;
```
vladispavlov, line 4962: Wouldn't it be better to split this to separate converters? On the one hand, they will be reusable for other devices. On the other hand, there maybe no more devices with the same set of features on the same datapoints.

lib/tuya.js
Koenkk: This is fine

converters/toZigbee.js
Koenkk: This is also fine

vladyspavlov:
@Koenkk Could you reopen pr? I want to finish it. Or it will be better to create a new pr?

vladyspavlov:
In work

ianclegg:
I have one of these, how is the development going, do you need any assistance?

vladyspavlov:
> I have one of these, how is the development going, do you need any assistance?

I don't have a lot time to finish all features of the thermostat. Now I stuck on schedule feature, have ideas how to do it, but maybe later. Seems like all other features are working for now.
I'll commit some fixes today and mark pull request ready to review.

vladyspavlov:
I sniffed traffic and determined packet with schedule data looks like this:
`00f11e000020060000cd0800000f0b1e000f0d1e000f110000161600000f060000141600000f`

It can be splitted to periods and decoded:
`00f11e0000200` - tuya "header": seq, cluster and data type.
___
Weekday setting
`060000cd` - Period 1
0x06 = 06 | 0x00 = 00 | 0x00cd = 205
which means 06:00, 20.5 degrees

`0800000f` - Period 2
0x08 = 08 | 0x00 = 00 | 0x000f = 150
08:00, 15.0 degrees

`0b1e000f` - Period 3
0x0b = 11 | 0x1e = 30 | 0x000f = 150
11:30, 15.0 degrees

`0d1e000f` - Period 4
13:30, 15.0

`11000016` - Period 5
17:00, 22.0

`1600000f` - Period 6
22:00, 15.0
___
Day off Setting
`06000014` - Period 1 
06:00, 20.0

`1600000f` - Period 2
22:00, 15.0

So we need just to decode this data  in converter to text or any other expose and vice versa. It can be done similarly as here:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/aa44881e0ee01bcc8724f0f92c69ddf03899e518/converters/toZigbee.js#L4934
https://github.com/Koenkk/zigbee-herdsman-converters/blob/aa44881e0ee01bcc8724f0f92c69ddf03899e518/converters/fromZigbee.js#L4729

Maybe someone can do faster than me.

vladyspavlov:
There is also an issue on setting `upper_temp`. `current_heating_setpoint` writes temperature to value of `upper_temp` divided by 10. So if you set `upper_temp` to 40, then `current_heating_setpoint` will be 4.0 degrees. I sniffed traffic to determine how it works and all settings are sent one by one and after that it sends `current_heating_setpoint` with value before setting `upper_temp` to rewrite it.
So I think it can be fixed via converter using globalStore. Put `current_heating_setpoint` value to the store before setting `upper_temp`, then set temp and get from the store value of `current_heating_setpoint` and overwrite it.
Or leave it as it is now and the user will do it manually. I don't think anyone will change this option very often, so it's not a big deal.

vladyspavlov:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/12849

vladyspavlov:
> There is also an issue on setting `upper_temp`

Fixed in https://github.com/Koenkk/zigbee-herdsman-converters/pull/3840/commits/e6d335f2f0eaeebcd543683363c06f1bf9076e31

vladyspavlov:
Added converters for schedule feature in https://github.com/Koenkk/zigbee-herdsman-converters/pull/3840/commits/3c3ce1abd1d87768a6eb46deae04dac68d85a823
Full functionality of the thermostat are now supported.

@Koenkk can you review, please?

Koenkk:
Thanks!

sheaur:
Wrong time synchronization and no periodical time synchronization on thermostat
hello, my device, the problem has not been solved: _TZE200_2ekuz3dz, z2m 1.28.1 zStack3x0 20220219
ieeeAddr: 0xa4c13851bf0f0033

PR #3848, opened 2022-02-09
TS011F_plug(NAS-WR01B) update for OTA

zvldz:


zvldz:
and https://github.com/Koenkk/zigbee-OTA/pull/94

PR #3855, opened 2022-02-11
Add Profalux support.

rclsilver:
Related to https://github.com/Koenkk/zigbee2mqtt/issues/11362


rclsilver:
ZOE device isn't recognized with the fingerprint. Any idea to fix that?

I didn't find model name from the manufacturer, if someone find them, we can comment!

devices/profalux.js
```diff
@@ -0,0 +1,91 @@
+const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
+const exposes = require('zigbee-herdsman-converters/lib/exposes');
+const utils = require('zigbee-herdsman-converters/lib/utils');
+const reporting = require('zigbee-herdsman-converters/lib/reporting');
+const e = exposes.presets;
+const ea = exposes.access;
+
+const tzCover = {
```
Koenkk, line 8: This looks very similar to https://github.com/Koenkk/zigbee-herdsman-converters/blob/325a960821dba3597cb510246a34aa39dbff8a7f/converters/toZigbee.js#L296 , could you use the existing converter instead?

rclsilver: Hum, cover_via_brightness do not allow to send the STOP command.
Any hint to refactor this code?

rclsilver: I can add a condition in cover_via_brightness to handle the stop command and use it if you preferer?

Koenkk: > I can add a condition in cover_via_brightness to handle the stop command and use it if you preferer?

yes please

Koenkk:
thanks!

PR #3856, opened 2022-02-11
Added support for AWOX 33952

9mrcookie9:
Hello, 
Methods for color and refresh are based on debug info from device.

### DIFF
--- a/devices/awox.js
+++ b/devices/awox.js
@@ -31,7 +31,7 @@ const fzLocal = {
             const color = awoxRemoteHelper.convertToColorName(msg.data);
             if (color != null) {
                 return {
-                    action: 'button_'+color,
+                    action: color,
                 };
             }
         },
@@ -42,7 +42,7 @@ const fzLocal = {
         convert: (model, msg, publish, options, meta) => {
             if (msg.data.movemode === 1 && msg.data.rate === 12) {
                 return {
-                    action: 'button_refresh_colored',
+                    action: 'refresh_colored',
                 };
             }
         },
@@ -53,7 +53,7 @@ const fzLocal = {
         convert: (model, msg, publish, options, meta) => {
             if (awoxRemoteHelper.isRefresh(msg.data)) {
                 return {
-                    action: 'button_refresh',
+                    action: 'refresh',
                 };
             }
         },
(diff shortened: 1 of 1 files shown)


devices/awox.js
```diff
@@ -1,4 +1,73 @@
...
+    refresh: {
+        cluster: 'genLevelCtrl',
+        type: ['raw'],
+        convert: (model, msg, publish, options, meta) => {
+            if (awoxRemoteHelper.isRefresh(msg.data)) {
+                return {
+                    action: 'button_refresh',
+                };
+            }
+            return {
```
Koenkk, line 65: In this case don't return anything (also for the other converters)

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1232/files

kodifan:
hallo 
is there any way to avoid (ignoring) sending additional power on/off message from device after pressing certain buttons? 
i.e.:
Debug Received Zigbee message from 'Awox_zigbee_pilot', type 'raw', cluster 'genLevelCtrl', data '{"data":[17,5,16,1,1],"type":"Buffer"}' from endpoint 1 with groupID 32776
Debug Received Zigbee message from 'Awox_zigbee_pilot', type 'commandOn', cluster 'genOnOff', data '{}' from endpoint 1 with groupID 32776

9mrcookie9:
@kodifan unfortunately this behaviour comes from the motherboard itself.

devices/awox.js
9mrcookie9: Done :)

devices/awox.js
```diff
@@ -8,6 +68,23 @@ module.exports = [
...
+            ]},
+        ],
+        model: '33952',
+        vendor: 'AwoX',
+        description: 'Remote controller',
+        fromZigbee: [fz.command_on, fzLocal.colors, fzLocal.refresh, fzLocal.refreshColored, fz.command_off,
+            fz.command_step, fz.command_move, fz.command_stop, fz.command_recall, fz.command_step_color_temperature],
+        toZigbee: [],
+        exposes: [e.action(['on', 'off', 'button_red', 'button_refresh', 'button_refresh_colored', 'button_blue', 'button_yellow',
+            'button_green', 'brightness_step_up', 'brightness_step_down', 'brightness_move_up', 'brightness_move_down', 'brightness_stop',
```
Koenkk, line 85: Last note (sorry I didn't mention this initially). It is better to remove the `button_` prefix, this make it more consistent with the other actions (e.g. `on` isn't called `button_on`)

9mrcookie9: No problem, fixed :) 
Are these rules perhaps written down in some file?
Sorry for my ignorance, I didn't search before.

9mrcookie9:
The naming has been corrected, we are up to date.

Koenkk:
thanks!

PR #3863, opened 2022-02-13
Add support for Profalux Visio covers

glamirand:
https://github.com/Koenkk/zigbee2mqtt/issues/11206


devices/profalux.js
```diff
@@ -0,0 +1,34 @@
+const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
+const tz = require('zigbee-herdsman-converters/converters/toZigbee');
+const exposes = require('zigbee-herdsman-converters/lib/exposes');
+const reporting = require('zigbee-herdsman-converters/lib/reporting');
+const e = exposes.presets;
+const ea = exposes.access;
+
+module.exports = {
+    fingerprint: [{type: 'Router', manufacturerID: 4368, endpoints: [{ID: 1, profileID: 260, deviceID: 512},]}],
```
Koenkk: Why not also add the input/output cluster like https://github.com/Koenkk/zigbee2mqtt/issues/11206#issuecomment-1035290270?

devices/profalux.js
```diff
@@ -0,0 +1,46 @@
...
+            default:
+                throw new Error(`Value '${value}' is not a valid cover position (must be one of 'OPEN' or 'CLOSE' or 'STOP')`);
+            }
+        },
+    },
+};
+
+module.exports = [
+    {
+        fingerprint: [{type: 'Router', manufacturerID: 4368, endpoints: [{ID: 1, profileID: 260, deviceID: 512}]}],
```
Koenkk, line 32: Why not also add the input/output cluster like https://github.com/Koenkk/zigbee2mqtt/issues/11206#issuecomment-1035290270?

glamirand: It did work well, but I tried to describe the fingerprint as close as other devices are when not using fingerprint (mainly using model and manufacturer name) so I modified the fingerprint in order to have only characteristics which seems releveant to me (manufId, used EP, profiletId and deviceId)
Do you think we'd better use the fingerprint you wrote?

Koenkk: > Do you think we'd better use the fingerprint you wrote?

Yes, this reduces the chances of conflicts with other fingerprints

glamirand: Ok then
It seems theere's a duplication of the profalux file : 
https://github.com/Koenkk/zigbee2mqtt/issues/11362
and
https://github.com/Koenkk/zigbee-herdsman-converters/pull/3855

I'll let you merge the requests.

Note I also added support for SLV LED drivers : https://github.com/Koenkk/zigbee-herdsman-converters/pull/3863/commits/110f677b269492bb6447afeecbcc8a592f26898c
Not sure I did the GitHub part properly but the code is working so I let you merge this request as well.

devices/slv.js
```diff
@@ -0,0 +1,23 @@
...
+        vendor: 'SLV',
+        description: 'VALETO CCT LED Driver',
+        extend: extend.light_onoff_brightness_colortemp({colorTempRange: [153, 370]}),
+    },
+    {
+        zigbeeModel: ['1002994'],
+        model: '1002994',
+        vendor: 'SLV',
+        description: 'VALETO Remote',
+        fromZigbee: [],
```
Koenkk, line 19: I would expect some commands here, does it send anything to the coordinator when pressing the buttons?

glamirand: Actually this a multi group remote delivered with the panels and I didn't work on i'ts integration more than just identifying it on the map (it's supposed th be directly binded to the LED panels).
Forget this device if you don't want to populate the file with a not yet really usefull description ;)

ahuillet:
Was this actually merged? The current version of https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/profalux.js does not reflect the contents of the PR, yet it doesn't seem it was reverted either. @glamirand was there a mishap?

PR #3867, opened 2022-02-14
Support running_state for TS0601_thermostat

krasnoukhov:
Device: https://www.zigbee2mqtt.io/devices/TS0601_thermostat.html

It supports `position` (also known as `valvePos`) of valve being open in %, so if it's non-zero it would now report running_state as `heat`, which is quite useful in HA interface.

This will likely bring support to `GS361A-H04` as well since it exposes position:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/5e61dd36b5fbfe60dc1d52386efe70cfca102d23/devices/siterwell.js#L34

As well as running state:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/5e61dd36b5fbfe60dc1d52386efe70cfca102d23/devices/siterwell.js#L37

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4453,7 +4453,7 @@ const converters = {
             case tuya.dataPoints.ecoTemp:
                 return {eco_temperature: value};
             case tuya.dataPoints.valvePos:
-                return {position: value};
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks!

PR #3872, opened 2022-02-15
Fixes Miboxer FUT039Z

plasticrake:
The existing configuration for this device was not working for me. Color temperatures were reversed, and setting color was not working. See Koenkk/zigbee2mqtt#10923

### DIFF
--- a/devices/miboxer.js
+++ b/devices/miboxer.js
@@ -12,14 +12,8 @@ module.exports = [
         model: 'FUT039Z',
         vendor: 'Miboxer',
         description: 'RGB+CCT LED controller',
-        toZigbee: [tz.on_off, tz.tuya_led_control, tz.tuya_do_not_disturb, tz.tuya_color_power_on_behavior],
-        fromZigbee: [fz.on_off, fz.tuya_led_controller, fz.brightness, fz.ignore_basic_report],
-        exposes: [e.light_brightness_colortemp_colorhs([153, 500]).removeFeature('color_temp_startup'),
-            exposes.binary('do_not_disturb', ea.STATE_SET, true, false)
-                .withDescription('Do not disturb mode'),
-            exposes.enum('color_power_on_behavior', ea.STATE_SET, ['initial', 'previous', 'cutomized'])
-                .withDescription('Power on behavior state'),
-        ],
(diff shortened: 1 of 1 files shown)


kirovilya:
wait. where is 'color_power_on_behavior' and 'do_not_disturb' ?

plasticrake:
@kirovilya I set `disableColorTempStartup: false` as in my testing it didn't seem the device supported it when it was on. This model `TS0505B` appears throughout this repo and I set this version of it similarly to how all the others were setup.

I do see that in moes.js it is setup a little differently than the rest, I can test this with my device and see if it works.
```js
        toZigbee: extend.light_onoff_brightness_colortemp_color().toZigbee.concat([
            tz.tuya_do_not_disturb, tz.tuya_color_power_on_behavior,
        ]),
        meta: {applyRedFix: true, enhancedHue: false},
        fromZigbee: extend.light_onoff_brightness_colortemp_color().fromZigbee,
        exposes: extend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}).exposes.concat([
            exposes.binary('do_not_disturb', ea.STATE_SET, true, false)
                .withDescription('Do not disturb mode'),
            exposes.enum('color_power_on_behavior', ea.STATE_SET, ['initial', 'previous', 'cutomized'])
                .withDescription('Power on behavior state'),
        ]),
```

Also, is `'cutomized'` how it should appear? Is the misspelling of `customized` deliberate?

plasticrake:
When configured as in the `moes.js` file `color_power_on_behavior` is now working in my testing. However I don't know if `do_not_disturb` is working for my device. I do not see any difference regardless if it is set to On or Off. Perhaps I don't understand how it is supposed to work.

kirovilya:
Checked. Very bad red.
applyRedFix doesn't help
The rest - it works.

plasticrake:
@kirovilya I agree the red has too much blue, regardless if `applyRedFix` is true or false. How do you fix that?

kirovilya:
can't think of a quick fix

kirovilya:
if change the 'color' exposes from 'XY' to 'HS', then the red color will be correct.

plasticrake:
I think I have it correct now. The red looks better for me

Koenkk:
thanks!

PR #3895, opened 2022-02-18
PAPP shall be exposed in single and 3 phases modes

H3buss:


Koenkk:
FYI @vk496

vk496:
Thnks @H3buss 

There may be some other bugs, since I'v tried to follow the Zlinky docs and not the Enedis directly

PR #3897, opened 2022-02-19
Fixed Matsee Garage Door opener being stuck

H3buss:
Some of us were unable to trigger the command when the door status is false. (zigbee2mqtt #10358)

This commit fixed it by switching the command, istead of sending true always.


JacekKac:
Not sure what to do - i have the same problem with my matsee garage door opener. I looked into the code of converters and i can see in devices.js - tuya that patch is applied, but i search in tozigbee.js and  there is nothing for this device. @H3buss can you check it please? 
i am on last dev version

PR #3904, opened 2022-02-21
Add support for GIDEALED RGBWW controller

ancestor:


Koenkk:
thanks!

Cyberbeni:
@ancestor shouldn't the color temp range maximum be 370 instead of 500, like it is mentioned in the linked issue?

PR #3905, opened 2022-02-21
Fix Aqara WS-EUK switch status will not be reported when it is changed.

z-x-f:
Fix https://github.com/Koenkk/zigbee2mqtt/issues/11507
The device may have to be re-paired for the change to take effect.


Koenkk:
thanks!

PR #3906, opened 2022-02-21
Expose BASE for 3 phases mode

H3buss:
Note: There are probably other data that should be exposed in 3 phases mode. I don't have access to the official Linky TIC doc to review that.


Koenkk:
thanks!

PR #3907, opened 2022-02-21
Add Schneider Electric Wiser CCT595011 motion sensor

vaizki:
Initial support for the SE Wiser celing / wall motion sensor.

This supports the motion alarm (occupancy) well but illuminance reporting is refusing to get configured and then if somehow gets through the sensor will spam illuminance values every few seconds, draining the battery. So more investigation is needed and I will create another PR when it works.


vaizki:
Wasn't sure how to handle the relationship between having a read-only attribute that reports number of levels and rw-attr for setting the current sensitivity level. Especially because reading anything from this device is agonizingly unrealiable no matter how it's triggered. So the value of 4 levels is now hardcoded because that is the only result I've ever gotten.

converters/toZigbee.js
```diff
@@ -6331,6 +6331,17 @@ const converters = {
             await entity.read('lightingBallastCfg', ['wiserControlMode'], {manufacturerCode: herdsman.Zcl.ManufacturerCode.SCHNEIDER});
         },
     },
+    wiser_motion_sensitivity: {
+        key: ['sensitivity'],
+        convertSet: async (entity, key, value, meta) => {
+            // Sensitivity levels are not documented so we cannot map to an enum
+            await entity.write('ssIasZone', {'currentZoneSensitivityLevel': value});
```
Koenkk, line 6338: What about merging this with TS0210_sensitivity? (https://github.com/Koenkk/zigbee-herdsman-converters/blob/8212f77eca0bcef157c31510a0e1d35fad5a87b3/converters/toZigbee.js#L6193)

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1255

converters/toZigbee.js
vaizki: TS0210 seems to use 0,2,6 as the sensitivity levels, I am not sure what the levels exactly are on the Wiser (that's why there's no enum) but the range seems to be 0-3 so the options don't match with the TS0210. I don't have the schneider gateway unfortunately so reverse-engineering is a pain and also the sensor just ignores 99% of all commands sent to it even if I trigger them right after a report or reset / pairing.

Koenkk: Although the levels don't match, we can still use it via numbers by changing `const sens = {'high': 0, 'medium': 2, 'low': 6}[value];` to `const sens = typeof value === 'string' ? {'high': 0, 'medium': 2, 'low': 6}[value] : value;`

converters/fromZigbee.js
```diff
@@ -7600,6 +7600,37 @@ const converters = {
...
+                result.sensitivity = value;
+            }
+            if (data && data.hasOwnProperty('numZoneSensitivityLevelsSupported')) {
+                const value = data.numZoneSensitivityLevelsSupported;
+                result.sensitivity_levels = value;
+            }
+            return result;
+        },
+    },
+    wiser_motion_alarm: {
```
Koenkk, line 7621: This converter is exactly the same as ias_occupancy_only_alarm_2

Koenkk:
Added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/3948

PR #3908, opened 2022-02-21
Expose humidity for Stelpro SMT402 series thermostats

dipcore:
### Description
Stelpro SMT402 thermostats provides humidity value. We can just expose its value so it can be easily used in home automation. ie turn on/off vent in bathroom, control humidifiers in another rooms etc without additional devices.

### Screenshot
Home assistant sensors
image


Koenkk:
thanks!

PR #3909, opened 2022-02-21
Add unit to angle and angle_axis

claha:
I'm using homeassistant and have a sensor that exposes an angle measurement but without any unit of measurement. This should fix that.


Koenkk:
thanks!

PR #3916, opened 2022-02-22
Fix `warning_simple`

EXP-Carthage:
Added the return of the `alarm` state so that it can be stopped in the case of a test run.
Which is the case for this [device](https://www.zigbee2mqtt.io/devices/SMSZB-120.html#develco-smszb-120).

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -378,6 +378,7 @@ const converters = {
                 {startwarninginfo: info, warningduration: 300, strobedutycycle: 0, strobelevel: 0},
                 utils.getOptions(meta.mapped, entity),
             );
+            return {state: {'alarm': value}};
         },
     },
     squawk: {


Koenkk:
I don't think this is correct, only a state should be returned here and this is an action.

EXP-Carthage:
For the device mentioned above, this allows to overcome the problem that it displays on the frontend 'OFF ? START'.
And on HA, only the 'START' command is available because the action button remains permanently on OFF

Koenkk:
But after the alarm stops, what is the value of `state`? Is it updated to reflect that the arm is off?

EXP-Carthage:
Here are the logs of the manual triggering of the alarm :
```
Zigbee2MQTT:debug 2022-02-24 14:34:12: Received MQTT message on 'zigbee2mqtt/0x0015bc003100b82a/set' with data '{"alarm":"START"}'
Zigbee2MQTT:debug 2022-02-24 14:34:12: Publishing 'set' 'alarm' to '0x0015bc003100b82a'
Zigbee2MQTT:info  2022-02-24 14:34:14: MQTT publish: topic 'zigbee2mqtt/0x0015bc003100b82a', payload '{"alarm":"OFF","battery":100,"battery_low":false,"enrolled":true,"fault":false,"ias_cie_address":"0x00124b0024c9ef42","last_seen":"2022-02-24T13:34:14.384Z","linkquality":29,"max_duration":10,"reliability":"no_fault_detected","restore_reports":true,"smoke":false,"supervision_reports":true,"temperature":22,"test":false,"update":{"state":"idle"},"voltage":3000,"zone_id":23}'
```

The `alarm` state remains `OFF` even though the device is ringing.
No log return once the alarm is stopped

Koenkk:
And what is the log after this change? Does it stay  START?

EXP-Carthage:
The logs I put above are the ones right after running the 'START' command. There is no log to indicate the state when the command is launched or even when the alarm stops

Koenkk:
Is https://github.com/Koenkk/zigbee-herdsman-converters/pull/3916#issuecomment-1049871648 before or after your change?

EXP-Carthage:
Before

Koenkk:
Thanks, does the value of  ALARM go back to OFF once the ALARM is finished?

EXP-Carthage:
No, it stays on START

Koenkk:
Is any zigbee message received once the alarm stops? You can check this by enabling the debug logging: 

EXP-Carthage:
No zigbee message is received once the alarm is over. And I reassure you, I always leave my `log_level` on `debug` ;)

EXP-Carthage:
Screenshot_20220224_175936
I just came across this information. It is therefore normal that the alarm status does not go up.
The original document is [here](https://www.smarthome-store.de/media/documents/smoke-alarm-technical-manual.pdf)

Koenkk:
I see, I think we should completely remove the `alarm` from the payload (since we cannot provide a meaningful state for it)

EXP-Carthage:
I think that this PR can be closed because indeed, the state of the alarm does not reflect the reality when the alarm is ordered.
On the other hand, as indicated in the device documentation, the alarm status, `alarm` is only triggered if the device detect smoke or fire.

PR #3918, opened 2022-02-22
Support Philips 9290022169 as 4080130P6

cadavre:
Philips Hue bulb light model 9290022169 is also discovered as 4080130P6.

This simple commit adds support to this bulb otherwise discovered as "Not supported" device.

Tested with external converter and confirmed to work.


Koenkk:
thanks!

PR #3921, opened 2022-02-23
Updating xiaomi batteries calculation to logarithmic percentage from 2850 to 3200

zen2:
for Issue #8499

- add a log boolean parameter to function toPercentage to select a logarithm or linear percentage result (default)
- create a new option type 3V_2850_3000_log
- use this option type for xiaomi batteries end devices

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -56,7 +56,7 @@ module.exports = [
         description: 'Aqara P1 door & window contact sensor',
         fromZigbee: [fz. xiaomi_contact, fz.ias_contact_alarm_1, fz.aqara_opple, fz.battery],
         toZigbee: [],
-        meta: {battery: {voltageToPercentage: '3V_2850_3200'}},
+        meta: {battery: {voltageToPercentage: '3V_2850_3000_log'}},
         exposes: [e.contact(), e.battery(), e.battery_voltage()],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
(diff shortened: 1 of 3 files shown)


Koenkk:
Found it didn't work with 0 yet, fixed this now and added some test cases. thanks!

zen2:
Good catch for log(0) !
Thanks ! now I know the PR workflow ^^

leroyloren:
What specifically makes this better? All my Aqara sensors have changed 30-40% (realistically) to 99-00% with this commit.

obrazek

zen2:
> What specifically makes this better? All my Aqara sensors have changed 30-40% (realistically) to 99-100% with this commit.
> 
> obrazek

It's not especially related to this commit. I suppose that you were on stable branche and so you get the new minimum value of 2850 that have appeared in the dev branch before this commit.
The last PR use logarithmic value instead of linear. It means that battery % decrease slowly at high voltage and decrease quicker for low voltage but references points for 0% and 100% are stricly the same.

On my side, the minimum value of 2850 is too low: I got several THP sensors that was on last stable at 33% and now get 0% and I know these sensors will works for a long time (total life is at least 18 months).

leroyloren:
I think it's really related to this PR, I primarily use Edge, because I'm interested in the news (before stable), with this PR today, all battery states of Aqara sensors have changed to 99-100%.

leroyloren:
Can't it be related to the fact that 3200 mV, now 3000, was originally calculated?

https://github.com/Koenkk/zigbee-herdsman-converters/pull/3921/files#diff-5dfe38baf287dcf756a17c2dd63483781b53bf4b669e10efdd01e74bcd8e780aR173

PR #3922, opened 2022-02-23
Add new TS0502B revision

lug-gh:
model:TS0502B
description:Light controller
modelZigbee:TS0502B
type:Router
nwk:13754
manuf id:4417
manufacturer:_TZ3210_zwqnazkb


Koenkk:
thanks!

PR #3926, opened 2022-02-25
Add support for action group for trust remote

ikilledmypc:
I would like to add support for the action group property of the full state of the device:
{
    "linkquality": 33,
    "action": "",
    "action_group": 150,
    "action_rate": 30
}
This value will change depending on the current selected group on the remote and will allow controlling individual lights as well as all lights. Contains value 145 .... 150, where 150 is all groups


ikilledmypc:
I assumed this was the way to do it haven't been able to test it locally on my home assistant setup

Koenkk:
Thanks!

PR #3933, opened 2022-02-27
Add Waxman 8850100/leakSMART Valve 2.0

kevin-david:
Docs are here: https://github.com/Koenkk/zigbee2mqtt.io/pull/1269

I tested this locally with the 8850100 I have, but in theory it should work for all other models as well and I doubt they identify themselves differently.

### DIFF
--- a/devices/waxman.js
+++ b/devices/waxman.js
@@ -30,7 +30,7 @@ module.exports = [
         // 8850310 2"
         model: '8850100',
         vendor: 'Waxman',
-        description: 'leakSMART Automatic Water Shut-off Valve 2.0',
+        description: 'leakSMART automatic water shut-off valve 2.0',
         fromZigbee: [fz.battery, fz.on_off],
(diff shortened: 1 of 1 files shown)


devices/waxman.js
```diff
@@ -19,4 +20,26 @@ module.exports = [
...
     },
+    {
+        zigbeeModel: ['House Water Valve - MDL-TBD'],
+        // Should work with all manufacturer model numbers for the 2.0 series:
+        // 8850000 3/4"
+        // 8850100 1"
+        // 8850200 1-1/4"
+        // 8850300 1-1/2"
+        // 8850310 2"
+        model: '8850100',
```
kevin-david, line 31: Given the above comment, I'm not sure if this is the right way to "identify" this model. Open to whatever the existing pattern is here, if any.

Koenkk: This is fine!

Koenkk:
thanks!

PR #3943, opened 2022-03-01
invert door status for _TZE200_nklqjk62

st4ro:
inverted status reported - becouse garage_door: on means open, off means closed (ie door_contact = TRUE should be "closed")

still we needs to customize home assistant device class for this device - this should be done somehow in homeassistant.ts


Koenkk:
For all contact sensors the following applies:
- Contact = contact: true = ON
- No contact = contact: false = OFF

PR #3948, opened 2022-03-02
feat: add support for CCT595011_AS

larrifax:
Add support for Schneider Electric's [Wiser Motion Sensor](https://www.se.com/ww/en/product/CCT595011_AS/wiser-motion-sensor/), which is also white-labeled by other companies, like the [Elko SmartSensor](https://proff.elko.no/smartsensor-bevegelse-tradlos/category2957.html) (in Norwegian).


larrifax:
Just spotted #3907, which is adding the same device.
I'll leave the current PR open until either of them is merged.

Koenkk:
Thanks!

PR #3950, opened 2022-03-03
Fix Lixee Report for Three Phase

SilentT-FR:
Fix for Legacy Mode report
fix Sum delivred
DPM & FPM report

Using this reference: https://www.enedis.fr/media/2035/download


Koenkk:
thanks!

PR #3955, opened 2022-03-04
tuya_radar_sensor addded datapoint 104

Burki24:
Datapoint 104: Detection date. At the moment not sure, what it belongs to.


converters/fromZigbee.js
```diff
@@ -7929,6 +7929,9 @@ const converters = {
             case tuya.dataPoints.trsIlluminanceLux:
                 result = {illuminance_lux: value};
                 break;
+            case tuya.dataPoints.trsDetectionData:
+                result = {detection_data: value};
```
Koenkk, line 7933: We should also add this the the `expose` array of the definition.

Burki24: Added the expose to device/tuya.js - at the moment, I'm not sure, what this dp send. Several times, I get an warning: DP104 is unknown, that's the reason for integrating.

Thanks.

Koenkk: If we don't know what it means, it would be better to just ignore it.

Burki24: That's why I didn't made  an entry in exposes first.
TuYa Bridge sniffing says, it's called motion detection, but it isn't clear, what that means.
I only added it to catch the warning from z2m "unknown dp".
So, if it's okay for you, I will delete the expose, but will let the DP included in "fromzigbe.js".

Burki24: Ah, okay, thanks

Koenkk:
thanks!

PR #3813, opened 2022-02-03
Add support for energy metering on sengled e11-g13 bulbs

jmarshak:
I noticed my zigbee sengled e11-g13 bulbs were reporting power and energy usage values in zha.  So I added it for zigbee2mqtt ;-).  Tested using an external converter in zigbee2mqtt.

`npm run lint` and `npm test` both pass 100%

Screen Shot 2022-02-03 at 3 11 14 PM

PR #3958, opened 2022-03-04
Add power metering & OTA to various Sengled devices

benley:
- Add power metering for Sengled E21-N1EA and E1F-N5E, which are the
  only ones I can directly confirm support it.  I suspect there are
  several more Sengled models that do, but I don't have the hardware to
  test any others.

- Add OTA support to various Sengled devices based on info from
  https://zigbee.blakadder.com/

Tested using external converters to confirm power metering.  `npm run lint` and `npm test` pass cleanly.

(following #3813)


Koenkk:
thanks!

PR #3961, opened 2022-03-06
Update hive.js

russdan:
Added in `emergency_heating` to system mode water for SLR2/b/c devices to fix the following error which shows in Domoticz when using native MQTT device:

```
Error: Zigbee2MQTT: Climate device invalid/unknown mode received! (0x001e5e00000cd00_climate_water_zigbee2mqtt: emergency_heating)
```

system mode `emergency_heating` is shown in the MQTT message, but not currently handled:

```
MQTT publish: topic 'zigbee2mqtt/Hive Controller', payload '{"linkquality":15,"local_temperature_heat":21.38,"occupied_heating_setpoint_heat":21,"occupied_heating_setpoint_water":22,"running_state_heat":"idle","running_state_water":"idle","system_mode":"emergency_heating","system_mode_heat":"heat","system_mode_water":"off","temperature_setpoint_hold_duration_heat":65535,"temperature_setpoint_hold_duration_water":0,"temperature_setpoint_hold_heat":true,"temperature_setpoint_hold_water":false}}'
```

I think this is also a valid state for Endpoint heat too, when selecting 'Boost' of the heating from the SLT controller, but I only use this on water so have only tested on water at present

### DIFF
--- a/devices/hive.js
+++ b/devices/hive.js
@@ -291,7 +291,7 @@ module.exports = [
                 .withDescription('Period in minutes for which the setpoint hold will be active. 65535 = attribute not' +
                     ' used. 0 to 360 to match the remote display').withEndpoint('heat'),
             exposes.climate().withSetpoint('occupied_heating_setpoint', 22, 22, 1).withLocalTemperature()
-                .withSystemMode(['off', 'auto', 'heat']).withRunningState(['idle', 'heat']).withEndpoint('water'),
+                .withSystemMode(['off', 'auto', 'heat', 'emergency_heating']).withRunningState(['idle', 'heat']).withEndpoint('water'),
             exposes.binary('temperature_setpoint_hold', ea.ALL, true, false)
                 .withDescription('Prevent changes. `false` = run normally. `true` = prevent from making changes.' +
                     ' Must be set to `false` when system_mode = off or `true` for heat').withEndpoint('water'),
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

robertalexa:
@russdan and @Koenkk 

A few things to note on the above:
- emergency_heating is also supported on the heat endpoint
- the reason why i have not exposed this is because the device itself will not accept native boost mode if the commands needed to active it are not sent in 1 single payload. If the commands are split, the device fails to boost and sets wrong parameters or reverts to previous state. That is why i have described the boost functions in the online documentation instead of allowing people to press (in the frontend) random buttons :). So in essence, the device can be set to emergency_heating (boost) via mqtt or via thermostat remote, but the state will not be shown in the Frontend correctly (aware of)
- also, emergency_heating is an invalid thermostat state in home assistant

Once the weather gets better I plan on coming back to this project and finish my native boost automations via Home Assistant. At that point, I will review this and might have some more answers/approaches and if needed I will update the documentation of all the SLRs

russdan:
Hi @robertalexa, thanks...  I don't use Home Assistant which is possibly why these issues manifested themselves, as you can see from the error message in my original description above it just shows as an unhandled mode - I have successfully (in Python) send 3 separate messages:

```
msg = {temperature_setpoint_hold_duration_water": duration}
client.publish('zigbee2mqtt/Hive Controller/set', '{"system_mode_water": "emergency_heating"}')
client.publish('zigbee2mqtt/Hive Controller/set', json.dumps(msg))
client.publish('zigbee2mqtt/Hive Controller/set', '{"temperature_setpoint_hold_water": true}')
```
and this works, at least from Python it does - they have to be in this sequence with the "unhold" as the final command, and they are all send simultaneously, so possibly appear as a single command to the SLR2 which is why it works in code, but wouldn't work as you say if doing this via more spaced-out button presses.

Sending them as a single message also works, as you say, ie 
```
msg = {"system_mode_water": "emergency_heating","temperature_setpoint_hold_duration_water": duration, "temperature_setpoint_hold_water": "1"}
client.publish('zigbee2mqtt/Hive Controller/set', json.dumps(msg))
```
but I don't know what limitations HA places on how you do this, but I believe its possible to do both approaches, if that makes any difference!  The main thing I was trying to fix was the non-HA use-case, to remove the errors which were appearing in Domoticz logs, which this fix does - does this change create any issues for HA users?

PR #3968, opened 2022-03-08
Expose battery & position for Qmotion blinds (HDM40PV620)

strobejb:
Exposes shade position and battery percentage as well as reporting for Qmotion HDM40PV620 shades. This change is based exactly on the contribution made in (PR/issue #2643)

I made one small addition by including the following line at the top:

`const reporting = require('../lib/reporting');`

Everything else is the same (kudos to @pabergman). With this change I can now observe battery and position values in the zigbee2mqtt UI and mqtt gateway.


Koenkk:
thanks!

PR #3974, opened 2022-03-10
Add 801480

adam-moyer:
Adds support for the first generation of the white ambiance 4" retrofit recessed downlights.


Koenkk:
thanks!

PR #3975, opened 2022-03-11
Expose cover running state

Drafteed:
Related PR: Koenkk/zigbee2mqtt/pull/11789

The expose added for all devices where the `running` can be present.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -998,7 +998,7 @@ module.exports = [
         fromZigbee: [fz.tuya_cover, fz.ignore_basic_report],
         toZigbee: [tz.tuya_cover_control, tz.tuya_cover_options],
         exposes: [
-            e.cover_position().setAccess('position', ea.STATE_SET),
+            e.cover_position().withRunning().setAccess('position', ea.STATE_SET),
             exposes.composite('options', 'options')
                 .withFeature(exposes.numeric('motor_speed', ea.STATE_SET)
                     .withValueMin(0)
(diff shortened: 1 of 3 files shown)


Koenkk:
Thanks!

Koenkk:
Right after merging I found out this is a breaking change (since all running sensors will be removed from HA). To avoid that I propose to add the binary running sensor where applicable (like https://github.com/Koenkk/zigbee-herdsman-converters/blob/87f53ea94a97e89bf78961ed764ea4e1bdc4ae52/devices/xiaomi.js#L1288)

Drafteed:
Oops... sorry.

leroyloren:
I use an intermediate position to open the blind (open, intermediate position, closed) and this PR has made it impossible to work with the blind when the intermediate position is open. Resp. only full opening is active, closing from intermediate position is inactive.

Drafteed:
> I use an intermediate position to open the blind (open, intermediate position, closed) and this PR has made it impossible to work with the blind when the intermediate position is open. Resp. only full opening is active, closing from intermediate position is inactive.

Which device model?

leroyloren:
_TZE200_xuzcvlku

Drafteed:
Is `Running` value always true? [Here already wrote about it.](https://github.com/Koenkk/zigbee2mqtt/commit/8f1787ed4a32912ca5f5a0e2b9cc5ff8b6064fb8#commitcomment-68595849) I will remove the expose for this device.

Drafteed:
> and could motor_speed delete this device as well please? The device does not support this, but this option is in the settings.

I can't do it, because it may work [on other devices](https://github.com/Koenkk/zigbee-herdsman-converters/blob/4bed9034520c78827a8d0a5563aaae36b64d47a2/devices/tuya.js#L946-L977).

PR #3976, opened 2022-03-11
Fix Orvibo AM25 power source

gauch:
The battery is showing the correct value but is not recognized as a battery device.


Koenkk:
thnaks!

PR #3977, opened 2022-03-11
Add 929003046101 model Philips Hue Runner.

erlendsellie:
Philips Hue Runner Single Spotlight Model 929003046101 was not in zigbeeModel list. Tested locally with external converter.


Koenkk:
thanks!

PR #3978, opened 2022-03-11
Update hive.js

russdan:
Add missing local_temperature_water value from water endpoint which generates 'climate device unhandled current_temperature_template' errors via homeassistant messages


Koenkk:
thanks!

PR #3980, opened 2022-03-11
Expose cover running state

Drafteed:
Round II.

Related PR: https://github.com/Koenkk/zigbee2mqtt/pull/11789

### DIFF
--- a/lib/exposes.js
+++ b/lib/exposes.js
@@ -335,12 +335,6 @@ class Cover extends Base {
         this.features.push(new Numeric('tilt', access.ALL).withValueMin(0).withValueMax(100).withDescription('Tilt of this cover'));
         return this;
     }
-
-    withRunning() {
-        assert(!this.endpoint, 'Cannot add feature after adding endpoint');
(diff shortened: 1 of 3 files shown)


lib/exposes.js
```diff
@@ -335,6 +335,12 @@ class Cover extends Base {
         this.features.push(new Numeric('tilt', access.ALL).withValueMin(0).withValueMax(100).withDescription('Tilt of this cover'));
         return this;
     }
+
+    withRunning() {
```
Koenkk, line 339: I'm sorry for not being clear, what I mean was adding this to `exposes` like `exposes.binary('running', ea.STATE, true, false)` (so not making it part of the feature list)

Drafteed: In the first time I tried this, but can`t figure out how to check/find exposed value in this place: https://github.com/Koenkk/zigbee2mqtt/blob/7a2ddf24d32006bc05283c5ceb681dc9871b0347/lib/extension/homeassistant.ts#L380-L427

Koenkk:
thanks!

PR #3989, opened 2022-03-13
Exposes `battery` and `battery_voltage` for 3RSS008Z

ronaldheft:
The Third Reality switch 3RSS008Z was missing battery information. Confirmed the `3V_2100` voltage setting [from the ZHA quirks for this switch](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/thirdreality/switch.py#L27).


Koenkk:
thanks!

PR #3990, opened 2022-03-13
ZLinky_TIC: add setting to allow user to filter exposed values

antoinelibert:
Hello,

Device `ZLinky_TIC` can expose a lot of attributes in certain conditions (mostly users with STANDARD mode for theirs Linky energy meter). This PR adds a new setting (specific) `tic_command_whitelist` to allow user to chose the attributes (s)he wants to expose. The values are the TIC command available [here](https://github.com/fairecasoimeme/Zlinky_TIC#mode-historique). Values should be separated by comma.

@vk496, what do you think? My personal use case is that I only need 5 attributes of ~20 .

image
image

### DIFF
--- a/devices/lixee.js
+++ b/devices/lixee.js
@@ -509,6 +509,14 @@ function getCurrentConfig(device, options, logger=console) {
         break;
     }
 
+    // Filter exposed attributes with user whitelist
+    if (options && options.hasOwnProperty('tic_command_whitelist')) {
+        const tic_commands_str = options['tic_command_whitelist'].toUpperCase();
+        if (tic_commands_str !== 'ALL') {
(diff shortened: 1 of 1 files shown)


Koenkk:
If @vk496 could do a review I can merge this.

vk496:
Hi,

There is a option called "Filter attributes from publish payload." under options, which I supose is intended for that. But I tried it (with attribute name and mqtt attributes) and it didn't work for me. So @antoinelibert solution seems fine to me if there is no native way to limit what is sent to MQTT. @Koenkk

antoinelibert:
My solution also intended to avoid polling and setting up Reporting so there is also less Zigbee trafic :).

PR #3994, opened 2022-03-14
Refactor Xiaomi genBasic converter

McGiverGim:
The `genBasic` cluster is used by Xiaomi devices to send periodically info about the device state. We have a lot of converters to process them, that duplicate code, but now we have one utility method that contains all this attributes.
Usually (not always) each attribute number represents always the same, but for example for the attribute 100 (state) we need to translate it into something more "human readable" depending on the device.
This PR reduces the number of methods processing the genBasic cluster. I've tested it with several Xiaomi/Aqara devices that I have but it affects a lot more.

@Koenkk I don't know if you think is ok the direction taken by this refactor. I will not push any more refactor before the next version because this changes are big enough and I think needs to be tested some time by the users.


converters/fromZigbee.js
```diff
@@ -5300,31 +5312,6 @@ const converters = {
...
-                voltage = msg.data['65282']['1'].elmVal;
-            }
-
-            if (voltage) {
-                const payload = {
-                    voltage: voltage, // @deprecated
-                    // voltage: voltage / 1000.0,
-                };
-
-                if (model.meta && model.meta.battery && model.meta.battery.voltageToPercentage) {
```
Koenkk, line 5320: This logic is missing https://github.com/Koenkk/zigbee-herdsman-converters/blob/530b629569428661ccf5cc207a4ac490116648a5/lib/xiaomi.js#L412

converters/fromZigbee.js
```diff
@@ -5392,32 +5379,6 @@ const converters = {
...
-                    const illuminance = msg.data['65281']['11'];
-                    // DEPRECATED: remove illuminance_lux here.
-                    result.illuminance = calibrateAndPrecisionRoundOptions(illuminance, options, 'illuminance');
-                    result.illuminance_lux = calibrateAndPrecisionRoundOptions(illuminance, options, 'illuminance_lux');
-                }
-
-                if (msg.data['65281'].hasOwnProperty('3')) {
-                    let temperature = msg.data['65281']['3'];
-                    temperature = calibrateAndPrecisionRoundOptions(temperature, options, 'temperature');
-                    result.temperature = temperature;
```
Koenkk, line 5414: RTCGQ11LM is missing from https://github.com/Koenkk/zigbee-herdsman-converters/blob/530b629569428661ccf5cc207a4ac490116648a5/lib/xiaomi.js#L159

converters/fromZigbee.js
```diff
@@ -5550,27 +5511,6 @@ const converters = {
...
-        type: ['attributeReport', 'readResponse'],
-        options: [exposes.options.precision('temperature'), exposes.options.calibration('temperature')],
-        convert: (model, msg, publish, options, meta) => {
-            if (msg.data.hasOwnProperty('65281')) {
-                const result = {};
-                if (msg.data['65281'].hasOwnProperty('100')) {
-                    result.contact = msg.data['65281']['100'] === 0;
-                }
-
-                if (msg.data['65281'].hasOwnProperty('3')) {
```
Koenkk, line 5564: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/3994/files#r826181423

converters/fromZigbee.js
```diff
@@ -5581,39 +5521,6 @@ const converters = {
...
-                const humidity = parseFloat(msg.data['65281']['101']) / 100.0;
-
-                // https://github.com/Koenkk/zigbee2mqtt/issues/798
-                // Sometimes the sensor publishes non-realistic vales, filter these
-                if (temperature > -65 && temperature < 65) {
-                    result.temperature = calibrateAndPrecisionRoundOptions(temperature, options, 'temperature');
-                }
-
-                // in the 0 - 100 range, don't produce messages beyond these values.
-                if (humidity >= 0 && humidity <= 100) {
```
Koenkk, line 5603: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/3994/files#r826183158

converters/fromZigbee.js
```diff
@@ -5628,22 +5535,6 @@ const converters = {
             }
         },
     },
-    SJCGQ11LM_temperature: {
```
Koenkk, line 5631: SJCGQ11LM is missing from https://github.com/Koenkk/zigbee-herdsman-converters/blob/530b629569428661ccf5cc207a4ac490116648a5/lib/xiaomi.js#L159

converters/fromZigbee.js
McGiverGim: Yes, you are right. For others elements, when no defined, we use `batteryVoltageToPercentage(value, '3V_2100');` as default. Do you want that I add it here or you prefer to simply not declare percentage?

converters/fromZigbee.js
McGiverGim: No, as you can see, the condition is negative, so it will enter here. The major part of devices have a "3" attribute that shows temperature, I don't know why this three precisely don't have it.

converters/fromZigbee.js
```diff
@@ -5581,39 +5521,6 @@ const converters = {
...
-            exposes.options.precision('humidity'), exposes.options.calibration('humidity')],
-        convert: (model, msg, publish, options, meta) => {
-            if (msg.data['65281']) {
-                const result = {};
-                const temperature = parseFloat(msg.data['65281']['100']) / 100.0;
-                const humidity = parseFloat(msg.data['65281']['101']) / 100.0;
-
-                // https://github.com/Koenkk/zigbee2mqtt/issues/798
-                // Sometimes the sensor publishes non-realistic vales, filter these
-                if (temperature > -65 && temperature < 65) {
```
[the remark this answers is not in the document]
McGiverGim: You are right again. It saw it, but I thouht the condition was the inverse... I will fix.

converters/fromZigbee.js
McGiverGim: You are right again. It saw it, but I thouht the condition was the inverse... I will fix.

converters/fromZigbee.js
McGiverGim: Same than above, the condition is negative, so it will enter here.

converters/fromZigbee.js
McGiverGim: I suppose that is the same than above that you are talking about. The condition is precisely the opposite, it will publish the temperature.

converters/fromZigbee.js
Koenkk: whoops, missed this!

converters/fromZigbee.js
Koenkk: If no `meta.voltageToPercentage` is defined, no voltage -> battery % converting should take place.

McGiverGim: In the old `aqara_opple` converter, for the value "1" it used it directly:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/13f1160bb2a5406f7de5ae773d17eae03c371299/converters/fromZigbee.js#L5371-L5373

I saw other converters that use the 1 the with the meta, so in the new utility file, , I use the `meta.voltageToPercentage` if defined and if not I use the old conversion directly.

Do you want that I remove it from here too then? 
https://github.com/Koenkk/zigbee-herdsman-converters/blob/9427a03d3d53f198a5c9c22ec9ceaa856693eca5/lib/xiaomi.js#L151-L154

I think is cleaner. The worst thing that can happen is that some device that don't define the meta will not have now a percentage, and is easy to fix adding the correct meta.

McGiverGim:
Fixed the bug. Only the comment on the review left.

converters/fromZigbee.js
Koenkk: Yes, please. Only when `meta.voltageToPercentage` is defined, voltage -> battery % calculation should be done.

McGiverGim: Fixed!

McGiverGim:
@Koenkk all the issues fixed. :)

Koenkk:
Perfect, went through it all again and looks good.

Big thanks, this is a very nice cleanup! 🏅

PR #3997, opened 2022-03-14
Tuya Smart Housekeeper: Added Values pm25

Burki24:


devices/tuya.js
```diff
@@ -143,7 +143,8 @@ module.exports = [
         description: 'Smart air house keeper',
         fromZigbee: [fz.tuya_air_quality],
         toZigbee: [],
-        exposes: [e.temperature(), e.humidity(), e.co2(), e.voc(), e.formaldehyd().withUnit('ppm'), e.pm25()],
+        exposes: [e.temperature(), e.humidity(), e.co2(), e.voc(), e.formaldehyd().withUnit('ppm'),
+            e.pm25().withValueMin(0).withValueMax(0).withValueStep(1)],
```
McGiverGim, line 147: Is correct to have min and max values to zero?

Burki24: Thanks, that was a mistake. The Value is 0-999. I commited the correction right now.

PR #4002, opened 2022-03-15
Fix issue with Elko power metering added in #3511

Sabesto:
Fix an issue where 0W would be reported any time a device report without power reading was received

Previous pull request: https://github.com/Koenkk/zigbee-herdsman-converters/pull/3511
Issue mentioned: https://github.com/Koenkk/zigbee2mqtt/issues/7781#issuecomment-1061890264


Koenkk:
thanks!

PR #4004, opened 2022-03-15
Create the new definition for the RB 266

simonbaudart:
Hello

I got today a new version of the RB 265 and I tested this definition in my local home assistant lab, got it working.
The functions are the same as the RB 265 but the model is the RB 266.


simonbaudart:
Here is the result of the integration into my home assistant : 

image

Koenkk:
thanks!

PR #4007, opened 2022-03-15
add plugwise 106-03

JanJaapKo:
The Plugwise Tom thermostatic radiator valve. Documentation added in https://github.com/Koenkk/zigbee2mqtt.io/pull/1290


JanJaapKo:
OK, I tested this device as an external device as described in the documentation section, which works fine. The test now fails and unfortunately my knowledge is too little to solve this. Can someone fix this?

Koenkk:
thanks!

PR #4009, opened 2022-03-16
Read MCCGQ01LM contact from genBasic

McGiverGim:
From time to time, when I open and close my dorr, my Xiaomi door sensor detected the open but it does not detect the close action.
I suspected simply the close message was lost. But today I've captured one log in debug mode and I can see the open door, with the `genOnOff` cluster, as expected, and then, instead of another `genOnOff` with the close action, as usually, I receive the `genBasic` one with the close state.
So I added the option to detect the open/close using the `genBasic` cluster.


Koenkk:
thanks!

PR #4013, opened 2022-03-18
fixes Koenkk/zigbee2mqtt#6071

simons-public:
Newer ERIA lights from Amazon with the zigbeeModel AD-RGBW3001 need `colorTempRange` set and support HS.


Koenkk:
thanks!

PR #4016, opened 2022-03-18
fix xiaomi pressure value

nukusinji:
#3994 
fix pressure value


Koenkk:
Thanks! May fix https://github.com/Koenkk/zigbee-herdsman-converters/pull/4016

antoinelibert:
Hello,

I believe this change broke the `pressure_precision` setting (specific). 0 and 1 digit work but not 2 or 3

WSDCGQ11LM
1.24.0-dev commit: 0cd02c7

PR #4021, opened 2022-03-19
New device support: Namron Touch Termostat model 4512737

owangen:
Product description for this thermostat here https://www.namron.com/products/namron-zigbee-touch-termostat-16a-hvit/


Koenkk:
thanks!

PR #4025, opened 2022-03-20
Parameter verification

### LINKED ISSUE #4019: Unexpected behavior when attempting to read color temperature
From ioBroker, we are trying to trigger reading group member states after the corresponding group state has been set. This works well for the most states, but it fails when attempting to read the color temperature or color state

What confuses me is this:
- I verified that the converter is the correct converter (light_colortemp). (`converter is {"key":["color","color_temp","color_temp_percent"],"options":[{"type":"binary","name":"color_sync","property":"color_sync","access":2,"value_on":true,"value_off":false,"description":"When enabled colors will be synced, e.g. if the light supports both color x/y and color temperature a conversion from color x/y to color temperature will be done when setting the x/y color (default true)."},{"type":"numeric","name":"transition","property":"transition","access":2,"value_min":0,"description":"Controls the transition time (in seconds) of on/off, brightness, color temperature (if applicable) and color (if applicable) changes. Defaults to `0` (no transition)."}]}`)
- a call with `await converter.convertGet(entity.device.endpoints[0],'color_temp', {})` raises an error with no message (a try / catch goes into the catch branch) and message for color_temp is received,
- if the same call is made without `await`, the adapter crashes due to an uncaught exception within a async function raised with `TypeError: Cannot read property 'message' of undefined`
- a call with `await entity.device.endpoints[0].read('lightingColorCtrl', ['colorMode', 'colorTemperature']);` does not raise an exception and a message with the updated color_temp is received.

When checking the code, i can see no reason why the call to converter.convertGet fails as it should simply call the entity.read function.

On a side note: I do get the same 'raised error' behavior when attempting to read the  'color', but the code works without an issue when attempting to read through the 'brightness' or 'state' converters. I did not dig any deeper for the color situation as i suspect that the cause is similar.

I welcome any suggestion as to what to do to fix this behavior. Resorting to direct query of zigbee attributes instead of going through the converters confirmed for the device is something i consider 'bad code'

asgothian:
Verify the 'meta' parameter for readColorAttributes to ensure no exception is raised when it is not provided.

should fix #4019


Koenkk:
thanks!

PR #4027, opened 2022-03-21
Improve `running` and add `motor_direction` for ZNCLDJ11LM

Drafteed:
This PR will improve `running` state for ZNCLDJ11LM. Previously, this value could get stuck in the `true` state (watch gif below).

I also added and expose `motor_direction`, that represents 3-state: `STOP`, `FORWARD` and `REVERSE`.

**Details:**

Data `1028` from `genBasic` returns correct current motor state (direction), instead `61440` from `genAnalogOutput`.

<details>
<summary>Video comparison</summary>
Before:

After:

</details>


converters/fromZigbee.js
```diff
@@ -5662,23 +5662,25 @@ const converters = {
...
+        },
+    },
+    xiaomi_curtain_direction: {
+        cluster: 'genBasic',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const data = msg.data['1028'] || 0;
+            const lookup = {0: 'STOP', 1: 'FORWARD', 2: 'REVERSE'};
+            const direction = lookup[data] || lookup[0];
+            return {running: direction !== 'STOP', motor_direction: direction};
```
McGiverGim, line 5683: We are moving all the `genBasic` reports to a single converter.

You only need to add the `xiaomi_basic` converter to the device:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/4ed602f34946d0678eeb75aa540ac5d0d4ce2491/converters/fromZigbee.js#L5278-L5298

and add your code for the 1028 here:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/4ed602f34946d0678eeb75aa540ac5d0d4ce2491/lib/xiaomi.js#L408-L415

If you put the logs in debug mode you can see if there are other attributes in the message that can be published, in this way the utility method has all the attributes that Xiaomi can send.

Drafteed:
Done. I also convert `xiaomi_curtain_options` to `xiaomi_basic`.

Koenkk:
thanks!

Drafteed:
@Koenkk What do you think about unify `motor_state` and Cover `state`.

Koenkk:
@Drafteed I don't think that is possible, `state` should be either `OPEN` or `CLOSE`

McGiverGim:
It will be great unify it with some constants, but this is a breaking change. It is possible to do this kind of change, @Koenkk ?

surfingbytes:
Could the same be done for newer motor version ZNCLDJ12LM please? It doesn't expose "motor_state", and "running" is always null: https://github.com/Koenkk/zigbee2mqtt/issues/3652
Thank you

Drafteed:
@surfingbytes Please check recent PR #4236 for ZNCLDJ12LM.
Changes available in the dev branch (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html).

surfingbytes:
@Drafteed thanks, I did that, and it's not much better. "motor_state" is now exposed, but it's always null, same as "running". But now, battery, which works in stable version, shows also null in this dev branch. I removed the device and paired it again and still the same.

surfingbytes:
Also (when it's fixed), could the motor_state values be the same as ZNCLDJ11LM - closing, opening, stopped. This PR has "stop" instead of "stopped". Thank you

Drafteed:
> Also (when it's fixed), could the motor_state values be the same as ZNCLDJ11LM - closing, opening, stopped. This PR has "stop" instead of "stopped". Thank you

HA extension lookup any possible `STOP` names. https://github.com/Koenkk/zigbee2mqtt/blob/3f6a13786e67445840521cf8f41b030221a24dee/lib/extension/homeassistant.ts#L410

Drafteed:
> @Drafteed thanks, I did that, and it's not much better. "motor_state" is now exposed, but it's always null, same as "running". But now, battery, which works in stable version, shows also null in this dev branch. I removed the device and paired it again and still the same.

Please open new issue and tag @nukusinji.

PR #4037, opened 2022-03-24
Fix inverted Xiaomi Covers controls

Drafteed:
Xiaomi curtains are fully controlled through position and when inversion is enabled, we also need to invert the position when we send open or close. this is a bug in the current implementation, that breaks the control of curtains with buttons in HA when inverse is enabled. I don't know how people used it before :)

I also added reverse detection to `motor_state`.


Koenkk:
thanks!

hellcry37:
When I've reported this nobody cared :))

hellcry37:
Glad it's fixed thanks!

PR #4039, opened 2022-03-24
Fix Xiaomi Covers position jumping on stop

Drafteed:
When the stop command is sent, the `currentPositionLiftPercentage` of the `closuresWindowCovering` cluster sending the current position of the curtains. 

The problem is that the default converter `fz.cover_position_tilt` inverses the value and sets incorrect position (xiaomi expect 'open' to be 100, 'closed' to be 0, instead Zigbee specs).

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5690,6 +5690,24 @@ const converters = {
             return {position};
         },
     },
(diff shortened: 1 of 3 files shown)


Koenkk:
thanks!

PR #4047, opened 2022-03-26
Min- and max-brightness for tuya dimmer _TZE200_3p5ydos3

fschaeck:
Added min- and max-brightness to tz.tuya_dimmer_level and used it for Tuya dimmer _TZE200_3p5ydos3, refactored fz.tuya_data_point_dump (moved some code into functions in lib/tuya.js) and added some logging to fz.tuya_dimmer (using those new functions in lib/tuya.js).

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -433,42 +433,25 @@ module.exports = [
             {modelID: 'TS0601', manufacturerName: '_TZE200_dfxkcots'},
             {modelID: 'TS0601', manufacturerName: '_TZE200_ojzhk75b'},
             {modelID: 'TS0601', manufacturerName: '_TZE200_swaamsoy'},
+            {modelID: 'TS0601', manufacturerName: '_TZE200_3p5ydos3'},
             {modelID: 'TS0601', manufacturerName: '_TZE200_1agwnems'},
         ],
         model: 'TS0601_dimmer',
         vendor: 'TuYa',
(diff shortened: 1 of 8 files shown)


devices/tuya.js
```diff
@@ -454,6 +453,24 @@ module.exports = [
             {vendor: 'Mercator Ikuü', model: 'SSWD01'},
         ],
     },
+    {
+        fingerprint: [
+            {modelID: 'TS0601', manufacturerName: '_TZE200_3p5ydos3'},
+        ],
+        model: 'TS0601_dimmer',
```
Koenkk, line 460: I expect that the other `TS0601_dimmer` will also support this feature (if not we can always split them later)

fschaeck: I didn't dare changing the code for ALL Tuya dimmers, especially, since the min brightness of 3p5ydos3 is using the data point other dimmers are using as the brightness level...

Well... I moved the exposes definition of min- and max-brightness to the existing model TS0601_dimmer, but I think the better way would be to create a new model like TS0601_dimmer_with_min_max or alike...

fschaeck: Okay... I reverted the description, fixed the tests of tuya_data_point_dump in test/fromZigbee.test.js and corrected a bunch of Lint errors. All checks passed now.

Koenkk:
thanks!

PR #4048, opened 2022-03-26
Added TS0202_2 support

SignalStealer:
Another variant of TS0202 (reported & in print at the bottom of the device), slightly different than the existing 2. I added the device documentation under https://github.com/Koenkk/zigbee2mqtt.io/pull/1300


devices/tuya.js
```diff
@@ -324,6 +324,15 @@ module.exports = [
         vendor: 'TuYa',
         extend: extend.light_onoff_brightness(),
     },
+    {
+        fingerprint: [{modelID: 'TS0202', manufacturerName: '_TYZB01_dr6sduka'}],
+        model: 'TS0202_2',
```
Koenkk, line 329: Since it uses the same converter, this can be merged with https://github.com/Koenkk/zigbee-herdsman-converters/blob/3ff17db7d4cfcb326ab2ccc25cbc0f9a9e11e762/devices/tuya.js#L354 and be added as whitelabel of it.

SignalStealer: I was considering that, but my device does not expose battery nor battery voltage. If you prefer, I can still merge it with the existing model. Please let me know.

Koenkk: Could you add this to the definition, repair the device and wait 24 hours to see if it gets reported?

```js
            await reporting.batteryVoltage(endpoint);
        },
```

Make sure the configure succeeds after repairing it, you should see something like `Succesfully configured...` in the z2m logs.

SignalStealer: Thanks for the help! I will try this and report back

SignalStealer: Yes, after adding the configure method, the device started reporting battery % and voltage. Unfortunately I was not able to validate the values, since both batteries which report 3100 and 2900 in other devices report as 3000 in this one, but it might as well just be inaccurate. I do not have any more discharged batteries to test with. What would you suggest in this situation please?

Koenkk: Just add it for now, it is better than nothing.

SignalStealer: I hope I understood you well, I added the configure method to the PR now.

Koenkk: Since functionality is the same, can you merge it with `TS0202` now? https://github.com/Koenkk/zigbee-herdsman-converters/blob/2351faa29151a1d4ef34a1012bd5159f57454537/devices/tuya.js#L354

SignalStealer: I'm sorry for taking my dear time on this. I have merged it now with TS0202, I hope you didn't mean to also include the configure method there. If so, I will add that.

devices/tuya.js
```diff
@@ -354,7 +355,9 @@ module.exports = [
         model: 'TS0202',
         vendor: 'TuYa',
         description: 'Motion sensor',
-        whiteLabel: [{vendor: 'Mercator Ikuü', model: 'SMA02P'}, {vendor: 'TuYa', model: 'TY-ZPR06'}],
+        whiteLabel: [{vendor: 'Mercator Ikuü', model: 'SMA02P'},
+            {vendor: 'TuYa', model: 'TY-ZPR06'},
+            {vendor: 'Tesla Smart', model: 'Tesla Smart Sensor Motion'}],
```
Koenkk, line 360: For the `model`, please fill in the model number that is listed on the device.

SignalStealer: Label on the device reads TS0202, so I updated it to that

Koenkk:
Thanks!

SignalStealer:
I thank you for all the guidance!

PR #4052, opened 2022-03-28
Add support for the Niko Single connectable switch

sjorge:
koenkk/zigbee2mqtt#11762, look like it's just supports on/off.

As mentioned in the issue power_on_behavior doesn't seem to work.


Koenkk:
thanks!

PR #4054, opened 2022-03-29
Add Niko Double connectable switch

sjorge:
Also brought the model naming in line with the manual for the single connectable switch, they use X in the 2nd to last model number spot as that one changes depending on if it has clamped or screw mount.

Unfortunately, I was also able to verify that both are rather spares in features. You can't even rebind them switch to toggle a group/device like with the Ubisys S1/S2 and some Aqara devices.

I'll try to figure out a docs PR for this later and mention the fact that they can't be rebound which is probably a deal breaker for some people.

### DIFF
--- a/devices/niko.js
+++ b/devices/niko.js
@@ -77,7 +77,7 @@ module.exports = [
     },
     {
         zigbeeModel: ['Single connectable switch,10A'],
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

PR #4055, opened 2022-03-29
Updated to reworked adapter, and a fix

TheStigh:


devices/datek.js
```diff
@@ -34,46 +34,52 @@ module.exports = [
         zigbeeModel: ['Meter Reader'],
         model: 'HSE2905E',
         vendor: 'Datek',
-        description: 'Datek Eva AMS HAN power-meter sensor',
+        description: 'Datek Eva AMS HAN power-meter sensor Re-Worked',
```
Koenkk, line 37: I don't think adding re-worked here adds any value. Better to write something for the changelog.

Koenkk:
Thanks!

PR #4058, opened 2022-03-30
Add support for `JY-GZ-01AQ` and improve `JT-BZ-01AQ/A`

Otnow:
Fixes Koenkk/zigbee2mqtt#11986

This converter is designed to work on coordinator firmware with support for native Aqara logic.

OTA file for `JY-GZ-01AQ` added with Koenkk/zigbee-OTA#101

### DIFF
--- a/lib/xiaomi.js
+++ b/lib/xiaomi.js
@@ -308,9 +308,9 @@ const numericAttributes2Payload = (msg, meta, model, options, dataObject) => {
             break;
         case '160':
             if (['JT-BZ-01AQ/A'].includes(model.model)) {
-                 payload.gas = value === 1;
+                payload.gas = value === 1;
             } else if (['JY-GZ-01AQ'].includes(model.model)) {
-                 payload.smoke = value === 1;
+                payload.smoke = value === 1;
             }
             break;
         case '161':
@@ -389,9 +389,9 @@ const numericAttributes2Payload = (msg, meta, model, options, dataObject) => {
             break;
         case '314':
             if (['JT-BZ-01AQ/A'].includes(model.model)) {
-                 payload.gas = value === 1;
+                payload.gas = value === 1;
             } else if (['JY-GZ-01AQ'].includes(model.model)) {
-                 payload.smoke = value === 1;
+                payload.smoke = value === 1;
             }
             break;
         case '315':
--- a/lib/xiaomi.js
+++ b/lib/xiaomi.js
@@ -174,9 +174,11 @@ const numericAttributes2Payload = (msg, meta, model, options, dataObject) => {
             payload.switch_type = {1: 'toggle', 2: 'momentary'}[value];
             break;
         case '11':
-            payload.illuminance = calibrateAndPrecisionRoundOptions(value, options, 'illuminance');
-            // DEPRECATED: remove illuminance_lux here.
-            payload.illuminance_lux = calibrateAndPrecisionRoundOptions(value, options, 'illuminance_lux');
+            if (!['JT-BZ-01AQ/A'].includes(model.model)) {
+                payload.illuminance = calibrateAndPrecisionRoundOptions(value, options, 'illuminance');
+                // DEPRECATED: remove illuminance_lux here.
+                payload.illuminance_lux = calibrateAndPrecisionRoundOptions(value, options, 'illuminance_lux');
+            }
             break;
         case '100':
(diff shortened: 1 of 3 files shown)


lib/xiaomi.js
```diff
@@ -300,25 +302,52 @@ const numericAttributes2Payload = (msg, meta, model, options, dataObject) => {
...
+            if (['JT-BZ-01AQ/A', 'JY-GZ-01AQ'].includes(model.model)) {
+                payload.test = value === 1;
+            }
             break;
         case '163':
-            payload.mute = value === 1; // JT-BZ-01AQ/A
+            if (['JT-BZ-01AQ/A', 'JY-GZ-01AQ'].includes(model.model)) {
+                payload.mute = value === 1;
+            }
             break;
```
McGiverGim, line 334: The idea with this utility library is to try to not limit by model the codes, if it's possible.
I think that until we know of other devices that use the 163, for example, for other things we must not make it conditional.
In this way some devices will mostly work simply adding the converter and we only need to add/remove the values that can be wrong. If we add a conditional for each device, then we will need to add all the models for each attribute.
That's my opinion. @Koenkk what's your idea? We must filter by model always or only when needed?

Otnow: Unfortunately, Aqara allows itself not to adhere to standards, and even between its own devices, and this library is proof of that.

Examples:

- in current PR, I had to add an exception for the _[JT-BZ-01AQ/A](https://www.zigbee2mqtt.io/devices/JT-BZ-01AQ_A.html)_ gas sensor to _case 11_, because it certainly does not have the ability to measure illuminance:

  https://github.com/Otnow/zigbee-herdsman-converters/blob/59fffda00758f3d3e49718d8765860e95162aa0e/lib/xiaomi.js#L176-L182
  ```javascript
          case '11':
              if (!['JT-BZ-01AQ/A'].includes(model.model)) {
                  payload.illuminance = calibrateAndPrecisionRoundOptions(value, options, 'illuminance');
                  // DEPRECATED: remove illuminance_lux here.
                  payload.illuminance_lux = calibrateAndPrecisionRoundOptions(value, options, 'illuminance_lux');
              }
              break;
  ```
- in current PR, I had to specify model  in condition for _cases 165 and 166_ (both mean `linkage_alarm`), because gas sensor _[JT-BZ-01AQ/A](https://www.zigbee2mqtt.io/devices/JT-BZ-01AQ_A.html)_ has both of these cases, but `linkage_alarm` only corresponds to _case 166_, and _165_ means something else (not clear yet):

  https://github.com/Otnow/zigbee-herdsman-converters/blob/59fffda00758f3d3e49718d8765860e95162aa0e/lib/xiaomi.js#L342-L351
  ```javascript
          case '165':
              if (['JY-GZ-01AQ'].includes(model.model)) {
                  payload.linkage_alarm = value === 1;
              }
              break;
          case '166':
              if (['JT-BZ-01AQ/A'].includes(model.model)) {
                  payload.linkage_alarm = value === 1;
              }
              break;
  ```
- in _case 102_ for the _[RTCZCGQ11LM](https://www.zigbee2mqtt.io/devices/RTCZCGQ11LM.html)_ presence sensor, the value changes depending on the firmware version:

  https://github.com/Koenkk/zigbee-herdsman-converters/blob/2ae0857563959af75dbbe0e019eedff6dbf89463/lib/xiaomi.js#L254-L267

Therefore, I believe that you need to know exactly what each case means for a particular device and therefore explicitly indicate the model in the condition, otherwise there may be problems (for example, duplication of values).

McGiverGim: I think we don't must fix it before is broken. It will complicate and make the code a lot bigger. I understand that we need to add exceptions when we encounter something wrong, but in my opinion we must left the others without conditional.
The worst thing that can happen is that some device will have something bad in the state, but if it does not expose it, it will not be published.
If we decide too filter by model for all the attributes, maybe we must try to move the big switch to methods, one for each attribute, to maintain it under control.
But again, all of this is only my opinion.

Otnow: @Koenkk, I tried to add support for [Aqara smoke detector (JY-GZ-01AQ)](https://github.com/Koenkk/zigbee2mqtt/issues/11986) so that it would be included in today's zigbee2mqtt release, so if there are no fundamental objections to current PR, then please merge it so that it can be included in release.

If, according to remarks @McGiverGim, the conditions in some cases need to be removed, then if necessary, I will do this in a new PR after release of zigbee2mqtt.

Koenkk: Sorry, couldn't work on z2m yesterday and already did the 1.25.0 release this morning. Thanks for the changes!

PR #4061, opened 2022-03-30
Add operation_mode and action for Niko Switches

sjorge:
Merge after https://github.com/Koenkk/zigbee-herdsman/pull/515

### DIFF
--- a/devices/niko.js
+++ b/devices/niko.js
@@ -5,7 +5,7 @@ const reporting = require('../lib/reporting');
 const e = exposes.presets;
 const ea = exposes.access;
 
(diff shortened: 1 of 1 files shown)


sjorge:
The omission of value 0 for the tz converter for the opperationMode attribute is intentional, the device nicely spits out an INVALID error when writing unsupported value, however if you write 0, the device soft-bricks and needs to be removed from power and then rejoined. It seemed best to not allow writes of anything not in the set of known values.

Koenkk:
Thanks as always 👍

PR #4071, opened 2022-04-02
New device supported: Connecte Thermostat

owangen:
This device https://en.connecte.no/termostat

For now, only fromzigbee is working. 
Struggling with getting tozigbee and time sync working for this Tuya device, but will continue working on these issues. Discussed [here](https://github.com/Koenkk/zigbee2mqtt/issues/11895).


devices/connecte.js
```diff
@@ -0,0 +1,44 @@
...
+        toZigbee: [
+        ],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            // Do a "magic" read on the basic cluster to trigger the termostat start reporting.
+            await endpoint.read('genBasic', ['manufacturerName', 'zclVersion', 'appVersion', 'modelId', 'powerSource', 0xfffe]);
+        },
+        exposes: [
+            exposes.binary('state', ea.STATE, true, false).withDescription('State of thermostat'),
+            exposes.numeric('system_mode', ea.STATE),
```
Koenkk, line 24: `system_mode` is already included in the `exposes.climate()`, and also it is not a numerical value (`exposes.numeric`)

owangen: Yes, true. Thanks, will clean up.

owangen:
I realize that it's better to wait until I get tozigbee working before continuing.

PR #4074, opened 2022-04-03
Update: Tuya Smart Air Homekeeper - DP20 now with correct DP in lib/tuya (Formaldehyd)

Burki24:
Fix this Issue 
#12019


Koenkk:
thanks!

PR #4075, opened 2022-04-03
Device Support: Led Trading 9133

asgothian:
Support for Led Trading Push-button-coupler 9133
9126_zigbee-unterputz-modul-fuer-friends-of-hue


Koenkk:
thanks!

PR #4077, opened 2022-04-03
Add Ynoa Smart LED Gu10 RGB CCT (LA-GU10-RGBW) and 5 button control f…

mirakels:
…or RGBW light (LA-5KEY-RGBW)


Koenkk:
Thanks!

devices/ynoa.js
```diff
@@ -18,4 +19,27 @@ module.exports = [
...
+        extend: extend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 526], supportsHS: true}),
+    },
+    {
+        zigbeeModel: ['ZBT-RGBWSwitch-D0800'],
+        model: 'LA-5KEY-RGBW',
+        vendor: 'Ynoa',
+        description: '5 key control for RGBW light',
+        fromZigbee: [fz.command_on, fz.command_off, fz.command_move_to_color_temp,
+            fz.command_move_to_color, fz.command_move_to_level, fz.battery],
+        exposes: [e.battery(), e.battery_low(), e.action(['on', 'off', 'brightness_move_to_level',
```
fgdrf, line 36: Hi folks, Thank you for supporting this device!

I've just noticed, that the device has support for action-groups, means you can switch either the type (color_move, brightnes_move, and color_temperature_move) or group (with dedicated group-Ids 16388-16391).

Howeverfor the move commands its importend to get action_data as well, in case of brightness these are values from 13 to 255 which is a representation for 5% to 100% (13, 64, 127, 194, 255 -> 5, 25, 50, 75, 100%)

```
2024-04-12 23:49:06MQTT publish: topic 'zigbee2mqtt/ynoa_switch_01', payload '{"action":"brightness_move_to_level","action_group":16389,"action_level":255,"action_transition_time":0.03,"battery":100,"battery_low":null,"linkquality":176}'
```

In case of color_temperature_move another value is important, rather than action_level here its action_color_temperature with values [153, 200, 250, 379] - whatever it means.

on color_move its action_color with representation of two dimentional values x/y : 

```
2024-04-12 23:54:24MQTT publish: topic 'zigbee2mqtt/ynoa_switch_01', payload '{"action":"color_move","action_color":{"x":0.22,"y":0.04},"action_group":16389,"action_transition_time":10,"battery":100,"battery_low":null,"linkquality":112}'
```

How is it possible to use this attributes as well, since I cannot see this in home-assistant and don't know how to handle just "color_temperature_move" without any context in which direction ;)

It would be awesome to collaborate to improve this integration, but since I'm new to zigbee2mqtt I've no idea where to start from and how to test.

Where can we start from?

PR #4081, opened 2022-04-03
First stage of additional changes to Centralite pearl thermostat.

nkasrawi:
1) Add a new exposed value "thermostat_running_mode" which is different from both thermostat_system_mode
and thermostat_running_state in that it expresses whether the thermostat is currently cooling, heating, or off (e.g. system mode could be cooling, but if the system is not actually cooling - due to setpoints - running mode would be off), since this value is exposed by this thermostate, and is useful for reporting purposes.

2) Expose the value of thermostat_control_sequence_of_operation and make it changeable.

@alexisiglesias1 
@Koenkk 

I noticed that a new/almost identical configuration was added recently for a centralite 3157100-E and I'm curious what the difference between that thermostat and the 3157100 that I have is, as well as what the difference between fz.legacy.thermostat_att_report and fz.thermostat is.  The new config also supports thermostat_temperature_setpoint_hold which I know the 3157100 also support (I just hadn't bothered to add that setting).  I guess I'm wondering if it might be possible to merge these two configurations, and/or if it would make sense to copy over the various changes I have been making, to the new configuration.

### DIFF
--- a/devices/centralite.js
+++ b/devices/centralite.js
@@ -170,12 +170,14 @@ module.exports = [
             tz.thermostat_occupied_heating_setpoint, tz.thermostat_occupied_cooling_setpoint,
             tz.thermostat_setpoint_raise_lower, tz.thermostat_remote_sensing,
             tz.thermostat_control_sequence_of_operation, tz.thermostat_system_mode,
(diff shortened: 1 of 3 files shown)


Koenkk:
Looks good to me, I guess 3157100-E/3157100 are equal functionality wise but lets keep them separate because of the different model number.

PR #4083, opened 2022-04-03
Second stage of additional changes to centralite pearl thermostat

nkasrawi:
While debugging the activity of this thermostat, I noticed that the configuration
didn't handle the "aux" use case, in that first, there was no "aux" running state declared, and
second, I noticed that different numerical values produced by the thermostat for 
running state were being overloaded onto the value 'heat' - namely
the case where the system is in heating mode, and is about to go into                 
emergency heat mode.  At that point, the system reports heating being on, but the fan is off (briefly).
As a result I chose to call this state "heat_no_fan" but I'm open to other suggestions.

Note that this thermostat doesn't technically support aux mode, though the company
has provided tech support indications that it's OK to tie the emergency heat
and aux wires together, in which case selecting "emergency heat" mode will 
effectively the system into aux mode, and the system can also enter emergency heat mode automatically.  
What I observed was the reported value of running state was different in both of these cases, and different 
from the normal heating case.

As I mentioned in my other centralite PR (https://github.com/Koenkk/zigbee-herdsman-converters/pull/4081) 
I see that there is another almost identical thermostat recently added and the same questions apply - namely 
Whether the configs can be merged and/or whether I should apply my changes, to the newer config.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -57,8 +57,8 @@ const dayOfWeek = {
 
 const thermostatRunningStates = {
     0: 'idle',
-    1: 'heat',
-    2: 'cool',
+    1: 'heat_no_fan',
+    2: 'cool_no_fan',
     4: 'fan_only',
     5: 'heat',
     6: 'cool',
@@ -68,6 +68,7 @@ const thermostatRunningStates = {
     D: 'heat',
     10: 'cool',
     12: 'cool',
+    13: 'aux_heat',
     14: 'cool',
     15: 'cool',
     22: 'cool',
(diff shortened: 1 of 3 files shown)


lib/constants.js
```diff
@@ -57,8 +57,8 @@ const dayOfWeek = {
 
 const thermostatRunningStates = {
     0: 'idle',
-    1: 'heat',
-    2: 'cool',
+    1: 'heat_no_fan',
+    2: 'cool_no_fan',
```
Koenkk, line 61: Please don't change these since this will introduce a breaking change.

nkasrawi: I guess the problem is that means that there's no way of distinguishing the regular heat state with this special heat state.  Is there a way I can implement something specific to the centralite device that interprets this value differently?  The values are different, and the states they represent according to the zigbee spec is different - it's just unfortunate that they were given the same text value in the converters.

PR #4085, opened 2022-04-04
WIP: fix configure problem with haElectricalMeasurement MicroMatic zb250

geirra:
A user reported that a new device will throw an error, tested this and it seems to be correct.

``` log
error 2022-03-31 15:38:59: Failed to configure '0x04cd15fffeb16a98', attempt 3 (Error: Read 0x04cd15fffeb16a98/1 haElectricalMeasurement(["acFrequencyDivisor","acFrequencyMultiplier"], {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')
    at Endpoint.checkStatus (/app/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:317:28)
    at Endpoint.read (/app/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:480:22)
    at Object.readEletricalMeasurementMultiplierDivisors (/app/node_modules/zigbee-herdsman-converters/lib/reporting.js:25:5)
    at Object.configure (/app/node_modules/zigbee-herdsman-converters/devices/micromatic.js:22:13)
    at Configure.configure (/app/lib/extension/configure.ts:115:13))
```
    
By adding a new device i also discovered another potential problem, that I need input on.

The device has energy consumption measurement, but this is not updated in z2m even tough the device sends a report.
Based on my test device this seems to only be updated once it reached 1kwh.
But in this case the device is still at 0,21kwh so z2m is not updated, is this correct behaviour ?

```
Received Zigbee message from 'Sofa', type 'attributeReport', cluster 'seMetering', data '{"currentSummDelivered":[0,2103939]}' from endpoint 1 with groupID 0
```

### DIFF
--- a/devices/micromatic.js
+++ b/devices/micromatic.js
@@ -19,10 +19,6 @@ module.exports = [
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl', 'haElectricalMeasurement', 'seMetering']);
             await reporting.brightness(endpoint);
-            // Unit missing acFrequencyDivisor and acFrequencyMultiplier, so we need to target specific units in haElectricalMeasurement
-            await endpoint.read('haElectricalMeasurement', ['acPowerMultiplier', 'acPowerDivisor', 'acCurrentDivisor']);
-            await endpoint.read('haElectricalMeasurement', ['acVoltageMultiplier', 'acVoltageDivisor', 'acCurrentMultiplier']);
-            // end
             await reporting.readMeteringMultiplierDivisor(endpoint);
             await reporting.rmsVoltage(endpoint, {min: 10, change: 20}); // Voltage - Min change of 2V
             await reporting.rmsCurrent(endpoint, {min: 10, change: 10}); // A - z2m displays only the first decimals, change of 10 / 0,01A
--- a/devices/micromatic.js
+++ b/devices/micromatic.js
@@ -19,7 +19,6 @@ module.exports = [
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl', 'haElectricalMeasurement', 'seMetering']);
             await reporting.brightness(endpoint);
-            await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
             await reporting.rmsVoltage(endpoint, {min: 10, change: 20}); // Voltage - Min change of 2V
             await reporting.rmsCurrent(endpoint, {min: 10, change: 10}); // A - z2m displays only the first decimals, change of 10 / 0,01A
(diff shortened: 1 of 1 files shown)


devices/micromatic.js
```diff
@@ -19,7 +19,6 @@ module.exports = [
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl', 'haElectricalMeasurement', 'seMetering']);
             await reporting.brightness(endpoint);
-            await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
```
Koenkk, line 22: This could potentially screw up the reported measurements. Could you make sure it doesn't by force removing the device via the frontend, restart z2m and repair the device. Are the reported measurements correct?

geirra: @Koenkk is there any better way to avoid the error when including the device? Open for suggestions.

Also, did you see the semetering issue?

Koenkk: Did you check if the measurements are still correct after force removing, restart z2m and repair the device? If yes, it can be removed safely.

About the metering, this is because no divisor/multiplier is available, try adding `await reporting.readMeteringMultiplierDivisor(endpoint);` here.

geirra: Seems like your right @Koenkk , it needs the multipliers to get the correct sum.
My testing has been flawed here. 

Other suggestions to avoid the error messages for users? 
it works but not ideal i think
```await reporting.readMeteringMultiplierDivisor(endpoint); 

seems to have worked for usage also

Comment (geirra) [reply devices/micromatic.js:22]: found some examples for targeting multipliers/divisors that we need. 
Picked the units based on the interview from the logs.

**Comment by Koenkk** [comment]
I think your initial issue has been fixed by https://github.com/Koenkk/zigbee-herdsman-converters/pull/4086 . This would mean that without any changes the device works fine on the latest dev (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html), could you test that?

Then this pr only needs to add `await reporting.readMeteringMultiplierDivisor(endpoint);`

**Comment by geirra** [comment]
sorry for the delay here, did a test on the latest-dev from docker, excluded and rejoined the device. 
With default driver it connected without problems, so i think #4086 fixed the problem. 

So we should include the missing readMeteringMultiplierDivisor, and this could the fixed.
```
``` 

This PR is now a bit messy, should it be closed and make clean one ?

Koenkk:
Not needed, merged it

NOTE: made a small mistake while editing your pr, fixed this in https://github.com/Koenkk/zigbee-herdsman-converters/commit/dffd15d094b95e9f179af3accc82328f56e69a28

PR #3839, opened 2022-02-07
Added support for acFrequency for Develco SPLZB-132

antoinelibert:
Hello, my first PR!

This closes: https://github.com/Koenkk/zigbee2mqtt/issues/11213

- Added support for acFrequency for Develco SPLZB-132
- Add missing `acFrequencyDivisor` and `acFrequencyMultiplier` to `readEletricalMeasurementMultiplierDivisors()`
- Add expose new preset `ac_frequency`

Please note this is based on a custom external converter that I made based on the original one inside this repo. This PR is an updated "good way" version. I don't have access to a full working dev environment so please test before 🤣 .

The custom external converter is attached to the PR for comparison. 

```js
const definition = {
    zigbeeModel: ['SPLZB-132'],
    model: 'SPLZB-132',
    vendor: 'Develco',
    description: 'Power plug',
    fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fz.device_temperature],
    toZigbee: [tz.on_off],
    exposes: [e.switch(), e.power(), e.current(), e.voltage(), e.energy(), e.device_temperature(), exposes.numeric('ac_frequency', ea.STATE).withUnit('Hz').withDescription('Measured electrical AC frequency')],
    options: [
        exposes.options.precision(`ac_frequency`)
    ],
    configure: async (device, coordinatorEndpoint, logger) => {
        const endpoint = device.getEndpoint(2);
        await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement', 'seMetering', 'genDeviceTempCfg']);
        await reporting.onOff(endpoint);
        await reporting.deviceTemperature(endpoint);
        await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
        await reporting.activePower(endpoint, { change: 10 }); // Power reports with every 10W change
        await reporting.rmsCurrent(endpoint, { change: 20 }); // Current reports with every 20mA change
        await reporting.rmsVoltage(endpoint, { min: constants.repInterval.MINUTES_5, change: 400 }); // Limit reports to every 5m, or 4V
        await reporting.readMeteringMultiplierDivisor(endpoint);
        await reporting.currentSummDelivered(endpoint, { change: [0, 20] }); // Limit reports to once every 5m, or 0.02kWh
        await reporting.instantaneousDemand(endpoint, { min: constants.repInterval.MINUTES_5, change: 10 });

[...]

PR #4086, opened 2022-04-04
Fix for commit #3839

smulle48:
This PR fix, what problems commit #3839 introduced.

Not all devices support acFrequencyDivisor and acFrequencyMultiplier.

With the original PR, this would introduce an UNSUPPORTED_ATTRIBUTE error on for example Heiman HS2SK, and no readings could be fetched from the device.

To access the acFrequencyDivisor and acFrequencyMultiplier, the following line should be used in the device:

await reporting.readEletricalMeasurementMultiplierDivisors(endpoint, true);

Do NOT use the following line, because it will break it again:

await reporting.acFrequency(endpoint);

Co-Authored-By: corneyl <5320363+corneyl@users.noreply.github.com>

### DIFF
--- a/devices/develco.js
+++ b/devices/develco.js
@@ -46,14 +46,14 @@ module.exports = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement', 'seMetering', 'genDeviceTempCfg']);
             await reporting.onOff(endpoint);
             await reporting.deviceTemperature(endpoint);
-            await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
+            await reporting.readEletricalMeasurementMultiplierDivisors(endpoint, true); // Set to true, to access the acFrequencyDivisor and acFrequencyMultiplier attribute. Not all devices support this.
             await reporting.activePower(endpoint, {change: 10}); // Power reports with every 10W change
             await reporting.rmsCurrent(endpoint, {change: 20}); // Current reports with every 20mA change
             await reporting.rmsVoltage(endpoint, {min: constants.repInterval.MINUTES_5, change: 400}); // Limit reports to every 5m, or 4V
             await reporting.readMeteringMultiplierDivisor(endpoint);
             await reporting.currentSummDelivered(endpoint, {change: [0, 20]}); // Limit reports to once every 5m, or 0.02kWh
             await reporting.instantaneousDemand(endpoint, {min: constants.repInterval.MINUTES_5, change: 10});
-            await reporting.acFrequency(endpoint);
+            //await reporting.acFrequency(endpoint); // dont use this, because it breaks all devices that dont support it. Look above for correct usage.
         },
         endpoint: (device) => {
             return {default: 2};
--- a/lib/reporting.js
+++ b/lib/reporting.js
@@ -18,11 +18,17 @@ function payload(attribute, min, max, change, overrides) {
     return [payload];
 }
 
-async function readEletricalMeasurementMultiplierDivisors(endpoint) {
+// Fix the problem that commit #3839 introduced.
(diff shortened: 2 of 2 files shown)


devices/develco.js
```diff
@@ -46,14 +46,16 @@ module.exports = [
...
+            await reporting.readEletricalMeasurementMultiplierDivisors(endpoint, true);
             await reporting.activePower(endpoint, {change: 10}); // Power reports with every 10W change
             await reporting.rmsCurrent(endpoint, {change: 20}); // Current reports with every 20mA change
             await reporting.rmsVoltage(endpoint, {min: constants.repInterval.MINUTES_5, change: 400}); // Limit reports to every 5m, or 4V
             await reporting.readMeteringMultiplierDivisor(endpoint);
             await reporting.currentSummDelivered(endpoint, {change: [0, 20]}); // Limit reports to once every 5m, or 0.02kWh
             await reporting.instantaneousDemand(endpoint, {min: constants.repInterval.MINUTES_5, change: 10});
-            await reporting.acFrequency(endpoint);
+            // dont use this, because it breaks all devices that dont support it. Look above for correct usage.
+            // await reporting.acFrequency(endpoint);
```
corneyl, line 58: The `await reporting.acFrequency(endpoint)` reads the actual frequency and should still be in here uncommented, because this Develco plug supports it. This definition is only for this device so it shouldn't break functionality of other devices. `readEletricalMeasurementMultiplierDivisors()` reads the multipliers and divisors, not the actual frequency.

corneyl (commented):
Thanks for the PR! I think there's one change unneeded though.

devices/develco.js
corneyl: Oh wait, the `reporting.acFrequency(endpoint)` is still in there, so then the comment doesn't make too much sense?

smulle48: Ahh yeah, my bad :-)

I'll fix it asap :-)

lib/reporting.js
```diff
@@ -18,11 +18,17 @@ function payload(attribute, min, max, change, overrides) {
     return [payload];
 }
 
-async function readEletricalMeasurementMultiplierDivisors(endpoint) {
+// Fix the problem that commit #3839 introduced.
+// You can set readFrequencyAttrs = true if the device support acFrequencyDivisor and acFrequencyMultiplier
+// See Develco.js SPLZB-132 for example
+async function readEletricalMeasurementMultiplierDivisors(endpoint, readFrequencyAttrs = false) {
```
Koenkk, line 24: This will disable acFrequencyDivisor/Multiplier read for all devices except when explicitly enabled. I would opt for the reverse, by default it is read so `readFrequencyAttrs = false` -> `readFrequencyAttrs = true`

smulle48: Won't that require readFrequencyAttrs = false for all other devices that support the readEletricalMeasurementMultiplierDivisors but don't support acFrequencyDivisor and acFrequencyMultiplier ??

@corneyl found that only 2 devices so far supported the frequency attribute.

Personally I think it's quite a change for "only" 2 devices so far. I mean they probably work even if the attribute is disabled, but the other way around, almost nothing works on all the other devices

corneyl: @Koenkk As far as I can see there are about 19 devices that call `readEletricalMeasurementMultiplierDivisors()` of which only one actually uses the frequency attributes, which is the Develco SPLZB-132 device. So then reading them not by default would be a more logical choice, right?

Also, I'm afraid more than just this Heimann plug have now problems reporting the power usage, see e.g. 
- https://github.com/Koenkk/zigbee2mqtt/issues/11983
- https://github.com/Koenkk/zigbee2mqtt/issues/11943

These issues show the same UNSUPPORTED_ATTRIBUTE error in the logs.

Koenkk: You are right guys, only `SPLZB-132` sets up the ac frequency reporting. Thanks!

PR #4087, opened 2022-04-04
Add _TZ3000_zwaadvus to TS011F_2_gang_wall

jdoe0000000:
This adds support  for _TZ3000_zwaadvus, which appears to be a variant of TS011F_2_gang_wall.

The hardware has a different model number printed on its back: SM-PW801-VZ. I wanted to add an entry for it in `whiteLabel`, but I'm not sure what to put in the `vendor` field since it doesn't have any branding.


Koenkk:
thanks!

PR #4088, opened 2022-04-04
Fix file permissions for devices/home_control_as.js

jdoe0000000:
For some reason, devices/home_control_as.js is the only file in the devices folder to have execute permissions. This removes the execute bit (i.e. change the mode to 644) to match other files in the devices folder.


Koenkk:
Thanks!

PR #4096, opened 2022-04-06
Add Wiser water leakage detector CCT592011

albatorsk:
I don't really know what I'm doing, but by following https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html and doing as best I could I got the CCT592011 to work with zigbee2mqtt.


Koenkk:
thanks!

PR #4101, opened 2022-04-08
Add unsupported Livarno remote control

dumbo0001:
Add support for the Livarno remote control that is included in a Livarno Home LED ceiling light (see https://www.lidl.nl/p/livarno-home-led-plafondlamp-zigbee-smart-home/p100339376)
Same type buttons as the model FB20-002 but different look and layout.


Koenkk:
Thanks!

PR #4108, opened 2022-04-09
Support IKEA E2103

wshuman3:


Koenkk:
Thanks!

tauzN:
I just installed a TREDANSEN bought in Denmark. 
Mine does not expose E2103 over Zigbee.
Mine exposes "TREDANSEN block-out cellul blind"

Have others had the same experience?


A custom external converter with these edits works for me:
```
...
zigbeeModel: ['TREDANSEN block-out cellul blind'],
model: 'TREDANSEN block-out cellul blind',
...
```

wshuman3:
You are right.  It was a copy/paste error from my external converter.  I'll correct it in a new PR.

tauzN:
> You are right. It was a copy/paste error from my external converter. I'll correct it in a new PR.

I think there is a non-block-out version too. Im new to this config.

Would you add something like this? Or how does that work?
```
...
zigbeeModel: ['TREDANSEN block-out cellul blind', 'TREDANSEN cellul blind']
...
```

wshuman3:
It would be a separate model if that is how it is exposed.

tauzN:
I just noticed they have a different name `PRAKTLYSING`. They should be the exact same, except non-block-out.
They should also be added, when someone gets their hands on them 👍


ref.: https://www.ikea.com/dk/da/p/praktlysing-plissegardin-tradlost-batteridrevet-hvid-90514217/

tauzN:
> Do you have one of those? How is it exposed?

'genBasic': {"modelId":"TREDANSEN block-out cellul blind","manufacturerName":"IKEA of Sweden"}

olmozt:
> I just noticed they have a different name `PRAKTLYSING`. They should be the exact same, except non-block-out. They should also be added, when someone gets their hands on them 👍
> 
> ref.: https://www.ikea.com/dk/da/p/praktlysing-plissegardin-tradlost-batteridrevet-hvid-90514217/

Here is a pull request for the non-blockout version `PRAKTLYSING` https://github.com/Koenkk/zigbee-herdsman-converters/pull/4133

PR #4110, opened 2022-04-09
Adding a 3key and Vibration paste of zigbe herdsman coverters

yhwl8888:
Adding a 3key and Vibration paste of zigbe herdsman coverters


devices/ruijia.js
```diff
@@ -0,0 +1,101 @@
...
+                    3: 'arm',
+                };
+                const value = msg.data.armmode;
+                return {action: lookup[value] || `armmode_${value}`};
+            }
+        },
+    },
+	
+
+    REJIA_vibration_alarm_1_with_timeout: {
```
Koenkk, line 41: Looks the same as `ias_vibration_alarm_1_with_timeout`, please use that instead.

PR #4112, opened 2022-04-09
Adding new naming for TRADFRI bulb E27 C/WS opal 600

Eckii24:
When trying to add my Ikea bulb I realised, that it has a different model name.

image

I tried the provided change locally and it works as expected.


Koenkk:
thanks!

PR #4123, opened 2022-04-11
[lidl] Prefer HS for HG08633

jpbede:
Played a little with the Lidl HG08633 - The HS options give better color results, so prefer it. The light also supports the color temperature setting at startup.


Koenkk:
Thanks!

Koenkk:
Thanks!

Issue #2168, opened 2021-01-27
Multiple OTA providers


mkaay:
Hey Koenkk,
Is it possible, to have multiple OTA providers (lib/ota/*.js) that return data for the same manufacturer and image?
And what happens when they return different versions, will the highest version be chosen?
I had a quick look around the code, but I couldn't find the right entry point.
The reason I'm asking is that there is an API for the Hue BT app, that needs no authorization, but it only works for the BT enabled lights.
Would it be possible to have "base" firmware version from the zigbeeOTA.js and potentially newer versions via the API in a new provider?

The API is pretty simple:
https://firmware.meethue.com/v1/checkUpdate?deviceTypeId=100B-11A&version=0
the device type id is just the manufacturer id an image type as hex. You can leave version 0 to get the newest version or use the actual version to check if an update is available

CC @TheJulianJES

Koenkk:
Multiple OTA providers are not supported. For this case a new OTA provider should be created. To do this.

- Create a new OTA provider (use e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/lib/ota/zigbeeOTA.js as inspiration) and add it here: https://github.com/Koenkk/zigbee-herdsman-converters/tree/master/lib/ota , call it `philips.js`
- Add it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/lib/ota/index.js
- In `devices.js` set `ota: ota.philips` for the Hue Bluetooth devices.

mkaay:
Okay, thanks! I will try to think of an elegant solution for this.

PR #4124, opened 2022-04-12
Hue ota

henriknil:
Philips Hue OTA support for selected devices

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -642,7 +642,7 @@ module.exports = [
         description: 'Hue Filament Standard A60/E27 bluetooth',
         meta: {turnsOffAtBrightness1: true},
         extend: hueExtend.light_onoff_brightness(),
-        ota: ota.zigbeeOTA,
+        ota: ota.hue,
     },
     {
         zigbeeModel: ['LCB001'],
@@ -694,7 +694,7 @@ module.exports = [
         description: 'Hue white A60 bulb E27 bluetooth',
         meta: {turnsOffAtBrightness1: true},
         extend: hueExtend.light_onoff_brightness(),
-        ota: ota.zigbeeOTA,
+        ota: ota.hue,
     },
     {
         zigbeeModel: ['LWW003', 'LWF003'],
@@ -901,7 +901,7 @@ module.exports = [
         description: 'Hue white and color ambiance LightStrip plus',
         meta: {turnsOffAtBrightness1: true},
         extend: hueExtend.light_onoff_brightness_colortemp_color(),
-        ota: ota.zigbeeOTA,
+        ota: ota.hue,
     },
     {
         zigbeeModel: ['LCL002'],
@@ -2250,7 +2250,7 @@ module.exports = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff']);
             await reporting.onOff(endpoint);
         },
-        ota: ota.zigbeeOTA,
+        ota: ota.hue,
     },
     {
         zigbeeModel: ['LOM008'],
@@ -2395,9 +2395,9 @@ module.exports = [
         model: '929002241201',
         vendor: 'Philips',
         description: 'Hue white filament Edison E27 LED',
-        ota: ota.zigbeeOTA,
         meta: {turnsOffAtBrightness1: true},
         extend: hueExtend.light_onoff_brightness(),
+        ota: ota.hue,
     },
     {
         zigbeeModel: ['LTV002'],
--- a/lib/ota/common.js
+++ b/lib/ota/common.js
@@ -428,9 +428,9 @@ async function getNewImage(current, logger, device, getImageMeta, downloadImage)
     const download = downloadImage ? await downloadImage(meta, logger) :
         await getAxios().get(meta.url, {responseType: 'arraybuffer'});
 
-    const checksum = (meta.sha512 || meta.sha256);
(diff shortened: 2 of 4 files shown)


henriknil:
This might need some work before merging.
I don't know if it's possible to get support for or devices or possible implementing a fallback to zigbee-OTA.

Koenkk:
Would be really nice if this could be enabled for all devices, what is blocking this? 

Pinging @thejulianjes and @mkaay here since they have previously updated Hue otas to the zigbee-ota repo.

TheJulianJES:
To my knowledge, the ``https://firmware.meethue.com/v1/checkupdate`` endpoint is only used in the Hue Bluetooth app at the moment. This app will no longer be updated/supported soon-ish, as the main Hue app is already able to control supporting Philips Hue lights via Bluetooth (as well as Zigbee if the user has a Hue Bridge v2).
Because of this, the endpoint mentioned above only gives out updates for Bluetooth capable lights.
(I did not yet check which endpoint is used by the main Hue app when only Bluetooth devices are connected. It might be the same one from above.)
All links for OTA images coming from this endpoint start with ``https://firmware.meethue.com/storage/``.

If the Hue Bridge is used with the Hue system (which is required for older non-Bluetooth lights), the Bridge checks for the updates. The Hue Bridge, however, does not use the endpoint mentioned above. Instead, it uses "a more complicated API" which also requires sending "confidential" bridge-specific data. This basically makes it impossible to implement here.
Those links for OTA images seem to all start with ``http://fds.dc1.philips.com/firmware``.

In my experience, the updates from the Hue Bluetooth app (endpoint listed at the top) also sometimes differ from the updates provided via the Hue Bridge (most of the times, the app ones are a bit newer). I suspect this is to ensure stability for the bigger Hue Bridge/Zigbee user base and because the Bluetooth part isn't/wasn't as stable as the Zigbee part.

For ZHA, deCONZ and other users (who need to download all OTA files), it would be nice to have an up-to-date list of **all** Hue OTA files (which zigbee-OTA currently contains). I'm not sure if the approach in this PR would mean that some Hue image links would be "deleted" from the zigbee-OTA repo or at least get out of sync with images used by Z2M (through the mobile endpoint).
Using the mobile endpoint *for some Hue devices* but not for others isn't the best approach in my opinion. The images provided by the Hue Bridge currently contain images for all Hue devices.
At the moment, I would personally prefer to regularly (but "manually") update the Hue images in the zigbee-OTA repo.

@mkaay already touched on this subject here: https://github.com/Koenkk/zigbee-herdsman-converters/issues/2168

If Z2M would support multiple OTA providers (prioritizing newer images), it's possible that a clean solution could be made, as the zigbee-OTA repo could still have links to all Hue images provided by the Hue Bridge, but Z2M could still check for all devices if a (newer) image is available through the mobile endpoint. (Again, note that images provided through the mobile endpoint may differ/be a bit more "experimental" for the smaller user base.)

TL;DR:
- this approach works for some Hue lights (not all but more than in this PR)
- the images from this endpoint are for the Hue Bluetooth app and are -- in rare cases -- newer than the ones provided by the Hue Bridge (potentially more "experimental" for the small user base)
- it's nice for ZHA, deCONZ users to have a list of all Hue images in zigbee-OTA
- multiple OTA providers are not possible in Z2M => this makes it hard to create "a clean solution" for all Hue devices

henriknil:
@TheJulianJES , thanks for the reply. Maybe support for multiple OTA providers is a feasible way forward.
I will think a bit about it.
Is there any more information available on the workings of the bridge endpoint? I assume with a rooted device you could extract the bridge-specific data for the other endpoint?

mkaay:
The Auth for the API is individual for each bridge and I managed to root mine and reverse engineer the encryption and communication.

So we have these options in my opinion:
- Build and host an API proxy
- Automatic pull request on update via cronjob
- update check via API, but manual pull request (like right now)

From a reliability/maintenance standpoint the automatic pull request might the best. Since I also don't know how much traffic this will cause.

henriknil:
@mkaay , sounds like a good idea.
The reason for me to try to implement was that I think there is missing an older firmware for https://www.zigbee2mqtt.io/devices/324131092621.html#philips-324131092621. Is it possible for you to add this to zigbeeOTA?

mkaay:
If you have the imagetype the dimmer is requesting, me or @TheJulianJES can get it for you

henriknil:
I'm not at home but from checking index.json in zigbeeOTA I pretty sure it's imageType=265

TheJulianJES:
Is there any reason you think there's an older firmware OTA image for that dimmer? (Doesn't look like there's an older version than the one currently in the zigbee-OTA repo).

henriknil:
> Is there any reason you think there's an older firmware OTA image for that dimmer? (Doesn't look like there's an older version than the one currently in the zigbee-OTA repo).
> 
> Previously discussed: [Koenkk/zigbee2mqtt#10139 (comment)](https://github.com/Koenkk/zigbee2mqtt/issues/10139#issuecomment-1030628109) (ZGB_100B_0109/image type 265)

It was only an assumption due to that firmware upgrades doesn't seem to work and for two other imagetypes adding an older version was the solution. But since you now checked the available version for me I will do some further investigation and possible get back with comments in [Koenkk/zigbee2mqtt#10139].
I guess some sniffing will be needed :)

mkaay:
I already have such a script for the bridge API :)

PR #4125, opened 2022-04-13
Power on behavior for ZBMINI-L

kirovilya:


Koenkk:
Thanks!

Daantm:
My sonoff zbmini stopped working, after updage 1.25.1. Now reverted to 1.25.0, which works fine. Running in combination with mosquitto broker

PR #4127, opened 2022-04-15
ZM25TQ add data_point_test and remove speed ctrl

Gimlir:
Following to issue  "TS0601 _TZE200_fzo2pocs (Zemismart ZM25TQ) - position and motor speed issue" #11164
Separated definition of Zemismart ZM25TQ from big Tuya section with added tz.tuya_data_point_test capability and removal of speed control not supported by this device.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1084,7 +1084,6 @@ module.exports = [
             // Window pushers:
             {modelID: 'TS0601', manufacturerName: '_TZE200_g5wdnuow'},
             // Tubular motors:
-            {modelID: 'TS0601', manufacturerName: '_TZE200_fzo2pocs'},
             {modelID: 'TS0601', manufacturerName: '_TZE200_5sbebbzs'},
             {modelID: 'TS0601', manufacturerName: '_TZE200_zuz7f94z'},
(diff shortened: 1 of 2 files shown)


Koenkk:
Is the motor speed also not supported via the TuYa gateway? If it is, this can probably be supported if we found out the data points for it: https://www.zigbee2mqtt.io/advanced/support-new-devices/03_find_tuya_data_points.html#requirements-and-caveats

Foroxon:
I believe this device support just one speed. And "25" in the model name is the speed itself. Please check the attached photo.


ZM25TQ

Koenkk:
I see, please move it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/zemismart.js and change the vendor to `Zemismart`

Gimlir:
Confirmed with seller from Aliexpress (Store: Zemismart Direct Store, looks like relevant source of information) - this motor is fixed speed.

Gimlir:
> I see, please move it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/zemismart.js and change the vendor to `Zemismart`

Done

Koenkk:
Thanks!

PR #4129, opened 2022-04-15
fix swtich dont work Touch switch 1 gang (with power monitoring) XF-E…

DavidFFerreira:
…U-S100-1-M

fix swtich dont work Touch switch 1 gang (with power monitoring) XF-EU-S100-1-M


Koenkk:
thanks!

PR #4131, opened 2022-04-17
Add support for the 7602031U7 variation Philips Hue Go with BT

pathartl:
There is a variation of the Philips Hue Go with Bluetooth with the model number 7602031U7 (previous model was 7602031P7). This PR simply adds a Zigbee Model for the U7.


Koenkk:
thanks!

Issue #4040, opened 2022-03-24
Add _TZE200_rmymn92d to // Contains all covers which need their position inverted by default List


SergeantPup:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/d4d69219b132984d72dbe021381ad441a88e381b/lib/tuya.js#L167

Update line 167 from:
`const coverPositionInvert = ['_TZE200_wmcdj3aq', '_TZE200_nogaemzt', '_TZE200_xuzcvlku', '_TZE200_xaabybja'];` to
`const coverPositionInvert = ['_TZE200_wmcdj3aq', '_TZE200_nogaemzt', '_TZE200_xuzcvlku', '_TZE200_xaabybja', '_TZE200_rmymn92d'];`

I own models _TZE200_xaabybja AND _TZE200_rmymn92d and I can confirm that they both need to be reversed.  Please add _TZE200_rmymn92d to the reversal list of covers.  When I added _TZE200_rmymn92d to the inverted list, the device accurately tracks state.

SergeantPup:
Opened in wrong area.  I possess all the requirements for a PR and will post it there.

SergeantPup:
Turns out, I'm not great at submitting a PR and I should leave it to the experts.  Reopening this issue but it also contains the desired code change

Koenkk:
added.

Changes will be available in the dev branch in a few hours from now. (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html)

leroyloren:
Could you invert for this device _TZE200_gubdgai2 too? Please.

![obrazek](https://user-images.githubusercontent.com/57643470/163726691-24f9ddf8-6109-46e6-9714-b07fe870d99e.png)

At this moment is real open.

PR #4143, opened 2022-04-21
Invert position _TZE200_gubdgai2

leroyloren:
https://github.com/Koenkk/zigbee-herdsman-converters/issues/4040#issuecomment-1100924941


leroyloren:
@Koenkk yes, true.

obrazek

Koenkk:
Then it is already correct, so no need to invert.

leroyloren:
I don't know if we understand each other, but according to the attached picture it is 0 = state CLOSED, but the blind is open at that moment

In the used automations it is necessary to use the inverted state.

Koenkk:
I see, thanks for double checking.

leroyloren:
Thanks 👍

PR #4151, opened 2022-04-23
Add support for additional smart knob variant.  Looks identical, but has new name _TZ3000_ixla93vd.

eric-timeit:


eric-timeit:
Add new smart knob variant.

Koenkk:
Thanks!

PR #4152, opened 2022-04-23
Add child_lock and led_enable support to the Niko Outlets

sjorge:
I also briefly had LED color changing working on 1 of the 2 of these devices I owned but then it crashed and I had to reset it.
And could not get it to work after reliably. I might revisit that at a later date but I'm out of ideas for now on that one.


Koenkk:
Thanks!

PR #4154, opened 2022-04-23
Namron Touch thermostat: Added support for custom attributes

owangen:
PR includes some bug fixing, and implementation of all custom attributes this thermostat supports


owangen:
Don't understand why unittest is failing after last commit...

Koenkk:
Thanks!

PR #4157, opened 2022-04-24
Added _TZ3210_4ubylghk

Tozapid:
Good day,

I bought a dimmer `QS-Zigbee-D02-TRIAC-2C-L` -  https://www.aliexpress.com/item/1005002821229566.html?spm=a2g0o.order_list.0.0.59b4d2d6Xl93WZ

But according to the scan, he has a different model. It seems that just adding a model is enough and everything will work.

### DIFF
--- a/devices/lonsonho.js
+++ b/devices/lonsonho.js
@@ -127,7 +127,7 @@ module.exports = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #4160, opened 2022-04-25
Update danfoss.js

fraggle-iw:
Hi,
according to the Danfoss documentation the default value für load_estimate is -8000 therfore is added min and max for this expose.

### DIFF
--- a/devices/danfoss.js
+++ b/devices/danfoss.js
@@ -79,7 +79,8 @@ module.exports = [
                 .withDescription('Mean radiator load for room calculated by gateway for load balancing purposes (-8000=undefined)')
                 .withValueMin(-8000).withValueMax(2000),
             exposes.numeric('load_estimate', ea.STATE_GET)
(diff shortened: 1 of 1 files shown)


devices/danfoss.js
```diff
@@ -79,7 +79,8 @@ module.exports = [
                 .withDescription('Mean radiator load for room calculated by gateway for load balancing purposes (-8000=undefined)')
                 .withValueMin(-8000).withValueMax(2000),
             exposes.numeric('load_estimate', ea.STATE_GET)
-                .withDescription('Load estimate on this radiator'),
+                .withDescription('Load estimate on this radiator')
+                .withValueMin(-8000).withValueMax(65535),
```
Koenkk, line 83: Does the max value of 65535 make sense?

fraggle-iw: You are right. I read the wrong line in the documentation. Max should be 3600.

Koenkk:
Thanks!

PR #4161, opened 2022-04-25
Change max temp from 32 to 35 for Danfoss Ally

microraptor:
The technical specification lists a temperature range from 5 to 35 deg celsius. See here (MaxHeatSetpointLimit): https://assets.danfoss.com/documents/193613/AM375549618098en-000102.pdf
Also the user guide says the same and my personal device with the latest firmware goes up to 35.


Koenkk:
Thanks!

PR #4164, opened 2022-04-27
Ignore Xiami Basic state for ZNMS12LM

McGiverGim:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/12279

The sensor publishes `"100": 44,` that is being mapped to `STATE: OFF` by default, that is not correct. We ignore it until we know what this `44` implies.


Koenkk:
Thanks!

PR #4165, opened 2022-04-27
Update giderwel.js

Pla5ma:
Added A11 smart LED strip.


Pla5ma:
Can be closed. I will add another commit as I found several additional device functions.

PR #4166, opened 2022-04-27
Giderwel A11 LED strip support

Pla5ma:
Support for Giderwell A11 strip. Single ID white and color LED strip like the Gledopto. Taken large chunks of code from those devices and changed as needed. Tested on real devices.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -1826,6 +1826,79 @@ const converters = {
             }
         },
     },
+    giderwel_light_onoff_brightness: {
+        key: ['state', 'brightness', 'brightness_percent'],
+        options: [exposes.options.transition()],
+        convertSet: async (entity, key, value, meta) => {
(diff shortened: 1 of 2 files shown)


Koenkk:
What doesn't work correctly with the standard converters?

Pla5ma:
It is white and color LED stripe which is controlled by one ID. I tried with the standard converters but the white LEDs where not controllable in any combination. Searching for a possible solution I found that the used chipset and firmware is the same as the already implemented Gledopto LED stripe. So i took everything from Gledopto which was necessary, e.g. left some unused functions out which are used by other Gledopto stripes and changed the naming to fit in to the convention. That means I haven't really tested it to the end because I thought if it was already implemented that way for the same firmware and chipset there must be a reason.

Koenkk:
Could you switch to the standard converters step by step? I don't think all changes are needed to just fix the white not being controllable.

Pla5ma:
I do not have the devices anymore. I installed them at a friends home.

PR #4174, opened 2022-04-28
Add Philips Hue Runner triple model 929003046201

laurigates:
This model was missing. I'm not sure if adding the ZigBee models to the existing Philips Hue White Ambiance Runner triple spotlight entry is better or keeping it separate like this since it has a different model number.

https://www.philips-hue.com/en-gb/p/hue-white-ambiance-runner-triple-spotlight/8719514338388


Koenkk:
Thanks!

PR #4175, opened 2022-04-28
JM-TRH-ZGB-V1 new Temperature & humidity sensor with clock

kirovilya:
https://aliexpress.ru/item/1005003980647546.html
named as JM-TRH-ZGB-V1 in Tuya iot cloud

Extended the similar device converters to receive and send additional data


Koenkk:
Thanks!

PR #4180, opened 2022-04-29
Fantem ZB003-X: Expose new features and fixes

AlexKay88:
**Fantem ZB003-X:** 
**Expose new features:** `battery2` remaining percentage, `battery_voltage`
**Fix:** First battery values have been overwritten with `battery2` values
**Changes:** `keep_time` now from 0 to 480 seconds selectable
**Changes:** `illuminance_calibration, temperature_calibration, humidity_calibration` min/max range values same as Tuya app shows

**Fantem ZB006-X:
Fix:** `configure` error caused by `await reporting.onOff(endpoint);`


Koenkk:
I don't think it makes sense to have `battery2` and `battery`, can these be merged?

AlexKay88:
It makes sense, because the device can be operated with 1 or 2 batteries and both are measured separately. Battery1 (left) is mandatory for operation. Battery 2 (right) can be removed (results e.g in longer reporting interval)
Also Tuya app shows them separately along with percentage remaining.
Why not show `battery1 %` and `battery2 %` to clearly state which one has to be replaced if empty?

Koenkk:
Didn't know this device has 2 batteries, then it makes sense indeed, thanks!

AlexKay88:
THX for merging!
@Koenkk May I ask you for a favour? For better Home Assitant integration `battery2` needs `device_class: battery`.
In https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/extension/homeassistant.ts
is only one more line needed. Copy line
`battery: {device_class: 'battery', entity_category: 'diagnostic', state_class: 'measurement'},`
to
`battery2: {device_class: 'battery', entity_category: 'diagnostic', state_class: 'measurement'},`
Would that be possible without a new PR?

PR #4195, opened 2022-05-02
error inclusion: Dual Wireless Remote Legrand (067774)

apages2:
deletion of the creation of the reporting genPowerCfg endpoint2 for the Wireless double remote switch device because it does not exist and the inclusion is not done completely (purple light and not green)


Koenkk:
Thanks!

PR #4200, opened 2022-05-04
change expose to match true function (067771 and 067772)

apages2:
bticino L441C/N4411C/NT4411C and K4003C/L4003C/N4003C/NT4003C

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -5385,7 +5385,7 @@ const converters = {
                     payload.device_mode = 'unknown';
                 }
             }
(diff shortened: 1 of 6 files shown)


Koenkk:
What is the reason for this change?

apages2:
Hello,

In fact the "Permanent LED" option does not exist.

These are the "LED ON if light ON" option and the other "LED ON if light OFF" option.

Activating these two options simultaneously brings the "permanent LED" feature ;-)

it's just a story of semantics, but it's to be precise and more understandable

best regards

Koenkk:
Thanks!

PR #4203, opened 2022-05-04
New Device TS011F _TZ3000_7issjl2q  (model ATMS1601Z)

prokudin07:
That is other model of that device - [TuYa TS011F_din_smart_relay](https://www.zigbee2mqtt.io/devices/TS011F_din_smart_relay.html)
This is model without power monitoring and relay with dry contact
Model: ATMS1601Z
image


Koenkk:
Thanks!

PR #4205, opened 2022-05-04
Update kwikset.js

AstleyTomlinsonJr:
Model number 66492-001 seems to be very similar, if not identical to model 99140-139.  Added ['SMARTCODE_CONVERT_GEN1_W3'] which is what my model 99140-139 shows as zigbeeModel when I paired to Z2M, though it looks exactly like to 66492-001 ['SMARTCODE_CONVERT_GEN1'].  All working for me now with this change when applied as an external converter.


Koenkk:
Thanks!

PR #4208, opened 2022-05-04
Update device TuYa TS0601_smart_human_presense_sensor

prokudin07:
Parent device - [TuYa TS0601_smart_human_presense_sensor](https://www.zigbee2mqtt.io/devices/TS0601_smart_human_presense_sensor.html)
This is device looks external view  how other device, but
exposes the same  how parent device - only other manufactured name.
image


Koenkk:
Thanks!

Issue #1097, opened 2020-03-26
light.toggle(on) with a transition attribute set (after light.turn_off) don't set brightness to correct previous level 


htvekov:
Hi' Koenkk.

I'm quite sure this issue is related to the other, newly fixed, brightness issue #1073 
It's somewhat the 'opposite' error of what was previously fixed.

Actually Xavi, the developer of ControllerX app, found this bug after new tests with the updated z2m-edge:

Environment:
Hass.io - Home Assistant 0.107.7
zigbee2mqtt-edge (but also on latest z2m-stable version)
Ikea White Spectrum bulb on z2m integration

Case:
- Turn on the light from HA (service calls)
- Change the brightness to 50%
- Turn off the light from HA (service calls)
- Call light.toggle service with transition = 0.3 (e.g.)

Then the brightness is NOT set back to 50%, but to a 'different' brightness.
This can be repeated over and over again with same result.

There's actually a pattern to the 'brightness set error'. The brightness set is ALWAYS the PREVIOUS brightness, until an lights off or toggle off is called with a transition attribute set. Then this brightness becomes new 'previous brightness' (Hope this somewhat difficult sentence makes any sense ?)

However, if I remove the "transition" attribute, I do not experience this. Then light always turns on at correct previous brightness.

ControllerX Github issue reference: https://github.com/xaviml/controllerx/issues/58

Hope you can fix this as well, Koenkk 🙏😃🙋‍♂️

Ciao !

Koenkk:
I cannot reproduce this but probably I'm using slightly different commands. Could you provide the log when reproducing this bug and running with `DEBUG=zigbee:herdsman:controller*`?

Some background on why you get these bugs when using transition: Zigbee does not have commands to turn on/off lights with a transition (besides one with a fixed 0.8 transition). Therefore we turning a bulb off with a transition, we set the brightness to 0. When the bulb is turned on again we need to restore this transition.

htvekov:
Hi' Koenkk.

OK. Wasn't aware of that.

Thought that transition was so low level, that it was handled completely by bulbs firmware.
As I understand your answer, Ikea bulbs firmware defaults to transition: 0.8,  if no attribute is specified ? And remaining transitions are 'emulated' by zigbee converter, by transitioning to/from brightness: 0 and then turn on/off lights ? Clever ! 👍😎

Anyway I'm on hassio and tried googling how to enable herdsman logging ?
[...]

PR #4211, opened 2022-05-05
Rewrite the level_onoff_brightness converter

pyos:
Hopefully the comments should make the reasoning behind the code clearer.

This commit fixes two issues:

  1. `toggle` without transition after `off` with transition sets brightness to 1 (so #1097, but with `"state": "toggle"` instead of `"state": "on"`);

  2. `OnLevel` attribute is not respected by `on` with transition (which instead always restores the light to the previous level).

It also adds a new feature: publishing `{"brightness": ..., "state": null}` allows setting brightness without affecting state.

  * If the light is on and new brightness is 0, it will stay on, but switch to minimum brightness.

  * If the light is off and new brightness is not 0, the light might ignore the command or remember the brightness for the next time it is turned on (if `Options` has bit 0 set - see Koenkk/zigbee-herdsman#534).

Known remaining issues:

  * Sending "state": "on"/"off" with brightness or transition to a group may or may not affect the state of devices that don't support brightness. E.g. [Envilar ZG302-BOX-RELAY](https://www.zigbee2mqtt.io/devices/ZG302-BOX-RELAY.html) reacts to MoveToLevelWithOnOff commands despite not implementing the Level Control cluster, while the [LEDVANCE SMART+ plug](https://www.zigbee2mqtt.io/devices/AC10691.html) does not.

  * `OnLevel` is still ignored if the entity is a group.

  * When `on` has to use `MoveToLevelWithOnOff` due to previous `off` having used it too, the transition is 0 rather than `OnTransitionTime`.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -798,7 +798,7 @@ const converters = {
             const {message} = meta;
             const transition = utils.getTransition(entity, 'brightness', meta);
             const turnsOffAtBrightness1 = utils.getMetaValue(entity, meta.mapped, 'turnsOffAtBrightness1', 'allEqual', false);
-            let state = message.hasOwnProperty('state') ? message.state.toLowerCase() : undefined;
+            let state = message.hasOwnProperty('state') ? (message.state === null ? null : message.state.toLowerCase()) : undefined;
             let brightness = undefined;
             if (message.hasOwnProperty('brightness')) {
                 brightness = Number(message.brightness);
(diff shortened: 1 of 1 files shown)


pyos:
Also, found another bug in the current implementation thanks to these tests - when `{state: "on"}` is sent after `{state: "off", transition: any}`, the lamp turns on immediately rather than with a transition of OnTransitionTime.

Koenkk:
Reviewed the code now and this looks good. Would it be possible to fix the remaining 2 test cases? The use case for `Should turn device off when brightness 0 is send` is e.g. when you have a dimmer which only sends brightness; without this feature the dimmer is not able to turn the bulb off.

pyos:
The difference in these tests is how the device turns on afterwards - currently sending `{state: "on"}` (without transition) after `{brightness: 0}` turns the device on at minimum brightness, while `{state: "on", transition: 0}` restores the old brightness. With this PR the brightness is restored regardless.

Koenkk:
I see now, thanks!

PR #4214, opened 2022-05-05
Namron touch thermostat: Fix bug DisplayAutoOffEnable

owangen:


converters/fromZigbee.js
```diff
@@ -1699,7 +1699,7 @@ const converters = {
                 result.hysterersis = precisionRound(data[0x100A], 2) / 10;
             }
             if (data.hasOwnProperty(0x100B)) { // DisplayAutoOffEnable
-                const lookup = {0: 'enable', 1: 'disable'};
+                const lookup = {0: 'enable', 1: 'disabled'};
```
Koenkk, line 1702: to be consistent; should `enable` be renamed to `enabled`?

owangen: Yes, I agree. Will fix after weekend

PR #4217, opened 2022-05-06
IKEA bulbs also need color_options.execute_if_off restored

sjorge:
Merge after https://github.com/Koenkk/zigbee2mqtt/pull/12407

### DIFF
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -13,11 +13,15 @@ const {
     calibrateAndPrecisionRoundOptions, postfixWithEndpointName, getMetaValue,
 } = require('../lib/utils');
 
-const bulbOnEvent = async (type, data, device) => {
+const bulbOnEvent = async (type, data, device, options, state) => {
     /**
      * IKEA bulbs lose their configured reportings when losing power.
      * A deviceAnnounce indicates they are powered on again.
      * Reconfigure the configured reoprting here.
+     *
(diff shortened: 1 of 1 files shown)


sjorge:
Testing:
- removed my quirks extension that did the restore
- restart z2m with the changes from the z2m pr applied and these zhc changes applied
- pair a fresh (used before, but not paired device) GU10 IKEA spot
- turn off the bulb
- set color_temp
- turn on the bulb and note the ct did not change (expected)
- set color_options.execute_if_off to true
- turn off the bulb
- set color_temp
- turn on the bulb and note the ct did change (expected)
- disconnect the power to the bulb
- wait 10 seconds
- restore power to the bulb
- turn off the bulb
- set a different color_temp
- turn on the bulb and observe the color_temp did change (expected)

With this change my quirks extension is no longer needed.

Additional test:
- restore z2m and zhc to upstream dev and master branches
- double check my quirks extension is still missing
- cut power to the bulb
- restore power to the bulb
- turn off the bulb
- set color_temp
- turn on the bulb and verify ct did not change (expected)
- check cached state's color_options.execute_if_off == true (as expected)
- manually read `options` attribute of `ligthingColorCtrl`
- check that the (cached) state's color_options.execute_if_off == false (as expected, as the value on the device got lost when the power was cut)

Koenkk:
Thanks!

PR #4219, opened 2022-05-07
Zg204 zl luminance motion sensor

jaspernlp:
Added support for the ZG-204ZL Luminance Motion Sensor


Koenkk:
Thanks!

vantus1987:
Good afternoon! Am I correct in understanding that there is currently no fully-functional support for this device?

PR #4220, opened 2022-05-07
Add on_level and execute_if_off to level_config

pyos:
on_level: the brightness at which the light turns on. Differs from current_level_startup in that the latter is specifically for when the device is connected to power. Values: "previous", 1..254.

execute_if_off: whether setting `{brightness: VALUE, state: null}` should have an effect even while current state is off. If so and on_level is "previous", the next time the light is turned on it will use the new brightness; otherwise it keeps the brightness set the last time it was on. Values: true, false.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -535,12 +535,41 @@ const converters = {
                 Object.assign(state, {current_level_startup: startUpCurrentLevelValue});
             }
 
+            // onLevel - range 0x00 to 0xff - optional
+            //           Any value outside of MinLevel to MaxLevel, including 0xff and 0x00, is interpreted as "previous".
+            if (value.hasOwnProperty('on_level')) {
+                let onLevel = value.on_level;
+                if (typeof onLevel === 'string' && onLevel.toLowerCase() == 'previous') {
+                    onLevel = 255;
+                } else {
+                    onLevel = Number(onLevel);
+                }
+                if (onLevel > 255) onLevel = 254;
+                if (onLevel < 1) onLevel = 1;
+                await entity.write('genLevelCtrl', {onLevel}, utils.getOptions(meta.mapped, entity));
+                Object.assign(state, {on_level: onLevel == 255 ? 'previous' : onLevel});
+            }
+
+            // options - 8-bit map
+            //   bit 0: ExecuteIfOff - when 0, Move commands are ignored if the device is off;
+            //          when 1, CurrentLevel can be changed while the device is off.
+            //   bit 1: CoupleColorTempToLevel - when 1, changes to level also change color temperature.
+            //          (What this means is not defined, but it's most likely to be "dim to warm".)
+            if (value.hasOwnProperty('execute_if_off')) {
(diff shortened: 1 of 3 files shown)


converters/toZigbee.js
```diff
@@ -854,19 +883,26 @@ const converters = {
...
-                    const current = utils.getObjectProperty(meta.state, 'brightness', 254);
-                    brightness = globalStore.getValue(entity, 'brightness', current);
-                    try {
-                        const attributeRead = await entity.read('genLevelCtrl', ['onLevel']);
-                        // TODO: for groups, `read` does not wait for responses. If it did, we could still issue a single
-                        //  command if all values of `OnLevel` are equal, or split into one command per device if not.
-                        if (attributeRead !== undefined && attributeRead['onLevel'] != 255) {
-                            brightness = attributeRead['onLevel'];
+                    if (!meta.state.hasOwnProperty('level_config')) {
+                        await converters.level_config.convertGet(entity, 'level_config', meta);
```
Koenkk, line 888: If I understand correctly, this will read the `level_config` for all bulbs when turned on; are we sure that all bulbs support this?

pyos: This code only executes when we can't use On because CurrentLevel got clobbered or we need a custom transition duration. MoveToLevelWithOnOff does not default to OnTransitionTime, so we have to read it manually.

>are we sure that all bulbs support this?

level_config.convertGet ignores any unsupported attributes, so we'll just default to on_transition_time = "disabled", on_level = "previous", and on_off_transition_time = 0 if any of them are unsupported.

pyos: This code is somewhat questionable though, because
1. if you manually set part of level_config (e.g. execute_if_off), this code will never read the rest of it and assume other attributes are unsupported;
2. the read only takes effect the *second* time its results are needed, I'm guessing because `meta.state` we already have is immutable and not affected by the read
Not sure what to do about that.

pyos: Fixed that, I think. Also disabled reading level_config for groups, since it's not merged from members' states anyway.

Koenkk: > MoveToLevelWithOnOff does not default to OnTransitionTime, so we have to read it manually. If level_config.execute_if_off is enabled we can probably use MoveToLevel + On instead, though in case of custom transition OnLevel still needs to be read. 

What about not caring about the default transition time in the bulb? You can set a default transition time in the zigbee2mqtt config like this; https://www.zigbee2mqtt.io/devices/046677476816.html#options

pyos: I suppose that's an option too. In that case `transition.specified` is simply always true. I suspect it'd interact poorly with binding though - not only do you have to sync the attribute and the config option, turning the bulb off via z2m then on via a bound remote will consistently set it to minimum brightness because all z2m off commands will have a transition.

pyos: Ok, I guess it's not really necessary.

Koenkk:
Thanks again!

PR #4222, opened 2022-05-07
Add information to the description about the temporarily unsupported `trigger indicator` option for `RTCGQ14LM`

Otnow:
Judging by the screenshot of the sensor settings from the _Aqara Home_ app, there is also the `trigger indicator` option, the support of which has not yet been implemented, so we make an appropriate note in the description.

Motion-Sensor P1

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -919,7 +919,7 @@ module.exports = [
         model: 'RTCGQ14LM',
         vendor: 'Xiaomi',
         whiteLabel: [{vendor: 'Xiaomi', model: 'MS-S02'}],
-        description: 'Aqara P1 human body movement and illuminance sensor',
+        description: 'Aqara P1 human body movement and illuminance sensor (trigger indicator not supported for now)',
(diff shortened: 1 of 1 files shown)


Koenkk:
Do you have the Xiaomi gateway, and if yes, could you sniff the traffic when sending this? Shouldn't be too hard to implement.

Otnow:
The fact of the matter is that I have everything except the sensor itself, and I [added](https://github.com/Koenkk/zigbee2mqtt/issues/11990#issuecomment-1083365730) its basic support by analogy with **RTCGQ12LM** and **RTCGQ13LM**, with the help of real sensor owners.

If I had a sensor, then support for the `trigger indicator` option would have already been implemented 🙂

Koenkk:
i see, I think it would be better to add this info to the device page; if we do this for all features that are not supported the description will get cluttered over time.

Otnow:
Ok, created a [PR](https://github.com/Koenkk/zigbee2mqtt.io/pull/1377) to add information about feature support to the documentation.

PR #4232, opened 2022-05-11
Update tuya.js

RogerioVidigal:
I've included Model "_TZE200_di3tfv5b" to line 595 and fixed the Model name for the previous in the same line... It was missing the _ before the name "_TZE200_k6jhsr0q"


Koenkk:
Thanks!

PR #4237, opened 2022-05-11
Add OTA support, `battery` initialization and other improvements for `Xiaomi MCCGQ14LM`

Otnow:
Changes based on [log](https://github.com/Koenkk/zigbee2mqtt/issues/12372#issuecomment-1124179229) analysis provided by @Noltari

`e.tamper()` has been removed because the sensor does not physically have a tamper.

PR #4245, opened 2022-05-15
Update xiaomi.js

nukusinji:
no need fz.battery
exactly battery level comes fz.aqara_opple.

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -45,14 +45,10 @@ module.exports = [
         model: 'MCCGQ14LM',
         vendor: 'Xiaomi',
         description: 'Aqara E1 door & window contact sensor',
-        fromZigbee: [fz.ias_contact_alarm_1, fz.aqara_opple, fz.battery],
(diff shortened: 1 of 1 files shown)


Otnow:
```javascript
        fromZigbee: [ ... fz.battery],

            await endpoint.read('genPowerCfg', ['batteryVoltage']);
        },
```

I [added this code](https://github.com/Koenkk/zigbee-herdsman-converters/pull/4237) and it is needed to initialize the battery value during pairing, so please close your PR, because it degrades functionality and carries no payload.

nukusinji:
> 

I think no need read when pairing. because can get initial battery value when pairing from fz.aqara_opple.

Otnow:
> I think no need read when pairing. because can get initial battery value when pairing from fz.aqara_opple.

Unfortunately, this is not always the case, which is why this additional code has been added, the presence of which does not interfere with the basic functionality at all, but only complements it.

nukusinji:
> 

okay, i close

Otnow:
> okay, i close

Thanks for understanding.

PR #4247, opened 2022-05-15
ubibsys: power_on_behavior is working on firmware 2.0.0+

sjorge:
I reported issues with about a year ago and they got back to me it got fixed in the latest firmware, it is also listend in the change log so it's now safe to expose this.

> Bug-fix: Startup behavior of the OnOff cluster might be incorrect due to interactions with Level Control or Color Control clusters on the same application endpoint
-- https://www.ubisys.de/en/support/firmware/changelog-s1/

I've also tested that this now seems to with on my S1 and S2 devices.


Koenkk:
Thanks!

PR #4249, opened 2022-05-15
added ROB_200-050-0 and SR-ZG9023A-EU

corchat:
the ROB_200-050-0 is a rebranded SR-ZG9023A-EU
for lack of knowledge I used the same definitions in both files.
Is there a better way to do this (create the definition in the sunrise file end link to it in the robb file)


Koenkk:
Thanks!

PR #4256, opened 2022-05-17
Add TuYa CX-0726 Temperature and humidity sensor

koenhendriks:
Bought this from https://nl.aliexpress.com/item/1005003637487200.html and it was not supported. Using the 'Support new devices' guide added this to TuYa devices.


koenhendriks:
Apparently already added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/4212

PR #4260, opened 2022-05-18
Add support for Legrand Binary Input On Off (NLIS)

wbx13:
When pressing on a button of the NLIS switch, the switch immediately sends a genOnOff attributeReport, then a second one around 5 seconds later.
Between these two messages, every time one of the two buttons is pressed, genBinaryInput messages are sent to report the new state rather than genOnOff.

Adding this allows the two endpoints states to be updated instantly without having to wait for the next (and delayed) genOnOff message.


Koenkk:
Thanks!

PR #4261, opened 2022-05-18
Add support for Centralite HA Thermostat 3156105

bigun27:
Add support for this thermostat: [https://www.amazon.com/dp/B00IUN4K3C](https://www.amazon.com/dp/B00IUN4K3C)

### DIFF
--- a/devices/centralite.js
+++ b/devices/centralite.js
@@ -217,6 +217,36 @@ module.exports = [
         },
     },
     {
+        zigbeeModel: ['3156105'],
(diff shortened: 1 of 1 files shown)


bigun27:
For reference, other Centralite devices that are supported have setpoint ranges starting at 10C, but this device supports down to 44F (6.66C) which is why I selected that number.
Should the minimums be set back to 10C to match the others (3157100 and 3157100-E)?

Koenkk:
Thanks!

Issue #4257, opened 2022-05-17
The file fromZigbee.js is too large to be indexed by GitHub search


wixoff:
The size of the file is 396 KB, and per GitHub documentation the maximum size for search indexing is 384 KB.  See https://docs.github.com/en/search-github/searching-on-github/searching-code#considerations-for-code-search.

Background: I have been struggling to search for the root cause behind https://github.com/Koenkk/zigbee2mqtt/issues/12230.  A code search for the string "action_user" turned up no hits in the  zigbee2mqtt, zigbee-herdsman, or zigbee-herdsman-converters repositories.  It turns out the hits would have been in `fromZigbee.js` which isn't currently indexed.

The file `toZigbee.js` is approaching this threshold size as well.

Can we consider a way to split these files?

Koenkk:
> Can we consider a way to split these files?

Best way to do this is to move some manufacturer specific converters into the manufacturer file (e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/2da3d247108773e69d87310ddd8e72e7232fb370/devices/tuya.js#L48)

PR #4262, opened 2022-05-18
Move some of the ubisys only converters to devices/ubisys.js

sjorge:
As mentioned in #4257 converters/*Zigbee.js files are getting rather big and maybe some stuff should be split off.

I've moved all the ubisys specific stuff to devices/ubisys.js. After this I did a the following tests to check if everything was still good:
- recallibrated my office blinds (J1)
- reconfigure my office light switch button actions (S1)
- reconfigured my office blinds buttons (J1)


Koenkk:
Saves a lot, thanks!

PR #4171, opened 2022-04-28
Add OTA support, fix `battery` and other improvements for `WSDCGQ12LM`

Otnow:
Fixes Koenkk/zigbee2mqtt#9952

Sensor sends heartbeat reports through the _aqaraOpple_ cluster, not _genBasic_:

```javascript
cluster 'aqaraOpple', data '{"247":{"data":[1,33,238,11,3,40,24,4,33,168,19,5,33,43,0,6,36,2,0,0,0,0,8,33,28,1,10,33,0,0,12,32,1,100,41,183,9,101,41,149,18,102,41,224,3],"type":"Buffer"}}'
Processed buffer into data {"1":3054,"3":24,"4":5032,"5":43,"6":2,"8":284,"10":0,"12":1,"100":2487,"101":4757,"102":992}
```

PR #4267, opened 2022-05-20
Fix duplication temperature/device_temperature SSM-U01

McGiverGim:
The SSM-U01 now is publishing two sensors (`temperature` and `device_temperature`) that get the data from the same "physical" sensor. So it has a duplication in the sensors published.

image

The SSM-U01 does not publish the `temperature` (cluster `msTemperatureMeasurement`), in place it publishes the `device_temperature` (cluster `genDeviceTempCfg`). It publishes the same value too in the `aqaraOpple` cluster, attribute `3`.

This PR removes the `temperature` converter from the device, letting only the `device_temperature` one, and modifies the `aqara_opple` converter to publish the value as `device_temperature` in place of `temperature`.

I'm pretty sure some people will complain about this, because the `device_temperature` has no options to "calibrate" it. But we need to unify both sensors, in one sense or in the other.

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -1568,7 +1568,7 @@ module.exports = [
         fromZigbee: [fz.on_off, fz.device_temperature, fz.aqara_opple, fz.ignore_metering, fz.ignore_electrical_measurement,
             fz.xiaomi_power],
         exposes: [e.switch(), e.energy(), e.power(), e.device_temperature(), e.power_outage_memory(), e.power_outage_count(),
-            e.switch_type(), e.voltage(), e.temperature(), e.current(),
+            e.switch_type(), e.voltage(), e.current(),
         ],
         toZigbee: [tz.xiaomi_switch_type, tz.on_off, tz.xiaomi_switch_power_outage_memory, tz.xiaomi_led_disabled_night],
         configure: async (device, coordinatorEndpoint, logger) => {
--- a/lib/xiaomi.js
+++ b/lib/xiaomi.js
@@ -160,6 +160,8 @@ const numericAttributes2Payload = (msg, meta, model, options, dataObject) => {
         case '3':
             if (['WSDCGQ12LM'].includes(model.model)) {
                 // This temperature value is ignored because with greater accuracy it is reported in key №100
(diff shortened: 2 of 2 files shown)


Otnow:
I think that it is necessary to rename the value of _attribute 3_ for all from `temperature` to `device_temperature` or `chip_temperature`, because this is the service value of the chip temperature.

McGiverGim:
@Otnow maybe is the correct move, looking at debug logs from the [WSDCGQ11LM](https://www.zigbee2mqtt.io/devices/WSDCGQ11LM.html) and the [WSDCGQ01LM](https://www.zigbee2mqtt.io/devices/WSDCGQ01LM.html) nobody of them publish the the "3" attribute.

My only doubt is about the exceptions:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/88fa4b76373b3a208c688b5c795e3766d06028c6/lib/xiaomi.js#L161-L165

By the comment, I suppose the [WSDCGQ12LM](https://www.zigbee2mqtt.io/devices/WSDCGQ12LM.html) publishes the "3" value, but maybe is not too bad to publish it without less precision as `device_temperature` and continue publishing with high precision as `temperature`.

The other devices I suppose they publish the "3" attribute but it is not the temperature.

@Koenkk what do you think? Maybe simply change the `temperature` by `device_temperature` in the "3" attribute? Remove the exception for WSDCGQ12LM or let it?

Otnow:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/88fa4b76373b3a208c688b5c795e3766d06028c6/lib/xiaomi.js#L161-L162

I [added](https://github.com/Koenkk/zigbee-herdsman-converters/pull/4171) this exception, so if we rename `temperature` to `device_temperature`, then this condition can be removed, because in this case, there will be no intersection of the value with `temperature` from _attribute 100_ for **WSDCGQ12LM**.

McGiverGim:
Ok, perfect! Let's wait for @Koenkk revision to see if he is aligned with the change. I will search too for other devices with `gen_basic` or `aqara_opple` that expose `temperature`. Maybe some of them must start to expose `device_temperature`.

Koenkk:
Looks good, thanks!

Issue #4140, opened 2022-04-20
Tuya Cover: coverPosition handler should indicate that cover is moving.


nchieng:
This branch on the switch statement here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/converters/fromZigbee.js#L2148

```javascript
case tuya.dataPoints.coverPosition:
  return { running: true };
```

The [following switch statement case](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/converters/fromZigbee.js#L2149-L2161) should appropriately handle setting `running` back to `false` when the `coverArrived` message is received.

PR #4276, opened 2022-05-23
Report cover as running when new position reported

nchieng:
Issue #4140 

`coverPosition` is reported whenever the covers are currently moving to open or closed state. 
Hence this PR makes it so that it reports it as `running:true`

The case statement below that handles `coverArrived` should appropriately set the `running` flag back to false.


Koenkk:
Thanks!

Koenkk:
This broke the position reporting so I had to revert this. (https://github.com/Koenkk/zigbee2mqtt/issues/12719). 

I don't understand why this was needed in the first place. Given running is based on the dp: `const running = dp === tuya.dataPoints.coverArrived ? false : true;` it should always be set to `true` when the dp is `tuya.dataPoints.coverPosition`

tungmeister:
I've updated to the dev to resolve the position reporting issue however running is now constantly reported as true, is this expected?

Koenkk:
@tungmeister pushed a possible fix

PR #4279, opened 2022-05-23
Add additional Zigbee model name for Insta 57008000

felixstorm:
The Zigbee model name for Insta 57008000 seems to have changed (probably with a recent firmware update). This PR adds the new name (keeping the old one for backwards compatibility).

Apart from the model name the existing device definition still seems to work fine with the latest firmware.

"Nexentro" seems to be the brand name under which Insta GmbH sells its own Zigbee devices (see https://www.nexentro.com/legal-notice/).

### DIFF
--- a/devices/insta.js
+++ b/devices/insta.js
@@ -21,7 +21,7 @@ module.exports = [
         ota: ota.zigbeeOTA,
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #4282, opened 2022-05-23
Update colortemp range for 5996411U5

shawnheide:
Changed the color temperature range to the min/max values that the device will set.


Koenkk:
Thanks!

PR #4281, opened 2022-05-23
Add white version of Philips Enrave M

lorenz:
This adds the white version of the Philips Hue Enrave M, similar to the other Enrave models.

PR #4286, opened 2022-05-25
add white Enrave M

peterus:
fix for PR #4281 (just in case that there is no update coming) as I will need this in the next days.


Koenkk:
Thanks!

PR #4294, opened 2022-05-28
Add AUBESS ZXZIR-02 (_TZ3290_acv1iuslxi3shaaj) universal smart IR rem…

mak-42:
…ote control.
Issue: https://github.com/Koenkk/zigbee2mqtt/issues/11633

### DIFF
--- a/devices/aubess.js
+++ b/devices/aubess.js
@@ -1,0 +1,25 @@
+const exposes = require('../lib/exposes');
+const fz = require('../converters/fromZigbee');
+const tz = require('../converters/toZigbee');
+const ea = exposes.access;
+
+module.exports = [
+    {
+        fingerprint: [{modelID: 'TS1201', manufacturerName: '_TZ3290_acv1iuslxi3shaaj'}],
+        model: 'ZXZIR-02',
+        vendor: 'AUBESS',
+        description: 'Universal Smart IR Remote Control',
+        fromZigbee: [
+            fz.zosung_send_ir_code_00,fz.zosung_send_ir_code_01,fz.zosung_send_ir_code_02,fz.zosung_send_ir_code_03,
+            fz.zosung_send_ir_code_04,fz.zosung_send_ir_code_05
+        ],
+        toZigbee: [
+            tz.zosung_ir_code_to_send, tz.zosung_learn_ir_code
+        ],
+        exposes: [
+            exposes.switch().withState('learnIRCode',undefined,'Turn on to learn new IR code',ea.SET),
+            exposes.text('learnedIRCode', ea.STATE).withDescription('The IR code learned by device'),
+            exposes.text('IRCodeToSend', ea.SET).withDescription('The IR code to send by device'),
+        ],
+    }
+];
--- a/lib/zosung.js
+++ b/lib/zosung.js
@@ -1,0 +1,42 @@
+'use strict';
+
+function nextSeq(entity){
+    entity.seq = ((entity.seq || -1)+1) % 0x10000;
+    return entity.seq;
+ }
+ 
+function messagesGet(entity, seq) {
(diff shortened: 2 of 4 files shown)


devices/aubess.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable linebreak-style */
+const exposes = require('../lib/exposes');
+const fz = require('../converters/fromZigbee');
+const tz = require('../converters/toZigbee');
+const ea = exposes.access;
+
```
Koenkk, line 6: Please move all the additions to `toZigbee.js` and `fromZigbee.js` to here (`fzLocal`/`tzLocal` like https://github.com/Koenkk/zigbee2mqtt.io/blob/ce7d93ca3cf736b9a30f75fa73e7b4002d9bf8eb/docs/externalConvertersExample/freepad_ext.js#L20)

lib/zosung.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+function nextSeq(entity) {
```
Koenkk, line 3: Please move all code to `devices/aubess.js`

devices/aubess.js
```diff
@@ -0,0 +1,26 @@
...
+            fz.zosung_send_ir_code_00, fz.zosung_send_ir_code_01, fz.zosung_send_ir_code_02, fz.zosung_send_ir_code_03,
+            fz.zosung_send_ir_code_04, fz.zosung_send_ir_code_05,
+        ],
+        toZigbee: [
+            tz.zosung_ir_code_to_send, tz.zosung_learn_ir_code,
+        ],
+        exposes: [
+            exposes.switch().withState('learnIRCode', undefined, 'Turn on to learn new IR code', ea.SET),
+            exposes.text('learnedIRCode', ea.STATE).withDescription('The IR code learned by device'),
+            exposes.text('IRCodeToSend', ea.SET).withDescription('The IR code to send by device'),
```
Koenkk, line 23: Please make all the exposes snake case (e.g. `learnIRCode` -> `learn_ir_code`).

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1404

devices/aubess.js
mak-42: I plan (in nearest future) to support a MOES device which use same protocol to handle IR remote control (look at the source issue).  That's why I move this part of the code to `toZigbee.js` and `fromZigbee.js`.
I agree that jumped the gun, and I'll move it to `device/aubess.js`.

But could you explain shall I move that code back to `toZigbee.js` and `fromZigbee.js` when I add a MOES device, or I should just copy it to `moes.js` file?

Koenkk: I see, maybe better to move to `lib/zosung.js` then (btw where does the name zosung come from?)

mak-42: I have a small investigation and figure out that the devices are likely manufactured by Shenzhen Zhuoxing Smart Technology Co., Ltd. (Zosung, [device page](http://www.zosung.cn/en/productshow.asp?Id=263)) for different vendors like AUBESS and so on.

You suggestion look reasonable (less conflicts and so on). I'll move everything into `lib/zosung.js` and make a commit after checking I break nothing.

lib/zosung.js
mak-42: I left it in the file, but now it is a local function.

devices/aubess.js
mak-42: done

devices/aubess.js
mak-42: I moved everything to `lib/zosung.js`.
All the changes in `toZigbee.js` and `fromZigbee.js` has been reverted.

PR #4296, opened 2022-05-28
Added config for iluminize 511.524 device

Zaunmich:
This is my first time adding a device config to zigbee herdsman. The device works on my end, but I am not sure if some of the exposes are redundant. The 4 scenes buttons on the device can not be set by in the web-GUI, they have to be set on the device itself.


Koenkk:
Thanks!

PR #4300, opened 2022-05-30
Try catch around unsupported attribute

robertalexa:
Hi Koen,

You have already patched the majority of them as part of this issue:
https://github.com/Koenkk/zigbee2mqtt/issues/11872

There was however 1 more that is having the same behaviour. The attribute is not available on the Hive TRV, only on the official Danfoss firmware.


Koenkk:
Thanks!

PR #4301, opened 2022-05-30
Add new features for TS0004

vladyspavlov:
Based on changes in https://github.com/Koenkk/zigbee-herdsman/pull/546.

New features for TS0004:
- Power on behavior for each endpoint separate and for the whole device
- Backlight switch
I don't understand how it affects the device, but it present in the tuya app, so maybe someone will find use case.
- Indicate switch
Same as previous.


vladyspavlov:
Maybe TS0002 have the same features, but I don't have it to check and there are many different manufs, so I can't be sure. If someone can confirm it so I can also add it for those devices.

converters/toZigbee.js
```diff
@@ -3862,6 +3862,20 @@ const converters = {
             await entity.read('manuSpecificTuya_3', ['switchType']);
         },
     },
+    tuya_power_on_behavior: {
+        key: ['tuya_power_on_behavior'],
```
Koenkk, line 3866: Please rename to `power_on_behavior`

devices/tuya.js
```diff
@@ -1798,9 +1798,23 @@ module.exports = [
...
-        exposes: [e.switch().withEndpoint('l1'), e.switch().withEndpoint('l2'), e.switch().withEndpoint('l3'),
-            e.switch().withEndpoint('l4')],
+        fromZigbee: [fz.on_off, fz.moes_power_on_behavior, fz.tuya_power_on_behavior, fz.tuya_backlight_switch,
+            fz.tuya_indicate_switch, fz.ignore_basic_report],
+        toZigbee: [tz.on_off, tz.moes_power_on_behavior, tz.tuya_power_on_behavior, tz.tuya_backlight_switch,
+            tz.tuya_indicate_switch],
+        exposes: [
+            e.power_on_behavior(),
+            exposes.switch().withState('backlight_switch', true, ''),
+            exposes.enum('indicate_switch', ea.ALL, ['off', 'switch', 'position']),
```
Koenkk, line 1808: I don't understand what this options does, can you add a description to it? e.g.

```js
exposes.enum('indicate_switch', ea.ALL, ['off', 'switch', 'position']).withDescription('blabla')
```

converters/toZigbee.js
vladyspavlov: There is also `moes_power_on_behavior` converter used with this device which already uses `power_on_behavior` key. So if I rename it the `moes_power_on_behavior` converter will intercept `tuya_power_on_behavior` converter. Or do you meanconverter name, not a key?

devices/tuya.js
vladyspavlov: These options present in tuya app, but changing their states doesn't affect device. I think there are similar devices which have the same model and they uses these options.

<details>
<summary>Tuya app</summary>

</details>

Backlight turns on automatically when the device is powered.

As it uses 4 gang switch icon in tuya, maybe this devices uses options:
https://www.aliexpress.com/item/1005004218695401.html
<details>
<summary>icon</summary>

</details>

Of course, I can remove them if it's necessary

Koenkk: If it doesn't do anything and it is unclear what it should do I propose to remove it (it only creates confusion otherwise)

devices/tuya.js
```diff
@@ -43,6 +43,45 @@ const tzLocal = {
...
+            const backlight = value === 'ON' ? 1 : 0;
+            await entity.write('genOnOff', {tuyaBacklightSwitch: backlight});
+            return {state: {backlight}};
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('genOnOff', ['tuyaBacklightSwitch']);
+        },
+    },
+    power_on_behavior: {
+        key: ['tuya_power_on_behavior'],
```
Koenkk, line 72: Please rename this to `power_on_behavior`

vladyspavlov: This device also uses `moes_power_on_behavior` converter which also has `power_on_behavior` key. I have tested and if it is renamed key to `power_on_behavior` it doesn't work correctly, `moes_power_on_behavior` converter calls earlier, so that's why I added "tuya" prefix.

devices/tuya.js
vladyspavlov: Done, https://github.com/Koenkk/zigbee-herdsman-converters/pull/4301/commits/7c7789c6b848906b35ed05dabc0300acef69f0a4

devices/tuya.js
Koenkk: What is the reason to use two `power_on_behavior` converters? If I look at `moes_power_on_behavior` is provides the same functionallity.

vladyspavlov: `moes_power_on_behavior` uses `genOnOff` cluster. This converter, with prefixed key, uses `manuSpecificTuya_v3` cluster.
`moes_power_on_behavior` switches "power on bevahior" for the whole device and all channels, whereas this converter switches "power on behavior" for the each channel separately.

Koenkk: I see, thanks for the explanation, I think it would be better to expose just one `power_on_behaviour` (the `manuSpecificTuya_v3` one)

vladyspavlov: Done, https://github.com/Koenkk/zigbee-herdsman-converters/pull/4301/commits/e19b5d069e9474f6fdd56fdcf983791c91ed954f

Koenkk:
Thanks!

andresfilia:
@vladyspavlov the backlight is working? Because I can't find this option on zigbemqtt... :(

vladyspavlov:
> @vladyspavlov the backlight is working? Because I can't find this option on zigbemqtt... :( 

It seems it was removed. I'm not worrying about it, anyway it wasn't working, at least with my devices.

andresfilia:
I tested using the Dev Console on zigbemqtt and it worked.
Using the `tuyaBacklightSwitch`
`tuyaBacklightSwitch = 1`
Screenshot 2023-02-14 at 18 13 17
IMG_3787

`tuyaBacklightSwitch = 0`
Screenshot 2023-02-14 at 18 19 27
All light in this image is off.
IMG_3788

The one light is on.
IMG_3789

PR #4306, opened 2022-06-01
Set power source to battery for two Third Reality devices

guillochon:
Both of these devices have batteries and are not connected to mains. Hopefully this will fix the issue where these devices appear plugged in in the zigbee2mqtt GUI.


Koenkk:
Thanks!

PR #4308, opened 2022-06-01
Danfoss Ally - adaptation run control - Add default state and fix typo

robertalexa:
Hi Koen,

Just doing some maintenance on my HA and have discovered the following:

image

The device when not doing neither of these 2 actions reports that attribute with a value of `0` which was not defined in the constants definition. (There was also a type in the word `initiate` which was fixed)

As per the documentation:
image

The documentation fails to make note of the `0`value which is however supported and default as seen in the last column `0x00` but also in range.

This PR adds the default value to the constants and the device converter and will fix the error in HA

Thanks


Koenkk:
Thanks!

PR #4316, opened 2022-06-03
ikea unhandled promise rejection

arteck:
we have a bug here

message 

`unhandled promise rejection: Cannot read property 'color_options' of undefined`

not the state.color_options is undefined ..  

the query whether is undefined must be set to   state ..


sjorge:
It might need both, state can also be undefined.

arteck:
> It might need both, state can also be undefined.

i think no.. only the state is undefined.. but ok here is the corr solution

Koenkk:
Thanks!

Issue #4284, opened 2022-05-23
Wrong/unknown thermostat running states for Centralite 3156105


bigun27:
I have noticed that Z2M correctly interprets the thermostat's running state in its non-heat pump configuration, but gets it wrong when configured to use a heat pump. Turns out, this is because the thermostat sends completely different values for the running state - some of which Z2M is not set up for. Is this an issue that can be resolved? I have made a table below of all values given by the device, along with the proper function to which they correspond.

| Type     |    |    |
|----------|----|----|
| Cooling  | 22 | 7  |
| Heating  | 9  | 13 |
| Fan only | 4  | 5  |
| Idle     | 0  | 1  |

robertalexa:
These are the current mappings:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/a1b6a7ef34eaf1b337a071a80f2b336986fb29c0/lib/constants.js#L58-L74

From your table, here are some discrepancies I am seeing (but can't make an assessment over which ones are the correct ones)
7 - currently undefined
13 - currently undefined
5 - currently `heat` vs your recommendation of `fan_only`
1 - currently `heat` vs your recommendation of `idle`

Could you double check that your table is correct?

@Koenkk do you have any insights on how to double check what running_state the values belong to?

From the ZCL this is all i can find:
![image](https://user-images.githubusercontent.com/9019306/171032075-63f2bc5f-e72c-4d66-aaaa-709f965aea83.png)

bigun27:
> These are the current mappings:

I saw that file when I was looking around

> From your table, here are some discrepancies

Yes, they are all weird discrepancies, but I can't control what values the device produces.
To be clear, there are other values the device produces that are correct, but I have omitted those values and only included the incorrect ones.

> Could you double check that your table is correct?

It is correct. I don't know why the manufacturer deviated from the standard values half of the time

Koenkk:
@robertalexa seems the table is correct, so if the manufacturer deviates a custom converter should be created for it.

Koenkk:
Fixed by https://github.com/Koenkk/zigbee-herdsman-converters/pull/4317

PR #4317, opened 2022-06-03
Add Centralite 3156105 proper running states

bigun27:
As mentioned in #4284 this thermostat has non-standard running states. This should fix that issue.

### DIFF
--- a/devices/centralite.js
+++ b/devices/centralite.js
@@ -8,6 +8,36 @@ const e = exposes.presets;
 const ea = exposes.access;
 const constants = require('../lib/constants');
 
+const fzLocal = {
+    thermostat_3156105: {
+      cluster: 'hvacThermostat',
+      type: ['attributeReport', 'readResponse'],
+      convert: (model, msg, publish, options, meta) => {
+          const result = {};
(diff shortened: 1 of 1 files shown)


converters/fromZigbee.js
```diff
@@ -8575,6 +8575,89 @@ const converters = {
             return result;
         },
     },
+    centralite_3156105_thermostat: {
```
Koenkk, line 8578: Can you move this converter into `centralite.js` under `fzLocal` and do the same for the changes to `constants.js`? Example: https://github.com/Koenkk/zigbee2mqtt.io/blob/402126e4bac0aa2f84d716bdec4420b3ab55a661/docs/externalConvertersExample/freepad_ext.js#L20

To prevent code duplication, you can remove everything that is the same as in `fz.thermostat` by calling the converter here, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/c1304800b51c7b90417c8c0be985c66a96d6681f/converters/fromZigbee.js#L6462

bigun27:
@Koenkk 
I have made the suggested changes, thanks for the examples!

Koenkk:
Looks clean now, thanks!

PR #4318, opened 2022-06-03
Add GL-C-002P mini5in1 as whitelabel of GL-C-008P

MartB:
I think its safe to include this device as whitelabel now.
See: https://github.com/Koenkk/zigbee2mqtt/issues/11202


Koenkk:
Thanks!

PR #4321, opened 2022-06-03
Fix zigbeeModel of Müller Licht 404038

Pferdebockwurst:
Müller Licht 404038 reports zigbeeModel 'tint-ColorTemperature', identical to Müller Licht 404037.


Pferdebockwurst:
I don't know how to pass the check. Müller Licht devices 404037 and 404038 seem to be technically identical, the only difference is the shape of the bulb. They report the same zigbeeModel / model ID 'tint-ColorTemperature'

Koenkk:
Merged the two now, thanks!

PR #4324, opened 2022-06-04
Custom tz, fz for efekta devices

smartboxchannel:
update EFEKTA_miniPWS

### DIFF
--- a/devices/custom_devices_diy.js
+++ b/devices/custom_devices_diy.js
@@ -6,7 +6,7 @@ const extend = require('../lib/extend');
 const e = exposes.presets;
 const ea = exposes.access;
 
-const tz_custom = {
+const tzCustom = {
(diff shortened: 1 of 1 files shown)


smartboxchannel:
Maybe someone can help with advice on why this fails the test. I do not understand. It works well on my local machine.

smartboxchannel:
> Maybe someone can help with advice on why this fails the test. I do not understand. It works well on my local machine.

camelCase is a naming convention in which each word within a compound word is capitalized except for the first word. Software developers often use camelCase when writing source code.

camelCase is useful in programming since element names cannot contain spaces. The camelCase naming convention makes compound names more readable. For example, myOneMethod is easier to read than myonemethod.

Koenkk:
Thanks!

PR #4325, opened 2022-06-05
add cooling setpoint

gunnarsundberg:
Currently, there is only a setting for a heating setpoint. This adds the Zen thermostat's already exposed cooling setpoint to zigbee herdsman converters.


Koenkk:
Thanks!

PR #4331, opened 2022-06-06
Change in Inovelli firmware numbering

nathanfiscus:
This is a fix for a recent change in Inovelli's versioning to decimal type numbers.


Koenkk:
thanks!

PR #4332, opened 2022-06-06
Niko 552-72201 should report changes

sjorge:
Minor follow up so that the Dimmer reports changes when controlled with the physical buttons.

Ref: https://github.com/Koenkk/zigbee2mqtt/issues/12726#issuecomment-1147833385


Koenkk:
Thanks!

PR #4341, opened 2022-06-12
Add HZC Eletric's dual dimmer D086-ZG

andredahlqvist:
Adds basic support for HZC Eletric's dual smart dimmer with model number D086-ZG.


Koenkk:
Thanks!

PR #4344, opened 2022-06-13
Add support for battery from batteryVoltage for Develco WISZB-120

sjorge:
Just got one of these and the manual says it supports battery readout, after poking the device it seems only batteryVoltage can be read. A fresh set of batteries reports 30 (2xAAA). I put in an older set that read ~50% on my Hue Motion sensor and with 3V_2500 they read around 40% so seems to be the correct one.

The WISZB-121 should be similar but without the tamper and temperature but I do not have one so I cannot verify the genPowerCfg also lives on endpoint 35.

### DIFF
--- a/devices/develco.js
+++ b/devices/develco.js
@@ -234,13 +234,17 @@ module.exports = [
         model: 'WISZB-120',
         vendor: 'Develco',
         description: 'Window sensor',
-        fromZigbee: [fz.ias_contact_alarm_1, fz.temperature],
+        fromZigbee: [fz.ias_contact_alarm_1, fz.battery, fz.temperature],
         toZigbee: [],
-        exposes: [e.contact(), e.battery_low(), e.tamper(), e.temperature()],
+        exposes: [e.contact(), e.battery(), e.battery_low(), e.tamper(), e.temperature()],
+        meta: {battery: {voltageToPercentage: '3V_2500'}},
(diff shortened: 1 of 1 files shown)


sjorge:
```node
let v = [ 30, 29, 28, 27, 26, 25, 24, 23, 22, 21 ];
for (const i of v) {
  console.log(toPercentage((i * 100), 2200, 3000));
} 
```

```
100
88
75
63
50
38
25
13
0
0
```

Give a much smoother transition curve than 3V_2500
```
100 
80
60
40
20
0
0
0
0
0
```

@Koenkk I wonder if we need another converter to better deal with AA/AAA batteries?
There voltage usually ~1.5v for a full one and most die between 1.1-1.2v.

https://en.wikipedia.org/wiki/AAA_battery

I'm happy to add a `3V_AAA` variant or something like that to the batteryVoltageToPercentage

sjorge:
The same revised mapping should also be used for Develco's AQSZB-110 as this uses 2x AA which have the same voltage range before they are dead around 1.1v so similar floor than 2x AAA both devices seems to have the batteries in series as the batteryVoltage returned with fresh batteries is `30` for both devices.

sjorge:
nvm, this is good to go as is
image

Looks like 3V_2500 is the correct one to use. So this is good to go.

Koenkk:
Thanks!

PR #4346, opened 2022-06-14
Add support for TS0210 additional vendor name

NicolaiPetri:
Vendor name is different for this device. It works perfect with this patch applied.


Koenkk:
Thanks!

PR #4350, opened 2022-06-16
Add manufacturer name '_TZ3210_s6zec0of' to model 'TS0505B'.

hinnalear:
Tuya Zigbee RGB+CCT E27 Light Bulb.
[https://www.ebay.com/itm/184561060701](url)


Koenkk:
Thanks!

PR #4351, opened 2022-06-16
Expose colortemp capability of Hue Iris gen2 white

mdamen:
Make sure Hue Iris gen2 white has colortemp exposed, just like black edition.


Koenkk:
Thanks!

PR #4354, opened 2022-06-18
Update tuya.js by adding one more white label tuya dimmer

pakkardkaw:
Update tuya.js by adding one more white label tuya dimmer


devices/tuya.js
```diff
@@ -678,6 +679,7 @@ module.exports = [
             {vendor: 'Earda', model: 'EDM-1ZAB-EU'},
             {vendor: 'Earda', model: 'EDM-1ZBA-EU'},
             {vendor: 'Mercator Ikuü', model: 'SSWD01'},
+            {vendor: 'Moes', model: 'TS0601'},
```
Koenkk, line 682: Is there any model number written on the device itself?

pakkardkaw: It's 'ZS-USD'

Koenkk:
thanks!

PR #4355, opened 2022-06-18
Set turnsOffAtBrightness1 to true on some of the IKEA lights

Electry:
This is the default behavior on the latest IKEA firmware.
Without turnsOffAtBrightness1, setting the brightness to 1 turns off the light (while still showing the state as ON).

Due to this, I had to use workarounds in my home assistant automations that would limit the min brightness to 2, which was not ideal.

I only updated the bulbs and LED panels that I personally own and have tested with the latest OTA update (updated via zigbee2mqtt).


Koenkk:
Thanks!

PR #4356, opened 2022-06-18
Extend Aqara H1 no neutral (WS-EUK01, WS-EUK02) functionality

Electry:
This PR extends the functionality of the Aqara H1 (no neutral, EU version) in-wall light switches.

Newly exposed options/states are:
- led_disabled_night (true/false)
- mode_switch (anti_flicker_mode/quick_mode)
- power_outage_count
- device_temperature

I tested all the changes on the no-neutral variants of the switches (WS-EUK01/2). Unfortunately I don't own the ones with neutral, but I'll be buying them soon and I'll create a separate PR for them if needed.


Koenkk:
Thanks!

lcsjandrey:
Do you think this converter would work for the Aqara US Switch with neutral (WS-USC04)?

Chekedaudau:
Hello there, is it possible to extend the functionality of the Aqara H1 EU switches with the OTA feature because most of the non EU H1 switches have this feature? Thank you for your brilliant work and have a nice day Thomas

PR #4360, opened 2022-06-19
Support 1746330V7 - Philips Hue Appear Outdoor Wall Light

andrewortman:
1746330V7 should be an alias of 1746330P7


Koenkk:
Thanks!

PR #4364, opened 2022-06-21
Update tuya.js - Added 'TS0601' 3 gang switch from manufacturer '_TZE200_bynnczcb'

dmedina2015:
Added 'TS0601' 3 gang switch from manufacturer '_TZE200_bynnczcb'


Koenkk:
Thanks!

PR #4373, opened 2022-06-23
corr fz.battery to fz.xiaomi_basic

arteck:


devices/xiaomi.js
```diff
@@ -45,7 +45,7 @@ module.exports = [
         model: 'MCCGQ14LM',
         vendor: 'Xiaomi',
         description: 'Aqara E1 door & window contact sensor',
-        fromZigbee: [fz.ias_contact_alarm_1, fz.aqara_opple, fz.battery],
+        fromZigbee: [fz.ias_contact_alarm_1, fz.aqara_opple, fz.xiaomi_basic],
```
Otnow, line 48: Please explain why you want to replace `fz.battery` with `fz.xiaomi_basic`, given that they process _different messages from different clusters_?:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/0043ecc9433580e8bca5a82e75ef1125f27915f7/converters/fromZigbee.js#L313-L345

https://github.com/Koenkk/zigbee-herdsman-converters/blob/0043ecc9433580e8bca5a82e75ef1125f27915f7/converters/fromZigbee.js#L5551-L5574

arteck: you right.. this is only for sensor data..
not for battery

PR #4375, opened 2022-06-23
Added support for Tuya temp and humidity sensor

owangen:
It's Tuya compatible device, but it's branded as IHOMECAM bought from here https://www.aliexpress.com/item/1005003634353180.html?spm=a2g0o.store_pc_promotion.slider_2002889515586.1


owangen:
I guess it's not correct to label the manufacturer as IHOMECAM. Maybe Generic is more correct...

Koenkk:
- Please rename the `vendor` to `TuYa`
- Move the added converters to the `tuya.js` file (into `fzLocal`/`tzLocal`)

owangen:
@Koenkk all changes implemented as described in your comment

Koenkk:
Sorry for not noticing this earlier, but it seems the added converters are very similar to (`tz.nous_lcd_temperature_humidity_sensor` and `fz.nous_lcd_temperature_humidity_sensor`, can these be merged?

owangen:
Yes, it's correct they are similar. I started out with extending those converters, most worked, but noticed that some of the datapoints was handle differently on the device I have.

Koenkk:
Could you merge them? You can make an if/else based on the manufacturerName. e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/85f6c9bfadb9ec04a7f94106f221bd4016724a9f/converters/toZigbee.js#L3816

owangen:
Will fix. Didn't think of the possibility to use manufacturername in converters. More cleaner to do it like that than a lot of special converters for almost same devices.

owangen:
Closing this one since it was supported in this PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/4415

PR #4376, opened 2022-06-24
Fix TuYa TS011F_plug_2 wrongly detected as TS011F_plug_1

iglu-sebastian:
This switches `_TZ3000_nzkqcvvs` from `TS011F_plug_1` to `TS011F_plug_2` as I have the device here and it does not support energy monitoring at all. If there is a `_TZ3000_nzkqcvvs` out there that is indeed a `TS011F_plug_1` with energy monitoring then let me know and we need to find some other solution to properly detect it.


Koenkk:
Thanks!

PR #4386, opened 2022-06-26
Add new model to "Hue white ambiance Adore GU10 with Bluetooth (3 spots)"

antaflos:
Adds 929003056401 to 3418131P6.


Koenkk:
Thanks!

PR #4391, opened 2022-06-28
#4385: `tuya_switch_power_outage_memory` converters - update attribute code according to cluster definition changes

### LINKED ISSUE #4385: fromZigbee.tuya_switch_power_outage_memory doesn't work, because of attribute check
There is a check for attribute `0x8002`:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/a88c04ce95062e78c09a356bb94dc8a3959ec616/converters/fromZigbee.js#L7276

But this attribute is listed under name `moesStartUpOnOff` in cluster definition:
[moesStartUpOnOff](https://github.com/Koenkk/zigbee-herdsman/blob/772d2edcc0e442a84fe8faa91dff2f0ae36e81ad/src/zcl/definition/cluster.ts#L460)
so the method receives it as a string and the check always fails.

There is also newer `moes_power_on_behavior` converter, which looks similar, but uses correct attribute name:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/a88c04ce95062e78c09a356bb94dc8a3959ec616/converters/fromZigbee.js#L2561

At the `toZigbee` side converters look close also, but the `moes_power_on_behavior` variant contains the `get` converter in addition:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/a88c04ce95062e78c09a356bb94dc8a3959ec616/converters/toZigbee.js#L2580

https://github.com/Koenkk/zigbee-herdsman-converters/blob/a88c04ce95062e78c09a356bb94dc8a3959ec616/converters/toZigbee.js#L3583

gorbunovav:
Fixes #4385

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -7273,7 +7273,7 @@ const converters = {
         cluster: 'genOnOff',
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
-            const property = 0x8002;
+            const property = 'moesStartUpOnOff';
 
             if (msg.data.hasOwnProperty(property)) {
                 const dict = {0x00: 'off', 0x01: 'on', 0x02: 'restore'};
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #4395, opened 2022-06-29
Add Evvr.js IRS configuration file

Focalcrest-Madi:
Add Evvr.js IRS configuration file
Add product module ID: SRB01/SRB01A


Koenkk:
thanks!

PR #4398, opened 2022-06-29
add 929003045101_03 to 929003047101

hallgrimurp:
add 929003045101_03 to 929003047101 both are round hue milliskin lights


Koenkk:
Thanks!

PR #4403, opened 2022-07-01
Add support for Livarno LED hemisphere lights

LukeLR:
Livarno outdoor LED light chain exists in two versions, this adds the second one.


Koenkk:
Thanks!

PR #4406, opened 2022-07-02
power factor precision

Utyff:
Change precision for PowerFactor parameter from 1 to 3
PTVO approved


ptvoinfo:
Looks reasonable.

Koenkk:
Thanks!

PR #4407, opened 2022-07-02
Fix typo for BTicino L4411C

RyuzakiKK:
This device is called L4411C, as mentioned in the official BTicino
catalogue https://catalogue.bticino.com/BTI-L4411C-EN


Koenkk:
Thanks!

PR #4413, opened 2022-07-03
Extend Aqara H1 with neutral (WS-EUK03, WS-EUK04) functionality

Electry:
Similarly to #4356, this PR exposes some new options/states of the Aqara H1 in-wall light switch, this time for the EU version with neutral wiring.

Newly exposed options/states are:
- led_disabled_night (true/false)
- power_outage_count

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -2300,7 +2300,7 @@ const converters = {
         key: ['led_disabled_night'],
         convertSet: async (entity, key, value, meta) => {
             if (['ZNCZ04LM', 'ZNCZ15LM', 'QBCZ14LM', 'QBCZ15LM', 'QBKG19LM', 'QBKG20LM', 'QBKG25LM', 'QBKG26LM',
-                'QBKG31LM', 'QBKG34LM', 'DLKZMK11LM', 'SSM-U01', 'WS-EUK01', 'WS-EUK02'].includes(meta.mapped.model)) {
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

dziedzic99:
Hi, I would be happy to try to code functionality to disable the leds permanently in these switches by tricking them it's night but I can't seem to find where they get the idea it's night from. Can someone please point me in the right direction?

nulano:
Hi, unlike the no-neutral version, this PR did not add mode_switch (anti_flicker_mode/quick_mode). Is this change not supported for these switches?

Electry:
> Hi, unlike the no-neutral version, this PR did not add mode_switch (anti_flicker_mode/quick_mode). Is this change not supported for these switches?

No-neutral variants shouldn't have any issues with flickering lights as both the switch and the light are directly connected to both neutral and live.

My WS-EUK04 responds with UNSUPPORTED_ATTRIBUTE if I try to use the aqara_switch_mode_switch converter.

Electry:
@dziedzic99 

I think they keep track of time on their own, but I haven't really looked into it. Maybe you could trick them by faking the genTime response and always respond with a 00:00 value.

nulano:
> My WS-EUK04 responds with UNSUPPORTED_ATTRIBUTE if I try to use the aqara_switch_mode_switch converter.

Yeah, I've since tested that. I'm not concerned about lights flickering, but want a faster trigger for automations. I've asked Aqara support and they said only the wireless switch supports "quick" mode.

While looking for a way to enable "quick" mode using some other parameter I found some values that might be the ON and OFF time. The switches do keep their own time, but I have no idea how, and don't really want to mess with changing random values on them.

PR #4419, opened 2022-07-04
Add operation_mode support for  WS-USC01

boojew:
Simple change to xiaomi.js to update WS-USC01 to add operation_mode and anti-flicker support.


hellojunha:
Hello. After this update, I see unexpected 'stateless switch' on my Homebridge device list. How do I remove or ignore that behaviour?

PR #4420, opened 2022-07-05
add KTNNKG ZB1248-10A

swex:
Thanks to this device, my sliding gates are now 'smart'!
Probably it exposes something else, but I don't have experience how to investigate it...
see https://aliexpress.ru/item/1005002658541715.html


Koenkk:
Thanks!`

PR #4421, opened 2022-07-05
Make GL-D-006P ota aware, see also https://github.com/Koenkk/zigbee-O…

renoutg:
Make GL-D-006P ota aware, see also https://github.com/Koenkk/zigbee-OTA/pull/135


Koenkk:
Thanks!

PR #4422, opened 2022-07-05
Koenkk/zigbee2mqtt#12688: `ZNLDP13LM` - add `power_outage_memory` support

gorbunovav:
Fixes  Koenkk/zigbee2mqtt#12688


Otnow:
In order for the value of `power_outage_memory` to be retrieved when requested, `fz.aqara_opple` must also be added.

In turn, adding `fz.aqara_opple` will turn on the processing of the heartbeat-message (`'{"247":{"data": ... '`) from the device, so there will be new states that can be exposed (for example, `device_temperature`, `power_outage_count`, etc.).

Also check if the _100 (state) attribute_ from the heartbeat-message matches the actual state of the device, because it will be overwritten:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/bb01ed184425e7a000437c0a6ed253354edff3a7/lib/xiaomi.js#L188-L232

gorbunovav:
@Otnow thank you for the hint! I've checked - the `100` attribute indeed reports the state value in the expected manner. 
I've also added some tweaks for other reported attributes.

PR #4435, opened 2022-07-10
Added Mercator Ikuü' model SMD4109W-RGB-ZB' to tuya.js

LyndseyPaxton:
Added support for a downlight Mercator Ikuü' model SMD4109W-RGB-ZB' to tuya.js

The model already existed in Tuta.js so I ASSUME just updating this file in this manner is the correct approach.


Koenkk:
Please revert the changes done to `package-lock.json`.

PR #4437, opened 2022-07-10
Properly support _TZ3000_r6buo8ba and others with appVersion 160

stanionascu:
Turns out that devices with applicationVersion 160 (fw 2.2.0) also require
polling for electrical measurements. Even though currentSummDelivered,
complains that it's unreportable attribute, it still reports the energy
when the value changes (depending on the load - very infrequently).

These are round plugs with Model: BSD48 marking.

Fixing issue https://github.com/Koenkk/zigbee2mqtt/issues/12760

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1660,7 +1660,7 @@ module.exports = [
     },
     {
         fingerprint: [].concat(...TS011Fplugs.map((manufacturerName) => {
-            return [69, 68, 65, 64].map((applicationVersion) => {
+            return [160, 69, 68, 65, 64].map((applicationVersion) => {
                 return {modelID: 'TS011F', manufacturerName, applicationVersion};
             });
         })),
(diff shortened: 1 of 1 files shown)


devices/tuya.js
```diff
@@ -1677,7 +1677,11 @@ module.exports = [
             const endpoint = device.getEndpoint(1);
             // Enables reporting of physical state changes
             // https://github.com/Koenkk/zigbee2mqtt/issues/9057#issuecomment-1007742130
-            await endpoint.read('genBasic', ['manufacturerName', 'zclVersion', 'appVersion', 'modelId', 'powerSource', 0xfffe]);
+            try {
+                await endpoint.read('genBasic', ['manufacturerName', 'zclVersion', 'appVersion', 'modelId', 'powerSource', 0xfffe]);
```
Koenkk, line 1681: This is interesting, maybe TuYa changed their magic packet to something different. Do you have a TuYa gateway to check this with? (sniff traffic when the device pairs to the TuYa gateway: https://www.zigbee2mqtt.io/advanced/zigbee/04_sniff_zigbee_traffic.html#with-cc2531

stanionascu: Yes, I have the gateway. But only the CC2652P Sonoff dongle stick.

Koenkk: Asked it here now: https://github.com/Koenkk/zigbee2mqtt/issues/12760#issuecomment-1181811294

PR #4438, opened 2022-07-11
Add "withPreset" method to the text exposes

vladyspavlov:
It will be useful to have ability to add presets to the text exposes.

### DIFF
--- a/lib/exposes.js
+++ b/lib/exposes.js
@@ -180,6 +180,12 @@ class Text extends Base {
         this.property = name;
         this.access = access;
     }
+
+    withPreset(name, value, description) {
+        if (!this.presets) this.presets = [];
+        this.presets.push({name, value, description});
+        return this;
+    }
 }
 
 class Composite extends Base {


vladyspavlov:
I want to use it in this converter https://github.com/Koenkk/zigbee-herdsman-converters/pull/3840, to create a preset of the default schedule for the thermostat to be able easily revert to it. It will be useful in cases like this one https://github.com/Koenkk/zigbee-herdsman-converters/blob/5eae5396f8e5e5b0e1809a19cc0fd67753f17acd/devices/tuya.js#L1416

Koenkk:
I see, @nurikk could you add support for this in the frontend?

vladyspavlov:
I'll try to add this feature to frontend, in progress.

nurikk:
If I understand correctly, withPreset is z2m functionality, frontend works with simple json objects

vladyspavlov:
> frontend works with simple json objects

The frontend can't display presets for textual fields now, it only works with numerical fields.

https://github.com/nurikk/zigbee2mqtt-frontend/blob/2cb6d92a253817b64a136ecb2744569524a1dff7/src/components/range-editor/range-editor.tsx#L26
https://github.com/nurikk/zigbee2mqtt-frontend/blob/dev/src/components/textual-editor/textual-editor.tsx

PR #4440, opened 2022-07-12
WIP: add plugwise lisa

vhdirk:
This adds support for the plugwise Lisa zone thermostat: https://www.plugwise.com/product/lisa/?lang=en
Still WIP since I want to figure out how to display the boiler status on the display and make the scenes work.

@JanJaapKo Do you happen to have one of these, too? I can always use assistance :) Perhaps we can ask Plugwise support for some pointers from the get-go.

### DIFF
--- a/devices/plugwise.js
+++ b/devices/plugwise.js
@@ -149,4 +149,27 @@ module.exports = [
                 .withDescription('Calibrates valve on next wakeup'),
         ],
     },
+    {
+        zigbeeModel: ['158-01'],
+        model: '158-01',
+        vendor: 'Plugwise',
+        description: 'Lisa zone thermostat',
+        fromZigbee: [fz.thermostat, fz.temperature, fz.battery],
+        toZigbee: [
+            tz.thermostat_system_mode,
+            tz.thermostat_occupied_heating_setpoint,
+        ],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genBasic', 'genPowerCfg', 'hvacThermostat']);
+            await reporting.batteryPercentageRemaining(endpoint);
+            await reporting.thermostatTemperature(endpoint);
+        },
+        exposes: [e.battery(),
+            exposes.climate()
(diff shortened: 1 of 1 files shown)


vhdirk:
@JanJaapKo ah, too bad. Would you mind sharing the email address you got the info for the Tom from?

JanJaapKo:
Sure, it was helpdesk@plugwise.com

However I first had created a ticket in their support system. Not sure if mailing out of the blue will work (the subject always had the ticket number in it). I think I created it here: https://www.plugwise.com/contact-formulier/

vhdirk:
I've asked their R&D department for documentation on the boiler status indicators and scene modes.
Apparently, none of that is implemented. So this is effectively a very expensive thermostat display.

@Koenkk That means this should be ready as is.

JanJaapKo:
> I've asked their R&D department for documentation on the boiler status indicators and scene modes. Apparently, none of that is implemented. So this is effectively a very expensive thermostat display.
> 
Makes you wonder where their selling point is then....... Or does it come with some kind of coordinator where they hide all the magic?

JanJaapKo:
Well the Zigbee spec allows you to control a display, to a certain extend (which is indeed kinda stupid). So it's either that or they've hidden all these things in custom clusters they're not disclosing to you. You couldn't sniff it out, like with the Tom?

JanJaapKo:
Right. So all devices are pretty dumb and everything works via the Adam? In fact that is what their website promotes, doesn't it? 

Come to think of it, looks like I forgot to update the z2m documentation of the Tom, I'll make an update on that later......

vhdirk:
Well yes, that is what they promote. But the Lisa without Adam is useless (unless you roll your own zigbee network). It has a button to select scenes, but that's local per Lisa and does not sync with Adam. And the pictures have it showing boiler status (heating/warm water), which is not even implemented apparently.

Koenkk:
Thanks!

bouwew:
@JanJaapKo @vhdirk Plugwise has released a new device, the Emma Pro.
I have received one from Plugwise (mainly for integration in the HA Plugwise Integration), together with some extra info related to its ZigBee capabilities.

Can you please help me in adding support for this new device to ZigBee2MQTT?

PR #4445, opened 2022-07-14
Add A60S RGBW new Version of SMART+ Lamp E27 RGBW

joeyboey:
This is a new Version of the bulb. The max colorTempRange was checked according to the guide in the wiki.


joeyboey:
image

Maybe just to check if I entered everything correctly.

Koenkk:
Picture will be added with the next release on 1 August

PR #4446, opened 2022-07-14
Fixing Maximum temperature bug. Adding running state based on output_status

nagydavid:
the earlier converter restricted the maximum allowed temperature to 30 Celsius, it threw an error time-to-time, when the regulator send out an update where the maximum allowed temperature is 35. 
Also added a running_state attribute, based on the output status, so it can be visible on the thermostat card, if the actuator is active or not.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3603,12 +3603,12 @@ const converters = {
             //             msg.data['danfossOutputStatus'];
             // }
             if (msg.data.hasOwnProperty('danfossOutputStatus')) {
-                if (msg.data['danfossOutputStatus'] === 1) {
+                if (msg.data['danfossOutputStatus'] === true) {
                     result[postfixWithEndpointName('output_status', msg, model)] = true;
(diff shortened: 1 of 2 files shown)


nagydavid:
@robertalexa @Koenkk 
Hi I am trying to create a PR, but I have some error which does not give mi any insight.

nagydavid:
@robertalexa it is possible, since the solution was suggested by you :)

nagydavid:
@robertalexa  do you know how to force the external converter to be used until it will appear in edge

robertalexa:
> @robertalexa do you know how to force the external converter to be used until it will appear in edge?

Give your device a unique id in your external converter. Then edit your database file and change the id of your device to the new id. Restart z2m. This way your paired device will not match the supported device, but it will match the external converter.

Hope this helps

nagydavid:
Thanx, but what do you mean by "database file" . devices.yaml? or ....

nagydavid:
@robertalexa  @Koenkk 
I tested in Home Assistant as well, and it works

Screenshot 2022-07-15 at 15 57 06

PR #4449, opened 2022-07-15
preferHS for Osram gardenpole RGBW

jasperro:
This is an updated version of this pr: https://github.com/Koenkk/zigbee-herdsman-converters/pull/3708

Colors were not represented correctly with XY, so prefer HS mode.

Tested with home assistant as working, giving the correct colors.


Koenkk:
Thanks!

PR #4458, opened 2022-07-19
Inovelli VZM31SN - Read Multiplier and Divisors

nathanfiscus:
Reads the multiplier and divisor information for electrical measurements


Koenkk:
Thanks!

PR #4463, opened 2022-07-20
Add Philips AKARI Downlight (LCD004, 8719514382350)

asedov:
The same as https://www.zigbee2mqtt.io/devices/8719514344723.html


Koenkk:
Thanks!

PR #4464, opened 2022-07-21
Update bitron.js

lxeka:
i have modified also locally and it enavles also the cooling part for this type of thermostat model. So this thermostat can be both changed to cooling  and heating mode


Koenkk:
Thanks!

PR #4465, opened 2022-07-22
Update osram.js

Burki24:
AC03642 doesn't suppport color_temp_startup and temperature-range is 135-370 mired.

### DIFF
--- a/devices/osram.js
+++ b/devices/osram.js
@@ -83,7 +83,7 @@ module.exports = [
         model: 'AC03642',
         vendor: 'OSRAM',
         description: 'SMART+ CLASSIC A 60 TW',
-        extend: extend.ledvance.light_onoff_brightness_colortemp(),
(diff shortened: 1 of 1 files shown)


devices/osram.js
```diff
@@ -83,7 +83,7 @@ module.exports = [
         model: 'AC03642',
         vendor: 'OSRAM',
         description: 'SMART+ CLASSIC A 60 TW',
-        extend: extend.ledvance.light_onoff_brightness_colortemp(),
+        extend: extend.ledvance.light_onoff_brightness_colortemp({colorTempRange: [153, 370], disableColorTempStartup: true}),
```
Koenkk, line 86: disableColorTempSupport can be removed since it is already disabled by default now: https://github.com/Koenkk/zigbee-herdsman-converters/blob/ba4a9c7f5d3eba5ff9939608fcfd6b600d23840d/lib/extend.js#L136

Koenkk:
Thanks!

PR #4467, opened 2022-07-22
Device support

asgothian:
New Device: Heiman Smart Air Quality Monitor HS2AQ-EFA-3.0 (see [here](https://www.heimantech.com/product/?type=detail&id=106))
New Device: LED Trading FOH smart switch 9125 (see [here](https://www.sunricher.com/battery-powered-friends-of-hue-switch-sr-zgp2801k4-foh-e.html))

Note: I do not have an EnOcean PTM215Z Module, which seems to identify with the same IEEE stub, so there may be a cross effect with these modules unless they provide a `manufId` which is not `null`

A.


Koenkk:
Thanks!

PR #4471, opened 2022-07-23
Corrects Device Definition for Tuya IH-K009 being Recognized as WSD500A

timnolte:
The device with model ID `TS0201` & manufacturer name `_TZ3000_dowj6gyi` was incorrectly being identified as a `WSD500A` when in fact it is a `IH-K009` device. Device information provided here as well. https://zigbee.blakadder.com/Tuya_IH-K009.html A request for support was mentioned here. https://github.com/Koenkk/zigbee2mqtt/issues/12759


Koenkk:
Thanks!

PR #4474, opened 2022-07-23
New Device

asgothian:
New Device: Sunricher SR-ZG9092A
Device is similar to Namron 4512737/4512738:
- Automatic time setting, genTime cannot be written to
- no support for display brightness
- no support for child lock

Updated exposes: Added `unoccupied_heating_setpoint` and `unoccupied_cooling_setpoint` as allowed Setpoints to the climate expose.


Koenkk:
Thanks!

PR #4479, opened 2022-07-24
Add: zignito device for led strip taken from tuya device

haydenbroadhead:


Koenkk:
Can you provide a link of this device? (with picture)

haydenbroadhead:
This is where i bought it https://www.amazon.co.uk/gp/product/B079SN66K6/

Looks like the brand is actually Ajax Online on the listing, happy to change naming if that's preferable?

haydenbroadhead:
i can see the other devices in the ajax js have use zigbeeModel e.g `zigbeeModel: ['AJ-RGBCCT 5 in 1']` and no fingerprint like i have. Not sure what's right now.

haydenbroadhead:
I moved to ajax but now i'm getting a build failure, any idea what that could be?

Koenkk:
Fixed it, thanks!

PR #4485, opened 2022-07-26
Publication of the `illuminance` state from attribute 11 of the heartbeat-message only for Xiaomi `RTCGQ11LM`

Otnow:
Based on the [discussion](https://github.com/Koenkk/zigbee2mqtt/discussions/13284), when checking, it turned out that the _attribute 11_ of the heartbeat-message is the `illuminance` state only for Xiaomi `RTCGQ11LM`, and the rest of the Xiaomi sensors with `illuminance` exposure use _attributes 100 and 101_:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/b6321b3055193af1955864d1b26b37c1483a22c9/lib/xiaomi.js#L246-L248

https://github.com/Koenkk/zigbee-herdsman-converters/blob/b6321b3055193af1955864d1b26b37c1483a22c9/lib/xiaomi.js#L272-L276

### DIFF
--- a/lib/xiaomi.js
+++ b/lib/xiaomi.js
@@ -191,10 +191,7 @@ const numericAttributes2Payload = (msg, meta, model, options, dataObject) => {
             }
             break;
         case '11':
-            if (['JT-BZ-01AQ/A', 'QBKG25LM', 'ZNLDP13LM'].includes(model.model)) {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #4486, opened 2022-07-26
Add _TZ3000_qja6nq5z as manufacturerName for tuya smart knob

AWare:
Just got one of these and it has a different vendor name to the ones listed, model number on base matches this.


Koenkk:
Thanks!

PR #4487, opened 2022-07-26
Add support for updated Hue Go (7602031PU)

theadamfarah:
The updated Philips Hue Go Lamp has differing model numbers for EU and UK models - [https://www.assets.signify.com/is/content/Signify/Assets/signify/global/PHILIPS-HUE-PORTABLE-LUMINAIRES-202109-UK-DoC.pdf](url).

This adds support for UK models through adding the model number to the `zigbeeModel` array.

Tested working using an external convertor.

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -406,7 +406,7 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness(),
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #4466, opened 2022-07-22
use fwu.ubisys.de to resolve older but needed images

zivillian:
fixes Koenkk/zigbee2mqtt#12526

PR #4491, opened 2022-07-27
ubisys OTA: return latest firmware if multiple files match

zivillian:
This is a followup for #4466 

As mentioned in https://github.com/Koenkk/zigbee2mqtt/issues/12526 the current approach fails, if multiple firmware update files are valid.

This patch tries to return only the firmware with the highest fileVersion.


zivillian:
I've successfully updated my C4 from 20210305-DE-FB0 via 1.9.2 and 1.9.4 to 2.0.0

Koenkk:
Thanks!

Skaronator:
Just updated z2m to 1.27.0 and gave this a try. Successfully updated my J1 from 1.9.1 to 1.9.2 (and looks like I can update again).

Thanks for the fix @zivillian!

PR #4492, opened 2022-07-27
Improve profalux cover support

Kevkehz:
Hello,

I open a PR to fix my own issue : https://github.com/Koenkk/zigbee2mqtt/issues/12634.
I didn't receive any answer about usage of `modelId` as selector.

Tell me if it's ok.

### DIFF
--- a/devices/profalux.js
+++ b/devices/profalux.js
@@ -17,24 +17,9 @@ module.exports = [
         exposes: [],
     },
     {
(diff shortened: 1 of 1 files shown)


Kevkehz:
As i can see, it's not authorised to use the same `model`. 
Maybe the best way to solve this is to find the correct fingerprint ?

Koenkk:
I've merged it with the already existing NSAV061 entry, thanks!

Jeepmg:
Hello,

If this is easy to do, would it not be useful to show the lqi and especially the battery level of the NB102 remote control?

Thanks to the contributors of the Profalux sections

PR #4500, opened 2022-07-29
Error When Referring to fz.command_store

marioedani:
I built a custom external converter and did a reference (in "definitition.fromZegbee") to various converters from "fromZigbee.js". The only one that rises an error is "fz.command_store". So I looked into the code and I think I found the problem (see proposed changes).
I also tried to create a "fzLocal.command_store" in my external converter, copying the code from original file "fromZigbee.js", and applying my changes: it worked well.
Sorry for my english. I hope I have explained.


Koenkk:
Well found, thanks!

marioedani:
You are welcome! Thank you to YOU!

PR #4502, opened 2022-07-29
Support Philips LWE005 Filament Bulbs

Nixon506E:
A new revision of the filament bulb that supports white ambiance


Koenkk:
Thanks!

PR #4507, opened 2022-07-31
Iluminize Shutter: Update position+state based on liftPercentage

Sven286:
(My first PullRequest)
My Iluminize device do have 'cover_state_via_onoff', but it's giving me the state "Is the device currently OPENING or CLOSING".
Also the position was'nt read out. I had to report on the attribute 'currentPositionLiftPercentage'.
This attribute was given me what I needed.
So I had to update the position and the state based on this attribute.


Koenkk:
Thanks!

PR #4511, opened 2022-08-03
Fix thermostatRunningMode default reporting

sjorge:
Made a mistake when moving things from my external converter to lib/reporting.
runningMode is very much like systemMode, it just reflects the 'current' state e.g. heat/cool/off depending on if the system is operating or not, so we should use the same defaults.


Koenkk:
Thanks!

PR #101, opened 2018-10-18
Zigbee OnOff Controller

kirovilya:
This is a noname device I bought on eBay
https://www.ebay.com/itm/273428347127

And I found a similar product from the manufacturer http://ksentry.manufacturer.globalsources.com/si/6008837134660/pdtl/ZigBee-module/1162731630/zigbee-on-off-controller-modules.htm

Issue #10, opened 2018-06-03
Strange report from QBKG03LM wired wall switch


kirovilya:
I got strange for me unhandled attReport from QBKG03LM wired wall switch (router):
```
"type":"attReport",
"endpoints":[
    {
        "device":{
            "_id":4,
            "type":"Router",
            "ieeeAddr":"0x00158d0001f4d992",
            "nwkAddr":29351,
            "manufId":4151,
            "manufName":"LUMI",
            "powerSource":"DC Source",
            "modelId":"lumi.ctrl_neutral2",
            "epList":[1,2,3,4,5,6,8],
            "status":"online",
            "joinTime":1528024933,
        },
        "profId":260,
        "epId":1,
        "devId":6,
    }
],
"data":{
    "cid":"genBasic",
        "data":{
            "65281":{
                "1":3300,
                "3":33,
                "5":14,
                "8":9746,
                "10":0,
                "100":0,
                "101":1,
                "110":0,
                "111":2,
                "153":0,
                "155":0
            }
        }
    }
}
```
What do you think it means?
Maybe it's someone's passing message? or an error in Shepherd?

May be as plug state? :
```
state: data['100'] === 1 ? 'ON' : 'OFF',
power: precisionRound(data['152'], 2),
voltage: precisionRound(data['150'] * 0.1, 1),
consumption: precisionRound(data['149'], 2),
temperature: precisionRound(data['3'], 2),
```

May be "1":3300 - is internal battery?

sebastianheierhoff:
Pretty close to what we're experiencing with the ecozy thermostat, see [Link](https://github.com/Koenkk/zigbee2mqtt/issues/129#issuecomment-406814771). No idea where this is coming from yet.

kirovilya:
@sebastianheierhoff Sorry, no. I do not use wall switch. For testing purposes only

PR #4513, opened 2022-08-05
ZMAM02 Ignore undefined data points

dashdrum:
For the ZMAM02_cover definition in converters/fromZigbee.js, there are two data points that are continually sent by the device that are not handled in the definition - 10 (AM02TimeTotal) and 101(AM02AddRemoter).  This change will ignore these data points to supress the constant warnings.

ZMAM02_cover: Unhandled DP #10 for _TZE200_iossyxra: {"dp":10,"datatype":2,"data":{"type":"Buffer","data":[0,0,3,232]}}
ZMAM02_cover: Unhandled DP #101 for _TZE200_iossyxra: {"dp":101,"datatype":1,"data":{"type":"Buffer","data":[0]}}

A developer can later provide configuration for these DPs when the need arises.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -8445,6 +8445,9 @@ const converters = {
                     break;
                 }
                 break;
+            case tuya.dataPoints.AM02AddRemoter: // DP 101: Ignore until need is defined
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #4515, opened 2022-08-06
Fix #4514 for Ubisys H1

### LINKED ISSUE #4514: The ubisys H1 uses a pIHeatingDemand scale of 0-100 not 0-254 as per the ZCL
The ubisys H1 uses a pIHeatingDemand scale of 0-100 not 0-254 as per the ZCL, 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/f575f5789f48067cca3148615e5e970034cd5e95/converters/fromZigbee.js#L111-L114

Does a straight mapping, so I can either create a copy of that one specifically for ubisys that takes the raw value. Or introduce a meta property to not map the value.

I'm leaning towards the 2nd solution as it can probably be used for other devices too,  but I have no time this weekend to look at this so mostly a issue to track this so I do not forget.

sjorge:
This also updates the danfoss TRV and Sinope TRV as they also reports 'raw' 0-100 %

Note, the Sinope is full of lib/legacy.js stuff, I did not clean/rework any of that and just elimited the sinope_thermostat from the regular convertors and updated the one call to the custom from from lib/legacy.js.


Koenkk:
Thanks!

PR #4516, opened 2022-08-06
Add support for BR Avato Tuya Smart Switch with Power Monitoring

raphaeleduardo42:
Hello,

Just like @alanmgomes #4170, I've bought a Tuya Smart Switch from Avato on Aliexpress (https://pt.aliexpress.com/item/1005003059126398.html?spm=a2g0o.order_list.0.0.21efcaa4YXwVtL&gatewayAdapt=glo2bra) and it have a different model.

I've tested and it's looks like to be the same model with the same features.


Koenkk:
Thanks!

PR #4518, opened 2022-08-08
TH1123ZB TH1124ZB SystemMode  can't be set to auto

MattL0:
There is no  'auto' SystemMode  on these thermostats

+ occupied_heating_setpoint accept values from 5 to 30


Koenkk:
Thanks!

PR #4519, opened 2022-08-08
support ZVG1 timers and weather_delay

fhempy:
This PR supports setting and reading
 - normal timers,
 - cycle timers and
 - weather_delay
for ZVG1 irrigation control.


Koenkk:
Big thanks!

ajarosz:
Hi Dominik @fhempy 
I noticed that this PR removed the ability to set the value of `timer_state` property. Was it intentional? With this property being fixed on 'enabled', it seems not possible to keep the value opened indefinitely - it is getting closed after `timer` value elapses.

Regards,
Artur

PR #4520, opened 2022-08-09
GL-S-006P as same GL-S-007P

arteck:


Koenkk:
This device is already supported: https://github.com/Koenkk/zigbee-herdsman-converters/blob/90a787cf3069865298b4f778fe2390d08c150b3d/devices/gledopto.js#L334

PR #4523, opened 2022-08-11
added support tuya motion sensor TS0202 _TZ3000_lltemgsf

d51x:


devices/tuya.js
```diff
@@ -849,6 +849,20 @@ module.exports = [
         toZigbee: [],
         exposes: [e.occupancy(), e.battery_low(), e.tamper(), e.battery()],
     },
+    {
+        fingerprint: [{modelID: 'TS0202', manufacturerName: '_TZ3000_lltemgsf'}],
+        model: 'TS0202_2',
```
Koenkk, line 854: This can be merged with `TS0202_1`

d51x: ok, but _TZ3000_lltemgsf dosn't have battery and tamper
only occupancy, battery_low

Koenkk: @groenmarsmannetje does the `tamper` work for your TS0202_1 motion sensor?

groenmarsmannetje: @koenkk I do not think so. Will double check

groenmarsmannetje: I checked and the tamper sensor doesn’t turn on when I open the device to remove battery.

Koenkk: Thanks!

yormedia:
Hi, I think you removed a bit to much. Now the battery level percentage is also removed.

PR #4374, opened 2022-06-23
Giex water valve

raf02:
Adding support for a irrigation water valve

https://github.com/Koenkk/zigbee2mqtt/issues/12778

PR #4524, opened 2022-08-11
Extend fingerprint for GiEX QT06

JeanMertz:
See https://github.com/Koenkk/zigbee-herdsman-converters/pull/4374#issuecomment-1212480089.


Koenkk:
Thanks!

PR #4540, opened 2022-08-18
add csm-300v2

nukusinji:
Add new device.
The model name is same ex-device.
But New model has tof sensor.

App version of old version is 1-4.
App version of new version is 10-.


Koenkk:
Thanks!

PR #4548, opened 2022-08-20
OSRAM 4052899926110 doesn't support color temp startup

leroyloren:
`Publish 'set' 'color_temp_startup' to 'LED rack' failed: 'Error: Write 0x7cb03eaa00adf03a/3 lightingColorCtrl({"startUpColorTemperature":125}, {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')'`

### DIFF
--- a/devices/osram.js
+++ b/devices/osram.js
@@ -227,7 +227,7 @@ module.exports = [
         model: '4052899926110',
         vendor: 'OSRAM',
         description: 'Flex RGBW',
-        extend: extend.ledvance.light_onoff_brightness_colortemp_color({colorTempRange: [125, 666], supportsHS: true, preferHS: true}),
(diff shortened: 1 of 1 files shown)


Koenkk:
This should already be the case for all OSRAM, Ledvance and Sylvania bulbs but seems this didn't happen when `options` were passed through the `extend`. Fixed this now in https://github.com/Koenkk/zigbee-herdsman-converters/commit/31e845794d53ee09f1b89b21ed20e7f64a7d60a2

PR #4558, opened 2022-08-24
ZG-204ZL illuminance with unit lx

leroyloren:
Expose illuminance with lx unit.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -2975,7 +2975,7 @@ module.exports = [
         fromZigbee: [fz.ZG204ZL_lms],
         toZigbee: [tz.ZG204ZL_lms],
         exposes: [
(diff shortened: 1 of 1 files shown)


devices/tuya.js
```diff
@@ -2975,7 +2975,7 @@ module.exports = [
         fromZigbee: [fz.ZG204ZL_lms],
         toZigbee: [tz.ZG204ZL_lms],
         exposes: [
-            e.occupancy(), e.illuminance(), e.battery(),
```
Koenkk, line 2978: Please use `e.illuminance().withUnit(`lx`)` instead. This device exposes a `illuminance` property, not `illuminance_lux`.

Koenkk:
Thanks!

PR #4562, opened 2022-08-25
add _TZE200_0u3bj3rc tuya human presence sensor

Burki24:


Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1525/

devices/tuya.js
```diff
@@ -3073,4 +3073,21 @@ module.exports = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg', 'msTemperatureMeasurement', 'msRelativeHumidity']);
         },
     },
+    {
+        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_0u3bj3rc'}],
+        model: 'Human_presence_sensor',
```
Koenkk, line 3078: Is there any other model number written on the device? If not please use `TS0601_human_presence_sensor`.

Burki24: finished, thanks

Koenkk:
Thanks!

Burki24:
@Koenkk 

thanks for putting tz and fz into tzlocal and fzlocal. Is that way of adding  for all devices in the future? Then I will update only the device-scripts from now on and will not do changes in fromzigbee and tozigbee.

Koenkk:
@Burki24 for specific converters yes, otherwise `fromZigbee.js`/`toZigbee.js` are getting too big.

PR #4571, opened 2022-08-28
Add TZE200_b2u1drdv to TS0601_cover

phoenixxie0:


Koenkk:
Tahnks!

phoenixxie0:
But TZE200_b2u1drdv doesn't support "motor_speed". And It is DOOYA DT82TV/MA just like DT82LEMA-1.2N. But without battery.

PR #4580, opened 2022-08-30
Add UFO-R11 variant _TZ3290_ot6ewjvmejq5ekhl

Impact123:
As the title says. Seems to work fine so far.


Koenkk:
Thanks!

inertia666:
I have this Moes IR blaster but cannot figure out how to decrypt the received IR code. The base64 codes i receive are different every time.

I have tried a broadlink packet decryption script and it seems to almost be working but not quite. The messages are still different but do now share some common data, but not all the time. So i am unsure if the decryption key or algorithm is right for my device.

PR #4581, opened 2022-08-31
Add support for Yale YMC 420 D

dsantos21:
This PR seems unusual... The source code for yale.js uses fz lock_operation_event and, for some reason, this model 420 D needs data type to be 'raw', I've just create a clone, just changing data mapping and data type.

Yale.js has an edited version of this function in module.exports (I saw an example of this from fz "easycode_action" and device easyaccess.js).

Feel free to change it all! I was just trying to understand how does it all work, and trying to make a PR from this issue https://github.com/Koenkk/zigbee2mqtt/issues/8946.

Thanks for you incredible work here, @Koenkk !

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3112,35 +3112,35 @@ const converters = {
         },
     },
     yale_ymc_420_d_lock_operation_event: {
-      cluster: 'closuresDoorLock',
-      type: 'raw',
-      convert: (model, msg, publish, options, meta) => {
-        const lookup = {
-          0: 'keypad_unlock',
-          1: 'zigbee_unlock',
-          2: 'Auto_lock',
-          3: 'rfid_unlock',
-          4: 'finger_unlock',
-          5: 'Unlock_failure_invalid_pin_or_id',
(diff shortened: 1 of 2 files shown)


Koenkk:
Can you provide the herdsman debug log of the `yale_ymc_420_d_lock_operation_event` message? Looks that something goes wrong with the decoding of it.

See https://www.zigbee2mqtt.io/guide/usage/debug.html on how to enable the herdsman debug logging. Note that this is only logged to STDOUT and not to log files.

giovannirco:
Hey @Koenkk I want to take over this PR and add a slightly change to support the YMC420W version which seems to be the same lock but water resistant. Do I just need to enable debug logs and run some operations and paste it here? 
I have used an external converter with the script from the issue that originated this PR and was able to pair my yale lock just fine, but I have the username section to map IDs to name on homeassistant. do I need to revert to the other one or the herdsman debug logs would be the same?

Koenkk:
You can make a new PR with the logs

sfrique:
@giovannirco do you have the script working?
I just got a YMC420W and would like to test it

PR #4588, opened 2022-08-31
Fix endpoint added twice to Niko 552-721X2 action property

sjorge:
We probably also need this to fully fix https://github.com/Koenkk/zigbee2mqtt/issues/13737


Koenkk:
See https://github.com/Koenkk/zigbee2mqtt/issues/13737#issuecomment-1233154693

PR #4560, opened 2022-08-24
New device naviclim atlantic

Noneangel:
I tried to add a new device support: an HVAC embedded thermostat (naviclim).
It mostly works, except for the part I wrote.

Could someone help me to address these problems? I will comment the file for each of them with a wireshark log of the sniff I made with the native hub.

PR #4590, opened 2022-08-31
hvac swing mode

Noneangel:
I would like to use the [swing mode for hvac option of Home Assistant](https://developers.home-assistant.io/docs/core/entity/climate/#swing-modes). So I propose to add a `Climate.withSwingMode` methods.

### DIFF
--- a/lib/exposes.js
+++ b/lib/exposes.js
@@ -425,8 +425,6 @@ class Climate extends Base {
 
     withSwingMode(modes, access=a.ALL) {
         assert(!this.endpoint, 'Cannot add feature after adding endpoint');
-        const allowed = ['off', 'on', 'vertical', 'horizontal', 'both'];
-        modes.forEach((m) => assert(allowed.includes(m)));
         this.features.push(new Enum('swing_mode', access, modes).withDescription('Swing mode'));
         return this;
     }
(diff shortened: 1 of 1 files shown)


Koenkk:
Which device support this?

Noneangel:
See https://github.com/Koenkk/zigbee-herdsman-converters/pull/4560#issuecomment-1234694112 for details. Right now, no device support it. If you accept this request, I will add it to https://github.com/Koenkk/zigbee-herdsman-converters/pull/4560

lib/exposes.js
```diff
@@ -423,6 +423,14 @@ class Climate extends Base {
         return this;
     }
 
+    withSwingMode(modes, access=a.ALL) {
+        assert(!this.endpoint, 'Cannot add feature after adding endpoint');
+        const allowed = ['off', 'on', 'vertical', 'horizontal', 'both'];
```
Koenkk, line 428: Looking at the HA docs, there does not seem to be a limitation on the allowed values. I propose that we also don't put any limitations (https://www.home-assistant.io/integrations/climate.mqtt/)

Noneangel: I use `Climate.withFanMode` as template. It is not limited by [home-assistant](https://developers.home-assistant.io/docs/core/entity/climate/#fan-modes) but is limited [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/852b871de63bc557a850435f6c5fcdd6602f4ca6/lib/exposes.js#L420).

PR #4593, opened 2022-09-01
For Owon PCT504-Z

XLDom:
Hello,

i've cancel old pull request to create on with all patch.


XLDom:
Huummmm @Koenkk, do you know why test failed ?

EXP-Carthage:
In your owon.js file, you are using `e.keypad_lockout()` but it is not declared in `toZigbee: []`
Try adding `tz.thermostat_keypad_lockout`.

PR #4594, opened 2022-09-02
Add support for Mercator SSWF01G

ciedema:
Add support for Mercator SSWF01F AC Fan Controller.

### DIFF
--- a/devices/mecrator.js
+++ b/devices/mecrator.js
@@ -5,6 +5,8 @@ const reporting = require('../lib/reporting');
 const e = exposes.presets;
 const ea = exposes.access;
 
+
+
 module.exports = [
     {
         fingerprint: [{modelID: 'TS011F', manufacturerName: '_TZ3210_yvxjawlt'}],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

davidjb:
Just a note for this device: it reports `{fanModeSequence":1}` which according to the ZCL spec should mean the fan has speeds of Low/High, but it actually supports the full range of speeds in this PR (low/medium/high and on, which equals high). The hardware definitely supports 3 speeds because a) it's working 😄, b) the fan speed indicator has 3 LEDs, which represent the 3 speeds & light up correctly and c) the Mercator app provides the 3 speeds.

However, when I use the touch button on mine it only toggles between low/high. If the fan on medium, then the speed goes to high on pressing the button then continues toggling low/high after that. I'm curious @ciedema whether you see the same? Currently discussing with the manufacturer and they say the button should cycle between all 3 speeds.

ciedema:
> 

Mine work as expected - bottom button to turn light on/off, middle turns fan on/off, top cycles through the 3 speeds.

davidjb:
Thanks @ciedema, much appreciated. Seems 2 of 4 of mine work correctly and the other 2 only have low/high speed control at the switch, all straight out of the box and appear identical (same PCB numbers, dates, model, boxes etc). Will see what the manufacturer says.

PR #4596, opened 2022-09-02
Do not read onLevel again if already in state

pyos:
The device might have forgotten it for some reason. If we assume that the user is not modifying it with direct attribute writes, the read should be redundant anyway.

Also, if the device does not support OnLevel at all, publish level_config with on_level set to 'previous' in order to never attempt to read it again.

See Koenkk/zigbee2mqtt#13570

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -859,6 +859,7 @@ const converters = {
                 state = meta.message.state = brightness === 0 ? 'off' : 'on';
             }
 
+            let publishFakeOnLevel = false;
             let publishBrightness = brightness !== undefined;
             const targetState = state === 'toggle' ? (meta.state.state === 'ON' ? 'off' : 'on') : state;
             if (targetState === 'off') {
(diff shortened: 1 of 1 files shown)


sjorge:
Would it not make more sense to just read `currentLevel` after turning on the device before publishing ?

pyos:
What for?

sjorge:
My idea on the first pass was to just not have to deal with it and just read the currentLevel to return the proper value, but since this returns the state before the bulb actually turns on reading that would give the wrong value. Reading onLevel is indeed fine.

Sorry for the rather rude initial comment.

pyos:
Yeah so this entire code is basically a manual implementation of On/Off from ZCL:

>On:
>Temporarily store CurrentLevel.
>Set CurrentLevel to the minimum level allowed for the device.
>Change CurrentLevel to OnLevel, or to the stored level if OnLevel is not defined, over the time period OnOffTransitionTime.
>
>Off:
>Temporarily store CurrentLevel.
>Change CurrentLevel to the minimum level allowed for the device over the time period
>OnOffTransitionTime.
>If OnLevel is not defined, set the CurrentLevel to the stored level.

except the last step of Off, "set the CurrentLevel to the stored level", is infeasible on most bulbs (they don't allow external modifications to CurrentLevel while off), so instead the "stored level" is written to z2m's own storage and then used in the first step of the next On instead of the real CurrentLevel.

"OnLevel is not defined" in ZCL means that it is either unsupported or has a value of 255. The string "previous" is just an alias for 255.

Koenkk:
Thanks!

Koenkk:
I'm realising now that this will add the `level_config` property to every lights payload. Would it be possible to store this data somewhere else? We can use e.g. the `meta.device.meta` object.

PR #4598, opened 2022-09-03
Add new fingerprint for QS-Zigbee-D02-TRIAC-L

CaptFennec:
Add fingerprint for "1 gang smart dimmer switch module without neutral" by Lonsonho
Model QS-Zigbee-D02-TRIAC-L bought on AliExpress, which should be supported by Zigbee2MQTT
https://www.aliexpress.com/item/4001279149071.html

### DIFF
--- a/devices/lonsonho.js
+++ b/devices/lonsonho.js
@@ -134,7 +134,7 @@ module.exports = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #4602, opened 2022-09-04
feat: added Aeotec Extender

mmornati:
## What?

This PR is adding the Aeotec signal extender device.
image

https://store.aeotec.com/products/range-extender-zi-zga001

The device is not exposing or proposing anything special except the LQI signal.

This PR is tested locally in my Z2M as external converter:
image

image


Koenkk:
Thanks!

PR #4607, opened 2022-09-04
add LED2101G4

hugobloem:
I have added TRADFRI bulb E14 WS globe 470lm with model number LED2101G4 to the Ikea vendor file.


Koenkk:
Thanks!

PR #4609, opened 2022-09-05
Add coverInverted to Tuya TS130F

McGiverGim:
This fixes https://github.com/Koenkk/zigbee2mqtt/issues/13805 https://github.com/Koenkk/zigbee2mqtt/issues/13847 https://github.com/Koenkk/zigbee2mqtt/issues/13813 and maybe some more...

Is a breaking change and I'm not too sure if this is the correct solution...
Since this commits:
https://github.com/Koenkk/zigbee-herdsman-converters/commit/1cc7c2301d58233ef541d7e1ecb746414d02ca8a
https://github.com/Koenkk/zigbee-herdsman-converters/commit/2bc158c3cc20ee45bd8b39fdf84df49f21fc0175
The state for the Tuya TS130F is incorrect. It shows CLOSED when OPEN and the inverse.

I suppose the new meta tag must show when the state of the cover is inverted over the zigbee standard: if the device shows 100 when Open, the meta tag must be there indicating is reversed.

This is the case of the TS130F, and I've added this meta tag. The problem is that we added the `invert_cover` config option to all this covers, so this is a breaking change, and we must remove this config.

Is this correct or maybe there is a some bug in commits? I'm not too sure of the intention of the commits, maybe it must use the value "inverted" to get the state and not the original value, that is another option. What do you think @Koenkk ?

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1104,6 +1104,7 @@ module.exports = [
         description: 'Curtain/blind switch',
         fromZigbee: [fz.cover_position_tilt, fz.tuya_backlight_mode, fz.tuya_cover_options],
         toZigbee: [tz.cover_state, tz.cover_position_tilt, tz.tuya_cover_calibration, tz.tuya_cover_reversal, tz.tuya_backlight_mode],
+        meta: {coverInverted: true},
         whiteLabel: [{vendor: 'LoraTap', model: 'SC400'}],
         exposes: [e.cover_position(), exposes.enum('moving', ea.STATE, ['UP', 'STOP', 'DOWN']),
             exposes.binary('calibration', ea.ALL, 'ON', 'OFF'), exposes.binary('motor_reversal', ea.ALL, 'ON', 'OFF'),


convicte:
@McGiverGim 
When looking into it, would you mind considering the earlier issue I've created, which was never resolved:
https://github.com/Koenkk/zigbee2mqtt/issues/11860

It seems like there is a bit of a double negative, since the old way in which this device was working was actually also incorrect, just that the position was inverted and not the state.

Not sure if this could be resolved together?

McGiverGim:
@convicte I'm not sure how this must behave.

According to the documentation, the readme file: 
https://github.com/Koenkk/zigbee-herdsman-converters/blob/852b871de63bc557a850435f6c5fcdd6602f4ca6/README.md?plain=1#L25

says that for devices that sent a position 100 as open, the `coverInverted` must be `true`. Until here, this PR fixes that and the cover will be working without touching `invert_cover`.

No we have the `invert_cover`:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/852b871de63bc557a850435f6c5fcdd6602f4ca6/lib/exposes.js#L498

If I'm not wrong, after this PR, if you enable de `invert_cover`, it will publish now position 0 for Open, and position 100 for Close, so the percentage is inverted, but will continue showing the "correct" OPEN/CLOSE state based on the expected position of the real cover.

This fixes your issue?

convicte:
It seems like there is a disconnect between the position and open/close state.

The latest update messed up the state, but didn't flip the position expected for each state.

What I am trying to say is that beforehand it was just a matter of needing to set the ```invert_cover: true```,  becuase the native position of the cover reported when the parameter was set to false was actually open=0 abd close=100.

This is opposite to the documentation ( "Inverts the cover position, false: open=100,close=0, true: open=0,close=100 (default false)").

Now the issue at hand is worse because the shutters are going down (showing ```moving:down```) and at the end of their run are flipping to open, which is complete nonsense for this type of shutter. 
This said, if we could have the fully bottom state being fixed to close and fully up being open again, as it was prior to 1.27.1 and while this is being done the ```invert_cover:false``` actually being open=100,close=0 that would be fantastic.

However, if the state is the only thing that can get fixed, I can live with my OCD being slightly triggered by having to set the ```invert_cover:true``` just for it to be what is should at false. ;)

I hope that makes sense!

convicte:
> Thanks, just as a final check, without having `cover_invert` in the `configuration.yaml` after this change:
> 
> * setting position 100 fully opens the cover
> * settings position 0 fully closes the cover
Thank you so much for the prompt response.
That is exactly what would be expected—I'll give it a go after it's available and confirm here, for completeness.

convicte:
@Koenkk @McGiverGim 

In retrospect, I believe something may have been missed, but I am not sure if this falls under this bug/pull or not.

The TS130F comes in two flavors - one shutter (TS130F) and two shutter (TS130F_dual) variants.

I have a mixture of both, and looking at it now, the TS130F_dual is suffering from another issue on top of the state being flipped. The entity state for the two shutters, which show up in HA as two independent covers (_left and _right) are indicated as unknown.

image
Front lawn shutter has one TS130F_dual module

The state for both of these shutters (under a TS130F_dual) are given in attribute ```state_left``` and ```state_right``` and they are flipped incorrectly as is the case in the standard TS130F.

They are now open but listed as closed:
```jason

{
    "calibration": "OFF",
    "calibration_left": "OFF",
    "calibration_right": "OFF",
    "calibration_time": 29.1,
    "last_seen": "2022-08-12T18:02:29+02:00",
    "linkquality": 51,
    "motor_reversal": "OFF",
    "motor_reversal_left": "OFF",
    "motor_reversal_right": "OFF",
    "moving": "STOP",
    "moving_left": "STOP",
    "moving_right": "STOP",
    "position_left": 100,
    "position_right": 100,
    "state_left": "CLOSE",
    "state_right": "CLOSE"
}
```

But in general, each HA entity for the physical left and right shutter should have it's state populated correctly, which it currently isn't.

Please let me know if I should create another bug report and transfer it there or will one of the recently closed bugs be sufficient (if re-opened).

McGiverGim:
About the "unknown" state: seeing the code, it seems to me that it only publishes the `state_left` and `state_right`, but not a simple `state`, for this reason it appears as unknown. It seems to me a new bug and has nothing with this PR.

About the inverse state, do you have `invert_cover: true`? If true, to me it seems that is correct (value of 100 is CLOSED when `invert_cover` is true). If not there is some kind of bug. 

If it is open or closed is the device who sends this information. If you want it to be the inverse, the only thing we can do with the actual code is:
- Flip the wires in the device
- Configure it to invert the motor

In this way, it will continue going up and down as now, but it will send the opposite direction to z2m.

I think is better to open a new issue anyway, than continue the conversation in the PR merged. Link to the PR if you think is interesting in the new issue.

McGiverGim:
Only as information, to me it seems that the state in the dual must be taken from here:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/ac5b334b9475d5a5768c5de20d9eee18a5654264/devices/lonsonho.js#L80-L83

This shows the suffix of the state for each device.

McGiverGim:
@convicte only as information, it seems to me that this code does not take into account the possible multiple endpoints in the cover:
https://github.com/Koenkk/zigbee2mqtt/blob/7dc48fb6d657f2560b42b93bc47c569620b3582f/lib/extension/homeassistant.ts#L415
But I don't have any dual device to test it.

convicte:
Thanks for the prompt answer and as usual fantastically thorough investigation @McGiverGim !

1) It used to work just fine before the latest Z2m update, so this must have broken something. I'll open an issue for this variant of the device separately and reference your information listed above.

2) Regarding the ```invert_cover: true```, the issue is identical to what I described above for the single shutter variant. i.e. I have it set to true, to make it behave as it should in false. This is incorrect and should be fixed.

Now, when - ```invert_cover: false```
Shutter fully open
Position: 0

Now, when I set  ```invert_cover: true``
Shutter fully open
Position 100

And this shouldn't be the case, per this response:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/4609#issuecomment-1237780882
It should be the opposite, like described

So in other words, both the TS130F and TS130F_dual have the same issue where the behavior of ```invert_cover``` is reversed as compared to the documentation. This PR per @Koenkk should fix it for the TS130F, but will it also fix it for TS130F_dual?

3) Besides opening a new issue, which I will in just a second, I just wanted to be thorough in my answer and say that you may be misunderstanding the problem here - OPEN/CLOSED.

The whole issue from the beginning was that the OPEN/CLOSED was not being reported correctly, for the TS130F, which is also the case for TS130F_dual. So the same fix should be applied to it - which I will request in the new issue reported.

Since the update I haven't flipped the ```motor_reverse``` nor have I switched the wires, and the shutter is working correctly (showing ```moving: up``` when it moves up and vice versa), but when it reaches the bottom it says open and top, says closed.

In conclusion, there are 3 issues:
1) TS130F_dual - fails to report the state for each cover entity in HA
2) TS130F_dual has the ```invert_cover``` functionality switched compared to default (false gives the results described in the docs for true and vice versa). This PR fixes it for the TS130F, will it also do it for TS130F_dual?
3) TS130F_dual has the OPEN/CLOSED issue happening for it as well, which was fixed by this PR for TS130F , so I wonder if it will also fix it for TS130F_dual

I hope I am being clear in this, but please feel free to follow up on stuff.

jvcsw:
Hi, 

I'm confused.
I don't know if the TS130F issue is resolved. 

In version 1.27.2 this should work fine?
Should I just change invert_cover: false?

Thanks

McGiverGim:
> Hi,
> 
> I'm confused. I don't know if the TS130F issue is resolved.
> 
> In version 1.27.2 this should work fine? Should I just change invert_cover: false?
> 
> Thanks

You need to wait to October version, or install edge version of zigbee2mqtt.

convicte:
Unfortunately, I have some bad news - the latest EDGE installed and the only thing that was resolved is that ```cover_inverse``` is working as described in the documentation:

The covers for TS130F remain open when fully down (with ```motor_reversal: OFF```) and closed when fully up
image
image

In other words, the PR doesn't solve the issue listed for  https://github.com/Koenkk/zigbee2mqtt/issues/13847, at least for me.

McGiverGim:
This is worth cover_inverse enabled? If it is, it is working as intended. Capture a debug log and we can confirm that. And please, attach it in an open issue, not here.

PR #4615, opened 2022-09-05
Fix zigbee2mqtt/13861

sjorge:
https://github.com/Koenkk/zigbee2mqtt/issues/13861

This should handle the issue logged, I only added it to KMPCIL_RES005 which the reporter is using and to lumi.sen_ill.mgl01 which I have access to an used to test if this works.

Given these are battery powered devices, the request will probably timeout unless the device is awake, but it does work.

```
debug 2022-09-05 19:26:11: Publishing get 'get' 'illuminance' to 'lux/office'
debug 2022-09-05 19:26:16: Received Zigbee message from 'lux/office', type 'readResponse', cluster 'msIlluminanceMeasurement', data '{"measuredValue":27068}' from endpoint 1 with groupID 0
info  2022-09-05 19:26:16: MQTT publish: topic 'zigbee2mqtt/lux/office', payload '{"battery":100,"illuminance":27068,"illuminance_lux":509,"linkquality":51,"voltage":3200}'
```

I think it would be OK to add this everywhere where fz.illuminance is used, I can make the extra change if you also think this will be OK.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -1510,6 +1510,12 @@ const converters = {
             await entity.read('msTemperatureMeasurement', ['measuredValue']);
         },
     },
+    illuminance: {
+        key: ['illuminance', 'illuminance_lux'],
+        convertGet: async (entity, key, meta) => {
+            await entity.read('msIlluminanceMeasurement', ['measuredValue']);
+        },
+    },
     // #endregion
 
     // #region Non-generic converters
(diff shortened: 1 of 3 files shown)


schasj:
@sjorge, the KMPCIL_RES005 devices I have allow for an optional battery but all of mine are using USB power. I believe the battery option is there only to keep the device alive during relatively brief power outages.

Koenkk:
Thanks!

PR #4619, opened 2022-09-05
Add new TS011F plug manufacturer

Z0G1ab:
Add new TS011F plug manufacturer: "_TZ3000_dksbtrzs" - tested and seems to work as expected.


Koenkk:
Thanks!

PR #4623, opened 2022-09-06
Update zigbeemodel in novo.js

sreisi:


Koenkk:
This will detect all `TS0601` devices as this device which is not correct, please add an additional fingerprint instead.

sreisi:
@Koenkk ah okay now I understand. So the previous implementation was correct.
I will change the code.
We have a "productive" Zigbee2Mqtt Raspberry in use. Is the integration already available there after a merge or only after a new release?

devices/novo.js
```diff
@@ -7,9 +7,10 @@ const ea = exposes.access;
 module.exports = [
     {
         fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_swhwv3k3'}],
+        zigbeeModel: ['TS0601'],
```
Koenkk, line 10: Please remove this line and check if it still works.

Koenkk:
> We have a "productive" Zigbee2Mqtt Raspberry in use. Is the integration already available there after a merge or only after a new release?

After merge: directly in dev branch (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html), in release version only after next release (every 1st of the month)

devices/novo.js
sreisi: done!

Koenkk:
Besides the move of the description (https://github.com/Koenkk/zigbee-herdsman-converters/pull/4623/files) which is just a cosmetic change, there are no changed anymore in this MR so I assume it can be closed.

PR #4624, opened 2022-09-06
update ledvance.js

dhanukapulkit:
Added the new B40 E14 and PAR16 50 dimmable lamps


dhanukapulkit:
Will submit a new one with the corrections due to indentions here

PR #4627, opened 2022-09-06
Support for Namron Panel Heater

geirra:
Work in progress for Namron Zigbee Panel heater 400/600/800/1000W units. 

Will add the proprietary code in the next few days.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1807,7 +1807,6 @@ const converters = {
             if (data.hasOwnProperty(0x100A)) { // Hysterersis
                 result.hysterersis = precisionRound(data[0x100A], 2) / 10;
             }
-    
             return result;
         },
     },
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1783,6 +1783,34 @@ const converters = {
             return result;
         },
     },
+    namron_panelheater: {
+        cluster: 'hvacThermostat',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const result = {};
+            const data = msg.data;
+            if (data.hasOwnProperty(0x1000)) { // OperateDisplayBrightnesss
+                result.display_brightnesss = data[0x1000];
+            }
+            if (data.hasOwnProperty(0x1001)) { // DisplayAutoOffActivation
+                const lookup = {0: 'deactivated', 1: 'activated'};
+                result.display_auto_off = lookup[data[0x1001]];
+            }
+            if (data.hasOwnProperty(0x1004)) { // PowerUpStatus
+                const lookup = {0: 'manual', 1: 'last_state'};
+                result.power_up_status = lookup[data[0x1004]];
+            }
+            if (data.hasOwnProperty(0x1009)) { // WindowOpenCheck
+                const lookup = {0: 'enable', 1: 'disable'};
+                result.window_open_check = lookup[data[0x1009]];
+            }
+            if (data.hasOwnProperty(0x100A)) { // Hysterersis
+                result.hysterersis = precisionRound(data[0x100A], 2) / 10;
+            }
+    
+            return result;
+        },
+    },
     namron_hvac_user_interface: {
         cluster: 'hvacUserInterfaceCfg',
         type: ['attributeReport', 'readResponse'],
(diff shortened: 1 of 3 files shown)


Koenkk:
Marked it as draft, unmark once ready.

geirra:
Thanks, overlooked the draft tag.

Added the code as it's more than enough to make the device work.
I'm just missing the proprietary stuff for now

geirra:
@Koenkk I need some pointers for adding custom settings to this device.

In the Thermostat-0x0201(Server) definition the unit has the following Proprietary Attributes:
Screenshot 2022-09-14 at 23 52 04

These are bit similar to the namron_thermostat  Proprietary settings, but are different for this panel heater. 
So from what i see i need to create a definition in toZigbee 

Example here
 `else if (key==='OperateDisplayBrightnesss') {
                const payload = {0x1000: {value: lookup[value], type: herdsman.Zcl.DataType.enum8}};
                await entity.write('hvacThermostat', payload, manufacturerOptions.sunricher);`
                
What i don´t understand is the manufacturerOptions.sunricher, what is the source of this.

Koenkk:
> What i don´t understand is the manufacturerOptions.sunricher, what is the source of this.

This is defined here: 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/5538e76d23242c116e0e98d500db1aad76924290/converters/toZigbee.js#L14

The docs say you should use manufacturercode `0x1224` which is what `manufacturerOptions.sunricher` will do:

https://github.com/Koenkk/zigbee-herdsman/blob/8360464b5209c6f5e9ebccf0604bedb662ec91b7/src/zcl/definition/manufacturerCode.ts#L892

geirra:
@Koenkk Need some suggestions here I think.
I'm able to pull the exposed proprietary values and set them, but do these values actually need "reporting" ?


Example
```// display_auto_off
await endpoint.configureReporting('hvacThermostat', [{
    attribute: {ID: 0x1001, type: 0x30},
    minimumReportInterval: 0,
    maximumReportInterval: constants.repInterval.HOUR,
    reportableChange: null}]);
```

This will fail with
```
2022-09-26 23:54:42Failed to configure '0x3fff000a00000420', attempt 4 (Error: ConfigureReporting 0x3fff000a00000420/1 hvacThermostat([{"attribute":{"ID":4097,"type":48},"minimumReportInterval":0,"maximumReportInterval":3600,"reportableChange":null}], {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE') at Endpoint.checkStatus (/app/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:317:28) at Endpoint.configureReporting (/app/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:694:22) at Object.configure (/app/node_modules/zigbee-herdsman-converters/devices/namron.js:592:1) at Configure.configure (/app/lib/extension/configure.ts:115:13) at Configure.onMQTTMessage (/app/lib/extension/configure.ts:55:21))
Info 2022-09-26 23:54:42MQTT publish: topic 'zigbee2mqtt/bridge/response/device/configure', payload '{"data":{"id":"0x3fff000a00000420"},"error":"Failed to configure (ConfigureReporting 0x3fff000a00000420/1 hvacThermostat([{\"attribute\":{\"ID\":4097,\"type\":48},\"minimumReportInterval\":0,\"maximumReportInterval\":3600,\"reportableChange\":null}], {\"sendWhen\":\"immediate\",\"timeout\":10000,\"disableResponse\":false,\"disableRecovery\":false,\"disableDefaultResponse\":true,\"direction\":0,\"srcEndpoint\":null,\"reservedBits\":0,\"manufacturerCode\":null,\"transactionSequenceNumber\":null,\"writeUndiv\":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE'))","status":"error","transaction":"ta8rd-2"}'

**Comment by geirra** [comment]
@Koenkk This one is more or less complete now i think. Ideally i would like to add the two read only values as well, but i havent found any example how to expose the data as RO-values in z2m.

Do you have any examples here?

**Comment by Koenkk** [comment]
I'm not sure what you mean, but regarding exposes it has to be done like this, e.g.:

```js
exposes.numeric('display_brightnesss', ea.STATE)
```

geirra:
Should be ready for review now @Koenkk

Koenkk:
Thanks!

PR #4633, opened 2022-09-07
Add correct support to Salus Controls PS600 (variable zigbeeModel)

jestempablo:
Device can report its `modelId` as one of these weird unicode blocks, which may change on re-pairing or even on Z2M restart. I'd like to save other folks time needed to debug this. These values are copied from `database.db` each time after device was reported as Unsupported. Seems to be stable now, but I'll update if needed.


Koenkk:
Thanks!

PR #4639, opened 2022-09-08
Add OTA support for STARKVIND

sjorge:
Looks like I forgot this while adding support, this should fix https://github.com/Koenkk/zigbee2mqtt/issues/13904


Koenkk:
Thanks!

PR #4645, opened 2022-09-10
Added support for Blaupunkt TMST-S1 Thermostat

jestempablo:
This is not full support, missing parts:

- battery reporting, 
- fahrenheit/celsius toggle,
- toggling away mode,
- make use of unoccupied setpoints 

Unoccupied setpoints are probably related to away mode, but i didn't yet find how would you even set them with the thermostat itself. maybe that requires a gateway and vendor UI? Same goes for using "schedule" operation mode.

But at least core functionality is OK.

### DIFF
--- a/devices/blaupunkt.js
+++ b/devices/blaupunkt.js
@@ -24,4 +24,40 @@ module.exports = [
         },
         exposes: [e.cover_position().setAccess('state', ea.ALL)],
     },
+    {
+        zigbeeModel: ['TMSTB_00.00.03.07TC'],
+        model: 'TMST-S1',
+        vendor: 'Blaupunkt',
+        description: 'Blaupunkt Smart Thermostat TMST-S1',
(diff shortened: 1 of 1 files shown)


jestempablo:
I messed up linting a lot more 😅 will fix that soon

jestempablo:
Looking at the entity in Home Assistant I would also need to fix the climate to not use range when not in auto mode (that is - regardless of mode selected showing `[Mode] 20-28 °C` instead of `Heat 20 °C`, `Cool 28 °C` or `Auto 20-28 °C`. So a little bit more tinkering needed, which may take some time since I don't know a lot about setting up support.

Koenkk:
Marked it as draft, please let me know once ready (please also revert the formatting changes https://github.com/Koenkk/zigbee-herdsman-converters/pull/4645/files)

PR #4651, opened 2022-09-12
Added support for IKEA LED2009C3

ima-quack:
IKEA retails a model of their E12 250 lumen light as LED2009C3 rather than the LED1935C3 that is already supported by Zigbee2MQTT.


Koenkk:
Thanks!

PR #4655, opened 2022-09-12
Update enbrighten.js

majorsl:
Added the Enbrighten 43113 which is an updated version of the 43080 dimmer switch. Tested with my 8 43113's in my home.


Koenkk:
Thanks!

PR #4656, opened 2022-09-13
Added initial support for Yale ZYA-C4-MOD-S module

danieledwardgeorgehitchcock:
Resolves Koenkk/zigbee2mqtt#4778

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -8639,33 +8639,6 @@ const converters = {
             return result;
         },
     },
-    yale_lock_operation_event: {
-        cluster: 'genAlarms',
-        type: ['commandAlarm'],
-        convert: (model, msg, publish, options, meta) => {
-            const result = {};
-            if (msg.data.clusterid == 64512) {
-                // We need to read the lock state in case the alarm code is unknown
-                msg.endpoint.read('closuresDoorLock', ['lockState']);
-                const alarmcode = msg.data.alarmcode;
-                const lookup = {
-                    9: 'error_jammed',
-                    21: 'manual_lock',
-                    22: 'manual_unlock',
-                    24: 'lock',
-                    25: 'unlock',
-                    27: 'auto_lock',
-                };
-                if (!lookup[alarmcode]) {
-                    result.action = 'unknown';
-                    meta.logger.warn(`zigbee-herdsman-converters:Yale Lock: Unrecognized Operation Event (${alarmcode})`);
-                } else {
-                    result.action = lookup[alarmcode];
-                }
-            }
-            return result;
-        },
-    },
     // #endregion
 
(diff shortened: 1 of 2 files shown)


danieledwardgeorgehitchcock:
Depends on Koenkk/zigbee-herdsman#575

converters/fromZigbee.js
```diff
@@ -8639,6 +8639,33 @@ const converters = {
...
         },
     },
+    yale_lock_operation_event: {
+        cluster: 'genAlarms',
+        type: ['commandAlarm'],
+        convert: (model, msg, publish, options, meta) => {
+            const result = {};
+            if (msg.data.clusterid == 64512) {
+                // We need to read the lock state in case the alarm code is unknown
+                msg.endpoint.read('closuresDoorLock', ['lockState']);
```
Koenkk, line 8649: This needs to be awaited with a try/catch around it. Otherwise z2m will crash with an unhandled promise rejection when the request fails.

converters/fromZigbee.js
```diff
@@ -8639,6 +8639,33 @@ const converters = {
             return result;
         },
     },
+    yale_lock_operation_event: {
```
Koenkk, line 8642: Move this converter to `yale.js` `fzLocal` (like: https://github.com/Koenkk/zigbee-herdsman-converters/blob/82990cd19b72847f98f4d90ad4d462a12c9c8cba/devices/tuya.js#L291)

converters/fromZigbee.js
danieledwardgeorgehitchcock: I have put an await around it, and made it async again in my latest push but, could not find any existing async examples in fromZigbee to use as a template

devices/yale.js
```diff
@@ -20,6 +20,41 @@ const lockExtend = (meta, lockStateOptions=null, binds=['closuresDoorLock', 'gen
...
 };
 
+const fzLocal = {
+    c4_lock_operation_event: {
+        cluster: 'genAlarms',
+        type: ['commandAlarm'],
+        convert: async (model, msg, publish, options, meta) => {
+            const result = {};
+            if (msg.data.clusterid == 64512) {
+                // We need to read the lock state in case the alarm code is unknown
```
Koenkk, line 30: I've added the try catch. One thing I'm wondering:

> // We need to read the lock state in case the alarm code is unknown

Can this read move to `if (!lookup[alarmcode]) {`? (given that we only need to read when unknown)

danieledwardgeorgehitchcock: Thanks!

I could put it there and add another lookup / condition statement to determine lock state - I was just replicating how the module communicates with it's native hub.

Let me know if you want me to change it :-)

Koenkk:
Perfect, thanks!

PR #4658, opened 2022-09-14
Titan Products CO2 sensor supports OTA

leoluk:
This was verified with a beta OTA image provided by the vendor.
Once it is released, I will add it to the OTA repository.

image


Koenkk:
> Once it is released, I will add it to the OTA repository.

It makes only sense to merge this once the OTA is available in https://github.com/koenkk/zigbee-ota, otherwise users will get errors when checking for updates. Ping me once it is there.

PR #4666, opened 2022-09-16
Added Support for DTB-ED2011-014

swissbyte:


swissbyte:
Thanks for pointing this out. I have adjusted the URL's accordingly. 

Regarding your 2nd question. Im not an expert regarding zigbee. What my device does is, it represents 4 ON/OFF switches. 
If you touch one of the areas, it will report ON for this key. If you touch the same area again, it will report OFF. 

Since this state is not controllable from the coordinator, cause it is an end device, i thought it would be a good idea to represent it as a binary state. 

Im happy to hear other opinions / possibilities regarding this. 

Thank you.

Koenkk:
Thanks, moved the links to the docs.

PR #4679, opened 2022-09-19
no show setup value in people cnt

nukusinji:


converters/fromZigbee.js
```diff
@@ -8155,9 +8155,12 @@ const converters = {
         convert: (model, msg, publish, options, meta) => {
             const lookup = {'0': 'idle', '1': 'in', '2': 'out'};
             const value = precisionRound(parseFloat(msg.data['presentValue']), 1);
+            const people = precisionRound(msg.data.presentValue, 0);
             let result = null;
-            result = {people: precisionRound(msg.data.presentValue, 0), status: lookup[value*10%10]};
-            return result;
+            if (value <= 80) {
+                result = {people: people, status: lookup[value*10%10]};
+                return result;
```
danieledwardgeorgehitchcock, line 8162: I don't think this will work as you are returning your result inside the IF statement - you'd want a result returned regardless of if the statement is TRUE or FALSE

nukusinji: ignore value > 81. It works fine.
The device use command value and 'people'. So command value need ignore.

danieledwardgeorgehitchcock: Here's how I'd approach it:

```javascript
            const lookup = {'0': 'idle', '1': 'in', '2': 'out'};
            const value = precisionRound(parseFloat(msg.data['presentValue']), 1);
            const people = precisionRound(msg.data.presentValue, 0);
            const result = {};
            if (value <= 80) {
                result.people = people;
                result.status = lookup[value*10%10];
            }
            return result;
        },
    },
```

nukusinji: when value is above 81, don't need return value.
ex) this is people count device. when value is 1. 
if user is change device setting like led enable.  the value monentary change 83. 
the user don't need '83' value. 
if using your code people value will change 1 -> 83 -> 1.

nukusinji: And z2m show this error. 
Exception while calling fromZigbee converter: Cannot set properties of null (setting 'people')}

danieledwardgeorgehitchcock: > when value is above 81, don't need return value. ex) this is people count device. when value is 1. if user is change device setting like led enable. the value monentary change 83. the user don't need '83' value. if using your code people value will change 1 -> 'unavailable'-> 1 in HA logs..

Have you actually tried my code? People value will only be 'unavailable' until the conditions for the IF statement are met (exactly the same as what your code is trying to achieve). The difference between my code and yours is that mine returns / resolves whereas yours doesn't.

nukusinji: > 

I tried your code. Z2M can not use return NULL. 
Your code only return 0~80 or NULL. 
My code only return 0~80, Another is ignore.

If i use command, z2m show this error 'Exception while calling fromZigbee converter: Cannot set properties of null (setting 'people')}'

danieledwardgeorgehitchcock: I've just updated the code in my comment. Could you try it again?

nukusinji: > I've just updated the code in my comment. Could you try it again?

Exception while calling fromZigbee converter: Cannot set properties of null (setting 'people')}

danieledwardgeorgehitchcock: Are you removing the device, restarting Z2M, then re-adding it while testing these scenarios?

Try that and if that fails, could you post up your whole converter so we can compare and do further fault finding?

nukusinji: > Are you removing the device, restarting Z2M, then re-adding it while testing these scenarios?
> 
> Try that and if that fails, could you post up your whole converter so we can compare and do further fault finding?

image

danieledwardgeorgehitchcock: OK, this discussion seems to be going nowhere

I will leave someone else to review this PR

Koenkk: @nukusinji if I understand you correctly, when the device passes any value >= 81 it is NOT the people count but some magic command right?

nukusinji: > @nukusinji if I understand you correctly, when the device passes any value >= 81 it is NOT the people count but some magic command right?

Yes. more than 81 is command.

Koenkk:
Thanks!

PR #4681, opened 2022-09-20
Add _TZ3000_ay8hciaf to TS0004

evangelo18:


devices/tuya.js
```diff
@@ -2542,7 +2542,7 @@ module.exports = [
         toZigbee: [],
     },
     {
-        zigbeeModel: ['TS0004'],
+        fingerprint: [{modelID: 'TS0004', manufacturerName: '_TZ3000_ay8hciaf'}],
```
Koenkk, line 2545: This should make no difference, every TS0004 will already be detected as such.

evangelo18: my apologies. I'm new to this

PR #4684, opened 2022-09-20
Update lonsonho.js

marthubner:
Adding missing piece of functionality "min_brightness". I just copied the code from Tuya devices _TZ3000_xfs39dbf and _TZ3000_7ysdnebc and tested that it is working.

### DIFF
--- a/devices/lonsonho.js
+++ b/devices/lonsonho.js
@@ -148,7 +148,6 @@ module.exports = [
         fromZigbee: extend.light_onoff_brightness().fromZigbee.concat([fz.tuya_min_brightness]),
         toZigbee: extend.light_onoff_brightness().toZigbee.concat([tz.tuya_min_brightness]),
         exposes: [e.light_brightness().withMinBrightness()],
-        extend: extend.light_onoff_brightness(),
     },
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

itavero:
@marthubner Did you have to do anything to get this working?
There's nothing about it in the documentation on the Zigbee2MQTT website, but it doesn't seem to work for the Lonsonho QS-Zigbee-D02-TRIAC-2C-LN (Model: TS110F; Manufacturer: _TYZB01_v8gtiaed) dimmer I have.

marthubner:
Hi @itavero, I also have the _TYZB01_v8gtiaed. This adds a "min_brightness" setting as seen here it the screenhot.
IMG_20221004_132842
When set the light does not go bellow this value when dimmed using a wall switch. It affects dimming via the wall switch only (physical switch connected to S1 or S2 pins of the device).

itavero:
Ah, I don't use the S1/S2. I was under the impression that it would map 1-100% brightness to min_brightness-100%, like some of my other dimmers can.

PR #4425, opened 2022-07-06
Updates acova.js

molusk:
Older TAFFETAS2 devices have different model name.

PR #4685, opened 2022-09-20
Add zigbeeModel to Acova

siiick:
Add new zigbeemodel for Acova Percale 2 heater.

### DIFF
--- a/devices/acova.js
+++ b/devices/acova.js
@@ -5,7 +5,7 @@ const reporting = require('../lib/reporting');
 
 module.exports = [
     {
-        zigbeeModel: ['PERCALE2 D1.00P1.01Z1.00'],
+        zigbeeModel: ['PERCALE2 D1.00P1.01Z1.00', 'PERCALE2 D1.00P1.02Z1.00'],
         model: 'PERCALE2',
         vendor: 'Acova',
         description: 'Percale 2 heater',


Koenkk:
Thanks!

molusk:
@siiick did you check the data/database.db entry of this device ? I had an issue where the first definition of the heater stopped working when I added the second one because of trailing odd characters in the modelId. #4425 

I am currently working on a issue I have with Taffetas model and noticed Percale is the only model without the 7 trailing \u0000 but maybe this heater works differently.

molusk:
I use File Editor module to browse the filesystem. database.db is in the config/zigbee2mqtt folder.
The file is right next to the external converters.

siiick:
you're right :
`{"modelId": "PERCALE2 D1.00P1.02Z1.00\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000"}`

but my percale2 are supported the way I declared the model

molusk:
In my case (with 2 different versions of Taffetas2 heaters), only one worked when I did not add the trailing odd stuff.
When I added them to one of the 2 modelIds, the other one wasn’t recognized anymore.
I had to add the odd characters to both modelIds and then everything works fine.

Maybe I am wrong but the risk of kipping model definition as they are is that the first model won’t be recognized anymore.
The best way to be sure would be to ask another owner if everything is fine with the new definition file (the author of the first integration maybe ?).
@acesyde do you still have Percale2 heaters and use z2m ? Can you confirm they still work with the last version of Zigbee2MQTT ?

acesyde:
@molusk it still works with the latest version of zigbee2mqtt

PR #2082, opened 2021-01-14
Add support for Niko Smart plug with earthing pin

sjorge:
Although the compliance document for the 552-80699 looks nearly identical to 170-33505 (suspicious) and claims rmsVoltage/rmsCurrent is supported, we get UNSUPPORTED_ATTRIBUTE while trying to setup reporting or reading the  linked attributes in readEletricalMeasurementMultiplierDivisors.

seMetering also cleans support for a few properties, but same story there, although currentSummDelivered did seem to take a reporting config. It never reported anything for me.

Just configuring activePower separately does seem to yield a somewhat correct power reading. My 1000lm Tradfri bulb reads 13 watt, behind an Ubisys S1 it reads 12-13 watt depending on the device. So it looks to be within expected variance of ~4W < 100 watt load and ~5% for > 100 watt loads.

PR #4699, opened 2022-09-24
Add support for Niko smart plug 552-80698

ldekamps:
The manufacturer Niko currently has two different kind of smart plugs available: [552-80698](https://www.niko.eu/en/products/niko-home-control/products-for-traditional-wiring/smart-plug-with-side-earthing-zigbee-productmodel-niko-20230a0a-a358-5569-b843-453098349fd1) and [552-80699](https://www.niko.eu/en/products/niko-home-control/products-for-traditional-wiring/smart-plug-with-pin-earthing-zigbee-productmodel-niko-59574367-4d82-576a-8142-5de4ecdad7b1).

The main difference between both devices is that model 552-80698 uses side earthing, while model 552-80699 uses pin earthing. PL https://github.com/Koenkk/zigbee-herdsman-converters/pull/2082 adds support for version 552-80699. Over the last few days, I have tested this integration for model 552-80698 and everything seems to work fine. It looks like both components are identical to each other, except for the pin difference.

I'm pretty sure that adding `Smart plug Zigbee SE` to the `zigbeeModel[]` array of the 552-80699 integration would have worked as well, but because of the `model`/`description` difference and the possibility that the manufacturer might use different firmware versions for both models in the future, I chose to add a completely new integration.

Code has passed linting and testing.

### DIFF
--- a/devices/niko.js
+++ b/devices/niko.js
@@ -132,6 +132,29 @@ module.exports = [
         ],
     },
     {
+        zigbeeModel: ['Smart plug Zigbee SE'],
+        model: '552-80698',
+        vendor: 'Niko',
+        description: 'Smart plug with side earthing pin',
+        fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fz.power_on_behavior],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #4701, opened 2022-09-24
Add support for Philips 9290031345

daniel-wassenberg:
The 5/6" model of this light is already supported (Philips 9290031346), added support for the 4" model (Philips 9290031345).


Koenkk:
Thanks!

PR #4705, opened 2022-09-25
support Moes heat valve  TS0601 _TZE200_7yoranx2

asidorka:
cheers,
got Moes heat valve from aliexpress which is identical to _TZE200_e9ba97vf, that I already have in household
image

  but started getting errors
**warn  2022-09-25 17:13:34: Received message from unsupported device with Zigbee model 'TS0601' and manufacturer name '_TZE200_7yoranx2'**

I suppose it  is just new revision of original Valve
chnages were tested locally by creating custom converter with copy-pasted code,everything works fine


Koenkk:
Thanks!

PR #4713, opened 2022-09-27
Add battery % to Develco MOSZB-140

sjorge:
Got a MOSZB-140 to look into the missing support for the more sensitive occupancy that works via msOccupancySensing, looks like this is going to be rather challenging. But that aside we're also missing a battery % estimate, which is easy enough to add. Hence this PR :)

### DIFF
--- a/devices/develco.js
+++ b/devices/develco.js
@@ -290,16 +290,20 @@ module.exports = [
         model: 'MOSZB-140',
         vendor: 'Develco',
         description: 'Motion sensor',
-        fromZigbee: [fz.temperature, fz.illuminance, fz.ias_occupancy_alarm_1],
+        fromZigbee: [fz.temperature, fz.illuminance, fz.ias_occupancy_alarm_1, fz.battery],
         toZigbee: [],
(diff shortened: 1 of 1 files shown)


devices/develco.js
```diff
@@ -290,16 +290,20 @@ module.exports = [
...
         configure: async (device, coordinatorEndpoint, logger) => {
-            const endpoint1 = device.getEndpoint(38);
-            await reporting.bind(endpoint1, coordinatorEndpoint, ['msTemperatureMeasurement']);
-            await reporting.temperature(endpoint1);
-            const endpoint2 = device.getEndpoint(39);
-            await reporting.bind(endpoint2, coordinatorEndpoint, ['msIlluminanceMeasurement']);
-            await reporting.illuminance(endpoint2);
+            const endpoint1 = device.getEndpoint(35);
+            await reporting.bind(endpoint1, coordinatorEndpoint, ['genPowerCfg']);
+            await reporting.batteryVoltage(endpoint1, {min: constants.repInterval.HOUR, max: 43200, change: 100});
```
sjorge, line 300: These are taken from the suggested settings from Develco.

Koenkk:
Thanks!

PR #4715, opened 2022-09-28
Add manufacturer name _TZE200_1ibpyhdc for Tuya motion sensor ZG-204ZL

Pferdebockwurst:


Koenkk:
Just added it myself before seeing this so this is not needed anymore.

PR #4720, opened 2022-09-29
Update xiaomi.js to correct https://github.com/Koenkk/zigbee2mqtt/issues/14236

Danit2:
Correct https://github.com/Koenkk/zigbee2mqtt/issues/14236


Koenkk:
I can't see how this solves the issue. It looks like a timeout issue. Can you remove this change and check if it still works. If not, then please provide the debug log with and without your change.

Danit2:
I have now tested and played a lot with the smoke detector.
My PR doesn't help with the problem, but it makes the selftest button a bit friendlier. I like it better this way.

The problem with the smoke detector is that after a command it does not accept any more commands for about 2 to 3 minutes.

But you can press the button on the detector and the command will be executed immediately.

I think this can only be fixed via the firmware.

Koenkk:
> My PR doesn't help with the problem, but it makes the selftest button a bit friendlier. I like it better this way.

I see now, thanks!

PR #4723, opened 2022-09-30
Add Hue Flourish light model 929003052601

brycelambert:
Add Philips Hue Flourish table lamp model 929003052601. Note this is the North American model that comes with an E26 bulb and technically different from the existing entry for the European model with an E27 bulb, model 4090431P9.

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -701,6 +701,13 @@ module.exports = [
         extend: hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}),
     },
     {
+        zigbeeModel: ['929003052601'],
+        model: '929003052601',
+        vendor: 'Philips',
+        description: 'Hue Flourish white and color ambiance table light with Bluetooth',
+        extend: hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}),
+    },
+    {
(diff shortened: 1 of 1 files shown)


devices/philips.js
```diff
@@ -700,6 +700,13 @@ module.exports = [
         description: 'Hue Flourish white and color ambiance table light with Bluetooth',
         extend: hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}),
     },
+    {
+        zigbeeModel: ['929003052601'],
+        model: '929003052601',
+        vendor: 'Philips',
+        description: 'Hue Flourish white and color ambiance table light with Bluetooth',
+        extend: hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}),
```
brycelambert, line 708: TBH I am not sure about this color temp range. This device has a std Hue E26 bulb. I looked around and can't find any specs for the color temp range & there's nothing in the zigbee2mqtt logs. I added as the specs require it.

Koenkk: Read this to figure out the color temp range: https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_3-adding-converter-s-for-your-device

Koenkk:
Thanks!

PR #4725, opened 2022-09-30
Update novo.js

sreisi:
Remove ignore_basic_report


devices/novo.js
```diff
@@ -10,8 +10,8 @@ module.exports = [
         model: 'C10-3E-1.2',
         vendor: 'Novo',
         description: 'Curtain switch',
-        fromZigbee: [fz.tuya_cover, fz.ignore_basic_report],
-        toZigbee: [tz.tuya_cover_control, tz.tuya_cover_options],
+        fromZigbee: [fz.tuya_cover],
+        toZigbee: [tz.tuya_cover_control],
```
Koenkk, line 14: Why? (and why was `tz.tuya_cover_options` also removed?)

sreisi: There are control problems with the device, so I removed the ignore_basic_report

Koenkk: I don't expect removing that converter to change anything, did you check this?

sreisi: I have already added the "tuya_cover_options" again yesterday.
My problem is that the device partially no longer responds once it has been completely Open or Closed.
So I looked at the other device integrations that use the same chip/device (TS0601), there the "ignore_basic_report" converters are removed. 
How can I do a pre-check, I thought I could only do that by playing the master branch on the Zigbee2Mqttt device?
Thanks for your patience ;)

Koenkk: That cannot be related to `ignore_basic_report`. Please create an issue in the z2m issue tracker instead.

PR #4731, opened 2022-10-02
MOSZB-140 : Add configurable occupancy endpoints

sjorge:


sjorge:
While out on a walk I just though of a possible way to handle this cleanly in a custom tz converter that then calls the correct one. Will need to find a way to check the endpoints it was called on though to then figure out which one to call. But In theory the idea should work I think.

sjorge:
That idea was easier to implement than I though, time to start looking into how to configure the temp/illuminance logic.

sjorge:
Hitting a new problem:
```
Zigbee2MQTT:error 2022-10-02 16:54:50: Failed to configure '0x0015bc001a0238ec', attempt 1 (Error: Read 0x0015bc001a0238ec/34 msOccupancySensing(["develcoArmThresholdMinTemperature"], {"sendWhen":"active","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":4117,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')
    at Endpoint.checkStatus (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:317:28)
    at Endpoint.read (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:480:22)
    at Object.configure (/Users/sjorge/Documents/repos/zigbee-herdsman-converters/devices/develco.js:593:17)
    at Configure.configure (/opt/zigbee2mqtt/lib/extension/configure.ts:115:13))
```

https://www.develcoproducts.com/media/1697/moszb-130-14x-technical-manual-motion-sensor-040721.pdf
image

https://github.com/Koenkk/zigbee-herdsman/blob/371a58547499e0acb5ec475549fb81c9d59ebcb3/src/zcl/definition/cluster.ts#L2560-L2562

I'm sure I added them correctly to herdsman, I also tried reading them without the manufactureCode set but same result as with it set.

Koenkk:
It looks good indeed, but maybe we are missing something. Can you make a sniff of just the read request?

sjorge:
> It looks good indeed, but maybe we are missing something. Can you make a sniff of just the read request?

Sure, I'll try to grab it this week, busy week ahead so might be in the weekend.

sjorge:
My development coordinator, so not the same key as my production one.
```
cc:7c:3d:c3:4d:66:28:11:53:dc:54:6c:db:9a:aa:07
```

This one is it joining without the reading the troublesome attributes:
[join_moszb140.pcapng.zip](https://github.com/Koenkk/zigbee-herdsman-converters/files/9709768/join_moszb140.pcapng.zip)

Here is a read to one of the troublesome attributes: (`zbee_nwk.src == 0xad2f || zbee_nwk.dst == 0xad2f`)
[read_msOccupancySensing.develcoTargetLevel.zip](https://github.com/Koenkk/zigbee-herdsman-converters/files/9709896/read_msOccupancySensing.develcoTargetLevel.zip)

I did 2 reads, one with specifying the manufacturerCode and one without, I used 2 test converters for this so I can just share the exact code:
```
        test: {
            key: ['test'],
            },
                await entity.read('msOccupancySensing', ['develcoTargetLevel'], manufacturerOptions);
            },
        },
        test2: {
            key: ['test2'],
            },
                await entity.read('msOccupancySensing', ['develcoTargetLevel']);
            },
        },
```

Koenkk:
Everything looks correct to me; the cluster `0x0406`, manufcode `0x1015` and attribute `0xfc02`. So it looks their docs are either not correct or maybe there are multiple revisions/fws for this device.

sjorge:
Yeah, I’m assuming there docs are wrong.I’ll pull out a few general fixes from this into a separate PR and close this one, no need to add the other occupancies if they are just going to be the same except a different timeout.~ sjorgeOn 5 Oct 2022, at 16:53, Koen Kanters ***@***.***> wrote:﻿
Everything looks correct to me; the cluster 0x0406, manufcode 0x1015 and attribute 0xfc02. So it looks their docs are either not correct or maybe there are multiple revisions/fws for this device.

sjorge:
Closed in favor of https://github.com/Koenkk/zigbee-herdsman-converters/pull/4741, will re-open if I do somehow end up hearing back from develco on this.

PR #4733, opened 2022-10-02
Add emergency_heating for heating to SLR2b/c

georgeduckett:
SLR2c (which I have) supports system mode emergency_heating for the heating control as well as water. The SLR2b (which I don't have) has information about using emergency_heating for controlling heating, but not water. I assume that also supports it, but the possible value for withSystemMode has not been included by mistake. This pull request adds it to both.

### DIFF
--- a/devices/hive.js
+++ b/devices/hive.js
@@ -344,7 +344,7 @@ module.exports = [
         },
         exposes: [
             exposes.climate().withSetpoint('occupied_heating_setpoint', 5, 32, 0.5).withLocalTemperature()
-                .withSystemMode(['off', 'auto', 'heat']).withRunningState(['idle', 'heat']).withEndpoint('heat'),
+                .withSystemMode(['off', 'auto', 'heat', 'emergency_heating']).withRunningState(['idle', 'heat']).withEndpoint('heat'),
             exposes.binary('temperature_setpoint_hold', ea.ALL, true, false)
                 .withDescription('Prevent changes. `false` = run normally. `true` = prevent from making changes.' +
                     ' Must be set to `false` when system_mode = off or `true` for heat').withEndpoint('heat'),
@@ -396,7 +396,7 @@ module.exports = [
         },
         exposes: [
             exposes.climate().withSetpoint('occupied_heating_setpoint', 5, 32, 0.5).withLocalTemperature()
-                .withSystemMode(['off', 'auto', 'heat']).withRunningState(['idle', 'heat']).withEndpoint('heat'),
+                .withSystemMode(['off', 'auto', 'heat', 'emergency_heating']).withRunningState(['idle', 'heat']).withEndpoint('heat'),
             exposes.binary('temperature_setpoint_hold', ea.ALL, true, false)
                 .withDescription('Prevent changes. `false` = run normally. `true` = prevent from making changes.' +
                     ' Must be set to `false` when system_mode = off or `true` for heat').withEndpoint('heat'),


Koenkk:
Thanks!

robertalexa:
That mode was purposely not exposed for any of the Hive boiler controllers. `emergency_heating` is an invalid value in the context of Home Assistant. Instead of a system mode, a preset should be used, which is not really (easily) achievable.

On top of that, the mode was not exposed because it creates more hassled than it resolves. Even without exposing that mode, you can still set the device into the native boost functionality (because `emergency_heating` has been added to acceptable values in the constants, and because the devices knows how to handle it). BUT, and this is a big BUT, the native boost can only be done via mqtt commands which need to be chained into 1 single command, if the timing is wrong, the device does not do what you want it to do, and instead resets the mode, but keeps the timer - basically a major bug!.

Because you can never point-and-click via the UI, exposing this can create issues for casual users. Instead, only people that ACTUALLY read the documentation will be successful in using the native boost. And this was already documented on z2m.io for all the Hive SLR family devices.

@Koenkk my personal recommendation would be for this change to be reverted, maybe a comment to be added to all the SLR devices (they all support this mode, for both heat and water if they are dual channel) specifying to not be re-added. Unfortunately I have only spotted this now, after the release :(

Koenkk:
@georgeduckett can you comment on this?

georgeduckett:
Hi @robertalexa, @Koenkk. Sorry for any trouble this caused. It's a little while ago now, but I do remember reading the docs. about having to send both commands in the same MQTT message. I think my aim was more to allow for home assistant to correctly report the state that it's in, although it was getting colder and there was only so long I wanted to muck around with it before my wife complained about the cold! I guess what you refer to @robertalexa (a preset) would be something like translating a change to emergency_heating to do the two commands together. I'm new to Zb2MQTT so not sure if that's a thing or not. Happy for this to be reverted if that's the best option

Koenkk:
i've reverted it, the preset option seems better indeed.

robertalexa:
@georgeduckett Thank you for your comments, much appreciated. You will find that despite you doing the right thing in z2m, HA will completely crap the bed because the status is invalid altogether. It also wont allow you to set a native boost from HA's end either.

The preset thing was a thing i always wanted to look into, but i think z2m will have to "forge" quite a few things when communicating with HA about this device. We also need to handle when HA sets a preset, and z2m needs to re-interpret it to actually do other actions. I think there are other zigbee thermostats that actually handle presets as zigbee attributes (i might be wrong though) while Hive have deviated from the zigbee standard and created their own status? - again, it has been over a year since i sorted out these devices and my memories are a bit foggy.

George, if you ever look into this, feel free to tag me and we can have a join adventure. Maybe when the cold season ends, so our wifes do not kill us :)

phoenix1589:
Hi,

For the dual channel receivers, there is a discrepancy between the behaviour in the emergency_mode for the heating (this pull request discussion) and that of the hot water - which had emergency_heating added in PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/3961 by @russdan.

@robertalexa, I can see you commented on that PR and whilst there wasn't a decision to revert that change, it feels odd that a different decision has been reached for the different endpoints (which I think is worse).

Having some way to know when it's in emergency_heat would be good. Without being a good enough programmer to actually suggest any code, having a read-only attribute (like link quality) that gets updated to be the value of system_mode sent by the thermostat - and the selectable system_mode (as is) remains with only off/auto/heat.

Now that it's almost time to get the heating on again, I remember I never quite finished my heating automation from earlier in the year.

robertalexa:
Hi @phoenix1589 

In all reality, that other PR should probably be reverted too to be consistent. However, because that is a water endpoint i do not think it creates an issue in HA, my device is single channel, and i dont remember the ha exposes code on top of my head. If memory serves right, and ha is not affected, then there is technically no harm in exposing it as a visual indicator.

As to the native heating boost, i would personally like it to be a nice preset and work nicely with ha, but lacked time to come up with sth. My current approach is to not use the native boost at all, especially since heating boost is different to water boost, again if i remember correctly from talking to other users. The tenperature is basically faked etc.

So in my case, i chose to not use it at all for now, and instead handle a temperature increase and a timer in node red, which is what i use alongside ha. Then go back to the previous temp when the timer runs out.

Will try to spend time on this in the spring, it is already cold where i leave and messing with the heating is not a great idea now :)

phoenix1589:
I'm not sure about the code in HA, but for the water endpoint, I get an 'empty' option for what I presume is emergency heat:
image

> especially since heating boost is different to water boost, again if i remember correctly from talking to other users. The tenperature is basically faked etc.

Yes, exactly, the heating setpoint is always 32 and the local temperature is always 21 - this will never change.



> Will try to spend time on this in the spring, it is already cold where i leave and messing with the heating is not a great idea now :)

So that's what I said last year - then spring came round and I didn't need heating so I got distracted with other things!

robertalexa:
@phoenix1589 Thanks for the screenshot. When i first tried it you weren't even getting that empty line.

I suspect if you set the mode to `emergency_heating` from z2m's side (frontend or mqtt), HA will throw an error around the lines of "unexpected value" sth sth. Would you be kind to confirm?

georgeduckett:
@phoenix1589, @robertalexa I'm not sure how to verify that error, but I guess I get the same. For the duration of the emergency_heating the associated HA climate doesn't get updated; I guess because HA isn't happy with the HVACMode so dosen't update anything.

robertalexa:
That is correct and it is why the native boost mode was not exposed to HA. It is included in the documentation because z2m supports it, but hive s implementation of the boost is done as a mode instead of a preset. Z2m does have anything yet to make these conversions and expose them to ha as ha wants. This may be something i will look into in the spring, when heating is not needed anymore

georgeduckett:
@robertalexa Ahh ok, that makes sense, thanks for taking the time to explain.

PR #4734, opened 2022-10-03
Added the slower mode of the Setpoint command for scheduled heatpoint changes

glad-wolf:
Added the Danfoss setpoint command with parameter 0 which is useful for scheduled heatpoint changes.

For Details see [Programming guide ](https://assets.danfoss.com/documents/202524/AU417130778872en-000101.pdf) on[ Danfoss Ally™ support](https://www.danfoss.com/en/products/dhs/smart-heating/smart-heating/danfoss-ally/danfoss-ally-support/#tab-documents) website

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -3563,6 +3563,18 @@ const converters = {
             return result;
         },
     },
+    danfoss_thermostat_setpoint_scheduled: {
+        cluster: 'hvacThermostat',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const result = {};
+            if (msg.data.hasOwnProperty('occupiedHeatingSetpoint')) {
+                result[postfixWithEndpointName('occupied_heating_setpoint_scheduled', msg, model, meta)] =
+                    precisionRound(msg.data['occupiedHeatingSetpoint'], 2) / 100;
+            }
+            return result;
+        },
+    },
     danfoss_icon_battery: {
         cluster: 'genPowerCfg',
         type: ['attributeReport', 'readResponse'],
--- a/devices/danfoss.js
+++ b/devices/danfoss.js
@@ -17,7 +17,8 @@ module.exports = [
         description: 'Ally thermostat',
         whiteLabel: [{vendor: 'Danfoss', model: '014G2463'}],
         meta: {thermostat: {dontMapPIHeatingDemand: true}},
-        fromZigbee: [fz.battery, fz.thermostat, fz.thermostat_weekly_schedule, fz.hvac_user_interface, fz.danfoss_thermostat],
+        fromZigbee: [fz.battery, fz.thermostat, fz.thermostat_weekly_schedule, fz.hvac_user_interface,
+            fz.danfoss_thermostat, fz.danfoss_thermostat_setpoint_scheduled],
         toZigbee: [tz.danfoss_thermostat_occupied_heating_setpoint, tz.thermostat_local_temperature, tz.danfoss_mounted_mode_active,
             tz.danfoss_mounted_mode_control, tz.danfoss_thermostat_vertical_orientation, tz.danfoss_algorithm_scale_factor,
             tz.danfoss_heat_available, tz.danfoss_heat_required, tz.danfoss_day_of_week, tz.danfoss_trigger_time,
@@ -27,7 +28,7 @@ module.exports = [
             tz.thermostat_weekly_schedule, tz.thermostat_clear_weekly_schedule, tz.thermostat_programming_operation_mode,
             tz.danfoss_window_open_feature, tz.danfoss_preheat_status, tz.danfoss_adaptation_status, tz.danfoss_adaptation_settings,
             tz.danfoss_adaptation_control, tz.danfoss_regulation_setpoint_offset,
-            tz.danfoss_thermostat_occupied_heating_scheduled_setpoint],
+            tz.danfoss_thermostat_occupied_heating_setpoint_scheduled],
         exposes: [e.battery(), e.keypad_lockout(), e.programming_operation_mode(),
             exposes.binary('mounted_mode_active', ea.STATE_GET, true, false)
                 .withDescription('Is the unit in mounting mode. This is set to `false` for mounted (already on ' +
@@ -49,6 +50,11 @@ module.exports = [
                 .withDescription('Values observed are `0` (manual), `1` (schedule) or `2` (externally)'),
             exposes.climate().withSetpoint('occupied_heating_setpoint', 5, 35, 0.5).withLocalTemperature().withPiHeatingDemand()
                 .withSystemMode(['heat']).withRunningState(['idle', 'heat'], ea.STATE),
+            exposes.numeric('occupied_heating_setpoint_scheduled', ea.ALL)
+                .withValueMin(5).withValueMax(35).withValueStep(0.5).withUnit('°C')
+                .withDescription('Scheduled change of the setpoint. Alternative method for changing the setpoint. In the opposite ' +
+                  'to occupied_heating_setpoint it does not trigger an aggressive response from the actuator. ' +
+                  '(More suitable for scheduled changes.)'),
             exposes.numeric('external_measured_room_sensor', ea.ALL)
                 .withDescription('If `radiator_covered` is `true`: Set at maximum 30 minutes interval but not more often than every ' +
                 '5 minutes and 0.1 degrees difference. Resets every 35 minutes to standard. If `radiator_covered` is `false`: ' +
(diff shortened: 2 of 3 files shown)


devices/danfoss.js
```diff
@@ -26,7 +26,8 @@ module.exports = [
             tz.thermostat_keypad_lockout, tz.thermostat_system_mode, tz.danfoss_load_balancing_enable, tz.danfoss_load_room_mean,
             tz.thermostat_weekly_schedule, tz.thermostat_clear_weekly_schedule, tz.thermostat_programming_operation_mode,
             tz.danfoss_window_open_feature, tz.danfoss_preheat_status, tz.danfoss_adaptation_status, tz.danfoss_adaptation_settings,
-            tz.danfoss_adaptation_control, tz.danfoss_regulation_setpoint_offset],
+            tz.danfoss_adaptation_control, tz.danfoss_regulation_setpoint_offset,
+            tz.danfoss_thermostat_occupied_heating_scheduled_setpoint],
```
Koenkk, line 30: Please also add an `exposes.numeric` for this in the list below (such that you can easily set this value from the frontend under the exposes tab)

glad-wolf: @Koenkk I tried this in this branch [danfoss_expose_scheduled_setpoint](https://github.com/Koenkk/zigbee-herdsman-converters/compare/master...glad-wolf:zigbee-herdsman-converters:danfoss_expose_scheduled_setpoint) but it does not work well because it modifies the same value as the already exposed `occupied_heating_setpoint` does.

The new exposed value can be used to change the setpoint but it does not always reflects the current setpoint.

glad-wolf: @Koenkk I renamed the option as suggested. And I extended fromZigbee to also update the new option. Now it works well (also with ea.All).

converters/fromZigbee.js
```diff
@@ -3560,6 +3560,10 @@ const converters = {
                     result[postfixWithEndpointName('running_state', msg, model, meta)] = 'idle';
                 }
             }
+            if (msg.data.hasOwnProperty('occupiedHeatingSetpoint')) {
```
Koenkk, line 3563: This converter is also used by the Danfoss Icon thermostat, do you know if it also supports the `occupied_heating_setpoint_scheduled`, if yes then also add it to that device.

glad-wolf: @Koenkk  I separated the fromZigbee part (& squashed the last commits)

Koenkk:
Thanks!

robertalexa:
@glad-wolf I like your approach :) Just wanted to double check, or give you some background info even, maybe you it will help you using the device :)

Some time ago, all zigbee softwares where having the same issue about this TRV, a super slow response to changing the occupied heating setpoint, there are a lot of threads on the internet with sooo many people saying the TRV is bad purely on this late behaviour.

Turns out that as you have already figured it out, the command was passed as if the internal scheduler made that change. Which is something that the PID does when a schedule is programmed on the device, instead of the user made the change (which of course they did, they sent the mqtt command :) ) but the TRV didn't get told that.

In reality I will 100% recommed NOT to use the internal scheduler, with the major fault being that when you change your batteries, it all gets lost, and you will have to set it all again. That, combined with the fact that z2m has no "smart automation" capability, it basically means you would rely on a different platform to handle your schedule. At which point, you will 100% want to use the aggressive change, unless you want your house to warm up 2 to 3 hours later than your scheduled time :).

To be perfectly honest, I can't really think of why you would want to expose this. If you set an internal schedule, the PID will already set the setpoint using the schedule command. And if you want the schedule yourself, you will never want to use that command. But I like the approach and the fact you made it available if people trully want it - but from a usage perspective it would be wrong

glad-wolf:
Hi @robertalexa,

I used both commands for complete last winter in 3 rooms. (I simply added two buttons to my UI to turn the heating on - one for aggressive variant and one for the slow variant).

The **non aggressive variant** does exactly what it should do: it reaches the target temperature in the time defined by "Control algorithm scale factor" (I use a value of 5 wich means the traget temperature should be reached in 30 Minutes - which it does plus minus 5 minutes). I think the 'Adaption Run' must have been run successfully to achieve this behavior. More information can be found in the [programming guide](https://assets.danfoss.com/documents/202524/AU417130778872en-000101.pdf) on this [page](https://www.danfoss.com/en/products/dhs/smart-heating/smart-heating/danfoss-ally/danfoss-ally-support/#tab-documents).
e.g. cap. 2.1 states
> NOTE: The aggressive command (type 1) must not be used all the time for schedule changes!!! If done, it
> could reduce comfort, energy efficiency and battery lifetime!

The **aggressive variant** simply completely opens the valve for about 5 minutes and then the PID kicks in. In my environment (wich is not ideal, but I can't change it) this means it overshoots the traget temperature several °C and then it takes about 2-3h until the room reaches the (lower) target temperature (this wastes energy and to have really hot radiator in a cool room is not very comfortable)
It is still useful  if you wan't get room get warm quickly (or in other environments with lower inlet temperatures or smaller radiators).

I think both variants are useful - it really depends on the use case and the environment.

The first variant of this PR simply added a second toZigbee command (which was enough to trigger it via MQTT) without exposing it. But this makes hard for people to find it.

surrealo:
Hi @glad-wolf ,

I just noticed this.Thanks very much. I hope you don't mind my asking,  I'm using the home assistant zigbee2mqtt plugin and was wondering whether it's normal that when changing occupied_heating_setpoint_scheduled through the exposes interface it's normal that shortly afterwards occupied_heating_setpoint value changes to mirror the value.?

robertalexa:
Yes that is expected. The new attribute is setting the occupied setpoint but with a different level of aggressiveness. The 2 values would remain in sync :) the new entity does not refer to a scheduled setpoint, but rather a setpoint with the same aggressiveness as when the internal scheduler would do

surrealo:
cool, that functionality was needed, in the mornings the heating was way too aggressive.

bnapalm:
@glad-wolf 

Thanks for submitting this. I was very frustrated that it was impossible (without custom changes) to send a "normal" mode setpoint command, in both Zigbee2MQTT as well as ZHA, and only the "aggressive" mode was possible. In the end I used custom "extended" converter to enable this locally, but this works much better (mine was SET-only, so wouldn't show a value), and it's great that this option is there for the user to choose, without the need for custom tweaks.

robertalexa:
I remember a time when all the open source zigbee forums were unanymously upset and saying that Dafoss Ally (and the rebrands) are awful and that they take hours to kick in, boilers are burning gas heating water and the thermostats were super slowly opening the valve causing a waste. I changed the default aggressiveness, posted on forums, everyone else implemented this, and now we went full circle :)

I still believe that the slow valve opening is a bad approach and a waste of gas/money at boiler level and the over shooting should be taken into account as part of the temperature scheduling. This way you maximise the use of the hot water that the boiler worked for immediately, meaning your main thermostat room reaches the desired temp quickly and tells the boiler to stop heating more water. The temp overshooting, if it is happening, can be rectified by using different temperatures for the start of the day. For example, night time 18C, desired day temp, 21C. At 6am set the temperature to 19/20 (depending on how much overshooting you experience). At 8am set your temps to 21C. This way you get the same overall temperature just from the excess heat radiated from your rardiators instead of having a hard working boiler and radiators refusing to warm up, in the fear of overshooting.

I have 8 of these TRVs so I have seen the "slow" aggressiveness for a whole winter and I was never able to do a schedule were the rooms were ACTUALLY at the desired temperature, they always felt sluggish and the boiler was burning gas - frustrating.

I would however want to learn more about your approach using the slow aggressiveness, and maybe we can help each other have more efficient heating systems :)

rsporsche:
I second glad-wolf. Having the same problems with aggressive setpoint changes in my apartment. I use the scheduled mode for scheduled changes via home assistant but manual changes via the climate device in home assistant (and google home by extension) result in the same 2-3 hours of sweltering heat described by glad-wolf. I would like an option for the climate device to expose the non aggressive mode by default.

Also, the Danfoss ZigBee cluster specification mentions a flag for the algorithm scale factor attribute that can disable the quick open feature?

```
Range 1-10 (lower 4 bit allocated to scale factor)
Scale factor of setpoint filter timeconstant
("aggressiveness" of control algorithm) 1=5min(Quick) ...
5=30min(Moderate) ... 10=80min(Slow).
Bit4=Quick open feature disable. 1=disable. 0=enable
```

e.g. the aforementioned Bit4. Is/Can this be exposed somehow and does anyone know if that would change the behaviour of manual changes via physically turning the TRV?

surrealo:
I've been tinkering with my TRVs again and another scenario where occupied_heating_setpoint_scheduled seems to be useful is when a window is opened and then closed again. 

Since the open window feature is broken when radiator covered is enabled I've been setting the setpoint to something low  (12) and of course once the window is shut again and the temperature changes to the regular setting the TRV becomes way too aggressive and overshoots. Using scheduled things seem to be more reasonable. I'll have to see over time.

The main problem with the overshooting with aggressive temperatures is that the radiator stops heating for a long time which is unpleasant and leads to inconsistent heat.

Retr0Joe:
I think, adding the "slower" mode leaded to new problems, see: https://github.com/Koenkk/zigbee2mqtt/discussions/20910.

When you change temperature localy or remotely, every time "occupied:heating_setpoint_scheduled" will be set. New suggestion how to change this behaviour is at the topic i´ve made.

Issue #4737, opened 2022-10-05
Philips Hue Flourish white and color ambiance ceiling light with Bluetooth- 929003053501


SelfDestructor:
Found a similar issue (#4136) so I'm assuming that this fits here instead of the Zigbee2MQTT issue tracker.

Zigbee model 929003053501 is not supported.

Should be the same as:
    {
        zigbeeModel: ['4090531P9', '929003053601'],
        model: '4090531P9',
        vendor: 'Philips',
        description: 'Hue Flourish white and color ambiance ceiling light with Bluetooth',
        extend: hueExtend.light_onoff_brightness_colortemp_color(),
    },

SelfDestructor:
With the help of Rene over at the Zigbee2MQTT Discord I learned how to create a converter for the lamp and test it out. Created pull request  #4738 to add this Hue Flourish lamp.

PR #4738, opened 2022-10-05
Update philips.js

SelfDestructor:
Added Hue Flourish white and color ambiance ceiling light with Bluetooth model 929003053501  #4737


Koenkk:
Thanks!

PR #4742, opened 2022-10-05
Remove battery reporting and add tuya magic TS0041, TS0041, TS0042, TS0043

robertalexa:
This PR is an outcome of https://github.com/Koenkk/zigbee2mqtt/issues/8072

I have left feedback as part of that thread, for the sake of consistency.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1596,18 +1596,13 @@ module.exports = [
         exposes: [e.battery(), e.action(['1_single', '1_double', '1_hold', '2_single', '2_double', '2_hold',
             '3_single', '3_double', '3_hold', '4_single', '4_double', '4_hold'])],
         toZigbee: [],
-        configure: async (device, coordinatorEndpoint, logger) => {
-            const endpoint = device.getEndpoint(1);
-            await endpoint.read('genBasic', [0x0004, 0x000, 0x0001, 0x0005, 0x0007, 0xfffe]);
-            await endpoint.write('genBasic', {0xffde: {value: 0x13, type: 0x20}});
-
-            /*
-             * reporting.batteryPercentageRemaining removed as it was causing devices to fall of the network
-             * every 1 hour, with light flashing when it happened, extremely short battery life, 2 presses for
-             * action to register: https://github.com/Koenkk/zigbee2mqtt/issues/8072
-             * Initially wrapped in a try catch: https://github.com/Koenkk/zigbee2mqtt/issues/6313
-             */
-        },
+        configure: tuya.configureMagicPacket,
+        /*
+         * reporting.batteryPercentageRemaining removed as it was causing devices to fall of the network
+         * every 1 hour, with light flashing when it happened, extremely short battery life, 2 presses for
+         * action to register: https://github.com/Koenkk/zigbee2mqtt/issues/8072
+         * Initially wrapped in a try catch: https://github.com/Koenkk/zigbee2mqtt/issues/6313
+         */
     },
     {
         fingerprint: [{modelID: 'TS004F', manufacturerName: '_TZ3000_xabckq1v'}, {modelID: 'TS004F', manufacturerName: '_TZ3000_czuyt8lz'}],


Koenkk:
- Is battery % still reported after this change?
- Can you also try with `configure: tuya.configureMagicPacket,` like: https://github.com/Koenkk/zigbee-herdsman-converters/blob/9c6d8c67ccf258f865e646971149555bb8a3a4fc/devices/tuya.js#L1175

robertalexa:
Battery starts as null but is being populated after a few clicks
image

`configureMagicPacket` doesn't have `await endpoint.write('genBasic', {0xffde: {value: 0x13, type: 0x20}});` only the `read`. I am guessing you are asking me to see if the device still behaves as expected without the write?

robertalexa:
With just `configureMaficPacket` it appears the Battery value does not get populated - although i will leave this running for some time to confirm for sure. Will also confirm is the device does re-announcements every hour.

The really really weird one about these devices is that they refuse to pair with my network via router devices (Hive thermostat receiver, ikea light power supply, double socket, cheap ewelink smart power socket, ikea repeater, hive repeater, tuya gas sensor) in situ, i have to get it close to the coordinator, it pairs first try, then if i move it back in situ (away from coordinator) and it happily connects to a router device..... not a real problem, just a really annoying quirk

robertalexa:
Device has been stable after 2 hours of inactivity. Responded first press, no re-announcement, battery level is also being reported. Would you like me to update my PR to just use the magic method?

Koenkk:
>  Would you like me to update my PR to just use the magic method?

Yes please!

devices/tuya.js
```diff
@@ -1596,10 +1598,15 @@ module.exports = [
         toZigbee: [],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
-            await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg']);
-            try {
-                await reporting.batteryPercentageRemaining(endpoint);
-            } catch (error) {/* Fails for some*/}
+            await endpoint.read('genBasic', [0x0004, 0x000, 0x0001, 0x0005, 0x0007, 0xfffe]);
+            await endpoint.write('genBasic', {0xffde: {value: 0x13, type: 0x20}});
```
Koenkk, line 1602: I believe you missed this one.

robertalexa:
All done. I have left the comments in there though, just in case someone in future will touch this again. I know it is a bit much for each, but the devices have been separated from each other before - so leaving the comment on just the TS0041 might not be enough?

Happy to change this if you find it necessary.

PS: over 24h in production now and all the switches have reacted 1st attempt every time, i have not seen any re-joins. So it looks positive.

Of course i do not have the 2,3 and 4 click devices but i wouldn't expect it to be different at all. Looking at my devices the board already has traces, just missing buttons and faceplate. I assume also firmware, but i doubt it handles battery differently etc

Sorry about the other device, i have pushed another commit

Koenkk:
Thanks!

Koenkk:
No

jan--m:
Just wanted to mention that my battery indicators are now working! (with the z2m dev version)

Screenshot 2022-10-12 at 10 03 39

I assume re-enabling PowerCfg reporting would be a bad idea as that would drain the battery?

robertalexa:
That would indeed yes (with possibly a lack of the magic tuya configuration). The devices should be exposing the battery on their own. This is something that if paired from scratch will start without a value, an after a few minutes the device passes a value. I would expect that as the battery drains more the device will continue to publish an updated value. But because it has been little time since we did this, and the fact that the batteries are meant to last more than a year, we will have to wait and see. All of my devices are still at 100% battery but have had little use. If all of this is correct, than the actual reporting configuration every hour is unnecessary and as has been the case so far, a bad idea :) (the are other whitelabel devices with the same model number that simply refuse the reporting configuration. Others accept it but then crash and reboot when it is meant to do it.)

Hope this helps!

PR #26, opened 2018-06-30
device update to support nanoleaf smrt ivy bulb

matchett808:
https://nanoleaf.me/en-gb/consumer-led-lighting/for-owners/get-support/technical-specifications/

It's the NANOLEAF SMART IVY - homeassistant.js PR to come

PR #4744, opened 2022-10-05
Add min temperature read/write for Moes BHT-002

bwindsor:
This adds the ability to read/write the minimum temperature for the external sensor on the Moes BHT-002 when in AL sensor mode.

The `max_temperature_limit` range also seems to be incorrect so I have corrected that (was 0-35C, now set to 20-70C). The purpose of `max_temperature_limit` is to provide an override to cut the thermostat out if the external floor sensor exceeds this temperature. It appears to be only used when in AL sensor mode.

~~I'll submit a PR for the z2m docs updating this description as well.~~ I'll put it in the `exposes` description what this actually means so that the autogenerated docs get updated.


Koenkk:
Thanks!

csszokolai:
Hi Guys, after I installed the update, I have to check this improvement. I received this error:

`2022-11-01 21:36:08zigbee-herdsman-converters:Moes BHT-002: Unrecognized DP #26 with data {"dp":26,"datatype":2,"data":{"type":"Buffer","data":[0,0,0,5]}}`

ilvi80:
"The max_temperature_limit range also seems to be incorrect so I have corrected that (was 0-35C, now set to 20-70C). The purpose of max_temperature_limit is to provide an override to cut the thermostat out if the external floor sensor exceeds this temperature. It appears to be only used when in AL sensor mode."

This is not true. An external temperature sensor is used to monitor the floor temperature. The device supports a maximum of 35, so it is necessary to return to the previous value - 35. This setting gives a limit for the thermostat, that is, what maximum temperature can be set in current_heating_setpoint. Thus, if max_temperature_limit = 33, then current_heating_setpoint cannot be greater than 33. The device supports the maximum value for current_heating_setpoint = 35

ilvi80:
max_temperature_limit = (0-35), it is true

ilvi80:
> This adds the ability to read/write the minimum temperature for the external sensor on the Moes BHT-002 when in AL sensor mode.
> 
> The `max_temperature_limit` range also seems to be incorrect so I have corrected that (was 0-35C, now set to 20-70C). The purpose of `max_temperature_limit` is to provide an override to cut the thermostat out if the external floor sensor exceeds this temperature. It appears to be only used when in AL sensor mode.
> 
> ~I'll submit a PR for the z2m docs updating this description as well.~ I'll put it in the `exposes` description what this actually means so that the autogenerated docs get updated.

This is not true. An external temperature sensor is used to monitor the floor temperature. The device supports a maximum of 35, so it is necessary to return to the previous value - 35. This setting gives a limit for the thermostat, that is, what maximum temperature can be set in current_heating_setpoint. Thus, if max_temperature_limit = 33, then current_heating_setpoint cannot be greater than 33. The device supports the maximum value for current_heating_setpoint = 35 and max_temperature_limit = 35
max_temperature_limit = (0-35), it is true

bwindsor:
@ilvi80 I think you are right, I misunderstood. I've submitted #4887 to revert this change.

BRSS73:
> Hi Guys, after I installed the update, I have to check this improvement. I received this error:
> 
> `2022-11-01 21:36:08zigbee-herdsman-converters:Moes BHT-002: Unrecognized DP #26 with data {"dp":26,"datatype":2,"data":{"type":"Buffer","data":[0,0,0,5]}}`

I have this problem too and it locks up the thermostat after a few minutes.
Only a reboot will unlock it for a while.
I have switched back to version 1.28.0 in the meantime.

bwindsor:
Look like I should have included a change to `fromZigbee` to decode data point 26 as well, but I only added sending data point 26 to the device. I'm not sure why I didn't do that to be honest. I'll try to do this tomorrow.

BRSS73:
Hi @bwindsor, have you implemented the change?

bwindsor:
Hi @BRSS73 thanks for the reminder - looks like I must have forgotten about this. Now changed in #5029, thanks for picking up on it!

BRSS73:
Can't wait!
Thanks @bwindsor

PR #4747, opened 2022-10-06
Remove withSensor() from Climate in favour of separate select

bwindsor:
In line with not having too many things bundled together under `exposes.climate` (see https://github.com/Koenkk/zigbee2mqtt/pull/14368#issuecomment-1270332591) this creates a separate `exposes.enum` for sensor selection rather than mingling it in with the rest of the Climate related things.


Koenkk:
Thanks!

PR #4750, opened 2022-10-07
Several updates to TV02-Zigbee

vladyspavlov:
- Moved `heating_stop` to `system_mode` and added description.
- Added description to presets.
- Updated minimum temperature limit.
- Added `battery` converter and expose.
- Updated converter and description for `holiday_start_stop` (holiday mode) to be more clear.
- Updated `working_day` converter and expose.
- Added converter for schedule and description.
- Updated `online` expose description to be more clear.


vladyspavlov:
Maybe it will be better to move this converter from global converters file to fzLocal and tzLocal?

Koenkk:
I would prefer to migrate this device to the new TuYa format. You can find an example for a thermostat here: https://gist.github.com/Koenkk/116d770830217a1bca3ccbd456c6e94a

vladyspavlov:
Merge only if https://github.com/Koenkk/zigbee-herdsman-converters/pull/4755 will be merged

vladi1234:
Hello @vladyspavlov,

At the moment everything is working very well and correctly, why such big changes?
Do you have `schedule` is of course interesting, I have to test.

devices/tuya.js
```diff
@@ -1969,40 +1969,89 @@ module.exports = [
...
+        meta: {
+            tuyaDatapoints: [
+                [2, 'preset', tuya.valueConverterBasic.lookup({'auto': 0, 'manual': 1, 'holiday': 3}),
+                    {datatype: tuya.sendDataPointEnum}],
+                [8, 'open_window', tuya.valueConverter.onOff],
+                [10, 'frost_protection', tuya.valueConverter.onOff],
+                [16, 'current_heating_setpoint', tuya.valueConverter.divideBy10],
+                [27, 'local_temperature_calibration', tuya.valueConverter.localTempCalibration],
+                [31, 'working_day', tuya.valueConverterBasic.lookup({'mon_sun': 0, 'mon_fri+sat+sun': 1, 'separate': 2}),
+                    {datatype: tuya.sendDataPointEnum}],
```
Koenkk, line 2033: Instead of defining the dataType in meta I suggest the following.

- In `lib/tuya.js` define a class called `Enum`.
- Replace e.g. `'mon_sun': 0` with `'mon_sun': tuya.Enum(0)`
- In `toZigbee.js` do a check like `else if (typeof convertedValue === 'Enum') {` then use `tuya.sendDataPointEnum`

devices/tuya.js
```diff
@@ -1969,40 +1969,89 @@ module.exports = [
...
+                [106, 'schedule', tuya.valueConverter.thermostatScheduleDay],
+                [107, 'system_mode', tuya.valueConverterBasic.lookup({'heat': false, 'off': true})],
+                [115, 'online', tuya.valueConverter.onOff],
+                [108, 'schedule_monday', tuya.valueConverter.thermostatScheduleDay],
+                [112, 'schedule_tuesday', tuya.valueConverter.thermostatScheduleDay],
+                [109, 'schedule_wednesday', tuya.valueConverter.thermostatScheduleDay],
+                [113, 'schedule_thursday', tuya.valueConverter.thermostatScheduleDay],
+                [110, 'schedule_friday', tuya.valueConverter.thermostatScheduleDay],
+                [114, 'schedule_saturday', tuya.valueConverter.thermostatScheduleDay],
+                [111, 'schedule_sunday', tuya.valueConverter.thermostatScheduleDay],
```
Koenkk, line 2052: Checking with @bwindsor  (https://github.com/Koenkk/zigbee-herdsman-converters/pull/4735) 

It would be good to do some unification between the different TuYa thermostat schedule formats that are available now. But from what I see here the structure is different; Moes BHT-002 uses 1 datapoint and this thermostat uses multiple

vladyspavlov:
> At the moment everything is working very well and correctly, why such big changes?

Added schedule and related stuff as described earlier. Also moved `heating_stop` to `system_mode`, so in Home Assistant climate entity modes work correctly, I think it's more clear.
The lower temperature limits for each mode are also updated, because TRV supports the temperature in the range [5, 30].
Updated `holiday_start_stop` to be more clear.
Replaced `battery_low` with `battery`, as TRV reports a numeric value.
And additional description to explain how all these things work.

vladi1234:
>The lower temperature limits for each mode are also updated, because TRV supports the temperature in the range [5, 30].

I have already corrected 5-30 several times.
Yes, in the meantime, someone has changed from 5-30 to 0-30, that's no problem, it must be done correctly on 5-30.

>Replaced battery_low with battery, as TRV reports a numeric value.

For info Low Battery is not percentage and not numerical. You have to read Manual, see below:
grafik

I also tested this myself.

vladyspavlov:
> For info Low Battery is not percentage and not numerical. You have to read Manual

I know about description in manual, but I sniffed traffic and TRV sends messages like this one:
`00702302000400000064`
Datatype is 02 (4 byte value) and value is 64 which is 100 in dec.
Also I checked tuya iot portal and the battery datapoint has this description:
Min 0 Max 100 Step 1

So I think that it reports a percentage. Currently, I don't have the low voltage batteries to test with. If you can, please check it. Anyway even if it reports only low battery there is no issue here. I have seen situation many times in tuya where something isn't documented, but it works.

vladi1234:
Have you seen any other number except 0 and 100?
I have not seen.
Ich habe das auch selbst getestet.

devices/tuya.js
vladyspavlov: I have made changes according to your suggestions. Review, please.

vladi1234:
> Moved heating_stop to system_mode and added description.

Why is it necessary to change heating_stop to system_mode? I need heating_stop.

> Updated converter and description for holiday_start_stop (holiday mode) to be more clear.

What is currently unclear in 'holiday_start_stop'?
Then what does your `holiday_start_stop` look like visually to you?
Can you take screenshot?

vladyspavlov:
> Why is it necessary to change heating_stop to system_mode? I need heating_stop.

If you use a home assistant, you can control the TRV mode within climate entity without templating or something else. Why use a separate expose when there is already one that is perfect for it?

> What is currently unclear in 'holiday_start_stop'?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/df1bacac6100763cad6a896652828b59f1b7463c/converters/fromZigbee.js#L4116-L4118
vs
https://github.com/Koenkk/zigbee-herdsman-converters/blob/2c7cdd34c60613e49ad1d49ca2b5b7374fe24bb8/lib/tuya.js#L1256-L1258

Shorter form, no large spaces, it's easier to read and in combine with the description, easier to understand how to write data. I know, the toZigbee converter hasn't changed, but the previous form may confuse, as for me. Anyway, I've been trying to improve the converter and your work has helped a lot. I think we should wait for Koen's answer.

vladi1234:
> > Why is it necessary to change heating_stop to system_mode? I need heating_stop.
> 
> If you use a home assistant, you can control the TRV mode within climate entity without templating or something else. Why use a separate expose when there is already one that is perfect for it?

Why does a zigbee2mqtt converter have to have settings for HA? 
https://github.com/Koenkk/zigbee2mqtt/issues/8522#issuecomment-999039475

I and many other users do not Homeassistant.

vladyspavlov:
> Why does a zigbee2mqtt converter have to have settings for HA?

I provide the example of a home assistant, this is one of a bonuses that you get when you use exposures correctly
https://github.com/Koenkk/zigbee-herdsman-converters/blob/df1bacac6100763cad6a896652828b59f1b7463c/lib/exposes.js#L389-L395
It is better to use it than to create a new one.

vladi1234:
Hello @vladyspavlov,

I know that `system_mode` is very familiar to me.

According to the Koenkk, this particular fragment was undesirable a year ago. It was negotiated and arranged.

Therefore, the TRV functionality appeared as it was developed by the manufacturers, without adding additional HA functions.

Not the whole world revolves around Homeassistant!

I use `heating_stop` in other automation systems, and other users and fans have adapted their own automation systems in the same way for a long time, and they are very happy with it.

The TV02-Zigbee thermostat has been working without problems for about a year.

I repeat it already because you don't want to understand. This is already becoming very intrusive on your part.
It's not good.

Me, other users and fans have put so much work and effort into TV02-Zigbee.

And now you don't need to change all the converters at once, because, in your opinion, they are bad.

As I have already written, "You have a `schedule` and a `working_day`, which, of course, is very interesting, I have to check."

vladyspavlov:
> Have you seen any other number except 0 and 100?
> I have not seen.
> I also tested this myself.

I tested with different batteries twice and you are right. It doesn't reports anything except 0 or 100. Fixed it. Thanks!

devices/tuya.js
vladyspavlov: I think it's necessary to create several unified converters, because some devices have a similar data format, while others have a completely different data format.
For example, these converters can be combined into one with params to configure such things like the number of periods, temperature converting, etc:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/8df9506531ab2dab43174a826d58ecdb016a0c6c/converters/toZigbee.js#L3453-L3463
https://github.com/Koenkk/zigbee-herdsman-converters/blob/8df9506531ab2dab43174a826d58ecdb016a0c6c/converters/toZigbee.js#L5106-L5119
https://github.com/Koenkk/zigbee-herdsman-converters/blob/9379ccbad81a2087097ca3141ef572f6660f9487/devices/tuya.js#L253-L267
For example, them can be combined to function with the next params
`tuya.valueConverter.tuyaThermostatSchedule1({ periods: 12, convertTemp: (t) => t * 2 })`

Some converters look completely different, e.g.:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/8df9506531ab2dab43174a826d58ecdb016a0c6c/converters/toZigbee.js#L4811-L4822
It's difficult to combine with the previous ones.

Function chaining can be a way out. It can be created many little function-utils that transform data into the required form. Something like this, declarative form:
`tuya.valueConverter.split(' ').schedule({ periods: 4, periodSize: 3 }).condition({ temperature: { min: 5, max: 35 } }).convert({ temperature: (t) => t * 2 });`
And the schedule method should return a strict object, so it can be easily used in the next method e.g.:
`{ p1: { hour: 10, minute: 20, temperature: 20.5 }, p2: ... }`
But it seems a bit complicated to me, I don't think it's a good idea.

vladi1234:
> > Have you seen any other number except 0 and 100?
> > I have not seen.
> > I also tested this myself.
> 
> I tested with different batteries twice and you are right. It doesn't reports anything except 0 or 100. Fixed it. Thanks!
> 
> > Not the whole world revolves around Homeassistant!
> 
> Z2M is not a stand-alone system, it's a layer between raw Zigbee data and some understandable format. If this format will have better compatibility with some of home automation system, I think it should be used. If your home automation system doesn't support it, you probably shouldn't care if it's `heating_stop` or `system_mode`. So I see no reason why this is a bad decision.
> 
> > they are very happy with it.
> 
> I don't know how happy they are. I don't think it's good to talk about other people without any research to back that opinion up. Looks like manipulation. Speaking for myself, I would be much more comfortable using `system_mode`. I already explained why it is better.
> 
> > Me, other users and fans have put so much work and effort into TV02-Zigbee.
> > And now you don't need to change all the converters at once, because, in your opinion, they are bad.
> 
> I respect your work. It's a big deal. I don't know why you think I'm trying to ruin your work. I don't think that converters are bad. They are good and can be improved.
> 
> This discussion about `heating_stop` is turning into an endless debate.

**Yes, you're right.**
> This discussion about `heating_stop' is turning into an endless debate.

**You're not really listening to me!**

**Please read it again correctly.**

**Now in plain language:
I have no objection if system_mode is added.**

**I am very much against it if,** "Moved heating_stop to system_mode and added description".

**Again:
heating_stop must stay!**

vladi1234:
>>   they are very happy with it.

> I don't know how happy they are. I don't think it's good to talk about other people without any research to back that opinion up. Looks like manipulation. Speaking for myself, I would be much more comfortable using system_mode. I already explained why it is better.

This is a serious accusation.

You think in such a way that manipulation is, that it is your problems that you think in such a way.

We worked together with users and fans for about a year.

And you just jumped in yesterday. And you only make heavy accusations and a lot of wind.

vladi1234:
Dear @vladyspavlov ,

You only have to imagine in about a year how many thermostats were sold?
I think about XX-Thousands pieces!
And if suddenly there is no more heating_stop.
Then about XX-thousands of problems will come out of it.

rotilho:
I think `heating_stop` makes more sense be part of `system_mode` IMO but we should preserve `heating_stop` for compatibility reasons.

vladi1234:
Hello @vladyspavlov,
thanks for your code for schedule.
i refined and tested it a bit. It works, but there are still small errors.

image

vladyspavlov:
> i refined and tested it a bit. It works, but there are still small errors.

Could you explain more please? I don't see errors on your screen. As I see there are some differences with code that commited here, could you show your version? Maybe it will help to find issues. Or try using the current version of the code.

vladi1234:
> Could you explain more please? I don't see errors on your screenshot.

The temperature belongs to the other segment-time.
grafik

> As I see there are some differences with code that commited here, could you show your version?

My code not much different. I took your schedule code core. I refined cosmetics and connected converters correctly.
It will be watching on my side tomorrow.

> Or try using the current version of the code.

I have always collected corrections and low limit 5 - 30 I want to change tomorrow.
I can also, then make pull-request current version schedule-converter, just as it is.

vladi1234:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/4773

vladyspavlov:
> The temperature belongs to the other segment-time.

This is how it works in the tuya app. The first segment is the range from 00:00 to time in the provided schedule and the next segments are ranges from previous to next. 24:00 segment also required.
I described it in description, check it.

vladi1234:
Then I did right.

      `Only end time of segment is displayed. Example: 00:00 to 06:00/17 C°`

grafik

vladyspavlov:
> I think `heating_stop` makes more sense be part of `system_mode` IMO but we should preserve `heating_stop` for compatibility reasons.

@Koenkk What do you think about multiple keys for a signle datapoint in the new tuya format?
Would it be ok to allow an array of keys and an array of converters?
For example:
`[dp, ['key1', 'key2'], [tuya.valueConverter.onOff, tuya.valueConverter.trueFalse]]`
`'key1'` will be converted with `tuya.valueConverter.onOff` and `'key2'` with `tuya.valueConverter.trueFalse`

vladyspavlov:
> grafik

Maybe this format makes it easier to understand how it works the first time, but otherwise it's a part without any functionality that needs to be repeated every time in every schedule, so I think it's better to leave it in the description.

Koenkk:
@vladyspavlov here you can find an example on how to return multiple keys for a single dp:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/8a343597059db0f0189b5f91a42b3252792f5ee2/devices/tuya.js#L2322

https://github.com/Koenkk/zigbee-herdsman-converters/blob/8a343597059db0f0189b5f91a42b3252792f5ee2/lib/tuya.js#L1209

vladyspavlov:
> @vladyspavlov here you can find an example on how to return multiple keys for a single dp

@Koenkk Thanks. It works, but only from device, not to device, which I need.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/8a343597059db0f0189b5f91a42b3252792f5ee2/lib/tuya.js#L1224-L1225
I think the lines above can be replaced with these to allow "to" with null keys:
```
const dpEntry = datapoints.find((d) => d[1] === convertedKey || d[1] === null);
if (!dpEntry || dpEntry[1] === undefined) {
```
What do you think?

vladyspavlov:
@vladi1234 Blocked me, I can't leave comments in him PR, so I'll leave it here, maybe he will read it.

> it may take some time, but now there is a problem with a user vladyspavlov who wants to completely remove all thermostat modules and replace them with others.

As you can see in the original PR, I'm currently working on leaving `heating_stop` for compatibility. As I understand it, this is your main complaint, so here is a compromise to solve it.

I'm not replacing all the modules, it's just a migration to the new tuya format. You will have the same functionality with the same key names.

Koenkk:
@vladyspavlov 

> Thanks. It works, but only from device, not to device, which I need.

What if you add the data point multiple times? one with `system_mode` key and one with `heating_stop`. This should do it for the toZigbee part.

vladyspavlov:
@Koenkk 
> What if you add the data point multiple times? one with `system_mode` key and one with `heating_stop`. This should do it for the toZigbee part.

Oh, yes, it works like a charm. Thanks!
Another question, is there any way to send to multiple datapoints? In other words, is it possible to migrate parts like this one without any changes to the new format "architecture"?
https://github.com/Koenkk/zigbee-herdsman-converters/blob/96563772e92eb459bf52c10e241c3a725968bcab/converters/toZigbee.js#L3542-L3545
By default, if you turn off frost protection mode, it will reset to auto. This code prevents it. I don't think that this logic needs to be done on the converter side, but for compatibility it would be good to leave it.

rotilho:
@vladyspavlov out of curiosity, how does the step property from calibration entity appear to you? Does it show step 1 or step 0.1?

vladyspavlov:
> @vladyspavlov out of curiosity, how does the step property from calibration entity appear to you? Does it show step 1 or step 0.1?

@rotilho If we talk about z2m, it shows step 0.1. If about HA, it shows step 1

rotilho:
@vladyspavlov thanks man!

Koenkk:
> By default, if you turn off frost protection mode, it will reset preset to auto. This code prevents it. I don't think that this logic needs to be done on the converter side, but for compatibility it would be good to leave it. The same situation with heating stop.

The current code seems to set it to manual. What if a different system_mode was set before, does it set it to a different system_mode?

vladyspavlov:
> The current code seems to set it to manual. What if a different system_mode was set before, does it set it to a different system_mode?

It always changes `preset` to `auto` if you change `system_mode` to `heat`. It doesn't matter what mode or preset was before it. This is by design, how it works out of the box. This code sets `preset` to `manual`, because in some cases it's more convenient.

vladi1234:
> I would prefer to follow this principle then (follow the standard behaviour as much as possible). This also reduces converter complexity.

Thermostat has its own logic, even switched to auto as the default. If you leave it like that, then TRV runs after every switch to Schedule, which makes the whole z2m project a nothing, because then you can no longer control it with home automation.

vladi1234:
> I would prefer to follow this principle then (follow the standard behaviour as much as possible). This also reduces converter complexity.

Thermostat has its own logic, even switched to auto as the default. If you leave it like that, then TRV runs after every switch to Schedule, which makes the whole z2m project a nothing, because then you can no longer control it with home automation.

This makes TRV an advantage, so it is neutral and you can use both modes.

vladyspavlov:
> If you leave it like that, then TRV runs after every switch to Schedule, which makes the whole z2m project a nothing, because then you can no longer control it with home automation.

You can easily add an additional action in your automation to switch it back to manual preset. There is no problem, except for editing the automations that use it.

lib/tuya.js
```diff
@@ -1199,7 +1220,7 @@ const valueConverterBasic = {
 
 const valueConverter = {
     trueFalse: valueConverterBasic.lookup({1: true, 0: false}),
-    onOff: valueConverterBasic.lookup({'ON': true, 'OFF': false}),
+    onOff: {from: (v) => v ? 'ON' : 'OFF', to: (v) => v === 'ON'},
```
vladyspavlov, line 1223: @Koenkk
Datapoint with `online` key sometimes reports data with an enum datatype, so the current version of the converter doesn't handle it. Here are the messages I sniffed:
`01007301000101`
`00527301000100`
`00347304000101`

Is it ok to modify this converter or is it better to create a new one specifically for this case?

rotilho:
If it keep changing to automatic I think the logic to force manual makes sense.

I wish we had a way to collect feedback from smart devices, it's a pandora's box sometimes.

Koenkk:
@vladyspavlov @vladi1234  regarding the force back to manual. Seems there is no best choice here, both have it cons. Therefore I propose to stick to the default behavior of the thermostat (switching back to auto)

vladyspavlov:
> If it keep changing to automatic I think the logic to force manual makes sense.

And it forces the current temperature setpoint to 20 degrees, so anyway, after changing `system_mode`, you need to add an additional step for setting your own setpoint.
I don't see the reason why it can't be added one more step in automation to switch it back to manual alongside with setting setpoint.

And why should it be a manual preset? What if i prefer a holiday preset, which is also supported? Do I need to add this to the converter? No, of course. It's better to leave the default logic of the device and correct it with your automations to your tasks and preferences.

rotilho:
I disagree with the conclusion but understand the reasoning.

I'd prefer having same class devices behaving in similar way even if this require some magic in the converter side. This approach allows more cross agonistic automations and reduce time researching each device specifics.

vladyspavlov:
> I'd prefer having same class devices behaving in similar way even if this require some magic in the converter side. This approach allows more cross agonistic automations and reduce time researching each device specifics.

Devices are mostly added by the community and different community members may have different expectations, also each device may have (and mostly has) its own logic. Additional guidelines will raise the difficulty threshold for writing converters. I think it's will be something similar to ZHA and I can said that ZHA have much simpler converters with much less exposes. This a double-edged sword.

vladyspavlov:
@Koenkk 
It looks like there is an issue with the frontend, or maybe you can better clarify what it is. A dropdown changes the value only if it is applied. It looks like it's state-based and until you apply it, the state doesn't change, so the dropdown value doesn't change. Here's a video for demonstration:

https://user-images.githubusercontent.com/16994749/197086820-e6589644-ea60-4875-9609-ef3366eb221e.mp4

I think it's because of composite expose. Can I fix it in some way from the converter?

Koenkk:
This definitely looks like a bug in the frontend, can you open an issue here so @nurikk can take a look at it? https://github.com/nurikk/zigbee2mqtt-frontend/issues

vladyspavlov:
> can you open an issue here so @nurikk can take a look at it?

Yes, of course.

The converter is now fully functional (apart from the composite expose issue, which is not a converter issue) and I have nothing to add to it at this time. Can you do a review? Are there any corrections to be made or can PR be accepted?

devices/tuya.js
```diff
@@ -2078,40 +2078,94 @@ module.exports = [
...
+                'a preset week programming temperature time and temperature. Before using these properties, check `working_day` ' +
+                'property. Each day can contain up to 10 segments. At least 1 segment should be defined. Different count of segments ' +
+                'can be defined for each day, e.g., 3 segments for Monday, 5 segments for Thursday, etc. It should be defined in the ' +
+                'following format: `hours:minutes/temperature`. Minutes can be only tens, i.e., 00, 10, 20, 30, 40, 50. Segments should ' +
+                'be divided by space symbol. Each day should end with the last segment of 24:00. Examples: `04:00/20 08:30/22 10:10/18 ' +
+                '18:40/24 22:50/19.5`; `06:00/21.5 17:20/26 24:00/18`. The temperature will be set from the beginning/start of one ' +
+                'period and until the next period, e.g., `04:00/20 24:00/22` means that from 00:00 to 04:00 temperature will be 20 ' +
+                'degrees and from 04:00 to 00:00 temperature will be 22 degrees.'),
             exposes.text('schedule_monday', ea.STATE),
```
Koenkk, line 2124: What is the relation between the `schedule_DAY` and composite `schedule` above?

Koenkk:
> Can you do a review? Are there any corrections to be made or can PR be accepted?

Done, I have one more question and after that this can be merged 🥳

devices/tuya.js
vladyspavlov: As each day has its own DP and the new schedule should be sent to the other one, and with the new Tuya format limitations, it was decided to do it in a way like this. The `schedule_day` is used to display the schedule of the specific day, whereas `schedule` composite is used to set a schedule for the specific day.

Koenkk: I see, looks good 😄

narendrastark:
> > Thermostat has its own logic, even switched to auto as the default.
> 
> In auto preset it works according to a defined schedule.
> 
> > If you leave it like that, then TRV runs after every switch to Schedule, which makes the whole z2m project a nothing, because then you can no longer control it with home automation.
> 
> You can easily add an additional action in your automation to switch it back to manual preset. There is no problem, except for editing the automations that use it. Forcing it back to manual preset also resets setpoint to 20 degrees, even if it was different before.



> > Thermostat has its own logic, even switched to auto as the default.
> 
> In auto preset it works according to a defined schedule.
> 
> > If you leave it like that, then TRV runs after every switch to Schedule, which makes the whole z2m project a nothing, because then you can no longer control it with home automation.
> 
> You can easily add an additional action in your automation to switch it back to manual preset. There is no problem, except for editing the automations that use it. Forcing it back to manual preset also resets setpoint to 20 degrees, even if it was different before.

This implementation makes the use of the scheduler component, which until now was a very convenient way to create time-based automations via GUI, unusable. The scheduler component only allows to change the temperature if the system_mode is set to heat each time. But then the temperature configured for auto mode applies and not the one I set in the scheduler component. Setting temperature without operation mode may lead to the thermostat completely ignoring the command (if it is in operation mode off) so that is no option.

Hessenpower01:
> > I would prefer to follow this principle then (follow the standard behaviour as much as possible). This also reduces converter complexity.
> 
> Thermostat has its own logic, even switched to auto as the default. If you leave it like that, then TRV runs after every switch to Schedule, which makes the whole z2m project a nothing, because then you can no longer control it with home automation.
> 
> **This makes TRV an advantage, so it is neutral and you can use both modes.**

This is a disadvantage that makes it unusuable. I want to control it via HA, with all the additional functionality ad not only time based.

Illunis:
since my heating is now completely broken and going crazy. Is there an easy way, without digging deep into the code, to reset the thermostats without a full downgrade and stay there? I don't want to fiddle with my heating until after the winter

holantomas:
@Illunis You can switch back zigbee2mqtt to version 1.28.0 or this package to version v14.0.645. Both are releases before this madness was merged.

Illunis:
> @Illunis You can switch back zigbee2mqtt to version 1.28.0 or this package to version v14.0.645. Both are releases before this madness was merged.

I did this. 
do external converters overwrite the built-in ones?
I can update again, If I do my research and rebuild the old converter (as external converter) from the removed parts, right?  I have a bit of coding experience, but I'm quite new to the "zigbee (and zigbee2mqtt) universe".

holantomas:
@Illunis asking wrong person 😁 but I think if you using git installation, you just revert to tag 1.28.0, then delete npm modules and install them again(this will download correct version of converter). If you using docker, then just use tag 1.28.0 instead of lastest or whatever you using and all should be done automatically.

For git installation, it will be good to look what update.sh doing during update, it should help you to get downgrade steps

holantomas:
@Illunis I just successfully downgrade to 1.28.0. I'm using docker so just stopped container, pull old image, do data backup, and start container with old tag. Everythings works great like before :)

timderspieler:
Any news on this? Heating is completely broken right now. That’s not something I would like to deal with in this winter when literally everybody’s heating bill explodes. PLEASE can somebody take a look at this?

PR #4754, opened 2022-10-07
[SRTS-A01](Aqara Smart Radiator Thermostat E1) add external sensor

artem-sedykh:


devices/xiaomi.js
```diff
@@ -121,10 +139,93 @@ const tzLocal = {
...
+                    await entity.write('aqaraOpple', {0xfff2: {value: val1, type: 0x41}}, {manufacturerCode: 0x115f});
+                    await entity.write('aqaraOpple', {0xfff2: {value: val2, type: 0x41}}, {manufacturerCode: 0x115f});
+
+                    await entity.read('hvacThermostat', ['localTemp']);
+                }
+                break;
+            }
+            case 'sensor_temp':
+                if (meta.state['sensor'] === 'external') {
+                    const temperatureBuf = Buffer.allocUnsafe(4);
```
Koenkk, line 202: Why not use `Buffer.alloc`? (https://www.w3schools.com/nodejs/met_buffer_allocunsafe.asp)

devices/xiaomi.js
```diff
@@ -2282,16 +2383,28 @@ module.exports = [
...
+        exposes: [
+            exposes.switch()
+                .withState('state', true, 'Turn off the thermostat', ea.ALL, 'OFF', 'ON'),
+            exposes.climate()
+                .withSetpoint('occupied_heating_setpoint', 5, 30, 0.5)
+                .withLocalTemperature(ea.STATE)
+                .withPreset(['manual', 'away', 'auto']).setAccess('preset', ea.ALL),
+            e.temperature_sensor_select(['internal', 'external']).withAccess(ea.ALL),
+            exposes.binary('calibrated', ea.STATE, true, false)
+                .withDescription('Is the valve calibrated'),
```
Koenkk, line 2395: Can you explain a bit more what this does?

artem-sedykh: "calibrated": the sign of valve calibration has the values 1 - the valve is calibrated and 0 - the valve is not calibrated
"sensor": internal - internal temperature sensor is used; external - an external temperature sensor is used

devices/xiaomi.js
artem-sedykh: the allocUnsafe method works faster than alloc, I'll fix it for alloc

devices/xiaomi.js
Koenkk: But was does "valve calibrated" mean?

artem-sedykh: Valve finds extreme positions, first it closes completely, and then it opens completely, after that it is considered that the thermal head can control the temperature

Koenkk:
Thanks!

PR #4755, opened 2022-10-08
Updates to new TuYa format

vladyspavlov:
These commits extends current functionally to simplify creating of converters.

Currently, you don't have the ability to set data to a value type other than boolean or number. This commit adds more data types.
https://github.com/Koenkk/zigbee-herdsman-converters/commit/735f32f52e549e42b486fb922c2eb85f588c070f

It would be useful to be able to set custom data types for each datapoint. You can use already defined converter, but send it with a different data type. It also allows you to specify that you want to send enum or bitmap data, that can't currently be sent in the new format.
https://github.com/Koenkk/zigbee-herdsman-converters/commit/ea4eeee7f69059fe144c00a1c5055512b0f13ee2

Meta in a converter can be very useful. For example, it allows to use the state of another property and create more complex logic.
https://github.com/Koenkk/zigbee-herdsman-converters/commit/b43400bc4fa4de0a96d2ccabdf32dce8e90473d8

### DIFF
--- a/lib/tuya.js
+++ b/lib/tuya.js
@@ -1227,7 +1227,7 @@ const tzDataPoints = {
             }
             if (dpEntry[3] && dpEntry[3].skip && dpEntry[3].skip(meta)) continue;
             const dpId = dpEntry[0];
-            const convertedValue = dpEntry[2].to(value);
+            const convertedValue = dpEntry[2].to(value, meta);
             let sendDataPoint;
             if (dpEntry[3] && dpEntry[3].datatype) sendDataPoint = dpEntry[3].datatype;
             if (sendDataPoint !== undefined) {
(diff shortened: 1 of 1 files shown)


vladyspavlov:
An example of using more data types
https://github.com/Koenkk/zigbee-herdsman-converters/blob/7a0060eb6b8548a6dc221f5d9a476931c7192466/devices/tuya.js#L2043

An example of using custom data types
https://github.com/Koenkk/zigbee-herdsman-converters/blob/7a0060eb6b8548a6dc221f5d9a476931c7192466/devices/tuya.js#L2026

An example of using meta in converter
https://github.com/Koenkk/zigbee-herdsman-converters/blob/7a0060eb6b8548a6dc221f5d9a476931c7192466/lib/tuya.js#L1267

vladyspavlov:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/4750

PR #4760, opened 2022-10-09
[SRTS-A01](Aqara Smart Radiator Thermostat E1) Replace state with system mode

duvholt:
Before (in Home Assistant):
image
image

After:
image
image


Koenkk:
Thanks!

Growiel:
Hi there, sorry if this is not the right place, but with the current implementation, it seems that the color of the thermostat (orange) depends on the selected mode (off or heat).

Is that on purpose ? I would expect it to show me if the valve is currently opened (and heating) by being orange and turn gray when the temperature is reached.

PR #4762, opened 2022-10-09
Adding 'very high' and 'max' to level sensitivity of hue motion sensors

MattL0:
I think this is simpler, and works here for both old and new philips sensors.


Koenkk:
Clean, thanks!

SaschaDe:
Is it possible that the option is only available for the new version? With the older motion detectors, "max" and "very high" are not available.

PR #4764, opened 2022-10-10
Add _TZE200_ji1gn7rw, a Tuya compatible switch

ArthurMLago:
I bought some switches here in Brazil that showed up as unsupported. I started reading on how to add new devices,  but it looked too much like a generic Tuya compatible switch. I tried just adding it's manufacturer Name to the list in tuya.js. And it seems to have worked!
Pics of the switch:
mynewswitch
mynewswitch2

Please let me know if I'm doing something wrong, or should conduct further testing.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1326,7 +1326,10 @@ module.exports = [
         },
     },
     {
-        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_nkjintbl'}],
+        fingerprint: [
+            {modelID: 'TS0601', manufacturerName: '_TZE200_nkjintbl'},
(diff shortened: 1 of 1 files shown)


Koenkk:
Since this is a 2 gang switch, can you add it to `TS0601_switch_2_gang` instead of `TS0601_switch`?

ArthurMLago:
sure thing!(sorry for taking so long)

Koenkk:
I see it is still under `TS0601_switch` and not `TS0601_switch_2_gang `.

ArthurMLago:
Sorry, messed up last commit, should be fine now

Koenkk:
Thanks!

PR #4771, opened 2022-10-11
Added support for Philips 929003054601 Fair

danieledwardgeorgehitchcock:
Resolves Koenkk/zigbee2mqtt#14430

O/P didn't have color temperature range from device, but I have based values on a similar age product with the same kelvin range specification.


Koenkk:
Thanks again @danieledwardgeorgehitchcock !

PR #4776, opened 2022-10-11
Support more features for Sunricher ZG9092A and fix time sync

pvainio:
- Push time to device every 24h (from Namron branded same device)
- Read seMetering multiplier and divisor on init for energy reporting to work
- Add keypad lock, brightness, display auto off
- Tune reporting intervals to avoid constant traffic
- Fix display auto off boolean value reversed for Namron (same device, same firmware)
- Fix local temperature exposed twice, directly and through climate


Koenkk:
Thanks!

PR #4777, opened 2022-10-12
Add delta attribute for converters using simulated brightness

dsypniewski:
This PR adds additional `delta` attribute alongside `brightness` for simulated brightness related updates.

Background:
Just `brightness` is fine for devices that can't get out of sync but in any event the devices get out of sync relying only on `brightness` causes the state to jump considerably eg.
Remote sets light brightness to 80%, someone later changes that light brightness manually to 30% the next time someone uses the remote the brightness will jump back to around 80% because that's the state saved by this module.

This could have been hacked around by comparing from and to states but because brightness is limited to a range that leaves edge case (literal ones at the edges of the range :grin:) where the state couldn't be updated anymore because the difference would be 0 (simulated brightness is at 255 and someone is trying to increase it).

Adding this `delta` attribute opens way for more seamless automations which take into account the actual current state of device no matter how it was changed.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1218,9 +1218,12 @@ const converters = {
             addActionGroup(payload, msg, model);
 
             if (options.simulated_brightness) {
+                const currentBrightness = globalStore.getValue(msg.endpoint, 'simulated_brightness_brightness', 255);
                 globalStore.putValue(msg.endpoint, 'simulated_brightness_brightness', msg.data.level);
                 const property = postfixWithEndpointName('brightness', msg, model, meta);
                 payload[property] = msg.data.level;
+                const delta_property = postfixWithEndpointName('delta', msg, model, meta);
+                payload[delta_property] = msg.data.level - currentBrightness;
             }
 
             return payload;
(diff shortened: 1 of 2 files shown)


Koenkk:
What bulbs are you controlling with this? If these are Zigbee bulbs, it is much nicer to use the move command: https://www.zigbee2mqtt.io/devices/81809_81813.html#moving-stepping

dsypniewski:
Brightness was just an example, I'm using it to control media players volume which doesn't have such feature.
Also with using the start/stop events in HA I find that sometimes the stop event get's lost somehow and then the moving command would only stop after reaching either end.

Koenkk:
> Also with using the start/stop events in HA I find that sometimes the stop event get's lost somehow and then the moving command would only stop after reaching either end.

Getting no stop event is problematic, are you sure this is the case? Because if no stop event is generated the timer will run endlessly thus: https://github.com/Koenkk/zigbee-herdsman-converters/blob/96563772e92eb459bf52c10e241c3a725968bcab/converters/fromZigbee.js#L1300

dsypniewski:
> Getting no stop event is problematic, are you sure this is the case? Because if no stop event is generated the timer will run endlessly thus:
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/96563772e92eb459bf52c10e241c3a725968bcab/converters/fromZigbee.js#L1300

I haven't looked into actual cause of this but it probably was something in my HA automation.
Anyhow, I switched to simulated brightness and since then I haven't had this problem so I'm pretty sure that this extension is receiving the stop events just fine.

dsypniewski:
@Koenkk is there something I could do to help move this forward or at least make clear that this is not desired?

Koenkk:
I'm fine with merging this if the property name is changed the `action_brightness_delta`. The `action_` prefix ensures the property is not persisted in the payload (since this is not a state)

Koenkk:
Thanks!

PR #4784, opened 2022-10-14
Added new fingerprint for Fantem ZB006-X

danieledwardgeorgehitchcock:
Resolves Koenkk/zigbee2mqtt#14477

### DIFF
--- a/devices/fantem.js
+++ b/devices/fantem.js
@@ -9,10 +9,11 @@ const tuya = require('../lib/tuya');
 
 module.exports = [
     {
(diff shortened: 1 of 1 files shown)


bladehstream (changes requested):
Line 16 modifies the description to remove the reference to neutral being optional; neutral is still optional for this device.

danieledwardgeorgehitchcock:
If the device has an optional neutral, is it worth mentioning that in the description? I noticed that the converter itself handles the actual power method. You'll see in my previous push I tweaked the title before removing it as I Don think this needs to be stated (and will keep the converter more universal for other small revisipns)

Koenkk:
Thanks!

PR #4787, opened 2022-10-15
Fix gas reporting for HS3CG

mikevlz:
HS3CG reports iasZoneStatus 34 when gas detected/test button pressed, so it should  use fz.ias_gas_alarm_2 to report gas leak correctly 34 & 1 -> false (ias_gas_alarm_1)
34 & 1<<1 -> true (ias_gas_alarm_2)
Issue: https://github.com/Koenkk/zigbee2mqtt/issues/14495


Koenkk:
Thanks!

PR #4790, opened 2022-10-16
Develco: Read Sw/Hw version for MOSZB-140 and minor cleanup

sjorge:
Merge after https://github.com/Koenkk/zigbee-herdsman/pull/597


sjorge:
Some future work would probably be to move this to just read these during interview instead.

Koenkk:
Thanks!

PR #4726, opened 2022-09-30
Develco: cleanup + add led_control/occupancy_timeout for MOSZB-140

sjorge:
Requires https://github.com/Koenkk/zigbee-herdsman/pull/588

@Koenkk 
I think the led_control and occupancy_timeout is valid for the MOSZB-130 too as per de docs. I do not have one of these devices however. Seems like endpoint 35 is the default one on a lot of devices including the MOSZB-140, do you want me to add them to the MOSZB-130 in a followup PR? (I'll be done a few more to finish support for the 3 configurable msOccupancySensing endpoints.

PR #4793, opened 2022-10-17
Fix INVALID_DATA_TYPE setting Develco InterfaceMode

msp1974:
Since PR #4726, there is an error when setting interface mode and it cannot be set.  This PR corrects that error and allows setting of interface mode again.


sjorge:
LGTM, seems I forgot to drop the `.develco` when refactoring some stuff. Sorry about that.

I double checked and it was the only one I forgot to update.

Koenkk:
Thanks!

PR #4795, opened 2022-10-17
Fix wrong window status reporting for BRT-100-TRV

xworld21:
The window detector of the BRT-100-TRV comes out with the wrong state. I understand that the corresponding `fromZigbee` already maps `0` to `OPEN` and `1` to `CLOSED`, so the states should not be flipped again in the exposed sensor.

I suspect that the similar TuYa TRV has the same problem but I don't have the device to check.

(I am new here, I hope sending a PR directly without issue is ok!)

### DIFF
--- a/devices/moes.js
+++ b/devices/moes.js
@@ -369,7 +369,7 @@ module.exports = [
         exposes: [
             e.battery(), e.child_lock(), e.eco_mode(), e.eco_temperature(), e.max_temperature().withValueMax(45), e.min_temperature(),
             e.valve_state(), e.position(), e.window_detection(),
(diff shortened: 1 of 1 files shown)


Koenkk:
Did you confirm that changing this fixes the issue?

xworld21:
> Did you confirm that changing this fixes the issue?

Yes. To be clear, the frontend always reported the correct status. Rather, the status would appear flipped in Home Assistant because of `payload_off: "OPEN"`. With this patch, both frontend and Home Assistant show the correct status.

Koenkk:
I see, thanks!

PR #4800, opened 2022-10-18
Fix left/right controlled together for TuYa TS0012_switch_module. Fixes #14541

fvillena:


devices/tuya.js
```diff
@@ -2536,6 +2536,8 @@ module.exports = [
         },
         meta: {multiEndpoint: true},
         configure: async (device, coordinatorEndpoint, logger) => {
+            await device.getEndpoint(1).read('genBasic',
+                ['manufacturerName', 'zclVersion', 'appVersion', 'modelId', 'powerSource', 0xfffe]);
```
Koenkk, line 2540: Can you check if the device still reports it state when you:

- Remove
```js
            await reporting.bind(device.getEndpoint(2), coordinatorEndpoint, ['genOnOff']);
            device.powerSource = 'Mains (single phase)';
            device.save();
```
- Repair the device
- Check if state is still reported when you control the switch physically.

PR #4801, opened 2022-10-18
Inovelli VZM35-SN fix for energy reporting

nathanfiscus:
The VZM35-SN is missing some parameters in the metering cluster that are queried in the energy reporting helpers preventing the device from configuring.  This replaces the reporting helpers with the raw commands.


Koenkk:
Thanks!

PR #4804, opened 2022-10-20
Remove preferHS from Osram Flex RGBW

desertblade:
Remove preferHS: true from OSRAM Flex RGBW 4052899926110.  This causes compatibility issues for people running Firmware build date is 20160204N****, which is newer than the firmware available for OTA.

This change should still allow use of hs_color in Home Assistant automations and the use of the color wheel. 

Tested xy color in my environment with no issues. Unable to validate hs_color since my lights do not support it.

This resolves issue: https://github.com/Koenkk/zigbee2mqtt/issues/13990

### DIFF
--- a/devices/osram.js
+++ b/devices/osram.js
@@ -227,7 +227,7 @@ module.exports = [
         model: '4052899926110',
         vendor: 'OSRAM',
         description: 'Flex RGBW',
-        extend: extend.ledvance.light_onoff_brightness_colortemp_color({colorTempRange: [125, 666], supportsHS: true, preferHS: true}),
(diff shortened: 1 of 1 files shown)


Koenkk:
CC: @CorePoint  https://github.com/Koenkk/zigbee2mqtt/issues/13597 , seems not all versions support HS so the preferHS I added earlier has to be reverted. You can change the color mode used in HA by following https://www.zigbee2mqtt.io/guide/usage/integrations/home_assistant.html#changing-supported-color-modes

CorePoint:
@Koenkk 
Thank you very much for letting me know!

PR #4808, opened 2022-10-20
Added device RB-279-T, clone of RB-278-T

Status-418:
Got a new bulb RB 279 T with my last purchase.

I got it to work on my local HA lab with the config in this:
image

Here is a picture of what it looks like in my instance of HA
image
image

I've already submitted PR to update the documentation here: 
https://github.com/Koenkk/zigbee2mqtt.io/pull/1648


Koenkk:
Thanks!

PR #4809, opened 2022-10-21
develco: MOSZB-140 re-order configuration

sjorge:
Looks like older firmware 3.x.x does not support develcoLedControl and it timesout.

For now re-order the configuration steps so we handle msIlluminanceMeasurement, msTemperatureMeasurement first and we try to read develcoLedControl last.

### DIFF
--- a/devices/develco.js
+++ b/devices/develco.js
@@ -534,24 +534,25 @@ module.exports = [
             return {default: 35};
         },
         configure: async (device, coordinatorEndpoint, logger) => {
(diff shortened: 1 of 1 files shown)


sjorge:
I'm going to see if I can read 'develcoPrimarySwVersion', 'develcoPrimaryHwVersion' during device interview, so the information is then available once configure() is called. 

That should allow us to skip the reading of develcoLedControl on older firmwares.

Koenkk:
Thanks!

PR #4811, opened 2022-10-21
Added initial support for Vimar IoT Thermostat firmware

danieledwardgeorgehitchcock:
This firmware (according to Vimar) is currently in Beta. All Vimar IoT devices support either Bluetooth or Zigbee firmwares which can be changed via the Vimar View application.

The firmware currently doesn't report manual changes made on the device, or a running state so usability is limited at present - lets hope for more functionality later down the line..


Koenkk:
Thanks!

PR #4697, opened 2022-09-23
Add support for Leviton ZS057

kevin-david:
a.k.a. Wall Switch, 0-10V Dimmer, 120-277V, Lumina™ RF

Doc PR: https://github.com/Koenkk/zigbee2mqtt.io/pull/1602

PR #4813, opened 2022-10-21
Remove `ballast_physical_*_level`

kevin-david:
This is not settable, and the range is 0-255 not 1-254

```
Read result of 'lightingBallastCfg': {"physicalMinLevel":0,"physicalMaxLevel":255}
```

and
```
First occurred: October 16, 2022 at 11:22:54 PM (2670 occurrences)
Last logged: 2:44:12 PM

    Invalid value for number.0x25c823ffffa60700_ballast_physical_minimum_level: 0 (range 1.0 - 254.0)
    Invalid value for number.0x25c823ffffa60700_ballast_physical_maximum_level: 255 (range 1.0 - 254.0)
```

and

```
Publish 'set' 'write' to '0x25c823ffffa60700' failed: 'Error: Write 0x25c823ffffa60700/1 lightingBallastCfg({"physicalMinLevel":6}, {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'READ_ONLY')'
```

### DIFF
--- a/devices/leviton.js
+++ b/devices/leviton.js
@@ -100,11 +100,7 @@ module.exports = [
             exposes.numeric('ballast_maximum_level', ea.ALL).withValueMin(1).withValueMax(254)
                 .withDescription('Specifies the maximum brightness value'),
             exposes.numeric('ballast_power_on_level', ea.ALL).withValueMin(1).withValueMax(254)
-                .withDescription('Specifies the initialisation light level. Can not be set lower than "ballast_minimum_level"'),
-            exposes.numeric('ballast_physical_minimum_level', ea.STATE_GET).withValueMin(0).withValueMax(255)
-                .withDescription('Specifies the minimum light output the ballast can achieve.'),
-            exposes.numeric('ballast_physical_maximum_level', ea.STATE_GET).withValueMin(0).withValueMax(255)
-                .withDescription('Specifies the maximum light output the ballast can achieve.')],
+                .withDescription('Specifies the initialisation light level. Can not be set lower than "ballast_minimum_level"')],
         configure: async (device, coordinatorEndpoint, logger) => {
             await extend.light_onoff_brightness().configure(device, coordinatorEndpoint, logger);
             const endpoint = device.getEndpoint(1);
--- a/devices/leviton.js
+++ b/devices/leviton.js
@@ -101,9 +101,9 @@ module.exports = [
                 .withDescription('Specifies the maximum brightness value'),
             exposes.numeric('ballast_power_on_level', ea.ALL).withValueMin(1).withValueMax(254)
                 .withDescription('Specifies the initialisation light level. Can not be set lower than "ballast_minimum_level"'),
-            exposes.numeric('ballast_physical_minimum_level', ea.ALL).withValueMin(1).withValueMax(254)
+            exposes.numeric('ballast_physical_minimum_level', ea.STATE_GET).withValueMin(0).withValueMax(255)
                 .withDescription('Specifies the minimum light output the ballast can achieve.'),
-            exposes.numeric('ballast_physical_maximum_level', ea.ALL).withValueMin(1).withValueMax(254)
+            exposes.numeric('ballast_physical_maximum_level', ea.STATE_GET).withValueMin(0).withValueMax(255)
                 .withDescription('Specifies the maximum light output the ballast can achieve.')],
         configure: async (device, coordinatorEndpoint, logger) => {
             await extend.light_onoff_brightness().configure(device, coordinatorEndpoint, logger);
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -1588,12 +1588,6 @@ const converters = {
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
             const result = {};
-            if (msg.data.hasOwnProperty('physicalMinLevel')) {
-                result['ballast_physical_minimum_level'] = msg.data.physicalMinLevel;
-            }
-            if (msg.data.hasOwnProperty('physicalMaxLevel')) {
-                result['ballast_physical_maximum_level'] = msg.data.physicalMaxLevel;
-            }
             if (msg.data.hasOwnProperty('ballastStatus')) {
                 const ballastStatus = msg.data.ballastStatus;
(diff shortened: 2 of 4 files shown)


devices/leviton.js
```diff
@@ -101,9 +101,9 @@ module.exports = [
                 .withDescription('Specifies the maximum brightness value'),
             exposes.numeric('ballast_power_on_level', ea.ALL).withValueMin(1).withValueMax(254)
                 .withDescription('Specifies the initialisation light level. Can not be set lower than "ballast_minimum_level"'),
-            exposes.numeric('ballast_physical_minimum_level', ea.ALL).withValueMin(1).withValueMax(254)
+            exposes.numeric('ballast_physical_minimum_level', ea.STATE_SET).withValueMin(0).withValueMax(255)
```
kevin-david, line 104: @Koenkk what's the right way to make this actually read only? 

is it removing https://github.com/Koenkk/zigbee-herdsman-converters/blob/2103a26beb34e2dff1f11afc29e16d37b377db8f/converters/toZigbee.js#L581? 

I think this attribute _might_ be broadly read only, it definitely is for this model, but I haven't been able to dig up a lot of documents...

kevin-david: Hm, that didn't seem to work (https://github.com/Koenkk/zigbee-herdsman-converters/pull/4813/commits/a5561a910fc382c8834565396108ad34dabb5be5)

Koenkk: Read only = `ea.STATE_GET` 

- So the value can be found in the published state
- And you can do a `/get`

kevin-david: Whoops, I did `STATE_SET` by accident on the first one. Updated now.

Either way, how do I get the CI to pass now - is there something else I have to change? https://github.com/Koenkk/zigbee-herdsman-converters/actions/runs/3303623754/jobs/5451785989

Thanks!

Koenkk:
@sjorge do I remember correctly that you have a Ubisys D1? If yes, do you know if the `ballast_physical_minimum_level` is settable?

sjorge:
> @sjorge do I remember correctly that you have a Ubisys D1? If yes, do you know if the `ballast_physical_minimum_level` is settable?

The D1 is the only one I don’t have. I got a bunch of S1, S2, J1’s and one C4.

sjorge:
@Koenkk that being said, I had a look at the ZCL as this is just a standard cluster.

image

It looks like those are indeed read only, they are similar to the colorTemperature min/max stuff. It just describes what the devices actually capabilities are. I don't think it is very useful to have those readable as an attribute in the mqtt payload.

I think it would make sense to treat them like the colorTemperature min/max and use there values in the exposes information to correctly set the range.

Koenkk:
> It looks like those are indeed read only, they are similar to the colorTemperature min/max stuff. It just describes what the devices actually capabilities are. I don't think it is very useful to have those readable as an attribute in the mqtt payload.

I agree with this. The people using this attribute is probably very limited and they can always read it via the dev console.

@kevin-david could you remove the `ballast_physical_*_level` completely (so also the exposes) for all devices.

kevin-david:
@Koenkk 
> > It looks like those are indeed read only, they are similar to the colorTemperature min/max stuff. It just describes what the devices actually capabilities are. I don't think it is very useful to have those readable as an attribute in the mqtt payload.
> 
> I agree with this. The people using this attribute is probably very limited and they can always read it via the dev console.
> 
> @kevin-david could you remove the `ballast_physical_*_level` completely (so also the exposes) for all devices.

Done - I agree as well, the property isn't useful generally.

devices/leviton.js
kevin-david: Obsolete with direction https://github.com/Koenkk/zigbee-herdsman-converters/pull/4813#issuecomment-1288036529

kevin-david:
@Koenkk snuck one more commit in here, I missed `fz.on_off` when adding this device originally back in #4697 

I think this will fix the error I have locally: `No converter available for 'ZS057-D0Z' with cluster 'genOnOff' and type 'attributeReport' and data '{"onOff":0}'`

BTW, is there any easy way to test PRs via home assistant  other than setting up a custom converter temporarily? I saw https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html#home-assistant-addon which only applies to the dev branch and I assume there's some kind of publishing step...

Koenkk:
@kevin-david to test locally, get root access to your HA container, enter the z2m container by doing `docker exec -it Z2M_CONTAINER_ID /bin/bash` , then the zigbee-herdsman-converters files can be found under `node_modules/zigbee-herdsman-converters`

kevin-david:
@Koenkk 
> @kevin-david to test locally, get root access to your HA container, enter the z2m container by doing `docker exec -it Z2M_CONTAINER_ID /bin/bash` , then the zigbee-herdsman-converters files can be found under `node_modules/zigbee-herdsman-converters`

The changes seem to get stomped on when I restart the addon from HA, and restarting via the web frontend doesn't seem to have any effect. Doing a `kill -HUP` on the `node index.js` process also seems to kill and restart the container, which drops the changes as well.

I'm editing the path `/app/node_modules/zigbee-herdsman-converters` with `vi` while the container is running - does that sound right?

Thanks again!

Koenkk:
You should not restart via HA itself but via `docker restart CONTAINER_ID`. Restarting via HA will delete the container.

kevin-david:
> You should not restart via HA itself but via `docker restart CONTAINER_ID`. Restarting via HA will delete the container.

Thanks, that helped. The other thing I had to do was turn off the watchdog functionality on the addon so it wasn't automatically restarted. I think this is worth documenting, any suggestions on where?

Also, I broke out the Leviton specific changes into #4834 since they were unrelated

Koenkk:
@kevin-david thanks, I believe the best place is here: https://github.com/zigbee2mqtt/hassio-zigbee2mqtt/blob/master/README.md

kevin-david:
@Koenkk 

> @kevin-david thanks, I believe the best place is here: https://github.com/zigbee2mqtt/hassio-zigbee2mqtt/blob/master/README.md

Here's a PR: https://github.com/zigbee2mqtt/hassio-zigbee2mqtt/pull/406

PR #4816, opened 2022-10-22
feat(devices/phillips): Add updated Hue LightStrip controller

asmundg:
New Hue LightStrip reports a new ZigBee model, but retains same functionality from what I can see. SKU (8718699703424) is still the same.


Koenkk:
Thanks!

PR #4817, opened 2022-10-22
Add support TS011F device clone

saspol:
Model: Tongou TO-Q-SY1-JZT
https://www.aliexpress.com/item/1005004777816579.html


Koenkk:
Thanks!

PR #4818, opened 2022-10-22
feat(namron): Add move to hue command for 4512706 remote

asmundg:
This remote has a color wheel, which we want to pick up and convert.

### DIFF
--- a/devices/namron.js
+++ b/devices/namron.js
@@ -196,9 +196,9 @@ module.exports = [
         fromZigbee: [fz.command_on, fz.command_off, fz.command_step, fz.command_step_color_temperature, fz.command_recall,
             fz.command_move_to_color_temp, fz.battery, fz.command_move_to_hue],
         exposes: [e.battery(), e.action([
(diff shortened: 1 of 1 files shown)


devices/namron.js
```diff
@@ -194,7 +194,7 @@ module.exports = [
         vendor: 'Namron',
         description: 'Remote control',
         fromZigbee: [fz.command_on, fz.command_off, fz.command_step, fz.command_step_color_temperature, fz.command_recall,
-            fz.command_move_to_color_temp, fz.battery],
+            fz.command_move_to_color_temp, fz.battery, fz.command_move_to_hue],
         exposes: [e.battery(), e.action(['on', 'off', 'brightness_step_up', 'brightness_step_down', 'color_temperature_step_up',
```
Koenkk, line 198: Please also add the produced action here

asmundg: Done (I think). I'm not sure if there are steps I can take to validate this, but the exposed actions match what's being logged when using the color wheel.

Koenkk:
Thanks!

PR #4823, opened 2022-10-23
develco: always read fw info from correct endpoint.

sjorge:
Before this change `device.endpoints[0]` would be correct after the deviceInterview.  It would however be incorrect once a fromDatabase call was used, the order of `device.endpoints` is different after.

Reading from the wrong endpoint results in a timeout, luckily we can identify the right endpoint by checking which one has the modelId data populated! We now consistently read the same endpoint, we can now also 'refresh' the firmware info after a deviceAnnounce. If we ever get OTA support it will be good to keep this up to date also.

For reference these are the ep for my develco devices.
    MOSZB-140: 34
    AQSZB-110: 38
    WISZB-120: 35
    HESZB-120: 35

Seems before it only sort of worked by pure luck 😨

### DIFF
--- a/devices/develco.js
+++ b/devices/develco.js
@@ -31,24 +31,28 @@ const fwOnEvent = async (type, data, device, options, state) => {
      * Develco devices have a manufacturer specific field on genBasic
      *   for sw and hw versions.
      *
-     * We read those during deviceInterview
+     * We read those during deviceInterview and deviceAnnounce
      *   so that this information is usable during configure() calls
      *   to skip some features on older devices that might cause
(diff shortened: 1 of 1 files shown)


sjorge:
@Koenkk would it make more sense to just read this during configure() ?

I could just make a share function to call in the device that currently have the onEvent ?

sjorge:
Hold off with merging this for now, I've tweaked the external converter and asked @X4LD1M0 to test if this also works better for him. If it doesn't there is another one with the current in this PR change applied.

PR #4824, opened 2022-10-23
Fix color temp ranges

tomasbedrich:
Tested on my home bulbs according to the guide: https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_3-adding-converter-s-for-your-device


Koenkk:
Thanks!

PR #4826, opened 2022-10-23
Add a new fingerprint for zemismart TS0601 device

renatofilho:
Got this device from Benexmart store on Aliexpress that came with a different manufacturerName.


Koenkk:
Thanks!

PR #4828, opened 2022-10-24
develco: impove firmware reading

sjorge:
This PR moves away from using an OnEvent to read the Sw/Hw versions from the Develco specific attributes and instead does it during configuration.

Before we would not always pick the same endpoint and this was causing problems, we now grab the first endpoint we find with genBasic. Before we somtimes read from an endpoint without genBasic cluster!

I've also added this to the configure of AQSZB-110, it also correctly returned the additional info.

I think it is probably safe to just do this for all Develco devices but I did not add this, I only added it to devices I have my hands on to test or to the ones that were already reading it before my fun adventure with the MOSZB-140.

### DIFF
--- a/devices/develco.js
+++ b/devices/develco.js
@@ -25,35 +25,34 @@ const develcoLedControlMap = {
     0xFF: 'both',
 };
 
-
-const fwOnEvent = async (type, data, device, options, state) => {
-    /**
-     * Develco devices have a manufacturer specific field on genBasic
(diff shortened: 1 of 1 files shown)


sjorge:
This should finally and properly fix zigbee2mqtt#14277 (I know, famous last words, but me and @X4LD1M0 iterated over this with a external converter so 🤞 )

Koenkk:
great, thanks!

PR #4833, opened 2022-10-25
WBMSW3: Removed reading values when configuring

wb-adegtyarev:
WBMSW3:
- Removed reading values when configuring, because sometimes there were errors. The sensor itself will send the value in a few seconds.
- Added clarification about CO2 sensor model.


Koenkk:
Thanks!

PR #4834, opened 2022-10-26
Improve Leviton ZS057 reporting behavior

kevin-david:
I broke this out from #4813 because it really is a different problem entirely...

I figured out the problem I mentioned here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/4813#issuecomment-1288254692 - see the added comment

I'm still having a big problem with timeouts on this device despite what I _think_ is acceptable LQI (50+), getting errors like this:

```
error 2022-10-25 23:38:15Publish 'set' 'state' to '0x25c823ffffa60700' failed: 'Error: Command 0x25c823ffffa60700/1 genLevelCtrl.moveToLevelWithOnOff({"level":0,"transtime":0}, {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":false,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Timeout - 29522 - 1 - 48 - 8 - 11 after 10000ms)'
```

It seems _partially_ working because the brightness change takes effect, but it gives that error / times out.

Strangely, using `moveToLevel` instead doesn't seem to work well (makes the lights flash)... still digging in, any advice apprechiated!


devices/leviton.js
```diff
@@ -91,10 +91,12 @@ module.exports = [
         model: 'ZS057-D0Z',
         vendor: 'Leviton',
         description: 'Wall switch, 0-10V dimmer, 120-277V, Lumina™ RF',
-        extend: extend.light_onoff_brightness({disableEffect: true, noConfigure: true}),
-        fromZigbee: [fz.brightness, fz.identify, fz.lighting_ballast_configuration, fz.level_config],
-        toZigbee: [tz.light_onoff_brightness, tz.ballast_config, tz.level_config],
-        exposes: [e.light_brightness(),
+        // Do not use extend.light_onoff_brightness because that has fz.ignore_basic_report,
+        // which is what tells us when the physical switch is used
+        fromZigbee: [fz.on_off, fz.on_off_via_brightness, fz.brightness, fz.level_config, fz.identify,
```
Koenkk, line 96: Above this line, can you try adding `meta: {disableDefaultResponse: true},` and see if that solves the connection issues?

kevin-david: This helped, however it exposed another problem that I _think_ was caused by `tz.level_config` so I removed it. It's working kind of OK now: https://github.com/Koenkk/zigbee-herdsman-converters/pull/4834/commits/c5cf4b0f0f00f42d10aebf1d13062932db355aee

Very open to suggestions on the reporting thing that feels like a hack...

devices/leviton.js
```diff
@@ -102,10 +105,12 @@ module.exports = [
             exposes.numeric('ballast_power_on_level', ea.ALL).withValueMin(1).withValueMax(254)
```
kevin-david, line 105: I'm not sure why, but with the current set of settings `ballast_power_on_level` doesn't seem to be sticking well either

kevin-david: It definitely changes something for a time, but it's almost like it doesn't want to save on more than a transient basis sometimes (i.e. the state object comes back eventually and it's 0)

kevin-david: Yeah, guess so. The behavior seems to be you can set it to anything, but that gets reset to 0 when the light is turned off and set back to 32 when the light is turned on

I swear it was working at one point before I started messing around with it though, ugh...

Koenkk: I see, then maybe also remove it from the exposes

kevin-david: Done: https://github.com/Koenkk/zigbee-herdsman-converters/pull/4834/commits/3a13fa02049c999b4559e579fd6a0fc61ad0fc30

Left a comment in case I come back to really debug it :)

Koenkk:
is this ready for merge?

kevin-david:
@Koenkk

> is this ready for merge?

I think so, given I'm the only one likely using this device, the risk seems low :) 

Any ideas on that reporting-every-5-seconds "fix" I added? That's really the only thing I don't like about this change, but it does help the device work better.

devices/leviton.js
```diff
@@ -91,21 +91,26 @@ module.exports = [
         model: 'ZS057-D0Z',
         vendor: 'Leviton',
         description: 'Wall switch, 0-10V dimmer, 120-277V, Lumina™ RF',
-        extend: extend.light_onoff_brightness({disableEffect: true, noConfigure: true}),
-        fromZigbee: [fz.brightness, fz.identify, fz.lighting_ballast_configuration, fz.level_config],
-        toZigbee: [tz.light_onoff_brightness, tz.ballast_config, tz.level_config],
+        // Do not use extend.light_onoff_brightness because that has fz.ignore_basic_report,
+        // which is what tells us when the physical switch is used
```
Koenkk, line 95: > which is what tells us when the physical switch is used

I would expect that a new converter is added which publishes actions when something is published to the basic cluster.

kevin-david: Good point. The `extend:` doesn't seem to be working as I expect. If I keep it, then explicitly add `fz.on_off` things work fine, which is what I did this this seems to match everything else - but feels like I am missing something. Nevertheless, this works locally like this!

kevin-david: When I say mostly - really wish I could get some more advanced settings like power on level & min/max on the device itself. I suspect some of these are hidden behind these manufacturer-specific clusters (32776, the next one which is obviously wrong)?

image

Koenkk: >  I suspect some of these are hidden behind these manufacturer-specific clusters (32776,

If they are, you need to figure out how it works by sniffing the traffic with the original gateway: https://www.zigbee2mqtt.io/advanced/zigbee/04_sniff_zigbee_traffic.html#with-cc2531

kevin-david: Nice, I have a CC2652 so I might have to pick up a CC2531 and give it a shot. In the meantime, I think this improvement is ready to merge and if I get around to it, I'll do some more sniffing :)

Koenkk:
Thanks!

PR #4838, opened 2022-10-27
TV02-Zigbee expose fix

vladyspavlov:
I forgot to replace the `battery` expose with `battery_low` the last time. The device supports only `battery_low` which already exists in the converter, so this fixes it.


Koenkk:
Thanks!

PR #4844, opened 2022-10-28
Cloud Even Thermostatic Radiator Valve

kovvik:
add support for Cloud Even TRV


Koenkk:
Can this be merged with one of the already supported TuYa thermostats? (e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/a197538c53680aaa79b0b976213b67a620f56d1c/devices/tuya.js#L2019)

kovvik:
The closest match would be this: https://github.com/Koenkk/zigbee-herdsman-converters/blob/a197538c53680aaa79b0b976213b67a620f56d1c/devices/tuya.js#L2195

Even the actual hardware looks like the same as the HT-10 but some data points are different (localTemp, heatingSetpoint, errorStatus) and some of them are not implemented in etop_thermostat (ex: temp calibration):
https://github.com/Koenkk/zigbee-herdsman-converters/blob/141d1b77e4709f53a7e49dc684a594da783d715a/converters/fromZigbee.js#L4497

I would like to implement all of the features the TRV has. There are some datapoints that I still have to reverse engineer.

Koenkk:
Then lets keep it separate, can you refactor the code to use the new format? This reduces code size and makes it better maintainable. You can use this TuYa thermostat as an example:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/17eef48d6282ad50c89600e620109dd787993b1b/devices/tuya.js#L2098

kovvik:
I've done the refactoring. Is the scheduling setup ok for you?

kovvik:
Yes, I checked that. That assumes that there is a dedicated dp that I can send a schedule for any day (dp 100 for the TR02). But this TRV has no such dp, all schedules should be sent to its own data point (101-107).
The thermostatScheduleDay cannot be used either because this TRV uses different format.

Koenkk:
I've merged this into `tuya.js` and shared some code with the TV02.

- Can you do a final check? 
- Especially check the description of the `frost_protection` expose.

lib/tuya.js
```diff
@@ -1244,6 +1249,7 @@ const valueConverterBasic = {
 const valueConverter = {
     trueFalse: valueConverterBasic.lookup({1: true, 0: false}),
     onOff: valueConverterBasic.lookup({'ON': true, 'OFF': false}),
+    systemMode: valueConverterBasic.lookup({'heat': false, 'off': true}),
```
kovvik, line 1252: The values are the opposite here, sorry it was my mistake in the original file
```suggestion
    systemMode: valueConverterBasic.lookup({'heat': true, 'off': false}),
```

lib/tuya.js
```diff
@@ -1165,6 +1165,11 @@ const tuyaExposes = {
         .withDescription('Result of the self-test'),
     faultAlarm: () => exposes.binary('fault_alarm', ea.STATE, true, false).withDescription('Indicates whether a fault was detected'),
     silence: () => exposes.binary('silence', ea.STATE_SET, true, false).withDescription('Silence the alarm'),
+    frostProtection: () => exposes.binary('frost_protection', ea.STATE_SET, 'ON', 'OFF').withDescription('When Anti-Freezing function'+
+        ' is activated, the temperature in the house is kept at 8 °C, the device display "AF". Press the pair button to cancel.'),
```
kovvik, line 1169: The CloudEven TRV has no frost protection indication on its display. It can be checked and set in the service menu of the device.
```suggestion
        ' is activated, the temperature in the house is kept at 8 °C.'),
```

Koenkk:
Fixed the issues, thanks!

PR #4846, opened 2022-10-29
Add action list to Inovelli VZM31SN

nathanfiscus:
This exposes the button presses that are possible with the Inovelli VZM35-SN


Koenkk:
Thanks!

PR #4847, opened 2022-10-29
Sinope switch expose status LED control

cloutierjo:
Expose some more feature from the Sinope switch and dimmer

This is my first work on that code base so don't hesitate to make suggestions if I missed anything.

Also have two questions that I'm not sure if it could be done better.
1. In a device definition, it seems that the extend get overwritten when we define a section (like expose or to/from zigbee) and so I needed to add back what is defined inside the "parent device". Is that the right way?
2. I know the front end is out of scope of this repo, I was trying to get the color widget for both status led, unfortunately the FE directly use the exposed name to show the correct widget and so, both properties would be shown as 'color_xy' which is clearly misleading. Did I miss something to help the FE and also 3rd party automation software to be able to display a color selector based on what currently exist ( I can ask the question in the zigbee2mqtt repo if more appropriate)

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -4405,6 +4405,30 @@ const converters = {
             }
         },
     },
+    sinope_led_color_on: {
+        // DM2500ZB and SW2500ZB
+        key: ['led_color_on'],
+        convertSet: async (entity, key, value, meta) => {
+            const r = (value.r >= 0 && value.r <= 255) ? value.r : 0;
+            const g = (value.g >= 0 && value.g <= 255) ? value.g : 0;
+            const b = (value.b >= 0 && value.b <= 255) ? value.b : 0;
+
+            const valueHex = r + g * 256 + (b * 256 ** 2);
+            await entity.write('manuSpecificSinope', {ledColorOn: valueHex});
+        },
+    },
+    sinope_led_color_off: {
+        // DM2500ZB and SW2500ZB
+        key: ['led_color_off'],
+        convertSet: async (entity, key, value, meta) => {
+            const r = (value.r >= 0 && value.r <= 255) ? value.r : 0;
+            const g = (value.g >= 0 && value.g <= 255) ? value.g : 0;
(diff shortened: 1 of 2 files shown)


Koenkk:
> In a device definition, it seems that the extend get overwritten when we define a section (like expose or to/from zigbee) and so I needed to add back what is defined inside the "parent device". Is that the right way?

If the definition has toZigbee, fromZigbee and exposes; extend becomes obsolete, I removed it.

> I know the front end is out of scope of this repo, I was trying to get the color widget for both status led, unfortunately the FE directly use the exposed name to show the correct widget and so, both properties would be shown as 'color_xy' which is clearly misleading. Did I miss something to help the FE and also 3rd party automation software to be able to display a color selector based on what currently exist ( I can ask the question in the zigbee2mqtt repo if more appropriate)

Did you try using the description to clarify what the color_xy is about? (off vs on)

cloutierjo:
Thanks for the merge, about the color I did try, but with this code:
```
            exposes.composite('color_xy', 'led_color_on')
                .withFeature(exposes.numeric('r', ea.ALL))
                .withFeature(exposes.numeric('g', ea.ALL))
                .withFeature(exposes.numeric('b', ea.ALL))
                .withDescription('Control status LED intensity when load ON'),
            exposes.composite('color_xy', 'led_color_off')
                .withFeature(exposes.numeric('r', ea.ALL))
                .withFeature(exposes.numeric('g', ea.ALL))
                .withFeature(exposes.numeric('b', ea.ALL))
                .withDescription('Control status LED color when load OFF'),
```
I get this ui that is not optimal, but could work. I'm just worried in other 3rd party ui it's going to be even worst (
Screenshot_20221029_093922

But I did stop investigating that direction since the property sent back is always `color`:
`No converter available for 'color' ({"x":0.4105263157894737,"y":0.48947368421052634})` So afaik there is no way to distinguish both of them at converter level. And changing that behaviour would likely break ~24 other device that I cannot test.

Koenkk:
I would suggest leaving it as is for now, filling in the r/g/b values is not that bad.

PR #4848, opened 2022-10-29
Added battery status reporting to Osram AC01353010G

danieledwardgeorgehitchcock:
Resolves Koenkk/zigbee2mqtt#14682 - Awaiting feedback from OP to confirm it works

### DIFF
--- a/devices/osram.js
+++ b/devices/osram.js
@@ -302,15 +302,17 @@ module.exports = [
         model: 'AC01353010G',
         vendor: 'OSRAM',
         description: 'SMART+ Motion Sensor',
-        fromZigbee: [fz.temperature, fz.ias_occupancy_only_alarm_2, fz.ignore_basic_report],
+        fromZigbee: [fz.temperature, fz.ias_occupancy_only_alarm_2, fz.ignore_basic_report, fz.battery],
         toZigbee: [],
+        meta: {battery: {voltageToPercentage: '3V_2100'}},
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['msTemperatureMeasurement', 'genPowerCfg']);
             await reporting.temperature(endpoint);
             await reporting.batteryVoltage(endpoint);
+            await reporting.batteryAlarmState(endpoint);
         },
-        exposes: [e.temperature(), e.occupancy()],
+        exposes: [e.temperature(), e.occupancy(), e.battery(), e.battery_voltage(), e.battery_low()],
     },
     {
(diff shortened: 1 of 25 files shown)


devices/osram.js
```diff
@@ -4,6 +4,27 @@ const ota = require('../lib/ota');
 const reporting = require('../lib/reporting');
 const extend = require('../lib/extend');
 const e = exposes.presets;
+const utils = require('../lib/utils');
+
+const fzLocal = {
```
danieledwardgeorgehitchcock, line 9: @Koenkk - I am not particularly happy with this.

To do this 'properly' in my head, I need to add another item in the `batteryVoltageToPercentage` section of `utils.js` and I need add all of the `batteryAlarmState`'s as per 3.3.2.2.4.11 in the Zigbee Cluster Library specification.

For now, it seemed simpler to just drop it into it's own local converter...

Koenkk: > I need to add another item in the batteryVoltageToPercentage section of utils.js

Why not do this? 

>  need to refactor and add all of the batteryAlarmState's as per 3.3.2.2.4.11 in the Zigbee Cluster Library specification.

For the battery low, we take `1<<3` into account here:

 https://github.com/Koenkk/zigbee-herdsman-converters/blob/c35dfceeb1833f77fd9aa3f9146395ae96f25ef9/converters/fromZigbee.js#L344

danieledwardgeorgehitchcock: Happy to add the new batteryVoltageToPercentage value.

As for the batteryAlarmState, I don't think we do..? Correct me if I'm wrong, but arent we just taking the bitshifts of the first threshold values for battery 1, 2 and 3, but not the second, third or fourth as in the ZCL spec - this device uses the fourth four some unknown reason...

Koenkk: We can take all here (so add a `battery4Low` to the expression)

danieledwardgeorgehitchcock: Done and ready for full review :-)

lib/utils.js
```diff
@@ -179,6 +179,8 @@ function batteryVoltageToPercentage(voltage, option) {
     } else if (option === 'Add_1V_42V_CSM300z2v2') {
         voltage = voltage + 1000;
         percentage = toPercentage(voltage, 2900, 4100);
+    } else if (option === 'CR2_1900_3000') {
```
danieledwardgeorgehitchcock, line 182: Could potentially be replaced by Koenkk/zigbee-herdsman-converters#4863

danieledwardgeorgehitchcock:
Depends on Koenkk/zigbee-herdsman-converters#4863

Koenkk:
Thanks again!

PR #4849, opened 2022-10-29
Added support for Hue Garnea LTD007

danieledwardgeorgehitchcock:
Resolves Koenkk/zigbee2mqtt#14701 - Unsure of color temperature range, but based on units with same spec


Koenkk:
Awesome!

PR #4850, opened 2022-10-29
Update müller_licht.js

copystring:
Fix to make it work with https://github.com/o0shojo0o/ioBroker.zigbee2mqtt


Koenkk:
Removing the legacy converters creates a breaking change. You can disable the legacy behaviour by passing `legacy: false` to the converter options.

copystring:
Okay. With legacy: false the problem is that the exposed is wrong. For example one exposed is color_wheel which changes to color_move. https://github.com/o0shojo0o/ioBroker.zigbee2mqtt doesn't know about this change when connecting via websocket. Where or how would it get the correct exposed data?

copystring:
Right. I forgot exposed don't matter with legacy so I changed the pull request. Is this ok now? The exposed data was wrong but is working now.

Koenkk:
Perfect, thanks!

PR #4851, opened 2022-10-29
Add support for US Variant of Ikea LED2004G8

ThoolooExpress:
The US variant is only 800 lumens, but otherwise appears identical.


Koenkk:
Thanks!

PR #4861, opened 2022-11-01
Add system mode support for TZE200_dv8abrrz thermostat

muddy79:
So far, TZE200_dv8abrrz (haozee_thermostat)  has no support for setting the HVAC / system mode from HA, since the device's mode has been mapped to the preset entity, not to the system mode entity. 
This causes inability to use the standard HA Thermostat card:
- z2m does not report available modes, so the entity displays buttons for all possible modes (off, auto, heat, COOL, DRY, FAN ONLY)
- pressing any of these buttons does not cause any effect, neither calling climate.set_hvac_mode service
The only way to operate this entity is to select preset mode.

This small change maps three system modes to device modes:
- 'off' mode mapped to device's 'off' mode
- 'heat' mode mapped to device's 'on' mode (which causes 100% opening of the valve and full heating)
- 'auto' mode mapped to device's 'manual' mode (which allows to select the  desired temperature and heating is started/stopped automatically)
I've decided not to map 'auto' system mode to 'auto' preset, since it enables the built-in schedule, which is NOT supported (yet?) to be setup from z2m. I guess, this preset mode should be called 'schedule' instead of 'auto', but I've not changed the name due to backward compatibility.

As a result, system mode and preset are a bit overlapping, but I guess it's necessary, since there is no way to distinguish between auto mode with manual temperature set point and built-in scheduler.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4098,9 +4098,10 @@ const converters = {
             const dp = dpValue.dp; // First we get the data point ID
             const value = tuya.getDataValue(dpValue);
             const presetLookup = {0: 'auto', 1: 'manual', 2: 'off', 3: 'on'};
+            const systemModeLookup = {0: 'auto', 1: 'auto', 2: 'off', 3: 'heat'};
             switch (dp) {
             case tuya.dataPoints.haozeeSystemMode:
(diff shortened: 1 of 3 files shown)


Koenkk:
Thanks!

PR #4863, opened 2022-11-01
Added generic batteryVoltageToPercentage converter

danieledwardgeorgehitchcock:
It would be useful to have a 'generic' `voltageToPercentage` converter in utils to be able to handle device oddities without cluttering up the list with edge cases

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -142,6 +142,7 @@ function getKey(object, value, fallback, convertTo) {
 
 function batteryVoltageToPercentage(voltage, option) {
     let percentage = null;
+    const optionSplit = option.split("_",);
     if (option === '3V_2100') {
         if (voltage < 2100) {
             percentage = 0;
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -179,6 +180,21 @@ function batteryVoltageToPercentage(voltage, option) {
     } else if (option === 'Add_1V_42V_CSM300z2v2') {
         voltage = voltage + 1000;
         percentage = toPercentage(voltage, 2900, 4100);
+    // Generic converter that expects an option value to be given in the format of:
+    // GENERIC_%MinimumThresholdVoltage%_%MaximumThresholdVoltage%
+    // I.E. GENERIC_1900_3000
+    } else if (optionSplit[0] === 'GENERIC') {
```
Koenkk, line 186: I think it is better to make this an object.

So:

- `meta: {battery: {voltageToPercentage: {min: 1900, max: 3000}}},`
- `else if (typeof option === 'object') {`
- `percentage = toPercentage(voltage, option.min, option.max);`

Koenkk:
After this https://github.com/Koenkk/zigbee-herdsman-converters/pull/4848 can be updated.

lib/utils.js
danieledwardgeorgehitchcock: Was thinking of using that approach so will swap it out.

Thanks!

danieledwardgeorgehitchcock: Done!

lib/utils.js
```diff
@@ -179,6 +179,18 @@ function batteryVoltageToPercentage(voltage, option) {
     } else if (option === 'Add_1V_42V_CSM300z2v2') {
         voltage = voltage + 1000;
         percentage = toPercentage(voltage, 2900, 4100);
+    // Generic converter that expects an option object with min and max values
+    // I.E. meta: {battery: {voltageToPercentage: {min: 1900, max: 3000}}}
+    } else if (typeof option === 'object') {
+        if (option.min && option.max) {
```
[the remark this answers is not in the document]
danieledwardgeorgehitchcock: I've removed the error catches. I need to get my head around how the test framework works so, will submit that as a separate PR if that is ok?

Koenkk:
Thanks!

PR #4865, opened 2022-11-01
Support new lonsonho light bulb

piotrdzz:
Hello, 
Recently I have bought additional light bulb from the same seller that sold me a Lonsonho light bulb.  The new bulb was not recognised while the seller claimed it is the same model as previously sold to me.  I figured out that new bulbs may have different fingerprint. Thus I am adding it here, I confirm that it works. 

zigbee2mqtt conf panel showing entity:
image


Koenkk:
Thanks!

PR #4873, opened 2022-11-02
Refactored batteryAlarmState to align with ZCL spec

danieledwardgeorgehitchcock:
Split this into it's own PR to make working / reverting easier down the line

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -338,9 +338,24 @@ const converters = {
             }
 
             if (msg.data.hasOwnProperty('batteryAlarmState')) {
-                const battery1Low = (msg.data.batteryAlarmState & 1<<0) > 0;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

evilpie:
In the specification that I found `BatteryVoltageMinThreshold` for battery 2 is a bit 10. So why is the code using bit 9 for battery2?

danieledwardgeorgehitchcock:
> In the specification that I found `BatteryVoltageMinThreshold` for battery 2 is at bit 10. So why is the code using bit 9 for battery2?

I was thinking that too but as I don't know bitwise op's that well, I piggybacked off of what was already there - maybe it _is_ wrong - happy for it to be corrected..

This is what I used as reference. I am also thinking of a way I could bring through the alarm state(s) as a string too

ZCL batteryAlarmState

PR #4875, opened 2022-11-02
Also read sw/hw version for Develco HMSZB-110

sjorge:
Now that I have access to one, I verified you can also read the sw/hw versions. I also aligned the batteryVoltage reporting with what develco recommended.


sjorge:
image

Koenkk:
Thanks!

PR #4879, opened 2022-11-03
Update namron.js

SKJOAN:
The values for Hysterersis is wrong. Default should be 0.5. don't know about range but definitely less the 5-50. Maybe 0.1 to 5

image

### DIFF
--- a/devices/namron.js
+++ b/devices/namron.js
@@ -610,7 +610,7 @@ module.exports = [
             exposes.numeric('hysterersis', ea.ALL)
                 .withUnit('°C')
                 .withValueMin(5).withValueMax(50).withValueStep(0.1)
-                .withDescription('Hysterersis setting, range is 5-50, unit is 0.1oC,  Default: 5.'),
+                .withDescription('Hysteresis setting, between 0.5 and 2 in 0.1 °C. Default: 0.5.'),
             exposes.numeric('display_brightnesss', ea.ALL)
                 .withValueMin(1).withValueMax(7).withValueStep(1)
                 .withDescription('Adjust brightness of display values 1(Low)-7(High)'),


owangen:
@SKJOAN I would guess you would need to multiply hystereresis with 10 when sending zigbee to the device. Like it's done with the touch thermostat.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/0064e81aa66b8eb702571a76a2422ec0d8ed6c74/converters/toZigbee.js#L3178

Likewise when converting from zigbee I would guess you need to divide by 10.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/0064e81aa66b8eb702571a76a2422ec0d8ed6c74/converters/fromZigbee.js#L1769

SKJOAN:
@owangen hi, im not aware of this transformation as I'm also not used to working in repository. But I see in the old version of the code that the temperature regulation was off by 5 deg. As you see in the picture attached. I have not been able to test after the change. image

owangen:
> @owangen hi, im not aware of this transformation as I'm also not used to working in repository. But I see in the old version of the code that the temperature regulation was off by 5 deg. As you see in the picture attached. I have not been able to test after the change. image

Ok. Could you test with Dev version of zigbee2mqtt? If hysteris setting still is wrong, just comment here and I can fix it.

SKJOAN:
@owangen hello again. There seems to be somthing wrong here stil. When i try to set the hysteresis to 0.5 i get this error. I guess it is related to the needed multiplier?

Error 2022-11-08 09:21:02Publish 'set' 'hysterersis' to 'Kontor Panelovn' failed: 'Error: Write 0x3fff000400000254/1 hvacThermostat({"4106":{"value":0.5,"type":32}}, {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":4644,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'INVALID_VALUE')'

owangen:
@SKJOAN Yes, pretty sure this is fixed in my latest PR. So please test with Dev build after that PR is merged.

PR #4881, opened 2022-11-03
Update bitron.js

lxeka:
By this modification, the bitron /32 thermostat is able to control the cooling and heating features. With the actual version, it is enabled just in heating mode.

### DIFF
--- a/devices/bitron.js
+++ b/devices/bitron.js
@@ -189,7 +189,7 @@ module.exports = [
         model: 'AV2010/32',
         vendor: 'SMaBiT (Bitron Video)',
         description: 'Wireless wall thermostat with relay',
-        fromZigbee: [fz.legacy.bitron_thermostat_att_report, fz.battery, fz.hvac_user_interface],
+        fromZigbee: [fz.legacy.thermostat_att_report, fz.battery, fz.hvac_user_interface],
         toZigbee: [tz.thermostat_control_sequence_of_operation, tz.thermostat_occupied_heating_setpoint,
             tz.thermostat_occupied_cooling_setpoint, tz.thermostat_local_temperature_calibration, tz.thermostat_local_temperature,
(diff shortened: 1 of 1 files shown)


lxeka:
No. because there are not enabled all the  thermostat cluster attributes. The legacy part just changes the data structure when the device responds. I have modified also for myself locally and it works fine

Koenkk:
I see, thanks!

Koenkk:
@lxeka can you check this?

bmorfin:
@lxeka : Is the issue clear enough ? Do you need further information to fix it ?

lxeka:
> @lxeka : Is the issue clear enough ? Do you need further information to fix it ?

Can you send me the datapacket that you send to the thermostat. Which mode are you using to send the data?
Because the lates modifications provide you send to thermostat customised datapackets through mqtt.
For example to set on cooling mode you need firstly set the ""control_sequence_of_operation" so it can operate on both heat and cool, and then you can use system_mode to modify the modality of the thermostat.

bmorfin:
That's correct, but the issue I have is with the Zigbee2MQTT integration in home assistant. The system mode attribute value 'cool' is not in the list of possible values (based on the bitron.js file I guess) :

image

You get an 'Invalid Value' error when you click on 'auto' :

image

And you get an error messages when you send an MQTT command from Home Assistant to set this attribute value to 'cool' : 

image

Hoping that this clarifies my request to update the bitron.js file

bmorfin:
Hello,

According to the documentation of the manufacturer, the values for the system mode attribute are :

image

But in the file, the values are different  :

image


Would it possible to change the file content, so that it corresponds to the documentation for this attribute (Value 'cool' instaed of 'auto') ?

bmorfin:
@lxeka : Is the issue clear enough ? Do you need further information to fix it ?

bmorfin:
Hello, the manufacturer documentation gives the 3 possible values for system mode

image

I would suggest to change the bitron.js file accordingly.

PR #4887, opened 2022-11-03
Revert change to max temperature range for Moes BHT-002, should be 0-35 as originally

bwindsor:
Addresses https://github.com/Koenkk/zigbee-herdsman-converters/pull/4744#issuecomment-1301627283 and changes the range back to `0-35` as it was originally.

I had previously misunderstood how `max_temperature_limit` works.


Koenkk:
Thanks!

PR #4889, opened 2022-11-04
Add support for Dresden Elektronik FLS-A lp

l-mb:
The device identifies as "FLS-A2" over Zigbee, but FLS-A lp is the documented name by the vendor.

https://github.com/Koenkk/zigbee2mqtt/issues/14225

### DIFF
--- a/devices/dresden_elektronik.js
+++ b/devices/dresden_elektronik.js
@@ -1,6 +1,7 @@
 const exposes = require('../lib/exposes');
 const ota = require('../lib/ota');
 const extend = require('../lib/extend');
+const reporting = require('../lib/reporting');
 const e = exposes.presets;
 
 module.exports = [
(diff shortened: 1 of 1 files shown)


devices/dresden_elektronik.js
```diff
@@ -31,4 +32,31 @@ module.exports = [
...
+        extend: extend.light_onoff_brightness({noConfigure: true}),
+        exposes: [e.light_brightness().withEndpoint('l1'), e.light_brightness().withEndpoint('l2'),
+            e.light_brightness().withEndpoint('l3'), e.light_brightness().withEndpoint('l4')],
+        endpoint: (device) => {
+            return {'l1': 11, 'l2': 12, 'l3': 13, 'l4': 14};
+        },
+        meta: {multiEndpoint: true, disableDefaultResponse: true},
+        configure: async (device, coordinatorEndpoint, logger) => {
+            await reporting.bind(device.getEndpoint(11), coordinatorEndpoint, ['genLevelCtrl', 'genOnOff']);
+            await reporting.bind(device.getEndpoint(12), coordinatorEndpoint, ['genLevelCtrl', 'genOnOff']);
```
Koenkk, line 49: Does this device have any physical buttons on it? If not, then we don't add reporting (to limit the amount of messages on the network)

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1678

devices/dresden_elektronik.js
l-mb: It does not. Sorry, I copied that from an example without fully realizing what it would do. Fixed.

(Which actually addressed that the device wouldn't always show the current state properly on start. Thanks!)

Koenkk:
Thanks!

PR #4892, opened 2022-11-04
Update onesti.js

TheModin:
Add support for NimlyPRO connect module: https://nimly.io/sv/accessoarer/nimly-module
As discussed here: https://github.com/Koenkk/zigbee2mqtt/issues/14726

### DIFF
--- a/devices/onesti.js
+++ b/devices/onesti.js
@@ -8,7 +8,7 @@ const constants = require('../lib/constants');
 
 module.exports = [
     {
-        zigbeeModel: ['easyCodeTouch_v1', 'EasyCodeTouch', 'EasyFingerTouch'],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

bergetun:
Any chance of adding support for Nimly Indoor lock as well ?  All locks from Nimly uses the same module and have the same functionality, so it should be the same.   It seems like z2m doesnt recongince it since it identify as NimlyIn when the module is attached to Nimly Inndoor lock instead of a Nimly Pro (See screenshot)

This is how it looks like if i try to add nimbly connection module in my Nimly Indoor lock through z2m 

Skjermbilde 2023-08-03 kl  15 14 41

PR #4894, opened 2022-11-05
Create akuvox.js

Burki24:
This Sensor is a combination of a TS0201 Tuya-Sensor with Heiman datapoints. The heiman-datapoints also has changed (battery). If it will be connected to Zigbee2MQTT, it will recognized as TS0201 temperature & humidity Sensor with display. But it has no display and zigbee2mqtt will not receive any data from that devices.  That's the reason for me to build a new converter, hopefully, that's the right way to integrate this device into zigbee2mqtt.

### DIFF
--- a/devices/akuvox.js
+++ b/devices/akuvox.js
@@ -1,12 +1,11 @@
 const exposes = require('../lib/exposes');
-const fz = {...require('../converters/fromZigbee'), legacy: require('../lib/legacy').fromZigbee};
+const fz = require('../converters/fromZigbee');
 const reporting = require('../lib/reporting');
 const e = exposes.presets;
 
 module.exports = [
     {
         fingerprint: [{modelID: 'TS0201', manufacturerName: '_TYZB01_ujfk3xd9'}],
-        zigbeeModel: ['M423-9E'],
         model: 'M423-9E',
         vendor: 'Akuvox',
         description: 'Smart temperature & humidity Sensor',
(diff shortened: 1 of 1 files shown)


Koenkk:
Does this device also work with the following configure (make sure to factory reset it)

```js
            await endpoint.read('genBasic', ['manufacturerName', 'zclVersion', 'appVersion', 'modelId', 'powerSource', 0xfffe]);
        },
```

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1680

Burki24:
> ```js
> configure: async (device, coordinatorEndpoint, logger) => {
>             const endpoint = device.getEndpoint(1);
>             await endpoint.read('genBasic', ['manufacturerName', 'zclVersion', 'appVersion', 'modelId', 'powerSource', 0xfffe]);
>         },
> ```

We will test this. I tell you tomorrow.

Burki24:
@Koenkk thanks, it's running well, I changed the code.

Koenkk:
Great, I propose to merge it with https://github.com/Koenkk/zigbee-herdsman-converters/blob/8e98b34f70c3c401dd15ac6ff2581229d68d5dd2/devices/tuya.js#L1589 now since the code is exactly the same.

Burki24:
> Great, I propose to merge it with
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/8e98b34f70c3c401dd15ac6ff2581229d68d5dd2/devices/tuya.js#L1589
> 
> now since the code is exactly the same.

@Koenkk 

It's not the same code, and not the same device. As I wrote, the akuvox-device has no display and also, the following inserts in TS0201 doesn't run: fzLocal.TS0201_battery and fzLocal.TS0201_humidity (zigbee-herdsman-converters/devices/tuya.js: line 1593). For the akuvox-device, we need fz.battery and fz.humidity. So please don't merge it to TS0201.

thanks
Burkhard

Koenkk:
I see, thanks!

PR #4907, opened 2022-11-06
Update tuya.js

clausmuus:
Add power monitoring polling to _TZ3000_qeuvnohg din smart relay.

Unfortunately I don't know the function of these lines:
```
            try {
                await endpoint.read('genBasic', ['manufacturerName', 'zclVersion', 'appVersion', 'modelId', 'powerSource', 0xfffe]);
            } catch (e) {
                // Sometimes can fail: https://github.com/Koenkk/zigbee2mqtt/issues/12760#issuecomment-1165435220
            }
            //await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement', 'seMetering']);
            //await reporting.rmsVoltage(endpoint, {change: 5});
            //await reporting.rmsCurrent(endpoint, {change: 50});
```
From there, I don't know if those are needed either. Anyway, with the patch the device works for me.


clausmuus:
Alternatively, the TS011F_plug_3 model would also work. However, the assigned image is not correct.

Koenkk:
> Alternatively, the TS011F_plug_3 model would also work. However, the assigned image is not correct.

I prefert it to move it there, this creates a lot of code duplication.

clausmuus:
>

clausmuus:
That's also OK for me, since that also works.

PR #4909, opened 2022-11-06
essentials smart home thermostat converter

TAKeanice:
I have built this converter for the thermostat I own. Also, it is relevant for https://github.com/Koenkk/zigbee2mqtt/issues/14239 . Please tell me if there are any aspects of how I built this that may have to be changed to make it into the main repository.

### DIFF
--- a/devices/essentials.js
+++ b/devices/essentials.js
@@ -138,7 +138,7 @@ const essentialsValueConverter = {
     },
 };
 
-const device = {
+const essentialsThermostat = {
     fingerprint: tuya.fingerprint('TS0601', ['_TZE200_i48qyn9s']),
     model: 'Essentials Zigbee Radiator Thermostat',
     vendor: 'TuYa',
@@ -235,5 +235,5 @@ const device = {
     },
 };
 
-module.exports = device;
+module.exports = [essentialsThermostat];
 
--- a/devices/essentials.js
+++ b/devices/essentials.js
@@ -217,7 +217,7 @@ const device = {
             [104, 'local_temperature_calibration', tuya.valueConverter.localTempCalibration1],
             [105, 'schedule_override_setpoint', tuya.valueConverter.divideBy10],
             [106, null, null], // TODO rapid heating
-            [107, 'window_open', tuya.valueConverter.lookup({'YES': true, 'NO': false})],
+            [107, 'window_open', tuya.valueConverterBasic.lookup({'YES': true, 'NO': false})],
             [108, null, null], // TODO hibernate
             [109, 'monday', essentialsValueConverter.day_schedule('monday', 1)],
             [110, 'tuesday', essentialsValueConverter.day_schedule('tuesday', 2)],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks for the PR. Could you refactor this to the new TuYa format? You can find an example here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/f7ca8a893874ee6d68032bf8efd60b31a532338f/devices/tuya.js#L2189

TAKeanice:
Thanks for the feedback, I will try to do it this weekend.

TAKeanice:
I am still committed to this, but need some more time.

TAKeanice:
@Koenkk I think the error in the git checks is not in my changes

TAKeanice:
> @Koenkk I think the error in the github checks is not in my changes

I was wrong, I didn´t know that it is now necessary to return an array of devices from a manufacturer.js file

Koenkk:
Can we harmonise the schedule expose with one of the existing supported TuYa thermostats?

- https://github.com/Koenkk/zigbee-herdsman-converters/blob/4b8c153235b0fef0481efec41d07ad90ce317f61/devices/tuya.js#L2253

- https://github.com/Koenkk/zigbee-herdsman-converters/blob/4b8c153235b0fef0481efec41d07ad90ce317f61/devices/tuya.js#L2159

TAKeanice:
Not sure whether that will provide a benefit, especially if the user is required to type a string in a specific format instead of a composite decomposing the parts of the schedule setpoints. I will take a look whether I can do anything useful with the standard functions. Is there anything I should look at for the effects of a composes definition besides the UI built into Zigbee2Mqtt?

Koenkk:
@TAKeanice the main goal is to unify the TuYa thermostat schedules. It would be great if existing converters could be re-used (to reduce code size and improve maintainability). This is indeed mainly for the UI.

TAKeanice:
Even though the tuya schedules get more harmonized by the string format "HH:MM/Degrees", it's pretty unusable for all home automation tools. It is too easy to mistype and enter invalid values.

I will happily introduce a new function to lib/tuya.js for a schedule using a composite of days with a nested composite of hour, minute and temperature for each setpoint, as for the essentials thermostat. If you wish, I can make that function generic such that it can be used in the tuyaDatapoints definition.

One thing I noted about the tuyaDatapoints definition is that the converter entry in a datapoint does not have a definition for `get`, so refreshing a value manually is not possible as with the former `convertGet`. Is that on purpose or just not finished yet?

Koenkk:
> I will happily introduce a new function to lib/tuya.js for a schedule using a composite of days with a nested composite of hour, minute and temperature for each setpoint, as for the essentials thermostat. If you wish, I can make that function generic such that it can be used in the tuyaDatapoints definition.

That would be great

> One thing I noted about the tuyaDatapoints definition is that the converter entry in a datapoint does not have a definition for get, so refreshing a value manually is not possible as with the former convertGet. Is that on purpose or just not finished yet?

AFAIK the TuYa api does not allow reading.

TAKeanice:
I am going to make the schedule function I used more general and outbid into the main tuya lib for others to use. This should remain open.

TAKeanice:
I just saw #5550 and will see how it affects the plan to introduce a better week schedule for tuya thermostats, as I am still committed to implementing it.

PR #4910, opened 2022-11-06
Inovelli VZM31SN - fw 2.05 Support

nathanfiscus:
Adds new LED effect animations (added in 2.05)
Adds parameter 252 (added in 2.05) to disable clearing notifications from the switch
Fixes the value labels for parameter 251.
Fixes a state issue for enum display types
Binds the onOff and levelControl cluster during configuration


Koenkk:
Thanks!

PR #4916, opened 2022-11-08
Add coverInverted to Yookee D10110

markfox85:
I'm having the exact same issue as described in https://github.com/Koenkk/zigbee2mqtt/issues/13813 with the Yookee D10110.


Koenkk:
Thanks!

PR #4917, opened 2022-11-08
re-compute local temperature when calibration of BRT-100-TRV has changed

xworld21:
My best attempt at fixing https://github.com/Koenkk/zigbee2mqtt/issues/14179

In short: the BRT-100-TRV does *not* report the local temperature when one changes the calibration value, so the converter has to do the computation locally.

There is a danger: if the Z2M state gets out of sync with the valve for any reason, then changing the calibration may return wrong values. I can't predict exactly how, because I am not familiar with how the values are updated on e.g. reconnection, and in which order.

However, since with the current converter the value is wrong *every* time one changes the calibration, this patch should still be an improvement.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4086,7 +4086,12 @@ const converters = {
             case tuya.dataPoints.moesSvalvePosition:
                 return {position: value};
             case tuya.dataPoints.moesScompensationTempSet:
-                return {local_temperature_calibration: value};
+                return {
+                    local_temperature_calibration: value,
+                    ...(meta && meta.state && meta.state.local_temperature != null && meta.state.local_temperature_calibration != null) ?
+                        {local_temperature: meta.state.local_temperature + (value - meta.state.local_temperature_calibration)} :
+                        {},
+                };
             case tuya.dataPoints.moesSecoMode:
                 return {eco_mode: value ? 'ON' : 'OFF'};
             case tuya.dataPoints.moesSecoModeTempSet:
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -4088,6 +4088,8 @@ const converters = {
             case tuya.dataPoints.moesScompensationTempSet:
                 return {
                     local_temperature_calibration: value,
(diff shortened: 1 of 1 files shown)


converters/fromZigbee.js
```diff
@@ -4086,7 +4086,12 @@ const converters = {
             case tuya.dataPoints.moesSvalvePosition:
                 return {position: value};
             case tuya.dataPoints.moesScompensationTempSet:
-                return {local_temperature_calibration: value};
+                return {
+                    local_temperature_calibration: value,
+                    ...(meta && meta.state && meta.state.local_temperature != null && meta.state.local_temperature_calibration != null) ?
+                        {local_temperature: meta.state.local_temperature + (value - meta.state.local_temperature_calibration)} :
```
Koenkk, line 4092: Instead of `meta.state.local_temperature_calibration`, why not use `value`?

xworld21: If you are referring to the fourth line, it's because I need to subtract the old calibration value before adding the new one (assuming we know the old calibration value, hence the previous null checks). Do you want me to add a comment? I realise the code can be obscure, maybe a justification re the valve not updating the temperature is in order here.

Koenkk: Adding a note would be good. 

So if I understand correctly, currently the `local_temperature` is correct but it takes some time to update after changing the `local_temperature_calibration`?

xworld21: > So if I understand correctly, currently the `local_temperature` is correct but it takes some time to update after changing the `local_temperature_calibration`?

Yes, the valve applies the new calibration when it comes to opening or closing the valve, and it shows the re-calibrated value on its display instantly. It just doesn't report the new value over Zigbee until the (uncalibrated) measurement itself change.

v6ak:
Kudos for this fix! I had supposed that the temperature field is raw temperature, because a change in calibration doesn't change it. Based on this assumption, I've made some automations for careful recalibration. So, maybe the BRT-100 is not as dodgy as I thought…

PR #4918, opened 2022-11-08
Fix hysteresis setting for Namron panel heater

owangen:
Fix hysterersis setting, handling decimal to and from device
Relates to https://github.com/Koenkk/zigbee-herdsman-converters/pull/4879#issuecomment-1306845533


Koenkk:
Thanks!

PR #4924, opened 2022-11-10
Add GL-G-101P

Pferdebockwurst:
This adds the GL-G-101P, which is an updated version of the GL-G-001P and thus also a Zigbee 12W Garden Lamp RGB+CCT (pro).


Pferdebockwurst:
Sorry for the typo in the commit description, the device is in fact an updated version of the GL-G-001P, which is already supported and where the config was taken from.

The lamp works with this config. The color temperature range can't be read, when sending a read command the lamp doesn't answer. However, for the GL-G-001P there is also no range given.

Pferdebockwurst:
The manufacturer states (https://www.gledopto.com/h-col-374.html that the lamp supports a color temperature range from 2200K to 6500K. This page also contains some nice images which might be usable for the zigbee2mqtt.io device repository.

Koenkk:
Please add the color temperature range as described on https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_3-adding-converter-s-for-your-device

Pferdebockwurst:
> Please add the color temperature range as described on https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_3-adding-converter-s-for-your-device

Please see my comments above. The device doesn't respond to these commands.

Koenkk:
Missed that, thanks!

Pferdebockwurst:
> Missed that, thanks!

No problem, thanks for your work.

PR #4936, opened 2022-11-12
 Add _TZE200_bvu2wnxz to SEA801-Zigbee/SEA802-Zigbee

olexs:
Adding another manufacturerName for this device. Mine came whitelabeled as "ME167", but it's the exact same as others listed.


olexs:
After some testing, does not actually work with the SEA801/802 defaults, despite apparently being the exact same device. Will re-open once I've gotten it to fully work.

PR #4939, opened 2022-11-12
Update villeroy_boch.js

Rapid4:
Hi,

I got a new version of the sink light controller. Can you please add the new model ID? I tested the new ID on my HA instance and everything is working as it should.

Zigbee model number: 5991713

image
index


Koenkk:
Thanks!

PR #4942, opened 2022-11-13
Innr SP 234 : Configure cummulative energy attribute

jschornick:
Assuming I have the right attribute, this should add energy reporting to the Innr SP 234 during configuration. Reported in Koenkk/zigbee2mqtt#14824.


Koenkk:
Thanks!

PR #4946, opened 2022-11-13
Do not read onLevel again if already in state

pyos:
The device might have forgotten it for some reason. If we assume that the user is not modifying it with direct attribute writes, the read should be redundant anyway.

If the device responds with an error to a request for on_level, mark that in globalStore and always use 'previous' from then on.

zigbee2mqtt#13570

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -884,17 +884,28 @@ const converters = {
                 // TODO: same problem as above.
                 // TODO: if transition is not specified, should use device default (OnTransitionTime), not 0.
                 if (transition.specified || globalStore.getValue(entity, 'turnedOffWithTransition') === true) {
-                    const current = utils.getObjectProperty(meta.state, 'brightness', 254);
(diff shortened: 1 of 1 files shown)


converters/toZigbee.js
```diff
@@ -884,17 +884,27 @@ const converters = {
...
+                            }
+                        } catch (e) {
+                            // OnLevel not supported
                         }
-                    } catch (e) {
-                        // OnLevel not supported
+                    }
+                    if (onLevel === 0) {
+                        onLevel = 'previous';
+                        globalStore.putValue(entity, 'on_level_unsupported', true);
```
Koenkk, line 901: Instead of storing it in the globalStore (which is cleared on every z2m restart) I think it is better to save it in `entity.meta` (an object which is saved in `data/database.db`).

So this could would become:

```js
entity.meta.onLevelSupported = true
entity.save()
```

pyos: done

Koenkk:
Many thanks!

PR #4951, opened 2022-11-14
Added support for "auto" modes of Legrand (412171)

celestianx:
Added new cluster definition to fully support device: Legrand (412171) - DIN contactor module ( Bticino FC80CC), referenced here [14812](https://github.com/Koenkk/zigbee2mqtt/issues/14812)

### DIFF
--- a/devices/legrand.js
+++ b/devices/legrand.js
@@ -27,7 +27,9 @@ module.exports = [
         toZigbee: [tz.legrand_deviceMode, tz.on_off, tz.legrand_identify, tz.electrical_measurement_power],
         exposes: [exposes.switch().withState('state', true, 'On/off (works only if device is in "switch" mode)'),
             e.power().withAccess(ea.STATE_GET), exposes.enum('device_mode', ea.ALL, ['switch', 'auto'])
-                .withDescription('switch: allow on/off, auto will use wired action via C1/C2 on contactor for example with HC/HP')],
+                .withDescription('switch: allow manual on/off, auto uses contact\'s C1/C2 wired actions for Peak/Off-Peak electricity rates'),
+                exposes.enum('auto_mode', ea.ALL, ['off', 'auto', 'on_override'])
+                .withDescription('Off/auto/on (override) (works only if device is set to "auto" mode)')],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genIdentify', 'genOnOff', 'haElectricalMeasurement']);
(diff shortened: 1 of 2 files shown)


devices/legrand.js
```diff
@@ -15,6 +16,23 @@ const readInitialBatteryState = async (type, data, device) => {
...
+    auto_mode: {
+        key: ['auto_mode'],
+        convertSet: async (entity, key, value, meta) => {
+            const mode = {'off': 0x00, 'auto': 0x02, 'on_override': 0x03};
+            const payload = {data: Buffer.from([mode[value]])};
+            await entity.command('manuSpecificLegrandDevices3', 'command0', payload);
+            return {state: {'auto_mode': value}};
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('manuSpecificLegrandDevices3', [0x0000], manufacturerOptions);
```
Koenkk, line 31: In order for this to work there also needs to be an `fromZigbee` converter which parses the read response of this.

devices/legrand.js
```diff
@@ -24,10 +42,14 @@ module.exports = [
...
+        toZigbee: [tz.legrand_deviceMode, tz.on_off, tz.legrand_identify, tz.electrical_measurement_power, tzLocal.auto_mode],
         exposes: [exposes.switch().withState('state', true, 'On/off (works only if device is in "switch" mode)'),
-            e.power().withAccess(ea.STATE_GET), exposes.enum('device_mode', ea.ALL, ['switch', 'auto'])
-                .withDescription('switch: allow on/off, auto will use wired action via C1/C2 on contactor for example with HC/HP')],
+            e.power().withAccess(ea.STATE_GET),
+            exposes.enum('device_mode', ea.ALL, ['switch', 'auto'])
+                .withDescription('Switch: allow manual on/off, auto uses contact\'s C1/C2 wired actions for Peak/Off-Peak ' +
+                    'electricity rates'),
+            exposes.enum('auto_mode', ea.ALL, ['off', 'auto', 'on_override'])
+                .withDescription('Off/auto/on (override) (works only if device is set to "auto" mode)')],
```
Koenkk, line 52: In the description please explain what it does, the values it self are already defined in the list above. So what does `auto_mode` do?

devices/legrand.js
celestianx: Thanks. But this is only required if we can read the auto_mode state from the device? Not so sure if it can be done or not sure how to identify the attributes code.
Effectively, as of  now, your recommendations would be to remove it entirely?

Koenkk: yes, remove the convertGet, the new state is already returned by the convertSet

celestianx: Ok, thanks. Why did you remove the function from toZigbee.js and put it instead in Legrand.js?

devices/legrand.js
Koenkk: `auto_mode` is fine, thanks!

PR #4954, opened 2022-11-14
Update tuya.js

clausmuus:
Add power monitoring polling to _TZ3000_qeuvnohg din smart relay.

The device needs polling. Without polling there are only one update in 10 to 30 minutes.
With this change, the device works for me.


Koenkk:
Thanks!

PR #4955, opened 2022-11-14
typo fix for ocupancy was incomplete

sjorge:
#4944 is incomplete and breaks a few TRV, including the Ubisys H1.


Koenkk:
Thanks!

PR #4956, opened 2022-11-14
[WIP] Add HC-020-ZB

xworld21:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/13094

Still WIP, for two reasons:
1. the fingerprint overlaps with a previous device, but I am struggling to go sufficiently far back in the git history to figure out what's the story there
2. the `local_temperature_calibration` works normally in Celsius, but partially fails in Fahrenheit (= the device does not report the updated value over Zigbee, just like the BRT-100-TRV)

I don't know how to fix 1. I guess we need to find another value that discriminates between the old and the new device, hence find somebody with a BHT-002-GCZLB with the correct fingerprint, at least.

For 2, I'll fix it soon, it's trivial but I have run out of time this week!

Cc @KGS501 could you please test this with your hardware?

Edit: this is probably relevant for the fingerprint: `Read result of 'genBasic': {"zclVersion":3,"appVersion":68,"stackVersion":0,"hwVersion":1,"manufacturerName":"_TZE200_ztvwu4nk","modelId":"TS0601","dateCode":"","powerSource":1}`

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -288,6 +288,22 @@ const tzLocal = {
             }
         },
     },
+    // for HC-020-ZB, we must use different data points depending on thermostat_unit
+    hc020_temperatures: {
+        key: ['current_heating_setpoint', 'max_temperature', 'min_temperature'],
+        convertSet: async (entity, key, value, meta) => {
+            // dps[key] = [celsius data point, fahrenheit data point]
+            const dps = {
+                'current_heating_setpoint': [16, 17],
+                'max_temperature': [19, 18],
+                'min_temperature': [26, 20],
+            };
+            const fahrenheit = meta.state.thermostat_unit === 'fahrenheit' ? 1 : 0;
+            const temp = fahrenheit ?
+                utils.mapNumberRange(value, 0, 100, 32, 212) : value;
+            await tuya.sendDataPointValue(entity, dps[key][fahrenheit], temp);
+        },
+    },
 };
 
 const fzLocal = {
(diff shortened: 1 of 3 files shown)


devices/tuya.js
```diff
@@ -3595,4 +3611,126 @@ module.exports = [
...
+        model: 'HC-020-ZB',
+        vendor: 'TuYa',
+        description: 'Thermostat',
+        fromZigbee: [tuya.fzDataPoints],
+        toZigbee: [
+            tzLocal.hc020_temperatures,
+            tuya.tzDataPoints,
+        ],
+        onEvent: tuya.onEventSetLocalTime,
+        configure: async (device, coordinatorEndpoint, logger) => {
```
xworld21, line 3625: By the way, I copied this from the instructions to add new TuYa devices. I don't know if it's actually needed here.

Koenkk: Probably not needed, please replace with `configure: tuya.configureMagicPacket,`

xworld21: As far as I can tell, there is no need for `configure` at all, I removed it.

devices/tuya.js
```diff
@@ -288,6 +288,36 @@ const tzLocal = {
...
+                // when in fahrenheit mode, the thermostat does not report the re-calibrated value
+                // furthermore, the calibration value is applied in fahrenheit
+                // in this case, guess local_temperature based on the previous known calibration
+                // note that this will accumulate rounding errors if the calibration is changed too many times
+                const capped = utils.numberWithinRange(value, -9, 9);
+                await tuya.sendDataPointValue(entity, 27, capped);
+                const state = {local_temperature_calibration: value};
+                if (fahrenheit && meta.state.local_temperature != null && meta.state.local_temperature_calibration != null) {
+                    state['local_temperature'] = utils.precisionRound(meta.state.local_temperature +
+                        (capped - meta.state.local_temperature_calibration) * 5 / 9, 1);
```
xworld21, line 306: Unfortunate compromise here: the calibration unit changes with `thermostat_unit`! So if I want to keep `local_temperature` updated, I have to do an unfortunate rounding.

Another option is to toggle the `thermostat_unit`, which forces the thermostat to report the new temperature. Then we don't report strange fractional values. However, the thermostat will end up beeping *three* times...

devices/tuya.js
```diff
@@ -3595,4 +3625,122 @@ module.exports = [
                 .withDescription('Turns the onboard LED on or off'),
         ],
     },
+    {
+        fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_ztvwu4nk'}],
```
xworld21, line 3629: Just to have a separate thread for discussion: this fingerprint is also used for at least another device with *different* data points (one of the Moes thermostats). The PR cannot be merged until this is sorted out, and I am afraid I just don't know where to go here.

At worst, I'll put this up somewhere as external converter and hope for a proper solution in the future.

Koenkk: That would be strange (I've not seen this before), what datapoints are different and do we know the `_TZE200_ztvwu4nk` works correctly with the current converter?

xworld21: I did a quick check, among the overlapping data points, 20 has a different meaning (dead zone temp for BHT-002, fahrenheit min temp for HC-020), the others more or less agree except maybe `moesHold`. Quite a few data points just don't overlap (e.g. the schedule is different, plus the whole `thermostat_unit` thing does not exist in the BHT-002 converter). I will test the current converter again, but to my memory, HC-020 only half works, hence this PR.

I didn't keep notes, but my cursory search through old issues gave the impression that there are genuine BHT-002 devices with fingerprint `_TZE200_ztvwu4nk` which work with the current converter.

xworld21: PS: the X5H thermostat has more similar data points to HC-020, compared to the BHT-002.

Koenkk: Could it be that this device was somehow mislabeled? (wrong manufacturer name flashed) Have you seen any others with the same thermostat as you have?

xworld21: > Could it be that this device was somehow mislabeled? (wrong manufacturer name flashed) Have you seen any others with the same thermostat as you have?

I only know of @KGS501 who seems to have the exact same device, down to the colour.

Somehow the Smart Life app can tell the difference. It is reported as "Zigbee Temperature Controller Public Version 2". MCU is V2.0.0. Maybe TuYa can recognise the MAC address on their end, which would not be that great for Z2M.

Koenkk: @inext-gr  can you provide your data/database.db entry of your BHT-002-GCLZB ?

PR #4957, opened 2022-11-14
Add support for AwoX / Eglo 12242 CCW Filament bulb

mstratford:


mstratford:
https://www.eglo.com/en/bulb-12242.html

Koenkk:
Thanks!

PR #4960, opened 2022-11-15
Add support for legrand dry contactor (412173)

CP-a-u-l-o:
Hi,

please note that I'm a beginner at converters, and it is my first attempt adding new device.

This is a new DIN module from legrand (released in September 22 in my country).
For all intents and purposes, it works quite the same way as the contactor model 412171 except that it turns on/off a dry circuit. It keeps the capability of measuring instantaneous power with an additional sensor meant to be set on the load circuit. This part is quite similar as the DIN power consumption module 412015.

Therefore to create the corresponding converter, I copied the 412171 one and added the activePower reading from the 412015.

It's been running great since I added this configuration in my setup.

### DIFF
--- a/devices/legrand.js
+++ b/devices/legrand.js
@@ -17,13 +17,13 @@ const readInitialBatteryState = async (type, data, device) => {
 
 module.exports = [
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

PR #4973, opened 2022-11-18
Add Phillips Hue Go with Bluetooth

synsi23b:
japanese model number


synsi23b:
I think I added it to the wrong Model. It worked no problem using Zigbee, but I don't know. Reopened new request on the correct Version of the Hue Go

synsi23b:
Actually it was the correct one. I should have done propper testing. Both worked, but the other one didnt provide color controls.

Koenkk:
Thanks!

PR #4979, opened 2022-11-20
Adding unoccupied_heating_setpoint setting

Lepuk:
By default in auto mode, the heater drops gradually to the lowest available temperature when unoccupied (7 °C by default, which is freezing). This setting to help controlling the unoccupied temperature target


Koenkk:
Thanks!

PR #4982, opened 2022-11-20
Add support for Philips LWA028 bulb

joashc:
This is the 2022 model of the LWA018.

Product link is here: https://www.philips-hue.com/en-au/p/hue-white-ambiance-a60---b22-smart-bulb---1100/8719514393332

The model number is the same as the LWA018.


Koenkk:
Thanks!

PR #4985, opened 2022-11-21
Make SinopeThermostats part of code more readable

MattL0:
And easier to work with


MattL0 (commented):
little modification

MattL0 (commented):
wrongly added

PR #4995, opened 2022-11-22
Adjust behaviour of hive receiver commands

Prouser123:
Hi, at the moment, when selecting `system_mode: heat` for example, only that attribute is sent. This is not enough for the hive receiver to turn on the heating. This PR adjusts the behaviour of `system_mode` calls to mimic the data sent by the hive remote control

When selecting `system_mode: heat` this code sends a default temperature of 20*C. This mimics the actions I have observed being sent by my remote. On my remote (SLT6 / mini), when selecting "manual" mode for heating, it always sends an initial setpoint of 20 degrees.

`system_mode: emergency_heat` Appears to work, although it appears what when this is set, the hive ignores any changes from zigbee2mqtt to `tempSetPointHoldDuration`.

I would love to have your feedback on this. Thank you!

\- James

### DIFF
--- a/lib/exposes.js
+++ b/lib/exposes.js
@@ -394,7 +394,7 @@ class Climate extends Base {
         return this;
     }
 
-    withHiveSystemMode(modes, access = a.ALL, description = "Mode of this device") {
+    withHiveSystemMode(modes, access = a.ALL, description = 'Mode of this device') {
         assert(!this.endpoint, 'Cannot add feature after adding endpoint');
         const allowed = ['off', 'heat', 'auto', 'dry', 'fan_only', 'sleep', 'emergency_heating'];
         modes.forEach((m) => assert(allowed.includes(m)));
--- a/lib/exposes.js
+++ b/lib/exposes.js
@@ -394,6 +394,14 @@ class Climate extends Base {
         return this;
     }
 
+    withHiveSystemMode(modes, access = a.ALL, description = "Mode of this device") {
+        assert(!this.endpoint, 'Cannot add feature after adding endpoint');
+        const allowed = ['off', 'heat', 'auto', 'dry', 'fan_only', 'sleep', 'emergency_heating'];
+        modes.forEach((m) => assert(allowed.includes(m)));
+        this.features.push(new Enum('hive_system_mode', access, modes).withDescription(description));
+        return this;
+    }
+
     withRunningState(modes, access=a.STATE_GET) {
         assert(!this.endpoint, 'Cannot add feature after adding endpoint');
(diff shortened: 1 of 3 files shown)


Koenkk:
Is this a breaking change?

Prouser123:
I don't believe so. I think the only functional difference is that setting `system_mode: heat` now turns on the heating without the need for additional commands to be sent.

Something of note is that this adds the `emergency_heating` system mode to the SLR1c receiver/thermostat. In this PR's current state the mode works but the time cannot be adjusted. However setting this mode does not appear to be previously possible(?) so if that is the case I do not believe that would be breaking either.

Thanks,
\- James.

Prouser123:
I have now removed the case for `emergency_heating`. It will now correctly function over MQTT messages as before.

I believe this PR is no longer breaking as unlike `emergency_heating`, which requires all attributes to be sent in the same command, attributes for the `heat` mode can be modified in subsequent commands. Therefore the setting of `occupiedHeatingSetpoint` when selecting the `heat` system mode does not appear to introduce a breaking change. In my testing with a `SLR1c`, commands altering the setpoint work as before.

Thank you, James

P.S: Would it be worth extending the `hive_thermostat_system_mode` to other receivers? I am not able to test this as I do not have other receivers but I believe they are all compatible with each other.

lib/exposes.js
```diff
@@ -394,6 +394,14 @@ class Climate extends Base {
         return this;
     }
 
+    withHiveSystemMode(modes, access = a.ALL, description = 'Mode of this device') {
+        assert(!this.endpoint, 'Cannot add feature after adding endpoint');
+        const allowed = ['off', 'heat', 'auto', 'dry', 'fan_only', 'sleep', 'emergency_heating'];
+        modes.forEach((m) => assert(allowed.includes(m)));
+        this.features.push(new Enum('hive_system_mode', access, modes).withDescription(description));
```
Koenkk, line 401: Why add a new `system_mode`? This won't even work since the property returned by `tz.hive_thermostat_system_mode` is `system_mode` and not `hive_system_mode`

Prouser123: Hi,  I added a new `system_mode` as certain commands such as `heat` send additional data to the device. As for the property, it is incorrect (let me fix that!) but on my host, zigbee2mqtt still executes the new code despite the incorrect property key. Perhaps that is unintentional behaviour?

Thanks,

James

Koenkk: Probably because there is also a `withSystemMode()` which matches with `hive_thermostat_system_mode`

PR #4997, opened 2022-11-22
add store set value of csm-300

nukusinji:
add store set value of csm-300

csm-300 can setup some function, but it can not read or response.


Koenkk:
What is the reason to add it to the globalStore?

nukusinji:
> What is the reason to add it to the globalStore?

It's my mistake because I found it necessary for state recovery when rebooting.

I removed it.

PR #5000, opened 2022-11-22
Adding support for new Livolo curtain switch (VL-FC2NWZ-2GS21)

daneZz:


Koenkk:
This look more like a fix then new device support?

daneZz:
Fair point. This model is a new version of the Livolo curtain switch. However, functionality seems to work the same apart from this change. So i would lean towards 'Adding support'. Its up to you, im fine with changing the pull request title to 'fix'.

Koenkk:
I see, lets leave it as is, thanks!

Issue #4996, opened 2022-11-22
How To Handle Philips Hue Same zigbeeModel For Different models


danieledwardgeorgehitchcock:
I originally wrote the device converter for the Philips Hue Beyond fixture (zigbeeModel: LLM001).

This was intended for the Pendant / Suspension version of the device (model: 7121131PU):
![Pendant / Suspension](https://www.assets.signify.com/is/image/PhilipsLighting/cbdcacae27d74a7991a4a9b9009dc223?&wid=360&hei=360&$jpglarge$)

However, I have recently acquired the Ceiling version of the same range (model: 7120131PH) which has the exact same fingerprint as the Pendant and for all intents and purposes functions the same:
![Ceiling](https://www.assets.signify.com/is/image/PhilipsLighting/a4ebe3c75fe24a8c81d4a9b9009db07f?&wid=360&hei=360&$jpglarge$)

Furthermore, I have a suspicion that the Table Lamp version of the same range (model: 7120231PU) may also share the same fingerprint too:
![Table Lamp](https://www.assets.signify.com/is/image/PhilipsLighting/ce5674baf4c64b8f9beaa9b9009db96c?&amp;wid=360&amp;hei=360&amp;$jpglarge$)

As far as I am aware, Zigbee2MQTT only allows for string type model identifiers, and not list type; which makes complete sense, but how do we correctly code a converter for a device that shares the same fingerprint, yet completely different models / types?

Koenkk:
In this case we add all the models to the `model`, so it will become: `7121131PU/7120131PH/7120231PU`

danieledwardgeorgehitchcock:
The device image and description won't necessarily match though would it?

Koenkk:
Nope, we simply cannot distinguish these devices. What you can do is make the description more generic and put all 3 devices on the picture.

PR #5008, opened 2022-11-23
Roll together Models for Hue Beyond that share same fingerprint

danieledwardgeorgehitchcock:
Resolves Koenkk/zigbee-herdsman-converters#4996


Koenkk:
Thanks!

PR #5011, opened 2022-11-23
Add possibility to turn on away mode

pjanecekATlangmaster:
Currently there is no way (from GUI) to turn away mode on.


Koenkk:
Thanks!

ebeneezer:
Does not work ! Can not configure workday week programming any more only holiday setting is saved please correct

pjanecekATlangmaster:
This behavior is not related to added away mode. 
I had to push it using payload in developer console before this change. 

> Does not work ! Can not configure workday week programming any more only holiday setting is saved please correct

PR #5027, opened 2022-11-26
Update AC33898 config

astappiev:
When trying to use `power_on_behaviour` I'm getting the following error, so I guess it's not supported.
```
Publish 'set' 'read' to 'bedroom/light' failed: 'Error: Read 0xf0d1b800001c6f67/1 genOnOff(["onOff","startUpOnOff"], {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')'
```

As for name and model, it seems Ledvance uses barcode as model. E.g. this Amazon listing has barcode as the model https://amazon.de/gp/product/B07MJLWC59 and never mention `AC33898` model which is written only on energy label.

And it does support OTA updates according to documentation.

### DIFF
--- a/devices/ledvance.js
+++ b/devices/ledvance.js
@@ -5,10 +5,11 @@ const reporting = require('../lib/reporting');
 module.exports = [
     {
         zigbeeModel: ['A60S TW'],
-        model: 'AC33898',
+        model: '4058075208384',
         vendor: 'LEDVANCE',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #5028, opened 2022-11-26
Update Bosch Thermostat II

mmattel:
This is an update of the Bosch Thermostat II device configuration based on the work of @BBJake, see https://github.com/Koenkk/zigbee2mqtt/issues/14926#issuecomment-1327624341 adding / improving the configuration.

`npm run lint` runs fine, but `npm test` returns one error I do not know why.

Note that I have an issue adding my Bosch Thermostat II to zigbee with the "install code" button - the code gets rejected though it is 95 chars long or sending a mqtt push (Message: Add install code is not supported). Most likely because I am using a Conbee II.

@Koenkk mind to take a look in the failing test?
@niklasarnitz @bbjake mind to re-test

### DIFF
--- a/devices/bosch.js
+++ b/devices/bosch.js
@@ -1,10 +1,176 @@
+const herdsman = require('zigbee-herdsman');
 const exposes = require('../lib/exposes');
 const fz = require('../converters/fromZigbee');
 const tz = require('../converters/toZigbee');
 const reporting = require('../lib/reporting');
+const utils = require('../lib/utils');
 const e = exposes.presets;
+const ea = exposes.access;
 
-module.exports = [
+// Radiator Thermostat II
+const boschManufacturer = {manufacturerCode: 0x1209};
+
+// Radiator Thermostat II
+const operatingModes = {
+    'automatic': 0,
+    'manual': 1,
+    'pause': 5,
+};
+
+// Radiator Thermostat II
+const stateOffOn = {
+    'OFF': 0,
+    'ON': 1,
+};
+
+// Radiator Thermostat II
+const displayOrientation = {
+    'normal': 0,
+    'flipped': 1,
+};
+
+// Radiator Thermostat II
+const tzLocal = {
+    bosch_thermostat: {
+        key: ['window_open', 'boost', 'system_mode'],
+        convertSet: async (entity, key, value, meta) => {
+            if (key === 'window_open') {
+                value = value.toUpperCase();
+                utils.validateValue(value, Object.keys(stateOffOn));
+                const index = stateOffOn[value];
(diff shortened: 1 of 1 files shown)


niklasarnitz:
I decoded the install code from the QR Code, since the one written down besides it didn’t work.

And please keep in mind to use the right payload format for the install code mqtt endpoint!

mmattel:
@niklasarnitz thanks for the info. I scanned the BTH QR code and got a string back with 95 characters which I pasted into zigbee2mqtt. Do I have to convert this string? And in case how?

niklasarnitz:
You don’t have to convert the code but you need to check the compatibility with your zigbee stick and you need the correct json payload: 
Publish this code to zigbee2mqtt/bridge/request/install_code/add with payload {"value":"THE_CODE"}. 

(From the official docs)

mmattel:
Thanks, I already did so as you described. The issue is as written above related to the Conbee II stick. I already ordered a new Sonoff Z3.0 running the CC2652P chip known supported by zigbee2mqtt.

Could you do a test with the changes made in the meanwhile?

BBJake:
@mmattel 
Thanks for the pull request. I would have done this on my own, but had problems with my github access.

The failing test might come from this line:
```
.withSetpoint('occupied_heating_setpoint', 5, 30, 0.5, ea.STATE_SET)
```
Could you please try to remove ",es.STATE_SET"?
I guess there will be another error in this line:
```
.withSystemMode(['off', 'heat', 'auto'], ea.STATE_SET),
```
Please also try to remove ", ea.STATE_SET". I guess its not necessary!

Btw:
I use a Sonoff Dongle-E - This also works with the install code, but at the moment this seems to be very unstable.
I sometimes get errors in the log (route error) or device suddenly is no longer visible in Z2M.
If this happens I need to remove batteries from the thermostat and put them in again. The device automatically reconnects after that...

mmattel:
@BBJake 
I removed the `ea.STATE_SET`, that worked but one test error remains:
`BTH-RA - child_lock, supports get: true`
I tried to fix without success, maybe you have a clue.

BBJake:
@mmattel 
Ok, i fixed the last test failure.
See https://github.com/BBJake/zigbee-herdsman-converters/blob/master/devices/bosch.js
Replace
```
            e.child_lock(),
```
with
```
            exposes.binary('child_lock', ea.ALL, 'LOCK', 'UNLOCK')
                .withDescription('Enables/disables physical input on the device'),
```

BBJake:
Or even better - replace it with this line:
```
e.child_lock().setAccess('state', ea.ALL),
```

mmattel:
I updated the code and squashed the commits, local test an CI is green now :smile:

@Koenkk would be ready to merge

Koenkk:
Thanks!

PR #4914, opened 2022-11-08
X5H migration to the new Tuya format + time fix

vladyspavlov:
Initial, not tested at all.

An attempt to resolve https://github.com/Koenkk/zigbee2mqtt/issues/14037

PR #5033, opened 2022-11-27
Add support for TCP smart TRVs

joejoyce:
Hi,

This should add support for TCP TRVs. The ones I have don't appear to have a model number printed on them anywhere. There's an issue discussing support for these: https://github.com/Koenkk/zigbee2mqtt/issues/13820.

There's some discussion in the thread on this but I implemented the device "off" mode as part of the main device state and the "away_mode" as a separate flag, in a similar manner to etop thermostat since it appeared that that's how HA expected things to be. This does complicate (as raised in discussion on the issue) the mode handling. Is this how you'd expect this to be implemented?

There is a small issue I've noted with setup. Setup mode can be entered/exited using the boolean toggle from Zigbee2MQTT however if you then change the mode however the TRV gets into a state where it won't do anything else. This can be resolved by putting it back into setup again and completing the setup process so it's not a major problem.

Thanks
Joseph


Koenkk:
Thanks for the pr, can you refactor this to the new TuYa format? https://github.com/Koenkk/zigbee-herdsman-converters/blob/bdab93c59972c9d7e6c94eb8109f65615de79d3c/devices/tuya.js#L2078

joejoyce:
Yes, will do.

james-fry:
Thanks for this PR @joejoyce - looking fwd to trying this out.

Koenkk:
I don't see the new api being used, I would expect to see ` tuyaDatapoints`.

joejoyce:
> I don't see the new api being used, I would expect to see ` tuyaDatapoints`.

Ah, apologies, I'd misunderstood what was wanted. I've done most of the refactor now but am having some difficulty with the the away_mode/setup_mode/system_mode fields with the tuyaDatapoints API. Have I understood correctly that it's a 1:1 correspondence between a field exposed over MQTT and a data-point when using the new API?

Currently there are custom toZigbee converters to adjust both the on/off state and the system mode when setting the system_mode. Is there a way to translate this behavior to the new tuya API or should I leave those dataPoints using the old one?

Thanks

vladyspavlov:
@joejoyce Hi, you can check how to migrate from the old to the new api here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/4914

PR #5034, opened 2022-11-28
Update echostar.js

MattL0:


Koenkk:
`genOta` messages are already ignored globally: https://github.com/Koenkk/zigbee2mqtt/blob/e66fe8534e81b6199dbb4dfdacaf3e5b8ba4210b/lib/extension/receive.ts#L106

MattL0:
I can swear a do not receive ota messages in the log from the doorbell anymore with that. 
Maybe something is deactivated?

PR #5035, opened 2022-11-28
add abs_thermostat_min_heat_setpoint_limit

MattL0:
also abs_thermostat_max_heat_setpoint_limit

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -1465,6 +1465,40 @@ const converters = {
             await entity.read('hvacThermostat', ['maxHeatSetpointLimit']);
         },
     },
+    abs_thermostat_min_heat_setpoint_limit: {
+        key: ['abs_min_heat_setpoint_limit'],
+        convertSet: async (entity, key, value, meta) => {
+            let result;
+            if (meta.options.thermostat_unit === 'fahrenheit') {
+                result = Math.round(utils.normalizeCelsiusVersionOfFahrenheit(value) * 100);
+            } else {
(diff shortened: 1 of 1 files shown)


Koenkk:
Where are these converters used?

sjorge:
Just as a data point, doesn't look like all device supports this, e.g. the Ubisys H1 while it does not return UNSUPPORTED_ATTRIBUTE, it doesn't return a value either.

It seems to just silently discard the read request, it's still responds to other requests at the time so the device was not sleeping.

ZLC does list it as required though:
image


But blanket adding it all havcThermostat devices seems like a bad idea.

MattL0:
On  sinope thermostat . I didnn’t sent the pr . I am waiting for another one to be accepted before .

Here are those attributs are read and can be write. They are already in the fromzigbee converter

MattL0:
4B1119F8-C3F0-4352-8227-98D0303420C6
310D5F35-1A86-4625-9E5D-4B5DD38A6541
A88031DD-B564-4A4D-90F5-3799D2B156FA
1CA800B2-426B-4F05-88D4-E3B4DE891F89

MattL0:
A8182689-32E4-46BD-91C3-9B36D4E90E17
46D79208-5297-4431-A1B4-16204E66CB76
0F77326B-A35C-4997-8E37-F37B19766BED
0261689E-9623-4E37-AB39-AEEA99122790

MattL0:
What would be nice would be to NOT have a slidder bar for any one of them

PR #5036, opened 2022-11-28
Add new model tuya TA0601 "_TZE200_atpwqgml"

alesoft73:
Add a new model manufacturer 3 gang swicth.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1443,6 +1443,7 @@ module.exports = [
         fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_kyfqmmyl'},
             {modelID: 'TS0601', manufacturerName: '_TZE200_2hf7x9n3'},
             {modelID: 'TS0601', manufacturerName: '_TZE200_bynnczcb'}],
+            {modelID: 'TS0601', manufacturerName: '_TZE200_atpwqgml'}],
         model: 'TS0601_switch_3_gang',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

alesoft73:
> Thanks!

Thanks yopu for you big work for community!!!

I think the same 1/2/3 gang models should also be added.
What do you say ?

Koenkk:
> I think the same 1/2/3 gang models should also be added.


What are the manufacturer names of these?

alesoft73:
> > I think the same 1/2/3 gang models should also be added.
> 
> What are the manufacturer names of these?
This is in aliexpress..
https://it.aliexpress.com/item/1005001622518790.html

I think the name for 1..2 and 4 gang is the same "[TA0601 "_TZE200_atpwqgml"]

alesoft73:
> > I think the same 1/2/3 gang models should also be added.
> 
> What are the manufacturer names of these?
This is in aliexpress..
https://it.aliexpress.com/item/1005001622518790.html

I think the name for 1..2 and 4 gang is the same "[TA0601 "_TZE200_atpwqgml"]

Koenkk:
@alesoft73 they will have a different manufacturer name

alesoft73:
@Koenkk I don't known, attend when arrive another switch and in case I add the devices ;)
Thanks another for your attention !

PR #5043, opened 2022-11-29
Refactor 1123zb and 1124zb stats

MattL0:


devices/sinope.js
```diff
@@ -82,26 +87,46 @@ module.exports = [
...
+                .withPiHeatingDemand(ea.STATE_GET),
+            exposes.enum('thermostat_occupancy', ea.ALL, ['unoccupied', 'occupied'])
                 .withDescription('Occupancy state of the thermostat'),
-            exposes.enum('backlight_auto_dim', ea.SET, ['on demand', 'sensing'])
+            exposes.enum('backlight_auto_dim', ea.ALL, ['on demand', 'sensing'])
                 .withDescription('Control backlight dimming behavior'),
-            e.keypad_lockout(), e.energy(), e.power(), e.current(), e.voltage()],
+            exposes.enum('keypad_lockout', ea.ALL, ['unlock', 'lock1'])
+                .withDescription('Enables or disables the device’s buttons'),
+            exposes.numeric('power', ea.STATE_GET)
```
Koenkk, line 112: Instead of this use just `e.power()` (same for current/voltage/energy)

MattL0: yes but can e.power()  acess be set to STATE_GET?

Koenkk: `e.power().withAccess(ea.STATE_GET)`

MattL0: Done, had to set ti for all. Otherwise i get complaints

devices/sinope.js
```diff
@@ -14,26 +14,31 @@ module.exports = [
...
         description: 'Zigbee line volt thermostat',
         meta: {thermostat: {dontMapPIHeatingDemand: true}},
-        fromZigbee: [
-            fz.legacy.sinope_thermostat_att_report, fz.legacy.hvac_user_interface, fz.electrical_measurement, fz.metering,
+        fromZigbee: [fz.legacy.sinope_thermostat_att_report, fz.legacy.hvac_user_interface, fz.electrical_measurement, fz.metering,
             fz.ignore_temperature_report, fz.legacy.sinope_thermostat_state, fz.sinope_thermostat],
-        toZigbee: [
-            tz.thermostat_local_temperature, tz.thermostat_occupied_heating_setpoint, tz.thermostat_unoccupied_heating_setpoint,
+
+        toZigbee: [tz.thermostat_local_temperature, tz.thermostat_occupied_heating_setpoint, tz.thermostat_unoccupied_heating_setpoint,
```
Koenkk, line 20: Why was `tz.sinope_thermostat_time` removed here?

devices/sinope.js
```diff
@@ -44,36 +49,21 @@ module.exports = [
...
             await reporting.thermostatOccupiedHeatingSetpoint(endpoint);
-
-            try {
-                await reporting.thermostatSystemMode(endpoint);
-            } catch (error) {/* Not all support this */}
-
-            try {
-                await reporting.thermostatRunningState(endpoint);
-            } catch (error) {/* Not all support this */}
+            await reporting.thermostatSystemMode(endpoint);
```
Koenkk, line 52: I don't think the try/catch should be removed as not all thermostats seem to support this and this causes the configure to fail.

devices/sinope.js
MattL0: Sorry forgot to readd it.

devices/sinope.js
MattL0: Set back both try catches for both thermostats this time. 

Running state seems to be handle by piheatingdemand (if more than 10 = running). 
I set it back anyways

Koenkk:
Thanks!

PR #5044, opened 2022-11-29
Add support for Owon WSP402

adamkoch:
I purchased this [Oz Zigbee Smart Plug](https://www.ozsmartthings.com.au/products/oz-zigbee-smart-plug) which identified itself as WSP402 (even though label said WSP403). I tested support by adding my own .js and it worked fine using WSP404 definition.


Koenkk:
Thanks!

Issue #3279, opened 2021-11-04
Support for Immax NEO 07505L


paffeke:
Is it possible to create new device for "Immax NEO 07505L"?
Thank you so much

PR #5045, opened 2022-11-29
#3279 - Support for Immax NEO 07505L

JiriBalcar:


Koenkk:
Thanks for the pr, could you refactor this to the new TuYa format? 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/ee22549826f0e1c835427a10097ff6ef5b0621e4/devices/tuya.js#L850

JiriBalcar:
Hi @Koenkk, i refactored it to new format. Please check it if I understood what you wanted. I had to create new converter, because I did not find any existing which could be used. This device have separate dataPoints for every action so I need to return {action: "ACTION_NAME"}. 

Thanks

Koenkk:
Looks good, I did a small refactor, can you check if it still works?

Koenkk:
Thanks!

marijandomi:
Is it possible to show which code triggered the action? because 10 codes can be stored on the keyboard. but it makes no sense to use more than one code when they all trigger the same action (arm_away, disarm, arm_home, sos).

JiriBalcar:
Well the keypad sends "code ID" (code 1, code 2, code 3) in separate message so it could be added. I did not have time to do it. Maybe in couple of days I could try to add attribute "last code used"?

marijandomi:
That would be great.  Thank you.

marijandomi:
One more question... :) is there any message about the quick function... Because the keyboard can be programmed with the quick function and you don't need to enter a code... E.g.  To open the garage door with code+arm_away, and close it only with quick_arm_away, the Yale Linus Smart Lock keyboard has such a function.

JiriBalcar:
@marijandomi Do you have any documentation about the quick function settings on this keypad?

marijandomi:
I only have the instructions that came with the keyboard and they say:
Enter setup mode: Admin code + #
Then...

Enable/disable quick arm
Enable: 3610#
Disable: 3611#

...

Notices:
when enabling quick ARM/DISARM/SOS function, you can arm/disarm/SOS without typing user code.

I tried to do it myself: arm/disarm/SOS can be turned on even without a user code. But the zigbee2mqtt gives the standard output (arm_away, disarm, arm_home, sos). it would be good if he could send in a separate message or attribute as you said he can with codeID.

marijandomi:
@JiriBalcar Do you have any news of "code ID" message?

PR #5046, opened 2022-11-29
Support the second button of WDE002906

ferrix:
Fix for the issue I brought up in https://github.com/Koenkk/zigbee2mqtt/issues/15239

### DIFF
--- a/devices/schneider_electric.js
+++ b/devices/schneider_electric.js
@@ -763,7 +763,7 @@ module.exports = [
         fromZigbee: [fz.command_on, fz.command_off, fz.command_move, fz.command_stop, fz.battery],
         toZigbee: [],
         endpoint: (device) => {
-            return {'left': 21, 'right': 22};
+            return {'right': 21, 'left': 22};
         },
         meta: {multiEndpoint: true},
         exposes: [e.action(['on_left', 'off_left', 'on_right', 'off_right', 'brightness_move_up_left', 'brightness_stop_left',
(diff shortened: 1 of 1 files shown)


manfred-mp:
Hi, have you this device and tested it? To my eyes actions are a bit messed up and also battery reporting is not there anymore? Btw I'm the author of this 1 button version and now have also 2 button devices with working converter. This should not be merged as is imho. Sorry about not creating merge request about this earlier.

Paste of critical part here (my current code): 
```
    endpoint: (device) => {
        return {'s1': 21, 's2': 22, 's3': 23, 's4': 24};
    },
    exposes: [e.action(['on', 'off', 'brightness_move_up', 'brightness_move_down', 'brightness_stop']), 
        e.battery()],
    meta: {multiEndpoint: true},
```

ferrix:
I have the device and I have tested this. I am a recent migrant from ZHA to this device and previously created a Gist for HA with the same configuration:

https://gist.github.com/ferrix/ca7cda4c3b808419d0d6d52317e3a807

I did it there using the template for 'FLS/AIRLINK/4'.

https://community.home-assistant.io/t/schneider-wiser-remote/407990

My implementation is an adaptation of the AIRLINK version from the same schneider.js file and there IS the battery but only for the left button or endpoint 21 since there is no use duplicating it (and I am not actually sure if the device supports it in the other endpoints)

ferrix:
The code you pasted also has the flaw that even though the device apparently exposes endpoints 23 and 24, they are not really used and should not be published.

ferrix:
Oh, I reviewed it again. Apparently was missing a commit there fixing a search and replace issue.

Koenkk:
Thanks!

PR #5049, opened 2022-11-29
philips: support setting gradient

zegl:
API receives a of comma separated hex RGB values.

Added to 4080248U9 / 915005987201 to start with.

This updates https://github.com/Koenkk/zigbee2mqtt/issues/15243

Depends on https://github.com/Koenkk/zigbee-herdsman/pull/622

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -94,6 +94,11 @@ const fzLocal = {
                     const colors = philips.decodeGradientColors(input, options);
                     return {colors};
                 }
+                if (msg.data.hasOwnProperty('state')) {
+                    const input = msg.data['state'].toString('hex');
+                    const colors = philips.decodeGradientColors(input, opts);
+                    return {colors};
+                }
                 return {};
             },
         };
@@ -195,6 +200,9 @@ const tzLocal = {
                 const scene = philips.encodeGradientColors(value, opts);
                 const payload = {data: Buffer.from(scene, 'hex')};
                 await entity.command('manuSpecificPhilips2', 'multiColor', payload);
+            },
+            convertGet: async (entity, key, meta) => {
+                await entity.read('manuSpecificPhilips2', ['state']);
             },
         };
     },
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -41,6 +41,29 @@ const hueExtend = {
         toZigbee: extendDontUse.light_onoff_brightness_colortemp_color({supportsHS: true, ...options})
             .toZigbee.concat([tz.hue_power_on_behavior, tz.hue_power_on_error]),
     }),
+    light_onoff_brightness_colortemp_color_gradient: (options={}) => ({
+        ...extendDontUse.light_onoff_brightness_colortemp_color({supportsHS: true, noConfigure: true, ...options}),
+        ota: ota.zigbeeOTA,
+        meta: {turnsOffAtBrightness1: true},
+        configure: async (device, coordinatorEndpoint, logger) => {
+            await extendDontUse.light_onoff_brightness_colortemp_color({supportsHS: true, ...options})
+                .configure(device, coordinatorEndpoint, logger);
+            for (const ep of device.endpoints) {
+                await ep.bind('manuSpecificPhilips2', coordinatorEndpoint);
+            }
+        },
+        exposes: extendDontUse.light_onoff_brightness_colortemp_color({supportsHS: true, ...options}).exposes.concat([
+            exposes.enum('gradient_scene', ea.SET, Object.keys(gradientScenes)),
+            exposes.list('colors', ea.ALL, exposes.text('hex', 'Color in RGB HEX format (eg #663399)'))
+                .withLengthMin(1)
+                .withLengthMax(9)
+                .withDescription('List of RGB HEX colors'),
+        ]),
+        fromZigbee: extendDontUse.light_onoff_brightness_colortemp_color({supportsHS: true, ...options}).fromZigbee.concat(
+            [fzLocal.philips_hue_multicolor({reverse: true})]),
+        toZigbee: extendDontUse.light_onoff_brightness_colortemp_color({supportsHS: true, ...options}).toZigbee.concat(
+            [tz.hue_power_on_behavior, tz.hue_power_on_error, tzLocal.gradient_scene, tzLocal.philips_hue_multicolor({reverse: true})]),
+    }),
 };
 
 const fzLocal = {
@@ -83,7 +106,6 @@ const fzLocal = {
             return payload;
         },
     },
-
     philips_hue_multicolor: (opts = {reverse: false}) => {
         return {
             cluster: 'manuSpecificPhilips2',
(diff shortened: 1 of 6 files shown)


devices/philips.js
```diff
@@ -1636,9 +1673,16 @@ module.exports = [
...
-        exposes: [exposes.enum('gradient_scene', ea.SET, Object.keys(gradientScenes)),
-            ...hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}).exposes],
+        toZigbee: [
+            tzLocal.gradient_scene,
+            tzLocal.philips_hue_multicolor({reverse: true}),
+            ...hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}).toZigbee,
+        ],
+        exposes: [
+            exposes.enum('gradient_scene', ea.SET, Object.keys(gradientScenes)),
+            exposes.text('philips_hue_multicolor', ea.SET),
```
Koenkk, line 1683: - I suggest to rename this to `colors`
- We can use the new list exposes instead: https://github.com/Koenkk/zigbee-herdsman-converters/blob/4c961732b32d34905b9b04738018c7b7b3b1384a/lib/exposes.js#L117

zegl: Thanks for the quick review! I'll take a look and implement your suggested changes. :-)

zegl:
Thanks for your feedback, I've now incorporated your suggestions. 👍 

As a future improvement, I was thinking if it was worth exposing some metadata/property like a [min, max] number of supported colors. What would be the preferred way of achieving this? Would adding `withLengthMin`/`withLengthMax` to the `List` class be a good idea?

devices/philips.js
```diff
@@ -1636,9 +1672,16 @@ module.exports = [
...
-        exposes: [exposes.enum('gradient_scene', ea.SET, Object.keys(gradientScenes)),
-            ...hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}).exposes],
+        toZigbee: [
+            tzLocal.gradient_scene,
+            tzLocal.philips_hue_multicolor({reverse: true}),
+            ...hueExtend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}).toZigbee,
+        ],
+        exposes: [
+            exposes.enum('gradient_scene', ea.SET, Object.keys(gradientScenes)),
+            exposes.list('colors', ea.SET, exposes.text('color', ea.SET)),
```
Koenkk, line 1682: Can you add a description to the `color`? (required format example). Something like:

`exposes.text('color', ea.SET)` -> `exposes.text('color', ea.SET).withDescription('Color in hex e.g. #DEEB34')`

zegl: Yes, will do! I figured out that it probably makes more sense to use X/Y colors instead of RGB, as that's what the device is using, so I'm rewriting this now. I'll ping you once it's ready for another review.

zegl:
Changes since last time:

* Switched the API to XY color space, as this is what the device protocol uses
* Managed to get a fromZigbee working, seems to work pretty well!
* Added support for setting between 1 and 9 colors

Please take a look, and let me know if there is anything I should improve before this can be merged!

zegl:
The tests are failing, seems to be related to 56d18d1a49c3eaf3740d603ffe9a48465f6dee51.

Koenkk:
I think RGB (hex) colors might be more user friendly as it seems to be hard to figure out x/y values.

zegl:
Yeah, agreed that RGB is easier. I changed my mind because of consistency reasons, none of the other converters for Philips devices seems to have RGB APIs, seems like they all are using x/y or h/s. So it felt odd to add RGB/hex to the mix, but I can change it back if you want me to.

Koenkk:
Lets change it back, this is typically something that will be entered from the frontend.

zegl:
Ok, I think we're getting somewhere. I'm starting to get more comfortable with the codebase now, and have made things a lot better now. 😄 

1. I've added `convertGet` for the gradients (depends on https://github.com/Koenkk/zigbee-herdsman/pull/622 to define `0xFC03` / `0x0002` as "state").
2. Changed the API to use RGB HEX. I added ColorRGB.toHex().
3. Moved the parsing to lib/philips.js and added some basic tests for it. I think that this will be really useful down the line if we add support for more features of the gradient payload.

I've been doing end-to-end tests using the latest `master` branch of zigbee2mqtt, and on top of the PR I made to zigbee-heardsman. Everything seems to be working great!

After this is merged (this PR is starting to get a bit too big in my opinion). I'm planning to add this feature to all Hue Gradient devices, and fix up the list of devices a bit. I'll also add a gradient picker to the frontend project.

Thanks for your support so far, big thanks! Hope we can get this merged. :-)

devices/philips.js
```diff
@@ -82,6 +83,26 @@ const fzLocal = {
             return payload;
         },
     },
+
+    philips_hue_multicolor: (opts = {reverse: false}) => {
+        return {
+            cluster: 'manuSpecificPhilips2',
+            type: ['attributeReport', 'readResponse'],
+            convert: (model, msg, publish, options, meta) => {
+                if (msg.data.hasOwnProperty('2')) {
```
sjorge, line 92: Should the '2' here be 'multiColor' ?

zegl: I probably should have commented on it, it's for compatibility until https://github.com/Koenkk/zigbee-herdsman/pull/622 get's merged which maps `2` to `state` (used below).

zegl: Done!

Koenkk:
@zegl I did a small refactor, created a new extend for the gradient to reduce code when adding this for more devices (which I also did). 

- Can you do a final check if everything works?
- I'm wondering instead of `colors`, `gradient_colors` is a better name (since `gradient_scene` is similar to this)

zegl:
@Koenkk Nice refactor! I made a test and everything is working as it should. :-)

zegl:
@Koenkk I think that `colors` is cleaner, but I'll let you decide. I'd assume that we could deprecate `gradient_scene` soon, as `colors` supports everything that `gradient_scene` does and more.

Koenkk:
It seems the way Philips advertises the multicolor feature is gradient (e.g. https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-festavia-string-lights/046677580148). So I think it would be good to name it similar in z2m.

sjorge:
> It seems the way Philips advertises the multicolor feature is gradient (e.g. https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-festavia-string-lights/046677580148). So I think it would be good to name it similar in z2m.
> 
> The scene part can stay for people that are too lazy to make their own set of colors 😄

Maybe just `gradient` then and mark `gradient_scene` as deprecated?

Koenkk:
@sjorge that also sounds good!

zegl:
> Maybe just gradient then and mark gradient_scene as deprecated?

Sounds good, I'll make the change.

Koenkk:
Awesome, big thanks!

PR #5055, opened 2022-11-30
ubisys: D1 read energy

heinzma:
currentSummDelivered is read only. With this energy will be read when power is reported.


heinzma:
sorry do not have a development environment setup at the moment.

Koenkk:
Thanks!

PR #5057, opened 2022-12-01
wyze lock converter based on issue #3018

salman2learn:
Added wyze lock converter based on https://github.com/Koenkk/zigbee2mqtt/issues/3018#issuecomment-1165982160.


Koenkk:
Thanks!

PR #4763, opened 2022-10-09
[SRTS-A01](Aqara Smart Radiator Thermostat E1) Added smart schedule support

artem-sedykh:

PR #5058, opened 2022-12-01
Aqara SRTS-A01 smart schedule

protyposis:
Add smart scheduling support for Aqara Smart Radiator Thermostat E1 (SRTS-A01). This is a refactored and improved version of the stale https://github.com/Koenkk/zigbee-herdsman-converters/pull/4763.

This adds the following:
 * `schedule`: A toggle to turn on/off the smart schedule.
 * `schedule_settings`: A string representation of the schedule settings. Example (default): `mon,tue,wed,thu,fri,sat,sun|8:00,24.0|18:00,17.3|23:00,22.0|8:00,22.0`

I tried to expose the schedule as a configuration object which can be nicely managed through the frontend, but this is currently not possible due to various limitations (see discussion in https://github.com/Koenkk/zigbee-herdsman-converters/pull/5058#pullrequestreview-1201469128). Exposing it as a string is a first step to figure out if this is even useful.

A schedule consists of a selection of weekdays and exactly 4 events that set a specific temperature at a specific time. The temperatures of the first 3 events are those that can be controlled in the Aqara app, the last one is always set to the previous one in the Aqara app. I decided to explicitly expose it anyway, but I don't know if it's interpreted by the TRV.

I have successfully tested reading and writing the settings, and made sure the written data corresponds exactly to the factory-set data. However, I have not tested whether the schedule is actually executed by the TRV, i.e., if it actually works.


0ip:
Thanks a lot for your effort! Really hope this PR makes it through.

Have you been able to work out how `preset` behaves when the schedule is active?

devices/xiaomi.js
```diff
@@ -98,12 +99,24 @@ const fzLocal = {
...
+                    result['valve_alarm'] ={1: true, 0: false}[value.readUInt32LE(36)];
+                    result['occupied_heating_setpoint'] = value.readUint16LE(32)/100;
+                    result['local_temperature'] = value.readUint16LE(28)/100;
+                    break;
+                case 0x027d:
+                    result['schedule'] = {1: 'ON', 0: 'OFF'}[value];
+                    break;
+                case 0x0276: {
+                    const schedule = xiaomiUtils.readSchedule(value);
+                    result['schedule_settings_json'] = schedule;
```
Koenkk, line 112: Why expose this twice? My preference would be to drop `schedule_settings_json`

devices/xiaomi.js
```diff
@@ -2564,6 +2595,11 @@ module.exports = [
             exposes.binary('window_open', ea.STATE, true, false),
             e.valve_detection().setAccess('state', ea.ALL),
             e.away_preset_temperature().withAccess(ea.ALL),
+            exposes.switch().withState('schedule', true,
+                'When being ON, the thermostat will change its state based on your settings',
+                ea.ALL, 'ON', 'OFF'),
+            exposes.text('schedule_settings', ea.ALL)
```
Koenkk, line 2601: Can we make this expose a composite or list like: https://github.com/Koenkk/zigbee-herdsman-converters/blob/a588bdeba82fb38bb3d8c7a309779e73744f7609/devices/xiaomi.js#L2592

lib/xiaomiUtils.js
```diff
@@ -0,0 +1,233 @@
+function readTemperature(buffer, offset) {
```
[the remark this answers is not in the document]
protyposis: I renamed the file in 309a0b1 to make it clear that it only contains TRV-related functions. `xiaomi.js` seems to contain mainly generic functions. So in terms of separation of concerns, I don't see a reason why they should to go into the same file.

devices/xiaomi.js
protyposis: Compared to most settings, this has quite a complicated format. Therefore, `schedule_settings_json` is meant for programmatic use and power users, while `schedule_settings` is a simplified alternative for simple management through the frontend. What speaks against providing both?

devices/xiaomi.js
protyposis: Is there any detailed documentation on that, specifically how to create a list with a fixed number of composite entries, and how to handle the resulting data in `toZigbee`?

devices/xiaomi.js
Koenkk: I prefer the `schedule_settings_json` over the `schedule_settings` (drop `schedule_settings` and rename `schedule_settings_json` to `schedule_settings`).

If it is exposed correctly (https://github.com/Koenkk/zigbee-herdsman-converters/pull/5058/files#r1037429975) then it will work for all users.

protyposis:
> Have you been able to work out how `preset` behaves when the schedule is active?

Unfortunately not. I expected some dependencies, e.g., that the `preset` changes to `auto` when enabling the schedule, or the schedule enables when switching to `auto`, but I couldn't identify any such interdependencies. So I was actually hoping to merge this, gather feedback from users, and improve in a follow-up PR. Unless a motivated reviewer with the required knowledge comes along who investigates this.

devices/xiaomi.js
protyposis: That would be great and my preferred way too, but as I commented below, I am lacking the guidance to implement it like this.

devices/xiaomi.js
protyposis: Figured I can define 4 static composites instead of using the list expose. Is this what you meant?
image

Is it possible to 
 * change the "days" enum dropdown to a multiselect?
   * alternatively to the dropdown, is it possible to change the "days2" switches to checkboxes and/or position them horizontally so they don't waste so much space?
 * remove all individual "Apply" buttons and have only one for the whole composite? (This is all managed by a single Zigbee message so it doesn't make sense to apply them individually)
 * Add a "get" button to the overall composite instead of the individual properties? (same reason as above)

Still wondering how to submit this form data and process it in the converter.

Koenkk: This can get a bit complicated indeed, I'm wondering if the schedule can also set per day/entry? If yes, then we can do something like this: https://github.com/Koenkk/zigbee-herdsman-converters/blob/a069993cd8ca18dd4dc3d9198aa3bb52e7518467/devices/tuya.js#L2077 . In this case there is one expose just for setting the schedule, and below are multiple exposes (read only) showing the schedule per day.

protyposis: Nope, this is all-or-nothing.

Koenkk: Can you post the current exposes code?

protyposis:
```js
        exposes.composite('schedule_settings_composite')
            .withFeature(exposes.enum('days', exposes.access.STATE_SET, ['mon', 'tue', 'wed', 'thu', 'fri', 'sat', 'sun'])) // TODO multiselect
            .withFeature(exposes.composite('days2', ea.ALL)
                .withFeature(exposes.switch().withState('mon', true, undefined, ea.STATE_SET))
                .withFeature(exposes.switch().withState('tue', true, undefined, ea.STATE_SET))
                .withFeature(exposes.switch().withState('wed', true, undefined, ea.STATE_SET))
                .withFeature(exposes.switch().withState('thu', true, undefined, ea.STATE_SET))
                .withFeature(exposes.switch().withState('fri', true, undefined, ea.STATE_SET))
                .withFeature(exposes.switch().withState('sat', true, undefined, ea.STATE_SET))
                .withFeature(exposes.switch().withState('sun', true, undefined, ea.STATE_SET)))
            .withFeature(exposes.composite('start', ea.STATE_SET)
                .withFeature(exposes.numeric('hour', ea.STATE_SET))
                .withFeature(exposes.numeric('minute', ea.STATE_SET))
                .withFeature(exposes.numeric('temperature', ea.STATE_SET)))
            .withFeature(exposes.composite('intermediate1', ea.STATE_SET)
                .withFeature(exposes.numeric('hour', ea.STATE_SET))
                .withFeature(exposes.numeric('minute', ea.STATE_SET))
                .withFeature(exposes.numeric('temperature', ea.STATE_SET)))
            .withFeature(exposes.composite('intermediate1', ea.STATE_SET)
                .withFeature(exposes.numeric('hour', ea.STATE_SET))
                .withFeature(exposes.numeric('minute', ea.STATE_SET))
                .withFeature(exposes.numeric('temperature', ea.STATE_SET)))
            .withFeature(exposes.composite('end', ea.STATE_SET)
                .withFeature(exposes.numeric('hour', ea.STATE_SET))
                .withFeature(exposes.numeric('minute', ea.STATE_SET))
                .withFeature(exposes.numeric('temperature', ea.STATE_SET)))
            .withDescription('Smart schedule configuration'),
```

Koenkk: Thanks, I'm wondering wether this should become a list (this seems more logical to me). Similar to https://github.com/Koenkk/zigbee-herdsman-converters/blob/a588bdeba82fb38bb3d8c7a309779e73744f7609/devices/xiaomi.js#L2592 

So the list contains the schedule entries, each schedule entry contains a day, time and temperature.

protyposis: See my question above:

> Is there any detailed documentation on [...] how to create a list **with a fixed number of composite entries**?

This TRV has a very basic scheduler implementation that requires a fixed amount of 4 time/temperature entries, and the current list implementation doesn't seem to allow to require a fixed number - or did I miss this? Also, the schedule only allows a global selection of days - not multiple configurations for multiple days. Please check the description of this PR, it documents exactly how the schedule settings look like.

Koenkk: > Also, the schedule only allows a global selection of days - not multiple configurations for multiple days

Can't this be achieved by wrapping the `days` in a list?

@nurikk we need your help here.

> requires a fixed amount of 4 time/temperature entries

For the list type we can add a `min`/`max` entries, then for this usecase we set both to 4. Could this be supported in the frontend?

protyposis: > Can't this be achieved by wrapping the `days` in a list?

Do you mean like this?

```js
exposes.composite('scheduleTest')
  .withFeature(exposes.list('days', ea.STATE_SET, 
    exposes.enum('days', exposes.access.STATE_SET, ['mon', 'tue', 'wed', 'thu', 'fri', 'sat', 'sun'])))
```
This would allow invalid or at undesired combinations like `['wed', 'tue', 'mon']` (unexpected order) or `['tue', 'tue']` (unexpected duplicates). I know it can be filtered, but it still creates an odd user experience. The day selection needs something like this
image
or this
image

Koenkk: That would be better indeed, @nurikk can this be implemented in the frontend?

protyposis: @nurikk excuse me, what did you say? What exactly should be supported, and where?

protyposis: What is a "link expose"? The state you are looking for is documented in the PR description.

nurikk: sorry, my mobile spelling sucks. I meant `list exposes feature`

nurikk: Min/max, can be supported. Kindly raise an issue in zigbee2mqtt-frontend issue tracker, I’ll check it this weekend

protyposis: So does that mean that the mockups cannot be realized?

Koenkk: @protyposis can you provide me an example of what you publish over MQTT to set `schedule_settings_json`?

protyposis: An example is in the PR description:

> ```json
>  {
>      "repeat": ["mon", "tue", "wed", "thu", "fri", "sat", "sun"],
>      "events": [
>          {"time": 480, "temperature": 24},
>          {"time": 1080, "temperature": 17.3},
>          {"time": 1380, "temperature": 22},
>          {"time": 480, "temperature": 22},
>      ],
>  }
> ```

To accommodate it to the proposed composite structure in https://github.com/Koenkk/zigbee-herdsman-converters/pull/5058/files#discussion_r1038766761, I tried it like this:

```js
{
  days: { mon: 'ON', tue: 'ON', wed: 'ON', thu: 'ON', fri: 'ON', sat: 'ON', sun: 'ON' },
  start: { hour: 8, minute: 0, temperature: 24 },
  intermediate1: { hour: 18, minute: 0, temperature: 17.3 },
  intermediate2: { hour: 23, minute: 0, temperature: 22 },
  end: { hour: 8, minute: 0, temperature: 22 },
}
```

Koenkk: If I understood correctly, this is the full and only schedule, right?

e.g. is it possible to set a different schedule for a day? e.g.

```
 {
     "repeat": ["mon", "tue", "wed", "thu", "fri"],
     "events": [
         {"time": 480, "temperature": 24},
         {"time": 1080, "temperature": 17.3},
         {"time": 1380, "temperature": 22},
         {"time": 480, "temperature": 22}
     ],
 }
```

```
 {
     "repeat": ["sat", "sun"],
     "events": [
         {"time": 520, "temperature": 24},
         {"time": 1500, "temperature": 22}
     ],
 }
```

Koenkk: Thanks, I will check this with @nurikk

Koenkk: @protyposis wanted to let you know I did not forget about this, working in https://github.com/nurikk/zigbee2mqtt-frontend/pull/1638

itsSaad:
@protyposis This PR seems to be stuck same as #4763.
What needs to be done. Can I help you to unblock this PR.
I just acquired two of these TRVs and feel that many functionalities are missing.
* Schedules
* External Temp Sensor
* No indication if the Valve is closed or open. I expected an `idle` hvac_action/system_mode.
Not sure what other functionalities the aqara app has.

Im willing to test this if that will unblock this effort.

devices/xiaomi.js
protyposis: Great, thanks @Koenkk. Will follow up here when https://github.com/nurikk/zigbee2mqtt-frontend/pull/1638 is finalized.

protyposis:
@itsSaad this PR is only about the schedule. Please discuss unrelated features elsewhere, e.g. in https://github.com/Koenkk/zigbee2mqtt/issues/13993.

itsSaad:
@protyposis Im not discussing those but just mentioned them. My comment was mainly to offer help.

devices/xiaomi.js
Koenkk: Apologies for taking so long, but all should be good now (in the latest dev). 

You can use the following expose:

```js
exposes.composite('Schedule', 'schedule', ea.ALL)
    .withFeature(exposes.composite('Days', 'days')
        .withFeature(exposes.switch().withState('mon', true, undefined, ea.STATE))
        .withFeature(exposes.switch().withState('tue', true, undefined, ea.STATE))
        .withFeature(exposes.switch().withState('wed', true, undefined, ea.STATE))
        .withFeature(exposes.switch().withState('thu', true, undefined, ea.STATE))
        .withFeature(exposes.switch().withState('fri', true, undefined, ea.STATE))
        .withFeature(exposes.switch().withState('sat', true, undefined, ea.STATE))
        .withFeature(exposes.switch().withState('sun', true, undefined, ea.STATE)),
    )
    .withFeature(exposes.composite('Start', 'start')
        .withFeature(exposes.numeric('Hour', ea.STATE).withProperty('hour'))
        .withFeature(exposes.numeric('Minute', ea.STATE).withProperty('minute'))
        .withFeature(exposes.numeric('Temperature', ea.STATE).withProperty('temperature')))
    .withFeature(exposes.composite('Intermediate 1', 'intermediate1')
        .withFeature(exposes.numeric('Hour', ea.STATE).withProperty('hour'))
        .withFeature(exposes.numeric('Minute', ea.STATE).withProperty('minute'))
        .withFeature(exposes.numeric('Temperature', ea.STATE).withProperty('temperature')))
    .withFeature(exposes.composite('Intermediate 2', 'intermediate2')
        .withFeature(exposes.numeric('Hour', ea.STATE).withProperty('hour'))
        .withFeature(exposes.numeric('Minute', ea.STATE).withProperty('minute'))
        .withFeature(exposes.numeric('Temperature', ea.STATE).withProperty('temperature')))
    .withFeature(exposes.composite('End', 'end')
        .withFeature(exposes.numeric('Hour', ea.STATE).withProperty('hour'))
        .withFeature(exposes.numeric('Minute', ea.STATE).withProperty('minute'))
        .withFeature(exposes.numeric('Temperature', ea.STATE).withProperty('temperature')))
    .withDescription('Smart schedule configuration'),
```

On you use this and submit it through the z2m frontend -> exposes tab, you will see the structure in the z2m debug log.

protyposis: On the latest HA dev version I get `N/A` everywhere. if I change the `ea.STATE` to `ea.ALL`, I get the input fields but with reload buttons everywhere again. Any idea what's wrong?
image

Koenkk: Is the property in the correct structure? (can you provide the debug log when it starts showing `N/A`)

protyposis: Right, the `N/A` issue is fixed by supplying the correct structure, still there are 3 issues left:
 1. I need to use `ea.STATE_SET` instead of your suggested `ea.STATE` to get editable input fields instead of read-only text labels. Is this ok?
 2. When editing fields, other fields get unexpectedly cleared. For example, if I click into the `minute` field in `start` and change the value, the `hour` and `temperature` get unexpectedly cleared (emptied). Same behaviour in the other 3 time/temperature composites.
 3. The `Apply` button never gets enabled, even when all fields are filled, so the schedule can never be submitted.

To be perfectly honest, this PR is getting out of hand and I would be very much in favor if I could just clean it up and we merge it with the text configuration I initially proposed (it wouldn't be the only device with a text-config). If there's demand, which we don't even know yet, someone can later handle the UI aspect in a separate PR and make it more user friendly. Could you live with this iterative approach?

Koenkk: 1. Yes that is fine
2. I cannot reproduce it, any chance you create a GIF for this?
3. Same as 2, can you also make a GIF of this? It should be enabled when clicking next to the text field after filling in the last value.


> To be perfectly honest, this PR is getting out of hand

I agree, I'm also fine in accepting it with just a text expose but with the JSON structure that this expose will send. Because when this gets merged, we cannot change the format anymore (since it will give a breaking change).

protyposis: Here's a screen recording in Chrome containing both issues (disabled `Apply` button at end of video). Just changing focus from one text field to another already clears the values.

Recording 2023-01-24 at 08 10 00

> when this gets merged, we cannot change the format anymore (since it will give a breaking change)

Can't the value be upgraded later with a type check, e.g. 
```js
if (typeof value === 'string') {
  // Upgrade old string to new JSON representation to avoid breaking change
  value = upgradeStringToJson(value);
}```
?

Comment (Koenkk) [reply devices/xiaomi.js:2726]: Could it be that the thermostats publishes the value while changing this? (then this should solved by changing `exposes.composite('schedule'` to `exposes.composite('schedule_set'`, then change `ea.STATE_SET` to `ea.SET`

> Can't the value be upgraded later in convert or configure with a type check, e.g.

I'm fine with that.

Comment (protyposis) [reply devices/xiaomi.js:2726]: This unfortunately doesn't fix it either, I'm giving up on the UI for now and will revert to the configuration string.

Comment (Koenkk) [reply devices/xiaomi.js:2726]: Agree!

File: devices/xiaomi.js [hunk 4]
```diff
@@ -2689,6 +2720,11 @@ module.exports = [
...
+                .withDescription('Smart schedule configuration (also exposed as JSON object through "schedule_settings_json")'),
```
Koenkk, line 2727: Could you add an example to the description?

lib/xiaomiTrvUtils.js
```diff
@@ -0,0 +1,233 @@
+function readTemperature(buffer, offset) {
```
Koenkk, line 1: Could you move this file into https://github.com/Koenkk/zigbee-herdsman-converters/blob/8be00cc65b20e3eb0f2726d74948f490db6536db/lib/xiaomi.js#L864 ?

Koenkk:
@protyposis I left two minor comments, after that it is ready for merge. Can you also fix the merge conflict?

devices/xiaomi.js
protyposis: Added

lib/xiaomiTrvUtils.js
protyposis: Moved

Koenkk:
Thanks!

PR #5059, opened 2022-12-01
Remove tamper sensor

kbickar:
The SNZB-03 physically does not have a tamper sensor so the value of this is meaningless and should be removed


Koenkk:
Thanks!

PR #5061, opened 2022-12-02
Added support for status LED control on Niko light switches

FHannes:
I've implemented control of the LED on the Niko 552-721X1 and 552-721X2 switches. It seems to use the same attributes as the smart outlets. Setting the LED state works fine both with the relay being controlled directly or being decoupled, but obviously when it is being controlled (given that the led is not disabled), it will update to the match the relay state on the next action.

The Niko switch is really fast standalone or connected to Zigbee2Mqtt. When connected to the Niko hub, it is just as slow as the EasyWave variant of the switches. Which indicates to me that Niko probably always has the relay decoupled and fully controls both the relay and LED from the Hub based on the switch input. So with these additional properties it should be possible to mimic that behavior through an automation platform like HomeAssistant or OpenHAB. For example: I have all of my EasyWave switches set to inverse LED behavior, so the LED is on when the light is off (so you see the switch when it's dark in the room).

I have some duplication in the converters between the switches and the outlets, because I thought it may be prudent to keep them separated. But if it's preferable, they could be merged to some extent. I only have a 552-721X1 switch myself. The double and dimmer switches I have are all EasyWave, so I was not able to test with those. But given the existing config, it seems reasonable to assume that at least the double 552-721X2 will behave exactly the same. It will probably also go for the dimmer switch, but since those are still somewhat different in functionality, I figured I'd not add it if I can't test it.

Finally, 2 things I observed whilst testing this:
- Like the smart socket, the light switches seem to support the outletChildLock attribute, but I couldn't quite pin down how it's supposed to work. Maybe it's not supposed to work or has a firmware bug...
- The smart socket's LED can also be controlled, but it seems to reset to the active state of the socket automatically after a second or so.

### DIFF
--- a/devices/niko.js
+++ b/devices/niko.js
@@ -42,10 +42,10 @@ const local = {
             convert: (model, msg, publish, options, meta) => {
                 const state = {};
                 if (msg.data.hasOwnProperty('outletLedState')) {
-                    state['led_enable'] = (msg.data['outletLedState'] == 1);
+                    state['led_enable'] = (msg.data['outletLedState'] == 1 ? 'ON' : 'OFF');
                 }
                 if (msg.data.hasOwnProperty('outletLedColor')) {
-                    state['led_state'] = (msg.data['outletLedColor'] == 255);
+                    state['led_state'] = (msg.data['outletLedColor'] == 255 ? 'ON' : 'OFF');
                 }
                 return state;
             },
@@ -59,7 +59,7 @@ const local = {
                     state['child_lock'] = (msg.data['outletChildLock'] == 0 ? 'LOCK' : 'UNLOCK');
                 }
                 if (msg.data.hasOwnProperty('outletLedState')) {
-                    state['led_enable'] = (msg.data['outletLedState'] == 1);
+                    state['led_enable'] = (msg.data['outletLedState'] == 1 ? 'ON' : 'OFF');
                 }
                 return state;
             },
@@ -86,8 +86,8 @@ const local = {
         switch_led_enable: {
             key: ['led_enable'],
             convertSet: async (entity, key, value, meta) => {
-                await entity.write('manuSpecificNiko1', {'outletLedState': ((value) ? 1 : 0)});
-                return {state: {led_enable: ((value) ? true : false)}};
+                await entity.write('manuSpecificNiko1', {'outletLedState': ((value.toLowerCase() === 'off') ? 0 : 1)});
+                return {state: {led_enable: ((value.toLowerCase() === 'off') ? 'OFF' : 'ON')}};
             },
             convertGet: async (entity, key, meta) => {
                 await entity.read('manuSpecificNiko1', ['outletLedState']);
@@ -96,8 +96,8 @@ const local = {
         switch_led_state: {
             key: ['led_state'],
             convertSet: async (entity, key, value, meta) => {
-                await entity.write('manuSpecificNiko1', {'outletLedColor': ((value) ? 255 : 0)});
-                return {state: {led_state: ((value) ? true : false)}};
+                await entity.write('manuSpecificNiko1', {'outletLedColor': ((value.toLowerCase() === 'off') ? 0 : 255)});
+                return {state: {led_state: ((value.toLowerCase() === 'off') ? 'OFF' : 'ON')}};
             },
             convertGet: async (entity, key, meta) => {
                 await entity.read('manuSpecificNiko1', ['outletLedColor']);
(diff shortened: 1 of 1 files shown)


devices/niko.js
```diff
@@ -198,27 +232,31 @@ module.exports = [
...
             await endpoint.read('manuSpecificNiko1', ['switchOperationMode']);
+            await endpoint.read('manuSpecificNiko1', ['outletLedState']);
+            await endpoint.read('manuSpecificNiko1', ['outletLedColor']);
         },
         exposes: [
             e.switch(),
             e.action(['single', 'hold', 'release']),
             exposes.enum('operation_mode', ea.ALL, ['control_relay', 'decoupled']),
+            exposes.binary('led_enable', ea.ALL, true, false).withDescription('Enable LED'),
+            exposes.binary('led_state', ea.ALL, true, false).withDescription('LED State'),
```
Koenkk, line 250: Instead of a boolean, `ON`/`OFF` fits better here.

devices/niko.js
```diff
@@ -232,13 +270,21 @@ module.exports = [
             await reporting.onOff(ep2);
             await ep1.read('manuSpecificNiko1', ['switchOperationMode']);
             await ep2.read('manuSpecificNiko1', ['switchOperationMode']);
+            await ep1.read('manuSpecificNiko1', ['outletLedState']);
```
sjorge, line 273: You can probably read switchOperationMode, outletLedState, and outletLedColor in one read.
While messing around with initial support for these I managed to read 5+ attributes at a time. So something like:

```
await ep1.read('manuSpecificNiko1', ['switchOperationMode', 'outletLedState', 'outletLedColor']);
await ep2.read('manuSpecificNiko1', ['switchOperationMode', 'outletLedState', 'outletLedColor']);
```

devices/niko.js
sjorge: led_enable is already a boolean on the outlets, but swapping led_State to ON/OFF might make sense given most 'state' properties use ON/OFF (or LOCKED/UNLOCKED, ...)

devices/niko.js
sjorge: I am getting timeouts reading outletLedState and outletLedColor on the single switch one, but I had one of the very first units so it might have old firmware, it certainly has a ton of other bugs where I am able to completely brick the device.

Given I am not getting UNSUPPORTED_ATTRIBUTE I think it probably crashed the firmware on mine again 😞

FHannes: >Curious to what firmware version yours has though, mine is 313-687-00 (20210224)
That would be the same for all of my zigbee switches.

I did have some issues after trying to read some of the standard device properties, where it would just timeout. But once I physically pressed the switch, everything returned to normal.

FHannes:
Makes sense, I'll make the adjustments. I don't have that much experience yet writing these converters, so I appreciate the input.

devices/niko.js
sjorge: I’d personally leave led_enable as a boolean as we have other _enable’s that are boolean.

So only led_state should be ON/OFF I think, although consistency on this generally not great across z2m.

devices/niko.js
sjorge: Firmware on these is definitely not super stable. 😔

FHannes:
I changed the enabled property to an ON/OFF value for both switches and the outlet. I did the same for the LED state property of the switches.

I also added a read on the LED status property after setting the enabled state. From what I've observed, the switch will automatically turn off the LED when you disable the LED, as it should. But when the LED was previously on, the `outletLedColor` attribute will still return the value 255 (ON) when requested, until you re-enable the LED, after which it will return the correct value of 0 (OFF). This means that the converter does not accurate represent the state of the physical device, but it does accurately represent the state reported by the device through Zigbee. This is probably a bug in the firmware, but I am not sure what the proper course of action is here for the implementation of the converter. Should this be changed in the converter to correctly mimic the physical behavior of the switch?

devices/niko.js
FHannes: I did a reset on one of my switches and connected it to my Niko hub for a couple of hours, but the firmware was not updated, so I assume it's probably still the latest version.

devices/niko.js
FHannes: Makes sense. I personally also think a boolean for led_enable makes more sense. I will wait for @Koenkk to weigh in before reverting the change to led_enable. There do seem to be quite a few different approaches across z2m to handle this.

FHannes:
I have reverted the led_enable properties to booleans.

Koenkk:
@FHannes what is the state of this pr, ready for review?

FHannes:
Yes, I believe it is ready.

sjorge:
LGTM

PR #5031, opened 2022-11-27
Sinope TH1400ZB converter enhancement

faucherc:
I have add and adjust all the setting available for the Sinope TH1400ZB. The final result fits the Sinope app and/or documentation.

PR #5062, opened 2022-12-02
Clean lookups for sinope_thermostat

MattL0:
I got weird behaviors if this is not formated like this. This mainly change the lookup constants.  I think this should have been in the release ( or add the 5031pr below)

The code has been taken from https://github.com/Koenkk/zigbee-herdsman-converters/pull/5031, I forgot to add it my pr


Koenkk:
Thanks!

PR #5066, opened 2022-12-02
Adding Support for TS0601 Thermostat from vendor Avatto

stevenbuehner:
Adding Support for AVATTO TS0601 by adding vendor-id
(product i.e. here: https://de.aliexpress.com/item/1005003952150617.html)


clausinwi:
I'm afraid that the model doesn't fit. 
I have several of these devices. Using this version of the converter gives me the following error when setting the target temperature at the device:
`Datapoint 4 not defined for '_TZE200_bvu2wnxz' with data {"dp":4,"datatype":2,"data":{"type":"Buffer","data":[0,0,0,100]}}`
dp 4 ist setting the target temperature.
I have found the module m167.js in [https://github.com/twhittock/avatto_me167](url) to work with this device, just add it as an external converter. 
Is it possible to integrate avatto_me16 in tuya.js?

Koenkk:
@twhittock would you mind creating a PR?

PR #5069, opened 2022-12-02
Workaround to Address Axios Issue

nathanfiscus:
This addresses an issue in Axios 1.2.0 that causes data to be unreadable. See https://github.com/axios/axios/issues/5328

### DIFF
--- a/lib/ota/common.js
+++ b/lib/ota/common.js
@@ -458,8 +458,8 @@ function getAxios() {
         config = {
             proxy: false,
             httpsAgent: new HttpsProxyAgent(proxy),
-            headers:{
-                "Accept-Encoding": "*",
+            headers: {
+                'Accept-Encoding': '*',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

lib/ota/common.js
```diff
@@ -458,6 +458,9 @@ function getAxios() {
         config = {
             proxy: false,
             httpsAgent: new HttpsProxyAgent(proxy),
+            headers: {
+                'Accept-Encoding': '*',
```
Daniel15, line 462: This says that **every** encoding is supported, which may not actually be true. It's better to listed the supported encodings along with quality values (e.g. to prefer Brotli over deflate and gzip), for example `'Accept-Encoding': 'br, gzip;q=0.9, deflate;q=0.9`

PR #5071, opened 2022-12-03
Fix 'NHPB/SWITCH/1' endpoint number

ferrix:
I have this device, it autodetected correctly, but the state reporting never worked.

The device does not appear to have an endpoint 10 at all, but changing to 21 fixes it.


Koenkk:
Thanks!

PR #5084, opened 2022-12-04
Updating Hue Lucca Wall Light

blubaru14:
There are two models of this device:
* White
* White and Color Ambiance


Koenkk:
Thanks!

PR #5085, opened 2022-12-04
Add SETTI+ TRV001 thermostatic radiator valve

alufers:
https://setti.pl/produkty/klimat/grzejniki/glowica-termostatyczna-setti-trv001

This appears to be another white label of the Saswell SEA801. I have checked it and setting the temperature, and various settings works. Setting schedules, and reading historical data seems to be broken for all the supported devices in this converter, so I am not touching that.

Here are the Tuya datapoints I ahve extracted from the Tuya IoT platform: https://gist.github.com/alufers/05a15106d71b78af35dd90c0989dbac9

In the platform the device shows up as:

Product Name
Thermostat Setti+ TRV001
Device ID
bfb7cda7b62d0fbea7tlds
Product Category
wk


Koenkk:
Thanks!

PR #5088, opened 2022-12-05
IM6001-WLP01 battery is highly inaccurate

sjorge:
My IM6001-WLP01 was reporting switching between 20% and 72%, it looks like the device would very occasionally send an attributeReport for batteryPercentageRemaining would frequently send batteryVoltage. The inaccuracy seems to come from the voltage -> % mapping. 

To figure out if we needed to divide the value or not, I put in a fresh CR2 battery (I got because it was reporting 20%) and it jumped back to 100% with a value of 200 being reported. Switching back to the original battery made it report ~70% again with a value of around 140-144. So we can just do the standard mapping.


Koenkk:
Thanks!

PR #5090, opened 2022-12-05
setpoint TH1300ZB are 5-36 according to sinope

MattL0:
https://support.sinopetech.com/en/1.1.5.5/

Setpoint range: 5 °C to 36 °C (41 °F to 97 °F)


Koenkk:
Thanks!

PR #5022, opened 2022-11-26
Bug fix for all known bugs TV02 which were installed in 0.28.1.

vladi1234:

PR #5096, opened 2022-12-06
Bug fix for known bugs TV02 which were installed in 0.28.1.

vladi1234:
>    If I switch frost_protection to on, then 2 sec. later 'on' is no longer active.

I now know why it didn't work.

[107, null, valueConverter.TV02SystemMode],

What does 'null' mean in this line?

Why is it necessary to insert this line again, that all this works correctly?

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -2141,7 +2141,7 @@ module.exports = [
                 'degrees and from 04:00 to 00:00 temperature will be 22 degrees.'),
             ...tuya.exposes.scheduleAllDays(ea.STATE, 'HH:MM/C'),
             exposes.binary('online', ea.STATE_SET, 'ON', 'OFF').withDescription('The current data request from the device.'),
-            exposes.numeric('error_status', ea.STATE).withDescription('Error status'),
+            tuya.exposes.errorStatus(),
         ],
         meta: {
             tuyaDatapoints: [
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -2107,24 +2107,18 @@ module.exports = [
         exposes: [
             e.battery_low(), e.child_lock(), e.open_window(), e.open_window_temperature().withValueMin(5).withValueMax(30),
             e.comfort_temperature().withValueMin(5).withValueMax(30), e.eco_temperature().withValueMin(5).withValueMax(30),
-            tuya.exposes.errorStatus(), tuya.exposes.frostProtection('The device display "AF", press the pair button to cancel.'),
-            exposes.climate()
-                .withSystemMode(['off', 'heat'], ea.STATE_SET, 'When switched to the "off" mode, the device will display ' +
-                    '"HS" and the valve will be fully closed. Press the pair button to cancel or switch back to "heat" mode. Battery life' +
-                    ' can be prolonged by switching the heating off. After switching to `heat` mode, `preset` will be reset to `auto` and' +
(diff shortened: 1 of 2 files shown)


Koenkk:
> What does 'null' mean in this line?
Why is it necessary to insert this line again, that all this works correctly?

Is this solved, does it fully work now?

vladi1234:
Yes, `auto` problem is fixed. It works correctly now.

devices/tuya.js
```diff
@@ -2146,15 +2140,15 @@ module.exports = [
                 'period and until the next period, e.g., `04:00/20 24:00/22` means that from 00:00 to 04:00 temperature will be 20 ' +
                 'degrees and from 04:00 to 00:00 temperature will be 22 degrees.'),
             ...tuya.exposes.scheduleAllDays(ea.STATE, 'HH:MM/C'),
-            exposes.binary('online', ea.STATE_SET, 'ON', 'OFF').withDescription('Turn on this property to poll current data from the ' +
-                'device. It can be used to periodically fetch a new local temperature since the device doesn\'t update itself. ' +
-                'Setting this property doesn\'t turn on the display.'),
+            exposes.binary('online', ea.STATE_SET, 'ON', 'OFF').withDescription('The current data request from the device.'),
```
Koenkk, line 2143: The old description was more explanatory, can we keep this?

Koenkk:
@vladyspavlov I know I decided else in https://github.com/Koenkk/zigbee-herdsman-converters/pull/4750 but many people are reporting issues with the current integration so I think it would be good to reconsider this decision. (e.g. https://github.com/Koenkk/zigbee2mqtt/issues/15164). Can you take a look at this?

devices/tuya.js
vladi1234: tuya.exposes.errorStatus(),

This was already long ago in second correction tuya.js changed.

vladi1234:
> @vladyspavlov I know I decided else in #4750 but many people are reporting issues with the current integration so I think it would be good to reconsider this decision. (e.g. [Koenkk/zigbee2mqtt#15164](https://github.com/Koenkk/zigbee2mqtt/issues/15164)). Can you take a look at this?

That doesn't solve the problem.

vladi1234:
https://github.com/Koenkk/zigbee2mqtt/issues/11117

vladi1234:
Yes, understand that will be switched to new Tuya format.

I tried to warn that please do not break all functions.

I and all other users are very upset, because nevertheless "vladyspavlov" completely destroyed TV02 with its logic and thoughts.

Then should I help to solve problems and prove that I am right?

This is slowly demotivating.

Koenkk:
Thanks!

vladi1234:
Thanks!

Rixxxi:
Hi @Koenkk and @vladi1234 

First of all: Thank you for your hard work!
I'm trying to install TV02 thermostats and run into issues. After reading through pages of comments ( #5022 and [#15164](https://github.com/Koenkk/zigbee2mqtt/issues/15164) and others) I finally landed here.

Obvisously you commited a fix, but I'm a bit of a noop and don't know how I can download/install/apply (maybe even compile?) it. Can you please advise?
(I already browsed the [ba2e24f page](https://github.com/Koenkk/zigbee-herdsman-converters/commit/ba2e24f240a83f037d7bccc8b6efe3461d484237), but I couldn't download it, neather find the tuya.js files in my HS file system).

Thanks a lot it advance!

Issue #5068, opened 2022-12-02
Please add support occupancy/unoccupancy on Sinope TH1123/4ZB-G2 thermostats


rigorm:
Like the first Gen as now thermostat_occupancy It would be great if it was also added to the G2 (2nd gen) thermostats.

MattL0:
Do not update the dev branch yet. And try this one. I wwould like to know if thermostatUnoccupiedHeatingSetpoint could be reported by the G2, but not convinced :P. 

What i do here is poll the data every 10 minutes from mqtt. 

```js
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const fz = {...require('zigbee-herdsman-converters/converters/fromZigbee'), legacy: require('zigbee-herdsman-converters/lib/legacy').fromZigbee};
const tz = require('zigbee-herdsman-converters/converters/toZigbee');
const constants = require('zigbee-herdsman-converters/lib/constants');
const reporting = require('zigbee-herdsman-converters/lib/reporting');
const extend = require('zigbee-herdsman-converters/lib/extend');
const e = exposes.presets;
const ea = exposes.access;

module.exports = [
    {
        zigbeeModel: ['TH1123ZB-G2'],
        model: 'TH1123ZB-G2',
        vendor: 'Sinopé',
        description: 'Zigbee line volt thermostat',
        meta: {thermostat: {dontMapPIHeatingDemand: true}},
        fromZigbee: [fz.legacy.sinope_thermostat_att_report, fz.legacy.hvac_user_interface, fz.electrical_measurement, fz.metering,
            fz.ignore_temperature_report, fz.legacy.sinope_thermostat_state, fz.sinope_thermostat],
        toZigbee: [tz.thermostat_local_temperature, tz.thermostat_occupied_heating_setpoint, tz.thermostat_temperature_display_mode,
            tz.thermostat_keypad_lockout, tz.thermostat_system_mode, tz.thermostat_running_state, tz.sinope_thermostat_backlight_autodim_param, 
            tz.sinope_thermostat_time, tz.sinope_thermostat_enable_outdoor_temperature, tz.sinope_thermostat_outdoor_temperature,
            tz.thermostat_pi_heating_demand, tz.acvoltage, tz.sinope_time_format],
        exposes: [
            exposes.climate()
                .withSetpoint('occupied_heating_setpoint', 5, 30, 0.5)
                .withSetpoint('unoccupied_heating_setpoint', 5, 30, 0.5)
[...]

PR #5102, opened 2022-12-07
Fix TH1123-G2

MattL0:
This would solve https://github.com/Koenkk/zigbee-herdsman-converters/issues/5068

- Fix thermostat_occupancy, backlight_auto_dim , main_cycle_output. 

Add a method to get time when pairing at the begining.

- remove some try/catch since this is pretty new thermostat and all is available.

- add a try/catch at the end for running state ( this is not really needed anyways , because when piheahtingdemand is above 10 this change the running state to heat.  This could failed because the thermostat does not have enought memory to store more reports.


Koenkk:
Thanks!

PR #5105, opened 2022-12-07
tuya.js : new manufacturer device TS0210 fkxmyics known as vibration detector

phmarteau:
support new manufacturer device TS0210 fkxmyics known as vibrator detector


Koenkk:
Thanks!

PR #5107, opened 2022-12-08
Update ctm.js with some last-minute changes

soversk:
MBD-S: Remove device_enabled, add relay_state.
AX Valve Controller: Move ssIasZone to endpoint 2.
mStikk Outlet: Add extra Modell ID’s.

### DIFF
--- a/devices/ctm.js
+++ b/devices/ctm.js
@@ -72,6 +72,19 @@ const fzLocal = {
             return result;
         },
     },
+    ctm_relay_state: {
+        cluster: 'genOnOff',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const result = {};
+            const data = msg.data;
+            if (data.hasOwnProperty(0x5001)) {
+                result.relay_state = data[0x5001] ? 'ON' : 'OFF';
+            }
+
+            return result;
+        },
+    },
     ctm_temperature_offset: {
         cluster: 'msTemperatureMeasurement',
         type: ['attributeReport', 'readResponse'],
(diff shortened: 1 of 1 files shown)


devices/ctm.js
```diff
@@ -936,17 +961,17 @@ module.exports = [
...
+                attribute: {ID: 0x5001, type: dataType.boolean},
+                minimumReportInterval: 1,
                 maximumReportInterval: constants.repInterval.HOUR,
-                reportableChange: null}]);
+                reportableChange: 0}], {manufacturerCode: 0x1337});
         },
         exposes: [e.switch(), e.illuminance(), e.illuminance_lux(), e.occupancy(),
-            exposes.binary('device_enabled', ea.ALL, 'ON', 'OFF')
-                .withDescription('Turn the device on or off'),
+            exposes.binary('relay_state', ea.ALL, 'ON', 'OFF')
```
Koenkk, line 973: What's the difference between `state` (from `e.switch()`) and `relay_state`?

soversk: e.switch() now does what “device_enabled” did in the previous implementation. When off the product Is disabled. Meaning, when you walk past the motion sensor, it will not switch on the relay.
“relay_state” just switches the relay, and shows the current state of the relay.

Koenkk: Isn't `device_enabled` a better name then?

soversk: I agree. But the functionality got moved from the custom attribute it was on during development to the standard on/off cluster (last-minute change).
Is there an easy way to rename e.switch() without writing custom fz/tz?

Koenkk: > But the functionality got moved from the custom attribute it was on during development to the standard on/off cluster (last-minute change).

I don't see why this influences the attribute name. Renaming `relay_state` to `device_enabled` should be possible.


> Is there an easy way to rename e.switch() without writing custom fz/tz?

Can you explain this more? I don't know what you are trying to do.

soversk: What we did was:
move `device_enabled` from 0x2201 to 0x0000
and move `relay_state` from 0x0000 to 0x5001
So `device_enabled` is now controlled by `fz.on_off` and `tz.on_off`.
Renaming `relay_state` to `device_enabled` seems wrong.
Sorry for the mess, I added support for thit product too early.

Koenkk: I see, it's better to clarify it in the description: `Turn the relay on or off` -> `Permanently turns the relay off`

soversk: But it doesn’t turn the relay permanently off. That is what `device_enabeled/e.switch()` does.
`relay_state` just turns the relay off until next time the motions sensor Is triggered.

Koenkk: But this device also provides a `state` (through `fz.on_off`), whats the difference between `state` and `relay_state`?

Koenkk: So IMHO `state` should become `device_enabled` and `relay_state` should become `state`. This aligns with how every other z2m supported relay works.

Koenkk:
Thanks!

PR #5108, opened 2022-12-08
Update ubisys.js

Danit2:
Update Device D1:
- Change the `level_config` because the device do not support `onTransitionTime` and `offTransitionTime` it only support `onOffTransitionTime` and add `onLevel` and `ExecuteIfOff`
- Add the actions of the inputs. For this i add new converters to force multiendpoint without the meta option. This is because the device have one light endpoint and two inputs endpoint. When i add the `{multiEndpoint: true}`, the light endpoint get a prefix and this is not so nice. At the moment only the `move` and `toggle` command to test. the others i will add later. @Koenkk what do you like more: Add the new converters to the ubisys file or in the fromZigbee file? Perhaps someone can use the converter also for a other device? because there are not `manufacturerSpecific` 
- Add a better description to the `minimum_on_level`

Update Device S2:
-  Add the same energy reporting as @heinzma has add to the D1 device

Update Device J1:
- Add the same energy reporting as @heinzma has add to the D1 device

I add the energy reporting only on the device that i have. but i think all ubisys device will support this. 
should I add this to all devices?


sjorge:
Should work on an S1 too:

Screenshot 2022-12-08 at 09 06 17
Screenshot 2022-12-08 at 09 05 50

devices/ubisys.js
```diff
@@ -750,6 +843,12 @@ module.exports = [
             return {'default': 1, 'meter': 3};
         },
         onEvent: async (type, data, device) => {
+            if (data.type === 'attributeReport' && data.cluster === 'seMetering') {
```
sjorge, line 846: You can probably change this into a function, from a quick glance can grab the correct endpoint from the data object so no need to hardcode 3 and 5

sjorge:
Not sure how accurate those readings are, the S1 that is behind my fridge only reports a value < 0.5

devices/ubisys.js
Danit2: yes that would be a possibility. Unfortunately, I'm not that good at writing functions. and the endpoint is already hardcoded everywhere on all devices anyway.
```
        endpoint: (device) => {
            return {'default': 1, 's1': 2, 's2': 3, 'meter': 4};
```

Danit2:
> Not sure how accurate those readings are, the S1 that is behind my fridge only reports a value < 0.5

I don't know. But normaly the Ubisys devices works very accurate. But i have no device to verify this. All my dievices are for light. So there are difficult to say that.

Danit2:
Thank you for your quick write but unfortunately it does not work.
it gives an synax error because of `await`

sjorge:
> Thank you for your quick write but unfortunately it does not work. it gives an synax error because of `await`

I've update the patch above, the function probably needs to be async I think. (I always struggle with this too)

heinzma:
Tested the update including the function. Works fine with my D1. 

Only `execute_if_off` does not work for me. The light turns on even when I only set the brightness.

Danit2:
> Tested the update including the function. Works fine with my D1. 
> 
> 
> 
> Only `execute_if_off` does not work for me. The light turns on even when I only set the brightness.

Yes i konw. It is not working when you send the command over the ui. 
But when you send a command over mqtt loke this: 

{"brightness":15, state:null}

It works. 

I think this is a issue from the fronted.

heinzma:
Thanks, for the explanation. I tried mqtt but only `{"brightness":15, state:"OFF"}` and `{"brightness":15}`.

`{"brightness":15, state:null}` did the trick. 

But now I can't see a difference, it works even when `execute_if_off` is false.
Do I miss something? Again...

Danit2:
@heinzma for me it is working. 
You must have the onLevel on previos. 
Then you can send the command and in the ui it changes the brightness but when you turn on the light it comes the old brightness. 
When execute_if_off is on, then you can send the command and the light will turn on with the new brightness.

Danit2:
@sjorge your funcion is working. Thanks. I will add them to the file snd to all devices with a meter endpoint. Thanks a lot.

heinzma:
> @heinzma for me it is working. You must have the onLevel on previos. Then you can send the command and in the ui it changes the brightness but when you turn on the light it comes the old brightness. When execute_if_off is on, then you can send the command and the light will turn on with the new brightness.

It is working for me! 

What I tried to say is that to expose `execute_if_off` seem unnecessary to me because it seems to work even when `execute_if_off` is set to false. But I don't mind if it's there.

Danit2:
For me it is not working when it is off. 
When it is off it is like that: 
- Turn on
- Light go to previus value (200)
- Turn off
- Send brightness 100
- Turn on
- Light go to previus value (200)

When it is on it is like that:
- Turn on
- Light go to previus value (200)
- Turn off
- Send brightness 100
- Turn on
- Light go to new value (100)

heinzma:
That's exactly how I tested it. 
The logs confirm `"on_level":"previous"` and `"execute_if_off":false`. 
Nevertheless the previous brightness gets overridden, with the brightness I send beforehand, when light is turned on.

But when it's working for you then it's probably a problem on my side. I will have a look at it when I have time.

Thanks

Danit2:
I have found a better solution for the actions exposes. 
@Koenkk can you merge the two converters `brightness_force_single_endpoint` and `on_off_force_single_endpoint` from the PR #5097 ? 

Then i can use this connverters  in the D1 device. 
It is mutch easyer to use only this two converters and not to make 6 other new converters. 
Or  shoud i do a new PR with the two new converters?

Koenkk:
> Or shoud i do a new PR with the two new converters?

Please include them in this PR. I'm not sure if https://github.com/Koenkk/zigbee-herdsman-converters/pull/5097 will be merged in its current state.

Koenkk:
I've created https://github.com/Koenkk/zigbee-herdsman-converters/pull/5133 to replace https://github.com/Koenkk/zigbee-herdsman-converters/pull/5130, can you try it by applying the changes and add to the definition `meta: {multiEndpoint: true, multiEndpointSkip: ['state', 'brightness']}`?

Danit2:
Ohh, thats is a good alternative and mutch easyer.

Can you please merge your PR #5133 then i can test it with the dev version in my external converter. Beqause i don't have access to the file. I use the HA AddOn.

Koenkk:
thanks!

PR #5109, opened 2022-12-08
philips: decode more of the gradient payload

zegl:
Also fixes the encoding of gradients with less than 5 colors, and added more tests.

`gradient_extras` is not used for anything more than testing at the moment. The `offset` and `segments` values _does_ affect the light effect on the device, and have been added to both the encoder and the decoder. This does not include a public API to get/set those values, it might come in another PR if we can find a good design for it.


zegl:
Hehe this device is very fun to work with.

Koenkk:
Thanks!

PR #5115, opened 2022-12-09
Added additional manufacturerName and DP for Frankever valve

BradleyFord:


BradleyFord:
When I add this code to my local system everything works perfectly, when I set the Threshold the position feedback is perfect.
I am assuming in earlier models of the Frankever (that have the same model code), they used DP101. Whilst in the new models they use DP102...

How should I handle this the checks do not allow duplicates?
(error  Duplicate key 'frankEverTreshold'  no-dupe-keys)

Koenkk:
Fixed the error, updated your code. Can you check if everything still works?

PR #5116, opened 2022-12-09
Add support for a variant of the Tuya tubular motor

zsqduke:
{modelID: 'TS0601', manufacturerName: '_TZE200_udank5zs'}
IMG_1716
I own these and they are the same as the Zemismart Tubular motor that I also own (_TZE200_fzo2pocs)


Koenkk:
Thanks!

PR #5119, opened 2022-12-10
disablePowerOnBehavior:true for https://www.moeshouse.com/collections/smart-light/products/zb-td5-rcw-gu10-ms

Hadatko:
Signed-off-by: Dusan Cervenka <cervenka.dusan@gmail.com>

### DIFF
--- a/devices/moes.js
+++ b/devices/moes.js
@@ -331,35 +331,6 @@ module.exports = [
         },
     },
     {
-        // Since a lot of TuYa devices use the same modelID, but use different datapoints
-        // it's necessary to provide a fingerprint instead of a zigbeeModel
-        fingerprint: [
-            {
-                // The model ID from: Device with modelID 'TS0601' is not supported
-                // You may need to add \u0000 at the end of the name in some cases
-                modelID: 'TS0505B',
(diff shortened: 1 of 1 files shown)


Hadatko:
Hi, as mentioned here https://github.com/Koenkk/zigbee2mqtt/issues/15451 i have different device than is already supported. It looks like it is almost same, but i had to turn off `disablePowerOnBehavior:true` option as it was throwing error with unsupported option. If i remove my config file, then device is recognized like  https://www.zigbee2mqtt.io/devices/ZLD-RCW.html but i have https://www.moeshouse.com/collections/smart-light/products/zb-td5-rcw-gu10-ms

How can i distuingish between these two devices? I am curious if i can remove existing support somehow from code to maybe get more reliable information about product (something specific).

Hadatko:
I see previously it was edited exiting model: https://github.com/Koenkk/zigbee2mqtt/issues/12919
Should i add disable attribute to existing configuration and remove new one?

Koenkk:
Please edit the existing definition.

Hadatko:
@Koenkk Done.

Koenkk:
Thanks!

Hadatko:
Thank you too ;)

PR #5124, opened 2022-12-10
ikea: Enable turnsOffAtBrightness1 for LED1925G6

Electry:
This is the default behavior in the latest firmware

Signed-off-by: Michal Chvíla <michal@chvila.sk>


Koenkk:
Thanks!

PR #5132, opened 2022-12-12
lib/philips: parse xy and temperature modes

zegl:
The devices exposes color_xy and color_temp information on the same endpoint that has the gradient data.

This fixes an error where gradient data would be attempted to be decoded (null pointer exception) when the device is in a different color mode.

The extra decoded information is not used anywhere yet.


Koenkk:
Thanks!

PR #5134, opened 2022-12-12
align color_mode values for hue gradient lights

sjorge:
lib/color sets color_mode to 'xy', 'hs', or 'color_temp'.

This change aligns the newer gradient lights to also use 'xy' instead of 'color_xy' as color_mode value introduce in #5132


Koenkk:
Thanks!

PR #5139, opened 2022-12-13
Added Aqara SJCGQ13LM

avdept:
This PR adds support for  Aqara Air Monitoring Panel S1. It seems like pm25 and co2 values are returned in different units now, so I had to add separate convertors.

Attached screenshot from z2m device expose tab
Screenshot 2022-12-13 at 20 09 49

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -58,6 +58,23 @@ const daysLookup = {
 
 
 const fzLocal = {
+    aqara_s1_co2: {
+        cluster: 'msCO2',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

kiselio:
Hello @avdept, does the device have a dedicated CO2 sensor? Are the readings correct?

avdept:
> Hello @avdept, does the device have a dedicated CO2 sensor? Are the readings correct?

Yes it has dedicated co2, i think it's senseair s8. Readings are more or less correct, never had lower than 450, and reacts quickly when directly breathing out on the panel. Although I've noticed weird spikes over night when it drops from 800 down to 500 for like 20 mins and then goes up to previous number

kiselio:
Thank you for your answer. Maybe some calibrating is needed, if it's a Senseair s8.

PR #5144, opened 2022-12-13
Exposed apparent power for Legrand outlet (067775/741811)

Mephistofeles:
I've noticed that Legrand's connected outlet reports apparent power, but it's not exposed in zigbee2mqtt.
For example: "Received Zigbee message from 'living_room_tv_cabinet_socket', type 'attributeReport', cluster 'haElectricalMeasurement', data '{"activePower":60,"apparentPower":140}' from endpoint 1 with groupID 0"


Koenkk:
Thanks!

PR #5145, opened 2022-12-13
Add support for Aqara C2 curtain motor

deiger:


deiger:
[Product details (only in Chinese)](https://www.aqara.com/cn/curtain_controller_c2_overview).
Seems to be originally intended for the Chinese market, but is very common now on AliExpress. Spec and Zigbee reporting is similar to B1, albeit without the battery.

Koenkk:
Thanks!

JoooostB:
@deiger Do you think it would be an easy change to add `reset_limits` functionality as requested in [this issue](https://github.com/Koenkk/zigbee2mqtt/issues/17206)?

PR #5149, opened 2022-12-14
Expose activePower in Vimar 14593

mattiazeni:
Added haElectricalMeasurement to Vimar 14593 to monitor instant power (activePower)

### DIFF
--- a/devices/vimar.js
+++ b/devices/vimar.js
@@ -10,7 +10,7 @@ const fzLocal = {
         cluster: 'haElectricalMeasurement',
         type: ['attributeReport', 'readResponse'],
         convert: async (model, msg, publish, options, meta) => {
-            let payload = {};
+            const payload = {};
             if (msg.data.hasOwnProperty('activePower')) {
                 payload.power = msg.data['activePower'];
             }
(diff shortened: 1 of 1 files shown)


devices/vimar.js
```diff
@@ -42,10 +56,12 @@ module.exports = [
         model: '14593',
         vendor: 'Vimar',
         description: '16A outlet IoT connected',
-        extend: extend.switch(),
+        fromZigbee: [fz.on_off, fz.ignore_basic_report, fzLocal.vimar_active_power],
```
Koenkk, line 59: Why not use `fz.electrical_measurement` instead of `fzLocal.vimar_active_power`?

mattiazeni: Because the value reported is wrong, it seems is 10x the real one. We discussed [here](https://github.com/Koenkk/zigbee2mqtt/issues/15482).

Koenkk: You can manually set the divisor by adding the following code to `configure`:

```js
            endpoint.saveClusterAttributeKeyValue('haElectricalMeasurement', {
                acPowerMultiplier: 1, acPowerDivisor: 10,
            });
```

mattiazeni: Ah nice. Let me give it a try. Let you know later.

mattiazeni: I installed a brand new version of HA and the standard config works correctly, without setting `acPowerMultiplier` and `acPowerDivisor`. I added a commit to change this.

Koenkk:
Thanks!

PR #5150, opened 2022-12-14
Added valve state and reporting

BBJake:
- Added pi_heating_demand (valve opening) as a sensor
- Added reporting for KeypadLockMode, Operating mode, pi_heating_demand and window_open
- Fixed some typos

### DIFF
--- a/devices/bosch.js
+++ b/devices/bosch.js
@@ -4,6 +4,7 @@ const fz = require('../converters/fromZigbee');
 const tz = require('../converters/toZigbee');
 const reporting = require('../lib/reporting');
 const utils = require('../lib/utils');
+const constants = require('../lib/constants');
 const e = exposes.presets;
 const ea = exposes.access;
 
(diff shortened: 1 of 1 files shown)


devices/bosch.js
```diff
@@ -139,9 +140,12 @@ const fzLocal = {
                 result.boost = (Object.keys(stateOffOn)[data[0x4043]]);
             }
             if (data.hasOwnProperty(0x4007)) {
-                const opModes = {0: 'auto', 1: 'heat', 2: 'unknowm 2', 3: 'unknonw 3', 4: 'unknown 4', 5: 'off'};
+                const opModes = {0: 'auto', 1: 'heat', 2: 'unknown 2', 3: 'unknown 3', 4: 'unknown 4', 5: 'off'};
```
Koenkk, line 143: Can you make these snake case? e.g `unknown 2` -> `unknown_2`

BBJake: Sure. Changed it..

Nezz:
Do you know if there's a practical difference between the "auto" and "heat" operating mode, or is it just something we have for uniformity?

BBJake:
@Nezz 
The "heat" should better be called "manuell", but the SystemModes for thermostats just has this options: off/auto/cool/heat.... The Bosch thermostat supports "off/manual/auto". If set to auto, it uses the information (weekday/time) that you can set over the Bosch-App.
At the moment it is not possible to set these value over the z2m-interface, so it just makes sense to use the auto (manual) mode.

Nezz:
Thanks for the clarification. Did you mean this?
>so it just makes sense to use the ~~auto~~ heat (manual) mode

Koenkk:
Thanks!

BBJake:
> Thanks for the clarification. Did you mean this?
> 
> > so it just makes sense to use the ~auto~ heat (manual) mode

Yes. It only makes sense to use the **heat** mode for z2m-users. In the auto mode the thermostat would overwrite it with other values directly set in the thermostat.

PR #5152, opened 2022-12-14
ubisys: S1/S2 group energy with power in the expose data

sjorge:
This should also allow manual reading of the value like we can do with power.


sjorge:
I'm confused as currentsummdelivered should provide the needed support to read the attribute???
```node
...snip...
    currentsummdelivered: {
        key: ['energy'],
            await entity.read('seMetering', ['currentSummDelivered']);
        },
    },
...snip...
```

sjorge:
Interesting, only 1 of my 2 S2's will respond to reading the other one times out, as do all my S1's... even for reading power which was certainly working when I originally added it.

They are about 2 major firmware jumps higher now so maybe something changed.

sjorge:
I guess there are 2 issues:
1. the test fails, while it should not, as show above it is working.
2. bug in how the frontend deals with the expose data?

sjorge:
I figured out why the tests fails for energy but not for power!

```
                    if (expose.endpoint && expose.property.length > expose.endpoint.length) {
                        property = expose.property.slice(0, (expose.endpoint.length + 1) * -1);
                    }
```

We step into this block for energy but do not for power, which means property is '' for energy.

sjorge:
Tests now pass and the frontend now looks like:
image

image


I'll file an issue against the frontend.

devices/ubisys.js
```diff
@@ -536,15 +536,19 @@ module.exports = [
...
-        exposes: [e.switch(), e.power().withAccess(ea.STATE_GET).withEndpoint('meter').withProperty('power'),
+        exposes: [
+            e.switch(),
             e.action([
                 'toggle', 'on', 'off', 'recall_*',
                 'brightness_move_up', 'brightness_move_down', 'brightness_stop',
             ]),
-            e.power_on_behavior(), e.energy()],
+            e.power_on_behavior(),
+            e.power().withAccess(ea.STATE_GET).withEndpoint('meter').withProperty('power'),
```
Koenkk, line 546: You probably don't the `meter` endpoint postfix. This can be achieved now with https://github.com/Koenkk/zigbee-herdsman-converters/pull/5133.

e.g.
```
meta: {multiEndpointSkip: ['power', 'energy']}
```

sjorge:
Look interesting, i’ll have a closer look tomorrow or saturday, depending on how busy i’ll be.~ sjorgeOn 15 Dec 2022, at 20:20, Koen Kanters ***@***.***> wrote:﻿
@Koenkk commented on this pull request.



In devices/ubisys.js:
>              e.action([
                 'toggle', 'on', 'off', 'recall_*',
                 'brightness_move_up', 'brightness_move_down', 'brightness_stop',
             ]),
-            e.power_on_behavior(), e.energy()],
+            e.power_on_behavior(),
+            e.power().withAccess(ea.STATE_GET).withEndpoint('meter').withProperty('power'),

You probably don't the meter endpoint postfix. This can be achieved now with #5133.
e.g.
meta: {multiEndpointSkip: ['power', 'energy']}

devices/ubisys.js
sjorge: If i drop the withEndpoint the fronend still displays it fine (just not indented) but the get request still end up failing the same way as before, am i missing something obvious?

energy/power is already in the state without a suffix (it always has been) so  skip doesn’t seem to be doing anything new for them.

sjorge: OK this makes sense, as `postfixWithEndpointName` is the function that  consumes multiEndpointSkip and neither the power or energy converter uses `postfixWithEndpointName`.

So I'm guessing the current PR is fine! It's just the frontend that needs to properly use the expose information.

Koenkk: `fz.metering` does indeed not use `postfixWithEndpointName`. But if I understand correctly now this is only for the reading part. Ideally just `e.power().withAccess(ea.STATE_GET)` should be used. Can you try if in `endpoint()`  the following works?

```js
return {'l1': 1, 's1': 2, 'default': 3};
```

sjorge:
```[hyperon :: sjorge][~/Documents/repos/zigbee-herdsman-converters][patch-8 ✗]
[■]$ git diff
diff --git a/devices/ubisys.js b/devices/ubisys.js
index 1c60a6cf..23dca0a8 100644
--- a/devices/ubisys.js
+++ b/devices/ubisys.js
@@ -543,14 +543,14 @@ module.exports = [
                 'brightness_move_up', 'brightness_move_down', 'brightness_stop',
             ]),
             e.power_on_behavior(),
-            e.power().withAccess(ea.STATE_GET).withEndpoint('meter').withProperty('power'),
-            e.energy().withAccess(ea.STATE_GET).withEndpoint('meter').withProperty('energy'),
+            e.power().withAccess(ea.STATE_GET).withProperty('power'),
+            e.energy().withAccess(ea.STATE_GET).withProperty('energy'),
         ],
         fromZigbee: [fz.on_off, fz.metering, fz.command_toggle, fz.command_on, fz.command_off, fz.command_recall, fz.command_move,
             fz.command_stop, fz.power_on_behavior, ubisys.fz.configure_device_setup],
         toZigbee: [tz.on_off, tz.metering_power, tz.currentsummdelivered, ubisys.tz.configure_device_setup, tz.power_on_behavior],
         endpoint: (device) => {
-            return {'l1': 1, 's1': 2, 'meter': 3};
+            return {'l1': 1, 's1': 2, default: 3};
         },
             const endpoint = device.getEndpoint(3);
@@ -636,14 +636,14 @@ module.exports = [
                 'brightness_stop_s2']),
             e.power_on_behavior().withEndpoint('l1'),
             e.power_on_behavior().withEndpoint('l2'),
-            e.power().withAccess(ea.STATE_GET).withEndpoint('meter').withProperty('power'),
-            e.energy().withAccess(ea.STATE_GET).withEndpoint('meter').withProperty('energy'),
+            e.power().withAccess(ea.STATE_GET).withProperty('power'),
+            e.energy().withAccess(ea.STATE_GET).withProperty('energy'),
         ],
         fromZigbee: [fz.on_off, fz.metering, fz.command_toggle, fz.command_on, fz.command_off, fz.command_recall, fz.command_move,
             fz.command_stop, fz.power_on_behavior, ubisys.fz.configure_device_setup],
         toZigbee: [tz.on_off, tz.metering_power, ubisys.tz.configure_device_setup, tz.power_on_behavior, tz.currentsummdelivered],
         endpoint: (device) => {
-            return {'l1': 1, 'l2': 2, 's1': 3, 's2': 4, 'meter': 5};
+            return {'l1': 1, 'l2': 2, 's1': 3, 's2': 4, default: 5};
         },
         meta: {multiEndpoint: true},
```

This causes all sorts of breakage now, I can now read power/energy but for the S1 state and power_on_behavior is broken.

With as the PR is now, I can manualy everythign works fine, except the frontend's read calls as it just ignores the withEndpoint(), I'm pretty sure this is a bug frontend side.

Koenkk: Since this a a special case, an alternative solution would be to create a separate `tzLocal`/`currentsummdelivered` for these devices which always read from the correct endpoint? (`meter` endpoint)

sjorge: The expose data contains everything to do the read properly as it contains the correct endpoint, It’s just the frontend not properly parsing it.

Koenkk: I see, but we should wonder wether an endpoint property without the endpoint name in it should be allowed. (that is also the reason why the tests are adapted in this pr). Ubisys devices are the only supported devices which use such construction: (like `.withEndpoint('meter').withProperty('power')`)

sjorge: I’ll see if i can update both convertGet to be `multiEndpointSkip` aware somehow. Let me thing about it a bit more.

sjorge:
@Koenkk I took another pass, it seems to be working now.
- dropped the withEndpoint() on the expose for S1/S2
- added `multiEndpointSkip` to the README.md
- Added `utils.getMultiEndpoint()` which will fetch the endpoint to use if multiEndpointSkip is set and if it's a hash instead of an array.

The toZigbee functions for power/energy now what endpoint to use.

Danit2:
@sjorge Your PR looks good. Can you add them also to the other ubisys devices? I think they must work on every device.

Koenkk:
Approach looks good to me.

- I think it is better to not mix with `multiEndpointSkip`, this looks more like a `multiEndpointForce`
- `e.power().withAccess(ea.STATE_GET).withProperty('power')` can become `e.power().withAccess(ea.STATE_GET)` now (since the property is `power` by default)

sjorge:
SGTM, I'll do another pass over this tonight or tomorrow, depending on when I have time.

sjorge:
Done, retested my S1/S2 and J1 and they still work and read the value.

```
debug 2022-12-19 21:30:55: Received Zigbee message from 'switch/office/light', type 'readResponse', cluster 'seMetering', data '{"instantaneousDemand":1}' from endpoint 3 with groupID 0
debug 2022-12-19 21:30:55: Received Zigbee message from 'switch/office/light', type 'readResponse', cluster 'seMetering', data '{"currentSummDelivered":[0,756]}' from endpoint 3 with groupID 0
debug 2022-12-19 21:31:05: Received Zigbee message from 'switch/livingroom/light_top', type 'readResponse', cluster 'seMetering', data '{"instantaneousDemand":1}' from endpoint 5 with groupID 0
debug 2022-12-19 21:31:06: Received Zigbee message from 'switch/livingroom/light_top', type 'readResponse', cluster 'seMetering', data '{"currentSummDelivered":[0,475]}' from endpoint 5 with groupID 0
debug 2022-12-19 21:31:17: Received Zigbee message from 'cover/office/window', type 'readResponse', cluster 'seMetering', data '{"currentSummDelivered":[0,59]}' from endpoint 3 with groupID 0
debug 2022-12-19 21:31:18: Received Zigbee message from 'cover/office/window', type 'readResponse', cluster 'seMetering', data '{"instantaneousDemand":0}' from endpoint 3 with groupID 0
```

And to check if the normal _l1 _l2 attribute still work on an S2 I read the power_on_behavior for both
```
debug 2022-12-19 21:32:04: Received Zigbee message from 'switch/livingroom/light_top', type 'readResponse', cluster 'genOnOff', data '{"startUpOnOff":0}' from endpoint 2 with groupID 0
debug 2022-12-19 21:32:04: Received Zigbee message from 'switch/livingroom/light_top', type 'readResponse', cluster 'genOnOff', data '{"startUpOnOff":255}' from endpoint 1 with groupID 0
```

Koenkk:
Thanks!

PR #5153, opened 2022-12-15
Support for a new Moes dimmer (Tuya TS0601 dimmer)

WickedDurkie:
Today I got a new Moes dimmer which looked exactly the same as another moes dimmer that I already have. But it was not supported. The manufacturer name was not found. This adds support for the new dimmer,


Koenkk:
Thanks!

PR #5093, opened 2022-12-06
update thirdreality plug

weihuan1111:

PR #5154, opened 2022-12-15
Revert "update thirdreality plug"

weihuan1111:
Reverts Koenkk/zigbee-herdsman-converters#5093


Koenkk:
This is not only a revert, I see `3RSP02028BZ` was added. Is this on purpose?

weihuan1111:
> This is not only a revert, I see `3RSP02028BZ` was added. Is this on purpose?

@Koenkk yes, because we have change our plugv2(with power function) model to 3RSP02028BZ, 
    and plugv1(without power function) model still be 3RSP019BZ

Koenkk:
Thanks!

PR #5158, opened 2022-12-15
Disable power on behavior for TS0504B

Hadatko:
Signed-off-by: Dusan Cervenka <cervenka.dusan@gmail.com>


Hadatko:
Not tested but my device has error with this option:
image

Koenkk:
I assume that none of the TuYa TS050XX devices support this, disabled it for more.

PR #5162, opened 2022-12-16
Add 915005987301 - Philips Hue Signe gradient table lamp (white)

jvmiert:
Add support for the Philips Hue Signe gradient table lamp (white).
https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-gradient-signe-table-lamp/046677803483


Koenkk:
Thanks!

PR #5171, opened 2022-12-19
Add wisdom.js for dmz250 dimmer

SjonHortensius:


devices/wisdom.js
```diff
@@ -0,0 +1,21 @@
...
+const tuya = require('zigbee-herdsman-converters/lib/tuya');
+const utils = require('zigbee-herdsman-converters/lib/utils');
+const globalStore = require('zigbee-herdsman-converters/lib/store');
+const e = exposes.presets;
+const ea = exposes.access;
+
+const definition = {
+    zigbeeModel: ['HK-DIM-SW'],
+    model: 'dmz250',
+    vendor: 'Wisdom',
```
Koenkk, line 16: Looks this device is from [Profulux](https://profolux.com/product/led-dimmer-zigbee-3-0-5-250w/). Can you move it to `profulux.js`?

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1769

devices/wisdom.js
SjonHortensius: @Koenkk is fine by me, the device identifies itself as Wisdom - so I actually think Profolux is simply a reseller?

SjonHortensius: also, did you notice there is no profOlux yet (not profUlux), there is only profAlux which is yet another manufacturer

Koenkk: Can you provide a link of where you bought the device?

SjonHortensius: that was through https://www.ledlampaanbiedingen.nl/zigbee-smart-wifi-led-dimmer-0-250-watt-werkt-via.html but it seems they are both white-labeling the same hardware. Again, the device identifies itself as `Wisdom` which is why I picked it

Koenkk: I see, then let's keep it as is.

PR #5172, opened 2022-12-19
add zigbeeModel variation for Philips Hue Iris Copper 4th gen

lbarnkow:
I bought two Hue Iris lights in december 2021 (rosé & copper), however, the `zigbeeModel` for my copper variant was not supported. Added it locally to test with Home Assistant and it works flawlessly.


Koenkk:
Thanks!

PR #5174, opened 2022-12-19
Update woolley.js

wifiuk:
adding SA-030-1


wifiuk:
the zigbee model might be the same but the model number is different, that's what allows me to add it
what are your suggestions here?

Koenkk:
You can remove the added definition and change the existing model to `SA-030-1`

PR #5183, opened 2022-12-21
Add BTicino LN4570CWI and fix capitalization

giu1io:
Hi,

I've bought a [BTicino LN4570CWI](https://catalogo.bticino.it/BTI-LN4570CWI-IT) hoping to add the support for it. However it's already supported and recognised as [Legrand 064873](https://www.zigbee2mqtt.io/devices/064873.html). I've added it as a white label device for the Legrand 064873, so if anyone is searching the documentation it will show up as supported.

As I was doing this I've noticed that another white label device in the same file (BTicino 3584C) had the brand name spelled with the wrong capitalization, which caused the brand to show up twice in the documentation.  I've also updated that to match all other BTicino devices.

Thanks!

### DIFF
--- a/devices/legrand.js
+++ b/devices/legrand.js
@@ -262,7 +262,7 @@ module.exports = [
         ota: ota.zigbeeOTA,
         fromZigbee: [fz.identify, fz.on_off],
         toZigbee: [tz.on_off, tz.legrand_identify],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #5184, opened 2022-12-21
Add SR-ZG9010A to sunricher.js

asgothian:
Add SR--ZG9010A

Device looks like this (couldn't find any images on the net, the device was provided for integration by led-trading.de

IMG_3189
IMG_3190


Koenkk:
Thanks!

PR #5191, opened 2022-12-23
Update visonic.js - Add temp support to MCT-350 SMA

alsobrsp:
The MCT-350 SMA supports temperature reporting and battery voltage. This PR is only for temperature.


Koenkk:
Thanks!

PR #5192, opened 2022-12-23
hue: add tzLocal.effect that support candle, fireplace, colorloop

sjorge:
This will fix koenkk/zigbee2mqtt#15699 

@Koenkk basics work, but I'm having trouble figuring out what to target with the expose stuff.

I tested the bulbs I have LCA008, LCG002, and LTG002 all seem to work!

But I'm not sure it's safe to just add fzLocal.effect to light_onoff_brightness_colortemp,  light_onoff_brightness_color, light_onoff_brightness_colortemp_color, and light_onoff_brightness_colortemp_color_gradient.

Also that still leaves the issue that the original tz.effect will be included. So some hints on how to best go about this are welcome. For my testing I just hardcoded the set of toZigbee for the bulbs I was testing with.


Koenkk:
`hueExtend` has to be updated to pass `disableEffect: true` as an option, this will prevent `e.effect()` and `tz.effect` from being added. 

`e.effect()` has to be updated to allow for more effects, something like:

```js
        effect: (additionalEffects=[]) => new Enum('effect', access.SET, ['blink', 'breathe', 'okay', 'channel_change', 'finish_effect', 'stop_effect', ... additionalEffects]).withDescription('Triggers an effect on the light (e.g. make light blink for a few seconds)'),
```

In `hueExtend` then add `e.effect(['candle', AND ALL THE OHTER])`

sjorge:
Based on https://github.com/Koenkk/zigbee2mqtt/issues/15699#issuecomment-1364438284 we should probably not mix this in with effect.

Perhaps a ‘hue_scene’ property instead, we can then also add a ‘disableHueScebe’ to the hueExposes and selectively enable this for lights we know work.

sjorge:
What change:
- rename tzLocal.effect -> tzLocal.hue_scene
- update exposes to take disableHueScenes option
- set disableHueScenes: false for LTG002, LCG002, LCA008 (bulbs I own and was able to test)
- move hue specific converters to devices/philips.js

I also noticed a unused manufacturerOptions entry that somehow ended in devices/ubisys.js so I removed that one.

sjorge:
Oops, looks like I need another pass on this, LTG002 only seems to support the 'candle' scene, colorloop obviously doesn't work as it only has CT support. I think fireplace is also not working.

I wont have time until tuesday or maybe monday evening to give this another pass.

sjorge:
Fixed

Koenkk:
> Based on [Koenkk/zigbee2mqtt#15699 (comment)](https://github.com/Koenkk/zigbee2mqtt/issues/15699#issuecomment-1364438284) we should probably not mix this in with effect.
> 
> Perhaps a ‘hue_scene’ property instead, we can then also add a ‘disableHueScebe’ to the hueExposes and selectively enable this for lights we know work.

Code looks good to me, but I don't see a good reason to not mix this in with effect? (I've checked the comment)

sjorge:
It’s more like a list of premade scenes hue preloads. We can merge the gradient_scene with this one too I think, but I don’t have any gradient capable devices to test with.

Unlike the other effects defined in ZCL these special hue scenes don’t really stop after a short while or when you turn off/on the light.

They remain in effect until you set color manually or a different pre-made scene (or cut the power to the light)

Koenkk:
I'm not sure if these things fall under the category of gradient, to me the following makes sense

- `gradient`/`gradient_scene`: this is for the Hue lights that support setting different colors for different segments of the light (multi color)
- `hue_scene`/`effect`: this is for Hue lights supporting the Hue specific effects (like the fireplace/candle). My opinion is that `effect` is the best name, this is also how they are being called in the app ([screenshot from app](https://www.reddit.com/r/Hue/comments/sor6kn/new_candle_and_fireplace_effects_in_the_hue_app/)). The duration of the effect, a few seconds for ZCL ones or endless like the Hue specific effects, doesn't matter for the naming IMO.

CC: @zegl

zegl:
Nice work! 👍

I agree with that these new effects should be separated from the predefined list of gradients. 

They do have some in common, but it starts to get complicated when the list of effects/gradients becomes device dependent. And as the gradient list API is deprecated, it makes less sense long term to mix them.

`effect` sounds like the best name to me, if that is what Philips calls it.

sjorge:
For now to keep it simple I've made the assumption that fireplace and colorloop only work on bulbs that support color, that seems to be the case from my sample size.

This makes the hueExtends a bit simpler. I only set `disableHueEffect: false` on the bulbs I have to test on. I don't know if it's safe to blanket enable this for all bulbs and lights.

devices/philips.js
```diff
@@ -59,72 +92,20 @@ const hueExtend = {
                 .withLengthMin(1)
                 .withLengthMax(9)
                 .withDescription('List of RGB HEX colors'),
+            exposes.enum('effect', ea.SET, [
```
Koenkk, line 95: I think effect is added twice here because `exposes: extendDontUse.light_onoff_brightness_colortemp_color({supportsHS: true, ...options}).exposes.concat([` is called without `disableEffect`

sjorge: Yeah looks like you're right. I've fixed it in the new commit.

Koenkk:
Awesome, thanks again! 😄

PR #5193, opened 2022-12-23
Sonoff SA-030-1

ArvydasM1:
Added Sonoff smart plug SA-030-1. This is also branded as Woolley and eWeLink, but mine shows up as Sonoff. The image has already been added for the other brands

### DIFF
--- a/devices/sonoff.js
+++ b/devices/sonoff.js
@@ -256,4 +256,16 @@ module.exports = [
             device.save();
         },
     },
+	{
+        zigbeeModel: ['SA-030-1'],
+        model: 'SA-030-1',
+        vendor: 'SONOFF',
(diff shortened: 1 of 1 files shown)


Koenkk:
I believe ewelink is the real manufacturer (this is what you find in data/database.db), others are just white label so this is not needed.

ArvydasM1:
From the user's perspective, the device shows up as unsupported. Perhaps it would be more user-friendly to add all white labels?

Koenkk:
It's indeed not supported in the latest z2m release but is in the dev branch and next release on 1 Jan (https://github.com/Koenkk/zigbee-herdsman-converters/blob/37da23b3898ade170638bea19e08c7d25a4925a0/devices/ewelink.js#L24)

PR #5200, opened 2022-12-24
Correct state names for C4 Yale module

ChipWolf:
The states are not reported back correctly due to the -ED suffixes


Koenkk:
Thanks!

PR #5202, opened 2022-12-25
Added TS0601 | _TZE200_3ylew7b4 and used iteration for tuya_cover in fromZigbee

willieee802:
While adding support for a tuya tubular motor TS0601 _TZE200_3ylew7b4, the tuya.firstDpValue function logs that a for loop shall be used to process addtional datapoints in one zigbee message which is the exact case for this model, otherwise only retreiving the first dpvalue returns no useful information.

Have tested this model as well as another tuya curtain model which does not need such for loop implementation, and they both work well.


Koenkk:
Thanks!

PR #5204, opened 2022-12-25
Added TS0601 Smart knob dimmer

tefracky:
Added the TS0601 Smart knob dimmer from TuYa, branded/sold by Moes. Maybe the whitelabel is wrong. The device can be found here:

https://moeshouse.com/collections/smart-dimmer-switch/products/new-wifi-smart-rotary-light-dimmer-switch-schedule-timer-brightness-memory-eu

Unfortunately, I am not sure how to use the 'indicatorMode' correctly, since I didnÄt found the right vonverter.

PR for zigbee2mqtt.io: https://github.com/Koenkk/zigbee2mqtt.io/pull/1781

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1351,7 +1351,8 @@ module.exports = [
         fromZigbee: [tuya.fz.datapoints],
         toZigbee: [tuya.tz.datapoints],
         configure: tuya.configureMagicPacket,
-        exposes: [tuya.exposes.lightBrightness().withMinBrightness().setAccess('min_brightness', ea.STATE_SET), tuya.exposes.lightType(), /*tuya.exposes.indicatorModeNoneRelayPos()*/],
+        exposes: [tuya.exposes.lightBrightness().withMinBrightness().setAccess('min_brightness', ea.STATE_SET), tuya.exposes.lightType()
+		    /* tuya.exposes.indicatorModeNoneRelayPos() */],
         meta: {
             tuyaDatapoints: [
                 [1, 'state', tuya.valueConverter.onOff, {skip: tuya.skip.stateOnAndBrightnessPresent}],
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1344,6 +1344,29 @@ module.exports = [
         ],
     },
     {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_p0gzbqct']),
+        model: 'TS0601_dimmer',
+        vendor: 'TuYa',
+        description: 'Zigbee smart knob dimmer',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        exposes: [tuya.exposes.lightBrightness(), tuya.exposes.lightType()],
+        meta: {
+            tuyaDatapoints: [
+                [1, 'state', tuya.valueConverter.onOff, {skip: tuya.skip.stateOnAndBrightnessPresent}],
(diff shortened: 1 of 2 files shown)


Koenkk:
Can this be merged with https://github.com/Koenkk/zigbee-herdsman-converters/blob/8406cf67ba43050129e4d79bdb64d5ea23f38b16/devices/tuya.js#L1223?

tefracky:
I tried it first, but no, it is a different device/hardware and it didn't work. The TuYa datapoints must be used. Also 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/8406cf67ba43050129e4d79bdb64d5ea23f38b16/devices/tuya.js#L1268

didn't work due to the difference in the datapoints.

Koenkk:
I see, does this device also support the max brightness? (I see you added min brightness).

tefracky:
No, it don't. I added it also in the PR for zigbee2mqtt.io: https://github.com/Koenkk/zigbee2mqtt.io/pull/1781

Koenkk:
I just found https://github.com/Koenkk/zigbee-herdsman-converters/blob/4b8c153235b0fef0481efec41d07ad90ce317f61/devices/tuya.js#L1268 , datapoints seem similar (with some extras). Can you check that one?

tefracky:
Yes, I already checked, but 'max_brightness', 'countdown' and 'power_on_behavior' are not supported. I checked the datapoints using the TuYa IoT developer website.

tefracky:
I will first make some more tests, maybe it is possible to use some other device.

tefracky:
So, overall, I only was able to include the lightType. Unfortunately, minBrightness could not be set or read, even if the data point is correct (checked using the TuYa IoT Developer platform).

Same for the indicator light: This can be 'none', relay' or 'pos', but not set or read.


light_mode

WickedDurkie:
I have the same Moes dimmer (in my case a model ZS3L with id '_TZE200_p0gzbqct' ) and it does support the minimum brightness. I'm using the TS0601_dimmer_1 implementation for now, which works correctly. Only the timer and max brightness options do not work.

tefracky:
@WickedDurkie you are right, I forgot the `setAccess('min_brightness', ea.STATE_SET)` during testing.

WickedDurkie:
BTW I reverted my PR which uses the same fingerprint. After merging, the ci checks should succeed.

tefracky:
Ready to be merged :)

devices/tuya.js
```diff
@@ -1343,6 +1343,30 @@ module.exports = [
...
+        configure: tuya.configureMagicPacket,
+        exposes: [tuya.exposes.lightBrightness().withMinBrightness().setAccess('min_brightness', ea.STATE_SET), tuya.exposes.lightType()],
+        /* tuya.exposes.indicatorModeNoneRelayPos() */
+        meta: {
+            tuyaDatapoints: [
+                [1, 'state', tuya.valueConverter.onOff, {skip: tuya.skip.stateOnAndBrightnessPresent}],
+                [2, 'brightness', tuya.valueConverter.scale0_254to0_1000],
+                [3, 'min_brightness', tuya.valueConverter.scale0_254to0_1000],
+                [4, 'light_type', tuya.valueConverter.lightType],
+                // [21, 'indicator_mode', tuya.valueConverterBasic.lookup({0: 'none', 1: 'relay', 2: 'pos'})],
```
Koenkk, line 1362: Shouldn't this be uncommented?

tefracky: It can, becaus in theroy, it should work. But in practice, it didn't work in my test. However, maybe it will work for others, so I uncommented.

Koenkk:
Thanks!

PR #5205, opened 2022-12-25
Add additional model number for Hue Resonate

devinpitcher:
Adds [929003098701](https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-resonate-outdoor-wall-light/046677576080), which is the same as Philips [1746430P7](https://www.zigbee2mqtt.io/devices/1746430P7.html).


Koenkk:
Thanks!

PR #5209, opened 2022-12-26
add TS004F Moes version

markxroberts:


markxroberts:
Looks as though this has already been addressed in #15730 from koenkk/zigbee2mqtt

PR #5212, opened 2022-12-26
Add Develco Vibration Sensor WISZB-137

dlq84:
Add support for Develco Vibration Sensor

https://www.develcoproducts.com/products/sensors-and-alarms/vibration-sensor/


Koenkk:
Thanks!

PR #5216, opened 2022-12-27
Updated Datek HSE2927E and added support for Datek Window/Door sensor HSE2920E

TheStigh:
Added support for Led-On-Motion for Datek Motion Sensor HSE2927E

### DIFF
--- a/devices/datek.js
+++ b/devices/datek.js
@@ -85,7 +85,7 @@ module.exports = [
             e.voltage_phase_c(), e.temperature()],
     },
     {
-        zigbeeModel: ['Motion Sensor'],
+        fingerprint: [{modelID: 'Motion Sensor', manufacturerName: 'Eva'}],
         model: 'HSE2927E',
         vendor: 'Datek',
         description: 'Eva motion sensor',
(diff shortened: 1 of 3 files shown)


TheStigh:
Please merge #5216 first.

- Added support for Led-On-Motion for Datek Motion Sensor HSE2927E
- Updated Datek Motion Sensor (HSE2927E) and added support for  Datek Window Sensor (HSE2920E)

devices/datek.js
```diff
@@ -85,13 +85,32 @@ module.exports = [
             e.voltage_phase_c(), e.temperature()],
     },
     {
-        fingerprint: [{modelID: 'Motion Sensor', manufacturerName: 'Eva'}],
+        zigbeeModel: ['Motion Sensor'],
```
Koenkk, line 88: Was this removed on purpose? (preferable this should be kept to prevent conflicts with devices using the same modelID)

TheStigh: Yes it was. No reason to actually remove it, do you want me to put it back ?

Koenkk:
Thanks!

PR #5218, opened 2022-12-28
Added _TZE200_m9skfctm implementation

Jonno12345:
Added implementation for the _TZE200_m9skfctm smoke alarm.


devices/tuya.js
```diff
@@ -3774,4 +3774,29 @@ module.exports = [
...
+        model: 'TS0601_smoke_2',
+        vendor: 'TuYa',
+        description: 'Photoelectric Smoke Detector',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        onEvent: tuya.onEventSetTime,
+        configure: tuya.configureMagicPacket,
+        exposes: [
+            e.smoke(),
+            exposes.numeric('smoke_level', ea.STATE).withUnit('ppm').withDescription('Parts per million of smoke detected'),
```
Koenkk, line 3788: Is `smoke_concentration` a better name?

Jonno12345: I just had a look on the Tuya IOT page and they call it 'Detected Smoke', and I get it but that sounds like a bool value to me. I'll change it to concentration, yes. Also in doing so, I noticed on the Tuya implementation they are working the value /10 so I'll add that too.

Koenkk:
Thanks!

PR #5219, opened 2022-12-28
Removes GBL padded data validation

pedrolamas:
This removes the GBL padded data verification completely as I can't find any reference that ending bytes should be 0x0 or any other thing for that matter.

This should fix https://github.com/Koenkk/zigbee-OTA/issues/218.


Koenkk:
Thanks!

Koenkk:
Awesome, thanks!

PR #5226, opened 2022-12-29
philips: enable effects for all LCA-series devices

zegl:
Follow up to #5192 

I tested the effects on a `LCA001` device, and it's working great. 👍

Based on this test, and the earlier tests by @sjorge I think that it's safe to assume that these features are supported by all `LCA` devices, as they are pretty much the same device but with different sockets and brightness.

Thanks again @sjorge 🙏


Koenkk:
Thanks! 🎆

PR #5227, opened 2022-12-29
philips: decode effects

zegl:
Added a decoder for the effects (see #5192) from the Philips `state` attribute.

### DIFF
--- a/lib/philips.js
+++ b/lib/philips.js
@@ -30,6 +30,16 @@ const decodeScaledGradientToRGB = (p) => {
 const COLOR_MODE_GRADIENT = '4b01';
 const COLOR_MODE_COLOR_XY = '0b00';
 const COLOR_MODE_COLOR_TEMP = '0f00';
+const COLOR_MODE_EFFECT = 'ab00';
+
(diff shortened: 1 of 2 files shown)


lib/philips.js
```diff
@@ -123,15 +141,26 @@ function decodeGradientColors(input, opts) {
...
+        // Effect mode
+        const effect = input.slice(0, 4);
+        const name = knownEffects[effect] || `unknown_${effect}`;
         return {
-            color_mode: 'xy',
-            x: Math.round(x / 65535 * 10000) / 10000,
-            y: Math.round(y / 65535 * 10000) / 10000,
-            brightness,
-            on,
+            color_mode: 'effect',
```
Koenkk, line 160: For compatibility reasons we have to stick to the Home Assistant color modes: https://www.home-assistant.io/integrations/light.mqtt/#json-schema . I propose to leave this `xy`

zegl: Cool! I'll change it! :-)

Koenkk:
Thanks!

PR #5228, opened 2022-12-29
philips: add Sunrise and Sparkle effects

zegl:
Sunrise is supported by all gradient lights. Sparkle is supported only by the Festavia light.


Koenkk:
Thanks!

PR #5232, opened 2022-12-30
Remove _TZ3000_6ygjfyll from ZB003-X, Add _TZ3000_6ygjfyll to IH012-RT01

x-magic:
To resolve the fact that a newer revision of IH012-RT01 is not working with zigbee2mqtt. 

See: https://github.com/Koenkk/zigbee2mqtt/issues/9757#issuecomment-1367716461

There's no sign that this new model closely resemble Fantem ZB003-X, but rather the same sensor as Tuya IH012-RT01. 

Tested working as of now. Cheers


Koenkk:
Thanks!

PR #5234, opened 2022-12-30
Add tuya ZTH01 temp. & hum. Sensor

Burki24:
Solves the problems with 'TS0601', ['_TZE200_yjjdcqsq']

closes Koenkk/zigbee2mqtt#15729
closes Koenkk/zigbee2mqtt#14819

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1712,7 +1712,7 @@ module.exports = [
         fingerprint: tuya.fingerprint('TS0601', ['_TZE200_yjjdcqsq']),
         model: 'ZTH01',
         vendor: 'TuYa',
-        description: 'Temperature and humidity',
+        description: 'Temperature and humidity sensor',
         fromZigbee: [tuya.fz.datapoints, fzLocal.gateway_connection_status],
         toZigbee: [tuya.tz.datapoints],
         configure: tuya.configureMagicPacket,
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1722,7 +1722,7 @@ module.exports = [
                 [1, 'temperature', tuya.valueConverter.divideBy10],
                 [2, 'humidity', tuya.valueConverter.raw],
                 [3, 'battery_state', tuya.valueConverter.batteryState],
-                // [9, 'temperature_unit', tuya.valueConverter.raw], This DP is not properly supported by the device, therefore commented out here
+                // [9, 'temperature_unit', tuya.valueConverter.raw], This DP is not properly supported by the device
             ],
         },
     },
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -720,6 +720,18 @@ const fzLocal = {
             return result;
         },
     },
+    gateway_connection_status: {
+        cluster: 'manuSpecificTuya',
+        type: ['commandMcuGatewayConnectionStatus'],
+        convert: async (model, msg, publish, options, meta) => {
+            // "payload" can have the following values:
+            //  0x00: The gateway is not connected to the internet.
+            // 0x01: The gateway is connected to the internet.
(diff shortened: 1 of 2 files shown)


Burki24:
Solves this issue #15729

devices/tuya.js
```diff
@@ -1696,6 +1708,24 @@ module.exports = [
...
+        fromZigbee: [tuya.fz.datapoints, fzLocal.gateway_connection_status],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        exposes: [e.temperature(), e.humidity(), tuya.exposes.batteryState(), e.battery_low()],
+        meta: {
+            tuyaDatapoints: [
+                [1, 'temperature', tuya.valueConverter.divideBy10],
+                [2, 'humidity', tuya.valueConverter.raw],
+                [3, 'battery_state', tuya.valueConverter.batteryState],
+                // [9, 'temperature_unit', tuya.valueConverter.raw], Does not have any effect on the device
```
TheRealMephisto, line 1725: @Burki24 This line is  commented out, but the dp value 9 is added to lib/tuya.js
Why is that?

You write here in the comment "Does not have any effect on the device" -> What do you want to say with that?

devices/tuya.js
```diff
@@ -720,6 +720,18 @@ const fzLocal = {
             return result;
         },
     },
+    gateway_connection_status: {
+        cluster: 'manuSpecificTuya',
+        type: ['commandMcuGatewayConnectionStatus'],
+        convert: async (model, msg, publish, options, meta) => {
+            // "payload" can have the following values:
```
TheRealMephisto, line 727: Not sure about how @Koenkk handles comments in the code in this library.
For my part, I never have commented out lines in the code itself as a best practice. 

These lines here should belong in a documentation IMHO - @Koenkk I did not find a documentation covering the gateway connection status. Should be somewhere close to [this](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html) I think .

I can try to sum up my learnings from this PR and from the discussion in issue #15729 and create a PR to add some info about that in the documentation, if that is ok for you.

TheRealMephisto (commented):
@Burki24 Thanks a lot again for the work done here! 
Just one small question for you in my comments about the code.


Looking forward to see my sensors working :)

devices/tuya.js
Burki24: DP 9 stands for switching between °C and °F. But it's not really supported by this device. So also in the original APP it isn't usable. For that, we don`t need this DP inside the converter or as Expose.

TheRealMephisto: Ok, understood!
I'll suggest a change of the comment in the code, just in case someone else stumbles across that line in the future and asks the same question as I did.

One more question: Will the zigbee2mqtt log not show a "No converter available [...]" message for the DP 9, if we comment it out completely?
If so, maybe we should leave it in there and handle the useless data transmitted by it later, in order to avoid that log message

TheRealMephisto (commented):
Clarify statement about commented out DP

devices/tuya.js
Burki24: No, I get no warnings or something else from the outcommented DP 9. So please, let it stay as it is.

Burki24:
> Clarify statement about commented out DP
> 
> [Edit: First time that I am using the online editor in Github. Trying to figure out how to push the suggested code change to this PR :D]

Thanks.

devices/tuya.js
Koenkk: These are valid comments IMO, they are not code, just explains what it does. 

These lines are taken from the TuYa developer documentation.

Koenkk:
Thanks!

PR #5236, opened 2022-12-31
Add support for Hue 6 inch High Luminance White Ambiance Downlight

syncnj:
* Added support for new 6 inch White Ambiance High Luminance version of downlight.
* This is the Amazon/US link for reference: https://www.amazon.com/gp/product/B0B6L38LK4


Koenkk:
Thanks!

PR #5241, opened 2023-01-01
Adding timer for sinope DM2500 and SW2500

onixx:
Adding timer for sinope DM2500 and SW2500 plus add fz for features I previously added such as ledintensity, min brightness. also change to set only led color since contribution does not include get


Koenkk:
Thanks!

PR #5243, opened 2023-01-01
Expose led_in_dark setting for Legrand 067776

AntorFr:
Expose led_in_dark setting for Legrand 067776 (same as Legrand 067775/741811)

tested with my devices at home.

npm lint / npm test performed without issues.


Koenkk:
Thanks!

PR #5245, opened 2023-01-01
Enable Hue effects for Hue Surimu 929002966501

merlinschumacher:
The "Hue White and Color Ambiance Surimu rectangle panel 929002966501" also supports the additional effects. I've tried it on my hardware.


Koenkk:
Thanks!

PR #5249, opened 2023-01-02
philips: fix binding for gradient lights

zegl:
Looks like this broke in the refactoring in #5192.

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/15883

(The issue is that `noConfigure` was true on line 88, when creating the `configure` function).


Koenkk:
Thanks!

PR #5253, opened 2023-01-02
Support power_on_behavior for 3RSB015BZ

rccoleman:
I think there's a dependency on firmware 1.00.29 for power_on_behavior to actually work, but it's valid nonetheless.  I've tested this locally with several of my plugs running 1.00.29.


Koenkk:
I enabled this for all switches now in https://github.com/Koenkk/zigbee-herdsman-converters/commit/fdb755b5ac21d81127bfb20d2b8557de185961b1. Thanks!

PR #5195, opened 2022-12-24
Refactor Sinope.js (Cosmetics)

MattL0:

PR #5255, opened 2023-01-02
Fix TH1400ZB configure

MattL0:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/5195#issuecomment-1369207694


Koenkk:
Thanks!

PR #5257, opened 2023-01-03
Add support for _TZE200_khx7nnka illuminance sensor

ov1d1u:
This is similar to `_TZE200_yi4jtqq1`, it just needed the fingerprint to be recognised by z2m.

AliExpress link of the device: https://www.aliexpress.com/item/1005004472770710.html


Koenkk:
Thanks!

PR #5258, opened 2023-01-03
fix: change reporting intervals for Sonoff SNZB-0x to prevent device disconnect

blacknell:
Prevents them from being disconnected from the network

relates https://github.com/Koenkk/zigbee2mqtt/issues/13600


Koenkk:
Thanks!

PR #5263, opened 2023-01-03
Add device Legrand 067766 Centralized ventilation SW

AntorFr:
Add support for device Legrand 067766 Centralized ventilation SW (official link [here](https://www.legrand.fr/pro/catalogue/44198-micromodule-pour-vmc-connectee/micromodule-pour-rendre-une-vmc-2-vitesses-connectee-250w-max))

Tested with my own device, fully functional
Lint and Test passed


Koenkk:
Thanks!

PR #5264, opened 2023-01-03
Exposed apparent power for Legrand 412015

Mephistofeles:
Legrand DIN power monitoring module measures apparent power, but it's not exposed currently.


Koenkk:
Thanks!

PR #5265, opened 2023-01-03
Inoveilli VZM31-SN: Energy Report Interval

InovelliUSA:
Changing the default reporting intervals to their expected values.


Koenkk:
Thanks!

epower53:
15 seconds still feels pretty aggressive, especially for energy (seMetering). How about we put up a poll on the Inovelli community and use the results to set values most people will be happy with and won't have to modify? I just started using this new version of the converter, and had to go through every single switch to get back to values I wanted (much longer than the default. My guess is many people will end up doing this...

PR #5267, opened 2023-01-04
Added Hue Effects support to the Hue Gradient Lightstrip and Hue Ensis White and Black

marblecoma:
I noticed in the Philips devices file that none of the Hue Gradient lights have the Hue Effects enabled. I only have the Hue Gradient lightstrip, but this one does support the Hue Effects in the Hue app. Also, I only have the black Hue Ensis light, but I added the Hue Effect support for the white one, as well. I will create a pull request to update the functionality in the Zigbee2mqtt.io devices page, as well.


Koenkk:
Thanks!

PR #5268, opened 2023-01-04
philips: expose current effect

zegl:
This fixes https://github.com/Koenkk/zigbee2mqtt/issues/15891

* Exposes `effect` for all devices that have custom Philips effects (defaults to `null` if no effect is in use).
* Changes the `gradient` state to default to an empty list if the gradient no longer is in use (gradient capable devices only).
* Renamed `decodeGradientColors` -> `decodeState` and `fzLocal.gradient` -> `philipsState` as it's now clear that the same attribute does more than just gradients.
* Updates the binding to only bind `manuSpecificPhilips2` on endpoint 11. The binding on all endpoints was an error, back from when I first added gradient support.

One question:

It seems like Hue devices with older firmware versions doesn't always have a `manuSpecificPhilips2` cluster. Would it be a good idea to wrap the binding in a try/catch? Or does that cause more harm than it prevents.

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -33,9 +33,18 @@ const hueExtend = {
         result['ota'] = ota.zigbeeOTA;
         result['meta'] = {turnsOffAtBrightness1: true};
         result['toZigbee'] = result['toZigbee'].concat([tzLocal.hue_power_on_behavior, tzLocal.hue_power_on_error]);
+
+        const extendConfigure = result['configure'];
+        result['configure'] = async (device, coordinatorEndpoint, logger) => {
+            await extendConfigure(device, coordinatorEndpoint, logger);
+            const endpoint = device.getEndpoint(11);
+            endpoint.bind('manuSpecificPhilips2', coordinatorEndpoint);
+        };
+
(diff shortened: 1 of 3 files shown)


Koenkk:
> It seems like Hue devices with older firmware versions doesn't always have a manuSpecificPhilips2 cluster. Would it be a good idea to wrap the binding in a try/catch? Or does that cause more harm than it prevents.

What error is logged in this case?

zegl:
I'm stupid enough and upgraded the device to the latest version – I'll try to downgrade it and check again.

If I remember correctly, the error was that the cluster could not be bound, or something along those lines. But that was also at a point when I was trying to bind it on all endpoints, and not only on endpoint 11.

sjorge:
It's probably be better to use the common lightingColorCtrl cluster for the non hue specific stuff and just use the new one for gradients, effects, ... ?

e.g. doesn't look like you can set hue/saturation color_mode via the hue specific cluster.

jonrupell:
Was this ever implemented? All my hue bulbs in z2m still show null no matter if an effect is being played or not

PR #5272, opened 2023-01-05
philips: set colorTempRange for LTW013

zegl:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/15929

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1379,7 +1379,7 @@ module.exports = [
         model: '8718696598283',
         vendor: 'Philips',
         description: 'Hue white ambiance GU10',
-        extend: hueExtend.light_onoff_brightness_colortemp(),
+        extend: hueExtend.light_onoff_brightness_colortemp({colorTempRange: [153, 454]}),
     },
     {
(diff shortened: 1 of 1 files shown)


sjorge:
Is this actually what the device reports as supported? All hue lights i got my hands on report the same value when reading the color cluster.

You can check these by reading lightingColorCtrl.colorTempPhysicalMin and lightingColorCtrl.colorTempPhysicalMax.

zegl:
@sjorge I didn't check with a physical device, as I don't have one, but the claims made in the original issue seems to be true. I double checked it against [some documentation](Example: https://www.bauhaus.se/media/pdf/2008840B.pdf) that I found online, which claims a temperature range between 2200 and 6500 K.

zegl:
I'll update the PR to also include the LTG002.

sjorge:
Ideally we verify it on the actual hardware but since there is usually little variance with Hue and my ws also returns the same range/have matching info on the site. It’s indeed probably OK to do both.

Koenkk:
Thanks!

PR #5273, opened 2023-01-05
ubisys: avoid INSUFFICIENT_SPACE while reading config

sjorge:
https://github.com/Koenkk/zigbee2mqtt/issues/15875

- tz: read inputConfiguration and inputActions seperately
- fz: merge correctly because they are 2 reads now

This change allows me to at least read my C4 and S2, this doesn't address the weird binds being reset however.


Koenkk:
Thanks!

PR #5274, opened 2023-01-05
philips: Support Hue Gradient Signe floor lamp (wooden color)

moshemalawach:
This lamp has all the same caracteristics than other colors, just another zigbee id...


Koenkk:
Thanks!

moshemalawach:
Not sure if it's available already on nightly docker build? I tried it but it still appears as "unsupported". If it's integrated, it's perhaps because of the brand "Signify Netherlands B.V."? (for some reasons the models sold in europe have this brand instead of "Philips") 
image

Koenkk:
The zigbee model has to be `929003479701` (as shown on the picture), I've pushed a fix.

PR #5275, opened 2023-01-05
Fixed external_measured_room_sensor description

TomTonic:
Fixed #15869
Changed true and false in the External_measured_room_sensor explanation to match Radiator_covered and the Danfoss documentation.


Koenkk:
Thanks!

PR #5276, opened 2023-01-05
Added action for on/off switches (new device option switch_action, default false)

Mr-Groch:
Those wall switches (not switch modules) can be installed without wiring all buttons to lamp and used in automations to control lights wirelessly. Or even if wired it can be used to control other related lights in automations. Action topic is preferred to be used as trigger, rather than state (state topic changes also when last_seen, linkquality or other property changes, so it must be filtered)

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -802,7 +802,6 @@ const converters = {
                 const state = msg.data['onOff'] === 1 ? 'ON' : 'OFF';
                 payload[property] = state;
                 if (options && options.state_action) {
-                    console.log(meta, model);
                     payload['action'] = postfixWithEndpointName(state.toLowerCase(), msg, model, meta);
                 }
                 return payload;
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -794,37 +794,58 @@ const converters = {
     on_off: {
         cluster: 'genOnOff',
         type: ['attributeReport', 'readResponse'],
+        options: [exposes.options.switch_action()],
         convert: (model, msg, publish, options, meta) => {
             if (msg.data.hasOwnProperty('onOff')) {
+                const payload = {};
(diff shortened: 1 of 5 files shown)


Koenkk:
Such feature is not specific to this device, it can apply to every wall switch, right?

Mr-Groch:
Basically yes, but problem is with naming - single switches actions can be named on and off always, but multi button switches endpoints are sometimes named like left, center, right, sometimes l1, l2, others just 1, 2, etc. I ddon't know well z2m code, so don't know if I can get endpoint name in fromZigbee converter (I wasn't able to find such example)

Koenkk:
My point is that this is basically duplicate information. It's the switch state but then just in another form. Can't this be easily solved with an HA automation?

Mr-Groch:
In HA yes - there are device events like switched on / off or I can use trigger based on entity state - HA have logic do detect state changes, but I don't have HA in mind. Any other software that want to run something based on switch event needs to remeber actuall state to detect event - there is no possibility to make stateless software now to handle switch events

Koenkk:
Since this functionality is not specific to these TuYa devices, it should also be added to all other switches. (mainly the `on_off` fromZigbee converter. Here you can find code to retrieve the endpoint name: https://github.com/Koenkk/zigbee-herdsman-converters/blob/47828fbc293befba72aa572229f7304181c58217/lib/utils.js#L131

Mr-Groch:
OK, should I add action by defaut for all extend.switch() or like in this PR as option ?

Koenkk:
Not by default, this should be an option as not everyone will be interested in this. e.g. `switch_action`

Mr-Groch:
Changed PR topic, and commited publishing action topic for all on_off converters when new switch_action device option is set to true

Koenkk:
Thanks, I've renamed it to `state_action` since we can also use it for other device types than switches.

PR #5278, opened 2023-01-06
Remove uneeded space for TH1400ZB

MattL0:
Also use const 'e' instead of 'exposes.presets' on some exposes for the same thermostat exposes


Koenkk:
Thanks!

PR #5279, opened 2023-01-06
Alternate modelID for samotech SM323

simonses1:
New SM323 modelID appears to be 'HK_DIM_A'


benpjohnson:
Thank you @simonses1 I just wired one in this morning with the same model modelID as you.

Koenkk:
Thanks!

PR #5282, opened 2023-01-06
Update sunricher.js

DubhAd:
The mis-copy of the modelID from Terncy causes the Terncy devices to be [mis-identified and no longer function](https://github.com/Koenkk/zigbee2mqtt/issues/15979). I am assuming here that the modelID is correct, based on the other information provided.

### DIFF
--- a/devices/sunricher.js
+++ b/devices/sunricher.js
@@ -606,7 +606,7 @@ module.exports = [
         },
     },
     {
-        fingerprint: [{modelID: 'TERNCY-DC01', manufacturer: 'Sunricher'}],
+        fingerprint: [{modelID: 'SR-ZG9010A', manufacturer: 'Sunricher'}],
         model: 'SR-ZG9010A',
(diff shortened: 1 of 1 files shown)


devices/sunricher.js
```diff
@@ -606,7 +606,7 @@ module.exports = [
         },
     },
     {
-        fingerprint: [{modelID: 'TERNCY-DC01', manufacturer: 'Sunricher'}],
+        fingerprint: [{modelID: 'TERNCY-DC01', manufacturerName: 'Sunricher'}],
```
Koenkk, line 609: @asgothian `manufacturer` had to be `manufacturerName` , @DubhAd can you check if this works now?

DubhAd: I've just tested and it works now - my Terncy devices are still detected as such and work again.

Koenkk:
Thanks!

PR #5288, opened 2023-01-07
Disable power_on_behavior for OSRAM Plugs

stickpin:
Disable unsupported power_on_behavior for OSRAM Plugs AC10691 and AB3257001NJ


Koenkk:
Thanks!

PR #5289, opened 2023-01-07
Add missing functionality to Bticino switches

knovoselic:
The following functionality has been added:
- exposed controls for `Identify` function (blinks built-in LED 3 times to make identifying a specific switch easier)
- exposed `led_in_dark` control for 4027C type switches (turns on/off build-in LED indicator)


Koenkk:
Thanks!

knovoselic:
@Koenkk one note/question. The help text for `led_in_dark` for K4027C was intentionally change to 
```
Enables the built-in LED
```
instead of 
```
Enables the LED when the light is turned off, allowing to see the switch in the dark
```

because this is a shutter switch and there is not "light turned off" state. The option simply turns on LED forever, or turns it off. Any thoughts how to better word this, as the current help message is a bit misleading?

Koenkk:
Just `Enables the built-in LED` sounds good to me.

PR #5290, opened 2023-01-07
Add hue effects to Philips Liane/Being/Lily

Sabesto:
Added hue effects to:
Liane (4090230P9)
Being Black/White (3261031P6 / 3261030P6) (Only "Candle") Lily (1741530P7)

Other zigbeeModel's with the same product name probably supports these effects, but I only added those I own and could test.


Koenkk:
Thanks!

PR #5291, opened 2023-01-07
Fix move _TZE200_9yapgbuv to the correct model

Sherex:
### Moves the Tuya TS0601 from `WSD500A` to `ZTH01`.

When my Tuya temp and humidity sensor was recognized as `WSD500A` they stopped reporting temperatures and neither resetting, pressing the button or re-pairing worked. I also made sure there was temperature changes when testing, as it only reports when there has been a change.

At the time `_TZE200_9yapgbuv` was added to the `WSD500A` model (#4994), the `ZTH01` had not been created yet (#5234).
I'm not sure if it ever worked for @floxboy (#4994), but I couldn't get it to work in the current version (1.29.1).

It has worked perfectly now for the last ~24 hours.
Previously I used a hacked together external converter for this device which also worked well.

Other relevant issues/PRs:
- https://github.com/Koenkk/zigbee2mqtt/issues/15729
- https://github.com/Koenkk/zigbee2mqtt/issues/14819

This should close https://github.com/Koenkk/zigbee2mqtt/issues/15774

---

Thank you for this great software! It's one of the few which is a pleasure to use. :)

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1668,7 +1668,6 @@ module.exports = [
             {modelID: 'TS0201', manufacturerName: '_TZ3000_yd2e749y'},
             {modelID: 'TS0201', manufacturerName: '_TZ3000_6uzkisv2'},
             {modelID: 'TS0201', manufacturerName: '_TZ3000_xr3htd96'},
-            {modelID: 'TS0601', manufacturerName: '_TZE200_9yapgbuv'},
         ],
         model: 'WSD500A',
         vendor: 'TuYa',
@@ -1698,7 +1697,7 @@ module.exports = [
         exposes: [e.battery(), e.temperature(), e.humidity(), e.battery_voltage()],
     },
     {
-        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_yjjdcqsq']),
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_yjjdcqsq', '_TZE200_9yapgbuv']),
         model: 'ZTH01',
         vendor: 'TuYa',
         description: 'Temperature and humidity sensor',
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1701,18 +1701,10 @@ module.exports = [
         model: 'ZTH01',
         vendor: 'TuYa',
         description: 'Temperature and humidity sensor',
-        fromZigbee: [tuya.fz.datapoints, tuya.fz.gateway_connection_status],
-        toZigbee: [tuya.tz.datapoints],
+        fromZigbee: [fz.tuya_temperature_humidity_sensor],
+        toZigbee: [],
         configure: tuya.configureMagicPacket,
         exposes: [e.temperature(), e.humidity(), tuya.exposes.batteryState(), e.battery_low()],
-        meta: {
-            tuyaDatapoints: [
-                [1, 'temperature', tuya.valueConverter.divideBy10],
-                [2, 'humidity', tuya.valueConverter.raw],
-                [3, 'battery_state', tuya.valueConverter.batteryState],
-                // [9, 'temperature_unit', tuya.valueConverter.raw], This DP is not properly supported by the device
-            ],
-        },
     },
     {
         fingerprint: [{modelID: 'TS011F', manufacturerName: '_TZ3000_3zofvcaa'}],
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1705,6 +1705,14 @@ module.exports = [
         toZigbee: [],
         configure: tuya.configureMagicPacket,
         exposes: [e.temperature(), e.humidity(), tuya.exposes.batteryState(), e.battery_low()],
+        meta: {
+            tuyaDatapoints: [
+                [1, 'temperature', tuya.valueConverter.divideBy10],
+                [2, 'humidity', tuya.valueConverter.raw],
+                [3, 'battery_state', tuya.valueConverter.batteryState],
+                // [9, 'temperature_unit', tuya.valueConverter.raw], This DP is not properly supported by the device
+            ],
+        },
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks for the pr, the exposes also needs to be updated since this device doesn't support battery % (only battery_state and battery_low). You can find the correct `exposes()` here: https://gist.github.com/Koenkk/6878634d9cd8ba56d00e19d4938ae7c1

Sherex:
I believe it's only the `WSD500A` device (in [devices/tuya.js#L1677](https://github.com/Sherex/zigbee-herdsman-converters/blob/fb4339150b7b840bce817dfc862b93766c368343/devices/tuya.js#L1677)) which exposes the battery percentage and voltage. (I don't have this model, so can't check if that's correct. But I think it is based on it's age in this file)

And that `_TZE200_9yapgbuv` is now recognised as `ZTH01` which correctly only exposes `battery_state` and `battery_low` (in [devices/tuya.js#L1707](https://github.com/Sherex/zigbee-herdsman-converters/blob/fb4339150b7b840bce817dfc862b93766c368343/devices/tuya.js#L1707)). And it seems the battery_state has been extracted to it's own method `batteryState()` (in [lib/tuya.js#L1170](https://github.com/Sherex/zigbee-herdsman-converters/blob/fb4339150b7b840bce817dfc862b93766c368343/devices/tuya.js#L1170)), but implemented the same way.

But I might be missing something or misunderstood you, please correct me if I'm wrong. :)

criegit:
I have the same problem. Running Hass OS on RPi4.
How long until this fix is in the official repo?

I have successfully tried editing the tuya.js file in the docker container, commiting the changes, restarting, but the sensor was still recognized as WSD500A when rebinding...

Sherex:
I noticed it doesn't actually report `battery_low` using the current device config for `ZTH01`. And it turns out the sensor doesn't report it at all as far as I can see. It does report a DP 9, which should be the temperature unit.

I noticed I used `fromZigbee: [fz.tuya_temperature_humidity_sensor]` in my external converter, which uses `battery_state === 0` to create the `battery_low` bool.

To test I changed my device config to use `fromZigbee: [fz.tuya_temperature_humidity_sensor]` instead and it seems to work fine and it also creates the `battery_low` bool.

But currently this `ZTH01` device is using `fromZigbee: [tuya.fz.datapoints, tuya.fz.gateway_connection_status]` and then further specifying the types using the `meta` config.

---

1. What do you prefer, using the raw datapoints or the `fz.tuya_temperature_humidity_sensor`?

2. Is there a benefit of exposing both `battery_low` and `battery_state` compared to just `battery_state`?

---

Let me know your preferences and I'll update the PR :)

Sherex:
Hey @criegit!

> I have successfully tried editing the tuya.js file in the docker container, commiting the changes, restarting, but the sensor was still recognized as WSD500A when rebinding...

1. Did you also remove `_TZE200_9yapgbuv` from the `WSD500A` device config?

2. When commiting the changes in the container, the image including the current tag also has to be specified.  
Example: `docker commit addon_45df7312_zigbee2mqtt zigbee2mqtt/zigbee2mqtt-amd64:1.29.1-1`

3. And then the container has to be restarted. Either through the Addon menu in Home Assistant or with the CLI tool. (if only Z2M is restarted inside the container it won't pick up the changes from `node_modules`)

I hope it helps in the debugging! :)

criegit:
Hi, that helped a lot. 
I got it to change the model. 
First only first commit, then both. I am pretty sure Battery status showed then, but not the rest. 
The i tested the edge version, but no data at all.

Using the Edge version is somehow brokenin my quite fresh HA OS installation. It is not correctly installed into the addons. (and the standard one not removed by uninstalling)
I still got it running in edge, but it does not work better than the standard version.

Burki24:
> I noticed it doesn't actually report `battery_low` using the current device config for `ZTH01`. And it turns out the sensor doesn't report it at all as far as I can see. It does report a DP 9, which should be the temperature unit.

That's correct, but the device didn't accept changes on that DP. So it's not needed for the ZTH01-Devices. We have the DP in the converter but not activated at the moment. Not sure if it's planned to update the FW by the suplier.

devices/tuya.js
```diff
@@ -1698,22 +1697,14 @@ module.exports = [
...
         model: 'ZTH01',
         vendor: 'TuYa',
         description: 'Temperature and humidity sensor',
-        fromZigbee: [tuya.fz.datapoints, tuya.fz.gateway_connection_status],
-        toZigbee: [tuya.tz.datapoints],
+        fromZigbee: [fz.tuya_temperature_humidity_sensor],
+        toZigbee: [],
         configure: tuya.configureMagicPacket,
         exposes: [e.temperature(), e.humidity(), tuya.exposes.batteryState(), e.battery_low()],
-        meta: {
```
Koenkk, line 1709: Please undo these changes (`fromZigbee`/`toZigbee` should stay as is).

Sherex: Thanks for the review! :)

I have reverted the removal of meta.
But I don't quite understand why it is needed?
I thought the converter `fz.tuya_temperature_humidity_sensor` mapped the datapoints and did any computation needed (eg. `divideBy10`).

But this didn't quite explain why, only how to use it:
https://www.zigbee2mqtt.io/advanced/support-new-devices/02_support_new_tuya_devices.html#_4-mapping-the-datapoints

Koenkk: It does indeed, but the separate converter for each device is the old way of integrating TuYa devices. Now just  `tuya.tz.datapoints/`tuya.fz.datapoints` should be used. Do you experience any problems when doing that?

criegit:
Any idea when this fix will be integrated into the HA OS / Zogbee2mqtt plugin?

PR #5292, opened 2023-01-07
Added support for Schneider Wiser Smoke Alarm

dennisgranasen:
It seems to work, although I have never actually triggered the alarm (lucky me).

Unknown cluster 64516 is not implemented.
No functionality has been added to manually trigger/test siren. I assume it is possible to do.


devices/schneider_electric.js
```diff
@@ -169,6 +170,25 @@ const fzLocal = {
...
+            return {
+                smoke: (zoneStatus & 1) > 0,
+                tamper: (zoneStatus & 1<<2) > 0,
+                battery_low: (zoneStatus & 1<<3) > 0,
+                supervision_reports: (zoneStatus & 1<<4) > 0,
+                restore_reports: (zoneStatus & 1<<5) > 0,
+                trouble: (zoneStatus & 1<<6) > 0,
+                ac_status: (zoneStatus & 1<<7) > 0,
+                test: (zoneStatus & 1<<8) > 0,
+                battery_defect: (zoneStatus & 1<<9) > 0,
```
Koenkk, line 188: This is according to the spec and therefore can be added to the original `ias_smoke_alarm_1` (then this can be removed)

dennisgranasen:
Yes it is, BUT...

1) The original implementation does not support battery_defect despite that
the spec says so. My implementation does. Perhaps it should be incorporated
in the original version instead? I didn't know if there is a reason for why
it was left out so I opted to put it in a separate solution instead.

2) The comment highlights the real reason why I implemented a new version
-  I want it to fetch the values upon pairing so it needs to listen to
readResponse... Perhaps there is a better way to accomplish this?

 // added readResponse to get sensor to trigger status updates upon pairing
+        type: ['commandStatusChangeNotification', 'attributeReport',
'readResponse'],



Den sön 8 jan. 2023 kl 09:43 skrev Koen Kanters ***@***.***>:

dennisgranasen:
I made an update to fix missing temperature readings. 
Not really sure how the underlying stuff works yet, still learning - but I couldn't get it to work with the base exposes, so I had to do this to get the readings to show correctly. The calibration works, but the precision does not seem to do anything. It never shows any decimals.

devices/xiaomi.js
```diff
@@ -121,10 +121,10 @@ const fzLocal = {
                 case 0x027c: // 0x00
                 case 0x027d: // 0x00
                 case 0x0280: // 0x00/0x01
-                    meta.logger.debug(`zigbee-herdsman-converters:aqara_trv: Unhandled key ${key} = ${value}`);
+                    meta.logger.debug(`zigbee-herdsman-converters:aqara_trv: Unhandled key ${key} = 0x${value.toString(16)}`);
                     break;
                 default:
-                    meta.logger.warn(`zigbee-herdsman-converters:aqara_trv: Unknown key ${key} = ${value}`);
+                    meta.logger.warn(`zigbee-herdsman-converters:aqara_trv: Unknown key ${key} = 0x${value.toString(16)}`);
                 }
```
Koenkk, line 128: These changes should go into a different pr.

Koenkk:
@dennisgranasen you can integrate both changes to the original ias_smoke_alarm_1 converter.

devices/xiaomi.js
dennisgranasen: Yes, you are right. Sorry, forgot to switch branch ;)

dennisgranasen:
> @dennisgranasen you can integrate both changes to the original ias_smoke_alarm_1 converter.

Ok, thanks. I've done that now.

converters/fromZigbee.js
```diff
@@ -957,7 +957,7 @@ const converters = {
     },
     ias_smoke_alarm_1: {
         cluster: 'ssIasZone',
-        type: 'commandStatusChangeNotification',
+        type: ['commandStatusChangeNotification', 'attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
             const zoneStatus = msg.data.zonestatus;
```
Koenkk, line 962: For the `attributeReport`/`readResponse` the attribute is `zoneStatus`, not `zonestatus`. This line should become:

```js
const zoneStatus = msg.type === 'commandStatusChangeNotification' ? msg.data.zonestatus : msg.data.zoneStatus;
```

devices/schneider_electric.js
```diff
@@ -981,4 +983,27 @@ module.exports = [
...
+    {
+        zigbeeModel: ['W599001'],
+        model: 'W599001',
+        vendor: 'Schneider Electric',
+        description: 'Wiser smoke alarm',
+        fromZigbee: [fz.schneider_temperature, fz.battery, fz.ias_enroll, fz.ias_smoke_alarm_1],
+        toZigbee: [],
+        ota: ota.zigbeeOTA, // local OTA updates are untested
+        exposes: [e.smoke(), e.battery_low(), e.tamper(), e.battery(), e.battery_voltage(),
+            en('local_temperature', ea.STATE).withUnit('°C').withDescription('Current temperature measured on the device'),
```
Koenkk, line 995: To be in line with other devices, this should be either `temperature` or `device_temperature`.

dennisgranasen:
local_temperature comes from the schneider_temperature implementation. If
this is changed here, then it needs to be changed there as well. I was
afraid of breaking stuff if I meddled with that. Do you want me to do that,
or should that be a subsequent PR?

Den tis 10 jan. 2023 kl 16:39 skrev Koen Kanters ***@***.***>:

converters/fromZigbee.js
dennisgranasen: Oh, good catch. Sorry for that mistake.
(Perhaps it would be wise to name them systematically to avoid such confusion in the future)

Koenkk:
> local_temperature comes from the schneider_temperature implementation. If
this is changed here, then it needs to be changed there as well.

Instead of `fz.schneider_temperature`, `fz.temperature` can be used. Since this is a new supported device, it is not a breaking change.

devices/schneider_electric.js
```diff
@@ -981,4 +982,28 @@ module.exports = [
...
+    {
+        zigbeeModel: ['W599001'],
+        model: 'W599001',
+        vendor: 'Schneider Electric',
+        description: 'Wiser smoke alarm',
+        fromZigbee: [fz.temperature, fz.battery, fz.ias_enroll, fz.ias_smoke_alarm_1],
+        toZigbee: [],
+        ota: ota.zigbeeOTA, // local OTA updates are untested
+        exposes: [e.smoke(), e.battery_low(), e.tamper(), e.battery(), e.battery_voltage(),
+            // the temperature readings are unreliable and may need more investigation.
```
Koenkk, line 994: I saw you mail, what do you mean with:

> Anyway, temperature readings seem to work sporadically.

It may help to increase the reporting frequency (in `configure`), try with the code below, it should make it report every 5 seconds:

```
await reporting.temperature(endpoint, {min: 1, max: 5, change: 0});
```

dennisgranasen:
I've had all sorts of problems with the temperature readings, but I think
they can all be summarized into these three categories:
1) Sometimes after configure/bind, the device responds to binding and to
read requests, but the device incorrectly reports 0 degrees -- both on
reports and read requests.
                      --> Sometimes it is just one report/reading that is
zero, but I've also had to remove and pair again to get better readings. I
have not investigated further the reason for this.
2) Sometimes after configure/bind, the device responds (seemingly correct)
to read requests, but reports do not work (even though binding seems to be
correct as far as I can tell from the UI and the logs).
                      --> Solution is to remove the device from the z2m and
pair it anew.
3) Sometimes after configure/bind, the device does not respond to read
requests (on the msTemp cluster, I haven't verified with ssIasZone), and no
temperature reports are coming even though binding seems correct (at least
I cannot see anything wrong in the logs).
                      --> Solution is to remove the device from the z2m and
pair it anew.

Note that ssIasZone reports are working just fine in all scenarios above.

Perhaps you are right that more frequent reporting solves the problem.
Every five seconds seems like a waste of energy, but I suppose it would be
a good test to see if the device behaviour becomes more predictable. I'll
do that at once. Then again, this is a smoke detector, not a thermometer. I
highly doubt its readings are very reliable anyway, so if the speedier
reports do not improve anything, perhaps we should just accept that
the temperature readings do not work very well?


Den fre 13 jan. 2023 kl 11:06 skrev Koen Kanters ***@***.***>:

Koenkk:
@dennisgranasen these sounds like some hw/fw issues of the device itself. Thanks!

t112013:
Hello guys, one quick question. does anyone know if activating the siren will be done? or do something with the zones, since its almost a must to be able to interconnet the sensors when you have more than one.
Thanks :)

PR #5299, opened 2023-01-08
Fixes for Busch-Jaeger devices

klada:
This PR adds a number of improvements and fixes for Busch-Jaeger ZLL devices:

* Add support for Busch-Jaeger dimmer module (6715 U) by exposing device as a dimmable light
* Expose the correct endpoint for the light device (see #5097 for a discussion about this)
* Do not expose state for Busch-Jaeger switch-only devices (without `relay` endpoint)
* Do not remove the scenes cluster from battery-operated Busch-Jaeger switch (RB01) as this is not required for these devices

This PR has a few implications:

* ~~Home Assistant auto-discovery will be changed as the device type changes from switch to light (in order to support dimming)~~
* If people want to bind previously paired RB01 devices to ZigBee scenes, they need to remove and re-pair them to Z2M. However there is no actual need to re-pair these devices if this feature is not important for people and they will continue to work fine.
* Existing BJ devices will be reconfigured, which means that the default bindings will be created again. It should be noted in the release notes that users should check their bindings, if they have messed with them.

Fixes Koenkk/zigbee2mqtt#7077

### DIFF
--- a/devices/busch-jaeger.js
+++ b/devices/busch-jaeger.js
@@ -71,33 +71,37 @@ module.exports = [
                 }
             }
 
+            // The total number of bindings seems to be severely limited with some of these devices.
+            // In order to be able to toggle groups, we need to remove the scenes cluster from RM01.
+            const dropScenesCluster = device.modelID == 'RM01';
+
             const endpoint11 = device.getEndpoint(0x0b);
             if (endpoint11 != null) {
-                // The total number of bindings seems to be severely limited with these devices.
-                // In order to be able to toggle groups, we need to remove the scenes cluster
-                const index = endpoint11.outputClusters.indexOf(5);
-                if (index > -1) {
-                    endpoint11.outputClusters.splice(index, 1);
+                if (dropScenesCluster) {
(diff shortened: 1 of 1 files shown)


Koenkk:
Let me know once this is ready for merge.

klada:
I'll do more testing next week and I also want to push out some improvements to the documentation. I'll ping you as soon as I am ready ;)

Koenkk:
> Home Assistant auto-discovery will be changed as the device type changes from switch to light (in order to support dimming)

I think we can easily avoid this by keeping the `e.switch()` in (next to the added light expose)

klada:
> I think we can easily avoid this by keeping the `e.switch()` in (next to the added light expose)

@Koenkk Of course, you are completely right. By just keeping `e.switch()` we won't mess with existing setups. I have also added this in my docs PR so it is clear to people why there is a `switch` **and** a `light` entity.

Tetsted and works in my Home Assistant instance, so this is ready for merging from my side.

The question is if you want to get rid of the (wrong) switch definition at any time or if you'd rather keep it forever.

PR #5301, opened 2023-01-10
philips: add perfio lights

zegl:
Added the light from the open issue, plus the other different models from the same series that seems to have the same specs.

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/16079


Koenkk:
Thanks!

Issue #4116, opened 2022-04-10
No converter available for 'HG08164'


kvakulo:
I have paired a Lidl HG08164 smart button with zigbee2mqtt v. 1.25.0 and it looks like it should be supported according to https://www.zigbee2mqtt.io/devices/HG08164.html.

However, I'm getting these errors when pressing the button:
```
Received Zigbee message from 'Lidl Smart Button', type 'raw', cluster 'genOnOff', data '{"data":[1,66,253,0],"type":"Buffer"}' from endpoint 1 with groupID 0
```

So I guess some adjustment of the converter might be needed in https://github.com/Koenkk/zigbee-herdsman-converters/blob/3743cea45af1494ef6e1355ca5b7b70304f07eed/devices/lidl.js#L382. 

However I'm not sure how best to play around as I'm running zigbee2mqtt as a homeassistant add-on.

Please let me know how if I can do anything to help or I can somehow play around with the converter inside homeassistant 😸

kvakulo:
Alright, I have made a bit of progress.

1) I have copied `lidl.js` into `/config/zigbee2mqtt/` and added `lidl.js` under `Settings -> External converters` in zigbee2mqtt.
2) Then I added `fz.tuya_on_off_action` to the `fromZigbee` list.
3) And then added `single` + `double` to the list of exposed actions.

```js
{
        fingerprint: [{modelID: 'TS004F', manufacturerName: '_TZ3000_rco1yzb1'}],
        model: 'HG08164',
        vendor: 'Lidl',
        description: 'Silvercrest smart button',
        fromZigbee: [fz.command_on, fz.command_off, fz.command_step, fz.command_stop, fz.battery, fz.tuya_on_off_action],
        toZigbee: [],
        exposes: [e.action(['on', 'off', 'brightness_stop', 'brightness_step_up', 'single', 'double']), e.battery()],
    },
```

This makes single and double presses work.

ghhassta:
Officially fixed as of release [1.30.0](https://github.com/Koenkk/zigbee2mqtt/releases/tag/1.30.0).

PR #5305, opened 2023-01-11
HG08164 fix

ghhassta:
Made the necessary changes proposed here [https://github.com/Koenkk/zigbee-herdsman-converters/issues/4116] to allow this device to work correctly.  Have tested in my own local instance and this does indeed make the formerly non-working button work correctly.  I don't seem to get the brightness controls, but at least I can now use the single and double click actions.


Koenkk:
Thanks!

PR #5308, opened 2023-01-11
Support for Tuya 2 gang switch module

Paalap:
Full support for modelID: TS0002 manufacturerName: _TZ3000_zmy4lslw instead of generic setup

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1941,6 +1941,30 @@ module.exports = [
         },
     },
     {
+        fingerprint: [{modelID: 'TS0002', manufacturerName: '_TZ3000_zmy4lslw'}],
+        model: 'TS0002_switch_module',
+        vendor: 'TuYa',
+        description: '2 gang switch module',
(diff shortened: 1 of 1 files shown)


devices/tuya.js
```diff
@@ -1940,6 +1940,30 @@ module.exports = [
             await reporting.bind(device.getEndpoint(1), coordinatorEndpoint, ['genOnOff']);
         },
     },
+    {
+        fingerprint: [{modelID: 'TS0002', manufacturerName: '_TZ3000_zmy4lslw'}],
```
Koenkk, line 1944: This device is already supported: https://github.com/Koenkk/zigbee-herdsman-converters/blob/328ce185b1d74a57b23c2c925682aa3bb654576f/devices/tuya.js#L1975

Paalap: In my version we have also power-on behavior and switch type setting (quite important for me)

Koenkk: This is also supported (when using z2m 1.29.1)

PR #5310, opened 2023-01-11
Updated Datek sensors & new converter to read HW version

TheStigh:
- Added fingerprint to all except 1 sensor from Datek
Reason are that all their products are now rebranded to Eva

- Added new converter `fz.hw_version`
Product `HLU2909K` has been revised, only possible way to see difference are the HW version

- Updated HLU2909K to show HW version if `hasOwnProperty` has value


Koenkk:
Thanks!

PR #5312, opened 2023-01-12
Add more philips effect lights

deviantintegral:
This PR adds candle effect support for additional lights.

I have confirmed this works on:

* LTA008
* LWA003
* LWS002 - while this shows up in the Hue app, and the bulb changes brightness, to my eyes it doesn't really animate in the same way as other lights. Since the Hue app shows it, I think we should too.

I don't have LWW002 to test with. Since it's the same definition and title as LWA003 I assume it works and it's just a later revision?

As well, for testing this with an external converter required a ton of copy-paste. Would we consider exporting `hueExtend`?

Finally, I'm not sure if Philips makes lights that are a "bright white" color temperature only. If so, I assume they don't support Candle in which case I think we'd want to change what's `light_onoff_brightness` to something like `light_onoff_warm_brightness`.


Koenkk:
> I don't have LWW002 to test with. Since it's the same definition and title as LWA003 I assume it works and it's just a later revision?

That's fine!

> As well, for testing this with an external converter required a ton of copy-paste. Would we consider exporting hueExtend?


Yes please do so in a new `lib/philips.js` file just like `xiaomi.js` and `tuya.js`. 

Thanks for your contributions again!

lib/philips.js
```diff
@@ -249,4 +659,9 @@ function encodeGradientColors(value, opts) {
 module.exports = {
     decodeGradientColors,
     encodeGradientColors,
+    hueExtend,
```
Koenkk, line 662: Since this is in `philips.js`, we can rename this to just `extend`. (we do the same for TuYa)

devices/philips.js
```diff
@@ -3,987 +3,588 @@ const fz = {...require('../converters/fromZigbee'), legacy: require('../lib/lega
...
-const utils = require('../lib/utils');
-const libColor = require('../lib/color');
-const herdsman = require('zigbee-herdsman');
 const e = exposes.presets;
 const ea = exposes.access;
 
-// Make sure extend.light_* is not used (hueExtend should be used instead)
+// Make sure extend.light_* is not used (philips.hueExtend should be used instead)
 const extendDontUse = require('../lib/extend');
 const extend = {switch: extendDontUse.switch};
```
Koenkk, line 12: `const extendDontUse = require('../lib/extend');` can be removed, when changing `onst extend = {switch: extendDontUse.switch};` to `onst extend = {switch: require('../lib/extend').switch};`

lib/philips.js
```diff
@@ -249,4 +659,9 @@ function encodeGradientColors(value, opts) {
...
+    tzLocal,
+    fzLocal,
```
Koenkk, line 664: `Local` can be dropped (just like TuYa): https://github.com/Koenkk/zigbee-herdsman-converters/blob/75db090d32280e21caa8c02957488b16faefda31/lib/tuya.js#L1814

deviantintegral:
Thanks for the review. All done!

Koenkk:
Looks great, thanks!

PR #5313, opened 2023-01-12
Add SIN-4-1-20_PRO and SIN-4-2-20_PRO

Char-r:
Same as SIN-4-1-20 and SIN-4-2-20 converters just with a different name.


Koenkk:
Thanks!

PR #5316, opened 2023-01-12
Aqara SRTS-A01 fix missing OTA cluster

### LINKED ISSUE #5304: Aqara SRTS-A01 missing genOta output cluster
This Aqara device does not automatically configure the `genOta` output cluster, which blocks OTA firmware updates. It is possible to manually add it into the `database.db` file to run an update as documented in https://github.com/Koenkk/zigbee2mqtt/issues/13993#issuecomment-1346644669.

Is there a way to explicitly add this cluster, e.g. in `Definition.configure`?

protyposis:
Fix missing OTA update functionality by adding the missing `genOta` cluster as suggested in https://github.com/Koenkk/zigbee2mqtt/issues/13993#issuecomment-1346644669. Closes #5304.

Due to a lack of understanding how Z2M works, I do not know whether this is a proper fix or if the lack of automated discovery of this cluster hints at a deeper issue which should rather be fixed instead. I hope a reviewer can clear this up.

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -2696,6 +2696,9 @@ module.exports = [
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await endpoint.read('aqaraOpple', [0x040a], {manufacturerCode: 0x115f});
+
+            // This cluster is not discovered automatically and needs to be explicitly attached to enable OTA
+            utils.attachOutputCluster(device, 'genOta');
         },
         ota: ota.zigbeeOTA,
     },
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const globalStore = require('./store');
+const herdsman = require('zigbee-herdsman');
 
 function isLegacyEnabled(options) {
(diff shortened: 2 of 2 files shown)


Koenkk:
> Due to a lack of understanding how Z2M works, I do not know whether this is a proper fix or if the lack of automated discovery of this cluster hints at a deeper issue which should rather be fixed instead.

To double check this, can you provide the herdsman debug log when pairing this device? Before doing that, force remove the device through the z2m frontend and restart z2m.

See https://www.zigbee2mqtt.io/guide/usage/debug.html on how to enable the herdsman debug logging. Note that this is only logged to STDOUT and not to log files.

protyposis:
What sticks out to me is
```
zigbee-herdsman:deconz:adapter RECEIVING SIMPLE_DESCRIPTOR - addr: 0xa61d EP:1 inClusters: 0,1,3,64704,10,513 outClusters: 3,64704,513
```

Here is the full log file (it might contain logs related to other devices too): [aqara-trv-pairing-herdsman-debug.log](https://github.com/Koenkk/zigbee-herdsman-converters/files/10424872/aqara-trv-pairing-herdsman-debug.log)

Koenkk:
So the device doesn't advertise of the OTA cluster, then this fix is fine.

PR #5318, opened 2023-01-13
Extend support for Moes MS-105B

volodymyrpavlenko:
* Use generic tuya datapoints accessors in Moes MS-105B
* Add support for new attributes:
  * `min_brightness` 
  * `light_type`


lib/tuya.js
```diff
@@ -1250,6 +1245,7 @@ const valueConverter = {
     onOff: valueConverterBasic.lookup({'ON': true, 'OFF': false}),
     powerOnBehavior: valueConverterBasic.lookup({'off': 0, 'on': 1, 'previous': 2}),
     lightType: valueConverterBasic.lookup({'led': 0, 'incandescent': 1, 'halogen': 2}),
+    lightTypeToEnum: valueConverterBasic.lookup({'led': new Enum(0), 'incandescent': new Enum(1), 'halogen': new Enum(2)}),
```
volodymyrpavlenko, line 1248: The line above is possibly an existing bug, unless other tuya devices expect the value to be numeric.
In case of MS-105B, the `light_type` should be sent as enum.

volodymyrpavlenko:
Tested the change locally with help of an external converter

converters/toZigbee.js
```diff
@@ -6583,37 +6583,11 @@ const converters = {
...
-                    await tuya.sendDataPointValue(entity, brightnessKeyId, newValue, 'dataRequest', 1);
-                    break;
-                } else {
-                    throw new Error('Dimmer brightness is out of range 0..254');
-                }
-
-            default:
-                throw new Error(`Unsupported Key=[${key}]`);
+            if (key === 'brightness') {
+                meta.message['state'] = value > 0 ? 'ON' : 'OFF';
```
Koenkk, line 6589: Please create a new tuya brightness value converter with this logic (such that only `tuya.tz.datapoints` is needed)

volodymyrpavlenko: Thanks for the comment!

Maybe I misunderstood you, do you suggest that the value converter will modify thee `meta.message` state? From a readability point of view, will it be an unexpected side effect of the value converter?
Will it even work if we modify the `meta.message` during the iteration?
```
for (const [key, value] of Object.entries(meta.message)) {
```

I don't have much experience with JS, but in Java this would lead to a `ConcurrentModificationException`.

Koenkk: No, the value converter can also return other properties, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/61055d1fb479e7669695ff284c1dad2458366ea4/lib/tuya.js#L1276

volodymyrpavlenko: I believe this is only implemented properly in the fromZigbee implementation.

According to the logic in the tuja toZigbee implementation, there's no support for converters expanding the original message.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/adb17702ee9a52746e850adcfaf3182171e92d43/lib/tuya.js#L1596-L1613

Maybe it's a good idea to add it, but would it make sense to separate this trivial change from a refactoring?

I might be missing some obvious things :)

volodymyrpavlenko: Or do you recommend that a valueConverter does a side action of sending another command? 
E.g. as here
https://github.com/Koenkk/zigbee-herdsman-converters/blob/adb17702ee9a52746e850adcfaf3182171e92d43/lib/tuya.js#L1478-L1493

volodymyrpavlenko: Something like 
```
            // A set converter is only called once; therefore we need to loop
            const state = {};
            if (!meta.mapped.meta || !meta.mapped.meta.tuyaDatapoints) throw new Error('No datapoints map defined');
            const datapoints = meta.mapped.meta.tuyaDatapoints;
// A queue, where converters can add values by returning an object
            const messageEntries = Object.entries(meta.message);
            for (let i = 0; i < messageEntries.length(); i++) {
                const [key, value] = messageEntries[i];
                const convertedKey = meta.mapped.meta.multiEndpoint ? `${key}_${meta.endpoint_name}` : key;
                const dpEntry = datapoints.find((d) => d[1] === convertedKey);
                if (!dpEntry || !dpEntry[1]) {
                    throw new Error(`No datapoint defined for '${key}'`);
                }
                if (dpEntry[3] && dpEntry[3].skip && dpEntry[3].skip(meta)) continue;
                const dpId = dpEntry[0];
                const preValue = await dpEntry[2].to(value, meta);
                let convertedValue = undefined;
                if (typeof preValue === 'object') {
                    if (preValue.hasProperty(key)) {
                        convertedValue = preValue[key];
                        delete preValue[key];
                    }
                    messageEntries.push(...Object.entries(preValue));
                }
                if (convertedValue === undefined) {
                    // conversion done inside converter, ignore.
                } else if (typeof convertedValue === 'boolean') {
                    await sendDataPointBool(entity, dpId, convertedValue, 'dataRequest', 1);
                } else if (typeof convertedValue === 'number') {
                    await sendDataPointValue(entity, dpId, convertedValue, 'dataRequest', 1);
                } else if (typeof convertedValue === 'string') {
                    await sendDataPointStringBuffer(entity, dpId, convertedValue, 'dataRequest', 1);
                } else if (Array.isArray(convertedValue)) {
                    await sendDataPointRaw(entity, dpId, convertedValue, 'dataRequest', 1);
                } else if (convertedValue instanceof Enum) {
                    await sendDataPointEnum(entity, dpId, convertedValue.valueOf(), 'dataRequest', 1);
                } else if (convertedValue instanceof Bitmap) {
                    await sendDataPointBitmap(entity, dpId, convertedValue.valueOf(), 'dataRequest', 1);
                } else {
                    throw new Error(`Don't know how to send type '${typeof convertedValue}'`);
                }
                state[key] = value;
            }
            return {state};
        },
```
?

Koenkk: I'm wondering now, why is this needed in the first place? None of the other TuYa devices need this.

volodymyrpavlenko: This is the behavior that already existed. Basically, when you set brightness to 0, the plugin will also set the state of the light to off. Otherwise you'll see it as "ON" with brightness 0. I believe it makes sense, as other dimmers, e.g. Schneider, do so.

Koenkk: I believe the device should report its new state immediately after setting it, is this not the case? (I believe it is the case with other TuYa devices)

volodymyrpavlenko: Yes, device status does report it's status back, but MS-105B does not switch off the light once brightness is set to 0. 

I have just manually tested removing the custom code with my local MS-105B.
Original state:
image
After changing brightness to 0:
image

Notice that the state is still "ON". 
At the same time, the actual lamp, connected to the dimmer, is still on  at a very low brightness.

Please see the log below:
```
Info 2023-01-18 22:18:58MQTT publish: topic 'zigbee2mqtt/Office Wall Switch', payload '{"brightness_l1":0,"brightness_l2":18,"last_seen":"2023-01-18T21:18:57.180Z","light_type_l1":"incandescent","light_type_l2":"halogen","linkquality":54,"min_brightness_l1":0,"min_brightness_l2":1,"state_l1":"ON","state_l2":"ON"}'
Info 2023-01-18 22:18:58MQTT publish: topic 'zigbee2mqtt/Office Wall Switch/l1', payload '{"brightness":0,"light_type":"incandescent","min_brightness":0,"state":"ON"}'
Info 2023-01-18 22:18:58MQTT publish: topic 'zigbee2mqtt/Office Wall Switch/type_l1', payload '{"light":"incandescent"}'
Info 2023-01-18 22:18:58MQTT publish: topic 'zigbee2mqtt/Office Wall Switch/l2', payload '{"brightness":18,"light_type":"halogen","min_brightness":1,"state":"ON"}'
Info 2023-01-18 22:18:58MQTT publish: topic 'zigbee2mqtt/Office Wall Switch/type_l2', payload '{"light":"halogen"}'
Debug 2023-01-18 22:18:58Received Zigbee message from 'Office Wall Switch', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,0,0],"type":"Buffer"},"datatype":2,"dp":2}],"seq":17671}' from endpoint 1 with groupID 0
Info 2023-01-18 22:18:58MQTT publish: topic 'zigbee2mqtt/Office Wall Switch', payload '{"brightness_l1":0,"brightness_l2":18,"last_seen":"2023-01-18T21:18:58.385Z","light_type_l1":"incandescent","light_type_l2":"halogen","linkquality":57,"min_brightness_l1":0,"min_brightness_l2":1,"state_l1":"ON","state_l2":"ON"}'
```

volodymyrpavlenko: Also, this is not a new logic, same logic is present in master:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/converters/toZigbee.js#L6601-L6613

Koenkk: I see now, the only goal of this change is to turn the light OFF when you set brightness 0?

lib/tuya.js
Koenkk: I think others reported it works fine, so let's leave it as is.

converters/toZigbee.js
Koenkk: I moved the logic to the value converter, please apply the changes in this pr and check if it works. Especially check the scenario where the light is OFF and you turn it ON by just sending the brightness.

volodymyrpavlenko: I have just tested, and without an explicit "ON" command, the switch won't turn on on changing just a brightness.
```javascript
    brightness: (stateDataPoint) => {
        return {
            from: tuya.valueConverter.scale0_254to0_1000.from,
            to: async (v, meta) => {
                const entity = meta.device.endpoints[0];
                await tuya.sendDataPointBool(entity, stateDataPoint, v > 0, 'dataRequest', 1);
                return tuya.valueConverter.scale0_254to0_1000.to(v, meta);
            },
        };
    },
```

volodymyrpavlenko: I also found that in this scenario:
- set brightness to 0 (with turning off)
- turn on
result is brightness == 0, but not the previous value as with other switches. What do you think is the desired behavior?

Koenkk: There is a use case for both. 

A feature of not turning the device ON when you send brightness > 0 is that you can change the brightness before the bulb turns on. Imagine a scenario in the night where the light was previously at brightness 100%, now you want to turn it ON with just brightness 5%, this is not possible with the current code (as it will flicker).

What is the reason you cannot send a `state` with every `brightness` message? e.g. Home Assistant does this by default.

volodymyrpavlenko: As far as I know, Tuya does not support sending multiple attributes in the same message. It uses own datapoints message instead, where you can set one attribute in each message only.
MS-105b are rather slow to switch on, so I believe switching on before setting brightness might be not an issue in practice, but I will manually test it tomorrow.

volodymyrpavlenko: I tested it manually with the following actions:
- set brightness to max
- switch off
- Switch on by setting brightness to min

the light didn’t blink visually

Koenkk: Great, then please check if everything works with:

```js
                // When brightness 0 is set, the light does not turn off, force it here.
                if (v === 0) {
                    const entity = meta.device.endpoints[0];
                    await sendDataPointBool(entity, stateDataPoint, false, 'dataRequest', 1);
                } else if (meta.state.state === 'OFF') {
                    const entity = meta.device.endpoints[0];
                    await sendDataPointBool(entity, stateDataPoint, true, 'dataRequest', 1);
                }
                return valueConverter.scale0_254to0_1000.to(v, meta);
```

volodymyrpavlenko: No, doesn't work. `meta.state` has endpoint-based state variable. So we'd need something like this:
```javascript
                // When brightness 0 is set, the light does not turn off, force it here.
                if (v === 0) {
                    const entity = meta.device.endpoints[0];
                    await tuya.sendDataPointBool(entity, stateDataPoint, false, 'dataRequest', 1);
                } else if (meta.state[stateStateName] === 'OFF') {
                    const entity = meta.device.endpoints[0];
                    await tuya.sendDataPointBool(entity, stateDataPoint, true, 'dataRequest', 1);
                }
                return tuya.valueConverter.scale0_254to0_1000.to(v, meta);
```

volodymyrpavlenko: By the way, I decided to return the dimmer since it has very poor dimming capability, but I have it at home until tomorrow to test

Koenkk: I see, maybe we should address this in a future PR and support it like all other TuYa dimmers, agree?

volodymyrpavlenko: I believe it may be useful for others to have the added functionality for the minimum brightness, since this dimmer does not work well on the minimum level. 

I tested the last code snippet above, it worked fine. Don't think that MS-105B works as other TuYa dimmers, since the behavior is different, and the existing code in the master branch is doing extra state manipulations when brightness changes.

But I agree that without being able to test the merged release, we can introduce bugs.

PR #5319, opened 2023-01-13
Updated time units for BRT-100-TRV (TS0601)

tefracky:
Made time units for BRT-100-TRV (TS0601) compatible with home assistant.

This should solve the error `Entity sensor...._boost_heating_countdown (<class 'homeassistant.components.mqtt.sensor.MqttSensor'>) is using native unit of measurement 'Min' which is not a valid unit for the device class ('duration') it is using; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/home-assistant/core/issues?q=is%3Aopen+is%3Aissue+label%3A%22integration%3A+mqtt%22`


Koenkk:
Thanks!

PR #5324, opened 2023-01-14
Adding NUE-ZBFLB 'LXX60-FN27LX1.0'] device

fredskis:
New device being added, has the same config as 'LXN56-1S27LX1.2'


fredskis:
Will merge both definitions into one - sorry, new to the syntax and unique requirement for models (thought it was zigbeeModel that would be unique)

PR #5327, opened 2023-01-14
Handle `hvacFanCtrl` messages without `fanMode` data

davidjb:
This handles `hvacFanCtrl` messages where `fanMode` isn't present, such as `{"fanModeSequence":1}` (such as that generated by the Mercator SSWF01G fan controller, reported in https://github.com/Koenkk/zigbee2mqtt/issues/15186). Previously, without `fanMode` present in the message, the state was assumed to be on given the lookup implementation.

Section 6.4.2.2 of the ZCL mentions that `FanMode` and `FanModeSequence` should be mandatory but at least the SSWF01G sends these attributes in separate messages.


Koenkk:
Thanks!

PR #5336, opened 2023-01-15
Fix Lidl 14156408L supporting colors

MichelFR:
The Lidl 14156408L does not support colors as it only supports tempeture between 153 and 500.

More information:
https://www.lidl.de/p/livarno-home-deckenleuchte-mit-lichtfarbensteuerung-zigbee-smart-home/p100339376


Koenkk:
Thanks!

PR #5345, opened 2023-01-16
Add mains power to NUE-ZBFLB

fredskis:
Device showed up with "?" as power source. 
Changed config to display as connected to mains power


Koenkk:
Thanks!

PR #5334, opened 2023-01-15
current_heating_setpoint fix for _TZE200_5toc8efa

VoltickVL:
current_heating_setpoint that is 10 time lower than expected

PR #5348, opened 2023-01-17
current_heating_setpoint that is 10 time lower than expected

VoltickVL:
When set "current_heating_setpoint" to 10 *C, get 1 *C at device and etc...


Koenkk:
Please check my comment from https://github.com/Koenkk/zigbee-herdsman-converters/pull/5334/files#r1070590325

PR #5349, opened 2023-01-17
Error with setting the temperature in "current_heating_setpoint" that…

VoltickVL:
… is 10 time lower than expected

Bugfix error with setting the temperature in "current_heating_setpoint" that is 10 time lower than expectedIf set "current_heating_setpoint" to 20 *C at Z2M -> get 2 *C on device and etc...


VoltickVL:
https://github.com/Koenkk/zigbee2mqtt/issues/16044

PR #5352, opened 2023-01-17
Fix calibration limits of BTH-RA

Master92:
While using the BetterThermostat custom component, I kept getting errors during its calibration procedure with my Bosch Heizkörper-Thermostat II. Checking the TRV's capabilities in the Bosch app, I discovered it can only calibrated in a range from -5 to +5°C. Further, the converter only allowed steppings of .5°C, while really .1°C is possible.

See attached screenshots of the Bosch app for reference.

bosch


Koenkk:
Thanks!

PR #5356, opened 2023-01-17
Improve Lidl Parkside Watering Timer Support

Big4SMK:
This is adding the code developed in https://github.com/Koenkk/zigbee2mqtt/issues/7695 thereby adding a lot of functionality for the watering meter.

Code is created by @zinserjan, changed by @moritzbeck01 to change the frost guard switch to a sensor and a button. I confirmed the frost guard sensor and reset to work by freezing the device for a little bit so I think this code is ready to move from an external converter into mainline.

Let me know if I'm missing anything that still needs to be done.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -2551,12 +2551,6 @@ const converters = {
             }
         },
     },
-    lidl_watering_timer: {
-        key: ['timer'],
-        convertSet: async (entity, key, value, meta) => {
-            await tuya.sendDataPointRaw(entity, tuya.dataPoints.lidlTimer, tuya.convertDecimalValueTo4ByteHexArray(value));
-        },
-    },
     matsee_garage_door_opener: {
         key: ['trigger'],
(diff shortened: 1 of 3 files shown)


Big4SMK:
Sure thing, I hope to find some time within the next week or so.

Big4SMK:
Still planning on building this out

Big4SMK:
I've spent a couple of hours refactoring but couldn't get things to work. Maybe at some point someone with more knowledge of the tuya code cares enough to convert the current code to the new standard.

d0np3p3:
Okay thanks for the try, shouldn't the ticket kept open, until it works?

DenDeXTeR51:
Hi! When this code (additional fetaures) will be added to master branch? or maybe there is external converter so I can use it separatelly?

Big4SMK:
The converter can be found in the zigbee2mqtt topic, more specifically here: https://github.com/Koenkk/zigbee2mqtt/issues/7695#issuecomment-1086721715

PR #5358, opened 2023-01-17
Add philips floor lamp 929003516101

rabb13:
Another variant of Philips hue Signe gradient floor lamp wood/oak 
Model `929003516101` is listed here-  https://www.philips-hue.com/en-gb/p/hue-white-and-colour-ambiance-signe-gradient-floor-lamp/8719514445291#specifications

Similar to model 929003479701, tested working with custom converter.


Koenkk:
thanks!

PR #5361, opened 2023-01-18
Fix Sinope SP2600ZB & SP2610ZB energy const

MattL0:
Transfert to fzLocal

### DIFF
--- a/devices/sinope.js
+++ b/devices/sinope.js
@@ -7,7 +7,7 @@ const reporting = require('../lib/reporting');
 const extend = require('../lib/extend');
 const e = exposes.presets;
 const ea = exposes.access;
-const {precisionRound} = require('../lib/utils');
+const {precisionRound, calibrateAndPrecisionRoundOptions} = require('../lib/utils');
 
 const manuSinope = {manufacturerCode: 0x119C};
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Why is this needed?

MattL0:
This is to fix this issue.

https://github.com/Koenkk/zigbee2mqtt/issues/15950

There is something with the generic semetering converter that makes this report null. Divider and power is both 1 for this plug. 
Maybe this has something to do with Sinope only reporting ont he second integer ex : [0, 33625253] or [0, 23]
Also, the plug is reporting in w/h . So i devided to total by 1000

Koenkk:
The only reason this converter works and the official one doesn't is the factor. I see the metering divisors are not read which might cause this. Can you add `await reporting.readMeteringMultiplierDivisor(endpoint);` to the configure of both devices?

MattL0:
I'm closing this

MattL0:
> The only reason this converter works and the official one doesn't is the factor. I see the metering divisors are not read which might cause this. Can you add `await reporting.readMeteringMultiplierDivisor(endpoint);` to the configure of both devices?

aaa... i should buy glasses.... lol thanks aha

PR #5363, opened 2023-01-18
Aqara SRTS-A01 heartbeat decoding (fixes battery percentage and firmware version)

protyposis:
Add decoding of heartbeat messages (sent ~once per hour) of the Aqara TRV. This adds a few new properties, increases the update frequency of some, and fixes a few related issues. This PR is inspired by #4763 which proposed basic heartbeat decoding.

 * Add exposes of `setup` state, `power_outage_count`, and `device_temperature`.
 * Expose hidden `valve_alarm`.
 * Add description to `valve_detection` expose.
 * Improve description of `local_temperature` expose.
 * Remove the battery percentage-from-voltage calculation introduced by #4674. The device reports the percentage directly and more precisely. The calculation used a wrong scale where almost full batteries were classified as almost empty, e.g. 2.9V were converted to 33%, and it was based on the imprecisely calculated voltage in 100mV steps. This also fixes inconsistent battery percentages that jump between the reported percentage and the calculated percentage, e.g. jumping between 95% and 33% at 2.9V.
 * Update the device firmware version to the currently installed version instead of using the default "placeholder" version `0.0.0_0025` advertised by the basic cluster (which doesn't update with firmware updates).
 * Initialize the battery voltage at device configuration, which was initially added in #4674 but replaced in #4754 by the voltage conversion.


devices/xiaomi.js
```diff
@@ -114,9 +115,23 @@ const fzLocal = {
...
                     break;
+                case 247: {
+                    const heartbeat = xiaomiTrvUtils.decodeTrvHeartbeat(meta, model, value);
+
+                    meta.logger.debug(`${model.zigbeeModel}: Processed heartbeat message into payload ${JSON.stringify(heartbeat)}`);
+
+                    // Overwrite device firmware default version with actual version
+                    if (heartbeat.firmware_version) {
+                        // TODO How to trigger update in UI?
+                        meta.device.softwareBuildID = heartbeat.firmware_version;
```
protyposis, line 126: @Koenkk is there any way to trigger an update of the UI, e.g. programmatic reconfiguration? With the current implementation the firmware version label only updates after a restart or reconfiguration. Unfortunately this in not directly available/accessible in `configure` because the heartbeat message cannot be requested on demand.

Koenkk: Is the `softwareBuildID` different from this `firmware_version`?

protyposis: Yes, as written in the PR description.

protyposis: The 'firmware_version` here is the _real_ version that is actually installed.

Koenkk: Currently this is not possible and it will require some work to get this working. I guess you are talking about the `softwareBuildID` in the frontend which comes from `bridge/devices` (for this https://github.com/Koenkk/zigbee2mqtt/blob/bb3e8f6d414f126cd7f57686e0e9798f285eb78e/lib/extension/bridge.ts#L612 has to be called).

Since firmware updating is not done frequently, I feel it acceptable to state in the docs that you have to restart z2m some time after the update to get the actual version.

lib/xiaomiTrvUtils.js
```diff
@@ -0,0 +1,79 @@
+const xiaomi = require('./xiaomi');
```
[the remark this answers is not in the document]
protyposis: 9c00c0a

Koenkk:
clean, thanks!

devices/xiaomi.js
leroyloren: I updated the edge with this commit, the latest firmware version is showing up fine, but the OTA keeps offering to update to the new version, is there any way to cancel this so that the OTA is not offered?

leroyloren:
@protyposis Could you please fix the version prefix? With the new FW 1030 the version is displayed as 0.0.0_01030

obrazek

Otnow:
> @protyposis Could you please fix the version prefix? With the new FW 1030 the version is displayed as 0.0.0_01030
> 
> obrazek

In [PR](https://github.com/Koenkk/zigbee-herdsman-converters/pull/6983), among other things, formatting of firmware version has been fixed.

PR #5364, opened 2023-01-18
Aqara feeder C1 local time

kirovilya:
Aqara feeder C1 polls the time during the interview, need to send back the local time instead of the UTC.
At this moment, the device.definition has not yet been set - therefore the onEvent method does not work.

https://github.com/Koenkk/zigbee2mqtt/discussions/16157


Koenkk:
💯

PR #5367, opened 2023-01-19
Add Aqara Cube T1 Pro CTP-R01

JJPro:
## External Converter for Temporary Use 

For folks want to use the cube right now, use this [external converter](https://github.com/JJPro/CTP-R01-converter). 
This PR will stay open for a while for reason described [below](#issuecomment-1397441503). 

Please use the external converter and provide feedbacks. 

Thank You!

### 1. write to `0xffc0.0x0009` (aqaraOpple.mode)

```js
  ...
  // attr: 0x0009
  await endpoint.write('aqaraOpple', { mode: 1 }, manufacturerOptions.xiaomi);
  ...
}
```
This line is borrowed from other Xiaomi converters.

I can't find documentation on what `0x0009` (`mode` attribute) on `0xffc0` (`aqaraOpple` cluster) does, but most xiaomi remotes & switches have it in their converter definition. So I left it in for now, and it doesn't seem to cause any problems during my experiment. But Aqara's own hub never writes to this section during the pairing process. I'll probably remove this piece in official release. 

### 2. write to `0xffc0.0x00ff` (aqaraOpple.0x00ff)

```js
  ...
  await endpoint.write('aqaraOpple',
    {
      0x00ff: {
        value: [0x45, 0x65, 0x21, 0x20, 0x75, 0x38, 0x17, 0x69, 0x78, 0x53, 0x89, 0x51, 0x13, 0x16, 0x49, 0x58],
        type: 0x41
      }
    }
    , manufacturerOptions.xiaomi
  );
  ...
}
```

I observed this behavior from sniffing the pairing process with Aqara Hub M2. 

It wrote a different value for each pairing experiment. I randomly picked one to use. Not sure what it does though. 

## Sniff Logs

[Sniff logs are posted here](https://github.com/JJPro/CTP-R01-converter/tree/trunk/sniffs). Help is welcomed. 

---
resolves Koenkk/zigbee2mqtt#15652


Koenkk:
For both both, try if it works fine without (and make sure to factory reset after removing them), if yes, leave them out.

Note to self: after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1833

JJPro:
I can't verify that. 

My device is paired to Aqara hub prior to Zigbee2MQTT, probably Aqara hub did initialization stuff to it, so I never came across a configuration issue that some other folks had with my first converter attempt without the `configure` section ([shared](https://github.com/Koenkk/zigbee2mqtt/issues/15652#issuecomment-1367839008) about 3 weeks ago).

My device works fine without them, but I don't know if it's the same for others who got their devices fresh unpaired. I can't find a way to factory reset the device and this is the only cube I have. 

An external converter version of this PR is [posted](https://github.com/Koenkk/zigbee2mqtt/issues/15652#issuecomment-1396897544). 
I'll wait for others to try and provide feedback. 

Let's keep this PR open until we collect enough positive feedbacks from the community.

JJPro:
What's the legacy toggle for under settings(specific)? 
I see a lot of devices have this toggle, and they output the same values under different attribute names. 

E.g. for [MFKZQ01LM](https://www.zigbee2mqtt.io/devices/MFKZQ01LM.html#xiaomi-mfkzq01lm)
- legacy disabled: `{...action_from_side: 4...}`
- legacy enabled: `{...from_side: 4...}`

Koenkk:
@JJPro you can leave out the whole legacy part for new (read: this) converter. It is meant to prevent a breaking change.

JJPro:
@Koenkk hey you forgot to attach the link

Koenkk:
@JJPro my sentence was wrong, what I meant: legacy was added to prevent a breaking change when changing the `action` format of already supported devices. Since you are adding a new device, you can leave out the legacy part.

JJPro:
I see. thx

Koenkk:
@JJPro let me know once this pr is ready for review.

JJPro:
Of course, I'll ping you when it's ready. thx

JJPro:
@Koenkk for `defaultSendRequestWhen`, what's the diff between `active` vs `fastpoll`?

Koenkk:
`active` will send when the device becomes active (sends a message to Zigbee2MQTT), `fastpoll` only sends the message when the device send the poll control cluster fastpoll message. Given this is Xiaomi device, I think `active` is a better choice.

JJPro:
I'm trying to manually read attributes(EP: 1, cluster: aqaraOpple[0xfcc0], attribute: 0x148 [328]) through dev console: 
image

And I got this error: 
image

What am I doing wrong?

JJPro:
It's finally **READY** for review! 🎉

lib/xiaomi.js
```diff
@@ -142,10 +143,10 @@ const buffer2DataObject = (meta, model, buffer) => {
     return dataObject;
 };
 
-const numericAttributes2Payload = (msg, meta, model, options, dataObject) => {
+const numericAttributes2Payload = async (msg, meta, model, options, dataObject) => {
```
Koenkk, line 146: I would prefer to not make this method async, instead of `await callback();` can we use `.then`/`.catch`?

devices/xiaomi.js
```diff
@@ -2991,4 +3010,55 @@ module.exports = [
...
+            e.angle('action_angle'),
+        ],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            device.softwareBuildID = `0.0.0_00${device.applicationVersion}`;
+            device.save();
+
+            const endpoint = device.getEndpoint(1);
+            await endpoint.read('aqaraOpple', [0x148], {manufacturerCode: 0x115f, disableResponse: true});
+
+            setTimeout(() => logger.info('battery info takes longer to show up, ' +
```
Koenkk, line 3060: This is true for most battery powered device, I don't think we should show a message for this case.

converters/fromZigbee.js
```diff
@@ -6151,6 +6182,18 @@ const converters = {
             return result;
         },
     },
+    CTPR01_action_analog: {
```
Koenkk, line 6185: Could you move these into `devices/xiaomi.js`/`fzLocal`?

lib/xiaomi.js
JJPro: Sorry this is a bit lengthy. 

Unfortunately, in this case I do need the whole function to wait for the callback to finish. Otherwise, if we wait on a local level (`case '247'` in this case): 
```js
case '247':
    {
        const dataObject247 = buffer2DataObject(meta, model, value);
        if (['CTP-R01'].includes(model.model)) {
            // execute pending soft switch of operation_mode, if exists
            const opModeSwitchTask = globalStore.getValue(meta.device, 'opModeSwitchTask');
            if (opModeSwitchTask) {
                const {callback, newMode} = opModeSwitchTask;
                callback()
                    .then((_) => {
                        payload.operation_mode = newMode;
                        globalStore.putValue(meta.device, 'opModeSwitchTask', null);
                    });
            } else {
                payload.operation_mode = {0: 'action_mode', 1: 'scene_mode'}[dataObject247[155]];
            }
        }
        const payload247 = numericAttributes2Payload(msg, meta, model, options, dataObject247);
        payload = {...payload, ...payload247};
    }
    break;
```
the result payload wouldn't be the same. The execution will continue without waiting, thus `operation_mode` isn't updated. 

`dataObject247` report is triggered by user "throw", only then the cube will respond to write command (the callback's doing), and we need to wait for the command to be delivered to update `operation_mode` to desired `newMode`. 
If we don't wait for it, the payload would not contain updated `operation_mode`,  and user interaction experience breaks. 

See this [demo](https://github.com/JJPro/CTP-R01-converter/wiki/Demos-%26-My-Flows#demo) for user interaction. 
See [more on throw](https://github.com/JJPro/CTP-R01-converter/wiki/throw) for the concept of how the behavior is implemented.

JJPro:
hey are `disableDefaultReponse` and `disableResponse` the same thing? 

if I have `disableDefaultResponse: true` in `meta` definition, did it mean that I can omit `disableResponse` and `disableDefaultResponse` in e.g. `write` command: 
```js
await endpoint.write('aqaraOpple', {mode: 1}, 
     {manufacturerCode: 0x115f, disableDefaultResponse: true, disableResponse: true});
```

alexlukas:
> My device works fine without them, but I don't know if it's the same for others who got their devices fresh unpaired. I can't find a way to factory reset the device and this is the only cube I have.

If this helps: I tried to setup my brand new cube today, added it to zigbee2mqtt, and realized that it doesn't work out of the box. Afterwards, I found this issue and your repository with the external converter; following the steps I added the converter and again added the cube, which worked fine!

Koenkk:
@JJPro no they are not the same. But it's better to not use any of these (only if you get errors)

JJPro:
@Koenkk 
Update: 
1. [remove battery notification on configure ](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5367#discussion_r1088082937)
2. [move stuff from fz lib to xiaomi fzLocal](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5367#discussion_r1088083415)
3. I would insist on changing `numericAttributes2Payload` function to async, for [reason described above](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5367#discussion_r1088389567).

Koenkk:
Great, 

3: I see, then keep it `async` but make sure it is called with `await` everywhere.

JJPro:
updated!

PR #5368, opened 2023-01-19
added IKEA PILSKOTT pendant lamp (SILVERGLANS driver)

GsakuL:
https://www.ikea.com/de/de/p/pilskott-led-haengeleuchte-smart-schwarz-30477878/

It just has the driver "as-is" inside

### DIFF
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -556,7 +556,9 @@ module.exports = [
         model: 'ICPSHC24-30-IL44-1',
         vendor: 'IKEA',
         description: 'SILVERGLANS IP44 LED driver for wireless control (30 watt)',
+        whiteLabel: [{vendor: 'IKEA', model: 'Pendant lamp WW', description: 'PILSKOTT LED pendant lamp'}],
         extend: tradfriExtend.light_onoff_brightness(),
+        meta: {turnsOffAtBrightness1: true},
     },
     {
         zigbeeModel: ['FLOALT panel WS 30x30'],
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -556,7 +556,7 @@ module.exports = [
         model: 'ICPSHC24-30-IL44-1',
         vendor: 'IKEA',
         description: 'SILVERGLANS IP44 LED driver for wireless control (30 watt)',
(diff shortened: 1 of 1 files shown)


GsakuL:
is this the best way? It would be good to have it show/mention 'PILSKOTT' on zigbee2mqtt.io/devices, so people looking for it don't have to guess what driver is inside

Koenkk:
Please add is a separate device with the correct model. In this way it will also appear on the supported devices page.

GsakuL:
a) i saw a lot of other drivers/lamps having `turnsOffAtBrightness1`, so I tested it, and it seems to be the case here too, so i added that

b) is see the 'no duplicate model' error. I'm sorry, but the thing is, this is actually the correct number. In the documentation/ref sheet (in my case https://www.ikea.com/de/de/manuals/pilskott-led-haengeleuchte-smart-schwarz__AA-2345298-2-2.pdf Page 4) it also says `Power Supply Unit Type: ICPSHC24-30-IL44-1` (the same model number as SILVERGLANS). So it can't be an entirely different device. 
Would it instead be possible to maybe use the 'white-label' feature for the Doc?

Koenkk:
I see, so it really identifies itself as such.

GsakuL:
so. I initially saw in my log `warn  2023-01-21 10:03:55: Device '0xcc86ecfffe4393c2' with Zigbee model 'Pendant lamp WW' and manufacturer name 'IKEA of Sweden' is NOT supported, please follow https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html`
So i followed the steps. I copied ikea.json, except only left Silverglans in export, since I knew (from installing and seeing the driver) that it was most likely that. First I just added the zigbeeModel, which works. But then I thought it would be nice to have a separate description to find it better (see above). But since duplicate 'model's are not allowed, I thought with 'whiteLabel' it would exactly do that. So i changed it to whiteLabel. I did this on my local instance, in the copied ikea.json. 
Now i wanted to test the 'T2030' model, but noticed, that the device page did not change. It would still say 'Pendant lamp WW'
Since the docs only imply that you can always add stuff the the external_converter, I never thought it would remember all of this.

tl;dr 
1) whiteLabel did not do, what I excpected. Lamp does not work, if only whiteLabel given
2) model 'T2030'/'T2030 PILSKOTT' does not work. Only 'Pendant lamp WW'.

PR #5369, opened 2023-01-19
Add Bankamp device

ahemwe:
First of all, I'm sorry for providing a PR of such a bad quality. It's literally my very first.

I've followed the guide but struggled with introducing not only a new device but also a new vendor.

**Device**
https://www.bankamp.de/GRAZIA-Pendelleuchte/2189-1-92

The light changes it's color on it's own. (so called dim to warm technology (https://www.bankamp.de/technologie/dim-to-warm/)

With this configuration I am able to switch the light on and off and to change brightness.

**Power on behavior**
A try to change power on behavior had no effect but ended up with not receiving any data at all anymore.
After reconfiguration it works again.

Consequently, I'd like to remove the exposure of "power on behavior", but don't know how.

**Effect**
Effects work well.

**Brightness**
Lowest brightness value is 5, below the light is simply dark.

**Binding**
I've tried to bind the light to an Ikea dimmer, which was accepted by zigbee2mqtt but didn't work at all.

**Doc**
I'll open a PR for doc as well as soon as I can take a picture. Would it also be allowed to use a link to vendors website?


Koenkk:
Thanks!

PR #5372, opened 2023-01-20
current_heating_setpoint fix for _TZE200_5toc8efa

VoltickVL:
Bugfix error with setting the temperature in "current_heating_setpoint" that is 10 time lower than expected If set "current_heating_setpoint" to 20 *C at Z2M -> get 2 *C on device and etc...


Koenkk:
Thanks!

bartlebi:
Using this device i have x10 degrees bigger. When set 25 it looks like 250 in settings and etc.

Koenkk:
@bartlebi can you provide the data/database.db entry of this device?

PR #5373, opened 2023-01-20
Set ROB_200-016-0 to multiEndpoint

mdbraber:
Enable multiEndpoint by default for this device: https://github.com/Koenkk/zigbee2mqtt/issues/10236 - more info about the device https://www.robbshop.nl/robb-smarrt-rgbw-afstandsbediening-zigbee

### DIFF
--- a/devices/robb.js
+++ b/devices/robb.js
@@ -261,6 +261,10 @@ module.exports = [
             'color_move', 'color_temperature_move', 'hue_move', 'brightness_step_up', 'brightness_step_down',
             'recall_*', 'on', 'off', 'toggle', 'brightness_stop', 'brightness_move_up', 'brightness_move_down',
             'color_loop_set', 'enhanced_move_to_hue_and_saturation', 'hue_stop'])],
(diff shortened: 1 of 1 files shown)


devices/robb.js
```diff
@@ -261,6 +261,10 @@ module.exports = [
             'color_move', 'color_temperature_move', 'hue_move', 'brightness_step_up', 'brightness_step_down',
             'recall_*', 'on', 'off', 'toggle', 'brightness_stop', 'brightness_move_up', 'brightness_move_down',
             'color_loop_set', 'enhanced_move_to_hue_and_saturation', 'hue_stop'])],
+        meta: {multiEndpoint: true, publishDuplicateTransaction: true},
```
Koenkk, line 264: Does this change the exposed actions? e.g. `on` becomes `on_l1`?

mdbraber:
Yes it does - as there's no such thing as a 'standard' `on` action, because it will always originate from a specific channel

Koenkk:
Then the actions should also be updated.

PR #5375, opened 2023-01-20
Remove redundant and unused temperature parameter

pklokke:
- The local measured temperature is exposed via "local_temperature"


Koenkk:
Thanks!

PR #5379, opened 2023-01-21
add support for PILSKOTT pendent

GsakuL:
I was stupid, not realizing that whiteLabel is Metadata/Docs only: #5368
I'm sorry.

I force-removed the devices, and made sure that this time the 'zigbeeModel' is correct and makes the lamp work. Obviously still shows up as Silvergans, but it essentially just is one pre-packaged.


devices/ikea.js
```diff
@@ -552,7 +552,7 @@ module.exports = [
         extend: tradfriExtend.light_onoff_brightness(),
     },
     {
-        zigbeeModel: ['SILVERGLANS IP44 LED driver'],
+        zigbeeModel: ['SILVERGLANS IP44 LED driver', 'Pendant lamp WW'],
```
Koenkk, line 555: Since the device has a separate model, it can be distinguished from this led driver. Can you move it into a new definition?

GsakuL:
since it's now a new device, I will add an image and add a note (as described in the guide) referencing the fact about the internal/built-in driver.
Thus  Koenkk/zigbee2mqtt#16288 is not complete with this PR alone

PR #5384, opened 2023-01-22
Add power/energy metering to Samotech SM309-S

alexjfisher:
When the device support was originally added, it was a straight copy of the `SM309` code.  This newer variant exposes power/energy use.

The reporting `min` and `change` parameters shamelessly cargo culted from https://github.com/Koenkk/zigbee-herdsman-converters/blob/f130b98ab32491848a4beeaea1d8e0a50519baad/devices/micromatic.js#L24-L27 but they look like sensible enough settings to me.

Tested with home assistant and the zigbee2mqtt addon.  Seems to work just fine.

```json
{
    "brightness": 254,
    "linkquality": 140,
    "power_on_behavior": "previous",
    "state": "ON",
    "current": 1.15,
    "energy": 2.71,
    "power": 275.5,
    "voltage": 238.4
}
```

### DIFF
--- a/devices/samotech.js
+++ b/devices/samotech.js
@@ -1,5 +1,8 @@
+const fz = require('../converters/fromZigbee');
+const exposes = require('../lib/exposes');
 const reporting = require('../lib/reporting');
 const extend = require('../lib/extend');
+const e = exposes.presets;
 
 module.exports = [
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #5385, opened 2023-01-22
Fix type of Danfoss window_open_internal

Nezz:
This entity was exposed as a number, but it's actually an enum now:


Koenkk:
Thanks!

PR #5298, opened 2023-01-08
Aqara FP1 regions basic support

mdziekon:
Changelog:
- Adds basic support for `region_event`, exposing most recent detected region event, as handled by `fzLocal.aqara_fp1_region_events`
- Adds basic support for regions-related configuration, configured by issuing JSON-encoded commands in `regions_config` input field
  - Each command should follow [these type definitions](https://github.com/Koenkk/zigbee-herdsman-converters/compare/master...mdziekon:zigbee-herdsman-converters:aqara-fp1-regions-basic-support?expand=1#diff-2d16489d92b9b8bfd6140c0045ff1f2f7e9caf7df7e92445313f1d1a7609da1aR68-R81) to be correctly handled and sent to the device
  - I'm planning to add this to the device's documentation repo, so that users can read more about this on zigbee2mqtt.io
- No "Other regions" support yet, simply because I didn't have time for this today. I'll try to add this as well in the upcoming days.

Based on:
- research by @Otnow (and some other people) in this thread https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5928#issuecomment-1166545226
- initial implementation by @patmalcolm91 from this thread https://github.com/Koenkk/zigbee2mqtt/issues/13711#issuecomment-1356182773

PR #5386, opened 2023-01-22
Fix Aqara FP1 action field population (previously region_event)

mdziekon:
Fix for https://github.com/Koenkk/zigbee-herdsman-converters/pull/5298

While refactoring, the `region_event` field has been renamed to `action` (see: https://github.com/Koenkk/zigbee-herdsman-converters/pull/5298/commits/fd51312b9a3588c39bb3f55d0445d1a6c7e7cba9), however the event handler still tries to populate `region_event` in its payload, resulting in field not receiving updates.
This PR, as far as I understand other converters, ~should~ does fix that.


Koenkk:
Thanks!

PR #5390, opened 2023-01-24
new zigbeeModel of Popp 701721

olzibambolzi:
just bought a Popp 701721 and it had a new zigbeeModel version. Works as expected after adding it


Koenkk:
Thanks!

Issue #4783, opened 2022-10-14
There is one incorrect value for System Mode on thermostat Smabit model AV/2010/32 


bmorfin:
The wireless wall thermostat Smabit (Bitron Video) model AV/2010/32 has 3 possible values for the attribute "System mode". 

These 3 values are "Off", "Cool" and "Heat". 

But the System Mode attribute values in the Zigbee2MQTT definition for this device are listed as "Off", "Auto" and "Heat". 

The value "Auto" does not exist and should then be replaced by the value "Cool".

bmorfin:
Here is the part of the Technical Manual listing the valid values :

![System Mode](https://user-images.githubusercontent.com/62761219/198890605-b18a8ad8-8b50-45c9-b630-6e21e7d0ae73.png)

bmorfin:
Just to clarify : I assume that the fix should be done in the file **bitron.js**, as Smabit is the new brand name of the Bitron Video products

bmorfin:
Maybe this line of the bitron.js should file be updated 

![image](https://user-images.githubusercontent.com/62761219/201916251-8fd05976-24b2-44e0-8273-fd1845f897e3.png)

bmorfin:
Another part of the technical documentation showinf the same vales for System Mode 

![image](https://user-images.githubusercontent.com/62761219/204135187-2c22e3ba-1d86-4fcf-8e19-64e3d7e342e0.png)

bmorfin:
I assume that the fix should be done in the file bitron.js, as Smabit is the new brand name of the Bitron Video products

bmorfin:
Hello,

According to the documentation of the manufacturer, the values for the system mode attribute are :

![image](https://user-images.githubusercontent.com/62761219/211186467-73ac8ef1-b34d-486a-9c78-d559d5a80745.png)

But in the file, the values are different :

![image](https://user-images.githubusercontent.com/62761219/211186460-be19a3aa-be8f-4a1b-82b0-fa68154d9928.png)

Would it possible to change the file content, so that it corresponds to the documentation for this attribute (Value 'cool' instaed of 'auto') ?

PR #5399, opened 2023-01-26
SMaBiT Thermostat only support heating

sjorge:
In the previous PR I dropped 'auto' from system_mode as the device rejects this with INVALID_VALUE. The device also rejects 'cool' as a system_mode so there is no way for the running state to end up being 'cool' it can only be 'heat' or 'idle' when not heating/off.

```
error 2023-01-26 14:38:37: Publish 'set' 'system_mode' to 'thermostat/home' failed: 'Error: Write 0x000d6f0018c114b8/1 hvacThermostat({"systemMode":3}, {"sendWhen":"active","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'INVALID_VALUE')'
debug 2023-01-26 14:38:37: Error: Write 0x000d6f0018c114b8/1 hvacThermostat({"systemMode":3}, {"sendWhen":"active","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'INVALID_VALUE')
    at Endpoint.checkStatus (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:317:28)
    at Endpoint.write (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:397:22)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Object.convertSet (/opt/zigbee2mqtt/node_modules/zigbee-herdsman-converters/converters/toZigbee.js:1249:13)
    at Publish.onMQTTMessage (/opt/zigbee2mqtt/lib/extension/publish.ts:246:36)
```

### DIFF
--- a/devices/bitron.js
+++ b/devices/bitron.js
@@ -194,7 +194,7 @@ module.exports = [
             tz.thermostat_occupied_cooling_setpoint, tz.thermostat_local_temperature_calibration, tz.thermostat_local_temperature,
             tz.thermostat_running_state, tz.thermostat_temperature_display_mode, tz.thermostat_keypad_lockout, tz.thermostat_system_mode],
         exposes: [e.battery(), exposes.climate().withSetpoint('occupied_heating_setpoint', 7, 30, 0.5).withLocalTemperature()
-            .withSystemMode(['off', 'heat']).withRunningState(['idle', 'heat', 'cool'])
+            .withSystemMode(['off', 'heat']).withRunningState(['idle', 'heat')
             .withLocalTemperatureCalibration(-30, 30, 0.1), e.keypad_lockout()],
         meta: {battery: {voltageToPercentage: '3V_2500_3200'}},
         configure: async (device, coordinatorEndpoint, logger) => {
--- a/devices/bitron.js
+++ b/devices/bitron.js
@@ -194,7 +194,7 @@ module.exports = [
             tz.thermostat_occupied_cooling_setpoint, tz.thermostat_local_temperature_calibration, tz.thermostat_local_temperature,
             tz.thermostat_running_state, tz.thermostat_temperature_display_mode, tz.thermostat_keypad_lockout, tz.thermostat_system_mode],
         exposes: [e.battery(), exposes.climate().withSetpoint('occupied_heating_setpoint', 7, 30, 0.5).withLocalTemperature()
-            .withSystemMode(['off', 'heat']).withRunningState(['idle', 'heat')
+            .withSystemMode(['off', 'heat']).withRunningState(['idle', 'heat'])
             .withLocalTemperatureCalibration(-30, 30, 0.1), e.keypad_lockout()],
         meta: {battery: {voltageToPercentage: '3V_2500_3200'}},
         configure: async (device, coordinatorEndpoint, logger) => {


Koenkk:
Bedankt! 😄

bmorfin:
Hello,

Thanks for removing 'auto' from system mode, as it is not a valid value (I have reported this issue in #4783 and in #4881). 

But 'cool' is a valid value for system mode. You can see this value in the SmaBit documentation :

image

There are 3 valid values : 'cool', 'heat' and 'off'. 

Would it then be possible to add again 'cool' as a valid value ?

sjorge:
I keep getting invalid_value when trying to set the system_mode. From my experience documentation, is unfortunately not always correct :(

You should still be able to set it, as it's only gone from the expose data.

`mosquitto_pub -t zibgee2mqtt/<friendlyname>/set/system_mode -m cool` 

```
debug 2023-02-06 15:49:15: Received MQTT message on 'zigbee2mqtt/thermostat/home/set/system_mode' with data 'cool'
debug 2023-02-06 15:49:15: Publishing 'set' 'system_mode' to 'thermostat/home'
error 2023-02-06 15:49:16: Publish 'set' 'system_mode' to 'thermostat/home' failed: 'Error: Write 0x000d6f0018c114b8/1 hvacThermostat({"systemMode":3}, {"sendWhen":"active","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'INVALID_VALUE')'
```

Mine, at least rejects the value. There is no firmware available for it, but mine shipped with what seems to be a rather recent one.

```
Firmware build date: 2022122722361275
Firmware version: V1b440-20210209
```

bmorfin:
Hi,

I've attached the documentation I've got.


[SMaBiT AV2010-32 Technical Manual.pdf](https://github.com/Koenkk/zigbee-herdsman-converters/files/10664367/SMaBiT.AV2010-32.Technical.Manual.pdf)
 
I'll check with Sambit's support team about the actual possible values for this attribute. I'll let you know if I get an answer.

sjorge:
OK I did some more digging,

You can reconfigure it using ControlSequenceOfOperation and depending on that value it will accept different systemMode's!

I should have some time this sunday to play around with this, it seems it can at least also work as a cooling mode mode where it rejects heat.

@Koenkk I think this is another job for a dynamic expose function based on the value of `ControlSequenceOfOperation` but i'm not sure if there is a way to trigger a republish of expose data when `ControlSequenceOfOperation` is changed?

Koenkk:
@sjorge  there is indeed no such mechanism in place (and it would be some work to implement this), restarting seems to be an acceptable workaround for this case.

sjorge:
Had a closer look, I should be able to finish this on sunday unless something comes up.

I also noticed this device seems to supports programming a schedule, I might look into this later but i'm going to ignore that for now.

PR #5401, opened 2023-01-27
Expose 3RSS009Z battery percentage

grelca:
exposes battery percentage for 3RSS009Z and also fixes its incorrectly-reported power source

confirmed that this device exposes `batteryPercentageRemaining` with max 100 range
Screenshot 2023-01-26 at 8 36 59 PM


Koenkk:
Thanks!

PR #5402, opened 2023-01-28
Fix support for KAMI motion sensor

Tediore:
The [KAMI motion sensor](https://www.zigbee2mqtt.io/devices/N20.html) (which has the same zigbeeModel as the KAMI contact sensor) only sends "occupied" messages. It also doesn't keep sending occupied messages as motion continues to be detected (it only sends it once until motion is no longer detected for 5 seconds). As a result, the current converter for this device isn't a good fit. Also, the current converter defined occupancy when `msg.data[7]` is 0, but it's actually when it's 1, so I updated that. 

Because this device doesn't keep sending occupied messages as motion continues to be detected, I don't think it makes sense to use the occupancy_with_timeout converter with this device. I updated the device entry and converter for this device to simply expose an action entity that returns "motion" whenever motion is detected.

This is probably considered to be a breaking change for this device, but since the current converter doesn't actually work properly, it probably doesn't matter.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -6862,7 +6862,9 @@ const converters = {
         cluster: 'msOccupancySensing',
         type: ['raw'],
         convert: (model, msg, publish, options, meta) => {
-            return {occupancy: msg.data[7] === 0};
+            if (msg.data[7] === 1) {
+                return {action: 'motion'};
+            }
         },
     },
     DNCKAT_S00X_buttons: {
--- a/devices/kami.js
+++ b/devices/kami.js
@@ -7,9 +7,9 @@ module.exports = [
         zigbeeModel: ['Z3ContactSensor'],
         model: 'N20',
         vendor: 'KAMI',
-        description: 'Entry sensor',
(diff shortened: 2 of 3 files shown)


Koenkk:
@cherowley could you confirm this? (https://github.com/Koenkk/zigbee2mqtt/discussions/6154)

cherowley:
Yep all working 👍

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

Koenkk:
@cherowley so on the latest z2m (1.29.2) the sensor is working fine?

cherowley:
Yes thanks 😊

Koenkk:
@cherowley so I does send an `occupancy: false` message after detecting `occupancy` after some time?

Tediore:
Hmm... definitely interested in the answer. During my testing, I left mine face down for over an hour to see if it sent an `occupancy: false` message and it never did.

cherowley:
Ah, I’m not using that status…

Koenkk:
I see, thanks @Tediore !

PR #5405, opened 2023-01-29
Fix wrong backlight setting for Moes ZS-EUB_1gang

mcuelenaere:
The backlight mode is off/normal/inverted instead of low/medium/high.


Koenkk:
Thanks!

PR #5406, opened 2023-01-29
Add Lightsolutions 30W Zigbee CCT LED driver.

lbach:
Lightsolutions 30W Zigbee CCT LED driver is basically a Sunricher white-labeled device, similar to Sunricher HK-ZD-CCT-A.


Koenkk:
thanks!

PR #5407, opened 2023-01-29
Add Device Support for SV02-610-MP-1.0

curt7000:
Add support for another FW Version SV02-610-MP-1.0.  Tested locally and working!


Koenkk:
Thanks!

PR #5409, opened 2023-01-30
Fix Sinope RM3500ZB

embak:
Fix mispelled temperature sensor binding and water leak sensor has no battery.

### DIFF
--- a/devices/sinope.js
+++ b/devices/sinope.js
@@ -12,6 +12,18 @@ const {precisionRound} = require('../lib/utils');
 const manuSinope = {manufacturerCode: 0x119C};
 
 const fzLocal = {
+    ias_water_leak_alarm: {
+        // RM3500ZB specific
+        cluster: 'ssIasZone',
+        type: ['commandStatusChangeNotification', 'attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const zoneStatus = msg.data.zoneStatus;
+            return {
+                water_leak: (zoneStatus & 1) > 0,
+                tamper: (zoneStatus & 1<<2) > 0,
+            };
+        },
+    },
     thermostat: {
         cluster: 'hvacThermostat',
         type: ['attributeReport', 'readResponse'],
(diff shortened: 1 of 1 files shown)


devices/sinope.js
```diff
@@ -1179,21 +1191,26 @@ module.exports = [
...
             await reporting.bind(endpoint, coordinatorEndpoint, binds);
             await reporting.onOff(endpoint);
+            await reporting.temperature(endpoint, {min: 10, max: 301, change: 10}); // divider 100: 0.1C
             await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
-            await reporting.activePower(endpoint);
-            await reporting.rmsCurrent(endpoint);
-            await reporting.rmsVoltage(endpoint);
+            await reporting.activePower(endpoint, {min: 10, max: 305, change: 2}); // divider 1 : 2W
+            await reporting.rmsCurrent(endpoint, {min: 10, max: 306, change: 10}); // divider 1000: 0.01Arms
+            await reporting.rmsVoltage(endpoint, {min: 10, max: 307, change: 1}); // divider 1: 1Vrms
```
Koenkk, line 1208: Why 305/306 and 307?

embak: In the past, we observed a lot of traffic and packet drops caused by devices reporting multiple electric measurements updates. This happens when there is a large number of devices on the same Zigbee network. In my case I have over 15 thermostats of all kind and other devices. Event report packet storms are amplified by power restoration or voltage variations. The packet storm will repeat afterward at every maximum reporting interval.  
Also Sinope thermostats and other of their devices doing energy or electrical measurements are verry chatty and sensitive to voltage variations. I experienced this with their thermostats and applied different maximum reporting intervals to the TH1123ZB/TH1124ZB/TH1500ZB converters in previous PR. , Their new water heater controler (RM3500ZB) and their wall plug switch (SP2600ZB) are even more sensitive.
Using different maximum reporting intervals is a proven method also used by Smarthings converters and others to mitigate this problem.

Koenkk:
thanks!

PR #5410, opened 2023-01-30
Fix reporting for HZC Electric D086-ZG

andredahlqvist:
Fix reporting for D086-ZG. Previously turning off a channel or dimming it would not be reflected correctdly in Home Assistant.


Koenkk:
Thanks!

PR #5413, opened 2023-01-31
Add Support for 2W A60 bulbs on FW 1.3

TrueCyberAxe:
These bulbs on FW version 1.3 show as "AJ_ZBPROA6" so I have added that model to support that firmware version.


Koenkk:
Thanks!

PR #5416, opened 2023-02-01
Add Philips Hue Filament White Ambiance ST72 B22

mattkrins:
Add LTV003/8719514301528, Philips Hue Filament White Ambiance ST72 B22


Koenkk:
thanks!

PR #5417, opened 2023-02-01
Add support for ALCANTARA2 D1.00P1.01Z1 version

peix187:
I have successfull added this device version on my Z2M (conbee 2) with HA and it work well


Koenkk:
Thanks!

PR #5422, opened 2023-02-01
Update ajax_online.js

cooper645:
After still being unsupported, I re-read the documentation. I Used the manufacturer model name, not the one provided by the actual device. I think this was the issue. Added fingerprint from the debug also.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "zigbee-herdsman-converters",
-  "version": "15.0.39",
+  "version": "15.0.40",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "zigbee-herdsman-converters",
-      "version": "15.0.39",
+      "version": "15.0.40",
       "license": "MIT",
       "dependencies": {
         "axios": "^1.3.0",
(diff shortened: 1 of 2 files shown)


devices/ajax_online.js
```diff
@@ -39,8 +39,9 @@ module.exports = [
         meta: {applyRedFix: true, enhancedHue: false},
     },
     {
+        fingerprint: [{modelID: 'CCT Light', manufacturerName: 'ZB/Ajax Online', manufacturerID: 4137}],
         zigbeeModel: ['ZB-CCT_Filament'],
-        model: 'ZB-CCT_Filament',
+        model: 'CCT Light',
```
Koenkk, line 44: This is a different device, please create a new definition for it (you can copy paste the existing one and change the `description`)

cooper645: It's the same device, I have added the fingerprint from the converter when I initially created it to show where I think I went wrong.

I believe my error was entering `ZB-CCT_Filament` as the model as listed on the manufacturer website instead of `CCT Light`, which is what is listed in the fingerprint. I didn't realise that model had to match what was in the fingerprint.

PR #5423, opened 2023-02-01
Add support for EB-E27-ST64-CCT-FV light bulb

Chartreusito:
Add support for EB-E27-ST64-CCT-FV light bulb


Koenkk:
Thanks! Could you also provide a picture for it?

PR #5424, opened 2023-02-01
Update bankamp.js

ahemwe:
Remove grazia as description since other lights share the same zigbee device.

Suggested [here](https://github.com/Koenkk/zigbee2mqtt.io/pull/1860#issuecomment-1410718579)


Koenkk:
Thanks!

PR #5430, opened 2023-02-03
Fingerprint for _TZE200_dq1mfjug for Tuya TS0601 smoke detector ES63-D5Z

Leissson:
Fingerprint for Tuya TS0601 smoke detector ES63-D5Z

image


Koenkk:
Thanks!

PR #5431, opened 2023-02-03
Update ajax_online.js

cooper645:
After adding support for this device, I realise I made an error and have amended it, the device doesn't expose itself as `ZB-CCT_Filament`, the amendment I made now reflects my external converter exactly. This amendment now means that the filament bulb I tried to add will now actually be supported as it still came up as unsupported when I had converter as `zigbeemodel: ZB-CCT_Filament`.


Koenkk:
thanks!

PR #5432, opened 2023-02-03
Fix setting schedule to TS0601_thermostat_1

Delorien84:
Update TRV definition to use tyua datapoints. Cleanup the code and fix setting schedule. 

Fix https://github.com/Koenkk/zigbee2mqtt/issues/9820


devices/tuya.js
```diff
@@ -2477,41 +2477,71 @@ module.exports = [
...
                 'the device will enter the boost heating mode, and the ▷╵◁ will flash. The countdown will be displayed in the APP'),
-            exposes.numeric('boost_heating_countdown', ea.STATE_SET).withUnit('min').withDescription('Countdown in minutes')
+            exposes.numeric('boost_time', ea.STATE_SET).withUnit('min').withDescription('Countdown in minutes')
                 .withValueMin(0).withValueMax(1000),
         ],
+        meta: {
+            tuyaDatapoints: [
+                [1, 'system_mode',
+                    {
+                        from: (v) => {
```
Koenkk, line 2510: Can this be solved with the `tuya.valueConverterBasic.lookup`? (what is the problem when used)

lib/tuya.js
```diff
@@ -1523,6 +1499,19 @@ const valueConverter = {
     },
 };
 
+const valueConverterAdvance = {
+    thermostatScheduleDayMultiDPWithDayNumber: (dayNum) => {
```
Koenkk, line 1503: I don't believe it is necessary to put this in `valueConverterAdvance`, it should work fine in `valueConverter` since `valueConverter` is only referenced in the lambdas.

Koenkk:
Many thanks for this cleanup! 🙇

lib/tuya.js
Delorien84: will try

devices/tuya.js
Delorien84: I do not think so. Because I will need create map with two `auto` keys. The problem is that the TRV only reports `preset` not the `system_mode` so I need to do some mapping. If I do not report `system_mode` HVAC modes in HA is not set correctly and show `cooling` etc. And `system_mode` does not allow `manual` value. Any ideas about that?

Koenkk: You can also return the preset in this converter, to return multiple keys from `from` see e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/712ce49bff609b5b65516f5491f496993f95322a/lib/tuya.js#L1302

> And system_mode does not allow manual value. Any ideas about that?

How was this previously handled?

Delorien84: It works same as currently. When reading, set both `system_mode` and `preset`. When writing `system_mode` or `preset` set same value.

Delorien84: Yes you can, it will also change `preset` to `on`

Delorien84: In HA I only change `preset`. Does not use HVAC mode (`system_mode`).

Koenkk: I see, is this PR ready for merge now?

Delorien84: Yes, it is.

Koenkk:
many thanks!

PR #5433, opened 2023-02-03
Add Moes whitelabel to TuYa TS0044

olexs:
Got this switch from Amazon.de in the 4-gang variant, it's white labeled by Moes with "powered by TuYa" in the corner of the box, interviewed as Tuya TS0044 and works perfectly. Didn't find it under Moes it in the device compatibility list initially, so I thought I'd add the whitelabel and autogenerated docs pick it up from here.

There seem to also exist 1-, 2- and 3-gang variants that share the same box and manual, and presumably internals. However I couldn't quickly find exact matches in the existing TuYa devices (or for sale, for that matter), so I only added the 4-gang that I can confirm matches and works.


Koenkk:
Thanks!

PR #5442, opened 2023-02-04
[New device support]: Add support for Aqara E1 3 gang switch (with neutral)

wilsonkhlam:
It is to follow-up the [request](https://github.com/Koenkk/zigbee2mqtt/issues/14249). Base on the contribution from @fata1ityz, I have fixed the issue with reporting incorrect button action.


Koenkk:
Thanks!

PR #5447, opened 2023-02-06
hue: change default for disableHueEffects

sjorge:
As mentioned in https://github.com/Koenkk/zigbee2mqtt/issues/16553, I grepped all our zigbeeModels and the first 25 were all on the list. I then randomly grabbed like another 10 and they were also listed.

I the noticed we had some light_onoff_brightness extends, these were not on the list from the PDF (makes sense) so these were not changed.

### DIFF
--- a/lib/philips.js
+++ b/lib/philips.js
@@ -66,7 +66,7 @@ const extend = {
         return result;
     },
     light_onoff_brightness_colortemp: (options={}) => {
-        options = {disableHueEffects: true, ...options};
+        options = {disableHueEffects: false, ...options};
         if (!options.disableHueEffects) options.disableEffect = true;
         const result = extendDontUse.light_onoff_brightness_colortemp(options);
         result['ota'] = ota.zigbeeOTA;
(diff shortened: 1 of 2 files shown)


sjorge:
I briefly considered renaming and inverting disableHueEffects, but we also have disableEffects which we toggle. So I opted to keep both it as disableHueEffects with a default of `false` now.

I also removed all the options where we manual set disableHueEffects to false before.

lib/philips.js
```diff
@@ -52,21 +52,14 @@ const knownEffects = {
 
 const extend = {
     light_onoff_brightness: (options={}) => {
-        options = {disableHueEffects: true, ...options};
-        if (!options.disableHueEffects) options.disableEffect = true;
         const result = extendDontUse.light_onoff_brightness(options);
         result['ota'] = ota.zigbeeOTA;
         result['meta'] = {turnsOffAtBrightness1: true};
         result['toZigbee'] = result['toZigbee'].concat([philipsTz.hue_power_on_behavior, philipsTz.hue_power_on_error]);
-        if (!options.disableHueEffects) {
```
Koenkk, line 61: I don't think this should be removed, e.g. [`046677562229`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3656a900508768a935fdf0d5bf122ff42c01a5cf/devices/philips.js#L2793) is known to support the effects>

sjorge: So do we keep onoff_brightness with the old default or do we also flip it? The few random ones i checked were not in the pdf vs all of the colortemp and color ones which wete?

sjorge: 👍 will give this another pass tomorrow

Koenkk:
Thanks again! 🏅

lundyfpv:
Thanks for doing this!

PR #5452, opened 2023-02-07
SMaBiT AV2010/32 does support cool when configured via ctrlSeqOfOper

sjorge:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/5399#issuecomment-1419194233

This device will support cool or heat depending on the ctrlSeqOfOper value. The manual it has logic for heating_only and cooling_only.

image

The device now has a 'dynamic expose' function that returns the correct supported modes based on `ctrlSeqOfOper`. I also added batteryVoltage reading, fixed the battery calculation. I also extended `Climate` expose to add a withCtrlSeweofOper methode.

We should update the notes for this device that a restart of z2m is needed after changing ctrlSeqOfOper for the expose data to be updated.

### DIFF
--- a/converters/toZigbee.js
+++ b/converters/toZigbee.js
@@ -1266,8 +1266,12 @@ const converters = {
             if (val === undefined) {
                 val = utils.getKey(constants.thermostatControlSequenceOfOperations, value, value, Number);
             }
-            await entity.write('hvacThermostat', {ctrlSeqeOfOper: val});
-            return {state: {control_sequence_of_operation: value}};
+            const attributes = {ctrlSeqeOfOper: val};
+            await entity.write('hvacThermostat', attributes);
+            // NOTE: update the cluster attribute we store as this is used by
+            //       SMaBiT AV2010/32's dynamic expose function.
+            entity.saveClusterAttributeKeyValue('hvacThermostat', attributes);
+            return {readAfterWriteTime: 250, state: {control_sequence_of_operation: value}};
         },
         convertGet: async (entity, key, meta) => {
             await entity.read('hvacThermostat', ['ctrlSeqeOfOper']);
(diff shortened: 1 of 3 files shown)


sjorge:
ctrlSeqeOfOper set to 0 (cooling_only) and restarted z2m
Screenshot 2023-02-07 at 19 14 17

Yes, system_mode is still set to heat as the thermostat hasn't reported the new mode yet. But hitting cool or off does not result in a INVALID_VALUE.

After setting ctrlSeqeOfOper to 2 (heating_only) and restarting z2m:
image


I can now set system_mode to off or heat without an INVALID_VALUE message.

converters/toZigbee.js
```diff
@@ -1261,6 +1267,13 @@ const converters = {
                 val = utils.getKey(constants.thermostatControlSequenceOfOperations, value, value, Number);
             }
             await entity.write('hvacThermostat', {ctrlSeqeOfOper: val});
+            // NOTE: read the value instead of returning state
+            //       this way the internal device hvacThermostat.ctrlSeqeOfOper
+            //       we track is updated.
+            //
+            //       this is needed because some device, e.g. SMaBiT AV2010/32 have dynamic
+            //       expose data based on this attribute
+            await entity.read('hvacThermostat', ['ctrlSeqeOfOper']);
```
Koenkk, line 1276: To prevent an unnecessary read, we can also update to value directly here, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/ee8db717cd831dc807a5b70ee57358d51b177e74/converters/fromZigbee.js#L7531 (`msg.endpoint` becomes `entity`)

lib/exposes.js
```diff
@@ -449,6 +449,14 @@ class Climate extends Base {
         return this;
     }
 
+    withCtrlSeqeOfOper(modes, access=a.STATE) {
```
Koenkk, line 452: Please rename this to `withControlSequenceOfOperation`

sjorge:
@Koenkk I'll do some more testing during my lunch break and let you know if everything still looks good.

sjorge:
All good, still works as expected. After changing ctrlSeqeOfOpr and restarting.

Koenkk:
Thanks!

PR #5453, opened 2023-02-07
Add "_TYZB01_cc3jzhlj" for TS0210

nmr94:
My Tuya TS0210 has the manufacturer name set to "_TYZB01_cc3jzhlj" so I've added this to the list of recognised fingerprints.

This has been tested with these devices and works perfectly (see attached images).

SCR-20230207-ufc
SCR-20230207-uf2


Koenkk:
Thanks!

PR #5458, opened 2023-02-09
Adds support for 1743930V7 model Econic outdoor

registerednerd:
Previous version was missing a specific model number, likely a body color difference.


Koenkk:
Thanks!

PR #5461, opened 2023-02-10
Added reporting setpoints to TS0121

stokket:
Reason: To prevent Zigbee network to be flooded with information.
Also added whiteLabel to all four connecte wall socket devices.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -2541,7 +2541,13 @@ module.exports = [
         zigbeeModel: ['TS0121'],
         model: 'TS0121_plug',
         description: '10A UK or 16A EU smart plug',
-        whiteLabel: [{vendor: 'BlitzWolf', model: 'BW-SHP13'}],
+        whiteLabel: [
+            {vendor: 'BlitzWolf', model: 'BW-SHP13'},
+            {vendor: 'Connecte', model: '4500990'},
+            {vendor: 'Connecte', model: '4500991'},
+            {vendor: 'Connecte', model: '4500992'},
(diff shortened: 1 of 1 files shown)


Koenkk:
Thansk!

gsolem:
Still spamming my logs after update to 1.30.2. Does it mean I have to remove the device and re-pair? I have tried the reconfigure option without any luck. I have a Connecte 4500992.

stokket:
> Still spamming my logs after update to 1.30.2. Does it mean I have to remove the device and re-pair? I have tried the reconfigure option without any luck. I have a Connecte 4500992.

image

my configuration looks like this, if you dont want to re-pair, just copy the info from the picture.

gsolem:
That made it alot better, now I am getting reports every minute in stead of multiple times a second :)

RubenKelevra:
@stokket can we make something like this the default for this device? I mean, most users don't need/want so finely granular values anyway.

stokket:
> @stokket can we make something like this the default for this device? I mean, most users don't need/want so finely granular values anyway.

My custom config file for testing this device is identical to release. which is configuring as described in picture.. I havent had time to test newest release to check for difference. if you could provide information on how its paring on your end I can have a look at the configuration

RubenKelevra:
@stokket after updating the values are shown the same as in your config.

Just one thing looks weird to me:

Screenshot from 2023-03-02 20-46-58

So is it polling or do we get reports pushed?

I mean isn't “reporting” and the “change” setting not just a setting send to the device to adjust their pushes to the network?

PR #5462, opened 2023-02-11
Fix 4257050-RZHAC factor keys for power

bdd:
Fixes factor to be 0.1 instead of fallback of 1 leading to reporting 10X power use.

In commit 4aac49d7c7d11eadf2c9b91662dfab841b139edf, when adding the fallback for `haElectricalMeasurement` cluster current and power divisor and multiplier read, the fallback object uses the wrong factor prefix for power multiplier and divisor.

power{Multiplier,Divisor} -> acPower{Multiplier,Divisor}


Koenkk:
Thanks!

PR #5469, opened 2023-02-13
New device _TZE200_rhblgy0z

cserem:
This device seem to be identical to _TZE200_lsanae15, but it has reverse metering capabilites.

This is basically a copy-paste of _TZE200_lsanae15, with e.produced_energy() and datapoint 2 defined.

If it is more reasonable to merge the two devices into one, please do so.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -2819,6 +2819,55 @@ module.exports = [
         },
     },
     {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_rhblgy0z']),
+        model: 'TS0601_din_3',
+        vendor: 'TuYa',
+        description: 'Zigbee DIN energy meter',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        whiteLabel: [{vendor: 'XOCA', model: 'DAC2161C'}],
+        exposes: [tuya.exposes.switch(), e.energy(), e.produced_energy(), e.power(), e.voltage(), e.current(),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

fred-c1:
@cserem 
Thanks for adding the support of this new device.

I just received the same device:    _TZE200_rhblgy0z

It seems that the picture here:
https://www.zigbee2mqtt.io/devices/TS0601_din_3.html
is incorrect.

The picture should be same as 
https://www.zigbee2mqtt.io/devices/TS0601_din_2.html

Do you agree?

cserem:
> @cserem Thanks for adding the support of this new device.
> 
> I just received the same device: _TZE200_rhblgy0z
> 
> It seems that the picture here: https://www.zigbee2mqtt.io/devices/TS0601_din_3.html is incorrect.
> 
> The picture should be same as https://www.zigbee2mqtt.io/devices/TS0601_din_2.html
> 
> Do you agree?

Yes, ours also looks like TS0601_din_2:
image

Please do create a pull request.

Claudinetto2:
Good morning, I bought this device _TZE200_rhblgy0z but it send power value every 2 minutes, can i modify this time?

fred-c1:
I don't think this is possible.

Claudinetto2:
Bad news. Do you know an alternative?

fred-c1:
If you don't need the relay and the display:
https://www.zigbee2mqtt.io/devices/TS0601_bidirectional_energy_meter.html
Measurements are sent every 3 seconds when the current is above 0.1A
and every 37 seconds when the current is below 0.1A

Claudinetto2:
have you ALI link or other store?

fred-c1:
https://fr.aliexpress.com/item/1005005466228073.html

PR #5473, opened 2023-02-14
Ensure heating temp can be set and read for Tuya TV02-Zigbee Thermostat

mattheworiordan:
This PR ensures `current_heating_setpoint` can be set and read.

Incorrect access prevents this device from appearing in Hombridge as the device is not considered a thermostat unless the temperature can be set as well as read.

This pull fixes https://github.com/itavero/homebridge-z2m/issues/635

---

This is my first PR in the Zigbee ecosystem. Please shout if I have need to make changes - advice welcome.


Koenkk:
Thanks!

PR #5465, opened 2023-02-12
IKEA: try and be smart about dontDividePercentage

sjorge:
https://github.com/Koenkk/zigbee2mqtt/issues/16412

@Koenkk can you confirm 2.4.x is where IKEA change it?

My on/off ones are on 2.2.010 which is the last one with group binding support, dontDividePercentage=true is needed
My open/close that I got with my fyrtur are on 2.3.079, no group bind but they also seem to require dontDividePercentage=true

It seems the one form the issue was on 2.4.x so I picked that as the cutoff. For now anything < 2.4 has dontDividePercentage=true and everything else has it dontDividePercentage=false.

For my tests I had it at 2.3 so I could see if the code path was doing the right thing, and it was:

```
# 2.2.x
Zigbee2MQTT:debug 2023-02-12 14:17:19: Received MQTT message on 'zigbee2mqtt/remote/office/light_desk/get' with data '{"battery":""}'
Zigbee2MQTT:debug 2023-02-12 14:17:19: Publishing get 'get' 'battery' to 'remote/office/light_desk'
Zigbee2MQTT:debug 2023-02-12 14:17:22: Received Zigbee message from 'remote/office/light_desk', type 'readResponse', cluster 'genPowerCfg', data '{"batteryPercentageRemaining":74}' from endpoint 1 with groupID 0
Zigbee2MQTT:warn  2023-02-12 14:17:22: XXX: dontDividePercentage = true
Zigbee2MQTT:info  2023-02-12 14:17:22: MQTT publish: topic 'zigbee2mqtt/remote/office/light_desk', payload '{"battery":74,"linkquality":116,"update":{"installed_version":570492465,"latest_version":587699761,"state":"available"}}'

# 2.3.x
Zigbee2MQTT:debug 2023-02-12 14:17:33: Received MQTT message on 'zigbee2mqtt/remote/livingroom/side_back_window/get' with data '{"battery":""}'
Zigbee2MQTT:debug 2023-02-12 14:17:33: Publishing get 'get' 'battery' to 'remote/livingroom/side_back_window'
Zigbee2MQTT:debug 2023-02-12 14:17:37: Received Zigbee message from 'remote/livingroom/side_back_window', type 'readResponse', cluster 'genPowerCfg', data '{"batteryPercentageRemaining":87}' from endpoint 1 with groupID 0
Zigbee2MQTT:warn  2023-02-12 14:17:37: XXX: dontDividePercentage = false
[...]

PR #5475, opened 2023-02-15
ikea: blind should also use smart battery calulation

sjorge:
Looks like people with the new hub get updates we're not getting via z2m for the ikea smart blind series. The 24.x version also uses the new batteryPercentageRemaining values.

https://github.com/Koenkk/zigbee-herdsman-converters/pull/5465#issuecomment-1429605834


Koenkk:
Thanks!

PR #5483, opened 2023-02-16
 Filter erroneous values of Sonoff SNZB-02_EFEKTA 

MichelNDF:
Same fix as for SNZB-02 (#13640), but with slightly different values (filter temperature < -33C, > 100C)
As well as for the humidity which has this odd behaviour with this firmware as well (filter humidity > 99.75%)

### DIFF
--- a/devices/custom_devices_diy.js
+++ b/devices/custom_devices_diy.js
@@ -831,7 +831,7 @@ module.exports = [
         vendor: 'Custom devices (DiY)',
         description: 'Alternative firmware for the SONOFF SNZB-02 sensor from EfektaLab, DIY',
         fromZigbee: [fzLocal.SNZB02_temperature, fzLocal.SNZB02_humidity, fz.battery, fzLocal.termostat_config,
-                     fzLocal.hydrostat_config, fzLocal.node_config],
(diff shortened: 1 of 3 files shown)


devices/custom_devices_diy.js
```diff
@@ -143,6 +143,24 @@ const tzLocal = {
...
+    // https://github.com/Koenkk/zigbee2mqtt/issues/13640
+    SNZB02_temperature: {
+        ...fz.temperature,
+        convert: (model, msg, publish, options, meta) => {
+            if (msg.data.measuredValue > -3300 && msg.data.measuredValue < 10000) {
+                return fz.temperature.convert(model, msg, publish, options, meta);
+            }
+        },
+    },
+    SNZB02_humidity: {
```
Koenkk, line 156: Can you move these two conveters to `fromZigbee.js` and remove `sonoff.js/SNZB02_temperature` converter (to prevent code duplication)

MichelNDF: It uses, slightly, different filter values, is this a problem?

Koenkk: No, both are really extreme (-33 vs -100)

Koenkk:
Thanks!

PR #5485, opened 2023-02-16
Trim new lines in Zigbee Model Ids

mikeyyuen:
Handle newlines in Zigbee Model Ids by stripping them along side nulls and other characters in the model lookup.

Addresses some of the issues in https://github.com/Koenkk/zigbee2mqtt/issues/16272 that prevent PS600 being matched with just it's model name.

The PS600 generates model ids like `"modelId":"PS600\u0000\u0000 �\u0001 ��\u0000\u0000\t\u0001@\u0011�u\u0000\u0000\u0018\u0002\n\u0005\u0000B"` which can include new lines.

Validated the regex will behave as expected:

```javascript
"PS600\u0000\u0000 �\u0001 ��\u0000\u0000\t\u0001@\u0011�u\u0000\u0000\u0018\u0002\n\u0005\u0000B".replace(/\0(.|\n)*$/g, '').trim()
```

Which outputs the following in the javascriot console:

```
"PS600"
```


Koenkk:
Thanks!

devices/salus_controls.js
```diff
@@ -107,6 +107,7 @@ module.exports = [
     {
         zigbeeModel: [
             'SS909ZB',
+            'PS600',
             'PS600\u0000\u0000 �\u0001 ��\u0000\u0000\t\u0001@\u0011�u\u0000\u0000\u0018&\n\u0005\u0000B',
             'PS600\u0000\u0000 �\u0001 ��\u0000\u0000\t\u0001@\u0011�u\u0000\u0000\u0018\u0006\n\u0005\u0000',
             'PS600\u0000\u0000 �\u0001 ��\u0000\u0000\t\u0001@\u0011�u\u0000\u0000\u0018\u0006\n\u0005\u0000B',
```
normanr, line 113: Could these lines now be removed?

mikeyyuen: Yep, I wanted to make sure we didn’t break anything first and it looks like we didn’t! Should be all gone now!

PR #5490, opened 2023-02-17
New device support: QBKG28LM

pyatins:
Manage to add Aqara H1 Pro two-key switch without neutral (model: QBKG28LM).


Koenkk:
Thanks!

PR #5497, opened 2023-02-19
Added support for TRV Appartme APRM-04-001

Cosik:
Added support for Appartme APRM-04-001. It is similar to GS361A-H04/GTZ02 https://global.appartme.pl/en/23-glowica-termostatyczna.html


Koenkk:
Thanks!

PR #5501, opened 2023-02-20
Add support for _TZE200_zl1kmjqx temp and humidity sensor

ov1d1u:
This is the device: https://www.aliexpress.com/item/1005004930449537.html


Koenkk:
Thanks!

PR #5504, opened 2023-02-20
Better white/color behaviour of Tuya WZ5

YodaDaCoda:
Remove separate white brightness.
While it is technically possible to illuminate both RGB and CCT LEDs at the same time, this does not happen during normal operation (e.g. using Skydance R10 remote).

This makes the device show up in HomeAssistant much more nicely i.e. a single light with RGB and CCT modes, rather than a RGBCCT light and a separate white brightness slider.

TODO:
- [x] Consolidate entities & adjust behaviour such that there is no separate white_brightness
- [x] Ensure state is reflected correctly when changing colors/modes via remote (i.e. not via z2m)
  * device doesn't seem to send silvercrestSetColor when changing from CCT to RGB mode, but does send updates after adjusting RGB color/brightness
- [ ] Figure out strange color behaviour when low brightness is selected in HA


YodaDaCoda:
I promise I'm gonna come back to this

PR #5511, opened 2023-02-22
Fix missing power source for switch gen3, motion sensor, door sensor,…

llhappier:
… water sensor and smart button.


Koenkk:
Thanks!

PR #5513, opened 2023-02-22
Update tuya.js Device _TZ3000_4o16jdca doesn't support switchType and backlightModeOffOn 

landsmanv:
Device _TZ3000_4o16jdca doesn't support switchType and backlightModeOffOn like _TZ3000_vsasbzkf, _TZ3000_odzoiovu removed from
added new record for _TZ3000_4o16jdca

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -2042,7 +2042,24 @@ module.exports = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #5514, opened 2023-02-23
Fix color handling for LIDL light bar (Koenkk/zigbee2mqtt#13421)

AndreKR:


Koenkk:
Made some small changes, many thanks!

I'm sure many people will be thankful!

RubenKelevra:
Hey @Koenkk,

I think this has to be changed for the LIDL HG08010 (Livarno Home outdoor spotlight) as well, as 0 does something which looks like 1 second on them.

```
// A transition time of 0 would be treated as about 1 second, probably some kind of fallback/default
// transition time, so for "no transition" we use 1 (tenth of a second).
```

RubenKelevra:
Hey @AndreKR, can you please explain how the color handling looks like before the fix? I like to confirm that it isn't happening on my HG08010 lamps :)

AndreKR:
@RubenKelevra The light acts as a standard Zigbee light, but only in white mode. The RGB LEDs are controlled completely separate by custom commands that we added with this PR. If you try using the light as a normal Zigbee light without custom commands, if I remember correctly the color LEDs would turn on, but they didn't turn off, so you'd have color and white LEDs on at the same time.

PR #5515, opened 2023-02-23
ubisys: H1 tweak pi_heating_demand reporting

sjorge:
Since this uses 0-100 for the valve percentage, it also seems to only update in ~3% increments. With the defaults it doesn't always report the valve is closed, sometimes it takes close to 1 hour. Which is undesirable for example if you have other automations that trigger the boiler based on demand.

After this change we knock it down to 1% changes before it is allowed to report. However we do cap the minimal reporting interval at 15 so it doesn't spam messages when it's making a position change. (All my position changes seem to take around 5-10 seconds)


Koenkk:
Thanks!

PR #5516, opened 2023-02-23
Smart plug with power in Third Reality

llhappier:
Add missing message for smart plug with power in Third Reality.


Koenkk:
Thanks!

PR #5517, opened 2023-02-23
battery: reject invalid 255 values

sjorge:
As mentioned https://github.com/Koenkk/zigbee2mqtt/issues/15725
 0xff(255) is an invalid or unknown reading and we should not update battery or voltage.


Koenkk:
Thanks!

PR #5524, opened 2023-02-25
Add LEDVANCE 2x2 panel

nkiest:
US version of the Panel TW Z3. OTA does not seem to find any results, so I skipped it. Verified colorTempRange.


Koenkk:
Thanks!

PR #5526, opened 2023-02-26
Add converter for Frient Door entry sensor

Frederik91:
Website: https://frient.com/products/entry-sensor/


sjorge:
Those are made by Develco, we have the pro version here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/ac121be70a47856fadd25f3ffe04412db5fa6683/devices/develco.js#L493

sjorge:
LGTM

Frederik91:
It seems like the tamper function is not present on this device, at least I'm not able to provoke it or find any information about the feature. The closest I could find is that the pro version has temperature and tamper sensor, where as the non-pro has neither of these. Should I remove the tamper part of the config or leave it there?
The other thing I noticed is that the default temperature min rep change is 100, a whole degree celsius. this seems a bit high, at least for if this is to control a heater. Is it ok to override this to say, 0.3 C or around there?

sjorge:
Just drop e.tamper from the exposes list, you don't need to remove anything from fromZigbee as  it will always grab the tamper bit from the ias message.

You can indeed specify a differenr change value in the configure function. Although it might be best to leave it at default there and just override it via the frontend to your needs.

Just of interest, mine all report 1-2C difference from one of my Frient VOC or Humidity sensors i have more centrally located because mine are on my windows and sunlight or lack there of effects them a lot (glass is UV coated double glazed) so they might not be ideal to control a heater.

Frederik91:
Great, I’ve done that now. 
That’s a good point, it might be better to use the smoke detector instead, thanks 😊

Koenkk:
Thanks!

PR #5534, opened 2023-03-01
ikea: Add turnsOffAtBrightness1 for TRADFRI 10W & 30W drivers

Electry:
This is the behavior of the latest IKEA firmware.


Koenkk:
Thanks!

PR #5537, opened 2023-03-02
Add OTA support for Xiaomi `MCCGQ12LM`

Otnow:
OTA file added with https://github.com/Koenkk/zigbee-OTA/pull/268

### DIFF
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -2908,6 +2908,7 @@ module.exports = [
         toZigbee: [],
         meta: {battery: {voltageToPercentage: '3V_2850_3000'}},
         exposes: [e.contact(), e.battery(), e.battery_voltage()],
+        ota: ota.zigbeeOTA,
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

torandreroland:
@Otnow: Have you gotten OTA to work with MCCGQ12LM? Mine doesn't seem to support this cluster.

Otnow:
@torandreroland, I don't have this device, so unfortunately I can't check OTA.
You can try searching in issues or opening a new one with this problem.

torandreroland:
@Otnow: Can you recall the reason for opening this issue in the first place? Was it only done because we found the OTA file?

I'm thinking of removing the genOta cluster again since my hypothesis is that the device don't support it, but just trying to make sure.

Otnow:
@torandreroland, yes, that's right, only based on the OTA file.

PR #5540, opened 2023-03-03
Fix setting system_mode for TS0601_thermostat_1

Delorien84:
Fixed https://github.com/Koenkk/zigbee2mqtt/issues/16881


Koenkk:
Thanks for the quick fix!

PR #5542, opened 2023-03-03
Update PS600 Config

mikeyyuen:
Addresses all remaining issues raised in https://github.com/Koenkk/zigbee2mqtt/issues/16272

- Remove redundent zigbeeModels now #5485 is merged
- Add battery support in the config for 3v battery 2.5v LowVoltage Limit


Koenkk:
Thanks!

PR #5546, opened 2023-03-04
Add VOC index to IKEA Vindstyrka

0ip:
Requires new cluster: https://github.com/Koenkk/zigbee-herdsman/pull/674

Thanks to https://github.com/dresden-elektronik/deconz-rest-plugin/issues/6756#issuecomment-1453324669


converters/fromZigbee.js
```diff
@@ -4759,6 +4759,15 @@ const converters = {
             }
         },
     },
+    ikea_voc_index: {
```
sjorge, line 4762: Since this is IKEA specific, you should move this to devices/ikea.js where the other ikea specific from/toZigbee converters are.

lib/exposes.js
```diff
@@ -643,6 +643,7 @@ module.exports = {
         valve_detection: () => new Switch().withState('valve_detection', true).setAccess('state', access.STATE_SET),
         vibration: () => new Binary('vibration', access.STATE, true, false).withDescription('Indicates whether the device detected vibration'),
         voc: () => new Numeric('voc', access.STATE).withUnit('ppb').withDescription('Measured VOC value'),
+        voc_index: () => new Numeric('voc_index', access.STATE).withDescription('Sensirion VOC index'),
```
sjorge, line 646: Do we know if this measurement is also ppb? If it's better to reuse the above voc() exposes to keep things consistent.

0ip: It's a unitless, relative property invented by Sensirion. I wrote a bit about it [here](https://github.com/Koenkk/zigbee2mqtt/issues/16717#issuecomment-1436874652). There's also a short [summary PDF](https://sensirion.com/media/documents/02232963/6294E043/Info_Note_VOC_Index.pdf) from the company itself about it.

converters/fromZigbee.js
0ip: Thank you, fixed

lib/exposes.js
sjorge: Interesting, makes me wonder if we got the Develco one wrong also, I got my hands on this when adding support for it via their support e-mail.

[Sensirion_Gas_Sensors_SGP3x_TVOC_Concept.pdf](https://github.com/Koenkk/zigbee-herdsman-converters/files/10888459/Sensirion_Gas_Sensors_SGP3x_TVOC_Concept.pdf)

But it wouldn't surprise me that the values it's spewing is not ppb either, but it at least does not seem to be a relative value like you're describing IKEA is using.

Value range from 0 -> ~30k

image

So indeed separating them makes sense here.

0ip: So, in contrast to the SEN54 the Vindstyrka is using, the SGP30 appears to actually output tVOC in ppb.

Koenkk:
Thanks!

PR #5547, opened 2023-03-04
Develco AQSZB-110 should use μg/m3 for voc

sjorge:
- changed the expose.voc() default to use μg/m3 (now inline with things like pm10, pm25, ...)
- updated Develco AQSZB-110 to output μg/m3 for voc, fixing https://github.com/Koenkk/zigbee2mqtt/issues/16057  for this device
- Heiman HS2AQ-EM already seems to output μg/m3 from what I could find, should now also be fixed.
- LifeControl and Aqara devices using the expose now have withUnit('ppb') added as they all seem to return ppb values and I couldn't easily located a formulate to convert that to μg/m3.


Koenkk:
Enorm bedankt weer! 😄

Koenkk:
@sjorge done!

RobLeighton67:
@sjorge been working recently with both the Develco AQSZB-110 and the Heiman HS2AQ-EM. 
Just to clarify, the HS2AQ report HCHO in mg/m3. 
Even then the HS2AQ display shows for example 0.01 mg/m3 HCHO but Z2M publishes as HCHO : 0.1
All very confusing!

sjorge:
The HS2AQ probably needs a /100 on the value then.

PR #5548, opened 2023-03-04
Fix discarded 0 PM2.5 measurements

0ip:
If I am not mistaken, the current implementation prevents measurements with the value 0. In the case of PM2.5, however, such values can certainly be measured.


Koenkk:
Thanks!

PR #5549, opened 2023-03-04
Add another OEM variant of Tuya SM0201

vangual:
This adds support for an OEM variant of the Tuya SM0201 sensor sold under the Marmitek brand and named "Sense MO".
The device yields correct values once the different manufacturerName is recognized.

Looking at their website the brand carries a lot of the other Tuya OEM Zigbee devices as well but I can only verify the manufacturerName for this device currently.


Koenkk:
Thanks!

PR #5552, opened 2023-03-05
add support for play light tube

chrisdrackett:
There is a short version as well, but I'm not sure of the model numbers.


Koenkk:
thanks @chrisdrackett !

PR #5557, opened 2023-03-06
Add preheat support for thermostats with scheduling like Danfoss Ally

tbarnekov:
Danfoss Ally support an adaptive preheating function when running in scheduler mode. The adaptive preheating is enabled by setting bit 1 of programingOperMode to 1 (which corresponds to Economy/EnergyStar mode in the Zigbee Cluster Specification 6.3.2.2.3.6).


Koenkk:
Thanks!

PR #5564, opened 2023-03-10
Update Philips 915005987701 to support additional zigbee model

ahemagain:
I cannot confirm if the original zigbee model here is in error, or if Philips present the same lamp with multiple zigbeeModels.

I can however confirm that my version of this lamp does in fact use the amended model. 

This change therefore appends support for a known model that was otherwise unsupported.


Koenkk:
Thanks!

PR #5566, opened 2023-03-11
#5497 support for TRV Appartme APRM-04-001

bjab:
I have also this TRV and I had to add zigbeeModel, to make it discoverable


Koenkk:
Thanks!

PR #4929, opened 2022-11-11
Fix GiEX QT06 water valve units

drakhart:
I made some fixes and improvements to the GiEX QT06 water valve:
- Replaced minutes with seconds. I realized that this device measures in seconds as soon as I set 2 irrigation cycles of 1 minute each with 1 minute interval and it just went crazy with instant on and offs. I set 60 "minutes" instead and it behaved as expected.
- Adjusted the max values to `3600` (seconds in an hour) because the existing `1440` (minutes in a day) didn't make sense anymore and setting `86400` (seconds in a day) just sounded ridiculous to me.
- Spelled `litres` instead of `Litres` as uppercase is just for the abbreviation (L).
- Removed some unnecessary units, especially the incorrect `GMT+8` for times as the device uses GMT (Zulu time).
- Improved some descriptions, especially those referring to last irrigation info.

PR #5567, opened 2023-03-11
fix: restore sh1btabb spec, add a7sghmms

drakhart:
Apply fixes as discussed in https://github.com/Koenkk/zigbee-herdsman-converters/pull/4929
- Restore sh1btabb spec using minutes, added conversion from GMT+8 to local timezone.
- Add a7sghmms spec using seconds and local timezone.
- Small refactor: added constants, removed magic numbers...

### DIFF
--- a/devices/giex.js
+++ b/devices/giex.js
@@ -188,8 +188,7 @@ module.exports = [
                 .withValueMin(0)
                 .withValueMax(MINUTES_IN_A_DAY)
                 .withUnit('minutes or litres')
-                .withDescription('Irrigation target, duration in minutes or capacity in litres (depending on mode), ' +
-                    'set to 0 to leave the valve on indefinitely'),
+                .withDescription('Irrigation target, duration in minutes or capacity in litres (depending on mode)'),
             exposes.numeric(keys.giexWaterValve.cycleIrrigationInterval, ea.STATE_SET)
                 .withValueMin(0)
                 .withValueMax(MINUTES_IN_A_DAY)
--- a/devices/giex.js
+++ b/devices/giex.js
@@ -1,135 +1,203 @@
 const exposes = require('../lib/exposes');
 const tuya = require('../lib/tuya');
-const e = exposes.presets;
+
+const ep = exposes.presets;
 const ea = exposes.access;
 
+const CAPACITY = 'capacity';
+const DURATION = 'duration';
+const MINUTES_IN_A_DAY = 1440;
+const OFF = 'OFF';
+const ON = 'ON';
+const SECONDS_IN_12_HOURS = 43200;
+
+const toLocalTime = (time, timezone) => {
+    if (time === '--:--:--') {
+        return time;
+    }
+
+    const local = new Date(`1970-01-01T${time}.000${timezone}`);
+    return local.toTimeString().split(' ').shift();
+};
+
+const keys = {
+    giexWaterValve: {
+        battery: 'battery',
+        currentTemperature: 'current_temperature',
+        cycleIrrigationInterval: 'cycle_irrigation_interval',
+        cycleIrrigationNumTimes: 'cycle_irrigation_num_times',
+        irrigationEndTime: 'irrigation_end_time',
+        irrigationStartTime: 'irrigation_start_time',
+        irrigationTarget: 'irrigation_target',
+        lastIrrigationDuration: 'last_irrigation_duration',
+        mode: 'mode',
+        state: 'state',
+        waterConsumed: 'water_consumed',
+    },
+};
+
 const dataPoints = {
-    // GIEX Water Valve
-    giexWaterValveState: 2,
-    giexWaterValveMode: 1,
-    giexWaterValveIrrigationTarget: 104,
-    giexWaterValveWaterConsumed: 111,
-    giexWaterValveIrrigationStartTime: 101,
-    giexWaterValveIrrigationEndTime: 102,
-    giexWaterValveLastIrrigationDuration: 114,
-    giexWaterValveBattery: 108,
-    giexWaterValveCycleIrrigationNumTimes: 103,
-    giexWaterValveCycleIrrigationInterval: 105,
-    giexWaterValveCurrentTempurature: 106,
+    giexWaterValve: {
+        battery: 108,
+        currentTemperature: 106,
+        cycleIrrigationInterval: 105,
+        cycleIrrigationNumTimes: 103,
+        irrigationEndTime: 102,
+        irrigationStartTime: 101,
+        irrigationTarget: 104,
+        lastIrrigationDuration: 114,
+        mode: 1,
+        state: 2,
+        waterConsumed: 111,
+    },
 };
 
 const fzLocal = {
-    giex_water_valve:
+    giexWaterValve: (timezone) =>
+        ({
+            cluster: 'manuSpecificTuya',
+            type: ['commandDataResponse', 'commandDataReport'],
+            convert: (model, msg, publish, options, meta) => {
+                for (const dpValue of msg.data.dpValues) {
+                    const value = tuya.getDataValue(dpValue);
+                    const dp = dpValue.dp;
+                    switch (dp) {
+                    case dataPoints.giexWaterValve.state:
+                        return {[keys.giexWaterValve.state]: value ? ON: OFF};
+                    case dataPoints.giexWaterValve.mode:
+                        return {[keys.giexWaterValve.mode]: value ? CAPACITY: DURATION};
+                    case dataPoints.giexWaterValve.irrigationTarget:
+                        return {[keys.giexWaterValve.irrigationTarget]: value};
+                    case dataPoints.giexWaterValve.cycleIrrigationNumTimes:
+                        return {[keys.giexWaterValve.cycleIrrigationNumTimes]: value};
+                    case dataPoints.giexWaterValve.cycleIrrigationInterval:
+                        return {[keys.giexWaterValve.cycleIrrigationInterval]: value};
+                    case dataPoints.giexWaterValve.waterConsumed:
+                        return {[keys.giexWaterValve.waterConsumed]: value};
+                    case dataPoints.giexWaterValve.irrigationStartTime:
+                        return {[keys.giexWaterValve.irrigationStartTime]: timezone ? toLocalTime(value, timezone) : value};
+                    case dataPoints.giexWaterValve.irrigationEndTime:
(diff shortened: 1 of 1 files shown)


devices/giex.js
```diff
@@ -1,135 +1,203 @@
...
+                    case dataPoints.giexWaterValve.cycleIrrigationInterval:
+                        return {[keys.giexWaterValve.cycleIrrigationInterval]: value};
+                    case dataPoints.giexWaterValve.waterConsumed:
+                        return {[keys.giexWaterValve.waterConsumed]: value};
+                    case dataPoints.giexWaterValve.irrigationStartTime:
+                        return {[keys.giexWaterValve.irrigationStartTime]: timezone ? toLocalTime(value, timezone) : value};
+                    case dataPoints.giexWaterValve.irrigationEndTime:
+                        return {[keys.giexWaterValve.irrigationEndTime]: timezone ? toLocalTime(value, timezone) : value};
+                    case dataPoints.giexWaterValve.lastIrrigationDuration:
+                        return {[keys.giexWaterValve.lastIrrigationDuration]: value.split(',').shift()}; // Remove meaningless ,0 suffix
```
drakhart, line 82: Note: for `_TZE200_a7sghmms` I've always seen `,0` (i.e. `00:02:10,0` instead of a proper fraction of a second, that's why I'm removing this.

devices/giex.js
```diff
@@ -1,135 +1,203 @@
...
+                .withDescription('Last irrigation water consumption'),
+        ],
+    },
+};
+
 module.exports = [
+    // _TZE200_a7sghmms uses seconds, timezone is local
     {
+        ...exportTemplates.giexWaterValve,
+        model: 'QT06 (a7sghmms)',
```
drakhart, line 162: @Koenkk I have a conflict with the model name: if I reuse it then the `Verify definitions` test fails with a `Duplicate model QT06` error, but I if specify a different model name then the link to documentation is broken. Is there any elegant way to solve this? Or should we just add new documentation pages for these models? Since there are slight differences in behavior.

drakhart:
I'm converting this to draft until we clarify this doubt about model names (the current ones in the PR are quite unfriendly): https://github.com/Koenkk/zigbee-herdsman-converters/pull/4929#issuecomment-1465049068

WhimsySpoon:
@drakhart I've tried the new file as a custom converter and have noticed a couple of issues on a `QT06 (sh1btabb)`:

- The device image no longer appears in the Z2M interface
- The `irrigation_start_time` and `irrigation_end_time` are now GMT+1

Everything else looks great and it's correctly identified as a model QT06 (sh1btabb) in the UI.

devices/giex.js
Koenkk: I propose to name them `QT06_1` and `QT06_2`

drakhart:
> * The device image no longer appears in the Z2M interface

This is most likely related to the change in model names, as well as the broken documentation link I mention in this comment: https://github.com/Koenkk/zigbee-herdsman-converters/pull/5567#discussion_r1133153610

I guess we'll need to create new markdown files and images in the main Z2M repo.

> * The `irrigation_start_time` and `irrigation_end_time` are now GMT+1 (I'm GMT local)

That's weird. With this conversion I'm getting exactly 7 hours less than the time reported by the device, which is the proper one for me (I'm on CET / GMT+1).

The 1900 fix sounds a bit hacky to me since Unix epoch starts in January 1st 1970, and also feels flaky because setting that I get a weird time difference (i.e. it converts `13:25:08` to `05:10:08` instead of he correct `06:25:08`).

Can you please check the system time on the machine you're running Z2M in, and make sure it's set to the proper timezone? Example:

```
❯ date
Sun Mar 12 13:16:58 CET 2023
```

WhimsySpoon:
Agreed on the hackiness; I can't explain why it makes a difference. I did some debugging and the timezone is GMT+8 (I'm in GMT) and the time was coming through as 8 hours from now.

I'm running Z2M Add-on in Home Assistant OS and the time is correct:

image

drakhart:
This feels very, very weird. Would you mind putting a year somewhere in the middle of epoch time, let's say 2000, instead of 1900 or 1970? Just to make sure this is not some edge case caused by the conversion returning a value below epoch's minimum date.

WhimsySpoon:
I changed it to:

````
const local = new Date(`2000-01-01T${time}.000${timezone}`);
````

And it's calculating the timezone correctly. I don't understand it. The code as is, with 1970, works perfectly outside of this environment.

**Comment by drakhart** [comment]
I can't explain it either, just wanted to make sure. But seeing that 2000 works properly for both you and me I'll assume that it will also work for people in other timezones, so I'll apply it to the PR.

**Comment by WhimsySpoon** [comment]
Sounds good. Happy to test anything else as needed.

File: devices/giex.js (hunk 2 above)
Comment (drakhart) [reply devices/giex.js:162]: Ok, I'll push those names and open the related PR in zigbee2mqtt.io asap.

**Comment by drakhart** [comment]
Opened the following PR for the docs & images: https://github.com/Koenkk/zigbee2mqtt.io/pull/1947

File: devices/giex.js [hunk 3]
```diff
@@ -1,135 +1,202 @@
...
+        ],
+    },
+    // _TZE200_sh1btabb uses minutes, timezone is GMT+8
+    {
+        ...exportTemplates.giexWaterValve,
+        model: 'QT06_2',
+        fingerprint: [
+            {modelID: 'TS0601', manufacturerName: '_TZE200_sh1btabb'},
+        ],
+        timezone: '+08:00',
```
Koenkk, line 187: Please don't add extra props on the definition level. I suggest to use the following in the converter:

```
 const time = model.model == 'QT06_2' ? '+08:00' : null
```

drakhart: Hmmm... I didn't want to add this model logic directly into the converter, just in case it keeps growing. I'll look for some other solution without polluting the model object, like a config object or something. 👍

WhimsySpoon:
I've tested the current changes and the `irrigation_end_time` no longer changes to `--:--:--` when the tap is open.

Instead, when the tap opens, the irrigation_end_time is the `irrigation_start_time` \ now plus one minute.

drakhart:
> I've tested the current changes and the `irrigation_end_time` no longer changes to `--:--:--` when the tap is open.
> 
> Instead, when the tap opens, the irrigation_end_time is the `irrigation_start_time` \ now plus one minute.

It's happening for me too and I didn't notice, thanks for the heads up. It's really weird though, let me look into it.

WhimsySpoon:
I did have a brief go at debugging but didn't have enough time. All I can be sure of is that it's not making into the function that converts the time to local.

drakhart:
I'm getting the following behaviors:
- If I press the physical button on the device end time is `--:--:--` until I press it again, as expected.
- If I switch it to on from Z2M in capacity mode the end time is also `--:--:--`, as expected.
- If I switch it to on from Z2M in duration mode the end time is start time + duration, unless I manually stop it (it then updates to the new end time). I'm not sure, but I'd say this is expected too.

All in all, I don't see anything unusual.

Are you sure it doesn't get into the `toLocalTime` function? It works fine in my system if I change the model names (I don't own the GTM+8 model, so I need to force it somehow).

WhimsySpoon:
I'll need to double check when I get home tonight, but previous versions definitely had the end time as dashes whilst the tap was open. I'll have a play later and see if I can find out where `start_time`+00:01:00 is getting injected.

Koenkk:
Looks good now, please let me know once this is ready for merge.

WhimsySpoon:
@drakhart Apologies for the delay on this one. I've conducted some thorough testing and was really struggling to get my head around this. I effectively debugged each line of relevant code to see where it was getting overwritten, but could never see where `--:--:--` was being lost.

I then decided to roll-back to the previous version of the converter, and it was still doing it! I went as far as pulling the converter entirely. Still doing it. I then wondered if it was something set on the device itself, so I removed and repaired it. Still doing it.

After sleeping on the issue, I then toggled the `mode` which had at some point been set back to `duration`. Changing it back to `capacity` had it working back again, showing `--:--:--` in `irrigation_end_time`. It seems that having `mode` set to `duration` and `irrigation_target` at `0` meant that the `irrigation_end_time` would behave as I discovered above.

My huge apologies, @drakhart, there was nothing wrong with your code at all. I've just tested the latest code in this PR and it's working perfectly when using `mode` as `capacity`.

WhimsySpoon:
I think I spoke to soon. It looks like the issue is with it not respecting `irrigation_target` when it's set to `0`, regardless of mode, it only opens with the physical button.

When the `irrigation_target` is set to `0` and the valve is opened via Z2M, something is happening inside the device (a small click can be heard, but it's not the same sound as the valve opening up), it then immediately closes again. Using the physical button works fine.

I think this may be a limitation of this devices firmware which is possibly solved in the later version. The only work-around I can find is setting the `irrigation_target` to the largest possible value.

Either way, it's not a problem with this code as the issue is present with the current code and this release.

Apologies for the back-and-forth on this one, @drakhart, and thanks again for all of your efforts.

drakhart:
No problem, @WhimsySpoon, better be safe than sorry. Thanks for taking your time double-checking all these things. ❤️

It indeed sounds like a limitation with `sh1btabb`, because setting the irrigation target to zero on `a7sghmms` leaves the valve open until manually closed. Let me update `sh1btabb` description removing this mention to the zero value to avoid misleading the end user and we'll be good to merge.

drakhart:
@Koenkk we're all good, please merge when you have the time. 👍 Thanks.

Koenkk:
Thanks!

fedegrc:
I have tested the `_TZE200_a7sghmms` several times this week with no problems. Thank @drakhart! Good job

WhimsySpoon:
@drakhart The UK recently moved to British Summer Time, or BST which is GMT+1 and the local times are still showing as GMT, so an hour behind current.

You're not going to believe this, but changing the following back to 1970 fixes this:

````
const local = new Date(`2000-01-01T${time}.000${timezone}`);
````

I'm going to have a play with this and try and get to the bottom of it before I open a new PR.

drakhart:
> @drakhart The UK recently moved to British Summer Time, or BST which is GMT+1 and the local times are still showing as GMT, so an hour behind current.
> 
> You're not going to believe this, but changing the following back to 1970 fixes this:
> 
> ```
> const local = new Date(`2000-01-01T${time}.000${timezone}`);
> ```
> 
> I'm going to have a play with this and try and get to the bottom of it before I open a new PR.

Ok, please let me know your findings and let's see if we can get this sorted out once and for all. 🤞

PR #5572, opened 2023-03-12
Fix missing converters for pi_heating_demand on Bosch `BTH-RA` thermostat

DerDreschner:
The documentation for the Bosch BTH-RA states that the `pi_heating_demand` entity should be read- and set-able. In reality, this isn't the case due to missing converters between the entity and the Zigbee attribute 0x4020 in cluster 0x0201. This PR adds that conversion.

The valve position is fixed to this position for 15 minutes, regardless of the `occupied_heating_setpoint` or `system_mode` entity. Only enabling the `boost` entity overwrites this behavior. After 15 minutes without change to the `pi_heating_demand` entity, the thermostat starts to control the valve position again under consideration of the current settings. 

This issue fixes https://github.com/Koenkk/zigbee2mqtt/issues/16538. It's tested and working with firmware version v3.04.04. Please note, that some people report that the valve position isn't changed as stated by the `pi_heating_demand` entity in firmware versions earlier to v3.04.04. Therefore, it may be necessary to update the firmware.

### DIFF
--- a/devices/bosch.js
+++ b/devices/bosch.js
@@ -49,7 +49,7 @@ const displayOrientation = {
 // Radiator Thermostat II
 const tzLocal = {
     bosch_thermostat: {
-        key: ['window_open', 'boost', 'system_mode'],
+        key: ['window_open', 'boost', 'system_mode', 'pi_heating_demand'],
         convertSet: async (entity, key, value, meta) => {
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

PR #5573, opened 2023-03-13
Enhance functionality of Xiaomi `SP-EUC01`

DerDreschner:
Changes for `SP-EUC01` (tested with firmware `0043`):
- Expose already existing `consumer_connected` entity
- Enable functionality to disable LED at night time
- Make overload protection threshold configurable
- Enable auto-off feature after 20 minutes without significant power draw
- Make button on `SP-EUC01` lockable

Additionally, I did some cleanup by moving multiple used entities to `exposes.js` with variables instead of defining them from scratch every single time. In case an entity I worked with didn't had a get-converter, but is able to be read, I added them.


Koenkk:
thanks!

PR #5577, opened 2023-03-14
Add support for Lupus 12071 LUPUSEC main power meter

BuethSam:
Add Support for the Lupus main power meter

Required a modified metering converter because the divisor needed to be configurable by the user depending on the installed meter. The converter creates an option to configure the impulses per kWh of the meter. (eg. 10000 Imp./kWh in the image below) 

ArtNr_12071_Hauptstromzaehler_demo1_150dpi

I also open a PR to add an image https://github.com/Koenkk/zigbee2mqtt.io/pull/1952

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -738,6 +738,37 @@ const converters = {
             return payload;
         },
     },
+    impulse_metering: {
+        /**
+         * When using this converter also add the following to the configure method of the device:
+         * endpoint.saveClusterAttributeKeyValue('seMetering', {multiplier: 10});
+         */
+        cluster: 'seMetering',
+        type: ['attributeReport', 'readResponse'],
+        options: [
+            exposes.numeric('impulses', exposes.access.SET)
+                .withValueMin(1)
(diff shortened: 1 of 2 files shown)


converters/fromZigbee.js
```diff
@@ -738,6 +738,37 @@ const converters = {
...
+        options: [
+            exposes.numeric('impulses', exposes.access.SET)
+                .withValueMin(1)
+                .withValueStep(1)
+                .withUnit('Imp./kWh')
+                .withDescription('Configure the impulses per kWh of the power meter. '),
+        ],
+        convert: (model, msg, publish, options, meta) => {
+            if (utils.hasAlreadyProcessedMessage(msg, model)) return;
+            const payload = {};
```
Koenkk, line 757: Can you use the existing metering converter and apply the impulses afterwards? Use this here: `const payload = fz.metering(model, msg, publish, options, meta)`

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1952

PR #5585, opened 2023-03-16
Added light 1/3 of the Hue Runner triple spotlight

dschoorisse:
The Hue Runner spotlight comes with three Ambiance spots with (seemingly) ascending model numbers. However, just two of the three of them are in the device list and were recognized.

Scherm­afbeelding 2023-03-16 om 21 49 15


Koenkk:
Thanks!

PR #5594, opened 2023-03-19
WS-USC03 & ROM001: remove default behavior of binding genOnOff to coordinator

JJPro:
## Related Devices: 
- WS-USC03: Aqara wall switch (controlling my bathroom light)
- ROM001: Philips Hue Smart Button

## Problems I had & solution: 
Both devices binds `genOnOff` to coordinator by default on configuration, which causes unexpected behavior of Hue smart button toggling my bathroom light immediately after pairing to z2m, with no automations configured between them. 

I don't get the point of binding `genOnOff` (same for `genLevelCtrl`) at pairing by default, thus rendering this PR to remove them. 

Any thoughts?

### DIFF
--- a/devices/philips.js
+++ b/devices/philips.js
@@ -1919,8 +1919,6 @@ module.exports = [
         exposes: [e.battery(), e.action(['on', 'off', 'skip_backward', 'skip_forward', 'press', 'hold', 'release'])],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
-            await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl']);
-
             const options = {manufacturerCode: 0x100B, disableDefaultResponse: true};
             await endpoint.write('genBasic', {0x0031: {value: 0x000B, type: 0x19}}, options);
             await reporting.bind(endpoint, coordinatorEndpoint, ['manuSpecificPhilips', 'genPowerCfg']);
--- a/devices/xiaomi.js
+++ b/devices/xiaomi.js
@@ -1030,7 +1030,6 @@ module.exports = [
         extend: extend.switch(),
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
-            await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff']);
             await reporting.onOff(endpoint);
         },
         ota: ota.zigbeeOTA,


Koenkk:
I don't see how this can make your bathroom lights toggle. They should not be listening to address `0x0000` (coordinator address). Also, this may prevent `on`/`off`/`toggle` actions being published (as they are not send to the coordinator anymore).

JJPro:
Thanks for the info. I recovered default bindings and amazingly everything works fine now. Have been running for a night, the issue never recur. 

---
I have a doubt about the last sentence in your message: 
> this may prevent on/off/toggle actions being published (as they are not send to the coordinator anymore).
I think you were talking about **Reporting**, not **Binding**, were you?

If I understand correctly, **Binding** is a one-way thing, isn't it? Then I guess removing **Binding** shouldn't affect **Reporting**, right? 

If that's the case, I still don't get the point of binding `genOnOff` to Coordinator. What does it do? (Indeed many devices have this binding, browsing through the codebase)

JJPro:
Can you explain what does it do to send `onOff` to Coordinator? Thanks!

Koenkk:
It doesn't send `onOff`, but all commands of the `onOff` cluster which are supported by the device: (e.g. `commandOn` which produces  `{"action": "on"}`. https://github.com/Koenkk/zigbee-herdsman/blob/e264ce0634a49a53b06e7cf5abaf5198106aa225/src/zcl/definition/cluster.ts#L475

JJPro:
Sorry my question was not clear, I meant to ask what the effects are of sending, e.g. `commandOn`, to the Coordinator. 

I understand you bind a switch's `genOnOff` to a device, so to toggle the device directly without z2m; but I don't understand what it does to bind `genOnOff` to the Coordinator?

Koenkk:
When the device send commandOn the the coordinator, the coordinator will forward this to Zigbee2MQTT which will publish `{"action":"ON"}` (which can then be used in your automations). Without any binds an `action` will never be published.

JJPro:
Okay, I think I get how binding and reporting works for different devices now. 

Sensors always report to Coordinator, and the reportings are not configurable, so there is no binding and reporting configurations needed for sensors;
But in order for switches to report to Coordinator, you have to first bind it to the Coordinator (otherwise, the device won't know who to report to), then configure reporting in **Reporting** tab. 

Is that right?

Koenkk:
@JJPro correct!

JJPro:
THANK YOU!

PR #5596, opened 2023-03-19
add SONOFF SA-028 to definition of Z111PL0H-1JX

### LINKED ISSUE #5590: Duplicate Model number 'Z111PL0H-1JX'
I was attempting to add support for my Woolley SA-028, US-style switch plug. It reports model number Z111PL0H-1JX and vendor 'SONOFF'. Upon adding it to the [sonoff.js](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5589/files#diff-371be97b35202a0b01c5397ac28e33156275c238d29ce61d31fb1ed47797203a) and creating a pull request #5589, I was notified by a failing test that the model number was a duplicate. I see now that there is a [woolley.js](https://github.com/Koenkk/zigbee-herdsman-converters/blob/d2c37da6717dedafb1be8e879de3cb3e2b81fa23/devices/woolley.js#L6) in which it is listed.

It is my understanding that the SA-029 defined in the existing woolley.js is the EU version, whereas the SA-028 is the US version. Documentation also lists a model SA-030 which I suspect will have the same zigbee model number.
What is the best way to handle the different region devices using the same zigbee model number?

mikedm139:
… and move from woolley.js to sonoff.js

Closes #5590 and replaces PR #5589.


Koenkk:
Thanks!

PR #5597, opened 2023-03-19
Update Moes TS0601 thermostat max temperature and local temperature calibration (_TZE200_cpmgn2cf)

xamadevp:
This TRV's range for the max temperature is 16-70 °C and it supports 0.5 °C steps for local temperature calibration


Koenkk:
Thanks!

PR #5599, opened 2023-03-21
Bosch `BTH-RA`: Respond to DST attribute requests with value `0x00` to avoid non-responding behaviour

DerDreschner:
The Bosch `BTH-RA` not only wants the current time, it also requests the 3 DST related attributes `dstStart`, `dstEnd` and `dstShift`. It is provided by the Bosch Smart Home Controller when pairing the thermostat as well as regularly when requesting the current time:

image
image

According to the [Zigbee Cluster Library Specification](https://zigbeealliance.org/wp-content/uploads/2019/12/07-5123-06-zigbee-cluster-library-specification.pdf) (point 3.12.2.2.4 to 3.12.2.2.6), these fields are optional but have to be set together when they are provided. As it's optional and DST time changes are being considered in the `new Date()).getTimezoneOffset()` call in zigbee-herdsman (https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/model/device.ts#L232), these fields are not supported by default and are reported as `UNSUPPORTED_ATTRIBUTE`.

Although it is optional by spec, it appears the Bosch `BTH-RA` doesn't behave very well when these attributes are not set by the coordinator. The thermostat requests these information every 5 minutes up to at least every hour. After some time in this state (couldn't determine on what it depends), the thermostat stops processing attribute change requests all together which leads to an more-or-less useless device. Bringing back the device from this "dead" state requires a hard reset by removing the batteries.

As workaround, I implemented a respond with value `0x00` to these attributes which seems to work. My thermostats didn't got into these non-responding state again and work well for a week now. That's why I'm keen to submit my changes now.

### DIFF
--- a/devices/bosch.js
+++ b/devices/bosch.js
@@ -227,6 +227,21 @@ const fzLocal = {
             return result;
         },
     },
+    bosch_ignore_dst: {
+        cluster: 'genTime',
+        type: 'read',
+        convert: async (model, msg, publish, options, meta) => {
+            if (msg.data.includes('dstStart', 'dstEnd', 'dstShift')) {
+                const response = {
+                    'dstStart': {attribute: 0x0003, status: herdsman.Zcl.Status.SUCCESS, value: 0x00},
+                    'dstEnd': {attribute: 0x0004, status: herdsman.Zcl.Status.SUCCESS, value: 0x00},
+                    'dstShift': {attribute: 0x0005, status: herdsman.Zcl.Status.SUCCESS, value: 0x00},
+                };
+
+                await msg.endpoint.readResponse(msg.cluster, msg.meta.zclTransactionSequenceNumber, response);
+            }
+        },
+    },
     bosch_thermostat: {
         cluster: 'hvacThermostat',
         type: ['attributeReport', 'readResponse'],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

BBJake:
@DerDreschner 
Very interesting. Just saw this in the changelog of the latest Z2M release.
Have you seen this big discussion? https://github.com/Koenkk/zigbee2mqtt/issues/13478

The problems that are described there seem to affect the Bosch thermostats as well.
So I'm wondering why your changes might fix the problems, because this seems to be a general problem with Zigbee 3.x devices.

mldytech:
> @DerDreschner 
> Very interesting. Just saw this in the changelog of the latest Z2M release.
> Have you seen this big discussion? https://github.com/Koenkk/zigbee2mqtt/issues/13478
> 
> The problems that are described there seem to affect the Bosch thermostats as well.
> So I'm wondering why your changes might fix the problems, because this seems to be a general problem with Zigbee 3.x devices.
> 

I also hoped that this merge would fix the problems. Unfortunately, with the latest stable version, the issues with the bosch thermostats still persist.

Just curious: Is there a way to force a device to only connect with a router and not the coordinater? Or even to a specific router? This would be a nice quick and dirty fix since they are working flawless when not directly connected to the coordinater.

PR #5603, opened 2023-03-23
viessmann(TRV): exposing battery for HA mqtt discovering

ChrisChoke:
Hey Koen,

I moved to Home Assistant on weekend and i was wondering that the battery sensor was not created while mqtt discover.
would be nice to add it. many other devices we exposing battery sensor as default.

greetz
Chris


Koenkk:
Thanks!

PR #5607, opened 2023-03-24
Add FUT035Z+ support

agglrx:
This PR adds support for the Miboxer FUT035Z+ (2 in 1 variant of FUT035Z) device. 

https://miboxer.com/en/product/2-in-1-led-controller-zigbee-3-0-2-4g


Koenkk:
This device should already be supported and be detected as a TS0502B.

agglrx:
Unfortunately this was not the case. It showed up as "unsupported" until I added a new definition with this specific fingerprint (`manufacturerName: _TZB210_lmqquxus`).

Koenkk:
What z2m version are you running? Make sure to try with 1.30.2

PR #5613, opened 2023-03-26
Missing support for KnockautX Smoke sensor "SMOAL024"

### LINKED ISSUE #5609: Missing support for KnockautX Smoke sensor "SMOAL024"
The [KnockautX SMOAL024](https://www.brelag.com/shop/smoal024-knockautx-rauchwarnmelder-4714#attr=) is a smoke detector based on the popular [tuya stack](https://developer.tuya.com/en) and integration should not be too dificulat as there is already a freamework for integration of such devices present.

According to the manfuctuer/distributor the following data points are available:
dp

As support for a new devices does not touch the zigbee2mqtt core code I decided to create the issue here in the converters repo. Let me know if this is the wrong place.

As the additional of a new device is well documented I'll give it a go myself - but would like to create the issue here for tracking purpose in case someone else needs support for this device.

pgfeller:
Fixes #5609


Koenkk:
This device seems to be the same as https://github.com/Koenkk/zigbee-herdsman-converters/blob/17c4e26c6028cad5d77563a1bdc6eeebda511050/devices/tuya.js#L2669 (if yes, please merge with that)

pgfeller:
This device is quite similar - but not the same, as it also exposes `[4, 'tamper', tuya.valueConverter.raw],`; which is not supported by the knockautx sensor.

Therefore I think it makes sense to handle this sensor as its own definition.

let me know if you agree. With kind regards, Patrik

Koenkk:
Then I propose to:
- Add this to `tuya.js` with `model` `TS0601_smoke_2`
- Add whitelabel KnockautX to it, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/3232113d25042cf93aea054a7cbcc0189ef2800d/devices/tuya.js#L732

pgfeller:
I've a question to this approach: In Zigbee2MQTT the model property of the definition seems to be used to look-up the device picture and doc. Is there a way to keep this functionality? That's why I would like to keep the model to be `SMOAL024`.

Of course I can mode the definition to tuya.js - but the vendor/brand of the device is 
[brelag/knockautx] (https://www.brelag.com/knockautx). They use the tuya chipset - but also have other devices. That's why I've create a new file - so that all devices from this vendor would go there. And the tuya.js file already contains more than 4'000 lines - which can also be difficult to find things.

Of course you've the final call what we do ... but the device picture/docu in zigbee2mqtt would be nice 🙂.

Koenkk:
@pgfeller I'll think about doing something with the picture, but I want to prevent many duplications of the same device. E.g. next month there could be the same device from a different whitelabel manufacturer.

pgfeller:
> @pgfeller I'll think about doing something with the picture, but I want to prevent many duplications of the same device. E.g. next month there could be the same device from a different whitelabel manufacturer.

@Koenkk  thank you for the reivew. I've moved the code to `tuya.js` and added the `whitelabel` entry. Maybe an approach for the images could be that we add the `manufacturerName` to the `whiteLabel` object:

```javascript
fingerprint: [
    {modelID: 'TS0601', manufacturerName: '_TZE200_ux5v4dbd'},
],
whiteLabel: [
    {vendor: 'KnockautX', model: 'SMOAL024', manufacturerName: '_TZE200_ux5v4dbd'},
],
```
This would allow to lookup the `model` for the image and metadata from the `whiteLabel` definitions of a matching device (would make this optional to keep compatibility).

with kind regards,
Patrik

pgfeller:
Did I make a mistake to merge the changes from `master` to be up-to-date with my branch? As I see now changes in the `diff` that I would not have expected (as they should be in `master` already ... .

Let me know if I need to revert the commit "🔃 Merge master up to https://github.com/Koenkk/zigbee-herdsman-converters/commit/d3e63af79145e362e5b5618f4319dab0dad7018e".

Koenkk:
> Let me know if I need to revert the commit "🔃 Merge master up to https://github.com/Koenkk/zigbee-herdsman-converters/commit/d3e63af79145e362e5b5618f4319dab0dad7018e".

Please do so such that we have a clean diff.

pgfeller:
... done - the diff looks clean now 👍

lib/tuya.js
```diff
@@ -419,6 +419,11 @@ const dataPoints = {
     childLock: 7,
     coverChange: 7,
     runningState: 14,
+    /**
+     * Data Transfer Type: Send and Report
+     * Data Type: Bool
+     */
+    muffling: 16,
```
Koenkk, line 426: Why was this added? (I do not see it being used)

pgfeller: I was in contact with the vendor to check what DPs are supported this device - and they claimed that it is also possible to mute the device via DP16 - but I could not verify this function with my three devices. It seems not to be working. Anyhow - as this is a common data point I kept it in the code for future useage to make it easier to identify the dunction of data points. But we can also remove this if you think we should not keep it in (as it is currently not used).

Koenkk: Please remove it and add it to https://github.com/Koenkk/zigbee-herdsman-converters/pull/5613/files#diff-7ae23d3c13c2d7f7e37106590684c791aba8093f8be813b7c74a4f1484017c3bR2730 as a comment.

Koenkk:
Thanks!

PR #5614, opened 2023-03-27
Add WSP403-E device

Undef-a:
Power meter/smart switch similar to WSP402. Tested and working with the current docker release plus this patch.


Koenkk:
Thanks!

PR #5615, opened 2023-03-27
Add Curtain Motor Controller (4512747) Namron

thoj:
This is just a white-label of sunrichter SR-ZG9080A. The psychical appearance is identical, including SR-ZG9080A designation.


Koenkk:
Thanks!

PR #5617, opened 2023-03-28
Added: Basic support for QBKG29LM - Aqara H1 (No Neutral, Triple Rocker)

samuelthng:
Tested, works exactly like the QBKG26LM, features all supported when model added to converters.


samuelthng:
Apologies, I did not update variables used in the external converter. will correct it.

Koenkk:
Thanks!

PR #5618, opened 2023-03-28
IHSENO Smart Button IH-K663

kirovilya:
https://aliexpress.ru/item/1005005266369364.html

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -4151,4 +4151,14 @@ module.exports = [
                 .withDescription('Volume of the alarm'),
         ],
     },
+    {
+        fingerprint: tuya.fingerprint('TS0041', ['_TZ3000_fa9mlvja']),
+        model: 'IH-K663',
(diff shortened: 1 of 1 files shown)


devices/tuya.js
```diff
@@ -4151,4 +4151,14 @@ module.exports = [
                 .withDescription('Volume of the alarm'),
         ],
     },
+    {
+        fingerprint: tuya.fingerprint('TS0041', ['_TZ3000_fa9mlvja']),
```
Koenkk, line 4155: This seems to be the same as https://github.com/Koenkk/zigbee-herdsman-converters/blob/2998f7b1c91239594b9d76710643249e2429bd28/devices/tuya.js#L1763

kirovilya: but without 'hold' action

Koenkk: Ah I see, thanks!

PR #5620, opened 2023-03-29
Add new manufacturer for Tuya TS0002_switch_module_2

bimusiek:
The module I am using (bough in Poland) is incorrectly recognised as TS0002, where it should be TS0002_switch_module_2. It has `power_on_behaviour` but no `switch_type` properties.


Koenkk:
Thanks!

PR #5625, opened 2023-04-01
SMaBiT AV2010/32: allow get/set of hysteresis low and high

sjorge:
Completes implementation for https://github.com/Koenkk/zigbee2mqtt/issues/17194

I did not add these to the exposes values, it's something you set once and can forget about. I also don't think a lot of people will want to change the default.

### DIFF
--- a/devices/bitron.js
+++ b/devices/bitron.js
@@ -5,6 +5,55 @@ const reporting = require('../lib/reporting');
 const extend = require('../lib/extend');
 const e = exposes.presets;
 const ea = exposes.access;
+const herdsman = require('zigbee-herdsman');
+
+const manufacturerOptions = {manufacturerCode: herdsman.Zcl.ManufacturerCode._4_NOKS};
+
+const bitron = {
+    fz: {
+        thermostat_hysteresis: {
+            cluster: 'hvacThermostat',
+            type: ['attributeReport', 'readResponse'],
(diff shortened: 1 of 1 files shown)


Koenkk:
> I did not add these to the exposes values, it's something you set once and can forget about. I also don't think a lot of people will want to change the default.

Even if set once, it should be exposed. Otherwise it doesn't end up in the docs (and cannot be set via the frontend) and this becomes an undocumented feature.

sjorge:
> > I did not add these to the exposes values, it's something you set once and can forget about. I also don't think a lot of people will want to change the default.
> 
> Even if set once, it should be exposed. Otherwise it doesn't end up in the docs (and cannot be set via the frontend) and this becomes an undocumented feature.

I was going to add it to the docs manually like for the ctrlSeqOpr stuff.

I vaguely remember something about having a settings like expose too, perhaps this is something that could be sent that way?

Or is that one only for device options?

Koenkk:
> I vaguely remember something about having a settings like expose too, perhaps this is something that could be sent that way?

That is only for device options (applied at the z2m side)

sjorge:
I’ll refactor this a bit more so it can be one expose entry.

sjorge:
Now looks like this
image

Koenkk:
Clean, thanks!

PR #5626, opened 2023-04-01
Adding basic support for ubisys router R0

peterbachmann:
Adding basic support for ubisys R0 router (https://www.ubisys.de/wp-content/uploads/ubisys-r0-technical-reference.pdf).
I'm a novice at zigbee and z2m, thus I struggle to understand the possible configuration options. The router does not seem to expose anything, but it would allow some limited configuration.
With the configuration added here, I was able to join the ubisys R0 in z2m, see "last seen" and "lqi" in Homeassistant and update the firmware through OTA.

Any advice how to complete the config is very welcome (@sjorge seems to have done most of ubisys converters, I hope he sees this :) )

Once this PR is ready to merge, I will also add a picture and description to the z2m repo.

### DIFF
--- a/devices/ubisys.js
+++ b/devices/ubisys.js
@@ -928,10 +928,9 @@ module.exports = [
     {
         zigbeeModel: ['R0 (5501)'],
         model: 'R0',
-        vendor: 'ubisys',
-        description: 'Zigbee Router R0',
-        supports: 'Router',
-        fromZigbee: [],
+        vendor: 'Ubisys',
+        description: 'Zigbee Router',
+        fromZigbee: [fz.linkquality_from_basic],
         toZigbee: [],
         exposes: [],
         ota: ota.ubisys,
--- a/devices/ubisys.js
+++ b/devices/ubisys.js
@@ -935,5 +935,5 @@ module.exports = [
         toZigbee: [],
         exposes: [],
         ota: ota.ubisys,
-    }
+    },
 ];
--- a/devices/ubisys.js
+++ b/devices/ubisys.js
@@ -925,4 +925,15 @@ module.exports = [
         },
         ota: ota.ubisys,
     },
+    {
+        zigbeeModel: ['R0 (5501)'],
+        model: 'R0',
+        vendor: 'ubisys',
+        description: 'Zigbee Router R0',
+        supports: 'Router',
+        fromZigbee: [],
+        toZigbee: [],
+        exposes: [],
+        ota: ota.ubisys,
+    }
 ];


devices/ubisys.js
```diff
@@ -925,4 +925,15 @@ module.exports = [
...
+        zigbeeModel: ['R0 (5501)'],
+        model: 'R0',
+        vendor: 'ubisys',
+        description: 'Zigbee Router R0',
+        supports: 'Router',
+        fromZigbee: [],
+        toZigbee: [],
+        exposes: [],
+        ota: ota.ubisys,
+    }
```
sjorge:
```suggestion
    },
```

Missing trailing ,

sjorge (approved):
Aside from a missing `,` this looks good to me.

And this is correct, R0 will only acts as a router so  there are no toZigbee/fromZigbee converters.

I don't own one my self, but looking at the spec it looks like even the build in menu thingy is almost completely empty aside from join/leave options.

devices/ubisys.js
peterbachmann: wow, you answered and fixed my mistake faster than I could understand the output of the lint checker and make a new commit ;-)
I was very unsure about the configure or meta section you added on other ubisys devices. m2z gives an error when I click reconfigure on the R0. If you have some ideas, I would be happy to try them with my R0. Unless you think they are unnecessary.

sjorge: There is nothing needed in the meta block, also this device won't need a configuration function either, as there is nothing to configure for it.

peterbachmann: Allright. Thanks a lot :) 
I will try to update the devices description in z2m and link the PR here.

peterbachmann: If I have your attention already, maybe I can ask you an unrelated question about LQI reportet in Homeassistant. I can see the LQI of each device on the Devices and Map pages of m2z but in HA (and over MQTT) only the LQI of the routers are reportet, not the end devices. I have found conflicting posts online (although I don't have the links at hand), one where Koenkk explains that this is how zigbee works, another where someone ammended the config of a coordinator to forward endpoint LQIs - at least that is how I (mis)understood. Do you know  which one is correct?

sjorge: I've never really dug into LQI values at all. I vaguely remember something about z2m showing what the device reported.

peterbachmann:
PR with documentation for ubisys R0: https://github.com/Koenkk/zigbee2mqtt.io/pull/1982

Koenkk:
Thanks!

PR #5636, opened 2023-04-06
Add support for 4512728

KristianAndersson:
Add support for Namron Zigbee 1 channel switch K2 - Black (4512728)


Koenkk:
Thanks!

PR #5639, opened 2023-04-07
Fix ZNCLDJ11LM position issue.

fenfenS:
pretty much the same as `zigbee-herdsman-converters/pull/4163`.
    Test with my own ZNCLDJ11LM deivces.


Koenkk:
Thanks!

PR #5297, opened 2023-01-08
Improvements for Bosch BTH-RA

compujuckel:
- Add input for remote temperature sensor. Setting `remote_temperature` will override the value of `local_temperature`.
- Add `displayed_temperature` to choose which temperature is shown on the thermostat display
- Fix range of `local_temperature_calibration`
- Add running state

Unfortunately I haven't figured out yet how to "exit" this remote temperature mode. Writing 0 just sets the internal temperature to 0°C.

PR #5640, opened 2023-04-07
Update Bosch BTH2 - Add Input for Remote Temperature

sestriel:
Working improvements from @compujuckel  ( #5297 ) 

- Add input for remote temperature sensor. Setting remote_temperature will override the value of local_temperature.
- Add displayed_temperature to choose which temperature is shown on the thermostat display

Both Features are working well - as expected, using them now for more than 4 Weeks.

### DIFF
--- a/devices/bosch.js
+++ b/devices/bosch.js
@@ -559,7 +559,7 @@ const definition = [
                 .withSetpoint('occupied_heating_setpoint', 5, 30, 0.5)
                 .withLocalTemperatureCalibration(-5, 5, 0.1)
                 .withSystemMode(['off', 'heat', 'auto'])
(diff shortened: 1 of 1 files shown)


Koenkk:
Cool, thanks!

mldytech:
Hi, sorry if this is the wrong place to ask - and thanks for your awesome work.

Is it possible to adjust the remote_temperature to 0.1 decimal point? Right now, it seems that it's only rounded to 0.5 steps. 

Best regards,
mldy

sestriel:
Hello @mldytech, you mean when the local temperature (remote) is showed on the device itself?. 
Its the TRV itself that only goes in .5 steps - its just rounded for the display.

PR #5641, opened 2023-04-07
Fix backlight_mode values for TuYa TS130F

jcugat:
I have a [TuYa TS130F](https://www.zigbee2mqtt.io/devices/TS130F.html) device which always shows this error in Home Assistant:

image

I _think_ this should be the correct way to fix it, since it's just the case that is incorrect. I've checked it in the `State` tab for this device in Zigbee2MQTT WebUI and this is what I get:

```
{
    "backlight_mode": "low",
    "calibration": "OFF",
    "calibration_time": 23.1,
    "linkquality": 10,
    "motor_reversal": "OFF",
    "moving": "STOP",
    "position": 100,
    "state": "OPEN"
}
```

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -1450,7 +1450,7 @@ module.exports = [
         whiteLabel: [{vendor: 'LoraTap', model: 'SC400'}],
         exposes: [e.cover_position(), exposes.enum('moving', ea.STATE, ['UP', 'STOP', 'DOWN']),
             exposes.binary('calibration', ea.ALL, 'ON', 'OFF'), exposes.binary('motor_reversal', ea.ALL, 'ON', 'OFF'),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

convicte:
The issue persists, even into `ver.1.31.0` released today:
image

Invalid option for select.XXXXXXX_window_shutter_backlight_mode: '' (valid options: ['low', 'medium', 'high'])

Template variable warning: 'dict object' has no attribute 'backlight_mode' when rendering '{{ value_json.backlight_mode }}'

convicte:
BTW, these regular - non-curtain switches suffer from the same exact problem.
The whole series exposes backlight for in-wall modules:
https://www.zigbee2mqtt.io/devices/TS0003_switch_module_1.html#tuya-ts0003_switch_module_1

PR #5646, opened 2023-04-09
Nous L13Z module correct converter linked

djbiti1:
Nous L13Z 2 channel switch module was being recognized incorrectly as a touch switch and switch_type wasn't available. Found a converter that works I linked the fingerprint and added the "L13Z" as a white label.


Koenkk:
Thanks!

PR #5649, opened 2023-04-11
Add homeetec vendor

elad-bar:


devices/homeetec.js
```diff
@@ -0,0 +1,277 @@
+const fz = require('../converters/fromZigbee');
```
Koenkk, line 1: Please move this into `tuya.js` in the standard format (so no things like `MODEL_CONFIGURATIONS`, `HANDLERS` or `getCurtainWithGangDefintion`). Since the format is different then the standard format, it makes it quite hard to understand what is happening here.

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/1994

devices/homeetec.js
elad-bar: Tried to separate the configuration from the logic, no problem,
Just a question about moving code to Tuya, since it's labeled as different vendor shouldn't it locate in different file?
Thanks

elad-bar:
@Koenkk, i noticed that TS0601_cover_3 added but not available in zigbee2mqtt.io,  
to fix conflict here with master, I change mine to TS0601_cover_4.

in zigbee2mqtt.io repo I left it the same to avoid conflicts over there, pls let me know how to proceed.

thanks

devices/tuya.js
```diff
@@ -721,6 +721,26 @@ const fzLocal = {
             return result;
         },
     },
+    homeetec_cover: {
```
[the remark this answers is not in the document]
elad-bar: homeetec device consisted with 2 types of controls - curtain switch + 1|2 switches,
switches and additional controls (child lock, calibration, backlight) are being controlled by DP Codes while curtain control is the main device and being treated as curtain.

as mentioned in the comment, Meta.tuyaDatapoints will be handled by `tuya.fz.datapoints` instead of default functionality of tuya cover.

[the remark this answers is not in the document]
elad-bar: an additional reason, dp code 3 in tuya cover is positionArrived while in that device it represents functionality of calibration

devices/tuya.js
```diff
@@ -2274,6 +2377,97 @@ module.exports = [
...
     },
+    {
+        fingerprint: [
+            {modelID: 'TS0601', manufacturerName: '_TZE200_p6vz3wzt'},
+        ],
+        model: 'TS0601_cover_4',
+        vendor: 'TuYa',
+        description: 'Curtain / Blind switch',
+        fromZigbee: [fz.tuya_cover, fz.ignore_basic_report, fz.cover_position_tilt, tuya.fz.datapoints],
+        toZigbee: [tz.tuya_cover_control, tz.cover_position_tilt, tuya.tz.datapoints],
```
Koenkk, line 2388: Do not mix up the new (`tuya.tz.datapoints`) and old (`tz.tuya_cover_control`) api. So please map all the datapoints in `tuyaDatapoints` and remove `tz.tuya_cover_control` from this list. Same goes for `fz.tuya_cover`

elad-bar: thanks, much more convenient :)
will remove this PR and post a new clean

Koenkk: Thanks, let me know if I can help.

elad-bar: Opened new one, and actually there is something that you might know how to handle,
Now that I'm using datapoints for covers, I don't have the tilt control as I used to have, 
Do you know how should it be done to support it?
Thanks

Koenkk: Does the tilt setting go via the normal Zigbee commands and not the TuYa datapoint?

elad-bar: Yes, the problem is that the device holds both cover and 2 switches, switches are accessible only from datapoint (as far as I understand)

Koenkk: Then using the cover control converters is fine (`tz.cover_position_tilt`, not `tz.tuya_cover_control`!) since `tz.tuya_cover_control` can be mapped through `tuyaDatapoints`.

PR #5653, opened 2023-04-11
Allow OTA support for IKEA VINDSTYRKA

leander091:
There has recently been an OTA update from IKEA for VINDSTYRKA air quality sensor. However, Zigbee2mqtt does not support it right now.

This pull request will fix that.

### DIFF
--- a/devices/ikea.js
+++ b/devices/ikea.js
@@ -1115,6 +1115,7 @@ module.exports = [
         description: 'Vindstyrka air quality and humidity sensor',
         fromZigbee: [fz.temperature, fz.humidity, fz.pm25, ikea.fz.ikea_voc_index],
         toZigbee: [],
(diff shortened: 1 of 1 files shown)


MattWestb:
I have sniffing and building one OTA file from Dirigera that was updating my device and it was working OK but Tråfdfri GW is not having it in its feed so need downloading it from the ZHA discussion and doing it with local OTA files (its safe then the firmware is signed and its not broken = then the device is checking it before flashing it = OK).

Koenkk:
I already applied this in https://github.com/Koenkk/zigbee-herdsman-converters/commit/58313ebfb3a4d0b31c64b2811fbda324a2e13118 before seeing this, sorry! (credits will go to you in the next z2m release changelog 😄 )

PR #5654, opened 2023-04-12
Updates to match 2.14 firmware (previous 2.08)

michaelkkehoe:
2.10
* Add parameters P54(Double down to value of P56)
* Add parameter option 3 to P22(Switch Type) - Single-Pole Full Sine Wave

2.12
* Add parameters P55(Double Up to level)
* Add parameters P56(Double Down to level)

2.14
* Add P25 (Increase non-neutral power output. This may cause flickering or aux switch problems with some bulbs. Adjust max level to fix this)
* Modify P56 range 0 to 254 (allow double tap to turn device off)
* Add P100(LED Scaling ) 0- VZM31 ; 1- LZW31
* Add P123(Use different scene numbers via the aux switch)
* Add P125(Send “move to level on/off” to bound devices that matches default level remote and local)


devices/inovelli.js
```diff
@@ -312,6 +321,34 @@ const ATTRIBUTES = {
...
       'For use with Smart Bulbs that need constant power and are controlled via commands rather than power.',
     },
+    doubleTapUptoParam55: {
+        ID: 53,
+        dataType: BOOLEAN,
+        displayType: 'enum',
+        values: {'Disabled': 0, 'Enabled': 1},
+        description: 'Enable or Disable setting brightness to parameter 55 on double-tap UP.',
+    },
+    doubleTapUptoParam56: {
```
edwardgarza, line 331:
```suggestion
    doubleTapDowntoParam56: {
```

devices/inovelli.js
```diff
@@ -312,6 +321,34 @@ const ATTRIBUTES = {
...
+        dataType: BOOLEAN,
+        displayType: 'enum',
+        values: {'Disabled': 0, 'Enabled': 1},
+        description: 'Enable or Disable setting brightness to parameter 56 on double-tap DOWN.',
+    },
+    brightnessLevelForDoubleTapUp: {
+        ID: 55,
+        dataType: UINT8,
+        min: 0,
+        max: 100,
```
edwardgarza, line 342: shouldn't this be 255/254?

epower53: @EricM_Inovelli says params 55 and 56 are both 0-100. See his post here:
https://community.inovelli.com/t/blue-series-2-1-firmware-changelog-vzm31-sn/12326/372

edwardgarza: ok - in that case at least align line 349 to match this one

epower53 (changes requested):
Apparently the 0-100 ranges @EricM_Inovelli provided for parameters 55 & 56 were a result of him lifting the parameter definitions directly from the Hubitat driver, which uses percentages rather than a 2-byte number converted to decimal. Since the z2m converter does not convert to percentages, the 0-100 range for parameters 55 and 56 is not correct. Please adjust the max value for parameters 55 and 56 to be 254. Forum user \@mamber tested all the way to 255, and found that the switch only responds if the values for these parameters is in the range 0-254. See this forum post: https://community.inovelli.com/t/blue-series-2-1-firmware-changelog-vzm31-sn/12326/425

edwardgarza:
Yup, I think 255 gets remapped to 254 in a few places?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/1abb64d9613a3314ae6d8256073eea5078125e4c/devices/inovelli.js#L935

https://github.com/Koenkk/zigbee-herdsman-converters/blob/1abb64d9613a3314ae6d8256073eea5078125e4c/devices/inovelli.js#L1061

https://github.com/Koenkk/zigbee-herdsman-converters/blob/1abb64d9613a3314ae6d8256073eea5078125e4c/devices/inovelli.js#L980

devices/inovelli.js
InovelliUSA: > Apparently the 0-100 ranges @EricM_Inovelli provided for parameters 55 & 56 were a result of him lifting the parameter definitions directly from the Hubitat driver, which uses percentages rather than a 2-byte number converted to decimal. Since the z2m converter does not convert to percentages, the 0-100 range for parameters 55 and 56 is not correct. Please adjust the max value for parameters 55 and 56 to be 254. Forum user mamber tested all the way to 255, and found that the switch only responds if the values for these parameters is in the range 0-254. See this forum post: https://community.inovelli.com/t/blue-series-2-1-firmware-changelog-vzm31-sn/12326/425

Yep, sorry, I got those parameter definitions from the Hubitat driver. The actual permitted values are:

image

devices/inovelli.js
michaelkkehoe: Fixed. Thanks

michaelkkehoe:
@epower53 @InovelliUSA @edwardgarza  Could you confirm if you actually see the new attributes in the 'Dev Console'? When I install the converter, it isn't able to pull these values.

epower53:
> @epower53 @InovelliUSA @edwardgarza Could you confirm if you actually see the new attributes in the 'Dev Console'? When I install the converter, it isn't able to pull these values.

I don't see these in the Dev Console, either. I think that the manuSpecificInovelliVZM31SN cluster definition must also need an update. I had assumed/hoped that adding new parameters to the converter would do it, but it seems there's another file that needs additions. @InovelliUSA can you confirm this is correct, and if so point us in the direction we need to go?

InovelliUSA:
> @epower53 Yeah I'll start on it. @InovelliUSA could you please confirm the data-types.

Datatypes look good.

PR #5659, opened 2023-04-13
Update schneider_electric.js

Eilivers:
Adding running state an keypad lockout to EKO07259


Koenkk:
Looks good, thanks!

Agurkmix:
Running state is giving an error as of now on this device, casuing  interview to fail and HA to not be able to read if the thermostat is active or not.

image

PR #5662, opened 2023-04-13
fix dp code comparison

elad-bar:
meta.tuyaDatapoints hold dpId as string while data holds it as numeric

should address issue reported at [zigbee2mqtt repo #16966](https://github.com/Koenkk/zigbee2mqtt/issues/16966)

### DIFF
--- a/lib/tuya.js
+++ b/lib/tuya.js
@@ -1829,7 +1829,7 @@ const tuyaFz = {
             const datapoints = model.meta.tuyaDatapoints;
             for (const dpValue of msg.data.dpValues) {
                 const dpId = dpValue.dp;
-                const dpEntry = datapoints.find((d) => d[0] === dpId);
+                const dpEntry = datapoints.find((d) => d[0].toString() === dpId.toString());
                 if (dpEntry) {
                     const value = getDataValue(dpValue);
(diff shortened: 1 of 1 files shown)


Koenkk:
> meta.tuyaDatapoints hold dpId as string

This is not true, it is a number, e.g.:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/c8d56ba5566ba97d4c3d5f544b980fad4d06f8f0/devices/tuya.js#L2792

elad-bar:
You are absolutely right, and that is the issue, when it's being mapped the number becomes string

Koenkk:
> when it's being mapped the number becomes string

I don't understand where and when this happens, do you have some more pointers? The root cause should be fixed in this case, not the symptoms.

elad-bar:
Shared logs that I wrote since it didn't work, but it was at the time I mixed both datapoints and cover control, might be resulted from it?

https://github.com/Koenkk/zigbee2mqtt/issues/16966#issuecomment-1507466392

Koenkk:
I saw that log but still cannot explain it, does it happen when not mixing the cover and datapoint converters?

PR #5663, opened 2023-04-14
Support Schneider WDE002960

energizer91:
Add support for Schneider Electric WDE002960 Push button dimmer.

It's literally WDE002386, but new generation and works exactly the same as WDE002961 (Rotary push button).

[Link to device](https://www.se.com/se/sv/product/WDE002960/tryckdimmer-med-m%C3%B6jlighet-att-ansluta-neutralledare-exxact-styrs-via-wiser-by-se-zigbee-3-0-rc-0130w-rl-080w-vit/)

### DIFF
--- a/devices/schneider_electric.js
+++ b/devices/schneider_electric.js
@@ -361,6 +361,27 @@ module.exports = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

PR #5664, opened 2023-04-14
Added 5123.1110 by cloning 5120.2110

saseba:
Tested the Configuration on Device at Hand Iluminize 5123.1110. No Error or Warning yet.  I tested:
-On/Off
-Set Brightness
-Dimm via Push Switch

Device is a combination of adjustable Constant Current power supply with added Dimmer logic similar to said "Mini Dimming actuators" (5120.2110, 5120.1110) so it makes sense that they have the same internal configuration as they share the same push switch logic and probably compose of pwm dimming internaly.

### DIFF
--- a/devices/iluminize.js
+++ b/devices/iluminize.js
@@ -75,6 +75,19 @@ module.exports = [
         },
     },
     {
+        zigbeeModel: ['5123.1110'],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #5666, opened 2023-04-15
Add support for Aqara WXKG04LM remote switch

box-archived:
Add device WXKG04LM - Aqara double rocker wireless remote switch T1

Tested on device with following functions:
- Battery remain (voltage)
- Multistate actions (single, double, triple, hold, release)
- Each buttons (left, right, both)


Koenkk:
Thanks!

PR #5673, opened 2023-04-16
Add basic support for BMCT-SLZ Light/Shutter Control II

slugzero:
I could briefly get my hands on a [Bosch BMCT-SLZ Light/Shutter Control II] (https://www.bosch-smarthome.com/uk/en/products/devices/light-shutter-control/).

This device is a little bit tricky, because it supports two different operation modes: Light Control and Roller Shutter. After the initial join, the device expects a configuration message (see comments in the code). When it receives that message, the device leaves, and rejoins either as a switch or roller shutter.
I did not find another example of a device with two different operation modes, so I only set it up as a 2-channel switch now. Switching and reporting the on/off state works, additional features such as power metering, child lock etc. are not supported yet.

### DIFF
--- a/devices/bosch.js
+++ b/devices/bosch.js
@@ -741,7 +741,7 @@ const definition = [
         zigbeeModel: ['RBSH-MMS-ZB-EU'],
         model: 'BMCT-SLZ',
         vendor: 'Bosch',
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks again1

PR #5679, opened 2023-04-17
Fix action/operation_mode for Niko 552-721X2

sjorge:
Fixes:
- koenkk/zigbee2mqtt#13737
- koenkk/zigbee2mqtt#17346

actionMap was extended and expose updates so everything ends up under `action` as expected, also fixed operation_mode as it could only be set on l1 and effected both relays, with multiEndpointEnforce meta we can now also fix this.


Koenkk:
Thanks again!

PR #5681, opened 2023-04-18
fix(namron-4512737/4512738): remove `occupancy` reporting attribute

ruant:
After the new firmware upgrade for this device, the occupancy/away_mode issue has been fixed and this attribute is there for removed from the device.
The device will not configure itself correctly without this code change after the new firmware.

Firmware PR: https://github.com/Koenkk/zigbee-OTA/pull/294


Koenkk:
Thanks!

PR #5684, opened 2023-04-20
Add Keypad Action Reporting and Keypad lock feature for Sinope Switch SW2500ZB

ericgu08:
The Sinope Switch SW2500ZB can report keypad actions (clicks) single, double, hold (or long) press, up and down.  This function can be used to trigger automation or to call scenes.

It is also possible to disable (lock) the keypad.  In the lock configuration, the switch no longer controle the load.  Keypad actions are still reported.

The modifications add these 2 features.

Note: this is my first pull request.  I appologize in advance if I'm not doing it the right way.

### DIFF
--- a/devices/sinope.js
+++ b/devices/sinope.js
@@ -181,11 +181,11 @@ const fzLocal = {
             }
             if (msg.data.hasOwnProperty('actionReport')) {
                 const lookup = {2: 'key_up_pressed', 3: 'key_up_held', 4: 'key_up_pressed2x',
-                18: 'key_dn_pressed', 19: 'key_dn_held', 20: 'key_dn_pressed2x'};
+                    18: 'key_dn_pressed', 19: 'key_dn_held', 20: 'key_dn_pressed2x'};
                 result.action_report = lookup[msg.data['actionReport']];
             }
             if (msg.data.hasOwnProperty('keypadLockout')) {
-                const lookup = {0: 'unlock',1: 'lock'};
+                const lookup = {0: 'unlock', 1: 'lock'};
                 result.keypad_lockout = lookup[msg.data['keypadLockout']];
             }
             return result;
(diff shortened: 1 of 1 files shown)


devices/sinope.js
```diff
@@ -972,8 +993,11 @@ module.exports = [
         description: 'Zigbee smart light switch',
         fromZigbee: [fz.on_off, fz.electrical_measurement, fzLocal.sinope],
         toZigbee: [tz.on_off, tzLocal.timer_seconds, tzLocal.led_intensity_on, tzLocal.led_intensity_off,
-            tzLocal.led_color_on, tzLocal.led_color_off],
+            tzLocal.led_color_on, tzLocal.led_color_off, tzLocal.keypad_lock],
         exposes: [e.switch(),
+            exposes.enum('action_report', ea.STATE, ['key_up_pressed', 'key_up_pressed2x', 'key_up_held',
+                'key_dn_pressed', 'key_dn_pressed2x', 'key_dn_held'])
```
Koenkk, line 999: Does `dn` mean `down`? If yes, then please change it. Also rename `held` to `hold` and `pressed2x` to `pressed_2_times`

ericgu08: I did a quick search in the device folder, and it appears the terms `single` `double` `hold` are more common to describe the clicks.  Are you ok with the following proposition instead:
```
            exposes.enum('action_report', ea.STATE, ['up_single', 'up_double', 'up_hold',
                'down_single', 'down_double', 'down_hold'])
```
defined in a similar way in inovelli.js and letv.js.

devices/sinope.js
```diff
@@ -972,8 +993,11 @@ module.exports = [
         description: 'Zigbee smart light switch',
         fromZigbee: [fz.on_off, fz.electrical_measurement, fzLocal.sinope],
         toZigbee: [tz.on_off, tzLocal.timer_seconds, tzLocal.led_intensity_on, tzLocal.led_intensity_off,
-            tzLocal.led_color_on, tzLocal.led_color_off],
+            tzLocal.led_color_on, tzLocal.led_color_off, tzLocal.keypad_lock],
         exposes: [e.switch(),
+            exposes.enum('action_report', ea.STATE, ['up_single', 'up_double', 'up_hold',
```
Koenkk, line 998: I missed this on my initial review but please rename this to `action`  and use `e.action()` (e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/2861a72762fab96952a9077bb1551770a53b3200/devices/xiaomi.js#L871)

ericgu08: Here you go.  In my first attempt, I used `e.action`, but then I hesitate since `action` is also used when the option  `state_action` is active.  I asked the question in the [discussions](https://github.com/Koenkk/zigbee2mqtt/discussions/17287), but no one replied.  By the way, thanks for your great dedication.

Koenkk:
Thanks!

PR #5686, opened 2023-04-20
Support for Philips Hue - Xamento Hue

k2iuh:
The device added is a single packaged device (Philips Hue - Xamento Hue recessed black) which is the same as the one from the package with three lights ('929003526202_01', '929003526202_02', '929003526202_03').

I own all four of the devices and the change proposed enables my last bulb :)


Koenkk:
Thanks!

PR #5689, opened 2023-04-22
Allow for different definition model, vendor and description based on whitelabel

Koenkk:
**Problem**

Currently there is a lot of definition (code) duplication between different TuYa devices.

e.g. the Lidl HG06492B 
https://github.com/Koenkk/zigbee-herdsman-converters/blob/1413d092d8babf7f1f6fa6f9fd13afd7977b08da/devices/lidl.js#L788

is functionally the same as TuYa TS0502A

https://github.com/Koenkk/zigbee-herdsman-converters/blob/1413d092d8babf7f1f6fa6f9fd13afd7977b08da/devices/tuya.js#L1689

but with a different `vendor`/`description`/`model` value. This is done in order to generate a separate supported device page (zigbee2mqtt.io) and to show the correct vendor/description/model and picture (since that is based on the model) in the logging/discovery and frontend.

**Solution**
Override the definition model/vendor/description with the one from the whitelabel. To determine wether a whitelabel matches the device, the `fingerprint` property of the whitelabel is checked against the device. The `fingerprint` property is optional. This is because there are a lot of whitelabels already and we don't have all the information to define the missing fingerprints right now.

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,6 +131,22 @@ function findByZigbeeModel(zigbeeModel) {
 }
 
 function findByDevice(device) {
+    let definition = findDefintion(device);
+    if (definition && definition.whiteLabel) {
+        const match = definition.whiteLabel.find((w) => w.fingerprint && w.fingerprint.find((f) => isFingerprintMatch(f, device)));
+        if (match) {
+            definition = {
+                ...definition,
+                model: match.model,
+                vendor: match.vendor,
+                description: match.description || definition.description,
+            };
+        }
+    }
+    return definition;
+}
+
+function findDefintion(device) {
     if (!device) {
         return null;
     }
--- a/devices/lidl.js
+++ b/devices/lidl.js
@@ -785,16 +785,6 @@ module.exports = [
         },
     },
     {
-        fingerprint: [{modelID: 'TS0502A', manufacturerName: '_TZ3000_oborybow'}],
-        model: 'HG06492B',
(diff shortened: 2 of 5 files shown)


Koenkk:
CC: @kirovilya @nurikk @sjorge

sjorge:
So if I understand correctly, we can now can add a `whiteLabel` property to OEM devices and have, then zhc will auto 'clone' the definition but swap out the vendor, model and description for the ones defined so they appear to the user as a 'unique' device?

SGTM

Koenkk:
@sjorge exactly, but of course only if the whitelabel has a `fingerprint`. This test case shows how it works https://github.com/Koenkk/zigbee-herdsman-converters/pull/5689/files#diff-af95cb6f5f65fe70216ddce70325e785438d5aaff956a0666ebf46ef6402f526R446 , note that `HG06492B` has no separate definition and the definition from `TS0502A` is re-used.

index.js
```diff
@@ -131,6 +131,22 @@ function findByZigbeeModel(zigbeeModel) {
...
+                model: match.model,
+                vendor: match.vendor,
+                description: match.description || definition.description,
+            };
+        }
+    }
+    return definition;
+}
+
+function findDefintion(device) {
```
hnykda, line 149: (nitpick: typo FIY)

Koenkk: Fixed, thanks!

PR #5693, opened 2023-04-26
Add diverging model name HK_DIM_A

t3ch-fr3ak:
A few models report the name HK_DIM_A instead of HK-DIM-A. Additionally, the power supply is not recognized.


Koenkk:
Please merge these changes to the existing definition, no need to create a new one.

PR #5697, opened 2023-04-29
Add energy report to SW2500ZB and other minor fixes

ericgu08:
Following the previous modification [Koenkk#5684)](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5684)] and after a firmware update of light switch SW2500ZB, the `actionReport` (clicks) are no longer been reported.  The Specific Sinope cluster was added to the bind and `actionReport` is now compatible with new and old firmware version.

Corrected the converter for `led_intensity_on`, `led_intensity_off` and `timer_seconds` which was causing duplication of states (`ledIntensityOn`, `ledIntensityOff`, `dimmerTimmer`)

Added `connected_load` parameter for `energy` calculation
Added `energy` report via the metering cluster.


Koenkk:
Thanks!

PR #5699, opened 2023-04-29
Lidl Lamp added which is as well a TS0505B

wollo:
Only copied over the configuration of the sister device, changed the key and it worked


Koenkk:
Thanks! I've merged both in `tuya.js` as `TS0505B_2`, for the user there is no difference, picture/vendor/description will be the Lidl one.

PR #5700, opened 2023-04-29
Adding AC frequency to Develco SPLZB-131

RamonSmit:
As a new contributor, I'm hoping that I'm in the right direction for creating a pull request. 

The SPLZB-131 does support Ac Frequency measurements. I've locally changed the device in the code. Hope that this will be sufficient to make the change persistent. 
```
{"ac_frequency":49960,"current":1.23,"energy":0.33,"linkquality":36,"power":207,"state":"ON","voltage":226.75}
```

If I'm not doing this correctly, please make me aware of that. 

- Ramon

### DIFF
--- a/devices/develco.js
+++ b/devices/develco.js
@@ -292,7 +292,7 @@ module.exports = [
         description: 'Power plug',
         fromZigbee: [fz.on_off, develco.fz.electrical_measurement, develco.fz.metering],
         toZigbee: [tz.on_off],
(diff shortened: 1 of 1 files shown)


Koenkk:
Congrats with your first contribution!

emandtf:
I have many 131 and only few of them reports a valid "50 Hz" freq. The others (which have firmware before year 2022) reports "50.000 Hz".

PR #5710, opened 2023-05-02
Support Lidl / Parkside watering timer

zinserjan:
This adds full support for the lidl watering timer as requested in Koenkk/zigbee2mqtt#7695. Tuya dp converter format is used when possible.

Tested locally with npm link against the latest zigbee2mqtt:
- [x] on/off
- [x] timer
- [x] reporting of time left (manual watering + scheduled watering)
- [x] battery
- [x] time sync
- [x] scheduling / time slot config (even partial configs are possible to simplify creation of custom mqtt entities in HA)
- [x] frost lock

Open tasks (won't fix):
- prevention of overlapping time slot configurations (e.g. due to iterations + pause) 
  - device skips/ignores overlapping time slots, only the first one wins
  - reporting will be correct for the active slot (on/off, time left) 

This it how it looks like:
image

Btw while testing I discovered that the initial reporting of the device was unreliable in my production setup, but worked pretty good in my new dev setup. In my previous converter I got around this with a delay of one second before sending the tuya magic, but this doesn't help always.
The only real difference in my setup between dev and production was the used firmware. Since I updated the firmware in my production environment as well, the initial reporting seems to be working pretty solid, too.
- bumped cc2652p coordinator from 20220103 to 20221226
- bumped cc2652p router from 20220125 to 20221102 

<details>
<summary>Info: data restrictions by the original gateway</summary>

**Note:** Device may accepts higher values or invalid for some of these, but the following are the officially used ones.

- Timer: 1 - 599 min
- Scheduler: weekday (monday to sunday), periodic (every 1 -7 days) or disabled
- Timeslot (1  - 6):
  - start hour: 0 - 23 (255 seems to be reported for never used slots, but it's technically invalid)
  - start minute: 0 - 59 (255 seems to be reported for never used slots, but it's technically invalid)
  - timer: 1 - 599 min (0 seems to be reported for never used slots, but it's technically invalid)
  - pause: 0 - 599
  - iterations: 1 - 9 (0 seems to be reported for never used slots, but it's technically invalid)

</details>

### DIFF
--- a/devices/lidl.js
+++ b/devices/lidl.js
@@ -578,8 +578,8 @@ const valueConverterLocal = {
     },
     schedulePeriodic: {
         to: (value) => {
-            const scheduleValue = utils.numberWithinRange(value, 0, 6);
-            // Note: value of 0 switches to disabled weekday scheduler
+            const scheduleValue = utils.numberWithinRange(value, 0, 7);
+            // Note: mode value of 0 switches to disabled weekday scheduler
             const scheduleMode = scheduleValue > 0 ? 1 : 0;
             return [scheduleMode, scheduleValue];
         },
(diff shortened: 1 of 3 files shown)


zinserjan:
Ah linting fails against the merged state. Seems that utils was removed in the mean time. I'll fix this.

devices/lidl.js
```diff
@@ -140,6 +141,96 @@ const fzLocal = {
             }
         },
     },
+    watering_time_left: {
```
Koenkk, line 144: Could you refactor this to a valueconverter? (such that we only use `tuya.tz.datapoints` for this device)

lib/tuya.js
```diff
@@ -1686,6 +1694,9 @@ const tuyaTz = {
                 } else {
                     throw new Error(`Don't know how to send type '${typeof convertedValue}'`);
                 }
+
+                if (dpEntry[3] && dpEntry[3].optimistic === false) continue;
```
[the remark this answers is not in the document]
Koenkk: Does this always needs to be ignored?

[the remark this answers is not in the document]
zinserjan: We definitely need a way to skip this automatic state change for some DPs.
1. When toggling the ON/OFF state via Tuya DP the device reports the state on the same DP when it was successful. But when the battery is low or when watering is disabled due to frost lock, the device never reports back and just ignores the command. Without skipping this new state change, the device is marked as ON while it's still OFF. 
2. The DP for resetting the frost lock doesn't have any state and is just for writing a single value.

devices/lidl.js
zinserjan: Technically this should be possible, because this relies only on the ON/OFF DP and the current/previous state.
But then I need to replace the continues reporting of the remaining time by a JS interval. Is it possible to report state changes programmatically?

The current approach uses the continuous reporting of the battery level (every minute) to update the remaining time. But if there is way to do it programmatically, I can refactor this.


What I mean is something like that

```js
var timer;
function  valudeContverter() {
	if (ON) {
	  timer = setInterval(() => {
		// report state programmatically --> how?
		device.updateState({time_left: xxx});
		}, 1000 * 60);
	} else {
		clearInterval(timer);
		timer = undefined;
	}
}
```

zinserjan: I found a `publish` method, which is used [in another converter](https://github.com/Koenkk/zigbee-herdsman-converters/blob/c45a74c630baa319c64b5574497db70eb35f6cd1/converters/fromZigbee.js#L502) to publish the state programmatically. This should work here too, if we pass it down to the `DP.from` converter.

@Koenkk does it makes sense this way?

Koenkk: @zinserjan yes, feel free to pass the `publish` method

zinserjan: Works now :blush:

devices/lidl.js
```diff
@@ -529,12 +711,152 @@ module.exports = [
...
+                            .withDescription('Number of watering iterations. Works only if there is a pause.'),
+                    ),
+            ),
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [1, null, valueConverterLocal.wateringState],
+                // disable optimistic state reporting (device may not turn on when battery is low)
+                [1, 'state', tuya.valueConverter.onOff, {optimistic: false}],
+                [5, 'timer', tuya.valueConverter.range(1, 599)],
```
Koenkk, line 843: The range is already specified in the exposes, therefore there is no need to do it again here. I suggest to use `tuya.valueConverter.raw` here and remove `tuya.valueConverter.range`

devices/lidl.js
```diff
@@ -529,12 +711,152 @@ module.exports = [
...
-        toZigbee: [tz.on_off, tz.lidl_watering_timer],
-        onEvent: tuya.onEventSetTime,
-        configure: async (device, coordinatorEndpoint, logger) => {},
-        exposes: [e.switch(), exposes.numeric('timer', ea.SET).withValueMin(1).withValueMax(10000)
-            .withUnit('min').withDescription('Auto off after specific time.')],
+        fromZigbee: [fz.ignore_basic_report, fz.ignore_tuya_set_time, fz.ignore_onoff_report, tuya.fz.datapoints],
+        toZigbee: [
+            {
+                ...tuya.tz.datapoints,
+                key: ['state', 'timer', 'reset_frost_lock', 'schedule_periodic', 'schedule_weekday',
```
Koenkk, line 718: Please add these keys directly to `tuya.tz.datapoints`, this line should just be `toZigbee: [tuya.tz.datapoints]`

devices/lidl.js
zinserjan: Yeah, I used this before but I noticed there isn't any validation. At least I was able to set unsupported values via frontend and MQTT messages. But that's a different story.

Koenkk:
Many thanks!

BDV77:
Wow this looks great, can't wait to get this in the next release and test it out. 

Many thanks!!!

PR #5716, opened 2023-05-04
Moes Star Ring Scene Switch 4 Gang

kirovilya:
https://moeshouse.com/products/eu-star-ring-smart-light-switch?variant=44637975740731


Koenkk:
thanks!

PR #5719, opened 2023-05-05
fix wrongly filed presence sensor Tze204 sxm7l9xa

richardwonka:


Koenkk:
Did you check if this works?

evangelo18:
refer pull request #5761

tothgergely92:
While I was reading the codebase to find out some workaround to reduce the spamming messages of Tze204_sxm7l9xa device, I found that its name is misspelled here: [https://github.com/Koenkk/zigbee-herdsman-converters/blob/07e41d1ef5233e1b61335220ba9ab2fe7647a8f3/src/devices/tuya.js#LL4410C60-L4410C68](L4410), and looks like this PR could solve this, but the CI test says duplicate fingerprint...maybe because of [https://github.com/Koenkk/zigbee-herdsman-converters/blob/07e41d1ef5233e1b61335220ba9ab2fe7647a8f3/src/devices/tuya.js#L4437](L4437). @Koenkk can you confirm that this fingerprint should be there only once? And if so, what was the intention to reference it these 2 different places?

evangelo18:
@tothgergely92 what was your work around for the spamming messages? I will have about 15 of these devices on one network shortly and am concerned it will cause issues.

Koenkk:
@tothgergely92 it should indeed be present only once

tothgergely92:
> @tothgergely92 what was your work around for the spamming messages? I will have about 15 of these devices on one network shortly and am concerned it will cause issues.

You can read it here: [https://github.com/Koenkk/zigbee2mqtt/issues/17483#issuecomment-1566165109](https://github.com/Koenkk/zigbee2mqtt/issues/17483#issuecomment-1566165109)

PR #5720, opened 2023-05-05
Develco SPLZB-131 ac_frequency: needs multiplier/divider usage

emandtf:
Without this multiplier/divisor the AC freq it not normalized to standard "Hz".


Koenkk:
Thanks!

PR #5723, opened 2023-05-05
Develco SPLZB-131: power_reactive (VAR) reading

emandtf:
Useful to distinguish between Capacitive and Inductive load (negative values are common for negative-energy).

Maybe other Develco SPLZB-13x could have same capability, but I don't have any of them.

### DIFF
--- a/converters/fromZigbee.js
+++ b/converters/fromZigbee.js
@@ -778,6 +778,7 @@ const converters = {
                 {key: 'activePowerPhB', name: 'power_phase_b', factor: 'acPower'},
                 {key: 'activePowerPhC', name: 'power_phase_c', factor: 'acPower'},
                 {key: 'apparentPower', name: 'power_apparent', factor: 'acPower'},
+                {key: 'reactivePower', name: 'power_reactive', factor: 'acPower'},
                 {key: 'rmsCurrent', name: 'current', factor: 'acCurrent'},
                 {key: 'rmsCurrentPhB', name: 'current_phase_b', factor: 'acCurrent'},
                 {key: 'rmsCurrentPhC', name: 'current_phase_c', factor: 'acCurrent'},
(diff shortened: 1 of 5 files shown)


Koenkk:
thanks!

sk-pub:
@emandtf, it seems that this change caused incorrect power measurements. All my 5 SPLZB-131 devices now report 100 times less power.

emandtf:
> @emandtf, @Koenkk, All my 5 SPLZB-131 devices now report 100 times less power.

There isn't any relation about previous or other types of measurements. This commit regards a NEW value and nothing is changed to normal ones of the way them are retrieved/read/parsed.
I'm using this code/commit for all my 11 SPLZB-131 devices from middle of April and all is working well for all kind of measurements.

sk-pub:
> > @emandtf, @Koenkk, All my 5 SPLZB-131 devices now report 100 times less power.
> 
> There isn't any relation about previous or other types of measurements. This commit regards a NEW value and nothing is changed to normal ones of the way them are retrieved/read/parsed. I'm using this code/commit for all my 11 SPLZB-131 devices from middle of April and all is working well for all kind of measurements.

I have no better guess – it's the only change for this device in 1.31.0. Downgrading to 1.30.4 solved this problem for me.

emandtf:
> @emandtf,  now report 100 times less power.
Please open your "z2m/data/configuration.db" using a normal Text editor anche check for multiplier/divider values for those 131 devices, and check if they are correct or fix it if needed, the restart Z2M.
I had a similar issue 1 year ago when something (I never discovered who and why) changed those values and power was 10 and 100 times less than the real ones.

sk-pub:
> > @emandtf,  now report 100 times less power.
> 
> Please open your "z2m/data/configuration.db" using a normal Text editor anche check for multiplier/divider values for those 131 devices, and check if they are correct or fix it if needed, then restart Z2M. I had a similar issue 1 year ago when something (I never discovered who and why) changed those values and power was 10 and 100 times less than the real ones.

I'd rather stick to ver. 1.30.4 :)

PR #5724, opened 2023-05-06
Add more controls to Sinopé thermostats 2nd display and minor fix to SW2500ZB energy report

ericgu08:
Thermostat: TH1123ZB, TH1124ZB, TH1123ZB-G2, TH1124ZB-G2, TH11300ZB, TH1400ZB, TH1500ZB. 
_(Tested on TH1124ZB with older (2019) and latest firwmare)_ 
1) Add property `second_display_mode` for the thermostat Second display control, available with the latest thermostat firmware. There are 3 availables modes:  

 a) `auto`: The outdoor temperature (`thermostat_outdoor_temperature`) is displayed until considered expired by the thermostat.  The display then revert to the set point value.  The outdoor temp is considered expired when it has not been refreshed within `outdoor_temperature_timeout`.
b) `setpoint`: Display the set point value at all time.
c) `outdoor temp`: Same as auto, except the display blank when the outdoor temperature is expired. 

2) `thermostat_outdoor_temperature` is now exposed, along with `second_display_mode` and `outdoor_temperature_timeout`, converters were also added.

3) Property `enable_outdoor_temperature` is deprecated. It was emulating the function by controlling the `outdoor_temperature_timout`. Newer firmware should use the `second_display_mode` and older firmware without Second Display mode support should control the timeout directly instead.

Light Switch: SW2500ZB
1) Extended the `timer_seconds` range to 65536, as allowed by the light switch.  Added a preset named `Disabled` with the value 0.
2) Added the value 0 to the `connected_load` range. Starting at 1 was a typo.
3) Added `reporting.readMeteringMultiplierDivisor(endpoint)` required for the energy summation (`currentSummDelivered()`) to report a value.

### DIFF
--- a/devices/sinope.js
+++ b/devices/sinope.js
@@ -249,7 +249,7 @@ const tzLocal = {
             await entity.read('hvacThermostat', ['SinopeAuxCycleOutput']);
         },
     },
-    enable_outdoor_temperature: { //DEPRECATED: Use Second Display Mode or control via the timeout
+    enable_outdoor_temperature: { // DEPRECATED: Use Second Display Mode or control via the timeout
         key: ['enable_outdoor_temperature'],
         convertSet: async (entity, key, value, meta) => {
             if (value.toLowerCase() == 'on') {
@@ -609,7 +609,7 @@ module.exports = [
         },
     },
     {
-        zigbeeModel: ['TH1124ZB'],
+        zigbeeModel: ['TH1124ZA'],
         model: 'TH1124ZB',
         vendor: 'Sinopé',
         description: 'Zigbee line volt thermostat',
@@ -1085,7 +1085,7 @@ module.exports = [
         },
     },
     {
-        zigbeeModel: ['SW2500ZB'],
+        zigbeeModel: ['SW2500ZA'],
         model: 'SW2500ZB',
         vendor: 'Sinopé',
         description: 'Zigbee smart light switch',
@@ -1095,7 +1095,7 @@ module.exports = [
         exposes: [e.switch(),
             e.action(['up_single', 'up_double', 'up_hold', 'down_single', 'down_double', 'down_hold']),
             exposes.numeric('timer_seconds', ea.ALL).withUnit('seconds').withValueMin(0).withValueMax(65535)
-                .withPreset('Disabled',0).withDescription('Automatically turn off load after x seconds'),
+                .withPreset('Disabled', 0).withDescription('Automatically turn off load after x seconds'),
             exposes.numeric('led_intensity_on', ea.ALL).withUnit('%').withValueMin(0).withValueMax(100)
                 .withDescription('Control status LED intensity when load ON'),
             exposes.numeric('led_intensity_off', ea.ALL).withUnit('%').withValueMin(0).withValueMax(100)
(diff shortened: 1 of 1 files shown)


devices/sinope.js
```diff
@@ -987,7 +1085,7 @@ module.exports = [
         },
     },
     {
-        zigbeeModel: ['SW2500ZB'],
+        zigbeeModel: ['SW2500ZA'],
```
Koenkk, line 1088: I don't understand this change, this will make the `SW2500ZB` unsupported.

devices/sinope.js
```diff
@@ -564,16 +609,17 @@ module.exports = [
         },
     },
     {
-        zigbeeModel: ['TH1124ZB'],
+        zigbeeModel: ['TH1124ZA'],
```
Koenkk, line 612: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/5724/files#r1186673551

devices/sinope.js
ericgu08: Koen, sorry and thanks for catching that.  I had no intention to rename the zigbeeModel.  I use those locally as this is the only way I found to test my modifications.  I have to use a different name in the external converter (and in database.db) than the one in the official load. Thanks!

Koenkk:
Thanks!

onixx:
@ericgu08 have you had luck getting energy reporting on a DM2500ZB ? All mine do not report energy.

ericgu08:
> @ericgu08 have you had luck getting energy reporting on a DM2500ZB ? All mine do not report energy.

Hello @onixx, I don't own a DM2500ZB.  I'm 90% sure the same modification could be applied to the DM2500ZB, but since I had not way to test, I preferred leave it unmodified, and be sure no issue was introduce.

In the file sinope.ts, in definitions for DM2500ZB, for the  `fromZigbee`, you have to replace `fz.electrical_measurement` with `fz.metering`.  

in `exposes:` add:
```
            exposes.numeric('connected_load', ea.ALL)
                .withUnit('W').withValueMin(0).withValueMax(1800)
                .withDescription('Load connected in watt'),
            e.energy(),
```

In `configure`, for `const binds`, add `seMetering` 

and add:
```
            try {
                await reporting.currentSummDelivered(endpoint, {min: 10, max: 300, change: [0, 10]});
            } catch (error) {/* Do nothing*/}

```

onixx:
Thx, for the info @ericgu08 there might be a firmware or revision limitation as I am not getting data on my SW2500ZB either here. I will need to try to hook them up to the original hub to see if any updates can be done. Have you noticed this feature appearing after an update ?
This is what I have on the SW2500ZB
Read result of 'genBasic': {"swBuildId":"539\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000"}

ericgu08:
I think we have the same Firmware version.  I do see 539 under MQTT device info or in Z2M frontend, which correspond to 1.3.2 in Neviweb.  Did you set a value other than 0  for the Connected load?  The switch doesn't automatically detect the load, like in some other devices (thermostat).  If I set it to the max (1800W) and turn on the switch, after few seconds, I can see the Energy value increasing.  I just installed the latest Z2M release version and it is still working for me.

vinnygambiny:
@ericgu08 how you get the connected load if it's not automatically detected ?

ericgu08:
@vinnygambiny : you have to determine the load yourself, for example if the switch drives 2 x 100 Watt light bulbs, you would enter 200 W.  If you have a more dynamic load, then it won't be precise.

I see the sinope.ts (was sinope.js) format changed quite a bit since last year.  But that is what I would try:

Under the section 
`        zigbeeModel: ["DM2500ZB"],`

Replace
`        extend: [m.light({configureReporting: true})],`
with
`        extend: [m.light({configureReporting: true}),m.electricityMeter({cluster: "metering"})],`


After
`            tzLocal.led_color_off,`
Add:
`            tzLocal.connected_load,`

After
`                .withDescription("Control status LED color when load OFF"),`
Add
`            e.numeric("connected_load", ea.ALL).withUnit("W").withValueMin(0).withValueMax(1800).withDescription("Load connected in watt"),`

In the frontend, you should see: A line item with "Connected load" and a slider to input the load you have calculated.
You should also see a line item Energy reporting in kWh.  Set the load to 1800 W, turn on the switch and you should see the Energy value changing after few seconds.

I hope this helps.

PR #5730, opened 2023-05-08
Add fingerprint to the Sonoff SNZB-02D

GimpArm:
Add fingerprint to the Sonoff SNZB-02D because without one the device isn't identified.

I recently bought 5 of these devices from AliExpress and they were not identified with the current Edge version. After adding the fingerprint they were detected and function without problem. Maybe this isn't correct and there is another underlying issue why a fingerprint is needed but this works well for me.

### DIFF
--- a/devices/sonoff.js
+++ b/devices/sonoff.js
@@ -164,6 +164,9 @@ module.exports = [
         },
     },
     {
+        fingerprint: [
+            {type: 'EndDevice', manufacturerName: 'SONOFF', modelID: 'SNZB-02D'}
+        ],
         zigbeeModel: ['SNZB-02D'],
         model: 'SNZB-02D',
         vendor: 'SONOFF',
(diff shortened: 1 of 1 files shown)


sjorge:
Weird as the zigbeeModel matches the modelID of the figureprint so it should match.

@Koenkk ☝️  ?

Koenkk:
@sjorge you are right, this fix (shouldn't) do anything as `modelID` is the same as `zigbeeModel`

GimpArm:
That's what I was thinking that the fingerprint shouldn't be what makes it work but I do not know much about how the devices are identified. Should I instead close this PR and open a ticket? Can you offer some advice on what logs might be helpful because the only log message I see is that the device was not identified and it contains the same manufacturer and model ID.

GimpArm:
Ok, I feel dumb and I'm sorry for wasting your time.
I doubled checked my Edge version and the auto updater I made hasn't been working for some time. Which was the reason it wasn't detected.

PR #5732, opened 2023-05-09
Support Philips Hue Phoenix LED Pendant light

kruton:
This light presents itself as two different lights. One is the upward-facing bulb and the other is the downward-facing bulb.


Koenkk:
Thanks!

PR #5746, opened 2023-05-13
Fix rtcgq13lm OTA update

z-x-f:
I have two rtcgq13lm devices that encountered the issue `Failed to find endpoint which support OTA cluster` during OTA  updates. I tried force remove, restarting z2m, and re-pairing, but the problem is still. I found that the method mentioned in #5316  worked for me. Thanks @protyposis method can solve my problem.


Koenkk:
Thanks!

PR #5749, opened 2023-05-14
Add converter for SISW01

sbeca:
Add converter for [Mercator Ikuü inline switch (model SISW01)](https://www.ikuu.com.au/product/inline-switch/)


Koenkk:
Thanks!

PR #5750, opened 2023-05-14
Fix unknown power source for WSP403

sbeca:
The [Oz Smart Things whitelabel of WSP403](https://www.ozsmartthings.com.au/collections/zigbee/products/oz-zigbee-smart-plug) reports an unknown power source. The WSP403 is a smart plug that is always connected to mains so this PR adds a check for that, as well as adding a `whiteLabel` entry for good measure.

### DIFF
--- a/devices/owon.js
+++ b/devices/owon.js
@@ -109,6 +109,7 @@ module.exports = [
         zigbeeModel: ['WSP403-E'],
         model: 'WSP403',
         vendor: 'OWON',
+        whiteLabel: [{vendor: 'Oz Smart Things', model: 'WSP403'}],
         description: 'Smart plug',
         fromZigbee: [fz.on_off, fz.metering],
         toZigbee: [tz.on_off],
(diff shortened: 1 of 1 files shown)


devices/owon.js
```diff
@@ -121,6 +122,11 @@ module.exports = [
             await reporting.instantaneousDemand(endpoint, {min: 5, max: constants.repInterval.MINUTES_5, change: 2}); // divider 1000: 2W
             await reporting.currentSummDelivered(endpoint, {min: 5, max: constants.repInterval.MINUTES_5,
                 change: [10, 10]}); // divider 1000: 0,01kWh
+
+            if (device.powerSource === 'Unknown') {
+                device.powerSource = 'Mains (single phase)';
```
Koenkk, line 127: No need for this if check, for a power plug we can assume it is always main powered.

sbeca: I have just pushed a new commit that removes the check

Koenkk:
Thanks!

PR #5751, opened 2023-05-14
enable indicator_mode on Lidl smart plug HG06337

jlayec:


Koenkk:
Thanks!

Xyaren:
Heelo, it seems that "on" does not work on this switch.
the other 3 states, off, on/off and off/on do work though.

PR #5753, opened 2023-05-15
New 8 gang switch/relay

kirovilya:
https://aliexpress.ru/item/1005005453758738.html

Fixed getting convertedKey when endpoint_name comes already in an attribute name.

image

For some reason, several switches of one device have combined into a single block, accumulate and send changes as a single object.
If you switch the first switch, it will send:
`{"topic":"0xa4c1386e1e3d81cc/set","payload":{"state_l1":"ON"}}`
If you switch the second without going to another page, then it will send:
`{"topic":"0xa4c1386e1e3d81cc/set","payload":{"state_l1":"ON","state_l2":"ON"}}`
And so on, until you go to another page:
`{"topic":"0xa4c1386e1e3d81cc/set","payload":{"state_l1":"OFF","state_l2":"ON","state_l3":"OFF"}}`
If you do switching through the dashboard, this does not happen.

When the converter receives the object {"state_l1":"ON","state_l2":"ON"}, then the tuya-converter behaves inappropriately - an error occurs:
`Error: No datapoint defined for 'state_l2' at Object.convertSet (C:\Projects\zigbee2mqtt\node_modules\zigbee-herdsman-converters\lib\tuya.js:1671:27)`
I fixed it.

But! Several commands will be sent to the device at the same time, in my case 8. And not all will reach.

Seems to be related to the zigbee2mqtt-frontend.

### DIFF
--- a/devices/tuya.js
+++ b/devices/tuya.js
@@ -5177,4 +5177,44 @@ module.exports = [
             ],
         },
     },
+    {
+        fingerprint: [
+            {modelID: 'TS0601', manufacturerName: '_TZE200_vmcgja59'},
+        ],
+        model: 'TS0601_switch_8',
+        vendor: 'TuYa',
+        description: '8 Gang switch',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #5757, opened 2023-05-16
Bosch ```BSP-FZ2```: Enable OTA updates

DerDreschner:
Bosch released a firmware update for the ```BSP-FZ2``` today. To allow the update over Z2M, I had to modify the converter. This is for IT.


Koenkk:
Thanks!

PR #5758, opened 2023-05-17
Improvements related to PTVO DIY devices

ptvoinfo:
- Added multi-endpoint version for humidity, pressure, illuminance converters.
- Improved a converter for "ptvo.switch" (dynamic configuration and exposes).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,13 +32,13 @@ jobs:
       if: startsWith(github.ref, 'refs/tags/') && github.event_name == 'push'
       run: |
         npm publish
-        PACKAGE=zigbee-herdsman-converters
+        PACKAGE=$(node -p "require('./package.json').name")
         VERSION=$(node -p "require('./package.json').version")
         until [ $(npm view $PACKAGE --json | jq --arg version "$VERSION" -r '.versions[] | select (. == $version)') ];
         do
           echo "Waiting for publish to complete"
           sleep 5s
         done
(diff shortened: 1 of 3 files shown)


converters/fromZigbee.js
```diff
@@ -3316,6 +3316,60 @@ const converters = {
             return payload;
         },
     },
+    humidity2: {
+        cluster: 'msRelativeHumidity',
+        type: ['attributeReport', 'readResponse'],
+        options: [exposes.options.precision('humidity'), exposes.options.calibration('humidity')],
+        convert: (model, msg, publish, options, meta) => {
```
Koenkk, line 3323: Please integrate these changes into the existing converters.

devices/custom_devices_diy.js
```diff
@@ -280,6 +280,75 @@ const fzLocal = {
     },
 };
 
+function ptvoGetMetaOption(device, key, defaultValue) {
```
Koenkk, line 283: Could you use `utils.getMetaValue` instead here?

converters/fromZigbee.js
ptvoinfo: @Koenkk, I am  agree that if a device supports multiEndpoint, it should always append a suffix. But this update may break integration with HA or other systems for many devices. Therefore, I have added a separate converter.

Koenkk: I see, please move it into fzLocal of `custom_devices_diy.js`

ptvoinfo: @Koenkk Maybe, add it to fromZigbee.js and start moving to the correct implementation for new devices?

Koenkk: I want to go for a diff solution in the future, so lets move it to `custom_devices_diy.js` for now.

PR #5760, opened 2023-05-17
add new device support to TS0601 sensor

evangelo18:
new device _TZE204_sxm7l9xa support added


evangelo18:
the conflicting instance doesn't appear to work correctly.

PR #5761, opened 2023-05-17
move support for device _TZE204_sxm7l9xa

evangelo18:
moved code as previous entry was not recognised


Koenkk:
What do you mean not recognised? I believe this change will stop the device from working (https://github.com/Koenkk/zigbee2mqtt/issues/17483)

evangelo18:
The current entry shows the device as unsupported when connected. Not sure
why there is a separate device entry for this device. It's the same as the
others where I added it

Koenkk:
try with the latest dev branch without any changes, it should be detected as supported. (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html). The version you are running probably does not have this change yet.

PR #5762, opened 2023-05-17
Update nodon.js

drount:


sjorge:
@Koenkk can we use the same whitelabel trick here? It looks like this one and the already present _PRO variant are all functionally the same as SIN-4-RS-20 just with a slight different  exterior?

drount:
@sjorge do you mean to add both the SIN-4-RS-20_LEX and SIN-4-RS-20_PRO to SIN-4-RS-20's config zigbeeModel?

sjorge:
> @sjorge do you mean to add both the SIN-4-RS-20_LEX and SIN-4-RS-20_PRO to SIN-4-RS-20's config zigbeeModel?

That could work too, as they appear to be the same, aside appearance/sub branding…

But recently this got added  https://github.com/Koenkk/zigbee-herdsman-converters/pull/5689 So I was asking @koenkk if it would be a good usecase for this too, as then we can keep the alternate description with Pro and Lex markings and even a picture, but have them share the rest if tue config.

drount:
Oh I see, thanks for educating me :)

Looking forward to the feedback to procede. 

I'm stoked to have found a certified shutter zigbee motor controller at reasonable price in EU and I think more people can benefit from it!

Koenkk:
@sjorge this is fine as the zigbeeModel is different, the whitelabel feature is mostly meant for the TuYa devices.

PR #5768, opened 2023-05-20
Add new Legrand dimmer 199182

julien-lebot:
This PR adds the successor to the 067771 Legrand dimmer, apparently called 199182.

I found this device after ordering the "new" version by accident and found that the 067771 code worked as-is for it.
I don't know if there's way to not duplicate the code and instead use the the ID 199182 as an alias for 067771, feel free to guide me in the right direction.

### DIFF
--- a/src/devices/legrand.js
+++ b/src/devices/legrand.js
@@ -258,6 +258,35 @@ module.exports = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks, could you rebase this pr? (`legrand.js` was moved)

PR #5773, opened 2023-05-22
Update Acova Taffetas2 features

molusk:
added some features to Taffetas 2 heaters :
 - occupancy reporting using embedded motion sensor (stays on for 2 hours when triggered and is triggered by any change on the heater configuration : mode, set-point...)
 - unoccupied set-point control
 - temperature calibration control


Koenkk:
Thanks!

PR #5800, opened 2023-05-25
fix trailing space in SNZB-02P description

sjorge:
Not sure what the pr subject should be with the changes... I don't think it's `feat: ...` ?

Anyway, the recent PR that added support for SNZB-02P had a training space in the description, this one removes it.

### DIFF
--- a/src/devices/sonoff.js
+++ b/src/devices/sonoff.js
@@ -278,7 +278,7 @@ module.exports = [
         zigbeeModel: ['SNZB-02P'],
         model: 'SNZB-02P',
         vendor: 'SONOFF',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

sjorge:
Ah so `fix(ignore):` because it's a fix and should not end up in the release notes, got it. Will try to remember that for next time.

Koenkk:
@sjorge I used `fix(ignore):` because the device will also be added in the next release, an add and fix in the same release doesn't make sense. If this device was already supported, I would have used `fix:`

PR #5808, opened 2023-05-28
fix(ignore): fix SNZB-02D not showing power source correctly

octa22:
This PR fixes incorrectly displayed power source for Sonoff SNZB-02D


Koenkk:
Thanks!

PR #5809, opened 2023-05-28
fix(ignore): fix minor comment typo in src/lib/light.js

sjorge:
Spotted this while looking something up, used `fix(ignore):` as it really nothing note worthy at all wrt something a user would run into and it's not a code change either.


Koenkk:
Thanks!

PR #5811, opened 2023-05-29
options.supportsHS should be aligned with meta.supportsHueAndSaturation

sjorge:
While looking into koenkk/zigbee2mqtt#17816 I spotted some confusing naming wrt options.supportsHS vs meta.supportsHueAndSaturation.

Both try to do very similar things:
- options: adjusts the exposes data to show support or lack of support for hs mode in addition to xy mode
- meta: disables reading of the HS color attributes (although we somehow still commands what doe HS operations)

This PR tries to adres this by:
- Switching to a more consistent naming
  - options.supportsHS to options.supportsHueAndSaturation (to bring them inline naming wise)
  - options.preferHS to options.preferHueAndSaturation (to be more consistent in the naming)
  - meta.enhancedHue to meta.supportsEnhancedHue (to be more consistent in the naming)
- Throw and error when meta.supportsHueAndSaturation == false in the toZigbee converter when using HS color (fixes bulbs like the trådfri cws ignoring commands triggered by a color.{hue,sat} payload
- passthru of options.supportsHueAndSaturation -> meta.supportsHueAndSaturation in lib/extend (to avoid forgetting to set it in both places or setting both to a conflicting value when using extends, exception being e.g. trådfri which does not use the extends)

Open Questions:
- default for meta.supportsHueAndSaturation is true default for options.supportHS (now options.supportsHueAndSaturation) was false, these should be aligned
> Currently having the `options` default to false while most lights seem to supports it makes no sense, if most lights do not support it, we should change the default of the `meta`. In my test array of Hue, Hue BT, Tådfri, Innr, and Osram bulbs only Trådfri does not seem to support it, seeming to indicate that the `meta` default is the correct one?


src/lib/extend.js
```diff
@@ -79,10 +79,10 @@ const extend = {
     },
     light_onoff_brightness_color: (options={}) => {
         options = {
-            disableEffect: false, supportsHS: false, preferHS: false, disablePowerOnBehavior: false,
+            disableEffect: false, supportsHueAndSaturation: false, preferHS: false, disablePowerOnBehavior: false,
```
sjorge, line 82: @Koenkk I think this made it so we didn't noticed the wrong default being used in `light.readColorAttributes()`

It's certainly confusing. Can you give this PR an extra careful look as it's rather complex/confusing at times with all the layers of abstraction.

sjorge:
Tested if a random Hue bulb I had in my test box and an Osram one both still worked (Kind of annoying that the frontend uses `{"color": "#hexhere"}` for both color_hs and color_xy expose :/)

So tested with:
```
mosquitto_pub -t zigbee2mqtt/0x0017880104259333/set -m '{"color": {"hue": 128, "saturation": 100}}'
mosquitto_pub -t zigbee2mqtt/0x0017880104259333/set -m '{"color": {"x": 0.700, "y": 0.500}}'
```

I also read colorMode afterwards to see if it changed between 0 and 1, it did.

Koenkk:
supportHS is not a `meta` option, it is used here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/4649dccd59a1319fd8c0c81420fcb0491a9c52a3/src/lib/extend.js#L85

sjorge:
> supportHS is not a `meta` option, it is used here:
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/4649dccd59a1319fd8c0c81420fcb0491a9c52a3/src/lib/extend.js#L85

I'm happy to close this PR and do a new one just also adding `supportsHS` to the README.md, but it looks like `supportsHS` and `supportsHueAndSaturation` are really the same thing.

sjorge:
OK going to run some errands now, so have some time to think about this some more:

Before whats in this PR now:
- options.supportHS is used to indicate how the exposes data looks, just xy or xy + hs
- meta.supportsHueAndSaturation indictes if the device supports hs or not on the hw level
- by default we assume that most devices support xy and hs

Rework plan for this PR:
- undo the `light.readColorAttributes()` change, but updated README.md to reflect that the default is `true`
- make options.supportsHueAndSaturation (keeping the more consistent naming) also set meta.supportsHueAndSaturation in the extend

Indicating via options that there is no Hue/Saturation support would then do the same expose changes as before and in addition tell the converters/light.readColorAttributes() that there is no Hue/Saturation support, fixing the inconsistent views in exposes vs converter logic.

sjorge:
Found another issue with some a tradfri bulb from my test set (this is on current master, not this PR)

```
Zigbee2MQTT:debug 2023-05-29 13:41:37: Received Zigbee message from '0x000d6ffffe197fe9', type 'readResponse', cluster 'lightingColorCtrl', data '{"colorMode":1,"currentX":30015,"currentY":26870}' from endpoint 1 with groupID 0
Zigbee2MQTT:info  2023-05-29 13:41:37: MQTT publish: topic 'zigbee2mqtt/0x000d6ffffe197fe9', payload '{"color":{"hue":32,"saturation":81,"x":0.458,"y":0.41},"color_mode":"xy","linkquality":225,"update":{"installed_version":318846322,"latest_version":587753009,"state":"available"},"update_available":true}'
Zigbee2MQTT:info  2023-05-29 13:41:37: Read result of 'lightingColorCtrl': {"colorMode":1,"currentX":30015,"currentY":26870}
Zigbee2MQTT:debug 2023-05-29 13:41:56: Received MQTT message on 'zigbee2mqtt/0x000d6ffffe197fe9/set' with data '{"color": {"hue": 128, "saturation": 100}}'
Zigbee2MQTT:debug 2023-05-29 13:41:56: Publishing 'set' 'color' to '0x000d6ffffe197fe9'
Zigbee2MQTT:info  2023-05-29 13:41:56: MQTT publish: topic 'zigbee2mqtt/0x000d6ffffe197fe9', payload '{"color":{"hue":128,"saturation":100,"x":0.1668,"y":0.6399},"color_mode":"hs","linkquality":229,"update":{"installed_version":318846322,"latest_version":587753009,"state":"available"},"update_available":true}'
```

The color does not change, as expected as this build has meta.supportsHueAdnSaturation set to false.

```
Zigbee2MQTT:debug 2023-05-29 13:42:59: Received MQTT message on 'zigbee2mqtt/0x000d6ffffe197fe9/get' with data '{"color":""}'
Zigbee2MQTT:debug 2023-05-29 13:42:59: Publishing get 'get' 'color' to '0x000d6ffffe197fe9'
Zigbee2MQTT:debug 2023-05-29 13:42:59: Received Zigbee message from '0x000d6ffffe197fe9', type 'readResponse', cluster 'lightingColorCtrl', data '{"colorMode":1,"currentX":30015,"currentY":26870}' from endpoint 1 with groupID 0
Zigbee2MQTT:info  2023-05-29 13:42:59: MQTT publish: topic 'zigbee2mqtt/0x000d6ffffe197fe9', payload '{"color":{"hue":32,"saturation":81,"x":0.458,"y":0.41},"color_mode":"xy","linkquality":225,"update":{"installed_version":318846322,"latest_version":587753009,"state":"available"},"update_available":true}'
```

We currently only handle these for the get converter as to not hit the UNSUPPORTED_ATTRIBUTE, we should also take care of this when calling set, we can fall back to converting hs to zy. I'll also take this with me into the next pass of this PR.

sjorge:
OK hit another small roadblock, aside for philips.extend.light_* none of the regular extends set meta at all, there are two options forward:
- align defaults of meta.supportsHueAndSaturation with options.supportsHueAndSaturation(supportsHS) and make sure the non default case is set correctly in both options and meta for all devices
- somehow start setting meta from the general extends

The first one is the easiest, although I think the 2nd one is the cleanest, I am not sure how the extend + regular meta will interact.

src/converters/toZigbee.js
```diff
@@ -1077,9 +1077,20 @@ const converters = {
...
+            const supportsHueAndSaturation = utils.getMetaValue(entity, meta.mapped, 'supportsHueAndSaturation', 'allEqual', true);
+            const supportsEnhancedHue = utils.getMetaValue(entity, meta.mapped, 'supportsEnhancedHue', 'allEqual', true);
+
+            if (newColor.isRGB() || newColor.isXY() || (!supportsHueAndSaturation && newColor.isHSV())) {
+                // Convert RGB/HSV to XY color mode because Zigbee doesn't support RGB and the light does not support hue/saturation
+                let xy;
+                if (newColor.isRGB()) {
+                    xy = newColor.rgb.gammaCorrected().toXY().rounded(4);
+                } else if (newColor.isHSV()) {
+                    meta.logger.warn(`This device does not support Hue/Saturation, falling back to X/Y!`);
```
Koenkk, line 1089: Logging a warning on every color set in this case will be annoying (maybe logs as debug). But I'm wondering if we should do any conversion at all and just throw an error instead. I have not seen users asking for this feature yet.

sjorge: I was also leaning on just erroring, I will change this after work.

sjorge: Done:
```
Zigbee2MQTT:debug 2023-05-30 20:14:02: Received MQTT message on 'zigbee2mqtt/0x000d6ffffe197fe9/set' with data '{"color": {"hue": 128, "saturation": 100}}'
Zigbee2MQTT:debug 2023-05-30 20:14:02: Publishing 'set' 'color' to '0x000d6ffffe197fe9'
Zigbee2MQTT:error 2023-05-30 20:14:02: Publish 'set' 'color' to '0x000d6ffffe197fe9' failed: 'Error: This light does not support Hue/Saturation, please use X/Y instead.'
Zigbee2MQTT:debug 2023-05-30 20:14:03: Error: This light does not support Hue/Saturation, please use X/Y instead.
    at Object.convertSet (/Users/sjorge/Documents/repos/zigbee-herdsman-converters/src/converters/toZigbee.js:1085:23)
    at Object.convertSet (/Users/sjorge/Documents/repos/zigbee-herdsman-converters/src/converters/toZigbee.js:787:61)
    at Publish.onMQTTMessage (/Users/sjorge/Documents/repos/zigbee2mqtt/lib/extension/publish.ts:247:52)
    at EventEmitter.emit (node:events:523:35)
    at EventBus.emitMQTTMessage (/Users/sjorge/Documents/repos/zigbee2mqtt/lib/eventBus.ts:109:22)
    at MQTT.onMessage (/Users/sjorge/Documents/repos/zigbee2mqtt/lib/mqtt.ts:140:27)
    at MqttClient.emit (node:events:511:28)
    at MqttClient._handlePublish (/Users/sjorge/Documents/repos/zigbee2mqtt/node_modules/mqtt/lib/client.js:1547:12)
    at MqttClient._handlePacket (/Users/sjorge/Documents/repos/zigbee2mqtt/node_modules/mqtt/lib/client.js:535:12)
    at work (/Users/sjorge/Documents/repos/zigbee2mqtt/node_modules/mqtt/lib/client.js:438:12)
Zigbee2MQTT:info  2023-05-30 20:14:03: MQTT publish: topic 'zigbee2mqtt/bridge/log', payload '{"message":"Publish 'set' 'color' to '0x000d6ffffe197fe9' failed: 'Error: This light does not support Hue/Saturation, please use X/Y instead.'","meta":{"friendly_name":"0x000d6ffffe197fe9"},"type":"zigbee_publish_error"}'
```

Koenkk:
All looks good! I will merge it after the next release (probably friday)

Koenkk:
Bedankt!

PR #5815, opened 2023-05-29
Readded strange trailing characters in Taffetas2 zigbeeModel

molusk:
Revert to the previous definition of zigbeeModel for Taffetas2. I thought the strange trailing characters were not mandatory anymore... they are.


Koenkk:
Thanks!

PR #5844, opened 2023-06-05
fix: IKEA SYMFONISK gen2 does not need to divide the batteryPercentageRemaining

sjorge:
We have a `ikea.fz.battery` that tries to be smart based on the fwVersion if we need to divide or not, IKEA fixed there implementation to be in spec a while ago.

The SYMFONISK gen2 remote never had the problem and our fwVersion check checkes for a version that is high enough, but this remote has an good firmware even at major 1.

We can simply switch to the regular `fz.battery` to get the correct result here.

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/17616

### DIFF
--- a/src/devices/ikea.js
+++ b/src/devices/ikea.js
@@ -1185,7 +1185,7 @@ module.exports = [
         model: 'E2123',
         vendor: 'IKEA',
         description: 'SYMFONISK sound remote gen2',
-        fromZigbee: [ikea.fz.battery, fz.legacy.E1744_play_pause, ikea.fz.ikea_track_click, ikea.fz.ikea_volume_click,
+        fromZigbee: [fz.battery, fz.legacy.E1744_play_pause, ikea.fz.ikea_track_click, ikea.fz.ikea_volume_click,
             ikea.fz.ikea_volume_hold, ikea.fz.ikea_dots_click_v1, ikea.fz.ikea_dots_click_v2],
         exposes: [e.battery().withAccess(ea.STATE_GET), e.action(['toggle', 'track_previous', 'track_next', 'volume_up',
             'volume_down', 'volume_up_hold', 'volume_down_hold', 'dots_1_initial_press', 'dots_2_initial_press',


Koenkk:
Thanks!

MattWestb:
@sjorge The first shipped version was needing the doubling but Dirigera was updating it one week after it was released and TF is do it also now. I think it good not supporting the first shipped firmware version and the user shall updating  the firmware then this is the first device with the new firmware structure and have one old Zigbee stack that is updated to next latest certificated one.

Rodret series is in the pipe and is the new version of OnOff / Dimmer, OpenClose and Shortcut remotes and they shall having the same firmware status then its using the same program structure as Symfonisk 2 (Thes is ZCL R9  = Matter compatible).

sjorge:
OK so for those we will also need to use the regular fz.battery, or can we somehow detect the newer stack based on zclVersion/appVersion? If so, that might be worth adding to the ikea.fz.battery as a different conditional branch.

MattWestb:
I have only seen the 1.0.012 as shipped firmware and was needing battery doubling in ZHA and Z2M all newer is fixed so only need filtering on that and all shall working OK if IKEA is not having more version shipped with there new devices.

All IKEA controllers can being very old its depends witch way there was shipped and stored and the lock downs in china was braking the supply change completely but it shall start working more normal now.

RODRET serie is looking like thits https://www.ifun.de/ikea-rodret-neuer-mehrfach-dimmer-im-anmarsch-209671/ and only different is data settings in user data and the laser graved buttons if reading the FCC papers https://device.report/ikea-of-sweden-ab/e2201.

sjorge:
ooh the single AAA battery is a pretty good improvement, I hate the silly coincells, not sure it's worth replacing my 8 remotes though.

MattWestb:
The problem is normally you cant baying the OpenClose remotes then its only shipped with the blinds but shall being possible ordering as spare parts.

Itss possible converting the old OnOff dimmer to one OpenClose remote by writing the user data from one OpenClose then they is using the same firmware in the main flash (i have one running for test OTA files) but the new devices is 100% debug locked so cant reading and writing the flash but we have hacking the Somfonisk 2 and cave cloning it with one old bootloader plus one new OTA file and its working great.

Group binding is 100% not implanted in new devices :-(((
If you like you can flashing your old OpenClose (or OnOff) with SWD.probe and the last working group binding file that many ZHA and Z2M user have doing and its working.

PR #5848, opened 2023-06-05
[LIXEE] Add MOTDETAT

SilentT-FR:
In V13 Firmware a new value was added for legacy Mode of the meter 

MOTDETAT

@vk496  its ok ?


Koenkk:
Thanks!

PR #5856, opened 2023-06-07
fix(ignore): Added descriptions and support 1-3 phase variants of ZJSBL7-100Z

octa22:
this PR adds descriptions for exposed values and exposes power consumption on the each phase besides the overall power consumption.
there are more variants of ZJSBL7-100Z (1,2,3 phase) and therefore it rather checks if there is data of the power consumption on the each phase.


src/lib/legacy.ts
```diff
@@ -4676,6 +4676,15 @@ const fromZigbee1 = {
             }
             if (dp === dataPoints.hochActivePower) {
                 result.power = (value[2] | value[1] << 8) / 10;
+                if (value.length > 3) {
+                    result.power_1 = (value[5] | value[4] << 8) / 10;
```
Koenkk, line 4680: can you rename these to e.g. `power_l1`? (add `l`)

octa22: sure. done

Koenkk:
Thanks!

PR #5870, opened 2023-06-10
fix: correctly identify Mercator Ikuü SMI7040 (`_TZ3000_zw7wr5uo`)

davidjb:
This adds support for identifying the SMI7040 light and also normalises the vendor naming for two other white label devices in the same file while I was at it.


Koenkk:
Thanks!

PR #5871, opened 2023-06-10
fix: normalise Mercator Ikuü vendor and device naming

davidjb:
This cleans up the double-up of `Mercator Ikuü` vs `Mercator` in the device listings.
 
As best I can ascertain, the vendor's name is Mercator Ikuü (as described at https://www.ikuu.com.au/about-us/ - the brand name is always referred to in this form, plus it has its own dedicated domain). 

Product and device names are sourced from https://www.ikuu.com.au/products. Several products are EOL and names were obtained from web caches (e.g. Wayback Machine).


Koenkk:
Thanks!

PR #5875, opened 2023-06-11
add support new device with model 3RSNL02043Z

llhappier:
The new device picture: https://github.com/Koenkk/zigbee2mqtt.io/blob/master/public/images/devices/3RSNL02043Z.jpg

### DIFF
--- a/src/devices/third_reality.ts
+++ b/src/devices/third_reality.ts
@@ -227,7 +227,7 @@ const definitions: Definition[] = [
         zigbeeModel: ['3RSNL02043Z'],
         model: '3RSNL02043Z',
         vendor: 'Third Reality',
-        description: 'Zigbee multi-Function night light',
+        description: 'Zigbee multi-function night light',
         ota: ota.zigbeeOTA,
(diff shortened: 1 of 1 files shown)


src/devices/third_reality.ts
```diff
@@ -223,6 +223,17 @@ const definitions: Definition[] = [
...
         },
     },
+    {
+        zigbeeModel: ['3RSNL02043Z'],
+        model: '3RSNL02043Z',
+        vendor: 'Third Reality',
+        description: 'Zigbee multi-function night light',
+        ota: ota.zigbeeOTA,
+        extend: extend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 555]}),
+        endpoint: (device) => {
```
Koenkk, line 233: Can you check if everything also works without:

```
        endpoint: (device) => {
            return {'default': 1};
        },
```

llhappier:
With issue https://github.com/Koenkk/zigbee2mqtt/issues/17974, if  without endpoint: (device) => , it will send on/off to endpoint 3 but light endpoint is 1.

Koenkk:
I see, thanks!

PR #5877, opened 2023-06-11
Add support for MULTI-ZIG-SW device

mortenmoulder:
My custom device based on E18-MS1-PCB chip from Texas Instruments.


Koenkk:
Could you point to an English guide on how to make this device?

mortenmoulder:
@Koenkk Is the PR on the other repository enough? https://github.com/Koenkk/zigbee2mqtt.io/pull/2083

This device is custom and is only being sold in Denmark. It's not even released yet. I wanted to get Z2M support before I started shipping devices.

Koenkk:
I see, please move to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/custom_devices_diy.ts

src/devices/multi-zig-sw.ts
```diff
@@ -0,0 +1,116 @@
...
+    [key: string]: T;
+};
+
+const getKey = (object: Dictionary<number>, value: number) => Object.keys(object).find((key) => object[key] === value);
+
+const getListValueByKey = (source: Dictionary<number>, value: string) => source.hasOwnProperty(value) ? source[value] : parseInt(value, 10);
+
+const switchTypeExposes = (epName: string) => [e.enum('switch_type', exposes.access.ALL, Object.keys(switchTypesList)).withEndpoint(epName)];
+
+const fzLocal: { [key: string]: Fz.Converter } = {
```
Koenkk, line 21: Remove `{ [key: string]: Fz.Converter }` here

src/devices/multi-zig-sw.ts
```diff
@@ -0,0 +1,116 @@
...
+        convert: (model, msg, publish, options, meta) => {
+            const voltage = msg.data['batteryVoltage'] * 100;
+            const battery = (voltage - 2200) / 8;
+
+            return {
+                battery: battery > 100 ? 100 : battery,
+                voltage: voltage,
+            };
+        },
+    },
```
Koenkk, line 34: Add `} as Fz.Converter,` here for every converter

mortenmoulder:
> I see, please move to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/custom_devices_diy.ts

Done with that. I allowed myself to move some of my stuff into utils.ts, then I created a new file called customTypes.ts (for the Dictionary type, that I didn't feel like belonged in the types.ts file), and finally moved my stuff into the custom_devices.ts file.

Is there a reason why I needed to move it into a file? Would it be better if I reverted my changes, then renamed the original file into "my brand" (Smarthjemmet.dk)? I plan on making more devices in the future under my brand name. The reason I ask is because my device, as you can see, had some custom stuff I wrote.

src/lib/customTypes.ts
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable no-unused-vars */
+/* eslint-disable @typescript-eslint/no-namespace */
+export type Dictionary<T> = {
```
Koenkk, line 3: This can be moved to `lib/types.ts`

mortenmoulder:
@Koenkk Done.

src/devices/custom_devices_diy.ts
```diff
@@ -148,6 +153,22 @@ const tzLocal = {
...
         },
     } as Tz.Converter,
+    multi_zig_sw_switch_type: {
+        key: ['switch_type'],
+        convertGet: async (entity, key, meta) => {
+            await entity.read('genOnOffSwitchCfg', ['switchType']);
+        },
+        convertSet: async (entity, key, value, meta) => {
+            if (key == 'switch_type') {
+                const data = getListValueByKey(switchTypesList, value as string);
```
Koenkk, line 163: Sorry for not noticing earlier, but why not use `utils.getFromLookup(value, switchTypesList);` here?

src/devices/custom_devices_diy.ts
```diff
@@ -2,13 +2,18 @@ import * as exposes from '../lib/exposes';
...
+import {Definition, Tz, Fz, KeyValueAny, KeyValue, Zh, Expose, Dictionary} from '../lib/types';
 import * as reporting from '../lib/reporting';
 import extend from '../lib/extend';
 import * as constants from '../lib/constants';
 const e = exposes.presets;
 const ea = exposes.access;
-import {calibrateAndPrecisionRoundOptions, postfixWithEndpointName} from '../lib/utils';
+import {calibrateAndPrecisionRoundOptions, getKey, getListValueByKey, postfixWithEndpointName} from '../lib/utils';
+
+const switchTypesList = {
```
Koenkk, line 13: This is not a list, please rename to `switchTypeLookup`

src/devices/custom_devices_diy.ts
```diff
@@ -148,6 +153,22 @@ const tzLocal = {
             };
         },
     } as Tz.Converter,
+    multi_zig_sw_switch_type: {
+        key: ['switch_type'],
+        convertGet: async (entity, key, meta) => {
+            await entity.read('genOnOffSwitchCfg', ['switchType']);
+        },
+        convertSet: async (entity, key, value, meta) => {
+            if (key == 'switch_type') {
```
Koenkk, line 162: No need for this if statement, the key will always be `switch_type` because of `key: ['switch_type'],`

mortenmoulder:
@Koenkk Done. I've removed the Dictionary type as well, because there was no need for it after removing the getListValueByKey method. Instead I just added the type of the object directly in that one line:

```
const actionLookup: { [key: number]: string } = {0: 'release', 1: 'single', 2: 'double', 3: 'triple', 4: 'hold'};
```

Koenkk:
thanks!

Koenkk:
Can you test if the old local code still works and if yes provide it here?

PR #5884, opened 2023-06-15
Add 929003116101 and 929003115701 (Philips Hue Perifo)

StevenReitsma:
Adds 929003116101 and 929003115701, which are both Philips Hue Perifo lights. The white versions were already present, but this adds the black versions as well. I've also fixed the description of 929003115801, which was wrong.


Koenkk:
Thanks!

PR #5888, opened 2023-06-17
add support for EGLO FUEVA-Z (ceiling light IP44)

pswid:
It seems to be working, however preset `warm` is missing for `color_temp` and `color_temp_startup` (where are only `coolest`, `cool`, `neutral` and `warmest` visible on GUI). Why? I also noticed that both `neutral` and `warmest` pressets are equal 370. It should be fixed, but how?

### DIFF
--- a/src/devices/eglo.ts
+++ b/src/devices/eglo.ts
@@ -23,6 +23,13 @@ const definitions: Definition[] = [
         description: 'SALITERAS-Z ceiling light',
         extend: extend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 370]}),
     },
+    {
+        zigbeeModel: ['EGLO_ZM_TW_CLP'],
+        model: '98847',
+        vendor: 'EGLO',
+        description: 'EGLO FUEVA-Z ceiling light IP44',
+        extend: extend.light_onoff_brightness_colortemp({colorTempRange: [153, 370]}),
+    },
 ];
 
 module.exports = definitions;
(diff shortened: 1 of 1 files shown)


pswid:
There are five presets in the AwoX android application for this luminaire:
**cozy**: color=370, brightness=229
**warm**: color=370, brightness=178
**neutral**: color=304, brightness=178
**daylight**: color=213, brightness=178
**cold**: color=153, brightness=229

It seems that warm and warmest/cozy differ only in brightness. So, in z2m you only need to correct the predefined values for:
`cool`/daylight = 178 (now it is 250)
`neutral `= 304 (now it's 370)
`coolest` and `warmest` (warm?) are correct.

Koenkk:
There is no need to support all the presets from the AwoX app since users can easily make these themselves (we don't do this for any bulb)

pswid:
Ok, there is no need to support presets identical to the AvoX app, but at least each preset should have different value, don't you think? Now both `neutral` and `warmest` have the same value (=370) so, whichever you choose at frontend, both are displayed as "selected".

image

Koenkk:
Neutral will always be 370, if bulb support more than 370 you also will have warm (454) but the bulb does not support this. Warmest will always be the max of the range. Since in this case the max is 370 it will be the same as warmest, so that is expected behaviour.

pswid:
I see. However, in a situation like this, preset `neutral` does not work as the user expects. Maybe it's worth rethinking the approach and improve it? Ideally, the lamp definition could also include (optional, embedded in the code) values for `cool`, `neutral` and `warm`? Or maybe, instead of hard-coded values, just do: `cool=0.25*max`, `neutral=0.5*max` and `warm=0.75*max`? What do you think?

Koenkk:
Yes but that would be a breaking change, I think the current behaviour is good enough.

PR #5889, opened 2023-06-17
Add support for Lonsonho X702A

dimich99:
Added support for [Lonsonho X702A](https://www.aliexpress.com/item/4001138076397.html) (2-gang with neutral).
Added LED backlight settings
Added power_on_behavior for each channel
image

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1936,7 +1936,7 @@ const definitions: Definition[] = [
         endpoint: (device) => {
             return {'l1': 1, 'l2': 2};
         },
-        meta: { multiEndpoint: true },
+        meta: {multiEndpoint: true},
         configure: async (device, coordinatorEndpoint, logger) => {
             await tuya.configureMagicPacket(device, coordinatorEndpoint, logger);
             await reporting.bind(device.getEndpoint(1), coordinatorEndpoint, ['genOnOff']);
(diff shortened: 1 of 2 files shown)


src/devices/tuya.ts
```diff
@@ -1927,12 +1927,34 @@ const definitions: Definition[] = [
...
     {
         zigbeeModel: ['TS0002'],
         model: 'TS0002',
         vendor: 'TuYa',
         description: '2 gang switch',
-        whiteLabel: [{vendor: 'Zemismart', model: 'ZM-CSW002-D_switch'}, {vendor: 'Lonsonho', model: 'X702'},
+        whiteLabel: [
+            {vendor: 'Zemismart', model: 'ZM-CSW002-D_switch'},
+            {vendor: 'Lonsonho', model: 'X702'},
+            {vendor: 'Lonsonho', model: 'X702A'},
```
Koenkk, line 1957: Why was this added here? X702A = TS0002_switch_module_3, not TS0002, right?

dimich99: Should I remove it from there and create separate zigbeeModel section or just remove?

Koenkk: Just remove (because it is already under TS0002_switch_module_3)

Koenkk:
Thanks!

PR #5891, opened 2023-06-18
renamed ZY-M100-S to ZY-M100-L and TS0601_smart_human_presence_sensor_2 to ZY-M100-S

pswid:
ZY-M100-LvsS
ZY-M100-L
ZY-M100-S


Koenkk:
Thanks!

PR #5892, opened 2023-06-18
Update hvacThermostat attributes + invalid value handling

tomasbedrich:
Please review per-commit.

- a5be5d64657c04fd5d1a9ed55d1b0a00e213663f likely won't raise any significant discussion.
- 32f3888fd22a7f71a35a2b6c5802ae2c658e84f3 can potentially cause some backwards compatibility issues, if someone has an automation based on "expected wrong" value reported by their thermostat. On the other hand – `localTemperature` is so common (ZCL mandatory) that hopefully no/very-little devices returns invalid values.
- ad5ceb7405fb27306bf596d8b9a664d63462fd78 should be fixed because `0` can easily be a valid value for these attributes and by converting invalid values to `0` we loose the information. *The subject of discussion is if `null` is the right value to return in this case.*

Your thoughts are welcome.

### DIFF
--- a/src/converters/fromZigbee.js
+++ b/src/converters/fromZigbee.js
@@ -43,7 +43,7 @@ const converters = {
             const dontMapPIHeatingDemand = model.meta && model.meta.thermostat && model.meta.thermostat.dontMapPIHeatingDemand;
             if (msg.data.hasOwnProperty('localTemp')) {
                 let value = precisionRound(msg.data['localTemp'], 2) / 100;
-                value = value < -250 ? 0 : value;
+                value = value < -273.15 ? null : value;
                 result[postfixWithEndpointName('local_temperature', msg, model, meta)] = value;
             }
             if (msg.data.hasOwnProperty('localTemperatureCalibration')) {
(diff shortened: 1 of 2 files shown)


src/converters/fromZigbee.js
```diff
@@ -42,22 +42,26 @@ const converters = {
             const result = {};
             const dontMapPIHeatingDemand = model.meta && model.meta.thermostat && model.meta.thermostat.dontMapPIHeatingDemand;
             if (msg.data.hasOwnProperty('localTemp')) {
-                result[postfixWithEndpointName('local_temperature', msg, model, meta)] = precisionRound(msg.data['localTemp'], 2) / 100;
+                let value = precisionRound(msg.data['localTemp'], 2) / 100;
+                value = value < -273.15 ? null : value;
```
Koenkk, line 46: In case value is null, we should not add it to the result. This (and also the `null`s below should change to)

```js
if (value >= -273.15) {
  result[postfixWithEndpointName('local_temperature', msg, model, meta)] = value;
}
``

tomasbedrich:
As part of some future refactor, I'd propose creating an abstraction layer on top of Zigbee type parsing – to handle the invalid cases on a single place.

PR #5896, opened 2023-06-19
Add HK_DIM_A as model name

t3ch-fr3ak:
A few models report the name HK_DIM_A instead of HK-DIM-A.


Koenkk:

PR #5897, opened 2023-06-19
Adjust water_consumed unit

kamaradclimber:
It seems more standard to expose liters with `L` rather than `l`. I only found a single other example (giex valve) in this codebase, however international standard is `L` (and I suspect home assistant requires it to really consider water sensor to be used in energy dashboard).


Koenkk:
Thanks!

PR #5902, opened 2023-06-20
add support for phoscon hive

Backup7750:
please review, this is my first addon for zigbee2mqtt and my first commit in github. (worked for me as external addon in zigbee2mqtt)

### DIFF
--- a/src/devices/dresden_elektronik.ts
+++ b/src/devices/dresden_elektronik.ts
@@ -37,7 +37,7 @@ const definitions: Definition[] = [
         model: 'Hive',
         vendor: 'Phoscon',
         description: 'Battery powered smart LED light',
-        extend: extend.light_onoff_brightness_colortemp_color(),
+        extend: extend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 370]}),
     },
     {
         zigbeeModel: ['FLS-A lp (1-10V)'],
--- a/src/devices/dresden_elektronik.ts
+++ b/src/devices/dresden_elektronik.ts
@@ -33,10 +33,10 @@ const definitions: Definition[] = [
         ota: ota.zigbeeOTA,
     },
     {
-        zigbeeModel: ['Hive'], // The model ID from: Device with modelID 'lumi.sens' is not supported.
-        model: 'Hive', // Vendor model number, look on the device for a model number
-        vendor: 'Phoscon', // Vendor of the device (only used for documentation and startup logging)
-        description: 'Phoscon Hive battery powered smart LED light', // Description of the device, copy from vendor site. (only used for documentation and startup logging)
+        zigbeeModel: ['Hive'],
+        model: 'Hive',
+        vendor: 'Phoscon',
+        description: 'Battery powered smart LED light',
         extend: extend.light_onoff_brightness_colortemp_color(),
     },
     {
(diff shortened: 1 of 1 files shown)


src/devices/dresden_elektronik.ts
```diff
@@ -32,6 +32,13 @@ const definitions: Definition[] = [
         extend: extend.light_onoff_brightness(),
         ota: ota.zigbeeOTA,
     },
+    {
+        zigbeeModel: ['Hive'], // The model ID from: Device with modelID 'lumi.sens' is not supported.
+        model: 'Hive', // Vendor model number, look on the device for a model number
```
Koenkk, line 37: Is there any model number written on the device or product site? If yes use that here.

Backup7750: no unluciily not. 
via Zigbee2MQTT i can read the following Data
'genBasic': {"modelId":"Hive","manufacturerName":"Phoscon"}
on the website is also no product code
https://phoscon.de/en/hive

src/devices/dresden_elektronik.ts
```diff
@@ -32,6 +32,13 @@ const definitions: Definition[] = [
         extend: extend.light_onoff_brightness(),
         ota: ota.zigbeeOTA,
     },
+    {
+        zigbeeModel: ['Hive'],
+        model: 'Hive',
+        vendor: 'Phoscon',
+        description: 'Battery powered smart LED light',
+        extend: extend.light_onoff_brightness_colortemp_color(),
```
Koenkk, line 40: Please add the color temperature range as described on https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_3-adding-converter-s-for-your-device

Backup7750:
How can I add also the Power to the Template with the extend method?

i can read out the following function via Zigbee2mqtt
Read result of 'genPowerCfg': {"batteryPercentageRemaining":80}

Koenkk:
Updated the code to add this, please check if it works (https://github.com/Koenkk/zigbee-herdsman-converters/pull/5902/files)

Backup7750:
got the same error like in ci. zigbee2mqtt will not start.

/app/node_modules/zigbee-herdsman-converters/index.js:58
            assert.fail(`'${definition.model}' has configure in extend and device, this is not allowed`);
                   ^
AssertionError [ERR_ASSERTION]: 'Hive' has configure in extend and device, this is not allowed
    at Object.addDefinition [as addDeviceDefinition] (/app/node_modules/zigbee-herdsman-converters/index.js:58:20)
    at new ExternalConverters (/app/lib/extension/externalConverters.ts:15:17)
    at new Controller (/app/lib/controller.ts:84:58)
    at start (/app/index.js:106:18)

but can check this only as external converter.

Koenkk:
Fixed that, updated the code

Backup7750:
battery and lamp works fine now. Thanks a lot!
but device does not appear in update list.

Koenkk:
What do you mean with "update list"?

Backup7750:
in the panel "OTA" All devices are shown up which can be updated then you can search for firmware there.
what can I do to check if an update for this device is available if not in that "list"?

Koenkk:
Are you sure you added the ` ota: ota.zigbeeOTA,` line? If this is present it will show up under OTA updates.

Backup7750:
okay this line i missed. now it works... my fault :D

Koenkk:
Thanks!

Backup7750:
welcome
and a big thank you to you and this great project and the great support for that!

PR #5905, opened 2023-06-23
Fix MULTI-ZIG-SW

mortenmoulder:
I am no longer using input 1, but rather 2, 3, 4, and 5. This is because the device resets itself if input 1 is being held down for 10+ seconds.

### DIFF
--- a/src/devices/custom_devices_diy.ts
+++ b/src/devices/custom_devices_diy.ts
@@ -154,7 +154,7 @@ const tzLocal = {
         },
     } as Tz.Converter,
     multi_zig_sw_switch_type: {
-        key: ['switch_type'],
+        key: ['switch_type_1', 'switch_type_2', 'switch_type_3', 'switch_type_4'],
         convertGet: async (entity, key, meta) => {
             await entity.read('genOnOffSwitchCfg', ['switchType']);
         },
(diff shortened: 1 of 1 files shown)


mortenmoulder:
`Duplicate zigbee model MULTI-ZIG-SW` fails because I am updating the definition. The product is not out yet, so nobody but me will experience any issues.

`Exposes access matches toZigbee` not sure what this means. I simply updated the name in Z2M so they look like this:
image

`Check if all exposes have a color temp range` doesn't look like this error is mine.

Koenkk:
The keys of `multi_zig_sw_switch_type` should be updated (`switch_type_1`, `switch_type_2`)

mortenmoulder:
@Koenkk Done. Thanks.

I'm having a hard time testing the stuff I modify in custom_devices_diy.ts. When I build the file and get the compiled .js files, Z2M cannot use that file as a custom converter. So I have to copy stuff bath and forth into my own .js file, and remove the TypeScript stuff from it.

Is there a guide to getting `npm run build` files to work with Z2M as custom converters?

Koenkk:
Great, currently there is no guide but maybe we can support `ts` external converters in the future.

PR #5909, opened 2023-06-24
Add znqbkg25lm support

Whipsmith:


Koenkk:
Thanks!

Its-Manzzi:
> 

Hey,

Did you manage to resolve this? About to buy some of these but need decoupled mode to work in my use case. So keen to see if there was a resolution found?

SouthButovo:
Same problem.
Aqara H1M - action sensor doesn't distinguish between pressing the button right or left.

Whipsmith:
@Koenkk I'm a bit out of my depth here, but could it be related to this line:
`toZigbee: [tz.on_off, tz.xiaomi_switch_operation_mode_opple, tz.xiaomi_flip_indicator_light],`?
Should it potentially be:
`toZigbee: [tz.on_off, tz.xiaomi_switch_operation_mode_basic, tz.xiaomi_flip_indicator_light],`?

I can't figure out how to test this locally, so I don't want to make the change and potentially break things.

SouthButovo:
@Whipsmith Can I help you with testing?
Or is it better to make a separate issue with this problem?

Whipsmith:
@SouthButovo I have the devices, so I can test locally, I'm just not sure on how to run a different branch of the config in Home assistant.

I have another question though. Have you updated the firmware on your switch? I had a single rocker that I updated and it no longer works in decoupled mode.

Its-Manzzi:
Have you looked here to see if this helps?
https://github.com/Koenkk/zigbee2mqtt/issues/18483

Koenkk:
@Whipsmith 

You can test with an ext converter: https://gist.github.com/Koenkk/e1c01870ffb0a938d57a72f5e881f296 

- save this as file next to `configuration.yaml` as `ext_converter.js`
- add it to `configuration.yaml`:
```yaml
external_converters:
  - ext_converter.js
```
- start z2m, check if issue is fixed

barkas18:
Hi there!
I have 4-gang Zemismart switches. When I add it to Home Assistant, all buttons are detected, but when one is turned on, all 4 are turned on. Same with turning it off.
With further transfer to Homekit, the problem remains. Is there a solution?

PR #5913, opened 2023-06-25
Add note to src/ota/ubisys.ts

sjorge:
Just for future reference, it's probably good to have a small note in the source.


Koenkk:
Thanks!

PR #5914, opened 2023-06-26
add motion sensor and light sensor in 'Zigbee multi-function night li…

llhappier:
add motion sensor and light sensor in 'Zigbee multi-function night light'，model id is '3RSNL02043Z', ThirdReality.


Koenkk:
Thanks!

PR #5925, opened 2023-06-28
set min rep interval to 1 for Schneider Electric 545D6102 brightness

monkeyclass:
Setting the min. rep internal for Schneider Electric LK FUGA wiser wireless dimmer (545D6102) reporting of brightness to 1. Having the min. set to 0 generates a lot of zigbee traffic that result in brightness values arriving at the endpoint out of order resulting in "jumpy" controls from the physical dimmer. Setting min. rep interval to 1 gives correct and smooth control.

Exactly the same issue as described for [Candeo Zigbee LED smart dimmer switch (C202)](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/candeo.ts#L36).

### DIFF
--- a/src/devices/schneider_electric.js
+++ b/src/devices/schneider_electric.js
@@ -611,7 +611,7 @@ module.exports = [
             const endpoint = device.getEndpoint(3);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl', 'lightingBallastCfg']);
             await reporting.onOff(endpoint);
-            await reporting.brightness(endpoint);
+            await reporting.brightness(endpoint, {min: 1});
             // Configure the four front switches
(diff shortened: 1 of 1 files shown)


Koenkk:
A default of `0` doesn't make sense, so can you update this instead? (https://github.com/monkeyclass/zigbee-herdsman-converters/blob/e29d6b621ed6712e1a9406092d798f45766a8a13/src/lib/reporting.ts#L85)

monkeyclass:
> A default of `0` doesn't make sense, so can you update this instead (use 1 as default)? (https://github.com/monkeyclass/zigbee-herdsman-converters/blob/e29d6b621ed6712e1a9406092d798f45766a8a13/src/lib/reporting.ts#L85)

That makes sense. Corrected it to 1.

PR #5931, opened 2023-06-30
Fix naming inconsistency for vendor LED-Trading

pomeloy:
Just stumbled upon two forms of capitalization on the zigbee2mqtt.io device search by vendor. As I understand it, the markdown files are auto-generated, so this should fix it on the next docgen run


Koenkk:
Thanks!

PR #5939, opened 2023-07-02
Added support for TuYa 1 gang, 2 gang and 3 gang switches

krishnachytanya:


krishnachytanya:
Links to devices 

https://zigbee.blakadder.com/MakeGood_MG-ZG01W.html

https://zigbee.blakadder.com/MakeGood_MG-ZG02W.html

https://zigbee.blakadder.com/MakeGood_MG-ZG03W.html

src/devices/tuya.ts
```diff
@@ -1497,6 +1517,30 @@ const definitions: Definition[] = [
...
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_nh9m9emk']),
+        model: 'TS0601-2-gang-switchboard',
+        vendor: 'TuYa',
+        description: '2 gang switch',
+        exposes: [
+            e.switch().withEndpoint('l1').setAccess('state', ea.STATE_SET),
+            e.switch().withEndpoint('l2').setAccess('state', ea.STATE_SET),
+        ],
+        fromZigbee: [fz.ignore_basic_report, legacy.fromZigbee.tuya_switch],
+        toZigbee: [legacy.toZigbee.tuya_switch_state],
```
Koenkk, line 1530: For all converters, don't use legacy anymore. Use only datapoints fromZigbee/toZigbee conveters, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/4a8ab6fb8c19ba30d3596321b8a3fa2e5661036c/src/devices/tuya.ts#L1434

src/devices/tuya.ts
```diff
@@ -1467,6 +1467,25 @@ const definitions: Definition[] = [
             await reporting.bind(device.getEndpoint(1), coordinatorEndpoint, ['genOnOff']);
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_gbagoilo']),
+        model: 'TS0601-1-gang-switchboard',
```
Koenkk, line 1472: Please update this to the actual models, e.g. `MG-ZG01W`

src/devices/tuya.ts
```diff
@@ -1467,6 +1467,25 @@ const definitions: Definition[] = [
...
+        description: '1 gang switch',
+        exposes: [e.switch().withEndpoint('l1').setAccess('state', ea.STATE_SET)],
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        endpoint: (device) => {
+            return {'l1': 1};
+        },
+        meta: {
+            multiEndpoint: true,
```
Koenkk, line 1483: Please remove multiEndpoint here, remove `endpoint: (device) => {` function` and change `state_l1` to `state`

src/lib/tuya.ts
```diff
@@ -369,6 +369,15 @@ export const valueConverterBasic = {
     trueFalse: (valueTrue: number | Enum) => {
         return {from: (v: number) => v === valueTrue};
     },
+    legacyOnOff: ( ) => {
```
Koenkk, line 372: Why not use `onOff` instead?

krishnachytanya: The device sends out this message - Received Zigbee message from 'light', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,0,0],"type":"Buffer"},"datatype":2,"dp":21}],"seq":51200}' from endpoint 1 with groupID null. Using the onOff is causing this error - Error 2023-07-05 08:26:46Exception while calling fromZigbee converter: Value '0' is not allowed, expected one of true,false}. The implementation for this is not complete yet. Hence pulled the pull request to draft

src/devices/tuya.ts
krishnachytanya: This device is a multi endpoint device. The other endpoints are not active

Koenkk: That doesn't matter, only 1 endpoint is used from z2m perspective so this can be removed.

src/lib/tuya.ts
krishnachytanya: Not needed. reverted that change

Koenkk:
Thanks!

PR #5942, opened 2023-07-03
Add polling to 'Din smart relay'

muusc:
The din smart relay _TZ3000_qeuvnohg needs polling to get metering data.


muusc:
wrong account

PR #5946, opened 2023-07-04
Update ecodim to support all white label versions of the dimmer switch

KDanielK:
Added support for all white label versions of the Dimmer-Switch-ZB3.0 since they all have manufacturerID 4714 For example MoooN AB that is missing now

NOTE: most likely the other versions such as Namron can be removed now since they all have 4714 as manufacturerID


Koenkk:
looks good, can you remove the others?

KDanielK:
> looks good, can you remove the others?

Done! :)

Koenkk:
Thanks!

zeminator:
Still a noobie.. All my ecodim devices are failing to fetching firmware.. Getting error "no image available for imagetype '60'". I have both 07 and 10 devices.

PR #5948, opened 2023-07-04
fix: weekdays ordering in convertRawToTimer and convertRawToCycleTimer

ScratMan:
When setting timers in RTX ZVG1 and Saswell SAS980SWT-7-Z01 irrigation valves, the readout from Zigbee invert the weekdays ordering, getting SaFrThWeTuMoSu for the full week. This commit restores the weekdays ordering to something more human readable string like SuMoTuWeThFrSa.


Koenkk:
Thanks!

PR #5950, opened 2023-07-05
Update ecodim.ts

markbajaj:
Added Smart Dim manufacturer to ecodim device. Fixes issues with another white label device.

Candeo Dimmer Plug C210 stopped working on 1.32.0 #18218


KDanielK:
If it's a white label of Candeo should it not be part of https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/candeo.ts instead of ecodim? 

Only asking since I have a conflicting pr for this file https://github.com/Koenkk/zigbee-herdsman-converters/pull/5946

markbajaj:
> 

It is a Candeo C210 plug - I am not though sure how I would add it to this file? It has always been recognised as EcoDim Eco-Dim.07/Eco-Dim.10. Would I set up a new device called 'C210'?

Obviously it is better to have the correct device etc, just not sure of the code :)

Thanks for looking at this so quickly.

Koenkk:
This should not be needed anymore after https://github.com/Koenkk/zigbee-herdsman-converters/pull/5946

Please check if the dev branch is fine. 

PR #5952, opened 2023-07-05
Update NodOn / ADEO description files.

AlexisPolegato:
Update NodOn description files with proper device description.
Add OTA options on devices. 

Move ADEO Roller Shutter Controller from NodOn to ADEO description file.


src/devices/nodon.ts
```diff
@@ -103,12 +91,13 @@ const definitions: Definition[] = [
...
             await reporting.onOff(ep2);
         },
+        ota: ota.zigbeeOTA,
     },
     {
         zigbeeModel: ['SIN-4-2-20_PRO'],
         model: 'SIN-4-2-20_PRO',
         vendor: 'NodOn',
-        description: 'Double LED relay',
+        description: 'Lighting Relay Switch',
```
Koenkk, line 100: Can you normalize the casing for all descriptions? (e.g. `Lighting relay switch`)

Koenkk:
Thanks!

Pocky0013:
Have NodON SIN-4-2-20, now see update OTA can be done. Try to update, but have the problem : Update of 'CMD Kitchen' failed (No image available for imageType '258'),
Actual firmware is : 516 and try to update for 541. 
I have 10 SIN-4-2-20 and the problem is the same at all. 

Do you have a idea about solve the problem ?

scalmaa:
Hello. Same problem for me. I have 3 roller shutter nodon installed. Z2M  warn for an update, tell that the installed version is 523 and suggest to update to 541. I tried to udpate and the message is the same : Update of 'VR Chambre Arthur' failed (No image available for imageType '265'). I tried to verify is there was an available update via the nodon application. When I succeeded to detect the module with the application, the application inform me that the module was up to date.

Thank's for investigation if possible

PR #5953, opened 2023-07-06
add support for Lonsonho X703A

dimich99:
Added support for [Lonsonho X703A](https://www.aliexpress.com/item/4001138076397.html) (3 gang with neutral and backlight)
Added indicator_mode
Added power_on_behavior
image


Koenkk:
Thanks!

PR #5958, opened 2023-07-10
Added support for Hue LWO005

basschipper:
Added support for Philips Hue white G93 E27 filament globe (EAN: 8718699688882)


Koenkk:
Thanks!

PR #5960, opened 2023-07-11
Adding iFEEL Plus Comfort IFS-STD001

Ptr314:
Link to the device: https://ozon.ru/t/WG5g49o

Have been using this code as an external converter for about a year, but decided to make this request after a crasing with one of last Z2M updates.


Koenkk:
Thanks!

PR #5964, opened 2023-07-13
Add new device: IKEA RODRET E2201

thatslolo:
Koenkk/zigbee2mqtt.io#2129

### DIFF
--- a/src/devices/ikea.js
+++ b/src/devices/ikea.js
@@ -1223,5 +1223,5 @@ module.exports = [
             await reporting.bind(endpoint, coordinatorEndpoint, binds);
             await reporting.batteryPercentageRemaining(endpoint);
         },
-    };
+    },
 ];
--- a/src/devices/ikea.js
+++ b/src/devices/ikea.js
@@ -1207,4 +1207,22 @@ module.exports = [
             await reporting.batteryVoltage(endpoint1);
         },
     },
+    {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

MattWestb:
The current firmware is is supporting group binding but i have not testing if device binding is working or not in ZHA.

thatslolo:
In my testing group binding worked as well as device binding.

MattWestb:
Sound great then they is using EZSP stack 7.1.X and it shall being possible doing both and also ZHA is still getting all events to the coordinator after paring and adding one group so its looks IKEA have doing somthing good this time in there "kitchen".

coolspot18:
I just tried the latest 1.32.2-1 but the E2201 is not responding to any button presses? The action field is not returning anything?

MattWestb:
@martin-vl Witch model is you having then the PR is only doe the dimmer switch and not foe short cut and OpenCloce versions ?

martin-vl:
@MattWestb I see that I have the E2213 type. The picture looked the same ;). Maybe I'm going to try to add this one. Fingers crossed.

MattWestb:
I dont knowing how you was getting it then its not Zigbee certified but its in the pipe like the shortcut bottom, RGB LED Stripes and new motion and windows/door sensors but all is expected after new year

PR #5968, opened 2023-07-15
Missing poweroutagememory for SPP02G

aidbish:
the following tuya.exposes.powerOutageMemory() was missing from the SPP02G reference which is the same socket as the SPP04G but with 2 sockets


Koenkk:
Thanks!

PR #5979, opened 2023-07-18
Update calex.ts

JoyceKimberly:
Fixed recent errors when using Hue/Saturations with the rgb bulb. It always worked in the past, but it seems supportsHueAndSaturation: true needs to be set now.


Koenkk:
Thanks!

PR #5985, opened 2023-07-19
New presense sensor. Model YXZBRB58

kirovilya:
https://aliexpress.ru/item/1005005744570857.html

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -5382,12 +5382,10 @@ const definitions: Definition[] = [
         },
     },
     {
-        fingerprint: [
-            {modelID: 'TS0601', manufacturerName: '_TZE204_sooucan5'},
-        ],
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE204_sooucan5']),
         model: 'YXZBRB58',
         vendor: 'TuYa',
-        description: 'Smart Human Presence Sensor',
+        description: 'Smart human presence sensor',
         fromZigbee: [tuya.fz.datapoints],
         toZigbee: [tuya.tz.datapoints],
         exposes: [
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -5381,6 +5381,51 @@ const definitions: Definition[] = [
...
+                .withDescription('Maximum range').withUnit('m'),
+            e.numeric('detection_delay', ea.STATE_SET).withValueMin(0).withValueMax(10).withValueStep(0.1)
+                .withDescription('Detection delay').withUnit('s'),
+            e.numeric('fading_time', ea.STATE_SET).withValueMin(0).withValueMax(1500).withValueStep(1)
+                .withDescription('Fading time').withUnit('s'),
+            e.enum('radar_scene', ea.STATE_SET, ['default', 'bathroom', 'bedroom', 'sleeping'])
+                .withDescription('Presets for sensitivity for presence and movement'),
+        ],
+        meta: {
+            tuyaSendCommand: 'dataRequest',
```
Koenkk, line 5408: The default is `dataRequest` so this shouldn't be needed (please test)

kirovilya: yes, working

blakadder:
Seems minimum and maximum range dp's are reversed
image

fliespl:
@kirovilya I have some strange issues with this converted.

Presets set values that are out of scope for specific items. I.e. bathroom sets 60s detection time, while detection time (based on your investigation) should be within 0-10. 

Something is really off here.

blakadder:
> @kirovilya I have some strange issues with this converted.
> 
> Presets set values that are out of scope for specific items. I.e. bathroom sets 60s detection time, while detection time (based on your investigation) should be within 0-10. 
> 
> Something is really off here.

I PRd a fix yesterday for swapped dp's

PR #5987, opened 2023-07-20
findByModel function. Useful in external converters 

kirovilya:
it is easier to add new or similar devices in external converters without completely copying the description.
for example:
```
const findByModel = require('zigbee-herdsman-converters').findByModel;

const oldDefinition = findByModel('TS0001_fingerbot_1');
const definition = {
    ...oldDefinition,
    model: 'new_model',
    fingerprint: tuya.fingerprint('TS0001', ['_TZ3210_434534534'])
};

module.exports = definition;
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -221,28 +221,14 @@ function isFingerprintMatch(fingerprint, device) {
 
 function findByModel(model){
     /*
-    Search device description by model name.
+    Search device description by definition model name.
     Useful when redefining, expanding device descriptions in external converters.
     */
-    const modelToFind = model.toLowerCase();
-    let candidate;
-    for (const definition of definitions) {
-        if (definition.model.toLowerCase() == modelToFind) {
-            candidate = definition;
-            break;
-        }
-        if (definition.whiteLabel) {
-            for (const whiteLabel of definition.whiteLabel) {
-                const {vendor, model, description} = whiteLabel;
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -219,13 +219,42 @@ function isFingerprintMatch(fingerprint, device) {
     return match;
 }
 
+function findByModel(model){
+    /*
+    Search device description by model name.
+    Useful when redefining, expanding device descriptions in external converters.
+    */
+    const modelToFind = model.toLowerCase();
+    let candidate = findByZigbeeModel(modelToFind);
```
Koenkk, line 228: Can't we just do `const definition = definitions.find((d) => d.model == model));` here?

kirovilya: what if model in whitelabel ?

kirovilya: ah, just in case, we are looking for zigbee model

Koenkk: I think definition `model` is enough, zigbeeModel will just complicate stuff.

kirovilya: ok. waiting

Koenkk:
I refactored a bit, can you check if it still works?

kirovilya:
> I refactored a bit, can you check if it still works?

ok now

Koenkk:
Thanks!

PR #5989, opened 2023-07-20
added 4033931P6 which is the white colored version of the 4033930P6

clamb:
The 4033931P6 seems to be the white colored version of the black one with model 4033930P6.


Koenkk:
Thanks!

PR #5992, opened 2023-07-21
add new tuya device SPM02 by [BITUO TECHNIK]

script0803:


Koenkk:
Thanks, could you also provide a picture of this device?

PR #5998, opened 2023-07-23
Added support for Visonic GB-540 Acoustic Glass Break Sensor

gregwjacobs:
Added support for https://www.visonic.com/glass-break-detector-gb-540

* Tamper Detection
* Low Battery Alerts
* Glass break sensing / alerting (used devices test mode to confirm operation / correct signaling)


Koenkk:
Thanks!

PR #6003, opened 2023-07-24
Adds support for ZB-SmartPlugIR-1.0.0

dgomes:
Just the same switch but with a IR port (can't figure out how it works...) all other aspects work (switch, energy and power)


Koenkk:
Thanks!

PR #6006, opened 2023-07-24
Added WB-MSW-ZIGBEE v.4 device

aadegtyarev:
We have made a new version which is slightly different from version 3.


Koenkk:
Thanks!

PR #6009, opened 2023-07-25
enhance reporting of nimly pro door lock

slackspace-io:
Enhancing support for Nimly Pro smart lock. The original version does not report values for numerous fields - see https://github.com/Koenkk/zigbee2mqtt/issues/17205

Parsing messages 256 and 257 to return the following:
last_unlock_source: Whether self, keypad, fingerprint, rfid which unlocked the door
last_unlock_user: numerical value of which user unlocked the door
last_lock_source: Whether self, keypad, fingerprint, rfid which locked the door
last_lock_user: numerical value of which user locked the door

As well the last reported pin entry -- This is whatever numbers were entered into the keypad. 

This works for me as an external converter, this is my first time trying to merge in to your repo. Hopefully i've followed things correct :)

### DIFF
--- a/src/devices/onesti.ts
+++ b/src/devices/onesti.ts
@@ -5,7 +5,57 @@ import tz from '../converters/toZigbee';
 import * as reporting from '../lib/reporting';
 const e = exposes.presets;
 const ea = exposes.access;
-import * as constants from '../lib/constants';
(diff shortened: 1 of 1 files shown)


slackspace-io:
Sorry.. forgot the steps after test :X 

Will fix and try again

PR #6010, opened 2023-07-25
Enhance reporting of nimly pro door lock

slackspace-io:
Enhancing support for Nimly Pro smart lock. The original version does not report values for numerous fields - see https://github.com/Koenkk/zigbee2mqtt/issues/17205

Parsing messages 256 and 257 to return the following:
last_unlock_source: Whether self, keypad, fingerprint, rfid which unlocked the door
last_unlock_user: numerical value of which user unlocked the door
last_lock_source: Whether self, keypad, fingerprint, rfid which locked the door
last_lock_user: numerical value of which user locked the door

As well the last reported pin entry -- This is whatever numbers were entered into the keypad.

This works for me as an external converter, this is my first time trying to merge in to your repo. Hopefully i've followed things correct :)


src/devices/onesti.ts
```diff
@@ -1,19 +1,69 @@
...
+                }
+                attributes.last_used_pincode = pincode;
+            }
+
+            // Handle attribute 256
+            if (msg.data['256'] !== undefined) {
+                const hex = msg.data['256'].toString(16).padStart(8, '0');
+                const firstOctet = String(hex.substring(0, 2));
+                const lookup: { [key: string]: string } = {
+                    '00': 'MQTT',
```
Koenkk, line 31: - I proprose to change this to `zigbee`, this library is used by more projects than just Zigbee2MQTT.
- Please make everything lower case here

src/devices/onesti.ts
```diff
@@ -26,8 +76,11 @@ const definitions: Definition[] = [
             device.save();
         },
         exposes: [e.lock(), e.battery(), e.sound_volume(),
-            e.lock_action_source_name(), e.lock_action_user(),
-            e.action(Array.from(Object.values(constants.easyCodeTouchActions))),
+            e.text('last_unlock_source', ea.STATE).withDescription('Last unlock source'),
```
Koenkk, line 79: I missed this in my initial review, but this should be an `e.enum` with all the values from the lookup. (same for `last_lock_source`)

Koenkk:
Thanks!

stefancn:
Thanks for your superb works guys,  but our locks doesn't show zeros '0' in the 'last_used_pin_code'  eg. '440605' shows '4465' can   you fix this ....

Regards ... 

/Stefan

PR #6011, opened 2023-07-25
Correct errors on Schneider NHPB Switches

slackspace-io:
Fixing the issue reported here; https://github.com/Koenkk/zigbee2mqtt/issues/18429 

Disable powerOnBehaviour as it is unsupported on this device.
Change endpoint to '1' which is the correct endpoint for OnOff reading.


Koenkk:
Thanks!

PR #6013, opened 2023-07-26
Added support for Niceboy sensors

vsigler:
Added support for a few Niceboy sensors:
https://niceboy.eu/en/product/orbis-motion-sensor
https://niceboy.eu/en/product/orbis-vibration-sensor
https://niceboy.eu/en/product/orbis-water-sensor
https://niceboy.eu/en/product/orbis-wd-sensor

Images added in PR here: https://github.com/Koenkk/zigbee2mqtt.io/pull/2147

All of those are TuYa-based, just needed fingerprints and whitelabel markers added.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -540,6 +540,7 @@ const definitions: Definition[] = [
             {vendor: 'TuYa', model: 'iH-F001'},
             {vendor: 'Tesla Smart', model: 'TSL-SEN-DOOR'},
             {vendor: 'Cleverio', model: 'SS100'},
+            tuya.whitelabel('Niceboy', 'ORBIS Windows & Door Sensor', 'Door sensor', ['_TZ3000_qrldbmfn']),
         ],
         configure: async (device, coordinatorEndpoint, logger) => {
             try {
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -540,6 +540,7 @@ const definitions: Definition[] = [
             {vendor: 'TuYa', model: 'iH-F001'},
             {vendor: 'Tesla Smart', model: 'TSL-SEN-DOOR'},
             {vendor: 'Cleverio', model: 'SS100'},
+            tuya.whitelabel('Niceboy', 'ORBIS Windows & Door Sensor', 'Door sensor', ['_TZ3000_qrldbmfn']),
```
Koenkk, line 543: Instead of `ORBIS Windows & Door Senso` use the model number which is written on the device.

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/2147/files

src/devices/tuya.ts
vsigler: It is actually what is listed in both the manual and on the device itself as "model". I did not find any other number that I could use. It is the same for all of the sensors.

Koenkk:
I see, thanks!

PR #6023, opened 2023-07-27
feat: initial support for Ubisys H10

sjorge:
https://github.com/Koenkk/zigbee2mqtt/issues/18121

This adds the basic support on the 'switch' endpoints, will try to add the thermostat endpoints later when I get more feedback.


sjorge:
I'm confused on the lint error:
```
Error:   957:17  error  Inconsistently quoted property 'default' found  quote-props
```

All other instances I found at a quick glance also have the default ep unquoted?

PR #6024, opened 2023-07-27
Nimly Smart Lock - Two bugs during verification

slackspace-io:
1.) When changing to enums I did not add descriptions this results in exposes page of zigbee2mqtt to display incorrectly.
2.) misuse of 'pincode' vs 'pin_code' for a variable set. 

Unfortunately how i'm developing requires me to manually move changes from a js compliant external converter file to the ts file which I push to git.  Apologies for the bugs. I'll try to find a more robust way of local testing.

Referencing the ongoing improvement request for these devices; https://github.com/Koenkk/zigbee2mqtt/issues/17205


Koenkk:
Thanks!

PR #6027, opened 2023-07-30
Adding new manufacturerName for TuYa smoke sensor

wapolinar:
Just a very small change to support the new production line of Tuya smoke sensors.

image

Old ones (that I own) have TZE200, new ones (just bought) have TZE204.


Koenkk:
Thanks!

PR #6032, opened 2023-08-01
Fixed Legrand/Bticino LedInDark feature

FabianMangold:
This PR:
- Adds a missing FZ converter to handle Legrand / Bticino's "LedInDark" feature
- Adds the relevant configuration to the affected devices
- Removes Legrand's 067776_inverted switch


src/converters/fromZigbee.js
```diff
@@ -3930,6 +3930,13 @@ const converters = {
             }
         },
     },
+    legrand_ledInDark: {
```
Koenkk, line 3933: Please rename to `legrand_led_in_dark`

src/devices/legrand.js
```diff
@@ -153,31 +153,6 @@ module.exports = [
             exposes.binary('led_in_dark', ea.ALL, 'ON', 'OFF').withDescription(`Enables the LED when the power socket is turned off,
             allowing to see it in the dark`)],
     },
-    {
-        // Some require coverInverted:
-        // - https://github.com/Koenkk/zigbee2mqtt/issues/15101#issuecomment-1356787490
-        // - https://github.com/Koenkk/zigbee2mqtt/issues/16090
-        fingerprint: [
```
Koenkk, line 160: Why was this removed?

FabianMangold: That's correct, and it was not my intention to break anything.

Until shortly, this was required as there was no other alternative. Now that OTA is available, those devices work perfectly with the classic (non inverted) 067776.
I am preparing another PR with more extensive changes to the 067776, and wanted to avoid duplication.
The change has been reverted, hope that is OK.

Koenkk:
thanks!

PR #6042, opened 2023-08-04
remove colortemp, fix specific cluster and change multiendpoint to one endpoint

llhappier:
remove color temp, fix specific cluster and change multiendpoint to one endpont.


Koenkk:
Thanks!

PR #6044, opened 2023-08-04
Update lidl.js to add extra manufacturer for TY0203

chazsconi:
I have a TY0203 Silvercrest smart window and door sensor, but the manufacturer name is different to the existing one provided.


Koenkk:
Thanks!

PR #6048, opened 2023-08-04
Support Sercomm SZ-WTD03 leak sensor

gregwjacobs:
Adds low battery and leak detection / alert for Sercomm SZ-WTD03 and SZ-WTD03N water sensors. 

Triggered device leak detector contacts and confirmed alarming occurs in zigbee2mqtt and Home Assistant


Koenkk:
Thanks!

PR #6052, opened 2023-08-06
Add missing Bticino OTA support

RyuzakiKK:
Thanks to https://github.com/Koenkk/zigbee-OTA/pull/322, also the Bticino 4003C and 4411C should be marked as OTA supported.

The 4003C device uses the firmware NLL, while the 4411C the firmware NLFN.


Koenkk:
Thanks!

PR #6056, opened 2023-08-07
Bug Fixes TS0003 manid _TZ3000_odzoiovu

MityaDVG:
_TZ3000_odzoiovu has no backlightMode. I moved it to model: 'TS0003_switch_module_2' but I don't know if there is a "switchType" mode in _TZ3000_4o16jdca.


Koenkk:
Thanks!

PR #6058, opened 2023-08-08
other Amaris LED panels

arteck:


Koenkk:

arteck:
nope i have only a issue our side
https://github.com/ioBroker/ioBroker.zigbee/issues/1830

Koenkk:
Thanks, I needed the manufacturer name which I could find it that issue.

PR #6063, opened 2023-08-09
no 40DB alarm on this device

Chartreusito:
Just opened the 2 devices i own with this manufacturer ID, there is no buzzer or siren in sight.


Koenkk:
Thanks!

PR #6065, opened 2023-08-11
_TZ3000_msl6wxk9 AKA ZMS-102 rework

Chartreusito:
the _TZ3000_msl6wxk9 i own is named  ZMS-102 and looks and works just as the classic TS0202 here : https://www.zigbee2mqtt.io/devices/IH012-RT01.html#tuya-ih012-rt01


Koenkk:
Thanks!

PR #6068, opened 2023-08-11
Fix FP1 Firmware Version Display

Nixon506E:


Koenkk:
Can you give some more context to this?

Nixon506E:
Previously the OTA screen showed no value for the firmware version of the FP1

Koenkk:
I see, thanks for the fix!

ajunio3:
There's one firmware update available for Aqara FP1 (version 58) in Aqara home. How can we have it on OTA in Z2M?
FP1 New Firmware

PR #6078, opened 2023-08-16
Zen Add fan mode, keypad lockout, ota clusters. Remove pi heating demand

Codel1417:
Adds hvacFan mode (Auto/On)
Adds keypad lockout option
Adds generic ZigBee OTA (No files to test though, Just going on a claim by the manufacturer that the thermostat follows the spec)

Removes PI Heating Demand as its not supported by the thermostat.

Sadly this doesn't fix the 'No response received' errors which are common

TODO: Docs


Koenkk:
Thanks!

PR #6079, opened 2023-08-17
Update tuya.ts

oscrey:
adding a tamper sensor for Tuya water leak detector (TS0207_water_leak_detector, _TZ3000_85czd6fy )


Koenkk:
Thanks!

PR #6085, opened 2023-08-19
Update philips.js with new-version Philips Filament ST19 bulb

kattmang:
From https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#done

Tested on my home's zigbee2mqtt instance, works fine on these new bulb skus.


Koenkk:
Thanks!

PR #6087, opened 2023-08-19
Green power 3 support

Gautomax:


src/devices/legrand.js
```diff
@@ -447,6 +465,15 @@ module.exports = [
...
         exposes: [e.action(['press_once', 'press_twice'])],
     },
+    {
+        fingerprint: [{modelID: 'GreenPower_3', ieeeAddr: /^0x00000000005.....$/}],
+        model: '600087L',
+        vendor: 'Legrand',
+        description: 'Wireless and batteryless blind control switch',
+        fromZigbee: [fzlocal.legrand_600087l],
+        toZigbee: [],
+        exposes: [e.action(['cover_stop', 'cover_up', 'cover_down'])]
```
Koenkk, line 475:
```suggestion
        exposes: [e.action(['stop', 'up', 'down'])]
```

I think it makes sense to simplify the actions like this, ca you also update the fz converter?

PR #6092, opened 2023-08-19
Scene rename

vanviegen:
As we can name scenes, I think we should also be able to *re*name them.

A tiny patch to zigbee2mqtt is also required, in order to call `emitScenesChanged()` when a rename happens.

This pull requests also contains an unrelated patch, without which I'm unable to build zigbee-herdsman-converters. I don't understand why building doesn't work for me, while it presumable does for others (I tried multiple nodejs versions), but this solution seems safe and may save other a headache.

### DIFF
--- a/src/converters/toZigbee.js
+++ b/src/converters/toZigbee.js
@@ -4530,6 +4530,36 @@ const converters = {
             meta.logger.info('Successfully removed all scenes');
         },
     },
+    scene_rename: {
+        key: ['scene_rename'],
+        convertSet: (entity, key, value, meta) => {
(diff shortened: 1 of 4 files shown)


Koenkk:
Looks good, could you also make a pr for the docs?  (https://www.zigbee2mqtt.io/guide/usage/scenes.html, click *Help to make docu better* at the bottom)

vanviegen:
Sure! https://github.com/Koenkk/zigbee2mqtt.io/pull/2176

Koenkk:
Bedankt! 😄

PR #6095, opened 2023-08-20
Fix: Ikea 1744 - Add missing action values

Burki24:
Add the following values to e.action:
'rotate_left', 'rotate_right', 'rotate_stop', 'skip_backward', 'skip_forward', 'play_pause'


Koenkk:
The default now is `legacy: false`, if you set this these actions will not show up. (https://www.zigbee2mqtt.io/devices/E1744.html#legacy-integration)

PR #6101, opened 2023-08-23
Add support for Namron water leak sensor 4512764

einarjh:
This adds support for the Water leak sensor from Namron.


Koenkk:
Thanks!

PR #6102, opened 2023-08-24
fix: Override labels in exposes

Drafteed:
This PR overrides labels of some presets.

Let's discuss if I have forgotten something or if you have any doubts.


Koenkk:
Looks good,thanks!

PR #6103, opened 2023-08-24
Add missing battery reporting to Namron 4512764 (Waterleak sensor)

geirra:
This device was missing battery reporting, so added this.
Seems like a oem om HZC unit.


Koenkk:
Thanks!

PR #6106, opened 2023-08-25
feat(add): 929003054401

markkvdb:
The Philips Hue Fair suspension light is available under yet another model name.


markkvdb:
Will look tomorrow for the colour temperature range.

markkvdb:
Should I use the whiteLabel and attach the model to an existing Fair suspension light?

Koenkk:
Thanks! 

> Should I use the whiteLabel and attach the model to an existing Fair suspension light?

No (because this is a unique model)

PR #6108, opened 2023-08-26
Update müller_licht.ts

obrain17:
Add 'Store Scene' to Müller Licht ZBT-DIMController-D0800

Currently no converter is in place for handling this:

````
Debug 2023-08-07 21:49:37 Received Zigbee message from 'DimController', type 'commandStore', cluster 'genScenes', data '{"groupid":16387,"sceneid":1}' from endpoint 1 with groupID 0
````

In case this PR will get merged
- also the documentation needs to be adapted.
- I will do an extra PR in this case then


Koenkk:
Thanks, no need to update the docs manually, the action will automatically be added after the next release.

PR #6112, opened 2023-08-28
Added missing device identifier 929002376402 for Hue Iris Gold limite…

DrHauss:
…d Edition

Added missing device identifier 929002376402 for Hue Iris Gold limited Edition


Koenkk:
Thanks!

PR #6114, opened 2023-08-28
fix: Paulmann devices support Hue/Saturation

sjorge:
It looks like Paulmann devices support Hue/Saturation. (I made the assumption here that probably all of there color enabled devices do). This should fix https://github.com/Koenkk/zigbee2mqtt/issues/18780


Koenkk:
Thanks!

PR #6130, opened 2023-09-02
Add zigbeeModel 929003536001 variation for Philips Hue Iris Copper special edition

davis-griffin:
I picked up a Hue Iris lamp which came with a different zigbeeModel from the ones that were supported by zigbee2mqtt.  Tested this config with my Home Assistant and it seems to do the trick.


Koenkk:
Thanks!

PR #6131, opened 2023-09-02
_TZE204_sooucan5 swap dp's

blakadder:
Manufacturer set the dp names wrong, they need to be be swapped.

Despite having those names in the Tuya device properties it makes no sense for normal operation to have the fading time st to 0.2s while the detection delay is 100s with the default setting.

This PR swaps the names for those two dp's to look and behave correctly


Koenkk:
Thanks!

PR #6134, opened 2023-09-03
Fix Legrand 067755 adapter

jiquem:
The previous adapter was not able to be used with the 067755.

### DIFF
--- a/src/devices/legrand.js
+++ b/src/devices/legrand.js
@@ -46,10 +46,11 @@ const fzlocal = {
 
 module.exports = [
     {
-        fingerprint: [{modelID: ' Wireless Scenes Command\u0000\u0000\u0000\u0000\u0000\u0000\u0000'}],
+        zigbeeModel: [' Pocket remote\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000'+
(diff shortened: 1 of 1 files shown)


src/devices/legrand.js
```diff
@@ -46,11 +46,10 @@ const fzlocal = {
 
 module.exports = [
     {
-        zigbeeModel: [' Pocket remote\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000'+
-            '\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000'],
+        fingerprint: [{modelID: ' Wireless Scenes Command\u0000\u0000\u0000\u0000\u0000\u0000\u0000'}],
```
Koenkk, line 49: This removes support for the `Pocket remote`, please add `Wireless Scenes Command` to it.

jiquem: Sorry, I understand that this remove the support for Pocket Remote, but I don't get how to add Wireless Scenes Command to it. Could you please explain me ?

Koenkk: Can you confirm your device also has model `067755` written on it?

jiquem: Yes, it's written on a sticker on it

Koenkk:
thanks!

src/devices/legrand.js
jiquem: Thanks for the modification.

PR #6147, opened 2023-09-07
 Feat(add): Support Ikea LED2104R3

liquidacid:
New shipped model. BEWARE: The \u001a AND the 8 are really needed, these are also sent from the bulb. This was quite a struggle to figure out since this character is not shown in the debug log but it is shown in the database luckly after pairing.

### DIFF
--- a/src/devices/ikea.js
+++ b/src/devices/ikea.js
@@ -439,6 +439,13 @@ module.exports = [
         extend: tradfriExtend.light_onoff_brightness(),
     },
     {
+        zigbeeModel: ['\u001aTRADFRI bulb GU10 WW 345lm8'],
+        model: 'LED2104R3',
+        vendor: 'IKEA',
+        description: 'TRADFRI LED bulb GU10 WW 345 lumen, dimmable',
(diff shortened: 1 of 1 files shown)


liquidacid:
This is my first PR, can anyone maybe point me in the right direction? I tried google but i can't figure out what spaces the check are the issue.

Koenkk:
Thanks!

OZ1SEJ:
I see there is no pairing info on the page https://www.zigbee2mqtt.io/devices/LED2104R3.html - I can't pair my LED2104R3 bulb.

According to IKEAs instructions, the bulb can be factory reset by switching the power on and off six times. I allow all devices to join (all) via the frontend, and power cycle the bulb six times. It flashes quickly four times, and begins to slowly fade for a few minutes, but at no time do I get any indication of a new device found or interview being started. Am I doing something wrong?

liquidacid:
the pairing works the same as the other versions of the lamps. If its not detected at all there is something else going on, this has nothing to do with this part of the game

PR #6154, opened 2023-09-10
Fix voltage_phase_c missing unit

esplor:
Small PR to fix HA warning:
```
The unit of sensor.datek_eva_han_meter_voltage_phase_c (None) cannot be converted
to the unit of previously compiled statistics (V).
Generation of long term statistics will be suppressed unless the unit changes back to V or a compatible unit. 
```


Koenkk:
Thanks!

PR #6156, opened 2023-09-12
Update acuity_brands_lighting.ts

peetipablo:
Add ABL-LIGHT-Z-201 zigbee model (RB56SC).  Functionality is identical to existing ABL-LIGHT-Z-001 model, different chassis.

https://www.amazon.com/gp/product/B0BQ1VJLPL/ref=ppx_yo_dt_b_search_asin_image?ie=UTF8&th=1


Koenkk:
Thanks!

jtviolet:
I've just added this exact model light (exact purchase from that Amazon page, by chance). The light is showing up in Z2M as zigbee model ABL-LIGHT-Z-201 (Model RB56SC).

There are no controls showing up to control RGB, just brightness and temperature. Is there any way to fix this? Is it that I need to add this converter?

CattDamon:
Does anyone have issues with the ramp up with these. When they turn on to 100% they flicker then set to 100%. Tried to delete and repair them tried to set the transition to 10 seconds and the turn off transition is nice and gradual but the turn on is very abrupt.

PR #6159, opened 2023-09-14
feat: add power outage memory to Mercator Ikuü SPP02GIP

georg:
The SPP02GIP is the exterior version of the SPP02G. Power outage memory was added to SPP02G in #5968


Koenkk:
Thanks!

PR #6160, opened 2023-09-14
Detect TS000F / _TZ3000_skueekg3 as TuYa WHD02 Wall switch module

alray31:
Another one:
Same units WHD02 bought at the same time, one already suported, one unsupported with a different ModelID / ManufacturerName:

image

image

image

image


Koenkk:
Thanks!

PR #6163, opened 2023-09-15
feat: expose color options for Ikea lights

protyposis:
Expose `color_options.execute_if_off` for Ikea lights to simplify enabling color (temperature) adjustments while lights are off. Previously, an MQTT message had to be sent (https://github.com/Koenkk/zigbee2mqtt/issues/18699#issuecomment-1690094357).

Even though the `ToZigbeeConverter` for `color_options` is added to all lights by default, docs and issues discuss this setting only in context of Ikea lights, so I decided to add this only to Ikea lights. Implements https://github.com/Koenkk/zigbee2mqtt/issues/18717.

image


src/lib/exposes.ts
```diff
@@ -376,6 +376,18 @@ export class Light extends Base {
 
         return this;
     }
+
+    withColorOptions() {
```
Koenkk, line 380: Can you add this as a separate expose instead of integrating it in light?

protyposis: Yes I can. By what rule do you decide where it should be placed?

Koenkk: Usually wether it is part of the Home Assistant entity or not (e.g. https://www.home-assistant.io/integrations/light.mqtt/) (not always but it is a good guideline)

protyposis: 0fa62e5b51e515c1fcc804e5595f15c73a8046d1

Koenkk:
Clean, thanks!

PR #6169, opened 2023-09-17
feat(add): Orvibo curtain motor - W45CZ

EarthEclipse:
New version of an Orvibo curtain motor.

I was very inspired to remove a trailing space in the previous version's description, hence the patch looks like it does. Sorry about it. 
The patch should contain the same definition as for the previous motor, just with a different `model` and `zigbeeModel`.


Koenkk:
Thanks!

PR #6175, opened 2023-09-18
Restore power_on_behavior for Gledopto GL-C-006P and GL-LB-001P

ov1d1u:
Support for `power_on_behavior` was disabled in b0b91c724dea790816d4927b1050faf9256b9124 as a result of Koenkk/zigbee2mqtt#16873 for all the Gledopto devices, but from my tests it perfectly works on GL-C-006P and GL-LB-001P, at least.

This pull request restore it for these two devices, although I'd recommend to restore it for all the devices and only disable it for the device reported in Koenkk/zigbee2mqtt#16873.

### DIFF
--- a/src/devices/gledopto.ts
+++ b/src/devices/gledopto.ts
@@ -276,7 +276,7 @@ const definitions: Definition[] = [
         vendor: 'Gledopto',
         ota: ota.zigbeeOTA,
         description: 'Zigbee LED Controller RGB (pro)',
-        extend: gledoptoExtend.light_onoff_brightness_color({noConfigure: true}),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

sschwieb:
I stumbled across this since I just switched from Deconz to Zigbee2MQTT, and was wondering why all my Gledopto bulbs suddenly turn green after power on, no matter what color they were set to before. After tweaking `gledopto.ts` similar to this PR for `GL-S-005Z`, they now come back with the light settings they had before cutting power.

@Koenkk , would you want me to open a PR for that? I don't have any other Gledopto devices, but @ov1d1u 's recommendation to restore the support for `power_on_behaviour` for all of them might be best... 🤷

Koenkk:
@sschwieb yes, please create a PR

PR #6179, opened 2023-09-18
Add support for new Festavia with 100 LEDs

mundschenk-at:
This adds support for the new Festavia "V2" light string with 100 LEDs. There are also new variants with 250 LEDs and 400 LEDs. Unfortunately I don't know their model IDs and numbers. All of these should also support three new Hue effects (besides sparkle), namely `opal`, `glisten` and `prism`.

By trial and error I have found the settings for `opal` and `glisten`, but `prism` is apparently _not_ `2100010d`. Unfortunately I have no equipment for sniffing the traffic of the official Hue app.


Koenkk:
Cool, thanks!

PR #6183, opened 2023-09-19
feat: Add expose presets for weather stations

andrejsoucek:


Koenkk:
Then for this case you can keep them in your ext converter (which is fine)

andrejsoucek:
Yes sure but since we already have presets for pressure and humidity, I thought it can be helpful to have QNH and wind speeds. If you don't agree, that's legit, feel free to close this PR. ;)

Koenkk:
We only want to have exposes which are used in zigbee-herdsmsn-converters so I will close this.

PR #6185, opened 2023-09-20
Update lellki.ts

gdjang:
Add zigbeeModel E220-KR5N0Z0-HA

I have purchased the Lellki WP33. When pairing it with Zigbee, it is recognized as E220-KR5N0Z0-HA, and it does not function properly. I have used an external converter to add E220-KR5N0Z0-HA, but it seems that simply adding the Zigbee model to the existing lellki.ts file would solve the issue.


Koenkk:
Thanks!

PR #6190, opened 2023-09-22
Add missing min\max exposes for 3 gang dimmer

kuzaxak:
`_TZE204_1v1dxkck` which is `TS0601_dimmer_3` has the same modes as `TS0601_dimmer_2`. I tested all settings, and they are working fine with my device.


Koenkk:
Thanks!

Issue #5627, opened 2023-04-01
Help needed to add the last missing feature of RM3500ZB


alray31:
So with the new update of z2m today (1.30.3-1), we can now read/write to the manufacturer specific attributes that were recently added to the ZCL using the dev console. I was able to test the drConfigWaterTempMin attribute from the manuSpecificSinope cluster. 

On the Sinopé Calypso (RM3500ZB), we should be able to turn on or off the minimum temperature at which the device automatically restart to heat. This is done with drConfigWaterTempMin (attribute 0x0076 of cluster 0xff01). It's default value is 45 (data type t.uint8_t)) which will make the RM3500ZB restart to heat by itself when the water temperature reach 45 °C. Writing a value of 0 in this attribute does disable this automatic feature (tested with the dev console). Setting any other value will adjust the temperature at which the RM3500ZB will restart to heat by itself (also tested with the dev console)

How could we add a slider to the exposes tab that could be adjusted from 0 to 45 , and once set to a given value, will write the corresponding value to the drConfigWaterTempMin attribute?

Or at least, could we add a switch to the exposes tab to turn on or off this feature? I.e: ON would send the value  ''45'' in drConfigWaterTempMin and OFF  would send the value ''0''.

I tried a few things to make it work, but it always fails. I can test with an external converter on my side if someone more experienced in coding can help me.

Here is the existing converter from zigbee-herdsman-converters/devices/sinope.js 
```
        zigbeeModel: ['RM3500ZB'],
        model: 'RM3500ZB',
        vendor: 'Sinopé',
        description: 'Calypso smart water heater controller',
        fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fzLocal.ias_water_leak_alarm,
            fzLocal.sinope, fz.temperature],
        toZigbee: [tz.on_off],
        exposes: [e.switch(), e.power(), e.current(), e.voltage(), e.energy(), e.water_leak(), e.temperature()],
        configure: async (device, coordinatorEndpoint) => {
            const endpoint = device.getEndpoint(1);
            const binds = ['genOnOff', 'haElectricalMeasurement', 'seMetering', 'msTemperatureMeasurement', 'ssIasZone',
                'manuSpecificSinope'];
            await reporting.bind(endpoint, coordinatorEndpoint, binds);
            await reporting.onOff(endpoint);
            await reporting.temperature(endpoint, {min: 10, max: 301, change: 10}); // divider 100: 0.1C
[...]

PR #6201, opened 2023-09-27
Added missing feature ''Low water temp protection adjustment'' for RM3500ZB

alray31:
Tested locally by editing /node_modules/zigbee-herdsman-converter/devices/sinope.js
Had to reformat for typescript (sinope.ts), might want to double check just in case.

Complete explanation at https://github.com/Koenkk/zigbee-herdsman-converters/issues/5627

calypso

From the manufacturer website:
image


Koenkk:
Thanks!

PR #6202, opened 2023-09-27
philips hue go 7602031N6 supports colours

meeb:
The Philips Hue Go light (7602031N6) supports full RGB, minor tweak to enable colour changing support. Tested and working with my own 7602031N6 lights.


Koenkk:
Thanks!

PR #6208, opened 2023-09-28
fix(detect): Detect (TS0001, _TZ3000_h8ngtlxy) as TuYa Smart water/gas valve

garbagecollected:
I added the smart valve to zigbee2mqtt and it was originally detected as a 1-gang-switch, which was wrong.

So I added the zigbee model and manufacturer (as seen in the attached photo) to the fingerprint list and now it's detected properly.

description


Koenkk:
thanks!

PR #6218, opened 2023-10-02
Update bosch.ts

smarts4home:
Added new zigbee model RBSH-SWDV-ZB/BSEN-CV - Door/window contact II


Koenkk:
Thanks!

psdl76:
@smarts4home On the integration of BSEN-CV. In the Bosch App it has the ability to change the sensibility of the vibration detection vom "very high", "high", "middle", "low" to "very low". Can you integrate this in zigbee2mqtt. My TV is vibrating the glasses in my living room because the standard is set to "very high". :)

PR #6220, opened 2023-10-03
Update battery for Wiser TRV

jamesonuk:
Battery was not exposed and was not calculated accurately.,

I wasn't sure whether to remove withRunningState as this does not appear to be supported by the device 
```
 2023-10-03 14:29:25Publish 'set' 'read' to 'Kitchen TRV' failed: 'Error: Read 0x50325ffffe7f5cd7/1 hvacThermostat(["runningMode"], {"sendWhen":"active","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')'
```
but I wasn't sure if there are any considerations to just remove this.

It does appear to support systemMode 
```
2023-10-03 14:31:25Read result of 'hvacThermostat': {"systemMode":1}
```
but I wasn't sure about how to figure out which modes it actually supports ?

I figured the battery change should be simple and then I can update modes later


Koenkk:
Thanks!

PR #6231, opened 2023-10-04
Support for _TZ3000_mugyhz0q / 899WZ 

Chartreusito:
Converter for 899WZ which is similar to Tuya TS0207_water_leak_detector but without any physical tamper switch and with a 80DB alarm.
https://www.aliexpress.com/item/1005005942082875.html

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1128,6 +1128,7 @@ const definitions: Definition[] = [
             tuya.whitelabel('TuYa', 'TS0207_water_leak_detector_1', 'Zigbee water flood sensor + 1m probe cable', ['_TZ3000_ocjlo4ea']),
             tuya.whitelabel('TuYa', 'TS0207_water_leak_detector_2', 'Zigbee water leak sensor', ['_TZ3000_upgcbody']),
             tuya.whitelabel('TuYa', 'TS0207_water_leak_detector_3', 'Zigbee water leak sensor', ['_TYZB01_sqmd19i1']),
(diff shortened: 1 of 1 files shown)


Koenkk:
Please merge with `TS0207_water_leak_detector` and use a dynamic expose to filter out tamper (e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/997cb6469012c80d98a7f60c90b0fac16e8e2541/src/devices/tuya.ts#L607)

src/devices/tuya.ts
```diff
@@ -1136,7 +1137,14 @@ const definitions: Definition[] = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg']);
             await reporting.batteryPercentageRemaining(endpoint);
         },
-        exposes: [e.water_leak(), e.tamper(), e.battery_low(), e.battery()],
+        exposes: (device, options) => {
+            const exps: Expose[] = [e.water_leak(), e.battery_low(), e.battery()];
+            if (!device || device.manufacturerName === '_TZ3000_mugyhz0q') {
```
Koenkk, line 1142:
```suggestion
            if (!device || device.manufacturerName !== '_TZ3000_mugyhz0q') {
```

I guess you don't want to add tamper in case of `_TZ3000_mugyhz0q`?

Chartreusito: Thanks, that was exactly the idea. I love that's a possibility.

Koenkk:
Thanks!

PR #6234, opened 2023-10-04
Add SP02-ZB001 Infrared Motion Sensor from iAlarm

bdavis2-PCTY:


bdavis2-PCTY:
PR for image: https://github.com/Koenkk/zigbee2mqtt.io/pull/2247

src/devices/tuya.ts
```diff
@@ -6322,6 +6322,27 @@ const definitions: Definition[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_mgxy2d9f']),
```
bdavis2-PCTY, line 6326: Reason  for using a fingerprint is that I know there's other models of this mention sensor, I'm just not sure what the manufacturer names are. 

EX, I know that this motion sensor the Meian SP02 is the same: https://www.aliexpress.us/item/3256804475730796.html

Koenkk:
Thanks!

PR #6235, opened 2023-10-04
set MTG275-ZB-RL as whitelabel

Chartreusito:
I opened _TZE204_dtzziy1e to confirm it's a 24G sensor as advertised.
Works out of the box with the existing converter, just changed it as whitelabel to fix model and description.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6054,7 +6054,7 @@ const definitions: Definition[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint('TS0601', ['_TZE204_sbyx0lm6', '_TZE204_clrdrnya']),
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE204_sbyx0lm6', '_TZE204_clrdrnya','_TZE204_dtzziy1e']),
         model: 'MTG075-ZB-RL',
         vendor: 'TuYa',
         description: '5.8G human presence sensor with relay',
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6054,7 +6054,7 @@ const definitions: Definition[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -6054,10 +6054,13 @@ const definitions: Definition[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint('TS0601', ['_TZE204_sbyx0lm6', '_TZE204_dtzziy1e', '_TZE204_clrdrnya']),
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE204_sbyx0lm6', '_TZE204_clrdrnya']),
```
Koenkk, line 6057: `_TZE204_dtzziy1e` shouldn't be removed here as the device will not be detected as supported anymore.

Chartreusito: Dully noted, thanks !

Koenkk:
Thanks, could you also submit a pr for the picture?

Chartreusito:
Is it necessary if the device is visually identical to the main device ?
Its looks the same as this : https://www.zigbee2mqtt.io/devices/MTG075-ZB-RL.html#tuya-mtg075-zb-rl
Internally it's a different sensor

Koenkk:
Yes, otherwise the frontend won't show a picture for it.

Chartreusito:
Picture PR for MTG275-ZB-RL : https://github.com/Koenkk/zigbee2mqtt.io/pull/2255

striker54:
Soory in the first place if this isn't the right place.
When trying to adjust Illuminance threshold value it gives the following error **No converter available for 'illuminance_threshold'**
Anyone has the same problem?

djcrawleravp:
Anyone having the same "spammy network" problem with the MTG275-ZB-RL?

yfaykya:
> Anyone having the same "spammy network" problem with the MTG275-ZB-RL?

Did you ever find a solution?

PR #6236, opened 2023-10-04
Zg 2002 rf

Chartreusito:
Add fingerprint for ZG-2002-RF
It's this device : https://www.aliexpress.com/item/1005004888188521.html?algo_exp_id=8ba97e9b-dc1c-45df-b39a-163a3d7f574a-1
It's a 2 CH relay switch, with the possibility to enable interlock mode (not via zigbee sadly).


Koenkk:
Thanks!

PR #6252, opened 2023-10-07
Readability improvements: Legrand.ts + Bticino.ts

FabianMangold:
1. Improved the readability by having:
        - Reordered definition attributes on all devices (see pattern)
        - Prettified the code

Pattern:
> zigbeeModel
> model
> vendor
> description
> whiteLabel
> extend
> ota
> meta
> fromZigbee
> toZigbee
> exposes
> onEvent
> configure
> endpoint

2. Also removed the definition for the 067776_inverted device due to:
        - the fact the Legrand OTA now works reliably
        - the cover can be inverted on the classic version
        
>  Caution: This may break backwards compat. !
>  Please cherry pick from that PR in case the latter commit is not acceptable

### DIFF
--- a/src/devices/bticino.ts
+++ b/src/devices/bticino.ts
@@ -14,16 +14,18 @@ const definitions: Definition[] = [
         model: 'K4003C/L4003C/N4003C/NT4003C',
         vendor: 'BTicino',
         description: 'Light switch with neutral',
+        ota: ota.zigbeeOTA,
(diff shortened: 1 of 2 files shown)


Koenkk:
I propose to keep 067776_inverted to not break existing setups (even when they can OTA)

FabianMangold:
OK, makes sense!
PR updated, merge conflict solved.
Also took the liberty of applying the same changes to bticino.ts

Koenkk:
Thanks!

PR #6257, opened 2023-10-08
Update philips.ts with model 1744230P7.

kernsche:
Add support for the Phillips Hue Econic outdoor Post Light 1744230P7.

Its the same Light as the 1744130P7 but with a higher Post.


Koenkk:
Thanks!

PR #6258, opened 2023-10-09
Add Philips Hue PAR38 Outdoor Spotlight bulb support

rgregg:
Add support for the new Philips Hue White and Color PAR38 Outdoor Spotlight bulb

### DIFF
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -2507,6 +2507,13 @@ const definitions: Definition[] = [
         extend: philips.extend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}),
     },
     {
+        zigbeeModel: ['LCS002'],
+        model: '577262',
+        vendor: 'Philips',
+        description: 'Hue White and Color Ambiance PAR38 Outdoor Spotlight',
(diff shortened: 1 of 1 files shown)


Koenkk:
can you provide a link to a device?

rgregg:
Yep, this is the device: https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-par38---e26-smart-bulb/046677577261

src/devices/philips.ts
```diff
@@ -2506,6 +2506,13 @@ const definitions: Definition[] = [
         description: 'Hue Lily outdoor spot light',
         extend: philips.extend.light_onoff_brightness_colortemp_color({colorTempRange: [153, 500]}),
     },
+    {
+        zigbeeModel: ['LCS002'],
+        model: '577262',
```
Koenkk, line 2511: When googling on this number, I don't get this bulb, either:

- Use `046677577261` here (EAN)
- Or a model number that is written on the device/box, I expect something in the format of `1742830P7`

rgregg: Ah thanks, I updated this to match what is on the device itself. In Googling 9290031508, I see a lot of indications that it should have an X at the end (e.g. FCC website), but the device does not have the X on the model number.

Koenkk:
Thanks!

PR #6259, opened 2023-10-09
Repair：_TZE204_ijxvkhd0 

yuanxin32323:
Repair：The status displayed in the presence is opposite to the actual…
Repair: "detection_distance_max" The range of max should be 1.5-5.5
Repair: "detection_distance_min" The meaning of min should be target_distance


Koenkk:
Thanks!

PR #6261, opened 2023-10-09
Update bosch.ts

sestriel:
- Removed unsupported "display_orientation" (0x400b) from endpoint.read because it leads to error "UNSUPPORTED_ATTRIBUTE".
- Added binding for humidity (msRelativeHumidity), humidity reporting is now working
- Added reporting for keypadlockout
- Added reporting for "window_open"


Koenkk:
Thanks!

PR #6266, opened 2023-10-10
fix(detect): Detect _TZ3210_mja6r5ix and _TZ3210_it1u8ahz as TS0505B

felipejfc:
- (fix): Properly set model
- (fix): Control both lights with led_control so that brightness can be set when in color mode.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -932,6 +932,22 @@ const definitions: Definition[] = [
         },
     },
     {
+        fingerprint: tuya.fingerprint('TS0505B', ['_TZ3210_mja6r5ix', '_TZ3210_it1u8ahz']),
+        model: 'TS0505B',
+        vendor: 'TuYa',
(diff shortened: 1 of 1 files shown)


Koenkk:
I think this can be merged with https://github.com/Koenkk/zigbee-herdsman-converters/blob/80036cd6f7e9804f2e419a3798123f72661c0b4d/src/devices/tuya.ts#L920

felipejfc:
> I think this can be merged with
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/80036cd6f7e9804f2e419a3798123f72661c0b4d/src/devices/tuya.ts#L920

Thanks for reviewing. I thought it could be merged too, however, the min temp reported by the device is 143, while the one you sent specify 153. Is it still possible to merge despite this?

Koenkk:
Then set it to `143`

felipejfc:
done! ptal

Koenkk:
Thanks!

PR #6260, opened 2023-10-09
feat(ignore): Enable incremental TSC builds

FabianMangold:
Hello @Koenkk,

Could TSC incremental builds be considered (at least on the DEV branch) ?
This tremendously improves rebuilds, especially on low power hardware.

Thanks a lot

PR #6267, opened 2023-10-10
chore: Revert "feat(ignore): Enable incremental TSC builds"

Koenkk:
Reverts Koenkk/zigbee-herdsman-converters#6260

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,8 +13,7 @@
       "outDir": ".",
       "baseUrl": ".",
       "resolveJsonModule": true,
-      "allowJs": true,
-      "incremental": true
+      "allowJs": true
     },
     "include": ["./src/"],
     "exclude": []


Koenkk:
@FabianMangold unfortunately I have to revert this, after this change the published zigbee-herdman-converters package doesn't contain any source code anymore. Could you investigate this?

- https://github.com/Koenkk/zigbee2mqtt/pull/19234 (only diff between 15.92.0 and 15.92.1 was reverting this change)
- Screenshot 2023-10-10 at 22 19 52

FabianMangold:
@Koenkk I can reproduce the issue locally now.
Will investigate for sure and let you know the outcome, apologies for the inconvenience caused.

FabianMangold:
Hi @Koenkk,
Turns out that the fix is trivial ! We just need to delete the buildcache when performing the "clean" task.

The explanation is actually simple. The github CI action does the following:
- npm ci
- build
- test
- lint
- publish

The "publish" step is the one that broke the build. Indeed, "publish" internally makes use of "package", and thus runs the package "pre-" hook, which is defined as "_npm run clean && npm run build"_
When running clean, the output files were swept away, but the build cache file (tsconfig.tsbuildinfo" remained there, breaking subsequent builds.
TSC seems to rely 100% on the build cache on not actually verify the presence of files..

I'll create a PR shortly in order to re-enable the builds.

PR #6273, opened 2023-10-11
fix: Re-enable incremental TSC builds

FabianMangold:
With ref. to the last comment in [PR#6267](https://github.com/Koenkk/zigbee-herdsman-converters/pull/6267), this PR contains a fix to (re)enable incremental TSC builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "lint": "eslint .",
     "test": "ZHC_TEST=true jest test",
     "test-watch": "ZHC_TEST=true jest test --watch",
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks! Could you also make such pr for zigbee2mqtt?

FabianMangold:
Sure, at a first glane, I see that there is no clean or (pre)pack scripts defined.
Let me check how the build pipeline works and I'll try to come up with something.

What about zigbee-herdsman ?

Koenkk:
> What about zigbee-herdsman ?

Yes please 😄

PR #6277, opened 2023-10-12
Update bosch.ts

smarts4home:
Added Bosch Outdoor Siren


Koenkk:
Could you fix the issues which cause the build to fail?

PR #6281, opened 2023-10-13
fix(detect): Detect _TZ3000_empogkya and _TZ3000_lubfc1t5 as TuYa ZM-L03E-Z and M10Z

felipejfc:


src/devices/tuya.ts
```diff
@@ -2180,6 +2178,8 @@ const definitions: Definition[] = [
         },
         whiteLabel: [
             tuya.whitelabel('Lonsonho', 'X703A', '3 Gang switch with backlight', ['_TZ3000_rhkfbfcv']),
+            tuya.whitelabel('Zemismart', 'ZM-L03E-Z', '3 gang switch with neutral', ['_TZ3000_empogkya']),
+            tuya.whitelabel('TuYa', 'M10Z', '2 gang switch with 20A power socket', ['_TZ3000_lubfc1t5']),
```
Koenkk, line 2182: > 2 gang switch

This definition is for a 3 gang switch

felipejfc: This device is a 2 Gang with a power socket that is also threated as a switch (same controls). Regardless, I will split them

felipejfc: Do you think we need to split?

felipejfc:
Added image: https://github.com/Koenkk/zigbee2mqtt.io/pull/2267

src/devices/tuya.ts
Koenkk: Then this is fine!

Koenkk:
Thanks!

PR #6282, opened 2023-10-14
busch-jaeger new states

arteck:
Long print is supported, differentiation between the available elements (1-row/2-row/4-row. works with main or battery devices

### DIFF
--- a/src/devices/busch-jaeger.ts
+++ b/src/devices/busch-jaeger.ts
@@ -17,103 +17,54 @@ const definitions: Definition[] = [
         extend: extend.switch(),
     },
     {
-        // Busch-Jaeger 6735, 6736 and 6737 have been tested with the 6710 U (Power Adapter),
-        // 6711 U (Relay) and 6715 U (dimmer) back-ends. Unfortunately both the relay and the dimmer
-        // report as model 'RM01' with genLevelCtrl clusters, so we need to set up both of them
-        // as dimmable lights.
-        //
-        // The battery-powered version of the device ('RB01') is supported as well. These devices are
-        // sold as Busch-Jaeger 6735/01, 6736/01 and 6737/01.
-        //
-        // In order to manually capture scenes as described in the devices manual, the endpoint
-        // corresponding to the row needs to be unbound (https://www.zigbee2mqtt.io/information/binding.html)
-        // If that operation was successful, the switch will respond to button presses on that
-        // by blinking multiple times (vs. just blinking once if it's bound).
-        zigbeeModel: ['RM01', 'RB01'],
-        model: '6735/6736/6737',
+        fingerprint: [{modelID: 'RM01', endpoints: [{ID: 10}, {ID: 11}, {ID: 12}, {ID: 13}, {ID: 18}]}],
+        model: '6737',
         vendor: 'Busch-Jaeger',
-        description: 'Zigbee Light Link power supply/relay/dimmer/wall-switch',
+        description: 'Zigbee Light Link power supply relay/dimmer 4rows',
         endpoint: (device) => {
             return {'row_1': 0x0a, 'row_2': 0x0b, 'row_3': 0x0c, 'row_4': 0x0d, 'relay': 0x12};
         },
-        exposes: (device, options) => {
-            const expose = [];
-
-            // If endpoint 0x12 (18) is present this means the following two things:
-            //  1. The device is connected to a relay or dimmer and needs to be exposed as a dimmable light
-            //  2. The top rocker will not be usable (not emit any events) as it's hardwired to the relay/dimmer
-            if (!device || device.getEndpoint(0x12) != null) {
-                expose.push(e.light_brightness().withEndpoint('relay'));
-                // Exposing the device as a switch without endpoint is actually wrong, but this is the historic
-                // definition and we are keeping it for compatibility reasons.
-                // DEPRECATED and should be removed in the future
-                expose.push(e.switch());
-            }
-            // Not all devices support all actions (depends on number of rocker rows and if relay/dimmer is installed),
-            // but defining all possible actions here won't do any harm.
-            expose.push(e.action([
-                'row_1_on', 'row_1_off', 'row_1_up', 'row_1_down', 'row_1_stop',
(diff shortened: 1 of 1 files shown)


Koenkk:
Does the change the exposed `action` values?

arteck:
yes to
 'on_row_1', 'off_row_1', 'brightness_step_up_row_1', 'brightness_step_down_row_1', 'brightness_stop_row_1',
              'on_row_2', 'off_row_2', 'brightness_step_up_row_2', 'brightness_step_down_row_2', 'brightness_stop_row_2',

Koenkk:
That's a breaking change which we should prevent, can we revert to the old action names?

arteck:
you mean  

'row_1_on', 'row_1_off', 'row_1_up', 'row_1_down', 'row_1_stop',
                'row_2_on', 'row_2_off', 'row_2_up', 'row_2_down', 'row_2_stop',
                'row_3_on', 'row_3_off', 'row_3_up', 'row_3_down', 'row_3_stop',
                'row_4_on', 'row_4_off', 'row_4_up', 'row_4_down', 'row_4_stop',

i think yes.. its only new name for the action

Koenkk:
Then I propose to not merge this, it creates an unnecessary breaking change for users.

arteck:
its a pity,, ok we rename it to old structure.. 
the new one would be more appropriate

Koenkk:
Can you fix the linting issues?

src/devices/busch-jaeger.ts
```diff
@@ -139,6 +89,202 @@ const definitions: Definition[] = [
...
+            const endpoint11 = device.getEndpoint(0x0b);
+            if (endpoint11 != null) {
+                const index = endpoint11.outputClusters.indexOf(5);
+                if (index > -1) {
+                    endpoint11.outputClusters.splice(index, 1);
+                }
+                await reporting.bind(endpoint11, coordinatorEndpoint, ['genLevelCtrl']);
+            }
+            const endpoint12 = device.getEndpoint(0x0c);
+            if (endpoint12 != null) {
```
Koenkk, line 222: I think you can remove all these `if` checks on the endpoints, because the fingerprint will only match when the device has endpoint `12`

Koenkk:
Thanks!

klada:
@Koenkk I saw this PR randomly and I do see __a lot__ of issues and breaking setups with it. I don't know if this is the right place to raise concerns. If not I'll create an issue.

@arteck What were you trying to accomplish? Is there an issue you were trying to fix?

When looking at the code I do see the following issues:

1. Support for RM01 devices on a Busch-Jaeger power supply has been dropped (RM01 without EP18). Why is this?
2. A lot of code was duplicated. I think the converter is much harder to maintain now.
3. Many important comments have been removed. These Busch-Jaeger devices are very very special and the comments tried to describe _why_ things have been the way they were before. By reading the comments carefully one would have noticed that there are RM01 devices without endpoint 18 for example. This information is now lost.
4. Now endpoint 10 is bound for RM01 devices on a dimmer/relay. This does not have any effect on these devices. The binding is meaningless and nedlessly fills up the low device memory.
5. **This change will break all existing automations for people** as actions have been changed.

muhlatte:
Hey klada,

because I was mainly responsible for the changes, here some annotations concerning your question from my position:

1: Support of the power supplied is given, see 
image
image
We should change that the power supplied one (official describtion 6735) has no row1 action buttons, because this will not be send by the device. The batterie one (official describtion 6735/01) send it.

2: now there is a distinction between the different models (1/2/4 rows; power supplied/battery), before you only have one model with always 4 rows, relay and brightness states, so that were many meaningless actions. Due to it's simularity the code is not hard to understand, the maintainance of this converter should not be a problem in future.

3: I think we can reimport the comments if desired.

4: much better than before, see comment for point 2

5: There were complaints about the change before with the other/old actions, see kartracers issue topic https://github.com/ioBroker/ioBroker.zigbee/issues/1842 , me including. Especially because there was no possiblity to differenciate short and long press on the buttons. This was supported before (so already such a change in the history, now the role backwards) -> That is for me the main reason for the change.
We know that this change will break the existing automations. Myself has in total 23 of this devices. The change of all my javascripts/blocklies was done in 1 hour of work, so overall with the benefits of long press support, it is worth. Just convert your blocklies in javascript and do a search and replace.

Hope we could give you some backgroundinformations for the change and I hope you can live with the changes.

Best regards,
Daniel

klada:
@muhlatte First of all, thank you for your reply. I still think that we are not talking about the same thing and that some things have been overlooked. See my comments below:

> 1: Support of the power supplied is given, see

In the image I can see a 6715U (dimmer). I am talking about the power adapter (6710U). Support for the power adapter (RM01 without endpoint 18) as been removed entirely by this change.

> We should change that the power supplied one (official describtion 6735) has no row1 action buttons, because this will not be send by the device. 

The 6735/01, 6736/01 and 6737/01 all send a row1 action, but only when installed to a 6710U. When installed to a 6711U (relay) or 6715U (dimmer) the action is not sent. Please see the code and comments in the old version.

> 2: now there is a distinction between the different models (1/2/4 rows; power supplied/battery), before you only have one model with always 4 rows, relay and brightness states, so that were many meaningless actions. Due to it's simularity the code is not hard to understand, the maintainance of this converter should not be a problem in future.

What is the problem with the meaningless actions? They never get triggered, so they don't hurt. If this really was a problem, why not change this [single statement](https://github.com/Koenkk/zigbee-herdsman-converters/blob/5be8bf3d5d4eaa611b38b7701ad6f102fb0e6073/src/devices/busch-jaeger.ts#L54), depending on the endpoints present?

> 4: much better than before, see comment for point 2

Why is it better to create bindings which do not work instead of creating only the required bindings?

> 5: There were complaints about the change before with the other/old actions, see kartracers issue topic [ioBroker/ioBroker.zigbee#1842](https://github.com/ioBroker/ioBroker.zigbee/issues/1842) , me including. Especially because there was no possiblity to differenciate short and long press on the buttons. 

Even with the old behavior, long-press events were emitted properly through MQTT. If you cannot use them in iobroker, then this is an iobroker bug. I think there is no need to break all peoples setups because of this.

PR #6287, opened 2023-10-15
Add additional zigbee id for Philips Play Lightbar

chrishas35:
I found that one of my three Play Lightbars reports a different Zigbee id, PCM002.  Endpoints are all the same.


Koenkk:
Thanks!

PR #6289, opened 2023-10-15
Revert "fix: yookie cover position"

barlock:
Reverts Koenkk/zigbee-herdsman-converters#6211

Sorry, after using this for longer and still seeing some bizarre behavior I started looking into this more and reading more through this thread https://github.com/zigpy/zha-device-handlers/issues/1306

My _guess_ is that I happen to have old firmware that is particularly buggy and newer firmware likely doesn't have my problem.

Leaving this change as is, from what I can tell, is the correct thing for newer devices and I'll need to write a custom converter.


Koenkk:
Thanks!

PR #6295, opened 2023-10-16
Add support for Philips Hue Play Gradient Light strip (PC version)

j-simian:
https://www.philips-hue.com/en-gb/p/lightstrips-play-gradient-lightstrip-for-pc/8719514434530

This is for the 32-34" model, not sure if it would be different on the 24"-27"

The file I have as an external converter:

```js
const ota = require('zigbee-herdsman-converters/lib/ota');
const { extend } = require('zigbee-herdsman-converters/lib/philips');

module.exports = [{
    zigbeeModel: ['LCX006'],
    model: '929003498601',
    vendor: 'Philips',
    description: 'Hue gradient lightstrip',
    meta: { turnsOffAtBrightness1: true },
    extend: extend.light_onoff_brightness_colortemp_color_gradient({ colorTempRange: [153, 500] }),
    ota: ota.zigbeeOTA,
}];
```

### DIFF
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -2741,7 +2741,7 @@ const definitions: Definition[] = [
         zigbeeModel: ['LXC006'],
         model: '929003498601',
         vendor: 'Philips',
-        description: 'Play Gradient Lightstrip for PC (32-34)',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #6300, opened 2023-10-17
Update philips.ts surimu round panel

sorryusernameisalreadytaken:
add the third surimu panel, the round one.
Material-Number (12NC): 929003598101

Test it locally since 4 days and it worked so far. Include binding.


Koenkk:
Thanks!

PR #6303, opened 2023-10-17
Add TS004F / _TZ3000_nuombroo

konistehrad:
Closes https://github.com/Koenkk/zigbee2mqtt/issues/19327 by adding support for _TZ3000_nuombroo

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2003,42 +2003,19 @@ const definitions: Definition[] = [
          */
     },
     {
-        fingerprint: tuya.fingerprint('TS004F', ['_TZ3000_nuombroo']),
+        fingerprint: tuya.fingerprint('TS004F', ['_TZ3000_nuombroo', '_TZ3000_xabckq1v', '_TZ3000_czuyt8lz']),
         model: 'TS004F',
         vendor: 'TuYa',
         description: 'Wireless switch with 4 buttons',
         exposes: [
-            e.battery(), 
+            e.battery(),
             e.enum('operation_mode', ea.ALL, ['command', 'event']).withDescription(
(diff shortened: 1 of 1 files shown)


konistehrad:
Ah sorry I’m not entirely sure how to set this up it looks like. Should I remove the model firld from both objects?

src/devices/tuya.ts
```diff
@@ -2001,6 +2001,35 @@ const definitions: Definition[] = [
...
+        model: 'TS004F',
+        vendor: 'TuYa',
+        description: 'Wireless switch with 4 buttons',
+        exposes: [
+            e.battery(), 
+            e.enum('operation_mode', ea.ALL, ['command', 'event']).withDescription(
+                'Operation mode: "command" - for group control, "event" - for clicks'),
+            e.action(['on', 'off', 'brightness_step_up', 'brightness_step_down', 'brightness_move_up', 
+                'brightness_move_down', '1_single', '1_double', '1_hold', '2_single', '2_double', '2_hold',
+                '3_single', '3_double', '3_hold', '4_single', '4_double', '4_hold'])],
```
Koenkk, line 2015: Do the actions depend on the `operation_mode`? (`command = `on`/`off`, etc, `event` = `1_double`, etc.)?
If this is the case my previous statement was wrong and you can merge these.

konistehrad: Yeah, exactly. In `command` mode the device will broadcast the specific action messages to the group (`on`, `off` etc) but in `event` mode it sends special frames handled by `tuya_on_off_action` in fromZigbee.  

Regarding the duplicate model error: should I suffix the new model with a `_1` like the different `TS0505B`'s?

PR #6306, opened 2023-10-18
fix: Disable default response on eWeLink devices

chetbox:
Replaces the `skipDefultResponse` device option - which does not seem to be used for any other devices - with the `disableDefaultResponse` meta option


chetbox:
Ah I see, I misunderstood what this option was for. After testing it for a few days I can confirm it didn't fix what I was hoping it would fix: The device still seems to occasionally not respond.

I'll look into this further and create an issue.

By the way, thanks to @matt-oakes for his help with this.

PR #6311, opened 2023-10-19
Update fingerprints for Vesternet Zigbee devices

martyn-vesternet:
Added newer fingerprints to match recent firmware version changes in Vesternet Zigbee devices.


Koenkk:
Thanks!

PR #6319, opened 2023-10-21
fix: PSBZS A1 - reset frost lock

zinserjan:
resolves koenkk/zigbee2mqtt#19320

Background: I made a mistake converting my external converter to the new tuya format and used a number instead of enum.


Koenkk:
thanks!

PR #6321, opened 2023-10-21
feat: Add rotation data to Philips Hue Tap Dial

zansko:
* Exposed extra information from `Philips Hue Tap Dial` switch. This enables more precise automation controls as it exposes the data about the rotation.
* New attributes exposed: `time`, `direction`, `type`
* Updated documentation -> https://github.com/Koenkk/zigbee2mqtt.io/pull/2281


src/lib/philips.ts
```diff
@@ -417,17 +417,20 @@ export const philipsFz = {
...
             if (button === 'dial') {
+                const typeLookup: KeyValue = {1: 'step', 2: 'rotate'};
+                const type = typeLookup[msg.data['type']];
                 const adjustedTime = direction === 'right' ? time : 256 - time;
                 const dialType = 'rotate';
                 const speed = adjustedTime <= 25 ? 'step' : adjustedTime <= 75 ? 'slow' : 'fast';
                 payload.action = `${button}_${dialType}_${direction}_${speed}`;
+                payload.time = adjustedTime;
+                payload.direction = direction;
+                payload.type = type;
```
Koenkk, line 433: Please prefix these with `action`, e.g. `action_direction`

src/lib/philips.ts
```diff
@@ -417,17 +417,20 @@ export const philipsFz = {
...
             const time = msg.data['time'];
             const payload: KeyValue = {};
 
             if (button === 'dial') {
+                const typeLookup: KeyValue = {1: 'step', 2: 'rotate'};
+                const type = typeLookup[msg.data['type']];
                 const adjustedTime = direction === 'right' ? time : 256 - time;
                 const dialType = 'rotate';
                 const speed = adjustedTime <= 25 ? 'step' : adjustedTime <= 75 ? 'slow' : 'fast';
                 payload.action = `${button}_${dialType}_${direction}_${speed}`;
```
Koenkk, line 430: Instead of adding `type`, isn't it more logical to add it here instead of `dialType`?

zansko: I would say no. `dialType` is hardcoded to `rotate`, while the `type` that button exposes is an info whether is a first message in the action of turning the dial or consequent. Since `action` already interprets `time` value to a enum `dialType` provides only a nicely readable string to the end user, no extra information.

I want to achieve with this PR 2 things:
* Expose the raw data that the dial provides so I can use them in custom functions in node red
* Keep the backwards compatibility so I don't break existing implementations of other people (that means not changing `action`)

The button exposes those 3 attributes and I'm just mapping them to same name. Please let me know if you want further changes on the naming.

src/lib/exposes.ts
```diff
@@ -615,6 +615,9 @@ export const presets = {
     // Specific
     ac_frequency: () => new Numeric('ac_frequency', access.STATE).withLabel('AC frequency').withUnit('Hz').withDescription('Measured electrical AC frequency'),
     action: (values: string[]) => new Enum('action', access.STATE, values).withDescription('Triggered action (e.g. a button click)'),
+    action_type: (values: string[]) => new Enum('action_type', access.STATE, values).withDescription('Type of the rotation, value in the first message is `step` and in the next messages value is `rotate`'),
```
Koenkk, line 618: It is not needed to put them here as these are not generic, so in the definition `exposes` use e.g. 

```js
e.enum('action_type', access.STATE, ['step', 'rotate']).withDescription('Type of the rotation, value in the first message is `step` and in the next messages value is `rotate`')
```

zansko: Taken care of

Koenkk:
Thanks!

PR #6324, opened 2023-10-22
feat: add support for lifud led driver lf-gaz150a6250-24

cryptobench:
This PR adds support for the Lifud Zigbee dimable LED driver lf-gaz150a6250-24.


Koenkk:
Thanks!

PR #6325, opened 2023-10-22
Support for futurehome thermostat (TS0601_TZE204_e5hpkc6d)

Reinachan:
This PR adds basic support for the Futurehome thermostat based on the Tuya TS0601 chip.

## Supports

Based on my current findings, I've been able to add support for

### Setting

- target temperature
- temperature calibration
- child lock
- window detection
- turning it on/off (using system mode)
  - this is equivalent to holding the on/off button on the device, turning it entirely off
- presets
  - user
    - is equivalent to 'heat' in the futurehome app, however, this preset is used whenever the user sets a custom target temperature, so 'user' feels more accurate. It also restores the previously used user-overriden value if you select this preset
    - does not get a default icon in Home Assistant, however, "heat" does
  - home
    - gets default icon in Home Assistant
  - away
    - is equivalent to 'eco away' in the futurehome app, however, the eco part feels pointless imo. 
    - gets default icon in Home Assistant when using 'away' but not for 'eco away'
  - auto
    - does not get a default icon in Home Assistant

### Reading

- room temperature
- floor temperature
- current running state of the heating (idle or heat)
- the temperature sensor being used. Can only be read, not set.

## Investigation

I know that most of the device settings are missing. I've picked up on some more values, but haven't quite figured out what they mean yet. I'll probably make new PRs in the future if I can figure them out.

#### Implemented

- 1
- 2
- 16
- 24
- 28
- 30
- 101
- 102
- 103
- 104
- 106
- 107

#### Unimplemented

These were not in the original implementation, however, the device reports values for them. I don't know what they are.

- 120 (number)
- 121 (number)
  -  Has values like `[0,0,0,228]`, `[0,0,0,114]`, `[0,0,0,237]`, `[0,0,0,239]`. Changes on its own. Maybe power draw?
- 122 (number)
- 123 (number)

### Known Missing Functionality

- Power draw

#### Advanced Settings

Some of these may or may not be possible to configure through Zigbee

[source](https://www.futurehome.io/en/shop/thermostat-black) (navigate to "attachments" tab and download the "user manual" pdf)

- Floor sensor type / external
  - 0: 10 kOhm | 1: 12 kOhm | 2: 15 kOhm | 3: 47 kOhm
- Max guarding temperature
  - Manual says this has to be set on the device itself, so might not be possible to implement
- Turn - 1-9°C, Standard is 2 ° C
- Calibration: -9 to + 9 ° C, Standard is 0 ° C
- Highest set point: 35-90 ° C, Standard is 35 °C
- Lowest setpoint: 0-30 ° C, Standard is 5 °C
- Home mode setpoint: The standard is 22 ° C
- Setpoint for away mode: The default is 15 ° C
- Brightness in sleep mode: 1-10, default is 1
- Open window detection: 0 = Disabled / 1 = Active

## Additional Notes

I've reached out to the manufacturer about this device, and they've said they'll try to provide me with the protocol used. When I get that, I'll try to implement the missing stuff and correct anything I've potentially done wrong in a new PR 

Feel free to let me know if I need to make any changes


Reinachan:
The failing test doesn't seem related to my work. I did  `git checkout 512a595` (latest commit not done by me) and ran `npm test` there and got the same error

src/devices/futurehome.ts
```diff
@@ -0,0 +1,96 @@
...
+                }),
+            ],
+            [16, 'current_heating_setpoint', tuya.valueConverter.raw],
+            [24, 'local_temperature', tuya.valueConverter.raw],
+            [101, 'local_temperature_floor', tuya.valueConverter.raw],
+            [
+                102,
+                'temperature_sensor',
+                tuya.valueConverterBasic.lookup({
+                    'air sensor': tuya.enum(0),
```
Koenkk, line 79: Please make this snake case, e.g. `air_sensor`

src/devices/futurehome.ts
```diff
@@ -0,0 +1,96 @@
+import * as exposes from '../lib/exposes';
+import * as tuya from '../lib/tuya';
+
+const e = exposes.presets;
+const ea = exposes.access;
+
+const definition = {
```
Koenkk, line 7: Please change this to https://github.com/Koenkk/zigbee-herdsman-converters/blob/d69dc7a13fbf9254219c5cfda84e50f5629bf1d0/src/devices/tuya.ts#L514

(this is what makes the build fail)

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/2282

src/devices/futurehome.ts
Reinachan: I assume this goes for both the value converter here and the enum in the `exposes` section? If so, this should be resolved :)

src/devices/futurehome.ts
Reinachan: I think this is resolved

src/devices/futurehome.ts
```diff
@@ -0,0 +1,51 @@
+import {Definition} from 'src/lib/types';
+import * as exposes from '../lib/exposes';
+import * as tuya from '../lib/tuya';
+
+const e = exposes.presets;
+const ea = exposes.access;
+
+const definitions: Definition[] = [
+    {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE204_e5hpkc6d']),
```
Koenkk, line 10: It looks this device is already supported:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/d69dc7a13fbf9254219c5cfda84e50f5629bf1d0/src/devices/connecte.ts#L10

Reinachan: It seems like support was added last week when I look at the blame. I started work on this a couple months ago and back then there wasn't support for it at all. I'll try updating my zigbee2mqtt container and giving it a try assuming it's been updated with those changes yet.

That said, there is an argument to be made about approach here. The other implementation is using what is labeled as "legacy" APIs. Is that the best approach? While I haven't tested it yet, I'm also unsure what I feel about the MQTT interface of the other approach. I'll report back when I've been able to try it myself.

If I find my approach to be better in my own opinion, and you agree, I can implement what I missed based on the existing implementation and submit this as an improvement rather than adding support for something new. I'll get back with proper arguments for that if so.

Koenkk: I definitely like your approach, if you could update the existing converter that would be great!

Reinachan: Sounds good! I'll try to get around to it sometime this week then!

One question regarding this though. The manufacturer name. I looked into this a bit. Futurehome and Connecte actually merged back in 2022 and it seems that the devices are now sold under the Futurehome brand instead. Should I change the manufacturer name to Futurehome or keep it as Connecte like the old approach did?

Koenkk: I would show the vendor name whatever is on the product/box

Reinachan:
Sorry it took some time to get back to this. Was busy.

Anyway, I've gotten most of the way there, however,  there's a couple issues remaining that I can't figure out how to fix. For some reason, I can't get the `hysteresis`, `max_temperature_protection`, and `sensor` endpoints to work bi-directionally, despite how they work with the original implementation. Is there something I'm missing?

@Koenkk

Reinachan:
Once that's fixed, I just need to remove dead code and then this should be feature-par with some improvements over the current implementation.

Koenkk:
They need to be added to https://github.com/Koenkk/zigbee-herdsman-converters/blob/aec276094383f6e8b067e189ba7650124d855120/src/lib/tuya.ts#L857

Reinachan:
@Koenkk  I think the rewrite is complete now.  I'll make a separate PR later if I figure out what the other unimplemented datapoints are. I haven't gotten a response from the manufacturer yet, but they said they're going to send me the protocol eventually :) 

Some things to note

1. I only have the `_TZE204_e5hpkc6d` model to test on, so I can't verify if this works on the other variants of this device or not. I'm assuming they work the same, but I can't guarantee it.
2. I deleted the file with the old approach and moved this all over to `futurehome.ts` as that seems to be their new branding and is what they use in their Zigbee device certification now.
3. I haven't tested that the changes you mentioned in your previous comment actually works on the device. I'd have to mess with my config to be able to test that in action, which I'd prefer not to. I know everything else works on my device.

I have no idea what this actually does, but the old code had it, so I included it.
```ts
whiteLabel: [
    tuya.whitelabel('Futurehome', 'Co020', 'Smart thermostat', ['_TZE200_e5hpkc6d']),
],
```

How this differs from the old approach (ergo, the breaking changes from a user perspective):

<table>
<tr>
 <th>
 <th>Old approach
 <th>New approach
<tr>
 <td>on/off state
 <td>is its own binary control
 <td>is controlled by climate system mode
<tr>
 <td>device presets
 <td>spread across various different controls despite being an enum where only one can be true
 <td>is just climate presets like how the device works and how it works in Futurehome
<tr>
 <td>"heat" preset
 <td>is named "heat" despite how it functions as the user-overidden temperature target 
 <td>renamed to "user" as a preset instead of system mode
<tr>
 <td>temperature sensor
 <td>named "internal", "external", "both"
 <td>renamed to "room sensor", "floor sensor", "max guard" based on what the manual calls them
</table>

Reinachan:
Btw, off-topic, but having read through a lot of the source code now, I have some stuff I'd like to just mention.

```ts
// Bad practice
some_converter: {
  // ...
} as Fz.Converter

// Instead you should do
some_converter: {
  // ...
} satisfies Fz.Converter
```
The `as` keyword asserts that you know better than TypeScript what the type actually is, so TypeScript will skip typechecking it and just assume you know what you're doing. `satisfies` will typecheck it and let you know if it's incompatible with the type.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html

Other than that, adhering to your codestyle is kinda hard without something like a prettierconfig. I turned off my prettier autoformatting as it heavily conflicted with your preferences. I'd suggest adding a prettier config.

Koenkk:
Thanks!

- I didn't know about the `satisfies` yet, could you make a PR to use this in the whole code base?
- Regarding code style, currently eslint (I know prettier does much more), but I think it's good enough for now.

Reinachan:
> I didn't know about the satisfies yet, could you make a PR to use this in the whole code base?
@Koenkk 

I'll see if I get around to it sometime. It might be a big undertaking as I may have to fix a lot of stuff alongside it, which is hard without extensive knowledge of the project as a whole. But I'll have a look at it at least when I have time for it. Could be easier than I think :p

Thanks for the help in getting this merged <3

3lB4rto:
Hi, I see this has some time already closed, but after searching a lot it is the only place that has details about this futurehome thermostat. 

I tried the original futurehome hub for a few days and could see the power usage, so i am wondering how to get it implemented into Z2M.

@Reinachan I am wondering if there is any chance to implement the power draw? It seems like the Datapoint 121 you mentioned in the beginning reflects some kind of possible value.

I have 3 of these thermostat, if there is anything i can try to help, please let me know :)

PR #6326, opened 2023-10-22
send default response in tuya_multi_action as well

konistehrad:
`tuya_multi_action` also needs to send a default response using the sequence ID to prevent second-click-delays,  exactly like `tuya_on_off_action`. (It's parsing and replying to the same messages on the same endpoint)


konistehrad:
Tested this on both a [ERS-10TZBVK-AA](https://www.zigbee2mqtt.io/devices/ERS-10TZBVK-AA.html#tuya-ers-10tzbvk-aa) and a [ERS-10TZBVB-AA](https://www.zigbee2mqtt.io/devices/ERS-10TZBVB-AA.html#tuya-ers-10tzbvb-aa)

Koenkk:
Thanks!

PR #6333, opened 2023-10-24
Legrand 067776(A): Support calibration + Venetian mode

FabianMangold:
This PR:
- includes a small refactoring of Legrand related devices / libraries
- enables calibration on 067776 / 067776A devices
- adds support for Venetian (BSO) mode

@Koenkk: A couple of questions / remarks:
1.) The reporting does not work reliably yet. I am somewhat puzzled with the reporting of attributes. Can you give me pointers as how to fix this or where to look for potential issues ?
- The device seems to report attributes regardless of the configured reporting. (I assume there is not much we can do about that, unless we explicitly disregard those reports)
- While the manufacturer code is honoured for readAttribute requests, it is ignored for attribute reports.
As an example, this is an excerpt from database.db for one of the devices:
```
{
  "61444": 25,
  "currentPositionLiftPercentage": 100,
  "currentPositionTiltPercentage": 0,
  "tuyaMotorReversal": 4,
  "moesCalibrationTime": 0,
  "tuyaMovingState": 100,
  "tuyaCalibration" :1
  }
```
Attribute 61444 (0xF004) should read "calibrationMode" following [zigbee-herdsman PR#784](https://github.com/Koenkk/zigbee-herdsman/pull/784), which as mentioned, works for readAttributes but not for reports.

2.) In order to avoid code duplication, i wanted to merge 067776 / 067776A devices in the device definition.
Shouldn't the following work ?
```
  {
        zigbeeModel: [' Shutter switch with neutral\u0000\u0000\u0000'],
        model: '067776',
        vendor: 'Legrand',
        description: 'Netatmo wired shutter switch',
        whiteLabel: [
            {model: '067776A',
            vendor: 'Legrand',
            description: 'Netatmo wired shutter switch with level control (NLLV)',
            meta: {supportsLevelDetection: false},
            fingerprint: [{modelID: ' Shutter SW with level control\u0000', manufacturerID: 4129}]},
        ],
        ota: ota.zigbeeOTA,
        meta: {supportsLevelDetection: false},
```
Thanks a lot.

P.S: I'll take a look at the failing test shortly. Looks like:
- ~my new Meta definition~
- ~and including a function in "exposes"~
causes troubles.

### DIFF
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -6,7 +6,7 @@ import tz from '../converters/toZigbee';
 import * as reporting from '../lib/reporting';
 import extend from '../lib/extend';
 import * as ota from '../lib/ota';
-import {tzLegrand, fzLegrand, readInitialBatteryState} from '../lib/legrand';
+import {tzLegrand, fzLegrand, readInitialBatteryState, _067776} from '../lib/legrand';
 const e = exposes.presets;
 const ea = exposes.access;
 
@@ -133,20 +133,23 @@ const definitions: Definition[] = [
         vendor: 'Legrand',
         description: 'Netatmo wired shutter switch',
         ota: ota.zigbeeOTA,
-        fromZigbee: [fz.ignore_basic_report, fz.cover_position_tilt, fz.legrand_binary_input_moving, fz.identify, fz.legrand_led_in_dark],
-        toZigbee: [tz.cover_state, tz.cover_position_tilt, tz.legrand_identify, tz.legrand_settingEnableLedInDark],
+        fromZigbee: [fz.ignore_basic_report, fz.cover_position_tilt, fz.legrand_binary_input_moving, fz.identify,
+            fz.legrand_led_in_dark, fzLegrand.calibration_mode(false)],
+        toZigbee: [tz.cover_state, tz.cover_position_tilt, tz.legrand_identify, tz.legrand_settingEnableLedInDark, tzLegrand.calibration_mode(false)],
         exposes: [
-            e.cover_position(),
+            _067776.getCover(),
             e.action(['moving', 'identify']),
             e.enum('identify', ea.SET, ['blink'])
                 .withDescription('Blinks the built-in LED to make it easier to identify the device'),
             e.binary('led_in_dark', ea.ALL, 'ON', 'OFF')
                 .withDescription('Enables the built-in LED allowing to see the switch in the dark'),
+            _067776.getCalibrationModes(false),
         ],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genBinaryInput', 'closuresWindowCovering', 'genIdentify']);
-            await reporting.currentPositionLiftPercentage(endpoint);
+            await reporting.currentPositionLiftPercentage(endpoint, {max: 120});
+            await reporting.currentPositionTiltPercentage(endpoint, {max: 120});
         },
     },
     {
@@ -183,20 +186,23 @@ const definitions: Definition[] = [
         vendor: 'Legrand',
         description: 'Netatmo wired shutter switch with level control (NLLV)',
         ota: ota.zigbeeOTA,
-        fromZigbee: [fz.ignore_basic_report, fz.cover_position_tilt, fz.legrand_binary_input_moving, fz.identify, fz.legrand_led_in_dark],
-        toZigbee: [tz.cover_state, tz.cover_position_tilt, tz.legrand_identify, tz.legrand_settingEnableLedInDark],
+        fromZigbee: [fz.ignore_basic_report, fz.cover_position_tilt, fz.legrand_binary_input_moving, fz.identify,
+            fz.legrand_led_in_dark, fzLegrand.calibration_mode(true)],
+        toZigbee: [tz.cover_state, tz.cover_position_tilt, tz.legrand_identify, tz.legrand_settingEnableLedInDark, tzLegrand.calibration_mode(true)],
         exposes: [
-            e.cover_position(),
+            _067776.getCover(),
             e.action(['moving', 'identify']),
             e.enum('identify', ea.SET, ['blink'])
                 .withDescription('Blinks the built-in LED to make it easier to identify the device'),
             e.binary('led_in_dark', ea.ALL, 'ON', 'OFF')
                 .withDescription('Enables the built-in LED allowing to see the switch in the dark'),
+            _067776.getCalibrationModes(true),
         ],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genBinaryInput', 'closuresWindowCovering', 'genIdentify']);
-            await reporting.currentPositionLiftPercentage(endpoint);
+            await reporting.currentPositionLiftPercentage(endpoint, {max: 120});
+            await reporting.currentPositionTiltPercentage(endpoint, {max: 120});
         },
     },
     {
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -148,8 +148,11 @@ const definitions: Definition[] = [
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genBinaryInput', 'closuresWindowCovering', 'genIdentify']);
-            await reporting.currentPositionLiftPercentage(endpoint, {max: 120});
-            await reporting.currentPositionTiltPercentage(endpoint, {max: 120});
+            let p = reporting.payload('currentPositionLiftPercentage', 1, 120, 1);
+            await endpoint.configureReporting('closuresWindowCovering', p, {manufacturerCode: 4129});
+
+            p = reporting.payload('currentPositionTiltPercentage', 1, 120, 1);
+            await endpoint.configureReporting('closuresWindowCovering', p, {manufacturerCode: 4129});
         },
     },
     {
@@ -201,8 +204,11 @@ const definitions: Definition[] = [
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genBinaryInput', 'closuresWindowCovering', 'genIdentify']);
-            await reporting.currentPositionLiftPercentage(endpoint, {max: 120});
-            await reporting.currentPositionTiltPercentage(endpoint, {max: 120});
+            let p = reporting.payload('currentPositionLiftPercentage', 1, 120, 1);
+            await endpoint.configureReporting('closuresWindowCovering', p, {manufacturerCode: 4129});
+
+            p = reporting.payload('currentPositionTiltPercentage', 1, 120, 1);
+            await endpoint.configureReporting('closuresWindowCovering', p, {manufacturerCode: 4129});
         },
     },
     {
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -1,51 +1,14 @@
-import {Definition, Fz, OnEvent, Tz} from '../lib/types';
+import {Definition} from '../lib/types';
 import * as exposes from '../lib/exposes';
 import fz from '../converters/fromZigbee';
 import * as legacy from '../lib/legacy';
 import tz from '../converters/toZigbee';
 import * as reporting from '../lib/reporting';
 import extend from '../lib/extend';
-import * as utils from '../lib/utils';
 import * as ota from '../lib/ota';
+import {tzLegrand, fzLegrand, readInitialBatteryState} from '../lib/legrand';
 const e = exposes.presets;
 const ea = exposes.access;
-
-const readInitialBatteryState: OnEvent = async (type, data, device, options) => {
-    if (['deviceAnnounce'].includes(type)) {
-        const endpoint = device.getEndpoint(1);
-        const options = {manufacturerCode: 0x1021, disableDefaultResponse: true};
-        await endpoint.read('genPowerCfg', ['batteryVoltage'], options);
-    }
-};
-
-const tzLocal = {
-    auto_mode: {
-        key: ['auto_mode'],
-        convertSet: async (entity, key, value, meta) => {
-            const mode = utils.getFromLookup(value, {'off': 0x00, 'auto': 0x02, 'on_override': 0x03});
-            const payload = {data: Buffer.from([mode])};
-            await entity.command('manuSpecificLegrandDevices3', 'command0', payload);
-            return {state: {'auto_mode': value}};
-        },
-    } as Tz.Converter,
-};
-
-const fzlocal = {
-    legrand_600087l: {
-        cluster: 'greenPower',
-        type: ['commandNotification'],
-        convert: (model, msg, publish, options, meta) => {
-            const commandID = msg.data.commandID;
-            const lookup: {[s: number]: string} = {0x34: 'stop', 0x35: 'up', 0x36: 'down'};
-            if (commandID === 224) return;
-            if (!lookup.hasOwnProperty(commandID)) {
-                meta.logger.error(`GreenPower_3 error: missing command '${commandID}'`);
-            } else {
-                return {action: lookup[commandID]};
-            }
-        },
-    } as Fz.Converter,
-};
 
 const definitions: Definition[] = [
     {
@@ -100,7 +63,7 @@ const definitions: Definition[] = [
         extend: extend.switch(),
         ota: ota.zigbeeOTA,
         fromZigbee: [fz.identify, fz.on_off, fz.electrical_measurement, fz.legrand_cluster_fc01, fz.ignore_basic_report, fz.ignore_genOta],
-        toZigbee: [tz.legrand_deviceMode, tz.on_off, tz.legrand_identify, tz.electrical_measurement_power, tzLocal.auto_mode],
+        toZigbee: [tz.legrand_deviceMode, tz.on_off, tz.legrand_identify, tz.electrical_measurement_power, tzLegrand.auto_mode],
         exposes: [
             e.switch().withState('state', true, 'On/off (works only if device is in "switch" mode)'),
             e.power().withAccess(ea.STATE_GET),
@@ -538,7 +501,7 @@ const definitions: Definition[] = [
         model: '600087L',
         vendor: 'Legrand',
         description: 'Wireless and batteryless blind control switch',
-        fromZigbee: [fzlocal.legrand_600087l],
+        fromZigbee: [fzLegrand.legrand_600087l],
         toZigbee: [],
         exposes: [e.action(['stop', 'up', 'down'])],
     },
--- a/src/lib/legrand.ts
+++ b/src/lib/legrand.ts
@@ -7,11 +7,11 @@ const ea = exposes.access;
 const legrandOptions = {manufacturerCode: 4129, disableDefaultResponse: true};
 
 const shutterCalibrationModes: {[k: number]: {description: string, onlyNLLV: boolean}} = {
-    0: {description: 'Classic (NLLV)', onlyNLLV: true},
-    1: {description: 'Specific (NLLV)', onlyNLLV: true},
-    2: {description: 'Up/Down/Stop', onlyNLLV: false},
-    3: {description: 'Temporal', onlyNLLV: false},
-    4: {description: 'Venetian (BSO)', onlyNLLV: false},
+    0: {description: 'classic_nllv', onlyNLLV: true},
+    1: {description: 'specific_nllv', onlyNLLV: true},
+    2: {description: 'up_down_stop', onlyNLLV: false},
+    3: {description: 'temporal', onlyNLLV: false},
+    4: {description: 'venetian_bso', onlyNLLV: false},
 };
 
 const getApplicableCalibrationModes = (isNLLVSwitch: boolean): KeyValueString => {
--- a/src/lib/legrand.ts
+++ b/src/lib/legrand.ts
@@ -1,4 +1,4 @@
-import {Fz, Tz, OnEvent} from '../lib/types';
+import {Fz, Tz, OnEvent, KeyValueString} from '../lib/types';
 import * as exposes from './exposes';
 import * as utils from '../lib/utils';
 const e = exposes.presets;
@@ -6,18 +6,18 @@ const ea = exposes.access;
 
 const legrandOptions = {manufacturerCode: 4129, disableDefaultResponse: true};
 
-const shutterCalibrationModes = {
-    'Classic (NLLV)': {ID: 0, onlyNLLV: true},
-    'Specific (NLLV)': {ID: 1, onlyNLLV: true},
-    'Up/Down/Stop': {ID: 2, onlyNLLV: false},
-    'Temporal': {ID: 3, onlyNLLV: false},
-    'Venetian (BSO)': {ID: 4, onlyNLLV: false},
+const shutterCalibrationModes: {[k: number]: {description: string, onlyNLLV: boolean}} = {
+    0: {description: 'Classic (NLLV)', onlyNLLV: true},
+    1: {description: 'Specific (NLLV)', onlyNLLV: true},
+    2: {description: 'Up/Down/Stop', onlyNLLV: false},
+    3: {description: 'Temporal', onlyNLLV: false},
+    4: {description: 'Venetian (BSO)', onlyNLLV: false},
 };
 
-const getApplicableCalibrationModes = (isNLLVSwitch: boolean) => {
+const getApplicableCalibrationModes = (isNLLVSwitch: boolean): KeyValueString => {
     return Object.fromEntries(Object.entries(shutterCalibrationModes)
         .filter((e) => isNLLVSwitch ? true : e[1].onlyNLLV === false)
-        .map((e) => [e[0], e[1].ID]));
+        .map((e) => [e[0], e[1].description]));
 };
 
 export const readInitialBatteryState: OnEvent = async (type, data, device, options) => {
@@ -43,8 +43,8 @@ export const tzLegrand = {
             key: ['calibration_mode'],
             convertSet: async (entity, key, value, meta) => {
                 const applicableModes = getApplicableCalibrationModes(isNLLVSwitch);
-                utils.validateValue(value, Object.keys(applicableModes));
-                const idx = applicableModes[value as string];
+                utils.validateValue(value, Object.values(applicableModes));
+                const idx = utils.getKey(applicableModes, value);
                 await entity.write('closuresWindowCovering', {'calibrationMode': idx}, legrandOptions);
             },
             convertGet: async (entity, key, meta) => {
@@ -76,10 +76,10 @@ export const fzLegrand = {
             convert: (model, msg, publish, options, meta) => {
                 const attr = 'calibrationMode';
                 if (msg.data.hasOwnProperty(attr)) {
-                    const idx = msg.data[attr];
                     const applicableModes = getApplicableCalibrationModes(isNLLVSwitch);
-                    utils.validateValue(idx, Object.values(applicableModes));
-                    const calMode = utils.getKey(applicableModes, idx);
+                    const idx = msg.data[attr];
+                    utils.validateValue(String(idx), Object.keys(applicableModes));
+                    const calMode = applicableModes[idx];
                     return {calibration_mode: calMode};
                 }
             },
@@ -106,7 +106,7 @@ export const _067776 = {
     },
     getCalibrationModes: (isNLLVSwitch: boolean) => {
         const modes = getApplicableCalibrationModes(isNLLVSwitch);
-        return e.enum('calibration_mode', ea.ALL, Object.keys(modes))
+        return e.enum('calibration_mode', ea.ALL, Object.values(modes))
             .withDescription('Defines the calibration mode of the switch. (Caution: Changing modes requires a recalibration of the shutter switch!)');
     },
 };
--- a/src/lib/legrand.ts
+++ b/src/lib/legrand.ts
@@ -1,5 +1,22 @@
-import {Fz, OnEvent, Tz} from '../lib/types';
+import {Fz, Tz, OnEvent} from '../lib/types';
+import * as exposes from './exposes';
 import * as utils from '../lib/utils';
+const e = exposes.presets;
+const ea = exposes.access;
+
+const shutterCalibrationModes = {
+    'Classic (NLLV)': {ID: 0, onlyNLLV: true},
+    'Specific (NLLV)': {ID: 1, onlyNLLV: true},
+    'Up/Down/Stop': {ID: 2, onlyNLLV: false},
+    'Temporal': {ID: 3, onlyNLLV: false},
+    'Venetian (BSO)': {ID: 4, onlyNLLV: false},
+};
+
+const getApplicableCalibrationModes = (isNLLVSwitch: boolean) => {
+    return Object.fromEntries(Object.entries(shutterCalibrationModes)
+        .filter((e) => isNLLVSwitch ? true : e[1].onlyNLLV === false)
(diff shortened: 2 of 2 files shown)


src/devices/legrand.ts
```diff
@@ -170,20 +133,27 @@ const definitions: Definition[] = [
...
-            e.enum('identify', ea.SET, ['blink'])
-                .withDescription('Blinks the built-in LED to make it easier to identify the device'),
-            e.binary('led_in_dark', ea.ALL, 'ON', 'OFF')
-                .withDescription('Enables the built-in LED allowing to see the switch in the dark'),
-        ],
+        meta: {supportsLevelDetection: false},
+        fromZigbee: [fz.ignore_basic_report, fz.cover_position_tilt, fz.legrand_binary_input_moving, fz.identify,
+            fz.legrand_led_in_dark, fzLegrand.calibration_mode],
+        toZigbee: [tz.cover_state, tz.cover_position_tilt, tz.legrand_identify, tz.legrand_settingEnableLedInDark, tzLegrand.calibration_mode],
+        exposes: function() {
```
Koenkk, line 140: Why do we need a function here?

Koenkk:
> 1.) The reporting does not work reliably yet. I am somewhat puzzled with the reporting of attributes. Can you give me pointers as how to fix this or where to look for potential issues ?

Can you try to include the manufacturer code when configuring the attribute reporting?

> While the manufacturer code is honoured for readAttribute requests, it is ignored for attribute reports.
As an example, this is an excerpt from database.db for one of the devices:

Can you provide the herdsman debug log when this is reported?

FabianMangold:
Just to clarify the previous question (which you probably edited), the function in exposes is required due to the self reference to meta (i.e. _067776.getCalibrationModes(this.meta)). When not wrapped in a function, the scope is different, a "this" undefined.

- Yes, I had done that, by using "endpoint.configureReporting", the results are identical.
- I will check the logs ASAP (sorry currently on a trip)
- I believe the meta is actually required, as the 067776 is not an NLLV switch, while the 067776A is. Therefore, the newly introduced "supportsLevelDetection" meta attribute is required. 
Regardless, even if i get rid of the meta part, the switch gets detected as a bticino switch. Haven't debugged it, but does a device description have precendence over a whitelabel with a proper fingerprint ?

FabianMangold:
Regarding reporting:

I've updated the code, and added the manufacturer code back to the reporting.
Database.db shows that the reporting is configured accordingly (the manufacturer code is not visible in the FrontEnd):

```
[
  {
    "cluster": 258,
    "attrId": 8,
    "minRepIntval": 1,
    "maxRepIntval": 120,
    "repChange": 1,
    "manufacturerCode": 4129
  },
  {
    "cluster": 258,
    "attrId": 9,
    "minRepIntval": 1,
    "maxRepIntval": 120,
    "repChange": 1,
    "manufacturerCode": 4129
  },
  {
    "cluster": 258,
    "attrId": 61442,
    "minRepIntval": 1,
    "maxRepIntval": 120,
    "repChange": 1,
    "manufacturerCode": 4129
  }
]
```
The reported attributes remain "generic":
```
{
  "61444": 25,
  "currentPositionLiftPercentage": 100,
  "currentPositionTiltPercentage": 0,
  "tuyaMotorReversal": 4,
  "tuyaMovingState": 100,
  "tuyaCalibration": 1,
  "moesCalibrationTime": 0
}
```
I'll check the herdsman debug logs ASAP

FabianMangold:
I've isolated 2 frames that depict the situation pretty well.

The first is the reply to the readrequest that was requested via the "calibration_mode" TZ converter (specifying the manufacturer code):
`zigbee-herdsman:controller:log Received 'zcl' data '{"frame":{"Header":{"frameControl":{"frameType":0,"manufacturerSpecific":true,"direction":1,"disableDefaultResponse":true,"reservedBits":0},"transactionSequenceNumber":14,"manufacturerCode":4129,"commandIdentifier":1},"Payload":[{"attrId":61442,"status":0,"dataType":48,"attrData":4}],"Command":{"ID":1,"name":"readRsp","parameters":[{"name":"attrId","type":33},{"name":"status","type":32},{"name":"dataType","type":32,"conditions":[{"type":"statusEquals","value":0}]},{"name":"attrData","type":1000,"conditions":[{"type":"statusEquals","value":0}]}]}},"address":7729,"endpoint":1,"linkquality":100,"groupID":0,"wasBroadcast":false,"destinationEndpoint":1}' +169ms`

The second is a simple attribute report (command 10) that happens to report the same attribute (61442)
` zigbee-herdsman:controller:log Received 'zcl' data '{"frame":{"Header":{"frameControl":{"frameType":0,"manufacturerSpecific":false,"direction":1,"disableDefaultResponse":true,"reservedBits":0},"transactionSequenceNumber":200,"manufacturerCode":null,"commandIdentifier":10},"Payload":[{"attrId":61442,"dataType":48,"attrData":4}],"Command":{"ID":10,"name":"report","parameters":[{"name":"attrId","type":33},{"name":"dataType","type":32},{"name":"attrData","type":1000}]}},"address":39958,"endpoint":1,"linkquality":184,"groupID":0,"wasBroadcast":false,"destinationEndpoint":1}' +947ms`

Koenkk:
> but does a device description have precendence over a whitelabel with a proper fingerprint ?

The white label also contains a description, this will be used if the fingerprint matches


> As can bee seen, those attributes were never requested and are "generic", not Manufacturer specific.
Here is another one that matches my example:
Received Zigbee message from 'XXX, type 'attributeReport', cluster 'closuresWindowCovering', data '{"tuyaMotorReversal":4}' from endpoint 1 with groupID 0
This attribute was requested, but should read "calibrationMode"

I see the issue, I will check this later, or is this blocking this pr from being merged?

FabianMangold:
Please disregard my comment about the whitelabel, and merging the devices.
Albeit being very similar, both devices are based on different firmware code base (NLV vs NLLV), and behave slightly differently, Therefore, keeping them distinct is probably way better, if you agree on that.

No please merge, this is not blocking. The only feature that doesn't work (yet) is the reporting of the blind position when the switch is in "Venetian Mode". Other than that, everything is OK and unchanged.

I may have a patch for ZH regarding the reporting issue, i simply need a but more time.... unless you can fix it faster :)
Thanks a lot.

FabianMangold:
Hi @Koenkk,
A potential fix is available here: https://github.com/Koenkk/zigbee-herdsman/pull/787

test/index.test.js
```diff
@@ -313,7 +313,7 @@ describe('index.js', () => {
 
 
             if (device.meta) {
-                containsOnly(['disableActionGroup', 'multiEndpoint', 'multiEndpointSkip', 'multiEndpointEnforce', 'applyRedFix', 'disableDefaultResponse', 'supportsEnhancedHue', 'timeout', 'supportsHueAndSaturation', 'battery', 'coverInverted', 'turnsOffAtBrightness1', 'coverStateFromTilt', 'pinCodeCount', 'tuyaThermostatSystemMode', 'tuyaThermostatPreset', 'tuyaDatapoints', 'tuyaThermostatPresetToSystemMode', 'thermostat', 'separateWhite', 'publishDuplicateTransaction', 'tuyaSendCommand'], Object.keys(device.meta));
+                containsOnly(['disableActionGroup', 'multiEndpoint', 'multiEndpointSkip', 'multiEndpointEnforce', 'applyRedFix', 'disableDefaultResponse', 'supportsEnhancedHue', 'timeout', 'supportsHueAndSaturation', 'battery', 'coverInverted', 'turnsOffAtBrightness1', 'coverStateFromTilt', 'pinCodeCount', 'tuyaThermostatSystemMode', 'tuyaThermostatPreset', 'tuyaDatapoints', 'tuyaThermostatPresetToSystemMode', 'thermostat', 'separateWhite', 'publishDuplicateTransaction', 'tuyaSendCommand', 'supportsLevelDetection'], Object.keys(device.meta));
```
Koenkk, line 316: If I understand correctly, `supportsLevelDetection` is Legrand specific, I propose to name it `legrandSupportsLevelDetection`

src/devices/legrand.ts
```diff
@@ -220,20 +194,31 @@ const definitions: Definition[] = [
...
+            return [
+                _067776.getCover(),
+                e.action(['moving', 'identify']),
+                e.enum('identify', ea.SET, ['blink'])
+                    .withDescription('Blinks the built-in LED to make it easier to identify the device'),
+                e.binary('led_in_dark', ea.ALL, 'ON', 'OFF')
+                    .withDescription('Enables the built-in LED allowing to see the switch in the dark'),
+                _067776.getCalibrationModes(this.meta),
```
Koenkk, line 209: Is the whole `expose()` function only needed to read from `this.meta`? If yes, I propose to drop the whole `supportsLevelDetection` and do something like this:

```js
exposes: [_067776.getCover(), e.action(['moving', 'identify']), _067776.getCalibrationModes(true)]
```

Where `true` indicates it `supportsLevelDetection`

FabianMangold:
Hi @Koenkk,
Thanks for your feedback. I've modified the code accordingly and simplified it.
Cannot completely get rid of functions, as the supported calibration modes need to be determined in the FZ /TZ.
Is this approach better ?
Cheers.

src/lib/legrand.ts
```diff
@@ -0,0 +1,112 @@
...
+const ea = exposes.access;
+
+const legrandOptions = {manufacturerCode: 4129, disableDefaultResponse: true};
+
+const shutterCalibrationModes = {
+    'Classic (NLLV)': {ID: 0, onlyNLLV: true},
+    'Specific (NLLV)': {ID: 1, onlyNLLV: true},
+    'Up/Down/Stop': {ID: 2, onlyNLLV: false},
+    'Temporal': {ID: 3, onlyNLLV: false},
+    'Venetian (BSO)': {ID: 4, onlyNLLV: false},
```
Koenkk, line 14: Last thing before this can be merged, could you make these `snake_case`? e.g. `classic_nllv`, `venetian_bso`

FabianMangold: I need those strings in the frontend, but let me change the structure somewhat. I assume you with to avoid having strings containing special chars. as keys right ?

Koenkk: I meant the values itself (the structure is fine). No special characters indeed, no spaces and only lower case, e.g. `Venetian (BSO)` should become `venetion_bso`, `Up/Down/Stop` -> `up_down_stop`.

FabianMangold: Oh okay, sure. Let me change it then.
Just figured that it would look better as "text", like so:

grafik

FabianMangold: PR updated accordingly.

grafik

Koenkk:
Thanks!

src/lib/legrand.ts
Koenkk: Not all people use the frontend, e.g. when you publish directly to MQTT, you don't want to deal with special chars.

FabianMangold: Fair point, thanks @Koenkk

syntesys87:
Could it work also with BTicino K4027C?

jd1900:
Hi @FabianMangold , thank you for the addition!
Could you add some information regarding the calibration?
I've tried looking finding more about it, but I've found nothing. When I bought them (067776 without the A) they didn't have any calibration.
How would it work?

May it be useful for being able to stop at a certain position?

FabianMangold:
Hi  @syntesys87 
> Could it work also with BTicino K4027C?
I would think so indeed, as the K4027C seems to be using the NLLV firmware. Do you have an actual BTicino device ?
If so, we could give this a try.

syntesys87:
Hi @FabianMangold, I have all already configured, for the moment I'm using all the switches with zigbee2mqtt and the covers are configured with legrand gateway... In a moment, I can switch the covers to zigbee2mqtt!

syntesys87:
@FabianMangold I've ported the changes on bticino.ts and legrand.ts, but I don't know if it's correct. This evening (I'm from Italy) I'll test it, I've only seen that the two files have exension .ts, but in node_modules seems to be .js, both with .map... Do I need to compile something or can I import straight, modifying only extension?
Thank you.

philippk75:
hello, I use the Bticino N4027C switches. Since last night's update from Zigbee2MQTT in HA, my Bticino N4027C switches have had the Legrand picture automatically replaced (which I don't like) and I'm suddenly finding all the settings with calibration etc. The calibration of the blinds also works, but also with the blinds I find the tilt angle, which makes little sense (how can I remove it), with the venetian I have not managed to perform the calibration or change the tilt angle?

Greetings

Philipp

jd1900:
> FabianMangold

Thank you [FabianMangold](https://github.com/FabianMangold), very well explained.
It would be great to add to here: https://www.zigbee2mqtt.io/devices/067776.html @Koenkk

syntesys87:
> hello, I use the Bticino N4027C switches. Since last night's update from Zigbee2MQTT in HA, my Bticino N4027C switches have had the Legrand picture automatically replaced (which I don't like) and I'm suddenly finding all the settings with calibration etc. The calibration of the blinds also works, but also with the blinds I find the tilt angle, which makes little sense (how can I remove it), with the venetian I have not managed to perform the calibration or change the tilt angle?
> 
> Greetings
> 
> Philipp

I have also the same problem, my K4027C is recognized as 067776A and calibration does not work. I also tried to calibrate, but classic_nllv is not working, with specific it starts calibrating but fails, tried with temporal but fails... The firmware seems equal but I think there are some differences... The other problem I found is that after the association with Z2Mqtt, the LED does not change color from RED or power off...

syntesys87:
This is a log extraction:
Info 2023-11-02 19:24:11MQTT publish: topic 'zigbee2mqtt/0x0004740000b79969', payload '{"action":null,"calibration_mode":"specific_nllv","led_in_dark":null,"linkquality":128,"position":100,"state":"CLOSE","tilt":0,"update":{"installed_version":3556351,"latest_version":3556351,"state":"idle"},"update_available":null}'
Info 2023-11-02 19:24:12MQTT publish: topic 'zigbee2mqtt/0x0004740000b79969', payload '{"action":null,"calibration_mode":"specific_nllv","led_in_dark":null,"linkquality":116,"position":50,"state":"OPEN","tilt":0,"update":{"installed_version":3556351,"latest_version":3556351,"state":"idle"},"update_available":null}'
Info 2023-11-02 19:24:20MQTT publish: topic 'zigbee2mqtt/0x0004740000b79969', payload '{"action":null,"calibration_mode":"specific_nllv","led_in_dark":null,"linkquality":124,"position":0,"state":"OPEN","tilt":0,"update":{"installed_version":3556351,"latest_version":3556351,"state":"idle"},"update_available":null}'
Info 2023-11-02 19:24:49MQTT publish: topic 'zigbee2mqtt/0x0004740000b79969', payload '{"action":null,"calibration_mode":"specific_nllv","led_in_dark":null,"linkquality":120,"position":0,"state":"OPEN","tilt":0,"update":{"installed_version":3556351,"latest_version":3556351,"state":"idle"},"update_available":null}'
Info 2023-11-02 19:25:16MQTT publish: topic 'zigbee2mqtt/0x0004740000b79969', payload '{"action":null,"calibration_mode":"specific_nllv","led_in_dark":null,"linkquality":128,"position":100,"state":"CLOSE","tilt":0,"update":{"installed_version":3556351,"latest_version":3556351,"state":"idle"},"update_available":null}'
Info 2023-11-02 19:25:45MQTT publish: topic 'zigbee2mqtt/0x0004740000b79969', payload '{"action":null,"calibration_mode":"specific_nllv","led_in_dark":null,"linkquality":108,"position":0,"state":"OPEN","tilt":0,"update":{"installed_version":3556351,"latest_version":3556351,"state":"idle"},"update_available":null}'
Info 2023-11-02 19:27:15MQTT publish: topic 'zigbee2mqtt/0x0004740000b79969', payload '{"action":null,"calibration_mode":"specific_nllv","led_in_dark":null,"linkquality":116,"position":9,"state":"OPEN","tilt":0,"update":{"installed_version":3556351,"latest_version":3556351,"state":"idle"},"update_available":null}'
Info 2023-11-02 19:27:41MQTT publish: topic 'zigbee2mqtt/0x0004740000b79969', payload '{"action":null,"calibration_mode":"specific_nllv","led_in_dark":null,"linkquality":104,"position":100,"state":"CLOSE","tilt":0,"update":{"installed_version":3556351,"latest_version":3556351,"state":"idle"},"update_available":null}'
Info 2023-11-02 19:28:34MQTT publish: topic 'zigbee2mqtt/0x0004740000b79969', payload '{"action":null,"calibration_mode":"specific_nllv","led_in_dark":null,"linkquality":104,"position":61,"state":"CLOSE","tilt":0,"update":{"installed_version":3556351,"latest_version":3556351,"state":"idle"},"update_available":null}'

With K4027C classic_nllv not working, with specific it calibrates but I think some parameter is not he same as 067776...

FabianMangold:
Hi @philippk75,

This is very unforunate indeed. The 067776A were integrated in the Legrand device library using the following fingerprint:
```
fingerprint: [
            {modelID: ' Shutter SW with level control\u0000', manufacturerID: 4129},
],
```
Those attributes get reported by my actual Legrand 067776A devices. Given the fact the BTicino devices use the very same Legrand NLLV firmware (or it might be the otherway arround actually), i suppose that your BTicino devices report the same attributes.
Could you please send all attributes of your switche's GenBasic cluster, so that we can compare it between the BTicino and Legrand switches ?
Thanks

FabianMangold:
@philippk75 and @syntesys87,
While the BTicino and K4027C seems to share the same firmware, I have no information about the hardware.
Before being able to switch calibration modes, your switches are likely have worked in "Up_Down_Stop" mode right ?
Can you remember what was first reported by Z2M right after the update ?
Are the switches operating as "before" when switching back to the original mode ?

@philippk75: You are absolutely right regarding the tilt. This is only relevant when the switch is operating in venetian mode. Let's check if this option can be made conditional.

philippk75:
> Hi @philippk75,
> 
> This is very unforunate indeed. The 067776A were integrated in the Legrand device library using the following fingerprint:
> 
> ```
> fingerprint: [
>             {modelID: ' Shutter SW with level control\u0000', manufacturerID: 4129},
> ],
> ```
> 
> Those attributes get reported by my actual Legrand 067776A devices. Given the fact the BTicino devices use the very same Legrand NLLV firmware (or it might be the otherway arround actually), i suppose that your BTicino devices report the same attributes. Could you please send all attributes of your switche's GenBasic cluster, so that we can compare it between the BTicino and Legrand switches ? Thanks

hello, could you explan how send the attributes?

syntesys87:
> @philippk75 and @syntesys87, While the BTicino and K4027C seems to share the same firmware, I have no information about the hardware. Before being able to switch calibration modes, your switches are likely have worked in "Up_Down_Stop" mode right ? Can you remember what was first reported by Z2M right after the update ? Are the switches operating as "before" when switching back to the original mode ?
> 
> @philippk75: You are absolutely right regarding the tilt. This is only relevant when the switch is operating in venetian mode. Let's check if this option can be made conditional.

The problem is that in every mode I set, the switch works as up/down/close. With classic_nllv the calibration does not start, also not in temporal mode. The switch operates as before. I think I have logs before and after the update, if you need.

FabianMangold:
> hello, could you explan how send the attributes?

The easiest way is to use the Z2M frontend. Click on the device, open the "Dev Console" tabsheet, select the GenBasic Cluster, then select the attributes (please limit it to 4 items per iteration) and click Read.
The device will then send the attributes back, You may have to do this several times, as select too many attributes simultaneously may not fit in one single ZigBee message. Also, certain attributes may not be available, so don't worry about those.

philippk75:
> > @philippk75 and @syntesys87, While the BTicino and K4027C seems to share the same firmware, I have no information about the hardware. Before being able to switch calibration modes, your switches are likely have worked in "Up_Down_Stop" mode right ? Can you remember what was first reported by Z2M right after the update ? Are the switches operating as "before" when switching back to the original mode ?
> > @philippk75: You are absolutely right regarding the tilt. This is only relevant when the switch is operating in venetian mode. Let's check if this option can be made conditional.
> 
> The problem is that in every mode I set, the switch works as up/down/close. With classic_nllv the calibration does not start, also not in temporal mode. The switch operates as before. I think I have logs before and after the update, if you need.

hello, i confirm, my Bticino N4027C ist able to control the tile inclination :-)

FabianMangold:
@syntesys87: It could very well be, that the BTicino switches simply do not support those modes. Only Legrand / BTicino could tell really. Is there any way your could check if the official "App" supports those modes with those devices?
In a first step, it might be easier just to avoid using those modes on BItico devices, and we should open a new thread to check if those modes can be supported on those devices.
Again, the changes were only meant to be for Legrand devices, not the BTicino ones.

FabianMangold:
Oh interesting @philippk75 (CC: @syntesys87). So it does work on BTicino devices too.
Are you in Venetian Mode now ? How did you calibrate it ? Is the procedure i meantioned above Okay or did you change anything ?

philippk75:
> The easiest way is to use the Z2M frontend. Click on the device, open the "Dev Console" tabsheet, select the GenBasic Cluster, then select the attributes (please limit it to 4 items per iteration) and click Read.
> The device will then send the attributes back, You may have to do this several times, as select too many attributes simultaneously may not fit in one single ZigBee message. Also, certain attributes may not be available, so don't worry about those.

hello, this ist the log, is this enough? für he other attibutes i get an erroer while reading?

2023-11-02 20:54:42Read result of 'genBasic': {"zclVersion":2,"appVersion":0,"stackVersion":67,"hwVersion":13}
2023-11-02 20:57:25Read result of 'genBasic': {"manufacturerName":" Legrand\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000","modelId":" Shutter SW with level control\u0000"}
2023-11-02 20:57:56Read result of 'genBasic': {"dateCode":" \u000020230406\u0000\u0000\u0000\u0000\u0000","powerSource":1}
2023-11-02 20:58:15Read result of 'genBasic': {"dateCode":" \u000020230406\u0000\u0000\u0000\u0000\u0000","powerSource":1,"appProfileVersion":0,"swBuildId":"0036"}

philippk75:
> @syntesys87: It could very well be, that the BTicino switches simply do not support those modes. Only Legrand / BTicino could tell really. Is there any way your could check if the official "App" supports those modes with those devices?
> In a first step, it might be easier just to avoid using those modes on BItico devices, and we should open a new thread to check if those modes can be supported on those devices.
> Again, the changes were only meant to be for Legrand devices, not the BTicino ones.

ich verstehe, wir sind ja nach dem upddate unfreiwillig dazu gekommen und wir plötzlich nicht mehr bticino taster haben sondern legrand :-) kann man das rückgängig machen?

philippk75:
> @syntesys87: It could very well be, that the BTicino switches simply do not support those modes. Only Legrand / BTicino could tell really. Is there any way your could check if the official "App" supports those modes with those devices? In a first step, it might be easier just to avoid using those modes on BItico devices, and we should open a new thread to check if those modes can be supported on those devices. Again, the changes were only meant to be for Legrand devices, not the BTicino ones.

i understand, we were involuntarily added after the update and suddenly we no longer have bticino buttons but legrand :-) can this be reversed?

syntesys87:
The N4027C and K4027C seems to have the same firmware, for Zigbee2MQTT documentation. These are the screenshots of my configuration from Home+Control app of one switch and it's equal to all the others:
[Photos-001.zip](https://github.com/Koenkk/zigbee-herdsman-converters/files/13244025/Photos-001.zip)

I can provide also the logs from Netatmo integration.

philippk75:
> Oh interesting @philippk75 (CC: @syntesys87). So it does work on BTicino devices too.
> Are you in Venetian Mode now ? How did you calibrate it ? Is the procedure i meantioned above Okay or did you change anything ?

at the moment i can only change the inclination of the tile physically on the switch by holding longer the up or down buttom... i cannont selct a venetian mode in the driver of the device, ist there an option in the legrand driver?

FabianMangold:
Ja, deshlab frage ich nach den Details!
Thanks for the attribute report. Your device reports the same attributes than my Legrand ones, with the exception of the HW version.
Your BTicino ones report HW version 13,
My Legrand 067776 HW version 8
and the 078887A HW version 1

Those attributes can be used to differentiate the devices in Z2M.
Now we need to device if we need to revert back, or leave it as is.
I am confused, because you say that the device is able to control the tilt angle, which seems to be confirmed by @syntesys87 screenshots.

Should Bticino devices get the feature set as well, or should we "try" to revert back to the former behaviour ?
Just looking @Koenkk in here :)

FabianMangold:
@philippk75: The calibration procedure is necessary, otherwise the "switch" cannot apply the tilt angle.
Could you please attempt to calibrate one of your blinds ? Just did it on one of mine, and i confirm that the procedure works for me.

philippk75:
> @philippk75: The calibration procedure is necessary, otherwise the "switch" cannot apply the tilt angle. Could you please attempt to calibrate one of your blinds ? Just did it on one of mine, and i confirm that the procedure works for me.

I would like the bticino to be able to control the inclination of the slats. it would be nice to see the old picture back and the old name of the device :-) maybee you could me explan in german how to calibrate and how to set the ebenetian mode?

philippk75:
> @philippk75: The calibration procedure is necessary, otherwise the "switch" cannot apply the tilt angle.
> Could you please attempt to calibrate one of your blinds ? Just did it on one of mine, and i confirm that the procedure works for me.

maybee in german it's easier to understand? i have venetians and traditional roller shutter, it would be fine that the tile option would be visible only für venetians

philippk75:
> * Temporal:
>   Once the blind is in the up position, start the calibration by pressing and holding the down button, until the blind reaches to bottom stop. Once there, the relay should click. Then shortly press down down button again and subsequently press and hold the up button until the blind gets to the upper stop.
>   Repeat the cycle: Press and hold down until the blind reaches the bottom stop, then press and hold until it reaches the top position (Note that there is no need to shortly press the down button in the second sweep cycle)
> * Venetian / BSO
>   Similar to the temporal calibration method, with the exception of the second sweep cycle!
>   In essence, do the first sweep as mentioned in the "Temporal" section, and once done, press and hold the down button until the venetian blind reaches the botton position. Then press and hold the UP switch: this switch will then calibrate the blind tilt angle by activating the switch briefly 4 times in a sequence.

i tried your procedure for the venetian mode, but it doesn't work, maybee because i have an automatic calibration mode? for the calibration i only hd to push simultanly the up an down buttom, than the led lights yellow and the shutter movs up an down an then the led turn to blue.

FabianMangold:
Well, the procedure may differ. The problem is that those are not documented, and are the ones that worked for me, on Legrand devices, after trial and error.
Pressing Up and Down indeed puts the device in calibration mode, but this was not required for me. Changing the calibration mode with Z2M automatically put the device in that mode, and thus, the calibration could be started straightaway.
Can you please try the temporal one ? This procedure is the one i got from the Legrand App, and has proven to be reliable. I just want to check if this works on BTicino devices as well. Once calibrated, you won't be able to set the tilt, but to request a position..
If this works reliably, we can take it from there, and tacke the Venetian mode in a second step.

Regarding the remaining open items, I have to check with @Koenkk as Z2m doesn't support this (AFAIK) out of the box:
- Device identification: All attributes seem to be identical, only the HW version is different. Unforunately, this attribute cannot (yet) be used for fingerprinting a device
- Conditional Tilt: Would it be possible to dynamically re-evaluate the "exposes" array upon change of an attribute (i.e. the calibration_mode in this speific case ?)

syntesys87:
@FabianMangold how did you approach the debug of NLLV firmware and the tries you did for make this working? I would debug how it works in the app and replicate it in Zigbee2MQTT.

FabianMangold:
Hi @syntesys87;
I am sorry, but is a bit of our the scope here, and I think this should be moved over to another thread. 
This this is the wrong place for such discussions, as this thread is meant to address topics arround this PR.
Unless there is no major issues, new discussions threads should be started elsewhere.
Thanks for your understanding. This simplifies tracking for everyone.

We still need to tackle the two points above however:
- Device identification: I will submit a PR to address that
- Conditional "exposes": This is more a nice to have, and not directly linked to this PR, it's just inherent to the way Z2m works, and a bit of a corner case, as the device supports so many different features.

philippk75:
> Well, the procedure may differ. The problem is that those are not officially documented. The description above reflects the procedures that worked for me on Legrand devices,. Those were mostly determine by "trial and error". Pressing Up and Down indeed puts the device in calibration mode, but this was not required for me. Changing the calibration mode with Z2M automatically put the device in that mode, and thus, the calibration could be started straightaway. Can you please try the temporal one ? This procedure is the one i got from the Legrand App, and has proven to be reliable. I just want to check if this works on BTicino devices as well. Once calibrated, you won't be able to set the tilt, but to request a position.. If this works reliably, we can take it from there, and tacke the Venetian mode in a second step.
> 
> Regarding the remaining open items, I have to check with @Koenkk as Z2m doesn't support this (AFAIK) out of the box:
> 
> * Device identification: All attributes seem to be identical, only the HW version is different. Unforunately, this attribute cannot (yet) be used for fingerprinting a device
> * Conditional Tilt: Would it be possible to dynamically re-evaluate the "exposes" array upon change of an attribute (i.e. the calibration_mode in this speific case ?)

hello, temporal seems to work, at least the rollo corrket reacts accordingly... with classical.nllv and specific.nllv the calibration process also starts automatically and does everything by itself.

what are your next activities?

FabianMangold:
HI @philippk75 ,
So you are confirming that "everything" works with your BTicino devices ?
If so, could you please comment [this issue](https://github.com/Koenkk/zigbee2mqtt/issues/19509) here ? 
Could you please tell us how you managed to calibrate the switches in temporal mode ? @syntesys87's devices don't  (yet) seem to cooperate!

The next step is to identify the devices properly, so that the names match again. Would you mind opening another issue for that, so that we can work on it ?
I am still puzzled, as different users report different results. Therefore we need to wait for some more feedback.

Thanks

philippk75:
hello,

i said that the specific calibration mode works and i can stwich to positions, not only 0%, 50% or 100%. I can not check if the temporal mode works, after that the rollo shutters worked.

i found this document that explains the calibration for venetian in italian, the procedure sesms different: https://professionisti.bticino.it/sites/default/files/2022-08/2022_Novit%C3%A0%20impianto%20Smart_gestione%20tapparelle%20frangisole.pdf

philippk75:
> hello,
> 
> i said that the specific calibration mode works and i can stwich to positions, not only 0%, 50% or 100%. I can not check if the temporal mode works, after that the rollo shutters worked.
> 
> i found this document that explains the calibration for venetian in italian, the procedure sesms different: https://professionisti.bticino.it/sites/default/files/2022-08/2022_Novit%C3%A0%20impianto%20Smart_gestione%20tapparelle%20frangisole.pdf

hello, thanks, this means that we have to wait the next zibee2mqtt update? or can we start somethings manuel?

Crystian182:
> * Temporal:
>   Once the blind is in the up position, start the calibration by pressing and holding the down button, until the blind reaches to bottom stop. Once there, release the button, and shortly press the down button. The relay should then click. Subsequently press and hold the up button until the blind gets to the upper stop.
>   Repeat the cycle: Press and hold down until the blind reaches the bottom stop, then press and hold until it reaches the top position (Note that there is no need to shortly press the down button in the second sweep cycle)

hi @FabianMangold,
I calibrated my shutter with specific_nllv mode, by pressing up and down buttons for 7 seconds.
The calibration worked fine but i don’t trust the percentages i set via homeassistant, since they are different from what i visually perceive.
I would like to manually define my own percentages, so i was looking at the temporal mode, which should fit my case allowing me to define where’s the point at 0% and where’s 100%. Isn’t it?
If yes, should i send the temporal calibration command via homeassistant before starting the procedure you described? Or it is not neeeded and i could just start the procedure when the blind is at my target up position?

jbouv55151:
> > hello,
> > i said that the specific calibration mode works and i can stwich to positions, not only 0%, 50% or 100%. I can not check if the temporal mode works, after that the rollo shutters worked.
> > i found this document that explains the calibration for venetian in italian, the procedure sesms different: https://professionisti.bticino.it/sites/default/files/2022-08/2022_Novit%C3%A0%20impianto%20Smart_gestione%20tapparelle%20frangisole.pdf
> 
> hello, thanks, this means that we have to wait the next zibee2mqtt update? or can we start somethings manuel?

Hi,
Did you managed to use the tilt option with Venetian mode ?

PR #6336, opened 2023-10-25
Fix Elko Super TR: Avoid 1 second update

arnerek:
Partly fixes https://github.com/Koenkk/zigbee2mqtt/issues/18619


Koenkk:
Thanks!

PR #6340, opened 2023-10-26
Update bosch.ts to support Door/window contact II plus

cruelacid:
Corrected the name of BSEN-CV model and exposed vibration sensor for this product, including the converter to extract the vibration sensor data.

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -922,7 +922,7 @@ const definitions: Definition[] = [
         description: 'Door/window contact II plus',
         fromZigbee: [fzLocal.bosch_contact],
         toZigbee: [],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

psdl76:
In the Bosch App the Vibration Sensitivity can be changed from "very high", "high", "middle", "low" and "very low". I need this because the sensor has the highest sensibility when integrated in zigbee2mqtt. My TV does vibrate the glass of my living room door. :) And i don't know how to debug mqtt messages. Can you elaborate on this topic ?!

cruelacid:
The code that I submitted only interprets the mqtt coming from the sensor. To change the sensitivity you would need to figure out what to send to the sensor in the other direction. I have no idea what the bosch app sends to the sensor to change the settings - you would need to sniff what it's sending before you could code it here...

PR #6342, opened 2023-10-27
feat: Support setting vacation_mode on Ubisys H1

sjorge:
When updated the H1 supports setting vacation_mode, they actually got back to me on this as I requested this feature last year!


Koenkk:
Thanks!

PR #6371, opened 2023-10-31
Add _TZ3000_xsjsnzhz to TS0210.

vindolin:
I bought another TS0210 sensor which was not supported.
I saw the change @Koenkk made in 2051d41 and added this line.
After restarting the service my device now works.


Koenkk:
Thanks!

PR #6376, opened 2023-10-31
Update philips.ts for LTD016

thk-socal:
Adding support for the 4" recessed can lights Zigbee model LTD016 Product model 9290031347.


Koenkk:
Thanks!

PR #6378, opened 2023-11-01
new iotperfect valve fingerprint

ivanfmartinez:
Just received a new valve and it came with different identification, purchased from same place on aliexpress, appear to be a newer version.

Adding the new ID it was detected and work correctly


Koenkk:
Thanks!

PR #6383, opened 2023-11-01
fix: Sonoff TRV child lock status incorrectly set to UNLOCKED

photomoose:
I discovered a minor bug with the Sonoff TRV child lock functionality I recently exposed. 

When the TRV sends an attribute report for the same cluster ID, the child lock attribute is not always be present in the payload. My previous code set the child lock to 'UNLOCK' if the attribute was not truthy. As a result, when the TRV sent an attribute report for the cluster without the child lock attribute (e.g. when the temperature is remotely set), the child lock state was incorrectly set to UNLOCK, despite it actually being LOCK.

This PR fixes the issue by checking for the presence of the attribute.


Koenkk:
Thanks!

PR #6384, opened 2023-11-02
fix AduroSmart Eria 81812

levogevo:
I had used wrong model when adding (been a while since I made my personal converter). This fixes the model for the converters.

### DIFF
--- a/src/devices/adurosmart.ts
+++ b/src/devices/adurosmart.ts
@@ -130,9 +130,9 @@ const definitions: Definition[] = [
         meta: {applyRedFix: true},
     },
     {
-        zigbeeModel: ['AD-ColorTemperature3001'],
+        zigbeeModel: ['AD-81812', 'AD-ColorTemperature3001'],
         model: '81812',
-        vendor: 'AduroSmart Eria',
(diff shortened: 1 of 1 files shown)


src/devices/adurosmart.ts
```diff
@@ -130,9 +130,9 @@ const definitions: Definition[] = [
         meta: {applyRedFix: true},
     },
     {
-        zigbeeModel: ['AD-81812'],
+        zigbeeModel: ['AD-ColorTemperature3001'],
```
Koenkk, line 133:
```suggestion
        zigbeeModel: ['AD-81812', 'AD-ColorTemperature3001'],
```

The old one should stay, otherwise people suddenly have unsupported devices 😄

levogevo: Sorry, I am not the expert in this; Should I keep the old one and just append the "fixed" version or is there another way to have two model/vendor names neatly be linked together.

In either case, considering I added the `'AD-81812` converter in the first place, I have my doubts about whether it will affect anyone else. 

Thanks for your time.

Koenkk:
Thanks!

PR #6390, opened 2023-11-03
Consolidate Legrand TZ led handling

FabianMangold:
This PR:
- Consolidates the Legrand led_in_dark and led_if_on FZ handlers in a single manufacturer specific led_mode FZ converter.
- Adds the "Led If On" feature to the 067776 / 067776A switches


Koenkk:
Thanks!

PR #6410, opened 2023-11-05
Update osram.ts

los7702:
Added Colortemp on model 4062172044776_1
to support Tunable White as declared by Osram


los7702:
this is a general device and fixed min and max coulors is not available. it depends on the dali devices konnected to this Osram Dali converter

PR #6412, opened 2023-11-05
Update vesternet VES-ZB-SWI-015

devildant:
fix the impl, the measurement are provide on endpoint11 not 1

proof : 
paper documentation : 
image

result on Z2M:
image

### DIFF
--- a/src/devices/vesternet.ts
+++ b/src/devices/vesternet.ts
@@ -118,16 +118,18 @@ const definitions: Definition[] = [
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint1 = device.getEndpoint(1);
             const endpoint2 = device.getEndpoint(2);
-            await reporting.bind(endpoint1, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement', 'seMetering']);
+            const endpoint11 = device.getEndpoint(11);
+            await reporting.bind(endpoint1, coordinatorEndpoint, ['genOnOff']);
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

chris-1243:
I don't know if it is the right place but it seems the fix is not present on 1.34.0-dev commit: 87f6b9e.

I am exactly facing the same issue and the binding is not correct.

Any work around until the issue is solved ?
[sunricher_error.txt](https://github.com/Koenkk/zigbee-herdsman-converters/files/13649969/sunricher_error.txt)
[Zigbee-herdsman.txt](https://github.com/Koenkk/zigbee-herdsman-converters/files/13649970/Zigbee-herdsman.txt)

I have added some log file if it helps

martyn-vesternet:
@chris-1243 if you're using the VES-ZB-SWI-015 device from Vesternet then there's an external converter at https://github.com/vesternet/zigbee2mqtt-herdsman-converters which should fix this issue now. It should hopefully configure the device correctly during pairing based on its reported firmware version.

It seems to work OK for me on the two versions of test device that I have here, perhaps you can check it on yours?

chris-1243:
I just had a try on the latest-dev and it is not implemented.

Using your file as an external converter, the pairing went successful and the endpoints were ALL corrects even the reporting (endpoint 11 based on my firmware).

It seems to be perdectly working on my side.

Many thanks for your effort

PR #6413, opened 2023-11-06
Update philips.ts for Festavia Effects

thk-socal:
Added in supported extraEffects Glisten and Opal to the Festavia 250 and 500 count string lights.


Koenkk:
Thanks!

PR #6415, opened 2023-11-06
feat: add support for Sonoff TRV local temperature calibration

photomoose:
Adds support for Sonoff TRV local temperature calibration.

Also reworded the local temperature calibration description to make it slightly friendlier (in my opinion).

Issue being tracked here: https://github.com/Koenkk/zigbee2mqtt/issues/19269


Koenkk:
Thanks!

PR #6422, opened 2023-11-06
Add Cleverio CZWS-100 support

waroen:
Add Cleverio CZWS-100 support, it was previously detected as a TuYa water leak detector but with some errors

- Tampered with was always on
- No battery percentage reported

Removed those attributes as they were not working anyway.


Koenkk:
Please integrate these changes into `TS0207_water_leak`, e.g. the tamper can be filtered here: 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/7967cee56ec717b34eb11ced46038c470cd52d3c/src/devices/tuya.ts#L1168

waroen:
I can do that, I'll have to rebuild a bit more to also filter out the battery percentage as well in that case. 

The device never reports anything there that I can detect.

I'll try it out and get back with an update.

PR #6423, opened 2023-11-06
fix(ignore): Fixed BTicino K4027C family detection

FabianMangold:
* 067776A: added additional HW version for BTicino whitelabels

Reference:
https://github.com/Koenkk/zigbee2mqtt/issues/19509#issuecomment-1794455410

Hi @Koenkk,
One more use report for that shutter switch.

I am starting to question the approach. Only Legrand can actually tell how those switches can be identified properly in the end. I am okay to maintain a list of hardware Version in the fingerprint for now, but cannot tell 100% if that doesn't break things for other users.
What do you think ?

### DIFF
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -190,7 +190,7 @@ const definitions: Definition[] = [
         whiteLabel: [
             {
                 model: 'K4027C/L4027C/N4027C/NT4027C', vendor: 'BTicino', description: 'Shutter SW with level control',
-                fingerprint: [{hardwareVersion: 13}],
+                fingerprint: [{hardwareVersion: 9}, {hardwareVersion: 13}],
             },
         ],
         ota: ota.zigbeeOTA,


Koenkk:
> What do you think?

This is fine, let's iterate further based on the user feedback.

Bobsilvio:
i have a bticino (no tilt) and now is show legrand (with tilt)

Screenshot 2023-11-14 alle 18 03 32

Bobsilvio:
Please ceck. it is already attached

FabianMangold:
Hi @Bobsilvio,
Sorry, my mistake, didn't see your attachment. Could you please use the "DEV" branch of Z2M ?
All your switches with HwVersion 9 will then be displayed as BTicino devices.

The ones having HwVersion 4 are more problematic! As a matter of fact we have users reporting that devices with that HwVersion are actually Legrand 067776A switches (or more precisely 067726A devices) [here](https://github.com/Koenkk/zigbee2mqtt/issues/19108#issuecomment-1796409957)
@Koenkk: With reference to P[R#6423](https://github.com/Koenkk/zigbee-herdsman-converters/pull/6423), this is unfortunately the situation we wanted to avoid.
I compared Bobsilvio's, Diaour's and my database.db, there is strictly no criteria to differentiate those switches if fear.

@Bobsilvio: Do you have the original gateway / app ? Is it a Legrand App ? If so, do the switches show up as BTicino switches? I just wonder how the original GW / APP is able to tell them apart.

Koenkk:
@FabianMangold if that's the case, we cannot distinguish them indeed. In that case I propose to keep detecting them as Legrand with tilt.

Rendeik:
I also have 1 bticino 4027c which I see as a legrand.
I tried to do a factory reset and now I can no longer complete the pairing.
I have control of the up/down stop during the 4 minutes of pairing, then the red light remains fixed and I can only use it without connection.

PR #6424, opened 2023-11-06
fix - Inovelli VZW35\VZM31 lookup values for aux switch button presses

nathanfiscus:
This fixes an issue where a button press from an aux switch would show up as unknown because the lookup wasn't defined.


Koenkk:
Thanks!

PR #6425, opened 2023-11-06
feat: add support for Sonoff TRV frost protection

photomoose:
Adds support for setting/getting frost protection temperature on Sonoff TRV.

I decided to add a new `withFrostProtectionTemperature` builder function to the existing `Climate` expose object. As a result, in order for this to be discovered by Home Assistant, I've had to add a new `frostProtectionTemperature` `DiscoveryEntry` in  the Home Assistant extension in the main zigbee2mqtt repository - see PR #19606 in zigbee2mqtt repository.

In hindsight, it might have been easier/simpler just to create a new `e.frost_protection_temperature()` function in `exposes.ts` which simply wraps a `Numeric` rather than extending the existing `Climate` object. Let me know if you'd prefer this option, and I will make the adjustments accordingly.

Issue tracked at: https://github.com/Koenkk/zigbee2mqtt/issues/19269


src/lib/exposes.ts
```diff
@@ -542,6 +542,13 @@ export class Climate extends Base {
         this.features.push(schedule);
         return this;
     }
+
+    withFrostProtectionTemperature(min=4.0, max=35.0, step=0.5, access =a.ALL) {
```
Koenkk, line 546: Please move this out of the climate expose (and just add it to the `exposes: []`) of the device. We don't want to pollute the climate expose.

photomoose: No problem, will move it later this evening.

photomoose:
This should be good to go now.

Koenkk:
Perfect, thanks!

PR #6433, opened 2023-11-08
fix: :bug: fixed converters for LEL and CO values

marvinroman:
After receiving TuYa gateway and attaching device I found a difference in the way the values were reported. I have made adjustments to fix these values. See application screenshot below. 

image

Ran

- [x] `npm lint`  
- [x] `npm run build`  
- [x] `npm test`


Koenkk:
Thanks!

PR #6436, opened 2023-11-08
fix(refactor): Consolidated Legrand options

FabianMangold:
Small optimization.

Make use of the global device options (exported from the specific Legrand library), instead of redefining them in various parts of the code.


Koenkk:
ty!

PR #5550, opened 2023-03-04
[WIP] weekly_schedule expose

sjorge:
-  [x] Fixed a bug that snuck into the last PR
-  [x] Start calculated derived values: mode and numoftrans
   -  [x] if we encounter a heatSetPoint we add "heat" to mode
   -  [x] if we encouter a coolSetPoint we add "cool" to mode
   -  [x] numoftrans = transitions.length (we now check if it's an array and error out otherwise)
- [ ] expose data

Since we now calculate the derived values, this should make the exposes data simpler.

The following now works correctly:
```
mosquitto_pub -t zigbee2mqtt/trv/office/set -m '{"weekly_schedule":{"dayofweek":["saturday","sunday"],"transitions":[{"transitionTime":"00:00","heatSetpoint":16},{"transitionTime":"09:00","heatSetpoint":20},{"transitionTime":"22:00","heatSetpoint":16}]}}'
```

PR #6443, opened 2023-11-10
Add ability to set weekly schedule to Sonoff TRV

photomoose:
Adds ability to *set* weekly schedule on Sonoff TRV.

It looks like Sonoff TRVs do not support the "Get Weekly Schedule" command, so it is not possible to query for the existing schedule. The device does, however, send the schedule back over multiple "Get Weekly Schedule Response" commands (one message per day of week) after pairing the device and after setting a schedule. This causes issues with Z2M however, as it appears that the last message received "wins" and the state is updated only with the schedule for the day represented by the last message.

From what I can tell, Z2M appears to support only one schedule over multiple days rather than having the ability to create a different schedule for each day. If different schedules per day are indeed possible, please let me know and I will investigate further.

Issue tracked at: https://github.com/Koenkk/zigbee2mqtt/issues/19269

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -3,7 +3,7 @@ import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as constants from '../lib/constants';
 import * as reporting from '../lib/reporting';
-import {binary, enumLookup, forcePowerSource, numeric, onOff} from '../lib/modernExtend';
+import {binary, enumLookup, forcePowerSource, numeric, onOff, customTimeResponse} from '../lib/modernExtend';
 import {Definition, Fz, KeyValue, KeyValueAny, ModernExtend, Tz} from '../lib/types';
 import * as ota from '../lib/ota';
 import * as utils from '../lib/utils';
@@ -623,6 +623,7 @@ const definitions: Definition[] = [
                 access: 'STATE_GET',
             }),
             sonoffExtend.weeklySchedule(),
+            customTimeResponse('1970_UTC'),
         ],
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
@@ -632,22 +633,6 @@ const definitions: Definition[] = [
             await reporting.thermostatSystemMode(endpoint);
             await endpoint.read('hvacThermostat', ['localTemperatureCalibration']);
             await endpoint.read(0xFC11, [0x0000, 0x6000, 0x6002, 0x6003, 0x6004, 0x6005, 0x6006, 0x6007]);
-        },
-        onEvent: async (type, data, device) => {
-            device.skipTimeResponse = true;
-            // The Zigbee Cluster Library specification states that the genTime.time response should be the
-            // number of seconds since 1st Jan 2020 00:00:00 UTC.
-            // This device, however, expects it to be number of seconds since the Unix Epoch (1st Jan 1970 00:00:00 UTC).
-            // Disable the responses of zigbee-herdsman and respond here instead.
-            // https://github.com/Koenkk/zigbee2mqtt/issues/19269#issuecomment-1802898476
-            if (type === 'message' && data.type === 'read' && data.cluster === 'genTime') {
-                const secondsUTC = Math.round(((new Date()).getTime()) / 1000);
-                const secondsLocal = secondsUTC - (new Date()).getTimezoneOffset() * 60;
-                await device.getEndpoint(1).readResponse('genTime', data.meta.zclTransactionSequenceNumber, {
-                    time: secondsUTC,
-                    localTime: secondsLocal,
-                });
-            }
         },
     },
     {
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -44,7 +44,6 @@ const sonoffExtend = {
             cluster: 'hvacThermostat',
             type: ['commandGetWeeklyScheduleRsp'],
             convert: (model, msg, publish, options, meta) => {
-
                 const day = Object.entries(constants.thermostatDayOfWeek)
                     .find((d) => msg.data.dayofweek & 1<<+d[0])[1];
 
@@ -666,8 +665,6 @@ const definitions: Definition[] = [
         extend: [onOff()],
     },
 ];
-
-
 
 export default definitions;
 module.exports = definitions;
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -3,12 +3,12 @@ import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as constants from '../lib/constants';
 import * as reporting from '../lib/reporting';
-import {binary, forcePowerSource, numeric, enumLookup, onOff} from '../lib/modernExtend';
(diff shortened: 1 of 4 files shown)


src/devices/sonoff.ts
```diff
@@ -433,7 +433,11 @@ const definitions: Definition[] = [
                 .withLocalTemperature()
                 .withLocalTemperatureCalibration(-7.0, 7.0, 0.2)
                 .withSystemMode(['off', 'auto', 'heat'], ea.ALL, 'Mode of the thermostat')
-                .withRunningState(['idle', 'heat'], ea.STATE_GET),
+                .withRunningState(['idle', 'heat'], ea.STATE_GET)
+                // TRVZB does not appear to support the Get Weekly Schedule command (0x02).
```
Koenkk, line 437: Do you get an error when trying to get the schedule?

photomoose: Oh, I didn't realise this failed the tests. 

I've attached a trace showing the ZCL packets when pairing the TRV with Z2M. 

Packet 46 is Z2M sending the Get Weekly Schedule command
Packet 47 is the TRV responding with Unsupported Cluster Command (0x81) - which is why I removed the GET access.

Also notice how packets 35 to 41 show the TRV sending each day's schedule as a separate command - from what I can tell, this happens after pairing and after sending a Set Weekly Schedule command to the device. I haven't yet found a way to explicitly ask for the schedule, given the official command fails. The device supports the ability to configure different schedules for each day, which is why I guess it reports each day's schedule in a single packet. My understanding of Z2M is that this is currently not supported - from what I can tell, you can only configure a single schedule in Z2M and apply it to multiple days, and not have different schedules for different days. (I could be wrong though!) As a result, when Z2M receives these messages, the last message "wins":

Current state of the device in Z2M:

```
{
    "battery": 38,
    "battery_low": null,
    "child_lock": "UNLOCK",
    "frost_protection_temperature": 7,
    "linkquality": 184,
    "local_temperature": 16.5,
    "local_temperature_calibration": 0,
    "open_window": "ON",
    "running_state": "heat",
    "occupied_heating_setpoint": 19,
    "system_mode": "heat",
    "weekly_schedule": {
        "days": [
            "sunday"
        ],
        "transitions": [
            {
                "heating_setpoint": 5,
                "time": 0
            },
            {
                "heating_setpoint": 30,
                "time": 878
            },
            {
                "heating_setpoint": 5,
                "time": 879
            },
            {
                "heating_setpoint": 5,
                "time": 879
            },
            {
                "heating_setpoint": 5,
                "time": 879
            },
            {
                "heating_setpoint": 5,
                "time": 879
            }
        ]
    }
}
```

The `weekly_schedule` in the state above reflects the payload received in packet 41; the schedules in packets 35 to 40 were overwritten by packet 41.


[Sonoff Weekly Schedule.pcapng.zip](https://github.com/Koenkk/zigbee-herdsman-converters/files/13325265/Sonoff.Weekly.Schedule.pcapng.zip)

photomoose: [Sonoff Z2M Set Schedule.pcapng.zip](https://github.com/Koenkk/zigbee-herdsman-converters/files/13325307/Sonoff.Z2M.Set.Schedule.pcapng.zip)

Here's another little trace showing the current behaviour of when setting the weekly schedule from Z2M.

Packet 1 is a Set Weekly Schedule command from Z2M to the device for a schedule on Wednesday.
Packet 4 is the device responding with a Get Weekly Schedule Response command with Wednesday's schedule. (It only sends back a command for the days affected by the previous Set Weekly Schedule command).

Packet 5 is another Set Weekly Schedule command - I added Saturday to the list of days in Z2M.
Packets 8 and 9 are the responses from the TRV as they were the days affected by the previous command.

It feels like if we want to be able to fully support this, then Z2M might need to cater for the ability to configure separate schedules for each day.

Koenkk: If I understand correctly, when setting the schedule multiple times, you can configure different schedules per day currently? I know it will not be reflected in `weekly_schedule` but this is just a matter of updating https://github.com/Koenkk/zigbee-herdsman-converters/blob/c3e653bf4a221b121afba26f41f4702bd071b298/src/converters/toZigbee.js#L1226

photomoose: I've added a comment on the original PR to see whether the author considered the idea. I do not see any current way of setting different schedules per day in the frontend.

https://github.com/Koenkk/zigbee-herdsman-converters/pull/5550#issuecomment-1809814572

photomoose: @pawol pointed out that this Sonoff device is very similar to, if not a rebrand of, the Tuya TS0601 _TZE200_z1tyspqw TRV. It looks like this device supports different schedules for each day and has been implemented by[ exposing a text box for each day with multiple time/temperature strings](https://github.com/Koenkk/zigbee-herdsman-converters/blob/9ad8015a141c9afd38c75f7093bd14cac5441570/src/devices/tuya.ts#L3169).

If you're happy with this, I can try to replicate something similar for this device.

Koenkk: @photomoose the TS0601 uses the TuYa datapoints, while this device uses the standard ZCL Zigbee commands, so maybe it looks the same, but the software is definitely different.

photomoose: @Koenkk sure, I understand the devices are different. What I meant was I could write something similar to the Tuya device - i.e. expose 7 textboxes in the UI (one for each day) and parse the text as a series of times and temperatures.

For example:

Monday: "09:00/19 12:00/16 18:00/20"
Tuesday: "09:00/19 12:00/16 18:00/20"
...
Sunday: "09:00/19 12:00/16 18:00/20"

I think this might be easier to do than trying to make the current front end support this schedule format.

Koenkk: Agree!

photomoose:
I'm going to start looking into this.

photomoose:
@Koenkk this should be good for a review. UI looks like this:

image

Note that I have not found a way to query for the schedule (the getWeeklySchedule command is unsupported). It looks like the TRV only reports the schedule after pairing to a coordinator or after setting the schedule.

Koenkk:
Thanks for the updates, what was the reason not to use [`thermostat_weekly_schedule`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/f840df6e771282b14f5053a100c197e6f308970e/src/converters/toZigbee.ts#L1250)?

photomoose:
The reason for not using the existing `thermostat_weekly_schedule` is because the Sonoff TRV allows for a different schedule to be defined per day, whereas I believe the existing functionality applies a single schedule to one or many days.

When the TRV is paired, it sends out 7 `commandGetWeeklyScheduleRsp` messages, one for each day of the week with the schedule for that day. Likewise, in the Sonoff eWeLlink app, schedules are configured on a per day basis and so saving a schedule for a week would result in 7 `setWeeklySchedule` messages being sent to the TRV.

This PR sends out 7 `setWeeklySchedule` messages when the schedule is applied. When a `commandGetWeeklyScheduleRsp` is received for a sinlge day, it is merged with the existing state so that receiving 7 messages would build the state for the week. The current behaviour overwrites the state, so only the last received message "wins".

Koenkk:
Thanks!

PR #6453, opened 2023-11-12
Add support for ID Lock 202 Multi

andrebk:
ID Lock has released a new lock called "202 Multi". It uses the same zigbee module as the existing ID Lock 150 model which is already supported.

I was able to get it to work for my lock with an external converter that simply copied the existing converter for the 150 and changed the model name. As I have understood it that means I can just add the new model to the `zigbeeModel` array, but this is my first contribution in this repo so please let me know if that's wrong.


Koenkk:
Thanks, congrats with your first contribution!

PR #6458, opened 2023-11-12
tuya.ts: Supoort Immax Neo Lite 07732L TRV

Yenya:
I have Immax 07732B and wanted to buy a new one, but got under the same label and a similar appearance another type _TZE200_rufdtfyv. Just adding it to the same section as 07732B makes it work with zigbee2mqtt, only the calendar schedules are empty, so they might be different in this model.

So far I tested that the values are read correctly, boost mode, manual mode and, force open/close mode work as expected. So I guess as a first approximation this can be merged as is.


Koenkk:
Thanks!

PR #6459, opened 2023-11-12
Add tilt to Profalux covers supporting it

ghoz:
Profalux cover motors can be configured by the remote to act as 3 sort of covers : regular lifting covers, lift and tilt cover, and "soft" fabric covers using the associated remote.
Sniffing the remote during setup shows that the configuration is stored on the cover in attribute 0 of cluster 0xFC21 and thus requires an update to zigbee-herdsman cluster list.
PR : https://github.com/Koenkk/zigbee-herdsman/pull/800

| config | mode | setup on the remote |
| - | - | - |
| 0 | default hard cover (aka Volet)  | 2xF Up + Down | 
| 1 | cover using tilt (aka BSO) | 2xF Stop + Up |
| 2 | soft cover (aka store)     | 2xF Stop + Down |

This PR allows to retrieve at configure time the type of cover and exposes the lift feature if configured on the cover

The configuration itself is not  exposed as I don't think it would make sense for normal use

### DIFF
--- a/src/devices/profalux.ts
+++ b/src/devices/profalux.ts
@@ -40,9 +40,15 @@ const definitions: Definition[] = [
         description: 'Cover',
         fromZigbee: [fz.command_cover_close, fz.command_cover_open, fz.cover_position_tilt],
         toZigbee: [tz.cover_state, tz.cover_position_tilt],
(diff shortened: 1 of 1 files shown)


ghoz:
My typescript fu is too limited, not sure how to interpret the failing tests, 
I'm assuming the dynamic expose causes issues with the endpoint requests / read attribute

PR #6450, opened 2023-11-11
fix Aqara L1-350 (ZNXDD01LM) support

nklebedev:
Fix Aqara L1-350 ceiling light support. Current version has broken support of power on behaviour and color temperature restore at startup (report unsupported attributes).

 `aqaraOpple` cluster is added to support additional attributes. After this, 'device temperature' and 'power outage count' seems to have the correct values. 'power_outage_memory' is also operated via `aqaraOpple` cluster.

The following data was observed from `aqaraOpple` cluster:

```
{
    "3": 33,          <- device temperature
    "5": 106,         <- outage count
    "9": 256,         <- ?
    "10": 0,          <- ?
    "12": 10,         <- ?
    "13": 28,         <- ?
    "17": 1,          <- ?
    "100": 0,         <- current state
    "101": 254,       <- brightness
    "102": 153,       <- color temperature
    "103": 425986700, <- = 0x19640A8C 0x1964 - 6500, 0x0A8C - 2700, min and max color temp
    "154": 0,         <- ?
}
```

`{"1299":6500,"1300":2700}`

`{"238":0}`

PR #6461, opened 2023-11-12
fix: Restore forgotten in #6450 xiaomi converters (regress)

nklebedev:
fix: Restore forgotten in #6450 xiaomi converters (from baseline)


Koenkk:
Thanks!

PR #6470, opened 2023-11-14
add support for 81814

wkwhat:
AduroSmart Eria 81812 and 81814 share the same zigbeeModel idenfitier (AD-ColorTemperature3001).


Koenkk:
Thanks!

Issue #786, opened 2019-11-30
Ikea Lights do not report change when switched externally


asgothian:
Hi all,

i have been running some tests and came against a possible issue. It has been raised in the ioBroker forums as well:
Ikea  lamps will not report their status after they have been switched on or off, unlike the ikea plugs do. This may be the same for other light endpoints as well - i did not check this.

As such, this is not necessarily an issue. It becomes an issue if you have the lamp paired with the coordinator AND an ikea remote. In this situation, it is possible to switch the lamp without any action triggered by the coordinator. As such, the lamps status will not be shown correctly in whichever system is connected to the coordinator.

I have tested with an ikea transformer (the only ikea light i have available at this time) and was able to introduce a configuration which has the transformer report its status. Before i set a pull request, i wanted to put this to a larger number of people. 

Is the benefit (being able to switch lamps even if the underlying smart home system with the configurator for the zigbee network is not working) worth adding this kind of configuration to each ikea light ?

A.

sjorge:
What lamps are you using? My 'LED1624G9 TRADFRI bulb E27 CWS opal 600lm' reports changes for both on/off and brightness if I use a E1743 to control it.

```
## Turn off
zigbee2mqtt:info  2019-12-04 23:14:49: MQTT publish: topic 'zigbee2mqtt/bedroom/desk_lamp/remote', payload '{"linkquality":57,"battery":74,"click":"off"}'
zigbee2mqtt:info  2019-12-04 23:14:50: MQTT publish: topic 'zigbee2mqtt/bedroom/desk_lamp', payload '{"state":"OFF","color":{"x":0.3064,"y":0.2035},"brightness":254,"color_temp":153}'
zigbee2mqtt:info  2019-12-04 23:14:50: MQTT publish: topic 'zigbee2mqtt/bedroom/desk_lamp/bulb', payload '{"state":"OFF","linkquality":60,"color":{"x":0.3064,"y":0.2035},"brightness":254,"color_temp":153}'

## Turn on
zigbee2mqtt:info  2019-12-04 23:14:52: MQTT publish: topic 'zigbee2mqtt/bedroom/desk_lamp/remote', payload '{"linkquality":60,"battery":74,"click":"on"}'
[...]

PR #6472, opened 2023-11-14
Fix reporting for mTouch One

soversk:
Fix reporting for mTouch One. Broken by " Fix reporting for Elko Super TR https://github.com/Koenkk/zigbee-herdsman-converters/issues/786 ".
Plus add RunningState.


Koenkk:
Thanks!

PR #6429, opened 2023-11-07
Update SIN-4-FP-21 and SIN-4-FP-20 from NodOn

AlexisPolegato:
Update SIN-4-FP-21 from NodOn to integrate Fil Pilote Cluster
Add SIN-4-FP-20 from NodOn

PR #6473, opened 2023-11-14
bad cast in toZibee nodon converter, take into account the device button, reverse stop/anti-freeze 

PoloAlexis:
Following "Update SIN-4-FP-21 and SIN-4-FP-20 from NodOn #6429" topic

I found 2 bugs and a missing feature

Bugs;
1- anti-freeze and stop values must be reversed
2- the toZibee for nodon use a buffer instead of a uint8 making the command not effective

Missing feature:
The device can be forced to a mode when using the dedicated button
The current implem does not take this into account

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -6330,10 +6330,10 @@ const converters2 = {
             const payload: KeyValueAny = {};
             const mode = msg.data['mode'];
 
-            if (mode === 0x00) payload.mode = 'stop';
+            if (mode === 0x00) payload.mode = 'anti-freeze';
             else if (mode === 0x01) payload.mode = 'comfort';
             else if (mode === 0x02) payload.mode = 'eco';
-            else if (mode === 0x03) payload.mode = 'anti-freeze';
+            else if (mode === 0x03) payload.mode = 'stop';
             else if (mode === 0x04) payload.mode = 'comfort_-1';
             else if (mode === 0x05) payload.mode = 'comfort_-2';
             else {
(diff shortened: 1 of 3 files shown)


Koenkk:
Thanks!

AlexisPolegato:
@PoloAlexis  Are you sure about the anti-freeze and stop values reversed ?

PoloAlexis:
Hello @AlexisPolegato 

Without the modification, I noticed that the "anti-freeze" mode used to change the "on-off" switch to "off" while the "stop" mode did not. 

So I checked with my oscilloscope.
I found that with the original implementation, the "stop" command leads to passthrough the negative periods.
and "anti-freeze" command leads to passthrough positive period only

This is the reverse of how the wire should be driven.

Based on that I have provided the fix.
Additionnaly, with the fix, when selecting "stop" mode, the switch state goes to "off"

I can give a try (by tomorrow) on one of my heat device and let you know.

AlexisPolegato:
> Without the modification, I noticed that the "anti-freeze" mode used to change the "on-off" switch to "off" while the "stop" mode did not.

Yes, this is normal ; this is the expected behavior.

> So I checked with my oscilloscope. I found that with the original implementation, the "stop" command leads to passthrough the negative periods. and "anti-freeze" command leads to passthrough positive period only
> 
> This is the reverse of how the wire should be driven.

Yes you're right, stop should be positive perdiods and anti-freeze should be negative periods, but then it depends how you do the measure.

> Based on that I have provided the fix. Additionnaly, with the fix, when selecting "stop" mode, the switch state goes to "off"
> 
> I can give a try (by tomorrow) on one of my heat device and let you know.

Ok, would be interesting to have the final result on your heater.

Issue #6426, opened 2023-11-07
bulbOnEvent crashed


arteck:
we have a problem with  ikea bulbs after power on (on wall switch) 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/5be8bf3d5d4eaa611b38b7701ad6f102fb0e6073/src/devices/ikea.ts#L41

```
d023-11-01 19:11:26.277	error	Caught by controller[1]: at Object.bulbOnEvent [as onEvent] (/opt/iobroker/node_modules/zigbee-herdsman-converters/src/devices/ikea.ts:41:36)
2023-11-01 19:11:26.277	error	Caught by controller[1]: TypeError: Cannot read properties of undefined (reading 'color_options')
2023-11-01 19:11:26.275	error	Caught by controller[0]: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:

```

Koenkk:
I think you need to `await` the `onEvent` call.

arteck:
i think the state is null

Koenkk:
is this with z2m? if yes, can you provide the debug log?

arteck:
nope its from iobroker ... but its directly from zigbee-herdsman-converters

arteck:
```
2023-11-14 19:47:47.745 - error: zigbee.0 (16611) unhandled promise rejection: Cannot read properties of undefined (reading 'color_options')
2023-11-14 19:47:47.879 - error: zigbee.0 (16611) TypeError: Cannot read properties of undefined (reading 'color_options')
at Object.bulbOnEvent [as onEvent] (/opt/iobroker/node_modules/zigbee-herdsman-converters/src/devices/ikea.ts:41:36)
at DeviceEvent.callOnEvent (/opt/iobroker/node_modules/iobroker.zigbee/lib/zbDeviceEvent.js:43:26)

```

the color_options restore is the problem
look your info

 * Additionally some other information is lost like
     *   color_options.execute_if_off. We also restore these.
     *
     * NOTE: binds are not lost so rebinding is not needed!
     */

arteck:
i'm not TS expert but i thing this works
https://github.com/Koenkk/zigbee-herdsman-converters/pull/6475

Koenkk:
I saw the pr, but this issue should be fixed at the iobroker side, the `state` value should never be `null`, if it's null pass `{}` instead.

PR #6475, opened 2023-11-15
state is undefined

arteck:
not all bulbs have color_options and crashed


Koenkk:
Discussing in https://github.com/Koenkk/zigbee-herdsman-converters/issues/6426

PR #6479, opened 2023-11-15
Add OTA to SPLZB-13x

qp68:
According to Develco [specs](https://cdn.bitelec.ch/7/1/2/c/712c1adcfb25809300a11eac200b22956c7cae58_smart_plug_mini_datasheet_v44.pdf), SPLZB-131, SPLZB-132, SPLZB-134 and SPLZB-137 are the same, only plug type is different. Thus #6462 should also be applied to SPLZB-132, SPLZB-134 and SPLZB-137.

### DIFF
--- a/src/devices/develco.ts
+++ b/src/devices/develco.ts
@@ -319,6 +319,7 @@ const definitions: Definition[] = [
         description: 'Power plug',
         fromZigbee: [fz.on_off, develco.fz.electrical_measurement, develco.fz.metering, develco.fz.device_temperature],
         toZigbee: [tz.on_off],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

tomtom13:
Hi,
Thanks for that PR, however I'm experiencing some dificulty with it merged into the mainstream.
After updating to newest Z2B, those became available for OTA, and firmware got uploaded. 
Thou current firmware is empty and for build date I've got different dates per device. What's a bit more strange is that those dates are different per device, might be a manufacturing date ? screenshot  attached. 
Screenshot 2023-12-01 at 16 16 21

PR #6483, opened 2023-11-15
fix: IKEA LED2201G8 dimming to off

excitare:
This new lamp seems to have an odd behavior when dimming to off with `transition` set to anything but `0`.
Depending on the current brightness value and transition set, the light might not actually turn off when executing the off command.
Changing the level to anything but 0 works with any transition just fine.

Seems like this is the pattern:
It can work with less than 1 second per 20 brightness steps to dim down.
Here are a few examples (0 <= brightness <= 254):
1. brightness: 50, transition=2 -> the lamp will take 2 seconds to dim down and turn off
2. brightness: 95, transition=6 -> the lamp will take 6 seconds to dim down and **not** turn off.
3. brightness: 10, transition=0.5 -> the lamp will take 0.5 seconds to dim down and **not** turn off. (Especially annoying at nighttime with Adaptive Lighting)

So for the light to actually turn off the requirements are something like this:
brightness < 21 -> transition 0
brightness < 41 -> transition < 2
brightness < 61 -> transition < 3
... and so on

This PR fixes the issue, but not when turning off during a transition.
For example with transition=10
- Start: brightness 100
- Level to brightness 50
- At 3 seconds I turn the light off, the assumed real brightness is ~85
- It takes 10 seconds to dim the light, but it will not turn off, as the brightness value used for the calculation is 100, so the calculated transition is 5, instead of the necessary 4 to be below `85 / 21`

I don't know if that is even such a big issue or a workaround is possible.

I wasn't really sure where to put the code for this, as it is 99% `light_onoff_brightness` logic 😄


src/converters/toZigbee.js
```diff
@@ -1001,6 +1001,151 @@ const converters = {
...
+            }
+
+            await entity.command(
+                'genLevelCtrl',
+                state === null ? 'moveToLevel' : 'moveToLevelWithOnOff',
+                {level: Number(brightness), transtime: scaledTransition},
+                utils.getOptions(meta.mapped, entity),
+            );
+
+            const result = {state: {}, readAfterWriteTime: scaledTransition * 100};
```
excitare, line 1132: Relevant part

Koenkk:
what if you remove the transition entirely?

excitare:
With transition=0 everything works as expected, and the lights don't show this behavior.

Koenkk:
In it's current form this PR is not acceptable (due to the code duplication). But let's first see if others also report this issue, maybe a firmware update of the bulb will fix it.

PR #6486, opened 2023-11-16
Add NodOn mfr specific attribute

AlexisPolegato:
Add NodOn mfr specific attribute to control Impulse Mode


src/devices/nodon.ts
```diff
@@ -47,24 +72,26 @@ const definitions: Definition[] = [
         vendor: 'NodOn',
         description: 'Multifunction relay switch',
         extend: extend.switch(),
+        fromZigbee: [fzLocal.impulse_time],
+        toZigbee: [tzLocal.impulse_time],
+        exposes: [e.numeric('transition_time', ea.SET).withValueMin(0).withValueMax(5000).withUnit('ms').withDescription('Impulse time')],
```
Koenkk, line 77: How does the exposes tab in the frontend show now? Is the switch still there?

src/devices/nodon.ts
```diff
@@ -38,33 +63,44 @@ const definitions: Definition[] = [
...
     },
     {
-        zigbeeModel: ['SIN-4-1-20_PRO'],
         model: 'SIN-4-1-20_PRO',
+        zigbeeModel: ['SIN-4-1-20_PRO'],
         vendor: 'NodOn',
         description: 'Multifunction relay switch',
-        extend: extend.switch(),
+        fromZigbee: [fz.identify, fz.on_off, fz.command_toggle, fz.command_on, fz.command_off, fz.power_on_behavior, fzLocal.impulse_time],
+        toZigbee: [tz.on_off, tz.power_on_behavior, tzLocal.impulse_time],
```
Koenkk, line 97: I see nothing is added to expose for the `impulse_time`, instead of adding a tz/fz and expose, could you use the modernExtend instead? Example https://github.com/Koenkk/zigbee-herdsman-converters/blob/4ce6450b2a55761f7e9f38688b645748885b876d/src/devices/xiaomi.ts#L987

PR #6488, opened 2023-11-16
fix(ignore): Revert "Add vendor Imhotep creation and its devices support"

Koenkk:
Reverts Koenkk/zigbee-herdsman-converters#6478

@nicolasaxenco this breaks the z2m tests because `withLocalTemperature()` is missing for the `e.climate()`, could you add this?


nicolasaxenco:
No problem, thanks, we will add it and make a new pull request

PR #6491, opened 2023-11-17
fix(detect): detect `_TZ3000_saiqcn0y` as WSD500A

gtalusan:
Hello, I picked up a couple of TuYa temperature/humidity sensors from [aliexpress](https://www.aliexpress.com/item/1005003707857013.html) and they're being misdetected by zigbee2mqtt as a TuYa temperature sensor with display.


Koenkk:
thanks!

PR #6492, opened 2023-11-17
Make state invertable

donbobka:


Koenkk:
I prefer to use the already existing `invert_cover` option instead, do you think that is feasible?

Koenkk:
Looks good! If ready for merge, please move out of draft.

donbobka:
I tested it on my one motor that set in one position and expect more to come next week to test it in one go in different configurations.

donbobka:
Tested at left and right side with under roll + left side over roll

Koenkk:
thanks!

PR #6494, opened 2023-11-17
update QS-Zigbee-SEC02-U

nakata5321:
Update QS-Zigbee-SEC02-U. Inheritance from Tuya switch class. This will give new options to switch - "Power-on behavior' and "Switch type".


Koenkk:
Thanks!

PR #6497, opened 2023-11-17
New mini switch and dimmer modules, smoke detector

cserem:


src/devices/tuya.ts
```diff
@@ -1396,6 +1396,77 @@ const definitions: Definition[] = [
...
+            tuya.exposes.lightBrightnessWithMinMax().withEndpoint('l2'),
+            tuya.exposes.countdown().withEndpoint('l1'),
+            tuya.exposes.countdown().withEndpoint('l2'),
+            tuya.exposes.lightType().withEndpoint('l1'),
+            tuya.exposes.lightType().withEndpoint('l2'),
+            e.enum('power_on_behavior', ea.STATE_SET, ['off', 'on', 'previous']),
+        ],
+        meta: {
+            multiEndpoint: true,
+            tuyaDatapoints: [
```
Koenkk, line 1418: Looks very similar to https://github.com/Koenkk/zigbee-herdsman-converters/blob/f8e98d97fad60ce1085ccc869d33bb874767f551/src/devices/tuya.ts#L1312 , does e.g. `[21, 'backlight_mode', tuya.valueConverter.backlightMode],` also work for this device?

src/devices/tuya.ts
```diff
@@ -1396,6 +1396,77 @@ const definitions: Definition[] = [
...
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        exposes: [
+            tuya.exposes.lightBrightnessWithMinMax(),
+            e.enum('power_on_behavior', ea.STATE_SET, ['off', 'on', 'previous']),
+            tuya.exposes.countdown(), tuya.exposes.lightType(),
+        ],
+        meta: {
+            tuyaDatapoints: [
```
Koenkk, line 1455: Looks similar to `TS0601_dimmer_1`, please do the same check for this device.

src/devices/tuya.ts
cserem: All 4 switch / dimmer units in my PR are in-wall "brick" modules to connect to to an existing switch and lights. Backlight mode is not available. I guess the extra tuya DP (21) wouldn't cause issues, but the extra expose for backlight mode would confuse the end user. I haven't seen any devices where the behaviour (tuya datapoint and exposes) is "overwritten" in case of a specific sub-type of the device, this is why i added them as new devices.

src/devices/tuya.ts
cserem: Same as above.

src/devices/tuya.ts
Koenkk: I see, then this is fine!

src/devices/tuya.ts
```diff
@@ -2163,6 +2234,18 @@ const definitions: Definition[] = [
             await reporting.bind(device.getEndpoint(1), coordinatorEndpoint, ['genOnOff']);
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0001', ['_TZ3000_veu2v775']),
+        model: 'TS0001_switch_module_2',
+        vendor: 'TuYa',
+        description: '1 gang mini switch module',
+        whiteLabel: [{vendor: 'MOES', model: 'ZM-104-M'}],
```
Koenkk, line 2242: Could you move these and the device below to `moes.js` and remove the whiteLabel from it? (update the vendor/model accordingly)

src/devices/tuya.ts
```diff
@@ -2163,6 +2234,18 @@ const definitions: Definition[] = [
...
+        extend: tuya.extend.switch({switchType: true}),
```
Koenkk, line 2243: Can be merged with `TS0001_switch_module`? Adding the `tuya.whitelabel` should be enough

src/devices/tuya.ts
```diff
@@ -2227,6 +2310,23 @@ const definitions: Definition[] = [
             await reporting.bind(device.getEndpoint(2), coordinatorEndpoint, ['genOnOff']);
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0002', ['_TZ3000_qaa59zqd']),
+        model: 'TS0002_switch_module_4',
```
Koenkk, line 2315: Can be merged with TS0002_switch_module

src/devices/tuya.ts
```diff
@@ -3403,6 +3503,33 @@ const definitions: Definition[] = [
...
+            tuyaDatapoints: [
+                [1, 'smoke', tuya.valueConverter.trueFalse0],
+                [9, 'self_test', tuya.valueConverterBasic.lookup({'checking': 0, 'check_success': 1, 'check_failure': 2})],
+                [14, 'battery_state', tuya.valueConverter.batteryState],
+                [15, 'battery', tuya.valueConverter.raw],
+                [16, 'silence', tuya.valueConverter.onOff],
+            ],
+        },
+        whiteLabel: [
+            {vendor: 'Moes', model: 'ZSS-HM-SSD01'},
```
Koenkk, line 3530: Could you move this to `moes.js`?

cserem: I have tried my best, please check :)

Koenkk:
Thanks, could you also make a PR for the docs to add the pictures/

Koenkk:
yes!

PR #6498, opened 2023-11-17
Update yale.ts

ForrsITe:
Added the YRD652 Assure Lock SL which replaces the YRD256 Assure Lock.  The locks have the same functionality but the 652 is the upgraded version.


Koenkk:
Thanks!

PR #6504, opened 2023-11-18
Add support for _TZE204_e5m9c5hl

pantherale0:
Adds support for Wenzhi WZ-M100-Z human presence sensor - https://www.aliexpress.us/item/3256805013855490.html


pantherale0:
Just realised that fingerprint is already in

PR #6513, opened 2023-11-20
Nodon cable outlet: revert stop and anti-freeze values

PoloAlexis:
Revert stop and anti-freeze value
Tested on a heating device
Following https://github.com/Koenkk/zigbee-herdsman-converters/pull/6473


Koenkk:
thanks!

PR #6469, opened 2023-11-13
Sonoff TRV - valve attributes

photomoose:
Exposes the following diagnostic attributes, after confirmation from Sonoff support:

image
 

Issue tracked at: https://github.com/Koenkk/zigbee2mqtt/issues/19269

PR #6457, opened 2023-11-12
Extended settings for Aqara dual relay module T2 and Aqara LED strip T1

kirovilya:

PR #6519, opened 2023-11-20
feat: Extend generators

Koenkk:
**Problem**
Currently when e.g. adding support for a new attribute a new `toZigbee`, `fromZigbee` and `exposes` has to be created and added to the definition. When a lookup is used, it's values are also duplicate in `toZigbee`, `fromZigbee` and `exposes`. This creates a lot of code like https://github.com/Koenkk/zigbee-herdsman-converters/pull/6457 and https://github.com/Koenkk/zigbee-herdsman-converters/pull/6469.

**Solution**
Make extend more generic by adding extend generators, the extend generator generates the `toZigbee`, `fromZigbee` and `exposes` code. I've added an example for the Xiaomi switch type. Later we can also add a function which dynamically determines the extend for a certain device (e.g. an unsupported bulb).

Please review: @kirovilya @photomoose @sjorge @nurikk

### DIFF
--- a/src/lib/extend.ts
+++ b/src/lib/extend.ts
@@ -2,7 +2,10 @@ import * as exposes from './exposes';
 import tz from '../converters/toZigbee';
 import fz from '../converters/fromZigbee';
 import * as light from './light';
-import {Extend} from './types';
+import {Expose, Fz, Tz, Extend} from './types';
+import {Enum, access} from './exposes';
+import {KeyValue} from './types';
+import {getFromLookupByValue, isString, getFromLookup} from './utils';
 const e = exposes.presets;
 
 const extend = {
@@ -143,6 +146,38 @@ const extend = {
         }
 
         return result;
+    },
+
+    enumLookup: (args: {
+        name: string, lookup: KeyValue, cluster: string, attribute: string | {id: number, type: number}, description: string,
+        zigbeeOptions?: {manufacturerCode: number},
+    }): Extend => {
+        const {name, lookup, cluster, attribute, description, zigbeeOptions} = args;
+        const attributeKey = isString(attribute) ? attribute : attribute.id;
+        const expose: Expose = new Enum(name, access.ALL, Object.keys(lookup)).withDescription(description);
+        const fromZigbee: Fz.Converter[] = [{
+            cluster,
+            type: ['attributeReport', 'readResponse'],
+            convert: (model, msg, publish, options, meta) => {
+                if (attributeKey in msg.data) {
+                    return {[expose.property]: getFromLookupByValue(lookup, msg.data[attributeKey])};
+                }
+            },
+        }];
+        const toZigbee: Tz.Converter[] = [{
+            key: [name],
+            convertSet: async (entity, key, value, meta) => {
+                const payloadValue = getFromLookup(value, lookup);
+                const payload = isString(attribute) ? {[attribute]: payloadValue} : {[attribute.id]: {value: payloadValue, type: attribute.type}};
+                await entity.write(cluster, payload, zigbeeOptions);
+                return {state: {[key]: value}};
+            },
+            convertGet: async (entity, key, meta) => {
+                // @ts-expect-error TODO fix zh type
+                await entity.read(cluster, [attributeKey], zigbeeOptions);
+            },
+        }];
+        return {exposes: [expose], fromZigbee, toZigbee};
     },
 };
 
--- a/src/devices/xiaomi.ts
+++ b/src/devices/xiaomi.ts
@@ -10,6 +10,7 @@ const e = exposes.presets;
 const ea = exposes.access;
 import * as globalStore from '../lib/store';
 import * as xiaomi from '../lib/xiaomi';
+import xiaomiExtend2 from '../lib/extend/xiaomi';
 import * as utils from '../lib/utils';
 import {Definition, OnEvent, Fz, KeyValue, Tz, Extend} from '../lib/types';
 const {printNumbersAsHexSequence} = utils;
@@ -2559,6 +2560,7 @@ const definitions: Definition[] = [
         exposes: [e.switch().withEndpoint('l1'), e.switch().withEndpoint('l2'),
             e.power(), e.current(), e.energy(), e.voltage(), e.device_temperature(),
         ],
+        extend: [xiaomiExtend2.switchType],
         ota: ota.zigbeeOTA,
     },
     {
--- a/src/index.ts
+++ b/src/index.ts
@@ -61,23 +61,31 @@ function validateDefinition(definition: Definition) {
 
 function addDefinition(definition: Definition) {
     if ('extend' in definition) {
-        const {extend, ...definitionWithoutExtend} = definition;
-        if (extend.hasOwnProperty('configure') && extend.configure !== undefined && definition.hasOwnProperty('configure')) {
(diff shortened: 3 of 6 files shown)


sjorge:
I'll take a closer look tomorrow when I am more awake...

But would having 'extendable configures' also be something to incorporate into this?

Say for something to work you need to setup reporting, if you bundle it with the extend you can just stack them. And the device definition can then in some cases just be left out as there is nothing to go in there.

That might mean more 'fragmentarion' but it would be more stackable.

E.g. having the toZigbee/fromZigbee/configure fragment for example hvacThermostat.occupiedHeatingSetpoint all bundled.

Then you just make that as supported for a device and everything got that feature gets added.

Or is my half asleep brain missing/misunderstanding something?

kirovilya:
I agree that perhaps the best solution would be to operate entirely with the functionality element:
- cluster, attribute, type, restrictions, transformation/converters (in both directions), presentation/exposes, reporting, binding

Each device is filled with such functional elements, and then comes the interpretation of such a description.

Koenkk:
Yes that makes a lot of sense, the extend should also setup the reporting for it. We can even add the onEvent.

sjorge:
After looking over this now I am more awake, this seems fine and like a good idea still.

I suppose if we want to go very dynamic we could also give this a 'onProbe' function were we return true/false depending if this can be attached. And manual/during interview we can fire these off.

E.g. back to the hvacThermostat.occupiedHeatingSetpoint the probe function could test if it's readable/writable and depending on the error e.g. ATTRIBUTE_UNSUPPORT return false or true. It could also do things like figure out the lower and upper limit for the expose callback, these are stored in the minHeatSetpointLimit and maxHeatSetpointLimit attributes.

I guess this 'probe' functionality could be opt in (or auto enabled for unsupported devices), at least until we have eventually converted all devices too it.

Not sure how we would handle manifacturer specfic stuff and quirks though, I guess a blacklist of sort could work.

kirovilya:
image
or "Discover Attributes Extended Command" to undestand attribute access control

kirovilya:
we can store the answers to these queries in the database, just like we store available endpoints and clusters

Koenkk:
Yes that would be great! 😄

photomoose:
This certainly looks like it reduces the amount of boilerplate code in the device definition. Forgive my ignorance, I'm still fairly new to this repo - but in the case of #6469, would we be looking at some sort of `NumericLookup` equivalent to the `EnumLookup`? If these lookups are to be generic and used by multiple devices, then perhaps the `access` ought to be configurable as some attributes can be read-only (as in the case with some of the Sonoff attributes).

Koenkk:
@photomoose yes a numeric variant should also be added indeed, good point about the access! I will add it.

Koenkk:
This PR is now ready for testing

- @kirovilya could you rebase https://github.com/Koenkk/zigbee-herdsman-converters/pull/6457 and see if it works for that? (instead of the attributes converter)
- @photomoose could you rebase https://github.com/Koenkk/zigbee-herdsman-converters/pull/6469 to use this and test if it works? Example:

```js
extend: [
    extend.numeric({
        name: 'valve_closing_limit_voltage',
        cluster: 0xFC11,
        attribute: {id: 0x6007, type: 0x20},
        description: 'Valve motor running voltage',
        unit: 'mV',
        readOnly: true,
    }),
],
```

src/lib/extend.ts
```diff
@@ -146,5 +149,75 @@ const extend = {
...
+        zigbeeCommandOptions?: {manufacturerCode: number}, readOnly?: boolean, unit?: string,
+    }): ModernExtend => {
+        const {name, cluster, attribute, description, zigbeeCommandOptions} = args;
+        const attributeKey = isString(attribute) ? attribute : attribute.id;
+
+        let expose = new Numeric(name, args.readOnly ? access.STATE_GET : access.ALL).withDescription(description);
+        if (args.unit) expose = expose.withUnit(args.unit);
+
+        const fromZigbee: Fz.Converter[] = [{
+            cluster,
```
photomoose, line 195: For some reason (not sure why), if `cluster` is a number here, the inbound messages from the Sonoff TRV do not get processed.  If it is a string representation of the decimal cluster ID, however, then it works as expected. Any ideas why this might be the case?

For example, in `fzLocal`:

...the following works:

```
    child_lock: {
        cluster: '64529',
```
whereas the following does not:

```
    child_lock: {
        cluster: 0xFC11,
```

photomoose: If I change the `extend.numeric` to the following, then all the attributes in #6469 are populated correctly:

```        
const fromZigbee: Fz.Converter[] = [{
      cluster: cluster.toString(),
```

photomoose: If I use the string representation in the extend definition, as follows:

```
extend.numeric({
                name: 'idle_steps',
                cluster: '64529',
                attribute: {id: 0x6003, type: 0x21},
                description: 'Number of steps used for calibration (no-load steps)',
                readOnly: true,
            })
``` 
then I get the following error:

image

So I need to use a numeric cluster ID when sending messages to the device, but a decimal string when handling the incoming messages. (See my previous `'tzLocal` and `fzLocal` examples in `sonoff.ts`).

kirovilya:
@Koenkk [Made the first adaptation](https://github.com/kirovilya/zigbee-herdsman-converters/commit/2f719e024cb853a4bbc39783f211afee6bcaec1c) to the new ModernExtend.
Great idea, but needed some tweaking. For example, we need enpoints in extend - I tried to implement this in code. But two handlers appear for one event and they differ only in enpoint.
I also had to add a binary extend.

And then a difficulty arose for me - I need to combine the old extend with modern...
image

kirovilya:
@Koenkk I finished replacing “my attributes” with ModernExtend. Added "scale" to the numeric extend.
https://github.com/Koenkk/zigbee-herdsman-converters/pull/6554

src/lib/extend.ts
Koenkk: Fixed!

photomoose: I can confirm this now works. Raised #6559 which would allow me to move the frost protection into the modern numeric extend.

Koenkk: @photomoose for the value scaling, you can use the `scale` parameter which was added by @kirovilya  in https://github.com/Koenkk/zigbee-herdsman-converters/pull/6554

photomoose: Nice, `scale` will work in this case, I'll use that once the PR is merged.

photomoose:
@Koenkk @kirovilya I just tried using the `scale` functionality in the `numeric` extend. Looks like the result is not actually being used - I've raised #6564 as a fix.

src/lib/extend.ts
photomoose: Raised #6566 for some of the Sonoff attributes.

photomoose:
@Koenkk I need a switch if I want to move child lock and open window enable/disable functionality to modern extend. Would you like me to create a modern `switch` extend?

Koenkk:
@photomoose I will create a modern extend switch but it should be only used for ON/OFF switches (e.g. plugs/wall sockets). Could you use the `modernExtend.binary` for that?

sjorge:
Just curios, can we combine extends? Like switch + powerMeter to make a smart plug?

Koenkk:
@sjorge exactly! Which we also configure al the reporting/bindings 😄

PR #6526, opened 2023-11-21
Update tuya.ts for _TZE204_g2ki0ejr

xs400dohc:
If I changed 'preset' got an error that 'comf'. 'man', 'aut', 'hol' was expected. With the change no error-message appears and change of 'preset' is working. 4th attempt of pull-request. I stopped the other because they didn't passed the autotest, hope my first pull-requests made no trouble. I'm not very familiar with github and how it works.


Koenkk:
Thanks!

PR #6527, opened 2023-11-21
update `as` assertions to `satisfies`

Reinachan:
The `as` keyword in TypeScript asserts that "hey, I know that this type actually is what I'm saying it is. You can skip this and assume I'm right". However, this is usually a bad idea to do, as you can do unsafe assumptions leading to bugs. This is why the `satisfies` keyword was introduced. It makes sure that the object or array actually satisfies the type you want it to be.

In this PR, I'm changing all instances of the keyword `as` to `satisifies` where it makes sense to do so.

Beware, that while the first commit only includes uncontroversial changes (ergo, only type-level changes), some of the types I've changed locally don't actually satisfy the type it's asserted to, and in those cases, we have to figure out what needs changed. Those could be indicators of bugs.

### DIFF
--- a/src/devices/kurvia.ts
+++ b/src/devices/kurvia.ts
@@ -1,4 +1,4 @@
-import {Definition, Tz} from '../lib/types';
+import {Definition} from '../lib/types';
 import tz from '../converters/toZigbee';
 import extend from '../lib/extend';
 
@@ -11,8 +11,7 @@ const definitions: Definition[] = [
         vendor: 'KURVIA',
         description: 'GU10 GRBWC built from AliExpress',
         extend: extendData,
-        // FIXME: This doesn't satisfy the constraints of Tz.Converter[] and may be an unsafe assertion
-        toZigbee: ([tz.on_off] as Tz.Converter[]).concat(extendData.toZigbee),
+        toZigbee: [tz.on_off, ...extendData.toZigbee],
         meta: {applyRedFix: true, supportsEnhancedHue: false},
     },
 ];
--- a/src/devices/kurvia.ts
+++ b/src/devices/kurvia.ts
@@ -11,6 +11,7 @@ const definitions: Definition[] = [
         vendor: 'KURVIA',
         description: 'GU10 GRBWC built from AliExpress',
         extend: extendData,
+        // FIXME: This doesn't satisfy the constraints of Tz.Converter[] and may be an unsafe assertion
         toZigbee: ([tz.on_off] as Tz.Converter[]).concat(extendData.toZigbee),
         meta: {applyRedFix: true, supportsEnhancedHue: false},
     },
--- a/src/devices/sprut.ts
+++ b/src/devices/sprut.ts
@@ -93,9 +93,6 @@ const fzLocal = {
         },
     } satisfies Fz.Converter,
     co2_mh_z19b_config: {
-        // FIXME: There is no "key" on type Converter
-        // @ts-expect-error
-        key: ['co2_autocalibration', 'co2_manual_calibration'],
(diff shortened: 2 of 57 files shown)


Reinachan:
@Koenkk this is now close to complete, however, I'm having some issues converting the tuya files and don't know how to best handle them.

So, basically, the `Tz.Converter` type expects the `convertGet` method to return a `Promise<void>`, however, `entity.read()` returns a `Promise<KeyValue>`. These types are incompatible.

One potential solution would be to return void from said method, but what if there's anything that depend on the value existing?

Another would be to update the converter to accept a type `Promise<KeyValue>`, however, then you may have to narrow the type everywhere it's used by checking for a null.

So yeah, I'm not really sure how to best handle this.

The erroring files are `src/devices/tuya.ts` and `src/lib/tuya.ts`

Reinachan:
Also, for the record, whenever you get around to reviewing this, I suggest reading the commit messages, as I've noted down what changes I've made to the code that could *potentially* break stuff. They *shouldn't*, but as I don't know the source code as well as you do, I can't guarantee they won't.

src/devices/sprut.ts
```diff
@@ -63,36 +63,38 @@ const fzLocal = {
...
         type: ['readResponse', 'attributeReport'],
         convert: (model, msg, publish, options, meta) => {
             return {noise_detect_level: msg.data['noiseDetectLevel']};
         },
-    } as Fz.Converter,
+    } satisfies Fz.Converter,
     co2_mh_z19b_config: {
+        // FIXME: There is no "key" on type Converter
+        // @ts-expect-error
         key: ['co2_autocalibration', 'co2_manual_calibration'],
```
Koenkk, line 98: `key` is ignored for an Fz.Converter, it can be safely removed

src/devices/sprut.ts
```diff
@@ -104,8 +106,10 @@ const fzLocal = {
                 return {co2_manual_calibration: switchActionValues[msg.data['sprutCO2Calibration']]};
             }
         },
-    } as Fz.Converter,
+    } satisfies Fz.Converter,
     th_heater: {
+        // FIXME: There is no "key" on type Converter
+        // @ts-expect-error
         key: ['th_heater'],
```
Koenkk, line 113: Same here

src/devices/kurvia.ts
```diff
@@ -11,6 +11,7 @@ const definitions: Definition[] = [
         vendor: 'KURVIA',
         description: 'GU10 GRBWC built from AliExpress',
         extend: extendData,
+        // FIXME: This doesn't satisfy the constraints of Tz.Converter[] and may be an unsafe assertion
         toZigbee: ([tz.on_off] as Tz.Converter[]).concat(extendData.toZigbee),
```
Koenkk, line 15: Does `toZigbee: [tz.on_off, ...extendData.toZigbee],` work here?

Koenkk:
Thanks for the very clean PR, really appreciate it! Left 3 small comments.

src/devices/sprut.ts
Reinachan: Fixed

src/devices/kurvia.ts
Reinachan: That worked. Marking as resolved

Reinachan:
Fixed the issues you mentioned. I also figured out the issue with the Tuya files. They were using arrow functions, and arrow functions will by default return the value without needing to write `return`.

With that, this PR is ready for merge if you approve the latest changes :)

@Koenkk

Koenkk:
Many, many thanks! 😄

PR #6528, opened 2023-11-21
Update namron.ts - Added Namron 4512767

SelfDestructor:
This adds support for the Namron Zigbee Smart Plug 16A sort 4512767, the black version of 4512766.

There is no difference feature wise to the white model so I don't think there's a need to create a whole new device for it.

The model number shown will be wrong in the Z2M Devices list but in the about device page it shows the correct model.

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -923,7 +923,7 @@ const definitions: Definition[] = [
         },
     },
     {
-        zigbeeModel: ['4512766'],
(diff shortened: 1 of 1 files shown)


Koenkk:
This device is already supported https://github.com/Koenkk/zigbee-herdsman-converters/blob/0673b8adbd76a56e0154a8467fd4ae6eff014272/src/devices/namron.ts#L156

PR #6532, opened 2023-11-22
support easyiot ZB-IR01

xsp1989:
support easyiot ZB-IR01


src/lib/easyiot.ts
```diff
@@ -0,0 +1,51 @@
+import * as exposes from './exposes';
+import {Fz, Tz} from './types';
+const ea = exposes.access;
```
Koenkk, line 3: Since this will likely not be used from other files, please move everything into the `devices/easiot.ts` and remove this file.

src/lib/easyiot.ts
```diff
@@ -0,0 +1,51 @@
...
+                },
+                {disableDefaultResponse: true});
+            meta.logger.debug(`Sending IR command success.`);
+        },
+    } as Tz.Converter,
+};
+
+
+export const presetsEasyiot = {
+    ir_info: () => e.text('ir_info', ea.STATE).withDescription('ZB-IR01 is an infrared remote control with a local code library.'+
```
Koenkk, line 43: This is documentation and not an expose. Please create a PR for the Z2M docs instead.

src/lib/easyiot.ts
```diff
@@ -0,0 +1,51 @@
...
+        'If you need technical information, please send an email to support@easyiot.tech'),
+    ir01_recv_command: () => e.text('ir01_recv_command', ea.STATE).withDescription('Received infrared control command'),
```
Koenkk, line 45: I would call this `received_command`

src/lib/easyiot.ts
```diff
@@ -0,0 +1,51 @@
...
+    ir01_send_command: () => e.text('ir01_send_command', ea.SET).withDescription('Send infrared control command'),
```
Koenkk, line 46: I prefer to call this `send_command`

Koenkk:
Thanks, could you also make a PR for the docs + add a picture here?

xsp1989:
> Thanks, could you also make a PR for the docs + add a picture here?

Thank you for your support of the open source cause. I am planning to add docs, but this infrared remote control is relatively complex and supports many devices. I am considering whether to further develop zigbee-herdsman-converters, or take out the protocol documents first and wait for other Human supplementary functions. By the way, can I add a Google Cloud Drive link to the docs? If possible, I would first send the device's protocol document to Google Drive.

Koenkk:
Linking to any other docs from the device page is also fine. Also need a picture for this device

PR #6533, opened 2023-11-22
Add more dp's to Avatto ZWT198 and split and check the schedule format 

marcelhoogantink:
Hello,

Here my PR for a further improvement of the Avatto ZWT198 thermostat.


src/devices/tuya.ts
```diff
@@ -20,8 +20,100 @@ const fzZosung = zosung.fzZosung;
...
+        }
+    }
+    return message;
+}
 
 const tzLocal = {
+
+    // Since some keys needed, are not defined in datapoints, the <datapoints> is extended here with needed datapoints
+
+    ZWT198_datapoints:
```
Koenkk, line 70: Please add the missing keys to the `tuya.datapoints` converter

src/devices/tuya.ts
```diff
@@ -4232,6 +4391,131 @@ const definitions: Definition[] = [
...
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_viy9ihs7']),
+
+        model: 'ZWT198 Avatto',
+        vendor: 'TuYa',
+        description: 'Avatto ZWT198 Wall Thermostat',
+        onEvent: tuya.onEvent({timeStart: '1970'}),
+
+        fromZigbee: [
+            tuya.fz.datapoints,
+            fzLocal.ZWT198_tuya_thermostat_schedule,
```
Koenkk, line 4408: Why is this needed? This should go via `tuya.fz.datapoints` instead

src/devices/tuya.ts
```diff
@@ -4166,6 +4324,7 @@ const definitions: Definition[] = [
         ],
         onEvent: tuya.onEventSetTime,
     },
+    /* OLD ONNDO defintion
```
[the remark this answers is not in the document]
marcelhoogantink: Done

src/devices/tuya.ts
marcelhoogantink: Changed key to an existing key in tuya.datapoints

src/devices/tuya.ts
marcelhoogantink: datapoint dp109 results in 4 different states, due to splitting up the original schedule into two parts (workdays and holidays schedule, and two syntax check states, displaying possible syntax error for these schedules. I did not see a possibility to do this within` tuya.fz.datapoints`. If there is, I am open to it. Please advice!

Koenkk: To return multiple properties at once, see https://github.com/Koenkk/zigbee-herdsman-converters/blob/3a10293dcd0563345b33849b3a45889312af0c3e/src/lib/tuya.ts#L515

marcelhoogantink:
I have (hopefully) a last question. When checking the schedule-format, they can be wrong. When the format is wrong, I do not send it to the device, but I like to update some state settings (to display the message). Is it possible to set/update/change (multiple) state-values (in state.json) without sending to the device ?
I tried with change the meta.state but that only seems to change on set not on change?


```
            meta.state.schedule = {
                schedule_workdays: valueWorkdays,
                schedule_holidays: valueHolidays,
                message_workdays: (resultWorkdays.length ? resultWorkdays : 'Workdays schedule format OK'),
                message_holidays: (resultHolidays.length ? resultHolidays : 'Holidays schedule format OK')
            };
```

Koenkk:
In that case you should do a `throw new Error('format is invalid')` (or with a better message)

marcelhoogantink:
I understand, but in the end I use Home Assistant, and I wondered how to get such an error visible in HA.
So thought I do it with some extra state-items. Do you have any clue for this, when not, I will use the error-option.

Koenkk:
We don't have a mechanism for this in place, the error is the best option for now.

marcelhoogantink:
Ok, but I got it working by bypassing the datapoints.
You asked to use these datapoints, is that mandatory?

https://github.com/Koenkk/zigbee-herdsman-converters/pull/6533#discussion_r1402645595

marcelhoogantink:
So there is the answer on your question, why I needed this and that I didn't use the datapoints way here.
I hope you accept this solution.

Koenkk:
@marcelhoogantink yes please use the `tuya.fz.datapoints`, the logic in `ZWT198_tuya_thermostat_schedule` can be moved to a custom value converter in `lib/tuya.ts`

PR #6541, opened 2023-11-22
Add support for Iluminize 5121.10

Ru4rc:
Add support for Rotary dimmer with integrated Zigbee 3.0 dimming actuator, this time with correct state reports


Koenkk:
Thanks!

PR #6546, opened 2023-11-23
Fix JetHome WS7 template. Add OTA support.

adeepn:
Update converter for JetHome WS7:
- fix actions
- add OTA support


Koenkk:
Thanks!

PR #6547, opened 2023-11-23
Update philips.ts

cflints:
Added LWA024


Koenkk:
This device is already supported; https://github.com/Koenkk/zigbee-herdsman-converters/blob/6569bad47a8b2c3b90b67f1cde1090d3a5509bb4/src/devices/philips.ts#L933

cflints:
That's the a60, mine was the b22 that came up unsupported.

________________________________

Koenkk:
Then update the description of the existing definition instead.

PR #6552, opened 2023-11-24
Fix Tuya TS0001 _TZ3000_skueekg3 detected as TS0001

pannal:
TS000F with this manufacturerName was already in the WHD02 list. I've just encountered another one with TS0001 as model. Ref: https://github.com/Koenkk/zigbee2mqtt/issues/12005#issuecomment-1824541797


Koenkk:
Thanks!

PR #6553, opened 2023-11-24
fix: Update configuration for Vesternet Zigbee devices

martyn-vesternet:
Split configuration based on firmware versions for battery powered devices, since later versions report 0 - 200 rather than 0 - 100 for battery percentage.

Split configuration based on firmware versions for 2 channel switch device, since later version report measurements on a different endpoint to the earlier version.

### DIFF
--- a/src/devices/vesternet.ts
+++ b/src/devices/vesternet.ts
@@ -69,7 +69,7 @@ const definitions: Definition[] = [
         },
     },
     {
-        fingerprint: [{modelID: 'ZGRC-KEY-013', softwareBuildID: '2.5.3_r20'}, {modelID: 'ZGRC-KEY-013', softwareBuildID: '2.7.6_r25'}],
+        fingerprint: [{modelID: 'ZGRC-KEY-013', softwareBuildID: '2.5.3_r20'}],
         model: 'VES-ZB-REM-013',
         vendor: 'Vesternet',
         description: 'Zigbee remote control - 12 button',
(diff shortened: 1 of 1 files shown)


martyn-vesternet:
@Koenkk this PR is failing checks, is it not possible to have multiple configurations where the model field is the same but specific to different firmware versions?

I could amend the model field with something to make it unique, but it does seem a valid case for the model to be the same?

Koenkk:
I propose to create a custom battery converter instead, we could check for the software build ID there and divide correctly. This will prevent a lot of code duplication (and will also fix the duplicate model issue)

martyn-vesternet:
@Koenkk can I simply return true / false for dontDividePercentage using "something" in the meta section instead?

I was looking for inspiration and found this for example:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/f9025cf4dbbba573e220f0187993967ee89cd78d/src/devices/m%C3%BCller_licht.ts#L79


I'm not sure on the syntax, but could I do something like:

`meta: {battery: {dontDividePercentage: (device: Zh.Device) => device.softwareBuildID == '2.5.3_r20' ? true : false}}`

That would seem much simpler?

Koenkk:
Agree, but then also `fromZigbee.ts`/`battery` has to be updated

martyn-vesternet:
@Koenkk why would that need updating?

I think I just need to be able to populate dontDividePercentage via a check on the softwareBuildID ?

Koenkk:
A check needs to be added to check if is a function or static value here:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/6e0e6990e2f08324f7aee6b90fe0799d0552f1dd/src/converters/fromZigbee.ts#L356C23-L356C43 

Otherwise the function is not executed.

martyn-vesternet:
I don't think I've explained it very well, but the value for dontDividePercentage would be set from the inline function in the meta section rather than wanting "dontDividePercentage" to be a function itself.

Like this one does for "supportsEnhanced":

 https://github.com/Koenkk/zigbee-herdsman-converters/blob/f9025cf4dbbba573e220f0187993967ee89cd78d/src/devices/m%C3%BCller_licht.ts#L79

Have I misunderstood what that does?  It seems to set "supportsEnhanced" based on that inline function after it?

If that's right, then to allow the same for dontDividePercentage, I think I'd need to update this:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/87d635954e10a43e6210a5728e5c75bc4422d62d/src/lib/types.ts#L65

To be something like:

`battery?: {voltageToPercentage?: string | {min: number, max: number}, dontDividePercentage?: boolean | ((device: Zh.Device) => boolean)},`

Which would then allow:

`meta: {battery: {dontDividePercentage: (device: Zh.Device) => device.softwareBuildID == '2.5.3_r20' ? true : false}}`


Or am I mistaken?

Koenkk:
Yes that's correct! But the `fz.battery` converter also has to be updated to execute the function.

```js
// This
const dontDividePercentage = model.meta && model.meta.battery && model.meta.battery.dontDividePercentage;
// Should be changed to this:
const dontDividePercentage = utils.getMetaValue(msg.endpoint, model, 'battery');
```

Then the `battery` can be made a function, e.g. `battery: (entity: Zh.Endpoint) => ({dontDividePercentage: false})`

martyn-vesternet:
@devildant I've finally been allocated some time to look at this again.

The original idea above was to somehow set dontDividePercentage during pairing, based on the firmware version reported from the device, since that would seem to be the simplest route to take.

After going down a bit of a rabbit hole with that though and not getting very far, I've reverted to simply processing the battery reports in a custom battery converter.

There's an external converter at https://github.com/vesternet/zigbee2mqtt-herdsman-converters which should do just that and it seems to work OK for me on the two versions of test device that I have here.  Perhaps you can check it on yours?

devildant:
> @devildant I've finally been allocated some time to look at this again.
> 
> The original idea above was to somehow set dontDividePercentage during pairing, based on the firmware version reported from the device, since that would seem to be the simplest route to take.
> 
> After going down a bit of a rabbit hole with that though and not getting very far, I've reverted to simply processing the battery reports in a custom battery converter.
> 
> There's an external converter at https://github.com/vesternet/zigbee2mqtt-herdsman-converters which should do just that and it seems to work OK for me on the two versions of test device that I have here. Perhaps you can check it on yours?

Hi,

I don't have time to test at the moment, but I see the code and for me it's OK for 015.. I abandoned the 004, because I changed my solution

PR #6558, opened 2023-11-24
UNSUPPORTED_ATTRIBUTE

ashottonoyan:
Keeps throwing error and spamming the network


Koenkk:
Is it also removed from the exposes?

ashottonoyan:
Actually no, I'm assuming since it's extending xiaomi, needs to be disabled explicitly. Not sure how to do that.

Koenkk:
Use `extend: xiaomiExtend.light_onoff_brightness_colortemp({disableSwitchPowerOutageMemory: true}),`

ashottonoyan:
I think it's rather `disablePowerOnBehavior: true`. Pushed.

Koenkk:
Thanks!

PR #6559, opened 2023-11-25
More args to numeric extend

photomoose:
We could add more functionality to the modern numeric extend to allow for something like this:

```
extend.numeric({
    name: 'frost_protection_temperature',
    cluster: 0xFC11,
    attribute: {id: 0x6002, type: 0x29},
    description: 'Minimum temperature at which to automatically turn on the radiator, if system mode is ' +
        'off, to prevent pipes freezing.',
    min: 4.0,
    max: 35.0,
    step: 0.5,
    unit: '°C',
    fromZigbee: (value) => value / 100,
    toZigbee: (value) => utils.toNumber(value) * 100,
}),
```

### DIFF
--- a/src/lib/extend.ts
+++ b/src/lib/extend.ts
@@ -194,20 +194,24 @@ const modernExtend = {
     },
     numeric: (args: {
         name: string, cluster: string | number, attribute: string | {id: number, type: number}, description: string,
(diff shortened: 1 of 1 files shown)


photomoose:
Closing in favour of #6554

PR #6560, opened 2023-11-25
Update tuya.ts

cruelacid:
Add support for Moes Zigbee Smart Dimmer Switch Module Timer Diy Breaker 2 Gang.


Koenkk:
This device is already supported: https://github.com/Koenkk/zigbee-herdsman-converters/blob/a93da0b31a4cc29a601150220c0460ceb1a1b440/src/devices/tuya.ts#L1400

PR #6561, opened 2023-11-25
Update namron.ts

Aeliassen:
Added support for 4512768 (I wrote a code for this, but when trying to upload it I found this code in an issue where someone asked for it to be added.)

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -942,6 +942,36 @@ const definitions: Definition[] = [
             await reporting.activePower(endpoint);
         },
     },
+    {
+        zigbeeModel: ['4512768'],
+        model: '4512768',
+        vendor: 'Namron',
+        description: 'Zigbee 2 channel switch Namron',
+        fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fz.power_on_behavior, fz.ignore_genOta,],
+        toZigbee: [tz.on_off, tz.power_on_behavior],
+        exposes: [e.switch().withEndpoint('l1'), e.switch().withEndpoint('l2'),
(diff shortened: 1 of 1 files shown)


src/devices/namron.ts
```diff
@@ -942,6 +942,36 @@ const definitions: Definition[] = [
...
+            e.numeric('power_l2', ea.STATE).withUnit('W').withDescription('Phase 2 power')],
+        endpoint: (device) => {
+            return {'l1': 1, 'l2': 2};
+        },
+        meta: {multiEndpoint: true, publishDuplicateTransaction: true},
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint1 = device.getEndpoint(1);
+            const endpoint2 = device.getEndpoint(2);
+            await reporting.bind(endpoint1, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement', 'seMetering']);
+            await reporting.bind(endpoint2, coordinatorEndpoint, ['genOnOff']);
```
Koenkk, line 969: Since you are only binding endpoint 1, do energy measurements work for both endpoints?

Aeliassen:
I'm sotty, I'm new to Github, i seemed to close the comment and can't reply on it anymore. Yes for some reason energy measurement works for both endpoints even if just one is called.

src/devices/namron.ts
Aeliassen: Yes, for some reason it works for both endpoints. 
Screenshot_2023-11-26-09-37-56-721_io homeassistant companion android

Koenkk:
Thanks!

PR #6564, opened 2023-11-26
Actually use the modern extend scaled value

photomoose:
When `scale` is applied to the value received from zigbee, the result is not actually used and the original value is reported in the UI. This fixes it.


Koenkk:
Thanks!

PR #6566, opened 2023-11-26
Make use of modern extend for Sonoff TRV

photomoose:
Refactor some of the Sonoff TRV attributes to make use of modern extend.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -65,6 +65,20 @@ const fzLocal = {
             return result;
         },
     } satisfies Fz.Converter,
+    valve_motor_running_voltage: {
+        cluster: '64529',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            const result: KeyValueAny = {};
+            const data = msg.data;
+
+            if (data.hasOwnProperty(0x6007)) {
+                result.valve_motor_running_voltage = data[0x6007] / 1000;
+            }
+
+            return result;
+        },
+    } satisfies Fz.Converter,
 };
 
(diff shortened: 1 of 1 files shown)


src/devices/sonoff.ts
```diff
@@ -3,8 +3,8 @@ import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as constants from '../lib/constants';
 import * as reporting from '../lib/reporting';
-import * as utils from '../lib/utils';
-import extend from '../lib/extend';
+import legacyExtend from '../lib/extend';
```
Koenkk, line 6:
```suggestion
import extend from '../lib/extend';
```

src/devices/sonoff.ts
```diff
@@ -467,18 +439,74 @@ const definitions: Definition[] = [
...
+            tz.thermostat_running_state,
+            tzLocal.child_lock,
+            tzLocal.open_window,
+        ],
+        extend: [
+            extend.numeric({
+                name: 'frost_protection_temperature',
+                cluster: 0xFC11,
+                attribute: {id: 0x6002, type: 0x29},
+                description: 'XXX Minimum temperature at which to automatically turn on the radiator, ' +
```
Koenkk, line 463: I don't understand why you added `XXX` here.

photomoose: Ha ha, good spot. I temporarily put the 'XXX' in there as a sanity check whilst testing to ensure the UI was rendering the modern expose and not the legacy expose. I forgot to remove it when committing last night.😴

Koenkk:
Thanks!

PR #6568, opened 2023-11-26
Add support for Namron kanalbryter 4512772/4512773

geirra:
This adds support for the Namron Kanalbryter in white(4512772) and black(4512773)


src/devices/namron.ts
```diff
@@ -942,6 +942,24 @@ const definitions: Definition[] = [
             await reporting.activePower(endpoint);
         },
     },
+    {
+        zigbeeModel: ['4512772', '4512773'],
+        model: '451277(2/3)',
```
Koenkk, line 947: Could you not use `(2/3)` here but use a whitelabel for the other device instead? Like: 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/d0aafe21dfadd399a98cc9176d2276d318cbf2a7/src/devices/namron.ts#L917

geirra: Shure, hopefully this looks better now ?

src/devices/namron.ts
```diff
@@ -942,6 +942,25 @@ const definitions: Definition[] = [
             await reporting.activePower(endpoint);
         },
     },
+    {
+        zigbeeModel: ['4512772', '4512773'],
+        model: '4512773',
+        vendor: 'Namron',
+        description: 'Zigbee Kanalbryter Sort',
```
Koenkk, line 949: Please make the descriptions English

geirra: Sorry about that, i used the product name in this case.
should be fixed now

Koenkk:
Thanks!

PR #6571, opened 2023-11-26
Second try, something went wrong on the first:Added New Avatto ZWT198 updat including scedule, reset and brightness…

marcelhoogantink:
Added New Avatto ZWT198 updat including scedule, reset and brightness in devices/tuya.ts;  lib/exposes.ts end lib.tuya.ts

Hello,

Here my second try with a PR for a further improvement of the Avatto ZWT198 thermostat. 
The first one went wrong in one or another way....

So here is my new try.
All code is in the right place as you asked.
And I got the format message-option running within the tuya.tz.datapoints.

Just 2 remarks:

1) 
I used `composite `for the `schedule`, (combining `schedule_weekday`, `schedule_holiday`, `message_weekday` and `message_holiday`) and I noticed that the text fields (`schedule_weekday `and `schedule_holiday`) are not directly updated after the device send the info back on a change. In the `state.json` file the info from the device is correctly displayed. Just after a page renew, the returned data is shown correctly. Before, without the `composite ` it was OK. So I think `composite `components are not well updated on screen. Should I create a issue for this?

2)
Having it all working within Zigbee2mqtt, the data is not well handled Home Assistant:
(see below the `state.json` file from zigbee2mqtt)
```
{
    "child_lock": "UNLOCK",
    "linkquality": 150,
    "local_temperature": 20.5,
    "factory_reset": "OFF",
    "system_mode": "off",
    "current_heating_setpoint": 22,
    "preset": "manual",
    "max_temperature_limit": 60,
    "local_temperature_calibration": 0,
    "running_state": "idle",
    "frost_protection": "ON",
    "working_day": "6-1",
    "deadzone_temperature": 1,
    "schedule": {
        "message_holiday": "Holidays schedule format OK",
        "message_weekday": "Workdays schedule format OK",
        "schedule_holiday": "08:00/22.0°C 23:00/16.0°C",
        "schedule_weekday": "06:00/20.0°C 08:00/16.0°C 11:30/16.0°C 12:30/16.0°C 17:00/22.0°C 22:00/16.0°C"
    },
    "backlight_mode": "high"
}
```
The `schedule `part in Home Assistant is shown as a json-string with the four items. Can that be converted into the four separate items? 
If yes, how do I manage that?

regards

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -511,121 +511,6 @@ const fzLocal = {
 
 const definitions: Definition[] = [
     {
-        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_viy9ihs7']),
-
-        model: 'ZWT198 Avatto',
-        vendor: 'TuYa',
-        description: 'Avatto ZWT198 Wall Thermostat',
-        onEvent: tuya.onEvent({timeStart: '1970'}),
-
-        fromZigbee: [
-            tuya.fz.datapoints,
-        ],
-        toZigbee: [
-            tuya.tz.datapoints,
-        ],
-        configure: tuya.configureMagicPacket,
-        exposes: [
-            e.binary('factory_reset', ea.STATE_SET, 'ON', 'OFF')
-                .withDescription('Full factory reset, use with caution !'),
-            e.child_lock(),
-            // e.battery(), // device seems not to give battery-info ..??
-            e.climate()
-                .withSystemMode(['off', 'heat'], ea.STATE_SET)
-                .withPreset(['auto', 'manual', 'temporary_manual'])
-                .withSetpoint('current_heating_setpoint', 5, 35, 0.5, ea.STATE_SET)
-                .withRunningState(['idle', 'heat'], ea.STATE)
-                .withLocalTemperature(ea.STATE)
-                .withLocalTemperatureCalibration(-9.9, 9.9, 0.1, ea.STATE_SET),
-
-            e.binary('frost_protection', ea.STATE_SET, 'ON', 'OFF')
-                .withDescription('Antifreeze function'),
-
-            e.max_temperature_limit()
-                .withUnit('°C')
-                .withValueMin(15)
-                .withValueMax(90)
-                .withValueStep(0.5)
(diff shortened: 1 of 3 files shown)


Koenkk:
> So in the end, I have to get rid of the syntax-messages to get in into Home Assistant?

I propose so, error messages should also not be part of the state, use `throw new Error('my error')` for the error messages.

marcelhoogantink:
Composite() is removed, made erors of the messages. This is were my final changes for this device.

src/devices/tuya.ts
```diff
@@ -4167,65 +4167,97 @@ const definitions: Definition[] = [
...
+                .withValueMin(0.5)
+                .withValueStep(0.5)
+                .withPreset('default', 1, 'Default value')
+                .withDescription('The delta between local_temperature (5<t<35)and current_heating_setpoint to trigger Heat'),
+            e.enum('backlight_mode', ea.STATE_SET, ['off', 'low', 'medium', 'high'])
+                .withDescription('Intensity of the backlight'),
+            e.enum('working_day', ea.STATE_SET, ['disabled', '6-1', '5-2', '7'])
+                .withDescription('Workday setting'),
+            e.text('schedule_weekday', ea.STATE_SET).withDescription('Workdays (6 times `hh:mm/cc.c°C`)'),
+            e.text('schedule_holiday', ea.STATE_SET).withDescription('Holidays (2 times `hh:mm/cc.c°C)`'),
```
Koenkk, line 4207: I already merged this to be in time for the 1 December release, but I think it will be good to add an example to the description, e.g.

```
e.text('schedule_holiday', ea.STATE_SET).withDescription('Holidays (2 times `hh:mm/cc.c°C`, e.g. `02:04/18.0°C,18:20/19.0°C`)'),
```

marcelhoogantink: I will ..
Tested it with ZIgbee2MQTT-edge.
It seems you missed somehow some changes in /lib/tuya.ts:

image

theré I add a datapoint key 'schedule_holiday' and in the merge (and the Edge-version that's missing

Koenkk: fixed!

PR #6573, opened 2023-11-26
Namron 4512761

enrawi:
Add support for namron 4512761

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -983,9 +983,8 @@ const definitions: Definition[] = [
         configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['genBasic', 'genOnOff', 'haElectricalMeasurement', 'seMetering']);
-            await endpoint.read('haElectricalMeasurement', ['acVoltageMultiplier', 'acVoltageDivisor']);
-            await endpoint.read('haElectricalMeasurement', ['acPowerMultiplier', 'acPowerDivisor']);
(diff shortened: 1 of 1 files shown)


src/devices/namron.ts
```diff
@@ -972,6 +972,24 @@ const definitions: Definition[] = [
...
+        model: '4512761',
+        vendor: 'Namron',
+        description: 'Namron ZigBee rele 16A',
+        fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fz.power_on_behavior],
+        toZigbee: [tz.on_off, tz.power_on_behavior],
+        exposes: [e.switch(), e.power(), e.current(), e.voltage(), e.energy(), e.power_on_behavior()],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genBasic', 'genOnOff', 'haElectricalMeasurement', 'seMetering']);
+            await endpoint.read('haElectricalMeasurement', ['acVoltageMultiplier', 'acVoltageDivisor']);
```
thordy, line 986: I would suggest switching to use the convenience method `readEletricalMeasurementMultiplierDivisors` instead of reading each of these. 
This also doesn't report any power consumption (for me at least), so you should also add `readMeteringMultiplierDivisor`, leaving the final configure as 
```` 
        await reporting.bind(endpoint, coordinatorEndpoint, ['genBasic', 'genOnOff', 'haElectricalMeasurement', 'seMetering']);
    },
```` 

That successfully reports `Power`, `Voltage`, `Current`, and `Energy`.

@enrawi will you update your PR, or want me to submit a new one?

thordy (commented):
Added some comments regarding `configure` which doesn't correctly report `energy` consumption

Koenkk:
Thanks!

PR #6580, opened 2023-11-27
Sonoff TRV: populate attributes during configuration

photomoose:
Populate the exposed attributes during device configuration.


Koenkk:
Thanks!

PR #6584, opened 2023-11-29
Develco SPLZB-131: support for device_temperature attr

emandtf:
I have 4 Develco SPLZB-131 but few of them seems don't have this value.
This behaviour seems firmware indipendant (it uses dates and not version numbers) as following table:

Firmware date = HAS_TEMPERATURE?
=======================

- 2018-01-16 16:06 = YES
- 2018-01-16 23:19 = NO!!
- 2021-06-08 03:51 = NO!!
- 2022-03-11 19:21 = YES

Eventually the "device_temperature" will be "null" when not supported........


Koenkk:
Thanks!

PR #6594, opened 2023-12-01
Update orvibo.ts to support ZL1-EN

btittelbach:
The [Orvibo SW30](https://www.zigbee2mqtt.io/devices/SW30.html) is also sold as `ZL1-EN` by IMOU and still manufactured by `Hangzhou Huacheng Network Technology Co.,Ltd.`

It looks exactly the same as the SW30, except that the label `Orvibo` is replaced by `IMOU`.
It works perfectly with the exact same device definition as the SW30.

It advertises itself as zigbeeModel `ZL1-EN` by manufacturer `MultIR`, and interestingly `MultiIR`was already listed as a possible zigbeeModel for the device.

### DIFF
--- a/src/devices/orvibo.ts
+++ b/src/devices/orvibo.ts
@@ -310,7 +310,7 @@ const definitions: Definition[] = [
         exposes: [e.water_leak(), e.battery_low(), e.tamper()],
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #6595, opened 2023-12-01
add some private function in sonoff devices

liangjia2019:
1.SNZB-03P add battery,occupancy_ult_timeout
2.SNZB-04P add tamper
3.SNZB-06P add occupancy_ult_timeout,occupancy_ult_sensitivity, illumination.


Koenkk:
Could you use the `modernExtend` instead? This replaces `fromZigbee`/`toZigbee` and `exposes, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/77da13d3db98de610ccc16d627cbbcf11b88ab7e/src/devices/xiaomi.ts#L1028

liangjia2019:
> Could you use the `modernExtend` instead? This replaces `fromZigbee`/`toZigbee` and `exposes, example:
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/77da13d3db98de610ccc16d627cbbcf11b88ab7e/src/devices/xiaomi.ts#L1028
Of course, if you wish me to do so, I will follow it myself

liangjia2019:
> I will follow it myself

Also, I would like to know when the next version of Zigbee2mqtt will be planned

Koenkk:
Great, the next release is on 1 January.

Koenkk:
> If you use get method, there will be a read button on the front-end, but most of the time it will not succeed because the device is in sleep mode

This is the case for all end devices, document this in the `description` instead.

PR #6602, opened 2023-12-02
M elec ml st r200

benswinney:
Updated code so that ML-ST-R200 is seen as a Light entity / device, instead of a switch within HA.

Also included code, so that the light switch position is reported is toggled manually.


benswinney:
Apologies, I didn't see that  ... could have saved myself some time haha 😄 

Closing this PR as it's no longer needed.

PR #6603, opened 2023-12-02
SPLZB-131 Alternative Configure

mhetzi:
Hi,
The old report settings do flood my network quite bad.
I changed these values via the webinterface and they work better.


Koenkk:
Thanks!

PR #6607, opened 2023-12-02
Improve TS0601_din_1 and correct Sunricher/Vesternet white-label

dduransseau:
- Add reactive power, energy fields to Tuya TS0601_din_1
- Correct Zemismart brand for SPM01-D2TZ (manufactured by bituo-technik)
- Reverse the Vesternet/Sunricher manufacturer for SR-ZG9101SAC-HP-SWITCH-B and SR-ZG9101SAC-HP-SWITCH-2CH


dduransseau:
Sure, @martyn-vesternet do you agree that both:
https://www.sunricher.com/16-amps-zigbee-relay-on-off-light-device-sr-zg9101sac-hp-switch-b.html
https://www.sunricher.com/zigbee-ac-2-gang-in-wall-switch-sr-zg9101sac-hp-switch-2ch.html
Are the main device and are sold by Vesternet as white label as (VES-ZB-HLD-017 and VES-ZB-SWI-015) ?

martyn-vesternet:
@dduransseau @Koenkk 

Vesternet sells various devices from various manufacturers (Sunricher, HZC, etc) and these are branded as Vesternet.

I'd prefer to keep the configurations separate in our own files so that they do not stomp on other configurations, either for Sunricher direct or for other vendors that choose to re-brand them (for example ROBB, Samotech, etc).  This also hopefully prevents others from making changes that might potentially break existing Vesternet devices.

Merging everything into the main Sunricher will be problematical IMO, because I personally only test against the device versions that Vesternet sell and these might not have the same functionality as others, even though the reported model number might be the same.

Where possible I try to constrain the Vesternet configuration files to fingerprint match on the model number and the specific firmware versions that I have tested, or firmware versions that I have been advised by people that have purchased the Vesternet devices.

If there is a better way to handle this then please let me know?

martyn-vesternet:
Yes unfortunately these "rebranded" devices often still have the same "modelID" which is difficult to account for.  I've been using the "softwareBuildId" to try and constrain the configurations to only match the firmware versions of devices that Vesternet has definitely sold.

Your PR for the 2-channel switch seems to be trying to match this:

`{modelID: 'ON/OFF(2CH)', softwareBuildID: '2.9.2_r54'}`

I've only seen two firmware versions for the Vesternet version of that device:

`{modelID: 'ON/OFF(2CH)', softwareBuildID: '2.5.3_r2'}, {modelID: 'ON/OFF(2CH)', softwareBuildID: '2.9.2_r3'}`

So unless you are certain that there exists a Vesternet version of that device with 2.9.2_r54 firmware I would rather leave our configuration file as-is (especially as there is a pending PR that accounts for differing behaviours between versions 2.5.3_r2 and 2.9.2_r3).

For your PR for the High Load Switch device, I don't see what that's trying to achieve other than to change the model / vendor around the configuration appears identical, is there something there that's changed?

From Vesternets' point of view the vendor is Vesternet and the model is VES-ZB-HLD-017 because that's what's physically on the device / box / manual.  I would prefer to leave that as-is, because future updates are easier in one place rather than across different files.



Of course, one option for the Vesternet devices might be to revert back to supplying custom converters via our own Github like we do for other platforms (Hubitat for example).  This way we just have to tell Vesternet customers to use those instead, from my side it's much less work but from the customer side it's a little less integrated.

martyn-vesternet:
I would suggest creating your own dedicated device in the Sunricher file to suit that particular firmware version.

You'll want to make sure that the configuration for the Vesternet version works correctly with your version, especially as there is a pending PR that accounts for differing behaviours between versions 2.5.3_r2 and 2.9.2_r3.

PR #6617, opened 2023-12-04
add pulse configuration for Frient Electricity Meter Interface 2

ywaf:
should fix https://github.com/Koenkk/zigbee-herdsman-converters/commit/b7b678862a3871b53606022daf6bd1e1ad8f3e06#commitcomment-133972455

(note: ive never written a converter, might be wrong way to do it)


src/devices/frient.ts
```diff
@@ -1,18 +1,61 @@
...
         zigbeeModel: ['EMIZB-141'],
         model: 'EMIZB-141',
         vendor: 'frient',
         description: 'Smart powermeter Zigbee bridge',
-        fromZigbee: [fz.metering, fz.battery],
-        toZigbee: [],
-        exposes: [e.battery(), e.power(), e.energy()],
+        fromZigbee: [fz.metering, fz.battery, frient.fz.pulse_configuration],
+        toZigbee: [frient.tz.pulse_configuration],
+        exposes: [
```
Koenkk, line 52: Could you use the modernExtend for this instead? Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/68eee46b7c3fbd79a08d90124c94c368de15041f/src/devices/xiaomi.ts#L987

dawiinci:
Until this is working what is the workaround for this? I can't get the energy readings. 
Thanks

ywaf:
Sorry for being inactive here, been a bit busy, is there any docs for this "modernExtend"? I'm complete shit at javascript so not 100% sure exactly what needs to be modified :)

Koenkk:
Not yet because it is extremely new (2 weeks old now) but basically you can drop the expose and to/fromZigbee converters when using:

```js
extend: [
            numeric({
                name: 'pulse_configuration',
                valueMin: 0,
                valueMax: 65535,
                valueStep: 1,
                scale: 1,
                unit: '',
                cluster: 'seMetering',
                attribute: 'develcoPulseConfiguration',
                description: 'Pulses per kwh. Default 1000 imp/kWh. Range 0 to 65535',
            }),
]

TheMugge:
If I would like to contribute to the creation of this updated converter for the Meter interface, is this possible? How can I test the converter before pushing? I've tried to add it as an external converter in zigbee2mqtt (in home assistant) but since there is already a converter for the meter it seems zigbee2mqtt always uses the internal converter over the external? Do I need to recompile zigbee2mqtt to test the changes in this PR?

Koenkk:
> but since there is already a converter for the meter it seems zigbee2mqtt always uses the internal converter over the external?

Z2M always prefers external converters over internal ones.

dawiinci:
I can confirm this. Maybe there is a typo in the config?

Hopefully this PR will he released soon, so that might fix it for you as well.

PR #6622, opened 2023-12-04
Adds calibration offsets for tilt on DJT11LM sensor

ghoz:
the DJT11LM seems to be very poorly calibrated,
* provide raw accelerometer values : raw_x, raw_y and raw_z used for calibration after calibration can be removed from the payload by adding "raw_[xyz]" to "filtered_attributes" of the devices
* allows a simple offset based calibration. not perfect , but much better than without.

one sensor went from ~15° to a few degrees off

will update the documentation on how to calibrate, 
but in a nutshell:
put the sensor on a flat surface, set the offset for x and y to the opposite of raw_x and raw_y 
put the sensor on its side, do the same thing for z

### DIFF
--- a/src/devices/xiaomi.ts
+++ b/src/devices/xiaomi.ts
@@ -2429,8 +2429,8 @@ const definitions: Definition[] = [
         exposes: [
             e.battery(), e.device_temperature(), e.vibration(), e.action(['vibration', 'tilt', 'drop']),
             e.numeric('strength', ea.STATE), e.enum('sensitivity', ea.STATE_SET, ['low', 'medium', 'high']),
-            e.angle_axis('angle_x'), e.angle_axis('angle_y'), e.angle_axis('angle_z'), e.battery_voltage(), e.power_outage_count(false),
-            e.x_axis(), e.y_axis(), e.z_axis(),
+            e.angle_axis('angle_x'), e.angle_axis('angle_y'), e.angle_axis('angle_z'),
+            e.x_axis(), e.y_axis(), e.z_axis(), e.battery_voltage(), e.power_outage_count(false),
         ],
     },
     {


src/converters/fromZigbee.ts
```diff
@@ -4911,10 +4911,20 @@ const converters1 = {
...
+                let x = ((data['1'] << 16) >> 16);
+                let y = (data['1'] >> 16);
                 // left shift first to preserve sign extension for 'z'
-                const z = ((data['0'] << 16) >> 16);
+                let z = ((data['0'] << 16) >> 16);
+
+                // raw accelrometer values
+                result.raw_x=x;
+                result.raw_y=y;
+                result.raw_z=z;
```
Koenkk, line 4922: Why would you still want to have the `raw` values if you have the calibrated ones?

src/converters/fromZigbee.ts
```diff
@@ -4911,10 +4911,20 @@ const converters1 = {
...
+
+                // simple offset calibration
+                x=calibrateAndPrecisionRoundOptions(x, options, 'raw_x');
+                y=calibrateAndPrecisionRoundOptions(y, options, 'raw_y');
+                z=calibrateAndPrecisionRoundOptions(z, options, 'raw_z');
```
Koenkk, line 4927: I think it's better to call the options `x` (no `raw_`)

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/2375

src/converters/fromZigbee.ts
ghoz: it the only way I found to get the raw values that need calibration

src/converters/fromZigbee.ts
ghoz: oh sorry you were talking about the option name
yes I can rename (the question for the raw values is still valid though)

Koenkk: Please rename

src/converters/fromZigbee.ts
Koenkk: But you already have the raw values before doing any calibration?

ghoz: We do have the raw values, but right now, they are not available to users, we use them as is to compute and expose angles.
In order to know what offset to set, we need a way for the user to get the values when the sensor is on a level surface to compensate the values
(eg sensor  flat  I get "raw_x":60,"raw_y":-13,"raw_z":-704, => I set in the options the offset for x to -60, y to 13 
with the sensor on one side :  I get "raw_x":-942,"raw_y":-11,"raw_z":225, => offset for z ~ -225 )
so to reiterate as a user in order to find the offsets I need the raw values one way or another

To do an automated calibration witout the raw values, we would have to ask the user to put the device on 2 perpendicular faces, and wait for the values to update between each change, then compute the offsets for each axis and push them in the config or somewhere else, but not sure how to approach this tbh

src/converters/fromZigbee.ts
ghoz: renamed.
I think It might be confusing as they are offsets for the accelerometer underlying /raw values, not the exposed angle_x y z that are computed...

PR #6623, opened 2023-12-04
Improve TS0601_din_1 and add Sunricher SR-ZG9101SAC-HP-SWITCH-B fingerprint

dduransseau:
- Add reactive power, energy fields to Tuya TS0601_din_1
- Correct Zemismart brand for SPM01-D2TZ (manufactured by bituo-technik)
- Add Sunricher SR-ZG9101SAC-HP-SWITCH-B/ Vesternet VES-ZB-SWI-015 firmware fingerprint

Relative to previous PR #6607

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -5730,7 +5730,7 @@ const definitions: Definition[] = [
     {
         fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE200_bcusnqt8'}],
         model: 'SPM01-D2TZ',
-        vendor: 'Yagusmart',
+        vendor: 'Zemismart',
         description: 'Smart energy monitor for 1P+N system',
         fromZigbee: [tuya.fz.datapoints],
         toZigbee: [tuya.tz.datapoints],
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -3557,15 +3557,15 @@ const definitions: Definition[] = [
                 [1, 'energy', tuya.valueConverter.divideBy100],
                 [6, null, tuya.valueConverter.phaseVariant1], // voltage and current
                 [16, 'state', tuya.valueConverter.onOff],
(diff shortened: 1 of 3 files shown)


Koenkk:
@martyn-vesternet could you check the changes to `vesternet.ts`?

src/devices/tuya.ts
```diff
@@ -3557,15 +3557,15 @@ const definitions: Definition[] = [
                 [1, 'energy', tuya.valueConverter.divideBy100],
                 [6, null, tuya.valueConverter.phaseVariant1], // voltage and current
                 [16, 'state', tuya.valueConverter.onOff],
+                [101, 'total_energy', tuya.valueConverter.divideBy100], // total energy produced + consumed
                 [102, 'produced_energy', tuya.valueConverter.divideBy100],
                 [103, 'power', tuya.valueConverter.raw],
                 [105, 'ac_frequency', tuya.valueConverter.divideBy100],
+                [109, 'energy_reactive', tuya.valueConverter.divideBy100], // reactive energy in VArh
+                [110, 'power_reactive', tuya.valueConverter.raw], // reactive power
```
Koenkk, line 3565: Please update the `exposes` accordingly.

martyn-vesternet:
@dduransseau I would suggest creating your own dedicated device in the Sunricher file to suit that particular firmware version.

You'll want to make sure that the configuration for the Vesternet version works correctly with your version, especially as there is a pending PR that accounts for differing behaviours between versions 2.5.3_r2 and 2.9.2_r3 (measurements are on different endpoints).

dduransseau:
@martyn-vesternet done. 
I see your point and seems a good thing that there is a new endpoint for measurement. On my device I see measure on each channel endpoint but values are exactly the same which should not be the case at least for power. Having a dedicated endpoint for shared value would make more sense.

martyn-vesternet:
@Koenkk for this one yes, I think the changes to vesternet.ts were removed so should be OK.

Koenkk:
thanks!

PR #6626, opened 2023-12-05
Update lixee.ts

fairecasoimeme:
Add battery percentage information to ZiPulses devices. (Available with 4000-0006 version)


Koenkk:
Thanks!

PR #6632, opened 2023-12-05
Add device: IKEA PARASOLL door/windows sensor E2013

Noettore:
I was able to get an IKEA PARASOLL door/window sensor before being released.

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -1273,7 +1273,7 @@ const definitions: Definition[] = [
         zigbeeModel: ['PARASOLL Door/Window Sensor'],
         model: 'E2013',
         vendor: 'IKEA',
-        description: 'PARASOLL Door/Window Sensor',
+        description: 'PARASOLL door/window Sensor',
         fromZigbee: [fz.ias_contact_alarm_1],
         toZigbee: [],
         exposes: [e.battery_low(), e.tamper(), e.contact()],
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

rjblake:
See you've added e.tamper to the sensor, but I'm not sure there is any tamper sensor for this. The IKEA documentation and press release does not mention this ability.

Noettore:
Hi @rjblake,
you're right, my bad.
I'll create a new PR removing the e.tamper and also the e.battery_low, as I don't believe that exists either.

Thanks,
Ettore

ArdaXi:
@Noettore If you're going to be creating a new PR, you might consider adding the standard battery converter too, since it does seem to expose a standard `genPowerCfg` cluster. That way you can still see the battery percentage.

rufferson:
And ota
```
{
    zigbeeModel: ['PARASOLL Door/Window Sensor'],
    model: 'E2013',
    vendor: 'IKEA',
    description: 'PARASOLL Door/Window Sensor',
    fromZigbee: [fz.ias_contact_alarm_1, fz.battery],
    toZigbee: [],
    ota: ota.tradfri,
    exposes: [e.battery(), e.contact()],
}
```

rjblake:
> @Noettore If you're going to be creating a new PR, you might consider adding the standard battery converter too, since it does seem to expose a standard `genPowerCfg` cluster. That way you can still see the battery percentage.

Seems like there is some battery info - found this post [https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7417#issuecomment-1845192248](url)

PR #6633, opened 2023-12-06
add Aqara WS-USC01 power_o_memory

caryyu0306:
[https://github.com/Koenkk/zigbee2mqtt/issues/19884](url)
link here

Because I tested the External converter using the WS-USC02 2-gang model to trick the power_outage_memory parameter into working, I located key areas in the source code where I added the model and power_outage_memory parameters. I'm not sure if this approach is correct; could you please verify?


Koenkk:
Already made the changes in https://github.com/Koenkk/zigbee-herdsman-converters/commit/3b07bf1d7b87491a3275163cc6acc71c40a17dab before seeing this, sorry!

PR #6634, opened 2023-12-06
fix: giving the _TZE204_5toc8efa own settings separated from other _TZE2**

DasRed:
Pull Request for the problem described here: https://github.com/Koenkk/zigbee2mqtt/issues/19991


Koenkk:
If I understand correctly, only the step size is different, please use the [`expose()` function instead](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3b07bf1d7b87491a3275163cc6acc71c40a17dab/src/devices/tuya.ts#L569)

PR #6637, opened 2023-12-06
Update innr.ts

z647230:
added support for innr sp-240 device:

        zigbeeModel: ['SP 240'],
        model: 'SP 240',
        vendor: 'Innr',

### DIFF
--- a/src/devices/innr.ts
+++ b/src/devices/innr.ts
@@ -1,6 +1,7 @@
 import {Definition} from '../lib/types';
 import * as exposes from '../lib/exposes';
 import fz from '../converters/fromZigbee';
+import tz from '../converters/toZigbee';
 import * as reporting from '../lib/reporting';
 import extend from '../lib/extend';
 const e = exposes.presets;
@@ -707,7 +708,6 @@ const definitions: Definition[] = [
         options: [exposes.options.measurement_poll_interval()],
 
         exposes: [e.switch(), e.power(), e.current(), e.voltage(), e.energy()],
-        onEvent: (type, data, device, options) => tuya.onEventMeasurementPoll(type, data, device, options, true, false),
     },
 ];
 
--- a/src/devices/innr.ts
+++ b/src/devices/innr.ts
@@ -689,25 +689,11 @@ const definitions: Definition[] = [
         zigbeeModel: ['SP 240'],
         model: 'SP 240',
         vendor: 'Innr',
-        description: 'Innr Smartplug with Powermetering',
-        fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fz.ignore_basic_report],
-        toZigbee: [tz.on_off],
-        configure: async (device, coordinatorEndpoint, logger) => {
-            const endpoint = device.getEndpoint(1);
-            await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'haElectricalMeasurement', 'seMetering']);
-            endpoint.saveClusterAttributeKeyValue('seMetering', {divisor: 100, multiplier: 1});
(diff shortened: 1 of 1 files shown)


src/devices/innr.ts
```diff
@@ -684,6 +685,30 @@ const definitions: Definition[] = [
             }
         },
     },
+    {
```
Koenkk, line 688: Can you try with the following instead:

```
{
        zigbeeModel: ['SP 240'],
        model: 'SP 240',
        vendor: 'Innr',
        description: 'Smart plug',
        extend: [
            onOff(),
            electricityMeter(),
        ],
    },
```

heresiarch:
You are so fast unbeliebable, thank you @Koenkk  . In the meantime I tried this successfully as workaround with a custom converter:
```javascript
//const extend = require('zigbee-herdsman-converters/lib/extend');
const {light, onOff, electricityMeter} = require('zigbee-herdsman-converters/lib/modernExtend');
//const e = exposes.presets;
//const ea = exposes.access;

const definition = {
zigbeeModel: ['SP 240'], // The model ID from: Device with modelID 'lumi.sens' is not supported.
model: 'SP 240', // Vendor model number, look on the device for a model number
vendor: 'inrr', // Vendor of the device (only used for documentation and startup logging)
description: 'smart plug innr', // Description of the device, copy from vendor site. (only used for documentation and startup
extend: [
        onOff(),
        electricityMeter({current: {divisor: 1000}, voltage: {divisor: 1}, power: {divisor: 1}, energy: {divisor: 100}}),
        ],
//ota: ota.zigbeeOTA,
};

module.exports = definition;

```

src/devices/innr.ts
z647230: it works but...
looking at the "exposes" screen of the home assistant/zigbee2mqtt it shows adjusters on the right side of the screen as if you could change voltage.  (see attachment)
innrsp240exposescreen

The way I implemented/tested your sample is by converting it into an external converter. (see attachment 2)
innrsp240

(ps. last time I programmed was in old fashioned assembler, Cobol and later in C. Now learning on how to work with all the modern stuff ;-) )

Koenkk:
what error message does show up?

src/devices/innr.ts
Koenkk: Can you try with:

```js
{
        zigbeeModel: ['SP 240'],
        model: 'SP 240',
        vendor: 'Innr',
        description: 'Smart plug',
        extend: [
            onOff({powerOnBehavior: false}),
            electricityMeter({current: {divisor: 1000}, voltage: {divisor: 1}, power: {divisor: 1}, energy: {divisor: 100}}),
        ],
    },
```

z647230: Configure goes ok, the powerOnBehavior i tried with and without your suggestion as it already worked. But the function to disable, like in your example, works. So i tested with poweronbehavior enabled and all settings off/on/toggle/restore work.
I noticed that the default Zigbee icon also is replaced by a plug jpeg image.
i think that makes it pretty complete unless you want to test other stuff on it ?

Koenkk:
Thanks!

z647230:
You’re welcome.But like said before, the powerInBehavior does work correctly so no need to disable it.Rgdz, MartinGrtz, MartinOn 10 Dec 2023, at 20:50, Koen Kanters ***@***.***> wrote:﻿
Thanks!

Koenkk:
It's disabled: https://github.com/Koenkk/zigbee-herdsman-converters/blob/c7ee83322f9cd4f5c83da7f042e2a5ae5468ec09/src/devices/innr.ts#L668

z647230:
Yes, I noticed but it should be enabled as the powerOnBehavior is supported correctly by the device.I fully tested it with all the states.Grtz, MartinOn 11 Dec 2023, at 19:22, Koen Kanters ***@***.***> wrote:﻿
It's disabled: https://github.com/Koenkk/zigbee-herdsman-converters/blob/c7ee83322f9cd4f5c83da7f042e2a5ae5468ec09/src/devices/innr.ts#L668

Koenkk:
I see, enabled it!

PR #6647, opened 2023-12-07
Develco MOSZB-140: adjusted min occupancy_timeout

emandtf:
Forcing minimum value to 20 was not the right value.
Default value (by manufacturer value) is 12 but after a test it can be lowered down to 5.


Koenkk:
Thanks!

PR #6649, opened 2023-12-07
Added RSH-HS06 _TZ3000_zl1kmjqx TUYA device - temperature and humidity sensor

panteLx:
Added support for RSH-HS06 _TZ3000_zl1kmjqx TUYA device - temperature and humidity sensor

It wont show the modelID so I'll leave it blank. External converter is working. When I trust the package the modelID is RSH-HS06.

Maybe there is a way to get the modelID but my z2mqtt wont show the devices as supported (with external converter) if I type in RSH-HS06 as modelID. It will work without the modelID


Koenkk:
Thanks!

PR #6651, opened 2023-12-07
Add support for BHI-US Bosch Universal Switch II

slugzero:
Adds support for the Bosch BHI-US Universal Switch II.

This converter requires a new manufacturer specific cluster:
https://github.com/Koenkk/zigbee-herdsman/pull/820

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -6,6 +6,7 @@ import * as reporting from '../lib/reporting';
 import * as utils from '../lib/utils';
 import * as constants from '../lib/constants';
 import * as ota from '../lib/ota';
+import * as globalStore from '../lib/store';
 import {Tz, Fz, Definition, KeyValue} from '../lib/types';
 const e = exposes.presets;
 const ea = exposes.access;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

mldytech:
Hi @slugzero, first of all thanks for adding support for this device. I noticed that the actual button presses are not exposed in the zigbee2mqtt webinterface and are also not available (at least not out of the box) in e. g. homeassistant.

Nevertheless, when pressing a button, it seems working and the MQTT messages are correct (e. g. for short press upper left):
`zigbee2mqtt/friendly-name/action: button_top_left_release`

Since I'm not very familiar with this project, is there anything I could do to get this device working as intended? I would be grateful for any suggestions and am also willing to do some work / get into the project.

slugzero:
Hi @mldytech, you are right, the `action` expose is missing from the device definition. Note that it should work in Home Assistant nonetheless: Make sure you have sent all actions at least once (i.e. press and long press all four buttons) while HA is running. After that, the entries should show up in the list of device triggers.

It will take me a little while to change this, because the device is in my "productive" network rn. But if you'd like to give it a try yourself: Adding  
```
e.action(['button_top_left_release',
	'button_top_right_release',
	'button_bottom_left_release',,
	'button_bottom_right_release',
	'button_top_left_longpress',
	'button_top_right_longpress',
	'button_bottom_left_longpress',
	'button_bottom_right_longpress',
	'button_top_left_longpress_release',
	'button_top_right_longpress_release',
	'button_bottom_left_longpress_release',
	'button_bottom_right_longpress_release']),
``` 
to the list of exposes in https://github.com/Koenkk/zigbee-herdsman-converters/blob/da65b1aeffd96527df02725b49de61e453fee059/src/devices/bosch.ts#L1886 should fix it.

The simplest way to do and test this is probably to copy over the device definition to a custom converter file as described here: https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html

mldytech:
@slugzero Thank you very much for your answer. You're right, the entries were already visible in HA.

I appreciate your tips / guide and will definitely give it a try on the weekend!

PR #6654, opened 2023-12-07
[BAC-002-ALZB] FIX mode get/set

a2nt:
Sorry, made some mistakes using previous commit, but now I have found a way to test it with real device and did it:
+ manually switching mode on thermostat using buttons 
+ switching mode using Zigbee2MQTT

Now it works perfect


Koenkk:
Thanks!

PR #6662, opened 2023-12-08
feat(add): Moes ZSS-X-GWM-C (white label TuYa TS0203 w/o tamper).

raineth:
Also split out the list of manufacturerNames used for models without tamper sensors into a separate const as it's starting to get lengthy.


Koenkk:
Thanks!

PR #6666, opened 2023-12-09
Random tidying

ghoz:
While browsing the source, some random oddities :
* 2 devices use `Tuya` as vendor name rather than `TuYa`
* Xiaomi MCCGQ13LM contact sensor using `battery_cover` rather than the much more common `tamper`
* a double calibration of  pm25 of STARKVIND Air purifier


ghoz:
Let me know if you prefer that I split the PR...

Koenkk:
Thanks!

PR #6681, opened 2023-12-11
Add OTA for Sonoff device TRVZB

lcheng33775823:


src/lib/ota/common.ts
```diff
@@ -324,8 +324,14 @@ export async function updateToLatest(device: Zh.Device, logger: Logger, onProgre
 
     return new Promise((resolve, reject) => {
         const answerNextImageBlockOrPageRequest = () => {
-            const imageBlockRequest = endpoint.waitForCommand('genOta', 'imageBlockRequest', null, 150000);
-            const imagePageRequest = endpoint.waitForCommand('genOta', 'imagePageRequest', null, 150000);
+            let imageBlockOrPageRequestTimeoutMs: number = 150000;
+
+            if ( request.payload.manufacturerCode == 4742 && request.payload.imageType == 8199 ) {
```
Koenkk, line 329: Could you add a comment here explaining why this is needed?

lcheng33775823: TRVZB, it consists of two MCUS (Zigbee Mcu + Mcu 2), so when doing OTA, I combined the OTA files of the two MCUS into one. When the Zigbee device receives the last packet of Block Response, Mcu 2 will be upgraded first. The upgrade End Request will not be sent until the Mcu 2 upgrade is successful, but the upgrade time is between 40 minutes and 1 hour, which will cause Zigbee2Mqtt to report an error. That's why I need to add this change.

Koenkk: Please document that above the `if` statement using comments.

lcheng33775823: Thanks for your help, I have resubmitted code comment

Koenkk:
Thanks!

PR #6683, opened 2023-12-11
fix: add color for Philips 929003598001

Shobba:
Color option is currently missing.
See: https://www.philips-hue.com/de-de/p/hue-white---color-ambiance-surimu-quadratisches-panel-30x30-weiss/8720169159136


Koenkk:
Thanks!

PR #6689, opened 2023-12-12
add Wireless switch with 6 buttons, 'TS004F', ['_TZ3000_r0o2dahu']

waterlou:
It is a 6 buttons wireless panel, but the model ID is TS004F, that suppose to be a 4 buttons panel in the definition of fz.tuya_on_off_action.  So I created another function fz.tuya_on_off_action_8 for this.  Actually I dunno why fz.tuya_on_off_action need to do the modelID check because it is just a simple int -> string conversion.

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -2253,6 +2253,21 @@ const converters1 = {
             return {action: `${button}${clickMapping[msg.data[3]]}`};
         },
     } satisfies Fz.Converter,
+    tuya_on_off_action_8: {
+        cluster: 'genOnOff',
+        type: 'raw',
+        convert: (model, msg, publish, options, meta) => {
+            if (hasAlreadyProcessedMessage(msg, model, msg.data[1])) return;
+            const clickMapping: KeyValueNumberString = {0: 'single', 1: 'double', 2: 'hold'};
+            const buttonMapping: KeyValueNumberString = {1: '1', 2: '2', 3: '3', 4: '4', 5: '5', 6: '6', 7: '7', 8: '8'};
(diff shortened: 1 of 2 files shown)


src/converters/fromZigbee.ts
```diff
@@ -2253,6 +2253,21 @@ const converters1 = {
             return {action: `${button}${clickMapping[msg.data[3]]}`};
         },
     } satisfies Fz.Converter,
+    tuya_on_off_action_8: {
+        cluster: 'genOnOff',
+        type: 'raw',
+        convert: (model, msg, publish, options, meta) => {
+            if (hasAlreadyProcessedMessage(msg, model, msg.data[1])) return;
+            const clickMapping: KeyValueNumberString = {0: 'single', 1: 'double', 2: 'hold'};
+            const buttonMapping: KeyValueNumberString = {1: '1', 2: '2', 3: '3', 4: '4', 5: '5', 6: '6', 7: '7', 8: '8'};
```
Koenkk, line 2262: Please merge this with `tuya_on_off_action` above

src/devices/tuya.ts
```diff
@@ -2143,6 +2143,42 @@ const definitions: Definition[] = [
             await reporting.batteryPercentageRemaining(endpoint);
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS004F', ['_TZ3000_r0o2dahu']),
+        model: 'TS004F',
```
Koenkk, line 2148:
```suggestion
        model: 'TS004F_6_buttons',
```

src/converters/fromZigbee.ts
Koenkk: @waterlou I saw you removed the original content, instead a new `else if` statement should be added for this device.

PR #6690, opened 2023-12-12
add support for sikom thermostat

atdhebela:
Adding support for zigbee device model "300-9715V10" made by vendor Sikom


Koenkk:
Thanks!

PR #6687, opened 2023-12-12
Add new IKEA JETSTROM light panel

aoaite:
Add new IKEA JETSTROM light panel

PR #6694, opened 2023-12-13
Add missing color temp values

aoaite:
Add missing color temp values as requested in a comment in previous PR

Dev console output:

`info 2023-12-13 01:05:29Read result of 'lightingColorCtrl': {"colorTempPhysicalMin":250,"colorTempPhysicalMax":454}`

PR #6687


Koenkk:
This is not needed since `250, 454` is the default range for all TRADFRI devices.

PR #6695, opened 2023-12-13
update about SPM01&SPM02 New version by [BITUO TECHNIK]

script0803:


script0803:
update about SPM01&SPM02 New version by [BITUO TECHNIK]

src/devices/tuya.ts
```diff
@@ -5832,6 +5832,71 @@ const definitions: Definition[] = [
...
+        model: 'SPM02V2',
+        vendor: 'TuYa',
+        description: 'Smart energy monitor for 3P+N system',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        exposes: [tuya.exposes.voltageWithPhase('X'), tuya.exposes.voltageWithPhase('Y'), tuya.exposes.voltageWithPhase('Z'),
+            tuya.exposes.powerWithPhase('X'), tuya.exposes.powerWithPhase('Y'), tuya.exposes.powerWithPhase('Z'),
+            tuya.exposes.currentWithPhase('X'), tuya.exposes.currentWithPhase('Y'), tuya.exposes.currentWithPhase('Z'),
+            e.energy().withDescription('Total forward active energy'),
```
Koenkk, line 5873: Instead of using `X`/`Y`/`Z`  I propose to use `a`/`b`/`c` since we also do that for other devices.

src/devices/tuya.ts
```diff
@@ -5832,6 +5832,71 @@ const definitions: Definition[] = [
...
+            e.produced_energy().withDescription('Total reverse active energy'),
+            e.power_factor().withDescription('Total Power factor'), tuya.exposes.powerWithPhase('total'),
```
Koenkk, line 5875: Why change the description?

Koenkk, line 5875: Instead of `total`, we can use `abc` here.

PR #6699, opened 2023-12-13
fix: HS2AQ correct HCHO value to match what is on the display

sjorge:
[Reported](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5547#issuecomment-1854187611) by @RobLeighton67 that we seem to be off by a factor of 10.

Display shows 0.01 but z2m shows 0.1, update the convertor to device by 1000.0 instead of 100.0 to align these.


Koenkk:
Thanks!

PR #6700, opened 2023-12-13
Added missing on/off for Namron outlet switch

gdybvik:
Namron Switch 4512767 presented no on/off under "exposes" in the GUI.


Koenkk:
Thanks!

PR #6701, opened 2023-12-13
No energy consumption and wrong current for _TZ3000_okaz9tjs and _TZ3000_typdpbpg

DataGhost:
Fixes [20028](https://github.com/Koenkk/zigbee2mqtt/issues/20028).

I've confirmed this to be working on _TZ3000_okaz9tjs. This PR includes a likely fix for _TZ3000_typdpbpg which seems to be affected the same way, still pending confirmation in that issue thread.


Koenkk:
Many thanks!

PR #6711, opened 2023-12-15
fix getAxios() in ota.common: Add support for HTTP 302 redirects

adeepn:
Fix getAxios() in ota.common: Add support for HTTP 302 redirects, when no hostname in Location header.

Fix me, if we need to parse more options than `responseType` to `axios.get()`


Koenkk:
Thanks!

PR #6716, opened 2023-12-15
fix: HeatIt Zig Dim 250 reporting and binding

geirra:
Config changes so the dimmer report changes when dimming manually.


Koenkk:
Thanks!

PR #6718, opened 2023-12-15
Update philips.ts - added colorTemp range to Centris 2 spots white

wbsantos:
Philips Hue Centris with 2 spots white was missing the colorTemp range. I have this product and already checked (it is the same as most of the philips hue products)


Koenkk:
Thanks!

Issue #6721, opened 2023-12-16
VOCKQJK11LM Improvements and switch to modernExtends


sjorge:
The PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/6709 was getting a bit big and messy, I've decided to break it up in a few [smaller PRs ](https://github.com/Koenkk/zigbee-herdsman-converters/pull/6709#issuecomment-1858788532)to get this in in smaller steps.

- [x] basic improvements  (phase 1)
   - [x] wait for https://github.com/Koenkk/zigbee-herdsman/pull/829
      - [x] use more readable attribute names
   - [X] implement air_quality as modernExtend
   - [X] move voc to modernExtend
     - [X] cleanup fz.xiaomi_tvoc
   - [X] move display_unit to modernExtend
      - [X] cleanup fzLocal.VOCKQJK11LM_display_unit
      - [X] cleanup lib.xiaomi numericAttributes2Payload
   - [x] Update modernExtend.numeric and modernExtend.enumLookup
      - [x] read attribute during configure (Do we need a toggle for this? I think we always want to do  this)
      - [x] configure attribute reporting (Only when reporting config is set in args)
    - [x] setup reporting of batteryVoltage (until phase 3 where we can use a modernExtend) 
- [x] allow reading of additional attributes, we get this for free with modernExtends! (phase 2)
   - [ ] ~~integrate precision and calibration in the numeric modernExtend and auto expose these options~~
   - [x] add temperature + humidity modernExtend presets beased on numeric modernExtend
     - [ ] ~~add a 'filter' function so we don't need to reinvent the wheel for things like `fz.SNZB_temperature`~~
   - [x] move temperature to modernExtend (don't forget about the reporting setup and options!)
   - [x] move humidity to modernExtend (don't forget about the reporting setup and options!)
- [ ] want modernExtend for battery things
   - [ ] add batteryPercentageRemaining modernExtend (do reporting, reading, ...)
   - [ ] add batteryVoltage modernExtend (do reporting, reading, have the 'battery type configrable', ...)
   - [ ] add xiaomiBatteryVoltage modernExtend (use aqaraOpple attribute as it offers better precision)
      - [ ] add to cluster.ts
      - [ ] update all devices and test I own
- [ ] move more stuff from lib.xiaomi's numericAttributes2Payload to modernExtends (phase 4)
   - [ ] move device_temperature to modernExtend
   - [ ] ...

sjorge:
[...]

PR #6723, opened 2023-12-16
VOCKQJK11LM Improvements and modernExtend conversion (1/4)

sjorge:
Waiting on https://github.com/Koenkk/zigbee-herdsman/pull/829 to switch to easier to read attribute names, draft until then.

More details in https://github.com/Koenkk/zigbee-herdsman-converters/issues/6721, but the bulk of this should at least be  reviewable.


sjorge:
I'm not sure how to fix the ci failure, I'm guessing I messed up the typing somewhere.

sjorge:
@Koenkk the CI failures seem to be related to https://github.com/Koenkk/zigbee-herdsman-converters/pull/6723/commits/76a8f425a43252a8115f85b7fd2f0e8010669a10 but if I drop that commit I get the error again:

```
Zigbee2MQTT:debug 2023-12-16 21:37:50: Reading for 0x54ef441000684954/1 aqaraOpple [{"attribute":{"id":297,"type":32}}]
Zigbee2MQTT:error 2023-12-16 21:37:50: Failed to configure '0x54ef441000684954', attempt 2 (Error: Cluster 'aqaraOpple' has no attribute 'undefined'
    at Object.getAttribute (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/zcl/utils.ts:93:19)
    at Endpoint.read (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:446:87)
    at setupAttributes (/opt/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:43:24)
    at configure (/opt/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:325:11)
    at Object.configure (/opt/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/index.ts:138:31)
    at Configure.configure (/opt/zigbee2mqtt/lib/extension/configure.ts:117:13))
```

Basically using `{attribute: {id: 0x0129, type: Zcl.DataType.uint8} ` works everywhere except in setupAttributes, but if i use `{attribute: {ID: 0x0129, type: Zcl.DataType.uint8}` everything **except** setupAttributes is broken (e.g. reading via get, it also doesn't seem to parse the readResponse/attributeReport that way.

sjorge:
@Koenkk this first PR should be ready now.

I also double checked and neither displayUnit nor airQuality support reporting, having it on the later would have been nice. But it seems trigger a report anyway everytime the value changes 🤷

src/lib/xiaomi.ts
```diff
@@ -1381,6 +1375,40 @@ export const xiaomiModernExtend = {
         attribute: 'presentValue',
         ...args,
     }),
+    aqaraVoc: (args?: Partial<modernExtend.NumericArgs>) => modernExtend.numeric({
+        name: 'voc',
+        cluster: 'genAnalogInput',
+        attribute: 'presentValue',
+        configureReporting: {minimumReportInterval: 10, maximumReportInterval: constants.repInterval.HOUR, reportableChange: 5},
```
Koenkk, line 1382: Let's remove `maximumReportInterval` and use the default from modernExtend instead (`repInterval.MAX`))

src/lib/modernExtend.ts
```diff
@@ -358,17 +381,26 @@ export function numeric(args: NumericArgs): ModernExtend {
         },
     }];
 
-    return {exposes: [expose], fromZigbee, toZigbee, isModernExtend: true};
+    const configure: Configure = async (device, coordinatorEndpoint, logger) => {
+        // call setupAttributes if we have a configureReporting payload or configureSkipRead is not set, but then use readOnly=true
+        const ep = device.getEndpoint(isNumber(configureEndpointId) ? configureEndpointId : 1);
```
Koenkk, line 386: The `configureEndpointId` option is not needed, the `setupAttributes` already configures it on the correct endpoint.

src/lib/modernExtend.ts
```diff
@@ -312,17 +320,32 @@ export function enumLookup(args: EnumLookupArgs): ModernExtend {
...
+    endpoint?: string, configureEndpointId?: number, configureReporting?: Partial<ZHConfigureReportingItem>, configureSkipRead?: boolean,
     valueMin?: number, valueMax?: number, valueStep?: number, scale?: number,
 }
 export function numeric(args: NumericArgs): ModernExtend {
-    const {name, cluster, attribute, description, zigbeeCommandOptions, unit, readOnly, valueMax, valueMin, valueStep, endpoint, scale} = args;
-    const attributeKey = isString(attribute) ? attribute : attribute.id;
+    const {
+        name, cluster, attribute, description,
+        zigbeeCommandOptions, readOnly, unit, endpoint,
+        configureEndpointId, configureReporting, configureSkipRead,
```
Koenkk, line 345: The `configureSkipRead` option doesn't seem to be used anywhere, if that's the case I propose to leave it out for now.

src/lib/modernExtend.ts
```diff
@@ -312,17 +320,32 @@ export function enumLookup(args: EnumLookupArgs): ModernExtend {
         },
     }];
 
-    return {exposes: [expose], fromZigbee, toZigbee, isModernExtend: true};
+    const configure: Configure = async (device, coordinatorEndpoint, logger) => {
```
Koenkk, line 323: Only add configure to the result when requested, e.g.

```
const result = {exposes: [expose], fromZigbee, toZigbee, isModernExtend: true}
if (configureReporting) {
  result.configure = ...
}
return result;

File: src/lib/xiaomi.ts (hunk 1 above)
Comment (sjorge) [reply src/lib/xiaomi.ts:1382]: I actually copied those from what is currently in the configure() function and those seemed like a decent value. (which I forgot to cleanup, rebasing that commit to also clean this up)
```
const payload = reporting.payload('presentValue', 10, constants.repInterval.HOUR, 5);
```

Dropping this sets it to `65000` which seems a bit high, I would not set it longer than 1x 3h.

File: src/lib/modernExtend.ts (hunk 2 above)
Comment (sjorge) [reply src/lib/modernExtend.ts:386]: Sadly that does not work for Xiaomi/Aqara devices ... see the 2nd point of this comment: https://github.com/Koenkk/zigbee-herdsman-converters/issues/6721#issuecomment-1858909096

We should probably not default to 1 though, and let it figure it out when unset.

File: src/lib/modernExtend.ts (hunk 3 above)
Comment (sjorge) [reply src/lib/modernExtend.ts:345]: I added it in case you can't read an attribute. I can drop it again.

File: src/lib/xiaomi.ts (hunk 1 above)
Comment (Koenkk) [reply src/lib/xiaomi.ts:1382]: > Dropping this sets it to 65000 which seems a bit high, I would not set it longer than 1x 3h.

I see, then this is fine

File: src/lib/modernExtend.ts (hunk 4 above)
Comment (sjorge) [reply src/lib/modernExtend.ts:323]: Does this still make sense if we drop `configureSkipRead` ?
- we have a `configureReporting` payload -> we call `setupAttributes()`
- we do not have a `configureReporting` payload -> we call `setupAttributes(...readOnly=true)`

We will always end up with a configure in this case, either to setup reporting or to just read the value ?

File: src/lib/modernExtend.ts (hunk 2 above)
Comment (sjorge) [reply src/lib/modernExtend.ts:382]: Just double check, without specifying ep1 it indeed fails but only for the aqara device I have. It works fine for a sonoff SNZB02 I had also temporary added a numeric modernExtend too.

```
error 2023-12-17 18:04:59: Failed to configure 'airquality/livingroom', attempt 1 (Error: Device 0x54ef441000684954 has no input cluster aqaraOpple
    at getEndpointsWithInputCluster (/opt/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:23:15)
    at setupAttributes (/opt/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:34:55)
    at configure (/opt/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:323:17)
    at Object.configure (/opt/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/index.ts:138:31)
    at Configure.configure (/opt/zigbee2mqtt/lib/extension/configure.ts:117:13)
    at Configure.onMQTTMessage (/opt/zigbee2mqtt/lib/extension/configure.ts:55:21))
```

I will leave configureEndpointId in place and set it to 1 in src/lib/xiaomi.ts for those 3 modernExtends but have it default to grabbing all the endpoints.

File: src/lib/xiaomi.ts [hunk 5]
```diff
@@ -1381,6 +1375,43 @@ export const xiaomiModernExtend = {
...
+        unit: 'ppb',
+        readOnly: true,
+        ...args,
+    }),
+    aqaraAirQuality: (args?: Partial<modernExtend.EnumLookupArgs>) => modernExtend.enumLookup({
+        name: 'air_quality',
+        lookup: {'excellent': 1, 'good': 2, 'moderate': 3, 'poor': 4, 'unhealthy': 5},
+        cluster: 'aqaraOpple',
+        attribute: 'airQuality',
+        configureEndpointId: 1,
```
sjorge, line 1394: Since we still call setupAttributes() but in readOnly when there is no configureReporting payload we still need to set it for these devices to avoid the error of it not finding aqaraOpple cluster.

Perhaps we should name this `endpointId` instead ?

sjorge:
@Koenkk can you take another look at this, I think most things are now resolved except 1 new question I added and a follow up w.r.t. 

> Only add configure to the result when requested

Given that with configureSkipRead is now gone, we will always call setupAttributes, either with a reporting payload or without one but with readOnly=true set (which does the read)

src/lib/modernExtend.ts
Koenkk: Fine with that!

Koenkk:
>  But once we start doing some more complex devices like one of the Ubisys H10 thermostats for example we will this this issue. (Although for a non battery device it's probably not the end of the world), although fixing that is probably out of scope for this PR, but something to keep in mind once we do have such a device.

I see, let's tackle this later indeed.

src/lib/xiaomi.ts
Koenkk: Let's call it `endpointID` indeed

src/lib/modernExtend.ts
```diff
@@ -358,17 +378,23 @@ export function numeric(args: NumericArgs): ModernExtend {
...
```
Koenkk, line 381: Please also add the same `configure` to `export function binary` (so `binary`, `numeric` and `enumLookup` have the same, I guess it would be good to create a function instead to prevent code duplication.

sjorge:
Removed and repaired the device, still seems to work and get configured.

sjorge:
Extra device came to make testing easier, paired fine first time.

Koenkk:
Prefect, made a small change regarding the `cluster.toString` and added a test case.

Bedankt! 😄

sjorge:
I'm generally really bad writing mock tests, thanks for adding one.

If I understand the test correctly, once I change temp and humidity to modernExtends in a layer PR, it should still pass (assuming I did the change correctly) ?

Koenkk:
Yes, but the order might change so then test needs to be updated.

Mar1usW3:
after upgrading to 1.3.5 i get the following errors in HA:
image

sjorge:
@Koenkk is there some sort of encoding issue with modernExtend vs the old style ?

Koenkk:
Yes, I've already fixed it

freeDom-:
Fixed it for me

PR #6733, opened 2023-12-18
WIP feat: Tuya datapoints modernExtend

kirovilya:
Examples:
```
    {
        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_fjjbhx9d', '_TZE200_e3oitdyu', '_TZE200_gwkapsoq', '_TZE204_zenj4lxv']),
        model: 'TS0601_dimmer_2',
        vendor: 'TuYa',
        description: '2 gang smart dimmer',
        meta: {multiEndpoint: true},
        endpoint: (device) => {
            return {'l1': 1, 'l2': 1};
        },
        whiteLabel: [
            {vendor: 'Moes', model: 'ZS-EUD_2gang'},
            {vendor: 'Moes', model: 'MS-105B'}, // _TZE200_e3oitdyu
            tuya.whitelabel('Moes', 'ZS-SR-EUD-2', 'Star ring smart dimmer switch 2 gangs', ['_TZE204_zenj4lxv']),
        ],
        extend: [
            tuya.modernExtend.dpLight({
                state: {dp: 1, type: tuya.dataTypes.bool, valueOn: ['ON', true], valueOff: ['OFF', false], skip: tuya.skip.stateOnAndBrightnessPresent},
                brightness: {dp: 2, type: tuya.dataTypes.number, scale: [0, 254, 0, 1000]},
                min: {dp: 3, type: tuya.dataTypes.number, scale: [0, 254, 0, 1000]},
                max: {dp: 5, type: tuya.dataTypes.number, scale: [0, 254, 0, 1000]},
                endpoint: 'l1',
            }),
            tuya.modernExtend.dpNumeric({
                name: 'countdown',
                dp: 6, type: tuya.dataTypes.number,
                expose: tuya.exposes.countdown(),
                endpoint: 'l1',
            }),
            tuya.modernExtend.dpLight({
                state: {dp: 7, type: tuya.dataTypes.bool, valueOn: ['ON', true], valueOff: ['OFF', false], skip: tuya.skip.stateOnAndBrightnessPresent},
                brightness: {dp: 8, type: tuya.dataTypes.number, scale: [0, 254, 0, 1000]},
                min: {dp: 9, type: tuya.dataTypes.number, scale: [0, 254, 0, 1000]},
                max: {dp: 11, type: tuya.dataTypes.number, scale: [0, 254, 0, 1000]},
                endpoint: 'l2',
            }),
            tuya.modernExtend.dpNumeric({
                name: 'countdown',
                dp: 12, type: tuya.dataTypes.number,
                expose: tuya.exposes.countdown(),
                endpoint: 'l2',
            }),
            tuya.modernExtend.dpEnumLookup({
                name: 'power_on_behavior',
                dp: 14, type: tuya.dataTypes.enum,
                lookup: {'off': 0, 'on': 1, 'previous': 2},
                description: 'Controls the behavior when the device is powered on after power loss',
            }),
            tuya.modernExtend.dpEnumLookup({
                name: 'backlight_mode',
                dp: 21, type: tuya.dataTypes.enum,
                lookup: {'off': 0, 'normal': 1, 'inverted': 2},
                description: 'Mode of the backlight',
            }),
        ],
    }

```
--------------
```
    {
        fingerprint: [
            {modelID: 'TS0601', manufacturerName: '_TZE200_gz3n0tzf'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_nthosjmx'},
        ],
        model: 'WZ5_cct',
        vendor: 'Skydance',
        description: 'Zigbee & RF 5 in 1 LED controller (CCT mode)',
        extend: [
            tuya.modernExtend.dpLight({
                state: {dp: 1, type: tuya.dataTypes.bool, valueOn: ['ON', true], valueOff: ['OFF', false]},
                brightness: {dp: 3, type: tuya.dataTypes.number, scale: [0, 254, 0, 1000]},
                colorTemp: {dp: 4, type: tuya.dataTypes.number, range: [250, 454], scale: [0, 454, 250, 1000]},
            }),
        ],
    },
    {
        fingerprint: [
            {modelID: 'TS0601', manufacturerName: '_TZE200_9hghastn'},
            {modelID: 'TS0601', manufacturerName: '_TZE200_9mt3kgn0'},
        ],
        model: 'WZ5_rgb',
        vendor: 'Skydance',
        description: 'Zigbee & RF 5 in 1 LED controller (RGB mode)',
        extend: [
            tuya.modernExtend.dpLight({
                state: {dp: 1, type: tuya.dataTypes.bool, valueOn: ['ON', true], valueOff: ['OFF', false]},
                brightness: {dp: 3, type: tuya.dataTypes.number, scale: [0, 254, 0, 1000]},
                color: {dp: 5, type: tuya.dataTypes.string},
            }),
        ],
    },
```
--------------
```
    {
        fingerprint: tuya.fingerprint('TS0601',
            ['_TZE200_yjjdcqsq', '_TZE200_9yapgbuv', '_TZE204_upagmta9', '_TZE200_cirvgep4', '_TZE200_upagmta9', '_TZE204_yjjdcqsq']),
        model: 'TS0601_temperature_humidity_sensor_2',
        vendor: 'TuYa',
        description: 'Temperature and humidity sensor',
        onEvent: tuya.onEvent({queryOnDeviceAnnounce: true}),
            await tuya.configureMagicPacket(device, coordinatorEndpoint, logger);
            // Required to get the device to start reporting
            await device.getEndpoint(1).command('manuSpecificTuya', 'dataQuery', {});
        },
        whiteLabel: [
            tuya.whitelabel('TuYa', 'ZTH01', 'Temperature and humidity sensor', ['_TZE200_yjjdcqsq', '_TZE204_yjjdcqsq']),
            tuya.whitelabel('TuYa', 'ZTH02', 'Temperature and humidity sensor', ['_TZE200_9yapgbuv']),
            tuya.whitelabel('TuYa', 'ZTH05', 'Temperature and humidity sensor', ['_TZE204_upagmta9', '_TZE200_upagmta9']),
            tuya.whitelabel('TuYa', 'ZTH08-E', 'Temperature and humidity sensor', ['_TZE200_cirvgep4']),
        ],
        extend: [
            tuya.modernExtend.dpTemperature({dp: 1, scale: 10}),
            tuya.modernExtend.dpHumidity({dp: 2}),
            tuya.modernExtend.dpBatteryState({dp: 3}),
            tuya.modernExtend.dpTemperatureUnit({dp: 9}),
        ],
    },
    {
        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_nvups4nh']),
        model: 'TS0601_contact_temperature_humidity_sensor',
        vendor: 'TuYa',
        description: 'Contact, temperature and humidity sensor',
        whiteLabel: [
            tuya.whitelabel('Aubess', '1005005194831629', 'Contact, temperature and humidity sensor', ['_TZE200_nvups4nh']),
        ],
        extend: [
            tuya.modernExtend.dpContact({dp: 1}, true),
            tuya.modernExtend.dpBattery({dp: 2}),
            tuya.modernExtend.dpTemperature({dp: 7, scale: 10}),
            tuya.modernExtend.dpHumidity({dp: 8}),
        ],
    },
```
--------------------
```
    {
        fingerprint: tuya.fingerprint('TS0601', ['_TZE204_vmcgja59']),
        model: 'ZYXH',
        vendor: 'TuYa',
        description: '8 gang switch',
        endpoint: (device) => {
            return {'l1': 1, 'l2': 1, 'l3': 1, 'l4': 1, 'l5': 1, 'l6': 1, 'l7': 1, 'l8': 1};
        },
        meta: {multiEndpoint: true},
        extend: [
            tuya.modernExtend.dpOnOff({dp: 1, endpoint: 'l1'}),
            tuya.modernExtend.dpOnOff({dp: 2, endpoint: 'l2'}),
            tuya.modernExtend.dpOnOff({dp: 3, endpoint: 'l3'}),
            tuya.modernExtend.dpOnOff({dp: 4, endpoint: 'l4'}),
            tuya.modernExtend.dpOnOff({dp: 5, endpoint: 'l5'}),
            tuya.modernExtend.dpOnOff({dp: 6, endpoint: 'l6'}),
            tuya.modernExtend.dpOnOff({dp: 0x65, endpoint: 'l7'}),
            tuya.modernExtend.dpOnOff({dp: 0x66, endpoint: 'l8'}),
        ],
    },
```


Koenkk:
Cool! Could you give an example on how this would look like for a device?

kirovilya:
added example to first post.

kirovilya:
but not everything works correctly yet :)

Another problem is that when you need to group several values into one dp package, you have to come up with something new...

Koenkk:
> Another problem is that when you need to group several values into one dp package, you have to come up with something new...

Could you give an example of this?

src/lib/tuya.ts
```diff
@@ -1312,8 +1312,211 @@ const tuyaExtend = {
...
+export interface TuyaLightArgs {
+    state: {dp: number, type: number, valueOn: [string | boolean, unknown], valueOff: [string | boolean, unknown], skip?: (meta: Tz.Meta) => boolean},
+    brightness: {dp: number, type: number, scale?: number | [number, number, number, number]},
+    max?: {dp: number, type: number, scale?: number | [number, number, number, number]},
+    min?: {dp: number, type: number, scale?: number | [number, number, number, number]},
+    endpoint?: string,
+}
+
+const tuyaModernExtend = {
+    enumLookup(args: TuyaEnumLookupArgs): ModernExtend {
```
Koenkk, line 1422: I would prefix all of these with `datapoint`, e.g. `datapointLight` because not all TuYa devices work via datapoints.

kirovilya: maybe a shorter prefix? for example "dp". e.g. "dpLight", "dpEnumLookup" :)

Koenkk: That's fine!

kirovilya:
There are some minor problems with the color:
in some cases, we also need to set the current color mode https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/legacy.ts#L8511

also, when controlled from HA, changes in two attributes (onoff and brightness) are received, so several sendDataPoints commands will be sent and this is not always processed correctly by the device - we need to be able to group them into one.

Koenkk:
The `tuya_light_wz5` converter is really complicated, let's not migrate that device to modernExtend for now.

kirovilya:
> The tuya_light_wz5 converter is really complicated, let's not migrate that device to modernExtend for now.

I think this needs to be done right away to get rid of legacy

> Isn't this also a problem now?

I don’t know if this is exactly it or not, but there is a problem here https://github.com/Koenkk/zigbee2mqtt/issues/19874

I tried just a complex case to understand what might be needed. 
Now I'll try to rewrite simple devices, not RGB...

Koenkk:
> I think this needs to be done right away to get rid of legacy

Agree, but I think it's best to start with some simple devices first, like some contact/temperature sensors and lights with onoff/brightness.

kirovilya:
I have a problem with modernExtend for Tuya... When there are several similar attributes, but related to different endpoints.
Now it looks like this:
```
endpoint: (device) => {
    return {'l1': 1, 'l2': 1};
},
datapoins = [
    [1, 'state_l1', tuya.valueConverter.onOff],
    [2, 'state_l2', tuya.valueConverter.onOff],
]
exposes: [
    tuya.exposes.switch().withEndpoint(`l1`),
    tuya.exposes.switch().withEndpoint(`l2`),
],
```
I'm trying to replace it with code like:
```
endpoint: (device) => {
    return {'l1': 1, 'l2': 1};
},
extend: [
    tuya.modernExtend.dpOnOff({dp: 1, endpoint: 'l1'}),
    tuya.modernExtend.dpOnOff({dp: 2, endpoint: 'l2'}),
],
```
And the problem is that in my case 2 toZigbee handlers are created with the same 'state' key, but for different enpoints.
z2m, when receiving a command in the topic '{"state_l2":"OFF"}', expands it to "state": "OFF", "endpoint": "l2"
and looks for a toZigbee handler with the key "state". finds the first instance and executes it.

In my case, this instance belongs to endpoint "l1" and it is not executed. the second handler will no longer be called.
Previously, the selection of the required DP by enpoint was carried out internally, but now the handler does not know about all DPs of the device, because they were removed.

Possible solutions:
1. modernExtend should continue to supplement the list of datapoints, and handlers should take them inside. then everything remains as before.
2. remove fake enpoints from the device definition and create extensions for each DP separately, calling them "state_l1", "state_l2". then the handler keys will be unique and with suffixes.
3. call several toZigbee handlers with the same keys, so that the decision to execute the command is made inside the handler. then it will be clear inside the handler for which DP should be executed and for which not.

Koenkk:
We can extend the toZigbee converter to have an additional option: `endpoint`, then z2m will only call that converter if the endpoint matches. Filtering in z2m has to be done [here](https://github.com/Koenkk/zigbee2mqtt/blob/56589dccf156915dcc78e9c7ffd669d7e5597db0/lib/extension/publish.ts#L211C54-L211C55)

kirovilya:
good idea. make this parameter optional

kirovilya:
@Koenkk Can you review the latest changes?

src/lib/tuya.ts
```diff
@@ -429,6 +430,91 @@ export const valueConverterBasic = {
     trueFalse: (valueTrue: number | Enum) => {
         return {from: (v: number) => v === valueTrue.valueOf()};
     },
+    color1000: () => {
```
Koenkk, line 433: @kirovilya will this converter be used? I propose to not add it here until it is removed from `legacy.ts` (`silvercrest_smart_led_string`)

src/lib/tuya.ts
```diff
@@ -1312,8 +1398,285 @@ const tuyaExtend = {
 };
 export {tuyaExtend as extend};
 
+
+function getHandlersForDP(name: string, dp: number, type: number, converter: Tuya.ValueConverterSingle,
```
Koenkk, line 1402: `converter: Tuya.ValueConverterSingle`, we also have `ValueConverterMulti`, could it also be added here?

src/lib/tuya.ts
```diff
@@ -1312,8 +1398,285 @@ const tuyaExtend = {
...
+        options: (definition) => {
+            const result = [];
+            if (name in utils.calibrateAndPrecisionRoundOptionsDefaultPrecision) {
+                const type = utils.calibrateAndPrecisionRoundOptionsIsPercentual(name) ? 'percentual' : 'absolute';
+                result.push(exposes.options.precision(name), exposes.options.calibration(name, type));
+            }
+            return result;
+        },
+        convert: (model, msg, publish, options, meta) => {
+            const result: KeyValue = {};
```
Koenkk, line 1417: This function can be simplified to:

```js
const result: KeyValue = {};
const dpValue = msg.data.dpValues.find((d) => d.dp === dp);
if (dpValue) {
    const value = converter.from(getDataValue(dpValue));
    const result = {[keyName]: value};
    if (name in utils.calibrateAndPrecisionRoundOptionsDefaultPrecision) {
        const valueNumber = utils.toNumber(value, keyName);
        result[keyName] = utils.calibrateAndPrecisionRoundOptions(valueNumber, options, keyName);
    }
    return result;
}
```

Koenkk:
Looks good, added 3 minor comments, after that it can be merged.

Really happy with this refactor 😄 

Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt/pull/20356

src/lib/tuya.ts
kirovilya: OK. I can temporarily comment out this converter until the next improvements.
I'll leave only the brightness and color temperature in "light" extend, without color.

src/lib/tuya.ts
kirovilya: yes, I can, but I don’t find an example of such a converter

Koenkk:
Thanks!

Kranzes:
Was this ever completed?

I expected the following to just work, but I still needed to set `tuyaDatapoints`

```typescript
tuya.modernExtend.dpBinary({
    name: "manual_mode",
    dp: 106,
    type: tuya.dataTypes.enum,
    valueOn: ["ON", 0],
    valueOff: ["OFF", 1],
}),
```

PR #6737, opened 2023-12-19
Distinguish between ZG-102ZL and ZG-102Z

martinsefcik:
ZG-102ZL is Luminance Door Sensor and ZG-102Z is only Door Sensor. Currently both models were identified as ZG-102ZL which was not correct.


Koenkk:
Thanks!

PR #6738, opened 2023-12-19
Add new device(GAS Valve) for ShinaSystem

shinasys:
- GAS Valve : GCM-300Z


src/devices/shinasystem.ts
```diff
@@ -694,6 +760,71 @@ const definitions: Definition[] = [
             await reporting.onOff(device.getEndpoint(3));
         },
     },
+    {
+        zigbeeModel: ['GCM-300Z'],
+        model: 'GCM-300Z',
+        vendor: 'ShinaSystem',
+        description: 'SiHAS Gas Valve',
+        fromZigbee: [fzLocal.GCM300Z_on_off, fz.battery],
```
Koenkk, line 768: Could you use the new modern extends instead? this replaces the fromZigbee/toZigbee/exposes and configure section. Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/54690b8f15235a94cc6d81799c9d31a01dbdd13d/src/devices/sonoff.ts#L371

shinasys: yes. It has been modified below.
gas_valve_state uses ['genOnOff', 'onoff'].
Changing this to 'binary' of modern extend cannot be used because it is read-only.
This attribute requires entity.command, not entity.write.
And 'onOff()' of modern extend was not used because the state description could not be changed.
So I used 'expose' instead of 'modern extend' for gas_valve_state.

Koenkk:
Thanks!

PR #6739, opened 2023-12-19
VOCKQJK11LM Improvements and modernExtend conversion (2/4)

sjorge:
Still very WIP but I though I'd open as draft already.

I plan on extending the modernExtend numeric with a 'calibration' argument that will toggle the calibration options.
It would be nice to have that code available for all numerics so we can easily enable where it makes sense.

### DIFF
--- a/src/devices/xiaomi.ts
+++ b/src/devices/xiaomi.ts
@@ -6,7 +6,10 @@ import * as ota from '../lib/ota';
 import * as constants from '../lib/constants';
 import * as reporting from '../lib/reporting';
 import extend from '../lib/extend';
-import {light, numeric, binary, enumLookup, forceDeviceType} from '../lib/modernExtend';
+import {
+    light, numeric, binary, enumLookup, forceDeviceType,
+    temperature, humidity,
+} from '../lib/modernExtend';
 const e = exposes.presets;
 const ea = exposes.access;
 import * as globalStore from '../lib/store';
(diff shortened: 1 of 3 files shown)


src/lib/modernExtend.ts
```diff
@@ -483,3 +483,34 @@ export function forceDeviceType(args: {type: 'EndDevice' | 'Router'}): ModernExt
...
     return {configure, isModernExtend: true};
 }
+
+// XXX: add options
+export function temperature(args?: Partial<NumericArgs>) {
+    return numeric({
+        name: 'temperature',
+        cluster: 'msTemperatureMeasurement',
+        attribute: 'measuredValue',
+        configureReporting: {minimumReportInterval: 10, maximumReportInterval: repInterval.HOUR, reportableChange: 100},
```
sjorge, line 493: I took the same values as in src/lib/reporting.ts

src/devices/xiaomi.ts
```diff
@@ -3109,20 +3112,20 @@ const definitions: Definition[] = [
...
         description: 'Aqara TVOC air quality monitor',
-        fromZigbee: [fz.battery, fz.temperature, fz.humidity, xiaomi.fromZigbee.aqara_opple],
+        fromZigbee: [fz.battery, xiaomi.fromZigbee.aqara_opple],
         meta: {battery: {voltageToPercentage: '3V_2850_3000'}},
-        exposes: [e.temperature(), e.humidity(), e.device_temperature(), e.battery(), e.battery_voltage()],
+        exposes: [e.device_temperature(), e.battery(), e.battery_voltage()],
         extend: [
             aqaraAirQuality(),
             aqaraVoc(),
+            temperature({endpointID: 1}),
```
Koenkk, line 3121: What is the data/database.db entry when pairing this device?

src/lib/modernExtend.ts
Koenkk: I think you also need ` scale: 100,` here?

sjorge: Yes, It's already fixed locally. I'll rebase and push, still working on the moving over the calibration code.

sjorge:
@Koenkk looks like somehow with the current code the options for temperature calibration are already somehow exposed. I don't understand where it's getting it from though.

Koenkk:
@sjorge I don't understand it either, but anyway the whole calibration should be moved up (outside of the converters)

sjorge:
> @sjorge I don't understand it either, but anyway the whole calibration should be moved up (outside of the converters)

Where do you want the calibration to end up ?

sjorge:
> There should be some "post process" function (out of scope for this PR), thinking about it now, you can remove the whole calibrate from your code, since the calibrate options will be discovered automatically based on the exposes.

Having a post process chain would be very nice, it could also fix things like `fz.SNZB_temperature()` as we can just add filter function to the chain.

So I just drop the calibrate stuff and the PR is good to go I guess ?

sjorge:
Cleaned up the calibrate stuff.

Koenkk:
Thanks!

PR #6741, opened 2023-12-19
fix: Use polling instead of reporting for TS011F_plug appVersion 192

sestriel:


Koenkk:
Without this change, do you get no values at all for energy, voltage, power and current?

sestriel:
Without change i didnt get energy and power. But instead i get these errors:
`Error: ConfigureReporting 0xa4c13834dff92aba/1 seMetering([], {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Timeout - 10402 - 1 - 39 - 1794 - 7 after 10000ms)
    at Timeout._onTimeout (/opt/iobroker/node_modules/zigbee-herdsman/src/utils/waitress.ts:64:35)
    at listOnTimeout (node:internal/timers:569:17)
    at processTimers (node:internal/timers:512:7)`
`TypeError: Cannot read properties of undefined (reading 'state')`

last one from tuya.js line 489
`            if (meta.state.state === 'OFF' || meta.state.current === 0) {
                result.power = 0;
            }`
didnt get last error when commenting out this block

Koenkk:
is this with z2m or iobroker?

sestriel:
Yes iobroker, am I on the wrong way?

sestriel:
Ok, im sorry. When i remove this device vom iobroker zigbee - so it pulls everything directly from herdsman-converters - everything works fine without changes here.

Koenkk:
iobroker should fix this error first: `TypeError: Cannot read properties of undefined (reading 'state')`, they should always pass a `meta.state`, please create an issue in their issue tracker to fix this first.

PR #6742, opened 2023-12-20
Bosch BTH-RA (radiator thermostat II): Fix inability for the heating algorithm to learn when using remote temp

DerDreschner:
The BTH-RA is "famous" for over-/undershooting temperatures. After some changes, [Bosch announced that there are no more planed changes to the heating algorithm](https://community.bosch-smarthome.com/t5/funktionen-und-nutzung/hinweis-und-umfrage-zum-heizk%C3%B6rper-thermostat-ii/m-p/42026/highlight/true#M11770). This resulted in quite a shitstorm where some folks plan to move away from the BTH-RA or Bosch Smart Home altogether.

While reading the threads, I recognized quite a pattern: A lot of the folks complaining use Home Assistant. As some of them posted valve opening charts as well (what seem to be impossible using the official Bosch bridge via API), they must use Zigbee2MQTT. I'm just not sure if they use a remote temperature reading to control the BTH-RA or not. Anyway, after experiencing an overshoot of 2°C during nighttime the last days, I tried to see if there is something wrong with the converter code.

And I guess I made some success. The thermostat seem to react at ±0.4° and ±0.1°. But the code in the herdsman converter rounds the remote temp value to ±0.5°. This appears to remove the ability for the heating algorithm to learn how much throughput would be enough to achieve the given temperature. It's more of a "open the valve complete" or nothing. It's far better with the changes applied (target temperature didn't changed in the shown time-frame).

Screenshot from 2023-12-20 06-37-06

As I don't work for https://github.com/BoschSmartHome and they seem quite in-cooperative when it comes to open source solutions instead of their proprietary bridge, all of this are just best guesses from my side. I can't say if these changes will really resolve, or at least improve, the over-/undershoot thematic. But if it does, well... Then Bosch should overthink their policies as this whole thematic made quite some negative PR. In case you need more man-power for that: hit me up, I'm open for a new workplace!

This is a list of the changes:
- Change the remote temperature code to round the value to ±0.1° instead of ±0.5°. Additionally, allow even more exact temperature readings to be input by round them before transmission. This removes the need for rounding in the automation templates when using MQTT publish messages.
- Change the descriptions for local and remote temperature. Before, it wasn't clear that the BTH-RA uses a fallback to the device temperature reading when there is no new remote temperature reported within 30 minutes. Personally, I used the following temperature trigger to achieve this. Will update the device doc in the next days to include that.
- Don't expose the "auto" mode as there is no way to control the applied heating schedule via Z2M anyway (at least right now).
- Other small changes (set "heat" status with any valve opening value > 0 instead of >= 10, use ZCL to get the manufacturer code, some refactoring and added value checks).

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -945,7 +945,7 @@ const definitions: Definition[] = [
             e.numeric('remote_temperature', ea.ALL)
                 .withValueMin(0)
                 .withValueMax(35)
-                .withValueStep(0.01)
+                .withValueStep(0.1)
                 .withUnit('°C')
                 .withDescription('Input for remote temperature sensor. ' +
                     'This must be set at least every 30 minutes to prevent fallback to internal temperature reading!'),
@@ -956,7 +956,7 @@ const definitions: Definition[] = [
             e.numeric('display_brightness', ea.ALL)
                 .withValueMin(0)
                 .withValueMax(10)
-                .withDescription('Specifies the brightness value of the display'),
+                .withDescription('Specifies the brightness level of the display'),
             e.enum('displayed_temperature', ea.ALL, Object.keys(displayedTemperature))
                 .withDescription('Temperature displayed on the thermostat'),
             e.child_lock().setAccess('state', ea.ALL),
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -589,7 +589,7 @@ const fzLocal = {
             const result: KeyValue = {};
             const data = msg.data;
             if (data.hasOwnProperty(0x4040)) {
-                result.remote_temperature = utils.precisionRound(data[0x4040] / 100, 2);
+                result.remote_temperature = utils.precisionRound(data[0x4040] / 100, 1);
             }
             if (data.hasOwnProperty(0x4042)) {
                 result.window_open = (Object.keys(stateOffOn)[data[0x4042]]);
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -599,7 +599,7 @@ const fzLocal = {
             if (data.hasOwnProperty(0x4020)) {
                 const demand = data[0x4020] as number;
                 result.pi_heating_demand = demand;
-                result.running_state = demand >= 10 ? 'heat' : 'idle';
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

mldytech:
@DerDreschner Thank you - I really appreciate your work and awareness of the suboptimal regulation of the thermostats. Sorry if this is the wrong place to ask.

As far as I understand, your made changes won't improve the situation when adapting the temperature via. the classic offset function - right? Would you recommend using the remote temperature function or the offset functionality? 
Currently I'm struggling to find a satisfying solution..

Best regards, mldy

DerDreschner:
@mldytech : Well, would be better if Bosch would acknowledge the sub-optimal regulation as I can't do that much about it. :joy: And don't worry, there's not a wrong place to ask. :smile: 

Yes, your understanding is correct. This PR only affects the remote temperature functionality.

It's hard to recommend anything as it depends on the concrete conditions you have. We have to keep in mind that ±0.4° and ±0.1°C are quite small changes and a lot of temperature sensors don't report that often by default as it drains the battery quite fast. So, my considerations are the following:
- If you're able to change the reporting for your sensor and being comfortable with the higher battery drain - use the external temperature functionality and see if it enhances your situation.
- If you can't change the reporting settings or want maximum energy efficiency on your temperature sensors, you should use the offset. My feeling is that you will have better results as the Thermostat II is able to check the internal temperature more often due to the bigger batteries then e.g. my Aqara temperature sensors.

Another thing to try is "[Better Thermostat](https://better-thermostat.org/)", at least if you're using Home Assistant and being comfortable with using [HACS](https://hacs.xyz/). In my own experience, I were able to prevent a too slow thermostat reaction if the temperature is about to fall below the set-point. The slight over-shoot is still there, but at least it's an improvement.

But, as a lot of folks out there, I would expect the Bosch Thermostat II to do that out-of-the-box...

bnowakow:
@DerDreschner thanks for keeping up with this. Sorry for lame question. Up until now I was setting mode to auto and set eg. 22 degrees during day and auto to 19 degrees during night. After this change should I change it to heat to 22 degrees and 19 degrees respectively?

Or in another words if remote temperature is at 23 and target is heat 22 it would't open valve since temperature is higher than target?

PR #6746, opened 2023-12-20
fix(ignore): Small refactor to modernExtend reporting config

Koenkk:
Some small improvements to ensure typing and reduce verbosity.

CC: @sjorge


sjorge:
LGTM, allows us to skip importing contants in a few places!

PR #6748, opened 2023-12-21
Fix mcuVersionRequest log flood

donbobka:
I think it's not necessary to have onEvent callback for these blinds (i copy pasted it from another blinds). So it fixes the problem of log flood.


Koenkk:
Thanks! I've improved the fix a little bit, please check if everything is OK

PR #6754, opened 2023-12-22
fix: ubisys H10 vacation_mode after firmware upgrade

sjorge:
Also a ton of bug fixes this firmware, everyone should probably upgrade.

This needs https://github.com/Koenkk/zigbee-herdsman/pull/835 to be merged.

I plan on doing a PR later when I have a bit more time introduce some of the missing functionality now that I have the technical reference. (And convert some of this to modernExtends while I'm digging around in there)


Koenkk:
Thanks!

PR #6758, opened 2023-12-23
Update philips.ts to support Hue GU5.3/MR16/LCG005

thk-socal:


Koenkk:
Thanks!

src/devices/philips.ts
mundschenk-at, line 1: I am a bit surprised that these would be separate models - all GU5.3 devices are in the MR16 form factor and they were just introduced, so it is doubtful that they have a new revision/hardware platform already.

thk-socal: Well my MR16, here in the US, were LCG005 model number. Not sure why, but that is the model reported by zigbee2mqtt when joining them.

mundschenk-at: Interesting. I think we should join the definitions though, no?

Issue #6752, opened 2023-12-22
In the Z2M Edge version (16.15.1 or 16.16?) BTH-RA gives errors in configuration


bullmastiffo:
Zigbee Model
RBSH-TRV0-ZB-EU
Zigbee Manufacturer
BOSCH
Description
Radiator thermostat II
Firmware build date
20220627
Firmware version
3.02.05

In the Edge version logs: 
```text
Zigbee2MQTT:info  2023-12-22 10:45:01: Configuring 'Bosch Valve Living Room'
2023-12-22T09:45:01.984Z zigbee-herdsman:controller:endpoint Bind 0x18fc260000a197e4/1 genPowerCfg from '0xbc026efffe228982/1'
2023-12-22T09:45:01.984Z zigbee-herdsman:adapter:ezsp:driv ZDO Bind_req params: {"sourceEui":{"_value":{"type":"Buffer","data":[24,252,38,0,0,161,151,228]}},"sourceEp":1,"clusterId":1,"destAddr":{"addrmode":3,"ieee":{"_value":{"type":"Buffer","data":[188,2,110,255,254,34,137,130]}},"endpoint":1}}
2023-12-22T09:45:01.986Z zigbee-herdsman:adapter:ezsp:ezsp ==> setExtendedTimeout: {"remoteEui64":{"_value":{"type":"Buffer","data":[188,2,110,255,254,34,137,130]}},"extendedTimeout":true}
2023-12-22T09:45:01.987Z zigbee-herdsman:adapter:ezsp:ezsp ==> {"_cls_":"setExtendedTimeout","_id_":126,"_isRequest_":true,"remoteEui64":{"_value":{"type":"Buffer","data":[188,2,110,255,254,34,137,130]}},"extendedTimeout":true}

...

2023-12-22T09:45:17.187Z zigbee-herdsman:controller:endpoint Bind 0x18fc260000a197e4/1 hvacThermostat from '0xbc026efffe228982/1' failed ({"address":42360,"clusterId":32801,"sequence":29} after 10000ms)
Zigbee2MQTT:error 2023-12-22 10:45:17: Failed to configure 'Bosch Valve Living Room', attempt 3 (Error: Bind 0x18fc260000a197e4/1 hvacThermostat from '0xbc026efffe228982/1' failed ({"address":42360,"clusterId":32801,"sequence":29} after 10000ms)
[...]

PR #6761, opened 2023-12-23
Bosch BTH-RA: Possible workaround for OTA problems and adding remote valve calibration

DerDreschner:
In the last few months, there were several people with problems during/after the OTA process (see i.e. https://github.com/Koenkk/zigbee2mqtt/issues/17740). The thing is, that the native Bosch implementation works totally different then Z2M. With the native system, the new firmware gets uploaded to the device automatically within 2 days and will only be executed when triggered manually via the app. This is the same for all devices in the Bosch eco-system.

After the logfiles from https://github.com/Koenkk/zigbee-herdsman-converters/issues/6752 (which seems to be affected as the displayed firmware version in Z2M doesn't match the actual firmware version), I suspect some kind of timeout issue in heavily congested networks as the device never got an `upgradeEndRequestResponse`. To mitigate this, I advanced the timeout to 1 hour and will continue to monitor the situation.

Additionally, I added support for remote valve calibration (`valve_adapt_process`). This is necessary after an update - so, the possibility to do that from Z2M itself instead of walking from device to device is quite handy. It's required to have https://github.com/Koenkk/zigbee-herdsman/pull/836 merged. I hope this false/true implementation is okay as there is no pure "one button triggers something" in the UI. Since a calibration is only possible when the device requests for it, I've added a value to see if that is the case or not (`valve_adapt_status`). In the future, I'm thinking about triggering the calibration process automatically after an update. In the meantime, an automation in i.e. HA should be sufficient in case someone is missing this feature. 

Other changes:
- Adjusted the default reporting settings to match the default Bosch values. I only changed the already supported attributes. Bosch itself sets up quite more attributes no-one identified yet. Those are not added with this PR - just to avoid any confusion.
- Added support for monitoring the setpoint source (`setpoint_change_source`)

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -970,6 +970,13 @@ const definitions: Definition[] = [
             await reporting.thermostatKeypadLockMode(endpoint);
             await reporting.batteryPercentageRemaining(endpoint);
 
+            // Report setpoint_change_source
+            await endpoint.configureReporting('hvacThermostat', [{
+                attribute: {ID: 0x0030, type: Zcl.DataType.enum8},
+                minimumReportInterval: 0,
+                maximumReportInterval: constants.repInterval.HOUR * 12,
+                reportableChange: 1,
+            }]);
             // report operating_mode (system_mode)
             await endpoint.configureReporting('hvacThermostat', [{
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #6765, opened 2023-12-24
Fix for passing decimal hue values to Silvercrest smart led string

pierrebateau:
This PR fixes an issue when sending a decimal hue value to the Silvercrest smart led string, causing weird artifacts. The hue will now be rounded, just like the saturation and brightness.


Koenkk:
Bedankt!

PR #6766, opened 2023-12-24
Add full support for TRV602

lompi:
This pull request adds full support for the TRV602 device. Until now, support for this device was added using the existing configuration, but this device does not support boost mode and gives the possibility to control screen orientation and display brightness.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -11,7 +11,7 @@ import * as globalStore from '../lib/store';
 import {ColorMode, colorModeLookup} from '../lib/constants';
 import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
-import {KeyValue, Definition, Tz, Fz, Expose, KeyValueAny, KeyValueNumberString, KeyValueString} from '../lib/types';
+import {KeyValue, Definition, Tz, Fz, Expose, KeyValueAny, KeyValueString} from '../lib/types';
 import {electricityMeter, onOff} from '../lib/modernExtend';
 
 const e = exposes.presets;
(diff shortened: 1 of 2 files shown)


src/devices/tuya.ts
```diff
@@ -3240,6 +3239,78 @@ const definitions: Definition[] = [
...
+                .withRunningState(['idle', 'heat'], ea.STATE),
+            ...tuya.exposes.scheduleAllDays(ea.STATE_SET, 'HH:MM/C HH:MM/C HH:MM/C HH:MM/C'),
+            e.enum('display_brightness', ea.STATE_SET, ['high', 'medium', 'low']).withDescription('Display brightness'),
+            e.enum('screen_orientation', ea.STATE_SET, ['up', 'right', 'down', 'left']).withDescription('Screen orientation'),
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [1, null,
+                    {
+                        from: (v) => {
```
Koenkk, line 3277: Could you move this to a value converter in `lib/tuya.ts` and also use it for `TS0601_thermostat_1`?

lompi: Ok. I moved that. I hope that it's ok, because js isn't my technology and I did it by feel.

Koenkk:
Thanks!

BAUBLITZ:
the TRV602 has an additional function that is still missing in tuya.js.
The “eco” and “comfort” functions. In "eco" the thermostat regulates the temperature by completely opening and closing the valve. With the "comfort" function, the thermostat regulates the temperature much more precisely because the valve is opened or closed to different extents depending on the difference between the set and the actual temperature.

To do this, the following must be added under exposes:

e.enum('mode', ea.STATE_SET, ['comfort', 'eco']).withDescription('Hysteresis - comfort > switches off/on exactly at reached temperature with valve smooth from 0 to 100%, eco > 0.5 degrees above or below, valve either 0 or 100%'),

an under tuyaDatapoints:

[114, 'mode', tuya.valueConverterBasic.lookup({'comfort': tuya.enum(0), 'eco': tuya.enum(1)})],

Can someone still integrate this?

Koenkk:
@BAUBLITZ done!

PR #6767, opened 2023-12-24
Added device id for Kojima Home door sensor

ahydrax:
Hi there!

This PR addresses issue https://github.com/Koenkk/zigbee2mqtt/issues/19939
Nothing new, just added device id which is already supported https://www.zigbee2mqtt.io/devices/ZG-102ZL.html


Koenkk:
Thanks!

PR #6769, opened 2023-12-25
feat!: Refactor calibrate and precision round options

Koenkk:
- Detect calibrate/precision options from exposes (and remove options from converters)
- Do calibration/precision round as a post process step instead of in converters

This simplifies the code and improves consistency

TODO: 
- [x] Update Z2M to call `postProcessFromZigbeeMessage`
- [x] Run docgen to see if anything was missed

CC: @sjorge @kirovilya


sjorge:
can we also add a generic 'postProcessing' hook to modernExtends (ideally an list of ones to process in order) ?
Ideally the calibration one is inserted at the end of the list we get from the modernExtend.

That should make things like  https://github.com/Koenkk/zigbee-herdsman-converters/blob/f194f49f3de5e11c3c571d9618383d092851ebc0/src/converters/fromZigbee.ts#L5717-L5731 obsolete.

Koenkk:
I think for these case we need to be able to pass a custom `from`/`to` conversion method to the numeric extend

PR #6776, opened 2023-12-27
Update shinasystem.ts

shinasys:
fix typo


src/devices/shinasystem.ts
```diff
@@ -740,7 +740,7 @@ const definitions: Definition[] = [
                 valueMin: 1,
                 valueMax: 540,
                 valueStep: 1,
-                unit: 'Miniute',
+                unit: 'Minute',
```
Koenkk, line 743: The unit should be [min](https://en.wikipedia.org/wiki/Minute)

shinasys:
@Koenkk I don't know why it failed? Please confirm.

Koenkk:
It's not because of this change, thanks!

PR #6777, opened 2023-12-27
fix: Fix power source/type for Xiaomi JTQJ-BF-01LM/BW

hongliangl:
for https://github.com/Koenkk/zigbee2mqtt/issues/18597

### DIFF
--- a/src/devices/xiaomi.ts
+++ b/src/devices/xiaomi.ts
@@ -2318,8 +2318,8 @@ const definitions: Definition[] = [
             e.numeric('gas_density', ea.STATE), e.enum('selftest', ea.SET, ['']),
         ],
         configure: async (device, coordinatorEndpoint, logger) => {
-            device.powerSource = 'Battery';
-            device.type = 'EndDevice';
+            device.powerSource = 'Mains (single phase)';
+            device.type = 'Router';
             device.save();
         },
     },


hongliangl:
I've discovered that this device requires a 12V 70mA DC adapter instead of direct AC power (`'Mains (single phase)'` in code), which means that the power source should be `DC source`. I think I made the wrong code change. Could you give some help to set the power source to `DC source` for the device? Thanks a lot! @Koenkk

Koenkk:
Isn't this device battery powered?

hongliangl:
> Isn't this device battery powered?

For the device produced in mainland China, it is DC powered, not battery powered. To be compatible with the device produced outside mainland China, I think it is okay to keep the existing code. Maybe I could add a external converter for this device to set it power to DC.

VivantSenior:
@hongliangl, I bought two of them and they are AC powered and work as routers using [this external converter](https://github.com/Koenkk/zigbee2mqtt/issues/18597#issuecomment-1925305670).

> I've discovered that this device requires a 12V 70mA DC adapter instead of direct AC power (`'Mains (single phase)'` in code), which means that the power source should be `DC source`. I think I made the wrong code change. Could you give some help to set the power source to `DC source` for the device? Thanks a lot! @Koenkk

12V DC 70 mA is just working voltage / current (according to the text at the back of the device). Out of the box I received 230V power adapter with it - therefor powered from mains and can be a router in the zigbee.

> Isn't this device battery powered?

No, it gets power from mains.
image

I was able to find a [manual](https://www.manualslib.com/manual/1620622/Honeywell-Jtqj-Bf-01lm-Bw.html) for the device and it mentions powering the device from mains several times:
[jtqjbf01lmbw.pdf](https://github.com/Koenkk/zigbee-herdsman-converters/files/14154287/jtqjbf01lmbw.pdf)
image
image

As a counter argument:
image

Supposedly the battery should be sell together with the device but there is no place to put the battery in ;-)

PR #6780, opened 2023-12-28
Bosch BTH-RA: Automatic valve adaptation after firmware update

DerDreschner:
This PR is a continuation of #6761. It triggers the valve adaptation process automatically after an successful firmware update on the Bosch BTH-RA thermostat.

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -11,7 +11,7 @@ import {Tz, Fz, Definition, KeyValue} from '../lib/types';
 const e = exposes.presets;
 const ea = exposes.access;
 
-const manufacturerOptions = {manufacturerCode: Zcl.ManufacturerCode.ROBERT_BOSCH_GMBH};
+export const manufacturerOptions = {manufacturerCode: Zcl.ManufacturerCode.ROBERT_BOSCH_GMBH};
 
 const sirenVolume = {
     'low': 0x01,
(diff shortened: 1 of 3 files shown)


Koenkk:
- Why is this needed?
- Can't we make a `toZigbee` converter for this and let the user trigger it manually?

DerDreschner:
It's not really needed, more like a "quality of life" improvement. As you mentioned, I already implemented a `toZigbee` converter to trigger the adaptation manually. But I'm not sure if everyone knows that a valve adaptation is necessary after an upgrade. If it's not done right after the update, the room could overheat as the valve is opened to 100% until the adaptation process is triggered. Therefore, I personally would prefer an automatic solution. But feel free to close the PR if you think otherwise. :smiley:

Koenkk:
Does the thermostat send a `deviceAnnounce` when batteries are re-inserted? If yes, we can trigger on that instead in an `onEvent` (I expect you also get a `deviceAnnounce` after the update, as the device typically sends that when booting). [Example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/c28c703469fbaa9279b4e8a11dd87a4f2c341f6a/src/devices/ikea.ts#L30) on how to do something on a `deviceAnnounce`

u20p17:
would be good if the adoption could be done automatically after a firmware update... after the last update of the bathrooms thermostat  the temperatur rised above 30 deg until I recognized it...  haha

DerDreschner:
@u20p17 : Someone complained about the same issue on the bosch community. I totally forgot about it and will work on it soon.

@Koenkk: Using the `deviceAnnounce` as trigger is a good idea. That should work, not only after firmware updates, but also e.g. changed batteries.

@burmistrzak : Do you plan to implement those changes on your rewrite? Or can/should I implement it?

PR #6781, opened 2023-12-28
Fix avatto wall socket identification

tiagobnoronha:
Avatto wall socket was been identified as smart plug. So, Manufacture Model "_TZ3000_pvlvoxvt" was added in tuya.ts at definitions array for device model 'TS011F_2_gang_2_usb_wall'. Before that, the device was being identified as 'TS011F_plug_1'


Koenkk:
Thanks!

PR #6783, opened 2023-12-28
Change model of the gas sensor with manufacture model _TZE204_yojqa8xn

tiagobnoronha:
The gas sensor with manufacture model _TZE204_yojqa8xn was in definition array as model TS0601_gas_sensor_1, but some information, such as gas value, was not exposed. I have tested the sensor with the converter to TS0601_gas_sensor_2, and all features are working (gas value, melody selection, alarm time, and preheat status).


Koenkk:
Thanks!

PR #6784, opened 2023-12-29
add powerOnBehavior: false

nukusinji:
add powerOnBehavior: false

light switch of dawon dns is not support poweronbehavior.


Koenkk:
Thanks!

PR #6786, opened 2023-12-29
Fixes for inovelli parameters.

efaden:
Fixing parameter names for the fan controller, and also adding parameter 26 to the regular 2-in-1

### DIFF
--- a/src/devices/inovelli.ts
+++ b/src/devices/inovelli.ts
@@ -146,7 +146,7 @@ const attributesToExposeList = (ATTRIBUTES: {[s: string]: Attribute}, exposesLis
  * Some of the descriptions, max, min or value properties may be overridden for each device
  */
 const COMMON_ATTRIBUTES: {[s: string]: Attribute} = {
-    dimmingSpeedUpRemote: {
+    speedUpRemote: {
         ID: 1,
         dataType: UINT8,
         min: 0,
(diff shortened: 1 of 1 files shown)


Koenkk:
@InovelliUSA please review

InovelliUSA:
Lot's of changes here so I'll need a little bit to test them. One thing that we have been doing is not putting access to parameter 26 in drivers, converters, etc. This is because changing the setting can damage the device if a certain type of load is connected to it. You can change this config option with a button combo though.

efaden:
Didn't realize it was intentional.

jsievenpiper:
Sorry if this is the wrong place for discussion, but since it's been at least mentioned here...

@InovelliUSA would you be willing to allow for the leading/trailing edge dimming settings for something like the canopy module? As a Z2M user, I'm not actually sure how to change this setting on the canopy module. Trailing edge dimming was an absolute godsend for the switches. Button taps on the switches seem a reasonable compromise, but not sure if/how to accomplish that on the canopy (and working in the fan canopy with live power is arguably sketchier than button presses on the switch 🙂).

Would love your thoughts!

InovelliUSA:
@jsievenpiper yeah, I am ok with this parameter being added to the vzm36. Since it is more difficult to change with the fan module and most people will only connect the load to standard light bulbs. Maybe we should close this PR and create a new one with the parameter added for the vzm36?

PR #6789, opened 2023-12-29
Add 'schedule', 'calibration' and more to ZWT07

tvoyat:
Add schedule (ZWT07_schedule converter) , child_lock and some other changes to thermostat ZWT07 ( TS0601/_TZE200_g9a3awaj)


src/devices/tuya.ts
```diff
@@ -4229,22 +4229,41 @@ const definitions: Definition[] = [
...
-        onEvent: tuya.onEvent({timeStart: '1970'}),
+        onEvent: tuya.onEventSetLocalTime,
         configure: tuya.configureMagicPacket,
         exposes: [
-            e.climate().withSetpoint('current_heating_setpoint', 5, 60, 0.5, ea.STATE_SET)
+            tuya.exposes.frostProtection(),
+            e.battery(),
+            e.child_lock(),
+            e.text('schedule_weekday', ea.STATE_SET).withDescription('Workdays (6 times `hh:mm/cc.c`)'),
+            e.text('schedule_weekend', ea.STATE_SET).withDescription('weekend (2 times `hh:mm/cc.c)`'),
```
Koenkk, line 4239: What divisor should be used between the times?

src/devices/tuya.ts
```diff
@@ -4229,22 +4229,41 @@ const definitions: Definition[] = [
         description: 'Wall-mount thermostat',
         fromZigbee: [tuya.fz.datapoints],
         toZigbee: [tuya.tz.datapoints],
-        onEvent: tuya.onEvent({timeStart: '1970'}),
+        onEvent: tuya.onEventSetLocalTime,
```
Koenkk, line 4232: Why was this change needed?

src/lib/tuya.ts
```diff
@@ -855,6 +855,64 @@ export const valueConverter = {
             return {preset: presetLookup[v], system_mode: systemModeLookup[v]};
         },
     },
+    ZWT07_schedule: {
```
Koenkk, line 858: This looks the same as the already existing `ZWT198_schedule`

src/devices/tuya.ts
tvoyat: Values are space separated.
<code>Received MQTT message on 'zigbee2mqtt/zwt07/set' with data '{"schedule_weekday":"06:00/21.0°C 08:00/18.0°C 11:30/21.0°C 12:30/18.0°C 17:00/21.0°C 22:30/18.0°C"}'</code>
Note that this value is pre-populated in the GUI (or HA) with the current value of zwt07. Changes are more intuitive.

src/devices/tuya.ts
tvoyat: Without this change, ZWT07 and zigbee2mqtt enter infinite loop with :
<code>No converter available for 'ZWT07' with cluster 'manuSpecificTuya' and type 'commandMcuVersionResponse' and data '{"seq":28928,"version":65}'</code>
See example output : https://gist.github.com/tvoyat/c70369ae3093f217ff9bf2691bedee0f
I don't know why it doesn't do it anymore with `tuya.onEventSetLocalTime`...

src/lib/tuya.ts
tvoyat: ZWT198_schedule use hard-codeid dpId (`const dpId = 109;`) different and incompatible from ZWT07 (67).
I didn't want to introduce any side effects to the ZWT198 devices by modifying this converter.

Koenkk: Then I propose to move the logic into a function and add the dpid as an argument

tvoyat: Logic in the new converter is slightly different and doesn't use `dpId` at all (acts like simple valueconverter) and is more generic.
I think `ZWT07_schedule` can replace `ZWT198_schedule` as is (but I have nothing to check) and is a better way than extracting the function

Koenkk: will address in other comments.

src/lib/tuya.ts
```diff
@@ -855,6 +855,64 @@ export const valueConverter = {
...
+        from: (value: number[], meta: Fz.Meta, options: KeyValue) => {
+            const programmingMode = [];
+            for (let i = 0; i < 8; i++) {
```
Koenkk, line 861: This for loop is exactly the same, I propose to move it to a function `parseScheduleProgrammingMode`

src/lib/tuya.ts
```diff
@@ -855,6 +855,64 @@ export const valueConverter = {
...
+            let scheduleWeekend :string;
+            if (meta.message.hasOwnProperty('schedule_weekday')) {
+                scheduleWeekday = value;
+                scheduleWeekend = meta.state['schedule_weekend'] as string;
+            } else {
+                scheduleWeekday = meta.state['schedule_weekday'] as string;
+                scheduleWeekend = value;
+            }
+
+            function scheduleToRaw(key: string, input: string, number: number, payload: number[], meta: Tz.Meta) {
```
Koenkk, line 884: This function is exactly the same and does not have to be duplicated.

PR #6791, opened 2023-12-29
lixee.ts : explode relais value for virtual relays

mildis:
Create 8 additional entries 'relais1' to 'relais8' mapping the virtual relays states.
0 is open, 1 is closed.
relais1 reflects the mechanical real relay.
Maybe @fairecasoimeme would like to take a look.


Koenkk:
Could you also add this to the `exposes`?

Koenkk:
thanks!

PR #6792, opened 2023-12-30
Update neo.ts

siellet:
Good morning,

I just purchased the Neo Model NAS-AB06B2 siren, the latest version. Unfortunately the latest version does not appear in neo.ts, but the old model does. Suddenly my new Neo siren model appears unknown in my Zigbee2mqtt. it is the same characteristics as the first except that the name of the Zigbee Builder has changed and therefore not currently recognized. I made the modifications in the neo.ts file, could you validate it so that in my Zigbee2mqtt my siren is recognized. THANKS


Koenkk:
Thanks!

PR #6796, opened 2023-12-30
Fix: Disable powerOnBehavior for Sunricher ZG9101SAC-HP-Switch

slugzero:
Configuration of the Sunricher ZG9101SAC-HP-Switch fails with error
`Failed to configure 'Light1', attempt 1 (Error: Read 0x000xxxxxxxxxxx/1 genOnOff(["startUpOnOff"], {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')`, so I disabled powerOnBehavior.

### DIFF
--- a/src/devices/sunricher.ts
+++ b/src/devices/sunricher.ts
@@ -207,7 +207,7 @@ const definitions: Definition[] = [
         model: 'ZG9101SAC-HP-Switch',
         vendor: 'Sunricher',
         description: 'Zigbee AC in wall switch',
-        extend: [onOff()],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #6802, opened 2024-01-01
Fix HG06467 effect firework

popy2k14:
The firework effect is the same as the updown.
It seems it's an copy and paste issue.
Ref: [[Z2M Thread](https://github.com/Koenkk/zigbee2mqtt/issues/20462)]


Koenkk:
Thanks!

PR #6804, opened 2024-01-02
Moes 2 Gang switch incorrectly identified

liamstears:
I have a 2 gang Moes light switch that works correctly but is identified as Moes ZS-US2-WH-MS but it should be Moes ZS-EUB_2gang

I hope I have done this correctly and have included all information needed to get this correctly identified in Z2M however please let me know if I have done anything wrong or if anything else is needed

`const {onOff, onOff} = require('zigbee-herdsman-converters/lib/modernExtend');

const definition = {
zigbeeModel: ['TS0012'],
model: 'TS0012',
vendor: '_TZ3000_18ejxno0',
extend: [onOff({powerOnBehavior: false}), onOff({powerOnBehavior: false})],
};

module.exports = definition;`

I have created pull request both here and on [zigbee2mqtt.io](https://github.com/Koenkk/zigbee2mqtt.io)

ZS-EUB_2gang


src/devices/moes.ts
```diff
@@ -388,6 +388,24 @@ const definitions: Definition[] = [
             device.save();
         },
     },
+    {
+        fingerprint: [{modelID: 'TS0012', manufacturerName: '_TZ3000_18ejxno0'}],
```
Koenkk, line 392: Please add this as a new `tuya.whitelabel` to `TS0012` instead.

liamstears: Sorry but I am new to this and don't really understand what to do...

Koenkk:
@hndrk since you also have a `_TZ3000_18ejxno0`, can you confirm the model should be `ZS-EUB_2gang` and not `ZS-US2-WH-MS` as in https://github.com/Koenkk/zigbee-herdsman-converters/pull/6804

liamstears:
Is there an update on when this can be added? Looks like some checks did not go through?

Koenkk:
Didn't get a reply so will merge this, thanks!

anergiz:
This unfortunately broke a similar device I have, it is also Moes two gang switch but it only reports 2 action per button, single and double click. I have a custom converter defined as
```
const e = exposes.presets;
const ea = exposes.access;

const definition = {
    // Since a lot of TuYa devices use the same modelID, but use different datapoints
    // it's necessary to provide a fingerprint instead of a zigbeeModel
    fingerprint: [
        {
            modelID: 'TS0021',
            manufacturerName: '_TZ3210_3ulg9kpo',
        },
    ],
    model: 'TS0021_new',
    vendor: 'TuYa',
    description: 'Wireless switch with 4 buttons',
    //fromZigbee: [fz.tuya_on_off_action, fz.battery],
    //to_zigbee: [],
    //onEvent: tuya.onEventSetTime, // Add this if you are getting no converter for 'commandMcuSyncTime'
    exposes: [e.battery(), e.action(['1_single','1_double','1_hold','2_single','2_double','2_hold']),
    ],
    meta: {
        // All datapoints go in here
        tuyaDatapoints: [
                [1, 'action', tuya.valueConverterBasic.lookup({'1_single': tuya.enum(0), '1_double': tuya.enum(1)})],
                [2, 'action', tuya.valueConverterBasic.lookup({'2_single': tuya.enum(0), '2_double': tuya.enum(1)})],
        ]
    },
};

module.exports = definition;
```

But when I update zigbee2mqtt to latest containing this change, some of these devices stopped working. I don't quite understand how fingerprints conflict in this case though.

eyesonly09:
I was wrong, although the device looked the same, the issue is for devices identified as
Zigbee Model
TS0042
Zigbee Manufacturer
_TZ3000_kt7obmnn
https://www.zigbee2mqtt.io/devices/TS0042.html#tuya-ts0042

But still the suffix on the manufacturer doesn't match to the change, I am not sure why the device identification changes in this case.

Koenkk:
@eyesonly09 this is only a cosmetic change so it cannot break as you described, please create an issue in the z2m issue tracker instead

PR #6806, opened 2024-01-02
BMCT-SLZ: fix z2m crash if device is not reachable on startup

slugzero:
I noticed that in some cases z2m crashes right after startup, which is caused by an uncaught exception when there is no route to the device. So we better catch everything.

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -1292,9 +1292,9 @@ const definitions: Definition[] = [
             if (type === 'start' || type === 'deviceOptionsChanged') {
                 if (options.device_mode) {
                     const index = utils.getFromLookup(options.device_mode, stateDeviceMode);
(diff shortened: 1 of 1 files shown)


slugzero:
I agree, this should not be necessary. For other errors (`UNSUPPORTED_ATTRIBUTE`, `MAC no ACK`) the error was caught properly. If you'd like to dive deeper into this, I shared the full log privately: https://github.com/slugzero/z2m-logs/blob/main/z2m-20240102.log

However, for that particular device, I'd like to add these additional `catch` statements anyway: They also suppress some unnecessary error messages during configuring. Since the device must be configured to behave either as roller shutter or switch, one of the `read` commands always returns `UNSUPPORTED_ATTRIBUTE`.

Koenkk:
I see, then this can be merged indeed!

PR #6809, opened 2024-01-02
Ubisys J1: Adapt exposes according to CoveringType

slugzero:
I have a little bit of spare time to fix some of the issues in my setup wichich are only mildly annoying: one is that the Ubisys J1 is exposing lift and tilt, even if it is configured as roller shutter with lift only.

This PR fixes this by adapting the exposes dynamically according to the `windowCoveringType`.

Note: I believe this could be useful for other devices as well since `windowCoveringType` is a standard ZCL type. Therefore it might be good to add it directly to the `cover()` expose. But I don't want to break things for other devices without testing

### DIFF
--- a/src/devices/ubisys.ts
+++ b/src/devices/ubisys.ts
@@ -815,11 +815,38 @@ const definitions: Definition[] = [
         toZigbee: [tz.cover_state, tz.cover_position_tilt, tz.metering_power,
             ubisys.tz.configure_j1, ubisys.tz.configure_device_setup,
             tz.currentsummdelivered],
-        exposes: [
-            e.cover_position_tilt(),
-            e.power().withAccess(ea.STATE_GET),
-            e.energy().withAccess(ea.STATE_GET),
-        ],
+        exposes: (device, options) => {
+            const coverExpose = e.cover();
+            if (device) {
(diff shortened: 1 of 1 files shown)


sjorge:
Nice! This was actually on my very very long list of things I wanted to eventually tacked, every since I did dynamic expose for the SMaBiT 902010/32. Thanks for tackling this one. I can remove it from the list :)

src/devices/ubisys.ts
```diff
@@ -815,11 +815,38 @@ const definitions: Definition[] = [
...
+                    break;
+                }
+            } else {
+                coverExpose.withPosition().withTilt();
+            }
+            return [
+                coverExpose,
+                e.power().withAccess(ea.STATE_GET),
+                e.energy().withAccess(ea.STATE_GET),
+                e.linkquality()];
```
sjorge, line 848: For readability you might want to move the `]` to it's own line.

sjorge (approved):

src/devices/ubisys.ts
sjorge: Might need to push another lint fix as you also need a `,` after `e.linkquality()`

slugzero: yes, just noticed that, too. thanks :-)

Koenkk:
Cool, thanks!

PR #6811, opened 2024-01-02
Add support for Namron Touch Termostat 2.0

esandoe:
Implemented config for tuya thermostat Namron Zigbee Touch Termostat 16A (Issue: https://github.com/Koenkk/zigbee2mqtt/issues/20428).

Device page: https://www.namron.com/products/namron-zigbee-touch-termostat-16a-sort-760/

I have mapped all tuya datapoints I could find through z2m logs as I do not have a TuYa gateway. The thermostat has a few more features that I was unable to find datapoints for (such as screen brightness, floor sensor type (10K, 12K, etc.), but this config should still go a long way to support the basic functions of the device.

Had to add new keys for a few datapoints in `tuyaTz` as I couldn't find any similar keys that would fit. I dont know if this is how I'm supposed to do it, but it seems to work!

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -10,6 +10,7 @@ import * as ota from '../lib/ota';
 import * as utils from '../lib/utils';
 import extend from '../lib/extend';
 import {forcePowerSource, light, onOff} from '../lib/modernExtend';
+import * as tuya from '../lib/tuya';
 
 const ea = exposes.access;
 const e = exposes.presets;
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks!

nedrehagen:
Thanks Koenkk - Works almost as expected!

I use floor sensor and the current solution shows Temperature in cards and not the floor temp.
Is there a way to solve this in the code, to use temperature_floor as "main" temp if floor sensor is selected?

Screenshot_3

nedrehagen:
Never mind

Found a solution with added lines in configuration.yaml

  '0x3425b4fffecf70a8':
    friendly_name: Hall
    homeassistant:
      name: Hall
      current_temperature_template: '{{ value_json.local_temperature_floor }}'
    legacy: false
    optimistic: true
    thermostat_unit: celsius

PR #6812, opened 2024-01-02
fix: LGYCDD01LM color supported again

Nerivec:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/20497

Introduced in 3db5698a6b34d92bce5069862fcb8f735fd9a557 (Nov 27), subsequent updates never restored 'color'.

```diff
-        extend: extend.light_onoff_brightness_colortemp_color(
-            {disableEffect: true, disablePowerOnBehavior: true, disableColorTempStartup: true, colorTempRange: [153, 370]}),
+        extend: [
+            modernExtend.light({effect: false, powerOnBehaviour: false, colorTemp: {startup: false, range: [153, 370]}}),
```

### DIFF
--- a/src/devices/xiaomi.ts
+++ b/src/devices/xiaomi.ts
@@ -977,7 +977,7 @@ const definitions: Definition[] = [
         whiteLabel: [{vendor: 'Xiaomi', model: 'RLS-K01D'}],
         description: 'Aqara Zigbee 3.0 LED strip T1',
         extend: [
-            light({effect: false, powerOnBehaviour: false, colorTemp: {startup: false, range: [153, 370]}, color: {modes: ['xy']}}),
+            light({effect: false, powerOnBehaviour: false, colorTemp: {startup: false, range: [153, 370]}, color: true}),
             xiaomiPowerOnBehavior(),
             numeric({
                 name: 'length',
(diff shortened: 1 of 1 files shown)


src/devices/xiaomi.ts
```diff
@@ -977,7 +977,7 @@ const definitions: Definition[] = [
         whiteLabel: [{vendor: 'Xiaomi', model: 'RLS-K01D'}],
         description: 'Aqara Zigbee 3.0 LED strip T1',
         extend: [
-            light({effect: false, powerOnBehaviour: false, colorTemp: {startup: false, range: [153, 370]}}),
+            light({effect: false, powerOnBehaviour: false, colorTemp: {startup: false, range: [153, 370]}, color: {modes: ['xy']}}),
```
Koenkk, line 980:
```suggestion
            light({effect: false, powerOnBehaviour: false, colorTemp: {startup: false, range: [153, 370]}, color: true}),
```

Does this also work?

Nerivec: Should [do the same](https://github.com/Koenkk/zigbee-herdsman-converters/blob/79795781289bd5b2a576b4e18cdfc1abfdf58264/src/lib/modernExtend.ts#L230), I just instinctively used the [old value](https://github.com/Koenkk/zigbee-herdsman-converters/blob/79795781289bd5b2a576b4e18cdfc1abfdf58264/src/lib/exposes.ts#L688).

PS: I don't have that device, I just found the bug.

Koenkk:
Great, many thanks!

Issue #6585, opened 2023-11-29
Fixing Quotra QV-RGBCCT


mikeodr:
I've just moved to Zigbee2MQTT from ZHA and it is working fantastic, one exception is my Quotra QV-RGBCCT is not functioning correctly. Turning it on with a colour fails in a similar way I had to patch in [Home Assistant Core](https://github.com/home-assistant/core/commit/21eee1b371b5bc39e26ff2e02a036014b0cbb308), the state of the device indicates that it is on, but the LED strip physically does not light up at all.

I'm going to try and get a patch together on my own, but am currently struggling to understand how to disabled `move_to_level_with_on_off` as this device apparently doesn't support it in the ZCL.

Specific device source needing a patch: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/quotra.ts#L13

I'm going to continue to try and resolve this on my own and post a PR, but am still trying to understand how to flag that as unsupported somehow.

mikeodr:
I think the solution is to not call `extend.light_onoff_brightness_colortemp_color` here so as not to pass `tz.light_onoff_brightness` into `toZigbee`

mikeodr:
I've dug some more into exactly what home-assistant/core is doing [here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/light.py#L236-L311) and [here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/light.py#L352-L369).

Depending on the state of the light it forces it on before attempting to transition to the right brightness/colour.

I'm still trying to familiarize myself with the zigbee2mqtt/herdsman code, so I'm not sure if a similar workaround for specific hardware is feasible vs just saying the B07JHL6DRV doesn't support these features.

Still investigating in my spare time.

Koenkk:
Could you provide the debug log when it should turn on but it doesn't?

See [this](https://www.zigbee2mqtt.io/guide/usage/debug.html) on how to enable debug logging.

mikeodr:
I've not had time to do this and the workarounds have been fine.

Sorry for the delays.

PR #6813, opened 2024-01-02
QV-RGBCCT does not support power restore state.

mikeodr:
Disable power on restore state and color temp startup for Quotra QV-RGBCCT, they are not actually supported.

In my bid to try and resolve https://github.com/Koenkk/zigbee-herdsman-converters/issues/6585 I noticed these were enabled but they are not actually supported.

### DIFF
--- a/src/devices/quotra.ts
+++ b/src/devices/quotra.ts
@@ -14,7 +14,7 @@ const definitions: Definition[] = [
         model: 'B07JHL6DRV',
         vendor: 'Quotra',
         description: 'RGB WW LED strip',
-        extend: [light({colorTemp: {range: [150, 500]}, color: true})],
+        extend: [light({colorTemp: {range: [150, 500]}, color: true, powerOnBehaviour: false})],
     },
 ];
 


src/devices/quotra.ts
```diff
@@ -14,7 +22,9 @@ const definitions: Definition[] = [
         model: 'B07JHL6DRV',
         vendor: 'Quotra',
         description: 'RGB WW LED strip',
-        extend: [light({colorTemp: {range: [150, 500]}, color: true})],
```
Koenkk, line 17: Use this instead: `extend: [light({colorTemp: {range: [150, 500]}, color: true, powerOnBehaviour: false})],`

mikeodr:
Thanks for the review, updated.

I may have more to follow in, still digging into https://github.com/Koenkk/zigbee-herdsman-converters/issues/6585 any thoughts there would be appreciated if applicable, or if it's not a supported idea please feel free to close.

Koenkk:
Thanks!

PR #6727, opened 2023-12-17
Sinope RM3500ZB - Improve reporting configuration

SF6789:

PR #6814, opened 2024-01-02
Fix: Disable powerOnBehavior for RM3500ZB

alray31:
Disable powerOnBehavior (unsupported) for RM3500ZB introduced in #6727 when switching to extend.
image
image


Koenkk:
Thanks!

PR #6819, opened 2024-01-03
Fix problem with set `preset` and `system_mode` for thermostats

lompi:
Changes to fix errors from issue https://github.com/Koenkk/zigbee2mqtt/issues/20476.


Koenkk:
Thanks!

PR #6821, opened 2024-01-03
Support for TV06 Zigbee Thermostatic Radiator Valve (TRV)

lars010101:
First PR in my life, i hope im doing it right. Thought I share what works for me:

Bought two of these TRV's, and the datapoints from debug log appear to align precisely with this pre-existing setup so I just add my devices fingerprint.

I confirmed temp readings, set temps, presets, anti-freeze, child_lock, heating stop. Didnt confirm scheduling but assume it works as the debug messages show datapoint alignment.


Koenkk:
Thanks!

PR #6822, opened 2024-01-03
Handling action 1 and 17 for sinope switch.

onixx:
The lookup for actionReport was not handling actions 1 and 17 which were generating these errors in the debug log
Error: Expected one of: 2, 3, 4, 18, 19, 20, got: '17'
Error: Expected one of: 2, 3, 4, 18, 19, 20, got: '1'

Added the up_clickdown and down_clickdown actions which are indicating the button is "down" and not yet released. Once released, one of the existing 2, 3, 4, 18, 19, 20 is generated.


Koenkk:
Thanks!

PR #6825, opened 2024-01-03
Add basic support for BMCT-RZ

danielpetrovic:
On/off function works. As this is my first converter, can anyone help?

There are quite some options, many are probably overlapping with the Shutter/Light control II but the internal relays are different and this one has a single relay that is Potential Free.

It supports on/off and pulse mode, default mode seems to be on/off.

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -1285,6 +1285,8 @@ const definitions: Definition[] = [
         model: 'BMCT-RZ',
         vendor: 'Bosch',
         description: 'Relay, potential free',
+        fromZigbee: [],
+        toZigbee: [],
         exposes: [e.switch()],
     },
     {
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -1287,7 +1287,8 @@ const definitions: Definition[] = [
         description: 'Relay, potential free',
         fromZigbee: [fz.on_off, fz.power_on_behavior],
         toZigbee: [tz.on_off, tz.power_on_behavior],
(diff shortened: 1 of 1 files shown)


Koenkk:
Can you update this with the [generated external definition](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_2-creating-the-external-definition)?

danielpetrovic:
> Can you update this with the [generated external definition](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_2-creating-the-external-definition)?

```
const {onOff} = require('zigbee-herdsman-converters/lib/modernExtend');

const definition = {
    zigbeeModel: ['RBSH-MMR-ZB-EU'],
    model: 'RBSH-MMR-ZB-EU',
    vendor: 'BOSCH',
    extend: [onOff({powerOnBehavior: false})],
};

module.exports = definition;
```

Koenkk:
Can you try with the generated one and update this PR with it?

danielpetrovic:
Yes, it works the same. But when I change the PR to this code it gives an error that the OnOff is not recognised.

I’ve only changed vendor and description from the generated one, to be in line with the other Bosch devices.

Koenkk:
Add the following to the top of the file: `import {onOff} from '../lib/modernExtend';`

danielpetrovic:
> Add the following to the top of the file: `import {onOff} from '../lib/modernExtend';`

Done, it finished without errors now.

PR #6826, opened 2024-01-04
Draft: Add switch_type to Tuya TS0052

gaetancollaud:
The tuya TS0052 supports different switchType. 
Using zigbee2mqtt, I was able to set it using the dev console and the cluster `manuSpecificTuya_3`, but it should be exposed in the specific settings.

This is still a draft because I have to test it properly


gaetancollaud:
I never took the time to properly test it and I don't think I will have in the near future, I will close this and reopen later maybe.

PR #6828, opened 2024-01-04
MTG035-ZB-RL support

mikelauskas:
added MTG035-ZB-RL support (_TZE204_mtoaryre)


src/devices/tuya.ts
```diff
@@ -6841,6 +6841,71 @@ const definitions: Definition[] = [
             ],
         },
     },
+    {
```
Koenkk, line 6844: This looks like a copy-paste of `MTG075-ZB-RL`, can you add `_TZE204_mtoaryre` to that definition instead of creating a new one?

mikelauskas: Sure, I'll do that. Thanks for all the work with the repo!

Koenkk:
Thanks!

PR #6831, opened 2024-01-04
feat: support custom CA certs for OTA firmware downloads

kaechele:
Part of the fixes for https://github.com/Koenkk/zigbee-OTA/issues/420

### DIFF
--- a/src/lib/ota/common.ts
+++ b/src/lib/ota/common.ts
@@ -11,8 +11,6 @@ import {Zcl} from 'zigbee-herdsman';
 import https from 'https';
 import tls from 'tls';
 let dataDir: string = null;
-let insecureTLS = false;
-const caBundle: string[] = null;
 const maxTimeout = 2147483647; // +- 24 days
 const imageBlockResponseDelay = 250;
 const endRequestCodeLookup: KeyValueNumberString = {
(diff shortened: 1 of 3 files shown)


Koenkk:
Instead of setting the ca from [Zigbee2MQTT](https://github.com/Koenkk/zigbee2mqtt/pull/20571/files#diff-e5b1e24112c361ca744f57cf46fba74c4e6722ffec31be818b7cc8254cbd2a10R359), can we set it from [zigbeeOTA.ts/`getAxios()`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/9c7f6dd30a1e96b238584ac0e66339da9b5fafa3/src/lib/ota/zigbeeOTA.ts#L4)?

kaechele:
Sure.

I somehow thought it may be useful to allow the user to change it from the settings. But I would be willing to admit the use case may be rather obscure ;-)

I can prepare a change that adds an optional parameter to getAxios to take the URI of a CA bundle, which we can then set from the location you indicated.

kaechele:
Sorry, to clarify: I interpreted your question as removing the setting from Zigbee2MQTT and just hardcode the CA bundle path in `zigbeeOTA.ts` the same way it's currently done for the `index.json`. Is that what you meant?

Koenkk:
>  Is that what you meant?

I think that's enough for now indeed, I currently don't see the benefit of making this configurable.

Koenkk:
Many thanks! 🥳

PR #6574, opened 2023-11-26
Add new device: IKEA SOMRIG E2213

thatslolo:
I was able to snatch a few of the yet (to my knowledge) unreleased SOMRIG Shortcut buttons and got it successfully configured.
The button actions are transferred by `heimanSpecificScenes`. Not sure how the handle them for now.
Advice how to use them would be greatly appreciated.

PR #6834, opened 2024-01-04
Fix exposes property of E2213

krishoog:
The newly added device E2213 (see #6574) has the wrong values for the exposes property:

1. the actions are prefixed with `dots_`, which should not be the case
2. three actions are missing

This small change fixes that.


Koenkk:
@krishoog bedankt, leuk om je hier te zien! 😄

PR #6836, opened 2024-01-05
Add OTA to vallhorn

mik9:


mik9:
Added it via custom definition and looks like it works. No new firmware available at the moment though. But reported fw_image_type is present in json.

Знімок екрана 2024-01-05 о 10 37 11

Koenkk:
Thanks!

PR #6840, opened 2024-01-05
feat: want a way to add missing input and output clusters

sjorge:
From https://github.com/Koenkk/zigbee-herdsman-converters/issues/6721, this also allowed us to drop the endpointID override for the VOCKQJK11LM device.

We can and probably should inject these on other LUMI based devices too but I do not own an others.

It is important to add the `quirkAddEndpointCluster` extend BEFORE other extends that might depend on the missing information.

Binding of the added clusters now also seems to work, which makes my TRV's happy 😄 

Also added a `quirkSendWhenActive` which has dramatically improved the experience, gone are the occoasion timeouts during configure if one is to sleep to push the button to keep the device awake. 

Edit:
Also fixes the `Error: Expected one of: 1, 2, 3, 4, 5, got: '0'` issue. It was hard to track down because most of the time the device would already have a airQuality value by the time we finish the interview and hit the configure() functions. That would also explain why I got it mostly on my dev network as at the time there is only this device on there, so no other traffic which results in tighter timings during interviewing.


sjorge:
I tested with a already paired device, for some reason the modernExtend configure functions that get added removed do not trigger a reconfigure. But manually re-configuring added the missing clusters.

When pairing the device in question, everything works as expected as the missing clusters get added before the other modernExtend's need them.

sjorge:
I'm not sure why it is still complaining `device.save is not a function` it doesn't complain about the other uses in the file.

src/devices/xiaomi.ts
```diff
@@ -3118,11 +3119,16 @@ const definitions: Definition[] = [
...
+            quirkAddEndpointCluster({endpointID: 1, inputCluster: 'genAnalogInput'}),
+            quirkAddEndpointCluster({endpointID: 1, inputCluster: 'aqaraOpple'}),
             aqaraAirQuality(),
             aqaraVoc(),
-            temperature({endpointID: 1}),
-            humidity({endpointID: 1}),
+            temperature(),
+            humidity(),
             aqaraDisplayUnit(),
+            reconfigureReportingsOnDeviceAnnounce(),
```
sjorge, line 3131: I wasn't able to differentiate a rejoin where reporting was still working, or a rejoin where the device has been disconnected for more than 1h, were reporting was missing.

It shouldn't hurt to reconfigure them in both cases though.

sjorge: Going to remove this one, it seems to forget eveytime power_outage_count increases, but there is no announce/join to be found in the logs :(

I'll need to come up with a better way to do this one.

sjorge:
Looks like the test failure is because we're missing a mock for the save() function. Progess :)

sjorge:
@Koenkk I might need some help with the tests, I don't think just using fn.test() in mockDevice for save() is the solution as now, it correctly runs the quirkAddEndpointCluster() modernExtend, but when we actually setup the reporting during testing the clusters form quirkAddEndpointCluster() are ofcourse not present as save() did nothing.

src/lib/modernExtend.ts
```diff
@@ -526,6 +527,57 @@ export function forcePowerSource(args: {powerSource: 'Mains (single phase)' | 'B
     return {configure, isModernExtend: true};
 }
 
+export interface QuirkAddEndpointClusterArgs {
+    endpointID: number, inputCluster?: string | number, outputCluster?: string | number,
```
Koenkk, line 531: We should allow passing a list here

Koenkk:
@sjorge for the test case we can do something similar to [`saveClusterAttributeKeyValue`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/2b8688af2cb83516566e44119e918592a7f686fb/test/utils.ts#L47)

src/lib/modernExtend.ts
sjorge: I also considered that yes, but I opted to keep the code simpler and just add more to the extend: [] block on the device.

I'll change it to accept a list and rewrite the function tonight.

sjorge: Looks like I at least got better at typescript lately, the difference is not that big.

sjorge:
> @sjorge for the test case we can do something similar to [`saveClusterAttributeKeyValue`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/2b8688af2cb83516566e44119e918592a7f686fb/test/utils.ts#L47)

```
        save: jest.fn().mockImplementation(() => {
            console.log(JSON.stringify(this));
            //device.inputClusters = this.inputClusters;
            //device.outputClusters = this.outputClusters;
        }),
```

Seems I can't access the current device object to update the mockDevice's object, it complains `this` is shadowed.

Koenkk:
Could you commit your code? Then I can take a look.

sjorge:
> Could you commit your code? Then I can take a look.

Done

sjorge:
Don't think I would have found that solution 😬 

I'll retest this on my dev network and report back.

sjorge:
Looks good, ready to merge.

Koenkk:
Great, thanks!

> I wonder if we can somehow bundle some of the reads during interview to speed things up.

The problem is that if we chunk them into e.g. 3, if one attribute is not supported, all 3 will not be read.

sjorge:
Have we seen any devices that do not have all mandatory genBasic attributes implemented?

If not we could bundle those but leave a fastInterview = false exit hatch.

hmm actually the exit hatch won’t work as we need the info to be able to use the hatch 😑

Koenkk:
I don't know to be honest, but I think it's better to be a bit conservative here.

sjorge:
Also binding now works great, got my non testing one paired with my TRV in that room.

Just need to figure out how to restore reporting/bindings when the outage counter goes up :(

PR #6843, opened 2024-01-06
Improved OTA errors & logging

Nerivec:
A couple of fixes, more error checking, more/better logging, and "softer fail" when a device is yet to have images available (at least via Z2M). I went with a `logger.warn` for that last one; it's not quite an error, but at least it grabs the user's attention (and it will respond with `no update available` standard message frontend-wise, avoiding panic...).

Addresses concerns in https://github.com/Koenkk/zigbee2mqtt/issues/20582, https://github.com/Koenkk/zigbee2mqtt/issues/16618 and [likely more](https://github.com/search?q=org%3AKoenkk+%22No+image+available+for+imageType%22&type=issues)...

I assumed previous error-checking (URL responses & co) was valid, since I don't have all these devices to test :smile:

PS: Unless I missed something, it should have no impact on z2m or zh as far as typing.


src/lib/ota/common.ts
```diff
@@ -313,40 +327,53 @@ function callOnProgress(startTime: number, lastUpdate: number, imageBlockRequest
...
 export async function isNewImageAvailable(current: Ota.ImageInfo, logger: Logger, device: Zh.Device, getImageMeta: Ota.GetImageMeta) {
+    const currentS = JSON.stringify(current);
+    logger.debug(`OTA: Is new image available for '${device.ieeeAddr}', current '${currentS}'`);
     const meta = await getImageMeta(current, logger, device);
-    const [currentS, metaS] = [JSON.stringify(current), JSON.stringify(meta)];
-    logger.debug(`Is new image available for '${device.ieeeAddr}', current '${currentS}', latest meta '${metaS}'`);
+
+    // Soft-fail because no images in repo/URL for specified device
+    if (!meta) {
+        logger.warn(`OTA: Images currently unavailable for device '${device.modelID}', hardwareVersion '${device.hardwareVersion}', ${currentS}'`);
```
Koenkk, line 367: Please also log the `device.manufacturerName` here (it's relevant for TuYa devices)

Koenkk:
Looks good! Could you rebase this PR and take a look at the comment?

Koenkk:
Thanks!

Nerivec:
vscode pulled some crap on me when I rebased, I was half-expecting a fail, but looks like I tamed it.

About the concern I had mentioned, about frontend "seemingly" not responding because of log level (if level > info, no toast appears if no update available) when the check firmware button is pushed, are you passing it along, or should I open an issue there?

Koenkk:
@Nerivec that issue should be fixed in the frontend indeed. Z2M already responds [correctly](https://www.zigbee2mqtt.io/guide/usage/ota_updates.html#manually-check-if-an-update-is-available) to this.

LarsStegman:
I think this PR introduced the problem we are seeing in https://github.com/Koenkk/zigbee2mqtt/issues/20658

PR #6847, opened 2024-01-06
Remove % to power_factor and add SPM01-D2TZ-U01

dduransseau:
Following discussion [#20593](https://github.com/Koenkk/zigbee2mqtt/discussions/20593) remove the % to power_factor and specify it on tuya device.
Might consider divideBy100 on tuya device in futur to have consistency with standard zigbee.
Add SPM01-D2TZ-U01 device that support power_factor between 0 and 1


Koenkk:
Thanks!

PR #6855, opened 2024-01-07
fix: Add categories to several config/diagnostic exposes

slugzero:
I need a handful of exposes with the `category` flag set to implement a test for https://github.com/Koenkk/zigbee2mqtt/pull/20663.

After that PR is also merged, this will fix a few exposes which are currently displayed in the wrong category in Home Assistant after discovery.


sjorge:
Seems i cant leave feedback on a line from the mobile app.

For a xiaomi device you make the left operation_mode with a category, below it there is also one for the right endpoint, should they not both be marked?

slugzero:
> For a xiaomi device you make the left operation_mode with a category, below it there is also one for the right endpoint, should they not both be marked?

@sjorge oops, fixed. Thanks!

Koenkk:
Thanks!

PR #6856, opened 2024-01-08
Change default energy reporting change threshold to 0.01kWh

### LINKED ISSUE #6787: [Regression] TS011F_plug_1 reporting is broken across the board
Pretty sure this is due to https://github.com/Koenkk/zigbee-herdsman-converters/commit/5f511d34cc1f28d3a8690639b87fbaa47bb791f4. I'm still not familiar enough with the codebase (especially that new part) to craft an external converter that works with these plugs again. So I can only share my observations. 

The plugs work fine under even the newest version IF they were added to the network in a version prior to 16.12.0. Reconfiguring the plugs also works, but removing them and re-adding them (or buying new plugs) will cause them to report erroneous or even no values. That's why I didn't notice anything until today. This seems like a big regression to me.

Affected TS011F_plug_1 manufacturers are at least: _TZ3000_okaz9tjs, _TZ3000_gvn91tmx, _TZ3000_ss98ec5d. Note that I have had the latter two since November 16 and they have always worked perfectly. I received the _TZ3000_okaz9tjs on November 30 and they required https://github.com/Koenkk/zigbee2mqtt/issues/20028 in order to work properly.

Last known working version: 1.34.0-dev e0f5021, with zhc 16.11.0 and zh 0.26.1
In this one voltage, current, power, energy all report as they should for each of those plugs.

I have a nightly pull for the latest-dev docker image, so while I might not have all of the intermediate versions I did search through my cached images to get to the first known non-working version: 1.34.0-dev 65d7f5f, with zhc 16.12.0 and zh 0.27.1
On this one, voltage, current, power, energy all still report as they should for each of those plugs. But removing and re-adding them results in:
* In the reporting tab, haElectricalMeasurement rmsVoltage+rmsCurrent, seMetering currentSummDelivered and haElectricalMeasurement activePower (if applicable) disappeared, only OnOff onOff remains
* _TZ3000_okaz9tjs: Voltage ok, current *1000, power ok, energy N/A. currentSummDelivered messages still appear in the logs but are ignored.
* _TZ3000_gvn91tmx: Voltage ok, current 0, power 0, energy N/A. haElectricalMeasurement rmsCurrent and activePower do NOT appear in the logs, currentSummDelivered messages still appear but are ignored.
* _TZ3000_ss98ec5d: Voltage ok, current 0, power ok, energy N/A. haElectricalMeasurement rmsCurrent does NOT appear in the logs, currentSummDelivered messages still appear but are ignored.

Reverting to 16.11.0 restores the plugs to their former glorious state without having to remove/add them.

Also not working is the current latest-dev: 1.34.0-dev a3a87f4, with zhc 18.1.0 and zh 0.29.1
Same story here, voltage, current, power, energy all good (and still updating). But removing/adding:
* _TZ3000_okaz9tjs: Same as 1.34.0-dev 65d7f5f, with zhc 16.12.0 and zh 0.27.1
* _TZ3000_gvn91tmx: Same as _TZ3000_okaz9tjs (so current is *1000 instead of 0 and power is ok instead of 0)
* _TZ3000_ss98ec5d: Same as _TZ3000_okaz9tjs (so current is *1000 instead of 0)
So slightly different but overall kind of the same story.

I didn't really see any new warnings/errors in the logs (although I didn't have herdsman debugging on, just Z2M in loglevel debug) so I'm not sure attaching a big chunk of noise is useful here unless you're looking for something specific. I have reverted to the 16.11.0 version for now.

DataGhost:
Fixes #6787 (imo)

Decided to just make a PR in case you feel this is a sane default too. The current value of 0.1kWh is pretty large especially compared to other more easily fluctuating values being reported with changes as little as 5W or 0.05A. In contrast, it takes two full minutes at 3kW to use 0.1kWh and since most of these devices (I think) are going to be smart plugs that usually only switch "small" loads, it doesn't seem like something that's going to happen often. My dryer uses only half that power generally. Even at a 0.01kWh change that same load is only going to cause a message every 12 seconds so it's not going to flood the network.

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -158,8 +158,8 @@ export function electricityMeter(args?: ElectricityMeterArgs): ModernExtend {
         seMetering: {
             // Report change with every 5W change
             power: {attribute: 'instantaneousDemand', divisor: 'divisor', multiplier: 'multiplier', forced: args.power, change: 5},
-            // Report change with every 0.1kWh change
-            energy: {attribute: 'currentSummDelivered', divisor: 'divisor', multiplier: 'multiplier', forced: args.energy, change: 0.1},
+            // Report change with every 0.01kWh change
+            energy: {attribute: 'currentSummDelivered', divisor: 'divisor', multiplier: 'multiplier', forced: args.energy, change: 0.01},
         },
     };
 


Koenkk:
> Even at a 0.01kWh change that same load is only going to cause a message every 12 seconds so it's not going to flood the network.

The question is if this is useful, might also be personal preference, but maybe the default should be 50W? (0.50)

DataGhost:
Not entirely sure what you mean. For 50W to equal "0.50" the unit has to be hectowatts and I'm not sure that's used anywhere? If you meant 0.50 for the energy that's 500Wh, not 50Wh, and that's a massive amount. Most of the devices I'm monitoring struggle to use that much in a week. I'm just arguing that 0.01kWh (or 10Wh) is a large enough value in most cases to warrant reporting changes about, and it's basically been the default until now IIRC, whether intentional or not. So changing it back only restores the default behaviour to how it was before.

Either way, it's probably a good idea to properly expose it in the web interface under reporting as well, that would have pointed me to the issue much sooner. It now just shows an empty box instead of the threshold value.

Koenkk:
Sorry I meant 0.05W (so 0.05kW), what I tried to do with the modernExtend is to provide better defaults for the reporting. I think an energy report a few times a day should be good enough for most of the uses cases (of course users can change it)

DataGhost:
0.05kWh seems a better default than 0.10kWh anyway. Not my personal preference but it's adjustable so that's fine. I'll open another ticket for the UI side of things since my biggest concern turned out to be not knowing what value was used.

Koenkk:
Could you update this PR and fix the tests? Then this can be merged.

Issue #6859, opened 2024-01-08
Investigate why e089bf242b228209cfae800be44807aa7eafb33a broke thermostat schdules setting in some cases


sjorge:
At first glance it looks like a TypeScript issue where something should be `string | number` where we currently just have `string` but I can't figure out why at first glance.

Need to dig further into this.

```
error 2024-01-08 11:19:47Publish 'set' 'weekly_schedule' to 'trv/office' failed: 'Error: 'transitionTime hour' is not a number, got string (00)'
```

sjorge:
More specifically

```
                    // accept 24h time notation (e.g. 19:30)
                    if (typeof elem['transitionTime'] === 'string') {
                        const time = elem['transitionTime'].split(':');
                        utils.assertNumber(time[0], 'transitionTime hour');
                        utils.assertNumber(time[1], 'transitionTime minute');
                        if ((time.length != 2) || isNaN(time[0]) || isNaN(time[1])) {
                            meta.logger.warn(
                                `weekly_schedule: expected 24h time notation (e.g. 19:30) but got '${elem['transitionTime']}'!`,
                            );
                        } else {
                            elem['transitionTime'] = ((parseInt(time[0]) * 60) + parseInt(time[1]));
                        }
```

We call parseInt later on and the assert fails now.

sjorge:
```typescript
let x = "10";
console.log(isNaN(x));
console.log(x);

let y = "xx";
console.log(isNaN(y)));
console.log(y));
```

This works as expected, isNaN can be used to check if a string is a valid number (before parsing)

sjorge:
Since we know we're deeling with a tring because of the typeof check, I think we can drop the asserts or just use assertString().

PR #6860, opened 2024-01-08
fix: weekly_schedule was broken by e089bf242b228209cfae800be44807aa7eafb33a

sjorge:
Closes https://github.com/Koenkk/zigbee-herdsman-converters/issues/6859


sjorge:
Hmm it seems `isNaN` does not take a stirng according to typescript types, but it does and it works as expected.

sjorge:
Internet says to use Number.isNaN but that doesn't behave the same as just isNan

```typescript
let x = "10";
console.log( Number.isNaN(x) );
console.log( isNaN(x) );
console.log(x);

let y = "abc";
console.log( Number.isNaN(y) );
console.log( isNaN(y) );
console.log(y);
```

That produces:
false
true
"10"

false
false
"abc"

sjorge:
Seems we can just call parseInt on a string though without it failing.

sjorge:
Alright, rebased and now both `npm run lint` and `npm run build` are happy. And it also still works :D

Koenkk:
Thanks!

PR #6867, opened 2024-01-09
Add Sinope TH1320ZB-04 Sinope

NicDesrochers:
Add SInope TH1320ZB-04 Sinope from Hilo, clone code from TH1300ZB

### DIFF
--- a/src/devices/sinope.ts
+++ b/src/devices/sinope.ts
@@ -1000,6 +1000,91 @@ const definitions: Definition[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Updated the code to use whiteLabel to prevent code duplication, thanks!

NicDesrochers:
> Updated the code to use whiteLabel to prevent code duplication, thanks!

OK great whitelabel is a way better!
Thank's for the quick merge

PR #6868, opened 2024-01-09
ptvo.switch: fixed an error message when Z2M tries to read an OnOff

ptvoinfo:
ptvo.switch: fixed an error message when Z2M tries to read an OnOff state from a non-existent cluster.


Koenkk:
Thanks!

PR #6869, opened 2024-01-09
Feature: Ability to trigger events from converters on incoming Zigbee messages

slugzero:
When working on the [dynamic exposes](https://github.com/Koenkk/zigbee-herdsman-converters/blob/0e38f1cb1ae7af020ff90aa8cd0574363f00183b/src/devices/bosch.ts#L1335) for the BMCT-SLZ, I noticed that there is currently no way to programmatically trigger an event from inside a zigbee-herdsman-converter (or at least none that I could find). So in order to reevaluate the exposes, either zigbee2mqtt has to be restarted, or an event has to be triggered manually from outside the converters (e.g. by changing a device option). I therefore decided to make the `device_mode` an option, because changing this option leads to an immediate reevaluation of the exposes and immediate update of the exposes in the frontend, Home Assistant, etc.

However, I am not happy with that solution: The `device_mode` is an internal state which can be read from the device, and duplicating this as an option requires keeping them in sync somehow. I would rather directly use the device internal state to determine the exposes, but then, with the current implementation, the exposes would only be updated after a zigbee2mqtt restart. It is not a big thing, but it is still inconvenient and not really intuitive for the user.  

I looked at other devices, and I noticed that many of the other function exposes also use a device internal state to determine their exposes, e.g.:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/0e38f1cb1ae7af020ff90aa8cd0574363f00183b/src/devices/lixee.ts#L700
https://github.com/Koenkk/zigbee-herdsman-converters/blob/0e38f1cb1ae7af020ff90aa8cd0574363f00183b/src/devices/bitron.ts#L237
https://github.com/Koenkk/zigbee-herdsman-converters/blob/0e38f1cb1ae7af020ff90aa8cd0574363f00183b/src/devices/ubisys.ts#L796
https://github.com/Koenkk/zigbee-herdsman-converters/blob/0e38f1cb1ae7af020ff90aa8cd0574363f00183b/src/devices/profalux.ts#L51

That means that also these devices will only have a set of default exposes immediately after joining (when the device state is not read yet), and the correct exposes according to the device state are only available after a z2m restart or after some other event triggers a reevaluation of the exposes.

I therefore suggest to introduce a mechanism to allow converters to trigger events on their own, e.g. if a device internal state changes or a device sends a particular message, etc.. This will probably also be useful for other purposes than the dynamic exposes. 

This PR introduces an new `Trigger` for `fromZigbee` converters, which is similar to the `Converter`, but instead of converting the incoming Zigbee message to a MQTT payload, it triggers a set of predefined events based on the incoming Zigbee message.

The implementation does not interfere with or change any of the existing behavior, and it only gives converters access to some predefined events (currently only `devicesChanged`). For devices that don't have any Triggers in their definition, nothing changes.

The PR also includes an "example" implementation of a `Trigger` for the BMCT-SLZ I mentioned.

The corresponding change that would be required in zigbee2mqtt is here: 
https://github.com/slugzero/zigbee2mqtt/commit/4dcd10eb4b5db2a4bab08d4262043594b5e87907
I will prepare another PR for z2m if this PR is accepted.

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -250,6 +250,7 @@ const tzLocal = {
             if (key === 'device_mode') {
                 const index = utils.getFromLookup(value, stateDeviceMode);
                 await entity.write('manuSpecificBosch10', {deviceMode: index});
+                await entity.read('manuSpecificBosch10', ['deviceMode']);
                 return {state: {device_mode: value}};
             }
             if (key === 'switch_type') {
@@ -577,6 +578,20 @@ const fzLocal = {
             return result;
         },
     } satisfies Fz.Converter,
+    bmct_trigger: {
+        cluster: 'manuSpecificBosch10',
+        attribute: 'deviceMode',
+        event: 'devicesChanged',
+        getEvents: (data, device) => {
+            if (data.hasOwnProperty('deviceMode')) {
+                const deviceMode = data['deviceMode'];
+                if (deviceMode !== device.meta.deviceMode) {
+                    device.meta.deviceMode = deviceMode;
+                    return ['devicesChanged'];
+                }
+            }
+        },
+    } satisfies Fz.Trigger,
     bwa1_alarm_on_motion: {
         cluster: '64684',
         type: ['attributeReport', 'readResponse'],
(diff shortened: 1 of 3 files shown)


Koenkk:
What if instead of this pass an extra property to the `meta` argument to trigger events? Than this can also be easily applied at other places like `configure`/`toZigbee` and `onEvent`. It's similar to the already existing [`publish`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/f88f026fb9094e4881dcdf39b9265d04081aa54d/src/converters/fromZigbee.ts#L516)

slugzero:
You mean I pass the [`emit()` function](https://github.com/slugzero/zigbee2mqtt/blob/4dcd10eb4b5db2a4bab08d4262043594b5e87907/lib/extension/receive.ts#L167) from `receive.ts` directly to convert() (maybe rename it to `emitEvent()`) and then call it via e.g. `meta.emitEvent('devicesChanged')`? Yes, that would also work. If you prefer that, I will change it. 
But I don't understand how we would get it to toZigbee/configure/onEvent from there? Move `emitEvent()` to a separate helper file? I could of course pass the whole event bus in `meta`, but I guess that's a little bit too much freedom for the converters...

Koenkk:
> But I don't understand how we would get it to toZigbee/configure/onEvent from there? 

It should be passed from z2m to those functions.

But maybe for now we should just publish a `deviceExposesChanged` callback to the meta? We shouldn't leak z2m eventbus specific things to zigbee-herdsman-converters.

slugzero:
Yes, that makes sense as long as we have only one event anyway. I updated the PR, hope that is what you had in mind.

Koenkk:
Looks good, could you also make a PR for z2m?

slugzero:
z2m PR is now here: https://github.com/Koenkk/zigbee2mqtt/pull/20776
Took a bit longer because I needed to add some additional tests for coverage

Koenkk:
Thanks!

PR #6872, opened 2024-01-10
FIX: Add new detect names for IKEA Japan Tradfri LEDs

kenneyd:
I'm trying to convert from ZHA to zigbee2mqtt, and found several of my LED bulbs purchased from IKEA Japan are not listed in the `ikea.ts` file, so I've added them here. They are all existing devices, just with a new name.

New mappings are from:
- `LED2002G5` -> `'TRADFRI bulb E17 WS globe 440lm', 'TRADFRIbulbE17WSglobeopal470lm'`
- `LED1738G7` ->  `'TRADFRI bulb E17 WS opal 600lm'`
- `LED1935C3` ->  `'TRADFRIbulbE17WWclear250lm'`

Also, a weird note I found is that the device with a reporting name of `TRADFRIbulbE17WSglobeopal470lm` actually has 440lm silkscreened on the device 🤔 

image

image

image

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -633,7 +633,7 @@ const definitions: Definition[] = [
         extend: [tradfriLight({colorTemp: true})],
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #6865, opened 2024-01-08
fix: Replace `defaultSendRequestWhen` with `pendingRequestTimeout`

slugzero:
`defaultSendRequestWhen` is removed with https://github.com/Koenkk/zigbee-herdsman/pull/860.
This PR replaces it with `pendingRequestTimeout`

PR #6874, opened 2024-01-10
fix: move other devices from 6865 to modernExtend.quirkPendingRequestTimeout

sjorge:
Didn't need to add anything to the type, as it still takes number :) read over that yesterday.

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -26,7 +26,6 @@ const timeLookup = {
     '1_HOUR': 3600,
     '30_MINUTES': 1800,
     '10_SECONDS': 10,
-    'IMMEDIATE': 0,
 };
 
 type ReportingConfigTime = number | keyof typeof timeLookup;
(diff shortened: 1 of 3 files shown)


sjorge:
Wow I'm really not awake, as you change it in a later commit, and simply accepting both is gonna be a pain.

src/lib/modernExtend.ts
```diff
@@ -22,10 +22,11 @@ function getEndpointsWithInputCluster(device: Zh.Device, cluster: string | numbe
 }
 
 const timeLookup = {
-    '1_HOUR': 3600,
     'MAX': 65000,
+    '1_HOUR': 3600,
     '30_MINUTES': 1800,
     '10_SECONDS': 10,
+    'IMMEDIATE': 0,
```
Koenkk, line 29: I propose to not add it here but instead add it to the function and map it to `0` there (since this is not really a time):

```
export function quirkPendingRequestTimeout(timeout: keyof typeof timeLookup | 'immediate'): ModernExtend {
```

sjorge: Yeah that why i tried to get it to also just take a number, I’ll revisit again and if that fails add the mapping in the function.

sjorge: Came up with a solution on my evening walk, should have gone earlier but it's been cold lately 🥶

Koenkk:
Looks good, thanks!

sjorge:
There is still some follow up work for zh I discovered https://github.com/Koenkk/zigbee-herdsman-converters/pull/6865#issuecomment-1887372083 @slugzero is gonna look at it. I might have time later next week too if they don't.

PR #6875, opened 2024-01-10
Partially add Schneider Electric AC Fan module support

lucasteligioridis:
# What?

Add partial support for the Schneider Electric AC Fan module, as discussed in https://github.com/Koenkk/zigbee2mqtt/issues/18620

This patch includes the fan module by Schneider electric in a partial form. All of the practical use-cases of the fan are here and working:
- turning on/off (where on defaults to low speed)
- setting speeds to low, medium or high

What is not included is controlling the LED, this requires a little bit more effort and packet sniffing as the standard endpoint of the other Wiser switches do not work for this particular model and/or firmware.

Further detail is described in the issue linked, good idea to just get something in a working form added and when a packet sniffer is flashed and data inspected we can further enhance this with the LED changes.

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -42,6 +42,14 @@ const tzLocal = {
             await endpoint.read(0xFF17, [0x0000], {manufacturerCode: 0x105e});
         },
     } satisfies Tz.Converter,
+    fan_mode: {
+        ...tz.fan_mode,
+        convertSet: async (entity, key, value, meta) => {
+            utils.assertString(value);
+            if (value.toLowerCase() === 'on') value = 'low';
+            return tz.fan_mode.convertSet(entity, key, value, meta);
+        },
+    } satisfies Tz.Converter,
 };
 
 const fzLocal = {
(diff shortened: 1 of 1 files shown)


src/devices/schneider_electric.ts
```diff
@@ -475,6 +483,21 @@ const definitions: Definition[] = [
             await reporting.onOff(endpoint);
         },
     },
+    {
+        zigbeeModel: ['CHFAN/SWITCH/1'],
+        model: '41ECSFWMZ-VW',
+        vendor: 'Schneider Electric',
+        description: 'Wiser 40/300-Series Module AC Fan Controller',
+        fromZigbee: [fz.on_off, fz.fan],
+        toZigbee: [tzLocal.fan_mode, tz.on_off],
```
Koenkk, line 492: What does the `tz.on_off` do here? I think only `fan_mode` should be here (same for `fromZigbee`)

lucasteligioridis: You're right! I thought this was required from when I first added this integration but seems`fan_mode` has all the correct endpoint/exposed attributes.

I removed the `onOff` from the reporting as well as this seems to be part of the `fanMode` binding? I tested it on my fan and all the data appears correctly and fan works as expected.

lucasteligioridis: Also added `off` to the array just so you can get the option for it in the drop down box within the Zigbee UI, even though the `off` option is defaulted into the array in the backend.

Koenkk:
Thanks!

PR #6882, opened 2024-01-11
Lixee : explode STGE into sub-values

mildis:
No need to expose these values are they are computed and STGE is already exposed by itself.

### DIFF
--- a/src/devices/lixee.ts
+++ b/src/devices/lixee.ts
@@ -208,6 +208,139 @@ const fzLocal = {
                             result[at_snake + (i+1)] = (val & (1<<i)) >>> i;
                         }
                         break;
+                    case 'statusRegister':
+                        // contact sec
+                        result[at_snake + '_contact_sec'] = (val & 0x1) == 1 ? 'ouvert' : 'ferme';
+                        // organe de coupure
+                        switch ((val >>> 1) & 0x7) {
+                        case 0:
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

ALaDoffe:
Hello, 
Just test and "status_register_tempo_jour" reports "UNDEF" instead of "BLANC"
Not yet test "status_register_tempo_demain" as it publishes at 08:00PM

thx

mildis:
Hi @ALaDoffe, PR #6997 solves this.
status_register is sent as a String representing an hex value but I didn't managed to convert it to true hex before parsing.

ALaDoffe:
Thank you @mildis.

In fact, I already done a custom sensor to decode status register in HA for "status_register_tempo_demain".
But I get an error on each HA restart that is why I am interrested in your component.
For information, during dev of this sensor I realize that values are not equal as described in documentation.
Indeed, Status_register for "couleur lendemain contrat tempo"  (bit 26-27) never take value 1 for "BLEU" it always keep value 0.
Here below what i do and it works for a while:

      {%- if today_at("06:00") < now() < today_at("20:00") -%}
            Not Yet Defined
      {%- else -%}
        {%- if TC == 2 -%}
            Jour BLANC
        {%- elif TC == 3 -%}
            Jour ROUGE
        {%- else -%}
            Jour BLEU
        {%- endif -%}
      {%- endif -%}

PR #6885, opened 2024-01-12
New Device Support for yale. Add NF-YRD622 TSDB

arunbala81:


src/devices/yale.ts
```diff
@@ -195,6 +195,13 @@ const tzLocal = {
 };
 
 const definitions: Definition[] = [
+    {
+        zigbeeModel: ['NF-YRD622 TSDB'],
+        model: 'NF-YRD622 TSDB',
```
Koenkk, line 200: When googling for this, I don't find any results, please use the model number which is listed on device/product box

src/devices/yale.ts
```diff
@@ -195,6 +195,13 @@ const tzLocal = {
...
+        vendor: 'Yale',
+        description: 'Automatically generated definition',
```
Koenkk, line 202: Please use a real description here.

src/devices/yale.ts
```diff
@@ -195,6 +195,13 @@ const tzLocal = {
...
+        extend: lockExtend(),
```
Koenkk, line 203: Does the automatically generated definition also provide a `pinCodeCount`?

PR #6892, opened 2024-01-12
added battery reporting for PARASOLL

thatslolo:
changed battery reporting from voltage to percent for SOMRIG

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -1271,7 +1271,7 @@ const definitions: Definition[] = [
         vendor: 'IKEA',
         description: 'PARASOLL door/window Sensor',
         fromZigbee: [fz.battery, fz.ias_contact_alarm_1],
-        toZigbee: [tz.battery_percentage_remaining],
+        toZigbee: [],
         ota: ota.tradfri,
         exposes: [e.battery(), e.contact()],
         configure: async (device, coordinatorEndpoint, logger) => {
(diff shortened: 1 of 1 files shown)


thatslolo:
should be ready now

thatslolo:
> should be ready now

now ;)

Koenkk:
Thanks!

mrskycriper:
@thatslolo Hello, is there any reason for `genBasic` and `ssIasZone` binds? It should work without them. I can't seem to find any issues that necessitated them

thatslolo:
Afaik I needed the Gennadi for battery reporting. For the other I don’t remember. Would there be a benefit in removing them?

mrskycriper:
> Afaik I needed the genBasic for battery reporting. For the other I don’t remember. Would there be a benefit in removing them?

Battery reporting is done using `genPowerCfg` cluster. So it should be bound instead. But since it works now, this bind would be unnecessary too.

I just would like to remove unnecessary binds since I'm doing a big IKEA converters cleanup. Can you please test it without those 2 binds?

thatslolo:
I can test it tomorrow.

mrskycriper:
> I can test it tomorrow.

Thanks!

thatslolo:
@mrskycriper tested it with joining a new PARASOLL sensor. Everything works as expected. Actually, I don't have a clue why I added these bindings.

mrskycriper:
> @mrskycriper tested it with joining a new PARASOLL sensor. Everything works as expected. Actually, I don't have a clue why I added these bindings.

Thanks, I'll just remove those since they aren't really needed, which is to be expected. Since Ikea more or less follows the Zigbee specifications

thatslolo:
@mrskycriper since I used the modernExtend I think all the config for the battery reporting is also redundant…

mrskycriper:
> @mrskycriper since I used the modernExtend I think all the config for the battery reporting is also redundant…

I've moved everything to modern extends. You can check it out in this PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/7220

PR #6893, opened 2024-01-12
Added support for Hiladuo blinds sold on Amazon

juchong:
Added support for Hiladuo blinds sold on Amazon ([link](https://a.co/d/edjIpxs)).

These blinds use a white label Tuya motor and seem to work well with the existing cover_3 configuration.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2802,7 +2802,7 @@ const definitions: Definition[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_68nvbio9', '_TZE200_pw7mji0l', '_TZE200_cf1sl3tj', '_TZE200_nw1r9hp6']),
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -2822,6 +2822,7 @@ const definitions: Definition[] = [
             tuya.whitelabel('Zemismart', 'ZM16EL-03/33', 'Cover motor', ['_TZE200_68nvbio9']),
             tuya.whitelabel('Zemismart', 'ZM25EL', 'Cover motor', ['_TZE200_pw7mji0l']),
             tuya.whitelabel('Zemismart', 'ZM85EL-2Z', 'Roman Rod I type U curtains track', ['_TZE200_cf1sl3tj', '_TZE200_nw1r9hp6']),
+            tuya.whitelabel('Hiladuo', '', 'Motorized Roller Shade', ['_TZE200_9p5xmj5r']),
```
Koenkk, line 2825: Please add a model number here instead of `''` (if it doesn't have any use the Amazon EAN code)

juchong: > Please add a model number here instead of `''` (if it doesn't have any use the Amazon EAN code)

I'm in the US and the blinds were assembled in the US, so an EAN isn't available  (I couldn't find it in the listing). I added the ASIN instead.

Koenkk:
Thanks!

PR #6906, opened 2024-01-14
fix: remove deprecated `sendWhen` from converters

slugzero:
This PR removes the deprecated `sendWhen` from zigbee-herdsman-converters. This is a mere refactor, devices will all behave the same way as they do since at least the beginning of December.

**Note:**
The following converters/devices made use of `sendWhen=active` and I could not find any information whether they have the genPollCtrl cluster or not. If these devices have the genPollCtrl cluster, everything is fine. If not, it might be worth adding a  `quirkCheckinInterval('1_HOUR')` to improve performance. I cannot tell because I don't own any of these devices, and adding this to devices without knowing might do more harm than good. 

A quick search through open issues did not show up anything and potential issues with these devices should be there since at least beginning of December. So I am just leaving this here in case someone searches for these devices in the future. 

* `LS21001_alert_behaviour`: Linkind A001082 (LS21001)        
* ~~`ZM35HQ_attr`: TuYa TS0202 (ZM-35H-Q, IH012-RT01, IH012-RT02)~~
* `wiser_vact_calibrate_valve`, `wiser_sed_thermostat_keypad_lockout`, `wiser_sed_thermostat_local_temperature_calibration`: Schneider Electric EH-ZB-VACT (EER53000)
* `ctm_group_config`: CTM Lyng mTouch Bryter, mTouch_Astro, mSwitch_Mic (Mikrofon)
* `plugwise_calibrate_valve`: Plugwise 106-03

The Schneider Electric CCTFR6400 also made use of `sendWhen`, but according to https://github.com/Koenkk/zigbee2mqtt/discussions/16951#discussion-4938246, it does have the genPollCtrl Cluster, so it should work fine.

### DIFF
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -2211,7 +2211,7 @@ const converters2 = {
         convertSet: async (entity, key, value, meta) => {
             const lookup = {'siren_led': 3, 'siren': 2, 'led': 1, 'nothing': 0};
             await entity.write('genBasic', {0x400a: {value: utils.getFromLookup(value, lookup), type: 32}},
-                {manufacturerCode: 0x1168, disableDefaultResponse: true, sendWhen: 'active'});
+                {manufacturerCode: 0x1168, disableDefaultResponse: true});
             return {state: {alert_behaviour: value}};
         },
     } satisfies Tz.Converter,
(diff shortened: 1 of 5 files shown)


sjorge:
@Koenkk do you have an TuYa devices? I gave the 2 I had away. But I remember them not having genPollCtrl clusters. For those it might make sense to add the quirk

Koenkk:
I have multiple but only 1 TuYa end device ([this one](https://www.zigbee2mqtt.io/devices/ZTH08-E.html#tuya-zth08-e)), it does not have `genPollCtrl`. @slugzero does that help?

slugzero:
@Koenkk thanks for looking it up. I'm almost certain that the device in question also won't have `genPollCtrl`. I'll have a look tomorrow in my pile of scrapped devices, maybe I'm lucky. I actually think I had one of these, but it was broken.

slugzero:
Ok I actually found one and could bring it back to life. It does not have genPollCtrl. If not triggered, it sends a status message roughly every 4 hours. So I added a 4h 10 min timeout.

Koenkk:
Thanks!

PR #6907, opened 2024-01-14
Add optional `state_action` for contact sensors

slugzero:
This adds the `state_action` option to contact sensors (disabled by default). If enabled, the sensor publishes an additional 'open'/'close' action every time the contact sensor changes its state.

This is something I am using for a while now in a custom converter, which I think could also be useful for others: When using the contact sensor as trigger to increment counter (e.g. I turned my gas and water utility meters smart with MCCGQ1xLMs), both counting state transitions in the MQTT payload and counting MQTT state messages themselves is inaccurate by 1-5%. The former because every lost device message immediately leads to a lost increment, the latter because regular state reportings would also lead to an unwanted increment. 

Triggering on the edges using `action` allows more elaborate counting, e.g. count all rising edges plus all falling edges when the state was already "closed". With a bit of additional debouncing, I am now getting near-accurate counts.

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -980,17 +980,22 @@ const converters1 = {
     ias_contact_alarm_1: {
         cluster: 'ssIasZone',
         type: 'commandStatusChangeNotification',
+        options: [exposes.options.state_action()],
         convert: (model, msg, publish, options, meta) => {
             const zoneStatus = msg.data.zonestatus;
             const contactProperty = postfixWithEndpointName('contact', msg, model, meta);
             const tamperProperty = postfixWithEndpointName('tamper', msg, model, meta);
             const batteryLowProperty = postfixWithEndpointName('battery_low', msg, model, meta);
 
-            return {
+            const payload: KeyValueAny = {
                 [contactProperty]: !((zoneStatus & 1) > 0),
                 [tamperProperty]: (zoneStatus & 1<<2) > 0,
                 [batteryLowProperty]: (zoneStatus & 1<<3) > 0,
             };
+            if (options && options.state_action) {
+                payload['action'] = postfixWithEndpointName((zoneStatus & 1) > 0 ? 'open' : 'close', msg, model, meta);
+            }
+            return payload;
         },
     } satisfies Fz.Converter,
     ias_contact_alarm_1_report: {
@@ -3903,10 +3908,16 @@ const converters1 = {
         },
     } satisfies Fz.Converter,
     xiaomi_contact: {
+        options: [exposes.options.state_action()],
         cluster: 'genOnOff',
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
-            return {contact: msg.data['onOff'] === 0};
+            const contact = msg.data['onOff'] === 0;
+            const payload: KeyValueAny = {contact: contact};
+            if (options && options.state_action) {
+                payload['action'] = contact ? 'close' : 'open';
+            }
+            return payload;
         },
     } satisfies Fz.Converter,
     W2_module_carbon_monoxide: {
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -3908,7 +3908,6 @@ const converters1 = {
         },
     } satisfies Fz.Converter,
     xiaomi_contact: {
-        options: [exposes.options.state_action()],
         cluster: 'genOnOff',
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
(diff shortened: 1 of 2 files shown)


Koenkk:
Instead of doing this in the converter, I propose to move it to [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/cfc2937f612b687080b4b2557a167bbb02a78c46/src/index.ts#L227), otherwise all converters need to be updated to support this.

slugzero:
Thanks, I wasn't aware of that possibility yet.

I just had a quick look at the devices under question and I have concerns that with your proposal, some of the devices would not show the expected behavior (i.e. send out an action only on an actual change of the contact state): 

There are two converters in `fromZigbee.ts`: `ias_contact_alarm_1` for `commandStatusChangeNotification` and `ias_contact_alarm_1_report` for `attributeReport`. Both return a `contact` payload. Most devices have only one of these, but some, e.g. [this](https://github.com/Koenkk/zigbee-herdsman-converters/blob/6ef352835316d3d1cfc8d6b1e264a40263dd9823/src/devices/smartthings.ts#L80) or [this]( https://github.com/Koenkk/zigbee-herdsman-converters/blob/6ef352835316d3d1cfc8d6b1e264a40263dd9823/src/devices/lidl.ts#L301) have both. 
I don't own any of these, but I believe that they probably use `attributeReport` for cyclic reporting, while `commandStatusChangeNotification` is the actual state change. So for these devices, the action should only be sent out for `commandStatusChangeNotification` messages. I didn't see a way in the `postProcessConvertedFromZigbeeMessage` function how to know what type of ZigBee message we got, but maybe I missed it?

So, I think there are the following two options: 
* Putting it to `postProcessConvertedFromZigbeeMessage`: All devices would have the option to send out a state action. But for some of these devices (e.g. those with both converters, and possibly also some that send `attributeReports` only), we would get false action messages on status reports.
* Putting it only to `ias_contact_alarm_1`: This will lead to a few devices without that option. I.e. the ones who use an `attributeReport` for the actual state change notification and the ones with a device specific converter. I counted around 10 other converters there, some of which could also be refactored back into `ias_contact_alarm_1`. But at least the devices that will have the option will also have proper actions (and a proper count for my use case).

Therefore I would rather leave it with the second option and then add it one-by-one to the converters as needed (as I already did for `xiaomi_contact`). But I you prefer the first one, just let me know and I can move it.

Koenkk:
>  and possibly also some that send attributeReports only),

For now you assume that `xiaomi_contact` is only sent on a state change, but I don't know if this is true for all devices using it.

I think we can trigger on the state change (e.g. [`meta.state`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/ad2d1114d4f7ae9bb516c6dc48e5959ff49cbec3/src/converters/toZigbee.ts#L829) but this is not passed yet. What do you think about it?

Thinking about that now, isn't that what you could also do instead of adding this? (check if contact changes from true -> false or false -> true?)

slugzero:
> For now you assume that `xiaomi_contact` is only sent on a state change, but I don't know if this is true for all devices using it.

Fair point. I have only tested the MCCGQ11LM (and 14, which is using `ias_contact_alarm_1`). I would assume that the 01, 12 and 13 behavior is not fundamentally different, but to be on the safe side, I could move the option from the converter to the MCCGQ11LM definition instead.
 
I believe triggering on the state change would not add any additional information compared to the state change in the `contact` payload itself. And as you point out correctly, then I could as well trigger on the payload itself.
That was also what I tried in the first place for my "counter" use case, but then I added the `action` for better precision: 
I have a ~1% chance of a single message from the device being lost, so e.g. a "closed" message after a device state change gets lost and the state in z2m remains "open". I can recover this lost message by also counting "opened" messages when the z2m state is already "open". This cuts my error rate to the probability of 2 consecutive messages being lost, i.e. from 1% to roughly 0.01%. When only looking at the state change in z2m, the "opened" message would also be lost because the state would change from "open" to "open".

slugzero:
Ok, I thought I found a way to move over the functionality to `postProcessConvertedFromZigbeeMessage`, but it does not work properly with the MCCGQ11LM: The device sends changes of the `contact` state in the `genOnOff` cluster, but it also sends a `contact` payload in the `genBasic` status messages (once every few hours or when pressing the button on the device). So an action should only be published for `contact` payload from the `genOnOff` cluster, but I did not figure out a way to do this in `postProcessConvertedFromZigbeeMessage`. @Koenkk  do you have an idea?

Koenkk:
> but it also sends a contact payload in the genBasic status messages (once every few hours or when pressing the button on the device).

Why is this a problem, assuming you trigger on the state change only, these messages should always be ignored?

slugzero:
> Why is this a problem, assuming you trigger on the state change only, these messages should always be ignored?

Triggering on the state change only would not add any extra information to the state in the payload itself and therefore would not justify any change at all imo. Therefore I want to trigger directly on the device messages (I tried to explain why in more detail in my [post from yesterday](https://github.com/Koenkk/zigbee-herdsman-converters/pull/6907#issuecomment-1894591796)), but only on the "right" ones. That was easy to do in the converters for both the MCCGQ11LM and MCCGQ14LM, but I am struggling to find a generic way to do this in the postprocessing.

Koenkk:
Maybe we should keep this as a external converter for now? I think it would also be good to submit a feature request in the Z2M issue tracker to see how many people are interested in this.

slugzero:
Fine for me. So let's wait if someone else requests that feature, too.

PR #6921, opened 2024-01-18
Update philips.ts for additional model Being Light

thk-socal:
Add support for '929003053901' of the Being Alum white ambiance light and change Mired range to 153-454 per spec and light capability.


Koenkk:
Thanks!

Issue #6887, opened 2024-01-12
 Bugs in converter for Aqara SRTS-A01


mrskycriper:
It seems that converter for Aqara SRTS-A01 uses `switch` type for `valve_detection`, `window_detection` and `schedule` instead of `binary` type for exposing the entities.

- [`schedule`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/346ef5bad0018d8eaa50a5035e85bec9c4a2a79d/src/devices/xiaomi.ts#L3404) as [`switch_`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/346ef5bad0018d8eaa50a5035e85bec9c4a2a79d/src/lib/exposes.ts#L647C9-L647C9) 
- [`valve_detection`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/346ef5bad0018d8eaa50a5035e85bec9c4a2a79d/src/devices/xiaomi.ts#L3393C24-L3393C24) as [`valve_detection`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/346ef5bad0018d8eaa50a5035e85bec9c4a2a79d/src/lib/exposes.ts#L761C14-L761C14)
- [`window_detection`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/346ef5bad0018d8eaa50a5035e85bec9c4a2a79d/src/devices/xiaomi.ts#L3391) as [`window_detection`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/346ef5bad0018d8eaa50a5035e85bec9c4a2a79d/src/lib/exposes.ts#L779C12-L779C12)

The [guide](https://www.zigbee2mqtt.io/guide/usage/exposes.html) states that `switch` is used for a primary switch entity and `binary` is used for settings. 

For the end user this results in missing entity names in UI.

![pic](https://github.com/Koenkk/zigbee-herdsman-converters/assets/32013287/00c75419-3521-48f6-8510-8380882395b8)

Also it seams that `schedule` is missing it's name, which results in it being named after the device itself. This was very confusing and took some time to deduce it's actual name.

PR #6922, opened 2024-01-18
Improving Aqara SRTS-A01 converter and general export tidying

mrskycriper:
This PR attempts to fix issues, found in #6887. All secondary exports received `device_category`.

However, I can't seem to figure out, how to set `device_category` for `child_lock` and `lock` in general. 
I've tried setting their `access` to `ALL` on defenition, using `withAccess` and using `setAccess`. But generateDefenition test always fails like this:
```
AssertionError [ERR_ASSERTION]: Cannot add access if not defined yet

      39 |
      40 |     withAccess(a: number) {
    > 41 |         assert(this.access !== undefined, 'Cannot add access if not defined yet');
         |               ^
      42 |         this.access = a;
      43 |         this.validateCategory();
      44 |         return this;
```

Also I need help with testing in general, is there a way to use this modified converter on an active installation as HA OS Add-on without buying additional hardware? External converters seem to be limited and can't handle the usage of additional npm packages in the source code.


Koenkk:
For child_lock, I propose to use the `binary` instead of the `lock` expose for this (this should fix your issue)

> External converters seem to be limited and can't handle the usage of additional npm packages in the source code.

They can access all packages that zhc has access to, can you provide the details?

mrskycriper:
> For child_lock, I propose to use the binary instead of the lock expose for this (this should fix your issue)

@Koenkk Thanks for the suggestion about lock.

> They can access all packages that zhc has access to, can you provide the details?
For the buffer you can use import {Buffer} from 'buffer';

Buffer was stubborn, but I managed to import it using `const {Buffer} = require(‘node:buffer’)`. 

Did some testing and it looks to be working. Still need some work done though. I kinda want to replace stateless enum with something, that would show as a button entity (that can be pressed from the interface) in Home Assistant. How is it defined in the converter context?

mrskycriper:
Looks good to go, except for the proposed stateless enum (`calibrate` – used to send device into calibration procedure) to something else, so it could work as a button entity.

mrskycriper:
Added device categories from [here](https://github.com/Koenkk/zigbee2mqtt/pull/20746) and did a lot of general export tidying.

src/lib/utils.ts
```diff
@@ -584,6 +584,31 @@ export function getFromLookup<V>(value: unknown, lookup: {[s: number | string]:
     return result ?? defaultValue;
 }
 
+// Silly hack, but boolean is not supported as index
+export function getFromLookupBool(value: unknown, lookup: {[s: string | number]: string | number}): boolean | number {
```
Koenkk, line 588: Instead of adding a new function, could we maybe extend `getFromLookup` like this: 

```
export function getFromLookup<V>(value: unknown, lookup: {[s: number | string]: V}, defaultValue: V=undefined, keyIsBool: boolean=false): V {
```

mrskycriper: Merged them together

Koenkk:
Thanks!

protyposis:
Thanks for these nice improvements. Unfortunately, the renaming of `sensor_temp` to `external_temperature_input` basically breaks all documented ways on how to connect the TRV with external temperature sensors, e.g. in https://github.com/Koenkk/zigbee2mqtt/discussions/14291 (since they all rely on the `sensor_temp` name).

I agree the old name was ambiguous and renaming is a good idea. Also, this isn't the first name change that happened in Z2M, so what I'm asking generally applies to all changes of this type, and I'm using this PR only as an example and placeholder for any similar case. Would it make sense to find a way of handling such breaking changes more gracefully and user-friendly? What's your opinion on this, @Koenkk?

mrskycriper:
Oh, I didn’t realize that this introduced breaking changes. I even made some duplicate exports to leave other trvs alone because they used some of those broken exposes. But somehow forgot that name change will definitely brake some things. 

I think there should be a breaking changes section in the release notes. And they probably should be minimized. But in this case I intended to share it as a standard temperature input expose for other devices too.

protyposis:
> I think there should be a breaking changes section in the release notes.

Yes, this would be a great start. Maybe also contributor guidelines with the best practices that you mentioned (focus on minimization and standardization).

In the longer term, there could also be a kind of deprecation period, e.g., supporting old and new names in parallel for a few months, displaying deprecation notices in the UI, or detecting the use of old/deprecated names and displaying update prompts in the UI (similar to [HA Repairs](https://www.home-assistant.io/integrations/repairs/)).

Anyway, I didn't mean to hijack this PR. In case this fits the roadmap of Z2M, we may continue in a dedicated discussion thread.

mrskycriper:
> locale temperature no longer shows the external sensor temperature when sensor is set to EXTERNAL. I rolled back to previous version and there all is working fine.

I've had a lot of problems with cashing after updating this converter. Had to restart, re-pair and clear mqtt topics several times before the new converter was actually applied. Maybe you have the same problem? No code, responsible for temperature data was changed, only the name of temperature input from `sensor_temp` to `external_temperature_input`.

mescalum:
> > locale temperature no longer shows the external sensor temperature when sensor is set to EXTERNAL. I rolled back to previous version and there all is working fine.
> 
> I've had a lot of problems with cashing after updating this converter. Had to restart, re-pair and clear mqtt topics several times before the new converter was actually applied. Maybe you have the same problem? No code, responsible for temperature data was changed, only the name of temperature input from `sensor_temp` to `external_temperature_input`.

Thanks, I'll try that. But for me it is clearly now showing the local sensor temperature on the SRTS instead of the external sensor temperature. Doesn't sound to be caching to me but more of ignoring passing on the external sensor temparuture to the local sensor reading. I use Better Thermostat on top of this and have more communication and adjusting going on since this update which reduces batterie lifetime even more than it already is. My external sensors are more stable in regards to reported temperature while the SRTS build in sensor changes a lot obviously.

mrskycriper:
> > > locale temperature no longer shows the external sensor temperature when sensor is set to EXTERNAL. I rolled back to previous version and there all is working fine.
> > 
> > 
> > I've had a lot of problems with cashing after updating this converter. Had to restart, re-pair and clear mqtt topics several times before the new converter was actually applied. Maybe you have the same problem? No code, responsible for temperature data was changed, only the name of temperature input from `sensor_temp` to `external_temperature_input`.
> 
> Thanks, I'll try that. But for me it is clearly now showing the local sensor temperature of the SRTS instead of the external sensor temperature. Doesn't sound to be caching to me but more of ignoring passing on the external sensor temparuture to the local sensor reading. I use Better Thermostat on top of this and have more communication and adjusting going on since this update which reduces batterie lifetime even more than it already is. My external sensors are more stable in regards to reported temperature while the SRTS build in sensor changes a lot obviously.

If you look at entity id in HA and name of the temperature input in z2m dashboard, are they updated to the new name?

mescalum:
Thanks external_temperature_input was key. Am in the process of changing all my automations to the new sensor name. I Wish this could have been more obvious as a breaking change. Best would have been to hae kept the old sensor name for a while to let users transition.

PR #6930, opened 2024-01-18
feat: better generator support for multi-endpoint devices

ffenix113:
* Fix pressure extender defaults to match ZCL spec
* Better defaults for reporting on temp, pressure and humidity extenders
* Better definition generator support for devices with multiple endpoints

Previously if multiple endpoints had second endpoint generated definition would still try to fetch attribute value from the first endpoint. Fetching attribute value from developer console worked properly when using correct endpoint.


test/generateDefinition.test.ts
```diff
@@ -95,7 +95,7 @@ describe('GenerateDefinition', () => {
             ]},
             configureReporting: {
                 1: [
-                    ['msTemperatureMeasurement', [reportingItem('measuredValue', 10, repInterval.HOUR, 100)]],
+                    ['msTemperatureMeasurement', [reportingItem('measuredValue', 10, repInterval.HOUR, 1)]],
```
Koenkk, line 98: This will make the sensor report changes of `0.01`C, I think thats too much.

src/lib/modernExtend.ts
```diff
@@ -67,13 +67,15 @@ async function setupAttributes(
 
 export function setupConfigureForReporting(
     cluster: string | number, attribute: ReportingConfigAttribute, config: ReportingConfigWithoutAttribute, access: Access,
+    endpointID?: number,
```
Koenkk, line 70: Why is this needed? `setupAttributes` automatically check the clusters.

test/generateDefinition.test.ts
ffenix113: You are right, reverted this change.

src/lib/modernExtend.ts
ffenix113: It is true, but if device has, say, 10 endpoints with temperature cluster - then `setupAttributes` function currently will iterate all endpoints and for each endpoint with temperature cluster it will try to bind it, setup reporting and possibly read the value of the attribute. As temperature cluster is present in 10 endpoints - `setupAttributes` will be called 10 times. This results in 10^2 * 3 = 300 operations.

If instead endpoint will be provided - `setupAttributes` will only do necessary operations for each endpoint and cluster exactly once, reducing operations to 10 * 3 = 30.

Does this sound reasonable?

src/lib/generateDefinition.ts
```diff
@@ -4,30 +4,53 @@ import {getClusterAttributeValue} from './utils';
...
 import {philipsLight} from './philips';
+import {Device} from 'zigbee-herdsman/dist/controller/model';
 
-interface GeneratedExtend {extend: ModernExtend, source: string, lib?: string}
+interface GeneratedExtend {extend?: ModernExtend, extendFn?: (a: object) => ModernExtend, args?: object, source: string, lib?: string}
 type ExtendGenerator = (endpoint: Zh.Endpoint) => Promise<GeneratedExtend[]>;
 type Extender = [string[], ExtendGenerator];
 
-function generateSource(device: Zh.Device, generatedExtend: GeneratedExtend[]): string {
+function generateSource(zigbeeModel: string, definition: Definition, generatedExtend: GeneratedExtend[]): string {
```
ffenix113, line 13: `zigbeeModel` here would not be necessary, but TS fails to understand that `definition` contains `zigbeeModel` field, thus failing the build. 

If there is a better way to fix it - would be glad to do it.

src/lib/generateDefinition.ts
```diff
@@ -4,30 +4,53 @@ import {getClusterAttributeValue} from './utils';
 import * as m from './modernExtend';
 import * as zh from 'zigbee-herdsman/dist';
 import {philipsLight} from './philips';
+import {Device} from 'zigbee-herdsman/dist/controller/model';
 
-interface GeneratedExtend {extend: ModernExtend, source: string, lib?: string}
+interface GeneratedExtend {extend?: ModernExtend, extendFn?: (a: object) => ModernExtend, args?: object, source: string, lib?: string}
```
ffenix113, line 9: I would prefer it to have fields like `extendFn?: (a: T) => ModernExtend, args?: T`, but I am not sure if it is possible to achieve this.

ffenix113:
Also, while this MR works properly with reported attributes and fetching attributes from dev console, when you try to fetch the attribute value from the `Exposes` tab in the device - message `No converter available for '<attrbiute>_<endpointID>' ("")` is displayed as notification.

I think this is related to `zigbee2mqtt` repo and how it searches for converters, and not this one.

src/lib/modernExtend.ts
Koenkk: But why are there 10 endpoints with the temperature cluster on the device?

ffenix113: Just an arbitrary example.

Still, it is possible that device could have more than one endpoint with same cluster. And if it would be cluster like temperature or pressure(or any numeric one) - would trigger this behavior.

As another example: device has multiple environmental sensors attached, and those sensors have overlapping and distinct clusters. I.e. all sensors can have temperature cluster, but only one of them will have CO2, another one can have VOC, etc. Think like board can have attached Bosch BME280 & Sensirion SCD4X sensors. They both have temp & humidity clusters, but BME280 additionally has pressure, while SCD4X has CO2.

Koenkk: If I understand correctly, for your use-case it's important that the generate definition works without requiring any external converters. How would the generate definition decide for which endpoints to setup temperature reporting for if e.g. 10 of the endpoints have the temperature cluster? It sounds like we are fixing something in z2m here while it should be fixed at the device side instead.

ffenix113: > How would the generate definition decide for which endpoints to setup temperature reporting for if e.g. 10 of the endpoints have the temperature cluster?

If we provide specific endpoint we can then tailor extender for that specific cluster & endpoint(would it be cluster attribute value or endpoint to bind to).

> It sounds like we are fixing something in z2m here while it should be fixed at the device side instead.

Maybe we have a bit of misunderstanding.
For my example I chose temp cluster, because I tested with it, but consider other examples of clusters that also rely on this functionality for generated definition: 
* [onOff](https://github.com/Koenkk/zigbee-herdsman-converters/blob/6a0797ab702db6d92113bd1389f1ca29933c4ede/src/lib/modernExtend.ts#L117)
* [light](https://github.com/Koenkk/zigbee-herdsman-converters/blob/6a0797ab702db6d92113bd1389f1ca29933c4ede/src/lib/modernExtend.ts#L313)
* [lock](https://github.com/Koenkk/zigbee-herdsman-converters/blob/6a0797ab702db6d92113bd1389f1ca29933c4ede/src/lib/modernExtend.ts#L355)
* [electricityMeter](https://github.com/Koenkk/zigbee-herdsman-converters/blob/6a0797ab702db6d92113bd1389f1ca29933c4ede/src/lib/modernExtend.ts#L226)

I am suggesting that device can have multiple endpoints with the same cluster, and we should try to support it by generated definitions as well.
For example devices that have multiple endpoints with same cluster:
* [FUGA compatible switch from Smarthjemmet.dk](https://www.zigbee2mqtt.io/devices/QUAD-ZIG-SW.html)
* [Namron 4512768](https://www.zigbee2mqtt.io/devices/4512768.html)
* [Aurora Lighting AU-A1ZBDSS](https://www.zigbee2mqtt.io/devices/AU-A1ZBDSS.html)

Koenkk: > I am suggesting that device can have multiple endpoints with the same cluster, and we should try to support it by generated definitions as well.

Yes that would be nice, but instead of having e.g. `temperature(endpointID: 1)`, `temperatature(endpointID: 2`), we can do something similar to what we already do for [onOff](https://github.com/Koenkk/zigbee-herdsman-converters/blob/6a0797ab702db6d92113bd1389f1ca29933c4ede/src/devices/atsmart.ts#L10). The only thing that needs to be added multiple times is the expose, to/fromZigbee converters don't need to be duplicated.

ffenix113: Ah, I see.
Thank you for explanation, will uldate the implementation then!

ffenix113: Sorry, took a bit longer then I expected. Now `from` and `to` entities are not duplicated.

kirovilya:
oops. It turns out I also made support for multi endpoints... maybe I shouldn’t have done it if it works here too.
then I suggest you look at my code - maybe you’ll find something useful
https://github.com/Koenkk/zigbee-herdsman-converters/pull/6968

src/lib/generateDefinition.ts
Koenkk: Why not?

src/lib/generateDefinition.ts
```diff
@@ -4,92 +4,178 @@ import {getClusterAttributeValue} from './utils';
...
-interface GeneratedExtend {extend: ModernExtend, source: string, lib?: string}
-type ExtendGenerator = (endpoint: Zh.Endpoint) => Promise<GeneratedExtend[]>;
+interface GeneratedExtend {extend?: ModernExtend, extendFn?: (a: object) => ModernExtend, args?: object, source: string, lib?: string}
+type ExtendGenerator = (endpoints?: Zh.Endpoint[]) => Promise<GeneratedExtend[]>;
 type Extender = [string[], ExtendGenerator];
 
-function generateSource(device: Zh.Device, generatedExtend: GeneratedExtend[]): string {
+function generateSource(zigbeeModel: string, definition: Definition, generatedExtend: GeneratedExtend[]): string {
     const imports: {[s: string]: string[]} = {};
+    const importsDeduplication = new Map<string, boolean>();
```
Koenkk, line 15: Why can't this be a `Set`?

src/lib/generateDefinition.ts
Koenkk: That's because `definition` does not always have a `zigbeeModel`, I suggest to create a new local type here:

```
type DefinitionWithZigbeeModel = Definition & {zigbeeModel: string[]};
```

src/lib/generateDefinition.ts
```diff
@@ -113,25 +199,25 @@ async function extenderOnOffLight(endpoint: Zh.Endpoint): Promise<GeneratedExten
...
-        return [{extend: m.onOff({powerOnBehavior: false}), source: 'onOff({powerOnBehavior: false})'}];
     }
+
+    return generated;
 }
 
-async function extenderElectricityMeter(endpoint: Zh.Endpoint): Promise<GeneratedExtend[]> {
+async function extenderElectricityMeter(endpoints: Zh.Endpoint[]): Promise<GeneratedExtend[]> {
+    // TODO: Support multiple endpoints
+    const endpoint = endpoints[0];
```
Koenkk, line 214: Please throw an error if multiple endpoints are passed

src/lib/generateDefinition.ts
```diff
@@ -4,92 +4,178 @@ import {getClusterAttributeValue} from './utils';
...
 
 const outputExtenders: Extender[] = [
-    [['genIdentify'], async (endpoint) => [{extend: m.identify(), source: 'identify()'}]],
+    [['genIdentify'], async () => [{extend: m.identify(), source: 'identify()'}]],
 ];
 
-async function extenderLock(endpoint: Zh.Endpoint): Promise<GeneratedExtend[]> {
+async function extenderLock(endpoints: Zh.Endpoint[]): Promise<GeneratedExtend[]> {
+    // TODO: Support multiple endpoints
+    const endpoint = endpoints[0];
```
Koenkk, line 158: Please throw an error if multiple endpoints are passed

test/generateDefinition.test.ts
```diff
@@ -105,9 +105,56 @@ const {temperature, onOff} = require('zigbee-herdsman-converters/lib/modernExten
...
+            },
+            externalDefintionSource: `
+const {temperature, onOff} = require('zigbee-herdsman-converters/lib/modernExtend');
+
+const definition = {
+    zigbeeModel: ['combo'],
+    model: 'combo',
+    vendor: '',
+    description: 'Automatically generated definition',
+    extend: [temperature({"endpoints":["1","2"]}), onOff({"powerOnBehavior":false})],
```
Koenkk, line 156: For other modernExtend the endpoints passed is an object, e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/e4b88b8704bc6baa00671a685eec0ce12c6781b0/src/devices/atsmart.ts#L10

This allows to specify a friendly name for the endpoint

ffenix113: Honestly I think that in this particular example the way that `onOff` handles some of its argument is "excessive".
I would propose to split it in to more extenders, like `[onOff({endpoints: ["left","center","right"]}), deviceEndpoints({left:1, center:2, right:3}), ota(someDefinition)]`, as I think it should not define which endpoints device has, nor, even more so, define the OTA configuration.

What do you think?

src/lib/generateDefinition.ts
ffenix113: I am not strong at Typescript(that's one reason), but from other languages I know that if I would to add a type parameter here, then it would mean that in all uses of this interface there should be a type parameter specified.

For example, if `GeneratedExtend` would be `GeneratedExtend<T>`, then what this should look like: 
```ts
type ExtendGenerator = (endpoints?: Zh.Endpoint[]) => Promise<GeneratedExtend[]>;
```
? If to provide single type here then we will not be able to have extender functions that will accept different set of argument types.

Maybe the other way is to provide a class that has two methods like
```ts
class Provider {
    getModernExtend(): ModernExtend
    getSource(): string
}
```
and to have this be implemented by other type, that is generic?

What would you suggest?

src/lib/generateDefinition.ts
ffenix113: Thank you. Updated.

src/lib/generateDefinition.ts
ffenix113: Thank you. Updated.

src/lib/generateDefinition.ts
ffenix113: Updated, please check.

src/lib/generateDefinition.ts
ffenix113: Updated, please check.

src/lib/generateDefinition.ts
Koenkk: I tried to do it with the generic but failed to do so. For functions like `extenderOnOffLight` it's also quite complicated as args can be either `OnOffArgs` or `LightArgs`. Therefore the provider class sounds like the best solution

ffenix113: Updated. Please check if solution is good.

test/generateDefinition.test.ts
ffenix113: Updated. I am not sure what specific check to do, so I didn't do any. If you would suggest which checks to do - I can give them a go.

I also didn't update current usages of `onOff` with endpoints, as it would increase the size of this MR. I can do it in the next MR, if this one will be merged.

Koenkk: Sounds good! Already merged this to get it into the 1 Feb release.

src/lib/modernExtend.ts
```diff
@@ -403,32 +412,60 @@ export function enumLookup(args: EnumLookupArgs): ModernExtend {
...
+    endpoint?: string, endpoints?: string[], reporting?: ReportingConfigWithoutAttribute,
     valueMin?: number, valueMax?: number, valueStep?: number, scale?: number, label?: string,
 }
 export function numeric(args: NumericArgs): ModernExtend {
     const {
-        name, cluster, attribute, description, zigbeeCommandOptions, unit, endpoint, reporting, valueMin, valueMax, valueStep, scale, label,
+        name, cluster, attribute, description, zigbeeCommandOptions, unit, reporting, valueMin, valueMax, valueStep, scale, label,
     } = args;
+
+    let endpoints = args.endpoints;
```
Koenkk, line 423: Maybe for a next pr, can remove the `endpoint` arg completely? If you want a single endpoint, just pass `endpoints` with a single element

PR #6934, opened 2024-01-19
Enable Identify for Vallhorn and Parasoll sensors

LarsStegman:
Enable Identify for Vallhorn and Parasoll sensors. Can you confirm I am using this right? I am able to call identify manually on the sensors and they will start blinking.


Koenkk:
Thanks!

PR #6945, opened 2024-01-23
Add Wireless switch with 6 buttons, 'TS004F', ['_TZ3000_r0o2dahu']

waterlou:
It is a 6 buttons wireless panel with model id TS004F.


waterlou:
Also added support for Aqara Z1 buttons

Koenkk:
Thanks!

PR #6949, opened 2024-01-24
feat: add Develco/Friend HMSZB-120

sjorge:
https://github.com/Koenkk/zigbee2mqtt/issues/20946

User confirmed it should have a temperature sensor, so it looks like a newer version of the HMSZB-110.

### DIFF
--- a/src/devices/develco.ts
+++ b/src/devices/develco.ts
@@ -728,7 +728,10 @@ const definitions: Definition[] = [
         exposes: [e.occupancy(), e.battery_low()],
     },
     {
-        zigbeeModel: ['HMSZB-110'],
+        whiteLabel: [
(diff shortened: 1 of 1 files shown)


src/devices/develco.ts
```diff
@@ -728,8 +728,9 @@ const definitions: Definition[] = [
         exposes: [e.occupancy(), e.battery_low()],
     },
     {
-        zigbeeModel: ['HMSZB-110'],
-        model: 'HMSZB-110',
+        whiteLabel: [{vendor: 'Frient', model: 'HMSZB-120'}, {vendor: 'Frient', model: 'HMSZB-110'}],
```
Koenkk, line 731: Please keep `model` as is and only add `HMSZB-120` as a whitelabel with a fingerprint containing a `zigbeeModel`

Koenkk:
Thanks!

PR #6951, opened 2024-01-24
fix: Ubisys H1 - remote_tempeature is not useful

sjorge:
I noticed the attribtute was not reportable making it kind of useless. And local_temp still had the local value making it hard to get up to date values for the temperature the TRV was working off.

In the next minor fw update they will fix this.

```
> I also got remote temperature binding working fine with zigbee2mqtt, I did notice that the custom remote temperature attribute does not support reporting unlike the local temperature variant. It would be nice if the remote temperature one would also support reporting.

With the next firmware we made a change, so that the LocalTemperature attribute consistently reflects the value of a remote temperature measurement when available, the overall outcome is equivalent to having the remote temperature as reportable.
```

Since remote_temperature is mostly useless and they are gonna update local_temperature in the next release, we can simple remove it.


Koenkk:
Thanks!

PR #6953, opened 2024-01-24
Update the _TZE200_mja3fuja

Luligu:
Hi, 
the Smart air box _TZE200_mja3fuja doesn't have the PM25 and in the DP 2 has the formaldehyde.
This update the device.


Koenkk:
Thanks!

PR #6954, opened 2024-01-24
IKEA: fix restoration of levelConfig.on_level

accek:
I noticed that after losing power, the Tradfri driver (ICPSHC24-10EU-IL-1) didn't properly restore the on_level property.
Should fix https://github.com/Koenkk/zigbee2mqtt/issues/19211.
It looks to me like a typo in a refactor mentioned in this comment: https://github.com/Koenkk/zigbee2mqtt/issues/19211#issuecomment-1765932843

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -58,7 +58,7 @@ const bulbOnEvent: OnEvent = async (type, data, device, options, state: KeyValue
             if (onLevel > 255) onLevel = 254;
             if (onLevel < 1) onLevel = 1;
 
-            device.endpoints[0].write('genLevelCtrl', {onLevel: onLevelRaw});
+            device.endpoints[0].write('genLevelCtrl', {onLevel: onLevel});
         }
     }
 };


Koenkk:
Many thanks!

mrmaximas:
@accek after power failure on_level = 0
You want to set it to 1, but this means that the brightness will be the lowest. How correct is this? It seems to me that 255 (previous) would be more correct.

accek:
Here's [a suggestion that the value reverts to `previous`](https://github.com/Koenkk/zigbee2mqtt/discussions/14779).
If it reverts to 0, this sounds suspisious (it's an invalid value according to Zigbee spec).

Maybe z2m should interpret 0 reported by the device as `previous`.
It's already suggested [in a comment](https://github.com/Koenkk/zigbee-herdsman-converters/blob/aebfa997703e22019e460e22bc0a6693186a63a9/src/converters/fromZigbee.ts#L589), but the implementation does not match the comment.
@pyos It looks like the initial implementation of `on_level` [came from you](https://github.com/Koenkk/zigbee-herdsman-converters/commit/682903ef9fa0559833a42bf866b381b1f838468c). Do you remember, if the implementation not matching the comment was intentional?

I can prepare another fix, but would rather clear up the confusion first.

mrmaximas:
> Here's [a suggestion that the value reverts to `previous`](https://github.com/Koenkk/zigbee2mqtt/discussions/14779). If it reverts to 0, this sounds suspisious (it's an invalid value according to Zigbee spec).
> 
> Maybe z2m should interpret 0 reported by the device as `previous`. It's already suggested [in a comment](https://github.com/Koenkk/zigbee-herdsman-converters/blob/aebfa997703e22019e460e22bc0a6693186a63a9/src/converters/fromZigbee.ts#L589), but the implementation does not match the comment. @pyos It looks like the initial implementation of `on_level` [came from you](https://github.com/Koenkk/zigbee-herdsman-converters/commit/682903ef9fa0559833a42bf866b381b1f838468c). Do you remember, if the implementation not matching the comment was intentional?
> 
> I can prepare another fix, but would rather clear up the confusion first.

I can only confirm that with the current fix I have no problems and everything works fine.

accek:
Good to hear. So I'd rather keep it as is at the moment unless there is a better reason not to. :)

mrmaximas:
> Good to hear. So I'd rather keep it as is at the moment unless there is a better reason not to. :)

I forgot to add that after restarting the brightness is set to 100% - 254.
I would like to go back to the previous brightness, i.e. 255 ;)

pyos:
technically the range of OnLevel is specified in the ZCL as 0x01-0x0f so 0 is simply not a valid value for it at all

src/devices/ikea.ts
```diff
@@ -58,7 +58,7 @@ const bulbOnEvent: OnEvent = async (type, data, device, options, state: KeyValue
             if (onLevel > 255) onLevel = 254;
```
JudahBella, line 58: Hope you don't mind me asking, why would onLevel be more than 255 and why isn't it clamped to 255 instead?

Also, under what circumstance(s) would `Number(onLevelRaw)` return a number greater than 255?

PR #6955, opened 2024-01-25
fix: correct scaling on Ubisys H1 local temperature offset

sjorge:
Should fix https://github.com/Koenkk/zigbee2mqtt/issues/20962, did not hear back from Ubisys on this one. But will verify with the display when I get home if they line up.

Will leave it in draft until I can verify.


Koenkk:
thanks!

PR #6958, opened 2024-01-25
[Feature] Add energy support for GS (HEIMAN) SKHMP30-I1

mrskycriper:
Found out that `SKHMP30-I1` has standard `seMetering` cluster and can report energy readings. I've verified the numbers and they are at least as accurate as provided power measurements.

Also replaced `vendor` from `HEIMAN` to `GS` as it is a white label device and `GS` vendor already [exists](https://www.zigbee2mqtt.io/supported-devices/#v=GS) in a database.

## TODO

- [x] add energy
- [x] observe and verify energy readings
- [x] refactor as GS device
- [x] convert to modernExtend

### DIFF
--- a/src/devices/gs.ts
+++ b/src/devices/gs.ts
@@ -1,6 +1,7 @@
 import {Definition} from '../lib/types';
 import * as exposes from '../lib/exposes';
 import fz from '../converters/fromZigbee';
+import tz from '../converters/toZigbee';
 import * as reporting from '../lib/reporting';
 import {light} from '../lib/modernExtend';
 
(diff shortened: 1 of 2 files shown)


mrskycriper:
Deiced to take a final look to numbers and it seems that default reporting intervals are enormous, previously the device would report every 30 seconds, but now it waited for 20 minutes and didn't register any changes at all.

mrskycriper:
Am I missing something? This config doesn't match with the numbers in the actual system after configuration

```ts
const configureLookup = {
        haElectricalMeasurement: {
            // Report change with every 5W change
            power: {attribute: 'activePower', divisor: 'acPowerDivisor', multiplier: 'acPowerMultiplier', forced: args.power, change: 5},
            // Report change with every 0.05A change
            current: {attribute: 'rmsCurrent', divisor: 'acCurrentDivisor', multiplier: 'acCurrentMultiplier', forced: args.current, change: 0.05},
            // Report change with every 5V change
            voltage: {attribute: 'rmsVoltage', divisor: 'acVoltageDivisor', multiplier: 'acVoltageMultiplier', forced: args.voltage, change: 5},
        },
        seMetering: {
            // Report change with every 5W change
            power: {attribute: 'instantaneousDemand', divisor: 'divisor', multiplier: 'multiplier', forced: args.power, change: 5},
            energy: {attribute: 'currentSummDelivered', divisor: 'divisor', multiplier: 'multiplier', forced: args.energy, change: 0.1},
        },
    };
```

Снимок экрана 2024-01-26 в 23 14 39

Power, Current and Voltage change intervals are 100 times bigger, than they should be

mrskycriper:
Ok, it looks like `haElectricalMeasurement` changes are scaled differently than `seMetering` and I don't quite understand where or how this conversion occurs

Koenkk:
The change is computed based on the divisor/multiplier in `modernExtend.ts`, this looks good, thanks!

PR #6962, opened 2024-01-26
 chore(type): Various typedoc improvements

krzysztof-kwitt:
Hello,

I wanted to move docs from README.md to corresponding typescript file, but I noticed that we still use jsdoc regardless we use typescript, so I migrated from jsdoc to tsdoc. Thanks to it, we don't have type duplication.


Koenkk:
Nice improvements, thanks!

PR #6964, opened 2024-01-26
[Hotfix] Return detection_interval for Aqara RTCGQ14LM

mrskycriper:
Somehow completely missed the fact, that I forgot to put `detection_interval` back into exposes during refactoring. This hotfix brings the option back


Koenkk:
Thanks!

PR #6967, opened 2024-01-27
[Fix] Add entity_category for more entites

mrskycriper:
Add `entity_category` to `detection_interval` and both Wirenboard sensors settings


Koenkk:
Thanks!

PR #6968, opened 2024-01-27
Allow multi endpoint for generateDefinition

kirovilya:
example of generating a description of a device with two sensors and three switches
```
const {temperature, temperature, onOff} = require('zigbee-herdsman-converters/lib/modernExtend');

const definition = {
    zigbeeModel: ['ptvo.switch111'],
    model: 'ptvo.switch111',
    vendor: 'ptvo.info',
    extend: [temperature({"endpoint":"l1"}), temperature({"endpoint":"l2"}), onOff({"powerOnBehavior":false,"endpoints":{"l1":1,"l2":2,"l3":3}})],
};

module.exports = definition;
```


Koenkk:
Let's check in https://github.com/Koenkk/zigbee-herdsman-converters/pull/6930 first.

PR #6974, opened 2024-01-28
Add support for Sunricher HK-SENSOR-4IN1-A

Crosenhain:
Adding initial support for the Sunricher HK-SENSOR-4IN1-A

### DIFF
--- a/src/devices/sunricher.ts
+++ b/src/devices/sunricher.ts
@@ -610,6 +610,30 @@ const definitions: Definition[] = [
         toZigbee: [],
         exposes: [e.contact(), e.battery()],
     },
+    {
+        zigbeeModel: ['HK-SENSOR-4IN1-A'],
+        model: 'HK-SENSOR-4IN1-A',
+        vendor: 'Sunricher',
+        description: 'Smart Home Sensor',
+        fromZigbee: [fz.occupancy, fz.ias_occupancy_alarm_1, fz.ias_occupancy_alarm_1_report, fz.battery, fz.humidity, fz.temperature, fz.led_on_motion, fz.occupancy_timeout, fz.illuminance],
+        toZigbee: [],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const firstendpoint = device.getEndpoint(1);
+            const thirdendpoint = device.getEndpoint(3);
+            const fourthendpoint = device.getEndpoint(4);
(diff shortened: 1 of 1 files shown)


Koenkk:

Crosenhain:
Sure, I'll give it a shot. The reason I did it the way I did it was because this sensor uses 5 endpoints in the clusters section and I didn't see a better way to handle it.

Crosenhain:
It produced this from the `Dev console`:
```const {batteryPercentage, temperature, humidity} = require('zigbee-herdsman-converters/lib/modernExtend');

const definition = {
    zigbeeModel: ['HK-SENSOR-4IN1-A'],
    model: 'HK-SENSOR-4IN1-A',
    vendor: 'Sunricher',
    extend: [batteryPercentage(), temperature(), humidity()],
};

module.exports = definition;
```

And that does produce useful temperature, humidity, and battery results. It doesn't have anything for illuminance or occupancy which I assume it because those aren't a part of modernExtend. It also sets up 3 binds between the endpoints and coordinator that contain the battery/temperature/humidity, but not for the occupancy or illumination endpoints. 

I've made another commit that still works, but uses what it can (and what worked for me) from modernExtend.

Koenkk:
These still need to be added to `modernExtend`, could you add them [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/982af17ef5d9847015f81ee7c2faae7fd5d9d867/src/lib/modernExtend.ts#L660)?

Crosenhain:
I'll give it a go, do you want that in a different pull request?

Koenkk:
Yes please!

Crosenhain:
With the work in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7083 by @mrskycriper this PR is no longer required for this sensor to work, I have re-paired it a few times since updating to the version containing that PR and everything looks good.

mrskycriper:
@Crosenhain does occupancy work as expected? Because not all sensors send motion end.

Also, does this sensor work without the help of external converter? I can’t seem to find it in the code base, so you should probably reopen the PR. With the modern extends implementation you can easily add this sensor permanently.

Crosenhain:
@mrskycriper Yes it does, I have been watching the MQTT topic for this sensor and it does seem the occupancy works as expected. 

You're right about the external converter - I'll sort that out now.

PR #6939, opened 2024-01-21
 LYWSD03MMC: add temp/humi callibration, show smile, comfort parameters

krzysztof-kwitt:
Hello,

I would like to add the following parameters:
* show_smile - Whether to show a smile on the device screen.
* temperature_offset - Temperature offset, in 0.1° steps
* humidity_offset - The humidity offset is set in 0.1 % steps.
* Comfort parameters

I am particularly interested in the offset configuration attributes, as I have 10 devices and I would like to calibrate them so that they return similar values.

I tested this on v.1.35.1-1 as custom converter: 
```js
const {Zcl} = require('zigbee-herdsman');
const {batteryPercentage, temperature, humidity, enumLookup, binary, numeric} = require('zigbee-herdsman-converters/lib/modernExtend');
const {isString} = require('zigbee-herdsman-converters/lib/utils');


function quirkAddEndpointCluster(args) {
    const {inputClusters, outputClusters} = args;

    const configure = async (device, coordinatorEndpoint, logger) => { 
        const endpoint = device.getEndpoint(1);
        inputClusters?.forEach((cluster) => {
            const clusterID = isString(cluster) ?
                Zcl.Utils.getCluster(cluster, device.manufacturerID).ID :
                cluster;

            if (!endpoint.inputClusters.includes(clusterID)) {
                endpoint.inputClusters.push(clusterID);
            }
        });

        outputClusters?.forEach((cluster) => {
            const clusterID = isString(cluster) ?
                Zcl.Utils.getCluster(cluster, device.manufacturerID).ID :
                cluster;

            if (!endpoint.outputClusters.includes(clusterID)) {
                endpoint.outputClusters.push(clusterID);
            }
        });

        device.save();
    }; 

    return {configure, isModernExtend: true}; 
} 

const definition = {
    zigbeeModel: ['LYWSD03MMC'],
    model: 'LYWSD03MMC (CUSTOM)',
    vendor: 'Xiaomi (CUSTOM)',
    description: 'Xiaomi temperature & humidity sensor with custom firmware',
    extend: [
        quirkAddEndpointCluster({
            endpointID: 1,
            outputClusters: [
                'hvacUserInterfaceCfg',
            ],
            inputClusters: [
                'genPowerCfg',
                'msTemperatureMeasurement',
                'msRelativeHumidity',
                'hvacUserInterfaceCfg',
            ],
        }),
        batteryPercentage(), 
        temperature({reporting: {min: 10, max: 300, change: 10}}), 
[...]

PR #6976, opened 2024-01-29
MHO-C401N: add new custom parameters

latargere:
Apply the same logic from this PR:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/6939
to add the new parameters to the mho-c401n


Koenkk:
Thanks!

PR #6978, opened 2024-01-30
feat: Add Compacta International, Ltd

jcgoette:
Please check me closely. Not very familiar with this.


jcgoette:
Right, 16 different irrigation channels.

Koenkk:
Updated the code, check if it works and add the remaining endpoints.

PR #6982, opened 2024-01-31
[Refactoring][Part 2] Updating Lumi (Aqara and Xiaomi) specific converters

mrskycriper:
# TODO

- [x] move Lumi related converters from `lib/legacy.ts`, `converters/fromZigbee.ts` and `converters/toZigbee.ts` to `lib/lumi.ts`
- [x] Update converter usages in `devices/lumi.ts`
- [x] Rename converters to `lumi_*`
- [x] Combine device specific converters into generic ones
- [x] Verify, what to do with `legacy` converters

# Questions and thoughts

- Can `legacy` functions be updated?
- Should they be updated? 
- It looks like some older devices will not work without legacy mode.


Koenkk:
> Can legacy functions be updated?
yes

> Should they be updated?
as minimal as possible

> It looks like some older devices will not work without legacy mode.
what do you mean?

mrskycriper:
> what do you mean?

I mean that right now they mostly don't work if legacy mode is not enabled.

Like here:
```js
lumi_buttons: {
    cluster: 'genOnOff',
    type: ['attributeReport'],
    options: [exposes.options.legacy()],
        if (isLegacyEnabled(options)) {
            const mapping: KeyValueAny = {4: 'left', 5: 'right'};
            const button = mapping[msg.endpoint.ID];
            if (button) {
                payload[`button_${button}`] = msg.data['onOff'] === 1 ? 'release' : 'hold';
                return payload;
            }
        }
    },
} satisfies Fz.Converter
```
But most importantly, what are the general differences between normal and legacy converters and how to update them?

Koenkk:
> But most importantly, what are the general differences between normal and legacy converters and how to update them?

These converters were kept to prevent a breaking change, therefore these were moved to legacy. On new installations `legacy: false` is automatically set in the z2m `configuration.yaml`, so legacy converters will not be used.

mrskycriper:
> These converters were kept to prevent a breaking change, therefore these were moved to legacy. On new installations legacy: false is automatically set in the z2m configuration.yaml, so legacy converters will not be used.

I've merged them all into non legacy ones. Does this look reasonable or they should be kept is separate but consolidated legacy converters?

Koenkk:
Let's leave all legacy in `legacy.ts`, this file will be completely removed in the future. Spend as little time as possible on it.

mrskycriper:
> Let's leave all legacy in legacy.ts, this file will be completely removed in the future. Spend as little time as possible on it.

@Koenkk left legacy alone but moved to `lib/limi.ts` for easier refactoring later.

Moved all Lumi related converters `lib/lumi.ts`, renamed and combined where possible. Probably not ideal and some more generalization is possible, but I feel like this is enough.

Merged a lot of conflicts, hopefully without regressions. Let me know if I can help with code review in some way as it's and enormous diff.

PR #6985, opened 2024-01-31
add new device ZS-SR-EUC

tdelsupexhe:


Koenkk:
Does it also work with the `TS0601_cover_5` definition from `tuya.ts`/

tdelsupexhe:
I just tried and couldn't get it to work

Koenkk:
What doesn't work?

tdelsupexhe:
I'll try to replace the definition from the one in TS0601_cover_5 and nothing work

Koenkk:
The should be the same:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/aebfa997703e22019e460e22bc0a6693186a63a9/src/devices/tuya.ts#L5927

tdelsupexhe:
Sorry I misunderstood what you have said. If I replace this line only, it works perfectly.

tdelsupexhe:
Done

Koenkk:
I mean removing it from `moes.ts` and merging it with `TS0601_cover_5`

tdelsupexhe:
Ok, I can do this but with the TS0601_cover_5 definition the backlight mode and the child lock does not work. I do not have a TuyaDatapoint 7 and 103 but I have a 14.

Koenkk:
I see, then this is fine!

PR #6990, opened 2024-02-01
fix: manuSpecificLumiicLumi should be manuSpecificLumi

sjorge:
Should fix https://github.com/Koenkk/zigbee2mqtt/issues/21120


sjorge:
@Koenkk should probably be merge before the februari release.

PR #6992, opened 2024-02-01
Detect ZNXNKG02LM button state during rotation

kryptan:
Aqara knob reports whether button is pressed during rotation. Currently rotating pressed knob results in events without the `action` field. 

* Fix `action` field - `start_rotating`, `rotation`, `stop_rotating` are now also used when rotating pressed knob.
* Add `action_rotation_button_state` field which is set to `pressed` or `released` during rotation.


Koenkk:
Thanks!

kryptan:
Looks like part of my commit was accidentally reverted in https://github.com/Koenkk/zigbee-herdsman-converters/commit/e122451393c87fffd63c1afd34b4bfe3b167ad68

Koenkk:
Whoops, fixed it!

PR #6993, opened 2024-02-01
fix: Niko 552-72301 state

VoltX:
Fix Niko 552-72301 State

https://github.com/Koenkk/zigbee2mqtt/discussions/20972


Koenkk:
Thanks!

PR #6996, opened 2024-02-01
Add essential features to BTH-RM

danielpetrovic:
Added heating_state entities, so you can actually use it as a thermostat. Currently the BTH-RM implementation doesn’t report the Heating state correctly. It is not available as entity and HA doesn’t know it.

This should work the same as the BTH-RM230Z.


danielpetrovic:
How can I install this change to my local Zigbee2MQTT install? To properly test it?

When I try to load the Bosch.ts through external converters it fails to load it.

Koenkk:
Thanks!

src/devices/bosch.ts
tsloms, line 1: After this update the BTH-RA (RBSH-TRV0-ZB-EU) work firmware 3.05.09, the window open switch is broken for me. Not sure how to support you to sort it out. Let me know

danielpetrovic: No, this patch was not the cause. The problem was caused by changing another file.

The issue is fixed in Z2M 1.36 and if you run Edge version.

tsloms: Thanks for your feedback, good do know! Than I'll wait for 1.36 being released for the homeassistant addon before I update.

salopette:
https://github.com/Koenkk/zigbee2mqtt/issues/21208#issuecomment-1973879602

PR #6997, opened 2024-02-01
Correct stge parsing by converting String to raw hex number

mildis:
Correct stge parsing by converting String to raw hex number


mildis:
@Koenkk Sorry for having missed this. Value is provided as a String representing an hex value. I missed the conversion, resulting in wrong results on parsing.

Koenkk:
Thanks!

PR #7001, opened 2024-02-02
Soft fail on multiple Electricity Meter endpoints

Nerivec:
Apparently throwing here prevents Z2M from starting. I set it to assert in the logs instead, but I'll let you change as needed @Koenkk since you co-authored the [commit](https://github.com/Koenkk/zigbee-herdsman-converters/commit/a1b6c282e6f98447028bcd912b7a9ac6953e05b2). It will just register the first passed endpoint for now, but I assume this is fine, instead of breaking Z2M, until proper support can be implemented?

Ref: 
https://github.com/Koenkk/zigbee2mqtt/issues/21170


Koenkk:
Thanks! Improved the PR a bit by using the new global logger.

PR #7003, opened 2024-02-03
fix: set 0 as transition for new devices

krzysztof-kwitt:
Hello,

I bought a new [IKEA bulb](https://www.zigbee2mqtt.io/devices/LED1836G9.html) and I thought it wasn't working because it had the transition value set to `""`(empty string), which caused the error as below:
```
error 2024-02-02 22:28:16: Publish 'set' 'brightness' to 'Zarówka Adama' failed: 'Error: 'transition' is not a number, got string ()'
debug 2024-02-02 22:28:16: Error: 'transition' is not a number, got string ()
    at toNumber (/app/node_modules/zigbee-herdsman-converters/src/lib/utils.ts:555:15)
    at Object.getTransition (/app/node_modules/zigbee-herdsman-converters/src/lib/utils.ts:408:28)
    at Object.convertSet (/app/node_modules/zigbee-herdsman-converters/src/converters/toZigbee.ts:1024:38)
    at Publish.onMQTTMessage (/app/lib/extension/publish.ts:249:52)
    at EventEmitter.wrappedCallback (/app/lib/eventBus.ts:174:23)
    at EventEmitter.emit (node:events:529:35)
    at EventBus.emitMQTTMessage (/app/lib/eventBus.ts:115:22)
    at MQTT.onMessage (/app/lib/mqtt.ts:141:27)
    at MqttClient.emit (node:events:517:28)
    at handlePublish (/app/node_modules/mqtt/src/lib/handlers/publish.ts:172:11)
```
This fix causes `""` (empty string) to be interpreted as 0, the default value.

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -405,7 +405,7 @@ export function getTransition(entity: Zh.Endpoint | Zh.Group, key: string, meta:
         const time = toNumber(message.transition, 'transition');
         return {time: time * 10, specified: true};
     } else if (options.hasOwnProperty('transition') && options.transition !== '') {
-        const transition = toNumber(options.transition || 0, 'transition');
+        const transition = toNumber(options.transition, 'transition');
         return {time: transition * 10, specified: true};
     } else {
         return {time: 0, specified: false};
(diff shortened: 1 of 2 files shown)


src/lib/utils.ts
```diff
@@ -405,7 +405,10 @@ export function getTransition(entity: Zh.Endpoint | Zh.Group, key: string, meta:
         const time = toNumber(message.transition, 'transition');
         return {time: time * 10, specified: true};
     } else if (options.hasOwnProperty('transition')) {
```
Koenkk, line 407:
```suggestion
    } else if (options.hasOwnProperty('transition') && options.transition !== '') {
```

transition == 0 is something different than no transition (`''`), I propose to make this change instead

Koenkk:
Thanks!

PR #7004, opened 2024-02-03
fix: Use writeStructure for Ubisys input configuration

sjorge:


sjorge:
Doesn't seem this is enough:
```
debug 2024-02-03 17:30:09: Error: WriteStructured 0x001fee0000005baa/232 manuSpecificUbisysDeviceSetup([{"attrId":{"ID":1,"type":72,"name":"inputActions"},"selector":{},"dataType":72,"elementData":{"elementType":"octetStr","elements":[[0,7,1,6,0,2],[0,134,1,8,0,1,0,50],[0,198,1,8,0,1,1,50],[0,11,1,8,0,3],[1,7,2,6,0,2],[1,134,2,8,0,1,0,50],[1,198,2,8,0,1,1,50],[1,11,2,8,0,3]]}}], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Data request failed with error: 'MAC no ack' (233))
    at ZStackAdapter.sendZclFrameToEndpointInternal (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/adapter/z-stack/adapter/zStackAdapter.ts:415:23)
    at Queue.execute (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/utils/queue.ts:35:20)
    at Request.send (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/controller/helpers/request.ts:79:20)
    at Endpoint.writeStructured (/opt/zigbee2mqtt/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:716:13)
    at Object.convertSet (/opt/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/devices/ubisys.ts:537:18)
    at Publish.onMQTTMessage (/opt/zigbee2mqtt/lib/extension/publish.ts:249:36)
    at EventEmitter.wrappedCallback (/opt/zigbee2mqtt/lib/eventBus.ts:174:17)
```

But the config worked once, now on a 2nd attempt I get the mac error and the device disapeared from the network :/

sjorge:
Sadly code does not seem to work :(

My C4 also seems dead, but I'm not sure this is related as it already dropping out of the network before this. I can't seem to reset it though.

sjorge:
Ubisys confirmed that there is no guard on the write in the current fw so I did indeed mess up my device.

Koenkk:
@sjorge does a factory reset bring it back to live?

sjorge:
> @sjorge does a factory reset bring it back to live?

Still going back and forth with them but it seems there is no guard and I just wrong incorrect payload (aka garbage) to part of the persistent storage. Even the reset button did not work, as it fails to power on.

I'm probably done poking this for now, as those devices are on the expensive (really expensive) side.

They seem to be interested in adding a fix for this in a future firmware. If they do and we have better writeStructure support I might pick this back up.

I did link the engineer this PR so hopefully they can replicate, and then prevent it.

PR #7011, opened 2024-02-04
FIX: fan-control for BAC-002-ALZB only

a2nt:
Fix for reported issue: https://github.com/Koenkk/zigbee2mqtt/issues/20619


Koenkk:
Thanks!

PR #7012, opened 2024-02-04
Add hint as an additional description for expose items

grafalex82:
Some devices may expose quite a lot of items. If these items are supplied with a description, the interface may get cluttered. If the description has certain amount of text describing values of the field, the interface gets too busy.

For example, 3-4 elements occupy the whole screen (even after the text was squeezed already):
image

The idea of the proposed change is to have a `hint` field in addition to the `description` field. The `description` may be used for a short 1-line description of the entity, while the hint may include big amount of text describing entity values or more detailed description of the setting. The `hint` text is not visible by default, but is shown in a hint/bubble when mouse is over the element.

This is the zigbee-herdsman-converter part of the improvement. Once approved, there will be also a corresponding change in the zigbee2mqtt-frontend project.

### DIFF
--- a/src/lib/exposes.ts
+++ b/src/lib/exposes.ts
@@ -15,6 +15,7 @@ export class Base {
     endpoint?: string;
     property?: string;
     description?: string;
+    hint?: string;
     features?: Feature[];
     category?: 'config' | 'diagnostic';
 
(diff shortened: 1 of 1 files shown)


Koenkk:
@nurikk what do you think about this? Doesn't it make more sense to truncate the text on the frontend?

nurikk:
> @nurikk what do you think about this? Doesn't it make more sense to truncate the text on the frontend?

Yeah, that’s definitely the fronted side issue

grafalex82:
Added UI part of the change https://github.com/nurikk/zigbee2mqtt-frontend/pull/1962

PR #7015, opened 2024-02-04
fix Xiaomi LYWSD03MMC converter

devbis:
Fix/revert of commit https://github.com/Koenkk/zigbee-herdsman-converters/commit/6de8b7c41efd9defffa4c17f88c69b67195d0d20 which was added for a fork instead of a original firmware.

Add features for fw version 1.1.0.

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/21165

Corresponding OTA PR: https://github.com/Koenkk/zigbee-OTA/pull/440

cc: @krzysztof-kwitt

### DIFF
--- a/src/devices/custom_devices_diy.ts
+++ b/src/devices/custom_devices_diy.ts
@@ -1,3 +1,4 @@
+import dataType from 'zigbee-herdsman/dist/zcl/definition/dataType';
 import * as exposes from '../lib/exposes';
 import fz from '../converters/fromZigbee';
 import * as legacy from '../lib/legacy';
@@ -1282,6 +1283,7 @@ const definitions: Definition[] = [
         },
     },
     {
+        // https://github.com/devbis/z03mmc/
         zigbeeModel: ['LYWSD03MMC'],
         model: 'LYWSD03MMC',
         vendor: 'Custom devices (DiY)',
@@ -1289,9 +1291,7 @@ const definitions: Definition[] = [
         extend: [
             quirkAddEndpointCluster({
                 endpointID: 1,
-                outputClusters: [
-                    'hvacUserInterfaceCfg',
-                ],
+                outputClusters: [],
                 inputClusters: [
                     'genPowerCfg',
                     'msTemperatureMeasurement',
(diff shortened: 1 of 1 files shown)


src/devices/custom_devices_diy.ts
```diff
@@ -27,6 +27,15 @@ const switchTypesList = {
     'multi-click': 0x02,
 };
 
+const dataType = {
```
Koenkk, line 30: Please use `import dataType from 'zigbee-herdsman/dist/zcl/definition/dataType';` instead

src/devices/custom_devices_diy.ts
```diff
@@ -1312,77 +1318,100 @@ const definitions: Definition[] = [
...
+                attribute: {ID: 0x0102, type: dataType.int16},
+                valueMin: -100.0,
+                valueMax: 100.0,
+                scale: 100,
+                description: 'Comfort parameters/Temperature minimum, in 0.01°C steps.',
             }),
             numeric({
                 name: 'comfort_temperature_max',
-                unit: 'C',
+                unit: 'ºC',
```
Koenkk, line 1370:
```suggestion
                unit: '°C',
```

src/devices/custom_devices_diy.ts
```diff
@@ -1312,77 +1318,100 @@ const definitions: Definition[] = [
...
+                scale: 100,
+                description: 'Comfort parameters/Humidity maximum, in 0.01% steps.',
             }),
         ],
         ota: ota.zigbeeOTA,
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            const bindClusters = ['msTemperatureMeasurement', 'msRelativeHumidity', 'genPowerCfg'];
+            await reporting.bind(endpoint, coordinatorEndpoint, bindClusters);
+            await reporting.temperature(endpoint, {min: 10, max: 300, change: 10});
```
Koenkk, line 1404: This should not be needed, e.g. this line is already handled by `temperature({reporting: {min: 10, max: 300, change: 10}}),`

src/devices/custom_devices_diy.ts
devbis: Nice catch! Thanks!

src/devices/custom_devices_diy.ts
devbis: Done

src/devices/custom_devices_diy.ts
devbis: Thanks! Changed.

src/devices/custom_devices_diy.ts
Koenkk: The configure is still there.

devbis: The configure is to read additional attributes optionally. They don't exist in older firmware versions.
Otherwise, there are errors in a loop and device cannot be configured.

Soukyuu:
A question about the calibration settings: what is the difference between these and the ones on the "settings (specific) tab? Are they the same, device local vs. sent, or additive?

PR #7017, opened 2024-02-05
added support for Danfoss Icon2 devices

JensenInnovation:
added support for Danfoss Icon2 devices as part of icon model as interfaces is close to identical


Koenkk:
thanks!

PR #7018, opened 2024-02-05
Aqara Z1 improvement

waterlou:
- use binary instead of enum for lock relay state
- use boolean instead of string for lock relay state
- decode lock relay payload


waterlou:
add led_disabled_night for z1 switch

src/lib/lumi.ts
```diff
@@ -1659,25 +1663,25 @@ export const toZigbee = {
...
                 break;
             case 'serving_size':
                 // @ts-expect-error
-                await sendAttr(0x0e5c0055, value, 4);
+                sendAttr(0x0e5c0055, value, 4);
                 break;
             case 'portion_weight':
                 // @ts-expect-error
-                await sendAttr(0x0e5f0055, value, 4);
+                sendAttr(0x0e5f0055, value, 4);
```
Koenkk, line 1684: Please keep all `await`

waterlou: fixed

Koenkk:
Thanks!

PR #7020, opened 2024-02-05
feat: Adds support for Vesternet Zigbee Sensors

martyn-vesternet:
VES-ZB-DOR-025 Door and window sensor, exposes contact, tamper, battery and link quality
VES-ZB-PIR-021 Motion sensor, exposes occupancy, tamper, battery and link quality
VES-ZB-TEM-027 Temperature and humidity sensor, exposes temperature, humidity, battery and link quality


src/devices/vesternet.ts
```diff
@@ -200,6 +200,55 @@ const definitions: Definition[] = [
             await reporting.batteryPercentageRemaining(device.getEndpoint(1));
         },
     },
+    {
+        fingerprint: [{modelID: 'DoorSensor-ZB3.0', softwareBuildID: '2.01'}, {modelID: 'DoorSensor-ZB3.0', softwareBuildID: '2.1'}, {modelID: 'DoorSensor-ZB3.0', softwareBuildID: '2.03'}, {modelID: 'DoorSensor-ZB3.0', softwareBuildID: '2.04'}],
```
Koenkk, line 204: Do we really need to specify all `softwareBuildID`? I prefer to have a single `{modelID: 'DoorSensor-ZB3.0', manufacturerName: 'MANUFACTURER_NAME_FROM_DATABASE_DB'}`

martyn-vesternet: Similar to other Vesternet devices, these devices are rebranded from the underlying hardware manufacturer, so I can only validate and test against firmware versions that have been supplied under the Vesternet branding.  

They might be different to those supplied to other companies that also rebrand them, so I would prefer to only submit code that is matched explicitly to the versions that Vesternet sell.

(I know for example that firmware version 2.02 in the contact sensor had issues with the IAS Cluster and didn't work correctly, hence I've excluded that version).

Koenkk:
I refactored the code a bit, can you check if all is ok?

martyn-vesternet:
Those changes seem to impact the reporting I'm configuring for each device though?

I try to set configuration so that it's a compromise on functionality / battery life and so that it matches that which I set on other platforms (Hubitat, SmartThings & Homey) for the Vesternet devices.

I prefer these for the humidity / temperature sensors:

`min: 60, max: 1800, change: 100` ( 1 degree / percent change, every 60 seconds at most, every 30 minutes at least ) 

whereas the config from "modernExtend" seems to be:

`min: '10_SECONDS', max: '1_HOUR', change: 100`

IMHO 10 seconds is too frequent, it will generate many reports in environments where the temperature / humidity changes rapidly.

And for battery:

`min: 600, max: 21600, change: 2` ( 1 percent change, every 10 minutes at most, every 6 hours at least ) 

(I just noticed I didn't set those correctly in my commit, so I will update them)

whereas the config from "modernExtend" seems to be:

`min: '1_HOUR', max: 'MAX', change: 10`

Which seems to only report on a 5 percent change and only report every 18 hours if no change?

Koenkk:
> I prefer these for the humidity / temperature sensors:

Agree about these ones, please update the defaults.

> Which seems to only report on a 5 percent change and only report every 18 hours if no change?

Yes, because it's unlikely that a sensors consumes more than 5 percent of battery every 18 hours

PR #7027, opened 2024-02-06
Fix definition of Vesternet VES-ZB-REM-013

### LINKED ISSUE #7024: Use of wildcard in list of possible values for enum expose
The Vesternet VES-ZB-REM-013 device use some wildcards like "on_*"
 in the definition of its Action expose 
https://github.com/Koenkk/zigbee-herdsman-converters/blob/0fcaeef8210c6c2635abe85c37e2a26c6f17d4f3/src/devices/vesternet.ts#L77-L78

This is causing some problems in the integration of Z2M with HomeSeer as it doesn't expect any wildcard here, and it prevents it to know exactly the number of buttons for this device. Is the use of wildcards standard here? I don't think I saw that for any other products.

Couldn't this be replaced by enumerating all the possible values instead, like for this other Vesternet device (ZG2833K8_EU05)

https://github.com/Koenkk/zigbee-herdsman-converters/blob/0fcaeef8210c6c2635abe85c37e2a26c6f17d4f3/src/devices/vesternet.ts#L187-L191

spudwebb:
Enumerate all possible action values instead of using wildcards for Vesternet VES-ZB-REM-013 device.
Fixes #7024


spudwebb:
Not sure why the review_request check returns an error?

Koenkk:
@martyn-vesternet  could you review this?

spudwebb:
@martyn-vesternet , did you get a chance to look at this fix?

martyn-vesternet:
Apologies for the delayed reply, this change is fine (not sure why I originally used wildcards here, since the other button devices don't).

Koenkk:
Thanks!

PR #7029, opened 2024-02-06
Fix error in BTH-RM after 6996.

danielpetrovic:
Included the wrong endpoint in 6996, which is not supported in this model.
Also adds Boost function, although it doesn't seem to do anything but the function is present.

The Running State is now working as intended which was fixed with 6996.

Found how to test it properly now, it fixes the error.

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -1191,23 +1191,22 @@ const definitions: Definition[] = [
                 maximumReportInterval: constants.repInterval.HOUR,
                 reportableChange: 1,
             }], manufacturerOptions);
-            // report pi_heating_demand (valve opening)
-            await endpoint.configureReporting('hvacThermostat', [{
-                attribute: {ID: 0x4020, type: Zcl.DataType.enum8},
-                minimumReportInterval: 0,
-                maximumReportInterval: constants.repInterval.HOUR,
-                reportableChange: 1,
-            }], manufacturerOptions);
-            // report is window_open
+            // report window_open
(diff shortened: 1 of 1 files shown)


danielpetrovic:
With the change it gives the following error. Is there a way for me to test this new change properly without merging it?

```
Zigbee2MQTT:error 2024-02-06 21:47:05: Failed to configure 'Kantoor Thermostaat', attempt 3 (Error: ConfigureReporting 0x001e5e090290f1cb/1 hvacThermostat([{"attribute":{"ID":16416,"type":48},"minimumReportInterval":0,"maximumReportInterval":3600,"reportableChange":1}], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":4617,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')
    at Endpoint.checkStatus (/app/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:319:28)
    at Endpoint.configureReporting (/app/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:672:22)
    at Object.configure (/app/node_modules/zigbee-herdsman-converters/src/devices/bosch.ts:1195:13)
    at Configure.configure (/app/lib/extension/configure.ts:120:13)
    at EventEmitter.wrappedCallback (/app/lib/eventBus.ts:174:17))
```

Koenkk:
Thanks!

PR #7032, opened 2024-02-07
Adding EuControls LCM-1C09-ZB

avwuff:
Hey, I'm probably doing this wrong but here's a PR to support this device.  Tested with actual hardware.


Koenkk:
Thanks!

PR #7038, opened 2024-02-07
Fix error in #7021 - the unwanted values are negative

svhelge:
The issue was not resolved with #7021 , the unwanted values are negative -2^31 = -0x80000000


Koenkk:
Thanks!

PR #7042, opened 2024-02-08
Slip relais and status register sub-values into their own JSON structure

mildis:
Relais details and status register details are moved to their own JSON, so it does not clutter the first-level values


Koenkk:
Thanks!

PR #7050, opened 2024-02-10
Schneider Electric MEG5126-0300/MEG5171-0000 and MEG5116-0300/MEG5171-0000

Nailik:
- added support for Schneider Electric MEG5126-0300/MEG5171-0000
- added switch_action and indicator_mode support to MEG5116-0300/MEG5171-0000

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -12,23 +12,55 @@ import {onOff, light, electricityMeter, identify, enumLookup} from '../lib/moder
 const e = exposes.presets;
 const ea = exposes.access;
 
-const indicatorMode = enumLookup({
-    name: 'indicator_mode',
-    lookup: {'reverse_with_load': 2, 'consistent_with_load': 0, 'always_off': 3, 'always_on': 1},
-    cluster: 'clipsalWiserSwitchConfigurationClusterServer',
-    attribute: {ID: 0x0000, type: 0x30},
-    description: 'Set Led Indicator Mode.',
-});
+function indicatorMode(endpoint?: string) {
+    let description = 'Set Indicator Mode.';
+    if (endpoint) {
+        description = 'Set Indicator Mode for ' + endpoint + ' Button.';
+    }
+    return enumLookup({
+        name: 'indicator_mode',
+        lookup: {
+            'reverse_with_load': 2,
+            'consistent_with_load': 0,
+            'always_off': 3,
+            'always_on': 1,
+        },
+        cluster: 'clipsalWiserSwitchConfigurationClusterServer',
+        attribute: {ID: 0x0000, type: 0x30},
+        description: description,
+        endpoint: endpoint,
+    });
+}
 
-const switchAction = enumLookup({
-    name: 'switch_action',
-    lookup: {0: 'light', 254: 'light_opposite', 1: 'dimmer', 253: 'dimmer_opposite', 2: 'standard_shutter', 252: 'standard_shutter_opposite',
-        3: 'schneider_shutter', 251: 'schneider_shutter_opposite', 4: 'scene', 5: 'toggle_light', 6: 'toggle_dimmer', 7: 'alternate_light',
-        8: 'alternate_dimmer', 127: 'not_used'},
-    cluster: 'clipsalWiserSwitchConfigurationClusterServer',
-    attribute: {ID: 0x0001, type: 0x30},
-    description: 'Set Switch Action.',
-});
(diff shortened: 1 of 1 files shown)


src/devices/schneider_electric.ts
```diff
@@ -15,6 +15,9 @@ const ea = exposes.access;
 const exposesLocal = {
     indicator_mode: e.enum('indicator_mode', ea.ALL, ['consistent_with_load', 'reverse_with_load', 'always_off', 'always_on'])
         .withDescription('Led Indicator Mode'),
+    switch_action: e.enum('switch_action', ea.ALL, ['light', 'light_opposite', 'dimmer', 'dimmer_opposite', 'standard_shutter',
```
Koenkk, line 18: Instead of adding the expose/fromZigbee/toZigbee converter, can you use the new `enumLookup` instead? Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/aeea2fc1b056c61c22e0bcd876e383825c883ccb/src/devices/custom_devices_diy.ts#L1305

This reduces code and improves consistency.

Nailik: I changed it, hope i did it correct as i was unable to test because i don't know how to import `modernExtend` into my custom converter.
I also updated the `indicator_mode`

Koenkk: do `const {enumLookup} = require('zigbee-herdsman-converters/lib/modernExtend');

Nailik: thanks - so doesn't quite work yet i always get

```
Error: Write 0x540f57fffe306b17/3 clipsalWiserSwitchConfigurationClusterServer({"1":{"value":1,"type":48}}, {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":4190,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')
    at Endpoint.checkStatus (/app/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:319:28)
    at Endpoint.write (/app/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:392:22)
    at Object.toZigbee.convertSet (/app/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:399:13)
    at Publish.onMQTTMessage (/app/lib/extension/publish.ts:249:36)
    at EventEmitter.wrappedCallback (/app/lib/eventBus.ts:174:17)
```

Can't really figure out the issue. Is there any way to activate verbose logging so i can see what z2mqtt sends when i use the old code - because it worked even though it didn't "look" as good.

Nailik: Also the `endpoint` parameter of `enumLookup` seems to behave weird because when i add it i get the error
`No converter available for 'switch_action_21' ...`.
Is it possible to use something as `withEndpoint` with `enumLookup` ? Else i would need to define it multiple times for different endpoints.

Nailik: So it's the wrong endpoint i think - how do i set the correct endpoint to `enumLookup`?

Nailik: sorry found my issue i was missing `meta: {multiEndpoint: true},` and 
```
endpoint: (device) => {
        return {
            'l3': 3, 'l21': 21, 'l2': 22,
        };
    },
```

Nailik:
finished the code and tested it on my devices

Koenkk:
Thanks!

alexeiw123:
@Nailik in the Z2MQTT dev channel, my Schneider 2AX switches (CH2AX/SWITCH/1) have dropped exposes for switch state and led indicator mode.
Could this be related to changes on those functions form this PR?

Dev release vs. stable channel screenshots below.

image
image

alexeiw123:
> i'll take a look

Assuming you don't have 2AX switches, let me know if I can help test.

Nailik:
@alexeiw123 i think i found it, i missed added `e.switch` after removing `extend.switch({ ...`.
I am not sure if you can test my master?

alexeiw123:
> @alexeiw123 i think i found it, i missed added `e.switch` after removing `extend.switch({ ...`. I am not sure if you can test my master?

If there is a way, it's probably beyond by git skill (I did my first PR just yesterday!)
If you merge I should be able to test in the next merge, which I think is daily?

If not perhaps using an external convertor.

Nailik:
@DanielNagy added it to my pr

PR #7051, opened 2024-02-10
Update neo.ts: Add support to TS0601 / _TZE204_q76rtoa9

pabloaugusto:
Here in Basil EKAZA manufacturer is selling this model of siren with this specs:

TS0601  
_TZE204_q76rtoa9

So, this commit will add support for it.

based on my reply to thread:
https://github.com/Koenkk/zigbee2mqtt/discussions/18649#discussioncomment-8427852

### DIFF
--- a/src/devices/neo.ts
+++ b/src/devices/neo.ts
@@ -36,8 +36,8 @@ const definitions: Definition[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #7053, opened 2024-02-10
Register support for "ROB_200-003-1" of ROBB

gkfirst8:
This commit will add support for ROB_200-003-1. It's technically the same as the already supported ROB_200-003-0 but then the local wired switch can be a normal switch. Images + notes have been registered via PR https://github.com/Koenkk/zigbee2mqtt.io/pull/2563


Koenkk:
Thanks!

PR #7056, opened 2024-02-11
feat: support _TZE204_ztqnh5cg

MondoGao:
## Description

Support _TZE204_ztqnh5cg Human Presence Detector. There is no model on the box but only ZT3L chipset model. I reuse a similar device's configuration and pass sanity test.

CleanShot 2024-02-10 at 22 23 51

## Verification
- Target Distance: Yes
- Illumination: Yes
- Presence: Yes (but my device is broken and only reporting 1 nor 0)

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -5113,7 +5113,7 @@ const definitions: Definition[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #7060, opened 2024-02-12
feat: Add new identify expose 

mrskycriper:
This PR introduces `identify` button with configurable timeout for future use in all devices with `genIdentify` cluster.

New `identify_timeout` option is added to `identify` converter. By default timeout is assumed to be 3 seconds.


mrskycriper:
I'm unsure of `device_category` set as `diagnostic`. ZHA clearly exposes it's identify button as `diagnostic` but I don't know if legacy select will work in such case.

Koenkk:
Could you also update https://github.com/Koenkk/zigbee-herdsman-converters/blob/8231fad9c0c5f58560683bd18c0ff60fc390f8b6/src/lib/modernExtend.ts#L93?

mrskycriper:
> Could you also update
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/8231fad9c0c5f58560683bd18c0ff60fc390f8b6/src/lib/modernExtend.ts#L93
> ?

@Koenkk Like this?
```js
export function identify(): ModernExtend {
    return {
        toZigbee: [tz.identify],
        exposes: [e.identify()],
        isModernExtend: true,
    };
}
```

Koenkk:
Yes, please also update the tests

mrskycriper:
@Koenkk fixed, now it will probably work with auto generated definitions.

src/converters/toZigbee.ts
```diff
@@ -231,8 +231,18 @@ const converters2 = {
     } satisfies Tz.Converter,
     identify: {
         key: ['identify'],
+        options: [exposes.options.identify_timeout()],
         convertSet: async (entity, key, value, meta) => {
-            await entity.command('genIdentify', 'identify', {identifytime: value}, utils.getOptions(meta.mapped, entity));
+            // External value takes prioroty over options for comapatability
+            let identifyTimeout = value;
```
Koenkk, line 237: I think this can be simplified to

```js
const time = value ?? meta.options.identify_timeout ?? 3
```

src/converters/toZigbee.ts
```diff
@@ -231,8 +231,18 @@ const converters2 = {
     } satisfies Tz.Converter,
     identify: {
         key: ['identify'],
+        options: [exposes.options.identify_timeout()],
         convertSet: async (entity, key, value, meta) => {
-            await entity.command('genIdentify', 'identify', {identifytime: value}, utils.getOptions(meta.mapped, entity));
+            // External value takes prioroty over options for comapatability
```
Koenkk, line 236:
```suggestion
            // External value takes priority over options for compatibility
```

src/converters/toZigbee.ts
mrskycriper: Simplified it as suggested

Koenkk:
Thanks!

PR #7062, opened 2024-02-12
fix bug with MEG5126-0300/MEG5171-0000

Nailik:
Very sorry!

I found a small issue with https://github.com/Koenkk/zigbee-herdsman-converters/pull/7050

MEG5126-0300/MEG5171-0000  reports brightness_l3 and state_l3 instead of brightness and state


Koenkk:
Thanks!

PR #7066, opened 2024-02-13
Rename MOES to Moes to unify vendor

tefracky:
Currently, MOES and Moes are used which duplicates the vendor name


Koenkk:
Thanks!

PR #7073, opened 2024-02-14
Add STPH-4-1-00 Device from NodOn

AlexisPolegato:
Add STPH-4-1-00 Device from NodOn

### DIFF
--- a/src/devices/nodon.ts
+++ b/src/devices/nodon.ts
@@ -147,6 +147,23 @@ const definitions: Definition[] = [
             await reporting.currentSummDelivered(ep);
         },
     },
+    {
+        zigbeeModel: ['STPH-4-1-00'],
+        model: 'STPH-4-1-20',
+        vendor: 'NodOn',
+        description: 'Temperature & humidity sensor',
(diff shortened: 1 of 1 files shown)


Koenkk:

src/devices/nodon.ts
```diff
@@ -147,6 +147,14 @@ const definitions: Definition[] = [
             await reporting.currentSummDelivered(ep);
         },
     },
+    {
+        zigbeeModel: ['STPH-4-1-00'],
+        model: 'STPH-4-1-20',
+        vendor: 'NodOn',
+        description: 'Temperature & humidity sensor',
+        extend: [batteryPercentage(), temperature({'endpoints': ['1']}), humidity({'endpoints': ['1']})],
```
Koenkk, line 155:
```suggestion
        extend: [batteryPercentage(), temperature(), humidity()],
```

Does this also work?

AlexisPolegato: Yes

Koenkk:
Thanks!

PR #7078, opened 2024-02-16
device VIYU_GU10_350_CCT_10454470

ExeErik:
add hornbach device VIYU_GU10_350_CCT_10454470

### DIFF
--- a/src/devices/hornbach.ts
+++ b/src/devices/hornbach.ts
@@ -79,6 +79,13 @@ const definitions: Definition[] = [
         description: 'FLAIR Viyu smart LED bulb CCT E27',
         extend: [light({colorTemp: {range: [220, 500]}})],
     },
+	{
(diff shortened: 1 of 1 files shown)


src/devices/hornbach.ts
```diff
@@ -79,6 +79,13 @@ const definitions: Definition[] = [
         description: 'FLAIR Viyu smart LED bulb CCT E27',
         extend: [light({colorTemp: {range: [220, 500]}})],
     },
+	{
+        zigbeeModel: ['VIYU_GU10_350_CCT_10454470'],
+        model: '10454470',
+        vendor: 'HORNBACH',
+        description: 'FLAIR Viyu smart GU10 CCT lamp',
+        extend: [light({colorTemp: {range: undefined}})],
```
Koenkk, line 87: Can you add the color temp range here? It should be part of the generated converter.

PR #7079, opened 2024-02-16
Remove tamper from different leak detector

Chartreusito:
No physical tamper switches on those devices. I personally own those 4 different devices and tested this.


Koenkk:
Thanks!

PR #7081, opened 2024-02-17
fix aqara z1 1 gang not working

waterlou:


src/devices/lumi.ts
```diff
@@ -2916,21 +2916,21 @@ const definitions: Definition[] = [
         ],
         meta: {multiEndpoint: true},
         endpoint: (device) => {
-            return {'button': 1};
+            return {'button_1': 1};
```
Koenkk, line 2919: Since this is a single rocker, why is multi endpoint needed in the first place? Can you try without `meta` and `endpoint: (device)`?

mrskycriper:
@waterlou do you have this device on hand? Could you test it without endpoints? I'm planing on converting Z1 to modern extend alongside Z1 Pro an it should work without endpoints as single rocker uses only one for switch and action

PR #6998, opened 2024-02-02
Initial work to get illuminance and occupancy into modernExtend

Crosenhain:
Attempting to modify modernExtend for use with [6974](https://github.com/Koenkk/zigbee-herdsman-converters/pull/6974)

Not yet tested, I am going to be setting up a test Z2M instance with a spare Zigbee radio to test this out.

PR #7083, opened 2024-02-17
feat: modern extend enhancements

mrskycriper:
This PR is based on #6998 and inspired by @Crosenhain.

### TODO
- [x] `ScaleFunction` as an alternative to linear value scaling for `numeric` extend
- [x] `illuminance` extend
- [x] `occupancy` extend
- [x] `dviceEndpoints` extend update
- [x] `entityCategory` option for `binary`, `numeric` and `enum` extends

`ScaleFunction` is passed as an alternative to fixed number for value scaling inside `numeric` extend. Argument `type` provides a way to distinguish between `fromZigbee` and `toZigbee` value conversions if they are asymmetrical.


mrskycriper:
Also I was thinking that maybe generic secondary exposes should have `entity_category` set by default?

src/lib/modernExtend.ts
```diff
@@ -118,7 +120,7 @@ export function onOff(args?: OnOffArgs): ModernExtend {
 
     const result: ModernExtend = {exposes, fromZigbee, toZigbee, isModernExtend: true};
     if (args.ota) result.ota = args.ota;
-    if (args.endpoints) {
+    if (args.endpoints && args.disableMutliEndpointOverride !== true) {
```
Koenkk, line 123: Can you give me a use case for this option? Why just not leave out `args.endpoints` if you don't want the endpoints?

src/lib/modernExtend.ts
```diff
@@ -439,6 +451,11 @@ export function numeric(args: NumericArgs): ModernExtend {
         if (valueMax !== undefined) expose = expose.withValueMax(valueMax);
         if (valueStep !== undefined) expose = expose.withValueStep(valueStep);
         if (label !== undefined) expose = expose.withLabel(label);
+        if (entityCategory) {
+            if ((args.access === 'ALL' && entityCategory === 'config') || (args.access !== 'ALL' && entityCategory === 'diagnostic')) {
```
Koenkk, line 455: Please move this logic to a function and also use it in `enumLookup`

src/lib/modernExtend.ts
```diff
@@ -410,15 +417,20 @@ export function enumLookup(args: EnumLookupArgs): ModernExtend {
     return {exposes: [expose], fromZigbee, toZigbee, configure, isModernExtend: true};
 }
 
+// reverseToZigbee provides a way to distinguish between fromZigbee (default) and toZigbee value conversions if they are asymmetrical
+export type ScaleFunction = (value: number, reverseToZigbee?: boolean) => number;
```
Koenkk, line 421:
```suggestion
export type ScaleFunction = (value: number, type: 'from' | 'to') => number;
```

I would make this a bit more explicit

src/lib/modernExtend.ts
```diff
@@ -751,3 +778,74 @@ export function pressure(args?: Partial<NumericArgs>): ModernExtend {
         ...args,
     });
 }
+
+export function illuminance(args?: Partial<NumericArgs>): ModernExtend {
```
Koenkk, line 782: Please also add to the definition generator: https://github.com/Koenkk/zigbee-herdsman-converters/blob/72c45d9ce26fafd4f6df5ca99e8c6fac17df097d/src/lib/generateDefinition.ts#L166

src/lib/modernExtend.ts
mrskycriper: Without it I can’t add modern onOff with multiple endpoints and another extend with endpoint at the same time. Because other extends require string endpoint name which requires endpoints to be defined. And this override brakes it because it’s only for onOff switches. If I add all endpoints here it will be confusing an not explicit as to what endpoints are specifically for onOff. Is it reasonable solution to this problem?

src/lib/modernExtend.ts
```diff
@@ -751,3 +778,74 @@ export function pressure(args?: Partial<NumericArgs>): ModernExtend {
...
+
+    const result: ModernExtend = illiminanceLux;
+    result.fromZigbee.concat(rawIllinance.fromZigbee);
+    result.toZigbee.concat(rawIllinance.toZigbee);
+    result.exposes.concat(rawIllinance.exposes);
+
+    return result;
+}
+
+export function occupancy(args?: Partial<BinaryArgs>): ModernExtend {
```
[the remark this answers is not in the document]
mrskycriper: Done

src/lib/modernExtend.ts
mrskycriper: Done

src/lib/modernExtend.ts
mrskycriper: Done

src/lib/modernExtend.ts
mrskycriper: Done

mrskycriper:
@Koenkk Maybe there should be just modern extend for endpoint definition and then all other extends will just accept string names of endpoints? onOff will still have this override option without new flag for compatability

mrskycriper:
`skipEndpointPostfix` implementation is debatable, because I can't understand, how to retrieve device `meta.multiEndpointSkip` inside modern extends.

Koenkk:
> Maybe there should be just modern extend for endpoint definition and then all other extends will just accept string names of endpoints?

This is already possible, [example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/b983a11799833a7c94ead0bbe867ec1548392279/test/generateDefinition.test.ts#L156)

mrskycriper:
> This is already possible, example

My bad, it's really there. Though it doen't set `multiEndpoint` and `multiEndpointSkip`. I'll remove the duplicate.

> The only reason you need this is to use the correct endpoint in the generated toZigbee converter, right?

Well, while working WB-MSW-ZIGBEE v.4 I had to add a `genBinaryOutput` cluster on a separate endpoint and thus it has endpoint name attached. But it's the only cluster of this type being exposed. The main endpoint also has `genBinaryOutput` that turns Zigbee data transfer off in favor of Modbus and so isn't useful for Z2M. I really want some way to remove this endpoint postfix automatically, so it looks just like any other setting and not as a 4th main switch in disguise.

src/lib/modernExtend.ts
```diff
@@ -21,6 +23,14 @@ function getEndpointsWithInputCluster(device: Zh.Device, cluster: string | numbe
     return endpoints;
 }
 
+function validateEntityCategory(category: 'config' | 'diagnostic' | undefined, access: 'STATE' | 'STATE_GET' | 'ALL'): boolean {
+    if ((access === 'ALL' && category === 'config') || (access !== 'ALL' && category === 'diagnostic')) {
```
Koenkk, line 27: This check is already done when calling `expose.withCategory`, so no need to do it here.

src/lib/modernExtend.ts
```diff
@@ -100,12 +110,17 @@ export function identify(): ModernExtend {
...
 export interface OnOffArgs {
     powerOnBehavior?: boolean, ota?: DefinitionOta, skipDuplicateTransaction?: boolean, endpoints?: {[s: string]: number},
-    configureReporting?: boolean,
+    configureReporting?: boolean, endpointNames?: string[],
 }
 export function onOff(args?: OnOffArgs): ModernExtend {
     args = {powerOnBehavior: true, skipDuplicateTransaction: false, configureReporting: true, ...args};
 
-    const exposes: Expose[] = args.endpoints ? Object.keys(args.endpoints).map((ep) => e.switch().withEndpoint(ep)) : [e.switch()];
+    let exposes: Expose[];
```
Koenkk, line 118: Do you already need this, otherwise we can move this to another PR? I propose to first get rid of `endpoints?: {[s: string]: number}` to not have similar arguments (required refactoring all users to use `deviceEndpoint()`)

src/lib/modernExtend.ts
mrskycriper: Ok, I’ll remove the function then and will set it if it’s not undefined

Koenkk:
> Well, while working WB-MSW-ZIGBEE v.4 I had to add a genBinaryOutput cluster on a separate endpoint and thus it has endpoint name attached.

What modernExtend where you using for this? Because when no endpoint is specified, it will use the first endpoint by default.

src/lib/modernExtend.ts
mrskycriper: I wanted to use it in my refactoring, but this makes sense too

mrskycriper:
> > Well, while working WB-MSW-ZIGBEE v.4 I had to add a genBinaryOutput cluster on a separate endpoint and thus it has endpoint name attached.
> 
> What modernExtend where you using for this? Because when no endpoint is specified, it will use the first endpoint by default.

Local binary modern extend in the device file. The default endpoint is 1, but this indicator controls from endpoint 5. Endpoint 1 has the same cluster that is controlling modbus. So if I won’t specify the endpoint I get a sensor disable button. Which is understandably not what’s needed here. With the 5th endpoint specified the attribute is labeled `activity_indicator_l4`. And I simply wish to remove that postfix on the z2m side

Koenkk:
> With the 5th endpoint specified the attribute is labeled activity_indicator_l4. And I simply wish to remove that postfix on the z2m side

I see, but removing the postfix will make the messages send to endpoint 1 (since that is the default endpoint) (or am I wrong here?)

mrskycriper:
> > With the 5th endpoint specified the attribute is labeled activity_indicator_l4. And I simply wish to remove that postfix on the z2m side
> 
> I see, but removing the postfix will make the messages send to endpoint 1 (since that is the default endpoint) (or am I wrong here?)

But the endpoint is still set for the expose. I’m only changing the name, so it shouldn’t really affect things

mrskycriper:
@Koenkk I've stashed the changes to endpoint application inside onOff for now. Entity validator was removed. Also stashed prefix removal for now, I'll think about this logic and will probably implement it later as to not block the rest of the PR from merge. I think that's it for now, should be able to merge in a current state.

Koenkk:
All good, thanks!

PR #7084, opened 2024-02-17
Adding [TS0601 / _TZE200_ntcy3xu1] and [TS0205 / _TZ3210_up3pngle]

marcelhoogantink:
Both devices have identical cases but different electronics and logic.


src/devices/tuya.ts
```diff
@@ -3709,6 +3725,25 @@ const definitions: Definition[] = [
             ],
         },
     },
+    {
+        fingerprint: [
```
Koenkk, line 3729: As all code is the same except the picture, you can just use a `tuya.whitelabel` (look in this file for examples)

src/devices/tuya.ts
```diff
@@ -537,6 +537,23 @@ const definitions: Definition[] = [
         toZigbee: [],
         exposes: [e.gas(), e.tamper()],
     },
+    {
+        fingerprint: tuya.fingerprint('TS0205', ['_TZ3210_up3pngle']),
+        model: 'TS0205_smoke_2',
```
Koenkk, line 542: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/7084/files#r1493791719

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/2579

src/devices/tuya.ts
marcelhoogantink: Hi, battery is different:
```battery.low()```  vs ```batteryState()``` : binary vs. enum

File: src/devices/tuya.ts (hunk 2 above)
Comment (marcelhoogantink) [reply src/devices/tuya.ts:542]: Also differs in battery behavior.

Comment (Koenkk) [reply src/devices/tuya.ts:542]: I think battery is also exposed for the other, can you merge them?

Comment (marcelhoogantink) [reply src/devices/tuya.ts]: Done, merged TS0205 smoke detector.
Expose of battery seems to be done by ```batteryPercentage()```; I see a battery percentage, but I did not add a expose:

afbeelding


Also ```fz.ias_smoke_alarm_1``` returns more information than the devices reports, Should I create a ```fz.ias_smoke_alarm_2``` with only the return I see in the device debug in TuYA IOT ??

Comment (Koenkk) [reply src/devices/tuya.ts]: > And how to remove the blue/white update arrow button, since this does not work when the device is sleeping. (it is needed and works OK to ge battery-percentage when pairing)

This is the case for all devices exposing battery percentage, you need to wake-up the device right before sending it

> Also fz.ias_smoke_alarm_1 returns more information than the device _TZ3210_up3pngle reports according TuYa IoT, Should I create a fz.ias_smoke_alarm_2 with only the returns I see in the device debug in TuYA IoT, and then split this TS0205 again??

If all the exposed information is correct, no

File: src/devices/tuya.ts [hunk 3]
```diff
@@ -545,11 +545,20 @@ const definitions: Definition[] = [
...
+
+        ],
+        fromZigbee: [
+            fz.ias_smoke_alarm_1,
+            fz.ignore_basic_report,
         ],
-        fromZigbee: [fz.ias_smoke_alarm_1, fz.battery, fz.ignore_basic_report],
         toZigbee: [],
-        exposes: [e.smoke(), e.battery_low(), e.tamper()],
+        exposes: [
```
Koenkk, line 556: If it works, you can keep the `battery_low` expose in

src/devices/tuya.ts
marcelhoogantink: OK, so it is done!

src/devices/tuya.ts
marcelhoogantink: OK, I will do that

Koenkk:
Thanks!

src/devices/tuya.ts
marcelhoogantink: Battery-low is ```N/A``` after update and testing. So it is not working, or we have to wait for it (24 hours max?)

marcelhoogantink: Seems to work, it is 'onwaar' false now

PR #7086, opened 2024-02-18
Add trigger counter for MCCGQ11LM and SJCGQ11LM

Drafteed:
This will add trigger counter for [Aqara door sensor (MCCGQ11LM)](https://www.zigbee2mqtt.io/devices/MCCGQ11LM.html) and [Aqara Water leak sensor (SJCGQ11LM)](https://www.zigbee2mqtt.io/devices/SJCGQ11LM.html). Trigger count indicates how many times the sensor/gercon was triggered. Perhaps other devices can also provide that value, but I don’t have them to test.

The counter is reset after a power outage or repairing. 
The value is send approximately every 40-50 minutes (or by pressing a button on the device).

**Use case:**
Using this, we can make a door sensor, as a pulse counter for an electric/water/gas meter.


Koenkk:
Thanks

PR #7088, opened 2024-02-18
Update samotech.ts

alexeiw123:
Added device support for SM325-ZG

### DIFF
--- a/src/devices/samotech.ts
+++ b/src/devices/samotech.ts
@@ -103,16 +103,8 @@ const definitions: Definition[] = [
         zigbeeModel: ['SM325-ZG'],
         model: 'SM325-ZG',
         vendor: 'Samotech',
-        description: 'Zigbee Smart Pull Cord Dimmer Switch',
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

alexeiw123:
@Koenkk (I thinkk) your commits removed the options that disable effects and power on behaviour, which didn't work in my testing of the device. I was just working on another PR that would tidy this code of mine up a lot, but yours is much more concise.

suggest replacing
`extend: [light({configureReporting: true})],`
with
`exposes: extend.light_onoff_brightness({disablePowerOnBehavior: true, disableEffect: true}).exposes,`

I'm not sure if `configureReporting: true` is also required or not, I don't understand the options well enough.

alexeiw123:
@Koenkk yes, after updating dev channel, these two non functional options have returned.

image

alexeiw123:
I have attempted to fix this here https://github.com/Koenkk/zigbee-herdsman-converters/pull/7095

Testing in my local converter is a bit different to the way it's done in this git though.

PR #7097, opened 2024-02-20
Update the Namron thermostat socket plug 4512749-N to operate as thermostat

ok-67:
I have tested this as an external converter for some time, and seems to be operating as expected. The thermostat outlet socket can now operate as a thermostat by setting the manufacturer specific attribute. I have added the necessary fields to set the temperature and calibration and also to switch between thermostat mode and on/off mode.


src/devices/namron.ts
```diff
@@ -96,6 +111,80 @@ const tzLocal = {
             }
         },
     } satisfies Tz.Converter,
+
+    namronPlugg: {
```
Koenkk, line 115: Instead of adding from/toZigbee and exposes, can you use the new `modernExtend` instead? Example https://github.com/Koenkk/zigbee-herdsman-converters/blob/9658b12fb629735dd2b57647d443be1b6b8b67c5/src/devices/sonoff.ts#L559

ok-67: I'll have a look at it, but I have never programmed in java script before, but I am still learning :-) , so it might take some time

PR #7098, opened 2024-02-20
Update samotech.ts to remove unsupported exposes

alexeiw123:
Recent merge added support for device SM325 but changes made on merge added additional exposes for "Power On Behaviour" and "Effects". These aren't supported by the device and this maintains functionality but removes those exposes.

Previous merge is here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7088

Before this change, non functional exposes shown:
image

After this change:
image


Koenkk:
Thanks!

alexeiw123:
> Thanks!

Well that's certainly a lot tidier. I tried that approach but couldn't test it locally.

PR #7099, opened 2024-02-21
Updates to Linptech ES1ZZ

AntoineGS:
fix(linptech): start values for motion_detection_distance at 0
fix(linptech): start values for fading_time at 0
fix(linptech): use of correct unit for presence_keep_time (minutes)

I have tested the first two and they are honored by the device, though I am not sure if "minutes" is the correct keyword here, it does seem used in other devices.
The last one the following screenshot confirms that it is in fact minutes not seconds:
image


Koenkk:
thanks!

AntoineGS:
Thanks to you for the awesome project ;)

PR #7104, opened 2024-02-21
fix system mode for Acova PERCALE 2 and TAFFETAS 2 

cperriot:
merge percale and taffetas heater as they are sharing some ZigbeeModel anyway.
make the order "heat" works as it seems Acova invert it with "Off"
 fix rimraf --glob for windows npm clean to work

It remains what seems an anomaly to me, the local_temperature is the temperature set on the heater itself (override) and not a measure of the room temperature.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "lint": "eslint .",
     "test": "ZHC_TEST=true jest test",
     "test-watch": "ZHC_TEST=true jest test --watch",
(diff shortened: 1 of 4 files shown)


Koenkk:
Thanks!

florian02820:
Hello, sorry for my English, I have a taffetas2 and the radiator temperature is the same as the setpoint temperature. I am starting with GitHub.

PR #7106, opened 2024-02-22
fix(detect): Detect 500.46 as Paulmann 500.46

einFreak:
500.46 was reported as 929.63 with wrong description etc. before 

files for documentation are created:
Koenkk/zigbee2mqtt.io#2588


Koenkk:
Thanks!

PR #7107, opened 2024-02-22
fix: use deviceEndpoints extend instead of endpoint defenitions

mrskycriper:
Implemented stashed idea from #7083.

All modern extend now should accept `endpointName` or `endpointNames`. Endpoint defenition is only added using `deviceEndpoints` extend.


Crosenhain:
I think I'm hitting what this fixes when adding in my occupancy sensor: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7107/files#diff-47eee2b61258a9d48e4b9104d706a4aeacec2e1baba7d21f34db7d2963357d28R825

PR #7109, opened 2024-02-24
Support setting motor speed for Aqara roller shade driver E1

chaayac:
Adds support for exposing the motor speed for these drivers, addressing a few feature requests in the z2m repo:
- https://github.com/Koenkk/zigbee2mqtt/issues/18076
- https://github.com/Koenkk/zigbee2mqtt/issues/18431
- https://github.com/Koenkk/zigbee2mqtt/issues/16734

Screenshot of exposes interface:
image

I tested this with my own motors and it works well (and achieves what I wanted it to - to make the motor noise quieter!).

This is my first contribution here and I mostly got this working with trial-and-error, so please let me know if anything needs changing or if something should be done differently. Cheers!

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -1936,7 +1936,7 @@ const definitions: Definition[] = [
         fromZigbee: [lumi.fromZigbee.lumi_curtain_position, lumi.fromZigbee.lumi_curtain_status,
             fz.ignore_basic_report, lumi.fromZigbee.lumi_specific],
         toZigbee: [lumi.toZigbee.lumi_curtain_position_state, lumi.toZigbee.lumi_curtain_battery,
-            lumi.toZigbee.lumi_curtain_charging_status],
+            lumi.toZigbee.lumi_curtain_charging_status, lumi.toZigbee.lumi_curtain_motor_speed],
         onEvent: async (type, data, device) => {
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks! Could you refactor this to a modernExtend enumLookup? After that you can remove the toZigbee converter and exposes, example:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/266689131e0ccdb1c672830632dc82f6a9b9905b/src/lib/lumi.ts#L1323

chaayac:
> Thanks! Could you refactor this to a modernExtend enumLookup? After that you can remove the toZigbee converter and exposes, example:
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/266689131e0ccdb1c672830632dc82f6a9b9905b/src/lib/lumi.ts#L1323

Done, thanks for the tip.

Koenkk:
Thanks!

Pirateguybrush:
@chaayac - Can this be made to work with this similar driver from the same manufacturer?

https://www.zigbee2mqtt.io/devices/ZNCLBL01LM.html

EDIT: In case anyone finds their way here while trying to achieve the same thing, it's not possible - the curtain driver has no motor speed control.  I paired it to an Aqara hub to test the slow opening, and rather than slowing the motor, it just moves it by 1% every X seconds over the opening duration.

Something similar could probably be achieved via a HA script/automation

PR #7111, opened 2024-02-24
Improve `trigger_count` for MCCGQ11LM and SJCGQ11LM

Drafteed:
As follow-up of https://github.com/Koenkk/zigbee-herdsman-converters/pull/7086:

- Added filter for incorrect values;
- The counter is reset after each scheduled report (50 min), and not just after a power outage as I initially assumed;
- After scheduled report and reset, the counter will be = 1, so we decrement the value by one.


Koenkk:
Thanks!

PR #7113, opened 2024-02-25
Add color support to Acuity Brands Lighting RB56SC and RB56AC

zhihengq:
Adding color support to RB56SC and RB56AC.


Koenkk:
Thanks!

PR #7115, opened 2024-02-25
feat: Expose Breeze Mode to Inovelli VZM35

nathanfiscus:
- Expose Breeze Mode Composite to VZM35
- Remove condition in on\off converter that isn't allowing the switch to use the rampRateOffToOnRemote parameter.  (This is a reported fix from the Inovelli forums, some more cleanup of the logic here is probably needed.)


Koenkk:
Thanks!

PR #7117, opened 2024-02-25
lumi: add quirkCheckinInterval(1_HOUR) to battery devices

sjorge:
Should be the 2nd piece for https://github.com/Koenkk/zigbee2mqtt/issues/21446 I own none of these devices though, I just looked for once with battery info in meta or a e.battery expose.


Koenkk:
Thanks!

PR #7121, opened 2024-02-25
FIX: PC321 sum current

dduransseau:
Fix PC321 sum current since it only an integer register.
Add produced_energy attribute to seMetering cluster


src/lib/modernExtend.ts
```diff
@@ -174,6 +174,7 @@ export function electricityMeter(args?: ElectricityMeterArgs): ModernExtend {
             power: {attribute: 'instantaneousDemand', divisor: 'divisor', multiplier: 'multiplier', forced: args.power, change: 5},
             // Report change with every 0.1kWh change
             energy: {attribute: 'currentSummDelivered', divisor: 'divisor', multiplier: 'multiplier', forced: args.energy, change: 0.1},
+            // produced_energy: {attribute: 'currentSummReceived', divisor: 'divisor', multiplier: 'multiplier', forced: args.energy, change: 0.1},
```
Koenkk, line 177: This is commented out so won't have any effect.

dduransseau: Yes I revert this addition since it block the test, I have to investigate the cause.

PR #7127, opened 2024-02-27
Fix errors thrown by PTM 216Z GP commands without payload

Nerivec:
```log
debug 2024-02-27 08:51:01: Received Zigbee message from '0x000000000155f47a', type 'commandNotification', cluster 'greenPower', data '{"commandFrame":{},"commandID":104,"frameCounter":971,"options":0,"payloadSize":0,"srcID":22410362}' from endpoint 242 with groupID 2948
info  2024-02-27 08:51:01: MQTT publish: topic 'zigbee2mqtt/0x000000000155f47a', payload '{"action":"internal_104","device":{"friendlyName":"0x000000000155f47a","ieeeAddr":"0x000000000155f47a","manufacturerID":null,"model":"PTM 216Z","networkAddress":62586,"type":"GreenPower"},"elapsed":104886,"linkquality":216}'
info  2024-02-27 08:51:01: MQTT publish: topic 'zigbee2mqtt/0x000000000155f47a', payload '{"action":"","device":{"friendlyName":"0x000000000155f47a","ieeeAddr":"0x000000000155f47a","manufacturerID":null,"model":"PTM 216Z","networkAddress":62586,"type":"GreenPower"},"linkquality":216}'
info  2024-02-27 08:51:01: MQTT publish: topic 'zigbee2mqtt/0x000000000155f47a/action', payload 'internal_104'
```

Credits to @dduransseau & @chris-1243 😉 
https://github.com/Koenkk/zigbee2mqtt/discussions/21462#discussioncomment-8510459

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -3173,6 +3173,10 @@ const converters1 = {
             if (hasAlreadyProcessedMessage(msg, model, msg.data.frameCounter, `${msg.device.ieeeAddr}_${commandID}`)) return;
             if (commandID === 224) return;
 
+            if (!msg.data.commandFrame.raw) {
+                return {action: `internal_${commandID}`};
+            }
+
             // Button 1: A0 (top left)
             // Button 2: A1 (bottom left)
             // Button 3: B0 (top right)
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -3173,10 +3173,6 @@ const converters1 = {
             if (hasAlreadyProcessedMessage(msg, model, msg.data.frameCounter, `${msg.device.ieeeAddr}_${commandID}`)) return;
             if (commandID === 224) return;
 
-            if (!msg.data.commandFrame.raw) {
-                return {action: `internal_${commandID}`};
-            }
-
             // Button 1: A0 (top left)
             // Button 2: A1 (bottom left)
             // Button 3: B0 (top right)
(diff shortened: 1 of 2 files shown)


src/converters/fromZigbee.ts
```diff
@@ -3173,6 +3173,10 @@ const converters1 = {
             if (hasAlreadyProcessedMessage(msg, model, msg.data.frameCounter, `${msg.device.ieeeAddr}_${commandID}`)) return;
             if (commandID === 224) return;
 
+            if (!msg.data.commandFrame.raw) {
```
Koenkk, line 3176: Instead of doing this, can we change:

```
const ID = `${commandID}_${msg.data.commandFrame?.raw.slice(0, 1).join('_')}`; // added a ? here
```

And add it to the lookup? 

```
'104_undefined': 'short_press_2_of_2'
```

Or alternatively, we change `if (commandID === 224) return;` to `if (commandID === 224 || commandID == 104) return;`, which is my preference as this only seems to be send once during pairing?

Koenkk: If it's just this giving the error I propose to add it, it's much more user friendly.

Nerivec: Add them all to the lookup you mean?

Koenkk: This code applies to just the PTM 216Z, so not sure what actions are send.

I've updated the PR
- It will not give the error anymore, but log a `PTM 216Z: missing command` when there is no such command
- Added `short_press_2_of_2`

What do you think?

Nerivec: Sounds good. This way users can report commands that log errors, and support can be added as needed.
Those 3 GP converters (215-216) cover lots of actual devices from what I've seen.

Koenkk:
Thanks!

PR #7139, opened 2024-03-01
Reference attributes directly from cluster

lucasteligioridis:
# What?

Change to referencing the attributes directly from the cluster object.

Full disclosure I've looked at the code and from what I understand this is how we should be referencing the data from the cluster, but I would also really need your help here to validate that this is correct. The current implementation doesn't seem to work correctly as hitting the "refresh" button didn't seem to update the value of these LEDs.

Can you please just confirm both this PR and the https://github.com/Koenkk/zigbee-herdsman/pull/942 are compatible and that the integration should work in this way?


Koenkk:
Looks good, thanks!

PR #7145, opened 2024-03-01
Add support for RGBGenie ZB-3008

ejpenney:
This is the black version of the RGBGenie ZB-3009, I noticed that the ZB-3009 definition doesn't specify this is an endpoint device, but, at least for the ZB-3008, this seems required to get endpoints from the device.  In any event, this PR is working for mi ZB-3008 and I can see the separate endpoint events.

Documentation PR: https://github.com/Koenkk/zigbee2mqtt.io/pull/2606


Koenkk:
Thanks!

PR #7149, opened 2024-03-02
feat: Hue remote does not expose genPollControl

sjorge:
So we set a reasonable one via quirkCheckInterval, this has made pairing more reliable for me.


Koenkk:
> this has made pairing more reliable for me.

I guess you mean the `configure` here?

sjorge:
Yes the configure after the interview, before i was hitting timeouts.

Koenkk:
Thanks for clarifying!

Issue #7146, opened 2024-03-01
SchneiderElectric Indicator Mode not working


lucasteligioridis:
HI @Koenkk looks like I've introduced this error in these recent changes:
https://github.com/Koenkk/zigbee-herdsman/pull/942
https://github.com/Koenkk/zigbee-herdsman-converters/pull/7139

MAYBE this PR is required but I just have no idea how this is impacting the lights like below since it doesn't even use this:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/7144

I'm just not sure how this isn't working, I followed the work that was existing and this should be working but I'm stumped. I am on the latest release of 1.36.0 which includes all of above from what I've seen, but it feels like the cluster is just not available? Anything I can do here to debug more? I just find it hard to test this stuff.

Is it because there are 2 clusters in `zigbee-herdsman` that have the same ID and now that we're trying to reference the attribute by name it's using the first cluster ID and failing to lookup?
https://github.com/Koenkk/zigbee-herdsman/blob/882f2e47dff5e21c989c6f1df6238c15202b8db5/src/zcl/definition/cluster.ts#L5005
https://github.com/Koenkk/zigbee-herdsman/blob/882f2e47dff5e21c989c6f1df6238c15202b8db5/src/zcl/definition/cluster.ts#L5019

This is the error when trying to change the indicator mode now. Is there a typo in the attribute name I have missed?? 
```
Publish 'set' 'indicator_mode' to 'GuestRoom/Light/Ceiling/Globe' failed: 'Error: Write 0xdc8e95fffef6ed28/1 manuSpecificSchneiderLightSwitchConfiguration({"ledIndication":2}, {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":4190,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')'
```

The light used in this example is below, but I suspect its lights/devices using this will be broken: https://www.zigbee2mqtt.io/devices/41E2PBSWMZ_356PB2MBTZ.html

Koenkk:
I also don't see yet why it is not working, can you try with `attribute: {ID: 0x0000, type: 0x30},` (reverting https://github.com/Koenkk/zigbee-herdsman-converters/pull/7139/files partially)

PR #7150, opened 2024-03-02
Revert referencing the attribute name

lucasteligioridis:
This was changed in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7139 and seemed to break this integration. Unsure as to why this is not working anymore as everything logically looks correct.

For now lets revert this change and hope this resolves the bug and gets this part of the integration working again.

Partially fixes: https://github.com/Koenkk/zigbee-herdsman-converters/issues/7146


lucasteligioridis:
Closing this for now, I have tested this locally and still get the same failure! I suspect a previous PR has broken this, I never tested this recently.

PR #7152, opened 2024-03-03
Fix the broken LED indicator for Schneider Electric wiser devices

lucasteligioridis:
# What?

@Koenkk figured out that issue 👍🏼 
Adds endpoint argument to the indicator mode function so we can target a different endpoint for where this cluster is actually located. By default this will target the same endpoint the device is defaulted on, we need to target a separate endpoint where this LED indicator funcitonality exists.

I'm not sure if this is the best way to achieve this change here, but it 100% works from me testing it.

Also removed this power on feature from these devices as is incompatible and just floods the expose tab with things that don't work anyway.

A minor thing, which I'm sure you will be able to explain, in the UI it renders with an indent for some reason, not obvious to me, but feel free to change it if its obvious to you. No functionality is impacted, just cosmetic:
Screenshot 2024-03-03 at 12 12 58 pm

# Why?

Fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/7146
Broken in a previous PR: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7094

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -502,7 +502,7 @@ const definitions: Definition[] = [
         vendor: 'Schneider Electric',
         description: 'Wiser 40/300-Series module switch 2AX',
         ota: ota.zigbeeOTA,
-        extend: [onOff({powerOnBehavior: true}), indicatorMode('smart')],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #7153, opened 2024-03-04
ADD new device-SPM01-U01 & SPM02-U01 by BITUO-TECHNIK

script0803:
Add some expose about power/ power reactive/ power apperant and three phase.
Since the calculation of cluster in our product is different from other devices, I wrote local fz in case it affects other users.


src/devices/bituo_technik.ts
```diff
@@ -1,25 +1,144 @@
...
+import * as utils from '../lib/utils';
+import {
+    precisionRound,
+    postfixWithEndpointName, calibrateAndPrecisionRoundOptions,
+} from '../lib/utils';
+import {KeyValueAny, Fz, Definition} from '../lib/types';
 const e = exposes.presets;
 
+const fzLocal = {
+    seMetering: {
```
Koenkk, line 12: Please extend the existing converters in fromZigbee.ts to prevent code duplication.

script0803: Yes, I understand this, but we found that our converter is different from the one in fromZigbee. If we modify it in fromZigbee, it will affect the use of other devices, so I chose to add my own converter in the localfz way.

Koenkk: >  but we found that our converter is different from the one in fromZigbee 

It looks generic, so I then we can add it.

script0803: OK, I will resubmit the request, thank you

PR #7159, opened 2024-03-05
fix(ws-usc01-powersource): set powerSource to Mains for Aqara WS-USC01

gtalusan:
Hello, I recently installed an Aqara WS-USC01 single rocker no-neutral switch.  I noticed the user interface is showing "Battery ?" as the power source.  This patch will set it to mains.


Koenkk:
Thanks!

PR #7162, opened 2024-03-05
feat: Add led indicator mode on Schneider Electric S520559

Minims:
Hi,

I have a Schneider Electric S520559 detected as [EKO09716](https://www.zigbee2mqtt.io/devices/EKO09716.html#schneider%2520electric-eko09716)

I'm able to change le LED Indicator mode from the Dev Console while setting `manuSpecificSchneiderFanSwitchConfiguration` / `ledIndication` to :

 - 0: reverse_with_load
 - 1: consistent_with_load
 - 2: always_off
 - 3: always_on
 
It's quite strange, but yes this socket use `manuSpecificSchneiderFanSwitchConfiguration`
 that is supposed to be for FAN.

FYI, `fanIndicatorMode()` function have also `4` and `5` as possible value and they are not working on this socket. 
So i have duplicate the code, rename the function and change the enum.

I don't understand how make an external converters to validate this PR, but it might works as I understand.

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -32,6 +32,22 @@ function indicatorMode(endpoint?: string) {
     });
 }
 
+function socketIndicatorMode() {
+    const description = 'Set Indicator Mode.';
+    return enumLookup({
+        name: 'indicator_mode',
+        lookup: {
+            'reverse_with_load': 0,
+            'consistent_with_load': 1,
+            'always_off': 2,
+            'always_on': 3,
+        },
+        cluster: 'manuSpecificSchneiderFanSwitchConfiguration',
+        attribute: 'ledIndication',
+        description: description,
+    });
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Nailik:
@Minims good find! I actually tried `indicatorMode` but it didn't work - how did you find out that this socket uses `manuSpecificSchneiderFanSwitchConfiguration` ?

Minims:
> @Minims good find! I actually tried `indicatorMode` but it didn't work - how did you find out that this socket uses `manuSpecificSchneiderFanSwitchConfiguration` ?

I have just look at the Cluster tabs in Z2M to see all available input cluster, then I have check every cluster to see what action are available.

I think it works because `manuSpecificSchneiderFanSwitchConfiguration` already exists in Z2M for other devices.

PR #7164, opened 2024-03-05
fix: Inovelli bug fixes

nathanfiscus:
- Fixes an issue with readResponses that aren't in the attribute list of the device.
- Fixes an issue with VZM36 fan_mode converter
- Updates the fan speed values per Inovelli's spec


InovelliUSA:
These look good. Thanks!

Koenkk:
Thanks!

PR #7165, opened 2024-03-06
Add New device: SPM01-U01&SPM02-U01 by BITUO-TECHNIK

script0803:
Hello, in this request, I added semetering to fromzigbee.ts, and changed the original electrical_measurement in fromzigbee.
Adding apparentPowerPhB, apparentPowerPhC, reactivePowerPhB, reactivePowerPhC, and powerfactorPhB, powerfactorPhC. 
After testing, it will not conflict with the existing convert. Thank you for your previous guidance, thank you!


src/converters/fromZigbee.ts
```diff
@@ -728,6 +728,36 @@ const converters1 = {
             return payload;
         },
     } satisfies Fz.Converter,
+    seMetering: {
```
Koenkk, line 731: could we also merge this with https://github.com/Koenkk/zigbee-herdsman-converters/blob/4a18e2966318e3153cff32198c71dc554528d72a/src/converters/fromZigbee.ts#L697 ?

PR #7166, opened 2024-03-06
added converter for Danfoss Icon2 devices

TautrimasM:
I have created and added converter for Danfoss Icon2 devices, and removed Icon2 devices modelId's from Icon converter, as Icon2 devices support a wider range of features. Tested reading, writing values and reporting - working fine with both Icon2 thermostat and Icon2 Main controller.

### DIFF
--- a/src/devices/danfoss.ts
+++ b/src/devices/danfoss.ts
@@ -219,12 +219,12 @@ const definitions: Definition[] = [
     },
     {
         fingerprint: [
-            { modelID: '0x8020', manufacturerName: 'Danfoss' }, // RT24V Display
-            { modelID: '0x8021', manufacturerName: 'Danfoss' }, // RT24V Display  Floor sensor
-            { modelID: '0x8030', manufacturerName: 'Danfoss' }, // RTbattery Display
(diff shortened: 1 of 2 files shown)


src/devices/danfoss.ts
```diff
@@ -350,6 +348,127 @@ const definitions: Definition[] = [
                 'danfossMultimasterRole'], options);
         },
     },
+    {
+        zigbeeModel: ['0x0210', '0x8040', '0x8041', '0x8042'],
+        model: 'Icon2 MC or RT',
```
Koenkk, line 353:
```suggestion
        model: 'Icon2',
```

Can we use this here?

src/devices/danfoss.ts
```diff
@@ -350,6 +348,127 @@ const definitions: Definition[] = [
                 'danfossMultimasterRole'], options);
         },
     },
+    {
+        zigbeeModel: ['0x0210', '0x8040', '0x8041', '0x8042'],
```
Koenkk, line 352: Please use the `fingerprint` (same as for the `Icon` where you copied it from)

TautrimasM: Hi, thanks for suggestions, I have implemented them now.

Koenkk:
Thanks!

PR #7168, opened 2024-03-06
feat(ignore): Refactor more to modernExtend

Koenkk:


martyn-vesternet:
@Koenkk generally speaking I'm not averse to any changes as such, however have they actually been validated / tested against the Vesternet devices?

I've not looked at the code in depth, but typically when I submit for the Vesternet devices my code has been tested against as many revisions of the devices that I have to hand.

For example this particular change seems to remove the explicit exposes definitions and reporting configuration of the dimmer module and replace it with something generic.  Do these generic changes work the same way that my original code did with this device?

I'm not saying there's anything wrong with any of the code changes, but it's hard to keep track of how devices performed during testing that may have occurred months or even years ago.  There are currently around 25 Vesternet devices (Zigbee and Z-Wave) across 4 platforms (hubitat, smartthings, homey & home assistant) and I try to ensure that configuration & behaviour is identical across all of them.

There's an outstanding PR at https://github.com/Koenkk/zigbee-herdsman-converters/pull/7020 for example that I haven't gone back to yet for a similar reason.

From a commercial perspective it can cause issues if customers subsequently report problems with their devices caused by code that has not been tested by Vesternet.  

Unfortunately I don't have (paid) time to retest every single code change that comes through myself to be sure that nothing breaks. 

What do you think is the best way forwards?

Would it be better to revert to just providing code via the Vesternet Github like I do for other platforms - https://github.com/vesternet

Koenkk:
@martyn-vesternet this was just a general refactor, functionality is not impacted and should be the same.

Koenkk:
I missed the level config indeed, added it back! Btw I believe with the old converter level config was not working as it was missing `tz.level_config`, that's fixed now.

PR #7173, opened 2024-03-07
BAC-002-ALZB & BAC-003 fan control units support

robvanoostenrijk:
Updated support using new Tuya datapoints for both `BAC-002-ALZB` and `BAC-003` fan control units.

- Obsoletes legacy hgkg.ts for `BAC-002` devices
- Full climate control states with `off`, `cool`, `heat` & `fan_only` states, allowing Home Assistant mushroom cards to use the climate control directly without a separate on/off state switch
- Full schedule programme support with manual/schedule mode switch
- Local temperature calibration
- Deadzone temperature setting


Koenkk:
Bedankt!

PR #7175, opened 2024-03-07
Update tuya.ts

dhc25:
Added Tuya TS004F model _TZ3000_b3mgfu0d

This model is identical to _TZ3000_czuyt8lz and all reporting is the same.

It is worth noting that neither "_TZ3000_czuyt8lz" nor "_TZ3000_b3mgfu0d" support battery level reporting.

However this change is simply to add "_TZ3000_b3mgfu0d" to the supported devices.


Koenkk:
Thanks!

PR #7179, opened 2024-03-08
Extend SMSZB-120 alarm fix to HESZB-120

Rickth64:
In the past a fix was done for disabling the siren on SMSZB-120 devices that got updated to firmware 4.0.5 and higher. Now that OTA updates are supported for HESZB-120 (very similar device) and they are now on 4.0.8, this fix should be applied for this device as well.

Fixes Koenkk/zigbee2mqtt#21721


Koenkk:
Thanks!

PR #7195, opened 2024-03-11
New Moes Star Ring Smart switch (Light+Sence)

kirovilya:
SR-ZS model https://aliexpress.ru/item/1005006573223382.html

Some DP-based ModernExtend


src/lib/modernExtend.ts
```diff
@@ -584,6 +584,37 @@ export function actionEnumLookup(args: ActionEnumLookupArgs): ModernExtend {
...
+    const {actionLookup: lookup, command, attribute, cluster, buttonLookup} = args;
+
+    const actions = Object.keys(lookup).map((a) => args.endpointNames ? args.endpointNames.map((e) => `${a}_${e}`) : [a]).flat();
+    // allows direct external input to be used by other extends in the same device
+    if (args.extraActions) actions.push(...args.extraActions);
+    const expose = e.enum('action', ea.STATE, actions).withDescription('Triggered action (e.g. a button click)');
+
+    const fromZigbee: Fz.Converter[] = [{
+        cluster: cluster.toString(),
+        type: [command],
```
Koenkk, line 601: It looks that everything is the same as `actionEnumLookup` except this line, can we extend `actionEnumLookup` with a `type` argument which is `['attributeReport', 'readResponse']` by default?

kirovilya: hmm... really. At first I wanted to make that the commands did not have attributes, so I made a separate extension. but then I added an attribute and it became similar to the original one.

Koenkk:
Thanks!

PR #7197, opened 2024-03-11
Update namron.ts

ok-67:
Update the definition of Namron Thermostat outlet socket 4512749-N to operate as thermostat.

Note 1: Thermostat mode must be selected for this operation.
Note 2: System Mode must be set to heat before changing the temperature set point.

This should probably be added to the description, but I have not found how and where to do this.

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -8,7 +8,7 @@ import * as reporting from '../lib/reporting';
 import * as globalStore from '../lib/store';
 import * as ota from '../lib/ota';
 import * as utils from '../lib/utils';
-import {forcePowerSource, light, onOff} from '../lib/modernExtend';
+import {forcePowerSource, light, onOff, binary, temperature, electricityMeter} from '../lib/modernExtend';
 import * as tuya from '../lib/tuya';
 
 const ea = exposes.access;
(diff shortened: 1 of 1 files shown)


src/devices/namron.ts
```diff
@@ -98,6 +98,75 @@ const tzLocal = {
...
+            const payload = {0x0010: {value: Math.round(value*10), type: 0x28}};
+            await endpoint2.write('hvacThermostat', payload);
+            return {state: {local_temperature_calibration: value}};
+        },
+        convertGet: async (entity, key, meta) => {
+            const endpoint2 = meta.device.getEndpoint(2); // Thermostat cluster is located on endpoint 2
+            await endpoint2.read('hvacThermostat', ['localTemperatureCalibration']);
+        },
+    } satisfies Tz.Converter,
+    thermostat_running_state: {
```
Koenkk, line 161: Can we create new modernExtend for this instead? It's a lot of code duplication.

ok-67: OK, and I also see there is a bug, there is some problem with the access: 'ALL' to the binary operation_mode
So it should not be comited yet.

I actually need an option to only SET the attribute, the device do not support to read the state.

ok-67: Are there a modernExtend for thermostat? I have searched thru the code, but could not find any.

ok-67: > OK, and I also see there is a bug, there is some problem with the access: 'ALL' to the binary operation_mode So it should not be comited yet.
> 
> I actually need an option to only SET the attribute, the device do not support to read the state.

Actually when I made the first version without modernExtend I used ea.ALL and that did not make any problem, but when using modernExtend and with access: 'ALL' it gives this error vhen it is configuring the device, if the device already is configured with STATE, and I then recompile the program with ALL and restart without adding or removing the device I get no error. Actually it tries 3 times and give the same error message and then everything seems to be OK and it operates as expected.:

024-03-11 20:15:03: Configure reporting: false, read: true for 0x7fff0001000002c5/1 0 [{"attribute":{"ID":4096,"type":48},"min":-1,"max":-1,"change":-1}]
Zigbee2MQTT:debug 2024-03-11 20:15:04: Received Zigbee message from '0x7fff0001000002c5', type 'readResponse', cluster 'genBasic', data '{}' from endpoint 1 with groupID 0
Zigbee2MQTT:error 2024-03-11 20:15:04: Failed to configure '0x7fff0001000002c5', attempt 3 (Error: Read 0x7fff0001000002c5/1 genBasic([4096], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')
    at Endpoint.checkStatus (/zigbee2mqtt/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:319:28)
    at Endpoint.read (/zigbee2mqtt/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:469:22)
    at setupAttributes (/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:65:13)
    at configure (/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:86:17)
    at Object.configure (/zigbee2mqtt/node_modules/zigbee-herdsman-converters/src/index.ts:141:25)
    at Configure.configure (/zigbee2mqtt/lib/extension/configure.ts:120:13)
    at EventEmitter.wrappedCallback (/zigbee2mqtt/lib/eventBus.ts:174:17))

Koenkk: > Are there a modernExtend for thermostat? I have searched thru the code, but could not find any.

Not yet, but these have to be added.

ok-67: As a beginner in typescript, I'm afraid that task is not for me. 
But I will see if I can sort out the bug that makes an error when configure the Namron device.

ok-67: @Koenkk 
As I wrote earlier,  I discovered a bug when connecting to the device. Caused by the converter trying to read the state of the proprietary attribute that sets thermostat mode or on/off mode for the socket. This attribute is not readable, only writable, and this seems to causes the error 'UNSUPPORTED_ATTRIBUTE'

One way to solve it is to redefine the BinaryArgs in modernExtend.ts and add 'SET' to the available access modes:

export interface BinaryArgs {
    name: string, valueOn: [string | boolean, unknown], valueOff: [string | boolean, unknown], cluster: string | number,
    attribute: string | {ID: number, type: number}, description: string, zigbeeCommandOptions?: {manufacturerCode: number},
    endpointName?: string, reporting?: ReportingConfig, access?: 'SET' | 'STATE' | 'STATE_GET' | 'ALL', entityCategory?: 'config' | 'diagnostic',
}

Then the operation_mode for the device can be configured like this:

           binary({
                name: 'operation_mode',
                cluster: 0x0000,
                attribute: {ID: 0x1000, type: 0x30},
                description: 'Set operation mode as ON/OFF or as Thermostat.',
                valueOn: ['Thermostat', 0x01],
                valueOff: ['ON/OFF', 0x00],
                access: 'SET',
                entityCategory: 'config',
                zigbeeCommandOptions: {manufacturerCode: 0x1224},
            }),

Will this cause any problems for the rest of the code? or is there a better way of doing this?

Koenkk: Adding `set` is fine!

PR #7202, opened 2024-03-12
Fix FLSZB-110 temperature reporting

samuli-as:
Currently the Develco/Frient water leak detector's temperature reporting is not working. It only shows null. I looked at the way it is configured for the smoke and heat sensors and used the code from there.

One problem is that changing the reporting interval does not seem to work for this device so whatever values are set in the configuration effectively become hard-coded, unless there is some other way about this? If not, the minimum should perhaps be higher than 1 minute.

Related (stale and closed) issue: https://github.com/Koenkk/zigbee2mqtt/issues/17802


Koenkk:
Thanks!

PR #7204, opened 2024-03-12
fix: properly define generated device to be multiEndpoint

ffenix113:
Currently generated definition clears
all endpoint information,
effectively breaking configured binds.

Also add a better check if device
should be treated as multiEndpoint.

P.S.: amazing to see all the work around device definition generation!

### DIFF
--- a/test/generateDefinition.test.ts
+++ b/test/generateDefinition.test.ts
@@ -87,7 +87,7 @@ describe('GenerateDefinition', () => {
             meta: undefined,
             fromZigbee: [expect.objectContaining({cluster: 'msTemperatureMeasurement'}), fz.on_off],
             toZigbee: ['temperature', 'state', 'on_time', 'off_wait_time'],
-            exposes: ['linkquality', 'switch(state)', 'temperature'],
+            exposes: ['linkquality', 'switch_1(state)', 'temperature'],
             bind: {1: ['msTemperatureMeasurement', 'genOnOff']},
             read: {1: [
                 ['msTemperatureMeasurement', ['measuredValue']],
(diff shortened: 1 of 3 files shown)


test/generateDefinition.test.ts
```diff
@@ -107,10 +107,46 @@ const definition = {
     model: 'combo',
     vendor: 'vendor',
     description: 'Automatically generated definition',
-    extend: [temperature({"endpointNames":["1"]}), onOff({"powerOnBehavior":false})],
+    extend: [temperature({"endpointNames":["1"]}), onOff({"powerOnBehavior":false,"endpointNames":["1"]})],
```
Koenkk, line 110: Since this device has only 1 endpoint, I didn't expect the `endpointNames` here, why was this added?

Koenkk: I think that would be nicer, as most of the devices will have just one single endpoint. Note that z2m defaults to the first endpoint, that's not always endpoint ID 1

ffenix113: > I think that would be nicer, as most of the devices will have just one single endpoint

I agree. Though there are already some generators that have endpoints added as an argument in any case. For example temperature, humidity and other simple numeric ones. 

Also, for generators first endpoint then should be passed in, to check if generator should add the endpoint as an argument or not, and each generator should do this check by itself(some generators can be composed of multiple extenders, such as for `genOnOff` & `lightingColorCtrl`.

Just something to think about. For now I will update `onOff` extender not to include the first endpoint.

ffenix113: Updated. Thought of a way to not provide endpoint argument if only first device endpoint is passed in to generator, so it should clean up those simple numeric cases I mentioned(i.e. temperature, humidity).

And added some functions for generators to check if they should use endpoints argument or not, added a comment about it just as an information.

Koenkk:
Thanks!

PR #7223, opened 2024-03-15
[high-conflict] Update manufacturer codes to use ZCL definition

Nerivec:
AKA less magic numbers 😉

Remains these, which apparently are not officially assigned to them:

```typescript
    LIVOLO: 0x1AD2,
    LYTKO: 0x7777,
    PERENIO: 0x007B,
```

Like `SPRUT` I suppose they will have to be added as "custom" in herdsman before being changed here. Might I suggest using the `CUSTOM_` prefix, just to be clear and avoid any potential conflict on future ZCL def changes?


Koenkk:
> Might I suggest using the CUSTOM_ prefix, just to be clear and avoid any potential conflict on future ZCL def changes?

Yes please!

PR #7227, opened 2024-03-17
hotfix: reverse alarm state for contact zone

mrskycriper:


src/lib/modernExtend.ts
```diff
@@ -1052,12 +1052,25 @@ export function iasZoneAlarm(args: IasArgs): ModernExtend {
             };
 
             if (bothAlarms) {
-                payload = {[alarm1Name]: (zoneStatus & 1) > 0, ...payload};
-                payload = {[alarm2Name]: (zoneStatus & 1 << 1) > 0, ...payload};
+                if (args.zoneType === 'contact') {
```
Koenkk, line 1055: This line is duplicated 3 times, I suggest to add a variable below `const exposeList`: `const inverseAlarm = args.zoneType === 'contact'`

mrskycriper:
@Koenkk cleaned up the condition a bit. Also this gave me idea for later: option to reverse the payload from the GUI. Might be useful in certain installations.

Also there seems to be a bug with duplicate Tuya models in the main branch

Koenkk:
Thanks!

PR #7228, opened 2024-03-17
fix: Fix pairing of Legrand devices failing

Koenkk:
Requires https://github.com/Koenkk/zigbee-herdsman/pull/982

Fixes https://github.com/Koenkk/zigbee-herdsman/issues/979


sjorge:
tsc is not willing to build this

sjorge:
With the way it is done now, I guess we could also move this to a modernExtend we attach to all legrand/biticino/netamo devices ?

Koenkk:
>  I guess we could also move this to a modernExtend we attach to all legrand/biticino/netamo devices ?


We cannot, since this happen before interview finishes (until interview finishes, it is not linked against a definition).

Does the fix work?

sjorge:
Ah good point! We need it during interviewing too!

Yes, this code as is works.

Koenkk:
Great, will fix the tests and merge this.

PR #7230, opened 2024-03-17
Add Calibration attributes for SIN-4-RS-20

AlexisPolegato:
Add:
- Calibration command for SIN-4-RS-20
- Calibrations attributes for SIN-4-RS-20


src/devices/nodon.ts
```diff
@@ -54,6 +108,60 @@ const definitions: Definition[] = [
             await reporting.currentPositionLiftPercentage(endpoint);
             await reporting.currentPositionTiltPercentage(endpoint);
         },
+        extend: [
+            enumLookup({
```
Koenkk, line 112: Since this is duplicated, can we define a `const nodonModernExtend` in the top of the file and move them to there?

Koenkk:
Thanks!

PR #7234, opened 2024-03-19
New device ZBMicro is added.

liubotongs:
Added Sonoff manufacturer's new device ZBMicro.


src/devices/sonoff.ts
```diff
@@ -865,6 +926,26 @@ const definitions: Definition[] = [
...
+        extend: [
+            onOff(),
+            binary({
+                name: 'rf_turbo_mode',
+                cluster: 0xFC11,
+                attribute: {ID: 0x0012, type: 0x29},
+                zigbeeCommandOptions: {manufacturerCode: 0x1286},
+                description: 'Enable/disable Radio power turbo mode',
+                valueOff: ['DISABLE', 0x09],
+                valueOn: ['ENABLE', 0x14],
```
Koenkk, line 944:
```suggestion
                valueOff: [false, 0x09],
                valueOn: [true, 0x14],
```

liubotongs: I see that there is no ZBMicro content in the newly merged version, what went wrong?

Koenkk: This review comment still needs to be solved before merging, can you take a look at it?

Koenkk:
Thanks!

PR #7236, opened 2024-03-19
Add Zigbee 3.0 "1CH Mini Smart Switch"

cbrunnkvist:
(There is no other more specific model code anywhere on the device)


cbrunnkvist:
Mine is one of these nondescript Tuya-based devices. It has large button on its back (or "front", if you will) and blue indicator LED for triggering the switch. 

I'd say it qualifies as pretty much identical to the one pictured in [WHD02](https://www.zigbee2mqtt.io/devices/WHD02.html#tuya-whd02), aside from the label.

The only photo I have looks quite mucky, because I had to pull tape off of it since it had already been installed 😵‍💫

Screenshot 2024-03-19 at 21 36 01

Koenkk:
Thanks!

PR #7237, opened 2024-03-19
Fix for showing 200% battery instead of 100% for easy*-locks (Onesti Products AS)

newlund:
This commit reverts this:
https://github.com/Koenkk/zigbee-herdsman-converters/commit/ec956f7a1aeb8162f5188d995ff763260fef1c35
for easy*-locks

And breaks out Nimly locks to a new zigbeeModel section since only they seem to be reporting battery percentage wrong.

@Koenkk  - I'm really outside my comfort zone with this. Please close this if it is the wrong way of doing it :)

### DIFF
--- a/src/devices/onesti.ts
+++ b/src/devices/onesti.ts
@@ -58,7 +58,37 @@ const fzLocal = {
 
 const definitions: Definition[] = [
     {
-        zigbeeModel: ['easyCodeTouch_v1', 'EasyCodeTouch', 'EasyFingerTouch', 'NimlyPRO', 'NimlyCode', 'NimlyTouch', 'NimlyIn'],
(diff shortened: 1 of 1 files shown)


newlund:
Since the CI failed I guess it was not the right way to do it :)

model is 'easyCodeTouch_v1' and vendor is 'Onesti Products AS' for all locks. Only zigbeeModel is different.

newlund:
Ok, I put a new model name for the Nimly locks. Seems to pass CI checks now. :)

Koenkk:
Thanks!

msandvold (commented):
Result: 2024-03-30 10:55:51Read result of 'genPowerCfg': {"batteryPercentageRemaining":200}

PR #7220, opened 2024-03-14
fix: convert IKEA to modern extend

mrskycriper:
## TODO

- [x] revise device names for consistency
- [x] fix incorrectly assigned models
- [x] sort definitions into regions
- [x] use modern extends
- [x] turn converters into modern extends
- [x] turn configurations into modern extends
- [x] move IKEA functions and converters to `lib/ikea.ts`
- [x] create zigbee2mqtt.io PR with renamed images for changed model names

## Additional
[Modern extend improvements](https://github.com/Koenkk/zigbee-herdsman-converters/pull/7239)

Partial solution for https://github.com/Koenkk/zigbee2mqtt/issues/21951

### Model changes

- `LED1923R5/LED1925G6` to `LED1923R5` (LED1925G6 already exists)
- `ICPSHC24-10EU-IL-1` to `ICPSHC24-10EU-IL-1/ICPSHC24-10EU-IL-2` (same Zigbee model)
- `ICPSHC24-30EU-IL-1` to `ICPSHC24-30EU-IL-1/ICPSHC24-30EU-IL-2` (same Zigbee model)
- `LED1934G3_E26` and `LED1934G3_E27` to `LED1934G3` (same bulbs, different sockets)
- `LED1937T5_E26` and `LED1937T5_E27` to `LED1937T5` (same bulbs, different sockets)
- `LED1903C5/LED1835C6` to `LED1835C6` (LED1903C5 is a dumb dimmable LEDARE bulb)

Images for renamed models: https://github.com/Koenkk/zigbee2mqtt.io/pull/2662

PR #7239, opened 2024-03-19
feat: modern extend improvements

mrskycriper:
## TODO

- [x] sort Modern Extends roughly according to ZCL documentation chapters
- [x] update `identify` extend
  - [x] separate sleepy devices
  - [x] remove compatibility with numeric identify as it will be removed in #7220
- [x] add `linkquality` extend
- [x] add `pm25` extend
- [x] add `commandsOnOff` extend
- [x] add `commandsLevelCtrl` extend
- [x] add `commandsColorCtrl` extend
- [x] add `commandsWindowCovering` extend
- [x] add `windowCovering` extend
- [x] add `deviceTemperature` extend
- [x] add `flow` extend
- [x] add `soilMoisture` extend
- [x] merge multiple action exposes into one on processing
- [x] add GET converters to battery extend
- [x] fix meta in battery extend
- [x] move `identify` to inputExtenders
- [x] add `ota` extend to definition generator


src/index.ts
```diff
@@ -103,9 +104,26 @@ function processExtensions(definition: Definition): Definition {
             assert.fail(`'${definition.model}' has function exposes which is not allowed`);
         }
 
+        const actionExposes: Expose[] = [];
+
+        if (exposes) {
+            const filteredExposes: Expose[] = [];
```
Koenkk, line 110: Instead of doing the filtering both here and in `if (ext.exposes) {`. Can we do it just one time above `if (actionExposes.length > 0) {`?

src/lib/exposes.ts
```diff
@@ -634,7 +634,7 @@ export const presets = {
     switch_: () => new Switch(),
     // Specific
     ac_frequency: () => new Numeric('ac_frequency', access.STATE).withLabel('AC frequency').withUnit('Hz').withDescription('Measured electrical AC frequency'),
-    action: (values: string[]) => new Enum('action', access.STATE, values).withDescription('Triggered action (e.g. a button click)'),
+    action: (values: string[]) => new Enum('action', access.STATE, values).withDescription('Triggered action (e.g. a button click)').withCategory('diagnostic'),
```
Koenkk, line 637: Why is this a `diagnostic` `category`?

src/lib/modernExtend.ts
```diff
@@ -1,16 +1,20 @@
 import {Zcl} from 'zigbee-herdsman';
 import tz from '../converters/toZigbee';
 import fz from '../converters/fromZigbee';
-import {Fz, Tz, ModernExtend, Range, Zh, Logger, DefinitionOta, OnEvent, Access} from './types';
```
Koenkk, line 4: It's hard to review the diff of this file, can you summarise  what you changed here? You just added more modernExtends and re-ordered them?

src/lib/exposes.ts
mrskycriper: Because action isn’t something to show on a dashboard, as it will be cleared instantly. The main function of the entity_category is exactly that. To sort by get/set and to hide almost everything unnecessary from the dashboards by default

src/lib/modernExtend.ts
mrskycriper: Yes, the size was getting out of hands. I almost added pressure extend twice. They were sorted roughly according to ZCl documentation chapters. All additions and fixes are listed as todos. Mostly just new extends

src/index.ts
mrskycriper: Done, not sure why I did this in the first place.

src/lib/modernExtend.ts
sjorge: Would it make sense to break this up into a separate file per ZCL chapter and maybe 1 file for the custom stuff that is not in the ZCL ?

mrskycriper: There are not a lot of generic extends left to add (switch configuration and hvac I think) so I don’t see a point in splitting them any further. All manufacturer specific stuff is in their files anyway

src/index.ts
```diff
@@ -137,6 +139,33 @@ function processExtensions(definition: Definition): Definition {
             }
         }
 
+        const actionExposes: Expose[] = [];
```
Koenkk, line 142:
```suggestion
        const actionExposes = exposes.filter((e) => e.name === 'action'));
        exposes = exposes.filter((e) => e.name !== 'action'));
```

mrskycriper: Thanks! I'll change that now

src/index.ts
```diff
@@ -1,6 +1,7 @@
 import * as configureKey from './lib/configureKey';
-import * as exposes from './lib/exposes';
+import {options, presets} from './lib/exposes';
```
Koenkk, line 2: I would prefer to keep the old as `options` and `presets` are generic, it's not clear they come from exposes in the code.

src/index.ts
Koenkk: I updated the code a bit more, please check again.

src/lib/modernExtend.ts
```diff
@@ -25,12 +29,26 @@ function getEndpointsWithInputCluster(device: Zh.Device, cluster: string | numbe
     return endpoints;
 }
 
+function getEndpointsWithOutputCluster(device: Zh.Device, cluster: string | number) {
```
Koenkk, line 32: A lot of the code is the same as `getEndpointsWithInputCluster`, lets merge them into `getEndpointsWithCluster` and add `type: 'input' | 'output'` as an argument of it.

src/index.ts
mrskycriper: Exposes from the definition conflicts with exposes as an import. That is why I changed it. Any suggestions on how to handle this better?

src/lib/modernExtend.ts
mrskycriper: Makes sense

src/lib/modernExtend.ts
```diff
@@ -102,12 +120,217 @@ export function setupConfigureForReporting(
...
+        access: 'STATE_GET',
+        entityCategory: 'diagnostic',
+        ...args,
+    });
+}
+
+export function identify(args?: {isSleepy: boolean}): ModernExtend {
+    args = {isSleepy: false, ...args};
+    const normal: Expose = e.enum('identify', ea.SET, ['identify']).withDescription('Initiate device identification').withCategory('config');
+    const sleepy: Expose = e.enum('identify_sleepy', ea.SET, ['identify_sleepy'])
```
Koenkk, line 312: I think only the description should vary for sleepy devices, not the key (`identify_sleepy`)

mrskycriper: I added it to better indicate that it is sleepy in external systems (which don't have support for descriptions). But maybe you are right.

src/lib/modernExtend.ts
```diff
@@ -151,108 +374,274 @@ export function onOff(args?: OnOffArgs): ModernExtend {
...
-                    assertNumber(divisor, property.divisor);
-                    const multiplier = endpoint.getClusterAttributeValue(cluster, property.multiplier);
-                    assertNumber(multiplier, property.multiplier);
-                    let change: number | [number, number] = property.change * (divisor / multiplier);
-                    // currentSummDelivered data type is uint48, so reportableChange also is uint48
-                    if (property.attribute === 'currentSummDelivered') change = [0, change];
-                    items.push({attribute: property.attribute, min: '10_SECONDS', max: 'MAX', change});
+    if (args.bind) {
+        result.configure = async (device, coordinatorEndpoint, logger) => {
+            if (args.endpointNames) {
```
Koenkk, line 411: This `if/else` is duplicated quite some times, please move into a function.

src/lib/modernExtend.ts
```diff
@@ -151,108 +374,274 @@ export function onOff(args?: OnOffArgs): ModernExtend {
...
+                    if (start === '1970_UTC') {
+                        const time = Math.round(((new Date()).getTime()) / 1000);
+                        payload.time = time;
+                        payload.localTime = time - (new Date()).getTimezoneOffset() * 60;
+                    } else if (start === '2000_LOCAL') {
+                        const oneJanuary2000 = new Date('January 01, 2000 00:00:00 UTC+00:00').getTime();
+                        const secondsUTC = Math.round(((new Date()).getTime() - oneJanuary2000) / 1000);
+                        payload.time = secondsUTC - (new Date()).getTimezoneOffset() * 60;
+                    }
+                    data.endpoint.readResponse('genTime', data.meta.zclTransactionSequenceNumber, payload).catch((e) => {
```
Koenkk, line 449: This reverts my fix: https://github.com/Koenkk/zigbee-herdsman-converters/commit/4d0dc21f642833cec196d4082bcb3b3f287a81ca

src/lib/modernExtend.ts
```diff
@@ -362,780 +751,1213 @@ export function light(args?: LightArgs): ModernExtend {
...
+    const defaultSimulatedBrightness = 255;
+
+    const simulatedBrightness = [e.composite('simulated_brightness', 'simulated_brightness', ea.SET)
+        .withDescription(`Simulate a brightness value.` +
+        `If this device provides a brightness_move_up or brightness_move_down action it is possible to specify the update interval and delta.` +
+        `The action_brightness_delta indicates the delta for each interval.`)
+        .withFeature(e.numeric('delta', ea.SET).withValueMin(0).withDescription('Delta per interval, 20 by default'))
+        .withFeature(e.numeric('interval', ea.SET).withValueMin(0).withUnit('ms').withDescription('Interval duration'))];
+
+    const fromZigbee: Fz.Converter[] = [
```
Koenkk, line 781: It seems we duplicate a lot of code from existing converters here, can't we just use e.g. `const fromZigbee = [fz.command_stop, fz.command_step, ...` here?

src/lib/modernExtend.ts
```diff
@@ -362,780 +751,1213 @@ export function light(args?: LightArgs): ModernExtend {
...
-                const endpoint = endpoints?.find((e) => getEndpointName(msg, model, meta) === e);
-                if (endpoints && !endpoint) {
-                    return;
-                }
-
-                let value = msg.data[attributeKey];
-                assertNumber(value);
-                if (scale !== undefined) {
-                    value = typeof scale === 'number' ? value / scale : scale(value, 'from');
+    const fromZigbee: Fz.Converter[] = [
```
Koenkk, line 966: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/7239/files#r1536813772

src/lib/modernExtend.ts
```diff
@@ -362,780 +751,1213 @@ export function light(args?: LightArgs): ModernExtend {
...
+    const exposes: Expose[] = [coverExpose];
 
-    let actions = Object.keys(lookup).map((a) => args.endpointNames ? args.endpointNames.map((e) => `${a}_${e}`) : [a]).flat();
-    // allows direct external input to be used by other extends in the same device
-    if (args.extraActions) actions = actions.concat(args.extraActions);
-    const expose = e.enum('action', ea.STATE, actions).withDescription('Triggered action (e.g. a button click)');
+    const invertCover = [e.binary(`invert_cover`, ea.SET, true, false)
+        .withDescription(`Inverts the cover position, false: open=100,close=0, true: open=0,close=100 (default false).`)];
 
     const fromZigbee: Fz.Converter[] = [{
```
Koenkk, line 1210: Use `fz.cover_position_tilt` here instead of duplicating it.

src/lib/modernExtend.ts
```diff
@@ -362,780 +751,1213 @@ export function light(args?: LightArgs): ModernExtend {
...
                 }
-                return {[expose.property]: value};
             }
+            return result;
         },
     }];
 
-    return {exposes: [expose], fromZigbee, isModernExtend: true};
-}
+    const toZigbee: Tz.Converter[] = [
```
Koenkk, line 1243: Here you can use `tz.cover_state, tz.cover_position_tilt`

src/lib/modernExtend.ts
mrskycriper: I was thinking of removing those separate generic fz and tz after converting everything to modern extends to concentrate everything inside extends. What is the general strategy on keeping/removing separate converters? I see a lot of new devices being added that use those instead of existing modern extends

Koenkk: I propose to move them only once no devices uses them directly. So for now just refer to them to prevent code duplication.

src/index.ts
Koenkk: `import * as exposesLib from './lib/exposes';`

src/lib/modernExtend.ts
mrskycriper: Got it, will re-use them for now

src/lib/modernExtend.ts
mrskycriper: Fixed

src/lib/modernExtend.ts
mrskycriper: Added new `setupConfigureForBinding` for this

mrskycriper:
@Koenkk implemented all suggested changes

Koenkk:
Many thanks for another great refactor 😄

Koenkk:
Some devices might report it on their own, but the question is wether we want to batteryAlarmState to be reported when we already have the battery %

PR #7244, opened 2024-03-19
New device Zucchini from Llama Devices is added

grebenyuk51:
this is my personal device. Will add photo to zigbee2mqtt later


src/devices/llama_devices.ts
```diff
@@ -0,0 +1,169 @@
...
+const configs: Config[] = [
+    switchTypeConf,
+    switchAtionConf,
+];
+
+function addEndpointPrefix(ep: number) {
+    return `${prefixEnpointName}${ep}`;
+}
+
+function createSwitchExposes(epc: number) {
```
Koenkk, line 69: Instead of doing this, please create a modernExtend for this, [example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/6b3911a45601e5b01e916c74bcf51b9848d2e757/src/lib/lumi.ts#L1305)

src/devices/llama_devices.ts
```diff
@@ -0,0 +1,169 @@
...
+const definitions: Definition[] = [
+    {
+        zigbeeModel: ['Zucchini'],
+        model: 'Zucchini',
+        vendor: 'Llama Devices',
+        description: 'On/Off output device',
+        fromZigbee: [fzLocal.on_off_action, fz.battery, fz.temperature, fz.humidity, fzLocal.on_off_switch_config],
+        toZigbee: [tzLocal.on_off_switch_config],
+        exposes: [
+            e.action([]),
```
Koenkk, line 140: There are no actions here, it can be removed?

src/devices/llama_devices.ts
```diff
@@ -0,0 +1,169 @@
...
+            e.battery(),
+            e.temperature(),
+            e.humidity(),
```
Koenkk, line 143: Please use modern extend for all of these, [example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/6b3911a45601e5b01e916c74bcf51b9848d2e757/src/devices/third_reality.ts#L200), after that you can remove the configure, toZigbee and fromZigbee.

src/devices/llama_devices.ts
```diff
@@ -0,0 +1,165 @@
...
+                    for (let i = 0; i < configs.length; i++) {
+                        features.push(mapModernEnum(configs[i], addEndpointPrefix(ep)));
+                    }
+                }
+                return features;
+            })(endpointCount),
+        ],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const ep1 = device.getEndpoint(1);
+            await reporting.bind(ep1, coordinatorEndpoint, ['genPowerCfg']);
```
mrskycriper, line 144: Modern Extends already handle configure inside of them

src/devices/llama_devices.ts
```diff
@@ -0,0 +1,165 @@
...
+                .forEach(async (ep) => {
+                    await ep.bind('genOnOff', coordinatorEndpoint);
+                    await ep.read('genOnOffSwitchCfg', configs.map((c) => c.attribute));
+                });
+            await reporting.batteryPercentageRemaining(ep1, {min: 0, max: 3600, change: 0});
+            await reporting.temperature(ep1, {min: 0, max: 3600, change: 0});
+            await reporting.humidity(ep1, {min: 0, max: 3600, change: 0});
+            device.save();
+        },
+        endpoint: (device) => {
```
mrskycriper, line 157: There is already `deviceEndpoints` Modern Extend for this.

src/devices/llama_devices.ts
grebenyuk51: less code and still works. thanks

src/devices/llama_devices.ts
grebenyuk51: value for actions generates in function, not in exposes. anyway, I`ll use extend

src/devices/llama_devices.ts
grebenyuk51: battery, temperature and humidity done. Actions I cannot move to modernExtend. I don't have any attributes in commands. endpoint, cluster and command without attributes
image

src/devices/llama_devices.ts
grebenyuk51: yes, you are right. But, I don't want min 10 seconds, I want 0 seconds for min. Battery handles it, but other hardcoded `min: '10_SECONDS'`

src/devices/llama_devices.ts
grebenyuk51: yes, you are right. But, I don't want min 10 seconds, I want 0 seconds for min. Battery handles it, but other hardcoded min: '10_SECONDS'

src/devices/llama_devices.ts
grebenyuk51: but `deviceEndpoints` doesn't provide `device` instance. I don't want hardcode. I already hardcoded `enpointCount` and it hearts... :(

mrskycriper: Ah, I see, oh well. Will be refactored later I suppose

src/devices/llama_devices.ts
grebenyuk51: wait a minute, i can do it!
```
modernExtend.battery({
    percentageReporting: true,
    voltageReporting: false,
    percentageReportingConfig: {change: 0, max: 3600, min: 0},
}),
modernExtend.temperature({
    reporting: {min: 0, max: 3600, change: 0},
}),
modernExtend.humidity({
    reporting: {min: 0, max: 3600, change: 0},
}),
```
thanks a lot! configure now smaller :)

P.S. I'm from .NET world. TS for me is a new language

src/devices/llama_devices.ts
```diff
@@ -0,0 +1,150 @@
...
+    } satisfies Fz.Converter,
+};
+
+const definitions: Definition[] = [
+    {
+        zigbeeModel: ['Zucchini'],
+        model: 'Zucchini',
+        vendor: 'Llama Devices',
+        description: 'On/Off output device',
+        fromZigbee: [fzLocal.on_off_action],
```
Koenkk, line 115: Can't you use `fz.command_on, fz.command_off`, etc... here instead of creating a new converter?

src/devices/llama_devices.ts
Koenkk: But still actions need to be filled in here.

src/devices/llama_devices.ts
```diff
@@ -0,0 +1,150 @@
...
+        exposes: [
+            e.action([]),
+        ],
+        extend: [
+            modernExtend.battery({
+                percentageReporting: true,
+                voltageReporting: false,
+                percentageReportingConfig: {change: 0, max: 3600, min: 0},
```
Koenkk, line 123: Why override the reporting configs?

src/devices/llama_devices.ts
```diff
@@ -0,0 +1,150 @@
...
+            name: 'toggle',
+            value: 2,
+        },
+    ],
+};
+const configs: Config[] = [
+    switchTypeConf,
+    switchAtionConf,
+];
+
```
Koenkk, line 62: All these functions make the code hard to read, I propose to just do:

```
const switchCfg = (epName) =>  modernExtend.enumLookup({
        name: cfg.name,
        lookup: mapObject(cfg.values, (k) => k.name, (v) => v.value),
        endpointName: epName,
        cluster: 'genOnOffSwitchCfg',
        attribute: cfg.attribute,
        description: cfg.description,
        access: 'ALL',
        entityCategory: 'config',
    })
```

Remove all the `cfg.` and add them directly there.

src/devices/llama_devices.ts
```diff
@@ -0,0 +1,150 @@
...
+            ...createSwitchExpends(endpointCount),
+        ],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            device.endpoints.filter((ep) => ep.supportsInputCluster('genOnOffSwitchCfg'))
+                .forEach(async (ep) => {
+                    await ep.read('genOnOffSwitchCfg', configs.map((c) => c.attribute));
+                });
+            device.save();
+        },
+        endpoint: (device) => {
```
Koenkk, line 140: Use the `deviceEndpoints()` modern extend instead of this.

src/devices/llama_devices.ts
grebenyuk51: for what purpose? in this place they have no effect at all

src/devices/llama_devices.ts
grebenyuk51: i use one universal converter for all the commands and you want me to use three converters instead?

src/devices/llama_devices.ts
grebenyuk51: because i want to receive measurement as quicky as temperature/etc changes

src/devices/llama_devices.ts
grebenyuk51: I use values in configs in several places. you want me hardcode same values in different places?

src/devices/llama_devices.ts
grebenyuk51: deviceEndpoints() modern extend does not provide device instance. i have to hardcode endpoints?

grebenyuk51: I already hardcoded enpointCount and it hearts... :(

Koenkk:
> this is my personal device. 

I missed this initially, but is the device available for sale somewhere and are there instructions on how to make it?

grebenyuk51:
in progress

src/devices/llama_devices.ts
grebenyuk51: done

src/devices/llama_devices.ts
grebenyuk51: done

src/devices/llama_devices.ts
grebenyuk51: done

Koenkk:
That should be available first before this can be merged.

src/devices/llama_devices.ts
grebenyuk51: done

src/devices/llama_devices.ts
grebenyuk51: done

grebenyuk51:
ok

grebenyuk51:
ok

PR #7245, opened 2024-03-19
Update code that converts brightness to fan mode

InovelliUSA:
This fixes an issue where the vzm35-sn and vzm36 was not correctly reporting fan speed.


Koenkk:
Thanks!

PR #7248, opened 2024-03-21
Add support for IKEA Jetstrom 3030

jamesonuk:
This supports colour unlike the other two Jetstrom lights which have already been added.
Rather bizarrely this is a wall light but the model comes up as ceiling but the other two are ceiling lights but don't.....

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -1140,7 +1140,7 @@ const definitions: Definition[] = [
     },
     {
         zigbeeModel: ['JETSTROM 3030 ceiling'],
-        model: 'L2205, L2206 JETSTRÖM',
+        model: 'L2206',
         vendor: 'IKEA',
         description: 'JETSTRÖM LED wall light panel, smart dimmable/wired-in colour and white spectrum, 30x30cm',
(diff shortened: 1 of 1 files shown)


src/devices/ikea.ts
```diff
@@ -1138,6 +1138,13 @@ const definitions: Definition[] = [
         description: 'STOFTMOLN ceiling/wall lamp 15 warm light dimmable',
         extend: [tradfriLight()],
     },
+    {
+        zigbeeModel: ['JETSTROM 3030 ceiling'],
+        model: 'L2205, L2206 JETSTRÖM',
```
Koenkk, line 1143:
```suggestion
        model: 'L2205/L2206',
```

Can we use this here?

jamesonuk: I got the model from the manual (https://www.ikea.com/gb/en/manuals/jetstroem-led-wall-light-panel-smart-dimmable-wired-in-colour-and-white-spectrum__AA-2416878-1-100.pdf) but thinking about it, it is only one product so not sure where the other model comes into it.   Looking at the back it of the device it says L2206 so just going with that

Koenkk:
Thanks!

PR #7251, opened 2024-03-22
Generic zclCommand handler allowing it to be sent from mqtt

kirovilya:
For example:
```
zigbee2mqtt/0xa4c138b8c414612e/1/set
 {"zclcommand":{"command": "discover", "cluster": 6, "payload": {"startAttrId": 1, "maxAttrIds": 255}, "options":{}}}
```


src/converters/toZigbee.ts
```diff
@@ -238,6 +239,15 @@ const converters2 = {
             await entity.command('genIdentify', 'identify', {identifytime: identifyTimeout}, utils.getOptions(meta.mapped, entity));
         },
     } satisfies Tz.Converter,
+    zclCommand: {
+        key: ['zclcommand'],
+        convertSet: async (entity, key, value, meta) => {
+            utils.assertObject(value, key);
+            const payload = (value.hasOwnProperty('payload') ? value.payload : {});
+            await (entity as Endpoint).zclCommand(value.cluster, value.command, payload, (value.hasOwnProperty('options') ? value.options : {}));
```
Koenkk, line 247:
```suggestion
            utils.assertEndpoint(entity);
            await entity.zclCommand(value.cluster, value.command, payload, (value.hasOwnProperty('options') ? value.options : {}));
```

Koenkk:
@kirovilya could you take a look at the failing tests? After that this can be merged.

Koenkk:
Thanks!

klebom:
Nice addition, but the command result isn't propagated back to the caller
The same goes for the normal ```/set {"command":...}``` and ```/set {"read":...}```.
I've patched /set { "command": ...} handling with this additional logging for my personal use.
```
meta.logger.info(`Command result of '${value.cluster}.${value.command}': ${JSON.stringify(result)}`);
```
also I include ```${meta.options.friendly_name}``` when logging the command

```meta.logger.info(`Invoked command '${value.cluster}.${value.command}' on '${meta.options.friendly_name}' with payload '${JSON.stringify(value.payload)}'`);```

However I don't think using the log for these results is omptimal. A separate topic could be used for results like zigbee2mqtt/bridge/commandresults, or a topic for the result could be included with the command.

kirovilya:
Yes, here you need to expect an response command from the device and it is different for different commands.
we still need to improve it or make our own converters

PR #7252, opened 2024-03-22
Adding some parameters and fixing some ranges

InovelliUSA:
Some vzm35-sn parameters also should be on the vzm31. P55 & 56 range needs to be updated.


Koenkk:
Thanks!

PR #7253, opened 2024-03-22
fix: set color temperature range for BDHM8E27W70-I1 and expose battery voltages

mrskycriper:
- set color temperature range for BDHM8E27W70-I1 as per manual (and testing) to 2700-6500K
- expose battery voltage for SOHM-I1 and SWHM-I1
- sort extends


Koenkk:
Thanks!

PR #7256, opened 2024-03-22
feat(add): NAS-PS09B2

trinode:
Add support for the tuya based Neo NAS-PS09B2 Also sold under the brand Haozee.

https://www.aliexpress.com/i/1005005946786561.html

I ran this config for a while and it seems to be behaving itself so I'm raising this to add it out of the box instead of having to manually configure it with a custom JS file.

Let me know if anything is missing / wrong, but happy to have it corrected if it's quicker / easy.

Issue relating to this (which lead to this PR):
https://github.com/Koenkk/zigbee2mqtt/issues/19292

Other Issues related:
https://github.com/Koenkk/zigbee2mqtt/issues/19828
https://github.com/Koenkk/zigbee2mqtt/issues/19925
https://github.com/Koenkk/zigbee2mqtt/issues/20062
https://github.com/Koenkk/zigbee2mqtt/issues/20188

Note that the `occupancy` and `human motion state` might appear as null, however this appears to be due to the sensor being insanely sensitive and already being in the occupied state from before the device was configured and not regularly sending state updates, only on change. A quick fix is to just unplug the sensor and plug it back in and it'll send the correct states (that or leave the room for a long time and make sure it's pointing somewhere where it won't see micro movements (it seems like it can see through some materials), so it shows the unoccupied state.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -7433,6 +7433,42 @@ const definitions: Definition[] = [
         },
     },
     {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE204_kyhbrfyl']),
+        model: 'NAS-PS09B2 ',
+        vendor: 'Neo',
+        description: 'Human presence sensor',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        exposes: [
+            e.occupancy(),
+            e.enum('human_motion_state', ea.STATE, ['none', 'small', 'large'])
+                .withDescription('Human Motion State'),
+            e.numeric('departure_delay', ea.STATE_SET).withUnit('s').withValueMin(3)
+                .withValueMax(600).withValueStep(1).withDescription('Presence Time'),
+            e.numeric('radar_range', ea.STATE_SET).withUnit('cm').withValueMin(150).withValueMax(600)
+                .withValueStep(75).withDescription('Motion Range Detection'),
+            e.numeric('radar_sensitivity', ea.STATE_SET).withValueMin(0).withValueMax(7)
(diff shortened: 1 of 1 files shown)


Koenkk:
Cool, thanks!

IronicRebound77:
> Add support for the tuya based Neo NAS-PS09B2 Also sold under the brand Haozee.
> 
> https://www.aliexpress.com/i/1005005946786561.html
> 
> I ran this config for a while and it seems to be behaving itself so I'm raising this to add it out of the box instead of having to manually configure it with a custom JS file.
> 
> Let me know if anything is missing / wrong, but happy to have it corrected if it's quicker / easy.
> 
> Issue relating to this (which lead to this PR): [Koenkk/zigbee2mqtt#19292](https://github.com/Koenkk/zigbee2mqtt/issues/19292)
> 
> Other Issues related: [Koenkk/zigbee2mqtt#19828](https://github.com/Koenkk/zigbee2mqtt/issues/19828) [Koenkk/zigbee2mqtt#19925](https://github.com/Koenkk/zigbee2mqtt/issues/19925) [Koenkk/zigbee2mqtt#20062](https://github.com/Koenkk/zigbee2mqtt/issues/20062) [Koenkk/zigbee2mqtt#20188](https://github.com/Koenkk/zigbee2mqtt/issues/20188)
> 
> Note that the `occupancy` and `human motion state` might appear as null, however this appears to be due to the sensor being insanely sensitive and already being in the occupied state from before the device was configured and not regularly sending state updates, only on change. A quick fix is to just unplug the sensor and plug it back in and it'll send the correct states (that or leave the room for a long time and make sure it's pointing somewhere where it won't see micro movements (it seems like it can see through some materials), so it shows the unoccupied state.

Thanks, this worked for me, set sensitivity to "0" on all sliders in Z2M and it works.  very quick and responsive with delay set to "3"

PR #7259, opened 2024-03-23
Try to fix unrealistic pressure readings from WSDCGQ12LM.

rotdrop:
Similar to issue koenkk/zigbee2mqtt#798 my instance of the WSDCGQ12LM sporadically returns unrealistic high pressure readings > 30000 hPa. This PR tries to fix this in the same manner as issue koenkk/zigbee2mqtt#798 was fixed in commit 9bdebcd.

### DIFF
--- a/src/lib/lumi.ts
+++ b/src/lib/lumi.ts
@@ -2025,16 +2025,10 @@ export const fromZigbee = {
     lumi_pressure: {
         cluster: 'msPressureMeasurement',
         type: ['attributeReport', 'readResponse'],
-        convert: (model, msg, publish, options, meta) => {
-            let pressure = 0;
-            if (msg.data.hasOwnProperty('scaledValue')) {
-                const scale = msg.endpoint.getClusterAttributeValue('msPressureMeasurement', 'scale') as number;
-                pressure = msg.data['scaledValue'] / Math.pow(10, scale) / 100.0; // convert to hPa
-            } else {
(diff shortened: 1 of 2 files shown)


src/lib/lumi.ts
```diff
@@ -2022,6 +2022,22 @@ export const fromZigbee = {
             }
         },
     } satisfies Fz.Converter,
+    lumi_pressure: {
+        cluster: 'msPressureMeasurement',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            let pressure = 0;
+            if (msg.data.hasOwnProperty('scaledValue')) {
```
Koenkk, line 2030: I think this part can to into the official `fz.pressure`

Then call it from here, example:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/8dce09a4caa4258040384f68bbdcb77a22b0a55e/src/devices/tuya.ts#L352

rotdrop: Yes, true enough.

rotdrop:
Oops. The last commit has a bug in it ...

Koenkk:
Thanks!

rotdrop:
Wait! The tab char was not the problem, the comparison is broken!

Koenkk:
Whoops, reverted it, could you make a new pr?

rotdrop:
Yes, of course.

PR #7260, opened 2024-03-24
Try to fix unrealistic pressure readings from WSDCGQ12LM.

rotdrop:
I have squashed the commits. I think it should work now.


Koenkk:
Thanks!

PR #7267, opened 2024-03-26
Update tuya.ts

Jumaikhanova:
Added support to Beok TS0601 (TDR89B-ZB-EP)


src/devices/tuya.ts
```diff
@@ -4712,6 +4712,7 @@ const definitions: Definition[] = [
     },
     {
         fingerprint: tuya.fingerprint('TS0601', ['_TZE200_viy9ihs7', '_TZE204_lzriup1j']),
+        zigbeeModel: ['TS0601', 'ZWT198/ZWT100-BH'],
```
Koenkk, line 4715: Please add the manufacturer name to `fingerprint` instead.

PR #7269, opened 2024-03-26
fix: set MCLH-02 color temp range and convert LifeControl to modern extend 

mrskycriper:
- Set color temperature range for `MCLH-02` as per instruction to 167-333 mired (3000-6000K)
- Update device names in accordance to their actual retail names from the manufactures web store
- Convert all devices to Modern Extend
- Added `reporting` parameter to the `electricityMeter` extend to disable reporting if needed

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -1150,9 +1150,10 @@ export interface ElectricityMeterArgs {
     power?: false | MultiplierDivisor,
     voltage?: false | MultiplierDivisor,
     energy?: false | MultiplierDivisor
+    reporting?: boolean
 }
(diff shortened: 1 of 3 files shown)


src/lib/modernExtend.ts
```diff
@@ -1150,9 +1150,10 @@ export interface ElectricityMeterArgs {
     power?: false | MultiplierDivisor,
     voltage?: false | MultiplierDivisor,
     energy?: false | MultiplierDivisor
+    reporting?: boolean
```
Koenkk, line 1153:
```suggestion
    configureReporting?: boolean
```

To stay consistent with other modern extends

mrskycriper: Done!

PR #7272, opened 2024-03-26
Inovelli Fixing some non-working parameters

InovelliUSA:
Quick Start for light and fan not working. Removing high non neutral otupt from vzm35 (not available on that device).


Koenkk:
Thanks!

InovelliUSA:
@Koenkk Curious if these updates that we have submitted over last month will make it to production? Or what is the timeline?

Koenkk:
yes, will all go into the 1 April release

PR #7279, opened 2024-03-28
Moved EFEKTA devices to a separate file, refactored and added another one device

kirovilya:


Koenkk:
Thanks!

NDF-MDJ:
The SNZB-02_EFEKTA reports spikes in the temperature again. Was the filter removed?

PR #7281, opened 2024-03-29
Improve Danfoss Icon2 support

tomash345:
- Add support for the infrared floor sensor
- Add the ability to read humidity attribute manually
- Add model ID for the Advanced Main Controller
- Change reporting values of some attributes to default
- Slight clean-up


Koenkk:
Thanks!

PR #7286, opened 2024-03-31
Improve support for Bosch Twinguard

burmistrzak:
- fix: Add missing convertGet keys for bosch_twinguard
- feat: Improve bosch_twinguard_measurements converter
- feat: Add VOC sensor

Also includes the obligatory _fix typo_. 😉

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -797,7 +797,7 @@ const fzLocal = {
                 result.temperature = parseFloat(msg.data['temperature']) / 100.0;
             }
             if (msg.data.hasOwnProperty('illuminance_lux')) {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

daloki83:
Maybe thats the wrong place, but with Z2M 1.36.1 and the Twinguard 8750001213 the Device stopped working and only shows old values for Temperature, Humidity, Illuminance (and maybe some others). Factory reset, deleted in Z2M, repairing etc didn't solve it. Tried multiple times. Than i rolled back to 1.36.0-1 and everything was fine again after a few minutes. I would say something is wrong with that code change. If i can support in any way, please let me know.

PR #7287, opened 2024-03-31
Update philips.ts for black Centris 4 spot

thk-socal:
Added support for 929003621301 Hue Centris 4 spot black

### DIFF
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -1971,6 +1971,13 @@ const definitions: Definition[] = [
         extend: [philipsLight({colorTemp: {range: [153, 500]}, color: true})],
     },
     {
+        zigbeeModel: ['929003621301_01', '929003621301_02', '929003621301_03', '929003621301_04', '929003621301_05'],
+        model: '929003621301',
+        vendor: 'Philips',
+        description: 'Hue White & Color ambience Centris ceiling light (4 spots) (black)',
+        extend: [philipsLight({colorTemp: {range: [153, 500]}, color: true})],
+    },
+    {
         zigbeeModel: ['5060830P7_01', '5060830P7_02', '5060830P7_03', '5060830P7_04'],
         model: '5060830P7',
         vendor: 'Philips',


Koenkk:
Thanks!

Koenkk:
@thk-socal I believe you were having issues with the generated external converter? Could you provide it here so I can debug it? ([generated external converter](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_2-creating-the-external-definition))

thk-socal:
@Koenkk was not an external converter at all. When this light is either joined, or part of the zigbee network, and I am running the docker latest-dev image, Z2M crashes with the following error:

```
Zigbee2MQTT:error 2024-03-30 14:28:30: Exiting...
Zigbee2MQTT:error 2024-03-30 14:28:30: AssertionError [ERR_ASSERTION]: '929003621301_01' has multiple 'ota', this is not allowed
    at processExtensions (/app/node_modules/zigbee-herdsman-converters/src/index.ts:124:28)
    at prepareDefinition (/app/node_modules/zigbee-herdsman-converters/src/index.ts:173:18)
    at findDefinition (/app/node_modules/zigbee-herdsman-converters/src/index.ts:285:16)
    at Object.findByDevice (/app/node_modules/zigbee-herdsman-converters/src/index.ts:257:22)
    at Device.resolveDefinition (/app/lib/model/device.ts:36:31)
    at Zigbee.start (/app/lib/zigbee.ts:69:13)
    at Controller.start (/app/lib/controller.ts:109:27)
    at start (/app/index.js:107:5)
```

When I put in :latest, then it is correctly identified as unsupported, but does NOT crash Z2M. Actually it is worse that crashing, Z2M:latest-dev will not even start. So although the error is likely valid, I think Z2M should continue to function when this happens and just disable that device.

I am hoping by adding the necessary code to philips.ts, that the light will now function and not crash Z2M.

Koenkk:
@thk-socal pushed the fix!

PR #7288, opened 2024-03-31
feat: add support for Legrand 067761A

Diaoul:
# Add support for Legrand 067761A
* Device is described [here](https://www.legrand.fr/pro/catalogue/commande-sans-fils-pour-vmc-connectee-celiane-with-netatmo-blanc-sans-plaque)
* Documentation PR: https://github.com/Koenkk/zigbee2mqtt.io/pull/2664

# Questions
1. Currently this devices is described as "on/off" and I imagine this is enough but maybe there is something more semantically close to the buttons? e.g. fan high speed and fan low speed
2. The device also supports `genLevelCtrl` but in the context of this particular switch I think it's not supposed to work. Anyways, if I want to add support for it, what should I do? The closest I found is `brightness_move_up` which is very far from what this is supposed to do :sweat_smile: 
3. How do I add support for OTA? It should be the same as the `Wireless remote switch` from Legrand as well as it's on the NLT device list from [here](https://developer.legrand.com/production-firmware-download/)

### DIFF
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -659,6 +659,23 @@ const definitions: Definition[] = [
             await reporting.onOff(endpoint);
         },
     },
+    {
+        zigbeeModel: [' Remote fan controller\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000'],
+        model: '067761A',
+        vendor: 'Legrand',
+        description: 'Exhaust wireless switch',
+        fromZigbee: [fz.identify, fz.command_on, fz.command_off, fz.battery],
+        toZigbee: [],
+        exposes: [
+            e.battery(),
+            e.action(['identify', 'on', 'off']),
+        ],
+        onEvent: readInitialBatteryState,
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genPowerCfg', 'genOnOff']);
+        },
+    },
 ];
 
 export default definitions;


Koenkk:
Could you try with the [generated external converter](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_2-creating-the-external-definition) instead? Try this with the 1 April release tomorrow.

Diaoul:
What is the difference supposed to be? Would that be better? I tried to stick to the way Legrand devices work.
IIRC I had an array `extend` with a few things inside, let me check

mrskycriper:
> It does not seem that it detected anything really useful :shrug: 
> 
> _Side note: The weird characters are not escaped correctly_
> 
> 
> 
> ```js
> 
> const {batteryPercentage, identify} = require('zigbee-herdsman-converters/lib/modernExtend');
> 
> 
> 
> const definition = {
> 
>     zigbeeModel: [' Remote fan controller'],
> 
>     model: ' Remote fan controller',
> 
>     vendor: ' Legrand',
> 
>     description: 'Automatically generated definition',
> 
>     extend: [batteryPercentage(), identify()],
> 
>     meta: {},
> 
> };
> 
> 
> 
> module.exports = definition
> 
> ```

I've recently added necessary modern extends and they will be in release version soon. It is strongly advised to use them instead of older split style converters.

Diaoul:
Sure :+1:

FabianMangold:
Hi all,
Sorry to jump in here , just wanted to get some more context about this and check on how to proceed in future.
(I know this may not be the right place for such questions, thus happy to follow it elsewhere!).

While the advantages of ModernExtends are obvious, what are the intentions on the long run ?
 Provide "initial" support, or replace split definitions altogether ?
 
 Legrand devices are somewhat different, in that they do not "always" follow standards. To address this, we've implemented those quirks in vendor specific libs.
 How should we address this in future ?
 Wouldn't a vendor specific template, that could be extended a slightly more flexible approach ? 

Thanks a lot for your insights.

Diaoul:
Generated code with latest:
```js
const {identify, battery, commandsOnOff, commandsLevelCtrl} = require('zigbee-herdsman-converters/lib/modernExtend');

const definition = {
    zigbeeModel: [' Remote fan controller'],
    model: ' Remote fan controller',
    vendor: ' Legrand',
    extend: [identify(), battery(), commandsOnOff(), commandsLevelCtrl()],
    meta: {},
};

module.exports = definition;
```

Is that better?

Koenkk:
@Diaoul if it works, please update the PR with it.

PR #7311, opened 2024-04-02
Fix `args` not used & casing for link quality modern extend.

Nerivec:
Casing change to match other functions. _Has no impact on codebase since this function is not yet used._


mrskycriper:
It's used in my IKEA PR, but I agree with the name changes

Koenkk:
Thanks!

Issue #7274, opened 2024-03-27
Bosch Twinguard exposes only CO2 instead of VOC levels


burmistrzak:
Hi folks,
I've been preparing to finally migrate a handful of Bosch Twinguards to Z2M and saw that the current implementation exposes the VOC sensor readings as CO2 (ppm) instead of VOC (µg/m³).

While @safakaltun did absolutely stellar work, reverse engineering Bosch's ZCL extension, their choice to expose the VOC readings as CO2-equivalent (?) doesn't seem quite right to me.

- First, the [mentioned](https://github.com/Koenkk/zigbee2mqtt/issues/14355#issuecomment-1339677829) [data sheet](https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bme680-ds001.pdf) for the [BME680 sensor](https://www.bosch-sensortec.com/products/environmental-sensors/gas-sensors/bme680) explicitly states that the sensor _cannot_ directly measure CO2 levels (p. 25), instead having to estimate them.
- Second, it's a bit unclear, to me at least, how @safakaltun came up with that [ppm conversion](https://github.com/Koenkk/zigbee-herdsman-converters/blob/8a3f6442f4dcd88ff7ff26bf5c1669125c43a27c/src/devices/bosch.ts#L786) formula. I guess they tried matching the value from the Bosch Smart Home app? 🤔
- And third, exposing only a CO2 (and no VOC) sensor in Z2M, has some unintended consequences for downstream systems like homebridge-z2m. Namely e.g. a missing VOC sensor in HomeKit, something the native integration from Bosch (via their Smart Home Controller) already supports.

Now, what to do about this situation?
Well, simply exposing the raw AQI value isn't super useful because, among other things, there are **many different, national AQI scales** out there. AFAICT, the Twinguard uses the German AQI internally.

@Koenkk What do you think about exposing an additional VOC level, so to not break backwards compatibility while adding to the overall usefulness of the device? 👀

Koenkk:
Im fine with that!

safakaltun:
Hi, “alarm” is an action. Doesn’t have a state until you trigger it. Once you do that, you will see “siren state” changing accordingly.

safakaltun:
[...]

PR #7313, opened 2024-04-02
Change AQI scale for Bosch 8750001213 Twinguard 

burmistrzak:
As it turns out, the IAQ (indoor air quality) index used by the BME680 gas sensor, is not linear. Great... 👀

The first implementation #7286 assumed a linear relationship between IAQ level and TVOC density, and that obviously [wasn't correct](https://github.com/Koenkk/zigbee-herdsman-converters/issues/7274#issuecomment-2030138081). This is an attempt to improve the overall accuracy of the conversion, and align the estimated TVOC density to its corresponding IAQ level.

We'll have to keep in mind, that we're dealing only with estimates here, and **not** with lab-grade sensor readings.

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -790,6 +790,8 @@ const fzLocal = {
                 }
             }
             if (msg.data.hasOwnProperty('airpurity')) {
+                const iaq = parseInt(msg.data['airpurity'];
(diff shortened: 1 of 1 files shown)


Koenkk:
Let me know once this is ready for merge.

PR #7322, opened 2024-04-03
Added BTicino FC80GCS and 4411C

xraver:
Hello
I've added BTicino FC80GCS (legrand 412172) and dimmer switch 4411C in the description.


Koenkk:
Thanks!

PR #7328, opened 2024-04-03
feat(Sber): separate devices and add SBDV-00079

mrskycriper:
## TODO
- [x] add `tuyaOnOffAction` and `tuyaOnOffActionLegacy` modern extends
- [x] Move Sber devices to `sber.ts`
- [x] Add support for Smart temperature and humidity sensor (SBDV-00079)
- [x] Update images

Images https://github.com/Koenkk/zigbee2mqtt.io/pull/2679


src/lib/modernExtend.ts
```diff
@@ -1464,11 +1465,17 @@ export function actionEnumLookup(args: ActionEnumLookupArgs): ModernExtend {
             if (attributeKey in msg.data) {
                 let value = getFromLookupByValue(msg.data[attributeKey], lookup);
                 // endpointNames is used when action endpoint names don't overlap with other endpoint names
-                if (args.endpointNames) value = postfixWithEndpointName(value, msg, model, meta);
+                if (args.endpointNames) {
+                    value = addEndpointName(value, msg, model, meta, args.actionTemplate === 'action_endpoint' ? 'postfix' : 'prefix');
```
Koenkk, line 1469: For consistency purposes, I would always postfix it

mrskycriper: But what to to with existing backwards implementations then?

Koenkk: It's only used by `tuyaOnOffAction` -> `SBDV-00032` (new device), so we don't have to consider this yet?

mrskycriper: > It's only used by `tuyaOnOffAction` -> `SBDV-00032` (new device), so we don't have to consider this yet?

All other similar Tuya based buttons have this structure

mrskycriper: @Koenkk tuyaOnOffAction is really just modern version of tuyaFz.on_off_action

Should I just revert all changes with postfix/prefix and use old tuyaFz.on_off_action instead?

Koenkk: Could we have a `tuyaOnOffAction` and `tuyaLegacyOnOffAction`?

mrskycriper: > Could we have a `tuyaOnOffAction` and `tuyaLegacyOnOffAction`?

Sounds reasonable

mrskycriper: Done

src/devices/sber.ts
```diff
@@ -0,0 +1,60 @@
...
+        ],
+    },
+    {
+        fingerprint: [{modelID: 'TS0041A', manufacturerName: '_TYZB01_ub7urdza'}],
+        model: 'SBDV-00032',
+        vendor: 'Sber',
+        description: 'Smart button',
+        extend: [
+            tuyaMagicPacket(),
+            tuyaOnOffActionLegacy({actions: ['single', 'double', 'hold']}),
```
Koenkk, line 35: Since this is a new device, why not use `tuyaOnOffAction`?

mrskycriper: It's not new, I pulled it out of tuya whitelabels list.

Koenkk:
Thanks!

PR #6862, opened 2024-01-08
feat: bosch smoke intruder and smoke alarm triggers

vjonas:
I ran into formatting issues. It logs too many changes. Not sure how to align that, how should I do that? @Koenkk
Potentially we could enable some formatter as Prettier to the codebase?
This should allow users to send a mqtt topic to toggle the alarm and intruder state of the smoke detector II.

Add these 2 commands to trigger the intruder and alarm state:
image

PR #7329, opened 2024-04-04
Fix alarm states for Bosch BSD-2

burmistrzak:
Based on [nice work](https://github.com/Koenkk/zigbee2mqtt/issues/21351) by @N3rdix, all credits to them for the contents of the first two commits! 🙏

- Adds the ability to properly `/get` and `/set` the current state both alarms, i.e. intrusion & smoke.

~~🚧 Further improvements under investigation...~~


N3rdix:
LGTM, the only thing I was not sure why it was there is the ` device.save ` in the configure. 
At least it works on my side without it 😀 

Thanks for saving me some time by creating this PR 👍

Koenkk:
`device.save` is not needed int his case

N3rdix:
Thanks for the quick implementation @burmistrzak 👍  Yes it is possible to send both commands, although only one alarm is ringing they are queued (see also https://github.com/Koenkk/zigbee-herdsman-converters/pull/6862#discussion_r1451455918).

That's the reason why I added that case explicitly to reflect correct switch states. Should be good to proceed.

Koenkk:
@burmistrzak only when doing mutations directly on the device object like here:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/e8d4e77a6cb1f26a2363957e4d62df77b63b9ddb/src/devices/dawon_dns.ts#L87

vjonas:
> @vjonas 👋 Do you remember if there was a specific reason for [unbinding](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3141809acc05722032eb14be81925f5c7fd1ff3a/src/devices/bosch.ts#L1022) the `genPollCtrl` cluster?

I do not remember properly, sorry. 
Thank you for the extra polishing work guys!

N3rdix:
> @N3rdix Were you ever able to silence a smoke alarm (that you manually triggered via Z2M), on the BSD-2 itself?
> 
> According to Bosch, that _should_ work... 🤔
> 
> >What happens in case of a Smoke alarm when I press a button of another networked Smoke alarm?
> 
> In this case, the alarms will stop at those networked Smoke alarms to which the alarm was forwarded. Only the alarm at the triggering Smoke alarm remains active.
> 
> 
> 
> 
> 
> Btw. our patch is working flawlessly! 🙌

Good question, i never tried that. I will give it a try the next days.
Did you test short- and long-press?

N3rdix:
that sounds reasonable, I was struggling with the bitmask coding as I am not a native TS/JS developer and therefore went (first) for the current implementation style.
But it sound you know what to do, I am happy to help and test. Have now 2 BSDs to also test some more complex cases 😃 

Hopefully I'll find some time next week to play around with the attributeReport/readResponse and the bitmask

N3rdix:
Nice findings, appreciate your work 👍 
With this information we should be able to extend the fromZigbee accordingly 😀

safakaltun:
Hi, they should be correct. I didn’t try sending commands myself, but those were the outcome of the sniffs. The implementation @vjonas made was working fine for me starting/stopping the fire and intrusion alarms externally. 

What I remember from the sniffs is that the controller was broadcasting the same command to the entire network. I don’t recall the devices replying back with a status after doing whatever was requested. I am not sure if what you are trying to do is possible without Bosch updating the firmware of these devices. 

By the way, I had a look at the table you created. There shouldn’t be any tamper protection in these devices.

safakaltun:
Interesting. Remotely triggered intrusion and smoke make different sounds, but based on this it is the same command sent. Weird. 

I might have the sniffs, but I am traveling at the moment and won’t be able to check that for some time. I don’t think I sniffed the part where pressing a button in one stops all sirens anyway.

safakaltun:
Is command “0x80” already defined in the iaszone cluster?

safakaltun:
> Was there a special procedure to get the SHC Zigbee Keys? 

I either used a bosch smart outdoor siren or a twinguard to get the install code. I don’t remember which one it was but I am sure I noted it in the github issue of one of them. :)

safakaltun:
> > Is command “0x80” already defined in the iaszone cluster?
> 
> Yep, [it is](https://github.com/Koenkk/zigbee-herdsman/blob/caaf520e31b816155aa45180929c47bbc0bf94af/src/zcl/definition/cluster.ts#L2731). 👀

Did you try it with a different data type instead of uint16? 

Also, not sure if it will make a difference but did you try using ‘ instead of “ for **data**?

 `parameters: [{name: "data", type: DataType.uint16}]`

safakaltun:
Maybe try what’s below. It is from one of the manuspecificbosch clusters I had created. This one was specific to the intruder alarm. 

`{name: 'data', type: DataType.uint8}`


I found the status notification sniffs. Can’t you connect these to the trigger buttons for the sirens? When the new status goes to clear, your controller can send a command to stop the sirens.


https://github.com/Koenkk/zigbee2mqtt/issues/17229#issuecomment-1824402822

vjonas:
> trying

Hey guys, I got them from sniffing the traffic locally. I used a bosch smart home controller. Even more weird I got different results.

vjonas:
> sniffer

It was indeed a pain in the ass to get through the sniffing process. But should be "documented" there.

PR #7336, opened 2024-04-04
transfer zigbeeCommandOptions into setupAttributes from enumLookup

grebenyuk51:
fixed issue when zigbeeCommandOptions not transferred from enumLookup to read attribute method
in this case modernExtend does not work with manufacturer specific attributes
unfortunately I don't know how to modify unit-test :(

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -59,7 +59,7 @@ function convertReportingConfigTime(time: ReportingConfigTime): number {
 
 export async function setupAttributes(
     entity: Zh.Device | Zh.Endpoint, coordinatorEndpoint: Zh.Endpoint, cluster: string | number, config: ReportingConfig[],
-    configureReporting: boolean=true, read: boolean=true,
+    configureReporting: boolean=true, read: boolean=true, options?: {manufacturerCode?: number, disableDefaultResponse?: boolean},
 ) {
     const endpoints = isEndpoint(entity) ? [entity] : getEndpointsWithCluster(entity, cluster, 'input');
     const ieeeAddr = isEndpoint(entity) ? entity.deviceIeeeAddress : entity.ieeeAddr;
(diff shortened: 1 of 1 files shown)


sjorge:
It should in theory figure out it needs the manufacturerCode from the cluster definition. 

What cluster.attribute is it failing for is it failing for?

grebenyuk51:
@sjorge 
here is my code:
```
modernExtend.enumLookup({
                                name: 'switch_mode',
                                lookup: {
                                    'toggle': 0,
                                    'alternate': 2,
                                },
                                endpointName: 'button_1',
                                cluster: 0xfcb0,
                                attribute: {
                                    ID: 0x0a00,
                                    type: 0x30,
                                },
                                description: 'Operation method of the switch',
                                access: 'ALL',
                                entityCategory: 'config',
                                zigbeeCommandOptions: {
                                    manufacturerCode: 0x4c44,
                                },
                            })
```


in [modernExtend.ts](https://github.com/Koenkk/zigbee-herdsman-converters/blob/8257c320d5c53703570a917fb2b1b738a566e285/src/lib/modernExtend.ts#L1308) calls method 
 without `options` param, than in [setupConfigureForReporting](https://github.com/Koenkk/zigbee-herdsman-converters/blob/8257c320d5c53703570a917fb2b1b738a566e285/src/lib/modernExtend.ts#L108) calls setupAttributes without optins param, and finally calls [endpoint.read](https://github.com/Koenkk/zigbee-herdsman-converters/blob/8257c320d5c53703570a917fb2b1b738a566e285/src/lib/modernExtend.ts#L84) without transferring `manufacturerCode`

sjorge:
Ah you are using a cluster/attribute not yet in zh's cluster.ts

Adding it there would have made zh pick the right manufacturerCode.

@Koenkk having per device cluster would also fix this, but it might still be useful for debugging/while adding support for a new device. On the other hand it does add more code to modernExtend to maintain.

I feel once we have per device cluster definitions we could probably do away with handing of numeric cluster/attrib/dataType stuff in modernExtend and make it a bit simpler/more readable.

grebenyuk51:
> Ah you are using a cluster/attribute not yet in zh's cluster.ts
yeah, and, as far as I understand the code, cannot add it using external-converter. Right now I cannot publish my device (working on it) thus cannot add it to z2m

Koenkk:
It seems it's better to wait until custom cluster device definitions are supported, I already started working on this in https://github.com/Koenkk/zigbee-herdsman/pull/1011

grebenyuk51:
sure, be in touch

PR #7340, opened 2024-04-05
add support for Ekaza EKAC-T3095Z

karlprieb:
This is a TS110E white label device sold in Brazil. It works similar to Lonsonho ones with the exception that these don't support light type. https://store.ekaza.com.br/interruptort3095z/p


src/devices/tuya.ts
```diff
@@ -5852,6 +5852,27 @@ const definitions: Definition[] = [
             await reporting.onOff(endpoint);
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS110E', ['_TZ3210_ysfo0wla']),
+        model: 'TS110E_1gang_3',
```
Koenkk, line 5857:
```suggestion
        model: 'EKAC-T3095Z',
```

src/devices/tuya.ts
```diff
@@ -5852,6 +5852,27 @@ const definitions: Definition[] = [
...
+        vendor: 'TuYa',
```
Koenkk, line 5858:
```suggestion
        vendor: 'Ekaza',
```

src/devices/tuya.ts
```diff
@@ -5852,6 +5852,27 @@ const definitions: Definition[] = [
...
+        description: '1 channel dimmer',
+        whiteLabel: [
```
Koenkk, line 5860: Remove `whiteLabel` here.

Koenkk:
Thanks!

> Not sure why we should handle this as a Ekaza vendor instead of TuYa. Can you explain? Thank you!

Because that's what you added in the whiteLabel, and a definition with 1 wihteLabel does not make any sense, therefore I asked to merge it.

PR #7343, opened 2024-04-06
hotfix(ikea): update `genLevelCtrl` target cluster type

mrskycriper:
- Fix `genLevelCtrl` target cluster type to output
- Remove redundant redirects in E1744 legacy extends

Hopefully this will finally fix https://github.com/Koenkk/zigbee2mqtt/issues/22084


Koenkk:
Thanks!

PR #7346, opened 2024-04-06
fix: split tuya fan devices for different fan ctrl scale

rodrigobutta:
Adding a device that it´s almost identical to [TuYa TS0601_fan_and_light_switch](https://www.zigbee2mqtt.io/devices/TS0601_fan_and_light_switch.html) but the `FAN LEVEL` is handled by 3 buttons (low, mod, high), while the device that I have, has 2 buttons, with a 7-diode display that handles 5 levels.

I´ve splitted those two models, and change the fan level for an enum that was already existed for other tuya fan device.

### The device´s image:
https://ae01.alicdn.com/kf/Aba9acc25999340518a5346d45681fe1ap.jpg

### Note
I've setted the `fan_speed` 0 array position to `101` because the original one was that, but it could be `3` as the (no 5 level) has. Not sure what it means.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6012,7 +6012,7 @@ const definitions: Definition[] = [
         ],
     },
     {
-        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_hmqzfqml', '_TZE200_lawxy9e2']),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #7349, opened 2024-04-06
fix: TS0505B missrepresentation as WZ5_dim_2

java-devil:
In response to:
https://github.com/Koenkk/zigbee2mqtt/issues/22095


Koenkk:
Did you test if this works?

java-devil:
Sadly I thought I left this PR in draft mode...

Are you able to refer me to a doc/post on how to best replace Herdsman with my own build on a running HAOS for testing purposes? I'm still new to this entire ecosystem I'm afraid.

Alternatively... If the procedure is involved/undocumented... As amateurish as it sounds... Could we perhaps get this merged - so that I can test the edge version - and revert if I don't confirm it works properly in ~24 hours?

Koenkk:
You can test this by using an external definition, see the z2m docs how to support new devices page on how to do that

java-devil:
I did check those docs and related issues.

However if you were to check the linked issue:
https://github.com/Koenkk/zigbee2mqtt/issues/22095

And your own snippet from the past:

java-devil:
I **did** prepare an external definition after checking those docs and related issues, before submitting this PR :)

However if you were to check the linked issue:
https://github.com/Koenkk/zigbee2mqtt/issues/22095

And your own snippet from the past:
https://github.com/Koenkk/zigbee2mqtt/issues/13466#issuecomment-1209662389

You would see this isn't so simple / possible anymore / I don't know what I'm doing (very probable):
- in my existing external definition I have the line: ```extend: [light({color: {modes: ["hs"], applyRedFix: true}})]```
- however what I seem to be missing is {"preferHS": "true"} which seems to not be supported anymore by `modernExtend`
- As it currently happens - I'm able to set the color via an HS color picker in Z2A, but it still ends up being send as XY color values to the device xD

In other words - am I able to impose the below "properties" onto my device using an external definition and modern extends?
```   fromZigbee: [legacy.fz.tuya_light_wz5],
        toZigbee: [legacy.tz.tuya_dimmer_state, legacy.tz.tuya_light_wz5],
        exposes: [
            e.light().withBrightness().setAccess('state', ea.STATE_SET).setAccess('brightness',
                ea.STATE_SET).withColor(['hs']).setAccess('color_hs', ea.STATE_SET),
        ],
```

Koenkk:
Can you post your external converter which works fine here? `preferHS` == `modes: ["hs"]`

java-devil:
If indeed `preferHS` == `modes: ["hs"]` I would expect the below snippet to result in HS formatted MQTT requests (in the form of `{"color":{"hue":360,"saturation":100}}`):

```
const definition = {
  fingerprint: [{modelID: 'TS0505B', manufacturerName: '_TZB210_3zfp8mki'}],
  model: 'TS0505B',
  vendor: 'TuYa',
  description: 'Zigbee RGB+CCT Light',
  extend: [light({color: {modes: ["hs"], applyRedFix: true, enhancedHue: false}})],
};
```

Is my reasoning correct?

While in reality in debug mode of Z2M I see (**even when using the HS Color picker from the "Exposes" tab that wasn't present before adding the external definition**):
```Received MQTT message on 'zigbee2mqtt/L01/set' with data '{"color":{"hex":"#ff0004"}}'```

And yes - this results in extremely de-saturated colors as expected.

All the while the state of my device clearly contains: `"color_mode": "xy",`

However if I **do** manually send a properly formatted color as expected by https://www.zigbee2mqtt.io/devices/WZ5_rgbcct.html (which seem to think that they are WZ5_dim_2s) the color_mode swaps to "hs" and all is well until I decide to modify the color once more in HA/Z2M.

**Comment by java-devil** [comment]
I'm afraid I decked out my ENTIRE home with those lights :P

I'm willing to go the extra mile, but I fall short in the technical specifics of this (amazing) ecosystem and unfortunately require assistance :/

**Comment by Koenkk** [comment]
What do you see in the exposes tab? I would expect something like this:

**Comment by java-devil** [comment]
When the device is recognized as WZ5_dim_2 (oddly enough I don't see such a device anywhere in this repository - only WZ5_dim_1) I **only** see the upper color picker.

When I apply the below external definition I **only** see the lower color picker (which I guess is expected since I only specified `modes: ["hs"]` versus `modes: ["hs", "xy"]`):

```
const definition = {
  fingerprint: [{modelID: 'TS0505B', manufacturerName: '_TZB210_3zfp8mki'}],
  model: 'TS0505B',
  vendor: 'TuYa',
  description: 'Zigbee RGB+CCT Light',
  extend: [light({color: {modes: ["hs"], applyRedFix: true, enhancedHue: false}})],
};
```

The really meaningful bit from my perspective though is that **both** result in the same form of message send over MQTT:
` Received MQTT message on 'zigbee2mqtt/L01/set' with data '{"color":{"hex":"#ff0004"}}'`

Which would imply that both simply have to behave the same desaturated way (which they do).

What I would however expect is that using lower color picker (post applying the external definition) would result in a message containing... well... hue & saturation ;)

Koenkk:
This is an issue of the frontend, for the HS color picker it should send HS values, not `hex`. Could you open an issue here: https://github.com/nurikk/zigbee2mqtt-frontend/issues ?

Note that this only applies to the frontend, if you use e.g. HA correct values should be send.

java-devil:
I could open an issue no problem... however... when using the "Light" "Service" in HA with the above external definition I still observe exactly the same behavior as when using the Z2M frontend... Meaning this is probably an issue with `modernExtend` after all?

Koenkk:
- How do you call the light service?
- Can you provide the debug log when setting the color via the HA color picker?

java-devil:
After additional testing I am happy to conclude that **you were indeed correct kind Sir, and I was fortunately mistaken.**

This IS indeed a Z2M web UI issue. And After applying the external definition from https://github.com/Koenkk/zigbee-herdsman-converters/pull/7349#issuecomment-2052139922 I **AM** getting vibrant colors while setting them in HA!

I will ofc open an appropriate issue in https://github.com/nurikk/zigbee2mqtt-frontend/issues

However our work here isn't unfortunately entirely done. The RGB bit of my WZ5 RGB+CCT now works correctly after applying the above external definition. However in consequence I did completely lose the ability to set the CCT color ;) I assume pushing me in the correct direction should be relatively obvious by now?

Regardless. **This is an amazingly well maintained piece of OSS software.** How can I best contribute to it financially?

Koenkk:
Use `extend: [light({colorTemp: {range: undefined}, color: {modes: ["hs"], applyRedFix: true, enhancedHue: false}})],` 

> How can I best contribute to it financially?

Many thanks! You can do so by clicking the sponsor button [here](https://github.com/koenkk/zigbee2mqtt)

java-devil:
I'm happy to confirm that this works.

Minor issue to potentially still take care of - my state now contains:
`"white_brightness": 254`

Which seems to be completely ignored (it's always fixed at 254) and regular `brightness` seems to be preferred even in CCT mode. I don't know if this is a configuration problem, but it sounds as it might be indicative of one?

Other than that - how do we get this external definition merged for others to benefit? :)

Koenkk:
Can you stop z2m, remove `white_brightness` from `data/state.json` and see if the issue is fixed after that?

java-devil:
Yes. That did the trick. Probably some leftover from my experimentations with external definitions in that case.

I wrote the issue to the fronted repository that you requested @Koenkk

Is there anything left I can do now to help in getting this external definition merged in a more maintainable form for other users to benefit?

Koenkk:
Thanks!

PR #7359, opened 2024-04-07
Hotfix for missing `ssIasZone` cluster on Bosch BWA-1

burmistrzak:
Includes the following, important changes:

- Always bind `ssIasZone` to BWA-1 (without, no alarm... Apparently. 🥸)
- Set `reportableChange` to `0` instead of `null`


Koenkk:
thanks!

PR #7363, opened 2024-04-08
Update ihorn.ts

cloudyngcloudy:
Rectify the code


Koenkk:
What's not working with the extend code? (this is the new way of doing this)

cloudyngcloudy:
There is no exposes discovered under the device.

在 2024年4月9日週二 上午2:49，Koen Kanters ***@***.***> 寫道：

Koenkk:
Are you using z2m 1.36.1?

mrskycriper:
Those modern extends were tested and are definitely working as intended. I personally use some devices that were used to create those extends. They are functionality equal to the older style. 

Are there any errors in the logs?

PR #7372, opened 2024-04-08
Hotfix(es) for Bosch BSD-2 & BWA-1

burmistrzak:
Hotfix(es) 🔥 for Bosch BSD-2 & BWA-1, includes:

- Remove `manufacturerOptions` from `ssIasZone` cluster, otherwise Bosch BSD-2 won't configure during setup
- Update attribute reporting for BWA-1, read initial attribute values during configuration


Koenkk:
Thanks!

PR #7374, opened 2024-04-09
fix(configure): use endpoint ids for filtering instead of friendly names (help needed)

mrskycriper:
During the testing of IKEA refactoring it was noted that `endpointNames` were not implemented properly. Right now the filter assumes that friendly names are equal to their numeric id, which is only partially true.

But, there is a problem. Currently there is no way to access `endpoint` object from `configure`. And it seems to be the only way to read those friendly names.

I'm proposing to add `definition` as the 3rd argument for `configure`. But I cant seem to find, where is `configure` being called inside Zigbee2MQTT so I could also add this argument there.

Also partial fix https://github.com/Koenkk/zigbee2mqtt/issues/22084 (the 4th one).


src/index.ts
```diff
@@ -159,9 +159,9 @@ function processExtensions(definition: Definition): Definition {
 
         let configure: Configure = null;
         if (configures.length !== 0) {
-            configure = async (device, coordinatorEndpoint) => {
+            configure = async (device, coordinatorEndpoint, configure_definition) => {
```
Koenkk, line 162:
```suggestion
            configure = async (device, coordinatorEndpoint, configureDefinition) => {
```

Koenkk:
>  Right now the filter assumes that friendly names are equal to their numeric id, which is only partially true.

Could you give me an example of this problem? I don't fully understand yet what this PR fixes.

Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt/pull/22154

mrskycriper:
> > Right now the filter assumes that friendly names are equal to their numeric id, which is only partially true.
> 
> Could you give me an example of this problem? I don't fully understand yet what this PR fixes.
> 
> Note to self; after this merge [Koenkk/zigbee2mqtt#22154](https://github.com/Koenkk/zigbee2mqtt/pull/22154)

Currently it compares endpoint names with endpoint ids when filtering. Which is just wrong. This was discovered because IKEA SYMFONISK remote has endpoints 1, 2 and 3. But uses multiendpoint only for 2 and 3 under the names of 1 and 2. And upon executing configure step, it tries to use endpoints with id 1 and 2 because of those friendly names.

src/lib/modernExtend.ts
```diff
@@ -116,9 +121,14 @@ export function setupConfigureForReporting(
 }
 
 export function setupConfigureForBinding(cluster: string | number, clusterType: 'input' | 'output', endpointNames?: string[]) {
-    const configure: Configure = async (device, coordinatorEndpoint) => {
+    const configure: Configure = async (device, coordinatorEndpoint, definition) => {
         if (endpointNames) {
-            const endpointsMap = new Map<string, boolean>(endpointNames.map((e) => [e, true]));
+            const definitionEndpoints = definition.endpoint(device);
+            const endpointIds: string[] = [];
```
Koenkk, line 127: `const endpointIds = endpointNames.map((e) => definitionsEndpoints[e])`

Would this work?

src/lib/modernExtend.ts
```diff
@@ -116,10 +118,12 @@ export function setupConfigureForReporting(
...
 export function setupConfigureForBinding(cluster: string | number, clusterType: 'input' | 'output', endpointNames?: string[]) {
-    const configure: Configure = async (device, coordinatorEndpoint) => {
+    const configure: Configure = async (device, coordinatorEndpoint, definition) => {
         if (endpointNames) {
-            const endpointsMap = new Map<string, boolean>(endpointNames.map((e) => [e, true]));
-            const endpoints = device.endpoints.filter((e) => endpointsMap.has(e.ID.toString()));
+            const definitionEndpoints = definition.endpoint(device);
+            const endpointIds = endpointNames.map((e) => definitionEndpoints[e]);
+            const endpointsMap = new Map<number, boolean>(endpointIds.map((e) => [e, true]));
+            const endpoints = device.endpoints.filter((e) => endpointsMap.has(e.ID));
```
Koenkk, line 126:
```suggestion
            const endpoints = device.endpoints.filter((e) => endpointIds.includes(e.ID));
```

And then you can remove `const endpointsMap = new Map<number, boolean>(endpointIds.map((e) => [e, true]));`

mrskycriper: Thanks!

Koenkk:
Thanks!

PR #7376, opened 2024-04-09
fix(legacy): use legacyAction in modern extends to avoid problems from converter duplication

mrskycriper:
This PR should prevent any problems from happening when legacy click was present next to modern extend.

Also partial fix https://github.com/Koenkk/zigbee2mqtt/issues/22084 (yes, the 3rd one).

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -394,10 +394,10 @@ export function onOff(args?: OnOffArgs): ModernExtend {
 }
 
 export interface CommandsOnOffArgs {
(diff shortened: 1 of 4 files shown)


src/lib/modernExtend.ts
```diff
@@ -779,6 +787,48 @@ export function commandsLevelCtrl(args?: CommandsLevelCtrl): ModernExtend {
         fz.command_stop,
     ];
 
+    if (args.legacyClick) {
```
Koenkk, line 790:
```suggestion
    if (args.legacyAction) {
```

Isn't this a better name?

Koenkk:
Thanks!

PR #7378, opened 2024-04-09
Fix `alarm_on_motion` option for Bosch BWA-1

burmistrzak:
Changes the payload structure for `write` operations because we're now using a defined cluster and attribute.

The previous payload format does not work with named clusters/attributes, meaning you can disable the motion alarm, but cannot enable it again.


Koenkk:
thanks!

PR #7379, opened 2024-04-10
FIX: Bosch BTH-RA

Burki24:
display_ontime didn't run correct. Expose would be found two times in payload with different name (display_ontime vs. display_onTime). And only display_onTime was reacting on changes made via frontend. After fix, the right expose "display_ontime" recognize the changes.

```
{
"battery":100,
"boost":"OFF",
"child_lock":"UNLOCK",
"display_brightness":9,

"display_onTime":28,
"display_ontime":24,

"display_orientation":"normal",
"displayed_temperature":"measured",
"last_seen":1712716480664,
"linkquality":168
...
```


Burki24:
Not needed, sorry.

Koenkk:
Thanks!

PR #7380, opened 2024-04-10
Update paulmann.ts with Smart switch 4 (Model: 501.37)

thomasm54:


src/devices/paulmann.ts
```diff
@@ -6,6 +6,17 @@ import {light, onOff} from '../lib/modernExtend';
 const e = exposes.presets;
 
 const definitions: Definition[] = [
+    {
+        zigbeeModel: ['501.37'],
+        model: '501.37',
+        vendor: 'Paulmann',
+        description: 'Smart switch 4 buttons white',
+        fromZigbee: [fz.command_on, fz.command_off, fz.battery, fz.command_move, fz.command_stop],
```
Koenkk, line 14: Can you convert this to modern extend?

Example:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/a0a1bff6e4b39d30fcb1a516dcefc791a0900dc9/src/devices/ikea.ts#L780

Koenkk: I see now that it will not fully work, so will merge this, thanks!

thomasm54: ok, thanks!

PR #7383, opened 2024-04-11
Minor housekeeping for Bosch 8750001213 Twinguard 

burmistrzak:
Includes the following minor changes for the Bosch Twinguard converter:

- Use named clusters/attributes instead of raw IDs for ZCL operations. Improves overall legibility.
- Expose binary switches instead of enum where appropriate.


Koenkk:
Thanks!

PR #7386, opened 2024-04-11
Minor improvements for Bosch BWA-1, BSD-2, BSEN-C2 & BSEN-CV

burmistrzak:
Includes the following minor fixes for the aforementioned devices:

- Add `fz.battery` to BSEN-C2 & BSEN-CV
- Remove cluster binding/reporting for `ssIasZone` & `manuSpecificBosch11` from BWA-1
- Remove cluster binding/reporting `ssIasZone` from BSD-2

Reporting is not required for IAS (`ssIasZone`), instead `commandStatusChangeNotification` is used to notify the coordinator. Attribute reporting is unnecessary for BSD-2 and BWA-1 because the respective attributes can't be changed outside of Z2M (e.g. on the device itself).

While it initially seemed like BWA-1 requires `ssIasZone` to be bound, that's actually not the case. It just takes a minute or so, after pairing, for the device to be armed. Duh! 😅


Koenkk:
Thanks!

PR #7391, opened 2024-04-12
Add support for the Sunricher HK-SENSOR-4IN1-A.

maginawin:
Remove zigbee model HK-SENSOR-4IN1-A from Namron because it's made of Sunricher.


src/devices/sunricher.ts
```diff
@@ -56,6 +68,23 @@ async function syncTime(endpoint: Zh.Endpoint) {
 }
 
 const definitions: Definition[] = [
+    {
+        zigbeeModel: ['HK-SENSOR-4IN1-A'],
+        model: 'HK-SENSOR-4IN1-A',
+        vendor: 'Sunricher',
+        description: '4IN1 Sensor',
+        fromZigbee: [fz.ias_occupancy_alarm_1, fz.battery, fz.temperature, fz.humidity, fz.illuminance],
```
Koenkk, line 76: `fromZigbee`, `toZigbee` and `deviceEndpoints({'endpoints': {'1': 1, '2': 2, '3': 3, '4': 4, '5': 5}}),` can be safely removed, please try if it works after that.

maginawin: Hello, thanks for your advice, the device works fine after that.

src/devices/sunricher.ts
```diff
@@ -56,6 +68,20 @@ async function syncTime(endpoint: Zh.Endpoint) {
...
+    {
+        zigbeeModel: ['HK-SENSOR-4IN1-A'],
+        model: 'HK-SENSOR-4IN1-A',
+        vendor: 'Sunricher',
+        description: '4IN1 Sensor',
+        extend: [
+            battery(),
+            identify(),
+            occupancy(),
+            iasZoneAlarm({'zoneType': 'generic', 'zoneAttributes': ['alarm_1', 'alarm_2', 'tamper', 'battery_low']}),
```
Koenkk, line 80:
```suggestion
            iasZoneAlarm({'zoneType': 'occupancy', 'zoneAttributes': ['alarm_1', 'tamper', 'battery_low']}),
```

Missed this initially, but can you try with this and remove `occupancy()`?

maginawin: Hello, thanks for your help. After removing the `occupancy()`, I took several tests and found that the `iasZoneAlarm()`  on this device isn't working properly, the device won't report occupancy status. 
I retained the `occupancy()` and removed the `iasZoneAlarm()`, and now the device is working properly.

Koenkk:
Thanks!

PR #7392, opened 2024-04-12
fix(Aqara): refactor rotary knobs to modern extend

mrskycriper:
Fixed recently added `ZNXNKG01LM` to modern extend and updated it's name to be in line with the wireless version. Also updated `ZNXNKG02LM` to modern extend.


Koenkk:
Thanks!

PR #7397, opened 2024-04-14
add _TZE200_mby4kbtq

b2un0:
relates to #Koenkk/zigbee2mqtt/issues/17684

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1003,6 +1003,33 @@ const definitions: Definition[] = [
         },
     },
     {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_mby4kbtq']),
+        model: 'TS0601_gas_sensor_4', // _TZE200_mby4kbtq looks like TS0601_gas_sensor_2
+        vendor: 'TuYa',
+        description: 'Gas sensor',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        exposes: [
+            e.gas(),
+            tuya.exposes.gasValue().withUnit('LEL'),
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -1002,6 +1002,33 @@ const definitions: Definition[] = [
...
+            e.gas(),
+            tuya.exposes.gasValue().withUnit('LEL'),
+            e.binary('preheat', ea.STATE, true, false).withDescription('Indicates sensor preheat is active'),
+            tuya.exposes.faultAlarm(),
+            e.binary('alarm_switch', ea.STATE_SET, true, false),
+            tuya.exposes.silence(),
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [1, 'gas', tuya.valueConverter.trueFalse0],
```
Koenkk, line 1023: Data points are similar to `TS0601_gas_sensor_2`, can you see if this device also supports the additional data points that `TS0601_gas_sensor_2` defines (e.g. `[7, 'alarm_time', tuya.valueConverter.raw],`)

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/2706

src/devices/tuya.ts
b2un0: However, if the "datapoints" match those of "TS0601_gas_sensor_2", the "alarm_switch" on DP 13 would be missing (is not activated in "TS0601_gas_sensor_2").

b2un0: these are the datapoints from the tuya dev console
https://github.com/b2un0/z2m-device-converter/blob/main/tuya/_TZE200_mby4kbtq.json5

Koenkk: Then let's leave it as is! Thanks!

PR #7398, opened 2024-04-14
feat(add): ZDMS16-1 and ZDMS16-2 Avatto dimmer switches

Suxsem:
As discussed here: https://github.com/Koenkk/zigbee2mqtt/issues/22175

Aliexpress page: https://it.aliexpress.com/item/1005006523496245.html

These dimmers uses different enum values for switchType comparing to standard tuya dimmers, so a new tuya converter (switchType2) has been introduced in lib/tuya.ts

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1699,10 +1699,10 @@ const definitions: Definition[] = [
         toZigbee: [tuya.tz.datapoints],
         configure: tuya.configureMagicPacket,
         exposes: [
-                tuya.exposes.lightBrightnessWithMinMax(),
-                tuya.exposes.countdown(),
-                tuya.exposes.switchType(),
-                e.power_on_behavior().withAccess(ea.STATE_SET),
+            tuya.exposes.lightBrightnessWithMinMax(),
+            tuya.exposes.countdown(),
+            tuya.exposes.switchType(),
+            e.power_on_behavior().withAccess(ea.STATE_SET),
         ],
         meta: {
             tuyaDatapoints: [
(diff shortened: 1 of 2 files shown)


src/devices/tuya.ts
```diff
@@ -1685,6 +1685,81 @@ const definitions: Definition[] = [
...
+            tuya.exposes.countdown().withEndpoint('l1'),
+            tuya.exposes.switchType().withEndpoint('l1'),
+            tuya.exposes.lightBrightnessWithMinMax().withEndpoint('l2'),
+            tuya.exposes.countdown().withEndpoint('l2'),
+            tuya.exposes.switchType().withEndpoint('l2'),
+            e.power_on_behavior().withAccess(ea.STATE_SET),
+        ],
+        meta: {
+            multiEndpoint: true,
+            tuyaDatapoints: [
```
Koenkk, line 1743: Data points seem to be the same as `TS0601_dimmer_4`, can you confirm and add it as a `tuya.whitelabel` to `TS0601_dimmer_4`?

src/devices/tuya.ts
```diff
@@ -1685,6 +1685,81 @@ const definitions: Definition[] = [
...
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        exposes: [
+            tuya.exposes.lightBrightnessWithMinMax(),
+            tuya.exposes.countdown(),
+            tuya.exposes.switchType(),
+            e.power_on_behavior().withAccess(ea.STATE_SET),
+        ],
+        meta: {
+            tuyaDatapoints: [
```
Koenkk, line 1708: This one looks to be the same as `TS0601_dimmer_5`?

src/devices/tuya.ts
Suxsem: hi, it's not. the new one doesn't support lightType and uses endpoint 4 and 8 for switchType.
also the switchType enum has a different mapping.

src/devices/tuya.ts
Suxsem: same as before, it's not. the new one doesn't support lightType and uses endpoint 4 for switchType.
also the switchType enum has a different mapping.

Koenkk:
Thanks!

EyalRotem:
Hi, i bought the Avatto ZDMS16-1 but it was unsupported since i guess the zigbee manufacturer (_TZE204_nqqylykc) is different, how can i it and submit a PR?

image

PR #7399, opened 2024-04-15
Rework `zoneStatus` converter for Bosch BSD-2

burmistrzak:
Includes the following major changes for the Bosch smoke detector:

- Use custom `zoneStatus` converter (because Bosch's implementation is different from default ZCL)
- Properly decode bitmask instead of fixed values
- Introduce `alarm_smoke`, `alarm_intruder` & `alarm_silenced` states
- Use different command to remotely trigger smoke alarm


N3rdix:
> @N3rdix Feel free to give it a try. 🤞

I'll give it a try this afternoon and update my local converter from your draft 👍

N3rdix:
I've tested the new converter and:
✅ triggering the manual alarms still works perfectly
✅ silencing a manually triggered smoke alarm works with a short press
❌ silencing a manually triggered intruder alarm does not work with a short or long press. Need to toggle the switch
🚧 need to test when faking a real alarm with a candle

Koenkk:
I think it can be done with this call: [example](https://github.com/Koenkk/zigbee-herdsman/blob/206b0f9b4f553a0a89fcffb5a2c21cc87a610353/src/controller/model/endpoint.ts#L614), but currently it's not exposes on the interface.

You can call the adapter directly by hacking a bit, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/da646ed04b0d21e9bead5185320e4457583d5f70/src/devices/schneider_electric.ts#L258

vjonas:
> @vjonas Do you remember what you did (settings on SHC, etc.) to successfully sniff [those](https://github.com/Koenkk/zigbee-herdsman-converters/pull/7329#issuecomment-2054223012) four commands you've implemented? I haven't been able to find them in my own packet captures just yet... 👀

I believe it had something to do with little endian VS big endian conversion. 
You probably sniffed the same values no?

vjonas:
> > I believe it had something to do with little endian VS big endian conversion. You probably sniffed the same values no?
> 
> Nope, haven't seen any of them, no matter the byte order... Did you trigger some sort of action on the SHC? 👀
> 
> What I saw were the exact same commands @safakaltun already captured.

Hmm not sure, I explored and fooled around with the public api first:
https://local.apidocs.bosch-smarthome.com/#/States/put_devices__deviceId__services_Alarm_state

But the values most likely came from sniffing the traffic. Let me check on the windows sniffing pc for any remaining files..

vjonas:
[zigbee sniff and network key.zip](https://github.com/Koenkk/zigbee-herdsman-converters/files/15008473/zigbee.sniff.and.network.key.zip)
This should be the sniff and network key @burmistrzak

N3rdix:
> Seems like Bosch use some aggressive debouncing/timings in their firmware, making especially button actions more unreliable. On that note, it's quite interesting, the official manual only mentions the option to remotely silence smoke alarms from within the Bosch app. Take that as you will.

Will the broadcast work with this current state? I don't see the mentioned `sendZclFrameToAll` in the commits.

> @N3rdix When this PR is merged, silencing (almost) all networked smoke alarms _should_ finally be possible.

I will definitely give it a try, will copy your changes to my local converter in case the merge takes longer.

----
_p.s.:
🏆 appreciate all your hard work, I am thrilled by your commitment. Thanks so far 👏_

Koenkk:
@burmistrzak I think it makes sense to add a new `toZigbee` command instead (which sends the required broadcast message). 

Can this pr be merged?

Koenkk:
Thanks!

Koenkk:
> I mean something more along the lines of e.g. a new topic zigbee2mqtt/broadcast to send arbitrarily commands as broadcast from Node-RED & Co.

I would prefer something more user friendly here, where users can e.g. publish `zigbee2mqtt/DEVICE_ID/set` `{"silence":""}` (even though it will not apply to just that `DEVICE_ID`

Koenkk:
sounds good!

vjonas:
> @Koenkk Might be worth noting that removing `smoke_alarm_state` and `intruder_alarm_state`, while unlikely already in use by anyone, is _technically_ a breaking change. Just FYI. ☝️

Thanks for pointing that out. I might have used it already in a automation to disable all alarms with a button. Nothing to worry about though.

PR #7406, opened 2024-04-18
Fix me167 schedule

StefanLobbenmeier:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/19787
Based on https://github.com/sychu/avatto_me167_TZE200_p3dbf6qs/tree/main


src/devices/tuya.ts
```diff
@@ -3571,13 +3570,71 @@ const definitions: Definition[] = [
...
+                [47, 'local_temperature_calibration', tuya.valueConverter.localTempCalibration2],
+            ],
+        },
+    },
+    {
+        fingerprint: tuya.fingerprint('TS0601', [
+            // '_TZE200_bvu2wnxz', /* model: 'ME167', vendor: 'AVATTO' */
+            // '_TZE200_6rdj8dzm', /* model: 'ME167', vendor: 'AVATTO' */
+            '_TZE200_p3dbf6qs', /* model: 'ME167', vendor: 'AVATTO' */
+            // '_TZE200_rxntag7i', /* model: 'ME168', vendor: 'AVATTO' */
```
StefanLobbenmeierObjego, line 3592: The patch worked for me with the fingerprint _TZE200_p3dbf6qs, I have no idea for the other devices. I will ask around in the ticket if someone needs an additional fingerprint here

StefanLobbenmeier, line 3592: The patch worked for me with the fingerprint _TZE200_p3dbf6qs, I have no idea for the other devices. I will ask around in the ticket if someone needs an additional fingerprint here

src/devices/tuya.ts
```diff
@@ -3571,13 +3570,71 @@ const definitions: Definition[] = [
...
+        ]),
+        model: 'TS0601_thermostat_5',
+        vendor: 'TuYa',
+        description: 'Thermostatic radiator valve',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        whiteLabel: [
+            tuya.whitelabel('AVATTO', 'ME167', 'Thermostatic radiator valve', ['_TZE200_p3dbf6qs']),
```
StefanLobbenmeier, line 3600: this makes the tests fail apparently, because there is another whitelabelled ME167. I am not sure how to deal with that though - do I just come up with another name for it?

StefanLobbenmeier:
Good question. I broke it into atomic commits so you can verify that each of them leaves functionality unchanged. Not sure if there is a way I can prove this with unit tests. 

The alternative is duplicating a lot of code, would you prefer that?

StefanLobbenmeier:
> The alternative is duplicating a lot of code, would you prefer that?

I mean that as a genuine question btw - if that is preferred I can change the PR to basically only contain extra lines, except for removing the fingerprint

Koenkk:
@DaaaDave since you just added a new TS0601 thermostat in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7407, could you test this change out?

src/lib/tuya.ts
```diff
@@ -811,52 +811,54 @@ export const valueConverter = {
...
-            }
-            for (const transition of transitions) {
-                const timeTemp = transition.split('/');
-                if (timeTemp.length != 2) {
-                    throw new Error('Invalid schedule: wrong transition format: ' + transition);
+    thermostatScheduleDayMultiDPN: (numberOfSchedules: number) => {
+        return {
+            from: (v: string) => {
+                const schedule = [];
+                for (let index = 1; index < 17; index = index + 4) {
```
sychu, line 818: numberOfSchedules should be also used in this loop

index < 4*numberOfSchedules + 1 (this +1 actually make no difference ;-)

src/lib/tuya.ts
```diff
@@ -811,52 +811,54 @@ export const valueConverter = {
...
-            const payload = [0];
-            const transitions = v.split(' ');
-            if (transitions.length != 4) {
-                throw new Error('Invalid schedule: there should be 4 transitions');
-            }
-            for (const transition of transitions) {
-                const timeTemp = transition.split('/');
-                if (timeTemp.length != 2) {
-                    throw new Error('Invalid schedule: wrong transition format: ' + transition);
+    thermostatScheduleDayMultiDPN: (numberOfSchedules: number) => {
```
sychu, line 814: What about setting default value for numberOfSchedules:

thermostatScheduleDayMultiDPN: (numberOfSchedules: number = 4)

With default value other thermostates definitions will not require update

sychu (changes requested):

src/lib/tuya.ts
StefanLobbenmeier:
```suggestion
                for (let index = 1; index < 4*numberOfSchedules + 1; index = index + 4) {
```

good catch

src/lib/tuya.ts
StefanLobbenmeier: Good idea. I think there still is a small change in the other thermostats since we have to call method instead of using the field right? 

(Will need to apply this another, on mobile right now)

sychu: Oh, indeed entry method for thermostats definition is:
thermostatScheduleDayMultiDPWithDayNumber, so it look like there also should be default value set.

After this failing tests should pass, but i think it is required to add additional tests to check new parameter anyway.

StefanLobbenmeierObjego: Applied the default value as you suggested. I will also add you to my fork feel free to push any changes directly into the branch

StefanLobbenmeier: Applied the changes you suggested, I will also add you to my branch (it is your code anyway), feel free to push to the branch directly.

Doable149:
Hello, there is a part in this PR that interests me about the schedules count parameter in the `thermostatScheduleDayMultiDP` function. Is this PR still active?

StefanLobbenmeier:
Yeah probably a good idea to start a new PR, since this one needed rebasing anyway and I never could figure out how to fix the CI

PR #7417, opened 2024-04-20
fix(GS): use alarm_2 for SGPHM-I1 and attribute cleanup

mrskycriper:
Multiple reports came in that SGPHM-I1 in fact uses `alarm_2` and was not reporting alarms correctly this whole time.

Also removed `tamper` and `battery_low` attributes form both gas sensors as they don't physically have tampers or batteries.


Koenkk:
Good as usual, thanks again! 👍

PR #7420, opened 2024-04-21
Add definition for Adurosmart 81898

Itay080:
I've added the automatically generated definition for Adurosmart 81898 which works fine for my module


Koenkk:
Thanks!

PR #7425, opened 2024-04-22
Add new variant of Innr E14 candle

bfayers:
Recently bought some new Innr E14 candles (dimmable, 2700k/warm white) but they were RB 243.


Koenkk:
Thanks!

PR #7427, opened 2024-04-22
Add broadcast alarm functionality for Bosch BSD-2

burmistrzak:
Quite an amazing feature of the Bosch smoke detectors. 🤩

By simply issuing a special Zigbee broadcast command, you'll be able to change the alarm/siren state _of all BSD-2 in your network **at once**_.
The main benefit here's that you can e.g. trigger a network-wide fire alarm when only one detector senses smoke, but also silence them all at once, except for the device that initially detected smoke.

Follow-up for #7418

**To-Dos**
- [x] https://github.com/Koenkk/zigbee-herdsman/pull/1028
- [x] ~~Ensure `adapter.sendZclFrameToAll` is called correctly~~
- [x] ~~Where do we get a valid `zclTransactionSequenceNumber` from?~~

Depends on https://github.com/Koenkk/zigbee-herdsman/pull/1036


src/devices/bosch.ts
```diff
@@ -174,6 +182,26 @@ Example: 30ff00000102010001`;
...
+                    Zcl.FrameType.SPECIFIC,
+                    Zcl.Direction.CLIENT_TO_SERVER,
+                    true,
+                    Zcl.ManufacturerCode.ROBERT_BOSCH_GMBH,
+                    71, // zclTransactionSequenceNumber
+                    'boschSmokeDetectorSiren',
+                    1280, // ssIasZone
+                    {data: index});
+                // @ts-expect-error
+                await meta.device.constructor.adapter.sendZclFrameToAll(255, broadcastFrame, 1, true);
```
Koenkk, line 200: I know I proposed to do it like this earlier, but thinking about it again it would be cleaner to add a `broadcastCommand` to [`controller.ts`](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/controller.ts) and use it here (we will modify `meta` to pass the `zh` `controller`)

Koenkk: If you have the time, please make a PR for it, maybe as part of https://github.com/Koenkk/zigbee-herdsman/pull/1028?

Nerivec:
Latest [herdsman commit](https://github.com/Koenkk/zigbee-herdsman/pull/1028/commits/11266fc3bc0ed8d6438488dcf0197dd87740d5e7) should now allow to model tz for broadcasts in the same manner as https://github.com/Koenkk/zigbee-herdsman-converters/blob/d9640a4d39fcc924d97e2a1034105c75452ff8c2/src/converters/toZigbee.ts#L243

Koenkk:
Thanks!

N3rdix:
Might be worth mentioning the behavior in the device documentation. Once I've tested the feature I can do that as well.

Nice work everybody 👍

PR #7429, opened 2024-04-23
Changes in the converter for the PTVO device

ptvoinfo:


src/converters/fromZigbee.ts
```diff
@@ -787,13 +790,16 @@ const converters1 = {
...
             }
             if (msg.data.hasOwnProperty('powerFactorPhB')) {
-                payload.power_factor_phase_b = precisionRound(msg.data['powerFactorPhB'] / 100, 2);
+                const property = postfixWithEndpointName('power_factor_phase_b', msg, model, meta);
+                payload[property] = precisionRound(msg.data['powerFactorPhB'] / 100, 2);
             }
             if (msg.data.hasOwnProperty('powerFactorPhC')) {
-                payload.power_factor_phase_c = precisionRound(msg.data['powerFactorPhC'] / 100, 2);
+                const property = postfixWithEndpointName('power_factor_phase_c', msg, model, meta);
+                payload[property] = precisionRound(msg.data['powerFactorPhC'] / 100, 2);
```
Koenkk, line 802: This will introduce issues for device which use this converter and have `meta: {multiEndpoint: true}`. Can you add `multiEndpointSkip: []` for all those device?

ptvoinfo: I've found only two devices that expose the 'power_factor' value (Tuya). I've added the corresponding definition there. But I didn't add it for device that configure reporting for it.

ptvoinfo (commented):
@Koenkk Could you please check what it is wrong with this commit? It is unclear with the details.

Koenkk:
I messed up when pushing to your branch, merged the changes in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7443, thanks!

PR #7430, opened 2024-04-23
Update Elko 4523430 max load value

Swampen:
## Changes 
* Update Elko 4523430 max load value 

According to the [device specification](https://www.elko.no/getfile.php/1343780-1712344203/ELKO.no/Produkter%20PIM/EKO05/NO-EKO05680/documents-custom/Datasheet4523430.pdf), the max load is 2300 W. I am currently running with more than 2000W on my device


Koenkk:
Thanks!

PR #7438, opened 2024-04-25
Add meta option to disable return value on `/set` for `cover_position_tilt`

burmistrzak:
Implements a lightly customized `cover_position_tilt` converter, that doesn't return the set value, because we already have reporting for `currentPositionLiftPercentage` configured.

Nothing more. 😊

Should help with https://github.com/itavero/homebridge-z2m/issues/852#issuecomment-2075542164


Koenkk:
> that doesn't return the set value,

Does this cause any problems?

src/devices/bosch.ts
```diff
@@ -1505,8 +1531,17 @@ const definitions: Definition[] = [
...
+            fz.power_on_behavior,
+            fz.cover_position_tilt,
+            fzLocal.bmct,
+        ],
+        toZigbee: [
+            tz.power_on_behavior,
+            tzLocal.bmct_cover_position_tilt,
+            tzLocal.bmct,
+        ],
         meta: {multiEndpoint: true},
```
Koenkk, line 1545: Instead of adding a new converter, I would use `meta: {multiEndpoint: true, coverPositionTiltDisableReport: true},` and use that in `tz.cover_position_tilt`

Koenkk:
Thanks!

PR #7441, opened 2024-04-25
fix(Modern Extend): rework and expand occupancy extend

mrskycriper:
- [x] Make standalone instead of binary derivative
- [x] Add delay and threshold settings
- [x] Update multi endpoint support
- [x] Add GET support
- [x] Update configure

Contact settings require cluster definition update https://github.com/Koenkk/zigbee-herdsman/pull/1031


Koenkk:
Thanks again!

PR #7444, opened 2024-04-25
refactoring(Modern Extend): replace single configure step with array

mrskycriper:
Replaces single configure step in Modern Extends with array to allow for more flexible extends


sjorge:
I'm confused why you want this? You can already have multiple modernExtend add a configure step to a device?

mrskycriper:
> I'm confused why you want this? You can already have multiple modernExtend add a configure step to a device?

I've encountered enough situations where single extend (for example battery) has multiple attributes to configure reports of. Single configure makes it unnecessarily difficult to do conditionally. And it's an artificial limitation since they are processed as an array anyway.

sjorge:
> > I'm confused why you want this? You can already have multiple modernExtend add a configure step to a device?
> 
> I've encountered enough situations where single extend (for example battery) has multiple attributes to configure reports of. Single configure makes it unnecessarily difficult to do conditionally. And it's an artificial limitation since they are processed as an array anyway.

You can just loop over them and call `setupConfigureForBinding()` multiple times in the configuration function you add, I'm pretty sure we already do that somewhere. 

Making it an array just makes it a bit more awkward to deal for no real gain IMHO. (But lucky for everybody my HO doesn't really matter)

src/devices/lifecontrol.ts
```diff
@@ -25,13 +25,15 @@ function airQuality(): ModernExtend {
...
 function electricityMeterPoll(): ModernExtend {
-    const configure: Configure = async (device, coordinatorEndpoint, definition) => {
-        const endpoint = device.getEndpoint(1);
-        await reporting.bind(endpoint, coordinatorEndpoint, ['haElectricalMeasurement', 'seMetering']);
-        await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
-        await reporting.readMeteringMultiplierDivisor(endpoint);
-        await reporting.currentSummDelivered(endpoint);
-    };
+    const configure: Configure[] = [
+        async (device, coordinatorEndpoint, definition) => {
```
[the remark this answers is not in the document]
mrskycriper: Perhaps, or maybe just reuse configure from electricity meter extend

sjorge:
> That is the problem, setup configure for reporting and binding returns full configure function. So it's either one step or adding them without setup to achieve multiple configure with no real reason to do it that way

```typescript
const configure: Configure = async (device, coordinatorEndpoint, definition) => {
  const attrA = setupConfigureForBinding('genPowerCfg', 'batteryPercentageRemaining');
  const attrB = setupConfigureForBinding('genPowerCfg', 'batteryVoltage');
  await attrA(device, coordinatorEndpoint, definition);
  await attrB(device, coordinatorEndpoint, definition);
};
```

I would expect the following to just work, optionally you can probably ditch the `await` to make it more async.

src/devices/lifecontrol.ts
```diff
@@ -25,13 +33,19 @@ function airQuality(): ModernExtend {
...
 
 function electricityMeterPoll(): ModernExtend {
-    const configure: Configure = async (device, coordinatorEndpoint, definition) => {
-        const endpoint = device.getEndpoint(1);
-        await reporting.bind(endpoint, coordinatorEndpoint, ['haElectricalMeasurement', 'seMetering']);
-        await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
-        await reporting.readMeteringMultiplierDivisor(endpoint);
-        await reporting.currentSummDelivered(endpoint);
-    };
+    const configure: Configure[] = [
```
Koenkk, line 36: I propose to just do an `endpoint.read()` here, this is the only place where `electricityMeterPoll` is used

mrskycriper: I had some plans for `setupConfigureForReading`, It should be useful for attributes that act as settings and reporting them is wasteful, but without reading them first their respective expose fields will be left empty. This confuses people and creates errors in the logs sometimes

Koenkk:
> I would expect the following to just work, optionally you can probably ditch the await to make it more async.

Never ditch the async, as it will create an unhandled promise exception when it fails (resulting in z2m to crash).

I think this is fine

Koenkk:
Extend is static, it doesnt have the device, but the exposed can be a function where device is one of the args

Koenkk:
@burmistrzak yes it's definitely something we should allow, would you mind making a PR for that? We can do something similar as for the [`configure`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/da65b1aeffd96527df02725b49de61e453fee059/src/index.ts#L170)

Koenkk:
Ah yes! That should be enough indeed.

PR #7446, opened 2024-04-26
device_support(Sber): add SBDV-00154

mrskycriper:
Image https://github.com/Koenkk/zigbee2mqtt.io/pull/2726

### DIFF
--- a/src/devices/sber.ts
+++ b/src/devices/sber.ts
@@ -53,7 +53,17 @@ const definitions: Definition[] = [
             battery({voltage: true, voltageReporting: true}),
         ],
     },
-    // Sber SBDV-00154 Smart leak sensor (fingerprint unknown)
+    {
+        fingerprint: [{modelID: 'TS0207', manufacturerName: '_TZ3000_c8bqthpo'}],
+        model: 'SBDV-00154',
+        vendor: 'Sber',
+        description: 'Smart water leak sensor',
+        extend: [
+            ignoreClusterReport({cluster: 'genBasic'}),
+            iasZoneAlarm({zoneType: 'water_leak', zoneAttributes: ['alarm_1', 'battery_low']}),
+            battery(),
+        ],
+    },
 ];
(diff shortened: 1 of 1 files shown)


src/devices/sber.ts
```diff
@@ -53,7 +53,17 @@ const definitions: Definition[] = [
             battery({voltage: true, voltageReporting: true}),
         ],
     },
-    // Sber SBDV-00154 Smart leak sensor (fingerprint unknown)
+    {
+        fingerprint: [{modelID: 'TS0207', manufacturerName: '_TZ3000_c8bqthpo'}],
```
Koenkk, line 57: Can this be merged with `TS0207_water_leak_detector`? (and add it as a `tuya.whitelabel` there)

Koenkk: > This makes it harder to maintain as you have to search for a specific one in the sea of others

Why do you think this is the case? I want to prevent that we have the same code duplicated a lot of times for the same device (`TS0207`), otherwise we have to add every whitelabel as a separate defintion.

mrskycriper: Then I'd have to delete `sber.ts` and move everything into `tuya.ts`. There are a lot of branded Tuya devices that have their own vendor files as well. I understand that this is code duplication, but a necessary one for more coherent grouping by vendor. Also, since Sber got their Tuya license revoked, it is very much possible that they will release non Tuya based devices in the future. So grouping them in a vendor file seems more logical that splitting up.

Koenkk: Then let's make an exception for this one. But I prefer to have whitelabels in `tuya.ts` to prevent a lot of code duplication.

mrskycriper: > Then let's make an exception for this one. But I prefer to have whitelabels in `tuya.ts` to prevent a lot of code duplication.

I'll keep this in mind, thanks!

PR #7447, opened 2024-04-26
device_support(Yandex): add all devices as Aqara whitelabels

mrskycriper:
Yandex devices have the exact same Zigbee models with only slight changes to the case print and packaging.


Koenkk:
Thanks!

PR #7450, opened 2024-04-27
fix: IKEA VINDSTYRKA uses different DataType for measuredValue

sjorge:
Now that we have https://github.com/Koenkk/zigbee-herdsman/pull/1019, we can properly fix the IKEA VINDSTYRKA.

- [x] modernExtend to easily inject them by abstracting the onEvent stuff
- [x] VINDSTYRKA should use correct DataType
- [x] make SONOFF device also use new modernExtend

This is the debug log of joining a VINDSTYRKA and reconfiguring the reporting for pm25Measurement.measuredValue via the frontend.

[vindstyrka_join_and_configreport.log](https://github.com/Koenkk/zigbee-herdsman-converters/files/15137706/vindstyrka_join_and_configreport.log)

PR #7451, opened 2024-04-27
fix(ignore): move ubisys clusters/attributes out of zh

sjorge:
~~Depends on modernExtend added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7450, rebase and merge after that one.~~

~~Also depends on https://github.com/Koenkk/zigbee-herdsman-converters/pull/7463, this PR needs a rebase merge.~~

After this, merge https://github.com/Koenkk/zigbee-herdsman/pull/1034


sjorge:
Doing one last test pass on my devices. Forgot to open as draft.

sjorge:
I don't understand the test failure, I don't think I've changed that code ?

src/lib/ubisys.ts
```diff
@@ -1,12 +1,151 @@
...
+                ubisysCoolingDemandEnableThreshold: {ID: 0x000C, type: Zcl.DataType.uint8,
+                    manufacturerCode: Zcl.ManufacturerCode.UBISYS_TECHNOLOGIES_GMBH},
+                ubisysCoolingDemandDisableThreshold: {ID: 0x000D, type: Zcl.DataType.uint8,
+                    manufacturerCode: Zcl.ManufacturerCode.UBISYS_TECHNOLOGIES_GMBH},
+            },
+            commands: {},
+            commandsResponse: {},
+        },
+    ),
+    addClusterClosuresWindowCovering: () => deviceAddCluster(
```
sjorge, line 94: @Koenkk nog een vraagje ivm de merge order...

Hoe loopt dit als ik hier top level `manufacturerCode: Zcl.ManufacturerCode.UBISYS_TECHNOLOGIES_GMBH,` bij zet? Dan kan het in theory weg bij de attributen.

Maar het was me niet duidelijk wat er dan gebeurd met de standaard ZCL attributen in closuresWindowCovering.

Koenkk: The manufacturer code from the custom cluster will always be used ([code](https://github.com/Koenkk/zigbee-herdsman/blob/86c080ec8e522db7e774025d8592e4f27272c1de/src/controller/model/device.ts#L823C17-L823C60)). So this PR is OK.

sjorge:
I'll rebase after https://github.com/Koenkk/zigbee-herdsman-converters/pull/7450 lands.

src/lib/ubisys.ts
```diff
@@ -222,4 +222,37 @@ export const ubisysModernExtend = {
 
         return {exposes: [expose], fromZigbee, toZigbee, configure, isModernExtend: true};
     },
+    cover: (): ModernExtend => {
+        // XXX: in the future we should split fromZigbee/toZigbee, onEvent,... and have these as seperate modernExtends
+
+        const exposes: Expose[] = [];
+        exposes.push((device: Zh.Device, options) => {
```
sjorge, line 229: @koenkk as talked about on discord, I don't think this will work. Also I'm not sure what type `options` should be here. 

But aside from that, Expose type is not allowed to be a function.

Koenkk: Should work with https://github.com/Koenkk/zigbee-herdsman-converters/pull/7463, please check.

sjorge: I rewrote it to this:
```
    cover: (): ModernExtend => {
        // XXX: in the future we should split fromZigbee/toZigbee, onEvent,... and have these as seperate modernExtends

        const dynamicCover: DefinitionExposes = (device: Zh.Device, options: KeyValue): Expose[] => {
            const result: Expose[] = [];
            const coverExpose = e.cover();
            const coverType = (device?.getEndpoint(1).getClusterAttributeValue('closuresWindowCovering', 'windowCoveringType') ?? undefined);
            switch (coverType) { // cf. Ubisys J1 Technical Reference Manual, chapter 7.2.5.1 Calibration
            case 0: // Roller Shade, Lift only
            case 1: // Roller Shade two motors, Lift only
            case 2: // Roller Shade exterior, Lift only
            case 3: // Roller Shade two motors exterior, Lift only
            case 4: // Drapery, Lift only
            case 5: // Awning, Lift only
            case 9: // Projector Screen, Lift only
                coverExpose.withPosition();
                break;
            case 6: // Shutter, Tilt only
            case 7: // Tilt Blind, Tilt only
                coverExpose.withTilt();
                break;
            case 8: // Tilt Blind, Lift & Tilt
            default:
                coverExpose.withPosition().withTilt();
                break;
            }

            result.push(coverExpose);
            result.push(e.power().withAccess(ea.STATE_GET));
            result.push(e.energy().withAccess(ea.STATE_GET));
            result.push(e.linkquality());
            return result;
        };

        return {exposes: dynamicCover, isModernExtend: true};
    },
```

But I can't get it to work.

sjorge: Dropping the last commit, now does with with https://github.com/Koenkk/zigbee-herdsman-converters/pull/7463, which would be my preferred way forward for now.

A modernExtend conversion for src/devices/ubisys.js is on my list of things to do, but preferably not as part of this PR as there is a lot of code here that needs to be refactored.

sjorge: I'm testing this PR, with the last commit dropped now... I'll now how well it works in a bit.

sjorge:
All seems good, I tested both TRV vacationMode changes, S1+C4+S2 input_action configuration and a full J1 calibration

```
[2024-04-28 22:05:01] warning:  zhc:ubisys: ubisys:   Done - will now read back the results.
[2024-04-28 22:05:01] warning:  zhc:ubisys: ubisys: Cover configuration read: {"windowCoveringType":0,"physicalClosedLimitLiftCm":65535,"physicalClosedLimitTiltDdegree":65535,"installedOpenLimitLiftCm":0,"installedClosedLimitLiftCm":240,"installedOpenLimitTiltDdegree":0,"installedClosedLimitTiltDdegree":900}
```

All seems to still work, but the J1 calibration is the most complex so I focused mostly on that

sjorge:
> @sjorge So the idea here is for ZH to eventually only include what's actually in the ZCL spec, right? What do you think, does it still make sense to add new Bosch-specific definitions to ZH, or should I wait until I can also refactor `bosch.ts` in a similar fashion?

You can probably start by adding the new ones in zhc on the devices that need them, then moving the other ones out of zh into zhc can come after. It's fine to have some in zhc and some still in zh, so new ones don't get added to zh IMHO.

Koenkk:
Merged https://github.com/Koenkk/zigbee-herdsman-converters/pull/7463 !

sjorge:
I only give it another quick retest after changing the casing of the DataType, I just turned on/off some of the switches. I did not do a full J1 reconfigure as it's rather noisy to make the blinds go up and down a few times during calibration and it's already pretty late.

Koenkk:
Thanks!

PR #7454, opened 2024-04-27
Add definition for SE Wiser CCT5010-0003

dflemstr:
Newest version of the SE Wiser dimmer puck module. As far as I can tell, it behaves mostly the same as the other puck versions, the only main difference being that it supports a neutral lead.  However more testing might reveal other differences?


Koenkk:
Thanks!

PR #7455, opened 2024-04-27
Improvements to PJ-1203A

schauveau:
Various improvements to PJ-1203A described in https://github.com/Koenkk/zigbee2mqtt/issues/22248


schauveau:
All support functions and objects are placed in a PJ1203A object  in `devices/tuya.ts`. An alternative could be to put them in `lib/tuya.ts` but they are all very specific to that device.  

I created new exposes such as `powerWithChannel` instead of reusing `powerWithPhase`. They are pretty much identical except that 'Phase' is not really the appropriate term for the PJ1203A (i.e. dual-channel energy meter vs triple-phase meter).

schauveau:
Here are notes to be added to the PJ-1203A documentation. I will create a pull request once this one is merged. 

## Notes

In some or all versions of the `PJ-1206A`, the energy flow datapoints are incorrectly emitted during the next update phase. In practice, that means that the energy flow direction is incorrect during `update_frequency` seconds. This is addresses by the `late_energy_flow_a` and  `late_energy_flow_b` options that cause all other channel datapoints to be delayed by the same amount. Be aware that the additional delay does not occur when the power/current is measured to zero.

When the options `signed_power_a` and `signed_power_b` are enabled, the energy flow directions are not reported in `energy_flow_a` and `energy_flow_b` as `consuming` or `producing`. Instead, they are set to `sign` and a negative power values are used when producing.

src/devices/tuya.ts
```diff
@@ -25,6 +25,266 @@ const fzZosung = zosung.fzZosung;
 const tzZosung = zosung.tzZosung;
 const ez = zosung.presetsZosung;
 
+const PJ1203A = {
```
Koenkk, line 28: Could you move this to `src/lib/tuya.ts`

src/devices/tuya.ts
```diff
@@ -25,6 +25,266 @@ const fzZosung = zosung.fzZosung;
...
+    //   - cache the datapoints for each channel and publish them together.
+    //   - (OPTIONAL) solve the issue described above by waiting for the next energy flow datapoint
+    //     before publishing the cached channel data.
+    //   - (OPTIONAL) provide signed power instead of energy flow.
+    //   - detect missing or reordered Zigbee message using the Tuya 'seq' attibute and invalidate
+    //     cached data accordingly.
+    //
+
+    get_private_state: (device: Zh.Device) => {
+        let priv = globalStore.getValue(device, 'private_state');
```
Koenkk, line 71: You can use the third `default_unknown` argument here to return a default value instead of ` if (priv===undefined) {`

src/devices/tuya.ts
```diff
@@ -25,6 +25,266 @@ const fzZosung = zosung.fzZosung;
...
+    // The features implemented here are
+    //   - cache the datapoints for each channel and publish them together.
+    //   - (OPTIONAL) solve the issue described above by waiting for the next energy flow datapoint
+    //     before publishing the cached channel data.
+    //   - (OPTIONAL) provide signed power instead of energy flow.
+    //   - detect missing or reordered Zigbee message using the Tuya 'seq' attibute and invalidate
+    //     cached data accordingly.
+    //
+
+    get_private_state: (device: Zh.Device) => {
```
Koenkk, line 70: Please use camelCase for all methods, e.g. `getPrivateState`

src/devices/tuya.ts
```diff
@@ -7670,34 +7930,49 @@ const definitions: Definition[] = [
...
         exposes: [
             e.ac_frequency(), e.voltage(),
-            tuya.exposes.powerWithPhase('a'), tuya.exposes.powerWithPhase('b'), tuya.exposes.powerWithPhase('ab'),
-            tuya.exposes.currentWithPhase('a'), tuya.exposes.currentWithPhase('b'),
-            tuya.exposes.powerFactorWithPhase('a'), tuya.exposes.powerFactorWithPhase('b'),
-            tuya.exposes.energyFlowWithPhase('a'), tuya.exposes.energyFlowWithPhase('b'),
-            tuya.exposes.energyWithPhase('a'), tuya.exposes.energyWithPhase('b'),
-            tuya.exposes.energyProducedWithPhase('a'), tuya.exposes.energyProducedWithPhase('b'),
-            e.numeric('update_frequency', ea.STATE).withUnit('s').withDescription('Update frequency'),
+            tuya.exposes.powerWithChannel('a'), tuya.exposes.powerWithChannel('b'), tuya.exposes.powerWithChannel('ab'),
```
Koenkk, line 7945: Why not use `powerWithPhase` here?

schauveau: This is because a Phase and a Channel are two different things from an electrical point of view. The  PJ-1203A is a dual channel (or 2 gangs) device vs for instance the `TS0601_3_phase_clamp_meter` that is a 3-phase meter. 
That does not change anything from the perspective of Z2M. I was just a bit annoyed to see both `phase` and `device` in the documentation.

Another way could be to keep Phase in the function name but with an optional argument to customize the documentation. Something like that should work I think:
```javascript
    powerWithPhase: (phase: string, kind: string = 'phase') => e.numeric(`power_${phase}`, ea.STATE).withUnit('W')
        .withDescription(`Instantaneous measured power (${kind} ${phase.toUpperCase()})`),
```
and then
```
 tuya.exposes.powerWithPhase('a', kind='channel')
```

src/devices/tuya.ts
schauveau: Are you sure? I was under the impression that  `src/lib/tuya.ts` was for generic and potentially reusable things.
Everything in `PJ1203A` is really specific to a single device.

src/devices/tuya.ts
schauveau: Ok I'll do that. For information, I choose to do it that way for two reasons: First to avoid creating and discarding a new default state every time `get_private_state` is called (is there a way to customize the device at startup? I could not find how) and second because I was planning for future customization of the initial `private_state` for some versions of the device (e.g. different `sec_inc` or different ordering of the DPs).

src/devices/tuya.ts
schauveau: I have been pondering on that for a while and I have a proposal to better integrate `PJ1203A` in `devices/tuya.ts`. 

1. move `PJ1203A.sync_time_increase_seq` and `PJ1203A.strict_fz_datapoints` to `fzLocal` with a `PJ1203A_` prefix.
2. get rid of `PJ1203A.options` by declaring the two options in the device `Definition`.
    - or by putting them in a new local variable `optLocal` similar to `fzLocal` and `tzLocal`.
3. Move the content of `PJ1203A.valueConverters` into a new local variable `convLocal` 
4. Create a new local variable `storeLocal` to hold functions that provide a globalStore for a device and move  `PJ1203A.get_private_state` to `storeLocal.PJ1203A`.  
5 And finally, transform everything remaining in `PJ1203A` into function members of the private state (so within `storeLocal.PJ1203A`). That should probably be done anyway.

src/devices/tuya.ts
Koenkk: Looks good! Maybe call it `powerWithPostifx` then?

schauveau: But then I would have to do the renaming it in all devices. Also that would also break a lot of external converters.

More generally, if we go that way the converter could simply be `power` with an optional argument `suffix`. 
And some helper functions could be added to automate the process. Something like that 

```javascript
function extend(name,suffix) { 
   return suffix ? name+"_"+suffix : name ;
}

function group(kind: string, name: string) {
  return name ? ` (${kind} ${name})` :  "";
} 

power: (suffix: string = null, kind: string = 'phase') => e.numeric(extend("power",suffix), ea.STATE).withUnit('W')
        .withDescription("Instantaneous measured power"+group(kind,suffix.toUpperCase()) ,

```

Koenkk: Even better!

src/devices/tuya.ts
Koenkk: Sounds good, can you commit the code so that I can properly review it?

src/devices/tuya.ts
Koenkk: I see, then let's leave it as is.

src/devices/tuya.ts
schauveau: The proposed changes would impact several converters so, for now, I am switching back to the original `WithPhase` syntax.  Maybe for another pull request  ...

Koenkk:
Thanks!

PR #7457, opened 2024-04-27
ZigbeeTLc firmware devices support

robvanoostenrijk:
Add support for the [pvvx ZigbeeTLc ](https://github.com/pvvx/ZigbeeTLc/) custom firmware devices.

This custom firmware is available for several temperature / humidity sensors:

- [Xiaomi Mijia (LYWSD03MMC)](https://pvvx.github.io/ATC_MiThermometer)
- [Xiaomi Miaomiaoce (MHO-C401)](https://pvvx.github.io/MHO_C401)
- [Qingping Temp & RH Monitor (CGG1-Mijia)](https://pvvx.github.io/CGG1)
- [CGDK2 Qingping Temp & RH Monitor Lite](https://pvvx.github.io/CGDK2)
- [Xiaomi Mijia MJWSD05MMC](https://pvvx.github.io/MJWSD05MMC)
- [MHO-C122 (2AWMOMHOC122)](https://pvvx.github.io/MHO_C122)
- [Tuya ZigBee TS0201 TZ3000](https://github.com/pvvx/BLE_THSensor)
- [Tuya ZigBee TH03Z](https://github.com/pvvx/BLE_THSensor)

This device file supports all above variants (display + comfort, display and no display).
Note it requires ZigbeeTlc firmware v0.1.2.1 or higher.

### DIFF
--- a/src/devices/zigbeetlc.ts
+++ b/src/devices/zigbeetlc.ts
@@ -1,25 +1,31 @@
+// requires zigbee2mqtt v1.34+
+// external converter for ZigbeeTLc v0.1.2.1+ by pvvx
+// https://github.com/pvvx/ZigbeeTLc
+// based on external converter for devbis-Firmware
+// https://raw.githubusercontent.com/devbis/z03mmc/master/converters/lywsd03mmc.js
+
 import {Zcl} from 'zigbee-herdsman';
-import * as exposes from '../lib/exposes';
-import fz from '../converters/fromZigbee';
-import tz from '../converters/toZigbee';
+
 import * as types from '../lib/types';
 import * as constants from '../lib/constants';
 import * as reporting from '../lib/reporting';
-import * as ota from '../lib/ota';
-const e = exposes.presets;
-const ea = exposes.access;
 
 import {
+    battery,
     binary,
+    enumLookup,
+    humidity,
     numeric,
+    ota,
     quirkAddEndpointCluster,
+    temperature,
 } from '../lib/modernExtend';
 
 const NS = 'zhc:zigbeetlc';
 
 async function configure(device: types.Zh.Device, coordinatorEndpoint: types.Zh.Endpoint, definition: types.Definition) {
     const endpoint = device.getEndpoint(1);
-    const bindClusters = ['msTemperatureMeasurement', 'msRelativeHumidity', 'genPowerCfg', 'genPollCtrl'];
+    const bindClusters = ['msTemperatureMeasurement', 'msRelativeHumidity', 'genPowerCfg', 'genPollCtrl', 'genOta'];
 
     try {
         await reporting.bind(endpoint, coordinatorEndpoint, bindClusters);
@@ -35,20 +41,6 @@ async function configure(device: types.Zh.Device, coordinatorEndpoint: types.Zh.
 }
 
 const extend = {
-    endpointQuirk: quirkAddEndpointCluster({
-        endpointID: 1,
-        outputClusters: [
-            'genOta',
-        ],
-        inputClusters: [
-            'genBasic',
-            'genPowerCfg',
-            'genIdentify',
-            'hvacUserInterfaceCfg',
-            'msTemperatureMeasurement',
-            'msRelativeHumidity',
-        ],
-    }),
     comfortDisplay: binary({
         name: 'comfort_display',
         valueOn: ['show', 0],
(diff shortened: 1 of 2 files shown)


src/devices/zigbeetlc.ts
```diff
@@ -0,0 +1,246 @@
...
+            {modelID: 'TS0201-bz', manufacturerName: 'Tuya', model: 'TS0201'},
+            {modelID: 'TH03Z-z', manufacturerName: 'Tuya', model: 'TH03Z'},
+            {modelID: 'TH03Z-bz', manufacturerName: 'Tuya', model: 'TH03Z'},
+        ],
+        model: 'WSD500A',
+        vendor: 'Tuya',
+        description: 'Temperature & Humidity Sensor (pvxx/ZigbeeTLc)',
+        fromZigbee: [fz.temperature, fz.humidity, fz.battery],
+        toZigbee: [],
+        exposes: [e.battery(), e.temperature(), e.humidity()],
```
Koenkk, line 233: Please use the `modernExtend` version instead. I guess you did not use them because some of the attributes might fail to setup? What is the error thrown in this case? I think we should add a property `allowError` to the modernExtends to be able to ignore it.

src/devices/zigbeetlc.ts
```diff
@@ -0,0 +1,246 @@
...
+        configure: configure,
+    },
+    /*
+        ZigbeeTLc devices supporting:
+        - Temperature (+calibration)
+        - Humidity (+calibration)
+    */
+    {
+        fingerprint: [
+            {modelID: 'TS0201-z', manufacturerName: 'Tuya', model: 'TS0201'},
```
Koenkk, line 223: `model` is not used, please remove from all

src/devices/zigbeetlc.ts
```diff
@@ -0,0 +1,246 @@
...
+                .withDescription('The temperature unit displayed on the screen'),
+        ],
+        extend: [
+            extend.endpointQuirk,
+            extend.tempCalibration,
+            extend.humidityCalibration,
+            extend.display,
+            extend.measurementInterval,
+        ],
+        ota: ota.zigbeeOTA,
```
Koenkk, line 213: Please use modernExtend `ota` instead.

robvanoostenrijk:
Updated taking into account the review feedback.

src/devices/zigbeetlc.ts
```diff
@@ -0,0 +1,253 @@
...
+    humidity,
+    numeric,
+    ota,
+    quirkAddEndpointCluster,
+    temperature,
+} from '../lib/modernExtend';
+
+const NS = 'zhc:zigbeetlc';
+
+async function configure(device: types.Zh.Device, coordinatorEndpoint: types.Zh.Endpoint, definition: types.Definition) {
```
Koenkk, line 26: Can this one be removed now?

robvanoostenrijk: This has now been removed as suggested. Device is working as expected.

src/devices/zigbeetlc.ts
```diff
@@ -0,0 +1,227 @@
...
+        - Humidity (+calibration)
+    */
+    {
+        fingerprint: [
+            {modelID: 'TS0201-z', manufacturerName: 'Tuya'},
+            {modelID: 'TS0201-bz', manufacturerName: 'Tuya'},
+            {modelID: 'TH03Z-z', manufacturerName: 'Tuya'},
+            {modelID: 'TH03Z-bz', manufacturerName: 'Tuya'},
+        ],
+        model: 'WSD500A',
```
Koenkk, line 210:
```suggestion
        model: 'WSD500A_1',
```

`WSD500A` is already in use so I guess this makes the most sense.

robvanoostenrijk: This a custom firmware for exactly those devices. The ZigbeeTLc firmware vastly improves the devices performance and reporting characteristics.

The same logic applies to the LYWSD03MMC model in the converter as well.

Keeping the above in mind, would you still like to have a model suffix for the same physical device but with different firmware? The Zigbee IDs are suffixed with (-z and -bz) for the custom firmware compared to the default already.

Koenkk: Yes, because the `model` should be unique for every device.

robvanoostenrijk: I've updated the duplicate models with custom ZigbeeTLc support to use an unique model name (suffix -z, same as the ZigbeeModel)

Koenkk:
Thanks!

pvvx:
> Yes, because the model should be unique for every device.

All firmware v0.1.2.1 supports changing the device name by the user.
How to work with Z2M with two devices with the same name and different FW versions and functions? This is also found in Tuya variants.

Each Zigbee device has a firmware version number, HW number, and manufacturer number.
Changing the device name does not create any problems with Zigbee OTA and working with software that supports Zigbee 3.0. Binding and working on the Zigbee network occurs by device MAC, and not by name.

IamHEREfromU:
my xiaomi LYWSD03MMC have the FM version 2.1.1_0159 how can i flash to the Firmware to run on my HA? bcs now the HA ask for a Blindkey and i dont have it!

Bodengriller:
I think there's a [little bug in converter-file](https://github.com/pvvx/ZigbeeTLc/issues/106)

davidsarkany:
@robvanoostenrijk Can you add LKTMZL02 support?

PR #7460, opened 2024-04-28
add device TICMeter

xmow49:
Add the device "TICMeter" converter and the OTA.


src/devices/gmmts.ts
```diff
@@ -0,0 +1,848 @@
+import {Definition, Fz, Tz, KeyValue, Zh} from '../lib/types';
+/* eslint-disable camelcase */
+/* eslint-disable max-len */
```
Koenkk, line 3: Please don't do any `eslint-disable` here

src/devices/gmmts.ts
```diff
@@ -0,0 +1,848 @@
...
+    STD: 'STANDARD',
+};
+
+const mode_tic_enum = ['HISTORIQUE', 'STANDARD', 'AUTO'];
+const mode_elec_enum = ['MONOPHASE', 'TRIPHASE', 'AUTO'];
+const mode_contract_enum = ['AUTO', 'BASE', 'HCHP', 'EJP', 'TEMPO', 'PRODUCTEUR'];
+
+// prettier-ignore
+const ticmeter_options = [
+    e.numeric(`refresh_rate`, ea.SET).withValueMin(60).withDescription(`Temps d'actualisation des valeurs statiques (celles qui possèdent des boutons refresh). Par défaut: ${DEFAULT_POLL_INTERVAL} s`).withValueMin(60).withValueMax(3600),
```
Koenkk, line 61: There is a lot of French here, please use English instead.

src/devices/gmmts.ts
```diff
@@ -0,0 +1,848 @@
+import {Definition, Fz, Tz, KeyValue, Zh} from '../lib/types';
```
Koenkk, line 1: This is quite some code to support just a single device, would it be possible to simply somehow by relying on standard functionality like reporting?

xmow49: This device is a smart energy monitor, witch is pluged to a electric meter and read data.
Depending of the electricity contract, commincation mode, number of phase, eletricity producer or not, entitys appears or not.
I already use reporting for somes attributes.
I use custom fz for haElectricalMeasurement beacause the device dont have all attributes available.

I dont know how to reduce the code.

Thanks

src/devices/gmmts.ts
xmow49: This product is intended only for the French market (French electric meter). Is it possible to add translations?

src/devices/gmmts.ts
xmow49: Where I can use it ?

Koenkk: Just remove it

src/devices/gmmts.ts
Koenkk: Translations are not support for this, but please translate it to English (ZLinky_TIC is also in English)

src/devices/gmmts.ts
xmow49: Hi,
If I remove them, I have a lot of lint errors:
image

If I remove spaces; and cut the line at 150 characters; the code becomes very ugly.
Can I keep at least 
/* eslint-disable max-len */
/* eslint-disable no-multi-spaces */ 
?
Thanks

Koenkk: Then let's keep those

src/devices/gmmts.ts
xmow49: I added an option to change the language.

src/devices/gmmts.ts
xmow49: I edited some fzlocal to reduce the code size.

PR #7469, opened 2024-04-30
Add Tuya 2 gang switch

nakata5321:
Hello!

Add new switch tuya brand. This switch is already recognized as ["TS0002" switch](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts#L2839) but this converter doesn't use all available features. And all already added similar switches doesn't have available features too.  For example [this switch](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts#L2799)
Since I can't check different, already added models, I've decide to add new switch.

If there are any options how to add this new switch to existed with full support of available features, Please provide the idea how to do it and I will update PD.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6426,6 +6426,24 @@ const definitions: Definition[] = [
     },
     {
         fingerprint: [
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #7473, opened 2024-04-30
Fix Eurotronic Zigbee Spirit system mode/host flags

mzanetti:
I wanted to debug why the child_lock option wasn't showing up neither in home assistant nor in the z2m web interface despite seeing traces of it in the logs (child_protection) and also seeing related code. Along the way I noticed that the entire host flags attribute was rather buggy. Setting thermostat mode somewhat worked, but it would frequently report back the wrong state upon attribute reports coming in. Granted, there's a bug in the firmware of those devices which makes parsing of it quite confusing. I've added a comment in the code about that. It seems extremely unlikely that there will be a firmware update changing/fixing that, so I believe my workaround should be safe to use (better than the previous state anyways).

Also it seemed to have a legacy converter which wasn't really used at all any more, which I've cleaned up.

While being at it, I also added the mirror_display option which comes in handy occasionally for those devices.

I am aware that this is not 100% backward compatible (the state object doesn't have window_open and child_protection any more), however, given those were also suffering from the before mentioned bugs and also not properly exposed, I assumed that may be okay... Please let me know if it is not and I can add them back.

The relevant parts (the "exposes") should be backwards compatible and now work properly.

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -4921,21 +4921,22 @@ const converters2 = {
         convert: async (model, msg, publish, options, meta) => {
             const result = await converters1.thermostat.convert(model, msg, publish, options, meta);
             if (result) {
-                // system_mode is always 'heat', we set it below based on eurotronic_host_flags
-                delete result['system_mode'];
-
                 if (typeof msg.data[0x4003] == 'number') {
                     result.current_heating_setpoint = precisionRound(msg.data[0x4003], 2) / 100;
                 }
                 if (typeof msg.data[0x4008] == 'number') {
-                    result.child_protection = (msg.data[0x4008] & (1 << 7)) != 0;
-                    result.mirror_display = (msg.data[0x4008] & (1 << 1)) != 0;
-                    result.boost = (msg.data[0x4008] & 1 << 2) != 0;
-                    result.window_open = (msg.data[0x4008] & (1 << 4)) != 0;
-
-                    if (result.boost) result.system_mode = constants.thermostatSystemModes[4];
-                    else if (result.window_open) result.system_mode = constants.thermostatSystemModes[0];
-                    else result.system_mode = constants.thermostatSystemModes[1];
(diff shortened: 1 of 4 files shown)


mzanetti:
FWIW, I've just tested this with the [Eurotronic Zigbee Comet](https://eurotronic.org/wp-content/uploads/2023/04/BDA-Comet-Zigbee-DE-2023-04-06.pdf) (Spirits successor) too. It works just fine with the same code as the Zigbee Spirit. Also reports as model SPZB0001 although I got this warning during the network joining:

```
Zigbee2MQTT:warn  2024-05-01 20:36:40: Images currently unavailable for device 'SPZB0001', hardwareVersion '34', manufacturerName Eurotronic, {}'
```

Not sure where to add the image though...

Koenkk:
Thanks!

raveit65:
I am also using the spirit zigbee TVR, but i can't really test the change because summertime starts now in my country.
Is it not bit dangerous to push something like that outside the heating season?
I hope the TVRs do work in october.



> FWIW, I've just tested this with the [Eurotronic Zigbee Comet](https://eurotronic.org/wp-content/uploads/2023/04/BDA-Comet-Zigbee-DE-2023-04-06.pdf) (Spirits successor) too. It works just fine with the same code as the Zigbee Spirit. Also reports as model SPZB0001 although I got this warning during the network joining:
> 
> ```
> Zigbee2MQTT:warn  2024-05-01 20:36:40: Images currently unavailable for device 'SPZB0001', hardwareVersion '34', manufacturerName Eurotronic, {}'
> ```
> 
> Not sure where to add the image though... For now it will show up with the Spirits image.

I can confirm that firmware reading issue with comet zigbee TVR.

raveit65:
> Granted, there's a bug in the firmware of those devices which makes parsing of it quite confusing. 

I just want to note that there are 2 different firmware versions around.

a) Firmware-Date :  20190408
Firmware-Version:  18181120
Which displays temperature like this `22.82°C`

b) Firmware-Date:  20191014
Firmware-Version: 22190930
Which displays temperature like `22.5°C` (only half degree)

PR #7475, opened 2024-05-01
Add countdown to tuyaOnOff

schauveau:
This is an implementation of the Tuya feature described in https://github.com/Koenkk/zigbee2mqtt/issues/22182

The [Tuya Zigbee Connectivity documentation](https://developer.tuya.com/en/docs/connect-subdevices-to-gateways/Zigbee_2?id=Kcww7qppbe87m)  specifies that some categories of device (I believe Smart switch and Smart Metering Socket) implement a countdown by using the standard onWithTimeOff command but in a non-standard way.

This implementation adds a countdown attribute that triggers a toggle in the future. 
- The countdown is given in seconds (while the standard `onWithTimeOff` command expects 1/10th seconds). 
- A running countdown can be cancelled by setting countdown to 0 or by setting the state (even if it does not change). 
- The new Tz converter handles both 'state' and 'countdown' to insure that the commands are sent in the proper order so first `state` then `countdown`. Setting both attributes at once still requires that 2 commands are sent so there is still a risk that only one of them will take effect (or can they be reordered???). 
    - QUESTION: Is there a way to send two Zigbee commands in the same message? 
-  The device report countdown at every multiple of 60 so once per minute while a countdown is in progress. More reporting can be configured using cluster `genOnOff` and attribute `onTime`.   

Examples:

- `/set {"countdown": 100}` to  toggle in 100s 
- `/set {"state" :"toggle", "countdown": 100}`  to  toggle now and in 100s
- `/set {"state" :"off", "countdown": 100}`  to turn OFF immediately and turn ON in 100s
- `/set {"countdown": 0}` to cancel the current countdown if any

In theory, the implementation should support multiple end-points but that was not tested.

The countdown is only enabled on the few devices I own but that should work for a lot more.


src/lib/tuya.ts
```diff
@@ -1187,6 +1187,61 @@ const tuyaTz = {
             return {state: {do_not_disturb: value}};
         },
     } satisfies Tz.Converter,
+    on_off_countdown: {
+        // Note: This is the Tuya on-off countdown feature documented for switches and smart plugs
+        //       using the Zigbee 'onWithTimedOff' command in a non-standard way.
```
Koenkk, line 1192: What exactly is non-standard here?

schauveau: The standard behavior is described in the Zigbee standard. See sections **3.8.2.3.6 On With Timed Off Command** and **3.8.2.4 State Description ** in the [ZCL](https://zigbeealliance.org/wp-content/uploads/2019/12/07-5123-06-zigbee-cluster-library-specification.pdf). To summarize,  the standard `onWithTimedOff` behavior is `TURN ON, WAIT on_time/10 SECONDS AND THEN TURN OFF`.  The register `off_wait_time` can be used to add an additional delay during which the device cannot be turned on. This is why the current implementation of `on_off` in `converters/Zigbee.ts` only allows 'on_time' to be set with 'state="on"'. 

The Tuya behavior is a lot more generic : "WAIT on_time SECONDS THEN TOGGLE" so a delayed toggle. There is no restriction on the initial state and `on_time` is now given in seconds.

A device that uses the default 'on_off' converter can be tested by turning it off and then `/set { state:"on", "on_time": 6 }`. Because of the scaling by 10 in 'on_off' converter, the value 60 is written in the register `on_time`.

- If the device turns ON immediately and OFF after 6 seconds then the device is using the standard Zigbee behavior described in the ZCL  (probably because I only have Tuya devices)
- If the device stays OFF and turns ON after 60 seconds then the device is using the Tuya countdown behavior.

PR #7476, opened 2024-05-01
fix(lumi): correct attribute name for Aqara curtain hand open

aurimasniekis:
The attribute name for Aqara curtain hand open functionality in the lumi.ts file has been corrected to align with correct value in zigbee-herdsman cluster name. Tried out in external converter and it works okay.

https://github.com/Koenkk/zigbee-herdsman/commit/68d97d81ce32ca257483348bbb7d56bb45178a3c#diff-ec1a2d43181075d54fd33395e767a90a8d5e66cc2778f3944130d6aa537992b7R4641


Koenkk:
Thanks!

PR #7479, opened 2024-05-02
Add `colorloop` effect for color lights

slugzero:
When I tried to set up some effects for my color lights, I realized that there is no colorloop in the list of supported effects for most of them. Apparently, I am not the first one who is missing this (cf. https://github.com/Koenkk/zigbee2mqtt/issues/1791). So here's my proposal for the implementation. 

The change allows to trigger a continuous color loop with the command `zigbee2mqtt/FRIENDLY_NAME/set` with payload `{"effect": colorloop}`. This is in addition to the already-existing command `hue_move` and basically does the same.

As proposed in the feature request, it also accepts a `transition` value from 1...255 to set the time for a full loop, where 1 is shortest and 255 ist longest. I tested it with my lights and it roughly matches the seconds for a full loop, but this does not need to be the case for every light. 
If no value for `transition` is given, a default value of 15 is used.

Example: for the slowest-possible color loop, publish to `zigbee2mqtt/FRIENDLY_NAME/set` with payload `{"transition": 255, "effect":"colorloop"}`. To turn off the color loop, publish `{"effect":"stop_effect"}`.

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -754,7 +754,11 @@ export function light(args?: LightArgs): ModernExtend {
     const exposes: Expose[] = lightExpose;
 
     if (args.effect) {
-        exposes.push(e.effect());
+        const effects = e.effect();
+        if (args.color) {
+            effects.values.push('colorloop');
+        }
+        exposes.push(effects);
         toZigbee.push(tz.effect);
     }
 
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -756,7 +756,7 @@ export function light(args?: LightArgs): ModernExtend {
     if (args.effect) {
         const effects = e.effect();
         if (args.color) {
-            effects.values.push('colorloop');
+            effects.values.push('colorloop', 'stop_colorloop');
         }
         exposes.push(effects);
         toZigbee.push(tz.effect);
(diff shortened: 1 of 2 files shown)


src/lib/modernExtend.ts
```diff
@@ -754,7 +754,11 @@ export function light(args?: LightArgs): ModernExtend {
     const exposes: Expose[] = lightExpose;
 
     if (args.effect) {
-        exposes.push(e.effect());
+        const effects = e.effect();
+        if (args.color) {
+            effects.values.push('colorloop');
```
Koenkk, line 759:
```suggestion
            effects.values.push('start_colorloop', 'stop_colorloop`);
```

`stop_effect` should definitely be here, but since this only stops the color loop, I propose it to rename to `stop_colorloop`

slugzero: Thank you for your review.
Not sure if I got you right: I thought it would be best to extend the existing `stop_effect` (which is added in [`e.effect()`](https://github.com/slugzero/zigbee-herdsman-converters/blob/677035f39a350b2c1f3f7c18ca5c072ccde7df0d/src/lib/exposes.ts#L678)) to also stop the colorloop. 
So you would suggest to rather add a separate `stop_colorloop` which only stops the color loop?

slugzero: You are right, the additional command is also called for lights that don't support colors and fails. I could check with `supportsInputCluster` if the color control cluster is present, but not sure how this behaves e.g. for groups or if there is still a regression for individual devices. 
So I think it is safest to have a separate `stop_colorloop` command as you suggested. Note that I did not change `colorloop` to `start_colorloop` in order to be consistent with Philips devices (where it is called `colorloop` only), and also to be consistent with other Zigbee Gateways (Deconz, zha). If you prefer `start_colorloop`, lmk

Koenkk:
Thanks!

PR #7484, opened 2024-05-04
Improve support for Bosch  BTH-RM230Z

### LINKED ISSUE #7357: BTH-RM230Z - Missing Cooling Mode
The same goes for the **BTH-RM - Room thermostat II (Battery model)**. I don't have one but looking at the Manual it also should have a Cooling Mode.

I got the Bosch Room Thermostat II (230V) and I noticed there's no option for switching to cooling mode currently.
When factory resetting the Device and setting it up without a connection, I can choose between Cooling or Heating on the Device, but it doesn't expose the Mode nor does it let me switch it in Z2MQTT.

I can see the correct running state & cooling setpoint when changing it on the Device (since I can't change it in Z2MQTT)

    "occupied_cooling_setpoint": 21,
    "occupied_heating_setpoint": 22.5,
    "pi_heating_demand": 100,
    "running_state": "cool",
    "setpoint_change_source": "schedule",
    "system_mode": "auto",

EDIT: To give a bit more context (if needed)
There are a few Different Operation Modes and how you can run it.

A short excerpt from the Manual

"In the event that an initial start-up is required without 
a Smart Home system (e.g. new build), basic settings 
can be made on the device via an installer mode. 
To this end, press and hold the rotary knob for 
at least 3 seconds until the display shows “INS”. By 
turning the knob you can make a selection and 
confirm the selection by pressing it."

image

Once you manually set it, you're locked in and can only change the temperature (Basically). With the Bosch Smart Home Controller, you can change between the Operation Modes (not necessarily relevant here) and between Cooling/Heating, which unfortunately is not possible on the device itself. (Only has Status LEDs)

burmistrzak:
Improve support for the Bosch thermostat by utilizing more default clusters and attributes.

Includes:

- Introduce `operating_mode` [ schedule | manual | pause ] _(manufacturer-specific)_
- Properly expose `system_mode`
- Add `cool` mode
- Remove `auto` from `system_mode` _(not supported by device)_
- Convert to `modernExtend`

~~There's currently some duplicated code in `bosch.ts`, but until I get to the Bosch TRV, we'll unfortunately have to live with that fact. 😅~~

Fixes #7357
Relevant #6995


src/devices/bosch.ts
```diff
@@ -551,6 +551,79 @@ const tzLocal = {
             }
         },
     } satisfies Tz.Converter,
+    bosch_room_thermostat: {
```
Koenkk, line 554: Could you convert this to modernExtend, some examples: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/sonoff.ts#L831

This replace all the toZigbee/fromZigbee and exposes code.

Koenkk: > Has the cluster name in deviceAddCustomCluster to be unique (per device, globally)? 

Unique per device

>  Can modernExtend partially override/extend a default cluster with manufacturer-specific attributes? E.g.: 0x4007 in hvacThermostat for operating_mode?

Yes! Just use the existing cluster name for the `name` parameter.

Your PR looks good, can I merge it?

Koenkk: Looks good, thanks!

Koenkk:
Thanks!

Koenkk:
Could you do a `console.log(device.customClusters)` and see if the clusters are there? Could be that there is a race condition

Koenkk:
The manufacturer code is missing from the custom cluster definition, try with: `displayOntime: {ID: 0x403a, type: Zcl.DataType.ENUM8, manufacturerCode: Zcl.ManufacturerCode.ROBERT_BOSCH_GMBH},`

Koenkk:
No, zigbee-herdsman should use it automatically

Koenkk:
Yes, you can take Ubisys as an example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/597b7d3e716e3bec81b77b3f3115b88da84502cf/src/lib/ubisys.ts#L9

PR #7494, opened 2024-05-06
Add IHC8223AL

65698888:
IHC8223AL(z2m用)


65698888:
I refer to the writing method in tuya.ts, why can't mine pass.

65698888:
Was it in JavaScript format before and not modified to TypeScript format?

src/devices/honyar.ts
```diff
@@ -128,6 +181,36 @@ const definitions: Definition[] = [
             await reporting.onOff(endpoint3);
         },
     },
+	{
+		zigbeeModel: ['000a0abb\u0000', 'RH5000_SmartOutlet'],
+		model: 'IHC8223AL',
+		vendor: 'Honyar',
+		description: 'Smart Power Socket 10A (with power monitoring)',
+		fromZigbee: [fz.on_off, fzLocal.honyer_electrical_measurement, fzLocal.honyer_metering],
```
Koenkk, line 189: Could you use modernExtend instead here? Example:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/3553ddedf8cf4ae3751df924b7ea7c0babefacfd/src/devices/adeo.ts#L297

65698888: Continue to try to modify, this time there is no error. Please help to check if there are any mistakes.

Koenkk: I don't see the `extend` in this PR yet

65698888: Can't I do it the way I'm doing it?

Koenkk: No, only modernExtend is accepted since this reduces code.

65698888: I use modernExtend for the first time, but I am not sure whether it is successful. Please help to check it.

src/devices/honyar.ts
```diff
@@ -128,6 +129,38 @@ const definitions: Definition[] = [
...
+    {
+        zigbeeModel: ['000a0abb\u0000', 'RH5000_SmartOutlet'],
+        model: 'IHC8223AL',
+        vendor: 'Honyar',
+        description: 'Smart Power Socket 10A (with power monitoring)',
+        extend: [
+            onOff({powerOnBehavior: false}),
+            electricityMeter({cluster: 'honyar'}),
+        ],
+        onEvent: async (type, data, device) => {
```
Koenkk, line 141: Can you try without this? I hope this device support reporting so polling is not needed.

65698888: I've already tried. Proactively report a problem, you have to proactively poll.

src/converters/fromZigbee.ts
```diff
@@ -5140,6 +5140,54 @@ const converters2 = {
...
+    honyer_electrical_measurement: {
+        cluster: 'haElectricalMeasurement',
+        type: ['attributeReport', 'readResponse'],
+        options: [
+            exposes.options.calibration('power', 'percentual'), exposes.options.precision('power'),
+            exposes.options.calibration('current', 'percentual'), exposes.options.precision('current'),
+            exposes.options.calibration('voltage', 'percentual'), exposes.options.precision('voltage'),
+        ],
+        convert: (model, msg, publish, options, meta) => {
+            if (meta.device.dateCode === '20170621') {
```
Koenkk, line 5169: If I understand correctly, there is a bug for this particular firmware which has the wrong multiplier/divisors?

65698888: Yes, I tried the following method, but it didn't work.

electricityMeter({current: {divisor: 1000}, voltage: {divisor: 1}, power: {divisor: 1}}),

Koenkk: Did you try to do a reconfigure afterwards? This should work (use the yellow refresh icon on the frontend to do so)

65698888: Thanks for your guidance, the display is normal after using the yellow refresh button on the front.

src/converters/fromZigbee.ts
```diff
@@ -5140,6 +5140,23 @@ const converters2 = {
             return payload;
         },
     } satisfies Fz.Converter,
+    honyer_metering: {
```
Koenkk, line 5143: Can this one also be removed?

src/devices/honyar.ts
```diff
@@ -128,6 +129,38 @@ const definitions: Definition[] = [
...
+        vendor: 'Honyar',
+        description: 'Smart Power Socket 10A (with power monitoring)',
+        extend: [
+            onOff({powerOnBehavior: false}),
+            electricityMeter({cluster: 'honyar', current: {divisor: 1000}, voltage: {divisor: 1}, power: {divisor: 1}}),
+        ],
+        onEvent: async (type, data, device) => {
+            device.skipDefaultResponse = true;
+            const Endpoint = device.getEndpoint(1);
+            if (Endpoint == null) {
```
Koenkk, line 144: When is this `null`? I wouldn't expect this check here.

src/devices/honyar.ts
```diff
@@ -128,6 +129,38 @@ const definitions: Definition[] = [
...
+            } else if (!globalStore.hasValue(device, 'interval')) {
+                const interval = setInterval(async () => {
+                    try {
+                        await Endpoint.read('haElectricalMeasurement', ['activePower', 'rmsCurrent', 'rmsVoltage']);
+                        await Endpoint.read('seMetering', ['currentSummDelivered']);
+                        await Endpoint.read('genOnOff', ['onOff']);
+                    } catch (error) {
+                        // Do nothing
+                    }
+                }, 5*1000);// Every 5 seconds
```
Koenkk, line 159: Every 5 seconds is too much, default it to 60 seconds and make it configurable (search for `measurement_poll_interval` in this repo for examples)

src/devices/honyar.ts
Koenkk: Can you try reusing existing polling code?

Try with:

```suggestion
        options: [exposes.options.measurement_poll_interval()],
        onEvent: (type, data, device, options) => tuya.onEventMeasurementPoll(type, data, device, options),
```

src/converters/fromZigbee.ts
Koenkk: @65698888 could you check if this one can be removed?

65698888:
@Koenkk Hello, is there anything that needs to be modified?

Koenkk:
Yes, please check [this comment](https://github.com/Koenkk/zigbee-herdsman-converters/pull/7494/files#r1616323967)

PR #7497, opened 2024-05-08
Fix Backlight for TS0013

piwi3910:
I noticed for my TS0013 switches the backlight converting didn't work like it did work for the 1 gang and 2 gang switches. 
I noticed the wrong payload is used as the switches don't change the brightness, but do change the inversion.
I modified the code with the right payload.


Koenkk:
Thanks!

PR #7520, opened 2024-05-13
Add shared `modernExtend` for Bosch BTH-*

burmistrzak:
Remove lots of duplicated code by using `modernExtend` wherever possible, and sharing common extends across Bosch thermostat models.


Koenkk:
> What do you think about my general approach here?

Perfect, removes a lot of duplication, many thanks!

PR #7526, opened 2024-05-16
feat(yale): add support for YRD410 physical button variant

JefeDavis:
currently that Physical button variant of the yale assure 2 (yrd410-pb) model shows unsupported, this change add it to the existing definition for the yrd410-ble


Koenkk:
Thanks!

bennettrogers:
I have my YRD410-PB connected to Z2M, but it's still showing as unsupported even with this change. (Screenshot attached). Is there something else I need to change to get it to be recognized by the converter?

image

PR #7529, opened 2024-05-17
Fix ID checks for scene/group

Nerivec:
From [Zigbee spec](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf):

> 3.7.2.4.1 Generic Usage Notes
Scene identifier 0x00, along with group identifier 0x0000, is reserved for the global scene used by the OnOff cluster.

Since [Philips devices](https://github.com/Koenkk/zigbee2mqtt/issues/22231) appear to require usage of scene ID zero with non-zero group IDs, I'm assuming the spec has to be interpreted that way, and not just "prevent usage of scene ID zero".

_Also got rid of a few more `hasOwnProperty` that don't check for undefined._

### DIFF
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -3542,7 +3542,7 @@ const converters2 = {
         key: ['scene_store'],
         convertSet: async (entity, key, value: KeyValueAny, meta) => {
             const isGroup = utils.isGroup(entity);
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #7539, opened 2024-05-20
Add support for "Immax NEO Smart irrigation valve with solar panel"

bmansvk:
This is maybe just new "firmware" for currently supported [QOTO Valve](https://www.zigbee2mqtt.io/devices/QT-05M.html)

See [Immax NEO Smart irrigation valve with solar panel](https://www.immax.eu/immax-neo-smart-irrigation-valve-with-solar-panel-soil-moisture-sensor-zigbee-p18882/)

It works for me.

### DIFF
--- a/src/devices/qoto.ts
+++ b/src/devices/qoto.ts
@@ -8,7 +8,7 @@ const ea = exposes.access;
 
 const definitions: Definition[] = [
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #7542, opened 2024-05-20
Add support for the Gewiss Smoke Sensor (Develco rebranded)

mzanetti:
https://www.gewiss.com/ch/de/products/product.1000000.1008169.GWA1512


Koenkk:
thanks!

PR #7548, opened 2024-05-22
Convert Bosch BSD-2 to `modernExtend`

burmistrzak:
Do I have to say more? 😇


Koenkk:
Thanks!

Koenkk:
> @Koenkk I guess, we'll have to soon add a `src/lib/bosch.ts` to keep things nice & tidy? 😅

makes sense!

Koenkk:
@burmistrzak It's a hard one indeed, I suggest to leave it as is.

PR #7555, opened 2024-05-24
Siemens RDZ100ZB

VladislavVesely:
Hello,

i'm not sure how this procedure should work. All files here are in TS, but based on https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_2-2-adding-converter-s-for-your-device i should be using JS and in fact trying TS did not work as external converter for me at all.

### DIFF
--- a/src/devices/RDZ100ZB.js
+++ b/src/devices/RDZ100ZB.js
@@ -1,0 +1,194 @@
+const {identify, battery, ota} = require('zigbee-herdsman-converters/lib/modernExtend');
+const exposes = require('zigbee-herdsman-converters/lib/exposes');
+const reporting  = require('zigbee-herdsman-converters/lib/reporting');
+const constants  = require('zigbee-herdsman-converters/lib/constants');
+const utils  = require('zigbee-herdsman-converters/lib/utils');
+const tz = require('zigbee-herdsman-converters/converters/toZigbee');
+const {precisionRound, postfixWithEndpointName} = require('zigbee-herdsman-converters/lib/utils');
+const e = exposes.presets;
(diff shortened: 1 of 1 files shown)


src/devices/RDZ100ZB.js
```diff
@@ -0,0 +1,194 @@
...
+    model: 'RDZ100',
+    vendor: 'Siemens',
+    description: 'Siemens RDZ100 Wireless Thermostat',
+    fromZigbee: [
+        siemensFromZigbee,
+    ],
+    toZigbee: [
+	tz_local_temperature,
+	tz_control_sequence_of_operation,
+	tz_system_mode,
```
Koenkk, line 165: Can you switch to the standard converters, e.g. `tz.thermostat_system_mode`?

VladislavVesely: What about the timeout then? That device accept command one per minute, when i tried timeout value in top layer meta it does nothing and i do not see possibility to set it otherwise. Is there some better way to wait for incomming event first? Btw also would it be possible to move description of all generic stuff to optional properties with default value of current texts? (I can do that in different pull request)

Koenkk: > Is there some better way to wait for incomming event first? 

I've replied on the discussion

> Btw also would it be possible to move description of all generic stuff to optional properties with default value of current texts? (I can do that in different pull request)

Do you mean this one?

```
expose.push(e.enum('control_sequence_of_operation', ea.STATE_SET, Object.values(siemensOperationModes)).withDescription('Operation mode'));
```

PR #7557, opened 2024-05-25
Add missing `opal` and `glisten` effects to Hue Festavia LCX015

mundschenk-at:
- Adds the missing `opal` and `glisten` effects to the 2nd-gen Hue Festavia 250 (`LCX015`).
- Re-orders the various Hue Festavia versions to make it harder to miss anything in future adaptions.
- Marks the 1st-gen Hue Festavia 250 as such.


Koenkk:
thanks!

PR #7561, opened 2024-05-26
Enable OTA for some Develco devices

samuli-as:
Enables the OTA attribute for some devices which now have images added in https://github.com/Koenkk/zigbee-OTA/pull/502.
The humidity sensor and water leak sensor I have tested. I believe I picked the correct devices for the smart button and smart plug but I do not have them at hand to verify.

One thing I noticed after updating some devices that they were reporting every 30 seconds for some reason even though their reported values had not changed. This interval was not reflected in the Z2M reporting settings. Resetting them made them behave themselves. Any ideas where to look? I have some sensors left unpatched for further testing.

Edit: I wonder if this is related: https://github.com/Koenkk/zigbee2mqtt/issues/21669

### DIFF
--- a/src/devices/develco.ts
+++ b/src/devices/develco.ts
@@ -525,6 +525,7 @@ const definitions: Definition[] = [
         description: 'Power plug',
         fromZigbee: [fz.on_off, develco.fz.electrical_measurement, develco.fz.metering],
         toZigbee: [tz.on_off],
+        ota: ota.zigbeeOTA,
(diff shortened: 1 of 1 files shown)


mamrai1:
If you are using ember adapter that is probably the case... I already replaced the batteries in both the develco heat alarms and changed the adapter to ezsp.

Koenkk:
Can this be merged?

Koenkk:
Thanks!

PR #7562, opened 2024-05-26
Fix Vindstyrka OTA

samuli-as:
IKEA Vindstyrka OTA relies on zigbeeOTA as its images are not in the same IKEA first party repository as the rest of them. It was seemingly inadvertently changed to use ikeaOta as well in the modern extend refactoring, causing a no images available error.


Koenkk:
Thanks!

PR #7568, opened 2024-05-28
add EARU EAYCB-Z-2P with leak protection

Perfect-Web:


Koenkk:
Thanks!

stefangries:
Why is this device named "with leak protection"? It does not expose something, that indicates that the device has this feature. AFAIK there is no such device with water leakage protection.

PR #7573, opened 2024-05-29
Set entity category to configuration for Sonoff TRVZB open/close degree entities

### LINKED ISSUE #7564: Sonoff TRVZB attributes are not marked with device_category as configuration entities
In the recent enhancements to support custom cluster and new attributes for the TRV ([closing and opening degree](https://github.com/Koenkk/zigbee-herdsman-converters/pull/7130)) these are not marked as `device_category` = `config`

I've searched everywhere (I think!) particularly in [sonoff.ts](https://github.com/Koenkk/zigbee-herdsman-converters/pull/7130/files) but I can't see where this is defined. If I knew how to change this I would submit a PR.  For example, `local_temperature_calibration` does come through as config but I can't see how.

`valveOpeningDegree` & `valveClosingDegree` should be config and not therefore displayed in dashboards.

blacknell:
Closes #7564

Currently, the entity category is not set for open and close degree meaning the controls display in dashboards and elsewhere in HA. These should be configuration entities managed through admin UI.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -970,6 +970,7 @@ const definitions: Definition[] = [
                 name: 'valve_opening_degree',
                 cluster: 'customSonoffTrvzb',
                 attribute: 'valveOpeningDegree',
+                entityCategory: 'config',
                 description: 'Valve open position (percentage) control. ' +
                     'If the opening degree is set to 100%, the valve is fully open when it is opened. ' +
                     'If the opening degree is set to 0%, the valve is fully closed when it is opened, ' +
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

blacknell:
@Koenkk - I've since noticed a whole bunch more sensor entities that seem to me "useless" to the average end user. I think they should move to be `diagnostic` category rather than default `sensor` category. Their function isn't documented anywhere by Sonoff.

I have 14 TRV's so that's 56 entities that get added to auto dashboards which is silly.

If you're in broad agreement I'll create a fresh PR and submit that (move them to diagnostic).

They are:-
- Closing steps
- Valve closing limit voltage
- Valve motor running voltage
- Valve opening limit voltage

There's also a read-only sensor called `Frost Protection Temperature` that is by default disabled but listed as a `Sensor`. But this is superfluous as it just reflects (if enabled) the value of the writeable `control` entity of the same name. I could investigate this and find out why it's duplicated and remove.

Koenkk:
> If you're in broad agreement I'll create a fresh PR and submit that (move them to diagnostic).

Sounds good!

PR #7574, opened 2024-05-29
Remove last `manuSpecificBosch*` usage

burmistrzak:
While some conversions to `modernExtend` are still quite basic, we at least can now finally remove `manuSpecificBosch*` from ZCL. 😎


Koenkk:
thanks!

PR #7575, opened 2024-05-30
Expose voltage for WISZB-120

kranzj:
Context: 'voltage' is in state, but not exposed. I'm using ni-mh rechargeable batteries, so the percentage is not as useful.


Koenkk:
thanks!

kranzj:
Hello again! Just tried the new version; the change kind of works, but it has the wrong unit for the battery voltage and shows 2700V instead of 2700mV. I'm not sure how I configure the unit, could you give me a hint?

Koenkk:
Use `e.battery_voltage()` instead of `e.voltage()`, could you submit a PR for this?

kranzj:
Thanks a lot :). I created https://github.com/Koenkk/zigbee-herdsman-converters/pull/7597.

PR #7583, opened 2024-05-31
Enforce no floating promises with eslint.

Nerivec:
Fixed all related issues. I don't see a reason for any to not be awaited, but please review/confirm.
Also added simple `reject` for OTA to fit with the rule; might need some adjusting.

CC: @sjorge @mrskycriper


Koenkk:
Thanks!

PR #7585, opened 2024-05-31
Optimize OTA

Nerivec:
- Convert logic to async/await
- Add mock to test actual update logic
  - _for now local only, takes too long for CI, even with fake timers that bypass throttling_
- Move existing tests to TS
- Fix code styling, add typing, remove magic numbers for ZCL spec.

TODO:
- [x] Better mock/test to allow CI testing of update (see above)?
- [x] Test with actual device
- [x] Test with devices known to have OTA quirks (Legrand, Sonoff, Bosch...)


Nerivec:
Tests by @mundschenk-at (on network of ~80 devices)
- https://www.zigbee2mqtt.io/devices/929001953101.html
  - via coordinator
  - ~25min
- https://www.zigbee2mqtt.io/devices/929001953101.html
  - via router
  - ~30min

Notes:
- ETA appears to become pretty good around the 10% mark. A bit over-estimated before then.

src/lib/ota/common.ts
```diff
@@ -427,265 +508,242 @@ export async function isNewImageAvailable(current: Ota.ImageInfo, device: Zh.Dev
...
+        const imagePageRequest = endpoint.waitForCommand('genOta', 'imagePageRequest', null, imageBlockOrPageRequestTimeoutMs);
+        waiters.imageBlockOrPageRequest = {
+            promise: Promise.race([imageBlockRequest.promise, imagePageRequest.promise]),
+            cancel: () => {
+                imageBlockRequest.cancel();
+                imagePageRequest.cancel();
+            },
         };
 
-        const answerNextImageRequest = () => {
```
Koenkk, line 536: I see this code was removed without a replacement. I'm not sure why it was there though because the `requestOTA` is already done before it, so this would have never got the `queryNextImageRequest`

src/lib/ota/common.ts
```diff
@@ -427,265 +508,242 @@ export async function isNewImageAvailable(current: Ota.ImageInfo, device: Zh.Dev
...
+            let pageSize = 0;
+
+            const sendImageBlockResponse = async (imageBlockRequest: CommandResult) => {
+                // Reduce network congestion by throttling response if necessary
+                {
+                    const blockResponseTime = Date.now();
+                    const delay = (blockResponseTime - lastBlockResponseTime);
+
+                    if (delay < IMAGE_BLOCK_RESPONSE_DELAY) {
+                        await new Promise<void>((resolve) => setTimeout(resolve, IMAGE_BLOCK_RESPONSE_DELAY - delay));
```
Koenkk, line 590:
```suggestion
                        await utils.sleep(IMAGE_BLOCK_RESPONSE_DELAY - delay);
```

src/lib/ota/common.ts
Nerivec: `requestOTA` doesn't respond to `queryNextImageRequest`, which the device needs to continue and start asking for image blocks. That was duplicated with that function, but with the response. Now it's just called directly with `sendQueryNextImageResponse` after `getNewImage`.
_That was probably also responsible for the (sometimes) very bad ETA on start._

src/lib/ota/common.ts
Nerivec: Done.
PS: Seems like utils could use some typing (returns, Promise<>, etc).

PR #7587, opened 2024-06-01
fix: ubisys writeStructure for config was not using ZLC.DataType

sjorge:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/22785

I didn't save the pcaps, but I checked with wireshark that it looked the same as the earlier good capture after this change. The device also accepted and applied the config I pushed.


Koenkk:
Thanks!

PR #7589, opened 2024-06-01
fix(detect): _TZE200_7ytb3h8u as Giex Water Valve

wtaylor:
Bought a water valve that had support added a few months ago but despite looking identical, the `manufacturerName` is slightly different. Added the value reported as Zigbee Manufacturer in the z2m UI.

### DIFF
--- a/src/devices/giex.ts
+++ b/src/devices/giex.ts
@@ -68,6 +68,7 @@ const definitions: Definition[] = [
         fingerprint: [
             {modelID: 'TS0601', manufacturerName: '_TZE200_a7sghmms'},
             {modelID: 'TS0601', manufacturerName: '_TZE204_7ytb3h8u'},
+            {modelID: 'TS0601', manufacturerName: '_TZE200_7ytb3h8u'},
         ],
         exposes: [
             ...exportTemplates.giexWaterValve.exposes,


Koenkk:
Thanks!

mmirannalti:
hi @wtaylor
Thanks for your fix, it helped me as well.

Just out of curiosity, is your valve a QT06 or GX02? Grey or fluo green?
Mine, that came with the same model/manufacturer combination, is actually a fluo gren GX02.

wtaylor:
> hi @wtaylor Thanks for your fix, it helped me as well.
> 
> Just out of curiosity, is your valve a QT06 or GX02? Grey or fluo green? Mine, that came with the same model/manufacturer combination, is actually a fluo gren GX02.
> 
>  

Yes, mine is green too
IMG20240617100107

heidle78:
Just a question, I have the same device and it doesn´t show up correctly? What is going wrong?

mmirannalti:
Hi.
What version of Z2M are you running?
With version 1.39.1-1 of the HA add-on it works now.

I had to run for a while the EDGE version but now the relevant fixes are part of the official release.

heidle78:
I tried to delete the device and refind again - but I guess the system always make a link to the historic data

mmirannalti:
Strange, for me it worked right away, once I ran the right version of Z2M.

Are you sure you have the same Zigbee Model/Zigbee Manufacturer values?
Apparently they keep on changing over time, although the device seems to be the same...
I have 3 of them, they look the exactly the same, but they came with different Manufacturer values.

heidle78:
That is found:
image

Not sure if there is another way to find the correct device definition

heidle78:
Thanks for the feedback, seems the firmware makes the difference

PR #7593, opened 2024-06-01
Add _TZE204_myd45weu fingerprint to Tuya TS0601_soil sensor

micwiel:
Looks like there is a new fingerprint in Soil Sensors I bought recently which seems to have the same characteristics like `_TZE200_myd45weu` one.


Koenkk:
Thanks!

PR #7597, opened 2024-06-02
WISZB-120: Use "battery_voltage" instead of "voltage"

kranzj:
Context: Currently, the wrong unit (V instead of mV) is reported for the battery voltage.


Koenkk:
Thanks!

PR #7598, opened 2024-06-03
[New device support]: TuYa Universal Smart IR Control - UFO-R4Z

zycarlc:


src/devices/tuya.ts
```diff
@@ -6072,6 +6072,18 @@ const definitions: Definition[] = [
         toZigbee: [tzZosung.zosung_ir_code_to_send, tzZosung.zosung_learn_ir_code],
         exposes: [ez.learn_ir_code(), ez.learned_ir_code(), ez.ir_code_to_send()],
     },
+    {
+        fingerprint: [{modelID: 'TS1201', manufacturerName: '_TZ3290_rlkmy85q4pzoxobl'}],
```
Koenkk, line 6076: Since the code is exactly the same as `ZS06`, could you add it as a `tuya.whitelabel` there?

zycarlc: Sorry I did not realize tuya.whitelabel before. Thanks for your reminder. Has added as instructed

Koenkk:
Thanks!

Could you also submit a picture for the docs? This can be done by clicking [here](https://github.com/Koenkk/zigbee2mqtt.io/upload/master/public/images/devices). 

Make sure that:
- The size is 512x512

PR #7602, opened 2024-06-03
fix: STARKVIND child_lock was broken after refactor

sjorge:
As reported here https://github.com/Koenkk/zigbee2mqtt/discussions/8744#discussioncomment-9641181, the refactor to modernExtend inverted the logic for child_lock and also changed the type to boolean true/false while most child_lock exposes use LOCK/UNLOCK.

This change addresses this, by reverting back to the old behavior and LOCK/UNLOCK value.

Testing was easy, I observe the wrong and inverted behavior on my STARKVIND. And observed the correct behavior afterwards.


Koenkk:
Thanks!

PR #7611, opened 2024-06-06
feat(Kwikset): Adds pincode support for additional Kwikset model

timnolte:
- Adds pin code support for the Kwikset 99140-002.


Koenkk:
Thanks!

PR #7613, opened 2024-06-07
Added support for YINMIK YK-S03 Pool Monitoring

lauer:
This will add support for YINMIK YK-S03 Smart pH and Chlorine Tester for Swimming Pool

https://www.amazon.com/YINMIK-Chlorine-Swimming-Salinity-Inground/dp/B0C2T8YLYW

It look like the same device as BLE-YL01, but some params might be a bit different.
Koenkk/zigbee2mqtt#18704

Image added in this PR Koenkk/zigbee2mqtt.io#2796

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -8712,6 +8712,76 @@ const definitions: Definition[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_d9mzkhoq']),
+        model: 'YK-S03',
+        vendor: 'Tuya',
+        description: 'Smart pH and Chlorine Tester for Swimming Pool',
+        whiteLabel: [{vendor: 'YINMIK', model: 'YK-S03'}],
+        fromZigbee: [tuya.fz.datapoints],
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -8712,6 +8712,76 @@ const definitions: Definition[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_d9mzkhoq']),
```
Koenkk, line 8716: I propose to merge this definition with BLE-YL01 (add the missing features there, probably they are exactly the same devices). For the correct detection, use `tuya.whitelabel` (search in this file for examples)

lauer: What about the image? They looks a bit different.

Koenkk: It will show in the UI with the props defined in the whitelabel, including the picture.

lauer: I have merged the definition now, and added some whitelabels, as how I understod it works.

Is it okay now?

Koenkk:
Perfect, thanks!

PR #7614, opened 2024-06-08
feat: move manuSpecificIkeaAirPurifier to zhc

sjorge:
Merge before https://github.com/Koenkk/zigbee-herdsman/pull/1083

Still working fine after cluster has been moved:

```
[2024-06-08 15:11:41] debug:    z2m: Publishing get 'get' 'pm25' to 'airpurifier/bedroom'
[2024-06-08 15:11:41] debug:    z2m: Received Zigbee message from 'airpurifier/bedroom', type 'readResponse', cluster 'manuSpecificIkeaAirPurifier', data '{"particulateMatter25Measurement":65535}' from endpoint 1 with groupID 0
[2024-06-08 15:11:47] debug:    z2m: Publishing 'set' 'child_lock' to 'airpurifier/bedroom'
[2024-06-08 15:11:50] debug:    z2m: Publishing get 'get' 'child_lock' to 'airpurifier/bedroom'
[2024-06-08 15:11:50] debug:    z2m: Received Zigbee message from 'airpurifier/bedroom', type 'readResponse', cluster 'manuSpecificIkeaAirPurifier', data '{"childLock":1}' from endpoint 1 with groupID 0
```

PR #7615, opened 2024-06-08
Add manuSpecificIkeaUnknown

sjorge:
Rebase after https://github.com/Koenkk/zigbee-herdsman-converters/pull/7614

Shows up on newer ikea devices and some devices that had a FW update post DIRIGERA.
Potentially needs to be added to more devices, but I added it on the ones I could verify had it, or devices sharing a fw with devices I verified.
e.g. all ikea blinds seem to share the same base firmware, so I added it to all of them.

I also observed some devices like STARKVIND, GUNNARP panel 40*40, FLOALT panel WS 30x90, and FLOALT panel WS 30x30 not having it.
These devices have not had a new FW release since DIRIGERA to my knowledge. So we can't just add it to all devices.

Since there are no known attributes, it's purely cosmetic but my weird itch is scratched, no more random numeric clusters on my IKEA devices. If we do every somehow get incoming date from one it should be easier to spot now.

Screenshot 2024-06-08 at 16 32 17

As per discord, It's OK to close and not merge this. But I was in here for the other 2 clusters anyway so not much effort to just do this one too.

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -14,6 +14,7 @@ import {
     tradfriRequestedBrightness, tradfriCommandsOnOff,
     tradfriCommandsLevelCtrl, styrbarCommandOn,
     ikeaDotsClick, ikeaArrowClick, ikeaMediaCommands,
+    addCustomClusterManuSpecificIkeaUnknown,
     addCustomClusterManuSpecificIkeaAirPurifier,
     addCustomClusterManuSpecificIkeaVocIndexMeasurement,
 } from '../lib/ikea';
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

sjorge:
@Koenkk not sure you have may ikea devices I don't have, but if you do. Might be worth checking of others need it too. I presume some will.

Koenkk:
Im wondering if we should add all manuspecific IKEA clusters to all IKEA devices

sjorge:
> Im wondering if we should add all manuspecific IKEA clusters to all IKEA devices, that prevents maintenance even when devies get OTAed

I don't think it makes sense for some very specific ones, like the STARKVIND and VINDSTYRKA clusters, they don't even use the same one for there pm25 for example.

It might make sense specifically for manuSpecificIkeaUnknown though, It's either going to be present or not. Don't think it will be an issue if it's not present and we add it.

Koenkk:
Exactly, so that might be the easiest for now

sjorge:
> Exactly, so that might be the easiest for now

I'll do a PR

PR #7617, opened 2024-06-08
fix: max value for bosh shutter button calibration

boesec:
fix comma error for bosch shutter button calibration options


Koenkk:
Thanks!

PR #7633, opened 2024-06-13
Enable OTA for Bosch BTH-RM & housekeeping

burmistrzak:
Includes the following changes:

- Switch to `modernExtend` OTA
- Enable OTA for both BTH-RM models
- Remove unused `calibration` key from BMCT-SLZ
- Minor cleanup


Koenkk:
Thanks!

PR #7638, opened 2024-06-14
feat(add): MiBoxer TR-C1ZR dimmer https://github.com/Koenkk/zigbee2mq…

mlichvar:
This is an attempt to add support for the MiBoxer dimmer requested in https://github.com/Koenkk/zigbee2mqtt/issues/19141.

It depends on https://github.com/Koenkk/zigbee-herdsman/pull/1090.

I'm not sure if it this is the right approach. I don't really know what I'm doing here. At least one missing piece is reading the minimum brightness back. Suggestions welcome.

### DIFF
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -1735,7 +1735,7 @@ const tuyaModernExtend = {
             result.exposes = result.exposes.map((e) => utils.isLightExpose(e) ? e.withMinBrightness() : e);
         } else if (args.minBrightness === 'command') {
             result.toZigbee.push(tuyaTz.min_brightness_command);
-            result.exposes = result.exposes.map((e) => utils.isLightExpose(e) ? e.withMinBrightness().setAccess('state', ea.STATE_SET) : e);
+            result.exposes = result.exposes.map((e) => utils.isLightExpose(e) ? e.withMinBrightness().setAccess('min_brightness', ea.STATE_SET) : e);
         }
 
         if (args.color) {
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -1108,9 +1108,9 @@ const tuyaTz = {
         },
         // The response contains the value but as the data type, randomly
         // causing malformed messages
-        //convertGet: async (entity, key, meta) => {
+        // convertGet: async (entity, key, meta) => {
         //    await entity.read('lightingColorCtrl', [0xf102]);
-        //},
+        // },
     } satisfies Tz.Converter,
     color_power_on_behavior: {
         key: ['color_power_on_behavior'],
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -1083,7 +1083,7 @@ const tuyaTz = {
             await entity.write('genOnOff', {0x8000: {value: v, type: 0x10}});
         },
     } satisfies Tz.Converter,
-    min_brightness: {
+    min_brightness_attribute: {
         key: ['min_brightness'],
         convertSet: async (entity, key, value, meta) => {
             const number = utils.toNumber(value, `min_brightness`);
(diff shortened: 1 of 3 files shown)


src/lib/tuya.ts
```diff
@@ -1098,6 +1098,15 @@ const tuyaTz = {
             await entity.read('genLevelCtrl', [0xfc00]);
         },
     } satisfies Tz.Converter,
+    min_brightness2: {
```
Koenkk, line 1101:
```suggestion
    min_brightness_2: {
```

And could you rename the existing `min_brightness` to `min_brightness_1`

src/lib/tuya.ts
```diff
@@ -1098,6 +1098,15 @@ const tuyaTz = {
...
+        key: ['min_brightness'],
+        convertSet: async (entity, key, value, meta) => {
+            const number = utils.toNumber(value, `min_brightness`);
```
Koenkk, line 1104:
```suggestion
            utils.assertNumber(value, key);
            const payload = {minimum: value};
```

src/lib/tuya.ts
```diff
@@ -1684,8 +1693,8 @@ const tuyaModernExtend = {
         return tuyaModernExtend.dpEnumLookup({name: 'power_on_behavior', lookup: lookup, type: dataTypes.enum,
             expose: e.power_on_behavior(Object.keys(lookup)).withAccess(readOnly ? ea.STATE : ea.STATE_SET), ...args});
     },
-    tuyaLight(args?: modernExtend.LightArgs & {minBrightness?: boolean, switchType?: boolean}) {
-        args = {minBrightness: false, powerOnBehavior: false, switchType: false, ...args};
+    tuyaLight(args?: modernExtend.LightArgs & {minBrightness?: boolean, minBrightness2?: boolean, switchType?: boolean}) {
```
Koenkk, line 1696:
```suggestion
    tuyaLight(args?: modernExtend.LightArgs & {minBrightness?: 'attribute' | 'command', switchType?: boolean}) {
```

src/lib/tuya.ts
```diff
@@ -1721,6 +1730,11 @@ const tuyaModernExtend = {
             result.exposes = result.exposes.map((e) => utils.isLightExpose(e) ? e.withMinBrightness() : e);
         }
 
+        if (args.minBrightness2) {
+            result.toZigbee.push(tuyaTz.min_brightness2);
+            result.exposes = result.exposes.map((e) => utils.isLightExpose(e) ? e.withMinBrightness() : e);
```
Koenkk, line 1735:
```suggestion
            result.exposes = result.exposes.map((e) => utils.isLightExpose(e) ? e.withMinBrightness().setAccess('state', ea.STATE_SET) : e);
```

src/lib/tuya.ts
mlichvar: I added also "none" for devices that cannot set minimum brightness. Not sure if that is right.

mlichvar:
Thanks for the suggestions. I updated the commit.

I was able to capture the attribute that the Tuya gateway is requesting and I could add it to Z2M, but strangely the response contains the value in the position of the data type, causing errors like:
````
'Error: ZCL command 0xa4c138cda399565f/1 lightingColorCtrl.read([61698], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Read for '7' not available)'
````

With the `convertGet` removed, I still get `z2m: No converter available for 'get' 'min_brightness' ()` and the CI is failing. I'm not sure if that `STATE_SET` does what I think it does or if it is set correctly.

Koenkk:
Thanks!

PR #7640, opened 2024-06-14
fix TICMeter entity name

xmow49:
Hi;

Some fixes for the TICMeter device:

- Fix entity name by removing spaces and accents for mqtt topics.
- Fix uptime division.

Thanks


Koenkk:
Thanks!

PR #7641, opened 2024-06-15
Add "off" fan mode for Inovelli

MangoScango:
Proposed fix for https://github.com/Koenkk/zigbee2mqtt/issues/22987.  Define "off" mode for fans, and allow conversion of 0% brightness to "off" fan mode.

I am generally not set up to be able to test this change personally, but due to the nature of it I am fairly confident this should resolve the issue. (If someone could point me towards how I could easily test this change in a HassOS environment, I would appreciate it)


Koenkk:
Thanks!

PR #7642, opened 2024-06-15
fix(detect): Detect  as MiBoxer E2-ZR

ppanagiotis:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/21185

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1347,6 +1347,7 @@ const definitions: Definition[] = [
             tuya.whitelabel('Lidl', 'HG08131C', 'Livarno Home outdoor E27 bulb in set with flare', ['_TZ3000_q50zhdsc']),
             tuya.whitelabel('Lidl', 'HG07834C', 'Livarno Lux E27 bulb RGB', ['_TZ3000_qd7hej8u']),
             tuya.whitelabel('MiBoxer', 'FUT037Z+', 'RGB led controller', ['_TZB210_417ikxay', '_TZB210_wxazcmsh']),
+            tuya.whitelabel('MiBoxer', 'E2-ZR', '2 in 1 led controller', ['_TZB210_ayx58ft5']),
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

andrew01144:
Hi - I have this problem, and have not succeeded in fixing it.
I my E2-ZR successfully joined Z2M, but it appeared as Tuya TS0502B. (Z2M rev was ~01-Jun, that's before the fix).
I upgraded Z2M; herdsman now shows as 19.62.0.
I removed the device. Re-joined the device, but it still appears as Tuya TS0502B.
Am I missing a step?
Many thanks!

Also - setting 'transition' seems to have no affect on the actual transition.

pvyleta:
@ppanagiotis Are you sure this is correct? I can see you added the `E2-ZR` into the section of RGB+CCT lights, while in fact the `E2-ZR` is just Dual-White controller, similar to the Miboxer FUT035.

In fact, `E2-ZR` reports itself as `TS0502B` and not as `TS0505B` as your PR assumes. 

Is there any chance that you wanted to add the `E3-ZR` instead of the `E2-ZR` ?

ppanagiotis:
> @ppanagiotis Are you sure this is correct? I can see you added the `E2-ZR` into the section of RGB+CCT lights, while in fact the `E2-ZR` is just Dual-White controller, similar to the Miboxer FUT035.
> 
> In fact, `E2-ZR` reports itself as `TS0502B` and not as `TS0505B` as your PR assumes.
> 
> Is there any chance that you wanted to add the `E3-ZR` instead of the `E2-ZR` ?

I'm not sure. In the past few months, I didn't have the opportunity to test if it works properly. Feel free to modify or adjust it as needed.

pvyleta:
@ppanagiotis Thanks for the clarification! I therefore can let you know that my hunch was correct, and once I moved the definition of the device, it now works like a charm for me and for others. Thanks a lot for the initial contribution, I would not be able to locate the right place without your previous PR.

PR #7651, opened 2024-06-18
fix(detect): Detect TRADFRI bulb E26 WS globe 1100lm as IKEA LED2201G8

tongkl1:
I just bought two IKEA TRADFRI E26 1100lm bulbs and found them to be unsupported. After a little research it seems it's because this California-only version is not documented, hence this PR.

Edit: force-push due to lint issue


Koenkk:
thanks!

PR #7653, opened 2024-06-18
SPLZB-132: read develco specific sw/hw versions

sjorge:
Got one of these for testing, they also need the develco hack.

(I should really get these into a modernExtend sometime soonish, also I noticed hitting check OTA wipes the version as the default genBasic one is not responding, I wonder if there is a way to fix that too but don't see how for now, what does the post update read? I guess we could just not 'update' the value in case of null?

```
[2024-06-18 19:57:52] debug:    zh:controller:endpoint: Error: ZCL command 0x0015bc002f013891/2 genBasic.read(["dateCode","swBuildId"], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')
[2024-06-18 19:57:52] info:     z2m: Device 'outlet/bathroom/heater' was updated from 'null' to 'null'
```
)

### DIFF
--- a/src/devices/develco.ts
+++ b/src/devices/develco.ts
@@ -354,6 +354,8 @@ const definitions: Definition[] = [
             await reporting.currentSummDelivered(endpoint, {change: [0, 20]}); // Limit reports to once every 5m, or 0.02kWh
             await reporting.instantaneousDemand(endpoint, {min: constants.repInterval.MINUTES_5, change: 10});
             await reporting.acFrequency(endpoint);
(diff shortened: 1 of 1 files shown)


Koenkk:
I think the version fix should be applied in z2m otaUpdate.ts

PR #7654, opened 2024-06-18
Minor `modernExtend` updates for Bosch BMCT-SLZ

burmistrzak:
Includes the following changes:

- Fix for `calibrationButtonHoldTime`
- ~~Add `boschExtend.roomThermostat()`~~
- ~~Add `boschExtend.radiatorThermostat()`~~
- Remove `fromZigbee` & `toZigbee` for BMCT-SLZ
- Use `deviceEndpoints` for BMCT-SLZ


src/devices/bosch.ts
```diff
@@ -326,6 +327,66 @@ const boschExtend = {
             isModernExtend: true,
         };
     },
+    roomThermostat: (): ModernExtend => {
```
Koenkk, line 330: Since both `roomThermostat` and `radiatorThermostat` are generic, I think this can be moved to `modernExtend.ts` under a combined `thermostat`

Koenkk: I think it's fine to add it there, don't see any risks for now

Koenkk:
Thanks!

PR #7657, opened 2024-06-19
1-THZB switch with sensor

kirovilya:
TYZGTH1CH-D1RF
https://aliexpress.ru/item/1005007039660277.html


Koenkk:
Could you also submit a picture for the docs? This can be done by clicking [here](https://github.com/Koenkk/zigbee2mqtt.io/upload/master/public/images/devices). 

Make sure that:
- The size is 512x512

kirovilya:
done https://github.com/Koenkk/zigbee2mqtt.io/pull/2830
but... :)
image

DooMMasteR:
@kirovilya My best guess is, that the `TYZGTH4CH-D1RF` I added also supports switching based on Temperature and humidity thresholds.  
I have no Tuya Gateway so I could not implement this "rather hard to just find" feature for the integration.  
I might wonder if anyone actually would use it or if people would rather remote control the device based on it's outputs?

I might, in the future buy a Tuya gateway and try to find out which data points and how they are used on the 4 channel variant. #7661

PR #7659, opened 2024-06-19
Add new device ZG2819S-RGBW

chris-1243:
Add new device ZG2819S-RGBW

### DIFF
--- a/src/devices/sunricher.ts
+++ b/src/devices/sunricher.ts
@@ -333,19 +333,19 @@ const definitions: Definition[] = [
         },
     },
     {
-    zigbeeModel: ['ZG2819S-RGBW'],
-    model: 'ZG2819S-RGBW',
-    vendor: 'Sunricher',
-    description: 'Zigbee handheld remote RGBW 4 channels',
-    fromZigbee: [fz.battery, fz.command_on, fz.command_off, fz.command_step, fz.command_move, fz.command_stop, fz.command_recall,
-        fz.command_move_hue, fz.command_move_to_color, fz.command_move_to_color_temp],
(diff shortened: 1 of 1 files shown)


Koenkk:
This device is already supported

https://github.com/Koenkk/zigbee-herdsman-converters/blob/f9c6313e517b4d805a75940cb608f98a57338a42/src/devices/iluminize.ts#L161

chris-1243:
Well and in my opinion, we should modify the iluminze.ts file as the model should not be `ZG2819S-RGBW` but `511.344`. My model is from Sunricher. It would be great to have it recognized correctly.

As you can see, this remote is also supported as a RGB Genie [ZB-5028](https://www.zigbee2mqtt.io/devices/ZB-5028.html).

On the other hand, the CCT model is well defined as:
  - Sunricher [ZG2819S-CCT](https://www.zigbee2mqtt.io/devices/ZG2819S-CCT.html)
  - Iluminize [511.324](https://www.zigbee2mqtt.io/devices/511.324.html)
 
Why not keeping this difference and used my converter ?

Koenkk:
You cannot distinguish these devices as the model ID is exactly the same

chris-1243:
It works but in the expose tab you have tons of useless information which are not reported by the device. My external converter was following the way the Sunricher CCT model works and what kind of information is reported.

To be honest, I did not want to remove code lines in the iluminize.ts file while submitting my PR. I thought those two devices were different.

As it is your project, I let you decide what is the best for it

511 344_expose

chris-1243:
I also have a converter made with the modernextend

PR #7661, opened 2024-06-19
feat(add): TS000F based TYZGTH4CH-D1RF

DooMMasteR:
This change adds support for the Mumubiz TYZGTH4CH-D1RF
which is based on a Tuya ZT3L module and reports as TS000F.

The device has a connector that supports an external DHT-22 or alike one-wire temperature and humidity senor.
It might also have detection for DS18B20 sensors, but I can't verify it.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -8643,6 +8643,34 @@ const definitions: Definition[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS000F', ['_TZ3218_ya5d6wth']),
+        model: 'TYZGTH4CH-D1RF',
+        vendor: 'Mumubiz',
+        description: '4 channel changeover contact with temperature and humidity sensing',
+        extend: [
+            tuya.modernExtend.tuyaOnOff(
+                {
+                    powerOnBehavior2: true,
+                    onOffCountdown: true,
+                    endpoints: ['l1', 'l2', 'l3', 'l4'],
+                }
+            ),
+            tuya.modernExtend.dpTemperature({dp: 102, scale: 10}),
+            tuya.modernExtend.dpHumidity({dp: 103}),
+        ],
+        endpoint: (device) => {
+            return {'l1': 1, 'l2': 2, 'l3': 3, 'l4': 4};
+        },
+        exposes: [],
+        meta: {multiEndpoint: true},
+        configure: async (device, coordinatorEndpoint) => {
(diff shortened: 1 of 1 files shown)


DooMMasteR:
related doc: https://github.com/Koenkk/zigbee2mqtt.io/pull/2829

Koenkk:
Thanks!

DooMMasteR:
I might get a Tuya gateway soon to add all the remaining datapoints.

Stepyon:
Hey @DooMMasteR, it looks like the countdown implementation is not exactly correct or full for the device. 
According to the [doc](https://www.zigbee2mqtt.io/devices/TYZGTH4CH-D1RF.html#countdown-numeric-l1-endpoint) the countdown should turn a device switch off after a certain time but it is not what it actually does. What it actually does, when you set the countdown it toggles a device switch OFF or **ON** when the time passed depending on the current state and then resets back to 0.

I checked with Tuya hub and there we can set 2 timer values for each gang:
switch

- "count down" - starting a countdown to toggle the device ON or OFF. This is what we already have exposed.
- "timing" - this sets the auto OFF delay, the value is saved on the device, so works even without hub connection and after device reboot. This is not exposed currently and I really miss this useful setting. 

Do you know if we can expose it as well? I'm ready to help testing if needed. Thanks.

DooMMasteR:
The timing features did not originally exist, I think someone else added it later, I guess. 
I do not have a not in use device to develop against at the though so I cannot add/verity the feature.

DooMMasteR:
@Stepyon idelally adding `inchingSwitch: true` to the converter might already add the feature.
you can test this: 4652a6d6b3002ec84fbc26bb7032d5f570be2acf aka https://github.com/Koenkk/zigbee-herdsman-converters/pull/11948

DooMMasteR:
Hey @Stepyon :-) the next release will at least add the feature on l1-3; l4 seems weird (probably an offset issue) but that might only be fixed at a later time and is mostly out of my scope of knowledge.

Stepyon:
@DooMMasteR thanks for the fix. 
Do you have any idea why itching timer settings don't have entities, I can change them in Zigbee2MQTT Exposes tab (with an external converter used), but they are not available in device controls and I can't add them to a dashboard because there are no such entities. The countdown settings have them at the same time.
image

andrei-lazarov:
Iirc it's because HA does not support composite entities.

So anything that is nested in Z2M ui, a feature with multiple options, won't show up in HA. But you can still use MQTT to control it, of course

PR #7662, opened 2024-06-20
fix some bug about energy and produced energy

script0803:


DooMMasteR:
Could you explain what the issues were or why the change is needed?

Filtering out high readings >3600 and low ones seems like an odd decision, why 5 and 3600? that seems like an odd choice of values....

script0803:
The main reason is that the previous 'await readMeteringPowerConverterAttributes(endpoint)' was deprecated and changed to 'await reporting.readMeteringMultiplierDivisor(endpoint)' I neglected it when I wrote it last time, which led to the problem that energy and produced_energy could not be read. 
And I found that our device will only report data when energy and produced energy change by 0.1. Therefore, when the device only measures energy, produced energy will still not be reported, so I changed 'change' to 0, so that both energy and produced can be reported.
 As for the reporting interval of 5 to 3600, this is consistent with our other products and will not have a big impact.

script0803:
As you can see in the issuehttps://github.com/Koenkk/zigbee2mqtt/issues/22429, this update only fixes this bug and does not affect other functions.

DooMMasteR:
still weird to filter/limit the readings, as this could still create some "weird" corner cases when reporting jumps from import to export or the other way around, but I guess it's ok when your other devices behave the same.

src/devices/bituo_technik.ts
```diff
@@ -18,10 +18,12 @@ const definitions: Definition[] = [
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['haElectricalMeasurement', 'seMetering']);
-            await reporting.currentSummDelivered(endpoint);
-            await reporting.currentSummReceived(endpoint);
+            await reporting.readMeteringMultiplierDivisor(endpoint);
```
Koenkk, line 21: Why did you add both this line and `await endpoint.saveClusterAttributeKeyValue('seMetering', {divisor: 100, multiplier: 1});`, one overwrites the other.

src/devices/bituo_technik.ts
```diff
@@ -18,10 +18,12 @@ const definitions: Definition[] = [
...
+            await reporting.currentSummDelivered(endpoint, {min: 5, max: 3600, change: 0});
```
Koenkk, line 22: Only add `change: 0` here and comment the reason why you've added it:

> And I found that our device will only report data when energy and produced energy change by 0.1. Therefore, when the device only measures energy, produced energy will still not be reported, so I changed 'change' to 0, so that both energy and produced can be reported.

PR #7664, opened 2024-06-20
Inovelli - Fix fan mode issue with vzm36

InovelliUSA:
Also remove P121 for vzm31-sn as it is not a valid param for that device.


Koenkk:
Thanks!

PR #7668, opened 2024-06-21
Update iluminize.ts

chris-1243:
Update ModernExtend and update of 511.344 device for ModernExtend support (better support of reported commands) and suppression of not working reporting options


chris-1243:
It is a following of #7659

Koenkk:
Thanks!

PR #7671, opened 2024-06-21
add CTM Lyng Norlum Dim OP

geirra:
This adds support for the CTM Lyng Norlum rotary dimmer.

Product link: https://ctmlyng.no/produkter/norlum-dim-op/


Koenkk:
Thanks!

PR #7677, opened 2024-06-23
Omit battery_low for models that don't have it

deviantintegral:
The devices in this PR don't seem to actually ever send a battery alarm. I've only ever seen them report "normal" in z2m. I've been running a sniffer for 24 hours and `zbee_zcl_general.power_config.attr_id > 0x0021` shows no results. If I'm reading https://github.com/Koenkk/zigbee-herdsman/blob/73162478f5e5c3d9058bfada27dfff52b9d2c612/src/zspec/zcl/definition/cluster.ts#L71 right, if reported it should be reported with attribute ID `0x3E`.

Also, I've never had this value flip _before_ a battery dies, so I don't think it is just waiting for a failure state to send.


Koenkk:
Thanks!

PR #7679, opened 2024-06-23
BHI-US: Add missing expose of button actions

mldytech:
According to the discussion in the original PR #6651, add the missing expose of button actions for Bosch BHI-US Universal Switch II (BHI-US)


mldytech:
@burmistrzak yes, I own one and could definitely do that :-)

PR #7682, opened 2024-06-23
Use 2.1-3V range for Third Reality 3RSS008Z switches

deviantintegral:
This matches what ZHA does at https://github.com/zigpy/zha-device-handlers/blob/9df7c3bdd7b37f46cf6e269cec2032aedde698c1/zhaquirks/thirdreality/switch.py#L22.

While the PR at https://github.com/zigpy/zha-device-handlers/pull/208 doesn't say so, I think someone must have tested to get these numbers as ZHA overrides their defaults for it while we are currently just using ours.

### DIFF
--- a/src/devices/third_reality.ts
+++ b/src/devices/third_reality.ts
@@ -41,6 +41,7 @@ const definitions: Definition[] = [
         ota: ota.zigbeeOTA,
         fromZigbee: [fz.on_off, fz.battery],
         toZigbee: [tz.on_off, tz.ignore_transition],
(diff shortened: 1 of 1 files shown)


deviantintegral:
This device was previously shown at 26%. With this change, they came in at 8%, and my battery tester showed them at 10% (it only does 10% increments). These were two NiMH batteries, so I think it's safe to say that the existing reporting of 26% was certainly too high.

Koenkk:
Thanks!

PR #7683, opened 2024-06-23
Allow `exposes` in `ModernExtend` interface to be a function

burmistrzak:
Already works in `Definition`, now it does in `ModernExtend` as well!

Context here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7444#issuecomment-2183967705


src/lib/types.ts
```diff
@@ -168,7 +168,7 @@ export type OnEvent = (type: OnEventType, data: OnEventData, device: Zh.Device,
 export interface ModernExtend {
     fromZigbee?: Fz.Converter[],
     toZigbee?: Tz.Converter[],
-    exposes?: Expose[],
+    exposes?: DefinitionExposes,
```
Koenkk, line 171:
```suggestion
    exposes?: (Expose | DefinitionExposesFunction)[],
```

I think it has to be this

Koenkk:
I've implemented it in https://github.com/Koenkk/zigbee-herdsman-converters/commit/059dfb88464287ac63da5cff880db97ce1fe1ba4 now

PR #7630, opened 2024-06-12
add _TZE204_rzrrjkz2 Smart Sprinkler Timer

OgV1:
add _TZE204_rzrrjkz2 Smart Sprinkler Timer

PR #7688, opened 2024-06-24
Update fingerprint so gets recognized correctly

OgV1:
These DPs on [this ](https://github.com/Koenkk/zigbee-herdsman-converters/pull/7630) created from device with this fingerprint **_TZE204_rzrrjkz2**


Koenkk:
Thanks!

PR #7692, opened 2024-06-24
Add support for _TZ3000_huvxrx4i

marthubner:
https://www.aliexpress.com/item/1005004774422133.html?spm=a2g0o.order_list.order_list_main.23.1ccc1802LuVhqB

2gang version - tested and it seems to be the same device as already supported, only different code.


Koenkk:
Thanks!

PR #7694, opened 2024-06-25
Add Support For Leading/Trailing Edge Dimming Parameter For Inovelli Fan Canopy Module

jsievenpiper:
From #6786 adds Leading/Trailing edge dimming support to just the Fan Canopy module from Inovelli.
See more: https://github.com/Koenkk/zigbee-herdsman-converters/pull/6786#issuecomment-2187545577

@InovelliUSA thank you for your support!


Koenkk:
thanks!

PR #7696, opened 2024-06-25
Add endpoint number for P26

InovelliUSA:
I believe we need the endpoint to make sure the configuration gets sent to the correct one. At the least it will be consistent with the rest of the vzm36 params.


Koenkk:
Thanks!

PR #7698, opened 2024-06-26
ME168 requires 6 time slots instead of 4

Paalap:
ME168 requires 6 time slots which wasn't considered in the logics.

BR


Paalap:
alternative to  #7406 - just found it

Koenkk:
Could you make the number of time slots an argument similar as https://github.com/Koenkk/zigbee-herdsman-converters/pull/7406/files#diff-d45f20d30a43dc3e172b16bd55872492f4867c8831c5e03a8be9be65f1cd72d5R857 to prevent code duplication?

PR #7702, opened 2024-06-27
fix: Implement prettier

Koenkk:
CC: @Nerivec


src/devices/ubisys.ts
```diff
@@ -124,13 +124,9 @@ const ubisys = {
...
-                        operationalStatus = (
-                            await entity.read(
-                                'closuresWindowCovering',
-                                ['operationalStatus'],
-                            )
-                        // @ts-expect-error
-                        ).operationalStatus;
+                        operationalStatus =
+                            // @ts-expect-error
+                            (await entity.read('closuresWindowCovering', ['operationalStatus'])).operationalStatus;
```
Nerivec, line 129: Maybe something like this?
```typescript
                        const response = await entity.read('closuresWindowCovering', ['operationalStatus'])
                        // @ts-expect-error
                        operationalStatus = response.operationalStatus;
```

src/lib/types.ts
```diff
@@ -23,8 +23,8 @@ export interface KeyValueString {
 export interface KeyValueNumberString {
     [s: number]: string;
 }
-// eslint-disable-next-line
 export interface KeyValueAny {
+    // eslint-disable-next-line
```
Nerivec, line 27: `eslint-disable-next-line @typescript-eslint/no-explicit-any`
Best to be specific instead of disabling all. Might be other places.

src/devices/ubisys.ts
Koenkk: Yes, that's cleaner :)

src/lib/types.ts
Koenkk: Good point, also removed some

PR #7706, opened 2024-06-29
feat: refactor develcoSpecificAirQuality into manuSpecificDevelcoAirQuality

sjorge:
Cleanup some of the develco stuff:
- [x] move clusters from zh to zhc
- [x] move attributes from zh to zhc
- [x] convert the sw/hw reading into modernExtend
- [x] move some of the specific attrbute based converters to modernExtend

Merge before https://github.com/Koenkk/zigbee-herdsman/pull/1097

### DIFF
--- a/src/lib/develco.ts
+++ b/src/lib/develco.ts
@@ -8,6 +8,17 @@ import {Fz, Tz, ModernExtend, Configure} from './types';
 const NS = 'zhc:develco';
 
 export const develcoModernExtend = {
+    addCustomClusterManuSpecificDevelcoGenBasic: () =>
+        deviceAddCustomCluster('genBasic', {
+            ID: 0x0000,
+            attributes: {
+                develcoPrimarySwVersion: {ID: 0x8000, type: Zcl.DataType.OCTET_STR, manufacturerCode: Zcl.ManufacturerCode.DEVELCO},
+                develcoPrimaryHwVersion: {ID: 0x8020, type: Zcl.DataType.OCTET_STR, manufacturerCode: Zcl.ManufacturerCode.DEVELCO},
+                develcoLedControl: {ID: 0x8100, type: Zcl.DataType.BITMAP8, manufacturerCode: Zcl.ManufacturerCode.DEVELCO},
+            },
+            commands: {},
+            commandsResponse: {},
+        }),
     addCustomClusterManuSpecificDevelcoAirQuality: () =>
         deviceAddCustomCluster('manuSpecificDevelcoAirQuality', {
             ID: 0xfc03,
--- a/src/lib/develco.ts
+++ b/src/lib/develco.ts
@@ -64,4 +64,66 @@ export const develcoModernExtend = {
         ];
         return {configure, isModernExtend: true};
     },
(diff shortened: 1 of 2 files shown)


src/lib/develco.ts
```diff
@@ -7,6 +7,40 @@ import {logger} from './logger';
 const NS = 'zhc:develco';
 
 export const develcoModernExtend = {
+    addCustomClusterManuSpecificDevelcoGenBasic: () => deviceAddCustomCluster(
+        'genBasic',
+        {
+            ID: 0x0000,
+            attributes: {
+                zclVersion: {ID: 0, type: Zcl.DataType.UINT8},
```
sjorge, line 15: @Koenkk The merging of these is not 100% clear to do.

Can I just not repeat all the ZCL genBasic attributes here and only add the 3 develco ones (L30-L32) here and have it properly merged?

Koenkk: It will get merged by default, so no need to re-specify everything.

sjorge:
I might need to start over, i can't seem to resolve the merge conflict :(

sjorge:
Going to test this a bit more and then mark ready, there is some more cleanup that can be done but I'm trying to work out how to best do it.

I think some of these can be a more generic, like a lot of them are ignoring certain bogus values. Which can probably be added in the generic modernExtend as it could be useful for other devices. I'll need to think about this some more.

src/lib/develco.ts
```diff
@@ -0,0 +1,159 @@
...
+        numeric({
+            name: 'voc',
+            cluster: 'manuSpecificDevelcoAirQuality',
+            attribute: 'measuredValue',
+            reporting: {min: '1_MINUTE', max: '1_HOUR', change: 10},
+            description: 'Measured VOC value',
+            // from Sensirion_Gas_Sensors_SGP3x_TVOC_Concept.pdf
+            // "The mean molar mass of this mixture is 110 g/mol and hence,
+            // 1 ppb TVOC corresponds to 4.5 μg/m3."
+            scale: 4.5,
```
sjorge, line 77: @Koenkk klopt dit?

https://github.com/Koenkk/zigbee-herdsman-converters/pull/7706/files#diff-97dc2d48d7db27dd10bb142fd276583a4dfdff437808452d813cfb4bb65b2b59L161 het zou dit moeten doen dus `* 4.5` maar het lijkt eerder een `/ 4.5` te doen?

Koenkk: Scale doet een `/`: https://github.com/Koenkk/zigbee-herdsman-converters/blob/a60047c170b562be05888513a9c493143ad91d7f/src/lib/modernExtend.ts#L1869

sjorge:
@Koenkk Dit kan al gemerged worden, de commits waar ik het in discord over heb waar ik nog tegen problemen loop zitten hier niet bij. En dit moet daar niet op wachten.

src/lib/develco.ts
sjorge: Found a ScaleFunction did the trick nicely.

Koenkk:
Bedankt!

PR #7707, opened 2024-06-29
Add support for Livolo Illuminance and Sound Sensor

PaulVanSchayck:
This add support for the [Livolo Illuminance and Sound Sensor (VL-FCJZ-2WP)](https://livolo.in.ua/en/mekhanizm-datchik-zvuka-zigbee-belyy-livolo-vl-fcjz-2wp/).

A couple of remarks/questions:

* The illuminance value reports between 0 and 100%. I hope this current Expose definition works for that.
* The noise value can be 1, 2, 3 and 4. The device manual calls these "silent", "normal", "lively" and "noisy". In practice I've rarely seen 1 ("silent"), but do see 2, 3 and 4. Perhaps I should implement this as an Enum expose. Would that be the best solution?
* I still need to make the device entry PR for zigbee2mqtt.io

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -2455,6 +2455,23 @@ const converters1 = {
             }
         },
     } satisfies Fz.Converter,
+    livolo_illuminance_state: {
+        cluster: 'genPowerCfg',
+        type: ['raw'],
+        convert: (model, msg, publish, options, meta) => {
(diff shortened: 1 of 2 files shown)


Koenkk:
The noise values are fine, the bigger the more noise.

Thanks for the PR!

Make sure that:
- The size is 512x512

PR #7609, opened 2024-06-05
[New device support]: Tuya Soil sensor - TS0601 TZE284_g2e6cpnw 

rare-magma:
## Link

https://www.aliexpress.com/item/1005007016626219.html

## Database entry

`{"id":19,"type":"EndDevice","ieeeAddr":"0xa4c1382e2ab69997","nwkAddr":40942,"manufId":4417,"manufName":"_TZE284_g2e6cpnw","powerSource":"Battery","modelId":"TS0601","epList":[1],"endpoints":{"1":{"profId":260,"epId":1,"devId":81,"inClusterList":[4,5,61184,0,60672],"outClusterList":[25,10],"clusters":{"genBasic":{"attributes":{"65487":14400,"65503":"��-\u0019��-\u0007\u0000\u0000\u0000\u0000\u0011","65506":56,"65508":1,"65534":0,"appVersion":77,"modelId":"TS0601","manufacturerName":"_TZE284_g2e6cpnw","powerSource":3,"zclVersion":3,"stackVersion":0,"hwVersion":1,"dateCode":""}}},"binds":[],"configuredReportings":[],"meta":{}}},"appVersion":77,"stackVersion":0,"hwVersion":1,"dateCode":"","zclVersion":3,"interviewCompleted":true,"meta":{"configured":-903893911},"lastSeen":1717618244776}`

## Comments

Device `0xa4c1382e2ab69997` with Zigbee model `TS0601` and manufacturer name `TZE284_g2e6cpnw`.

I'm not sure if the `102` datapoint is correct. 
I got the following debug info:

```
Received Zigbee message from 'Soil sensor', type 'commandDataReport', cluster 'manuSpecificTuya', data:
 '{
  "dpValues": [
    { "data": { "data": [2], "type": "Buffer" }, "datatype": 4, "dp": 102 }
  ],
  "seq": 3840
}' 
from endpoint 1 with groupID 0
```

The device sends the following messages more frequently:
```
{
    "dpValues": [
      {
        "data": { "data": [0, 0, 0, 0], "type": "Buffer" },
        "datatype": 2,
        "dp": 3
      },
      {
        "data": { "data": [0, 0, 1, 19], "type": "Buffer" },
        "datatype": 2,
        "dp": 5
      },
      {
        "data": { "data": [0, 0, 3, 47], "type": "Buffer" },
        "datatype": 2,
        "dp": 110
      }
    ],
    "seq": 3072
  }
```

PR #7710, opened 2024-06-30
Fix for TS0601_soil_2 tze284 g2e6cpnw datapoint mapping

rare-magma:
I think the datapoint mappings need to be reviewed for this device following the discussion in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7609#issuecomment-2198030719


rare-magma:
@xpavli44 I just tried these changes via an external converter and got the following message from the device:

```
{
	"dpValues": [
		{
			"data": { "data": [0, 0, 0, 21], "type": "Buffer" },
			"datatype": 2,
			"dp": 3
		},
		{
			"data": { "data": [0, 0, 0, 254], "type": "Buffer" },
			"datatype": 2,
			"dp": 5
		},
		{
			"data": { "data": [0, 0, 3, 9], "type": "Buffer" },
			"datatype": 2,
			"dp": 110
		}
	],
	"seq": 9219
}
```

I think the datapoint 5 and 110 correspond to the same temperature reading, the first one is in celsius and the other in fahrenheit

xpavli44:
@rare-magma

> I think the datapoints 5 and 110 correspond to the same temperature reading, the first one is in celsius and the other in fahrenheit.

indeed I made the same observation independently

PR #7716, opened 2024-07-01
feat(add): Expose deviceRunTime for STARKVIND air purifier

indrajitr:
Expose `deviceRunTime` in the diagnostic category for STARKVIND air purifier.
Also, add 'minutes' as the unit for filter_age and device_age.


Koenkk:
Thanks!

PR #7717, opened 2024-07-01
fix(detect): Adjust STARKVIND air purifier stand and table variants

indrajitr:
The STARKVIND air purifier has two variants, one for the stand and one for the table.

The table variant has model E2006 and the stand variant has model E2007.

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -662,6 +662,7 @@ const definitions: Definition[] = [
         model: 'E2007',
         vendor: 'IKEA',
         description: 'STARKVIND air purifier',
+        whiteLabel: [{vendor: 'IKEA', model: 'E2006', description: 'STARKVIND air purifier table', fingerprint: [{modelID: 'E2006'}]}],
         extend: [addCustomClusterManuSpecificIkeaUnknown(), addCustomClusterManuSpecificIkeaAirPurifier(), ikeaAirPurifier(), identify(), ikeaOta()],
     },
     {


indrajitr:
The only difference between E2006 and E2007 is the image (table vs stand). Is a way to conditionally switch the image shown based on the detected model? If yes, the two definitions could be combined.

src/devices/ikea.ts
```diff
@@ -658,7 +658,14 @@ const definitions: Definition[] = [
...
     {
-        zigbeeModel: ['STARKVIND Air purifier', 'STARKVIND Air purifier table'],
+        zigbeeModel: ['STARKVIND Air purifier table'],
+        model: 'E2006',
+        vendor: 'IKEA',
+        description: 'STARKVIND air purifier table',
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), addCustomClusterManuSpecificIkeaAirPurifier(), ikeaAirPurifier(), identify(), ikeaOta()],
+    },
+    {
+        zigbeeModel: ['STARKVIND Air purifier'],
```
Koenkk, line 668: It may be better to use a `whiteLabel` here [example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/7df9894acb4b904cc1f5a3799d9798ec86eb0edd/src/devices/custom_devices_diy.ts#L277)

indrajitr: Thank you for the pointer!

Koenkk: Please also add the fingerprint.modelId part for proper detection.

indrajitr: Done. Apologies, I missed that somehow.

Koenkk:
Thanks!

PR #7720, opened 2024-07-01
fix: Fix Bosch BTH-RA Home Assistant discovery

Koenkk:
Required for https://github.com/Koenkk/zigbee2mqtt/pull/23075

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -209,7 +209,7 @@ export interface DefinitionMeta {
     /**
      * Override the Home Assistant discovery payload using a custom function.
      */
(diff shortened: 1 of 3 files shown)


gpayer:
Looks good! :+1:

src/lib/types.ts
```diff
@@ -209,7 +209,7 @@ export interface DefinitionMeta {
     /**
      * Override the Home Assistant discovery payload using a custom function.
```
Nerivec, line 210: Maybe add a line that informs on where this is executed?
```typescript
    /**
     * Override the Home Assistant discovery payload using a custom function.
     * Executed right before the payload is passed to `stringify`, before publishing.
     */
```

PR #7723, opened 2024-07-02
fix: Recategorize configurable attributes for Inovelli

indrajitr:
Move configurable attributes for Inovelli switches to 'config' category instead of lumping them under 'control'.

Also add support for 'identify' for Inovelli switches.


Koenkk:
Thanks!

PR #7725, opened 2024-07-02
(fix) Use ZHA voltage calculation for TS0203 and WSD500A

deviantintegral:
I'm splitting this out from https://github.com/Koenkk/zigbee-herdsman-converters/pull/7681 since this only affects a few devices, while the work in the other PR affects much more and has seen several revisions.


Koenkk:
Thanks!

PR #7726, opened 2024-07-02
Fixing all endpoints being triggered at once when controlling a single endpoint for ZMO-606-S2

fabergoncalves:
I had to add the endpoint return in line 368 and the async configuration for the switch to properly work, without that, when I controlled a single switch, all the other switches were controlled as well

### DIFF
--- a/src/devices/zemismart.ts
+++ b/src/devices/zemismart.ts
@@ -358,7 +358,7 @@ const definitions: Definition[] = [
         extend: [tuya.modernExtend.tuyaOnOff({powerOutageMemory: true, indicatorMode: true, onOffCountdown: true, endpoints: ['l1', 'l2', 'l3']}),],
         meta: {multiEndpoint: true},
         endpoint: (device) => {
-            return {'l1': 1, 'l2': 2, 'l3': 3};
+            return {l1: 1, l2: 2, l3: 3};
         },
         configure: async (device, coordinatorEndpoint) => {
             await tuya.configureMagicPacket(device, coordinatorEndpoint);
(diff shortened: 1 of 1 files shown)


src/devices/zemismart.ts
```diff
@@ -355,11 +355,18 @@ const definitions: Definition[] = [
...
-        extend: [
-            deviceEndpoints({endpoints: {l1: 1, l2: 2, l3: 3}}),
-            identify(),
-            tuya.modernExtend.tuyaOnOff({indicatorMode: true, onOffCountdown: true, endpoints: ['l1', 'l2', 'l3']}),
-        ],
+        extend: [tuya.modernExtend.tuyaOnOff({powerOutageMemory: true, indicatorMode: true, onOffCountdown: true, endpoints: ['l1', 'l2', 'l3']})],
+        endpoint: (device) => {
+            return {l1: 1, l2: 2, l3: 3};
+        },
+        configure: async (device, coordinatorEndpoint) => {
```
Koenkk, line 362: I think the only change needed to fix this issue is: `configure: tuya.configureMagicPacket,`

fabergoncalves: Thanks, I changed to use configure: tuya.configureMagicPacket but I had to keep the enpoint: (device) in line 359. Without this I would get the error "No converter available for 'state_l1' ("OFF")"

Koenkk: That's interesting because that's what `deviceEndpoints({endpoints: {l1: 1, l2: 2, l3: 3}}),` should do, did you check with z2m 1.39.0?

fabergoncalves: Oh I see what happened, the "No converter available for 'state_l1' ("OFF")" happened because I removed the `deviceEndpoints({endpoints: {l1: 1, l2: 2, l3: 3}}),` to include the `endpoint: (device) => {`. When I removed everything and changed to `configure: tuya.configureMagicPacket,`, the code was missing and throwing this error. Now everything works as expected... just pushed an update.

Koenkk:
Thanks!

PR #7727, opened 2024-07-02
Add noise state for Livolo TI0001-illuminance

PaulVanSchayck:
As already mentioned in PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/7707 , I was looking for a way to represent both the 4 different noise states the device reports, as well as a binary sensor when noise was detected. 

Would this be an acceptable way? This works quite nicely for me.


src/devices/livolo.ts
```diff
@@ -340,7 +341,11 @@ const definitions: Definition[] = [
         model: 'TI0001-illuminance',
         description: 'Zigbee digital illuminance and sound sensor',
         vendor: 'Livolo',
-        exposes: [e.noise_detected(), e.illuminance().withUnit('%').withValueMin(0).withValueMax(100)],
+        exposes: [
+            e.noise_detected(),
+            e.illuminance().withUnit('%').withValueMin(0).withValueMax(100),
+            new Enum('noise_state', access.STATE, ['silent', 'normal', 'lively', 'noisy']).withDescription('Detected noise level'),
```
Koenkk, line 347:
```suggestion
            e.enum('noise_level', access.STATE, ['silent', 'normal', 'lively', 'noisy']).withDescription('Detected noise level'),
```

`noise_level` looks better I guess, also update the fromZigbee converter.

PaulVanSchayck: Agreed. Fixed now

Koenkk:
Thanks!

PR #7728, opened 2024-07-03
fix(detect): Fix fingerprint.modelID for E2006

indrajitr:
Fix `fingerprint.modelID` that was set incorrectly in #7717.


Koenkk:
Thanks!

PR #7730, opened 2024-07-03
Add curtain motor _TZE200_libht6ua to tuya.ts

cptjhmiller:
Add curtain motor with ID _TZE200_libht6ua (same as _TZE204_1fuxihti)


Koenkk:
Thanks!

PR #7740, opened 2024-07-05
New device support. Ledron QS-Zigbee-D06-DC.

dinamik1488:
New device support. Ledron QS-Zigbee-D06-DC.
Dimmer 12-24-36v.
https://svetomaniya.ru/product/00000017044


Koenkk:
Thanks!

PR #7741, opened 2024-07-05
Add extra fingerprint for TS0601 soil sensor

syphernl:
My [TS0601 Soil Sensor](https://www.zigbee2mqtt.io/devices/TS0601_soil_2.html) identifies as `_TZE284_sgabhwa6`. After creating an external converter based off the `_TZE284_g2e6cpnw` it reports data:

image

Not 100% sure if they are the same as the Battery state is `null`


Koenkk:
Thanks!

PR #7743, opened 2024-07-06
Update tuya.ts 'silence' for PA-44Z

Koenkk:
Update tuya.ts 'silence' for PA-44Z


Koenkk:
Thanks!

src/devices/tuya.ts
RemyRoux, line 1: Would it be possible to do the same change for the "test" property in order to trigger the alarm?

PR #7748, opened 2024-07-07
zigbeetlc: Comfort parameters should scale at 100

robvanoostenrijk:
Users reported the Comfort parameters for temperature and humidity are wrongly scaled.

The ZigbeeTlc firmware multiplies temperature and comfort settings with a factor 10 internally, while the same is not applied to comfort parameters.

This adjusts the ZigbeeTlc converter to align with that implementation.

Thanks to @Bodengriller for pointing this out and debugging the cause.


robvanoostenrijk:
ZigbeeTlc firmware has received an update to resolve this issue instead.

PR #7753, opened 2024-07-08
ZigbeeTLc: Adjust parameter scaling to ZigbeeTLc firmware v0.1.2.2

robvanoostenrijk:
New ZigbeeTLc firmware v0.1.2.2 aligns the scaling of all parameters across the different supported sensor types.
This converter update matches the uniform scaling setting that is now applied in the firmware.


Koenkk:
Thanks!

PR #7755, opened 2024-07-10
Update tuya.ts

oroGithub:
Added 'datapoints.key' variables for new Tuya device '_TZE204_dapwryy7' ('5.8 GHz human presence sensor').


Koenkk:
Thanks!

PR #7762, opened 2024-07-12
Added new device ADEO ZB-SMART-PIRTH-V3

chpego:
Smart 4 in 1 sensor : measure and monitor temperature, humidity and brightness and/or detect a presence

PR with device image: https://github.com/Koenkk/zigbee2mqtt.io/pull/2889

### DIFF
--- a/src/devices/adeo.ts
+++ b/src/devices/adeo.ts
@@ -410,6 +410,13 @@ const definitions: Definition[] = [
             await reporting.batteryPercentageRemaining(endpoint);
         },
     },
+    {
+        zigbeeModel: ['ZB-SMART-PIRTH-V3'],
+        model: 'ZB-SMART-PIRTH-V3',
+        vendor: 'ADEO',
+        description: 'Smart 4 in 1 sensor',
+        extend: [battery(), identify(), illuminance(), temperature(), humidity(), iasZoneAlarm({"zoneType":"generic","zoneAttributes":["alarm_1","alarm_2","tamper","battery_low"]})],
+    },
 ];
 
 export default definitions;


src/devices/adeo.ts
```diff
@@ -410,6 +410,15 @@ const definitions: Definition[] = [
             await reporting.batteryPercentageRemaining(endpoint);
         },
     },
+    {
+        zigbeeModel: ['ZB-SMART-PIRTH-V3'],
+        model: 'ZB-SMART-PIRTH-V3',
```
Koenkk, line 415: Could you update this with the model number which is written on the device/product page?

src/devices/adeo.ts
```diff
@@ -410,6 +410,20 @@ const definitions: Definition[] = [
...
+        model: '83633205',
+        vendor: 'ADEO',
+        description: 'Smart 4 in 1 sensor',
+        fromZigbee: [fz.battery],
+        exposes: [e.battery()],
+        extend: [
+            illuminance(),
+            temperature(),
+            humidity(),
+            iasZoneAlarm({zoneType: 'generic', zoneAttributes: ['alarm_1', 'alarm_2', 'tamper', 'battery_low']}),
```
Koenkk, line 424: Could you check if both `'alarm_1', 'alarm_2'` are needed here? I expect that one can be removed.

Koenkk, line 424: What kind of dev
```suggestion
            iasZoneAlarm({zoneType: 'occupancy', zoneAttributes: ['alarm_1', 'alarm_2', 'tamper', 'battery_low']}),
```

This is an occupancy sensor it seems.

Koenkk:
Thanks!

PR #7768, opened 2024-07-15
Add support for GL-MC-002P

Ru4rc:
Added support for device: Gledopto Zigbee USB Mini LED Controller RGB+CCT (Pro)


Ru4rc:
I don't know why the check is failing, the device works though

Koenkk:
Thanks!

PR #7775, opened 2024-07-16
add "TOQCB2_settings" tuya.ts

bifi2090:
added "TOQCB2_settings"-section and link in definition "toZigbee" for device "TOQCB2-80"

(see: https://github.com/Koenkk/zigbee2mqtt/issues/21943#issuecomment-2228302694)

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -10099,7 +10099,7 @@ const definitions: Definition[] = [
                 .withUnit('µS/cm')
                 .withDescription('Electrical Conductivity maximal value')
                 .withValueMin(0)
-                .withValueMax(100),
+                .withValueMax(20000),
             e
                 .numeric('ec_min', ea.STATE_SET)
                 .withUnit('µS/cm')


src/devices/tuya.ts
```diff
@@ -525,6 +525,78 @@ const tzLocal = {
             }
         },
     } satisfies Tz.Converter,
+    TOQCB2_settings: {
```
Koenkk, line 528: Instead of adding a new converter, add them to `tuyaDatapoints` instead

Koenkk: Here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7775/files#diff-0157bb612c63e92c50f93116e1f50bf8e263030ed26d86caa28479159c2a9f4aR10660

bifi2090: ok, but i tought, they were allready in there? Could you give me an example, on how to do that ? Like i said, with the "tzsettings"-section added an used as external converter i can modify the relevant values. Out off the box on the other hand it didn´t worked.

Koenkk: I see they are already there indeed, so adding `TOQCB2_settings` shouldn't be needed, what doesn't work when removing it?

bifi2090: Ok, there is alot to write, so that I don't have to write everything again (😜)  please see 

https://github.com/Koenkk/zigbee2mqtt/issues/21943#issuecomment-2228302694

There i have explained the errors, respectively the differences between using a separate external-converter as a file and the expected native support in terms of making changes in the relevant settings.

Koenkk: This issue can be solved by adding the missing keys (e.g. `over_voltage_setting `) to https://github.com/Koenkk/zigbee-herdsman-converters/blob/7d058d9ae369853005a6533d8111fb390a1ed29e/src/lib/tuya.ts#L1090

bifi2090: Ok, lets test it out. I Close this and make a new PR. THX

PR #7779, opened 2024-07-18
Update wirenboard.ts

BurchuMikhail:
Fix missing data issue for "Occupancy" parameter


src/devices/wirenboard.ts
```diff
@@ -606,6 +606,7 @@ const definitions: Definition[] = [
         model: 'WB-MSW-ZIGBEE v.4',
         vendor: 'Wirenboard',
         description: 'Wall-mounted multi sensor',
+        fromZigbee: [fz.occupancy],
```
Koenkk, line 609: This shouldn't be needed due to ` occupancy(),` below

BurchuMikhail: When adding the line 'fromZigbee: [fromZigbee_1.default.occupancy],' to the file /node_modules/zigbee-herdsman-converters/devices/wirenboard.js, the Occupancy parameter shows the current value. Without adding this line, the Occupancy parameter is always NA. This behavior appeared since version 1.37.1.

Koenkk: Do you see e.g. `occupancy_l1` in the frontend -> device -> state tab which updates when not using `fz.occupancy`?

BurchuMikhail: At the moment (z2m 1.38.0) it looks like this:
```
{
 "activity_led_indicator": false,
 "humidity": 39.89,
 "illuminance": 25429,
 "illuminance_lux": 349,
 "last_seen": 1721803596549,
 "linkquality": 52,
 "noise": 41.69,
 "noise_detect_level": 50,
 "noise_detected": true,
 "noise_timeout": 60,
 "occupancy": false,
 "occupancy_default": true,
 "occupancy_level": 741,
 "occupancy_sensitivity": 50,
 "occupancy_timeout": 5,
 "state_l1": "OFF",
 "state_l2": "OFF",
 "state_l3": "OFF",
 "temperature": 29.74,
 "temperature_offset": 0,
 "th_heater": false,
 "update": {
 "installed_version": 62,
 "latest_version": 62,
 "state": "idle"
 },
 "update_available": false
}
```
Wherein:
`occupancy` - does not change,
`occupancy_default` - changes.

When making the above changes it looks like this:
`occupancy` - changes,
`occupancy_default` - changes.

The `occupancy_l1` is missing.

Koenkk: To fix it, remove `fz.occupancy` and use this below:

```
deviceEndpoints({
                endpoints: {default: 1, l1: 2, l2: 3, l3: 4, indicator: 5},
                multiEndpointSkip: ['occupancy'],
            }),
```

PR #7784, opened 2024-07-18
Tuya FS-05R dimmer support

juacas:
Tuya FS-05R dimmer support

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -528,6 +528,16 @@ const tzLocal = {
 };
 
 const fzLocal = {
+    FS05R_brightness: {
+        cluster: 'genLevelCtrl',
+        type: ['attributeReport', 'readResponse'],
+        convert: (model, msg, publish, options, meta) => {
+            if (msg.data.hasOwnProperty('61440')) {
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -794,6 +804,20 @@ const modernExtendLocal = {
 };
 
 const definitions: Definition[] = [
+    {
+        fingerprint: [
```
Koenkk, line 808: Doesn't this device work out of the box? It should be detected as https://github.com/Koenkk/zigbee-herdsman-converters/blob/1f1c553b9ce1c6bab4f53f365c0feba7cf66515b/src/devices/tuya.ts#L6820, for proper detection you can add a `tuya.whiteLabel` there (search `tuya.ts` for examples)

juacas: You're right. Sorry, my bad.

With zigbee2mqtt 1.36 this device had problems. I didn't realize that it was fixed in later versions.
Sorry again.

I'll try to add a tuya.whiteLabel for model FS-05R (I already added a photo to docs' repository).

PR #7786, opened 2024-07-19
Add typing for battery voltage.

Nerivec:
These names appear not very consistent, might want to also fix that?


src/lib/types.ts
```diff
@@ -88,6 +88,16 @@ export interface DiscoveryEntry {
     discovery_payload: KeyValue;
 }
 
+export type BatteryVoltage =
+    | '3V_2100'
```
Koenkk, line 92: To make the names consistent, I guess we can drop the voltage from it, and just use the range , e.g. `2100_3000`

Nerivec: Maybe the ones that are basically saying min_max but with a string should just be using min/max object instead?
`3V_2500`, `3V_2500_3200`, `3V_2850_3000`, `4LR6AA1_5v`.
Could even extend the min/max object and cover `3V_add_1V`, `Add_1V_42V_CSM300z2v2`.
That way, the string is reserved for special calculations.

```typescript
export function batteryVoltageToPercentage(voltage: number, option: BatteryVoltage | {min: number; max: number, vOffset?: number}) {
    let percentage = null;

    switch (option) {
        case '3V_2100': {
            if (voltage < 2100) {
                percentage = 0;
            } else if (voltage < 2440) {
                percentage = 6 - ((2440 - voltage) * 6) / 340;
            } else if (voltage < 2740) {
                percentage = 18 - ((2740 - voltage) * 12) / 300;
            } else if (voltage < 2900) {
                percentage = 42 - ((2900 - voltage) * 24) / 160;
            } else if (voltage < 3000) {
                percentage = 100 - ((3000 - voltage) * 58) / 100;
            } else if (voltage >= 3000) {
                percentage = 100;
            }

            percentage = Math.round(percentage);
            break;
        }
        case '3V_1500_2800': {
            percentage = 235 - 370000 / (voltage + 1);
            percentage = Math.round(Math.min(Math.max(percentage, 0), 100));
            break;
        }
        default: {
            if (typeof option === 'object') {
                // Generic converter that expects an option object with min and max values
                // I.E. meta: {battery: {voltageToPercentage: {min: 1900, max: 3000}}}
                percentage = toPercentage(voltage + (option.vOffset ?? 0), option.min, option.max);
                break;
            } else {
                // only to cover case where a BatteryVoltage is missing in this switch
                throw new Error(`Unhandled battery voltage to percentage option: ${option}`);
            }
        }
    }
}
```

Koenkk: Sounds good!

Koenkk:
Thanks!

PR #7789, opened 2024-07-21
feat(add): TS0601_cover_10 extra manufacturer

ajgon:
I have a few Z100 motors, and started connecting them recently. What I noticed is, they come in two flavors (two different manufacturer IDs), both working the same way (as far as I can tell).

Both are also wall-socket powered and are indistinguishable from outside, thus I'm adding second ID.


Koenkk:
Thanks!

PR #7803, opened 2024-07-27
delete TS0601_soil_2 in tuya.ts, and NAS-WV03B in neo.ts

baggiowu:
These two products belong to our company, but they have not achieved all functions, and the models are wrong. I will upload the products again in next pull request. Thanks.


Koenkk:
>  I will upload the products again in next pull request. Thanks.

baggiowu:
> > I will upload the products again in next pull request. Thanks.
> 
> Could you do that in this PR? The `_TZE284_g2e6cpnw` will become unsupported after this PR

OK. Thanks.

baggiowu:
`_TZE284_g2e6cpnw` added in commit fe94c5d5245079615782827a128950acfab1484c

Koenkk:
I don't see `_TZE284_g2e6cpnw` added in that commit?

src/devices/neo.ts
```diff
@@ -255,13 +277,178 @@ const definitions: Definition[] = [
...
+                [
+                    102,
+                    'humidity_alarm',
+                    tuya.valueConverterBasic.lookup({
+                        lower_alarm: tuya.enum(0),
+                        upper_alarm: tuya.enum(1),
+                        cancel: tuya.enum(2),
+                    }),
+                ],
+                [3, 'humidity', tuya.valueConverter.divideBy10],
```
setfire2, line 350: Shouldn't this be `[3, 'humidity', tuya.valueConverter.raw],`?
With divideBy10 I get measurements between 1.5% and 2.1%. That's almost like in the Sahara :-)
When I just watered the soil it shows 2.1%. I guess 21% makes much more sense.

src/devices/neo.ts
```diff
@@ -255,13 +277,178 @@ const definitions: Definition[] = [
...
+                [5, 'temperature', tuya.valueConverter.divideBy10],
+                [110, 'temperature_f', tuya.valueConverter.divideBy10],
+                [107, 'temperature_sensitivity', tuya.valueConverter.divideBy10],
+                [108, 'humidity_sensitivity', tuya.valueConverter.raw],
+                [103, 'max_temperature_alarm', tuya.valueConverter.divideBy10],
+                [104, 'min_temperature_alarm', tuya.valueConverter.divideBy10],
+                [105, 'max_humidity_alarm', tuya.valueConverter.raw],
+                [106, 'min_humidity_alarm', tuya.valueConverter.raw],
+                [109, 'schedule_periodic', tuya.valueConverter.raw],
+                [15, 'battery_percentage', tuya.valueConverter.raw],
```
setfire2, line 360: Battery level is not showing up for me in the Frontend even after hours.

setfire2 (changes requested):

src/devices/neo.ts
setfire2: By the way, I live in northern Germany, not the Sahara...

baggiowu: The value of humidity with this device  unit is 1%, and is range between 0% - 100%.

src/devices/neo.ts
baggiowu: > Battery level is not showing up for me in the Frontend even after hours.

Thanks. I have changed dpid 15 key words from 'battery_percentage' to 'battery',  and use e.battery() to expose it.

src/devices/neo.ts
baggiowu: > Shouldn't this be `[3, 'humidity', tuya.valueConverter.raw],`? With divideBy10 I get measurements between 1.5% and 2.1%. That's almost like in the Sahara :-) When I just watered the soil it shows 2.1%. I guess 21% makes much more sense.

Thanks. This should be '[3, 'humidity', tuya.valueConverter.raw]'.

setfire2: Very nice, thanks. The device now also notices differences of 1% humidity. With the "old" converter it showed changes only in steps of 5%.

src/devices/neo.ts
setfire2: I did that change as well in my local instance about 8 hours ago. Still no value for the battery level showing up in the frontend. But since the desciption says it might take up to 24 hours until a value shows up, it might make sense to wait a little longer.

setfire2:
@baggiowu: after days still no battery level showing up. Let me know if I can do anything to help debugging this. I'd love to see this working.

Issue #7807, opened 2024-07-28
Incorrect cluster in Ubisys dimmer specs


mberglun:
Should this cluster not be 0xFC01?
https://github.com/Koenkk/zigbee-herdsman-converters/blob/bbe18586cd13b500a7d97a5d177744f4e236b50c/src/lib/ubisys.ts#L149
based on the documentation of D1: https://www.ubisys.de/wp-content/uploads/ubisys-d1-technical-reference.pdf, page 16, section 7.2.

I have these devices, not quite sure how to test whether it works, but happy to do it if someone can provide instructions.

sjorge:
I might have messed this up with a refactor recently, let me double check.

sjorge:
Yep looks like I did, https://github.com/Koenkk/zigbee-herdsman/pull/1034/files#diff-ec1a2d43181075d54fd33395e767a90a8d5e66cc2778f3944130d6aa537992b7L4313

I'll do a PR to fix.

mberglun:
Thanks a lot for the quick fix!

PR #7809, opened 2024-07-30
fix: fix cluster id that got typoed in #7451

sjorge:
While moving the clusters out of zh into zhc in #7451 the cluster ID for manuSpecificUbisysDimmerSetup accidentally got changed from 0xfc01 -> 0xfc00. 

This PR corrects this.

Fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/7807


Koenkk:
Thanks!

PR #7815, opened 2024-08-01
Add Support TS0601 _TZE284_aao3yzhs soil sensor

Barneybaer84:


Koenkk:
Seems to be exactly the same as `TS0601_soil`, could you add it to that definition?

Barneybaer84:
TS0601_soil temperature have no divideBy10 but this sensor.

Koenkk:
I see, thanks!

Make sure that:
- The filename is `TS0601_soil_3.png`
- The size is 512x512

jjvelar:
Hi,
First of all, thanks for your work.
Any idea when this will be released into production?
Thanks and best regards

Koenkk:
Today!

jjvelar:
> Today!

Excellent timing! :-)

Quick question: I have been able to create a TS0601_soil_3.js file with the sensor configuration and include it in Z2M configuration.yaml as an external converter. Should I remove the js file and the external converter entry before updating Home Assistant's Z2M add-on, or it is not necessary to remove it at all?  

Thanks again.

jjvelar:
> Quick question: I have been able to create a TS0601_soil_3.js file with the sensor configuration and include it in Z2M configuration.yaml as an external converter. Should I remove the js file and the external converter entry before updating Home Assistant's Z2M add-on, or it is not necessary to remove it at all?

@Koenkk, I saw the update is available in Home Assistant. Thanks again.

I created a TS0601_soil_3.js file with the sensor configuration and included it in Z2M configuration.yaml as an external converter.

Should I remove the js file and the external converter entry before updating Home Assistant's Z2M add-on, or it is not necessary to remove it at all?

Koenkk:
You can remove the external converter after updating to z2m 1.40.0

jjvelar:
> You can remove the external converter after updating to z2m 1.40.0

Works like a charm. Thanks again.

PR #7812, opened 2024-08-01
add NAS-WV05B2 NAS-STH02B2

baggiowu:
I have fixed the errors in #7808. Thanks.

PR #7816, opened 2024-08-01
fix: _TZE284_g2e6cpnw / _TZE284_sgabhwa6 datapoint mappings

rare-magma:
The mapping applied in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7812 seems to match this device better than the previous one created in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7609

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1952,13 +1952,70 @@ const definitions: Definition[] = [
         fromZigbee: [tuya.fz.datapoints],
         toZigbee: [tuya.tz.datapoints],
         configure: tuya.configureMagicPacket,
-        exposes: [e.temperature(), e.soil_moisture(), e.battery(), tuya.exposes.batteryState()],
+        exposes: [
+            e.soil_moisture(),
+            e.numeric('temperature', ea.STATE).withUnit('°C').withValueMin(-10).withValueMax(60).withDescription('Soil temperature'),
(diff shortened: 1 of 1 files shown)


Koenkk:
The datapoints are very similar to those of the `NAS-STH02B2`, could try with that definition, and if it works move `NAS-STH02B2` to `tuya.ts` and add a `tuya.whitelabel` there?

rare-magma:
@Koenkk these 2 particular devices don't emit the datapoints 1 ('Soil electrical conductivity') and 4 ('Soil fertility'), I think it makes sense to have different definitions

Koenkk:
I see, thanks!

setfire2:
My `_TZE284_sgabhwa6` works really well with this patch. Only the battery state does not show up but I can live with it as long as the battery level in percent works, which it does. Soil Moisture changes are also noticed in 1% steps. Almost perfect, thanks a lot! Id' love to see this in the next release.

setfire2:
@goberhammer, @rare-magma: what is your battery level showing? Mine is showing 63,6% for a week already. This must not be wrong but I find it a little suspicious.

goberhammer:
Hi @setfire2 , I've waited a few days before responding to you to be sure: both my battery and my battery state read as "N/S" in the Exposes tab and are not present in the state packet

setfire2:
Thanks for the info @goberhammer. It seems that there still an issue with the battery information for that device. Here's what my panel for that sensor looks like:
moisturesensor

PR #7817, opened 2024-08-01
Make Lupus 12126 and 12127 OTA capable

cromefire:
There are OTA updates available and the devices should be and seem to be able to receive them (athough it seems minde is up-to-date, so I wasn't able to upload a new version)


cromefire:
They also don't seem to be able to configure the power on behavior, but I don't know how to fix that:
image

src/devices/lupus.ts
```diff
@@ -49,13 +49,15 @@ const definitions: Definition[] = [
         vendor: 'Lupus',
         description: '1 channel relay',
         extend: [onOff()],
```
Koenkk, line 51:
```suggestion
        extend: [onOff({powerOnBehavior: false})],
```

To fix the power on behaviour error

cromefire:
Applied the changes.

Koenkk:
Thanks!

PR #7818, opened 2024-08-02
Update Aqara WS-EUK03 power measurement

priitliivak:
Aqara exposes standard activePower attribute for their latest switches.


Koenkk:
Thanks!

PR #7823, opened 2024-08-03
Added references to Legrand 199142 (same as 064888) and BTicino 3577C (same as Legrand 067694)

xraver:


src/devices/legrand.ts
```diff
@@ -385,7 +385,7 @@ const definitions: Definition[] = [
     },
     {
         zigbeeModel: [' Micromodule switch\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000'],
-        model: '064888',
+        model: '064888 / 199142',
```
Koenkk, line 388: Maybe add it to `whiteLabel` instead?

xraver: Sorry for my late answer! I forgot this topic!
i did as suggested. this is much bettert! thanks

Koenkk:
Thanks!

PR #7824, opened 2024-08-04
Add _TZ3000_h1ipgkwn

rock3r:
A Tuya 2-way USB switch, sold under whitelabel by UNSH (no exact model number). See https://www.aliexpress.com/item/1005006406053419.html

Z2m detects it as https://www.zigbee2mqtt.io/devices/TS0002.html#tuya-ts0002, but that's obviously wrong. This device has no metering, too.

Auto-generated external definition, that I derived the PR from:

```js
const {deviceEndpoints, identify, onOff, electricityMeter} = require('zigbee-herdsman-converters/lib/modernExtend');

const definition = {
    zigbeeModel: ['TS0002'],
    model: 'TS0002',
    vendor: '_TZ3000_h1ipgkwn',
    extend: [deviceEndpoints({"endpoints":{"1":1,"2":2}}), identify(), onOff({"powerOnBehavior":false,"endpointNames":["1","2"]}), electricityMeter()],
    meta: {"multiEndpoint":true},
};

module.exports = definition;
```

I had to convert it since Tuya examples are in the old style. I added the external definition to z2m and tested it, seems to work fine.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -10953,6 +10953,27 @@ const definitions: Definition[] = [
             ],
         },
     },
+    {
+        fingerprint: [{
+                modelID: 'TS0002',
+                manufacturerName: '_TZ3000_h1ipgkwn',
+            },
+        ],
+        model: '2CH Zigbee USB Switch Module',
+        vendor: 'Tuya',
+        description: '2-way USB Smart Adapter',
+        configure: tuya.configureMagicPacket,
+        extend: [
+            tuya.modernExtend.tuyaOnOff({
+                endpoints: ['l1', 'l2']
+            }),
+        ],
+    	endpoint: (device) => {
+    		return {l1: 1, l2: 2};
+    	},
+    	meta: {multiEndpoint: true},
+    	whiteLabel: [tuya.whitelabel('UNSH', '[N/A]', '2CH Zigbee USB Switch Module', ['_TZ3000_h1ipgkwn'])],
+    }
 ];
 
 export default definitions;
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -10954,7 +10954,8 @@ const definitions: Definition[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


rock3r:
@Koenkk if you have any pointers on how to get this picked up by z2m, I'd be happy to improve the documentation, too.

rock3r:
BTW, I think that the power-on behaviour works, after testing it. I initially thought I'd exclude it, but I didn't find a way to do it. Looked at other devices but couldn't find similar ones.

I didn't look for Tuya specific codes but the basic functions seem to be fine (they were even when using the wrong device definition).

Koenkk:
> Z2m detects it as https://www.zigbee2mqtt.io/devices/TS0002.html#tuya-ts0002, but that's obviously wrong. This device has no metering, too.

What doesn't work with this definition? Because the code looks the same. If it's just the cosmetics, add the `tuya.whitelabel`

rock3r:
As mentioned in the description, it's an entirely different device (a multi-port USB switch). It doesn't support energy metering, despite the auto-generated definition saying it does. I tried poking all attributes and there is only one that can be read, but not written to.

rock3r:
If you want to test these by yourself, I am happy to cover the cost of purchase for the devices. Just ping me (my contacts are on https://sebastiano.dev/)

Koenkk:
Isn't it detected as this currently?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/4f10a79da513f3b573994df86a0201ca577d286a/src/devices/tuya.ts#L3637

rock3r:
Yes, it is. If you prefer only the white label, I can switch to that instead. I'll need to test it better but it should be ok — I don't have prior experience writing converters so I'll trust you on it :)

A question: should I still create a product page for the device, with image and everything?

Koenkk:
> A question: should I still create a product page for the device, with image and everything?

Yes, since that picture will be used in the z2m frontend and supported device page.

PR #7826, opened 2024-08-04
Add _TZ3000_mw1pqqqt

rock3r:
A Tuya 3-way USB switch, sold under whitelabel by UNSH (no exact model number). See https://www.aliexpress.com/item/1005006406053419.html

Z2m detects it as https://www.zigbee2mqtt.io/devices/TS0003.html, but that's obviously wrong. This device has no metering, too.

Auto-generated external definition, that I derived the PR from:

```js
const {deviceEndpoints, identify, onOff, electricityMeter} = require('zigbee-herdsman-converters/lib/modernExtend');

const definition = {
    zigbeeModel: ['TS0003'],
    model: 'TS0003',
    vendor: '_TZ3000_mw1pqqqt',
    extend: [deviceEndpoints({"endpoints":{"1":1,"2":2,"3":3}}), identify(), onOff({"powerOnBehavior":false,"endpointNames":["1","2","3"]}), electricityMeter()],
    meta: {"multiEndpoint":true},
};

module.exports = definition;
```

I added the external definition to z2m and tested it, seems to work fine.

> [!IMPORTANT]
> PR #7824 should be merged before this one.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -10953,6 +10953,27 @@ const definitions: Definition[] = [
             ],
         },
     },
+    {
+        fingerprint: [{
+                modelID: 'TS0002',
+                manufacturerName: '_TZ3000_h1ipgkwn',
+            },
+        ],
+        model: '2CH Zigbee USB Switch Module',
+        vendor: 'Tuya',
+        description: '2-way USB Smart Adapter',
+        configure: tuya.configureMagicPacket,
+        extend: [
+            tuya.modernExtend.tuyaOnOff({
+                endpoints: ['l1', 'l2']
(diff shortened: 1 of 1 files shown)


Koenkk:
> Z2m detects it as https://www.zigbee2mqtt.io/devices/TS0003.html, but that's obviously wrong. This device has no metering, too.

Same as the other, PR, what doesn't work with the TS0003 definition?

rock3r:
As mentioned in the description, it's an entirely different device (a multi-port USB switch). It doesn't support energy metering, despite the auto-generated definition saying it does. I tried poking all attributes and there is only one that can be read, but not written to.

PR #7829, opened 2024-08-06
fix: Use `endpointNames` for `electricityMeter` of ZGA003

vilaemail:
ZGA003 electricity meter readings do not work, as they are reported for each endpoint separately, but we are trying to read only one endpoint. This fix exposes all readings (supported by `electricityMeter`) and they work now.

See also https://github.com/Koenkk/zigbee2mqtt/discussions/23419#discussioncomment-10236839


Koenkk:
Thanks!

PR #7832, opened 2024-08-07
New 1 gang Tuya

nakata5321:
Hello!

Add new switch tuya brand. This switch is already recognized as ["TS0001" switch](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts#L3549) but this converter doesn't use all available features. And all already added similar switches doesn't have available features too. 
Since I can't check different with already added models, I've decide to add new switch.

If there are any options how to add this new switch to existed with full support of available features, Please provide the idea how to do it and I will update PD.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -3563,7 +3563,7 @@ const definitions: Definition[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Make sure that:
- The filename is `MODEL.png`
- The size is 512x512

PR #7833, opened 2024-08-07
detect _TZ3000_v4l4b0lp as TS0003_switch_3_gang_with_backlight

nakata5321:
Want to add this manufacturer number to this zigbee model to use all available features.


Koenkk:
Thanks!

PR #7835, opened 2024-08-07
update _TZE200_tsxpl0d0

electronicm:
Add ZM-105B-M as requested at https://github.com/Koenkk/zigbee2mqtt.io/pull/2933#issuecomment-2271932779


src/devices/tuya.ts
```diff
@@ -2282,7 +2282,8 @@ const definitions: Definition[] = [
             return {l1: 1, l2: 1};
         },
         whiteLabel: [
-            {vendor: 'Moes', model: 'MS-105B-M'},
+            tuya.whitelabel('Moes', 'MS-105B-M', '2 gang smart dimmer module', ['_TZE200_tsxpl0d0']),
```
Koenkk, line 2285: I expect that the `MS-105B-M` has a different manufname, so better to keep `{vendor: 'Moes', model: 'MS-105B-M'},` here

electronicm: Unfortunately, I did not find any indication of when to use [tuya.whitelabel](https://github.com/electronicm/zigbee-herdsman-converters/blob/919dbb15a3878309394d9c981c1a7fe5b0ef8856/src/lib/tuya.ts#L431-L437). So I took the line from `'KnockautX', 'FMD2C018'` and also rewrote `MS-105B-M` to standardize the code.
And now I have also seen my typo.

BTW: I have not found any indication that `MS-105B-M` is compatible with Zigbee. According to the manufacturer's website, datasheets, product pictures and manuals, it is a Wifi only device. Is it therefore not even out of place?

Koenkk: I've removed it, thanks for checking.

Koenkk:
Thanks!

PR #7837, opened 2024-08-07
Update giex.ts

ayasystems:


Koenkk:
Please add the new definition to the file (don't remove the existing entries)

PR #7839, opened 2024-08-07
chore: Convert ELKO 4523430 to use modern extend syntax

chdefrene:
Hi 👋

I'm preparing some modifications to this device, thought I'd clean it up a bit first by implementing the new `extend` syntax.

Please let me know if I overlooked anything (first committer and all that jazz...). I have this device installed in my house, so I have verified that it still functions as intended.

Finally, is my understanding correct that there does not currently exist any modern extend converters for the `text` and `climate` attributes?

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -1926,30 +1926,9 @@ const converters1 = {
             const result: KeyValueAny = {};
             const data = msg.data;
 
-            if (data.hasOwnProperty('elkoLoad')) {
-                // Load
-                result.load = data['elkoLoad'];
-            }
-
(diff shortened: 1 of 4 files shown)


Koenkk:
Nice cleanup, many thanks!

> Finally, is my understanding correct that there does not currently exist any modern extend converters for the text and climate attributes?

Not yet, feel free to create them.

overskeid:
I believe those changes to ctm.ts will break "mTouch One".

chdefrene:
> I believe those changes to ctm.ts will break "mTouch One".

Thanks for the heads up, I wasn't aware this was a separate device! I have reverted the changes in #7871.

PR #7841, opened 2024-08-08
Don't fail configure on unbind fail.

Nerivec:
Fix for https://github.com/Koenkk/zigbee2mqtt/issues/22492#issuecomment-2276204771 (trying to unbind non-existent bind)


Koenkk:
Let's create an option on the ZH endpoint unbind call to ignore `NO_ENTRY errors.

Nerivec:
Will be replaced with better logic in herdsman.

PR #7842, opened 2024-08-08
Add support for ClickSmart+ CSP042 2 gang switch

matthewsmithy:
Add support for ClickSmart+ CSP042 2 gang switch


Koenkk:
Thanks!

PR #7846, opened 2024-08-10
feat: Add modern extend for 'text' converter

chdefrene:
* Add modern extend for 'text' converter
* Convert ELKO 4523430 to use 'text' modern extend

Related fix:
* Use correct value for 'regulator_mode' for ELKO 4523430 (noticed different readings on the thermostat display vs. Z2M)


chdefrene:
I have not removed this `toZigbee` converter, since it is used for a device specific rule regarding max allowed characters.

https://github.com/Koenkk/zigbee-herdsman-converters/pull/7846/files#diff-51715221d06da600c3567799bdf03b1c7c25252c45217ff9af22702a4830dc32R43

Koenkk:
@chdefrene in order to get rid of the `toZigbee` converter, we can add an extra argument to the text modernExtend: `validate`, which is a optional function to validate the input.

chdefrene:
Sorry for the long wait, but I have applied the change now 🙂

Koenkk:
Thanks!

PR #7857, opened 2024-08-14
Add manufacturer for Tuya TS0601_fan_5_levels_and_light_switch

RafaelDiasCampos:
Added manufacturer for the Zigbee fan controller I bought. Tested and it works fine with the existing configuration.

image
image


Koenkk:
Thanks!

PR #7848, opened 2024-08-11
Add ZNCLDJ01LM Aqara C3 Curtain Motor

marthubner:
Adding support for Aqara C3 Curtain Motor based on my external converter presented [here](https://github.com/Koenkk/zigbee2mqtt/issues/19555#issuecomment-2269053169).

PR #7859, opened 2024-08-15
Fix Aqara C3 support

marthubner:
I accidentally switched a few things in my last pull request https://github.com/Koenkk/zigbee-herdsman-converters/pull/7848
This should fix it
Also removing redundant control converter as it causes some exceptions to pop up and the device can be controlled anyway


Koenkk:
Thanks!

PR #7860, opened 2024-08-15
Update GL-D-013P

marthubner:
This effectively adds supported Identify and Power-on-behavior features while retaining all functionality. Tested on GL-D-013P but it may be applicable to other Gledopto lights as there are many similar ones, but I only have this one and can't test it.


Koenkk:
Can you check if the transitions still work correctly. Especially if the time of the transition is correct.

marthubner:
Seems to be working correctly. When setting transition to 1, the light gradually turns on and off over 1 second period. Setting it to 5, it smoothly turns on and off over 5 seconds period. Anything else to check?

Koenkk:
Nope, thanks!

PR #7866, opened 2024-08-17
Fix E2013 from IKEA

baierjan:
Reintroduce both endpoint definition and bind the ssIasZone cluster to allow contact reporting right out of the box. The binding and endpoint definition was (probably unintentionally) removed in #7220 

Related to https://github.com/Koenkk/zigbee2mqtt/issues/22579

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -914,7 +914,8 @@ const definitions: Definition[] = [
         description: 'PARASOLL door/window sensor',
         extend: [
             addCustomClusterManuSpecificIkeaUnknown(),
-            bindCluster({cluster: 'genPollCtrl', clusterType: 'input'}),
+            deviceEndpoints({endpoints: {'1': 1, '2': 2}}),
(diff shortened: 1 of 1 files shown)


Koenkk:
Does this also fix the issue that the sensor goes offline?

baierjan:
It might, I can say for sure as I couldn't reproduce the issue in my setup. I suspect that the "goes offline" issue could be caused by the sensor not having anything to update.

Koenkk:
To double check, you post a screenshot of your sensors state tab? I want to see if there are any values there postfixed with `_1` or `_2`

baierjan:
This is what I currently see in the state tab:
```
{
    "ac_status": false,
    "battery": 83,
    "battery_defect": false,
    "battery_low": false,
    "contact": true,
    "last_seen": 1723979646618,
    "linkquality": 80,
    "restore_reports": false,
    "supervision_reports": false,
    "tamper": false,
    "test": false,
    "trouble": false,
    "update": {
        "installed_version": 16777241,
        "latest_version": 16777241,
        "state": "idle"
    }
}
```

Koenkk:
perfect, thanks!

PR #7868, opened 2024-08-18
add support for _TZ3000_prits6g4

DarkShortie:
Add _TZ3000_prits6g4 to TS0001_switch_module.

Device was tested and is currently operated with a copy of this definition as an external converter.


Koenkk:
Thanks!

PR #7871, opened 2024-08-18
Fixes for ELKO Super TR + ctm_thermostat

chdefrene:
Publishing fixes for the issues raised in #7839 and #7846 in a separate pull request, so they will get resolved more quickly.


Koenkk:
Thanks!

PR #7872, opened 2024-08-18
Update Gledopto Pro LED controllers to standard light ModernExtend

marthubner:
Tested with GL-C-008P Mini, GL-C-006P Mini and GL-C-301P which is a 5in1 device and switches its identity between 003P, 006P, 007P, 008P and 009P.
All of these devices support power on behavior and identify.
Also tested transition which was not working for colors with the old code and works after this change.


Koenkk:
Thanks!

PR #7881, opened 2024-08-20
Fix for activity led activation problem for Wiren Board

BurchuMikhail:
Update wirenboard.ts


src/devices/wirenboard.ts
```diff
@@ -245,22 +254,21 @@ const tzLocal = {
...
+            return {state: {[key]: value}};
+        },
+        convertGet: async (entity, key, meta) => {
+            await entity.read('genBinaryOutput', ['presentValue']);
+        },
+    } satisfies Tz.Converter,
 };
 
 const sprutModernExtend = {
-    sprutActivityIndicator: (args?: Partial<modernExtend.BinaryArgs>) =>
```
Koenkk, line 251: Can you debug what goes wrong when using the modern extend? It's the preferred way to do this.

BurchuMikhail: When using a modern extend up to version 1.39.0 (inclusive) everything works correctly. But in version 1.39.1, modernExtend.js added the "autoDetectInputEndpoint" function, which leads to an unexpected result:
zh:controller:endpoint: ZCL command 0x7cc6b6fffe5f655b/1 genBinaryOutput.write({"presentValue":1}, {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false})
and I expect:
zh:controller:endpoint: ZCL command 0x7cc6b6fffe5f655b/5 genBinaryOutput.write({"presentValue":1}, {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false})
As far as I understand, the command is now sent to endpoint 1 instead of 5. The sensor freezes.
Using the "entity" function as in previous versions fixes this.

Koenkk: Could you provide the data/database.db entry of this device?

BurchuMikhail: database in case of clean installation of z2m 1.39.1:
[database.db.zip](https://github.com/user-attachments/files/16706252/database.db.zip)

database in the case of proposed fix:
[database.db.zip](https://github.com/user-attachments/files/16706389/database.db.zip)

Koenkk: It puzzles me why the code you propose in this PR sends it to ep 5 and not to 1, could you provide the debug log when setting the `activity_led` with your fix included?

BurchuMikhail: [2024-08-23.05-03-32.zip](https://github.com/user-attachments/files/16723744/2024-08-23.05-03-32.zip)

Koenkk: I found the issue, attempted to fix it in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7896

BurchuMikhail: Checked it out, problem's gone. Thank you.

PR #7882, opened 2024-08-20
Update candeo.ts

dhc25:
Added additional model reference to C204.
New ref = C-ZB-DM204
This ref replaces C204 in coming months.


dhc25:
New ref will be used on new units only. Existing units maintain C204

Koenkk:
hanks!

PR #7888, opened 2024-08-21
Updated sonoff.ts to support a rare version of SNZB-01 button

zsqduke:
Adding support to a rare version of the SNZB-01 button, model KF01, that I own. It shares the same look as WB01 but only supports single click. It uses the "ssIasZone" endpoint to report button presses so copying the WB01 description didn't work. Now the button works.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -657,9 +657,8 @@ const definitions: Definition[] = [
     },
     {
         zigbeeModel: ['KF01', 'KF-01'],
-        model: 'SNZB-01',
+        model: 'SNZB-01-KF',
         vendor: 'SONOFF',
-        whiteLabel: [{vendor: 'eWeLink', model: 'KF01'}],
         description: 'Wireless button',
         exposes: [e.battery(), e.action(['off', 'single',]), e.battery_voltage()],
         fromZigbee: [fz.command_status_change_notification_action, fz.battery],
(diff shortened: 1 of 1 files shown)


src/devices/sonoff.ts
```diff
@@ -655,6 +655,22 @@ const definitions: Definition[] = [
             await reporting.batteryPercentageRemaining(endpoint, {min: 3600, max: 7200});
         },
     },
+    {
+        zigbeeModel: ['KF01', 'KF-01'],
+        model: 'SNZB-01',
```
Koenkk, line 660: The model name has to be unique, maybe use `SNZB-01-KF` here?

src/devices/sonoff.ts
```diff
@@ -655,6 +655,22 @@ const definitions: Definition[] = [
...
+        vendor: 'SONOFF',
+        whiteLabel: [{vendor: 'eWeLink', model: 'KF01'}],
```
Koenkk, line 662: Can be removed

zsqduke: 661 and 662, these two lines are better removed? Is that what you mean because I'm completely noob. I thought they were used to identify the device. 

After I change the model, as I understand it's a new model so I will need to do PR on zigbee2mqtt to update the device database?

Koenkk: Only the `whiteLabel` line can be removed. For the device database, a PR should be made with the picture:

Make sure that:
- The filename is `SNZB-01-KF.png`
- The size is 512x512

Koenkk:
Thanks!

PR #7898, opened 2024-08-26
Adding inching switch feature to Tuya devices

guidugli:
Tuya multi-gang switches may have extra features like inching switch, which turns off the switch after the specified amount of time. Tuya switches are supposed to be from 1 to 6 gangs.

This pull request implements inching switch and adds the feature to TS0002_zemismart device (the device also supports countdown, so it was added to it, since the feature was added recently).

REFERENCE:
https://developer.tuya.com/en/docs/connect-subdevices-to-gateways/tuya-zigbee-multiple-switch-access-standard?id=K9ik6zvnqr09m
https://developer.tuya.com/en/docs/connect-subdevices-to-gateways/tuya-zigbee-measuring-smart-plug-access-standard?id=K9ik6zvofpzqk#title-8-TUYA_COMMON_PRIVATE%20cluster
https://developer.tuya.com/en/docs/iot-device-dev/Jog-switch-Wi-Fi?id=Kaung27s713rc


src/devices/tuya.ts
```diff
@@ -8441,6 +8441,27 @@ const definitions: Definition[] = [
             tuya.whitelabel('RoomsAI', 'RoomsAI_37022454', '1 Gang switch with backlight', ['_TZ3000_w1tcofu8']),
         ],
     },
+    {
+        fingerprint: [{modelID: 'TS0002', manufacturerName: '_TZ3000_ywubfuvt'}],
+        model: 'TS0002_zemismart',
+        vendor: 'Tuya',
+        description: '2-Gang switch with backlight, countdown and inching',
+        extend: [
+            tuya.modernExtend.tuyaOnOff({powerOnBehavior2: true, backlightModeOffOn: true, onOffCountdown: true, endpoints: ['l1', 'l2']}),
```
Koenkk, line 8450: Since it's tightly coupled the on/off, could you add it as an option of the `tuyaOnOff` modern extend?

src/devices/tuya.ts
```diff
@@ -8441,6 +8441,27 @@ const definitions: Definition[] = [
             tuya.whitelabel('RoomsAI', 'RoomsAI_37022454', '1 Gang switch with backlight', ['_TZ3000_w1tcofu8']),
         ],
     },
+    {
+        fingerprint: [{modelID: 'TS0002', manufacturerName: '_TZ3000_ywubfuvt'}],
```
Koenkk, line 8445: I would say, let's not at a different definition but add it to the standard `TS0002`, worst case it only works for some

src/devices/tuya.ts
guidugli: No problem, will do it.

src/devices/tuya.ts
guidugli: Ok

guidugli: Any suggestion on how to unify entries for models TS0002_switch_module_3, TS0002_switch_module_1, TS0002, TS0002_switch_module, TS0002_switch_module_2, TS0002_switch_2_gang, TS0002_switch_module_4? I believe all these models could converge on few entries as TS0002 (all features), TS0002_basic (on/off), TS0002_limited (switch type, indicatormode, backlightmodeoffon)
Reasoning: any non-mapped device TS0002 would inherit all features. Some features do not impact the device like, on my device, although it does not have an indicator led, indicatormode has no effect (it does not break it). 
I have a 4 and 6 gang device that do not implement the more advanced features, so similar 2 gang devices could use TS0002_limited. If the device is so basic that does not even have features like poweronbehavior, then they can use the TS0002_basic.


  | on/off | poweronbehavior2 | indicatormode | backlightmodeoffon | onOffCountdown | SwitchType
-- | -- | -- | -- | -- | -- | --
TS0002 | X |   |   |   |   |  
TS0002_switch_2_gang | X | X |   | X |   |  
TS0002_switch_module | X |   |   |   |   | X
TS0002_switch_module_1 | X |   | X | X |   |  
TS0002_switch_module_2 | X |   | X |   |   | X
TS0002_switch_module_4 | X | X | X | X |   |  
TS0002_switch_module_3 | X | X | X |   |   |

Koenkk: Sounds good!

Koenkk:
Thanks!

PR #7906, opened 2024-08-26
added _TZE204_iuk8kupi tuya

shirou93:
added _TZE204_iuk8kupi. It's propably same device like _TZE200_iuk8kupi. Only new revision


Koenkk:
Thanks!

PR #7907, opened 2024-08-27
Standarize namron smart plugs

taraldb:
Devices are currently implemented differently. Attempt to standarize behaviour of the two plugs (white and black).

First time contributor, please check extra carefully!

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -4,8 +4,8 @@ import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as constants from '../lib/constants';
 import * as exposes from '../lib/exposes';
-import {forcePowerSource, light, onOff} from '../lib/modernExtend';
-import * as ota from '../lib/ota';
+import {forcePowerSource, light, onOff, ota, identify, electricityMeter} from '../lib/modernExtend';
+import * as ota_old from '../lib/ota';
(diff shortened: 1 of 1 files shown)


src/devices/namron.ts
```diff
@@ -139,27 +139,29 @@ const definitions: Definition[] = [
...
     },
+    {
+        zigbeeModel: ['4512766'],
+        model: '4512766',
+        vendor: 'Namron',
+        description: 'Zigbee smart plug 16A',
+        extend: [
+            ota(),
+            onOff(),
+            identify(),
```
taraldb, line 150: Might have to be removed? Same for L+163

src/devices/namron.ts
```diff
@@ -4,8 +4,8 @@ import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as constants from '../lib/constants';
 import * as exposes from '../lib/exposes';
-import {forcePowerSource, light, onOff} from '../lib/modernExtend';
-import * as ota from '../lib/ota';
+import {forcePowerSource, light, onOff, ota, identify, electricityMeter} from '../lib/modernExtend';
+import * as ota_old from '../lib/ota';
```
taraldb, line 8: renamed to avoid name conflict. There is probably another best practice somewhere that I'm not aware off.

Koenkk:
Thanks!

PR #7915, opened 2024-08-29
Add support for new ubisys S1-R (Series 2).

lgraf:
Adds support for the new ubisys [S1-R (Series 2)](https://smarthome-store.de/power-switch-s1-r.html). Currently a technical reference document for the device is not (yet) available. The main difference seems to be that there is no endpoint `4` in the new device and the energy related clusters are located in endpoint `1` (instead of `4` like the regular S1-R).

The device seems to be working fine:
```
{
    "energy": 0.01,
    "last_seen": "2024-08-29T19:51:12+02:00",
    "linkquality": 116,
    "power": 6.6,
    "state": "ON",
    "action": null
}
```

One thing  i am unsure about is if the fingerprint is working correctly. When i test this `Definition` as external converter with only the `fingerprint` defined (no `zigbeeModel` attribute), the normal S1-R `Definition` is used. As external converter the new `Definition` is only picked up when i trump with the `zigbeeModel` attribute (`S1-R (5601)`). Not sure if this is normal behaviour for external converters or if there something wrong about the fingerprinting?

**db entry (paired with the Definition of this PR)**
```
{
    "id": 9,
    "type": "Router",
    "ieeeAddr": "0x001fee000000a8ea",
    "nwkAddr": 34094,
    "manufId": 4338,
    "manufName": "ubisys",
    "powerSource": "Mains (single phase)",
    "modelId": "S1-R (5601)",
    "epList": [
        1,
        2,
        3,
        232,
        242
    ],
    "endpoints": {
        "1": {
            "profId": 260,
            "epId": 1,
            "devId": 266,
            "inClusterList": [
                0,
                3,
                4,
                5,
                6,
                1794,
                2820
            ],
            "outClusterList": [],
            "clusters": {
                "genBasic": {
                    "attributes": {
                        "modelId": "S1-R (5601)",
                        "manufacturerName": "ubisys",
                        "powerSource": 1,
                        "zclVersion": 8,
                        "appVersion": 1,
                        "stackVersion": 1,
                        "hwVersion": 16,
                        "dateCode": "20240612-DE-FB1",
                        "swBuildId": "2.5.0"
                    }
                },
                "seMetering": {
                    "attributes": {
                        "multiplier": 1,
                        "divisor": 10000
                    }
                },
                "genOnOff": {
                    "attributes": {
                        "onOff": 0
                    }
                }
            },
            "binds": [
                {
                    "cluster": 1794,
                    "type": "endpoint",
                    "deviceIeeeAddress": "0x00124b002c3a8c02",
                    "endpointID": 1
                },
                {
                    "cluster": 6,
                    "type": "endpoint",
                    "deviceIeeeAddress": "0x00124b002c3a8c02",
                    "endpointID": 1
                }
            ],
            "configuredReportings": [
                {
                    "cluster": 1794,
                    "attrId": 1024,
                    "minRepIntval": 5,
                    "maxRepIntval": 3600,
                    "repChange": 1,
                    "manufacturerCode": null
                },
                {
                    "cluster": 6,
                    "attrId": 0,
                    "minRepIntval": 0,
                    "maxRepIntval": 65000,
                    "repChange": 1,
                    "manufacturerCode": null
                }
            ],
            "meta": {}
        },
        "2": {
            "profId": 260,
            "epId": 2,
            "devId": 260,
            "inClusterList": [
                0,
                3
            ],
            "outClusterList": [
                3,
                5,
                6,
                8,
                768,
                64514
            ],
            "clusters": {},
            "binds": [
                {
                    "cluster": 6,
                    "type": "endpoint",
                    "deviceIeeeAddress": "0x001fee000000a8ea",
                    "endpointID": 1
                },
                {
                    "cluster": 6,
                    "type": "endpoint",
                    "deviceIeeeAddress": "0x00124b002c3a8c02",
                    "endpointID": 1
                },
                {
                    "cluster": 8,
                    "type": "endpoint",
                    "deviceIeeeAddress": "0x00124b002c3a8c02",
                    "endpointID": 1
                },
                {
                    "cluster": 768,
                    "type": "endpoint",
                    "deviceIeeeAddress": "0x00124b002c3a8c02",
                    "endpointID": 1
                }
            ],
            "configuredReportings": [],
            "meta": {}
        },
        "3": {
            "profId": 260,
            "epId": 3,
            "devId": 260,
            "inClusterList": [
                0,
                3
            ],
            "outClusterList": [
                3,
                5,
                6,
                8,
                768,
                64514
            ],
            "clusters": {},
            "binds": [
                {
                    "cluster": 6,
                    "type": "endpoint",
                    "deviceIeeeAddress": "0x00124b002c3a8c02",
                    "endpointID": 1
                },
                {
                    "cluster": 8,
                    "type": "endpoint",
                    "deviceIeeeAddress": "0x00124b002c3a8c02",
                    "endpointID": 1
                },
                {
                    "cluster": 768,
                    "type": "endpoint",
                    "deviceIeeeAddress": "0x00124b002c3a8c02",
                    "endpointID": 1
                }
            ],
            "configuredReportings": [],
            "meta": {}
        },
        "232": {
            "profId": 260,
            "epId": 232,
            "devId": 1287,
            "inClusterList": [
                0,
                61,
                64512,
                64599
            ],
            "outClusterList": [
                3,
                25
            ],
            "clusters": {},
            "binds": [],
            "configuredReportings": [],
            "meta": {}
        },
        "242": {
            "profId": 41440,
            "epId": 242,
            "devId": 97,
            "inClusterList": [],
            "outClusterList": [
                33
            ],
            "clusters": {},
            "binds": [],
            "configuredReportings": [],
            "meta": {}
        }
    },
    "appVersion": 1,
    "stackVersion": 1,
    "hwVersion": 16,
    "dateCode": "20240612-DE-FB1",
    "swBuildId": "2.5.0",
    "zclVersion": 8,
    "interviewCompleted": true,
    "meta": {
        "configured": 332242049
    },
    "lastSeen": 1724949732485
}
```

### DIFF
--- a/src/devices/ubisys.ts
+++ b/src/devices/ubisys.ts
@@ -735,11 +735,11 @@ const definitions: DefinitionWithExtend[] = [
                 endpoints: [{ID: 1, profileID: 260, deviceID: 266, inputClusters: [0, 3, 4, 5, 6, 1794, 2820], outputClusters: []}],
             },
         ],
-        model: 'S1-R (Series 2)',
+        model: 'S1-R-2',
         vendor: 'ubisys',
         description: 'Power switch S1-R (Series 2)',
         extend: [
-            deviceEndpoints({endpoints: {'1': 1, '2': 2, '3': 3, '232': 232}}),
+            deviceEndpoints({endpoints: {'1': 1, '2': 2, '3': 3, '232': 232}, multiEndpointSkip: ['state', 'power', 'energy']}),
             identify(),
             onOff({powerOnBehavior: false}),
             electricityMeter({cluster: 'metering', configureReporting: false}),
@@ -753,7 +753,6 @@ const definitions: DefinitionWithExtend[] = [
             await reporting.readMeteringMultiplierDivisor(endpoint);
             await reporting.instantaneousDemand(endpoint);
         },
-        meta: {multiEndpointSkip: ['state', 'power', 'energy']},
         onEvent: async (type, data, device) => {
             /*
              * As per technical doc page 18 section 7.3.4
--- a/src/devices/ubisys.ts
+++ b/src/devices/ubisys.ts
@@ -39,6 +39,15 @@ const ubisysOnEventReadCurrentSummDelivered = async function (type: OnEventType,
     }
 };
 
+const ubisysPollCurrentSummDelivered = async (type: OnEventType, data: OnEventData, device: Zh.Device, endpointId: number, options: KeyValue) => {
+    const endpoint = device.getEndpoint(endpointId);
+    const poll = async () => {
+        await endpoint.read('seMetering', ['currentSummDelivered']);
+    };
+
+    utils.onEventPoll(type, data, device, options, 'currentSummDelivered-poll', 60, poll);
+};
+
 const ubisys = {
     fz: {
         dimmer_setup: {
@@ -747,13 +756,14 @@ const definitions: DefinitionWithExtend[] = [
             commandsLevelCtrl({endpointNames: ['2', '3']}),
             commandsColorCtrl({endpointNames: ['2', '3']}),
         ],
+        options: [exposes.options.measurement_poll_interval()],
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ['seMetering']);
             await reporting.readMeteringMultiplierDivisor(endpoint);
             await reporting.instantaneousDemand(endpoint);
         },
-        onEvent: async (type, data, device) => {
+        onEvent: async (type, data, device, settings) => {
             /*
              * As per technical doc page 18 section 7.3.4
              * https://www.ubisys.de/wp-content/uploads/ubisys-s1-technical-reference.pdf
@@ -769,7 +779,7 @@ const definitions: DefinitionWithExtend[] = [
                 const ep2 = device.getEndpoint(2);
                 ep2.addBinding('genOnOff', ep1);
             } else {
-                await ubisysOnEventReadCurrentSummDelivered(type, data, device);
+                await ubisysPollCurrentSummDelivered(type, data, device, 1, settings);
             }
         },
     },
--- a/src/devices/ubisys.ts
+++ b/src/devices/ubisys.ts
@@ -7,6 +7,7 @@ import * as constants from '../lib/constants';
 import * as exposes from '../lib/exposes';
 import * as legacy from '../lib/legacy';
 import {logger} from '../lib/logger';
+import {commandsColorCtrl, commandsLevelCtrl, commandsOnOff, deviceEndpoints, electricityMeter, identify, onOff} from '../lib/modernExtend';
 import * as ota from '../lib/ota';
 import * as reporting from '../lib/reporting';
 import {DefinitionWithExtend, Fz, OnEventType, Tz, OnEventData, Zh, KeyValue, KeyValueAny} from '../lib/types';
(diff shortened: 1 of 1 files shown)


src/devices/ubisys.ts
```diff
@@ -725,6 +726,54 @@ const definitions: Definition[] = [
...
+             * https://www.ubisys.de/wp-content/uploads/ubisys-s1-technical-reference.pdf
+             *
+             * This cluster uses the binding table for managing command targets.
+             * When factory fresh, this cluster is bound to endpoint #1 to
+             * enable local control.
+             *
+             * We use addBinding to 'record' this default binding.
+             */
+            if (type === 'deviceInterview') {
+                const ep1 = device.getEndpoint(1);
```
Koenkk, line 769: This should be done in the `configure()`

src/devices/ubisys.ts
```diff
@@ -725,6 +726,54 @@ const definitions: Definition[] = [
...
+            deviceEndpoints({endpoints: {'1': 1, '2': 2, '3': 3, '232': 232}}),
+            identify(),
+            onOff({powerOnBehavior: false}),
+            electricityMeter({cluster: 'metering', configureReporting: false}),
+            commandsOnOff({endpointNames: ['2', '3']}),
+            commandsLevelCtrl({endpointNames: ['2', '3']}),
+            commandsColorCtrl({endpointNames: ['2', '3']}),
+        ],
+        configure: async (device, coordinatorEndpoint) => {
+            const endpoint = device.getEndpoint(1);
```
Koenkk, line 751: Everything inside here is not needed because of `electricityMeter`

src/devices/ubisys.ts
```diff
@@ -725,6 +726,54 @@ const definitions: Definition[] = [
...
+            await reporting.bind(endpoint, coordinatorEndpoint, ['seMetering']);
+            await reporting.readMeteringMultiplierDivisor(endpoint);
+            await reporting.instantaneousDemand(endpoint);
+        },
+        meta: {multiEndpointSkip: ['state', 'power', 'energy']},
```
Koenkk, line 756: Use `multiEndpointSkip` of `deviceEndpoints` instead

src/devices/ubisys.ts
```diff
@@ -725,6 +726,54 @@ const definitions: Definition[] = [
...
+    {
+        // S1-R Series 2 uses the same modelId as the regular S1-R, but the energy clusters are located in endpoint 1 (instead of 4, like the regular S1-R).
+        fingerprint: [
+            {
+                manufacturerName: 'ubisys',
+                modelID: 'S1-R (5601)',
+                endpoints: [{ID: 1, profileID: 260, deviceID: 266, inputClusters: [0, 3, 4, 5, 6, 1794, 2820], outputClusters: []}],
+            },
+        ],
+        model: 'S1-R (Series 2)',
```
Koenkk, line 738:
```suggestion
        model: 'S1-R-2',
```

Can we use this here? It's a bit more friendly without the spaces

src/devices/ubisys.ts
lgraf: I think its needed because `configureReporting` is set to `false` for `electricityMeter`. The reason is that the `currentSummDelivered` is not bindable for this device (as far i can tell for the most ubisys devices). As workaround only the `instantaneousDemand` attribute is binded manually and `currentSummDelivered` read in `onEvent`.

src/devices/ubisys.ts
lgraf: My understanding is that this default binding should only done during the initial device pairing (factory default).

src/devices/ubisys.ts
lgraf: done

src/devices/ubisys.ts
lgraf: done

src/devices/ubisys.ts
Koenkk: I see now, I think we can improve the `ubisysOnEventReadCurrentSummDelivered` a bit, I see 2 issues with it:
- In case the Ubisys devices is sending a lot of messages, a lot of reads are done
- In case no messages are send, the value becomes stale

We can solve this by polling at a certain interval, you can use the Tuya one as an [example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/cea2c45bd37d4408702e52a476dea5c2bab730d4/src/lib/tuya.ts#L165)

lgraf: Sounds good. I will take a closer look at the weekend when I have the device for testing at hand.

lgraf: I was a bit uncertain to change this for devices i do not own myself, so i added a separate method that can be used for polling.

Koenkk:
Thanks!

PR #7923, opened 2024-09-01
FIX: THE01860A

Burki24:
The reason behind the fix:
The fz converter used in the part 'TS0222_temperature_humidity' does not return correct values. This device needs the fz converter: fromZigbee_1.default.humidity


Koenkk:
Thanks!

PR #7929, opened 2024-09-02
Improved battery level accuracy with updated voltage-to-percentage calculation

RubenKelevra:
- Updated the voltage-to-percentage conversion to provide a more realistic view of battery levels, especially at lower voltages.
- The previous method sometimes showed full or half-full batteries even when they were actually low.

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/8499

## Comparison

| Voltage | Old Conversion Percentage | New Conversion Percentage |
|---------|---------------------------|---------------------------|
| 2600    | 12.4                       | 0.0                       |
| 2650    | 14.4                       | 0.0                       |
| 2700    | 16.4                       | 10.0                      |
| 2750    | 19.5                       | 25.0                      |
| 2800    | 27.0                       | 40.0                      |
| 2850    | 34.5                       | 55.0                      |
| 2900    | 42.0                       | 70.0                      |
| 2950    | 71.0                       | 80.0                      |
| 3000    | 100.0                      | 90.0                      |
| 3050    | 100.0                      | 95.0                      |
| 3100    | 100.0                      | 97.5                      |
| 3150    | 100.0                      | 98.8                      |
| 3200    | 100.0                      | 100.0                     |

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -265,18 +265,20 @@ export function getKey<T>(object: {[s: string]: T} | {[s: number]: T}, value: T,
 
 export function batteryVoltageToPercentage(voltage: number, option: BatteryNonLinearVoltage | BatteryLinearVoltage): number {
     if (option === '3V_2100') {
-        let percentage: number = 100; // >= 3000
+        let percentage: number;
 
-        if (voltage < 2100) {
+        if (voltage >= 3200) {
+            percentage = 100;
+        } else if (voltage >= 3000) {
+            percentage = 90 + ((voltage - 3000) * (100 - 90)) / (3200 - 3000);
+        } else if (voltage >= 2900) {
+            percentage = 70 + ((voltage - 2900) * (90 - 70)) / (3000 - 2900);
+        } else if (voltage >= 2800) {
+            percentage = 40 + ((voltage - 2800) * (70 - 40)) / (2900 - 2800);
+        } else if (voltage >= 2700) {
+            percentage = 10 + ((voltage - 2700) * (40 - 10)) / (2800 - 2700);
+        } else {
             percentage = 0;
-        } else if (voltage < 2440) {
-            percentage = 6 - ((2440 - voltage) * 6) / 340;
-        } else if (voltage < 2740) {
-            percentage = 18 - ((2740 - voltage) * 12) / 300;
-        } else if (voltage < 2900) {
-            percentage = 42 - ((2900 - voltage) * 24) / 160;
-        } else if (voltage < 3000) {
-            percentage = 100 - ((3000 - voltage) * 58) / 100;
         }
 
         return Math.round(percentage);
--- a/test/utils.test.js
+++ b/test/utils.test.js
@@ -120,8 +120,12 @@ describe('lib/utils.js', () => {
         });
 
         it('gets non-linear', () => {
-            expect(utils.batteryVoltageToPercentage(2300, '3V_2100')).toStrictEqual(4);
-            expect(utils.batteryVoltageToPercentage(3000, '3V_2100')).toStrictEqual(100);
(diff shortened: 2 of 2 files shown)


RubenKelevra:
@Koenkk I guess it would make sense to adapt the "3V_1500_2800" conversion as well?

Koenkk:
Where is the calculation based on?

RubenKelevra:
> Where is the calculation based on?

I've explained that here:

https://github.com/Koenkk/zigbee2mqtt/issues/8499#issuecomment-2323616976

It's a mix of two things: Remaining battery energy capacity specs by battery manufacturers and the fact that most devices won't accept 1.5 V and will often struggle with 2.6 V. So the remaining battery energy capacity below 2.7 V is ignored - because it's not useable.

For specs see:

https://www.duracell.com/wp-content/uploads/2021/06/CR2032CH_245mAh-CS.pdf
https://www.cr2032.co/cms/prodimages/panasonic_cr2032_datasheet.pdf

Koenkk:
I see, note that this change won't affect Lumi devices (which are often referred to in https://github.com/Koenkk/zigbee2mqtt/issues/8499) since those use e.g. `meta: {battery: {voltageToPercentage: {min: 2850, max: 3000}}},`

RubenKelevra:
Where do min/max come from? Does the device send those?

Koenkk:
> Like is that the measuring width of the voltage meter on the device or something like that?

I believe these values came from https://github.com/Koenkk/zigbee2mqtt/issues/8499, everything below 2850 mv is considered dead

RubenKelevra:
Ah, alright.

I've looked through my dead battery container yesterday, and they were either below 1 V or between 2.6-2.7 Volt, measured with a multimeter. So most of my sensors must have kept going until 2.7 or 2.6 Volt — despite being made by Xiaomi.

So my suggestion would be to keep the minimum voltage variable, which can be set if we know a specific voltage at which the device will usually stop working.

If it is set the lower end of the percentage calculation is modified to reflect this.

The upper end however isn't needed.

Koenkk:
> So my suggestion would be to keep the minimum voltage variable, which can be set if we know a specific voltage at which the device will usually stop working.

Maybe I'm not understanding it, but that's what we do right? It may be a bit on the high side but the voltage at which the Xiaomi devices dies is somewhere between 2600-2850mv.

RubenKelevra:
> > So my suggestion would be to keep the minimum voltage variable, which can be set if we know a specific voltage at which the device will usually stop working.
> 
> Maybe I'm not understanding it, but that's what we do right? It may be a bit on the high side but the voltage at which the Xiaomi devices dies is somewhere between 2600-2850mv.

Well, I was thinking more about a way to specify the minimum operation voltage for a device. So if users report back, that a device always dies at 2900 mW instead of the expected 2650 mW, we can set a variable and the calculation adjusts to show 2900 mW as 0% for this device.

This would avoid that certain device types always die with 40% battery left or something like this.

Koenkk:
That's already possible, see https://github.com/Koenkk/zigbee-herdsman-converters/blob/00ed002588706b8bcb20f84681ff51dc0df098cc/src/devices/bosch.ts#L1124

RubenKelevra:
> That's already possible, see https://github.com/Koenkk/zigbee-herdsman-converters/blob/00ed002588706b8bcb20f84681ff51dc0df098cc/src/devices/bosch.ts#L1124

Okay, but isn't that a simple linear conversion? 🤔


I was thinking more about the ability for adding a low end where the device turns off, and the function recalculates the percentage based on the energy content until this point. 🤔

Koenkk:
It is a simple linear conversion indeed. 

> I was thinking more about the ability for adding a low end where the device turns off, and the function recalculates the percentage based on the energy content until this point.

Can't that achieved by setting the `min`?

PR #7931, opened 2024-09-02
Fix missing water leak  in RB-SRAIN01

rdorys:
add missing exposure of water leak in RB-SRAIN01 (newly supported device)


Koenkk:
Thanks!

PR #7935, opened 2024-09-02
Fix wrong model in GUI for giex GX02 (_TZE200_7ytb3h8u)

idogrf:
`giex GX02` was wrongly displayed as model `QT06_2` in the GUI when zigbee manufacturer is `_TZE200_7ytb3h8u`
this issue was reported here - https://github.com/Koenkk/zigbee2mqtt/issues/21844

before
image

after
image

### DIFF
--- a/src/devices/giex.ts
+++ b/src/devices/giex.ts
@@ -85,7 +85,7 @@ const definitions: DefinitionWithExtend[] = [
                 .withUnit('sec')
                 .withDescription('Cycle irrigation interval'),
         ],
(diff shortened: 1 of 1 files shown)


jochenruben:
Thanx! Do we have to pair again, or will it be sufficient to install the next release?

idogrf:
> Thanx! Do we have to pair again, or will it be sufficient to install the next release?

I don't think you'll have to re-pair.
It's a simple UI change, there's no real change in how the device works (my guess is that internally `QT06_2` and `GX02` are the same
Once this PR is approved and released, you can install the new version and it should appear as the right device.

Koenkk:
Thanks!

PR #7943, opened 2024-09-04
fix(detect): Detect `_TZB210_ayx58ft5` as MiBoxer E2-ZR #23379

pvyleta:
The original code that fixed #21185 added the device erroneously under TS0505B, which we are fixing here. Tested by manually editing the tuya.js on my running HomeAssistant, which allowed me to control my device successfully.


Koenkk:
thanks!

PR #7947, opened 2024-09-06
Add support for Paulmann 501.40 RGB remote control

Hykario:
Adds support for Paulmann 501.40 RGB remote control
image


Koenkk:

Make sure that:
- The filename is `501.40.png`
- The size is 512x512

EpicLPer:
Heya, first of all thanks for adding this @Hykario :)
I noticed a bit of an issue tho, not sure if it's a firmware quirk or if the implementation is a bit off in Z2M.

When enabling the group switches at the bottom everything works fine as long as just 1 is enabled. However enabling 2 or more groups only sends an action to the lowest group being enabled.

As example, if group 2 and 4 are switched on and you press power it only sends a command to group 2, but not group 4. If group 2 is then switched off and only 4 is enabled it sends a command to group 4 just fine.

freeDom-:
I agree with @EpicLPer. When multiple groups are selected the device receives an additional action unknown event in addition tk the others.

Screenshot_20241127-211723_1

Futthermore the refresh button is not working and also triggering an unknown action.

Would it be possible to also send events (for group 0) when no group is selected? This would increase the amount of programmable buttons. It could also be used for most commonly used stuff, because you would save one additional button press.

Hykario:
`z2m: Received Zigbee message from 'LightRemote', type 'commandOn', cluster 'genOnOff', data '{}' from endpoint 2 with groupID 0`  
`z2m: Received Zigbee message from 'LightRemote', type 'commandOn', cluster 'genOnOff', data '{}' from endpoint 3 with groupID 0`  
`z2m: Received Zigbee message from 'LightRemote', type 'commandOn', cluster 'genOnOff', data '{}' from endpoint 4 with groupID 0`  

seems like the remote is actually sending data for every endpoint. 
I will add this as soon as i got Time, hopefully this Weekend.

freeDom-:
Thanks for your analysis! Sad that the manufacturer did not see the potential for another 14 programmable buttons :-D.

CrazyBavarian:
Maybe somebody can help me how can i get action_hue status into an entity? i have only a entity from the action?
 
thanks for support
Screenshot 2024-11-28 225852
Screenshot 2024-11-28 230138

PR #7959, opened 2024-09-08
fix: add all device endpoints to fingerprint

lgraf:
In relation to: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7915

I only added the one different device endpoint to the fingerprint, thought that was enough. But the fingerprint does not match. I added now all device endpoints, so that the fingerprint will match.

Thanks and sorry for the inconvenience.


Koenkk:
Thanks!

PR #7965, opened 2024-09-09
Improve compatibility with _TZ3000_303avxxt DIN rail switch

octa22:
this PR improves compatibility with _TZ3000_303avxxt DIN rail switch
I have found out, that the behaviour is exactly the same as _TZ3000_zjchz7pd


Koenkk:
Thanks!

PR #7966, opened 2024-09-09
Bugfix for Tuya ZWT198/ZWT100-BH using _TZE204_xnbkhhdr, auto and manual are reversed in preset datapoint.

friyin:
I installed an Avatto wall thermostat (TS0601 _TZE204_xnbkhhdr) and I realized that auto and manual where reversed in preset datapoint in my device, checking the code I noticed that where if branches were the same.


Koenkk:
Makes sense, thanks!

PR #7968, opened 2024-09-10
feat(add): 929003116301, 929003116401, 929003116501, 929003116601

LaurentvdBos:
Add support for the Perifo light tubes. I have the small black one, generated an external definition and added gradient support like all other Philips gradient bulbs.

I took the opportunity to also add support for its larger and/or white siblings.


LaurentvdBos:
Small sidenote: z2m added `identify()` to the list of extends, but since that was missing for many of the Philips bulbs, I did not add it (it also adds it for Hue bulbs that I already have, which do not have such a block in converters).

I rarely use the identify button, so I won't miss it, but is this something to add to the bulbs that support it?

Issue #7669, opened 2024-06-21
Ubisys C4 - Commands not correctly sent to coordinator


chris-1243:
Some commands like `brightness_move_down_s_` or `brightness_stop_s_` are not read by the coordinator. After spending some time in the converter, the C4 is the only device from Ubisys using the `legacy.fz.ubisys.` in `fromZigbee`. All other devices are using the "standard" `fromZigbee` like `fz.command_move` or `fz.command_stop`.

I played with an external converter and the `brightness_move` commands are correctly read by the coordinator while using the non `legacy.fz.ubisys`.

PR #7972, opened 2024-09-10
Update ubisys.ts

chris-1243:
It is a following of #7669. I did not want to modify `legacy.fz.ubisys_c4`. It seems something is not working and I have the feeling it would be better to use what is already created.
This way, it is quite easy to roll back. In the future, it might be a good idea to clear the code for this device. I do not have the knowledge for this unfortunately.


chris-1243:
Anyone could check where the error is ? For the moment, I have mounted the whole ubysis.ts file in my docker container and it is working

```
Code style issues found in the above file
``` 

is the error I get during the check.

Koenkk:
Thanks

PR #7976, opened 2024-09-12
update 3r PLUG cluster

3reality-support:


3reality-support:
I don't know where the code is wrong, I checked punctuation, spaces, indentation marks

PR #7981, opened 2024-09-13
Update ubisys.ts

chris-1243:
- Final move commands for C4 to default `fz.command` converter
- Rename Vendor / Manufacturer S1-R-2 in order to have only one brand while searching in the documentation


Koenkk:
Thanks!

PR #7983, opened 2024-09-13
Document unit of external temperature sensor

magicus:
Added documentation of the value of the external sensor.

### DIFF
--- a/src/devices/danfoss.ts
+++ b/src/devices/danfoss.ts
@@ -139,7 +139,8 @@ const definitions: DefinitionWithExtend[] = [
                             'property *at most* every 5 minutes or every 0.1K change in measured room temperature.',
                     )
                     .withValueMin(-8000)
-                    .withValueMax(3500),
+                    .withValueMax(3500)
+                    .withUnit('0.01 °C'),
                 e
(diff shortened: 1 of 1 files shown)


magicus:
Based on https://github.com/Koenkk/zigbee2mqtt.io/pull/3017

src/devices/danfoss.ts
```diff
@@ -139,7 +139,8 @@ const definitions: DefinitionWithExtend[] = [
                             'property *at most* every 5 minutes or every 0.1K change in measured room temperature.',
                     )
                     .withValueMin(-8000)
-                    .withValueMax(3500),
+                    .withValueMax(3500)
+                    .withUnit('0.01 °C'),
```
Koenkk, line 143: This is not a valid unit, must be e.g. `°C`, maybe use that and in the description clarify it is `0.01 °C` to prevent a breaking change?

magicus: In that case I think it's better to leave the formal unit unspecified and just use a description on how the value is interpreted as I originally did.

Koenkk:
Thanks!

PR #7986, opened 2024-09-13
Increase poll interval of Livolo TI0001-hygrometer and TI0001-pir to 300 seconds

PaulVanSchayck:
I was having issues with the Livolo motion detector (TI0001-pir) having spurious detection events throughout the day. I noticed I was getting payload events every 10 seconds. On the same frequency as this poll interval was set. 

This is unlike the other Livolo devices in the same file, which are all have a 300 seconds interval. Except for the hygrometer.

After setting the TI0001-pir poll interval to 300 seconds, I don't have any false detection events anymore. I've noticed no adverse effects while testing the last 2 months. For consistency I've also put the hygrometer to 300 seconds.

Ps. I need to add `npm run build` to the `prepare` step inside `package.json`. Otherwise, I can't use zigbee-herdsman-converters from a github source in the docker build of zigbee2mqtt. Would it be worth adding that?

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
         "build": "tsc",
         "build-watch": "tsc --watch",
         "prepack": "npm run clean && npm run build",
-        "prepare": "husky && npm run build"
+        "prepare": "husky"
     },
     "author": "Koen Kanters",
     "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
         "build": "tsc",
         "build-watch": "tsc --watch",
         "prepack": "npm run clean && npm run build",
-        "prepare": "husky"
+        "prepare": "husky && npm run build"
     },
     "author": "Koen Kanters",
     "license": "MIT",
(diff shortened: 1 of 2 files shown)


Koenkk:
Bedankt!

> Ps. I need to add npm run build to the prepare step inside package.json. Otherwise, I can't use zigbee-herdsman-converters from a github source in the docker build of zigbee2mqtt. Would it be worth adding that?

It shouldn't be added to `prepare`, but you indeed need to `npm run build` before you can use it from z2m.

PaulVanSchayck:
> > Ps. I need to add npm run build to the prepare step inside package.json. Otherwise, I can't use zigbee-herdsman-converters from a github source in the docker build of zigbee2mqtt. Would it be worth adding that?
> 
> It shouldn't be added to `prepare`, but you indeed need to `npm run build` before you can use it from z2m.

Yes, but `zigbee-herdsman-converters` won't get build if you fetch it from a git source during the build of  `zigbee2mqtt`. Only when it's in the `prepare` step of `zigbee-herdsman-converters`. Maybe the correct way would be to copy (or mount) the `zigbee-herdsman-converters` directory into the container at build time, and point to this path in the package.json of `zigbee2mqtt`. I'll try that next time.

PR #7992, opened 2024-09-15
add status shutter for Legrand 067776

apages2:
modification of the converters file for Legrand in order to add the different states (up, down, stop) of the Legrand shutters


Koenkk:
Thanks!

effepiconcept:
but is it possible to set a position now? I still can't do it

PR #7995, opened 2024-09-16
fix: TO-Q-SY2-163JZT unable to set over voltage higher than 255

iaxon:
As described this comment
https://github.com/Koenkk/zigbee2mqtt/issues/21999#issuecomment-2053524181 it wasn't possible to set the over voltage protection over 255. That was due to a overflow.

The fix is tested in hassio-zigbee2mqtt.


Koenkk:
Thanks!

PR #7891, opened 2024-08-21
Modify the water shortage abnormal timeout valve setting and its name.

liubotongs:

PR #8002, opened 2024-09-18
Hint for required firmware for auto close valve in SONOFF SWV

N3rdix:
It was not mentioned in the PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/7891 that this feature requires a specific firmware version. This already caused some confusion on different channels.

This hint should help other users to solve the issue.


Koenkk:
Thanks!

PR #8004, opened 2024-09-18
Updated Tuya TS0205 (_TZ3210_up3pngle) (model TS0205_smoke_2) 

marcelhoogantink:
Updated Tuya TS0205 (manufacturer: '_TZ3210_up3pngle')  (model: 'TS0205_smoke_2') to use only ModernExtend.

After issues with some new bought smoke-sensors of this type, with last_seen and getting offline, I tried several configurations.
Long story short, I ended up with a working config using just modernExtend just a little different (more specific for this sensor)then the generated one.


src/devices/tuya.ts
```diff
@@ -811,6 +811,18 @@ const definitions: DefinitionWithExtend[] = [
...
+        fingerprint: [
+            {
+                modelID: 'TS0205',
+                manufacturerName: '_TZ3210_up3pngle',
+            },
+        ],
+        model: 'TS0205_smoke_2',
+        vendor: 'Tuya',
+        description: 'Smoke sensor',
+        extend: [battery(), iasZoneAlarm({zoneType: 'smoke', zoneAttributes: ['alarm_1', 'tamper']})],
```
Koenkk, line 824: Does the battery % work? It's not working for `TS0205`

marcelhoogantink: Yes it does 😃

Koenkk: Could you try with `battery({percentageReporting: false})` , re-pair the device and see if it still works? If yes, then you can apply these changes to the existing `TS0205` definition.

marcelhoogantink: First results  with ```battery({percentageReporting: false})```, after 5 hours:
Sensor is not reporting battery level by itself at moment of re-pair and has still not reported it after 5 hours.
image


(test on second sensor with this setting: after awakening this sensor ( trigger tamper by opening), and hitting the blue (image  )button, the battery % is shown. So it look like it does not work 100% with this ```battery({percentageReporting: false})``` setting. 

Now waiting for the first sensor to show tell it is still alive. (Normally the sensors did that after aprox. every 18 hours with uploaded settings)

Further results tomorrow ...

marcelhoogantink: So now, 29 hours after the re-pair, the device has not send anything; battery is still None/? and last-seen is still the same ...
I guess this shows that the ```battery({percentageReporting: false})``` does not work with the ```_TZ3210_up3pngle``` device. I checked the settings to be sure that all OK, and they are!

Last test: I woke up the device by opening it, triggering the tamper: ```last-seen``` is updated and battery line is added to the status (state.yaml) but value is still ```null```. After that, I woke up the device again and hit the (image  )button: battery-% appears (94%).

(PS: the other 4 devices, interviewed with my proposed settings wereworking OK and did send new battery % ans last-seen info during test-time.)

Koenkk: Then I would propose to use the following definition:

```
{
        zigbeeModel: ['TS0205'],
        model: 'TS0205',
        vendor: 'Tuya',
        description: 'Smoke sensor',
        whiteLabel: [
            {vendor: 'Tesla Smart', model: 'TSL-SEN-SMOKE'},
            {vendor: 'Dongguan Daying Electornics Technology', model: 'YG400A'},
            tuya.whitelabel('Tuya', 'TS0205_smoke_2', 'Smoke sensor', ['_TZ3210_up3pngle']),
        ],
        // Configure battery % fails
        // https://github.com/Koenkk/zigbee2mqtt/issues/22421
        extend: [battery({percentageReporting: false}), iasZoneAlarm({zoneType: 'smoke', zoneAttributes: ['alarm_1', 'tamper']})],
            if (device?.manufacturerName === '_TZ3210_up3pngle') {
                // Required for this version
                // https://github.com/Koenkk/zigbee-herdsman-converters/pull/8004
            }
        },
}
```

marcelhoogantink: So simplify by modernExtent is not a priority?

Koenkk:
Thanks!

PR #8007, opened 2024-09-19
Add a custom cluster for Perenio PEHPL0X

Stonos:
Fixes the `Cluster with name '64635' does not exist` error you would get when trying to update certain properties (https://github.com/Koenkk/zigbee2mqtt/issues/23995).


Koenkk:
Thanks!

PR #8009, opened 2024-09-20
idea with weekdays programs for tuya-moes thermostats

dukobpa3:
Just an idea but maybe we can go with it and improve.

Would be great to make proper control for this kind of scheduling because whole set of available thermostats have it.

Now we have 2 implementations and bot not much useful:
– long string for all values
– bunch of separate fields

### DIFF
--- a/src/devices/moes.ts
+++ b/src/devices/moes.ts
@@ -15,7 +15,7 @@ const fzZosung = zosung.fzZosung;
 const tzZosung = zosung.tzZosung;
 const ez = zosung.presetsZosung;
 
-const programText = (name: string) => e.text(name, ea.STATE_SET).withPattern(/\d\d:\d\d\/\d\d°C/);
+const programText = (name: string) => e.text(name, ea.STATE_SET).withPattern(/\d\d:\d\d\/\d\d°C/, 'Template is: 00:00/00°C');
 
 const definitions: DefinitionWithExtend[] = [
     {
--- a/src/devices/moes.ts
+++ b/src/devices/moes.ts
@@ -15,11 +15,7 @@ const fzZosung = zosung.fzZosung;
 const tzZosung = zosung.tzZosung;
 const ez = zosung.presetsZosung;
 
-const exposesLocal = {
-    hour: (name: string) => e.numeric(name, ea.STATE_SET).withUnit('h').withValueMin(0).withValueMax(23),
-    minute: (name: string) => e.numeric(name, ea.STATE_SET).withUnit('m').withValueMin(0).withValueMax(59),
-    program_temperature: (name: string) => e.numeric(name, ea.STATE_SET).withUnit('°C').withValueMin(5).withValueMax(35).withValueStep(0.5),
-};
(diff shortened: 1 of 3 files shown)


Dukobpa3:
I'm not using HA and any other visual dashboards except z2m so for me this whistles is important :)

src/lib/legacy.ts
```diff
@@ -6531,6 +6531,52 @@ const toZigbee2 = {
...
+            const getNumbers = (input: exposes.Text): number[] => {
+                input.checkPatternMatch();
+
+                const hourTemperature = input.property.split('/');
+                const hourMinute = hourTemperature[0].split(':', 2);
+                const clamp = (value: number, min: number, max: number): number => {
+                    return Math.max(min, Math.min(max, value));
+                };
+
+                return [clamp(parseInt(hourMinute[0]), 0, 23), clamp(parseInt(hourMinute[1]), 0, 59), clamp(parseInt(hourTemperature[1]) * 2, 5, 35)];
```
Dukobpa3, line 6558: I think here is not needed exception. Clamp is more than enough.

src/devices/moes.ts
```diff
@@ -122,7 +118,7 @@ const definitions: DefinitionWithExtend[] = [
             legacy.tz.moes_thermostat_deadzone_temperature,
             legacy.tz.moes_thermostat_max_temperature_limit,
             legacy.tz.moes_thermostat_min_temperature_limit,
-            legacy.tz.moes_thermostat_program_schedule,
+            legacy.tz.moes_thermostat_program_schedule_v2,
```
Koenkk, line 121: Let's not call it v2 but update the existing one.

src/lib/legacy.ts
```diff
@@ -6531,6 +6531,52 @@ const toZigbee2 = {
...
+
+            /* Merge modified value into existing state and send all over in one go */
+            const newProgram = {
+                // @ts-expect-error ignore
+                ...meta.state.program,
+                ...value,
+            };
+
+            const getNumbers = (input: exposes.Text): number[] => {
+                input.checkPatternMatch();
```
Koenkk, line 6550: I don't understand what this does, how can `input` be `exposes.Text` here? Isn't it a string?

src/devices/moes.ts
Dukobpa3: It is just draft but if you like the idea I'll change both moes/moesS thermostats for using this one with proper namings
(made pr as draft)

src/lib/legacy.ts
Dukobpa3: aww
Sure. I've just started to dive into it and missed some core concepts. I've thought there is component and not clean result.

But maybe you'll suggest where can I make this check? I wanted to try to incapsulate all info inside component but not sure where this information can be used. As I see a lot of this stuff doing in z2m-frontend?

Koenkk: I would you do a simple regex check here instead of ` input.checkPatternMatch();`

Dukobpa3: I can but main idea was to improve something and provide kind of better incapsulated solution

PR #8011, opened 2024-09-21
Improve support for the SUTON STB3L-125-ZJ DIN rail RCBO (TZE204_wbha…

octa22:
This PR improves support for the SUTON STB3L-125-ZJ DIN rail RCBO (TZE204_wbhaespm)

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -5567,6 +5567,69 @@ const definitions: DefinitionWithExtend[] = [
             tuya.exposes.currentWithPhase('c'),
             e.temperature(),
             e.binary('leakage_test', ea.STATE_SET, 'ON', 'OFF').withDescription('Turn ON to perform a leagage test'),
+            e
+                .binary('over_current_breaker', ea.STATE_SET, 'ON', 'OFF')
+                .withDescription('OFF - alarm only, ON - relay will turn off when threshold reached'),
+            e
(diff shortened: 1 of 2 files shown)


src/devices/tuya.ts
```diff
@@ -5529,7 +5669,7 @@ const definitions: DefinitionWithExtend[] = [
         vendor: 'SUTON',
         description: 'Zigbee DIN RCBO energy meter',
         fromZigbee: [tuya.fz.datapoints],
-        toZigbee: [tuya.tz.datapoints],
+        toZigbee: [tzLocal.TZE204_threshold, tuya.tz.datapoints],
```
Koenkk, line 5672: Please use just `tuya.tz.datapoints` instead of `TZE204_threshold`

octa22: OK, you mean to implement the logic in tuya.valueConvertor? Is there an example how to implement a setting logic for several exposed attributes behind a single dataPoint? Thanks

Koenkk: Yes, here is an example to send data points directly from the converter (https://github.com/Koenkk/zigbee-herdsman-converters/blob/615eb00f66267429fa8434a233b39f2259af3d1a/src/lib/tuya.ts#L977). With this you ca also send multiple.

Koenkk:
Thanks!

PR #8020, opened 2024-09-25
Add new Tuya 4-in-1 sensor model

ericvoshall:
PR to add new device model from issue reported in https://github.com/Koenkk/zigbee2mqtt/issues/24095.

Not sure if I need to do anything else for a PR or not.


Koenkk:
Thanks!

PR #8022, opened 2024-09-25
Change TS0601_soil_2 to NAS-STH01B2, and fix the dpid issues

baggiowu:
Changed model TS0601_soil_2 to NAS-STH01B2, and fix the TUYA DPID issues.


rare-magma:
@baggiowu I think it would be better to create another definition for the model you're having issues with. This converter is working fine for my device

src/devices/tuya.ts
```diff
@@ -2043,7 +2042,7 @@ const definitions: DefinitionWithExtend[] = [
                         cancel: tuya.enum(2),
                     }),
                 ],
-                [3, 'soil_moisture', tuya.valueConverter.raw],
+                [3, 'humidity', tuya.valueConverter.raw],
```
Koenkk, line 2045: Since this is a soil moisturising sensor, `soil_moisture ` looks fine to me

rare-magma:
@Koenkk / @baggiowu `humidity`, `battery_state` and `battery` are always reporting the same value on my device so I'm not sure if they actually work. The rest of the datapoints and calibration settings seem to be fine.

src/devices/tuya.ts
baggiowu: I think so too. But I've received some emails asking to change to humidity.

baggiowu:
> @Koenkk / @baggiowu `humidity`, `battery_state` and `battery` are always reporting the same value on my device so I'm not sure if they actually work. The rest of the datapoints and calibration settings seem to be [fine.](url)

image

PR #8028, opened 2024-09-26
Update efekta.ts

smartboxchannel:
added support for two new devices EFEKTA PST DUO V1 LR and EFEKTA PST DUO V1


src/devices/efekta.ts
```diff
@@ -762,6 +769,372 @@ const definitions: DefinitionWithExtend[] = [
...
+                valueOff: ['OFF', 0],
+                cluster: 'genPowerCfg',
+                attribute: {ID: 0x0205, type: 0x10},
+                description: 'Enable сontrol of comparison with previous data',
+            }),
+        ],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpointOne = device.getEndpoint(1);
+            const overrides0 = {min: 7200, max: 43200, change: 1};
+            await reporting.batteryAlarmState(endpointOne, overrides0);
```
Koenkk, line 952: I think this can also be added as a modernExtend?

smartboxchannel: It would be great, as long as it doesn't work through modernity. As soon as you implement it, I will add efekta.ts

Koenkk: Could you check if you can add this? Add `batteryAlarmStateReporting` to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/modernExtend.ts#L285

smartboxchannel:
so why was it closed? I initially published a fully working version of the converter.

src/devices/efekta.ts
smartboxchannel: image

Apparently there was something here, but now I don't see anything about batteryAlarmStateReporting.

PR #8031, opened 2024-09-26
Improve multi-endpoint support

### LINKED ISSUE #8027: `lib/philips.ts` and `lib/modernExtend.js` are only partially multi-endpoint aware
As a follow-up to https://github.com/Koenkk/zigbee2mqtt/issues/23652, I think I have discovered why the "naive" converter implementation only partially works. [From line 60 onwards](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/philips.ts#L60), `philipsLight` is not multi-endpoint aware.

The `power_on_behavior` expose handling in https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/modernExtend.ts#L1008 also does not take endpoints into account. (There might other functions affected.)

mundschenk-at:
Fixes #8027.

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -33,6 +33,7 @@ import {
     assertNumber,
     batteryVoltageToPercentage,
     configureSetPowerSourceWhenUnknown,
+    exposeEndpoints,
     flatten,
     getEndpointName,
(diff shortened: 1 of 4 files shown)


Koenkk:
Thanks!

PR #8036, opened 2024-09-27
Update tuya.ts

maniac-semal:
Added whitelabel for MTG235-ZB-RL (_TZE204_clrdrnya), which is currently identified as MTG075-ZB-RL.

Seperate pull request for a picture for this Type will also  be created. Hopefully I did it correctly.


Koenkk:
Thanks!

PR #8040, opened 2024-09-28
improvements for tint (Mueller Licht)

FernetMenta:
1) Expose action group for tint remote 404049x
2) Fix color temp warning when using predefined scene button on remote


Koenkk:
Thanks!

PR #8044, opened 2024-09-28
Update frient powermeter led 2

tbowmo:
This should fix https://github.com/Koenkk/zigbee2mqtt/issues/19042

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -740,8 +740,8 @@ const converters1 = {
         convert: (model, msg, publish, options, meta) => {
             if (utils.hasAlreadyProcessedMessage(msg, model)) return;
             const payload: KeyValueAny = {};
-            const multiplier = (msg.endpoint.getClusterAttributeValue('seMetering', 'multiplier') ?? 1) as number;
-            const divisor = (msg.endpoint.getClusterAttributeValue('seMetering', 'divisor') ?? 1000) as number;
+            const multiplier = msg.endpoint.getClusterAttributeValue('seMetering', 'multiplier') as number;
+            const divisor = msg.endpoint.getClusterAttributeValue('seMetering', 'divisor') as number;
             const factor = multiplier && divisor ? multiplier / divisor : null;
 
             if (msg.data.instantaneousDemand !== undefined) {
@@ -753,17 +753,15 @@ const converters1 = {
                 payload[property] = power;
             }
 
-            if (factor != null && (msg.data.currentSummDelivered !== undefined || msg.data.currentSummReceived !== undefined)) {
-                if (msg.data.currentSummDelivered !== undefined) {
-                    const value = msg.data['currentSummDelivered'];
-                    const property = postfixWithEndpointName('energy', msg, model, meta);
-                    payload[property] = value * factor;
-                }
-                if (msg.data.currentSummReceived !== undefined) {
-                    const value = msg.data['currentSummReceived'];
-                    const property = postfixWithEndpointName('produced_energy', msg, model, meta);
-                    payload[property] = value * factor;
-                }
+            if (msg.data.currentSummDelivered !== undefined) {
+                const value = msg.data['currentSummDelivered'];
+                const property = postfixWithEndpointName('energy', msg, model, meta);
+                payload[property] = value * (factor ?? 1);
+            }
+            if (msg.data.currentSummReceived !== undefined) {
+                const value = msg.data['currentSummReceived'];
+                const property = postfixWithEndpointName('produced_energy', msg, model, meta);
+                payload[property] = value * (factor ?? 1);
             }
 
             return payload;
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -740,8 +740,8 @@ const converters1 = {
         convert: (model, msg, publish, options, meta) => {
             if (utils.hasAlreadyProcessedMessage(msg, model)) return;
             const payload: KeyValueAny = {};
-            const multiplier = msg.endpoint.getClusterAttributeValue('seMetering', 'multiplier') as number;
-            const divisor = msg.endpoint.getClusterAttributeValue('seMetering', 'divisor') as number;
+            const multiplier = (msg.endpoint.getClusterAttributeValue('seMetering', 'multiplier') ?? 1) as number;
+            const divisor = (msg.endpoint.getClusterAttributeValue('seMetering', 'divisor') ?? 1000) as number;
             const factor = multiplier && divisor ? multiplier / divisor : null;
 
             if (msg.data.instantaneousDemand !== undefined) {
--- a/src/devices/frient.ts
+++ b/src/devices/frient.ts
@@ -9,7 +9,7 @@ const definitions: DefinitionWithExtend[] = [
         zigbeeModel: ['EMIZB-141'],
         model: 'EMIZB-141',
         vendor: 'Frient',
-        description: 'frient Electricity Meter Interface 2 LED',
+        description: 'Electricity meter interface 2 LED',
         extend: [
             ota(),
             electricityMeter({cluster: 'metering', power: {divisor: 1000, multiplier: 1}, energy: {divisor: 1000, multiplier: 1}}),
--- a/src/devices/frient.ts
+++ b/src/devices/frient.ts
@@ -9,7 +9,7 @@ const definitions: DefinitionWithExtend[] = [
     {
         zigbeeModel: ['EMIZB-141'],
         model: 'EMIZB-141',
-        vendor: 'Frient A/S',
+        vendor: 'Frient',
         description: 'frient Electricity Meter Interface 2 LED',
         extend: [
             ota(),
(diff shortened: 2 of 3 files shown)


src/devices/frient.ts
```diff
@@ -1,24 +1,120 @@
...
+            key: ['pulse_configuration'],
+            convertSet: async (entity, key, value, meta) => {
+                await entity.write('seMetering', {develcoPulseConfiguration: value}, utils.getOptions(meta.mapped, entity));
+                return {readAfterWriteTime: 200, state: {pulse_configuration: value}};
+            },
+            convertGet: async (entity, key, meta) => {
+                await entity.read('seMetering', ['develcoPulseConfiguration'], manufacturerOptions);
+            },
+        } satisfies Tz.Converter,
+        interface_mode: {
```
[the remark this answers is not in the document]
tbowmo: Hmmm.. perhaps, I'm searching for a more comprehensive example, where there are both getter and setter defined (ie both fromZigbee and toZigbee in the same modernExtend function).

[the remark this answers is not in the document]
tbowmo: @Koenkk I think I got it now

src/lib/develco.ts
```diff
@@ -170,4 +171,61 @@ export const develcoModernExtend = {
             valueIgnore: [0xffff, -0x8000],
             ...args,
         }),
+    current_summation: () =>
```
Koenkk, line 174: Could you refactor it to e.g. `numeric()`? Example can be found [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/ff9ad52aa88c7b275599ad568e66af9720aa5c0a/src/lib/develco.ts#L66)

src/devices/frient.ts
```diff
@@ -1,24 +1,31 @@
...
+            develcoModernExtend.readGenBasicPrimaryVersions(),
+            develcoModernExtend.pulse_configuration(),
+            develcoModernExtend.current_summation(),
+        ],
+
+        configure: async (device, coordinatorEndpoint, logger) => {
             const endpoint = device.getEndpoint(2);
             await reporting.bind(endpoint, coordinatorEndpoint, ['seMetering', 'genPowerCfg']);
+            await reporting.instantaneousDemand(endpoint);
+            await reporting.readMeteringMultiplierDivisor(endpoint);
```
Koenkk, line 28: I think the configure can be removed completely, both the binds and reporting are done by `electricityMeter()` and `batter()` modernExtend.

src/lib/develco.ts
tbowmo: current_summation is a write only entity, so I do not see any idea in converting that to `numeric()` as that seems like read/write?
But perhaps the pulse_configuration can be changed.. I'll look into it.

src/devices/frient.ts
tbowmo: What about endpoint specification for the modernExtend functions?

src/lib/develco.ts
Koenkk: We can add `access: 'SET'` option (also has to be added to `numeric`)

src/devices/frient.ts
Koenkk: If only endpoint 2 has the `seMetering` cluster, it should work out of the box, is this the case?

tbowmo: I think that is the case. So let's see how it goes without the configuration

src/lib/develco.ts
tbowmo: That is a bit our of my league at the moment.. Also looked into converting the pulse_config to something more generic (like numeric), but seems a bit overcomplicated for this pr (at the moment)

tbowmo:
@Koenkk I have tested the integration locally, and it works, at least I get energy etc. from the device

```json
{
    "battery": 100,
    "battery_low": false,
    "energy": 118.28,
    "interface_mode": "electricity",
    "linkquality": 112,
    "power": 421,
    "power_2": 357,
    "pulse_configuration": 1000,
    "update": {
        "installed_version": 196866,
        "latest_version": 196866,
        "state": "idle"
    },
    "voltage": 3100,
    "current_summation": null,
    "update_available": null
}
```

src/converters/fromZigbee.ts
```diff
@@ -740,8 +740,8 @@ const converters1 = {
         convert: (model, msg, publish, options, meta) => {
             if (utils.hasAlreadyProcessedMessage(msg, model)) return;
             const payload: KeyValueAny = {};
-            const multiplier = msg.endpoint.getClusterAttributeValue('seMetering', 'multiplier') as number;
-            const divisor = msg.endpoint.getClusterAttributeValue('seMetering', 'divisor') as number;
+            const multiplier = (msg.endpoint.getClusterAttributeValue('seMetering', 'multiplier') ?? 1) as number;
+            const divisor = (msg.endpoint.getClusterAttributeValue('seMetering', 'divisor') ?? 1000) as number;
```
tbowmo, line 744: @Koenkk one thought here, could it be an idea to convert this to Number, instead of doing `as number`?

like this instead
```
const divisor = Number(msg.endpoint.getClusterAttributeValue('seMetering', 'divisor') ?? 1000);
```

Koenkk: This value is always a number (or `undefined`) so it should not be needed.

src/devices/frient.ts
```diff
@@ -1,33 +1,32 @@
...
-        description: 'Smart powermeter Zigbee bridge',
-        fromZigbee: [fz.metering, fz.battery],
-        toZigbee: [],
-        extend: [ota()],
-        exposes: [e.battery(), e.power(), e.energy()],
-        configure: async (device, coordinatorEndpoint) => {
-            const endpoint = device.getEndpoint(2);
-            await reporting.bind(endpoint, coordinatorEndpoint, ['seMetering', 'genPowerCfg']);
-        },
+        description: 'frient Electricity Meter Interface 2 LED',
```
Koenkk, line 12:
```suggestion
        description: 'Electricity meter interface 2 LED',
```

src/lib/develco.ts
Koenkk: Can you try with:

```
        numeric({
            name: 'pulse_configuration',
            cluster: 'seMetering',
            attribute: 'develcoPulseConfiguration',
            description: 'Pulses per kwh. Default 1000 imp/kWh. Range 0 to 65535',
            access: 'ALL',
            unit: '%',
            valueMin: 0,
            valueMax: 65535,
        }),
```

src/converters/fromZigbee.ts
tbowmo: I had problems, because they where undefined, and so factor became null. Then on line 756, it checks if factor is not null, before calculating anything. That lead up to it reporting null as energy consumption.

When I added those default values, it was reporting like it should.

src/lib/develco.ts
tbowmo: I have added the suggestion. But am unable to test if it works at the moment (doesn't have access to the house where my device is located for at least a week)

src/converters/fromZigbee.ts
Koenkk: The divisor/multiplier can be forced in the configure. I would propose to have the same logic as `instantaneousDemand` here, expose raw value when no factor is defined

src/lib/develco.ts
```diff
@@ -170,4 +170,26 @@ export const develcoModernExtend = {
             valueIgnore: [0xffff, -0x8000],
             ...args,
         }),
+    current_summation: (args?: Partial<NumericArgs>) =>
```
Koenkk, line 173:
```suggestion
    currentSummation: (args?: Partial<NumericArgs>) =>
```

src/lib/develco.ts
```diff
@@ -170,4 +170,26 @@ export const develcoModernExtend = {
...
+            name: 'current_summation',
+            cluster: 'seMetering',
+            attribute: 'develcoCurrentSummation',
+            description: 'Current summation value sent to the display. e.g. 570 = 0,570 kWh',
+            access: 'SET',
+            valueMin: 0,
+            valueMax: 268435455,
+            ...args,
+        }),
+    pulse_configuration: (args?: Partial<NumericArgs>) =>
```
Koenkk, line 184:
```suggestion
    pulseConfiguration: (args?: Partial<NumericArgs>) =>
```

src/converters/fromZigbee.ts
tbowmo: hmm.. even though that I defined the metering entity like 
```js
  electricityMeter({cluster: 'metering', power: {divisor: 1000, multiplier: 1}, energy: {divisor: 1000, multiplier: 1}}),
```
I had trouble in the fromZigbee.metering function. This was the previous release though, so things might have changed now.

Anyways, I have changed the logic a bit, so it now returns the raw value.. I'll get around to test things in a couple of days.

tbowmo:
@Koenkk I have now tested the implementation, and it works. Both changing pulse counts, and setting current summation property.
So it's a go for a merge from me now.

Koenkk:
Thanks!

PR #8048, opened 2024-09-29
Fix a typo in `modernExtend.numeric.toZigbee.convertSet`

### LINKED ISSUE #8042: Scaling is ignored when `precision` is specified in a `modernExtend.numeric`.
In `toZigbee` scaling is effectively being ignored when `precision` is specified. I believe the correct approach would be to call `precisionRound()` _before_ scaling the value.

I'd be happy to make a PR although I do see an issue if the incoming value is not numeric. Not sure if that is a real concern... Any suggestions?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/f67d7ff214d452554d1de0dbc6afaa71ce8184c4/src/lib/modernExtend.ts#L2093-L2098

morgendagen:
Fixes #8042


Koenkk:
Thanks!

PR #8051, opened 2024-09-30
feat(add): add Tuya Air Quality sensor _TZE204_c2fmom5z

ernetas:
Not sure if the already existing `_TZE200_c2fmom5z` was not a typo, but `_TZE204_c2fmom5z` could be added as well.


Koenkk:
Thanks!

PR #8056, opened 2024-10-01
Support for _TZE200_rxq4iti9

nicoschreiner:
Add Avatto Thermostatic radiator valve with manufacturer id "_TZE200_rxq4iti9"


Koenkk:
Thanks!

PR #8060, opened 2024-10-01
Improve reporting E2206

Rembock:
Enables reporting of powerDivisor, so it's changes dynamically with the amount of power.
For details, see: https://github.com/Koenkk/zigbee2mqtt/issues/23961#issuecomment-2366733453


Rembock (commented):
Hopefully fixed the errors now

src/devices/ikea.ts
```diff
@@ -602,7 +608,30 @@ const definitions: DefinitionWithExtend[] = [
...
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), onOff(), identify(), ikeaOta(), electricityMeter()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), onOff(), identify(), ikeaOta()],
+        fromZigbee: [fz.electrical_measurement, fz.metering],
+        toZigbee: [tz.electrical_measurement_power, tz.acvoltage, tz.accurrent, tz.currentsummdelivered],
+        configure: async (device, coordinatorEndpoint) => {
+            const endpoint = device.getEndpoint(1);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['haElectricalMeasurement', 'seMetering']);
+            // Enable reporting of powerDivisor, needs to change dynamically with the amount of power
+            // For details, see: https://github.com/Koenkk/zigbee2mqtt/issues/23961#issuecomment-2366733453
+            await endpoint.configureReporting('haElectricalMeasurement', [
```
Koenkk, line 619: I'm wondering if we could keep everything as is with just this line added. I guess everything except the min and change should be the same. When pairing it, was `change` value do you see in the z2m frontend (note that it requires 1.40.2 as with older versions the change value appears empty on the frontend)

Rembock: You are absolutely right, I got rid of the unnecessary lines and was able to pair it successfully.
With version 1.40.2, the `change` value is now 50 for all attributes of `haElectricalMeasurement`, I assume you meant this? 
I had to provide attribute `acPowerDivisor` with a `min`, `max` & `change` value, as these were not filled automatically.

Koenkk: Does it make sense to change the `minimumReportInterval` to `1` or `0`? The reportable change makes sure the device does not report it constantly.

Rembock: Yes, if I leave it at 50 the `acPowerDivisor` won't change. It looks like it's reporting 2 values only, 1 and 10. So I thought this value was too high and changed it. Maybe there's another way, but this is the best result I could manage to get

Koenkk:
Thanks!

PR #8066, opened 2024-10-02
Bugfix for Avatto ZWT198 _TZE204_xnbkhhdr, fix reversed 6-1 and 5-2 in 'working_day' datapoint

B3WiN:
Problem: the 6-1 and 5-2 option is reserved with _TZE204_xnbkhhdr.

Fix the reserved working_day options for Avatto (Tuya) ZWT198 _TZE204_xnbkhhdr. It use the same method as _TZE204_lzriup1j.


Koenkk:
Thanks!

PR #8077, opened 2024-10-05
Fix 12 of the 13 Typescript inferred type cannot be named without a reference errors

clockbrain:
When compiling zigbee-herdsman-converters within zigbee2mqtt/node-modules I get 13 inferred type errors.

The steps to reproduce this compile error are simply:
```
cd zigbee2mqtt/node_modules
rm -rf zigbee-herdsman-converters
git clone https://github.com/Koenkk/zigbee-herdsman-converters.git
cd zigbee-herdsman-converters
npm ci
npm run build
```
The compile errors are:
```
Found 13 errors in 10 files.

Errors  Files
     2  src/converters/toZigbee.ts:4529
     1  src/lib/ledvance.ts:29
     1  src/lib/legacy.ts:8431
     1  src/lib/legrand.ts:118
     2  src/lib/lumi.ts:1432
     1  src/lib/ota/zigbeeOTA.ts:102
     1  src/lib/philips.ts:130
     1  src/lib/tuya.ts:1234
     2  src/lib/utils.ts:251
     1  src/lib/zosung.ts:211
```
I'm not 100% sure I have the correct fixes but at least the compiler doesn't complain with these additional type declarations.
One compile error remains that I wasn't able to fix:
```
src/lib/lumi.ts:1432:14 - error TS2742: The inferred type of 'lumiModernExtend' cannot be named without a reference to 'zigbee-herdsman-converters/node_modules/zigbee-herdsman/dist/controller/model'. This is likely not portable. A type annotation is necessary.

1432 export const lumiModernExtend = {
                  ~~~~~~~~~~~~~~~~
```

Interestingly these compile errors don't occur when downloading and building zigbee-herdsman-converters elsewhere, i.e. in a folder outside of zigbee2mqtt/node_modules.

### DIFF
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -4526,7 +4526,7 @@ const converters3 = {
     } satisfies Tz.Converter,
 };
 
-const converters = {...converters1, ...converters2, ...converters3};
+const converters: KeyValueAny = {...converters1, ...converters2, ...converters3};
 
 export default converters;
 module.exports = converters;
--- a/src/lib/ledvance.ts
+++ b/src/lib/ledvance.ts
@@ -26,7 +26,7 @@ export const ledvanceFz = {
     } satisfies Fz.Converter,
 };
 
-export const ledvanceTz = {
+export const ledvanceTz: KeyValue = {
     ledvance_commands: {
         /* deprecated osram_*/
         key: ['set_transition', 'remember_state', 'osram_set_transition', 'osram_remember_state'],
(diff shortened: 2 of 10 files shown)


Koenkk:
I would suggest not to put the zhc git repo in node_modules but create a symlink from node_modules instead.

clockbrain:
Yes, creating a symlink to zigbee-herdsman-converters works. So if there isn't any need for this PR I can just close it?

Koenkk:
Indeed, no need for a PR, thanks anyway!

PR #8083, opened 2024-10-05
Support for S32055 TW led driver/dimmer from The Light Group

TurboJens:
Tested, working great even as unsopported hardware.


Koenkk:
Thanks!

PR #8086, opened 2024-10-06
Adding PLUG COMPACT EU EM T

rj2s:
Working fine with ZB2MQTT but not yet supported

### DIFF
--- a/src/devices/ledvance.ts
+++ b/src/devices/ledvance.ts
@@ -46,7 +46,7 @@ const definitions: DefinitionWithExtend[] = [
         extend: [onOff(), electricityMeter()],
     },
     {
-        zigbeeModel: ['PLUG COMPACT OUTDOOR EU EM T'],
+        zigbeeModel: ['PLUG COMPACT OUTDOOR EU EM T', 'PLUG COMPACT EU EM T'],
         model: '4099854293276',
         vendor: 'LEDVANCE',
         description: 'SMART+ Compact outdoor plug EU with energy meter',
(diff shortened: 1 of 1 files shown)


src/devices/ledvance.ts
```diff
@@ -332,6 +332,15 @@ const definitions: DefinitionWithExtend[] = [
         description: 'SMART+ Outdoor Plug EU',
         extend: [ledvanceOnOff({powerOnBehavior: false})],
     },
+    {
+        zigbeeModel: ['PLUG COMPACT EU EM T'],
+        model: 'PLUG COMPACT EU EM T',
```
Koenkk, line 337: Use the model number written on the device here

src/devices/ledvance.ts
```diff
@@ -332,6 +332,15 @@ const definitions: DefinitionWithExtend[] = [
...
+        vendor: 'LEDVANCE',
+        description: 'Automatically generated definition',
```
Koenkk, line 339: Fill in a sensible description here.

src/devices/ledvance.ts
rj2s: Model is 4099854293276

src/devices/ledvance.ts
rj2s: Description should be : PLUG COMPACT EU EM T

Koenkk:
Merged it with the already existing 4099854293276, thanks!

PR #8093, opened 2024-10-07
fix legrand pairing attribute 0xf000

CodeForLove83:
fix legrand pairing attribute 0xf000 (operation time since the latest reset or power on)


Koenkk:
thanks!

PR #8094, opened 2024-10-08
Update shinasystem.ts

shinasys:
- PMM-300Z2 and PMM-300Z3 do not support acFrequencyDivisor and acFrequencyMultiplier attributes.


Koenkk:
Thanks!

PR #8099, opened 2024-10-08
Update tuya.ts

jonnysciar:
Add support for LoraTap SC400ZB-EU, same as LoraTap SC400 but supports different calibration times for up and down

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -2212,7 +2212,13 @@ const converters1 = {
             }
             if (msg.data.moesCalibrationTime !== undefined) {
                 const value = parseFloat(msg.data['moesCalibrationTime']) / 10.0;
-                result[postfixWithEndpointName('calibration_time', msg, model, meta)] = value;
+                if (meta.device.manufacturerName === '_TZ3000_cet6ch1r') {
+                    const endpoint = msg.endpoint.ID;
+                    const calibrationLookup: KeyValueAny = {1: 'up', 2: 'down'};
+                    result[postfixWithEndpointName(`calibration_time_'${calibrationLookup[endpoint]}'`, msg, model, meta)] = value;
+                } else {
+                    result[postfixWithEndpointName('calibration_time', msg, model, meta)] = value;
+                }
             }
             return result;
         },
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2556,13 +2556,24 @@ const definitions: DefinitionWithExtend[] = [
             tuya.whitelabel('Nous', 'B4Z', 'Curtain switch', ['_TZ3000_yruungrl']),
         ],
         exposes: (device) => {
-            const exps = [
+            const exps: Expose[] = [
                 e.cover_position(),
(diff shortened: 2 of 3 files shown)


src/devices/tuya.ts
```diff
@@ -2586,6 +2586,36 @@ const definitions: DefinitionWithExtend[] = [
...
+            tz.tuya_cover_reversal,
+            tuya.tz.backlight_indicator_mode_2,
+            tuya.tz.backlight_indicator_mode_1,
+        ],
+        meta: {coverInverted: true, multiEndpoint: true},
+        endpoint: (device) => {
+            return {up: 1, down: 2};
+        },
+        exposes: [
+            e.cover_position().withEndpoint('up'),
```
Koenkk, line 2608: It's not nice that also these exposes are under an endpoint. I propose to modify `tz.tuya_cover_calibration` instead to support up and down calibration. (e.g. `calibration_up`, `calibration_down`). Then we can also keep this in the original `TS130F` definition (add `calibration_up` and `calibration_down` exposes conditionally in the exposes function)

jonnysciar: I have implemented the feature following your suggestion.
I also update `fz.tuya_cover_options` with `calibration_time_up` and `calibration_time_down`

src/converters/fromZigbee.ts
```diff
@@ -2211,8 +2211,14 @@ const converters1 = {
                 result[postfixWithEndpointName('motor_reversal', msg, model, meta)] = reversalLookup[value];
             }
             if (msg.data.moesCalibrationTime !== undefined) {
+                const endpoint = msg.endpoint.ID;
                 const value = parseFloat(msg.data['moesCalibrationTime']) / 10.0;
-                result[postfixWithEndpointName('calibration_time', msg, model, meta)] = value;
+                if (endpoint === 2) {
```
Koenkk, line 2216: Let's only expose calibration time up/down for ``:

```
if (meta.device.manufacturerName === '_TZ3000_cet6ch1r') {
  ..
} else {
  ..
}
```

weichenb: Hi @jonnysciar,

In [issue #25996](https://github.com/Koenkk/zigbee2mqtt/issues/25996) there was a discussion to rename "calibration up" and "calibration down" into "calibration to open" and "calibration to close".

Currently I am working to include full support for LoraTap SC500ZB-v4. Beside dual way claibration, the extension will also support undocumented external switch type support.
Would you mind if I change the names accordingly?

Apropos LoraTap SC400 is still not included in device documentation. Will you still do?

weichenb: to provide a better impression, it could look like:
grafik

jonnysciar: Hi @weichenb, sorry for the late response.
To me it is okay to change 'Calibration Up' and 'Calibration Down' to 'Calibration Open' and 'Calibration Close'. It is more generic and reusable.

Regarding LoraTap SC400, in the documentation it should be mentioned as White-label in the [TS130F page ](https://www.zigbee2mqtt.io/devices/TS130F.html). I think it only misses the 'new' calibration fields, let me know if they should be added in that page or I have to create another page for the SC400.

PR #8104, opened 2024-10-09
Add Müller Licht Tint Aris

frauhottelmann:
Hey @Koenkk ,

Can you please check. I am pretty sure it won't work like this because the zigbeeModel already exists. But I don't know what else to do, because that is what I get from "generate_external_definition".

Thanks

### DIFF
--- a/src/devices/muller_licht.ts
+++ b/src/devices/muller_licht.ts
@@ -17,12 +17,12 @@ function mullerLichtLight(args: LightArgs) {
 
 const definitions: DefinitionWithExtend[] = [
     {
-    zigbeeModel: ['tint-ExtendedColor'],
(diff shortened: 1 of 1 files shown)


frauhottelmann:
Oh, and the manufacturer name is "MLI".

Koenkk:
This device should already be supported out of the box (it uses a very generic modelID so the picture/description will never be correct)

frauhottelmann:
> This device should already be supported out of the box (it uses a very generic modelID so the picture/description will never be correct)

Yes, it does work. Then I'll just close this. Do you still want to add a page to the documentation? I've created a pull request there as well. The reset is different to the Tint bulbs.

Koenkk:
Yes, but add it to `404036_45327_45317_45328.md` instead

PR #8108, opened 2024-10-10
Lidl HG06338 add feature On/Off USB

chpego:
Add the feature switch on/off of the 4 USBs that seems to have been omitted

From [ZHA](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts011f_plug.py#L358-L367), the quirk indicates the dndpoint number corresponding to the "socket" containing the 4 USBs

On ZHA it worked fine, and when I switched to Z2M I lost the On/Off function of its 4 USBs. I haven't yet been able to test it on Z2M, but I think it should be fine, don't you?

### DIFF
--- a/src/devices/lidl.ts
+++ b/src/devices/lidl.ts
@@ -353,16 +353,16 @@ const definitions: DefinitionWithExtend[] = [
         model: 'HG06338',
         vendor: 'Lidl',
         description: 'Silvercrest 3 gang switch, with 4 USB (EU, FR, CZ, BS)',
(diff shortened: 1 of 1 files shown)


chpego:
ok, but I'm 100% sure that the USB part is controllable since I was doing it with ZHA.

Koenkk:
Could you provide the data/database.db entry of your device?

chpego:
Sorry my Lidl HG06338 died a few days ago and there's nothing more I can do.

Koenkk:
If you paired it to z2m, it should still be in the data/database.db file.

PR #8112, opened 2024-10-10
Update Ikea TRADFRI LED1949C5 with E17/440 lm variant

eidolonjs:
These E17/440 lm variants of the already supported TRADFRI LED1949C5 bulbs are sold in Japan.


Koenkk:
Thanks!

PR #8119, opened 2024-10-11
Add Namron Edge dimmer 4512782 / 4512781

geirra:
This adds support for namron edge dimmer, some of the clusters doesnt report. So these are disabled


Koenkk:
Thanks!

PR #8120, opened 2024-10-12
Add ZigUSB_C6 support

xyzroe:


src/devices/xyzroe.ts
```diff
@@ -463,6 +483,117 @@ const definitions: DefinitionWithExtend[] = [
...
+                .binary('state', ea.ALL, 'ON', 'OFF')
+                .withProperty('state')
+                .withValueToggle('TOGGLE')
+                .withDescription('💡 Indicates the USB state via LED')
+                .withEndpoint('3'),
+        ],
+        extend: [
+            deviceEndpoints({endpoints: {1: 1, 2: 2, 3: 3}}),
+            identify(),
+            electricityMeter({
```
Koenkk, line 582: Please extend `electricityMeter` with `dc` support, then we can get rid of the `configure`

src/devices/xyzroe.ts
```diff
@@ -463,6 +483,117 @@ const definitions: DefinitionWithExtend[] = [
...
+                cluster: 'electrical',
+                configureReporting: false,
+                endpointNames: ['1'],
+            }),
+            temperature({
+                reporting: {
```
Koenkk, line 588: Why use custom reporting values here?

src/devices/xyzroe.ts
```diff
@@ -463,6 +483,117 @@ const definitions: DefinitionWithExtend[] = [
             ]);
         },
     },
+    {
+        zigbeeModel: ['ZigUSB_C6'],
+        model: 'ZigUSB_C6',
+        vendor: 'xyzroe',
+        description: 'Zigbee USB switch with monitoring',
+        fromZigbee: [fz.power_on_behavior, fz.on_off, fzLocal.iasZoneAlarm],
```
Koenkk, line 491: Both can be replaced with a modernExtend: `onOff` and `iasZoneAlarm`

src/devices/xyzroe.ts
xyzroe: Do you mean add support in file modernExtend.ts ? Because now as I see `electricityMeter` doesn't support `DC`.
Or can I just enable support in this section somehow?

src/devices/xyzroe.ts
xyzroe: For a more positive user experience, I would like to have a discomfort setting for reponding like this.

src/devices/xyzroe.ts
xyzroe: I started with this. But `onOff` doesn't work properly with `power_on_behavior` when there is more than one endpoint. It makes one expose `power_on_behavior` but without specifying endpoint, while `fz` and `tz` uses _N suffix all the time. So `power_on_behavior` option don't work. 

With `iasZoneAlarm` I will try again, OK. But it seems that I also had some unresolvable problem.

xyzroe:
Also I had some problems with setup reporting using extend `electricityMeter`, so I used `configureReporting: false`.
It setups `min` reportable correct, but for `max` and `change` is uses default values.

src/devices/xyzroe.ts
xyzroe: When I use 
```
    iasZoneAlarm({
      endpointNames: ["1"],
      zoneType: "generic",
      zoneAttributes: ["alarm_1"],
      description: "🔌 Indicates if overcurrent protection is triggered",
    }),
``` 
it doesn't change description. Should I also rework `function iasZoneAlarm(args: IasArgs): ModernExtend {` or I can leave as is?

xyzroe:
> Also I had some problems with setup reporting using extend `electricityMeter`, so I used `configureReporting: false`. It setups `min` reportable correct, but for `max` and `change` is uses default values.

I tried 
```
    electricityMeter({
      cluster: "electrical",
      reporting: electroReporting,
      endpointNames: ["1"],
    }),
```

with `const electroReporting = { min: 10, max: 90, change: 100 };`

but after configuration device on reporting tab I get 
Screenshot 2024-10-12 at 1 58 38 PM

src/devices/xyzroe.ts
Koenkk: Yes add support for DC in the electricityMeter modernExtend, is this something you could take a look at?

xyzroe:
Please look at my implementation to support for DC attributes in the electricityMeter within modernExtend. Additionally, I have added the ability to set up custom reporting settings.

xyzroe:
I hope my approach is acceptable in this project.

src/devices/xyzroe.ts
Koenkk: - Does this device support a power on behaviour per endpoint? If yes, we can add support for this in the modernExtend expose
- For iAsZoneAlarm, a new entry should be added to https://github.com/Koenkk/zigbee-herdsman-converters/blob/b53aef5c1d93f33f68835ea96d2902c5a93e8979/src/lib/modernExtend.ts#L1400

xyzroe: 1. Yes, only the first endpoint. I made a small fix that makes power_on_behaivor work, while multiendpoint (Added suffix, corresponding to the endpoint)

2. Ok, ok i can do it like this. But in my implementation we can pass a unique description. Is it not acceptable?

src/lib/modernExtend.ts
```diff
@@ -466,17 +466,29 @@ export interface OnOffArgs {
...
+        ? exposeEndpoints(e.switch(args.description), args.endpointNames)
+        : exposeEndpoints(e.switch(), args.endpointNames);
 
     const fromZigbee: Fz.Converter[] = [args.skipDuplicateTransaction ? fz.on_off_skip_duplicate_transaction : fz.on_off];
     const toZigbee: Tz.Converter[] = [tz.on_off];
 
     if (args.powerOnBehavior) {
-        exposes.push(e.power_on_behavior(['off', 'on', 'toggle', 'previous']));
+        const powerOnBehaviorExpose = args.description
+            ? e.power_on_behavior(['off', 'on', 'toggle', 'previous']).withDescription(args.description)
```
Koenkk, line 483: Please add a separate description for powerOnBehavior, change `powerOnBehavior?: boolean;` to `powerOnBehavior?: boolean | {description: string};`

src/lib/modernExtend.ts
```diff
@@ -466,17 +466,29 @@ export interface OnOffArgs {
...
+            : e.power_on_behavior(['off', 'on', 'toggle', 'previous']);
+        if (args.endpointNames) {
+            args.endpointNames.forEach((endpoint) => {
```
Koenkk, line 486: I think we can also use `exposeEndpoints` here

src/lib/modernExtend.ts
```diff
@@ -1448,7 +1461,17 @@ export function iasZoneAlarm(args: IasArgs): ModernExtend {
     let alarm2Name = 'alarm_2';
 
     if (args.zoneType === 'generic') {
-        args.zoneAttributes.map((attr) => exposes.push(exposeList[attr]));
+        args.zoneAttributes.map((attr) => {
+            if (args.description) {
```
Koenkk, line 1465: I think this can be changed to:

```
let expose = exposeList[attr];
if (args.description) {
  expose = expose.clone().withDescription(args.description);
}
exposes.push(expose);
```

src/devices/xyzroe.ts
```diff
@@ -463,6 +467,30 @@ const definitions: DefinitionWithExtend[] = [
...
+            electricityMeter({
+                cluster: 'electrical',
+                attributes: 'both',
+                configureReporting: electroReporting,
+                endpointNames: ['1'],
+            }),
+            temperature({
+                reporting: tempReporting,
+            }),
+            onOff({powerOnBehavior: true, endpointNames: ['1'], description: '🔌 Controls the USB port'}),
```
Koenkk, line 487: Please remove the emojis here, none of the other descriptions use it.

src/devices/xyzroe.ts
```diff
@@ -463,6 +467,30 @@ const definitions: DefinitionWithExtend[] = [
...
+            deviceEndpoints({endpoints: {1: 1, 2: 2, 3: 3}}),
+            identify(),
+            electricityMeter({
+                cluster: 'electrical',
+                attributes: 'both',
+                configureReporting: electroReporting,
+                endpointNames: ['1'],
+            }),
+            temperature({
+                reporting: tempReporting,
```
Koenkk, line 485: Why do we need custom reporting values here? (same for `electroReporting`)

src/lib/modernExtend.ts
```diff
@@ -1611,21 +1634,31 @@ export function iasWarning(args?: IasWarningArgs): ModernExtend {
 type MultiplierDivisor = {multiplier?: number; divisor?: number};
 export interface ElectricityMeterArgs {
     cluster?: 'both' | 'metering' | 'electrical';
+    attributes?: 'both' | 'ac' | 'dc';
```
Koenkk, line 1637:
```suggestion
    electricalMeasurementType?: 'both' | 'ac' | 'dc';
```

src/lib/modernExtend.ts
xyzroe: ok

src/lib/modernExtend.ts
xyzroe: ok, try

src/lib/modernExtend.ts
xyzroe: ok. try

src/devices/xyzroe.ts
xyzroe: This refers to the device itself, why can't I use the description that seems correct to me? I reworked the code so that it would use standard functions, ok I agree with that. But the description of the fields in the device is the decision of the device developer.

src/devices/xyzroe.ts
xyzroe: Because as a device developer I think it is more correct to use it with such settings. I perfectly understand and agree with your comments about the general code, but I can’t understand why the device should use the standard one??

src/lib/modernExtend.ts
```diff
@@ -1655,6 +1688,9 @@ export function electricityMeter(args?: ElectricityMeterArgs): ModernExtend {
     if (args.cluster === 'metering' && args.powerFactor) {
         throw new Error(`Power factor is not supported with cluster 'metering', use 'both' or 'electrical'`);
     }
+    if (args.cluster === 'electrical' && args.attributes === 'dc') {
+        throw new Error(`DC attributes are not supported with cluster 'electrical', use 'both' or 'ac'`);
```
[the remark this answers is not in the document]
xyzroe: Yep, my fault. I'll fix.

src/lib/modernExtend.ts
xyzroe: Ok

src/devices/xyzroe.ts
Koenkk: >  but I can’t understand why the device should use the standard one??

Because it's nice to have unification, why should 1 plug report every 1w and others every 10w? We should provide decent defaults, users can always configure it differently from the frontend (in case of a specific usecase)

xyzroe: But devices are different, with different measurement limits. In this case, the device measures the consumption of a USB device at 5V, and there cannot be values greater than 10 W in principle.

Koenkk: That justifies custom reporting values for the `electricityMeter` indeed!

What about the ones for temperature?

src/lib/modernExtend.ts
```diff
@@ -1611,21 +1628,31 @@ export function iasWarning(args?: IasWarningArgs): ModernExtend {
...
+    electricalMeasurementType?: 'both' | 'ac' | 'dc';
     current?: false | MultiplierDivisor;
     power?: false | MultiplierDivisor;
     voltage?: false | MultiplierDivisor;
     energy?: false | MultiplierDivisor;
     producedEnergy?: false | true | MultiplierDivisor;
     acFrequency?: false | true | MultiplierDivisor;
     threePhase?: boolean;
-    configureReporting?: boolean;
+    configureReporting?: false | true | ReportingConfigWithoutAttribute;
```
Koenkk, line 1639: I would keep this `false`/`true` and allow to pass different min/max/change per attribute (current/power/etc. like done in https://github.com/Koenkk/zigbee-herdsman-converters/pull/8123/files)

src/devices/xyzroe.ts
Koenkk: The description is fine, it's just the emojis which make it non standard (and I would like to keep thing similar)

src/devices/xyzroe.ts
xyzroe: I assumed that this parameter is insignificant in this device, since it essentially displays the temperature of the device, and in order to reduce the load on the network, I decided to reduce reporting relative to the standard values.
But if unification is more important, of course I can return the standard ones.

Koenkk: Let's return to the standard ones then

Koenkk:
Made some small improvements, thanks!

xyzroe:
Thanks!

PR #8045, opened 2024-09-28
Updated Sengled defininition to include E13 a21

clockbrain:
Works as an external converter but I haven't been able to test it within the sengled.ts file because of compile issues with zigbee-herdsman-converters in my local environment. Obviously there is something amiss with my setup. Here is the issue I've been having:
[https://github.com/Koenkk/zigbee2mqtt/discussions/21544#discussioncomment-10735454
](https://github.com/Koenkk/zigbee2mqtt/discussions/21544#discussioncomment-10735454
)

Also note that configuration for electrical metering reporting needs to be adjusted manually via the UI due I suspect to this issue:
[https://github.com/Koenkk/zigbee-herdsman-converters/issues/7917](https://github.com/Koenkk/zigbee-herdsman-converters/issues/7917)

Issue #7917, opened 2024-08-30
Report change value of instantaneousDemand


shinasys:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/8049aceee3cc6750779fc4827a8848107a1d0ad9/src/lib/modernExtend.ts#L1611-L1617

Shouldn't the value 5 in instantaneousDemand be 0.005? Both currentSummDelivered and instantaneousDemand belong to the seMetering cluster, so they should use the same unit.

Also, the value of 0.1kWh for currentSummDelivered seems too large. I think a value of 0.01kWh would be more appropriate.
Since standard electrical outlets generally consume less power than power meters, it seems necessary to lower the default value.

I would suggest making the following changes to the code.
```
seMetering: {
            // Report change with every 5W change
            power: {attribute: 'instantaneousDemand', divisor: 'divisor', multiplier: 'multiplier', forced: args.power, change: 0.005},
            // Report change with every 0.01kWh change
            energy: {attribute: 'currentSummDelivered', divisor: 'divisor', multiplier: 'multiplier', forced: args.energy, change: 0.01},
            // produced_energy: {attribute: 'currentSummReceived', divisor: 'divisor', multiplier: 'multiplier', forced: args.energy, change: 0.1},
        },
```

Alternatively, I think it would be more useful if the electricityMeter() could be modified to accept input parameters of report(min,max,change).

Koenkk:
> Shouldn't the value 5 in instantaneousDemand be 0.005?

Yes you are right, since the value is in kWh

> Also, the value of 0.1kWh for currentSummDelivered seems too large. I think a value of 0.01kWh 

That would make the device report with every 10 W change, would that be too often? I think every 100 W is fine. Note that if users for whatever reason want it more frequent, they can easily change it via the configure tab

> Alternatively, I think it would be more useful if the electricityMeter() could be modified to accept input parameters of report(min,max,change).

We want to unify the reporting, so preferably not.

shinasys:
> they can easily change it via the configure tab

That's a different issue, but The change value of uint48 type is not modified in the reporting tab of the configuration.
![image](https://github.com/user-attachments/assets/0129ec0f-8bae-499c-84c2-567f1175996a)

Koenkk:
> That's a different issue, but The change value of uint48 type is not modified in the reporting tab of the configuration.

Will be solved in the next z2m release!

PR #8123, opened 2024-10-13
Allow reporting minimum to be specified in device file for electricityMeter configuration

clockbrain:
The Sengled E13-A21 smart bulb will not report instantaneous demand unless the reporting min is set to 5 seconds or lower. The Sengled hub sets it to 3 seconds but testing shows that 5 seconds is ok. Setting it to 6 seconds or longer stops reporting.

This change allows the min value to be set in the device configuration. This was proposed here https://github.com/Koenkk/zigbee-herdsman-converters/pull/8045#issuecomment-2389510112

Also included is a change to the default change value for instantaneous demand from 5 to 0.005 (kWh) as raised here https://github.com/Koenkk/zigbee-herdsman-converters/issues/7917 

Note that the Sengled bulb has a max power consumption of 12.5W so a reporting change value of 5W is still rather large and it therefore only reports on big dimmer adjustments. However I can live with that and just set that change value via the UI. The PR for the min value however isn't about optimal value, the device just won't report with the current default of 10 seconds.


src/lib/modernExtend.ts
```diff
@@ -1609,12 +1609,13 @@ export function iasWarning(args?: IasWarningArgs): ModernExtend {
 
 // Uses Electrical Measurement and/or Metering, but for simplicity was put here.
 type MultiplierDivisor = {multiplier?: number; divisor?: number};
+type MultiplierDivisorMin = {multiplier?: number; divisor?: number; min?: number};
 export interface ElectricityMeterArgs {
     cluster?: 'both' | 'metering' | 'electrical';
     current?: false | MultiplierDivisor;
-    power?: false | MultiplierDivisor;
+    power?: false | MultiplierDivisorMin;
```
Koenkk, line 1616:
```suggestion
    power?: false | (MultiplierDivisor & ReportingConfigWithoutAttribute);
```

src/lib/modernExtend.ts
```diff
@@ -1836,7 +1837,7 @@ export function electricityMeter(args?: ElectricityMeterArgs): ModernExtend {
                             // Check if this property has a divisor and multiplier, e.g. AC frequency doesn't.
                             if ('divisor' in property) {
                                 // In case multiplier or divisor was provided, use that instead of reading from device.
-                                if (property.forced) {
+                                if (property.forced && property.forced.divisor) {
```
Koenkk, line 1840:
```suggestion
                                if (property.forced?.divisor || property.forced?.multiplier) {
```

src/lib/modernExtend.ts
```diff
@@ -1852,7 +1853,12 @@ export function electricityMeter(args?: ElectricityMeterArgs): ModernExtend {
                                 assertNumber(multiplier, property.multiplier);
                                 change = property.change * (divisor / multiplier);
                             }
-                            items.push({attribute: property.attribute, min: '10_SECONDS', max: 'MAX', change});
+                            let minval: ReportingConfigTime = '10_SECONDS';
+                            // In case min was provided, use that instead of default.
+                            if ('min' in property && 'forced' in property && property.forced && 'min' in property.forced && property.forced.min) {
+                                minval = property.forced.min ?? minval;
+                            }
+                            items.push({attribute: property.attribute, min: minval, max: 'MAX', change});
```
Koenkk, line 1861:
```suggestion
                            items.push({attribute: property.attribute, min: property.forced?.min ?? '10_SECONDS' , max: 'MAX', change});
```

Does this work? (then the rest can be removed)

src/lib/modernExtend.ts
clockbrain: I can't use ReportingConfigWithoutAttribute because its elements of min/max/change aren't optional so would need to allow all three to be specified. I was trying to make this PR the lightest touch possible. This works.
```
    power?: false | (MultiplierDivisor & {min?: ReportingConfigTime});
```

src/lib/modernExtend.ts
clockbrain: Even with the optional property chaining the compiler complains that: 

```
Property 'divisor' does not exist on type 'false | MultiplierDivisor | (MultiplierDivisor & { min?: ReportingConfigTime; })'.
  Property 'divisor' does not exist on type 'false'.ts(2339)
```

Only fix I could find was to include a test for forced:
```
if (property.forced && (property.forced.divisor || property.forced.multiplier)) {
```

src/lib/modernExtend.ts
clockbrain: The compiler also complains about the optional property chaining in the coalesce.
```
Property 'forced' does not exist on type '{ attribute: string; divisor: string; multiplier: string; forced: false | (MultiplierDivisor & { min?: ReportingConfigTime; }); change: number; } | ... 9 more ... | { ...; }'.
  Property 'forced' does not exist on type '{ attribute: string; change: number; }'.ts(2339)
```
 I tried many permutations but couldn't get it to accept anything in that single items.push statement. I even explored adding a placeholder forced property to power_factor being the only one without a forced but that caused other issues. Hence going back to the separate variable approach.

I'll update the PR with another commit bundling these changes.

src/lib/modernExtend.ts
```diff
@@ -1713,9 +1713,9 @@ export function electricityMeter(args?: ElectricityMeterArgs): ModernExtend {
         },
         seMetering: {
             // Report change with every 5W change
-            power: {attribute: 'instantaneousDemand', divisor: 'divisor', multiplier: 'multiplier', forced: args.power, change: 5},
+            power: {attribute: 'instantaneousDemand', divisor: 'divisor', multiplier: 'multiplier', forced: args.power, change: 0.005, min: 'min'},
             // Report change with every 0.1kWh change
-            energy: {attribute: 'currentSummDelivered', divisor: 'divisor', multiplier: 'multiplier', forced: args.energy, change: 0.1},
+            energy: {attribute: 'currentSummDelivered', divisor: 'divisor', multiplier: 'multiplier', forced: args.energy, change: 0.1, min: 'min'},
```
Koenkk, line 1718: Why was the `change` changed here? (also for the line above)

src/lib/modernExtend.ts
```diff
@@ -1852,7 +1852,12 @@ export function electricityMeter(args?: ElectricityMeterArgs): ModernExtend {
                                 assertNumber(multiplier, property.multiplier);
                                 change = property.change * (divisor / multiplier);
                             }
-                            items.push({attribute: property.attribute, min: '10_SECONDS', max: 'MAX', change});
+                            let minval: ReportingConfigTime = '10_SECONDS';
+                            // In case min was provided, use that instead of default.
+                            if ('min' in property && 'forced' in property && property.forced && 'min' in property.forced) {
+                                minval = property.forced?.min ?? minval;
+                            }
+                            items.push({attribute: property.attribute, min: minval, max: 'MAX', change});
```
Koenkk, line 1860:
```suggestion
                            items.push({attribute: property.attribute, min, max: 'MAX', change});
```

src/lib/modernExtend.ts
```diff
@@ -1852,7 +1852,12 @@ export function electricityMeter(args?: ElectricityMeterArgs): ModernExtend {
                                 assertNumber(multiplier, property.multiplier);
                                 change = property.change * (divisor / multiplier);
                             }
-                            items.push({attribute: property.attribute, min: '10_SECONDS', max: 'MAX', change});
+                            let minval: ReportingConfigTime = '10_SECONDS';
```
Koenkk, line 1855:
```suggestion
                            let min: ReportingConfigTime = '10_SECONDS';
```

src/lib/modernExtend.ts
clockbrain: I didn't change it for energy/currentSummDelivered. `change` for energy is still 0.1 as it has always been. I just added the property `min` at the end. 

I did however change the value for power/instantaneousDemand from 5kWh to 0.005kWh as mentioned in this issue https://github.com/Koenkk/zigbee-herdsman-converters/issues/7917#issuecomment-2323316989. With  a `change` value of 5 for power I get a front end reporting Min Rep Change value of 50,000 (because the divisor for my E13-A21 is 10,000).

Koenkk: I see, thanks!

Koenkk:
I updated the code a bit to also allow specifying the `max` and `change`, can you do a final check? Then this can be merged.

clockbrain:
Yes, this all looks good and it tests ok with my E13-A21 bulb. Thanks for your help with it.

Koenkk:
Thanks!

PR #8124, opened 2024-10-13
feat(add) add Heiman HS-720ES carbon monoxide like Moes ZC-HM

dzikus:
This is simple copy and paste from Moes ZC-HM with new manufacturerName and model/vendor.

### DIFF
--- a/src/devices/moes.ts
+++ b/src/devices/moes.ts
@@ -465,7 +465,7 @@ const definitions: DefinitionWithExtend[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
> This is simple copy and paste from Moes ZC-HM


Please add it as a `tuya.whiteLabel` there (to prevent code duplication), example:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/b53aef5c1d93f33f68835ea96d2902c5a93e8979/src/devices/tuya.ts#L936

dzikus:
done

PR #8132, opened 2024-10-14
Added wall switchgear.

liubotongs:
Three new devices are added, which are wall switches of one channel, two channel and three channel. One question I have is how the API: power_on_behavior supports multiple endpoints. I initially copied the TUYA TS0003 process in sonoff.ts, but it can only set the powerOnBehavior of the first Endpoint, and my device supports every Endpoint. Then I implemented sonoffOnOff functionality in sonoff.ts, but the functionality was not complete enough, such as reading the property response when the state was not synchronized to the Web UI display.


src/devices/sonoff.ts
```diff
@@ -545,6 +550,132 @@ const sonoffExtend = {
...
+        args = {powerOnBehavior: true, skipDuplicateTransaction: false, ...args};
+        const exposes: Expose[] = args.endpointNames ? args.endpointNames.map((ep) => e.switch().withEndpoint(ep)) : [e.switch()];
+        const fromZigbee: Fz.Converter[] = [args.skipDuplicateTransaction ? fz.on_off_skip_duplicate_transaction : fz.on_off];
+        const toZigbee: Tz.Converter[] = [tz.on_off];
+
+        if (args.powerOnBehavior) {
+            const behavior_string = ['off', 'on', 'toggle', 'previous'];
+            fromZigbee.push(fz.power_on_behavior);
+            toZigbee.push(tz.power_on_behavior);
+            if (args.endpointNames) {
```
Koenkk, line 669: This can be moved into the `onOff` `modernExtend`, then `sonoffOnOff` can be removed.

liubotongs: How should this be achieved? Since I tried to use onoff modernExtend before, only the first Endpoint is useful, while the others are not. Can you help me modify it?

Koenkk: It has been added in the meantime, now a power on behaviour is exposed for each endpoint (https://github.com/Koenkk/zigbee-herdsman-converters/blob/46c7522dac972e5f71fd01967dc7c24a932ffcf2/src/lib/modernExtend.ts#L536).


You should be able to use it now.

liubotongs: M
However, this does not solve my problem, I want each of the three channels to have different powerBehavior.

PR #8142, opened 2024-10-15
Update candeo.ts

dhc25:
•	Added Candeo LED strip controllers:
o	C-ZB-LC20-RGBCCT
o	C-ZB-LC20-RGBW
o	C-ZB-LC20-RGB
o	C-ZB-LC20-CCT
o	C-ZB-LC20-Dim

•	Fixed incorrect reference for C201 in last update. o	A recently submitted change to candeo.ts has been made to match “Candeo Zigbee Dimmer” to the C201. This has caused all C202 devices (a rotary dimmer) to now be presented as the wrong device (a micro module). o	We think this is wrong.
o	The C202 is a far more popular device, therefore we think it makes more sense to match Candeo Zigbee Dimmer to the most popular option, which is C202.

•	Added Candeo C202.1

Can we please request that @dhc25 and @candeodevelopment are notified if there are future changes to the Candeo devices?

### DIFF
--- a/src/devices/candeo.ts
+++ b/src/devices/candeo.ts
@@ -1,5 +1,5 @@
-import {electricityMeter, light, onOff} from '../lib/modernExtend';
-import {DefinitionWithExtend} from '../lib/types';
+import { electricityMeter, light, identify, onOff } from '../lib/modernExtend';
+import { DefinitionWithExtend } from '../lib/types';
 
 const definitions: DefinitionWithExtend[] = [
     {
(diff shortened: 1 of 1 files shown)


dhc25:
I've tried 5 different approaches to formatting these amends.
Any help would be appreciated..
Thanks

dhc25:
Thank you @Koenkk 
Going forwards, can we please request that @dhc25 and @candeodevelopment are notified (and possibly even asked to approve) if there are future changes to the Candeo devices?

Koenkk:
Thanks!

That would be nice indeed, @candeodevelopment can you consider this?

candeodevelopment:
@Koenkk yes that would be good, please add both @candeodevelopment and @dhc25 to nitpicks for the candeo.ts file!

Koenkk:
@candeodevelopment done!

PR #8144, opened 2024-10-16
update 3r product

3reality-support:


3reality-support:
This is a product that uses batteries, and I don't know if this modification can solve this problem

Koenkk:
Thanks!

PR #8157, opened 2024-10-19
Tuya BAC-003 & BAC-002-ALZB: Add current_cooling_setpoint

robvanoostenrijk:
When the airconditioning control (FCU) is set to mode `cool`, MatterBridge integration using Google Home sends `current_cooling_setpoint`, as opposed to `current_heating_setpoint`.

This change adapts the converter to handle `current_cooling_setpoint` and `current_heating_setpoint` equally, both map to tuya setpoint datapoint `16` on the device.


Koenkk:
Thanks!

PR #8174, opened 2024-10-23
fix: Prevent IKEA PARASOLL and BADRING being stuck on a previously reported state after it rapidly changes back and forth.

marazmarci:
Credit goes to @legsim for [discovering](https://github.com/Koenkk/zigbee2mqtt/issues/22579#issuecomment-2425173192) that configuring reporting for the `ssIasZone` cluster's `zoneStatus` attribute with `minimumReportInterval: 0` fixes a common problem with the PARASOLL and BADRING sensors where their state can become stuck on a previously reported state after it rapidly changes back and forth.

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -931,7 +931,7 @@ const definitions: DefinitionWithExtend[] = [
                 zoneType: 'contact',
                 zoneAttributes: ['alarm_1'],
                 // This is required to prevent the device's reported state being stuck after it quickly changed back and forth:
-                configureZoneStatusReportingEndpoint: 2,
+                zoneStatusReporting: true,
             }),
             identify({isSleepy: true}),
             battery(),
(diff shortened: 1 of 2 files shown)


marazmarci:
@Koenkk, please add the `hacktoberfest-accepted` label to the PR before merging it. :pray:

src/devices/ikea.ts
```diff
@@ -931,6 +932,13 @@ const definitions: DefinitionWithExtend[] = [
             battery(),
             ikeaOta(),
         ],
+        configure: async (device) => {
```
sjorge, line 935: This can probably extracted into a modernExtend quirk and added for these 2 devices.

e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/7e8a807a334c4f21a59ef8770ceee935a9fa4c3a/src/lib/modernExtend.ts#L2350-L2359 is quirk that also adds an addition configure step.

Koenkk: I think it would be better to add this to the already existing `iasZoneAlarm`, @marazmarci is this something you could change?

marazmarci: @Koenkk, I implemented the changes you requested. I'm no TypeScript dev, but I hope it is good and it works.

I didn't yet have time to actually test it, so I'm marking the PR as a draft in the meantime.

Koenkk: Made some small updates, let me know if this work!

marazmarci: I tested it, and it worked both for PARASOLL and BADRING! I saw the `zoneStatus` attribute with the correct endpoint on the `Reporting` tab.

Koenkk:
thanks!

PR #8175, opened 2024-10-23
Add missing `off_wait_time` to `light_onoff_brightness` converter

marazmarci:
I successfully used the [`onWithTimedOff` command](https://github.com/Koenkk/zigbee-herdsman-converters/blob/7e8a807a334c4f21a59ef8770ceee935a9fa4c3a/src/converters/toZigbee.ts#L47) for relays and smart plugs in my automations. For some devices, it's enough only to provide an `on_time` value in the payload, but for some devices (such as the Sonoff ZBMINI, see note [here](https://www.zigbee2mqtt.io/devices/ZBMINI.html#on-with-timed-off)), it's also required to provide an `off_wait_time` value. In https://github.com/Koenkk/zigbee2mqtt.io/pull/2392, a documentation section was added for all devices supporting `on_time` about the "on with timed off" feature, in which it is mentioned that the device may require to also provide an `off_wait_time` value. This documentation section is also shown for lights ([example](https://www.zigbee2mqtt.io/devices/LED2201G8.html#on-with-timed-off)), so as a user, I would expect it to work.

`on_time` works¹ for lights (I tried some IKEA LEDxxxxGx lights), but I noticed that `off_wait_time` produces the following error:
> No converter available for 'off_wait_time' (0)

¹ "works" in the sense that it doesn't produce an error in Z2M. However, for some models, it _actually_ works, but for some, it doesn't, that's why I started experimenting with `off_wait_time`.

In this PR, I added the `off_wait_time` key to the `light_onoff_brightness` converter. As far as I understand the code, this works when a `brightness` value is not provided, and `light_onoff_brightness` delegates to `on_off`, according to this part of the code:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/7e8a807a334c4f21a59ef8770ceee935a9fa4c3a/src/converters/toZigbee.ts#L1152-L1154

Some history: `on_time` was added to the same line of code in https://github.com/Koenkk/zigbee-herdsman-converters/commit/ef35333787a7d12bdee7b0c52502c30a74b32113 by @Koenkk after @sjorge reported https://github.com/Koenkk/zigbee2mqtt/issues/12276


marazmarci:
@Koenkk, please add the `hacktoberfest-accepted` label to the PR. :pray:

Koenkk:
Thanks!

PR #8176, opened 2024-10-23
Added support for TS0601 Thermosphere Thermostat

pink88:
Added support for TS0601 Thermosphere Thermostat, with exception to schedule as it requires additional work to create a converter.


Koenkk:
Thanks!

SteveOswald:
Hi @pink88,

I've just commited a PR for a TVR602Z wich has also the model id "TS0601", but another manufacturer name. I've written a new converter for the schedule, since the old once didn't work either. Maybe those will work for your device too. The converter can be found under "thermostatScheduleDayMultiDP_TRV602Z_WithDayNumber".

Cheers!

pink88:
Thanks Steve. It doesn't quite match with what I've got, but this
definitely helps me out a lot. I will be stealing some of this code 😀

PR #8187, opened 2024-10-25
Add NHMOTION/UNIDIM/1 (ELKO EKO06984, EKO06985 and EKO06986)

FSund:
I've combined the converters for Schneider WDE002961 (push button dimmer) and Schneider 545D6306 (PIR with relay) to come up with something that seems to work.

This supports on/off, dimming, ballast config, occupancy, and light level/illuminance.

Some testing still remains before it's ready for review.

Can't decide between ELKO or Elko, the database seems to have both entries.

### TODO
- [x] Verify that all functionality works after rewrite using only modern extend
- [x] Support dimming mode (seems to only support `0: RL-LED` and `3: Auto`)
- [x] Verify the dimming mode mapping is correct (`0: RL-LED` and `3: Auto` might be reversed)
- [x] Support `on_level`

Optional(?)
- [x] Sensitivity
- [x] Detection delay
- [x] Lux threshold


peroo:
Just a quick note that I think the vendor should probably be Schneider. The ELKO version is just a white-label of WDE002962.

Also, I have a list of commands sniffed from the gateway during pairing, and when executing commands in the app. Might help with getting all the functionality implemented.
```
When pairing:
- Set on level 255. Cluster 8, profile 104, endpoint 3, attribute 11, 8bit uint, on level 255.
- Configure reporting, endpoint 3, cluster 0x0b05, attribute 0x011d, 8bit signed int, 127
- Configure reporting, endpoint 3, cluster 0x0b05, attribute 0x011c, 8bit uint, 255
- Configure reporting, endpoint 3, cluster 0x0006, attribute 0x000, boolean, direction reported 0x00 - attribute 0x4001 16bit uint, on time 1.0s, direction reported 0x00
- Configure reporting, endpoint 3, cluster 0x0008, attribute 0x0000, direction reported 0x00, 8bit uint, current level 1
- Configure reporting, endpoint 3, cluster 0x0406, attribute 0x0000, direction reported 0x00, 8bit bitmap
- Configure reporting, endpoint 37, cluster 0x0400, attribute 0x0000, direction reported 0x00, 16bit uint, measured value 441
- Configure reporting, endpoint 37, cluster 0xff19, attribute 0x0000, direction reported 0x00, 16bit uint, 1
- Bind:
    - endpoint 3, cluster 0x0b05
    - endpoint 3, cluster 0x0006
    - endpoint 3, cluster 0x0008
    - endpoint 37, cluster 0x0406
    - endpoint 37, cluster 0x0400
    - endpoint 37, cluster 0xff19

App Commands:
- Dimming mode: endpoint 3, cluster 301, profile 104, attribute 0xe000
    - RL-LED: enum8 0
    - Auto: enum8 3
- Dimming range: endpoint 3, cluster 301:
    - Min: attribute 0x0010, 0-254
    - Max: attribute 0x0011, 0-254
- Toggle mode: endpoint 3, cluster 8:
    - 100%: attribute 0x0011, 8bit uint, on level 254
    - last level: attribute 0x0011, 8bit uint, on level 255
- Sensitivity settings: endpoint 37, cluster 0x0406
    - Medium: attribute 0xe003, uint8 75
    - Low: attribute 0xe003, uint8 50
    - High: attribute 0xe003, uint8 100
- Detection delay: endpoint 37, cluster 406
    - 15s: attribute 0x0010, 16bit uint, 15
    - 300s: attribute 0x0010, 16bit uint, 300
    - 13618: attribute 0x0010, 16bit uint, 300
- Lux threshold, endpoint 37, cluster 0xff19
    - Current lux (21lx): attribute 0x0000, 16bit uint, 13618 (0x3532)
    - 1000lx: attribute 0x0000, 16bit uint, 30001(0x7531)
```

src/devices/elko.ts
```diff
@@ -197,6 +197,48 @@ const definitions: DefinitionWithExtend[] = [
             device.save();
         },
     },
+    {
+        zigbeeModel: ['NHMOTION/UNIDIM/1'],
+        model: 'EKO06984',
+        vendor: 'ELKO',
+        description: 'SmartPir with push dimmer',
+        fromZigbee: [fz.on_off, fz.brightness, fz.illuminance, fz.level_config, fz.schneider_lighting_ballast_configuration, fz.occupancy],
```
Koenkk, line 205: Can you try with modern extend? E.g. `extend: [light({configureReporting: true})],`

FSund:
> Also, I have a list of commands sniffed from the gateway during pairing, and when executing commands in the app. Might help with getting all the functionality implemented.

From your list and some testing it seems like some custom tz an fz might be required to support all features.

For example, setting `dimmer_mode`

```
z2m: Publish 'set' 'dimmer_mode' to '0x70ac08fffe4959e0' failed: 
'Error: ZCL command 0x70ac08fffe4959e0/3 lightingBallastCfg.write(
{"57344":{"value":1,"type":48}}, 
{"timeout":10000,
  "disableResponse":false,
  "disableRecovery":false,
  "disableDefaultResponse":true,
  "direction":0,
  "reservedBits":0,
  "manufacturerCode":4190,
  "writeUndiv":false
}) 
failed (Status 'INVALID_VALUE')'
```

peroo:
Apologies, on closer inspection I've realized my switch is a slightly different version from yours. I have the `NHMOTION/DIMMER/1` - `EKO07252`, which is the multiwire version. That means that `WDE002962` is wrong for this device. I expect the commands to be the same, the difference from my understanding is just electrical, but it's hard to say. 

For what it's worth I'm seeing the same errors when setting dimmer_mode 1 or 2. In the past I've used the `tz.wiser_dimmer_mode` converter which is another version of the schneider dimmer mode, but confusingly auto/led values are opposite from what I've sniffed, which either indicates a difference or my notes are wrong. I can't think of a way to really test.

FSund:
> Apologies, on closer inspection I've realized my switch is a slightly different version from yours. I have the `NHMOTION/DIMMER/1` - `EKO07252`, which is the multiwire version. That means that `WDE002962` is wrong for this device. I expect the commands to be the same, the difference from my understanding is just electrical, but it's hard to say.
> 
> For what it's worth I'm seeing the same errors when setting dimmer_mode 1 or 2. In the past I've used the `tz.wiser_dimmer_mode` converter which is another version of the schneider dimmer mode, but confusingly auto/led values are opposite from what I've sniffed, which either indicates a difference or my notes are wrong. I can't think of a way to really test.

I see. I think EKO07252 is the replacement for EKO06984/85/86, as those are not sold anymore. 

I'll change the vendor back to Elko then I think, since I haven't been able to find any alternative producers for this older dimmer.

FSund:
Some updates:

- The device reports manufacturerName "Schneider Electric" during interview, so I've left it as that for now. But I'm not certain if this name comes from the device itself, or from my converter?

- I'm having some issues with setting up levelConfig, or specifically `on_level`. The device has the `genLevelCtrl` as an input cluster on endpoint 3, and a output cluster on endpoint 37. I'm not sure how to implement this in practice.


Here's the endpoints and clusters

```
endpoints:
  - id: 3
    output_clusters:
      - genOta
    input_clusters:
      - genBasic
      - genIdentify
      - genGroups
      - genScenes
      - genOnOff
      - genLevelCtrl
      - lightingBallastCfg
      - haDiagnostic
  - id: 37
    output_clusters:
      - genIdentify
      - genOnOff
      - genLevelCtrl
      - msOccupancySensing
    input_clusters:
      - genBasic
      - genIdentify
      - msIlluminanceMeasurement
      - msOccupancySensing
      - haDiagnostic
      - elkoOccupancySettingClusterServer
  - id: 242
    output_clusters:
      - greenPower
```

peroo:
> * The device reports manufacturerName "Schneider Electric" during interview, so I've left it as that for now. But I'm not certain if this name comes from the device itself, or from my converter?

That'll be from the device. You can query it in the dev console by reading `manufacturerName` from the `genBasic` cluster on endpoint 3.

Some updates on my end.
* I've verified that my notes had dimming mode backwards, so the existing `wiserDimmerControlMode` is correct.
* I've managed to cobble together occupancy sensitivity control, the only thing to note is that since it's manufacturer specific the `manufacturerCode` needs to be set to Schenider when configuring reporting and setting. `zigbee-herdsman` has cluster definitions for all these manufacturer specific clusters and attributes, but they were added as Elko specific, and have the wrong manufacturer code, which makes them unusable. I don't think any of these have ever been used anywhere, they were probably just added from documentation, so they should be renamed and updated to apply to Schneider, IMO.
* I spent hours debugging why the PIR lights green when attached to z2m. Turns out it lights green when the network is open for adoption. 🤦

I'm planning to submit a PR for my device since it's slightly different. We should be able to share most of the implementation.

BTW, there's some Schneider documentation floating around for dimmer switches. It doesn't include PIR functionality, but it's hugely useful for getting accurate documentation of dimmer functionality. 
[Connected_Universal_Push-button_Dimmer_04-2022_368225_1653575314693_en-US.pdf](https://github.com/user-attachments/files/17539138/Connected_Universal_Push-button_Dimmer_04-2022_368225_1653575314693_en-US.pdf)

FSund:
Anyone know of a way to avoid the "_<endpoint name>" postfix for the light entity?

With the current setup

https://github.com/Koenkk/zigbee-herdsman-converters/blob/014fb0ca1992c94e93498fe2b45624ccfea83488/src/devices/schneider_electric.ts#L2041-L2049

The light entity in HA is named `light.<device name>_light`, which is a bit clumsy. I tried endpoint name `default`, but then it got the postfix `_default`.

enoch85:
@FSund How do I your PR into my Prod ENV?

enoch85:
@FSund Could you please add support for the one I posted as well, or do you suggest a separate PR?

FSund:
@enoch85 I have not gotten on_level to work yet, but I suppose that's not required, and can be added later.

Does all functionality work with your Exxact device? Ballast config, illuminance/lux detection etc? In that case I can add the Swedish device to this PR.

enoch85:
@FSund 

Afaik, everything works. The only thing I need really is turning it off and on, and detect presence. Just tested ballast though, and it's also working! 👍🏼 Regarding the on_level I can see that it's on by a change in color on the switch. Please see pic for more info.

Screenshot_2024-11-10-09-19-44-53

enoch85:
Maybe would be nice with an icon of the device though, don't know if this PR covers that?

src/devices/schneider_electric.ts
```diff
@@ -2030,6 +2032,53 @@ const definitions: DefinitionWithExtend[] = [
             schneiderElectricExtend.visaKeyEventNotification('4'),
         ],
     },
+    {
+        zigbeeModel: ['NHMOTION/UNIDIM/1'],
```
enoch85, line 2036: Please notice that this is different on the Swedish model.

peroo: You have the newer multiwire version (`NHMOTION/DIMMER/1`). It's a separate device that has basically identical functionality except for improved dimming range. I'm working on a PR for it.

enoch85: Aah ok, please notify me then. Would be super if we could co-op on the implementation.

enoch85: This I'm missing from this PR though:

1. `dimmer_mode`: Auto, RL-LED
2. `on_level` (I would say "Change On Mode": 100%, Last level
3. Sensitivity: High, Medium,  Low
4. Detection delay: X seconds or minutes

FSund: @enoch85 `dimmer_mode` works for my Elko dimmer. `on_level`, sensitivity and detection delay I have not figured out how to support.

FSund:
I've added `levelConfig` to the light extend, but can't get it to respond via the GUI.

But if I set `on_level` via Dev console it works fine

image

image

I've tried setting it via the Exposes tab and reading via Dev console, but it doesn't seem to change. 

Any ideas what I'm doing wrong?

src/devices/schneider_electric.ts
enoch85: > @enoch85 `dimmer_mode` works for my Elko dimmer. `on_level`, sensitivity and detection delay I have not figured out how to support.

Yes, it's working, but I'm missing 'Auto', and I don't see RC in my Wiser app.

peroo:
I've pushed my PR which adds extends for ballast and the Schneider specific stuff. Everything should work, though I haven't 100% verified that yet.

https://github.com/Koenkk/zigbee-herdsman-converters/pull/8300

FSund:
I've finally gotten around to clean up this. All functionality is confirmed working on my devices.

Thanks a lot to @peroo for doing the heavy lifting here in #8300.

enoch85:
@peroo Your integration works perfect! I'm soo grateful, thank you!

src/devices/schneider_electric.ts
```diff
@@ -2128,6 +2128,36 @@ const definitions: DefinitionWithExtend[] = [
             {vendor: 'Exxact', model: 'WDE003962'},
         ],
     },
+    {
+        zigbeeModel: ['NHMOTION/UNIDIM/1'],
+        model: 'NHMOTION/UNIDIM/1',
```
Koenkk, line 2133: I would expect a different model number is written on product page or device, can you use this?

FSund: To be honest I'm not certain that this is actually a Schneider Electric white label product. It looks mostly identical to Elko EKO07250 and related devices, which is definitely produced by Schneider, but this device (EKO06984) seem like an earlier version.

In any case I haven't been able to find the Schneider product number anywhere.

I can change to `model: 'EKO06984'` and `vendor: 'Elko'`, but in that case I should perhaps move the definition over to `elko.ts`. In that case I would also have to duplicate a lot of the `schneiderElectricExtend` functionality.

FSund: The `zigbeeModel` is different, `NHMOTION/UNIDIM/1` vs. `NHMOTION/DIMMER/1`.

Koenkk: Then let's keep  `NHMOTION/UNIDIM/1`

Koenkk:

Make sure that:
- The filename is `NHMOTION-UNIDIM-1.png`
- The size is 512x512

FSund:
> Could you also submit a picture for the docs? This can be done by:
> 
> 1. Create a fork by clicking [here](https://github.com/Koenkk/zigbee2mqtt.io/fork)
> 2. Go to the `public/images/devices` directory, _Add file_ -> _Upload files_
> 3. Upload the files and press _Commit changes_
> 4. Press _Contribute_ -> _Open pull request_ -> update title/description -> _Create pull request_
> 
> Make sure that:
> 
> * The filename is `NHMOTION-UNIDIM-1.png`
> * The size is 512x512
> * The background is transparent (use e.g. [Adobe remove background](https://new.express.adobe.com/tools/remove-background))

Docs and images was already added https://github.com/Koenkk/zigbee2mqtt.io/pull/3285

But I see I perhaps used the wrong filenames. Should I rename the image and/or the markdown file?

Koenkk:
yes please rename

PR #8188, opened 2024-10-25
feat: speed-controlled fans

lorenz:
This implements converters for speed-controlled fans. These fans are
implemented as level-controlled devices on the ZigBee side (having a
onOff and levelCtrl cluster), thus no speed translation is necessary and
their state can be handled by the normal on_off converters.

Also adds my custom FanBee device as we need one device to run tests
against.

Co-dependent with https://github.com/Koenkk/zigbee2mqtt/pull/24483


src/devices/custom_devices_diy.ts
```diff
@@ -1175,6 +1175,17 @@ const definitions: DefinitionWithExtend[] = [
             }),
         ],
     },
+    {
+        zigbeeModel: ['FanBee1', 'Fanbox2'],
+        model: 'FanBee',
+        vendor: 'Lorenz Brun',
+        description: 'Fan with valve',
```
Koenkk, line 1182: Can you provide a link to this device? Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt/pull/24483

lorenz: This is a custom device built by me. I'll eventually open-source the design files and firmware, but right now there is no public docs or anything for it.

Koenkk: For those type of devices, you can use the external converters (there is no point of adding it to the list of officially supported devices if nobody can get it)

lorenz: I know, but for the tests we need a device which exercises this behavior. There is an Inovelli fan which looks like it could also use this fan control model but as I do not own one of them I didn't feel comfortable changing it over to the new model (also this is a breaking change for anyone using the MQTT API).

I do intend on making this device publicly available.

Koenkk: > I do intend on making this device publicly available.

Good, then I will merge it once this is done.

lorenz: https://github.com/lorenz/fanbee

src/converters/toZigbee.ts
```diff
@@ -1554,6 +1554,22 @@ const converters2 = {
...
+        key: ['fan_speed'],
+        convertSet: async (entity, key, value, meta) => {
+            utils.assertNumber(value);
+            await entity.command(
+                'genLevelCtrl',
+                'moveToLevelWithOnOff',
+                {level: value.toString(), transtime: 0},
+                utils.getOptions(meta.mapped, entity),
+            );
+            return {state: {fan_speed: value}, readAfterWriteTime: 0};
```
Koenkk, line 1567:
```suggestion
            return {state: {fan_speed: value}};
```

src/lib/exposes.ts
```diff
@@ -560,14 +560,21 @@ export class Fan extends Base {
     constructor() {
         super();
         this.type = 'fan';
-        this.addFeature(new Binary('state', access.ALL, 'ON', 'OFF').withDescription('On/off state of this fan').withProperty('fan_state'));
     }
 
     withModes(modes: string[], access = a.ALL) {
+        this.addFeature(new Binary('state', access, 'ON', 'OFF').withDescription('On/off state of this fan').withProperty('fan_state'));
```
Koenkk, line 566: This should not be here, add a `withState()` and update all existing `e.fan()` to call `.withState()`

Koenkk: Although I agree `fan_state` is not the nicest name, I think we should keep it consistent at least. Is there a strong reason to have `state` instead of `fan_state`?

lorenz: As far as I understand the codebase the on_off from/toZigbee converters want the property to be called `state`, right?

Koenkk: Yes, but we could add another `key` to the list (`fan_state`) to also support that.

lorenz: IMO the solution with `state` is a cleaner both from a code perspective (no vaguely-related keys in generic converters) and from an MQTT API perspective as any system which knows how to power on/off a device can now do that to a fan too without needing specific knowledge, similar to how this is designed with the the Zigbee clusters. I would prefer to convert the remaining `fan_state`  fields to `state` as well for similar reasons but that's a breaking API change so probably needs to wait until the next major release.

Koenkk: >  I would prefer to convert the remaining fan_state fields to state as well for similar reasons but that's a breaking API change so probably needs to wait until the next major release.

That cannot be done I think, there are devices which have both a `state` (for e.g. a light) and `fan_state` (for the fan). I think the cleanest solution is to have a `withState()` where the property name is specified

lorenz: But shouldn't that be two different endpoints / entities?

Koenkk: I don't remember/can find the device anymore, but that could be indeed, they are definitely different entities.

lorenz: So I had another look at this, and I see now that Z2M merges all entities together on the MQTT side. So regardless of endpoints everything ends up in one big object. I also checked the sources and all my other devices and pretty much everything uses `state` as its primary on/off/mode attribute. Lights, curtains, relay modules, ...

The first fan to be added to Z2M happened to be a Hampton Bay model which has both a light as well as a fan in a single device and endpoint which was the reason this got added as `fan_state`. Even on the [Deconz issue](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/932) this caused problems and the general consensus was that the manufacturer should've used two endpoints for this. For speed-controlled fans this must be on a separate endpoint as the on/off and level control clusters are already used for the fan. With multiEndpoint in Z2M we can then make the properties unique.

So I have a strong preference for using `state` at least for speed-controlled fans which do not have an API in Z2M yet. This is not a breaking change and stays consistent with the rest of Z2M at the cost of being different from mode-controlled fans. Mode-controlled fans cannot share any external control loops with speed-controlled fans and similar devices anyways so this is not that important.

If you insist I can change this back to `fan_state` but I really don't think we should.

Koenkk: I'm fine with naming it `state`, for consistency also rename `fan_speed` -> `speed`. I still think 

```js
this.addFeature(new Binary('state', access, 'ON', 'OFF').withDescription('On/off state of this fan').withProperty('state'));
```

should be moved out of `withSpeed()` (there should be a seperate `withState` where you can indicate the state property (`fan_state`/`state`)

lorenz: Renamed from `fan_speed` -> `speed` and moved the state property into its own helper.

Koenkk:
Looks good, I think https://github.com/Koenkk/zigbee2mqtt/pull/24483/files also needs to be updated?

lorenz:
Already did that, probably just pushed the same repo twice. Will fix later today.

Koenkk:
Thanks!

Make sure that:
- The filename is `FanBee.png`
- The size is 512x512

PR #8189, opened 2024-10-25
Smoke and heat alarm fixes for table full error (SMSZB-120 and HESZB-120)

mamrai1:
Bind only 4 clusters (genpollcntrl, ssIasZone, ssIasWd and powercfg)  to avoid 'Table full' error


Koenkk:
Thanks!

PR #8190, opened 2024-10-26
Define endpoint in definition.toZigbee containing key 'state'

Zitrium31:
Purpose of this change is described in [issue #24352](https://github.com/Koenkk/zigbee2mqtt/issues/24352) (repo: zigbee2mqtt)

This fix allows the selection of the proper converter among others using the same key `state` with multi endpoints device when sending a command _state_ at [zigbee2mqtt\lib\extension\publish.ts](https://github.com/Koenkk/zigbee2mqtt/blob/e06848db8c5ad247c27e959eb85465a30134769f/lib/extension/publish.ts#L262) at  line 262 `const converter = converters.find((c) => (!c.key || c.key.includes(key)) && (!c.endpoint || c.endpoint == endpointName));`

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -486,7 +486,7 @@ export function onOff(args?: OnOffArgs): ModernExtend {
         : exposeEndpoints(e.switch(), args.endpointNames);
 
     const fromZigbee: Fz.Converter[] = [args.skipDuplicateTransaction ? fz.on_off_skip_duplicate_transaction : fz.on_off];
-    const toZigbee: Tz.Converter[] = [tz.on_off];
+    const toZigbee: Tz.Converter[] = [{...tz.on_off, endpoint: args?.endpointNames}];
 
     if (args.powerOnBehavior) {
         exposes.push(...exposeEndpoints(e.power_on_behavior(['off', 'on', 'toggle', 'previous']), args.endpointNames));
@@ -961,7 +961,7 @@ export function light(args?: LightArgs): ModernExtend {
 
     const fromZigbee: Fz.Converter[] = [fz.on_off, fz.brightness, fz.ignore_basic_report, fz.level_config];
     const toZigbee: Tz.Converter[] = [
-        tz.light_onoff_brightness,
+        {...tz.light_onoff_brightness, endpoint: args?.endpointNames},
         tz.ignore_transition,
         tz.level_config,
         tz.ignore_rate,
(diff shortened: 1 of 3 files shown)


src/lib/modernExtend.ts
```diff
@@ -1243,12 +1243,19 @@ export function commandsColorCtrl(args?: CommandsColorCtrl): ModernExtend {
 
 export interface LockArgs {
     pinCodeCount: number;
+    endpointNames?: string[];
```
Koenkk, line 1246: I don't understand how this works, this is an array and `endpointName` is a `string` in `publish.ts`. Did you check if this works?

Zitrium31: Thanks much for your review 👍 

Actually, I tested it succesfully with the diy device describe in [issue #24352](https://github.com/Koenkk/zigbee2mqtt/issues/24352), that is to say made of the following 3 endpoints:
- Endpoint 8 : genBasic, genIdentify, msTemperatureMeasurement.
- Endpoint 10 : closuresWindowCovering ( `ep1` ) --> It uses key `state`
- Endpoint 12 : genOnOff ( `relay` ) --> It uses also key `state`

Your remark is more than relevant : My bad ! It works only if `endpointNames: string[]` contains only 1 item for sure :confused:

I believe `endpointNames` must definitely remain an array to allow multiple endpoints with the same cluster. Then I am gonna see how I can handle that in `publish.ts`. Please advise if you see a better way.

Koenkk: We should update `publish.ts` to check if the endpoint in the array is endpoint is a list.

Zitrium31: :ok_hand: I will come back to you with a PR to update `publish.ts`

Zitrium31: [PR #24525](https://github.com/Koenkk/zigbee2mqtt/pull/24525) created.

src/lib/types.ts
```diff
@@ -331,7 +331,7 @@ export namespace Tz {
     export interface Converter {
         key?: string[];
         options?: Option[] | ((definition: Definition) => Option[]);
-        endpoint?: string;
+        endpoint?: string | string[];
```
Koenkk, line 334:
```suggestion
        endpoints?: string[];
```

I'm thinking now, maybe always make this a `string[]`? Should simplify the z2m code: `(!c.key || c.key.includes(key)) && (!c.endpoints || c.endpoints.includes(endpointName)));`

Zitrium31: I fully agree with this idea to simplify `endpoints` as `string[]` only. 
Use of an array definitely makes sense as one or more `endpointName` can be associated to a given cluster. 
If I remember well, the actual `endpoint: string` is already used elsewhere : Before go further, I need to evaluate the impact.

Koenkk: I don't expect it is used in many places, changing the type here and doing a `npm run build` should show you the issues.

Zitrium31: Thanks for the advise. I'll do that. :thumbsup:

Zitrium31: Finally, your expectation is verified and I found only `endpoint` used in `tuya.ts`.

Koenkk:
Thanks!

PR #8196, opened 2024-10-27
Add Support for Chacon roller shutter module ZB-ERSM-01

Vecken:
Add Support for Chacon roller shutter module ZB-ERSM-01, using modern Extend
New manufacturer


Koenkk:
Thanks!

PR #8199, opened 2024-10-27
Update innr.ts RB 279 T As it does not support color

ed-wright:
This bulb does not support color only color temp


ed-wright:
This is to fix an issue i raised in Zigbee2MQTT repo https://github.com/Koenkk/zigbee2mqtt/issues/24513

Koenkk:
Thanks!

PR #8209, opened 2024-10-28
Bind genBinaryInput cluster instead of genpolllctrl 

mamrai1:
Develco smoke and heat detector ‘genBinaryInput’Cluster contains reliability and status flags.


Koenkk:
Thanks!

PR #8210, opened 2024-10-28
Add two private attributes for SR-ZG9040A/ZG9041A-D.

maginawin:
1. External Switch Type
2. Minimum PWM

### DIFF
--- a/src/devices/sunricher.ts
+++ b/src/devices/sunricher.ts
@@ -25,12 +25,149 @@ import {
 } from '../lib/modernExtend';
 import * as reporting from '../lib/reporting';
 import * as globalStore from '../lib/store';
-import {DefinitionWithExtend, Fz, Zh} from '../lib/types';
+import {DefinitionWithExtend, Expose, Fz, ModernExtend, OnEvent, OnEventData, OnEventType, Tz, Zh} from '../lib/types';
 import * as utils from '../lib/utils';
 
 const NS = 'zhc:sunricher';
 const e = exposes.presets;
 const ea = exposes.access;
+
+const sunricherManufacturerCode = 0x1224;
+
+async function readAttributeIfOnline(type: OnEventType, data: OnEventData, device: Zh.Device, attribute: number) {
+    let isReading = false;
+    if (type === 'deviceInterview') {
+        const interviewData = data as {status: string};
+        if (interviewData.status === 'successful') {
+            isReading = true;
+        }
+    } else if (type === 'deviceAnnounce') {
+        isReading = true;
+    }
+    if (isReading) {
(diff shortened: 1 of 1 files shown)


src/devices/sunricher.ts
```diff
@@ -25,13 +25,150 @@ import {
...
+            .withLabel('Minimum PWM')
+            .withDescription('Power off the device and wait for 3 seconds before reconnecting to apply the settings.')
+            .withValueMin(0)
+            .withValueMax(50)
+            .withUnit('%')
+            .withValueStep(1),
+    ];
+
+    const onEvent: OnEvent = async (type, data, device) => {
+        await readAttributeIfOnline(type, data, device, attribute);
```
Koenkk, line 160: I think these only  need to be read one time after the device has joined? If yes, put them in a `configure`.

src/devices/sunricher.ts
```diff
@@ -25,13 +25,150 @@ import {
...
+                    };
+                }
+                return undefined;
+            },
+        },
+    ];
+
+    const toZigbee: Tz.Converter[] = [
+        {
+            key: ['minimumPWM'],
```
Koenkk, line 134:
```suggestion
            key: ['minimum_pwm'],
```

src/devices/sunricher.ts
```diff
@@ -25,13 +25,150 @@ import {
...
+                await entity.write('genBasic', {[attribute]: {value: numericValue, type: dataType}}, {manufacturerCode: sunricherManufacturerCode});
+                return {state: {externalSwitchType: value}};
+            },
+            convertGet: async (entity, key, meta) => {
+                await entity.read('genBasic', [attribute], {manufacturerCode: sunricherManufacturerCode});
+            },
+        } satisfies Tz.Converter,
+    ];
+
+    const exposes: Expose[] = [e.enum('externalSwitchType', ea.ALL, ['Push button', 'Normal on/off', 'Three way']).withLabel('External Switch Type')];
```
Koenkk, line 98:
```suggestion
    const exposes: Expose[] = [e.enum('external_switch_type', ea.ALL, ['push_button', 'normal_on_off', 'three_way']).withLabel('External Switch Type')];
```

maginawin:
Hi @Koenkk , thank you for your detailed suggestions, I've modified them.

Gnarfoz:
@maginawin Hi, this change seems aimed at SR-ZG9040A/ZG9041A-D but it also affects https://www.zigbee2mqtt.io/devices/ZG9101SAC-HP-Switch.html.
The external switch attribute shows up, but I don't think that's configurable on these devices?
The PWM thing doesn't show up. The switch is not a dimmer, anyway.

PR #8212, opened 2024-10-29
Update AC01353010G minimum voltage

aioue:
Multiple of these devices consistently report 2200mV before ceasing to report.

Update the minimum voltage so that zigbee2mqtt reports a lower minimum battery % value, rather than ~27%, before the devices runs out of power and stops reporting.


Koenkk:
Thanks!

PR #8214, opened 2024-10-29
Correct local temperature calibration min and max value on SONOFF TRVZB

m-larin:
Experimentally found out the supported calibration range for SNOFF TRVZB. When sending a message to MQTT with subject zigbee2mqtt/0x0cae5ffffebc15dc/set/local_temperature_calibration and transmitting deliberately large values, I received an error explicitly specifying the acceptable range: The value of "value" is out of range. It must be >= -128 and <= 127. Received 160. Send this max 12.7 and minv -12.8 values work correct, error not raise.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -916,7 +916,7 @@ const definitions: DefinitionWithExtend[] = [
                 .climate()
                 .withSetpoint('occupied_heating_setpoint', 4, 35, 0.5)
                 .withLocalTemperature()
-                .withLocalTemperatureCalibration(-7.0, 7.0, 0.2)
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

m-larin:
> Hi. Sorry to write in a closed (and merged) PR. I've opened [zigpy/zha#300](https://github.com/zigpy/zha/pull/300) to make the exact change in ZHA, and someone suggested changing the step to 0.1. Do you have any thoughts on this?

Hi. The step affects how much the temperature calibration will change when pressing the up or down buttons in the web interface. When sending a number, you can use any value with an accuracy of one tenth.

fgsch:
@m-larin oh right, that's a very good point.

PR #8226, opened 2024-10-30
Update linptech.ts ES1ZZ

ryangu:
add led indicator and fix the illuminance calculation


Koenkk:
Thanks!

bisquit2003:
image
hello, doesn't work(

PR #8230, opened 2024-10-31
Avatto trv06

SANSTONE:
We have developed a standard zigbee intelligent temperature control valve，This submission is because some customers want to use this equipment through this project. And Thanks!


Koenkk:
Looks the same as `TS0601_thermostat_3`, could you add it there?

SANSTONE:
> Looks the same as `TS0601_thermostat_3`, could you add it there?
Agreed. Thanks for the suggestion

PR #8234, opened 2024-10-31
Add NAS-TH07B2

baggiowu:
Add NAS-TH07B2
[database.db.txt](https://github.com/user-attachments/files/17587329/database.db.txt)


Koenkk:
Thanks!

PR #8236, opened 2024-10-31
add support for _TZE204_m64smti7 TNCE DIN single phase energy meter

octa22:
This PR adds support for the TNCE DIN single phase energy meter
picture goes here https://github.com/octa22/zigbee2mqtt.io/pull/1

thanks


Koenkk:
Thanks!

PR #8238, opened 2024-10-31
Add hue centris 3 spots white v2

BogdanDarius:
Add new device.

https://www.philips-hue.com/de-de/p/hue-white---color-ambiance-hue-centris-cross-3er-spot-deckenleuchte/8720169318731


Koenkk:
Thanks!

PR #8241, opened 2024-11-01
device/zigbeetlc: Removed no longer needed endpoint quirk

robvanoostenrijk:
Removal of quirkAddEndpointCluster, the latest ZigbeeTlc firmware version no longer requires this.

Users might have to force remove and re-configure their device to resolve any Zigbee TABLE_FULL errors.


Koenkk:
Thanks!

PR #8246, opened 2024-11-01
Add support for Sedea eTH730 sensor

TomDev-Leg:
Hello, 
This PR is linked with https://github.com/Koenkk/zigbee2mqtt.io/pull/3175 .

I can say that i was able to detect and pair my sensor but battery/voltage was not exposed (empty values).
After some search I saw that my device was map to the wrong definition : 
Tuya TS0201 which use a fzlocal TS0201_battery, and this local definition was removing battery information because of this https://github.com/Koenkk/zigbee2mqtt/issues/11470

So i create instead this converter (which is not so different from other) to have battery/voltage values back.
If later i detect an issue, i will contribute to fix stuff of course.
Thank you


Koenkk:
Thanks!

PR #8247, opened 2024-11-01
fix temperature reading and description for _TZE204_m64smti7

octa22:
This PR fixes the temperature reading and description for _TZE204_m64smti7 DIN energy meter


Koenkk:
Thanks!

PR #8248, opened 2024-11-01
Switch E2206 to Zigbee OTA

jamesonuk:
This is OTA from newer IKEA URL.
OTA was added in https://github.com/Koenkk/zigbee-OTA/commit/35afc409fee009079ddf1f67a1163130084eb1cc Switch from IKEA OTA to zigbee OTA


Koenkk:
Thanks!

PR #8251, opened 2024-11-02
Remove IKEA OTA

jamesonuk:
As per 
https://github.com/Koenkk/zigbee2mqtt/discussions/24464
and 
https://github.com/Koenkk/zigbee-OTA/pull/581

IKEA firmware updates are going to be in the OTA repository rather than from IKEA URL so PR is to remove IKEA OTA functionality and replace it with standard OTA

Would negate the need for https://github.com/Koenkk/zigbee-herdsman-converters/pull/8248


Koenkk:
Thanks for the PR, but @Nerivec is already working on a bigger refactor of this. So I'm closing this in favour of that.

PR #8269, opened 2024-11-04
fix: changed device vendor name from Namron AS to Namron

SanderBlom:
Updated the vendor name of the newly added thermostat (#8119) to align with existing devices from the same manufacturer, which are all listed under the vendor name `Namron` rather than `Namron AS`.


Koenkk:
Thanks!

PR #8272, opened 2024-11-04
Restore divideBy10FromOnly for Tuya ZY-M100-24G

magicus:
This is a partial revert of https://github.com/Koenkk/zigbee-herdsman-converters/pull/8229. Apparently, the sensitivity controls for occupancy and move is doing some weird things; reading and publishing values in different magnitute.

This makes the value jump up and down when changing it (going between 0.6 and 6, or 6 and 60) but it seems to stabilize after a while. This is not ideal, but better than the fix in https://github.com/Koenkk/zigbee-herdsman-converters/pull/8229 which results in an endless stream of log messages.

I still hope it is possible to find an even better solution, but in the meantime, this fixes the regression.


Koenkk:
Thanks!

PR #8278, opened 2024-11-05
Add support for battery information to Legrand 067646

CodeForLove83:
Add support to retrieve battery percentage information to Legrand 067646


Koenkk:
Thanks!

PR #8280, opened 2024-11-05
feat: Add running state to Saswell TRV climate

barthez:
I happen to own this TRV (white-labeled as RTX). It was lacking the running state interface of climate (in HA) although it provides `heating` binary sensor. I kept the heating state attribute for backward compatibility.


Koenkk:
Thanks

PR #8283, opened 2024-11-07
Update shinasystem.ts

shinasys:
Due to the need for further testing, the OTA feature has been deactivated.


Koenkk:
It would be better to remove the OTA from the zigbee-OTA, this will only land in the next z2m release (1 December) (I assumed there is an issue with the OTA)

shinasys:
Thanks. I'll follow up on this later.

PR #8292, opened 2024-11-09
Add support for Philips A60 bulb (Model 9290002269A)

ICalin:
The zigbeeModel  was taken from another device from the repo, did not work and just increased by 1. I don't have the box of the bulb. I found this buld at uni in a box and I want to use it for my thesis. I only have the model which seems to be an old one.


Koenkk:
Thanks!

PR #8298, opened 2024-11-10
Update tuya.ts

Liionboy:
Correction for Auto, Manule and Temporary Manual Mode

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6895,7 +6895,7 @@ const definitions: DefinitionWithExtend[] = [
                     tuya.valueConverterBasic.lookup((_, device) => {
                         // https://github.com/Koenkk/zigbee2mqtt/issues/21353#issuecomment-1938328429
                         if (device.manufacturerName === '_TZE200_viy9ihs7') {
-                            return {auto: tuya.enum(1), manual: tuya.enum(0), temporary_manual: tuya.enum(2)};
+                            return {auto: tuya.enum(0), manual: tuya.enum(1), temporary_manual: tuya.enum(2)};
                         } else {
                             return {auto: tuya.enum(0), manual: tuya.enum(1), temporary_manual: tuya.enum(2)};
                         }


src/devices/tuya.ts
```diff
@@ -6895,7 +6895,7 @@ const definitions: DefinitionWithExtend[] = [
                     tuya.valueConverterBasic.lookup((_, device) => {
                         // https://github.com/Koenkk/zigbee2mqtt/issues/21353#issuecomment-1938328429
                         if (device.manufacturerName === '_TZE200_viy9ihs7') {
-                            return {auto: tuya.enum(1), manual: tuya.enum(0), temporary_manual: tuya.enum(2)};
+                            return {auto: tuya.enum(0), manual: tuya.enum(1), temporary_manual: tuya.enum(2)};
```
Koenkk, line 6898: Won't this break https://github.com/Koenkk/zigbee2mqtt/issues/21353#issuecomment-1938328429 again (CC: @MrEcosse) ?

Liionboy: now is not working wiyh this onfiguration, auto is manual and manual is auto, with convert file is work normally with this values

Koenkk: I see, but it was reversed on purpose, see https://github.com/Koenkk/zigbee2mqtt/issues/21353#issuecomment-1938328429, waiting for @MrEcosse to reply

MrEcosse: It depends on the model.  The ZWT 198 and ZWT100 need different definitions because the manufacturer has different code for the program mode on each.  As per 21353:

The ZWT198 is fingerprint _TZE200_viy9ihs7 In the datapoints section the preset should be enumerated as follows:
[4, 'preset', tuya.valueConverterBasic.lookup({'auto': tuya.enum(0), 'manual': tuya.enum(1), 'temporary_manual': tuya.enum(2)})],

The ZWT100 is fingerprint _TZE204_lzriup1j. In the datapoints section the preset should be enumerated as follows:
[4, 'preset', tuya.valueConverterBasic.lookup({'manual': tuya.enum(0), 'auto': tuya.enum(1), 'temporary_manual': tuya.enum(2)})],

I've noticed in the latest release the two thermostat definitions have been combined into one - in which case one or other of them will not work correctly because of the above.  They need to be two separate definitions based on the fingerprints as above.

image

Koenkk: The if else here makes sure the correct modes are used for each. 

> The ZWT198 is fingerprint _TZE200_viy9ihs7 In the datapoints section the preset should be enumerated as follows:
[4, 'preset', tuya.valueConverterBasic.lookup({'auto': tuya.enum(0), 'manual': tuya.enum(1), 'temporary_manual': tuya.enum(2)})],

Then the current logic is wrong: `return {auto: tuya.enum(1), manual: tuya.enum(0), temporary_manual: tuya.enum(2)};`, this PR corrects this so I will merge it.

PR #8300, opened 2024-11-10
Add support for Schneider NH3527A PIR Dimmer switch

peroo:
Supports all features of the device.

When pairing it throws an error which I've not been able to figure out, but the occupancy sensitivity attribute works fine otherwise.
`z2m: Exception while calling fromZigbee converter: Expected one of: 50, 75, 100, got: 'undefined'`

The PR includes a new generic extender for lighting ballast, as well as some fixes for `levelConfig` and better naming/descriptions for `pirOToUDelay`


src/devices/schneider_electric.ts
```diff
@@ -311,6 +317,83 @@ const schneiderElectricExtend = {
...
+                return Math.round(10000 * Math.log10(value) + 1);
+            }
+        };
+
+        const luxThresholdExtend = numeric({
+            name: 'ambience_light_threshold',
+            cluster: 'manuSpecificSchneiderOccupancyConfiguration',
+            attribute: 'ambienceLightThreshold',
+            reporting: {min: '10_SECONDS', max: '1_HOUR', change: 5},
+            description: 'Threshold above which occupancy will not trigger the light switch.',
```
peroo, line 364: I'm not 100% sure about this definition. The manual claims it can be adjusted to avoid false detections when there are large changes in environmental lighting.

src/devices/schneider_electric.ts
```diff
@@ -311,6 +317,83 @@ const schneiderElectricExtend = {
...
+        extend.exposes.push(...luxThresholdExtend.exposes);
+        extend.configure.push(
+            setupConfigureForReading('manuSpecificSchneiderOccupancyConfiguration', ['ambienceLightThreshold']),
+            setupConfigureForReading('msOccupancySensing', ['schneiderOccupancySensitivity']),
+        );
+
+        return extend;
+    },
+
+    dimmingOnLevel: (): ModernExtend => {
```
peroo, line 382: Schneider dimmers don't support fixed on_levels for some reason, just 100% or previous.

enoch85:
Thank you! This seems perfect!

src/devices/schneider_electric.ts
FSund: FYI: `on_level` works with arbitrary levels for NHMOTION/UNIDIM/1 (when set via Dev console), so I don't think this is true in general.

peroo: Thanks, you are very much correct. I guess I assumed it didn't work based on the app not supporting it.

I guess we could revert to the default levelControl implementation, but as you pointed out in your PR the UI for it seems broken. It looks like numeric expose with presets are broken. I'll poke at it a bit.

src/lib/exposes.ts
```diff
@@ -428,8 +428,8 @@ export class Light extends Base {
                 new Numeric('current_level_startup', access.ALL)
                     .withValueMin(1)
                     .withValueMax(254)
-                    .withPreset('minimum', 0, 'Use minimum permitted value')
-                    .withPreset('previous', 255, 'Use previous value')
+                    .withPreset('minimum', 'minimum', 'Use minimum permitted value')
+                    .withPreset('previous', 'previous', 'Use previous value')
```
peroo, line 432: When using numeric preset values the presets wouldn't get the correct active state in the UI since the feature state is the string value (255 !== 'previous').

This shouldn't have any bearing on the actual state since the relevant fromZigbee/toZigbee always convert to the string value anyway.

FSund:
Very nice work. I'll have to utilize some of this for the NHMOTION/UNIDIM/1 PR.

Is there an easy way to test a custom zigbee-herdsman-converters commit locally in my HA installation?

FSund:
> What I'm doing is building the package then copying it to my z2m vm and installing it.

I see. I'm running HAOS and using Supervisor, so it might not be that easy in that case. Thanks anyway :)

enoch85:
@peroo Would you consider this ready for production now?

Would be great if this was merged soon. 👍🏼

src/devices/schneider_electric.ts
peroo: I've double checked that this is how it works. Useful for not triggering the light during the day when the room is bright enough already.

enoch85:
@Koenkk Would be great if you could take a look at this. I have like 30 of these installed in the house.

Koenkk:
Thanks!

Koenkk:

Make sure that:
- The filename is `NH3527A.png`
- The size is 512x512

Issue #8289, opened 2024-11-08
Take into account source endpoint when skipping duplicate transactions


martyn-vesternet:
@Koenkk should the "hasAlreadyProcessedMessage" function in utils take into account the source endpoint that the message is received from?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/e1694de10a8a4e25ff62042bcda3e58922ea4348/src/lib/utils.ts#L138

Otherwise, messages are skipped if they happen to use the same sequence number but are from a different source endpoint, for example in the case of an "all on" or "all off" button on a remote control that sends the same command from 4 source endpoints:

![image](https://github.com/user-attachments/assets/bddd3219-52e0-4b1c-a00f-bbc50ea961d3)

Here, the "true" duplicates would be 31391 - 31402 only, however the existing code rejects 31385 - 31402 as they all have the same sequence number.

I know that I can set "publishDuplicateTransaction" to true, but then I get actual real duplicates like:

![image](https://github.com/user-attachments/assets/d942f059-ccd9-4146-9910-1f4c0e818b89)

I'm using this in a local converter which seems to work fine to take into account the endpoint, so thought it worth a mention:

```
function hasAlreadyProcessedMessage(msg, model, ID = null, key = null) {
    if (model.meta && model.meta.publishDuplicateTransaction) return false;
    const currentID = ID !== null ? ID : msg.meta.zclTransactionSequenceNumber;
    key = key || msg.device.ieeeAddr + '-' + msg.endpoint.ID;
    if (transactionStore[key]?.includes(currentID)) return true;
    // Keep last 5, as they might come in different order: https://github.com/Koenkk/zigbee2mqtt/issues/20024
    transactionStore[key] = [currentID, ...(transactionStore[key] ?? [])].slice(0, 5);
    return false;
}
```

Koenkk:
Yes it should, thanks for looking into this, could you make a PR?

PR #8301, opened 2024-11-11
Update hasAlreadyProcessedMessage in utils.ts

martyn-vesternet:
Take into account source endpoint when skipping duplicate transactions.

Fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/8289


Koenkk:
Thanks!

PR #8304, opened 2024-11-11
fix!: Remove `illuminance_lux`

Koenkk:
- Moved `illuminance_lux` to `illuminance`, this breaks in two ways: `illuminance_lux` has been removed and the value of `illuminance` changes. This affects the following devices: 83633205, MOSZB-140, MOSZB-153, EFEKTA_iAQ3, EFEKTA_PWS_Max, EFEKTA_PWS_MaxPro, EFEKTA_eON29wz, EFEKTA_eFlower_Pro, EFEKTA_iAQ, E2134, POK004, POK011, ROB_200-070-0, HK-SENSOR-4IN1-A, AU-A1ZBPIRS, RADON TriTech ZB, TS0601_illuminance_temperature_humidity_sensor_1, TS0202_3, S-LUX-ZB, TS0222_light, THE01860A, TS0222, LCZ030, MIR-HE200-TY, XFY-CGQ-ZIGB, TS0601_smart_human_presence_sensor_1, Y1_IN, ZY-M100-S_1, ZY-M100-S_2, WZ-M100, ZG-205ZL, ZG-204ZM, YXZBRB58, ZY-M100-24G, ZY-M100-24GV2, ZY-M100-24GV3, MTG075-ZB-RL, RT_ZCZ03Z, ZY-M100-S_3, ZG-205Z, 8750001213, HSE2927E, ZB003-X, 07502L, JS-MC-SENSOR-ZB, KK-ES-J01W, KMPCIL_RES005, RTCGQ11LM, RTCGQ12LM, b-parasite, 7A-SS-ZABC-H0, 5AA-SS-ZA-H0, RTCGQ14LM, RTCGQ15LM, ZNCLBL01LM, GZCGQ01LM, GZCGQ11LM, 07047L, 9290012607, 9290019758, 9290030675, 9290030674, 545D6306, CCT595011, ZSS-ZK-THL, 4512770, TERNCY-PP01, MBD-S, CTM_MBD_Dim, 3RSNL02043Z, WB-MSW-ZIGBEE v.3, MS100, USM-300ZB, PIR313-E


CollaVinilica:
Why would you something like this? WHY? It broke every single sensor with a lux sensor I have. Now I have to waste an afternoon changing back all my entities IDs. At least you could have made it so I wouldn't rename (and disable!) pre-existing entities, c'mon!

nivek1612:
The worst thing for me is node-red does not appear to see any illuminance sensor now. So I will be rolling back.

hardwareadictos:
Same here!!! My HA instance doesn't see any illuminance_lux or illuminance sensor (that last should appear at least as this merge states). Rolling back...

CollaVinilica:
> Same here!!! My HA instance doesn't see any illuminance_lux or illuminance sensor (that last should appear at least as this merge states). Rolling back...

Try and take a look under the device info, in my case all the lux sensor associated with Aqara devices where disabled.. I had to manually enable them back and then rename their entity IDs to get my automations to work again.

Schermbecker:
> Try and take a look under the device info, in my case all the lux sensors associated with Aqara devices where disabled.. I had to manually enable them back and then rename their entity IDs to get my automations to work again.

That's the solution, thank you so much! 👍I also wondered, why all my illuminance entities were gone.

nivek1612:
> Try and take a look under the device info, in my case all the lux sensors associated with Aqara devices where disabled.. I had to manually enable them back and then rename their entity IDs to get my automations to work again.

That was it thanks.

hardwareadictos:
> > Same here!!! My HA instance doesn't see any illuminance_lux or illuminance sensor (that last should appear at least as this merge states). Rolling back...
> 
> Try and take a look under the device info, in my case all the lux sensors associated with Aqara devices where disabled.. I had to manually enable them back and then rename their entity IDs to get my automations to work again.

That did the trick. Thanks!

sj3fk3:
_Moved illuminance_lux to illuminance, this breaks in two ways: illuminance_lux has been removed **and the value of illuminance changes.**_ 

Sorry, but where can I find HOW the value changes? I've got a couple of GZCGQ01LM and they now show the same high values they showed when first introduced and before the [whole discussion](https://github.com/Koenkk/zigbee2mqtt/issues/2837) on how to convert those values to "proper" Lux values. Was this intended? Just not clear on why this seems to be reverted?

redelschaap:
@Koenkk I get that this change was made. However, this does not only break integrations because of the changed name, it also breaks integrations beause the values of the two entities are way too different. In my case, I have to change more than 20 integrations because of the different values. This means that I also have to discover the correct threshold values again.

I have some Aqara GZCGQ11LM sensors. The previous illuminance values were much more detailed than the lux values. For some reason, it also seemed to be more reliable/consistent. See the screenshot below.

Could we at least get the raw value back, perhaps as a new entity? Or is there any way to do this ourselves?

image

Koenkk:
Let's discuss the `illuminance_raw` in https://github.com/Koenkk/zigbee2mqtt/issues/25379 (please upvote the OP)

redelschaap:
I don't get why the values are the same either. But that's why I want to be able to use the values from the top graph, they are much more realistic (and useful) to me.

Viald:
Same issue here. When the day breaks my external sensor goes from 4769 lux to 3011 lux, and from 3011 lux to 1 lux in 10 minutes. Before, each step was around 80 lux.
Before dusk was between 120 lux and 20 lux. And night was around 20 lux and lower. It's now impossible to detect this steps.
I don't really understand what conversion is applied by z2m 2.0

DirtyHerby:
> > Same here!!! My HA instance doesn't see any illuminance_lux or illuminance sensor (that last should appear at least as this merge states). Rolling back...
> 
> Try and take a look under the device info, in my case all the lux sensors associated with Aqara devices where disabled.. I had to manually enable them back and then rename their entity IDs to get my automations to work again.

Where do I find this settings to enable it manually? I use Hue motion sensors.

cyruz-git:
I see that this is fixable on HomeAssistant. Any idea how to fix this in Homebridge? I can see the values in Z2M but nothing in Homebridge and of course all broken automation in Homekit.

PR #8313, opened 2024-11-13
Elko Super Thermostat - Configurable reporting of local temperature based on sensor choice

torandreroland:


src/converters/fromZigbee.ts
```diff
@@ -1921,35 +1921,105 @@ const converters1 = {
...
         type: ['attributeReport', 'readResponse'],
+        options: [exposes.options.local_temperature_based_on_sensor()],
         convert: (model, msg, publish, options, meta) => {
             const result: KeyValueAny = {};
             const data = msg.data;
-
+            if (data.localTemp !== undefined) {
+                const value = precisionRound(msg.data['localTemp'], 2) / 100;
+                const valuesFloorSensor = ['floor', 'supervisor_floor'];
+                const sensorType = globalStore.getValue(msg.endpoint, 'sensor');
```
Koenkk, line 1931: It would be better to get the value from the `meta.state`, the `globalStore` is always reset after every z2m restart.

src/devices/elko.ts
```diff
@@ -37,7 +37,7 @@ const definitions: DefinitionWithExtend[] = [
         model: '4523430',
         vendor: 'ELKO',
         description: 'ESH Plus Super TR RF PH',
-        fromZigbee: [fz.elko_thermostat, fz.thermostat],
+        fromZigbee: [fz.elko_thermostat],
```
Koenkk, line 40: Why remove `fz.thermostat`? It requires a lot of code duplication in `fz.elko_thermostat`

src/converters/fromZigbee.ts
torandreroland: Thank you for the valuable input. I agree (now that I understand the meta.state capability) that this is much better and I've made the change.

src/devices/elko.ts
torandreroland: The reason is to have a have different handling of localTemp from that in fz.thermostat. Is there a way of having a handling of localTemp in fz.elko_thermostat taking precedence of that in fz.thermostat to avoid the code duplication?

Koenkk: Yes, see https://github.com/Koenkk/zigbee-herdsman-converters/blob/347eeb3a4ce88bf7f2eb9490f8881e1d45f5e04c/src/devices/plugwise.ts#L31 as an example

torandreroland: Thank you very much for the pointer! I've updated the pull request with this.

Koenkk:
Thanks!

PR #8319, opened 2024-11-15
Add another variation of Tuya HY368 radiator termostat

xdanik:
Seems like just another variation of the same product. I have manually edited `node_modules/zigbee-herdsman-converters/devices/tuya.js` on my local installation of Zigbee2Mqtt to include this line and it seems to work fine.


Koenkk:
Thanks!

PR #8321, opened 2024-11-16
Fixed LoraTap SC400ZB-EU calibration time reading

PedroQ:
Calibration times for my _TZ3000_cet6ch1r devices were not being read, so went looking at the logs and found the cause:

`"calibration_time_'down'":17,"calibration_time_'up'":17,"calibration_time_down":null,"calibration_time_up":null`

Removing these single quotes fixes the issue.


Koenkk:
thanks!

PR #8323, opened 2024-11-16
Add support for TECH VNTH-T2 Smart Radiator Valve (Tuya thermostat)

christophe-r:
Add support for Smart Radiator Valve (Tuya thermostat) :
Brand: **TECH**
Model: **VNTH-T2**
Fingerprint: **_TZE204_r7brscr6**

The image file has been added to zigbee2mqtt.io:
https://github.com/Koenkk/zigbee2mqtt.io/blob/master/public/images/devices/VNTH-T2.png

### DIFF
--- a/src/devices/tech.ts
+++ b/src/devices/tech.ts
@@ -11,7 +11,7 @@ const definitions: DefinitionWithExtend[] = [
         fingerprint: [{modelID: 'TS0601', manufacturerName: '_TZE204_r7brscr6'}],
         model: 'VNTH-T2',
         vendor: 'TECH',
-        description: 'Smart Radiator Valve',
+        description: 'Smart radiator valve',
         ota: ota.zigbeeOTA,
         onEvent: tuya.onEventSetLocalTime,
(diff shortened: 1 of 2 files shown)


Koenkk:
Please use `datapoints` instead, [example Tuya thermostat](https://github.com/Koenkk/zigbee-herdsman-converters/blob/f7110985024ac51bb1523ac389de843abf186f95/src/devices/tuya.ts#L4393)

christophe-r:
@Koenkk Thanks for your review.

It's done.

Indeed, it's much simpler :)

src/devices/tech.ts
```diff
@@ -0,0 +1,103 @@
+import {ota} from '../index';
```
Koenkk, line 1: Please also add this file to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts

Koenkk, line 1:
```suggestion
import * as ota from '../lib/ota';
```

christophe-r:
@Koenkk All done

PR #8324, opened 2024-11-16
Adding ZSS-QY-LS-C

saxel-cz:


Koenkk:
`_TZ3000_9kbbfeho` is already supported: https://github.com/Koenkk/zigbee-herdsman-converters/blob/3f3549f79a38e8daabbca41760672bc73e0c42c6/src/devices/tuya.ts#L7036

PR #8325, opened 2024-11-16
Fix Somfy blind

heisenberg2980:
Removed tilt and fixed position and status of the blind


heisenberg2980:
@Koenkk this PR has been merged, but I just removed the external converter from my instance and after restarting Z2M, it shows the blind as not recognised again. I already updated Z2M to version 2.0, is there anything I´m missing?

PR #8328, opened 2024-11-17
Detect `TRADFRIbulbG125E26WSopal440lm` as IKEA LED1936G5

audunmg:
I don't have a picture handy, but the model number LED1936G5 is printed on the base.
The bulb is bought in Japan, which is probably why it's different.


Koenkk:
thanks!

PR #8329, opened 2024-11-17
Fix overall power for TS0601_3_phase_clamp_meter

mwdmwd:
This fixes total power reporting for negative power on the `TS0601_3_phase_clamp_meter` (without relay) by making it also use the `power` converter added in f7dfbc4dcd9c8744973a3536d43893faae643fa3.


Koenkk:
thanks!

PR #8332, opened 2024-11-17
feat: Support live loading of external converters

Nerivec:
For https://github.com/Koenkk/zigbee2mqtt/pull/24764

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -80,40 +80,43 @@ function arrayEquals<T>(as: T[], bs: T[]): boolean {
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -98,17 +127,15 @@ const converterRequiredFields = {
     toZigbee: 'Array',
 };
 
-function validateDefinition(definition: Definition) {
+function validateDefinition(definition: Definition): asserts definition is Definition {
```
Koenkk, line 130: I think the validateDefinition can be made optional, only has to be done for external converters. Official converters are already covered by typescript, this stems from the JS era 😄  Should give some more perf improvements!

PR #8333, opened 2024-11-17
feat: Add occupancy timeout for Tuya IH012-RT01/ZMS-102

srett:
As discussed in the [referenced issue](https://github.com/Koenkk/zigbee2mqtt/issues/24621), add a safety timeout that would reset the occupancy of the mentioned PIR sensors if the recurring occupancy messages stop, and no clearance message was received.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1851,12 +1851,14 @@ const definitions: DefinitionWithExtend[] = [
         model: 'IH012-RT01',
         vendor: 'Tuya',
         description: 'Motion sensor',
-        fromZigbee: [fz.ias_occupancy_alarm_1, fz.ignore_basic_report, fz.ZM35HQ_attr, fz.battery],
+        fromZigbee: [fz.ignore_basic_report, fz.ZM35HQ_attr, fz.battery],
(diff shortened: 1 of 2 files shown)


Koenkk:
Could you add this to the modernExtend instead: https://github.com/Koenkk/zigbee-herdsman-converters/blob/0306b3969990effe612477f6f93aa47364004e32/src/lib/modernExtend.ts#L1464 ? Example how to use it: https://github.com/Koenkk/zigbee-herdsman-converters/blob/0306b3969990effe612477f6f93aa47364004e32/src/devices/gs.ts#L57

srett:
@Koenkk took me a bit longer to get back to this but here's my attempt. I thought about somehow unifying this with the existing timeout intended for sensors that *only* send occupation messages and never clearance, but I couldn't think of a nice way to do so, since we want the timeout configurable for these devices, but not for those that send both types of messages. I named it keepAliveTimeout to distinguish from the other one, and added an according comment.

While updating the according devices to use modernExtend, I also removed the tamper attribute since it's not used. The same might be true for the battery_low attribute, but the lowest one is as 27% currently so that might take some more time to verify.

Koenkk:
Thanks!

PR #8348, opened 2024-11-21
Notification of successful device update after re-interview for Yandex devices

kirovilya:


src/devices/yandex.ts
```diff
@@ -126,32 +127,63 @@ function YandexCluster(): ModernExtend {
...
+        {
+            cluster: 'genBasic',
+            type: ['attributeReport', 'readResponse'],
+            convert: (model, msg, publish, options, meta) => {
+                localMeta = meta;
+                return;
+            },
+        } satisfies Fz.Converter,
+    ];
+    const onEvent: OnEvent = async (type, data, device, settings, state) => {
```
Koenkk, line 156: I think it would be better to pass `deviceExposesChanged` as an argument of event, preferably add `meta` there which is an object containing `deviceExposesChanged`. Because currently the code will not work when no genBasic/genOnOff message has been received yet.

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
-pnpm run pretty:check
+#pnpm run pretty:check
```
Koenkk, line 1: Please enable

src/index.ts
```diff
@@ -453,7 +454,7 @@ export function findByModel(model: string) {
 
 // Can be used to handle events for devices which are not fully paired yet (no modelID).
 // Example usecase: https://github.com/Koenkk/zigbee2mqtt/issues/2399#issuecomment-570583325
-export async function onEvent(type: OnEventType, data: OnEventData, device: Zh.Device) {
+export async function onEvent(type: OnEventType, data: OnEventData, device: Zh.Device, meta: Fz.Meta) {
```
Koenkk, line 457: I think it's cleaner to make a new meta, `OnEvent.Meta`

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt/pull/24863

.husky/pre-commit
kirovilya: done

src/index.ts
kirovilya: added. right?

Koenkk:
Thanks!

PR #8352, opened 2024-11-21
fix: Detect ZY-M100-24G wall mounted variant

kriskbx:
I bought a different variant of the same sensor that was not detected properly. This PR fixes it. I tested the changes in zigbee2mqtt and the sensor works properly in my setup.


Koenkk:
Thanks!

Rldye:
I am new at this but for some reason, I cannot get _TZE200_ya4ft0w4 to setup for me.  I only get that this sensor cannot be configured.

PR #8360, opened 2024-11-23
I corrected the definition.

saxel-cz:


Koenkk:
It's already defined here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/52f0d4093469e8f8ab678168d2a4d3f1c6cfba5c/src/devices/tuya.ts#L7055

PR #8361, opened 2024-11-23
feat(add):  support Tuya `_TZ3210_bfwvfyx1` as TS0505B

will-tm:
## Description
This Pull Request adds support for Tuya 15W led light bulb E27 RGBCW from [AliExpress](https://www.aliexpress.us/item/1005005721435025.html).
There is currently no support for `_TZ3210_bfwvfyx1` fingerprint which causes this devices to be defaulted to `TS0505B_1` whose implementation is barely compatible and unable to properly control color.

Untitled

### Additional Info
In order to just get the color working `_TZ3210_bfwvfyx1` could just be added to `TS0505B_2` supported fingerprints but this new device accepts a color temperature range from `50` to `1000` instead so I took the initiative to give it its own definition.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1681,6 +1681,19 @@ const definitions: DefinitionWithExtend[] = [
         },
     },
     {
+        fingerprint: tuya.fingerprint('TS0505B', ['_TZ3210_bfwvfyx1']),
+        model: 'TS0505B_4',
+        vendor: 'Tuya',
+        description: 'Zigbee RGB+CCT light',
+        whiteLabel: [tuya.whitelabel('Tuya', 'TS0505B_4_4', 'Zigbee 3.0 15W led light bulb E27 RGBCW', ['_TZ3210_bfwvfyx1'])],
+        toZigbee: [tz.on_off, tzLocal.led_control, tuya.tz.do_not_disturb],
+        fromZigbee: [fz.on_off, fz.tuya_led_controller, fz.brightness, fz.ignore_basic_report],
+        exposes: [e.light_brightness_colortemp_colorhs([50, 1000]).removeFeature('color_temp_startup'), tuya.exposes.doNotDisturb()],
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -1680,6 +1680,19 @@ const definitions: DefinitionWithExtend[] = [
             device.getEndpoint(1).saveClusterAttributeKeyValue('lightingColorCtrl', {colorCapabilities: 29});
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0505B', ['_TZ3210_bfwvfyx1']),
+        model: 'TS0505B_4',
+        vendor: 'Tuya',
+        description: 'Zigbee RGB+CCT light',
+        whiteLabel: [tuya.whitelabel('Tuya', 'TS0505B_4_4', 'Zigbee 3.0 15W led light bulb E27 RGBCW', ['_TZ3210_bfwvfyx1'])],
```
Koenkk, line 1688:
```suggestion
        whiteLabel: [tuya.whitelabel('Tuya', 'TS0505B_4_1', 'Zigbee 3.0 15W led light bulb E27 RGBCW', ['_TZ3210_bfwvfyx1'])],
```

Koenkk:
Thanks!

Make sure that:
- The filename is `TS0505B_4.png`
- The size is 512x512

will-tm:
Will do, thank you for merging!

HDHristov:
Hi.

Prior to this addition, the bulbs were recognized as some sort of generic led controllers. This had the negative side effect that the color temperature range was limited.
Few things have changed with this addition:

1. Color temperature range is extended, but the color temperature in Kelvin doesn't match the actual color temperature.
2. There is no more a "transition time" setting in the Settings (specific) tab in z2m. I don't know if this disappeared from this particular addition or due to some other change.
3. There was some (limited) support for effects, which is now missing.

PR #8362, opened 2024-11-23
Busch-Jaeger 6735/6736/6737: Drop deprecated switch state

klada:
Since this device has breaking changes in Z2M 2.0 (payload changes) we can use this as a chance to drop the deprecated state as well.

These are multi-endpoint devices and therefore the state needs to be prefixed by the endpoint name (`state_relay`). The `state` attribute never worked properly (e.g. it did not update when device was turned on directly).

For Home Assistant users there has been a note in the [docs](https://www.zigbee2mqtt.io/devices/6735_6736_6737.html#home-assistant-device-discovered-both-as-light-and-switch) for quite a while that the `switch` entity needs to be ignored.

### DIFF
--- a/src/devices/busch_jaeger.ts
+++ b/src/devices/busch_jaeger.ts
@@ -45,10 +45,6 @@ const definitions: DefinitionWithExtend[] = [
             //  2. The top rocker will not be usable (not emit any events) as it's hardwired to the relay/dimmer
             if (!device || device.getEndpoint(0x12) != null) {
                 expose.push(e.light_brightness().withEndpoint('relay'));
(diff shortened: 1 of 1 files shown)


klada:
@Koenkk I think Z2M 2.0 is a good chance to get this `state` vs. `state_relay` fixed for these devices. What do you think? This would need to be added to the release notes, the 6735/6736/6737 devices are already on your list with breaking changes.

Koenkk:
Thanks, could you also make a PR to update the docs? Target the `feat/2.0.0` branch.

klada:
OK, I have updated the docs as well: https://github.com/Koenkk/zigbee2mqtt.io/pull/3260

PR #8369, opened 2024-11-24
Exposing action_brightness_delta to allow better blueprints

TheUnlimited64:
This enables home assistant users to have a better dimming experience using automations like, therefore zha is not needed anymore for smooth dimming experience.

[https://github.com/[TheUnlimited64/tuya_smart_knob_blueprint](https://github.com/TheUnlimited64/tuya_smart_knob_blueprint)](url)

Only issue currently (atleast with mine) is that there seems to be a cooldown after ~2s. If I dimm long enough it stops receiving. But I'm not sure how to mitigate it.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -7598,6 +7598,7 @@ const definitions: DefinitionWithExtend[] = [
                 'rotate_left',
                 'rotate_right',
             ]),
+            e.numeric("action_brightness_delta", ea.STATE).withValueMin(-255).withValueMax(255),
             e.numeric('action_step_size', ea.STATE).withValueMin(0).withValueMax(255),
             e.numeric('action_transition_time', ea.STATE).withUnit('s'),
             e.numeric('action_rate', ea.STATE).withValueMin(0).withValueMax(255),


Koenkk:
This expose should not always be present, since it is only there when the `simulated_brightness` option is activated. I would add it here instead: https://github.com/Koenkk/zigbee2mqtt/blob/69f728295224c06ce509eb3afdf611b37ca8fa37/lib/extension/homeassistant.ts#L1535 (` if (isDevice && entity.options.simulated_brightness) {`

TheUnlimited64:
@Koenkk thats a great Idea! Wasn't really sure how the exposing works, but I'm digging deaper. Is there maybe a dev docs? I already have an additional feature for this and potentially more sensors in mind.

But I'm very pleased that I can overwrite pretty much everything with a custom converter, makes developing/testing a breeze :)

Tomorrow or in a few days I'll add it when I figured everything out and put it in the right place

TheUnlimited64:
Hi @Koenkk,

I've updated my PR with a new feature implementation called `simulated_color_temperature`. 
The implementation is more straightforward compared to `simulated_brightness` as my sensor 
doesn't provide a `command_stop` functionality, so the interval would be useless currently.

Additionally, I'd appreciate some guidance on implementing the exposure in `homeassistant.ts`. 
I'm not entirely clear on the exposure line's functionality, which is preventing me from 
implementing the correct exposure method.

Could you please provide some direction on this?

Thanks!

src/converters/fromZigbee.ts
```diff
@@ -1424,6 +1425,22 @@ const converters1 = {
                 payload.action_transition_time = msg.data.transtime / 100;
             }
 
+            if (options.simulated_color_temperature) {
+                const deltaOpts = options.simulated_color_temperature_delta ?? 500;
```
Koenkk, line 1429: I assume the delta is in Kelvin, I propose to use mireds here as that is what the ZCL uses.

src/converters/fromZigbee.ts
```diff
@@ -1424,6 +1425,22 @@ const converters1 = {
...
+
+                if (globalStore.getValue(msg.endpoint, 'simulated_color_temperature') === undefined) {
+                    let color_temperature = globalStore.getValue(msg.endpoint, 'simulated_color_temperature_temperature', 2000);
```
Koenkk, line 1432:
```suggestion
                    let colorTemperature = globalStore.getValue(msg.endpoint, 'simulated_color_temperature_temperature', 2000);
```

TheUnlimited64:
> Can you clarify it a bit more? Then I can help.

You suggested instead of exposing the additional information `action_brightness_step` (and therefore also `action_color_step_temperature`) to expose it with the `homeassistant.ts` file. But I'm not sure how that would look like. 

Also wouldn't hardcoding it be a bad pattern? A more straight forward principle might be better for future solutions, doesn't it? Maybe we could define base exposes but overwrite them at device level? That way an user has control to expose any mqtt data that it needs. What do you think about this solution?

Also whats a bit ugly is that in the options descriptions often says like `default is 20`. But that 20 is hardcoded into the converter function itself. I did try to think of an solution to set an default at option level but that wouldn't work with the current architecture. Atleast I can't think of a good way of implementing such functionality and it would presuppose that an option is toggleable via the UI (because if they're default values any option would be turned on automatically)

Koenkk:
An alternative would be to dynamically add it to exposes, you cannot always add it there since not everyone is using this option. Then it can be added here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/b7877b7d0b10f64a827e19801d39a4b021043dd3/src/index.ts#L316


(which would mean `prepareDefinition` needs to receive the `options`)

TheUnlimited64:
Hi!

I've now pushed my proposal of the additional options properties being spread into the exposes. What do you think? I'm a bit confused that the classes have multiple use cases (option definition and exposes)

TheUnlimited64:
Also, how could I test my changes with in home assistant? Or do I have to setup zigbee2mqtt within docker myself?

Koenkk:
See https://github.com/zigbee2mqtt/hassio-zigbee2mqtt/blob/master/CONTRIBUTING.md#developing-against-the-home-assistant-container on how to make local changes.

TheUnlimited64:
Hi!

I've finally had time to implement it and test it. What do you think?

src/index.ts
```diff
@@ -376,6 +384,48 @@ function prepareDefinition(definition: DefinitionWithExtend): Definition {
         }
     }
 
+    //if options are defined, add them to the exposes
```
Koenkk, line 387: This can be done much easier, see https://github.com/Koenkk/zigbee-herdsman-converters/commit/3d98d3f54bc63a720111f58194269250f5fb324a#diff-47eee2b61258a9d48e4b9104d706a4aeacec2e1baba7d21f34db7d2963357d28R679 as an example

TheUnlimited64: Am I correct, that you're implying that it better should be done by the modernExtendRefactor script, to migrate all definitions to extend a simulated_light base? That sounds to be a great approach, but I'm a bit unsure if it could handle the edge cases like the D1 universal dimmer, that's an example of why this got this complicated, because it acts as a dimmer, but also as a light entity.

Is this what you're intending to push forward, or am I on the wrong track? If not I'll read into the code base and check how to integrate it

Koenkk: Sorry, I meant the changes of `src/lib/modernExtend.ts`

PR #8370, opened 2024-11-24
feat(add): Shyugj DoorSensor-ZB3.0

bobrippling:
This Pull Request adds support for the Shyugj Door Sensor. There is currently no support for this vendor, so a definition file has been created for it.

### DIFF
--- a/src/devices/shyugj.ts
+++ b/src/devices/shyugj.ts
@@ -3,10 +3,10 @@ import {DefinitionWithExtend} from '../lib/types';
 
 const definitions: DefinitionWithExtend[] = [
     {
-        zigbeeModel: ['DoorSensor-ZB3.0'],
-        model: 'DoorSensor-ZB3.0',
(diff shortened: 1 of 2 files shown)


Koenkk:
Could you provide the data/database.db entry of this device?

src/devices/shyugj.ts
```diff
@@ -0,0 +1,22 @@
+import {battery, iasZoneAlarm} from '../lib/modernExtend';
+import {DefinitionWithExtend} from '../lib/types';
+
+const definitions: DefinitionWithExtend[] = [
+    {
+        zigbeeModel: ['DoorSensor-ZB3.0'],
+        model: 'DoorSensor-ZB3.0',
```
Koenkk, line 7: Is there a better model number written on the device or product page?

bobrippling: The only other identifier I have is an item number on the setup documentation, `S901D-ZG`, if that feels more useful?

Koenkk:
Thanks!

PR #8371, opened 2024-11-24
Update Schneider PIR switch/Rotary dimmer config

peroo:
Replaces manual config with modern dimmer/occupancy extends. I've verified all functionality works correctly.

I've also updated the model numbers to use what I believe are the official Schneider model numbers (they are stamped on the back, and used in documentation). Schneider has a large number of country specific subsidiaries with different white label versions, so I think it's preferable to use the neutral model number.

Documentation update PR: https://github.com/Koenkk/zigbee2mqtt.io/pull/3253

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -758,33 +758,32 @@ const definitions: DefinitionWithExtend[] = [
     },
     {
         zigbeeModel: ['NHROTARY/UNIDIM/1'],
-        model: 'WDE002961',
+        model: 'NH3516A',
         vendor: 'Schneider Electric',
         description: 'Rotary dimmer',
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

FSund:
Can I ask how this works when the device has multiple endpoints? Is there some magic behind the scenes that figure out which endpoint to use for the different functions?

peroo:
> Can I ask how this works when the device has multiple endpoints? Is there some magic behind the scenes that figure out which endpoint to use for the different functions?

My understanding is if endpoints aren't specified modern extends will use whichever has a matching cluster. https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/modernExtend.ts#L254C17-L254C34

Presumably this won't work for e.g. a four-gang switch where all the endpoints are the same type, but it works fine for these devices since they have distinct endpoints.

PR #8380, opened 2024-11-25
Busch-Jaeger 6735/6736/6737: Don't expose legacy payload anymore

klada:
Since legacy mode is gone we should adjust the exposed actions to the actual values.


Koenkk:
Thanks!

Issue #6418, opened 2023-11-06
Thermostat TV02 throws error exeption on setting schedule for each day of the week


bruno4712:
Setting a  schedule for one day of the week in the format like this { "schedule": { "week_day": "monday", "schedule": "06:00/20 22:00/19 24:00/18" } } via field send_payload throws a exeption. It is not possible to set the value on respective object because the field working_day is not read from the object list here.

ioBroker.zigbee v1.8.23 zigbee-herdsman-converters v15.107.0

protocol:

Stack trace for null: Error: Invalid "working_day" property, need to set it before at Object.to (/opt/iobroker/node_modules/iobroker.zigbee/node_modules/zigbee-herdsman-converters/src/lib/tuya.ts:641:23) at Object.convertSet (/opt/iobroker/node_modules/iobroker.zigbee/node_modules/zigbee-herdsman-converters/src/lib/tuya.ts:897:57) at /opt/iobroker/node_modules/iobroker.zigbee/main.js:735:52

https://www.zigbee2mqtt.io/devices/TV02-Zigbee.html

This occurs on following thermostats: 
TV02-Zigbee / TS0601 manufacturer: _TZE200_hue3yfsn and _TZE200_mudxchsu

The processing of the field weekScheduleType must be fixed here.

hard copy from debugger:

![tuya js weekScheduleType v15 107 0](https://github.com/Koenkk/zigbee-herdsman-converters/assets/127872559/580ce6d0-7203-41ec-bc92-5e6d6698d9ba)

Patching the field weekScheduleType in the debugger as work around helps to fill the 7 schedule fields.

see also this issue https://github.com/ioBroker/ioBroker.zigbee/issues/1898

9Mad-Max5:
Didn't work at my side.
Still get the same error only hard coding it to a value helped at my situation.

bruno4712:
who can fix this?

9Mad-Max5:
I wasn't able to make a proper fix in my mind.
But I let it set default to "separate" this at least makes the code working.

Here is the according pull request.
https://github.com/Koenkk/zigbee-herdsman-converters/pull/8386

Main issue for me understanding is that during the send_payload from the iobroker.zigbee only the message itself is provided, this will unfortunately never contain the working_day setting.
With sticking with separate is unfortunate but at least give's the possibility to use the thermostats.

bruno4712:
In my new installation I use following versions of the npm modules.
[...]

PR #8386, opened 2024-11-26
added a default to set the temperature

9Mad-Max5:
I ran in to the same Issue as in https://github.com/Koenkk/zigbee-herdsman-converters/issues/6418
Unfortunately, I hadn't been able how to figure out a way to pass the information about the schedule type. Therefore, I felt back to a system with a default value of separate, which is the least restrictive.
It is not ideal but at least gets the things going.
Hopefully that helps the affected people, even if my frontend in form of iobroker.zigbee is still using version 19.


Koenkk:
Thanks!

PR #8391, opened 2024-11-28
Set the path to pnpm store inside the dev container

rohankapoorcom:
When running pnpm in the dev container, it doesn't have the path to the store directory set (it's not inherited from the environmental variables because it's located in a different volume. This can lead pnpm to create a store directory inside the repository itself which is not ideal. Setting it as follows in the `/home/node` folder avoids permissions errors during running the dev container. See https://github.com/pnpm/pnpm/issues/5803#issuecomment-1974820613 for more details.


Koenkk:
Thanks!

PR #8393, opened 2024-11-28
Improvement to W599001 Smoke Alarm

mattbyte:
Adding support to expose heat alarm and hush mode bringing support closer to parity with that of the manufacturer hub and app. The heat alarm is for detecting fires and will sound the device siren, therefore, it is important that it be exposed for use. I used fzLocal as the hush bit is appears to be manufacturer specific (not specified in ZCL).


src/devices/schneider_electric.ts
```diff
@@ -565,6 +565,26 @@ const fzLocal = {
             return ret;
         },
     } satisfies Fz.Converter,
+    ias_smoke_alarm_1_heat_alarm_2_with_hush: {
```
Koenkk, line 568: could you extend and use: https://github.com/Koenkk/zigbee-herdsman-converters/blob/2d229ffdae8f31c9dc1825ffa1ee4940d9cd029c/src/lib/modernExtend.ts#L1487 instead? 

Example usage: https://github.com/Koenkk/zigbee-herdsman-converters/blob/2d229ffdae8f31c9dc1825ffa1ee4940d9cd029c/src/devices/adeo.ts#L439

mattbyte: As in make a local version of the iasZoneAlarm modern extend that can do the hush mode bit decode?

Koenkk: Add the functionality to `modernExtend.ts`

mattbyte: No worries, done! Have tested and it correctly configures the reporting and binding as before. Only slight negative is no ability to tell the user that the second alarm is heat while the first is smoke.

image

One option would be to add a new argument maybe called secondaryZoneType?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/2d229ffdae8f31c9dc1825ffa1ee4940d9cd029c/src/lib/modernExtend.ts#L1480-L1486

I think it is also acceptable as is, at least both alarms are exposed.

src/lib/modernExtend.ts
```diff
@@ -1574,6 +1576,9 @@ export function iasZoneAlarm(args: IasArgs): ModernExtend {
                     if (args.zoneAttributes.includes('battery_defect')) {
                         payload = {battery_defect: (zoneStatus & (1 << 9)) > 0, ...payload};
                     }
+                    if (args.zoneAttributes.includes('hush')) {
```
Koenkk, line 1579: I reliase now that this option is device specific, therefore it cannot be put directly here. I think we should allow passing custom attributes to `iasZoneAlarm` instead.

mattbyte:
Have had a go at implementing passing of custom (manufacturer) zone attributes to `iasZoneAlarm`. Not yet fully tested as I wanted to confirm I'm on the right track first. Based on the `binary` modern extend of course.

Koenkk:
Well done, thanks!

PR #8397, opened 2024-11-29
Adding new ID for Tuya SOS button

aknarts:
Just purchased the MOES SOS button 

It is just rebranded Tuya SOS button - battery, one button 

Zigbee Model
TS0215A
Zigbee Manufacturer
_TZ3000_9r5jaajv


aknarts:
This device is strictly not an SOS button but works like one, the configuration works if I have it as an external converter.

Koenkk:
Thanks!

PR #8402, opened 2024-11-29
Reduce code duplication and clean up the Inovelli Converter

rohankapoorcom:
In preparation for the new VZM30-SN, I spent some time cleaning up the existing converter, reducing duplication and moving as much as possible to use modern extends.

@InovelliUSA please review when you have some time.

The power / energy monitoring multipliers don't seem to be working currently (I commented on the specific line below).

### DIFF
--- a/src/devices/inovelli.ts
+++ b/src/devices/inovelli.ts
@@ -1979,7 +1979,19 @@ const definitions: DefinitionWithExtend[] = [
         vendor: 'Inovelli',
         description: '2-in-1 switch + dimmer',
         exposes: exposesListVZM31.concat(identify().exposes as Expose[]),
-        extend: [inovelliExtend.addCustomClusterInovelli()],
+        extend: [
+            deviceEndpoints({
+                endpoints: {'1': 1, '2': 2, '3': 3},
+                multiEndpointSkip: ['state', 'power', 'energy', 'brightness'],
+            }),
+            inovelliExtend.addCustomClusterInovelli(),
+            electricityMeter({
+                current: false,
+                voltage: false,
+                power: {min: 15, max: 3600, change: 1},
+                energy: {min: 15, max: 3600, change: 0},
+            }),
+        ],
         toZigbee: [
             tzLocal.light_onoff_brightness_inovelli,
             tz.power_on_behavior,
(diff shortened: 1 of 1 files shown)


src/devices/inovelli.ts
```diff
@@ -2167,7 +1974,19 @@ const definitions: DefinitionWithExtend[] = [
...
+        extend: [
+            deviceEndpoints({
+                endpoints: {'1': 1, '2': 2, '3': 3},
+                multiEndpointSkip: ['state', 'power', 'energy', 'brightness'],
+            }),
+            inovelliExtend.addCustomClusterInovelli(),
+            electricityMeter({
+                current: false,
+                voltage: false,
+                power: {min: 15, max: 3600, change: 1},
```
rohankapoorcom, line 1986: The min/max/change here doesn't seem to be respected (I'm not sure why). When I run a reconfigure, it's added with `activePower` min: 10, max: 65000, change: 5 and `currentSummDelivered` min: 10, max: 65000, change: 0.1 instead of what's provided here (matching the custom call below)

src/devices/inovelli.ts
```diff
@@ -2178,46 +1997,28 @@ const definitions: DefinitionWithExtend[] = [
...
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
-            await reporting.bind(endpoint, coordinatorEndpoint, ['seMetering', 'haElectricalMeasurement', 'genOnOff', 'genLevelCtrl']);
+            await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl']);
             await reporting.onOff(endpoint);
 
             // Bind for Button Event Reporting
             const endpoint2 = device.getEndpoint(2);
             await reporting.bind(endpoint2, coordinatorEndpoint, ['manuSpecificInovelli']);
-            await endpoint.read('haElectricalMeasurement', ['acPowerMultiplier', 'acPowerDivisor']);
```
rohankapoorcom, line 2200: Are these two lines needed? Or does this happen automatically now?

I'm concerned that the correct multipliers/divisors are not being used anymore since the power is being measured as 500-600w when it previously was showing up as 50-60w (before this change)

Koenkk:
@InovelliUSA please review

src/devices/inovelli.ts
InovelliUSA: There are two reporting mechanisms in the device that can sometimes overlap. When we first created the device we took our configuration options from our Z-Wave switch not knowing (at the time) that Zigbee had its own built in energy / power reporting functionality. So, you can have different configurations in the exposes tab vs the reporting tab. If I could do it again, I would remove config options 17-19 and only used the built in Zigbee reporting. Anyway, I think that is why you may be seeing this. Try setting P17-19 to 0 and then test the reporting configuration.

rohankapoorcom: Sorry, I think my original message wasn't very clear.

I haven't changed any of the reporting parameters in the exposes tab. When the original code is running (see single commit diff: https://github.com/Koenkk/zigbee-herdsman-converters/pull/8402/commits/44b0a9afe73d97ce2acb0a59410ee4a2c1a292c4), it explicitly sets (manually) the reporting to the min/max/change values as described on line 2025 and 2026. This leads to the following screenshot:

image


I switched the code to use the modern extend instead and passed it along the same min/max/change values as before. However, this doesn't seem to be taking affect. Leading to the following screenshot:

```
electricityMeter({
    current: false,
    voltage: false,
    power: {min: 15, max: 3600, change: 1},
    energy: {min: 15, max: 3600, change: 0},
})
```
image

I see duplication of the reporting calls (but with the correct values). Is that expected? 

I think this is more of a question for @Koenkk.

src/devices/inovelli.ts
rohankapoorcom: To clarify @Koenkk, does this happen automatically when calling the `electricityMeter` modern extend?

```
await endpoint.read('haElectricalMeasurement', ['acPowerMultiplier', 'acPowerDivisor']);
```

I'm guessing it is not being applied since I'm seeing a 10x increase in power measurement after this code was removed.

Koenkk: > To clarify @Koenkk, does this happen automatically when calling the electricityMeter modern extend?

Yes  it does: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/modernExtend.ts#L1874

rohankapoorcom: Hmm that looks like it should be working... 

Any idea why the power is 10x higher than before these changes? I'm guessing I missed something.

Koenkk: Could it be that device reports the wrong divisor? You should see what it responds in the configure in the debug log


Koenkk: Looks good, it reports the value here:

```
[2024-12-03 11:04:13] debug: 	z2m: Received Zigbee message from 'Kitchen Lights Load Switch', type 'readResponse', cluster 'haElectricalMeasurement', data '{"acPowerDivisor":10,"acPowerMultiplier":1}' from endpoint 1 with groupID 0
```

rohankapoorcom: I swear it wasn't working when I originally posted this comment. Just tested it again and now it's working fine...

Koenkk:
Is this now ready for merge? (CC: @InovelliUSA )

src/devices/inovelli.ts
InovelliUSA: Oh yeah, that isn't the question I thought it was and would be better answered by @Koenkk lol.

InovelliUSA:
> Is this now ready for merge? (CC: @InovelliUSA )

Other than the question mentioned #8402 , I am ok with this merge.

rohankapoorcom:
@Koenkk are you good to merge this? I'd like to open the next PR (starting from this one) adding the new device (VZM30-SN).

Thanks!

Koenkk:
@rohankapoorcom yes, could you rebase it?

rohankapoorcom:
Rebased @Koenkk

Koenkk:
Thanks!

PR #8408, opened 2024-12-01
Poll interval battery fix for Smarthings Multipurpose Sensor 2018

torandreroland:
Fix for battery drain on Smarthings Multipurpose Sensor 2018 in line with [this change in Deconz](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/4163).


Koenkk:
Thanks!

PR #8409, opened 2024-12-01
Fix Philips Hue gradient color encoding.

LukeT92:
Color xy coordinates that encoded with leading zeroes were not padded correctly, leading to malformed commands that devices ignore.


Koenkk:
Thanks!

PR #8419, opened 2024-12-02
Fix bug 'HHST001'

sdbldn:
Fix the temperature compensation issue of 'HHST001' and add blind spot temperature function.


Koenkk:
Thanks!

PR #8424, opened 2024-12-02
Added device Dowsing & Reynolds

ahmet8282:


Koenkk:
Please also add the file to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts

Koenkk:
Thanks!

PR #8425, opened 2024-12-02
add support for TS0505

Hykario:
Hello,
this will add support for TS0505 

I Got another Question. This GU10 is supposed to be Tuya but it works with modern extends.
So i was wondering if the TS0505B and other versions of this also works without the Tuya convert.
I got one at Home and will try later.


Koenkk:
Does this device not work out of the box with z2m 1.42.0?

Koenkk:
I see, could you add this to `tuya.ts` instead and use `Tuya` as the vendor.

Hykario:
okey added it to tuya as tuya device. i hope it is correct if not let me know what to change

Koenkk:
Thanks!

Make sure that:
- The filename is `TS0505.png`
- The size is 512x512

fredericgermain:
Thanks for the PR!
I just wanted to add that the B22 here is also a TS0505
https://www.aliexpress.com/item/1005007212350943.html

PR #8427, opened 2024-12-03
fix: add SR-ZG9002KR12-Pro SR-ZG9002KR12-Pro configure and use extend instead only fromZigbee

niracler:
I apologize, but I formatted the README.md file using npm run pretty:write. However, I can't commit my code without reformatting it again.


Koenkk:
Thanks!

PR #8429, opened 2024-12-03
Add innr filament bulbs RF-271-T/RF-273-T

PickingUpPieces:
Add the new Innr filament bulbs RF 271 T and RF 273 T.
I've tested both, and they worked perfectly out of the box with the generated external definition — awesome work! 🚀
The only manual adjustment I had to make was adding the `turnsOffAtBrightness1` property.

There is a third model, RF 274 T, which I assume has a different bulb shape but should function the same as the other two. Unfortunately, I don’t own this type, so I can’t test this assumption. Should I add the model RF 274 T anyway?

Different filament bulb types: https://www.innr.com/wp-content/uploads/2024/06/innr-SAR-002-2002-UK-DoC-RF-262-271-T-273-T-274-T-v1.0.pdf

Documentation PR: https://github.com/Koenkk/zigbee2mqtt.io/pull/3286

### DIFF
--- a/src/devices/innr.ts
+++ b/src/devices/innr.ts
@@ -815,6 +815,13 @@ const definitions: DefinitionWithExtend[] = [
         description: 'Smart E26 LED bulb',
         extend: [light({turnsOffAtBrightness1: true})],
     },
+    {
+        zigbeeModel: ['RF 271 T'],
+        model: 'RF 271 T',
+        vendor: 'Innr',
+        description: 'Smart E27 filament LED globe light bulb',
+        extend: [light({colorTemp: {range: [153, 556]}, turnsOffAtBrightness1: true})],
+    },
 ];
 
 export default definitions;
--- a/src/devices/innr.ts
+++ b/src/devices/innr.ts
@@ -822,6 +822,13 @@ const definitions: DefinitionWithExtend[] = [
         description: 'Smart E27 filament LED globe light bulb',
         extend: [light({colorTemp: {range: [153, 556]}, turnsOffAtBrightness1: true})],
     },
+    {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Pferdebockwurst:
Hey @PickingUpPieces, sorry to hijack this commit. I am very interested in buying it these lamps. What are your experiences with them so far? Radio stability? Color impressions? I have pretty similar lamps from another vendor, which unfortunately lose connection to the zigbee network after a couple of days and need a power cycle for reconnection. I hope that's not the case with these lamps from innr? Best regards.

PickingUpPieces:
Hi @Pferdebockwurst,
Disclaimer: I haven’t used a filament bulb from another vendor, so I don’t have a direct comparison. 
That being said, I’ve been using these bulbs for about a month now, and so far, I’m really happy with their performance.

- After a power cut, they reconnect instantly and are available in Home Assistant without any issues. Over the last month, I’ve had no connection problems at all.
- They are perfect for use with adaptive lighting throughout the day with their color range from 1800K-6500K.
- The bulb can be dimmed to a very low brightness, which I find great for the evenings.
- One con to mention is the "two LED strips" inside the bulb—one for warm white and one for cool white light. At low brightness, you can see the two separate strips when looking directly at the bulb, but they blend well at higher brightness levels.
- I like the color impressions, especially the warm white color is very moody . But again, I don't have a comparison or high standards^^

Overall, I can definitely recommend these new filament bulbs. I bought the globe version during Black Friday at Kaufland, but unfortunately, the deal is no longer available.
Hope that helps! :)

PR #8435, opened 2024-12-04
Fix bug in tuyaModernExtend

mattbyte:
Bug introduced in #8190 and prevents `tuyaModernExtend` functioning when no endpoint is specified. Low impact as almost no converters use this yet.

When endpoint is undefined, array is still truthy. Correction to ensure value is falsy when zigbee2mqtt is matching the converter.

I now have a converter working for a garage door opener if `tuyaModernExtend` is ready to be used more widely?

### DIFF
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -1808,7 +1808,7 @@ export function getHandlersForDP(
         : [
               {
                   key: [name],
-                  endpoints: [endpoint],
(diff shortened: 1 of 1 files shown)


Koenkk:
> I now have a converter working for a garage door opener if tuyaModernExtend is ready to be used more widely?

tuyaModernExtend can be used, feel free to create a PR. Thanks for the fix!

PR #8439, opened 2024-12-04
Fix MAZDA TR-M2Z Get system mode with preset enabled

Nopraz:
In order to find the real system mode for home assistant, I'm using the preset enabled, when the preset is not off then I return 'heat' otherwise I return 'off'


Koenkk:
Thanks!

PR #8440, opened 2024-12-04
fix: Amina S boolean values

bmork:
Quoting causes non-working home assistant sensors, expecting string payloads instead of booleans.


Koenkk:
Thanks!

PR #8443, opened 2024-12-05
Add support for TRV603-WZ

Cosik:
Hi, I got some new TRV from Aliexpress. I added all data points from TUYA server but not exposed all of them, I'm not sure if it should be done like this.


Koenkk:
Thanks!

PR #8446, opened 2024-12-05
Update tuya.ts

Danieldiazi:
The if-else block has the same actions in both branches.

Into else branch must be: "
 return {auto: tuya.enum(1), manual: tuya.enum(0), temporary_manual: tuya.enum(2)};"


domome:
Hi
Now it works correctly for my device _TZE204_lzriup1j , thank you.

Koenkk:
Thanks!

PR #8448, opened 2024-12-06
Improvement to _tze204_nklqjk62

mattbyte:
Based on interrogation of the Tuya datapoints and testing, this PR exposes and more clearly documents the features of the Tuya PJ-ZGD01 garage door opener and also updates the implementation to use tuyaModernExtends. Care has been taken to avoid modifying the naming/behavior of the two existing exposed entities to avoid breaking changes.

Unsure on the purpose of `await reporting.bind(endpoint, coordinatorEndpoint, ['genBasic']);` in the original implementation as it seems to behave the same without it, including reporting the status of the contact roughly every minute and joining the network without issue.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -7359,7 +7359,8 @@ const definitions: DefinitionWithExtend[] = [
                 type: tuya.dataTypes.bool,
                 valueOn: [true, true],
                 valueOff: [false, false],
-                description: 'Request door to close (= false) or open (= true), will not pulse output if contact shows door is already in requested state',
+                description:
+                    'Request door to close (= false) or open (= true), will not pulse output if contact shows door is already in requested state',
             }),
             dpNumeric({
                 name: 'countdown',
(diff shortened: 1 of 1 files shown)


mattbyte:
For anyone using this with HA, adding this to your HA `configuration.yaml` will show the door as a cover to allow CarPlay integration via HomeKit. Assumes friendly name and HA entity name updated to "Garage Door" in Z2M.

```yaml
cover:
  - platform: template
    covers:
      garage_door:
        device_class: garage
        friendly_name: "Garage Door"
        unique_id: garage_door
        value_template: >-
          {% if is_state('binary_sensor.garage_door_garage_door_contact','off') %}closed
          {% else %}open
          {% endif %}
        open_cover:
          - condition: state
            entity_id: binary_sensor.garage_door_garage_door_contact
            state: "off"
          - action: switch.turn_on
            target:
              entity_id: switch.garage_door_trigger
        close_cover:
          - condition: state
            entity_id: binary_sensor.garage_door_garage_door_contact
            state: "on"
          - action: switch.turn_off
            target:
              entity_id: switch.garage_door_trigger
        icon_template: >-
          {% if states('binary_sensor.garage_door_garage_door_contact') == "on" %}
            mdi:garage-open-variant
          {% else %}
            mdi:garage-variant
          {% endif %}
```

Koenkk:
Thanks!

PR #8450, opened 2024-12-07
add Livarno Home HG08130B

RubenKelevra:
- Add the HG08130B E14 candle lamp, which are currently sold by Lidl, to the self reported Vendor strings
- Fixes Livarno Home, being called Livarno Lux. Which is, I think, either an older branding or the branding for non-smart home lamps, as it's neither printed on the lamps, the packaging or shown on the Website in relation to those lamps.


RubenKelevra:
By chance the device entry was used by a test, so I've updated the test as well.

Koenkk:
Thanks!

PR #8454, opened 2024-12-07
fix description of Livarno Home HG06104A

RubenKelevra:
- It's 2 meters, not 2.5 meters
- The branding is "Livarno Home", not "Livarno Lux"
- Added that it's an "RGBCCT" LED strip

Packaging:

IMG_20241206_170157

This supersedes https://github.com/Koenkk/zigbee2mqtt.io/pull/3301


RubenKelevra:
@Koenkk how about a PR to fix all those "Livarno Lux" entries and replace them with "Livarno Home", as "Livarno Lux" is the brand for non-smart lighting, and "Livarno Home" is the brand for smart lighting.

Koenkk:
Thanks!

PR #8457, opened 2024-12-08
Add support for Danfoss Icon Basic Main Controller

tomsommer:
Danfoss Icon Basic Main Controller (088U1142) has modelID 0x0200

### DIFF
--- a/src/devices/danfoss.ts
+++ b/src/devices/danfoss.ts
@@ -342,6 +342,7 @@ const definitions: DefinitionWithExtend[] = [
     },
     {
         fingerprint: [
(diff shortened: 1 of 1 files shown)


nagydavid:
I had the same, it means , that your Danfoss Icon needs to be restarted. unplugged and replugged, and also redo the connectivity test. After it should work

nagydavid:
Well, I am just telling my 2 cents, it happened with me, and this is how I resolved it. I have never encountered otherwise.

Koenkk:
Thanks!

Issue #8395, opened 2024-11-29
electricityMeter can only use the default settings


script0803:
extend: [identify(), onOff({"powerOnBehavior":false}), electricityMeter({
        threePhase: true,
        producedEnergy: true,
        acFrequency: true,
        powerFactor: true,
        })],

I used this code and applied it to my external converter, but none of these extra settings worked

Koenkk:
Are you using the latest z2m?

script0803:
Oh, I just updated my z2m and it works now, thank you very much.

In addition, I have another problem. I used
 
await endpoint.saveClusterAttributeKeyValue('haElectricalMeasurement', {
acPowerMultiplier: 1, acPowerDivisor: 1}); 

in the converter to change my power unit from kw to w, but it doesn't seem to work now.

Koenkk:
To set the divisor with modern extend, see https://github.com/Koenkk/zigbee-herdsman-converters/blob/7c895668fde39734e0faa61f29bd324e7673837f/src/devices/robb.ts#L102

script0803:
Hi， @Koenkk 

I tried these methods, but none of them worked. It seems that the divisor defined in my device will force overwrite what I set in the converter.  

So can I choose not to change the divisor but change the unit from w to kw?

<img width="330" alt="image" src="https://github.com/user-attachments/assets/254e25af-4823-478e-9a72-8951f03c3f54">

```javascript
const {electricityMeter} = require('zigbee-herdsman-converters/lib/modernExtend');
const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
const reporting = require('zigbee-herdsman-converters/lib/reporting');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const e = exposes.presets;

const definition = {
    zigbeeModel: ['SPM01X001'],
    model: 'SPM01X00111',
    vendor: 'BITUO TECHNIK',
    description: 'Automatically generated definition',
    configure: async (device, coordinatorEndpoint) => {
            const endpoint = device.getEndpoint(1);
            await reporting.bind(endpoint, coordinatorEndpoint, ['haElectricalMeasurement', 'seMetering']);
            await reporting.readMeteringMultiplierDivisor(endpoint);
            await endpoint.saveClusterAttributeKeyValue('haElectricalMeasurement', {
            acPowerMultiplier: 1, acPowerDivisor: 1,
        });
    },
[...]

PR #8463, opened 2024-12-10
Fix bug about power unit, add three new devices

script0803:
A converter is defined to change the factor of the power class parameter to convert the unit from kw to w.

### DIFF
--- a/src/devices/candeo.ts
+++ b/src/devices/candeo.ts
@@ -28,8 +28,15 @@ const definitions: DefinitionWithExtend[] = [
         extend: [light({configureReporting: true})],
     },
     {
-        zigbeeModel: ['C204', 'C-ZB-DM204'],
+        zigbeeModel: ['C204'],
         model: 'C204',
+        vendor: 'Candeo',
+        description: 'Zigbee micro smart dimmer',
+        extend: [light({configureReporting: true}), electricityMeter()],
+    },
+    {
+        zigbeeModel: ['C-ZB-DM204'],
+        model: 'C-ZB-DM204',
         vendor: 'Candeo',
         description: 'Zigbee micro smart dimmer',
         extend: [light({configureReporting: true}), electricityMeter()],


src/devices/bituo_technik.ts
```diff
@@ -1,84 +1,321 @@
...
-import {DefinitionWithExtend} from '../lib/types';
+import {DefinitionWithExtend, Fz, KeyValueAny} from '../lib/types';
+import * as utils from '../lib/utils';
+import {postfixWithEndpointName, precisionRound} from '../lib/utils';
 
 const e = exposes.presets;
+const ea = exposes.access;
 
+const bituo_fz = {
+    electrical_measurement: {
```
Koenkk, line 13: this is a lot of duplication, can't we add a new parameter the the electricityMeter modern extend for this?

script0803: As I mentioned in this [discussion](https://github.com/Koenkk/zigbee-herdsman-converters/issues/8395), adjusting the parameters of power in extend modern did not work for me, and if I change the electricityMeter modern extend, maybe 2 will affect others, and creating a converter in our brand's device file is the safest option. We will have more products that need to use this converter in the future, so I think it is worth it.

script0803: @Koenkk 
Hi, any news on the pull request? Let me know if there's anything else needed. Thanks!

Koenkk: > I change the electricityMeter modern extend, it might affect others.

This can be avoided by adding a parameters, this just duplicates too much code.

script0803: But I don't know how to add a new parameter for modern extend...

script0803: > > I change the electricityMeter modern extend, it might affect others.
> 
> This can be avoided by adding a parameters, this just duplicates too much code.

I checked the code of modern extend, divisor and mutiplier are just indexes, you still use electrical measurement as your converter, so I still need to rewrite the logic of electrical measurement, otherwise it will be the same as the result of using forced now, which does not work for my device. 

In addition, your modern extend does not support total power, I added this in my converter, so I don't think this is "useless duplication of code"

script0803: > > I change the electricityMeter modern extend, it might affect others.
> 
> This can be avoided by adding a parameters, this just duplicates too much code.

If you insist on asking me to change the modern extend instead of rewriting a "duplicate code" converter, please give me some help, because this problem is very tricky for this device, thank you very much

Koenkk: > which does not work for my device.

could you explain what exactly does not work?

> In addition, your modern extend does not support total power, 

I think that can be added to the original modernExtend?

script0803: > > which does not work for my device.
> 
> could you explain what exactly does not work?
> 
extend: [
electricityMeter({
fzMetering: fz.metering,
fzElectricalMeasurement: bituo_fz.electrical_measurement,
acFrequency: true,
powerFactor: true,
configureReporting: false,
producedEnergy: true,
power: {divisor: 1, multiplier: 1},
}),
], 
the power: {divisor: 1, multiplier: 1} in it does not work for my device, it cannot change my divisor and multiplier

Koenkk: I see, in that case we should debug why this doesn't work (as it should). Could you try to add a `console.log()` above https://github.com/Koenkk/zigbee-herdsman-converters/blob/ead17e18f26e824305face85c8b5338de11a7937/src/lib/modernExtend.ts#L1915

and see if that is logged?

script0803: > I see, in that case we should debug why this doesn't work (as it should). Could you try to add a `console.log()` above
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/ead17e18f26e824305face85c8b5338de11a7937/src/lib/modernExtend.ts#L1915
> 
> and see if that is logged?

I think the cause of this problem has been found in this [discussion](https://github.com/Koenkk/zigbee-herdsman-converters/issues/8395)

Also, could you please tell me how to modify my own modern extend files for zigbee2mqtt? I installed HomeAssistant on linux, and I have never tried to modify these files before. Thanks a lot.

script0803: > I think what we can do is reset the divisor/multiplier immediately after receiving the attribute report through an `onEvent` handler, see

Thank you very much! I will try it.

PR #8466, opened 2024-12-10
Fix Aqara H1M 3-gang not supported power_outage_memory and led_disabled_night

cralex96:
The Aqara H1M 3-gang (ZNQBKG26LM) model is included in the validations to perform power_outage_memory and led_disabled_night actions. The 2-gang model (ZNQBKG25LM) was already included, so the 3-gang should be included in the same way


Koenkk:
Thanks!

PR #8467, opened 2024-12-10
Minor SNZB-03P documentation fixe

Ten0:
That's actually not a delay for occupancy. When you enter the room it immediately says you enter the room. However there's a no-movement timeout before it says "unoccupied".

Tied to https://github.com/Koenkk/zigbee2mqtt.io/pull/3317.


Koenkk:
Thanks!

PR #8468, opened 2024-12-11
Add support for IKEA LED2111G6 TRADFRI bulb E17 CWS globe 810lm version

annadorottya:
This is the version sold in Japan.


Koenkk:
Thanks!

PR #8469, opened 2024-12-11
Add support for Inovelli VZM30-SN switch

rohankapoorcom:
CC: @InovelliUSA 

This switch is currently in beta. Adding the converter here to make it easy for beta testing now that the converter is in good shape.

Image for the docs is here: https://github.com/Koenkk/zigbee2mqtt.io/pull/3320

### DIFF
--- a/src/devices/inovelli.ts
+++ b/src/devices/inovelli.ts
@@ -3,7 +3,7 @@ import {Zcl} from 'zigbee-herdsman';
 import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as exposes from '../lib/exposes';
-import {deviceAddCustomCluster, deviceEndpoints, electricityMeter, identify} from '../lib/modernExtend';
+import {deviceAddCustomCluster, deviceEndpoints, electricityMeter, humidity, identify, temperature} from '../lib/modernExtend';
 import * as reporting from '../lib/reporting';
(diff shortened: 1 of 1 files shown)


src/devices/inovelli.ts
```diff
@@ -478,6 +478,8 @@ const COMMON_ATTRIBUTES: {[s: string]: Attribute} = {
         max: 127,
         readOnly: true,
         description: 'The temperature measured by the temperature sensor inside the chip, in degrees Celsius',
+        category: 'diagnostic',
+        unit: '°C',
```
rohankapoorcom, line 482: @Koenkk - I just realized that on all of the Inovelli devices, the internalTemperature parameter doesn't show up as the `device_class: temperature` when MQTT sends it to Home Assistant. I thought adding the unit and category would fix that, but it doesn't seem to be doing so. Is there something else that needs to be set?

Koenkk: For that it needs to be added here; https://github.com/Koenkk/zigbee2mqtt/blob/bc6eccf6152c4cd91e75309f7ea0b8ad4e8835d4/lib/extension/homeassistant.ts#L294

rohankapoorcom: Cool - since it's a separate repo, we can handle it separately after this is merged if that's okay with you.

Koenkk:
Thanks!

PR #8471, opened 2024-12-12
feat: Load converters on-demand

Nerivec:
_This is intended for post-2.0.0.
Will keep all conflict-prone changes to a minimum until ready to execute all scripts to update the entire repo._

## Changes

<details><summary>Logic before</summary>

- Load all definitions into memory on start
  - import all src/devices/* modules
  - process all definitions (process extensions, validate)
  - create global lookup (mapping `zigbeeModel` to array of definitions)
  - store definitions in global object
- Resolve definition:
  - for already joined devices on start _if needed_
  - on joining device _if needed_
  - on interviewing device _if needed_
  - on message from device _if needed_
  - on re-interview

#### Definition resolving

- find matching definitions in lookup by `zigbeeModel`
- find matching definition
  - if no match, and allowing to generate, generate external definition
  - if single match, return match
  - if multiple match, try identify by fingerprint, else return first match by zigbeeModel
- cache definition in Z2M `Device`

</details>
<details><summary>Logic after</summary>

- Load pre-built index (mapping `zigbeeModel` to array of [`module`, `definition index`]) into memory on start (index is JSON built at compile time)
- Resolve definition:
  - for already joined devices on start _if needed_
  - on interviewing device _if needed_, _when got modelID_
  - on message from device _if needed_
  - on re-interview

#### Definition resolving

- find matching indexes in pre-built index by `zigbeeModel`
- import matched src/devices/* modules (cached)
- get matched definitions by `index`
- find matching indexed definition
  - if no match, and allowing to generate, generate external definition
  - if single match, return match
  - if multiple match, try identify by fingerprint, else return first match by zigbeeModel
- process matching definition (process extensions, validate)
- cache definition in Z2M `Device`
- [after "resolving for all devices"] purge cached imported definitions modules (happens on start, or after an external converter changes)

</details>

### Other changes

- Export `src/devices/*` using named exports
- Export some more APIs from ZHC
- Rename some ZHC APIs to match new logic
  - Lookup now only used for external converters (skipped entirely if no external converter present)
- Rename some symbols in `generateDefinition.ts` to avoid clash with globals.

### TODO:

- [x] New tests
- [x] Execute scripts to update repo to new logic
- [x] [external] Update API usage in Z2M https://github.com/Koenkk/zigbee2mqtt/pull/26403
- [ ] [external] Review usage of `resolveDefinition` in Z2M (optimize as needed)
- [x] [external] Update API usage in Z2M docs https://github.com/Koenkk/zigbee2mqtt.io/pull/3527
- [x] [post] Remove no longer useful scripts

## Benchmarks

This being a deep change in the architecture, some in-depth testing was required. Since this is mostly about memory, tests mostly focus on memory...
_Tested using Zigbee2MQTT 2.0.0 (for both "Before" & "After")._
_I tried to minimize the impact of other systems while testing, but allow for some margin of error still..._

<details><summary>VSCode Profiler</summary>

With brand new network (already established on coordinator, no device).
Cold start.

| Item | Before | After | Diff |
| ---- | ------ | ----- | ----- |
| CPU time total (non-idle) | 1948 ms | 1242 ms | 64% |

Heap snapshot once settled after start (limited to items above 1 MB in "Before"):

| Item | Before | After | Diff |
| ---- | ------ | ----- | ----- |
| global / | 35.6 MB | 13.6 MB | 38% :rocket: |
| Object | 33.7 MB | 21.9 MB | 64% |
| (compiled code) | 25.6 MB | 18.8 MB | 73% |
| (string) | 18.1 MB | 11.5 MB | 63% |
| (closure) | 17.1 MB | 13.7 MB | 80% |
| Array | 16.1 MB | 2.44 MB | 15% :rocket: |
| (array) | 11.5 MB | 5.36 MB | 46% :rocket: |
| system / Context | 8.22 MB | 4.73 MB | 57% |
| Buffer | 6.42 MB | 6.43 MB | - |
| Numeric | 4.76 MB | 0 | - :rocket: |
| (unknown) | 3.35 MB | 2.72 MB | 81% |
| Light | 2.73 MB | 0 | - :rocket: |
| (concatenated string) | 2.57 MB | 903 KB | 34% :rocket: |
| Map | 1.69 MB | 815 KB | 47% :rocket: |
| Module | 1.65 MB | 1.31 MB | 79% |
| (system) | 1.54 MB | 1.23 MB | 79% |
| Enum | 1.42 MB | 1.5 KB | - :rocket: |
| --- | --- | --- | --- |
| TOTAL | 192 MB | 105 MB | 54% |

</details>

<details><summary>Clinic.js</summary>

With brand new network (already established on coordinator).
Cold start and pairing of 1 device.
Red=Before
Blue=After

mem-before
mem-after

<details><summary>Even Loop use</summary>

even-loop-use-before

even-loop-use-after

</details>

<details><summary>Event Loop delay</summary>

event-loop-delay-before
event-loop-delay-after

</details>

</details>

Overall, I'd say we can expect memory usage to be around 60% of what it was before. This is reflected throughout the application's runtime, and should remain pretty stable, with small spikes only expected around the addition of new devices (i.e. need to load some converters).


Nerivec:
@sjorge @kirovilya if you have some time, and can review the changes, mostly just the core logic changes in `src/index.ts`, the rest is utility. Would be good to have some feedback, tweak the approach as necessary.

src/index.ts
```diff
@@ -6,35 +6,41 @@ import {Zcl} from 'zigbee-herdsman';
...
     OnEventData,
     OnEventMeta,
     OnEventType,
     Option,
+    ResolvedDefinition,
     Tz,
     Zh,
 } from './lib/types';
 import * as utils from './lib/utils';
+import modelsIndexJson from './models-index.json';
```
sjorge, line 33: huh, neat trick I didn't know about, I always JSON.parse(fs.readFileSync(...

src/index.ts
```diff
@@ -81,43 +88,133 @@ function arrayEquals<T>(as: T[], bs: T[]): boolean {
...
+}
+
+export function addDefinition(definition: IndexedDefinition): void {
+    externalDefinitions.splice(0, 0, definition);
+
+    if (definition.externalConverterName) {
+        externalDefinitionsCount++;
+    }
+
+    if ('fingerprint' in definition && definition.fingerprint) {
```
sjorge, line 164:
```suggestion
    if (definition.fingerprint) {
```
Should be fine too, definition doesn't have a fingerprint property it will eval to false.

src/index.ts
```diff
@@ -378,109 +477,66 @@ export function postProcessConvertedFromZigbeeMessage(definition: Definition, pa
...
-
-        if (match) {
-            definition = {
-                ...definition,
-                model: match.model,
-                vendor: match.vendor,
-                description: match.description || definition.description,
-            };
+    if (definition) {
+        if (definition.whiteLabel) {
```
sjorge, line 490:
```suggestion
        if (definition?.whiteLabel) {
```

nit: can be folder to one if level

Koenkk:
@kirovilya could you also take a look at this?

src/index.ts
Nerivec: Thanks to this: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/tsconfig.json#L16 😉

https://www.typescriptlang.org/tsconfig/#resolveJsonModule

src/index.ts
Nerivec: I missed those, too much copy/pasting... 😝

src/index.ts
Nerivec: The return is outside the inner if, so it's needed (else would need a ternary on return, this way undefined is inferred).

src/index.ts
Nerivec: Fixed the ones that aren't used for narrowing 😉

Nerivec:
@sjorge Thanks for the feedback 👍
1. The new external converters in 2.0.0 already have a [prop](https://github.com/Koenkk/zigbee-herdsman-converters/blob/d78ce1da940eab7a17416e6d2eb4e508b8afb8dc/src/lib/types.ts#L276) that identifies them. It would need to be added to the [published payload](https://github.com/Koenkk/zigbee2mqtt/blob/5e71cf2cc84cd6c5c1c24d801a1f2dadee4e7068/lib/extension/bridge.ts#L800), then I guess it's just a matter of flipping a switch somewhere in frontend 😉
2. Probably good indeed. Though not in this PR to keep it only about the base logic change (it will already be huge once scripts are applied).

src/index.ts
```diff
@@ -96,43 +103,133 @@ function arrayEquals<T>(as: T[], bs: T[]): boolean {
...
+    }
+
+    if (definition.zigbeeModel) {
+        for (const zigbeeModel of definition.zigbeeModel) {
+            addToExternalConvertersLookup(zigbeeModel, definition);
+        }
+    }
+}
+
+let loadedModules: Record<string, IndexedDefinition[]> = {};
```
Nerivec, line 192: Note to self: remove this caching mechanism in favor of node's own module caching.

Nerivec:
I reverted the lambda part, it's not very useful for perf as it turns out, not worth the extra logic (and the breaking change that would need special handling until next major). It will also make this PR a lot easier to handle.

I moved the dupe checking from tests to the index builder script. It also seems the tests were not catching everything. @Koenkk I'll let you double-check the changes are sound, all in one commit: https://github.com/Koenkk/zigbee-herdsman-converters/pull/8471/commits/2c51dc94afe3cb008c5a12799a15c57e81877a6d 

@Koenkk can you check the remaining tests in `index.test.js`, I'm thinking some of that stuff was for JS code, TS typing should handle some (most?) of it without needing tests.

package.json
```diff
@@ -34,10 +34,11 @@
...
         "test:coverage": "vitest run --config ./test/vitest.config.mts",
         "test:watch": "vitest watch --config ./test/vitest.config.mts",
-        "clean": "rimraf --glob index* devices lib converters tsconfig.tsbuildinfo",
-        "build": "rimraf --glob index* devices lib converters && tsc && cp -R dist/* .",
-        "build-watch": "tsc --watch",
-        "prepack": "pnpm run clean && pnpm run build",
+        "clean": "rimraf --glob index* coverage dist devices lib converters tsconfig.tsbuildinfo",
+        "build": "rimraf --glob index* coverage dist devices lib converters && tsc && mv dist/* .",
+        "build:watch": "tsc --watch",
+        "build-index": "node build-models-index.mjs",
```
Koenkk, line 40:
```suggestion
        "build:index": "node build-models-index.mjs",
```

I'm also wondering wether it should be part of the `build` command

Nerivec: I went with a `-` because it isn't really a subset of `build` like the other scripts. But yea, I think we might just integrate it directly in `build` command. I needed more control when I ran the initial tests, but now it can be part of the regular `build`.

Koenkk:
Are the benchmarks up-to-date even after removing the `resolve()`?

build-models-index.mjs
```diff
@@ -0,0 +1,105 @@
...
+    for (let i = 0; i < definitions.length; i++) {
+        const definition = definitions[i];
+
+        if (addedModels.includes(definition.model.toLowerCase())) {
+            throw new Error(`Duplicate model ${definition.model}`);
+        }
+
+        addedModels.push(definition.model.toLowerCase());
+
+        if (definition.whiteLabel) {
```
Koenkk, line 58: I don't think whiteLabels have to be added to the index, the `fingerprint`/`zigbeeModel` always has to match before it will even be matched against a whiteLabel. (i.e. the whiteLabel fingerprint is always a superset of the definition `fingerprint`/`zigbeeModel`)

Nerivec: They are not added. This is the de-dupe logic.

build-models-index.mjs
```diff
@@ -0,0 +1,105 @@
...
+        if (definition.fingerprint) {
+            for (const fingerprint of definition.fingerprint) {
+                for (const addedFingerprint of addedFingerprints) {
+                    if (equal(addedFingerprint, fingerprint)) {
+                        throw new Error(`Duplicate fingerprint for ${definition.model}: ${JSON.stringify(fingerprint)}`);
+                    }
+                }
+
+                // type `modelID?: string;` means modelID can be undefined, ends up under index "null"
+                addToLookup(fingerprint.modelID, [module, i]);
```
Koenkk, line 81: The `TS0601` modelID is heavily overloaded, most Tuya device require both modelID + manufName for matching, I'm wondering if we could add that also to the lookup? e.g. `TS0601__TZE200_bq5c8xfe`

models-index.json
```diff
@@ -0,0 +1 @@
...[["yokis.js",10]],"tlc4-up":[["yokis.js",11]],"tlc8-up":[["yokis.js",12]],"tlm1-up":[["yokis.js",13]],"tlm503-up":[["yokis.js",13]],"tlm2-up":[["yokis.js",14]],"tlm2_503-up":[["yokis.js",14]],"tlm4-up":[["yokis.js",15]],"tlm4_503-up":[["yokis.js",15]],"galet4-up":[["yokis.js",16]],"d10110":[["yookee.js",0]],"zb-cl03":[["ysrsai.js",0]],"fb56-zcw20fb1.2":[["ysrsai.js",0]],"zb-ct01":[["ysrsai.js",1]],"zb-dl01":[["ysrsai.js",2]],"nuet56-dl27lx1.1":[["zemismart.js",1]],"lxt56-ls27lx1.6":[["zemismart.js",2]],"ts0302":[["zemismart.js",3]],"lxn56-ss27lx1.1":[["zemismart.js",5]],"zen-01":[["zen.js",0]],"lywsd03mmc-z":[["zigbeetlc.js",0]],"lywsd03mmc-bz":[["zigbeetlc.js",0]],"mho-c122-z":[["zigbeetlc.js",0]],"mho-c122-bz":[["zigbeetlc.js",0]],"mho-c401-z":[["zigbeetlc.js",0]],"mho-c401-bz":[["zigbeetlc.js",0]],"mho-c401n-bz":[["zigbeetlc.js",0]],"cgdk2-z":[["zigbeetlc.js",1]],"cgdk2-bz":[["zigbeetlc.js",1]],"ts0201-z":[["zigbeetlc.js",2]],"ts0201-bz":[["zigbeetlc.js",2]],"th03z-z":[["zigbeetlc.js",2]],"th03z-bz":[["zigbeetlc.js",2]],"zth01-z":[["zigbeetlc.js",2]],"zth01-bz":[["zigbeetlc.js",2]],"zth02-z":[["zigbeetlc.js",2]],"zth02-bz":[["zigbeetlc.js",2]],"zha-colorlight":[["zipato.js",0]]}
```
Koenkk, line 1: Can't we `.gitignore` this file?

Nerivec: It would break TS (missing file), and tests (would require building index locally before).

build-models-index.mjs
Nerivec: The lookup index is (currently) only a way to pinpoint a converter in a particular file.
This would introduce quite a bit more logic (detecting if manuf name present, if should consider it, etc.).

Do you see an efficient way to do this and not create problems with other converters?

models-index.json
Koenkk: > would require building index locally before

That's fine right? You have to build the index before a `tsc`

build-models-index.mjs
Koenkk: True

> Do you see an efficient way to do this and not create problems with other converters?

Reversing the question, would entries like `TS0601` with many definitions spread over many files give any problems?

src/index.ts
```diff
@@ -72,12 +78,13 @@ export {
...
 };
 export * as ota from './lib/ota';
+export {setLogger} from './lib/logger';
 
 export const getConfigureKey = configureKey.getConfigureKey;
 
 // key: zigbeeModel, value: array of definitions (most of the times 1)
-const lookup = new Map<string, Definition[]>();
-export const definitions: Definition[] = [];
+const externalConvertersLookup = new Map<string, DefinitionWithExtend[]>();
```
Koenkk, line 86:
```suggestion
const externalDefinitionsLookup = new Map<string, DefinitionWithExtend[]>();
```

src/index.ts
```diff
@@ -96,43 +103,128 @@ function arrayEquals<T>(as: T[], bs: T[]): boolean {
...
+    // local cache for models with lots of matches (tuya...)
+    const defs: Record<string, DefinitionWithExtend[]> = {};
+
+    for (const [module, index] of indexes) {
+        if (!defs[module]) {
+            logger.debug(`Loading module ${module}`, NS);
+
+            // NOTE: modules are cached by nodejs until process is stopped
+            // currently using `commonjs`, so strip `.js` file extension, XXX: creates a warning with vitest (expects static `.js`)
+            const {definitions} = (await import(`./devices/${module.slice(0, -3)}`)) as {definitions: DefinitionWithExtend[]};
```
Koenkk, line 200: Should we have a function to also clear loaded modules which can e.g. be called after z2m loaded all definitions? It seems to be possible with [`delete require.cache[moduleName]`](https://medium.com/@interviewer.live/understanding-module-caching-in-node-js-977dab968541#:~:text=of%20your%20application.-,Node.,to%20be%20reloaded%20if%20required.)

models-index.json
Nerivec: Vitest builds its own JS internally, the file will be missing in that case (unless explicitly building first to have the json file in place for vitest when it does its own build).
It will also cause permanent lint failure in editors due to the missing file (which may cause further issues with some editor-specific work flows).

It's annoying to have to have both because of the fact it requires a commit before a release, to sync-up, but I can't see a good way that doesn't mess with devs & editors.

_Note: You have to build the index _after_ `tsc`, it uses the JS._

src/lib/types.ts
```diff
@@ -266,14 +266,17 @@ type DefinitionBase = {
     vendor: string;
     description: string;
     whiteLabel?: WhiteLabel[];
+    generated?: true;
+    externalConverterName?: string;
+};
+
+type DefinitionConfig = {
```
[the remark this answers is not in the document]
Nerivec: I left the split from previous commits (resolve) where you have:
- matcher
- metadata
- config
- features

as separate types to build `Definition` & `DefinitionWithExtend`.
It's no longer necessary, but I thought it created a rather nice separation of "business"... It doesn't have any impact since it's only used locally to build the aforementioned types.
We can merge back Matcher, Base & Config if you prefer though.

src/index.ts
Nerivec: This is `require` only. I don't believe there is one for dynamic imports (yet?).

build-models-index.mjs
Nerivec: The index links `file.js` to index in `converters` export of said file, for every matching file, every matching index, so no problem, except for the fact that it needs to load more stuff than usual when you have more files involved than usual... They definitely have more impact on perf than singles, but that hasn't changed from before (only scaled to current perf).

src/lib/types.ts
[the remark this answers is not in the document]
Koenkk: Lets leave it

build-models-index.mjs
Nerivec: Here are some stats for ref (all keys in index with more than 5 entries):

```
ts0505b has 10 entries!
ts0201 has 11 entries!
ts0601 has 271 entries!
ts011f has 39 entries!
tlsr82xx has 6 entries!
null has 9 entries!
ts0003 has 14 entries!
dimmer-switch-zb3.0 has 10 entries!
greenpower_2 has 8 entries!
ts0202 has 11 entries!
ts0001 has 16 entries!
ts004f has 7 entries!
ts0222 has 6 entries!
ts130f has 6 entries!
ts0726 has 8 entries!
ts000f has 9 entries!
ts0002 has 6 entries!
ts110e has 8 entries!

Models index size: 3781
```

sjorge:
Thanks for all the work on this one! It’s a huge one. <3

ireun:
Just a quick report after update :)
Before:
image
After:
image

Zigbee2MQTT LXC container on Proxmox VE 88.3.4

MSL-DA:
I am surprised by the improvement ... 😱 I also see more than 50% reduction

defanator:
@Nerivec seeing around 30% RSS decrease in my (quite ascetic) setups on RPi 3, awesome! Thanks.

(Now, finally, z2m isn't the most greedy service, telegraf win!)

PR #8472, opened 2024-12-12
Add Support for TS0222 Light Vendor

susch19:
* because _TZ3000_l6rsaipj was not yet in the list but is supported with the same exposes

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/25141


Koenkk:
Thanks!

PR #8473, opened 2024-12-12
Fix: Danfoss eTRV - timestatus should be set to 1 when setting time

trexx:
It is the responsibility of the ZigBee coordinator, after writing to the "Time" attribute, to update "Time Status" "synchronized" bit to "1"

This is from the "Common ZigBee Cluster Specification" for the Danfoss eTRV.
https://assets.danfoss.com/documents/284715/AM375549618098en-000103.pdf


Koenkk:
Thanks!

PR #8475, opened 2024-12-13
Add switch scene support for the Miboxer FUT089Z remote

kvakulo:
The command is sent when double clicking the "I" part of the 1-7 buttons.


Koenkk:
Thanks!

PR #8478, opened 2024-12-14
fix(detect): Detect _TZE204_3ejwxpmu as TS0601_co2_sensor

drm87:
Extends fingerprint of TuYa TS0601_co2_sensor (NDIR co2 sensor). 
Testing with external definition was successful:

`const definition = {
    zigbeeModel: ['TS0601'],
    model: 'TS0601',
    vendor: '_TZE204_3ejwxpmu',
    extend: [],
    meta: {},
    fromZigbee: [legacy.fromZigbee.tuya_air_quality], // We will add this later
    toZigbee: [], // Should be empty, unless device can be controlled (e.g. lights, switches).
    exposes: [e.temperature(), e.humidity(), e.co2()],
};`

image

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1312,7 +1312,7 @@ const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #8479, opened 2024-12-14
danfoss: Update time regularly to account for loss and drift

kennylevinsen:
Danfoss eTRVs will regularly lose their time, such as on battery change, after updates, after hardware resets, etc., which leads to wildly inaccurate date and time of day. Even without these events, clock drift may lead to slowly growing clock errors.

This not only breaks scheduling, but also causes the automatic adaption runs which cut off heating to run at inappropriate times such as in the middle of the day when heat is required.

Follow the Danfoss recommendations and set time when the device connects, and at a one week interval.

/cc @filips

### DIFF
--- a/src/devices/danfoss.ts
+++ b/src/devices/danfoss.ts
@@ -5,10 +5,19 @@ import tz from '../converters/toZigbee';
 import * as constants from '../lib/constants';
 import * as exposes from '../lib/exposes';
 import * as reporting from '../lib/reporting';
-import {DefinitionWithExtend} from '../lib/types';
+import * as globalStore from '../lib/store';
+import {DefinitionWithExtend, Zh} from '../lib/types';
 
 const e = exposes.presets;
 const ea = exposes.access;
+
+const setTime = async (device: Zh.Device) => {
+    const endpoint = device.getEndpoint(1);
+    const time = Math.round((new Date().getTime() - constants.OneJanuary2000) / 1000);
+    // Time-master + synchronised
+    const values = {timeStatus: 1, time: time, timeZone: new Date().getTimezoneOffset() * -1 * 60};
+    await endpoint.write('genTime', values);
+};
 
 const definitions: DefinitionWithExtend[] = [
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Nerivec:
From what I've been told about this device, I wonder if it might be worth putting the `start` `setTime` in a `setTimeout` since they apparently are slow to reconnect on start and this results in lots of timeouts.
Something like this:
```typescript
else if (type === 'deviceAnnounce' || type === 'start') {
    if (type === 'start') {
        setTimeout(async () => await setTime(device), 300000);
    } else {
        await setTime(device);
    }

    // rest...
}
```
__Note: first line is just because we shouldn't create an array to check against two values_ 😉)

Though, if the device always `deviceAnnounce` on reconnect, it might not be needed on `start` at all.

CC: @antst

kennylevinsen:
I have 6 Ally TRVs on my network, and I don't have any logged timeouts from them across several z2m restarts. Maybe open an issue with logs and description?

> _Note: first line is just because we shouldn't create an array to check against two values 😉)

That's is just following the existing style, which I don't have an opinion on.

Nerivec:
I'm only relaying some observations from a previous debugging session with antst.
Since yours aren't affected, I guess there is a difference with @antst 
I'll let you guys check, might be a different firmware, or another device that's causing trouble for these in the network?

antst:
> I have 6 Ally TRVs on my network, and I don't have any logged timeouts from them across several z2m restarts. Maybe open an issue with logs and description?
> 
> > _Note: first line is just because we shouldn't create an array to check against two values 😉)
> 
> That's is just following the existing style, which I don't have an opinion on.

I run 16 of them for 2 years in a busy network. They are, in my opinion, the best TRVs (and I tried most of market can offer), and part of it is that they are very good on battery life. And part of being good on battery life is to don't rush with communication.
You can try to put your network down for couple of hours and start again, and see.  Especially will help, if half of them talk to coordinator through couple of routers, not directly  :)

antst:
> I'm only relaying some observations from a previous debugging session with antst. Since yours aren't affected, I guess there is a difference with @antst I'll let you guys check, might be a different firmware, or another device that's causing trouble for these in the network?

In my case it also fly now even with unchanged code. We noticed this issue after my network was down for sometime (due to the flashing and replacing adapters etc), so TRVs went to relaxed mode :)
Once it revitalized, now restart takes not time even with old code.
But very first time things will go wrong, old code will manifest with the same issue :)

kennylevinsen:
Ah well, if you restart z2m while the TRVs are unavailable in an error state where they won't reconnect for an hour, then getting a timeout is expected. You should only get one for the TRV, as the next attempt will be once the device announces or a week later, whichever comes first. You could also have e.g. setpoint automations, load balance, external temp sync attempt to write to the TRVs during this time, which would also time out.

> And part of being good on battery life is to don't rush with communication.

The device is only expected to sleep for a handful of seconds (or as configured in the poll control cluster, which can allow more extreme battery optimizations than what we're doing) in order to stay in spec and have reliable packet delivery. Being down for up to an hour sounds more like weird error handling (exponential backoff on reconnect?).

antst:
> Ah well, if you restart z2m while the TRVs are unavailable in an error state where they won't reconnect for an hour, then getting a timeout is expected. You should only get one for the TRV, as the next attempt will be once the device announces or a week later, whichever comes first. You could also have e.g. setpoint automations, load balance, external temp sync attempt to write to the TRVs during this time, which would also time out.

Yes, it is the case. I actually run load balancing and external temperature sensors :)

> > And part of being good on battery life is to don't rush with communication.
> 
> The device is only expected to sleep for a handful of seconds (or as configured in the poll control cluster, which can allow more extreme battery optimizations than what we're doing) in order to stay in spec and have reliable packet delivery. Being down for up to an hour sounds more like weird error handling (exponential backoff on reconnect?).

At least this is what I see, if I leave mesh without coordinator for quite some time, and then start coordinator back. 
And yes, I think they internally do something like exponential backoff or something else like that, not sure though that current behavior is what they really wanted :)

MSL-DA:
Does this change also take summer/winter time into account?

ussaka:
When setting time Danfoss recommends to write all 6 attributes once every week. This PR doesn't set `DstStart`, `DstEnd` and `DstShift`. Any idea how will the Ally behave without those attributes set? at least when DST starts/stops user could have wrong time for a week until the time is updated.

@kennylevinsen Could DST attributes be added?

kennylevinsen:
If they follow the spec, not setting the DST attributes should just mean that DST is handled manually by changing timezone - the timezone offset written includes the DST component, which needs to be removed if you want to implement the DST attributes.

Feel free to write a patch for it. This PR only focuses on setting the time more often as per the Danfoss documentation to avoid drift and random calibration runs, and did not change *what* we set as time.

PR #8481, opened 2024-12-15
Fix inverted `state` on Tuya motorized blinds

TzeWey:
When "Invert cover" is enabled, the state published when the covers reach the ends are reversed. This is because the inverted position is used to determine the `state` and is then inverted once more.

To resolve this, the `state` is determined using the final position only. The `state` will be published as CLOSE when the position is 0, otherwise OPEN is published.

Verified with the A-OK AM25 that is recognized as B09M3R35GC _TZE200_9p5xmj5r Unable to verify with a motor that is non-inverted.


Koenkk:
Thanks!

PR #8482, opened 2024-12-15
fix(ELKO): Resolve order of magnitude issue with floor_temp

chdefrene:
Fixes the following issues:
* https://github.com/Koenkk/zigbee2mqtt/issues/25022
* https://github.com/Koenkk/zigbee2mqtt/issues/25112


chdefrene:
I have tested this change on my own device, and it keeps the floor temp value as a decimal, while respecting the selected sensor for the local temperature value.

air
floor

Koenkk:
Thanks!

PR #8487, opened 2024-12-17
MAZDA MZV-T1Z Set manual mode by default when turning on instead off schedule mode and remove off preset

Nopraz:
@Koenkk Hello, before you watch this pull request I wanted to ask a question about managing system modes on devices that don't have pure "on" mode.

These TRVs only have presets and off is one of them, the problem is that Home assistant badly reacts when I change the preset to lets say "eco" mode and considers that the TRV is still off despite any other presets than "off" meaning "heat"

How can I manage correctly system_mode for this TRV? Are there any other way?


Koenkk:
I think this can be fixed by also updating the `system_mode` when the `preset` changes.

Nopraz:
@Koenkk thank you for the answer, I did that and now it works correctly on home assistant. I thought that it wasn't needed

src/devices/mazda.ts
```diff
@@ -56,25 +62,55 @@ const definitions: DefinitionWithExtend[] = [
...
+                            return utils.getFromLookup(v, lookup);
+                        },
+                    },
                 ],
                 [
                     2,
                     'system_mode',
                     {
-                        from: (v) => {
+                        from: (v: KeyValue) => {
```
Koenkk, line 100:
```suggestion
                        from: (v: tuya.Enum) => {
```

src/devices/mazda.ts
```diff
@@ -1,9 +1,15 @@
+import {KeyValue} from 'zigbee-herdsman/dist/controller/tstype';
```
Koenkk, line 1: Please use this one instead:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/8f313c3643b07fbba2d7a2e7447293590ee0f7e7/src/lib/types.ts#L17

Koenkk:
Thanks!

PR #8492, opened 2024-12-19
fix: add off system_mode to TYBAC-006

crazyserver:
Changes were purposed on https://github.com/Koenkk/zigbee2mqtt/issues/21647


src/devices/tuya.ts
```diff
@@ -11353,7 +11353,46 @@ const definitions: DefinitionWithExtend[] = [
...
+                    2,
+                    null,
+                    {
+                        // Map system_mode back to both 'state' and 'system_mode'
+                        from: (v: number) => {
+                            const modes = ['cool', 'heat', 'fan_only'];
+
+                            return {
+                                system_mode: modes[v],
+                                system_mode_device: modes[v],
```
Koenkk, line 11391: This is not exposed, I think it can be removed?

PR #8493, opened 2024-12-19
feat(add): ota and power on for Innr AE 270 T

AquaMorph:


AquaMorph:
I don't think this is working right. I see the following on my bulbs state when I test it out
```
  "update": {
        "installed_version": -1,
        "latest_version": -1,
        "state": null
    },
    "update_available": null
```

AquaMorph:
Nevermind I was running a version of Zigbee2MQTT that had a different syntax. I fixed that for my test and the OTA update worked fine for me.

Koenkk:
Thanks!

PR #8496, opened 2024-12-21
Update moes.ts - update ZC-HM fingerprint to include most recent version

Justblair:
Just needed to add '_TZE284_rjxqso4a' to the fingerprint and it starts working


Koenkk:
Thanks!

PR #8507, opened 2024-12-22
feat: Use vitest for testing

Nerivec:
I only included the minimum to convert to vitest:
- API changes
- Convert require to ES
- Small fixes

TODO (separate PR):
- Fix OTA tests (faulty section disabled) and re-enable coverage
- Convert all to TS
  - Requires fixing several imports/exports throughout the codebase and fixing several types
- Adding coverage for `src/lib` would be good


Koenkk:
Thanks!

PR #8509, opened 2024-12-23
feat: support Sunricher SR-ZG9030A-MW

niracler:
## Notably

While developing support for Sunricher devices, I discovered a ModelId conflict between two devices in their product line. To resolve this, I implemented an expose that allows users to switch the ModelId. This implementation required skipping the set/get access check in the test cases.

I'm not entirely confident if bypassing the test case through this approach is the best solution. I would appreciate your feedback and suggestions for potential improvements.

Reference code showing the ModelId conflict handling:

```typescript
const sunricherTz = {
    setModel: {
        key: ['model'],
            await entity.write('genBasic', {modelId: value});
            return {state: {model: value}};
        },
    } satisfies Tz.Converter,
};
```

And the test case bypass:

```typescript
if (device.toZigbee.includes(sunricher.tz.setModel)) return;
```


src/devices/sunricher.ts
```diff
@@ -361,7 +362,159 @@ async function syncTime(endpoint: Zh.Endpoint) {
...
+
+                const attribute = lookup[key as keyof typeof lookup];
+                await entity.write('genBasic', {[attribute.attribute]: {value: value, type: attribute.type}}, {manufacturerCode: 0x1224});
+                return {state: {[key]: value}};
+            },
+        },
+    ];
+
+    const exposes: Expose[] = [
+        e
```
Koenkk, line 420: Could you use modern extend for all of these? Examples: https://github.com/Koenkk/zigbee-herdsman-converters/blob/9bb99f0c972c5a339c846ac966bf2739d591fe42/src/devices/sonoff.ts#L1180

src/devices/sunricher.ts
```diff
@@ -361,7 +362,159 @@ async function syncTime(endpoint: Zh.Endpoint) {
...
+            deviceEndpoints({endpoints: {'1': 1, '2': 2, '3': 3}}),
+            light(),
+            occupancy(),
+            illuminance({endpointNames: ['3']}),
+            commandsOnOff(),
+            commandsLevelCtrl(),
+        ],
+        meta: {multiEndpoint: true},
+        toZigbee: [sunricher.tz.setModel],
+        exposes: [e.enum('model', ea.SET, ['HK-DIM', 'ZG9030A-MW']).withDescription('Model of the device')],
```
Koenkk, line 516: Where is this used for? (what does changing the `modelId` do)

niracler: Some `ZG9030A-MW` devices were mistakenly set with the modelId `HK-DIM` during manufacturing. Therefore, we need to provide a method that allows users to update the modelId of these `HK-DIM` devices to `ZG9030A-MW`.

src/devices/sunricher.ts
niracler: DONE

Koenkk:
Thanks!

src/devices/sunricher.ts
87insane: I updated to the 2 version of z2m. I deleted the ZG9030A-MW and re-paired it. It was recognized again as HK-DIM. When I want to change this via the web interface, nothing happens.

So I did this directly in the database and restarted z2m. The correct device will then be displayed in the web interface. But I don't see any movement. When I walk through the sensor, nothing is transmitted to my FHEM via MQTT.

image

PR #8512, opened 2024-12-24
Fix mode enum values for TRV603

Cosik:
During using of TRV603 I noticed that enum for setting mode is using values (0,2) instead of (0,1) which cause errors during usage.


Koenkk:
Thanks!

PR #8520, opened 2024-12-26
Update zigbeetlc.ts for MHO-C401

NuclearKitty:
Added definition für recently supported MiaoMiaoCe MHO-C401 (old version, without N) in src/devices/zigbeetlc.ts


Koenkk:
Thanks!

PR #8523, opened 2024-12-27
Feat: two new track lights from XAL (models 050-0131558M and 050-1212558H)

salgarelli:
Added vendor XAL, w/ two new track lights (models 050-0131558M and 050-1212558H)


Koenkk:
Please also update https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts

Koenkk:
Thanks!

PR #8524, opened 2024-12-27
Increase TS0601_thermostat_3 schedule transitions from 4 to 6

Schmallon:
This PR resolves an inconsistency in the scheduling logic for `TS0601_thermostat_3` by increasing the number of transitions passed to `thermostatScheduleDayMultiDPWithDayNumber` from 4 to 6, fixing [Koenkk/zigbee2mqtt#19787](https://github.com/Koenkk/zigbee2mqtt/issues/19787). Currently, the code enforces only four transitions even though the `scheduleAllDays` format string explicitly indicates six. This mismatch can cause problematic behavior, in particular thermostats unexpectedly requesting a temperature of 153.6.

I have tested this change on two devices (`_TZE284_o3x45p96` and `_TZE200_9xfjixap`), and it solves the issue for both. While further testing across additional models would be ideal, the existing code is demonstrably incorrect: it references six transitions yet imposes a four-transition limit. Merging this PR will rectify that inconsistency and reduce confusion for other users, so I recommend moving forward with this fix.


Koenkk:
Thanks!

PR #8526, opened 2024-12-27
Update philips.ts

asgothian:
Add 3 mtr Variant of Hue solo light strip.


asgothian:
seems it got added while I was testing.. closing this

PR #8528, opened 2024-12-28
Add support for Salus FC600

mkrle:
Feedback welcome!

TODO:

- `running_state` will report values not in `thermostatRunningStates` constants if the fan is running on med or high speed. Should I add those states to `thermostatRunningStates` in `src/lib/constants.ts` or is there a better way?
- OTA fails with `device.definition.ota.isUpdateAvailable is not a function`. I haven't been able to figure this out so far. If I can't make it work I'll remove it from the definition.
- I don't know how the "presets" like eco/off etc are configured. I think it's nice to have but I don't see this as a blocker

### DIFF
--- a/src/devices/salus_controls.ts
+++ b/src/devices/salus_controls.ts
@@ -1,4 +1,5 @@
 import fz from '../converters/fromZigbee';
+import tz from '../converters/toZigbee';
 import * as exposes from '../lib/exposes';
 import {electricityMeter, onOff} from '../lib/modernExtend';
 import * as reporting from '../lib/reporting';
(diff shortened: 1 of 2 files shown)


src/devices/salus_controls.ts
```diff
@@ -95,6 +96,55 @@ const definitions: DefinitionWithExtend[] = [
...
+            const binds = ['genBasic', 'genIdentify', 'genTime', 'hvacThermostat', 'hvacFanCtrl', 'hvacUserInterfaceCfg'];
+            await reporting.bind(endpoint, coordinatorEndpoint, binds);
+            await reporting.thermostatTemperature(endpoint);
+            await reporting.thermostatOccupiedHeatingSetpoint(endpoint);
+            await reporting.thermostatSystemMode(endpoint);
+            await reporting.fanMode(endpoint);
+            await reporting.thermostatRunningMode(endpoint);
+            await reporting.thermostatRunningState(endpoint);
+        },
+        ota: {manufacturerName: 'SalusControls'},
```
Koenkk, line 146:
```suggestion
        ota: true,
```

Koenkk:
> running_state will report values not in thermostatRunningStates constants if the fan is running on med or high speed. Should I add those states to thermostatRunningStates in src/lib/constants.ts or is there a better way?

Please add them there

mkrle:
> Please add them there

Thanks, done.

Koenkk:
I will squash them, thanks!

PR #8529, opened 2024-12-28
Add endpoint capability to lock type

Zitrium31:
Purpose of this change:
- add endpoint capability to lock in `fromZigbee` converter and `modernExtend` lock function
- add pincode capability for LOCK, UNLOCK, TOGGLE _state_ commands in `toZigbee lock` set converter
- add dummy garage door and lock  'GADOLOCK' device in `custom_devices_diy` definitions for test purpose. This device is made of  several endpoints using state key in `toZigbee` converter : onOff, windowCovering, lock
- fix proper toZigbee selection with device made of several endpoints using state key in `toZigbee` converter in _exposes access_ test

In short, this change supplements issue Koenkk/zigbee2mqtt#24352 for lock.

Note : A complementary PR will be submitted shortly in repo [Koenkk/zigbee2mqtt](https://github.com/Koenkk/zigbee2mqtt) to add endpoint support to lock type in lib/extension/homeassistant.ts :arrow_right: See PR https://github.com/Koenkk/zigbee2mqtt/pull/25359

:slightly_smiling_face: Thanks in advance for the review of this PR. Should you have any remark, please ask.

### DIFF
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -332,8 +332,20 @@ const converters2 = {
     lock: {
         key: ['state'],
         convertSet: async (entity, key, value, meta) => {
-            utils.assertString(value, key);
-            await entity.command('closuresDoorLock', `${value.toLowerCase()}Door`, {pincodevalue: ''}, utils.getOptions(meta.mapped, entity));
+            // If no pin code is provided, value is a only string. Ex: "UNLOCK"
+            let state = utils.isString(value) ? value.toUpperCase() : null;
+            let pincode = '';
+            // If pin code is provided, value is an object including command and code. Ex: {command: "UNLOCK", code: "1234"}
+            if (utils.isObject(value)) {
+                if (value.code) {
+                    pincode = utils.isString(value.code) ? value.code : '';
+                }
+                if (value.command) {
+                    state = utils.isString(value.command) ? value.command.toUpperCase() : null;
+                }
+            }
+            utils.validateValue(state, ['LOCK', 'UNLOCK', 'TOGGLE']);
+            await entity.command('closuresDoorLock', `${state.toLowerCase()}Door`, {pincodevalue: pincode}, utils.getOptions(meta.mapped, entity));
         },
         convertGet: async (entity, key, meta) => {
             await entity.read('closuresDoorLock', ['lockState']);
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -335,13 +335,13 @@ const converters2 = {
             // If no pin code is provided, value is a only string. Ex: "UNLOCK"
             let state = utils.isString(value) ? value.toUpperCase() : null;
             let pincode = '';
-            // If pin code is provided, value is an object including command and code. Ex: {command: "UNLOCK", code: "1234"}
+            // If pin code is provided, value is an object including new state and code. Ex: {state: "UNLOCK", code: "1234"}
             if (utils.isObject(value)) {
                 if (value.code) {
                     pincode = utils.isString(value.code) ? value.code : '';
                 }
-                if (value.command) {
-                    state = utils.isString(value.command) ? value.command.toUpperCase() : null;
+                if (value.state) {
+                    state = utils.isString(value.state) ? value.state.toUpperCase() : null;
                 }
             }
             utils.validateValue(state, ['LOCK', 'UNLOCK', 'TOGGLE']);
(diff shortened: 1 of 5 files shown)


src/devices/custom_devices_diy.ts
```diff
@@ -1193,6 +1195,18 @@ const definitions: DefinitionWithExtend[] = [
             }),
         ],
     },
+    {
+        zigbeeModel: ['GADOLOCK'],
+        model: 'GADOLOCK',
+        vendor: 'Custom devices (DiY)',
+        description: 'Garage door with lock',
```
Koenkk, line 1202: Are there any instructions available on how to build such lock? This is a requirement for custom devices to be integrated.

Zitrium31: Thank you for your review.

I have actually a similar DIY device at home that controls an existing garage door motor in addition to 4 electrical latches at each corners. This project is very specific and I have my own external definition. From this point of view, I can say there is no interrest to pollute zhc repo with such device.

I found no existing device made of such lock type with endpoints. Thus I felt it was valuable (ex: to avoid regression bug like issue https://github.com/Koenkk/zigbee2mqtt/issues/24352) to add this represantative `GADOLOCK` model to re-inforce regression test on such device made of several endpoints using state key in toZigbee converter : onOff, windowCovering, lock.

This being said, I understand without any problem that this approach does not fulfill zhc policy and this device can be removed from this PR without any problem. No matter for me 🙂

src/converters/toZigbee.ts
```diff
@@ -332,8 +332,20 @@ const converters2 = {
...
-            await entity.command('closuresDoorLock', `${value.toLowerCase()}Door`, {pincodevalue: ''}, utils.getOptions(meta.mapped, entity));
+            // If no pin code is provided, value is a only string. Ex: "UNLOCK"
+            let state = utils.isString(value) ? value.toUpperCase() : null;
+            let pincode = '';
+            // If pin code is provided, value is an object including command and code. Ex: {command: "UNLOCK", code: "1234"}
+            if (utils.isObject(value)) {
+                if (value.code) {
+                    pincode = utils.isString(value.code) ? value.code : '';
+                }
+                if (value.command) {
```
Koenkk, line 343:
```suggestion
                if (value.state) {
```

I think this is more consistent

src/devices/custom_devices_diy.ts
Koenkk: Yes please remove the device from this PR

test/index.test.js
```diff
@@ -441,7 +441,9 @@ describe('index.js', () => {
                     property = expose.property.slice(0, (expose.endpoint.length + 1) * -1);
                 }
 
-                const toZigbee = device.toZigbee.find((item) => item.key.includes(property));
+                const toZigbee = device.toZigbee.find(
```
Koenkk, line 444: Why was this change needed?

src/devices/custom_devices_diy.ts
Zitrium31: This is understood : I gonna remove this phantom device.

src/converters/toZigbee.ts
Zitrium31: Now I see your suggestion, I fully agree it is more consistent with a key named `state`. I will update that too.

test/index.test.js
Zitrium31: This change has the same purpose as the one we did here : https://github.com/Koenkk/zigbee2mqtt/blame/bc6eccf6152c4cd91e75309f7ea0b8ad4e8835d4/lib/extension/publish.ts#L263

Let's take an example : If you have for instance a device made of a winndowClosures cluster (Ex: A garage door) and a onOff cluster (Ex: A garage light), then you have here a mix up between toZigbee converter because `find` function stops at the first key `state` found if we don't use the additional endpoint criteria. This is also explained in issue https://github.com/Koenkk/zigbee2mqtt/issues/24352.

Today, with the removal of my `GADOLOCK` device, no false positive error will be triggered (even with all others thousand devices! ).

I hope I am clear enough. Please advise if you prefer keep this change preventively or remove it because considered not relevant. At my level (a simple z2m user), I will simply follow your recommendation 🙂

src/devices/custom_devices_diy.ts
Zitrium31: Done

src/converters/toZigbee.ts
Zitrium31: Done

PR #8530, opened 2024-12-28
Add support for RBSH-SWD2-ZB

leonw-04:
RBSH-SWD2-ZB is the same as RBSH-SWD-ZB but reports different model. RBSH-SWD2-ZB is the with matter and zigbee. Same as #8351


Koenkk:
Thanks!

PR #8535, opened 2024-12-29
fix: read pilot wire mode for Nodon and Equation modules

vmeurisse:
Moved the cluster and zigbee parsers to dedicated file to avoid confusion with manuSpecificAssaDoorLock in findClusterNameById.

Affect: SIN-4-FP-20, SIN-4-FP-21, SIN-4-FP-21_EQU

Fixes Koenkk/zigbee2mqtt#25333


Koenkk:
Thanks!

PR #8536, opened 2024-12-29
Detect _TZE284_locansqn as Nous TH01Z

kamaradclimber:
_TZE284_locansqn is the same as all other variants of TH01Z but report a different manufacturer. It features same LCD screen, same sensors and a clock.


Koenkk:
Thanks!

PR #8546, opened 2025-01-01
fix: Fix configure failure for Datek HSE2905E

mellbye:
The Datek HSE2905E meter reader stopped working properly after a refactor in e8883c0. The device doesn't actually expose an onOff cluster. It also doesn't support power readings on the HA cluster, so separately extend the `electricityMeter` object for SE Metering and HA Electricity Measurement.

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/25401

### DIFF
--- a/src/devices/datek.ts
+++ b/src/devices/datek.ts
@@ -5,7 +5,7 @@ import tz from '../converters/toZigbee';
 import * as constants from '../lib/constants';
 import {repInterval} from '../lib/constants';
 import * as exposes from '../lib/exposes';
-import {electricityMeter, onOff, temperature} from '../lib/modernExtend';
+import {electricityMeter, temperature} from '../lib/modernExtend';
 import * as reporting from '../lib/reporting';
 import {DefinitionWithExtend} from '../lib/types';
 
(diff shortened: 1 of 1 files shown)


src/devices/datek.ts
```diff
@@ -41,7 +41,22 @@ const definitions: DefinitionWithExtend[] = [
...
+                fzMetering: fz.metering_datek,
+                producedEnergy: true,
+            }),
+            electricityMeter({
+                threePhase: true,
+                cluster: 'electrical',
+                power: false,
+                voltage: {},
+                current: {},
+                fzMetering: fz.metering_datek,
```
Koenkk, line 56: I think this and voltage + current can be removed

mellbye: Agreee that the custom fromZigbee should only be set on the SE variant.

When it comes to voltage/current: These were previously (pre refactor) set up to report changes down to 0, while the default when extending is 5V and 0.05A respectively. Should I perhaps change them to `{change: 0}` instead of removing? I noticed that especially voltage becomes almost pointless when set to 5V, since the normal ripple around 230V is much smaller than that.

Koenkk: >  I noticed that especially voltage becomes almost pointless when set to 5V, since the normal voltage variation around 230V is much smaller than that.

Maybe we should reduce the default value of 5V? What about 3?

mellbye: I guess it depends on what the intent of the default is: Is it to capture significant voltage changes (e.g. brownouts), or to keep real-time eye on a healthy power supply? 

This is what I currently have in Grafana, based on the full 0.1V precision of the meter:

Screenshot 2025-01-02 at 14 20 40

mellbye: For now, I removed current and set voltage to `change: 0` to make the behavior for the Datek meter similar to how it worked in the past. Please let me know if you instead want me to change the default for all electricity meters.

Koenkk: > Is it to capture significant voltage changes

We should find a healthy balance here, but change `0` use cases are very specific, if a users wants that it can always be changed manually via the reporting tab. I think all devices should use the same default (for consistency purposes)

mellbye: I'm completely fine with removing it if that matches the project goals. It just seemed like a regression to me for data that was previously streamed over MQTT to stop getting emitted after an update. I'll update this PR to leave it at the default, and let a discussion about potentially changing the default happen separately.

Koenkk:
Thanks!

PR #8551, opened 2025-01-02
add preset for Salus FC600

mkrle:
Feedback welcome!

While trying to figure out how to control presets on Salus FC 600 (https://github.com/Koenkk/zigbee-herdsman-converters/pull/8528) I was able to map out parts of a manufacturer specific cluster it uses.

I was able to probe attributes `0x0004`-`0x0007` and `0x000f` (and their data types) but I could not figure out their purpose. I have labeled them `attr4`, `attr5` and so on using their decimal IDs, perhaps someone will figure out the purpose later on. Please let me know if this is not a good approach.

The attributes `0x0000`-`0x0002` don't exist in this device (FC600) but I believe they still belong to this cluster.  Since the cluster is now defined within the device I'm not sure if I should keep them them for documentation purpose, or leave them out.

The command performs a soft reset (reboot) of the thermostat, and the thermostat never sends back the response. I don't know if and how it's possible to flag this, so that the command doesn't time out with an exception when in fact it was successful.

### DIFF
--- a/src/devices/salus_controls.ts
+++ b/src/devices/salus_controls.ts
@@ -1,7 +1,9 @@
+import {Zcl} from 'zigbee-herdsman';
+
 import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as exposes from '../lib/exposes';
-import {electricityMeter, onOff} from '../lib/modernExtend';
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #8562, opened 2025-01-04
tuya.ts TOQCB2-80 fingerprints updated

jonaubf:
Add a _TZE204_tzreobvu fingerprint to support Tuya TOQCB2-80 3 poles device


Koenkk:
Thanks!

PR #8569, opened 2025-01-06
CCT5010-0001 was sold under the name 550B1012 in Denmark

poizan42:
Adding as white label model for lack of a better place. It was sold as a fully first-party product, albeit with additional "Lauritz Knudsen by Schneider Electric" branding.

Originally made as a PR to zigbee2mqtt.io: [zigbee2mqtt.io#3351](https://github.com/Koenkk/zigbee2mqtt.io/pull/3351):

> Not sure exactly how we usually document multiple model numbers for the same hardware when it's not quite a white label situation. It was sold with both Schneider Electric and "LK" (Lauritz Knudsen) branding on the device (see https://www.lk.dk/produkter/550B1012/wiser-tradlos-lysdaemper-for-indbygning-puck)
> 
> They exists as two different products in Schneider Electric's catalog
> 
>     * https://www.se.com/dk/da/product/CCT5010-0001/connected-dimmer-wiser-micro-module/
> 
>     * https://www.se.com/dk/da/product/550B1012/wiser-tr%C3%A5dl%C3%B8s-lysd%C3%A6mper-for-indbygning-puck/
> 
> 
> - but their ["dimmer guide"](https://www1.lk.dk/apps/dimmertool/dimmerGuide.ph7) makes it clear that they are in fact the same product with two different SKUs (or CRs whatever that means):
> 
> image

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -682,7 +682,7 @@ const definitions: DefinitionWithExtend[] = [
                 .enum('dimmer_mode', ea.ALL, ['auto', 'rc', 'rl', 'rl_led'])
                 .withDescription('Sets dimming mode to autodetect or fixed RC/RL/RL_LED mode (max load is reduced in RL_LED)'),
         ],
-        whiteLabel: [{vendor: 'Elko', model: 'EKO07090'}],
+        whiteLabel: [{vendor: 'Elko', model: 'EKO07090'}, {vendor: 'Schneider Electric', model: '550B1012'}],
     },
     {
         zigbeeModel: ['PUCK/SWITCH/1'],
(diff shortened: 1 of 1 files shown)


poizan42:
Actually there is precedence for adding multiple model numbers from the same vendor directly to model: https://github.com/Koenkk/zigbee-herdsman-converters/blob/fe08be8e1ae85ed98046fe5abe7728bfc486126d/src/devices/schneider_electric.ts#L689

poizan42:
Ugh, would it be possible to make the CI scripts actually print what prettier is unhappy about? I'm not really into the whole node world, but I couldn't see anything obvious from the documentation. Maybe it could call prettier with --write and print out the unified diff between working copy and PR branch?

I will try to set up a node development environment when I have the time to figure out what the problem is, but it feels like a bit of high barrier to just update some documentation :/

Koenkk:
Thanks!

PR #8570, opened 2025-01-06
Update tuya.ts

cslotay:
Added Tuya Dimmer  ZDMS16-US


Koenkk:
Can you try with the definition from `TS0601_dimmer_4`?

PR #8571, opened 2025-01-06
fix: Fix Centralite 4257050-ZHAC to ignore 'transition'

bdd:
Came across the error below getting logged when an HA automation was dimming the light down with a transition.

`error: z2m: No converter available for 'transition' (4)`

Added a `tz.ignore_transition` to reduce log spam.
Tested as an external converter using HA add-on running on the latest HAOS.


Koenkk:
Thanks!

PR #8572, opened 2025-01-06
fix Salus FC600 OTA

mkrle:
Device reports `zhc:tz: Read result of 'genBasic': {"manufacturerName":"SALUS"}` but in OTA index it's `SalusControls`.


Koenkk:
Thanks!

PR #8564, opened 2025-01-05
Add support for ID200W-ZIGB from ION Industries (_TZE200_ykgar0ow)

redelschaap:
Add support for the [ION Industries ID200W-ZIGB](https://www.ionindustries.com/winkel/led-zigbee-dimmer-0-3-200-watt/) dimmer.

Tested locally for a few weeks now, work perfectly.

PR #8574, opened 2025-01-07
Add Nous E6 10.24 revision

amadeuszxd:
Hello, I bought new revision of Nous E6 Temperature sensor, which is not supported because it has new fingerprint "_TZE200_wtikaxzs". I tested it and it has more options than normal e6, which i named by the revision 10-24 (I was it on the back of the device). Every of the option is locally tested by me.

### DIFF
--- a/src/devices/nous.ts
+++ b/src/devices/nous.ts
@@ -82,7 +82,7 @@ const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        fingerprint: tuya.fingerprint('TS0601', ['_TZE284_wtikaxzs', '_TZE200_nnrfa68v', '_TZE200_zppcgbdj']),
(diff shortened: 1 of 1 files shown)


Koenkk:
Could you merge this with the already existing Nous E6 definition and make the exposes a function? See https://github.com/Koenkk/zigbee-herdsman-converters/pull/8564/files for an example.

Marcoparigi:
When this will be deployed ?
Thank you 😉

Koenkk:
Thanks!

pubbouder:
Hello,

I have almost similar equipment from home.

const definition = {
    zigbeeModel: ['TS0601'],
    model: 'TS0601',
    vendor: '_TZE200_wtikaxzs',
    extend: [],
    meta: {},
};

module.exports = definition;

Can you add it to the general definition?
Thanks

Koenkk:
@pubbouder `_TZE200_wtikaxzs` will be supported in the 1 February release.

PR #8587, opened 2025-01-10
feat: add A-OK AM25 _TZE200_7shyddj3

aarriiooss:
I recently got some roller blinds installed that use the AM25 tube motors from A-OK. They showed up as unsupported but adding the new ID (`_TZE200_7shyddj3`) in an external converter has worked fine for the past few weeks.


Koenkk:
Thanks!

PR #8589, opened 2025-01-11
feat(add): add thermostat endpoints to Ubisys H10

sjorge:
Seems the thermostat stuff I had in a comment on https://github.com/Koenkk/zigbee2mqtt/issues/18121#issuecomment-2571346226 does seem to work.

So this just adds that, I don't have the device myself so cannot verify the functionality. So I did not rewrite this to modernExtends as I have no way of testing with an actual device.


Koenkk:
Thanks!

PR #8595, opened 2025-01-12
Fix inverted `position` on Tuya motorized blinds

TzeWey:
Resolves the inverted `position` on Tuya motorized blinds not being reflected by using the `coverPosition` converter for datapoint '3' (which is also the `position`)

Verified with the A-OK AM25 that is recognized as B09M3R35GC _TZE200_9p5xmj5r Unable to verify with a motor that is non-inverted.


TzeWey:
I missed this change when raising #8481

Koenkk:
Thanks!

PR #8616, opened 2025-01-15
Update wirenboard.ts

BurchuMikhail:
- add the ability to change baud rate
- add a flag for the presence of a modbus connection


Koenkk:
Thanks!

PR #8619, opened 2025-01-15
Update nedis.ts - added newer nedis radiator valve version TZE284

Schneekatzeeins:
Added newer nedis thermostat radiator valve with hardware version TZE284_ne4pikwm


Koenkk:
Thanks!

PR #8626, opened 2025-01-17
add Third Reality two types of plugs

3reality-support:


src/devices/third_reality.ts
```diff
@@ -417,6 +417,66 @@ const definitions: DefinitionWithExtend[] = [
...
+                acPowerDivisor: 10,
+            });
+            device.save();
+        },
+    },
+    {
+        zigbeeModel: ['3RSP02064Z'],
+        model: '3RSP02064Z',
+        vendor: 'Third Reality',
+        description: 'Zigbee / BLE smart plug with power',
```
Koenkk, line 454: Can you try removing frmoZigbee, toZigbee, exposes and configure and add:

```
extend: [onOff(), electricityMeter()],
```

(same for the other device)

PR #8628, opened 2025-01-17
Add support for Moes BHT series Thermostat device (TS0601/_TZE204_xalsoe3m)

egabor:
This device has the same functionality as `_TZE204_aoclfnxz`.


Koenkk:
Thanks!

SSG1111:
Thanks for the update! B ut this doesn't work properly.
attribute "temperature" has wrong scaling (1 instead of 0.1)
attribute "hvac_action" is not available, therefore CLIMATE visual doesn't display the status correctly

SSG1111:
No, no solution... we need to apply for the fix somehow. 
Wrong scailing - doesn't look nice in HA, but devices are still working as they should
Missing hvac_action - the has probably something to do with Tuya, not Z2M. New firmawre update maybe needed? I was thinking to create custom CLIMATE entity and use that one in HA interface.

https://www.home-assistant.io/integrations/generic_thermostat/

PR #8635, opened 2025-01-18
Add Niko switch action reporting functionality

svenjochems:
Some Niko switches are not reporting state updates as 'action' values.
Based on research in https://github.com/Koenkk/zigbee2mqtt/issues/13737 and https://github.com/zigpy/zha-device-handlers/pull/3701.

Changes:
* Add `switchActionReporting` attribute to enable or disable the reporting of `action` values
* Move custom Niko Clusters `manuSpecificNiko1` and `manuSpecificNiko2` from zigbee-herdsman to modernExtend `manuSpecificNikoConfig` and `manuSpecificNikoState`
* Change `ledState` property from `ON/OFF` to `ON/OFF/Blue/Red/Purple`
* Add `ledSyncMode` option
* Change `action` implementation to support pushing multiple buttons together
* action `null` values are not published anymore

### DIFF
--- a/src/devices/niko.ts
+++ b/src/devices/niko.ts
@@ -1,6 +1,9 @@
+import {Zcl} from 'zigbee-herdsman';
+
 import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as exposes from '../lib/exposes';
+import {deviceAddCustomCluster} from '../lib/modernExtend';
 import * as reporting from '../lib/reporting';
 import {DefinitionWithExtend, Fz, KeyValue, Tz} from '../lib/types';
 import * as utils from '../lib/utils';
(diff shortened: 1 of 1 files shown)


src/devices/niko.ts
```diff
@@ -330,6 +376,7 @@ const definitions: DefinitionWithExtend[] = [
                 'release_right_ext',
             ]),
             e.enum('operation_mode', ea.ALL, ['control_relay', 'decoupled']),
+            e.binary('action_reporting', ea.ALL, true, false),
```
Koenkk, line 379: Could you add a description here? (`.withDescription()` like below)

src/devices/niko.ts
```diff
@@ -23,14 +56,21 @@ const local = {
...
+                if (msg.data.switchActionReporting !== undefined) {
+                    const actionReportingMap: KeyValue = {0x00: false, 0x1f: true};
+                    state['action_reporting'] = utils.getFromLookup(msg.data.switchActionReporting, actionReportingMap);
+                }
                 if (msg.data.switchAction !== undefined) {
                     // NOTE: a single press = two separate values reported, 16 followed by 64
                     //       a hold/release cycle = three separate values, 16, 32, and 48
+                    // NOTE: these values should be interpreted bitwise
+                    //       when pushing multiple buttons at the same time, multiple bits can be set simultaneously and should generate multiple events
+                    //       currently, these values are not mapped and thus ignored
```
svenjochems, line 73: @sjorge @Koenkk Is this possible to implement in the converter? It seems 1 value always maps to 1 'event', but in this case it could map to multiple 'events' when pushing multiple buttons at once.

sjorge: My first reaction would be to have the actions end up in a list but I don't think that is an excepted value for action property.

The next best thing would be to decode them into a list, then if there are > 1 in there merge them and add these as extra possible values.

Koenkk: Would publishing multiple actions make sense here? To do that call `publish({"action": "x"})` here.

src/devices/niko.ts
```diff
@@ -23,66 +58,78 @@ const local = {
...
 
-                    state['action'] = actionMap[msg.data.switchAction];
+                    for (const button in buttonShift) {
+                        const shiftedValue = (msg.data.switchAction >> buttonShift[button]) & 0xf;
+                        for (const action in actions) {
+                            if (shiftedValue == actions[action]) {
+                                const buttonPostFix = button === '' ? '' : '_' + button;
+                                if ('null' === action) {
+                                    // publish without button name
+                                    publish({action: 'null'});
```
svenjochems, line 101: I wonder if we should actually push the null values or if we should just skip them. I added them to be backwards compatible.

src/devices/niko.ts
```diff
@@ -119,50 +166,100 @@ const local = {
...
             },
             convertGet: async (entity, key, meta) => {
-                await entity.read('manuSpecificNiko1', ['outletLedColor']);
+                await entity.read('manuSpecificNikoConfig', ['outletLedColor']);
+            },
+        } satisfies Tz.Converter,
+        switch_led_sync_mode: {
+            key: ['led_sync_mode'],
+            convertSet: async (entity, key, value, meta) => {
+                // This could be set endpoint specific, where the first 4 bits are used for the left button and the last 4 bits for the right button
```
svenjochems, line 226: Any pointers on how to make this endpoint specific are welcome. I need the value for both endpoints to write the correct value to the device..

src/devices/niko.ts
svenjochems: Thanks, implemented it this way!

src/devices/niko.ts
Koenkk: I think we should not

src/devices/niko.ts
Koenkk: Use the `meta.state` (this is identical to the state that is published to MQTT)

src/devices/niko.ts
svenjochems: I have removed it

src/devices/niko.ts
svenjochems: I managed to get this working for multiple endpoints. Thanks!

svenjochems:
@Koenkk @sjorge I have finished my changes. This PR is ready for review

Koenkk:
Looks good, thanks!

Magkoooh (approved):

Magkoooh (approved):

PR #8637, opened 2025-01-18
Workaround IKEA bulbs freezing during a brightness & color transition

bobrippling:
This PR resolves Koenkk/zigbee2mqtt#18574 by unfreezing IKEA bulbs (that we know to be frozen) before sending commands to them.

Any toZigbee handlers that use commands which will need the bulb to not be frozen are wrapped by `ikea_bulb_unfreeze()`.

`ikea_bulb_unfreeze()` will:
- detect if the key is likely to need an unfreeze (e.g. `state: off`) and if so:
- install a proxy the entity's `command()` method where:
        - a frozen bulb is unfrozen before submitting a command to it
        - remember that a bulb is frozne, if a freezing command is sent to the bulb (e.g. colour transition)

The unfreeze method used is `genLevelCtrl.stop`, which works for all bulbs (as far as we know) and crucially completes the previous transition, which is probably more beneficial than leaving the bulb in a halfway state.

[Brightness transitions] with a colour are still skipped (Koenkk/zigbee2mqtt#1810) - this PR does not alter that.

Tested with:
```sh
pub() {
  mosquitto_pub -t "$light"/set -m "$1"
}

light=zigbee2mqtt/...

echo "turning on, dim & cool"
pub '{"state": "on", "brightness": 2, "color_temp": 150, "transition": 0}'

sleep 2
echo "converting to bright & warm"
pub '{"brightness": 255, "color_temp": 500, "transition": 5}'

sleep 3
echo "turning off"
pub '{"state": "off"}'
```

[Brightness transitions]: https://github.com/Koenkk/zigbee-herdsman-converters/blob/b1ea96f097c53e78b363433e0b96e992afe7914d/src/lib/utils.ts#L464-L474=

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -33,7 +33,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED1545G12",
         vendor: "IKEA",
         description: "TRADFRI bulb E26/E27, white spectrum, globe, opal, 980 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: [
@@ -102,7 +102,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED1936G5",
         vendor: "IKEA",
         description: "TRADFRI bulb E26/E27, white spectrum, globe, opal, 440/450/470 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: [
@@ -114,14 +114,14 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED2003G10",
         vendor: "IKEA",
         description: "TRADFRI bulb E26/27, white spectrum, globe, opal, 1055/1100/1160 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: ["TRADFRIbulbE26WSglobeclear800lm", "TRADFRIbulbE27WSglobeclear806lm", "TRADFRIbulbE26WSglobeclear806lm"],
         model: "LED2004G8",
         vendor: "IKEA",
         description: "TRADFRI bulb E26/E27, white spectrum, globe, clear, 800/806 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: [
@@ -169,14 +169,14 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED1732G11",
         vendor: "IKEA",
         description: "TRADFRI bulb E26/E27, white spectrum, globe, opal, 1000 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: ["TRADFRI bulb E27 WW 806lm", "TRADFRI bulb E26 WW 806lm"],
         model: "LED1836G9",
         vendor: "IKEA",
         description: "TRADFRI bulb E26/E27, warm white, globe, opal, 806 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({turnsOffAtBrightness1: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({turnsOffAtBrightness1: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: ["TRADFRI bulb E27 WS clear 806lm", "TRADFRI bulb E26 WS clear 806lm"],
@@ -209,7 +209,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED2035G10",
         vendor: "IKEA",
         description: "TRADFRI bulb B22, white spectrum, globe, opal, 1055 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     // #endregion E26/E27/B22
     {
@@ -225,7 +225,7 @@ export const definitions: DefinitionWithExtend[] = [
         description: "TRADFRI bulb E12/E14/E26/E27, color/white spectum, globe, opal, 600 lm",
         extend: [
             addCustomClusterManuSpecificIkeaUnknown(),
-            ikeaLight({colorTemp: {range: [153, 500], viaColor: true}, color: true}), // light is pure RGB (XY), advertise 2000K-6500K
+            ikeaLight({colorTemp: {range: [153, 500], viaColor: true}, color: true, withUnfreezeSupport: true}), // light is pure RGB (XY), advertise 2000K-6500K
             m.identify(),
         ],
     },
@@ -249,7 +249,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED1835C6",
         vendor: "IKEA",
         description: "TRADFRI bulb E12/E14/E17, white spectrum, candle, opal, 450/470/440 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: ["TRADFRI bulb E14 WS globe 470lm", "TRADFRI bulb E12 WS globe 450lm"],
@@ -263,7 +263,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED1733G7",
         vendor: "IKEA",
         description: "TRADFRI bulb E14, white spectrum, globe, opal, 600 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: ["TRADFRI bulb E14 W op/ch 400lm", "TRADFRI bulb E12 W op/ch 400lm", "TRADFRI bulb E17 W op/ch 400lm"],
@@ -317,7 +317,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED1949C5",
         vendor: "IKEA",
         description: "TRADFRI bulb E12/E14/E17, white spectrum, candle, opal, 450/470/440 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: ["TRADFRI bulb E14 CWS globe 806lm", "TRADFRI bulb E12 CWS globe 800lm", "TRADFRI bulb E17 CWS globe 810lm"],
@@ -346,7 +346,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED1537R6/LED1739R5",
         vendor: "IKEA",
         description: "TRADFRI bulb GU10, white spectrum, 400 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, withUnfreezeSupport: true}), m.identify()],
     },
     {
         zigbeeModel: ["TRADFRI bulb GU10 W 400lm"],
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -33,7 +33,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED1545G12",
         vendor: "IKEA",
         description: "TRADFRI bulb E26/E27, white spectrum, globe, opal, 980 lm",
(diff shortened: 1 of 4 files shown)


src/converters/toZigbee.ts
```diff
@@ -3,6 +3,7 @@ import {Zcl} from 'zigbee-herdsman';
 import * as libColor from '../lib/color';
 import * as constants from '../lib/constants';
 import * as exposes from '../lib/exposes';
+import {unfreezeMechanisms, UnfreezeSupport} from '../lib/ikea';
```
bobrippling, line 6: I feel like vendor-specific imports shouldn't be here and the logic from `../lib/ikea` should be inlined into the `ikea_bulb_unfreeze` converter below

src/devices/ikea.ts
```diff
@@ -210,6 +211,7 @@ const definitions: DefinitionWithExtend[] = [
         vendor: 'IKEA',
         description: 'TRADFRI bulb B22, white spectrum, globe, opal, 1055 lm',
         extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        toZigbee: [tz.ikea_bulb_unfreeze],
```
bobrippling, line 214: This could specify different unfreeze mechanisms in the future (as in the [original code](https://github.com/Koenkk/zigbee2mqtt/issues/18574#issuecomment-1680923093)), but I only have this one bulb type to test with, so I'm just changing this for now

src/lib/ikea.ts
```diff
@@ -812,3 +813,66 @@ export function addCustomClusterManuSpecificIkeaUnknown(): ModernExtend {
...
+    constructor(entity: Endpoint | Group, mechanism: UnfreezeMechanism) {
+        this.entity = entity;
+        this.unfreeze = mechanism;
+    }
+
+    async command(clusterKey: string, commandKey: string, payload: KeyValue, options: KeyValue) {
+        const frozenUtil = globalStore.getValue(this.entity, 'frozenUntil');
+        if (frozenUtil != null) {
+            if (Date.now() <= frozenUtil) {
+                console.log('Light is frozen, will attempt to unfreeze');
```
bobrippling, line 864: This will be changed to a proper logger call

src/converters/toZigbee.ts
bobrippling: Moved

src/devices/ikea.ts
bobrippling: Hardcoded unfreeze mechanism, see PR description for details

src/lib/ikea.ts
```diff
@@ -825,3 +838,87 @@ export function addCustomClusterManuSpecificIkeaUnknown(): ModernExtend {
...
+    payload.movemode !== 0;
+
+// keys for whom we need to unfreeze the light before issuing a corresponding command to said light
+const keysNeedingUnfreeze = new Set(["state", "brightness", "brightness_percent", "color_temp", "color_temp_percent"]);
+
+const ikea_bulb_unfreeze = (next: Tz.Converter["convertSet"]) => {
+    const converter: Tz.Converter["convertSet"] = async (entity, key, value, meta) => {
+        if (!keysNeedingUnfreeze.has(key)) {
+            return await next(entity, key, value, meta);
+        }
```
bobrippling, line 881: @lbschenkel I'm not sure about this change I've introduced - I guess we want to perform the unfreeze regardless of the `key`, perhaps I should just drop this `if`.

Similarly, maybe I should unconditionally wrap all the `convertSet`s, instead of just ones mentioning specific keys? i.e. remove this condition:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/8ae9fc7cc23984ba829a053de8818c07771b0260/src/lib/ikea.ts#L97-L106

bobrippling:
This will also resolve home-assistant/core#85535, also possibly related: basnijholt/adaptive-lighting#49

src/lib/ikea.ts
```diff
@@ -825,3 +838,87 @@ export function addCustomClusterManuSpecificIkeaUnknown(): ModernExtend {
...
+    (payload.transtime as number) > 0 &&
+    clusterKey === "lightingColorCtrl" &&
+    // ...except for 'stop' commands:
+    payload.rate !== 1 &&
+    payload.movemode !== 0;
+
+// keys for whom we need to unfreeze the light before issuing a corresponding command to said light
+const keysNeedingUnfreeze = new Set(["state", "brightness", "brightness_percent", "color_temp", "color_temp_percent"]);
+
+const ikea_bulb_unfreeze = (next: Tz.Converter["convertSet"]) => {
```
Koenkk, line 877: Could you explain (very short) what the problem is and why this solution fixes it, also add a link to the issue (for future reference)

src/lib/ikea.ts
Koenkk: > I guess we want to perform the unfreeze regardless of the key, perhaps I should just drop this if.
I guess yes, given this is only called for converters which have a key in `keysNeedingUnfreeze`

> Similarly, maybe I should unconditionally wrap all the convertSets, instead of just ones mentioning specific keys? i.e. remove this condition:

I think this is fine for now

src/lib/ikea.ts
```diff
@@ -825,3 +838,87 @@ export function addCustomClusterManuSpecificIkeaUnknown(): ModernExtend {
...
+    const converter: Tz.Converter["convertSet"] = async (entity, key, value, meta) => {
+        if (!keysNeedingUnfreeze.has(key)) {
+            return await next(entity, key, value, meta);
+        }
+
+        const id = "deviceIeeeAddress" in entity ? entity.deviceIeeeAddress : entity.groupID;
+
+        const proxyCommand: Zh.Endpoint["command"] = async (clusterKey, commandKey, payload, options) => {
```
Koenkk, line 885: Can't we extend/override the toZigbee converter instead of changing the `command` here? I think that will be a more maintainable solution

src/lib/ikea.ts
```diff
@@ -93,6 +93,19 @@ export function ikeaLight(args?: Omit<m.LightArgs, "colorTemp"> & {colorTemp?: t
         result.exposes.push(presets.light_color_options());
     }
 
+    if (result.toZigbee) {
```
Koenkk, line 96: Given this will be applied to every IKEA bulb, are we sure that every IKEA bulb is affected by this? (even the newer ones with recent firmwares?)

src/lib/ikea.ts
lbschenkel: I am on a trip overseas, but I can look at the PR in 2 weeks or so (if it still makes sense).

bobrippling: Thanks - I'll leave it as-is, if this does get merged before then, we can always remove it in a future PR, which will appear to users as adding functionality

src/lib/ikea.ts
bobrippling: Yes I can't be certain - I'll limit it to the bulbs I can confirm with and those already confirmed in Koenkk/zigbee2mqtt#18574

protyposis: I tested/used quite a range of Ikea bulbs and contributed workarounds for Ikea bulbs to https://github.com/basnijholt/adaptive-lighting. Currently, I have the following bulbs in use and can confirm they are affected:

LED2003G10
LED1836G9
LED1949C5
LED1537R6/LED1739R5
LED1732G11
LED1835C6
LED1624G9
LED1733G7
LED1936G5
LED2004G8
LED1545G12

It seems likely that all models sold until 2023 are affected (haven't bought any newer models).

src/lib/ikea.ts
bobrippling: I agree it'd be more maintainable - the issue is that we want to detect when we're sending a `lightingColorCtrl` as the cluster key, which I don't think we can do at the toZigbee level (I may be wrong)

Happy to go with your request still, in which case we would pessimise the code and assume keys such as `color_temp`, `brightness` will generate `lightingColorCtrl`, and use those (instead of `lightingColorCtrl`) to detect if a bulb will be frozen

src/lib/ikea.ts
bobrippling: Sure - done

src/lib/ikea.ts
bobrippling: Thanks for the list, I also found it on LED2035G10 - I've opted these bulbs into unfreeze handling in 68695ef

src/lib/ikea.ts
Koenkk: `color_temp` will always generate a `lightingColorCtrl` right? `brightness` will always be `genLevelCtrl`, note that `state` can also sometimes generate a `genLevelCtrl` (in case a default transition is specified for the bulb)

src/lib/ikea.ts
lbschenkel: Chiming in: I have a *lot* of IKEA lights of different generations and sockets. All of them with latest firmware. Every single one of them have this bug. I'm convinced that all models are affected.

What I suggest is not to whitelist or blacklist any lights for now, but apply to all RGB IKEA lights. 

That said, as this functionality is somewhat "intrusive" and there is a fair chance of negatively affecting somebody's workflow, that the whole unfreeze logic be gated by a toggle. I'm neutral if the toggle should be on or off by default, but there should be a way to opt out of this workaround — just to be on the safe side, especially if a firmware change happens to improve things (not holding my breath though,).

I'm traveling without a laptop so I haven't inspected the diff, if a toggle is already implemented then just carry on... :-)

bobrippling: I've changed this to apply to all RGB IKEA lights - with the option to opt out too (using `meta.options`)

Except for one bulb - @protyposis, does `LED1836G9` definitely have the bug? It looks like it's configured as a non-color bulb:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/96efe558ebe43ba52a70a2a049d157a27cad15df/src/devices/ikea.ts#L176-L179

src/lib/ikea.ts
bobrippling: Sure, I've updated to go off those keys instead: cdd887ff5f2 (the large hunk is mostly whitespace)

src/lib/ikea.ts
lbschenkel: What this means in practice is that you may get away with doing inconditionally for all IKEA lights. The lights not affected are unlikely to be sent a command to change temperature or color with a transition, as they don't support it anyway and this capability is not advertised. Then the workaround will never trigger. This may result in simpler code which is easier to maintain.

protyposis: @bobrippling sorry, I forgot to clear `LED1836G9` from the list. It's not affected due to being single-color.

All other bulbs in my list are white-spectrum bulbs, except one which is RGB. All of them are affected.

I second what @lbschenkel suggested. It seems likely that all WS and RGB lights are affected, and kind of unlikely that Ikea is going to push a fix. They've already had a few years to do that, if it had been of any relevance to them (I'm still wondering why their Tradfri/Dirigera system seems unaffected).

bobrippling: No worries, I agree with you both.

To summarise:
- we think it's unlikely that we'll see a fix from IKEA
- we know that the workaround here is harmless for bulbs without a `color_temp` so a blanket application is ok
- because we have the option to turn this off, it's quite safe to deploy (assuming I've got the [option code](https://github.com/Koenkk/zigbee-herdsman-converters/commit/2403943073169ba858cfdcc93d5b3adf9d8e18f0) correct)

@Koenkk I'll revert to applying the unfreeze fix to `ikeaLight()` unconditionally if this is ok with you?

Koenkk: > I'll revert to applying the unfreeze fix to ikeaLight() unconditionally if this is ok with you?

Yes!

lbschenkel: > we know that the workaround here is harmless for bulbs without a color_temp so a blanket application is ok

Nitpick: the bug manifests with both color temperature commands and RGB/CIE/XYZ color change commands, but I guess that's what you meant.

lbschenkel:
Amazing job, @bobrippling!

src/lib/ikea.ts
bobrippling: Ah yes I think so, good point. I think the code is ok now in fact, we apply the fix for any light with color parameters too

bobrippling:
Thanks for the merge and thanks for the original work @lbschenkel!

LarsStegman:
Thank you so much @bobrippling the missing transitions have been annoying me for years!

protyposis:
Been testing this for two days now, and so far it works great, no more freezes. 🙌 Thanks for taking the time to contribute this!

Just a heads-up: while this fixes stuck transitions in general, Ikea bulbs still don’t properly support simultaneous color and brightness transitions (the brightness immediately jumps to the target value). So, while this allows deactivating unfreeze workarounds that simulate long transitions through a series of small, instantaneous change commands, the workaround that executes color and brightness transitions in series is still needed for visual smoothness.

I just spent quite some time debugging this, only to realize (again) that Ikea bulbs don’t support simultaneous brightness and color transitions... The upside is that I didn’t find any issues with this unfreeze workaround. :)

lbschenkel:
> Ikea bulbs still don’t properly support simultaneous color and brightness transitions (the brightness immediately jumps to the target value)

Let me clarify a bit:

There is **no** actual Zigbee command at the protocol (cluster) layer that can change both the brightness and color at the same time — you can do that in a roundabout way in one command only if you're recalling a scene). So there are always two commands involved, one to change brightness and one to change color/temperature. As this is a Zigbee limitation, it applies to everything that speaks Zigbee, be it Z2M or something else.

What happens with most lights is that they will get one command with a transition, then they will get another command (could be the same or different transition), and while they're processing the first they start "interpolating" both changes and doing them simultaneously. How well they interpolate it's up to the firmware. They could even process them in sequence, even though I never saw a light doing that.

Now we can talk about IKEA. With IKEA, the moment it gets the 2nd command, the 1st one is *cancelled*. Basically the 2nd command overwrites the 1st. This is yet another bug/shortcoming in the firmware.

The behaviour you're seeing is the result of yet another Z2M workaround: it knows the lights have this bug, so if it sees an incoming command with both brightness+color+transition, it will send the brightness command first with a transition of 0 (so it finishes immediately) and then the color+transition command. So this is not really due to IKEA, but due to the workaround done by Z2M to make the IKEA lamps behave less horribly.

(I have ideas on how to improve on this other workaround too, but that's a discussion for another day.)

protyposis:
> if it sees an incoming command with both brightness+color+transition, it will send the brightness command first with a transition of 0 (so it finishes immediately) and then the color+transition command

Thanks for the elaboration, I wasn’t aware of this detail.

> I have ideas on how to improve on this other workaround too, but that's a discussion for another day.

It would be great to discuss that, as this seems to be the last missing piece for a proper IKEA bulb experience, IMO. Don't want to hijack this PR either, so how about opening an issue or discussion for it?

For turn-on transitions, it might make sense to use `color_options.execute_if_off` to set the target color first, then transition to the target brightness... effectively skipping the color transition entirely, since fading from the previous to the new color at turn-on can look odd. For brightness+color+transition commands while the bulb is already on, do you have a better approach than splitting them into two consecutive transitions (one for color and one for brightness)?

lbschenkel:
> It would be great to discuss that, as this seems to be the last missing piece for a proper IKEA bulb experience, IMO. Don't want to hijack this PR either, so how about opening an issue or discussion for it?

I agree, please do open a [discussion](https://github.com/Koenkk/zigbee2mqtt/discussions) and cc me and we'll take it from there.

bobrippling:
Glad to hear it's been working alright - I've not found any problems either. If you cc me in the discussion too, I'm wondering if we can somehow use a scene on the IKEA bulbs to convince them into doing a brightness and temperature transition at the same time, going off:

> There is **no** [...] command [...] that can change both the brightness and color at the same time (you can do that in a roundabout way in one command only if you're recalling a scene)

Nice! I've experimented with this and we can get brightness and colour changes simultaneously:

```sh
#!/bin/sh

dev='<todo>'; echo "fill in \$dev"; exit 1
t=zigbee2mqtt/$dev/set
scene_name=tempbright_test

scene_add(){
  mosquitto_pub -t "$t" -m '{
      "scene_add": {
          "ID": 1,
          "name": "'"$scene_name"'",
          "transition": 5,

          "state": "ON",
          "brightness": 254,
          "color_temp": 454
      }
  }
  '
}

scene_recall(){
  mosquitto_pub -t "$t" -m '{ "scene_recall": 1 }'
}

case "$1" in
  add) scene_add ;;
  recall) scene_recall ;;
  *) exit 2 ;;
esac
```

So we'd need a translation layer that detects requests for transitions with brightness and temperature, create a temporary scene, use it to fulfil the request, then delete the scene (or leave the scene and overwrite on the next request)

lbschenkel:
Let us please continue this off-topic discussion in https://github.com/Koenkk/zigbee2mqtt/discussions/27325

PR #8638, opened 2025-01-18
Support _TZE200_jkfbph7l (AVATTO ME167)

jemx:
I bought six of these AVATTO ME167 thermostatic radiator valve, two of them came with a different vendor id (visually the same device). Here is the support for the devices with the `_TZE200_jkfbph7l` id.

Successful locally tested via an external converter.


Koenkk:
Thanks!

PR #8640, opened 2025-01-18
Add support for https://de.aliexpress.com/item/1005008269272176.html

oli-z:
Hi,

I recently bought this energy meter, which wasn't supported yet, so I tried to add support. The basics (voltage, current, power) work, but there are still some Tuya datapoints that are unmapped. I will try to figure out what they do after installing the device and maybe create another PR to add support for them later.

First time contributing to this project so I hope I did everything right 😅

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -14204,7 +14204,6 @@ const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-
         fingerprint: tuya.fingerprint('TS0601', ['_TZE204_goecjd1t']),
         model: 'TS0601_smart_energy_meter',
         vendor: 'Tuya',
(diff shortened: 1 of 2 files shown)


Koenkk:
Could you check if it works with the definition from PJ-MGW1203 ? That also exposes energy

oli-z:
> Could you check if it works with the definition from PJ-MGW1203 ? That also exposes energy

Seems like this one uses a different data point for energy than mine (101, which my device doesn't seem to have). It looks like 104 and 105 are the energy datapoints for mine, but they are offset:

image

Maybe one is the daily energy and one the total energy, I'll keep an eye on it... I also have datapoints 103, 106, 107 and 109 but no idea what they are doing. Some go up and down if the power goes up and down. Unfortunately, I don't have a Tuya ZigBee gateway, so I can't look there.

src/devices/tuya.ts
```diff
@@ -14203,6 +14203,25 @@ const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE204_goecjd1t']),
+        model: 'TS0601_smart_energy_meter',
```
Koenkk, line 14208:
```suggestion
        model: 'ZWPM16',
```

src/devices/tuya.ts
```diff
@@ -14203,6 +14203,25 @@ const definitions: DefinitionWithExtend[] = [
...
+        vendor: 'Tuya',
```
Koenkk, line 14209:
```suggestion
        vendor: 'AVATTO',
```

src/devices/tuya.ts
```diff
@@ -14203,6 +14203,25 @@ const definitions: DefinitionWithExtend[] = [
...
+        configure: tuya.configureMagicPacket,
+        exposes: [e.power(), e.voltage(), e.current()],
+        meta: {
+            tuyaDatapoints: [
+                [18, 'current', tuya.valueConverter.divideBy1000],
+                [19, 'power', tuya.valueConverter.divideBy10],
+                [20, 'voltage', tuya.valueConverter.divideBy10],
+            ],
+        },
+        whiteLabel: [{vendor: 'AVATTO', model: 'ZWPM16'}],
```
Koenkk, line 14223: This can be removed.

PR #8644, opened 2025-01-19
Fix CO2 value interpretation

Eyenseo:
The zigbee standard specifies that the values of section 4.13 Concentration Measurement are to be represented between 0 and 1 (section 4.13.2.1 Attributes). Further it is explained in section 4.13.2.1.1 MeasuredValue Attribute that "MeasuredValue represents the concentration as a fraction of 1 (one)."

Because the current implementation multiplies the given MeasuredValue by 1000000 (million) the published value is not only wrong, but changes in the inverse direction. The larger the actual measured value becomes (the larger the denominator is), the smaller MeasuredValue becomes.

| Device (PPM) | MeasuredValue | z2m (PPM) |
|--------------|---------------|-----------|
| 800          | 0.00125       | 1250      |
| 1250         | 0.00080       | 800       |

- [x] I have tried this change as external converter with my own co2 zigbee device.
- [ ] Dependent changes
   - `modernExtend.ts` seems to make similar assumptions too, but I do not know how / when `scale` and `change` are used - before or after the conversion took place.
   - some other place?
- [ ] Figure out if some existing devices actually implement MeasuredValue in the current way. And provide device specific workarounds.
   - I would need help with that, as I own no other co2 sensor.

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -503,7 +503,7 @@ const converters1 = {
         cluster: 'msCO2',
         type: ['attributeReport', 'readResponse'],
         convert: (model, msg, publish, options, meta) => {
-            return {co2: Math.floor(msg.data.measuredValue * 1000000)};
+            return {co2: Math.floor(1 / msg.data.measuredValue)};
         },
     } satisfies Fz.Converter,
     occupancy: {
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -941,15 +941,18 @@ export function occupancy(args?: OccupancyArgs): ModernExtend {
 }
 
 export function co2(args?: Partial<NumericArgs>) {
+    const fraction_of_1: ScaleFunction = (value: number, type: 'from' | 'to') => {
+        return 1 / value;
+    };
     return numeric({
         name: 'co2',
         cluster: 'msCO2',
         label: 'CO2',
         attribute: 'measuredValue',
-        reporting: {min: '10_SECONDS', max: '1_HOUR', change: 0.00005}, // 50 ppm change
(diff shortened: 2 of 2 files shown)


Koenkk:
> modernExtend.ts seems to make similar assumptions too, but I do not know how / when scale and change are used - before or after the conversion took place.

What device do you have? Maybe it's best to also refactor that to modernExtend.

Eyenseo:
DIY project based on nrf52840/zephyr and scd41.

I would like to use modernExtend - I just don't know how to do that / where the calculations take place. But I have also not searched for documentation too much. My guess was that they somehow still make use of the function that I modified, but based on your answer that is not correct, which would mean that they convert via `scale`?
I wanted to get the PR out as it took me some head scratching why the values where weird and I was hoping to get some help from the community to figure out what to do with this change.

Koenkk:
Using modern extend should be easy, see https://github.com/Koenkk/zigbee-herdsman-converters/blob/cdb1b50df2552594cd554dcd3226b8ed3a506d77/src/devices/diyruz.ts#L325 as an example.

The conversion is done here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/cdb1b50df2552594cd554dcd3226b8ed3a506d77/src/lib/modernExtend.ts#L2177

Eyenseo:
I'm unsure about `reporting: {min: '10_SECONDS', max: '1_HOUR', change: 0.00005}, // 50 ppm change`. I guess `change` is used before `scale` is applied and is usually used in a way of `abs(previous-new) > change`?

Koenkk:
Thanks!

BOPOHOP:
https://www.zigbee2mqtt.io/devices/DIYRuZ_AirSense.html becomes broken due to the change in calculation in fromZigbee.js

AlexMKX:
The same here.
all PPMs are now 400->2000 and so forth.

PR #8649, opened 2025-01-19
Add reporting for SLR1d

harveybolton:
Added missing reporting endpoints for Hive SLR1d

### DIFF
--- a/src/devices/hive.ts
+++ b/src/devices/hive.ts
@@ -372,6 +372,18 @@ const definitions: DefinitionWithExtend[] = [
                         ' used. 0 to 360 to match the remote display',
                 ),
         ],
+        meta: {disableDefaultResponse: true},
+        configure: async (device, coordinatorEndpoint) => {
+            const endpoint = device.getEndpoint(5);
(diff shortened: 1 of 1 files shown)


src/devices/hive.ts
```diff
@@ -372,6 +372,18 @@ const definitions: DefinitionWithExtend[] = [
                         ' used. 0 to 360 to match the remote display',
                 ),
         ],
+        meta: {disableDefaultResponse: true},
```
Koenkk, line 375: Is this really needed?

providentitsolutions: It's present in SLR1c the block above, I built with that and it worked

harveybolton: It's present in SLR1c the block above, I built with that and it worked

Koenkk: Can you try if it works without it?

si458:
@Koenkk please can this PR be reopened and merged?
I've had my old SLR1b replaced with a d version as it died and its completely missing the reporting but the model is infact the same as b and c models

I had to add them in manually in Z2M using this post on HA forums instead for the time being
https://community.home-assistant.io/t/hive-receiver-slr1d-does-not-publish-state-changes-to-z2mqtt/801090

PR #8651, opened 2025-01-19
feat(add): Avatto ME168

andreypuhovsky:
Following my previous pull request for this device, I've purchased a Tuya Gateway to make it 99% functional :)

image
image
image
image
image
image

tuya:
IMG_5826
IMG_5825
IMG_5824
IMG_5823

Additional things:
- fixes for max temperature (strange to not being able to set up up to 35 degrees)
- fix the all days schedule description to be more helpful (new format allows to copy-paste and modify for user's needs)


andreypuhovsky:
unfortunately I'm unable to run prettier to find the issues. if anyone will show me them I'll fix them asap :)

src/devices/tuya.ts
```diff
@@ -5517,6 +5516,153 @@ const definitions: DefinitionWithExtend[] = [
             await reporting.bind(endpoint, coordinatorEndpoint, ['genOnOff']);
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_ybsqljjg' /* model: 'ME168', vendor: 'AVATTO' */]),
+        model: 'TS0601_thermostat_5',
```
Koenkk, line 5521: Since there is only 1 whitelabel, use `ME168` here

src/devices/tuya.ts
```diff
@@ -5517,6 +5516,153 @@ const definitions: DefinitionWithExtend[] = [
...
+                .binary('scale_protection', ea.STATE_SET, 'ON', 'OFF')
+                .withDescription(
+                    'If the heat sink is not fully opened within ' +
+                        'two weeks or is not used for a long time, the valve will be blocked due to silting up and the heat sink will not be ' +
+                        'able to be used. To ensure normal use of the heat sink, the controller will automatically open the valve fully every ' +
+                        'two weeks. It will run for 30 seconds per time with the screen displaying "Ad", then return to its normal working state ' +
+                        'again.',
+                )
+                .withCategory('config'),
+
```
Koenkk, line 5571: Could you remove the empty lines here, same goes for `tuyaDatapoints`

src/lib/tuya.ts
```diff
@@ -1497,6 +1497,37 @@ export const valueConverter = {
             return data;
         },
     },
+    thermostatME168_systemModeAndPreset: (toKey: string) => {
```
Koenkk, line 1500: Could you re-use `thermostatGtz10SystemModeAndPreset` here? Maybe call it `thermostatSystemModeAndPreset` and pass the lookup via a parameter.

andreypuhovsky: we will need to pass 2-3 different maps (1 for `from` and 1-2 for `to` lookup) from the device definition.
is that acceptable?

src/lib/tuya.ts
```diff
@@ -1497,6 +1497,37 @@ export const valueConverter = {
...
+        return {
+            from: (v: string) => {
+                utils.assertNumber(v, 'system_mode');
+                const modeMap: {[mode: number]: {name: string; system_mode: string; preset: string}} = {
+                    0: {name: 'auto', system_mode: 'auto', preset: 'none'},
+                    1: {name: 'manual', system_mode: 'heat', preset: 'none'},
+                    2: {name: 'off', system_mode: 'off', preset: 'none'},
+                    3: {name: 'eco', system_mode: 'heat', preset: 'eco'},
+                    4: {name: 'comfort', system_mode: 'heat', preset: 'comfort'},
+                    5: {name: 'rapid', system_mode: 'heat', preset: 'boost'},
```
Koenkk, line 1510: Is the `rapid` preset not settable?

andreypuhovsky: rapid is the name in terms of tuya api, the z2m preset is boost

src/lib/tuya.ts
Koenkk: Yes

andreypuhovsky: agree. I think my implementation is a little bit better as it provides visibility to the real device mode (useful for diagnostics & debugging). 
however, I have 2 other thermostats here and will need to check how they implement different modes (such as holiday, comfort and so on)

src/lib/tuya.ts
Koenkk: Then I think `boost` makes more sense here (to keep it inline with other devices)

src/devices/tuya.ts
andreypuhovsky: unfortunately, I think it's not possible to use the actual model as it was white labeled somewhere else

```
 FAIL  test/index.test.js > index.js > Verify definitions
Error: Duplicate whitelabel zigbee model ME168
```

```ts
        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_ybsqljjg']),
        model: 'ME168',
        vendor: 'AVATTO',

```

andreypuhovsky: named as AVATTO_ME168 for now, let's discuss this further

src/lib/tuya.ts
andreypuhovsky: done

src/lib/tuya.ts
andreypuhovsky: done. gtz10 migrated, old function marked as deprecated

src/devices/tuya.ts
andreypuhovsky: done

Koenkk:
Thanks!

PR #8655, opened 2025-01-20
Custom system/preset mode mapping for TRV602Z

swiergot:
The default value converter `thermostatSystemModeAndPreset` is not suitable for this valve. Currently system mode "heat" changes preset mode to "comfort" while system mode "off" changes preset mode to "eco".

Please do not merge this PR yet. For some reason with these changes the valve will not respond to changing system or preset modes at all. I have zero past experience with this so I hope @Koenkk or @SteveOswald (as you made changes to this valve's support recently) can see where I made a mistake. At the moment I'm out of ideas. Nothing in the Z2M logs even with the debug level.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -5216,7 +5216,7 @@ const definitions: DefinitionWithExtend[] = [
                         on: tuya.enum(5),
                     }),
                 ],
-                [2, 'system_mode', tuya.valueConverter.thermostatSystemModeAndPreset('system_mode')],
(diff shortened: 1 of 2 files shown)


swiergot:
@Koenkk I think I've got the converter working. The key was adding a null-name datapoint, a few other thermostats do it for some reason. Does that make sense to you?

Koenkk:
That's fine but can we reduce code duplication by merging it with `thermostatGtz10SystemModeAndPreset` (all differences should be passed via a parameter)

PR #8658, opened 2025-01-22
Added TS0601_zigbee_3_phase_meter with all datapoints from tuya cloud

oooeeee:
Device link: https://aliexpress.com/item/1005007890988611.html

screenshots after addition:
screen_after
screen_exposes_after

database.db
```
{"id":20,"type":"Router","ieeeAddr":"0x842712fffe1fc410","nwkAddr":65091,"manufId":4098,"manufName":"_TZE284_a14rjslz","powerSource":"Mains (single phase)","modelId":"TS0601","epList":[1],"endpoints":{"1":{"profId":260,"epId":1,"devId":81,"inClusterList":[0,4,5,61184,0,60672],"outClusterList":[25,10],"clusters":{"genBasic":{"attributes":{"65503":"\u0011�#/i\u0011�#/i\u0012�#/i\u0012�#/i\u0013�#/i\u0013�#/iˡ#/i#/i","65506":31,"65508":0,"65534":0,"modelId":"TS0601","manufacturerName":"_TZE284_a14rjslz","stackVersion":0,"dateCode":"","zclVersion":3,"appVersion":74,"powerSource":1}}},"binds":[],"configuredReportings":[],"meta":{}}},"appVersion":74,"stackVersion":0,"hwVersion":1,"dateCode":"","zclVersion":3,"interviewCompleted":true,"meta":{"configured":217189513},"lastSeen":1737538880024}
```

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -14042,6 +14042,49 @@ const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint('TS0601', ['_TZE284_a14rjslz']),
+        model: 'TS0601_zigbee_3_phase_meter',
+        vendor: 'Tuya',
+        description: 'Zigbee 3 Phase Meter',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        whiteLabel: [
+            {vendor: 'Ourtop', model: 'ATMS10013Z3'},
+            {vendor: 'Ourtop', model: 'ATMS100133Z'},
+        ],
+        exposes: [
+            e.energy(),
+            tuya.exposes.voltageWithPhase('a'),
+            tuya.exposes.voltageWithPhase('b'),
+            tuya.exposes.voltageWithPhase('c'),
+            tuya.exposes.powerWithPhase('a'),
+            tuya.exposes.powerWithPhase('b'),
+            tuya.exposes.powerWithPhase('c'),
+            tuya.exposes.currentWithPhase('a'),
+            tuya.exposes.currentWithPhase('b'),
+            tuya.exposes.currentWithPhase('c'),
+            e.produced_energy(),
+            e.numeric('active_enery_total', ea.STATE).withUnit('kWh').withDescription('Total Active Energy'),
+            e.power(),
+            e.power_reactive(),
+            e.power_factor().withUnit('%'),
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [1, 'energy', tuya.valueConverter.divideBy100],  // "Forward Active Energy", 7685 -> 76.85kW
+                [6, null, tuya.valueConverter.phaseVariant2WithPhase('a')],  // "Phase A", CJMAAEMAAAI=
+                [7, null, tuya.valueConverter.phaseVariant2WithPhase('b')],  // "Phase B"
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -14042,6 +14042,49 @@ const definitions: DefinitionWithExtend[] = [
...
+            tuya.exposes.voltageWithPhase('b'),
+            tuya.exposes.voltageWithPhase('c'),
+            tuya.exposes.powerWithPhase('a'),
+            tuya.exposes.powerWithPhase('b'),
+            tuya.exposes.powerWithPhase('c'),
+            tuya.exposes.currentWithPhase('a'),
+            tuya.exposes.currentWithPhase('b'),
+            tuya.exposes.currentWithPhase('c'),
+            e.produced_energy(),
+            e.numeric('active_enery_total', ea.STATE).withUnit('kWh').withDescription('Total Active Energy'),
```
Koenkk, line 14069: Could you explain a bit more what this is?

oooeeee: - fixed typo in `enery` -> `energy`
- in physical terms, this represents the sum of consumed and produced energy. Since I don't generate any energy to send to the grid, it is effectively equal to the consumed energy alone. The term "Active Energy Total" comes directly from the datapoint in the Tuya Cloud platform, and I prefer to keep it as-is
- what means `ea.STATE` I don't know, sorry. But I copied it from code above

изображение

oooeeee: Also sorry, I don't understand what about you asking for. In github interface question assigned only to line with typo, but in email starting from `model`.
Models a bit strange too because:
- on seller page model looks like `...13Z3`, and in real life the same
- but in tuya cloud platform visible model is `...133Z`

So I decided to keep both of them

изображение

Koenkk: Could you use `ATM...` as the model then in the definition?

Koenkk: > in physical terms, this represents the sum of consumed and produced energy. Since I don't generate any energy to send to the grid, it is effectively equal to the consumed energy alone. The term "Active Energy Total" comes directly from the datapoint in the Tuya Cloud platform, and I prefer to keep it as-is

In another definition we call this `total_energy`, I think it's better to use that name (https://github.com/Koenkk/zigbee-herdsman-converters/blob/29dca37953258c80d10d422f72e892c6098e34d3/src/devices/tuya.ts#L5824)

oooeeee: - changed `active_energy_total` -> `total_energy`
- changed vendor and model as written on real device (to ourtop + model)

oooeeee: and looks like after merge of this code I need to rename file from https://github.com/Koenkk/zigbee2mqtt.io/pull/3448

Koenkk:
Thanks!

leuhinev:
hello, @Koenkk! I am testing such a device at home. I come across a problem
error: 	zhc: Failed to apply calibration to 'power': 'power_calibration' is not a number, got string ()

oooeeee:
I have no any calibration inside this device in tuya, so I did not exported
any endpoint to calibrate

чт, 30 янв. 2025 г., 21:04 Evgeniy ***@***.***>:

src/devices/tuya.ts
Lunden0: Hello, is there a guide on how to setup this device in Z2M?

oooeeee:
For me enough just to enable pairing mode on the device (pressing button
for five seconds) and allow connection of new devices in z2m

сб, 8 мар. 2025 г., 11:41 Lunden0 ***@***.***>:

PR #8661, opened 2025-01-22
TS0601 PIR Sensor (_TZE200_gjldowol)

omerfaruk-aran:
This PR adds support for the TS0601 PIR Sensor (_TZE200_gjldowol).
- Includes luminance, occupancy, battery, and sensitivity settings.

### DIFF
--- a/src/devices/TS0601_pir_sensor.ts
+++ b/src/devices/TS0601_pir_sensor.ts
@@ -1,8 +1,9 @@
-import {Definition} from '../lib/types';
 import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as exposes from '../lib/exposes';
 import * as tuya from '../lib/tuya';
+import {Definition} from '../lib/types';
+
 const e = exposes.presets;
 const ea = exposes.access;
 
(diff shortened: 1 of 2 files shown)


src/devices/TS0601_pir_sensor.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const definition: Definition = {
+    fingerprint: [
+        {
+            modelID: 'TS0601',
+            manufacturerName: '_TZE200_gjldowol',
+        },
+    ],
+    zigbeeModel: ['TS0601'],
+    model: 'TS0601_pir_sensor',
```
Koenkk, line 16: Is there any other model number written on the device or product page?

src/devices/TS0601_pir_sensor.ts
```diff
@@ -0,0 +1,57 @@
+import * as exposes from '../lib/exposes';
```
Koenkk, line 1: Please move this into `tuya.ts` and remove this file.

src/devices/TS0601_pir_sensor.ts
omerfaruk-aran: Thank you for your follow-up. During the Zigbee2MQTT detection process, the following information is returned:

```javascript
const definition = {
    zigbeeModel: ['TS0601'],
    model: 'TS0601',
    vendor: '_TZE200_gjldowol',
    extend: [],
    meta: {},
};

module.exports = definition;
```
However, the model information printed on the device packaging and user manual differs:

1. On the packaging: The model is listed as `IH012-RT01`.
2. In the user manual: The model is listed as `ZM-35ZH-Q`.

This discrepancy makes it difficult to ensure accurate integration with Zigbee2MQTT. Could this be due to vendor-specific modifications to the firmware or a generic implementation for TS0601-based devices?

IMG_5176
IMG_5177

Koenkk: Let's use `ZM-35ZH-Q` then, `_TZE200_gjldowol` is the internal model of the device.

PR #8662, opened 2025-01-23
feat: support SR-ZG9070A-SS from heiman to sunricher and fix correct zone types for sensors

niracler:


src/devices/sunricher.ts
```diff
@@ -407,43 +407,54 @@ async function syncTime(endpoint: Zh.Endpoint) {
...
+        description: 'Smart photoelectric smoke alarm',
+        extend: [m.battery(), m.iasZoneAlarm({zoneType: 'smoke', zoneAttributes: ['alarm_1', 'alarm_2', 'tamper', 'battery_low']}), m.iasWarning()],
+    },
     {
         zigbeeModel: ['HK-SENSOR-PRE'],
         model: 'SR-ZG9030F-PS',
         vendor: 'Sunricher',
         description: 'Smart human presence sensor',
-        extend: [m.illuminance(), m.occupancy(), m.commandsOnOff()],
+        extend: [m.illuminance({scale: (value) => value}), m.occupancy(), m.commandsOnOff()],
```
Koenkk, line 422: What kind of illuminance values does it provide after this change?

niracler: By default, Zigbee devices following the standard apply a logarithmic transformation to the raw illuminance value, as specified by the Illuminance Measurement Cluster. The formula used is:  

```
MeasuredValue = 10000 × log10(illuminance [lux]) + 1
```

However, this device does not follow that standard. It reports the `MeasuredValue` directly as the actual illuminance in lux without any transformation. Therefore, the value can be used as-is without additional scaling or conversion.

Koenkk:
Thanks!

PR #8676, opened 2025-01-25
SmartThings IM6001-MPP01: change constants.repInterval to 1 minute from 5 hours

natefelt:
Accelerometer sensors for SmartThings IM6001-MPP01 were all reporting NULL. I adjusted constants.repInterval to 1 minute from 5 hours which is how the other SmartThings devices with accelerometers are coded. Tested as external converter on my own setup and can confirm movement is reported properly.


Koenkk:
Thanks!

spry-salt:
I would be so grateful if this could be addressed.

PR #8679, opened 2025-01-26
Fix missing mqtt action messages for Legrand Wireless Shutter switch (0067646)

nicpullan:
Update Legrand Wireless Shutter switch (0067646)
Same fix as Legrand Wireless switches 067773/067774 as per [issue 14256](https://github.com/Koenkk/zigbee2mqtt/issues/14256 )
Tested as external converter and confirmed working


Koenkk:
thanks!

PR #8681, opened 2025-01-27
Invert direction for Smartwings shades

Korri:
Inspired by ZHA's approach, and feedback form SmartWings support, I confirmed that this converter works and am now proposing that we merge it to the main codebase.

For shades to work well, you still need to invert their Zigbee state (press Up & Down together for 5s, then Press the P button under the battery cover once).

Fixes https://github.com/koenkk/zigbee2mqtt/issues/22847
The above issue is where I got the external converter code from, I slightly adjusted it to be more consistent with the code in this repository, but logic should be identical.

ZHA's pull request for those shades has a lot of investigation and context on them: https://github.com/zigpy/zha-device-handlers/pull/2220

### DIFF
--- a/src/devices/smartwings.ts
+++ b/src/devices/smartwings.ts
@@ -5,13 +5,15 @@ import * as reporting from '../lib/reporting';
 import {DefinitionWithExtend, Tz, Zh} from '../lib/types';
 import {assertString, getFromLookup, getOptions} from '../lib/utils';
 
-const backwards_cover_state = {
-    key: ['state'],
-    convertSet: async (entity: Zh.Endpoint, key: string, value: number | string, meta: Tz.Meta) => {
-        const lookup = {open: 'downClose', close: 'upOpen', stop: 'stop', on: 'downClose', off: 'upOpen'};
-        assertString(value, key);
-        value = value.toLowerCase();
-        await entity.command('closuresWindowCovering', getFromLookup(value, lookup), {}, getOptions(meta.mapped, entity));
+const tzLocal = {
+    backwards_cover_state: {
+        key: ['state'],
+        convertSet: async (entity: Zh.Endpoint, key: string, value: number | string, meta: Tz.Meta) => {
+            const lookup = {open: 'downClose', close: 'upOpen', stop: 'stop', on: 'downClose', off: 'upOpen'};
+            assertString(value, key);
+            value = value.toLowerCase();
+            await entity.command('closuresWindowCovering', getFromLookup(value, lookup), {}, getOptions(meta.mapped, entity));
+        },
     },
 };
 
@@ -24,7 +26,7 @@ const definitions: DefinitionWithExtend[] = [
         vendor: 'Smartwings',
         description: 'Roller shade',
         fromZigbee: [fz.cover_position_tilt, fz.battery],
-        toZigbee: [backwards_cover_state, tz.cover_position_tilt],
+        toZigbee: [tzLocal.backwards_cover_state, tz.cover_position_tilt],
         meta: {battery: {dontDividePercentage: true}, coverInverted: true},
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

src/devices/smartwings.ts
```diff
@@ -13,7 +26,7 @@ const definitions: DefinitionWithExtend[] = [
         vendor: 'Smartwings',
         description: 'Roller shade',
         fromZigbee: [fz.cover_position_tilt, fz.battery],
-        toZigbee: [tz.cover_state, tz.cover_position_tilt],
+        toZigbee: [tzLocal.backwards_cover_state, tz.cover_position_tilt],
         meta: {battery: {dontDividePercentage: true}, coverInverted: true},
```
brettmiller, line 30: After this update depending on if the shade has the motor direction reversed or not the shade either has the state and open/close percentage reversed or the buttons are reversed (close opens the shade, open closes the shade). With the other inversions done in this change should  `coverInverted` also be inverted from `true` to `false` to correct the state?

quadcom:
This merge should have been a UI toggle so that users that have already flipped the Smartwings 'app' direction using the remote can revert this forced change. Everyone that has done this has woken up to broken shades now. This 'fix' is only useful for new adopters of Smartwing shades as they won't have to go through the 'app' direction flip procedure.

brettmiller:
This change has now caused someone to have physical damage done to their blinds. https://github.com/Koenkk/zigbee2mqtt/issues/26133#issuecomment-2631409319. Seem like it should be reverted asap until a better solution can be implemented that won't possibly harm peoples devices and that hopefully doesn't cause everyone already using these shades and blinds to have to factory reset them to have them work correctly and have the remotes function as expected(which is required to get the remotes to function correctly otherwise just switching motor will suffice).

java-devil:
@Koenkk as @brettmiller described this PR in it's current form introduces internally inconsistent behavior, and should be reverted ASAP as it risks causing physical damage until a fix is introduced.

@quadcom this toggle exists under the "Settings (specific)" of the device's tab. Also I don't find this sort of tone particularly helpful in an OSS project.

FYI @Korri

quadcom:
> @quadcom this toggle exists under the "Settings (specific)" of the device's tab. Also it needs to be said this sort of tone particularly helpful in an OSS project.

1. That option is not the 'app' flip. That is a a config that needs to be performed using the smartwings remote and is inaccessible from within Z2M. The setting flip switch, flips the 100% being displayed in either open or closed state. The 'app'  flip changes the behavior of the open/close direction. By default, pressing up on the remote equates to close in apps. The shades 'app' flip changes the action on the shade when the sends the up command (normally close) instead performs an open instead.

The last instruction on adding the shades to HA (or any other app) is to "Once the shade is successfully paired with Home Assistant, follow the steps below to ensure that the control command directions on Home Assistant match those on the remote control." 

https://cdn.shopify.com/s/files/1/0573/0215/5461/files/Zigbee_Motor_Programming_Guide.pdf?v=1729328292

2. What tone are you referring to? I simply pointed out that a change like that should not have been a forced one but an optional one. Inverting a configuration in the code is something that will definitely have an effect on existing users. If that wasn't obvious. 
 
 
PS: That last comment was tone

Koenkk:
Reverted!

Changes will be available in the [dev branch](https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) in a few hours from now and in the next release which is every 1st of the month.

ThrashinVictim:
I am going to add this here too.

I have 4 sets of shades, 1 I bought months before the other 3. 
This change (2.1.0) fixed the 3 shades I had and broke the 1 that worked before. 
So I have a choice to stick with the 2.1.0 and have three that work correctly with Alexa and HA (showing correct state) and have one that doesn't Or go with 2.1.1 and have 3 that show incorrect state and 1 that does.

KendallHopkins:
This commit also fixed my shades. So I'll continue using it via external_converters. I suspect depending on the version of the blinds it will either fix previous broken, or break previously working. Maybe this should be offered as an opt-in option, similar to the "Invert cover" option?

PR #8684, opened 2025-01-27
SNZB-02D - add support for comfort levels and temperature units.

photomoose:
Adds support for setting the temperature and humidity comfort levels, in addition to setting the display temperature units (Celsius/Fahrenheit) on the Sonoff SNZB-02D. Depending on these values, icons are displayed on the device as shown below:

IMG_4358
IMG_4359

The properties are exposed in the Z2M UI as shown below:

image

Note that the button on the back of the device needs to be pressed before changing the values in the Z2M UI. Not entirely sure why (presumably to wake it up to receive commands), the device sends a ZCL "On/Off: Toggle" command when the button is pressed. This is the same behaviour when using the Sonoff Bridge and the eWeLink app (the app waits for the button press before writing attributes to the device).

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -853,7 +853,80 @@ const definitions: DefinitionWithExtend[] = [
         model: 'SNZB-02D',
         vendor: 'SONOFF',
         description: 'Temperature and humidity sensor with screen',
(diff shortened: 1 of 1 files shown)


Koenkk:
Cool, thanks!

PsychoOW:
Amazing!! Thank you!!

lukasjakobi:
Awesome work! Great Feature!

PR #8691, opened 2025-01-29
TYBAC-006: Update system_mode to support "off", dropping "state"

robvanoostenrijk:
Resolves issue https://github.com/Koenkk/zigbee2mqtt/issues/21647, "Home Assistant not able to turn climate device off"


Koenkk:
Thanks again!

PR #8700, opened 2025-01-30
Add additional improvements for Inovelli devices

rohankapoorcom:
In preparation for the new VMZ32-SN, I spent some time fixing some issues with the existing converter as well as moving around some attributes to make the new integration simpler.

1. Made a bunch of the Inovelli attribute fz/tz converters cluster aware (since the new device will use multiple clusters with attributes that need to be exposed).
2. Updated the Inovelli attribute fz converter to handle reading attributes from multiple endpoints (needed for the VZM36 which previously failed). This fixes https://github.com/Koenkk/zigbee2mqtt/issues/22459.
3. Moved all of the dimmable light attributes to a separate object to allow for reuse with VZM32-SN.
4. Renamed `leadingTrailingEdge_1` to `dimmingMode` and made it visible (but readonly on all dimmers) while keeping it read/write on VZM36
5. Added a call at the end of `configure` for each device to read the values of all of the attributes for that device so that they are not blank requiring the user to manually hit refresh (or call the mqtt get api) for each one. Fixes: https://community.inovelli.com/t/null-states-in-z2m/18543

CC: @InovelliUSA

### DIFF
--- a/src/devices/inovelli.ts
+++ b/src/devices/inovelli.ts
@@ -2097,7 +2097,7 @@ const definitions: DefinitionWithExtend[] = [
 
             // Bind for Button Event Reporting
             const endpoint2 = device.getEndpoint(2);
-            await reporting.bind(endpoint2, coordinatorEndpoint, ['manuSpecificInovelli']);
+            await reporting.bind(endpoint2, coordinatorEndpoint, [inovelliCluster]);
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
@InovelliUSA could you review this?

InovelliUSA:
> @InovelliUSA could you review this?

Tested it and everything seems to be working for me.

rohankapoorcom:
Since this wasn't merged yet, I fixed a couple of additional bugs that Eric (@InovelliUSA) reported to me directly with this converter where the breeze mode and fan state where not being read correctly for the VZM36.

Koenkk:
Thanks!

PR #8705, opened 2025-02-01
Yokis : updating device definition

LaurentChardin:
This PR covers the following issues:
- https://github.com/LaurentChardin/zigbee-herdsman-converters/issues/14
- https://github.com/LaurentChardin/zigbee-herdsman-converters/issues/9

it also switch from the default `onOff` modernExtend to a custom one, because Yokis devices do not implement the timer function of the Zigbee specs (they came later in another revision). Deferencing the key in the `toZigbee` ensures wrong documentation is not generated on the website.

### DIFF
--- a/src/devices/yokis.ts
+++ b/src/devices/yokis.ts
@@ -1400,6 +1400,39 @@ const yokisCommandsExtend = {
     },
 };
 
+// Yokis specific definition
+
+// Yokis does not support the timer OnOff cluster. It uses a custom cluster instead.
+// Dereferencing the `on_time` and `off_wait_time` from the keys of the converter.
+const yokisTz = {
+    on_off: {
+        key: ['state'],
+        convertSet: async (entity, key, value, meta) => {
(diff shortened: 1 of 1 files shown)


src/devices/yokis.ts
```diff
@@ -1400,6 +1400,37 @@ const yokisCommandsExtend = {
     },
 };
 
+// Yokis specific definition
+
+// Yokis does not support the timer OnOff cluster. It uses a custom cluster instead.
+// Dereferencing the `on_time` and `off_wait_time` from the keys of the converter.
+const yokisTz = {
+    on_off: {
+        key: ['state'],
```
Koenkk, line 1409:
```suggestion
        key: ['state'],
        ...tz.on_off,
```

the rest of the code can be removed.

LaurentChardin: Updated : i had to put the spread operator first as it was overriding the key.

Koenkk:
Thanks!

PR #8715, opened 2025-02-02
fix: Adapt SP 24x converter for new firmware version

### LINKED ISSUE #8712: Divisors need updating after SP 242 firmware update
### What happened?

Firmware build 20250127 (1.9.28) for the Innr SP 242 has been released but it's changed the divisors.

@GerryInnr reported that:
> Before this firmware version, the device did not provide a multiplier or divisor for CurrentSummationDelivered, it only provided the CurrentSummationFormatting (because multiplier and divisor are optional in Zigbee and CurrentSummationFormatting is mandatory). The new version provides all three, with multiplier = 1 and divisor = 1000. If the converter just sets/overrides the divisor to 100 (which was the right value for the previous firmwares) without checking whether the device already provided a value, then yes, the converter would need fixing for the new firmware version to set it to 1000.

Energy should read 64.39 kWh but is off by a factor of 10 due to the hard coded divisors.
Image

Read the multiplier and divisor from the device (Remove the hard coded override?)

Update firmware, see that the values now report differently because of the hard coded values and the change to the defaults.

### Zigbee2MQTT version

2.1.0

mundschenk-at:
- Removes explicit `divisor`/`multiplier` settings.
- Removes `reconfigureReportingsOnDeviceAnnounce()` call for `SP 242`.
- Adds `SP 242` and `SP 244` as white label devices.
- Fixes #8712.

### DIFF
--- a/src/devices/innr.ts
+++ b/src/devices/innr.ts
@@ -787,33 +787,16 @@ const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        zigbeeModel: ['SP 240'],
+        zigbeeModel: ['SP 240', 'SP 242', 'SP 244'],
         model: 'SP 240',
         vendor: 'Innr',
-        description: 'Smart plug',
-        extend: [m.onOff(), m.electricityMeter({current: {divisor: 1000}, voltage: {divisor: 1}, power: {divisor: 1}, energy: {divisor: 100}})],
-        ota: true,
-    },
-    {
-        zigbeeModel: ['SP 242'],
-        model: 'SP 242',
-        vendor: 'Innr',
-        description: 'Smart plug',
-        extend: [
-            m.onOff(),
-            m.electricityMeter({current: {divisor: 1000}, voltage: {divisor: 1}, power: {divisor: 1}, energy: {divisor: 100}}),
-            // Device looses reporting config on power cycle
-            // https://github.com/Koenkk/zigbee-herdsman-converters/issues/6747
-            m.reconfigureReportingsOnDeviceAnnounce(),
+        description: 'Smart plug (EU)',
+        whiteLabel: [
+            {model: 'SP 242', vendor: 'Innr', description: 'Smart plug (UK)', fingerprint: [{modelID: 'SP 242'}]},
+            {model: 'SP 244', vendor: 'Innr', description: 'Smart plug (US)', fingerprint: [{modelID: 'SP 244'}]},
         ],
-        ota: true,
-    },
-    {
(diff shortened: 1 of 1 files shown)


mundschenk-at:
@Koenkk Ideally, fingerprinting should allow to specify a minimum/maximum firmware version to support both the old and the new FW. To keep things simple, I'm thinking of basing it on `dateCode` (instead of the `softwareBuildID`), does that sound right? To make sure there are no compatibility issues, I'd have to add separate constraint fields. Or would it be better to make a general mechanism here?

Koenkk:
> , I'm thinking of basing it on dateCode (instead of the softwareBuildID), does that sound right?

Currently we cannot do a greater/less than match in the fingerprint, I think it would be the easiest if everyone upgrades their plugs?

mundschenk-at:
Personally, I'm OK with that, but the greater/less than matching could be added easily (though not very elegant) with new fields.

Koenkk:
How many old versions are there? If there is just 1 or 2 we could easily do it with the fingerprint.

mundschenk-at:
I don't know (I've asked @GerryInnr) but at least four (`1.4.3`, `1.6.22`, `1.6.24`, and `1.7.23`).

AlexanderLeporiday:
> > , I'm thinking of basing it on dateCode (instead of the softwareBuildID), does that sound right?
> 
> Currently we cannot do a greater/less than match in the fingerprint, I think it would be the easiest if everyone upgrades their plugs?

I think it's fair to say that they must update the latest version for all fixes. I was just concerned about people who buy the item off the shelf in the future, which is running a much older firmware, and if they run it against a newer version of the converter, they will run into issues. But these issues would be resolved by doing an update of the firmware. I can't see any reason why people wouldn't be able to upgrade the firmware. 

I was more curious about how it's been handled with other products that introduce changes to the converter that older firmware would not be compatible with.

mundschenk-at:
I guess those probably too many to handle without conditional fingerprinting. I can try to add a less-than/greater-than condition to fingerprinting or everyone just has to update for the device to work properly. Both solutions are fine by me, your call @Koenkk

Koenkk:
> to update for the device to work properly

@GerryInnr also agrees that people just have to update, so let's go for that.

Koenkk:
Many thanks!

hmmbob:
For those updating to 2.1.1 with the new firmware and still seeing wrong values: 

You need to hit the "reconfigure" button of the device in the zigbee2mqtt frontend (it's the yellow arrows button) before the changes come into effect.

PR #8723, opened 2025-02-03
Prevent Tuya packets from being reprocessed by checking entire packet instead of `seq` only

jpbede:
Prevent duplicate Tuya packets from being reprocessed by comparing the entire packet instead of just referring to the `seq`.

There seem to be devices that repeatedly send their reports with the same sequence ID.

https://github.com/Koenkk/zigbee2mqtt/issues/25605 should still be fixed, as far as I have seen, the packages were identical.

See https://github.com/Koenkk/zigbee2mqtt/issues/26124 and probably https://github.com/Koenkk/zigbee2mqtt/issues/26145, https://github.com/Koenkk/zigbee2mqtt/issues/26148 also

### DIFF
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -1867,7 +1867,7 @@ const tuyaFz = {
         cluster: 'manuSpecificTuya',
         type: ['commandDataResponse', 'commandDataReport', 'commandActiveStatusReport', 'commandActiveStatusReportAlt'],
         convert: (model, msg, publish, options, meta) => {
-            if (utils.hasAlreadyProcessedMessage(msg, model, msg.data.seq)) return;
+            if (utils.hasAlreadyProcessedMessage(msg, model, msg.data)) return;
             const result: KeyValue = {};
             if (!model.meta || !model.meta.tuyaDatapoints) throw new Error('No datapoints map defined');
             const datapoints = model.meta.tuyaDatapoints;
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -1867,7 +1867,7 @@ const tuyaFz = {
         cluster: 'manuSpecificTuya',
         type: ['commandDataResponse', 'commandDataReport', 'commandActiveStatusReport', 'commandActiveStatusReportAlt'],
         convert: (model, msg, publish, options, meta) => {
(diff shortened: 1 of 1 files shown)


jpbede:
I probably wrote the PR description a bit misleading.

The issue is that there are devices that repeatedly send their reports with the same sequence ID, although data has changed. With the filter based on the `seq` in place, those reports get filtered.
 
```
[2025-02-03 19:50:39] debug: 	zh:controller: Received payload: clusterID=61184, address=52189, groupID=0, endpoint=1, destinationEndpoint=1, wasBroadcast=false, linkQuality=116, frame={"header":{"frameControl":{"frameType":1,"manufacturerSpecific":false,"direction":1,"disableDefaultResponse":false,"reservedBits":0},"transactionSequenceNumber":153,"commandIdentifier":6},"payload":{"seq":256,"dpValues":[{"dp":105,"datatype":2,"data":{"type":"Buffer","data":[0,0,1,241]}}]},"command":{"ID":6,"parameters":[{"name":"seq","type":33},{"name":"dpValues","type":1011}],"name":"activeStatusReport"}}

[2025-02-03 19:51:39] debug: 	zh:controller: Received payload: clusterID=61184, address=52189, groupID=0, endpoint=1, destinationEndpoint=1, wasBroadcast=false, linkQuality=116, frame={"header":{"frameControl":{"frameType":1,"manufacturerSpecific":false,"direction":1,"disableDefaultResponse":false,"reservedBits":0},"transactionSequenceNumber":156,"commandIdentifier":6},"payload":{"seq":256,"dpValues":[{"dp":105,"datatype":2,"data":{"type":"Buffer","data":[0,0,1,246]}}]},"command":{"ID":6,"parameters":[{"name":"seq","type":33},{"name":"dpValues","type":1011}],"name":"activeStatusReport"}}
```

Koenkk:
the `transactionSequenceNumber` seems to be different though, could you try with: `if (utils.hasAlreadyProcessedMessage(msg, model)) return;` ?

Koenkk:
Since quite some people are impacted I've updated it to use the `transactionSequenceNumber` now. Let me know if that is not correct.

jpbede:
Thanks, seems to work when bumping the `zigbee-herdsman-converts` manually and running `pnpm run start`!

PR #8724, opened 2025-02-03
Added type for LightArgs.levelConfig 

mccordscvs:
In modernExtend the interface LightArgs.levelConfig.disableFeatures uses a generic string[] type. This argument is tightly bound to 6 options but there is no autocomplete for it.

I exported a new type called LevelConfigFeatures form types. 
Applied the type to:

- exposes.Light.withLevelConfig
- modernExtend.LightArgs.LevelConfig.disableFeatures
- lib/ikea.ts ikeaLight Function

I thought the best option would be to apply the type directly to withLevelConfig and work out from there instead of just applying it the interface. I debated weather to export directly form extend.ts but decided types.ts was the best option.  If you want any changes let me know.


Koenkk:
Thanks!

PR #8727, opened 2025-02-04
Add support for light_brightness_move and light_brightness step to Inovelli switches

rohankapoorcom:
It turns out Inovelli switches support moving and stepping (except they have a bug in the firmware that prevents moving down from working). This exposes that functionality for the VZM30, VZM31, VZM36 that all support it. VZM35 is a fan switch and does not support it.

This was tested by both myself and @InovelliUSA


Koenkk:
Thanks!

PR #8728, opened 2025-02-04
Update Lixee.ts, need help

fairecasoimeme:
I tried to update the  converters but I don't know how to test in local.
* I used external converter (following tutos) but I can't access to logs because I get a 500 internal server message, so I can't see if it's ok or not
* When I try to do dev console read attribute, 'tariffPeriod' don't appear in 'lixeePrivate' so I think this is because cluster.ts take advantage ?

Can anyone help ?


src/devices/lixee.ts
```diff
@@ -13,6 +14,65 @@ const ea = exposes.access;
 const e = exposes.presets;
 
 const NS = 'zhc:lixee';
+
+const local = {
+	modernExtend: {
+        addCustomClusterManuSpecificLixee: () =>
+            deviceAddCustomCluster('liXeePrivate', {
```
Koenkk, line 21:
```suggestion
            deviceAddCustomCluster('liXeePrivateCustom', {
```

Could you check if this now work from the dev console? Otherwise there is a conflict with the already existing cluster.

Koenkk:
> ) but I can't access to logs because I get a 500 internal server message, so I can't see if it's ok or not

Where do you get this?

src/devices/lixee.ts
fairecasoimeme: Ok thank you for this

fairecasoimeme:
> > ) but I can't access to logs because I get a 500 internal server message, so I can't see if it's ok or not
> 
> Where do you get this?

image

fairecasoimeme:
So I don't know how to test.
I read that I have to copy my lixee.js in external_converters folder and the file is used when I restart Z2M but I have some errors like : 
* z2m: Failed to call 'ExternalConverters' 'start' (/app/data/extension/lixee.js:125
    } satisfies Fz.Converter,

Or others but I don't know if it's linked to js file, if i have to use tz file ...
I'd really like to contribute but it's really complicated

Koenkk:
Try changing `} satisfies Fz.Converter,` to `},`, basically you have to remove all the Typescript stuff since this is a Javascript file. Could you send me the whole external converter?

fairecasoimeme:
> Try changing `} satisfies Fz.Converter,` to `},`, basically you have to remove all the Typescript stuff since this is a Javascript file. Could you send me the whole external converter?

Thank you for the help. I removed all the typescript stuff and it's works now but how to compile to ts file ?

other question : I tried to enhance some exposes datas with composite and features but that's add an "apply button" how to do the same without button ? I didn't find the equivalent.

in attachment the lixee.js file (in txt format)
[lixee.js.txt](https://github.com/user-attachments/files/18737504/lixee.js.txt)

Koenkk:
> I tried to enhance some exposes datas with composite and features but that's add an "apply button" how to do the same without button ? I didn't find the equivalent.

Composite always has an apply button (since all attributes need to be submitted at once).

PR #8730, opened 2025-02-04
Tuya BAC-003: Add optional device state property as per issue request

robvanoostenrijk:
Resolution of discussion in koenkk/zigbee2mqtt#21647


Koenkk:
Thanks!

PR #8732, opened 2025-02-04
add support for Purmo/Radson Yali Parada Plus

sdebruyn:
Tested with my own Radson Yali Parada Plus. Probably similar models work as well, but I don't know their model numbers.

### DIFF
--- a/src/devices/purmo.ts
+++ b/src/devices/purmo.ts
@@ -1,0 +1,66 @@
+import fz from '../converters/fromZigbee';
+import tz from '../converters/toZigbee';
+import * as exposes from '../lib/exposes';
+import * as reporting from '../lib/reporting';
+import {DefinitionWithExtend} from '../lib/types';
+
+const e = exposes.presets;
+
+const definitions: DefinitionWithExtend[] = [
+    {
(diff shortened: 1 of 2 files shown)


src/devices/purmo.ts
```diff
@@ -0,0 +1,66 @@
+import fz from '../converters/fromZigbee';
```
Koenkk, line 1: Please also add to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts

sdebruyn:
Just documenting this here for anyone who might search for it (and for myself):

The Purmo/Radson Yali Plus series might show a ⚠️ 3 (warning sign with number 3) every few seconds on its display. This error code is not documented in the installation manual. Radson's technical staff is aware of the issue and will release an OTA update that should resolve it. The update would be released through their Unisenza gateway and app, but at that point we should check if we can get the OTA file and serve it through z2m as well. The warning should not be concerning as it's related to the zigbee connection and not to any heating/power element. The warning also appears for users using the Unisenza gateway. The zigbee connection does seem to work fine as far I can see.

Koenkk:
Thanks!

Make sure that:
- The filename is `Yali-Parada-Plus.png`
- The size is 512x512

PR #8739, opened 2025-02-04
Correction of mistake and added daily energy meter

SapozhnikUA:
* fix wrong parameter 
* added daily energy meters for each phases 

Please correct code - I'm not programmer :) 

image


Koenkk:
Thanks!

PR #8740, opened 2025-02-04
Update iluminize.ts

chris-1243:
Update model 511.344.

The aim is to see model ZG2819S-RGB from Sunricher in the supported device list


Koenkk:
Thanks!

PR #8746, opened 2025-02-05
PTVO converters. Added rounding for pressure, humidity, illuminance.

ptvoinfo:
PTVO converters. Added rounding for pressure, humidity, illuminance.


Koenkk:
Thanks!

PR #8673, opened 2025-01-25
Update lixee.ts

fairecasoimeme:
* add tariff period new value
* fix chunk default description
* change default chunk value to 2

PR #8748, opened 2025-02-05
fix: Fix LiXee integration

Koenkk:
Reverts Koenkk/zigbee-herdsman-converters#8673 (CC: @fairecasoimeme)

Should fix https://github.com/Koenkk/zigbee2mqtt/issues/26146


jojo01fr:
Hi, just made the update to 2.1.1, LTARF has the right value now, thanks for the good work. One can also extract that info and even other interesting, from STGE with a 32 bit decoder

PR #8749, opened 2025-02-05
fix: Enable OTA for newer Hue Motion Sensors

mundschenk-at:
- Enables OTA for Hue Motion Sensor `SML003` (an image is now available).
- Enables OTA for Hue Outdoor Motion Sensor `SML004` (I'm assuming an image is available as the firmware changelog states that there are updates for all accessories).


Koenkk:
Thanks!

PR #8756, opened 2025-02-06
Update innr.ts

chris-1243:


Koenkk:
Thanks!

AlexanderLeporiday:
@Koenkk Given that the docs were already generated with the wrong vendor I wasn't sure if they needed to be manually fixed or if docgen would correct the issue and preserve the notes so I raised https://github.com/Koenkk/zigbee2mqtt.io/pull/3499

Please close it down if it's not needed and this will be automatically fixed by this request.

Koenkk:
Docs will be auto generated with the next release.

PR #8757, opened 2025-02-06
Update tuya.ts

dorianmartinez31:
Support of QS-Zigbee-SEC01-DC (Mini 1 Gang Zigbee Switch Module) Fix of support of QS-Zigbee-SEC01-U after last firmware update, declaring module as Router instead of EndDevice. Still compatible with previous firmwares


Koenkk:
Thanks!

PR #8759, opened 2025-02-07
Update tuya.ts - 

TheLinx79:
Added another version of the round Tuya 24G Zigbee USB-C presence sensor received from Aliexpress.


Koenkk:
Thanks!

PR #8770, opened 2025-02-08
New devices EFEKTA ePST POW

smartboxchannel:
New devices EFEKTA ePST POW


smartboxchannel:
@Koenkk I don't understand what could be wrong with code style. I don't have the option to check through prettier --check. I've checked everything I can, the style is perfect. You need to at least add some kind of message with a link to the lines in which your code style check fails.

PR #8774, opened 2025-02-08
adding support for _TZE284_ogx8u5z6 thermostatic radiator valve

icecream42:
This is just another variant of the _TZE284_ogx8u5z6, used the same external converter with changed string as for the _TZE204_ogx8u5z6 before.


Koenkk:
Thanks!

PR #8785, opened 2025-02-09
Exposing action_group for EGLO 99099

szamozoli:
Exposing action group enables the usage of group selector buttons, enhancing the capabilities of the remote.


Koenkk:
Thanks!

PR #8788, opened 2025-02-10
Fix Aqara `DJT12LM` issues

z-x-f:
This is my DJT12LM information, it works fine
```js
   "device": {
        "applicationVersion": 25,
        "dateCode": "20200528",
        "hardwareVersion": 1,
        "manufacturerID": 4447,
        "manufacturerName": "LUMI",
        "model": "DJT12LM",
        "powerSource": "Battery",
        "softwareBuildID": "2019\u0000www.",
        "stackVersion": 2,
        "type": "EndDevice",
        "zclVersion": 3
    }
```

### DIFF
--- a/src/lib/lumi.ts
+++ b/src/lib/lumi.ts
@@ -2045,17 +2045,24 @@ export const lumiModernExtend = {
             ...args,
         }),
     lumiVibration: (): ModernExtend => {
-        const exposes: Expose[] = [e.action(['shake', 'triple_strike'])];
+        const exposes: Expose[] = [e.action(['triple_strike', 'movement'])];
 
         const fromZigbee: Fz.Converter[] = [
             {
-                cluster: 'ssIasZone',
+                cluster: 'genMultistateInput',
                 type: ['attributeReport', 'readResponse'],
                 convert: (model, msg, publish, options, meta) => {
-                    if (msg.data[45] !== undefined) {
-                        const zoneStatus = msg.data[45];
-                        const actionLookup: KeyValueNumberString = {1: 'shake', 2: 'triple_strike'};
-                        return {action: actionLookup[zoneStatus]};
+                    if (msg.data.presentValue !== undefined && msg.data.presentValue === 1) {
+                        return {action: 'triple_strike'};
+                    }
+                },
+            },
+            {
+                cluster: 'manuSpecificLumi',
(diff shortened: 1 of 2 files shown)


src/lib/lumi.ts
```diff
@@ -5247,6 +5254,26 @@ export const toZigbee = {
             }
         },
     } satisfies Tz.Converter,
+    lumi_sensitivity_adjustment: {
```
Koenkk, line 5257: Could you use modernExtend for this just like: https://github.com/Koenkk/zigbee-herdsman-converters/blob/9c3b9876056c84a94ae7083a36bedc7b6aafe9ec/src/lib/lumi.ts#L1534

src/lib/lumi.ts
```diff
@@ -2045,17 +2045,24 @@ export const lumiModernExtend = {
             ...args,
         }),
     lumiVibration: (): ModernExtend => {
-        const exposes: Expose[] = [e.action(['shake', 'triple_strike'])];
+        const exposes: Expose[] = [e.action(['triple_strike', 'movement'])];
```
Koenkk, line 2048: Please keep the actions the same to prevent a breaking change.

src/lib/lumi.ts
```diff
@@ -2045,17 +2045,24 @@ export const lumiModernExtend = {
...
 
         const fromZigbee: Fz.Converter[] = [
             {
-                cluster: 'ssIasZone',
```
Koenkk, line 2052: Maybe we should keep both converters? I'm not sure if this will break for others.

src/lib/lumi.ts
z-x-f: Oh, I was just following the naming in the Aqara app. If it's not appropriate, I'll revert back to the original action.
aqara app

z-x-f: @Koenkk Hi, Kk ,What do you think?

Koenkk: I would not go for a breaking change here, so please revert.

z-x-f: Since I'm not entirely sure what the original `shake` is meant to do, would it be okay to retain 'shake' and add `movement` as an additional action?

Koenkk:
Thanks!

PR #8789, opened 2025-02-10
Orvibo MixSwitch 4 gang converter update

vilaemail:
Changes:
- Add manufacturer specific custom commands (so they can be executed from dev console or by publishing to MQTT)
- Document commands in the code comments
- Add composite options for changing switch behavior (recalling scene and/or toggling a realy)
- Add support for updating state when switch recalls a scene (this should use standard scene capability, however at the moment we are unable to properly configure scenes for each endpoint, [see feature request](https://github.com/nurikk/zigbee2mqtt-frontend/issues/2393)).

Breaking(?) changes:
- Remove powerOnBehavior - as it is not supported by the switch, makes no difference to switch operation when powered off/on
- Rename endpoints L1 -> left_up, L2 -> left_down, L4 (this was wrong, it was L3 on the back of the switch) -> center_up, L4 (new endpoint) -> center_down, L5 -> right_up, L6 -> right_down

### DIFF
--- a/src/devices/orvibo.ts
+++ b/src/devices/orvibo.ts
@@ -3,11 +3,15 @@ import {Zcl} from 'zigbee-herdsman';
 import fz from '../converters/fromZigbee';
 import tz from '../converters/toZigbee';
 import * as exposes from '../lib/exposes';
+import {logger} from '../lib/logger';
 import * as m from '../lib/modernExtend';
 import * as reporting from '../lib/reporting';
-import {DefinitionWithExtend, Tz} from '../lib/types';
+import {DefinitionWithExtend, Fz, KeyValueAny, ModernExtend, OnEvent, Option, Tz} from '../lib/types';
+import {addActionGroup, hasAlreadyProcessedMessage, postfixWithEndpointName} from '../lib/utils';
 
 const e = exposes.presets;
+const ea = exposes.access;
+const NS = 'zhc:orvibo';
 
 const tzLocal = {
     DD10Z_brightness: {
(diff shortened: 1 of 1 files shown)


src/devices/orvibo.ts
```diff
@@ -310,7 +312,51 @@ const definitions: DefinitionWithExtend[] = [
...
+                        // To REMOVE relay from a scene execute on relay endpoint (1-4) cluster 5 command 2 with payload { "groupid":0,"sceneid":<SCENE_ID> }
+
+                        // COMMANDING RELAY AND RECALLING A SCENE
+                        // It is possible to configure a switch to command a relay (see setSwitchRelay command) and recall a scene (this command). It is important to execute commands in the following order - clearSwitchAction, then setSwitchRelay and then setSwitchScene.
+                        // This can be useful for a scenario where you have blinds motor set-up on relay 1 and 2, and would like to have switch 1 toggle relay 1, but turn off relay 2. You would command relay 1 with TOGGLE action and recall scene set-up for relay 2 to turn it off.
+                        ID: 0x04,
+                        parameters: [{name: 'data', type: Zcl.BuffaloZclDataType.BUFFER}],
+                    },
+                },
+                commandsResponse: {},
```
vilaemail, line 354: Is there a way to test if above custom commands received proper response from the device (from sniffing it is simple 1 byte with value 0 response), and if not report error in the logs?

src/devices/orvibo.ts
```diff
@@ -310,7 +312,51 @@ const definitions: DefinitionWithExtend[] = [
         model: 'T40W4Z',
         vendor: 'ORVIBO',
         description: 'MixSwitch 4 gangs',
-        extend: [m.deviceEndpoints({endpoints: {l1: 1, l2: 2, l4: 3, l5: 5, l6: 6}}), m.onOff({endpointNames: ['l1', 'l2', 'l4', 'l5', 'l6']})],
+        extend: [
+            m.deviceAddCustomCluster('manuSpecificOrvibo', {
+                ID: 0x0017,
+                attributes: {},
+                commands: {
```
vilaemail, line 319: @Koenkk Right now I am using these commands in custom script, which publishes MQTT messages. Do you see this as somehow exposable to the front end? It is not as simple as showing a property to be changed, as there are several commands which need to be executed, see example below where I am setting up switch (i.e. 1) to turn off relay 2 and toggle relay 1. Use case is blinds motor where relay 1 and relay 2 are up/down controls and they must not be on at the same time.

```
echo $'\t'"Clear scene $scene_id for R$off_relay"
mosquitto_pub -t "zigbee2mqtt/$address/$off_relay/set" -m '{ "command": {"cluster":5, "command":2, "payload": { "groupid":0,"sceneid":'$scene_id' }} }' -h core-mosquitto -p 1883 -u "$USER" -P "$PASSWORD"
sleep 6
    
echo $'\t'"Add scene $scene_id command R$off_relay OFF"
mosquitto_pub -t "zigbee2mqtt/$address/$off_relay/set" -m '{ "command": {"cluster":5, "command":0, "payload": { "groupid": 0, "sceneid": '$scene_id', "transtime": 0, "scenename": "'$scene_id'", "extensionfieldsets": [{"clstId": 6, "len": 1, "extField":[0]}] }} }' -h core-mosquitto -p 1883 -u "$USER" -P "$PASSWORD"
sleep 6
    
echo $'\t'"Clear SW$switch"
mosquitto_pub -t "zigbee2mqtt/$address/7/set" -m '{ "command": {"cluster":23, "command":2, "payload": { "data":['$switch',0,0] }} }' -h core-mosquitto -p 1883 -u "$USER" -P "$PASSWORD"
sleep 6
    
echo $'\t'"Add SW$switch recall scene $scene_id"
mosquitto_pub -t "zigbee2mqtt/$address/7/set" -m '{ "command": {"cluster":23, "command":4, "payload": {"data":['$switch',0,0,0,0,'$scene_id']}} }' -h core-mosquitto -p 1883 -u "$USER" -P "$PASSWORD"
sleep 6
    
echo $'\t'"Add SW$switch command R$on_off_relay ON/OFF"
mosquitto_pub -t "zigbee2mqtt/$address/7/set" -m '{ "command": {"cluster":23, "command":0, "payload": {"data":['$switch',0,0,'$reversed_address','$on_off_relay',4,1,6,0,1,2]}} }' -h core-mosquitto -p 1883 -u "$USER" -P "$PASSWORD"
sleep 6
```

address = IEEE address
off_relay = endpoint (i.e. 2)
switch = endpoint (i.e. 1)
on_off_relay = endpoint (i.e. 1)

Anyway, I am willing to document this + upload my bash script which eases configuring, just let me know where is the proper place? I was thinking notes on the docs page for the device, if that is appropriate?

src/devices/orvibo.ts
```diff
@@ -310,7 +312,51 @@ const definitions: DefinitionWithExtend[] = [
...
+                        // COMMANDING RELAY AND RECALLING A SCENE
+                        // It is possible to configure a switch to command a relay (see setSwitchRelay command) and recall a scene (this command). It is important to execute commands in the following order - clearSwitchAction, then setSwitchRelay and then setSwitchScene.
+                        // This can be useful for a scenario where you have blinds motor set-up on relay 1 and 2, and would like to have switch 1 toggle relay 1, but turn off relay 2. You would command relay 1 with TOGGLE action and recall scene set-up for relay 2 to turn it off.
+                        ID: 0x04,
+                        parameters: [{name: 'data', type: Zcl.BuffaloZclDataType.BUFFER}],
+                    },
+                },
+                commandsResponse: {},
+            }),
+            m.deviceEndpoints({endpoints: {left_up: 1, left_down: 2, center_up: 3, center_down: 4, right_up: 5, right_down: 6}}),
```
vilaemail, line 356: @Koenkk I have renamed endpoints, is kind of "breaking change" OK?

Koenkk: It's fine

src/devices/orvibo.ts
Koenkk: If they don't throw an error when sending it is OK

src/devices/orvibo.ts
Koenkk: For such cases you have to create a toZigbee converter, from there you can send anything you want (even multiple commands). You can find many examples in https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/converters/toZigbee.ts

vilaemail: @Koenkk 
**Configuring a switch**
How can I access other key/values when setting a particular key/value in the converter? Will `meta.message.<key_here>` be populated for other key/values?

Reasoning:
I need 2 user inputs (which relay to control - if any) and which scene to recall (if any). Would that be a converter with 2 key/value pairs? If so I would need to access both key/value pairs when setting any of those two keys, as the device API requires clearing and re-setting values. I see the meta argument, but am unsure where would I find such value?

**Setting scenes**
I think this should be exposed on the frontend, let me know if that makes sense? [Feature request I have opened](https://github.com/nurikk/zigbee2mqtt-frontend/issues/2393).

vilaemail: I think the scope of this PR is good for us to complete it now. The discussion from above can be a separate PR.

If it would help anyone, attaching the mentioned bash script.
[orvibo.txt](https://github.com/user-attachments/files/18768305/orvibo.txt)

Koenkk: `meta.message` indeed, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/dc90fb39cb6877af8e3b8be2a3f523f12c6e32d1/src/converters/toZigbee.ts#L32

> I need 2 user inputs (which relay to control - if any) and which scene to recall (if any). Would that be a converter with 2 key/value pairs? 

If both always need to provided at the same time, I would make it an `e.composite`, so you have to change e.g. `{"my_key": {"scene": x, "relay": y}}`, then you need only one toZigbee converter with the key being `my_key`.

vilaemail: @Koenkk  I've updated the PR (and initial description) based on this discussion. Please take a look and let me know if I need to change anything else. Thanks!

Koenkk:
Thanks!

cjim8889:
Thanks for the good work, I wonder whether it's also possible to let right_up and right_down button to control other types of remote devices such as curtain controller?

PR #8790, opened 2025-02-10
Change the inching_control_set function. If the value of inching_time…

liubotongs:
… is greater than 128, the setting is abnormal.


Koenkk:
Thanks!

PR #8798, opened 2025-02-11
Update iluminize.ts

chris-1243:
Update modernextend commands

This reflects which commands are actually sent by the device and not all the definitions


chris-1243:
You may completely disregard #8797 which I closed.

I am trying to learn how to use your tools to check my job. Kind of nightmare for a beginner but it seems I managed to use them correctly...

Koenkk:
Thanks!

PR #8799, opened 2025-02-11
TS0601 _TZE200_cxakecfo white label ME168

Anian-igor:
Whis PR will add support for _TZE200_cxakecfo white label TUYA thermostat ME168


Koenkk:
Thanks!

PR #8803, opened 2025-02-12
Support for ZHT-SR thermostat from moes

acasajus:
Still doesn't support programs but the basic functionality should be there

### DIFF
--- a/src/devices/moes.ts
+++ b/src/devices/moes.ts
@@ -597,6 +597,85 @@ const definitions: DefinitionWithExtend[] = [
             tuya.modernExtend.tuyaLedIndicator(),
         ],
     },
+    {
+        fingerprint: [
+            {modelID: 'TS0601', manufacturerName: '_TZE204_lpedvtvr'},
+        ],
+        model: 'ZHT-SR',
+        vendor: 'Moes',
+        description: 'Moes smart ring thermostat',
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        onEvent: tuya.onEventSetTime, 
+        configure: tuya.configureMagicPacket,
+        exposes: [
+            e
+            .climate()
+            .withSetpoint('current_heating_setpoint', 1, 90, 0.5, ea.STATE_SET)
+            .withLocalTemperature(ea.STATE)
+            .withLocalTemperatureCalibration(-10, 10, 1, ea.STATE_SET)
+            .withSystemMode(['off', 'heat'], ea.STATE_SET)
+            .withRunningState(['idle', 'heat'], ea.STATE)
+            .withPreset(['manual', 'temporary manual', 'program', 'eco']),
+            //eco_temperature,
+            new exposes.Numeric('eco_temperature', exposes.access.STATE_SET)
+                .withUnit('°C')
+                .withDescription(
+                    'Temperature for eco mode',
+                )
+                .withValueMin(10)
+                .withValueMax(30),
+            e.child_lock(),
+            e.deadzone_temperature().withValueMin(1).withValueMax(10).withValueStep(0.5),
+            e.max_temperature_limit().withValueMin(35).withValueMax(45).withValueStep(0.1),
+            e.min_temperature_limit().withValueMin(1).withValueMax(15).withValueStep(0.1),
+            e.temperature_sensor_select(['IN', 'AL', 'OU']),
+            //backlight_brightness
+            new exposes.Numeric('backlight_brightness', exposes.access.STATE_SET)
+                .withUnit('%')
+                .withDescription(
+                    'Backlight brightness',
+                )
+                .withValueMin(0)
+                .withValueMax(100)
+                .withValueStep(0.1),
+            //Screen time
+            new exposes.Enum('screen_time', exposes.access.STATE_SET, [10,20,30,40,50,60].map((secs) => `${secs} seconds`))
+                .withDescription('Screen on time')
+                .withCategory('config'),
+            e.binary('rgblight',exposes.access.STATE, true, false),
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [1, 'system_mode', tuya.valueConverterBasic.lookup({'off': false,'heat': true})],
+                [2, 'preset', tuya.valueConverterBasic.lookup({'manual': 0, 'temporary manual': 1, 'program': 2, 'eco': 3})],
+                [16, 'local_temperature', tuya.valueConverter.divideBy10], 
+                [18, 'min_temperature_limit', tuya.valueConverter.divideBy10],
+                [32, 'sensor', tuya.valueConverterBasic.lookup({'IN':0,'AL':1,'OU':2})],
+                [34, 'max_temperature_limit', tuya.valueConverter.divideBy10],
+                [39, 'child_lock', tuya.valueConverter.lockUnlock],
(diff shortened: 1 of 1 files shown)


src/devices/moes.ts
```diff
@@ -597,6 +597,104 @@ const definitions: DefinitionWithExtend[] = [
...
+                [16, 'local_temperature', tuya.valueConverter.divideBy10],
+                [18, 'min_temperature_limit', tuya.valueConverter.divideBy10],
+                [32, 'sensor', tuya.valueConverterBasic.lookup({'Internal sensor (IN)': 0, 'Both (AL)': 1, 'External sensor (OU)': 2})],
+                [34, 'max_temperature_limit', tuya.valueConverter.divideBy10],
+                [39, 'child_lock', tuya.valueConverter.lockUnlock],
+                [47, 'running_state', tuya.valueConverterBasic.lookup({heat: 0, idle: 1})],
+                [48, 'backlight_brightness', tuya.valueConverter.raw],
+                [50, 'current_heating_setpoint', tuya.valueConverter.divideBy10],
+                [101, 'local_temperature_calibration', tuya.valueConverter.raw],
+                [102, 'week_program_l3_1', tuya.valueConverter.raw],
```
Koenkk, line 669: All these are not exposed through `exposes`, I propose to remove them.

src/devices/moes.ts
```diff
@@ -597,6 +597,104 @@ const definitions: DefinitionWithExtend[] = [
...
+                [109, 'floor_temperature', tuya.valueConverter.divideBy10],
+                [110, 'deadzone_temperature', tuya.valueConverter.divideBy10],
+                [111, 'high_protect_temperature', tuya.valueConverter.divideBy10],
+                [112, 'low_protect_temperature', tuya.valueConverter.divideBy10],
+                [113, 'eco_temperature', tuya.valueConverter.divideBy10],
+                [
+                    114,
+                    'screen_time',
+                    tuya.valueConverterBasic.lookup({
+                        '10 seconds': 0,
```
Koenkk, line 685:
```suggestion
                        '10_sec': 0,
```

(do the same for all others values)

acasajus: That will make the selector show `10_sec` instead of `10 seconds`. Why not keep it more human readable?

src/devices/moes.ts
```diff
@@ -597,6 +597,104 @@ const definitions: DefinitionWithExtend[] = [
...
+                        '20 seconds': 1,
+                        '30 seconds': 2,
+                        '40 seconds': 3,
+                        '50 seconds': 4,
+                        '60 seconds': 5,
+                    }),
+                ],
+                [115, 'rbglight', tuya.valueConverterBasic.trueFalse(1)],
```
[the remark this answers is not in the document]
acasajus: Done

src/devices/moes.ts
acasajus: Removed

src/devices/moes.ts
Koenkk: It's the convention for every enum value, I'm also fine with `10_seconds`

acasajus: Done

Koenkk:
Thanks!

SNPN:
Guys,
Thanks for your work on implementing the device.
I guese there are few other datapoints available such as : 

```
DP ID | DP Name | DP Code | Data transfer type | Data Type | Data Attributes
-- | -- | -- | -- | -- | --
28 | Factory data reset | factory_reset | Issue and report | bool |  
102 | Week Program 13 1 | week_program_13_1 | Issue and report | raw |  
103 | Week Program 13 2 | week_program_13_2 | Issue and report | raw |  
104 | Week Program 13 3 | week_program_13_3 | Issue and report | raw |  
105 | Week Program 13 4 | week_program_13_4 | Issue and report | raw |  
106 | Week Program 13 5 | week_program_13_5 | Issue and report | raw |  
107 | Week Program 13 6 | week_program_13_6 | Issue and report | raw |  
108 | Week Program 13 7 | week_program_13_7 | Issue and report | raw |  
114 | Screen Time Set | screen_time_set | Issue and report | enum | Enumerated values：10S,20S,30S,40S,50S,60S
115 | Rgblight | rgblight | Issue and report | bool |  
```



(not sure DP 115 is useful since it does not have RGB I guess. But probably it allows to disable the rign light).

acasajus:
Hey,

Had all of those (except 28) but @Koenkk asked to remove 102-108 because there was still no support for programs. I don't know the format they expect and it's not like any other thermostat already implemented that I can see. 114 and 115 are there. And 28 seems risky to export. A misclick and it's out. You can always factory reset the device directly.

If you have the format for the week program I'm happy to implement it.

SNPN:
Would the comment in the source code help?

```
Function name : dp_download_week_program_13_1_handle
Function description : on DPID_WEEK_PROGRAM_13_1 processing function
Input parameter : value:Source data
        : length:Data length
Return parameter : Successful return:SUCCESS/Failed to return:ERROR
Instructions for use : Issue and report type,need to report the result to App after data is dealt with
```

Function is the following:

```
static unsigned char dp_download_week_program_13_1_handle(const unsigned char value[], unsigned short length)
{
    //Example: The current DP type isRAW
    unsigned char ret;
    /*
    //RAW type data processing
    
    */
    
    //There should be a report after processing the DP
    ret = mcu_dp_raw_update(DPID_WEEK_PROGRAM_13_1,value,length);
    if(ret == SUCCESS)
        return SUCCESS;
    else
        return ERROR;
}
```

acasajus:
I also go t that from moes. That unfortulately does not define how the value is encoded.

acasajus:
Ask them if they can sent what byte array  we need to send to those data points and what each byte means. I've also asked them but customer support doesn't seem to be able to get it. Maybe you'll have more luck.

SNPN:
Hi,
I tried to ask MOES but now they don't think to be as cooperative as they used to be and ask me info about the Thermostat via the official supported app...

Anyway, with the new update they are automatically detected and it looks like all the features are implemented.
Still I experience some bugs:
- Hour on the thermostat is correct ... but not the date (like, now it says 13h46 but Friday ...)
- It seems that I cannot send anything to the thermostat. When I change the heating setpoint for instance, the value do not change on the thermostat and on HA, it reverts back to what is set on the thermostat. On the contrary, any change made on the thermostat is replicated on HA. Looks like I can only READ values but not WRITE.
- The message on top of the screen never disappears (asks to connect the device to a network). I suspect it's linked to the official app, not a big deal but a bit annoying since it "looks like" it's not configured while it's perfectly connected th Z2M
Any idea how to fix?
Thanks a lot!

splichy:
@acasajus 
Hi, thank you for your work! 
Are you sure with (min/max)s for min_temperature_limit(1-15) and max_temperature_limit(35-45) as those actually affect what can be set by the knob directly on the device(at least in sensor mode IN). Those min/max values seem more appropriate for low_protect_temperature and high_protect_temperature(in this case 45 can be a bit too low e.g. for ceiling heating, where higher temperatures are usually used)
In that case descriptions are IMHO also swapped in https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/devices/ZHT-SR.md
I would expect desc:
> Cuts the thermostat out regardless of air temperature if the external floor sensor exceeds this temperature. Only used by the thermostat when in AL sensor mode..

to be at High protect temperature and not at Maximum temperature limit.

PR #8804, opened 2025-02-13
feat(lumiLight): custom color temp range

z-x-f:
Add support for a custom color temperature range in the lumiLight extension.
This change allows users to override the default range ([153,370]) with a custom one.
For example, for Opple MX480 the correct color temperature range is 175-370.

### DIFF
--- a/src/lib/lumi.ts
+++ b/src/lib/lumi.ts
@@ -1427,13 +1427,14 @@ export const lumiModernExtend = {
     lumiLight: (
         args?: Omit<modernExtend.LightArgs, 'colorTemp'> & {
             colorTemp?: true;
+            colorTempRange?: Range;
             powerOutageMemory?: 'switch' | 'light' | 'enum';
             deviceTemperature?: boolean;
             powerOutageCount?: boolean;
         },
     ) => {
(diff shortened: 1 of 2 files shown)


src/lib/lumi.ts
```diff
@@ -1427,13 +1427,14 @@ export const lumiModernExtend = {
...
             colorTemp?: true;
+            colorTempRange?: Range;
             powerOutageMemory?: 'switch' | 'light' | 'enum';
             deviceTemperature?: boolean;
             powerOutageCount?: boolean;
         },
     ) => {
         args = {powerOutageCount: true, deviceTemperature: true, ...args};
-        const colorTemp: {range: Range; startup: boolean} = args.colorTemp ? {startup: false, range: [153, 370]} : undefined;
+        const colorTemp: {range: Range; startup: boolean} = args.colorTemp ? {startup: false, range: args.colorTempRange || [153, 370]} : undefined;
```
Koenkk, line 1437:
```suggestion
        const colorTemp: {range: Range; startup: boolean} = args.colorTemp ? {startup: false, range: args.colorTempRange ?? [153, 370]} : undefined;
```

Koenkk:
Thanks!

Issue #8605, opened 2025-01-14
Centralite 3156105 running states are still wrong (sometimes)


triosniolin:
Issue https://github.com/Koenkk/zigbee-herdsman-converters/issues/4284 made the statement that `Z2M correctly interprets the thermostat's running state in its non-heat pump configuration, but gets it wrong when configured to use a heat pump`. 

PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/4317 was created to solve this issue, but it did not take into account that the running_state is actually **correct** except when the device is configured as a heat pump. My use case is for this device as a furnace thermostat and the "heating" `running_state` is not coming through because of this `running_state` remapping.

I do not know if the thermostat publishes its running configuration; I'm not a zigbee wizard. However, I would be happy to help if someone can talk me through how to query the device configuration so we can hopefully get both the standard (furnace/air conditioning) mode and the heat pump mode to display properly on this thermostat.

Devbymitch:
@triosniolin thanks for your research into this, it made fixing it much easier! Just got merged into master so assuming next release will include this.

Change `Heat pump mode` in `Settings (specific)` to false.

PR #8806, opened 2025-02-13
Centralite 3156105 - Add reporting for missing attributes & Fix for non-heat pump configurations

Devbymitch:
**Device:** `Centralite 3156105`

**Fixes:** Issues #8605 and [#25694 (zigbee2mqtt)](https://github.com/Koenkk/zigbee2mqtt/issues/25694)

**Summary:** 
Added attributes that do not report by default (`OccupiedHeatingSetpoint`, `SystemMode`) and added an option to disable heat pump mode conversion from this PR #4317 which made this thermostat not work correctly for any modes BUT heat pump mode.

**Possible Concerns:** 
- First time contributing to this repo so all comments / concerns are welcome.
- Not sure if adding a non-device option to `exposes` is the proper approach or if there's a better way to handle this.
- Unnecessary comments?

**Notes:**
- The setting defaults to true so it does not break existing setups relying on #4317 
- I was hoping there was something I could poll to get if it's in heat pump mode or not but I was not able to find that. 
- `ctrlSeqeOfOper` is also set to 6 on my thermostat (see #4317) and I couldn't get it to change no matter what settings I changed. I was not able to set the device to heat pump mode, I'm assuming because I do not have the required wires connected.
  - Wondering if this would be a good candidate for a similar option so the user could set their mode to be interpreted correctly downstream, in HA for example (heat only, cool only, etc)

### DIFF
--- a/src/devices/centralite.ts
+++ b/src/devices/centralite.ts
@@ -7,25 +7,40 @@ import * as exposes from '../lib/exposes';
 import * as m from '../lib/modernExtend';
 import * as reporting from '../lib/reporting';
 import * as globalStore from '../lib/store';
-import {DefinitionWithExtend, Fz} from '../lib/types';
+import {DefinitionWithExtend, Fz, Tz} from '../lib/types';
 
 const e = exposes.presets;
 const ea = exposes.access;
+
+const tzLocal = {
+    thermostat_3156105_hpm: {
+        key: ['heat_pump_mode'],
+        convertSet: async (_entity, key, value, meta) => {
+            // Store the setting in the Zigbee2MQTT state only (no device interaction)
+            meta.state[key] = value;
+            return {state: {[key]: value}};
+        },
+    } satisfies Tz.Converter,
+};
 
 const fzLocal = {
     thermostat_3156105: {
         cluster: 'hvacThermostat',
         type: ['attributeReport', 'readResponse'],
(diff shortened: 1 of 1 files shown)


src/devices/centralite.ts
```diff
@@ -7,25 +7,40 @@ import * as exposes from '../lib/exposes';
...
 import * as globalStore from '../lib/store';
-import {DefinitionWithExtend, Fz} from '../lib/types';
+import {DefinitionWithExtend, Fz, Tz} from '../lib/types';
 
 const e = exposes.presets;
 const ea = exposes.access;
 
+const tzLocal = {
+    thermostat_3156105_hpm: {
+        key: ['heat_pump_mode'],
```
Koenkk, line 17: This can be an option, e.g. 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/4ab9c5e424eaba39741175f008dd65749a528622/src/devices/lellki.ts#L73

Than instead of `const useTranslation = meta.state.heat_pump_mode ?? true;` do `const useTranslation = !!options.heat_pump_mode;`

Devbymitch: Yep that's exactly what I was missing, thanks. I added this inline since it's unique to this specific device. Would you rather me abstract it into `exposes`?

Koenkk: Nope, it's perfect!

PR #8811, opened 2025-02-15
Fix local temperature calibration for _TZE204_lpedvtvr 

acasajus:
The local temperature wraps around 256 instead of 4096. Also replaced enum for IN/AL/OU and added description to the exposed control


Koenkk:
Thanks!

PR #8813, opened 2025-02-15
Add support for ZBCMR-01 NovaDigital roller blind motor

LucasTor:
Added support for ZBCMR-01 roller blind motor sold under the NovaDigital brand in Brazil
I actually just used the same TS0601_cover_2 device, just added the fingerprint for it

Here is the generated external definition for reference:
Screenshot 2025-02-15 at 19 30 16

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -4127,10 +4127,10 @@ const definitions: DefinitionWithExtend[] = [
         fingerprint: [...tuya.fingerprint('TS0601', ['_TZE200_eegnwoyw']), ...tuya.fingerprint('TS0105', ['_TZE600_ogyg1y6b'])],
         model: 'TS0601_cover_2',
         vendor: 'Tuya',
-        description: 'Curtain motor fixed speed',
+        description: 'Curtain motor or roller blind motor with fixed speed',
         whiteLabel: [
-            {vendor: 'Zemismart', model: 'BCM100DB'},
-            {vendor: 'NovaDigital', model: 'ZBCMR-01'},
+            tuya.whitelabel('Zemismart', 'BCM100DB', 'Curtain Motor', ['_TZE200_eegnwoyw']),
+            tuya.whitelabel('NovaDigital', 'ZBCMR-01', 'Roller Blind Motor', ['_TZE600_ogyg1y6b']),
         ],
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -4124,11 +4124,14 @@ const definitions: DefinitionWithExtend[] = [
...
     {
-        fingerprint: tuya.fingerprint('TS0601', ['_TZE200_eegnwoyw']),
+        fingerprint: [...tuya.fingerprint('TS0601', ['_TZE200_eegnwoyw']), ...tuya.fingerprint('TS0105', ['_TZE600_ogyg1y6b'])],
         model: 'TS0601_cover_2',
         vendor: 'Tuya',
         description: 'Curtain motor fixed speed',
-        whiteLabel: [{vendor: 'Zemismart', model: 'BCM100DB'}],
+        whiteLabel: [
+            {vendor: 'Zemismart', model: 'BCM100DB'},
+            {vendor: 'NovaDigital', model: 'ZBCMR-01'},
```
Koenkk, line 4133: Could you use tuya.whitelabel here? You can find examples in this file

LucasTor: Done in [423e596](https://github.com/Koenkk/zigbee-herdsman-converters/pull/8813/commits/423e596643d0ad83ed3123e299898b61d4d794d6)

Koenkk:
Thanks!

Make sure that:
- The filename is `ZBCMR-01.png`
- The size is 512x512

capyvara:
Hey @LucasTor do you still have this device? I'm having an issue with it (regardless of pairing with z2m), the configured upper/lower limit drifts over time, gets lower ~1cm reach open/close cycle, making it unusable for automation, I'd like to know if this is something up to the device or I got defective units.

Thanks

LucasTor:
Hey @capyvara! I don't for this exact reason, it happens even after hard resetting it and using just the included RF remote

A bit of shame given how pricey those are, so I ended up returning them to the seller after finding this bug in all (3) of my units 

It seems it has no position feedback and relies purely on timing to set the limits

capyvara:
Thanks @LucasTor , same for me, 3 units with this issue, really seems to be timing based only, maybe not even a stepper motor as clearly when it reaches the limit it decelerates for about one second instead of instant stop.

I'll ask for a return so also., did you find any other good replacement model?

PR #8816, opened 2025-02-16
Update default ac frequency precision

ghoz:
AC frequency is tightly managed by grid operators, fluctuations are really small Curent precision of 0 means the values are not really useful (typical changes are in 0.02 Hz range)


Koenkk:
thanks!

PR #8820, opened 2025-02-16
HS2WD-E: remove unsupported features from exposes

ghoz:
* This [HS2WD-E](https://www.zigbee2mqtt.io/devices/HS2WD-E.html) siren has one tone only, no volume/level, no strobe level. might as well stop exposing those as it causes endless frustrations...
* Exposes a missing max_duration, that is discovered as config.
  * In practice this means that no matter the duration, it will not trigger for longer than the set value to avoid runaway alarms (device defaults to 240 s). 
  * tested, at last a feature that is properly implemented on the siren.

### DIFF
--- a/src/devices/heiman.ts
+++ b/src/devices/heiman.ts
@@ -302,7 +302,13 @@ const definitions: DefinitionWithExtend[] = [
         },
         exposes: [
             e.battery(),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #8825, opened 2025-02-17
[AVATTO Tuya] Fix inverted presets on another model (_TZE204_xnbkhhdr)

Nopraz:


Koenkk:
Thanks!

Nopraz:
Yes sorry about this pull request, i was running an older version...

Nopraz:
I reverted it

pvdw0310:
Hi, thanks for the previous change for the ZWT198/ZWT100-BH preset are inverted (Issue #26567)

Is it possible to look also at the temperature setting of the thermostat.
I like to set te temperature of the thermostat in an automation with "Set value for Eetplaats Cv thermostaat Temperature" instruction, but temperature is not changing.

If I go to developer tools => states => I give eetplaats cv thermostaat temperature in => I get a different value, the calibration value => number.0xa4c138d8d0d851f6_local_temperature_calibration

Is there a way to change that and to set the temperature of the thermostat in an automation.

Thanks in advance to change for the next update.

Peter

PR #8826, opened 2025-02-17
Fix multiple versions of the name Nova Digital

bernardesarthur:
Fix multiple versions of the name Nova Digital.


Koenkk:
thanks!

PR #8829, opened 2025-02-18
Fix the type of the dimmingMode parameter on Inovelli devices to uint8

rohankapoorcom:
Looks like this is not a boolean per @InovelliUSA but is rather an uint8.


Koenkk:
Let me know if this is OK @InovelliUSA

InovelliUSA:
Yep, this is ok.

Koenkk:
Thanks!

PR #8833, opened 2025-02-18
fix: cleanup exports

Nerivec:
- Remove old exports
- Cleanup unused legacy libs
- Cleanup a few JS-specific tests


Koenkk:
Thanks!

PR #8834, opened 2025-02-18
Track HS2WD-E siren in a virtual state

ghoz:
The HS2WD-E siren does not report back its state.
As the siren command has a duration (and a max duration config) , it will  stop at some point, but the triggering system as no way to know that.
Attempts to track current state by looking at the mode, duration (and strobe) when the command is sent.

This is not ready to merge.

2 open questions :
* How to  publish a state from a toZigbee ``convertSet`` context with a delay ? there is no easy ``publish`` available as in fromZigbee ``convert``
* Where to properly implement all this without duplicating tz.warning ? would adding a flag to ``meta`` be the right place to turn on this virtual state ?


src/devices/heiman.ts
```diff
@@ -10,6 +10,69 @@ import {DefinitionWithExtend, Reporting, Zh} from '../lib/types';
...
+
+            // XXX Track and maintain a virtual state of the siren
+            // is on if duration set, and either strobe with dutycycle or alarm is set
+            const siren_on = values.duration > 0 && (values.mode != 'stop' || (values.strobe && value.strobe_duty_cycle > 0));
+
+            // Stop existing timer because we received a new siren command
+            clearTimeout(globalStore.getValue(entity, 'state_timer'));
+            globalStore.clearValue(entity, 'state_timer');
+            if (siren_on) {
+                const duration = Math.min(values.duration, 240);
```
ghoz, line 63: 240 is a placeholder for the actual max duration for now

Koenkk:
> Where to properly implement all this without duplicating tz.warning ? would adding a flag to meta be the right place to turn on this virtual state ?

That's fine

> How to publish a state from a toZigbee convertSet context with a delay ? there is no easy publish available as in fromZigbee convert

What is the use case for this? Do you want to publish siren OFF after a certain time?

ghoz:
> What is the use case for this? Do you want to publish siren OFF after a certain time?

Yes, exactly . Not sure how other warning device return their state through ias Zone maybe, but this one doesn't return anything, which make it's use as an HA siren a bit tricky

I wanted to copy the idea of the occupancy sensors that don't reset their state, but as those timers are created from within fromZigbee convertSet they have access to ``publish``, which is not available here...

Koenkk:
We have to add a `publish` to https://github.com/Koenkk/zigbee2mqtt/blob/1accb8b66b95b99658c50e4c0209edcb946a5fe2/lib/extension/publish.ts#L230 in that case.  Would you mind making a PR for that?

ghoz:
Beside the naming of the meta and the virtual state,
Only quirk left is that when setting the maxDuration, it does update tge cache state, but not the cached zigbee attribute, so a manual read is needed for now

src/converters/toZigbee.ts
```diff
@@ -560,6 +561,27 @@ export const warning: Tz.Converter = {
             {startwarninginfo: info, warningduration: values.duration, strobedutycycle: values.strobeDutyCycle, strobelevel: values.strobeLevel},
             utils.getOptions(meta.mapped, entity),
         );
+
+        // Track and maintain a virtual state of the siren
+        // is on if duration set, and either strobe with dutycycle or alarm is set
+        const virtualState = meta.mapped.meta?.virtualState?.warning;
```
Koenkk, line 567: What holds us from not enabling this for every device?

ghoz: I simply have no clue how other warning devices return their state ... as this siren doesn't support most of the options, it might not be the best example device to generalize

Koenkk: I think it conforms against the ZCL spec so should be fine, let's try it.

ghoz: Made available to all warning devices🤞
The ``siren_state`` will still needs to be exposed

src/lib/types.ts
```diff
@@ -219,6 +219,12 @@ export interface DefinitionMeta {
      * Never use a transition when transitioning to off (even when specified)
      */
     noOffTransitionWhenOff?: boolean | ((entity: Zh.Endpoint) => boolean);
+    /**
+     * Maintain a virtual state for devices that don't report it
+     */
+    virtualState?: {
```
Koenkk, line 225: I think this can be removed now.

src/devices/heiman.ts
```diff
@@ -301,6 +301,7 @@ export const definitions: DefinitionWithExtend[] = [
             await endpoint.read("ssIasWd", ["maxDuration"]);
         },
         exposes: [
+            e.binary("siren_state", ea.STATE, "ON", "OFF").withDescription("Alarm state (Virtual)"),
```
Koenkk, line 304: This can be added to every device having `tz.warning`, maybe good to add it to `exposes.ts`

PR #8837, opened 2025-02-19
Added temperature control accuracy attribute for the trvzb

lcheng33775823:


src/devices/sonoff.ts
```diff
@@ -1318,6 +1319,24 @@ export const definitions: DefinitionWithExtend[] = [
                 valueStep: 1.0,
                 unit: '%',
             }),
+            m.numeric({
+                name: 'temp_accuracy',
```
Koenkk, line 1323:
```suggestion
                name: 'temperature_accuracy',
```

photomoose:
@lcheng33775823 this is great, I've been looking for a way to have more control over the hysteresis. Thanks.

FYI: I've also exposed the external temperature sensor attributes here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/8873

marazmarci:
Using the term "accuracy" can be a bit misleading, and can cause confusion without reading the description (for example only seeing the entity name in HA). A more accurate name could be "Temperature control hysteresis". Or would "hysteresis" be a too scientific term to use? 🤔

shaggee:
Guys do you know if temp accuracy atribute will be added to next release?

shaggee (approved):

photomoose:
> Guys do you know if temp accuracy atribute will be add

Assuming @lcheng33775823 doesn't reply before, I'll pull these changes down tomorrow, test it, and make the changes as suggested by @Koenkk. Then it will likely be in the next release.

bojanpotocnik:
> and make the changes as suggested by @Koenkk

@photomoos what is your opinions on https://github.com/Koenkk/zigbee-herdsman-converters/pull/8837#issuecomment-2677995022 above?
I also think _accuracy_ is a bit unusual wording for writable attribute representing _hysteresis_, _tolerance_, _deadband_, _margin_, ... Technically speaking, _precision_ would be more correct than _accuracy_ in this case :thinking:

Koenkk:
Thanks!

PR #8839, opened 2025-02-19
ubisys: we have OTA images for the H10

sjorge:
@cmaschke I'm 95% sure this will fix the OTA not available you're seeing for the H10.

I was tired last night and accidentally checked the H1 to see if it was enabled.


Koenkk:
thanks!

PR #8841, opened 2025-02-19
[FIX]: Increase TS0601_thermostat_4 maximum heating value and permissible calibration limits

qveensi:
Based on https://github.com/Koenkk/zigbee2mqtt/issues/19462

Hello, I recently [purchased](https://a.aliexpress.com/_Ew3QOXq) 3 thermal heads of this model and noticed that the set values ​​in the templates do not correspond to the actual ones.

My devices are detected as expected as TS0601_thermostat_4.
generated_external_definition:
```
const definition = {
    zigbeeModel: ['TS0601'],
    model: 'TS0601',
    vendor: '_TZE204_pcdmj88b',
    extend: [],
    meta: {},
};

module.exports = definition;
```
each set value is tested on hardware. so please review my pool request.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -4944,15 +4944,15 @@ export const definitions: DefinitionWithExtend[] = [
             e.battery_low(),
             e
                 .climate()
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #8843, opened 2025-02-19
Add SPM01V2 data report duration configuration

vilaemail:
With 3.1.3 firmware for SPM01V2 it is possible to "Data report duration set (Threshold value range 5~3600 seconds)". [See here](https://github.com/Koenkk/zigbee2mqtt/issues/22236#issuecomment-2325492263).

This change adds support for that into zigbee2mqtt.

This addresses [this issue](https://github.com/Koenkk/zigbee2mqtt/issues/26364).


vilaemail:
@Koenkk Not sure why GitHub is showing an error message for this PR

```
Merging is blocked
Merge is not an allowed merge method in this repository.
This branch must not contain merge commits.
```

Can you help me figure this out when you have a moment? Thanks.

src/devices/tuya.ts
```diff
@@ -9672,6 +9672,13 @@ export const definitions: DefinitionWithExtend[] = [
             e.produced_energy().withDescription('Total reverse active energy'),
             e.power_factor().withUnit('%'),
             e.ac_frequency(),
+            e
+                .numeric('data_report_duration', ea.SET)
```
Koenkk, line 9676:
```suggestion
                .numeric('data_report_interval', ea.SET)
```

Would this name make more sense?

vilaemail: I am ok either way, the reason for the name I used is that in Tuya app that is how it is called when updating firmware - [see screenshot here](https://github.com/Koenkk/zigbee2mqtt/issues/22236#issuecomment-2325492263).

Koenkk: Then let's keep it like this.

PR #8859, opened 2025-02-20
fix: Migrate from eslint/prettier to Biome

Koenkk:
CC: @Nerivec 

CI lint/format step went from 27 to 1 second! 😮


package.json
```diff
@@ -20,18 +20,9 @@
...
-        "wemo",
-        "zigbee",
-        "zigbee-shepherd"
-    ],
+    "keywords": ["aqara", "tradfri", "hue", "wemo", "zigbee", "zigbee-shepherd"],
     "scripts": {
-        "eslint": "eslint --max-warnings=0",
-        "pretty:write": "prettier --write .",
-        "pretty:check": "prettier --check .",
+        "biome:check": "biome check",
```
Nerivec, line 25: Should reduce the name of scripts as much as possible. It should at least be faster than calling the thing directly 😅

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
Koenkk: Added, I have no clue about Jetbrains but I assume most use VSCode nowadays.

package.json
Koenkk: You suggest calling it `biome`?

Nerivec: I went with `check` / `check:ci` in ZoH (write/no write) for now.
I guess could also go with the more widely used `format`.

src/index.ts
```diff
@@ -133,13 +133,13 @@
...
 function getFromExternalDefinitionsLookup(zigbeeModel: string | undefined): DefinitionWithExtend[] | undefined {
-    const lookupModel = zigbeeModel ? zigbeeModel.toLowerCase() : 'null';
+    const lookupModel = zigbeeModel ? zigbeeModel.toLowerCase() : "null";
 
     if (externalDefinitionsLookup.has(lookupModel)) {
         return externalDefinitionsLookup.get(lookupModel);
     }
 
-    return externalDefinitionsLookup.get(lookupModel.replace(/\0(.|\n)*$/g, '').trim());
+    return externalDefinitionsLookup.get(lookupModel.replace(/\0(.|\n)*$/g, "").trim());
```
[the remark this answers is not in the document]
T10876: مافام

PR #8860, opened 2025-02-20
Add Support for Enkin - ZDM150 Dimmer Module

encryptix:
https://enkin.com/products/ss-en-00-zdm150-150w-led-zigbee-dimmer-module

This is an external converter I have for a dimmer module and is working well.


src/devices/enkin.js
```diff
@@ -0,0 +1,33 @@
...
+
+const definitions = [{
+    zigbeeModel: ['ZDM150'],
+    model: 'ZDM150',
+    vendor: 'Enkin',
+    description: '150W Dimmer Module',
+    extend: [light({powerOnBehavior: false, effect: false, configureReporting: true}), forcePowerSource({powerSource: 'Mains (single phase)'})],
+
+    configure: async (device, coordinatorEndpoint) => {
+        device.powerSource = 'Mains (single phase)';
```
encryptix, line 13: I don't know if this line is needed when its also on L10

src/devices/enkin.js
```diff
@@ -0,0 +1,33 @@
...
+
+        const endpoint = device.getEndpoint(1);
+        await bind(endpoint, coordinatorEndpoint, ['genOnOff', 'genLevelCtrl']);
```
encryptix, line 16: I don't know what this line or the next line does, copied from other device files for lights.

src/devices/enkin.js
```diff
@@ -0,0 +1,33 @@
...
+
+const definitions = [
+    {
+        zigbeeModel: ['ZDM150'],
+        model: 'ZDM150',
+        vendor: 'Enkin',
+        description: '150W Dimmer Module',
+        extend: [light({powerOnBehavior: false, effect: false, configureReporting: true}), forcePowerSource({powerSource: 'Mains (single phase)'})],
+
+        configure: async (device, coordinatorEndpoint) => {
```
Koenkk, line 12: This shouldn't be needed as you have `configureReporting: true`

encryptix: I see modernExtend does the following automatically
```
                await setupAttributes(device, coordinatorEndpoint, 'genOnOff', [{attribute: 'onOff', min: 'MIN', max: 'MAX', change: 1}]);


            configureSetPowerSourceWhenUnknown('Mains (single phase)'),
``

Is it just these I remove? How do I keep configuration for `genLevelCtrl` ?

Koenkk: Reporting is also setup for `currentLevel` so that also can be removed.

Koenkk:
Thanks!

PR #8867, opened 2025-02-21
feat: Extend support for custom zigbee frames with `zcl_command`

chrivers:
The existing implementation is almost a generic "send zigbee frame" function, but falls short.

Notably, it:

 - Does not allow specifying a Zcl.FrameType value
 - Does not allow suppressing normal ACK reporting
 - Does not allow disabling log output in z2m

When working on implementing support for complicated Zigbee devices (Philips Hue, for instance), these limitations hit hard.

This reasonably simple change makes the `zcl_command` significantly more powerful, and should be fully backwards-compatible. 

The one possible exception is that the logging default is reversed. I'm testing devices that need to stream at a rate of 10-60 packets/sec, so being able to disable the log is actually a critical feature, surprisingly.

### DIFF
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -243,8 +243,18 @@ const converters2 = {
             utils.assertObject(value, key);
             const payload = value.payload !== undefined ? value.payload : {};
             utils.assertEndpoint(entity);
-            await entity.zclCommand(value.cluster, value.command, payload, value.options !== undefined ? value.options : {});
-            logger.info(`Invoked ZCL command ${value.cluster}.${value.command} with payload '${JSON.stringify(payload)}'`, NS);
+            await entity.zclCommand(
+                value.cluster,
+                value.command,
+                payload,
+                value.options !== undefined ? value.options : {},
+                value.log_payload ?? {},
+                value.check_status ?? false,
+                value.frametype ?? Zcl.FrameType.SPECIFIC,
+            );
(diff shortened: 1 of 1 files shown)


chrivers:
The motivation for this PR is to make it possible to run Hue Entertainment Mode over z2m.

See also: https://github.com/Koenkk/zigbee-herdsman/pull/1322

src/converters/toZigbee.ts
```diff
@@ -243,8 +243,18 @@ const converters2 = {
...
             const payload = value.payload !== undefined ? value.payload : {};
             utils.assertEndpoint(entity);
-            await entity.zclCommand(value.cluster, value.command, payload, value.options !== undefined ? value.options : {});
-            logger.info(`Invoked ZCL command ${value.cluster}.${value.command} with payload '${JSON.stringify(payload)}'`, NS);
+            await entity.zclCommand(
+                value.cluster,
+                value.command,
+                payload,
+                value.options !== undefined ? value.options : {},
+                value.logPayload ?? {},
```
Koenkk, line 251:
```suggestion
                value.log_payload ?? {},
```

Could you use snake case here? (same for the others)

chrivers:
You got it - one moment :)

chrivers:
Ready for another look - let me know if you'd like anything else fixed :)

Koenkk:
Thanks!

PR #8869, opened 2025-02-22
fix: Publish dist folder instead of root

Koenkk:
Also fixes `build:watch` that is currently not working.
 
CC: @Nerivec


Nerivec:
Note: `build:watch` can't work properly for ZHC, since changes to models/indexes in `devices` would require re-running the indexer.

Koenkk:
Fixed that!

PR #8870, opened 2025-02-22
Add the "ZY-M100-24GV3.1" human body sensor. The detection distance of this version of the human body sensor has been updated.

yuanxin32323:
The detection distance of this version of the human body sensor has been updated.


Koenkk:
Please use the `exposes()` function instead to prevent code duplication, example:

yuanxin32323:
> Please use the `exposes()` function instead to prevent code duplication, example:
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/54de561532d8f3400389031c84fce32c59eec069/src/devices/tuya.ts#L937

OK，I have already merged these two devices.

Koenkk:
Thanks!

PR #8872, opened 2025-02-22
fix: Add `manuSpecificPhilips3` to every Philips light

Koenkk:
Replaces https://github.com/Koenkk/zigbee-herdsman/pull/1322

CC: @chrivers


src/lib/philips.ts
```diff
@@ -53,80 +54,113 @@ export const knownEffects = {
...
-    return result;
-}
 
-export function philipsOnOff(args?: modernExtend.OnOffArgs) {
-    args = {powerOnBehavior: false, ota: true, ...args};
-    const result = modernExtend.onOff(args);
-    result.toZigbee.push(philipsTz.hue_power_on_behavior, philipsTz.hue_power_on_error);
-    return result;
-}
+        const customCluster = m.deviceAddCustomCluster("manuSpecificPhilips3", {
```
Koenkk, line 97: @chrivers lot's of refactoring here, but this is basically the thing that is needed. Let me know if this is OK and I will merge it.

chrivers: Wow! That's amazing @Koenkk :D

Let me just give this a look!

PR #8875, opened 2025-02-23
adds publish to Tz.Meta type

ghoz:
In order to be able to publish from the context of toZigbee  convertSet 
adds publish to Tz.Meta type

Required to implement a virtual siren state in https://github.com/Koenkk/zigbee-herdsman-converters/pull/8834

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -328,6 +328,7 @@ export namespace Tz {
         state: KeyValue;
         endpoint_name: string | undefined;
         membersState?: {[s: string]: KeyValue};
+        publish?: Publish;
     }
     // biome-ignore lint/suspicious/noConfusingVoidType: allow for converters
     export type ConvertSetResult = {state?: KeyValue; membersState?: {[s: string]: KeyValue}} | void;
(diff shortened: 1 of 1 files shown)


src/lib/types.ts
```diff
@@ -328,6 +328,7 @@ export namespace Tz {
         state: KeyValue;
         endpoint_name: string | undefined;
         membersState?: {[s: string]: KeyValue};
+        publish?: Publish;
```
Koenkk, line 331:
```suggestion
        publish: Publish;
```

I think it should not be optional

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt/pull/26490

src/lib/types.ts
ghoz: As it is used in other places I took the easy way out 😅

Koenkk: I think it should be mandatory, would this give any problems?

ghoz: Re checking it is used in z2m availabilty

Only problem I see would be compatibility for other users of zh /zhc

Koenkk: That is not a problem, next zhc release is a breaking one anyway

Koenkk:
Thanks!

PR #8877, opened 2025-02-23
Feat poll profalux remote battery

ghoz:
Profalux remotes do not report battery state, they do however support reading voltage.
As the remote is otherwise chatty (lots of OTA requests amongs others) , polls the battery at most once a day, when receiving a message .

Tested on the wallmounted version, the portable remotes should work too as it is the same electronic in a different case

Note : the % is estimated based on voltage, the remote stops working at 2.25v, but reports a min voltage of 2.0v
`` 'genPowerCfg': {"batteryVoltThres1":22,"batteryVoltThres2":25,"batteryVoltMinThres":20,"batteryVoltage":29}``

Slight problems :
* can't get the power source to appear as a battery in the UI, it is still a question mark
* using the modernExtend for the battery,  both voltage and calculated % allow get, not sure if it makes sense, and how to override the access to STATE only


Koenkk:
> Profalux remotes do not report battery state, they do however support reading voltage.

Does the voltage reporting also not work?

ghoz:
No reporting of voltage...
Just checked , it fails

src/devices/profalux.ts
```diff
@@ -148,5 +166,10 @@ export const definitions: DefinitionWithExtend[] = [
         fromZigbee: [],
         toZigbee: [],
         exposes: [],
```
Koenkk, line 168: fromZigbee/toZigbee/exposes can be removed

src/devices/profalux.ts
```diff
@@ -148,5 +166,10 @@ export const definitions: DefinitionWithExtend[] = [
...
+        extend: [
+            m.battery({voltage: true, voltageToPercentage: {min: 2200, max: 3100}, percentageReporting: false}),
+            m.forcePowerSource({powerSource: "Battery"}),
+        ],
+        onEvent: onEventBatteryPoll,
```
Koenkk, line 173: Please add a note here why this is needed

src/devices/profalux.ts
```diff
@@ -2,13 +2,26 @@ import fz from "../converters/fromZigbee";
...
 
 const NS = "zhc:profalux";
 const e = exposes.presets;
 const ea = exposes.access;
 
+const DAY = 86400000;
+
+async function onEventBatteryPoll(type: OnEventType, data: OnEventData, device: Zh.Device, options: KeyValue) {
+    if (type === "message" && Date.now() > globalStore.getValue(device, "battery_nextpoll", 0)) {
+        const endpoint = device.getEndpoint(2);
```
Koenkk, line 18:
```suggestion
        const endpoint = device.endpoints.find((e) => e.supportsInputCluster("genPowerCfg"));
```

ghoz:
Is there anything else to do for the power source to show as battery in the UI ?
I expected the ``forcePowerSource`` to do that...

Koenkk:
Thanks!

PR #8882, opened 2025-02-24
Ported Custom Philips Clusters From Herdsman To ModernExtends

danieledwardgeorgehitchcock:
This is a draft pull request intended to begin the moving of `manufacturerSpecific` clusters out of the Herdsman Clusters specification and into the `modernExtends` format in preparation for some larger re-factoring taking place for the work done by @chrivers in the Bifrost project.

At present, this is just a lift and drop of the cluster specs as they are in Herdsman (will create a separate PR in that repo for removal).

Once I have got this over the line, I will begin porting across the newly documented cluster attributes / commands / responses that have been recently documented

### DIFF
--- a/src/lib/philips.ts
+++ b/src/lib/philips.ts
@@ -94,7 +94,44 @@ const philipsModernExtend = {
             result.exposes.push(...exposeEndpoints(e.enum("effect", ea.SET, effects), args.endpointNames));
         }
 
-        const customCluster = m.deviceAddCustomCluster("manuSpecificPhilips3", {
+        const customCluster = m.deviceAddCustomCluster("manuSpecificPhilips", {
+            ID: 0xfc00,
+            manufacturerCode: Zcl.ManufacturerCode.SIGNIFY_NETHERLANDS_B_V,
+            attributes: {
+                config: {ID: 49, type: Zcl.DataType.BITMAP16},
+            },
+            commands: {},
+            commandsResponse: {
+                hueNotification: {
+                    ID: 0,
+                    parameters: [
+                        {name: 'button', type: Zcl.DataType.UINT8},
+                        {name: 'unknown1', type: Zcl.DataType.UINT24},
+                        {name: 'type', type: Zcl.DataType.UINT8},
+                        {name: 'unknown2', type: Zcl.DataType.UINT8},
+                        {name: 'time', type: Zcl.DataType.UINT8},
+                        {name: 'unknown2', type: Zcl.DataType.UINT8},
+                    ],
+                },
+            },
+        });
+
+        const customCluster2 = m.deviceAddCustomCluster("manuSpecificPhilips2", {
+            ID: 0xfc03,
+            manufacturerCode: Zcl.ManufacturerCode.SIGNIFY_NETHERLANDS_B_V,
+            attributes: {
+                state: {ID: 0x0002, type: Zcl.DataType.OCTET_STR},
+            },
+            commands: {
+                multiColor: {
+                    ID: 0,
+                    parameters: [{name: 'data', type: Zcl.BuffaloZclDataType.BUFFER}],
+                },
+            },
+            commandsResponse: {},
+        });
+
+        const customCluster3 = m.deviceAddCustomCluster("manuSpecificPhilips3", {
             ID: 0xfc01,
             manufacturerCode: Zcl.ManufacturerCode.SIGNIFY_NETHERLANDS_B_V,
             attributes: {},
@@ -122,8 +159,18 @@ const philipsModernExtend = {
             },
             commandsResponse: {},
         });
-        result.onEvent = [...(result.onEvent ?? []), ...customCluster.onEvent];
-        result.configure = [...(result.configure ?? []), ...customCluster.configure];
+        result.onEvent = [
+            ...(result.onEvent ?? []),
+            ...customCluster.onEvent,
+            ...customCluster2.onEvent,
+            ...customCluster3.onEvent
+        ];
+        result.configure = [
+            ...(result.configure ?? []),
+            ...customCluster.configure,
+            ...customCluster2.configure,
+            ...customCluster3.configure
+        ];
 
         return result;
     },
@@ -158,7 +205,7 @@ const philipsModernExtend = {
         return result;
     },
 };
-export {philipsModernExtend as m};
+export {philipsModernExtend as m};.edw
 
 const philipsTz = {
     gradient_scene: {


src/lib/philips.ts
```diff
@@ -94,7 +94,44 @@ const philipsModernExtend = {
...
+                        {name: 'type', type: Zcl.DataType.UINT8},
+                        {name: 'unknown2', type: Zcl.DataType.UINT8},
+                        {name: 'time', type: Zcl.DataType.UINT8},
+                        {name: 'unknown2', type: Zcl.DataType.UINT8},
+                    ],
+                },
+            },
+        });
+
+        const customCluster2 = m.deviceAddCustomCluster("manuSpecificPhilips2", {
```
chrivers, line 119: Suggestion: Name these consts after the cluster id (`customClusterFC03`, or such). That way, we never have to renumber them, if we find any new clusters.

chrivers: I don't know if it's acceptable to rename the clusters themselves. Maybe it's possible to add an alias name, and deprecate the `manuSpecificPhilipsX` names over time?

danieledwardgeorgehitchcock: See latest commit

danieledwardgeorgehitchcock: The cluster namings will be part of a larger re-factor. I have aired on the side of a lift and drop for now, with segmented changes in subsequent PRs to break up the monolithic nature of what we will need to do

src/lib/philips.ts
```diff
@@ -122,8 +159,13 @@ const philipsModernExtend = {
...
             commandsResponse: {},
         });
-        result.onEvent = [...(result.onEvent ?? []), ...customCluster.onEvent];
-        result.configure = [...(result.configure ?? []), ...customCluster.configure];
+        result.onEvent = [...(result.onEvent ?? []), ...customClusterFC00.onEvent, ...customClusterFC03.onEvent, ...customClusterFC01.onEvent];
+        result.configure = [
+            ...(result.configure ?? []),
+            ...customClusterFC00.configure,
+            ...customClusterFC03.configure,
+            ...customClusterFC01.configure,
```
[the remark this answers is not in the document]
danieledwardgeorgehitchcock: Done!

src/lib/philips.ts
```diff
@@ -94,7 +94,44 @@ const philipsModernExtend = {
...
+            commands: {
+                multiColor: {
+                    ID: 0,
+                    parameters: [{name: "data", type: Zcl.BuffaloZclDataType.BUFFER}],
+                },
+            },
+            commandsResponse: {},
+        });
+
+        const customClusterFC01 = m.deviceAddCustomCluster("manuSpecificPhilips3", {
```
chrivers, line 134: Any chance we can have `const customClusterFC03` appear after `const customClusterFC01`? I realize this means `manuSpecificPhilips3` will come before `manuSpecificPhilips2`, but now that we're getting a better and better idea what these do, hopefully we can (at some later point, in a future PR) rename the `manuSpecific...` to more meaningful names.

danieledwardgeorgehitchcock: Done!

src/lib/philips.ts
```diff
@@ -94,7 +94,29 @@ const philipsModernExtend = {
             result.exposes.push(...exposeEndpoints(e.enum("effect", ea.SET, effects), args.endpointNames));
         }
 
-        const customCluster = m.deviceAddCustomCluster("manuSpecificPhilips3", {
+        const customClusterFC00 = m.deviceAddCustomCluster("manuSpecificPhilips", {
```
Koenkk, line 97: This code is only executed for Philips devices using `philipsLight()`, there are also e.g. plugs and motion sensors that have these clusters. Maybe we should create a new `philipsCustomClusters()` modern extend?

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee-herdsman/pull/1326

src/lib/philips.ts
danieledwardgeorgehitchcock: Thanks for the review! I'll get right on it!

danieledwardgeorgehitchcock: > This code is only executed for Philips devices using `philipsLight()`, there are also e.g. plugs and motion sensors that have these clusters. Maybe we should create a new `philipsCustomClusters()` modern extend?

This is looking like a larger job than I expected.

My thoughts so far:

- Cluster `FC00` can be stripped out from Light as, I cannot find any light fixture that uses it
- I can port Philips only elements from `fromZigbee` and `toZigbee` into the new `modernExtends` light definition as part of this request

What I am hesitant on doing (for now) is to do as you suggested and create a new extends definition for Switches which uses the `FC00` cluster as part of this PR - I am trying to keep PR's as small and non-monolithic as possible so that things are easier to test (and revert if needed)

My suggestion (if you'd be happy), is for me to do the changes in the bullets above and keep that as the scope of this PR.

Actions for subsequent PR's:

- Create switch `modernExtends` and port FC00 cluster and port definitions from other areas
- Remove cluster definitions from Herdsman (done, PR raised)
- Re-Name clusters to more descriptive names
- Add additional cluster attributes / commands that have been documented by Christian in the Bifrost project

Koenkk: I don't see much risk with adding the custom clusters through the modern extend, could you elaborate more where you got stuck?

danieledwardgeorgehitchcock: Maybe I am just misunderstanding and going off on a tangent.

Here is what I have done so far:
https://github.com/danieledwardgeorgehitchcock/zigbee-herdsman-converters/tree/philips_switch_modern_extends

Koenkk: I would not expect this: https://github.com/danieledwardgeorgehitchcock/zigbee-herdsman-converters/commit/2b7cecf0d87467e93d1c519065dbb8c1556a5dae

What I expected:
- Add a new modern extend to https://github.com/Koenkk/zigbee-herdsman-converters/blob/4b7c15c91a5f8582c18ab7a4d46123a1db45d062/src/lib/philips.ts#L57, `customClusters` which only does a `m.deviceAddCustomCluster` for every cluster
- Add the modern extend to every Philips device, so every device in `philips.ts` has `extend: [philips.m.customClusters(), ...]`
- Also add it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/4b7c15c91a5f8582c18ab7a4d46123a1db45d062/src/lib/generateDefinition.ts#L303

danieledwardgeorgehitchcock: I did think it was a lot! 😂 Hence why I branched..

I will have another go..

There is some stuff I've done though that I may bring over to other PRs.. so, I don't think it's necessarily wasted effort

danieledwardgeorgehitchcock: Could you look at https://github.com/Koenkk/zigbee-herdsman-converters/blob/e8acd7fca48e9dce33b1d7da477cc9f8bc7b1202/src/lib/philips.ts#L203 where I have had a go at splitting it out? Please let me know if I am in the right ballpark - I am new to the `modernExtends` concept and out of my comfort zone here :-)

Koenkk: > where I have had a go at splitting it out?

I don't understand what you mean with this, could you elaborate a bit more on this?

chrivers:
This isn't stale, but I can't remove the label. @Koenkk, a little help please? :)

marazmarci:
FYI another PR was just merged: [Move manuSpecificPhilips and manuSpecificPhilips2 from ZH to ZHC #11648](https://github.com/Koenkk/zigbee-herdsman-converters/pull/11648)

I didn't dive into the topic too deeply, but I guess it achieves the same goal what this PR tried to achieve.

PR #8885, opened 2025-02-24
Cleanup exports in fz/tz

Nerivec:
- Remove local objects in favor of plain exports
- Remove default export (can `* as` if needed)
  - should allow removing default imports in future, for places where few items used (for now all moved to `* as`)
- Add type (instead of `satisfies`) to fz/tz and fix associated issues (in separate commit to easily identify changes)


Koenkk:
Thanks!

PR #8904, opened 2025-02-27
Modify product model

sdbldn:
This section of the thermostat is not only suitable for GC models, but also for multiple models including GA, GB, and GC, so this suffix is removed.


Koenkk:
Thanks!

PR #8905, opened 2025-02-27
Tuya Air Quality _TZE200_dwcarsat: Filter more strange values for PM2.5

McGiverGim:
My `_TZE200_dwcarsat` shows some strange peaks very out of range:

image

So I added the two values (43521 and 43522) that I see in the zigbee2mqtt log to the actual rule that filters this kind of values. Maybe its better to simply filter any value bigger than `0xaa00` (43520) or simply the max value of PM2.5 that for this device is 100.

### DIFF
--- a/src/lib/legacy.ts
+++ b/src/lib/legacy.ts
@@ -1939,7 +1939,9 @@ const fromZigbee = {
                 case dataPoints.tuyaSabCO2:
                     if (["_TZE200_dwcarsat", "_TZE204_dwcarsat"].includes(meta.device.manufacturerName)) {
                         // Ignore: https://github.com/Koenkk/zigbee2mqtt/issues/11033#issuecomment-1109808552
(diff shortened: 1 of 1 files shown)


McGiverGim:
Added the 9254 (0x2426), it appears in the logs as strange value too.

Koenkk:
> Maybe its better to simply filter any value bigger than 0xaa00 (43520)

This sounds like a more durable solution, could you update this PR?

McGiverGim:
> Added the 9254 (0x2426), it appears in the logs as strange value too.

@Koenkk there is at least one bad value smaller than 43520. I can put the 9254 or the maximum value of the valid range of the device. I will look in the manual and change it.

McGiverGim:
@Koenkk done.

I found more big bad values, so I ended filtering any value bigger than the range published in the manual, for PM2.5 the valid range is 0-1000 ug/m3.

Koenkk:
Thanks!

PR #8907, opened 2025-02-27
Update candeo.ts

candeodevelopment:
Updates Candeo C203:

1) Adds additional model ID to support historically named devices 
2) Limits Level Cluster features to suit device
3) Configures Level Cluster options to suit device 
4) Adds ability to change switch input type between momentary and toggle

### DIFF
--- a/src/devices/candeo.ts
+++ b/src/devices/candeo.ts
@@ -242,7 +242,9 @@ export const definitions: DefinitionWithExtend[] = [
         model: "C203",
         vendor: "Candeo",
         description: "Zigbee micro smart dimmer",
-        extend: [m.light({configureReporting: true, levelConfig: {disabledFeatures: ["on_transition_time", "off_transition_time", "execute_if_off"]}})],
+        extend: [
+            m.light({configureReporting: true, levelConfig: {disabledFeatures: ["on_transition_time", "off_transition_time", "execute_if_off"]}}),
+        ],
         fromZigbee: [fzLocal.switch_type, fz.ignore_genOta],
         toZigbee: [tzLocal.switch_type],
         exposes: [e.enum("external_switch_type", ea.ALL, ["momentary", "toggle"]).withLabel("External switch type")],
--- a/src/devices/candeo.ts
+++ b/src/devices/candeo.ts
@@ -242,21 +242,12 @@ export const definitions: DefinitionWithExtend[] = [
         model: "C203",
         vendor: "Candeo",
         description: "Zigbee micro smart dimmer",
-        fromZigbee: [fz.on_off, fz.brightness, fz.level_config, fz.power_on_behavior, fzLocal.switch_type, fz.ignore_genOta],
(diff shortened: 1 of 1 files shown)


src/devices/candeo.ts
```diff
@@ -180,10 +235,38 @@ export const definitions: DefinitionWithExtend[] = [
...
     {
-        fingerprint: [{modelID: "C203", manufacturerName: "Candeo"}],
+        fingerprint: [
+            {modelID: "C203", manufacturerName: "Candeo"},
+            {modelID: "HK-LN-DIM-A", manufacturerName: "Candeo"},
+        ],
         model: "C203",
         vendor: "Candeo",
         description: "Zigbee micro smart dimmer",
-        extend: [m.light({configureReporting: true})],
```
Koenkk, line 187: Why remove modern extend here? Please use that instead of the `configure` and `from/toZigbee` converters

candeodevelopment: How do we use modern extend to:

1) Set defaults for things like OnLevel, OnOffTransitionTime, etc, that we wish for our devices to have when initially paired
2) Configure reporting change / min / max to suit our devices
3) Add our fzLocal.switch_type, tzLocal.switch_type and e.enum() exposes to allow setting of the switch input between momentary and toggle?

Koenkk: 1 and 3. This you cannot (so keep as is)
2. I hope the default is OK, do you have any problems with this?

candeodevelopment: If it's not currently possible with modern extend, are we able to use the code from the PR in the meantime?

We already offer an external converter that does all of this for our customers, but would prefer the code built in to Z2M by default so that our devices work out-of-the-box with Z2M.

Koenkk: I think you can still use modern extends partially, e.g. this allows to get rid of:

```js
fromZigbee: [fz.on_off, fz.brightness, fz.level_config, fz.power_on_behavior
toZigbee: [tz.light_onoff_brightness, tz.level_config, tz.power_on_behavior, tz.identify
``` 

and some exposes

candeodevelopment: We're not able to get this working ourselves, can we use the code from the PR as is?  Or can you offer some guidance on how to achieve the end result using modern extend?

Koenkk: Updated the code, let me know if this is OK

candeodevelopment: We tested this but it doesn't seem to create the correct exposes for the device.

With our original code:

image

With your suggested code:

image

Koenkk: Updated it, could you check again?

candeodevelopment: Yes, that seems to work well now.

(I've just adjusted it slightly to remove the non-supported features)

Koenkk:
Thanks!

Issue #8901, opened 2025-02-27
About exposes.power of electricityMeter() in modernExtend


shinasys:
I have a question regarding exposes.power when using electricityMeter() in modernExtend.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/cba3082d1a9594b4aa74e44d6dccd0ba5d6e1a8b/src/lib/modernExtend.ts#L1888C9-L1905C49

 I want seMetering.power to be set, but electrical_measurement_power is being set instead. electrical_measurement_power is a signed int16 type, so it can only represent up to 32768W. I want to use seMetering.power, which is a signed int24 type, to represent larger values. 
Is there a way to do this?

Koenkk:
Let's continue in https://github.com/Koenkk/zigbee-herdsman-converters/pull/8909

PR #8909, opened 2025-02-28
Add configuration for the reporting type of power in electricityMeter…

shinasys:
https://github.com/Koenkk/zigbee-herdsman-converters/issues/8901#issue-2883172642

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -1666,6 +1666,7 @@ interface MeterArgs {
     configureReporting?: boolean;
     endpointNames?: string[];
     fzMetering?: Fz.Converter;
+    reportingPowerType?: "metering" | "electrical";
     // applies only to electrical
     electricalMeasurementType?: "both" | "ac" | "dc";
     voltage?: false | (MultiplierDivisor & Partial<ReportingConfigWithoutAttribute>);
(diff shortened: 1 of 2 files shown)


shinasys:
@Koenkk I encountered 'Merging is blocked.' How can I proceed?
Thanks.

src/lib/modernExtend.ts
```diff
@@ -1894,15 +1895,16 @@ function genericMeter(args?: MeterArgs) {
         if (args.producedEnergy !== false) exposes.push(e.produced_energy().withAccess(ea.STATE_GET));
         fromZigbee = [args.fzElectricalMeasurement ?? fz.electrical_measurement, args.fzMetering ?? fz.metering];
         toZigbee = [
-            tz.electrical_measurement_power,
+            args.reportingPowerType === "electrical" ? tz.electrical_measurement_power : tz.metering_power,
```
Koenkk, line 1898: Can't we base this off the `cluster` parameter?

shinasys: I don't have a good idea at the moment. I can't find a suitable parameter in the cluster. Do you have any good methods in mind? 
Thanks.

shinasys: Do you mean adding an argument such as 'both_meter' in the args.cluster parameter?

Koenkk: I see what you want to do now, maybe we can do something like this?

```js
power?: false | (MultiplierDivisor & Partial<ReportingConfigWithoutAttribute> & {cluster?: "metering" | "electrical"});
```

shinasys: Oh, I see now. Thanks for clarifying. I will apply this immediately.

src/lib/modernExtend.ts
```diff
@@ -1894,15 +1894,18 @@ function genericMeter(args?: MeterArgs) {
         if (args.producedEnergy !== false) exposes.push(e.produced_energy().withAccess(ea.STATE_GET));
         fromZigbee = [args.fzElectricalMeasurement ?? fz.electrical_measurement, args.fzMetering ?? fz.metering];
         toZigbee = [
```
Koenkk, line 1896:
```suggestion
        const useMetering = args.power !== false && args.power?.cluster === "metering";
        toZigbee = [
```

Then use it below (to avoid having the same condition twice.

Koenkk:
Thanks!

PR #8912, opened 2025-03-01
Support ZS05, universal smart IR remote control

Virusmater:
Add initial support, picture of the device:
image
Let me know if something should be changed. This is my first time contributing to this repo. 
I will update the docs website once that PR is accepted. Let me know if it should be done beforehand.

Thanks!

### DIFF
--- a/src/devices/wmun.ts
+++ b/src/devices/wmun.ts
@@ -1,0 +1,36 @@
+import * as fz from "../converters/fromZigbee";
+import * as exposes from "../lib/exposes";
+import * as reporting from "../lib/reporting";
+const e = exposes.presets;
+const zosung = require("zigbee-herdsman-converters/lib/zosung");
+const fzZosung = zosung.fzZosung;
+const tzZosung = zosung.tzZosung;
+const ez = zosung.presetsZosung;
+import type {DefinitionWithExtend} from "../lib/types";
+
+export const definitions: DefinitionWithExtend[] = [
+    {
+        fingerprint: [{modelID: "TS1201", manufacturerName: "_TZ3290_u9xac5rv"}],
+        model: "ZS05",
+        vendor: "WMUN",
+        description: "Universal smart IR remote control",
+        fromZigbee: [
+            fzZosung.zosung_send_ir_code_00,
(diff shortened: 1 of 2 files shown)


src/devices/wmun.ts
```diff
@@ -0,0 +1,36 @@
...
+const e = exposes.presets;
+const zosung = require("zigbee-herdsman-converters/lib/zosung");
+const fzZosung = zosung.fzZosung;
+const tzZosung = zosung.tzZosung;
+const ez = zosung.presetsZosung;
+import type {DefinitionWithExtend} from "../lib/types";
+
+export const definitions: DefinitionWithExtend[] = [
+    {
+        fingerprint: [{modelID: "TS1201", manufacturerName: "_TZ3290_u9xac5rv"}],
```
Koenkk, line 13: Please add to `tuya.ts` -> `ZS06` instead (use `tuya.whitelabel` for proper identification)

https://github.com/Koenkk/zigbee-herdsman-converters/blob/cba3082d1a9594b4aa74e44d6dccd0ba5d6e1a8b/src/devices/tuya.ts#L9212

Virusmater:
@Koenkk, that a good catch, but I think it deserves a separate entity because it is a battery-operated one.
What would be the best way to declare the battery level without messing up with USB powered models?

Koenkk:
Missed that, yes than keep it separate.

- Also add it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts
- Use `extend: [m.battery()]` instead of the exposes/configure/fromZigbee battery code. (search this repo for examples)

Virusmater:
Made changes. 

By the way - is there an easy way to test it before building? Placing the file to external_converters doesn't work because it is *.ts file and not *.js, therefore doesn't support imports.

Koenkk:
The import checking is covered by the TS compiler so should be good, thanks!

Make sure that:
- The filename is `ZS05.png`
- The size is 512x512

Virusmater:
Sure, done https://github.com/Koenkk/zigbee2mqtt.io/pull/3561

PR #8917, opened 2025-03-02
Update tuya.ts

1tigris1:
"The switch has been changed to DIN rail NOUS DIN Switch _TZ3000_2iiimqs9 as a common device with three variations."


Koenkk:
Thanks!

PR #8919, opened 2025-03-02
Add device : An infrared human presence sensor with long battery life, powered by two AAA batteries.

yuanxin32323:


src/devices/tuya.ts
```diff
@@ -12156,6 +12156,67 @@ export const definitions: DefinitionWithExtend[] = [
...
+	    ],
+	    meta: {
+	        tuyaDatapoints: [
+				[1, null, {
+					from: function(v, meta) {
+				
+						if (v == 0) {
+							return {
+								state: 'pir',
+								occupancy: true,
```
Koenkk, line 12195: How is this reset back to `false`? (or should it be called `presence`?)

src/devices/tuya.ts
```diff
@@ -12156,6 +12156,67 @@ export const definitions: DefinitionWithExtend[] = [
...
+	            .withDescription("Fade time"),
+	    ],
+	    meta: {
+	        tuyaDatapoints: [
+				[1, null, {
+					from: function(v, meta) {
+				
+						if (v == 0) {
+							return {
+								state: 'pir',
```
Koenkk, line 12194: What is the added value of `state` over `presence`?

src/devices/tuya.ts
yuanxin32323: In the document provided by the manufacturer, the data type of this parameter is "enum", and the enumerated values are: "pir", "none".

[24G人体存在传感器（电池版）(zaydjiac)_08_25.xlsx](https://github.com/user-attachments/files/19055424/24G.zaydjiac._08_25.xlsx)

yuanxin32323: > How is this reset back to `false`? (or should it be called `presence`?)如何将其重置回 `false` ？（或者应该称为 `presence` ？）

I'm really sorry. I wrote the parameter name incorrectly. It should all be "occupancy".

Koenkk:
@yuanxin32323 could you answer the questions? e.g. https://github.com/Koenkk/zigbee-herdsman-converters/pull/8919/files#r1976718849

PR #8920, opened 2025-03-02
Fix sensor select and backlight switch for _TZE204_lpedvtvr

szheliab:
Fixed sensor selector (IN/AL/OU) and converted backlight entity from sensor to switch. Previously both of these entities had 'null' value.


Koenkk:
Thanks!

PR #8925, opened 2025-03-03
update 3RSMR01067Z

3reality-support:


src/devices/third_reality.ts
```diff
@@ -152,7 +152,9 @@ export const definitions: DefinitionWithExtend[] = [
         model: "3RSMR01067Z",
         vendor: "Third Reality",
         description: "Smart motion sensor R1",
+        fromZigbee: [fz.ias_occupancy_alarm_1],
         ota: true,
+        exposes: [e.occupancy()],
         extend: [
```
Koenkk, line 158: Try with this in extend instead: `m.iasZoneAlarm({zoneType: "occupancy", zoneAttributes: ["alarm_1"]})`

3reality-support: ok，it has already been modified

Koenkk:
Thanks!

PR #8927, opened 2025-03-03
[FIX]: TS0601_thermostat_4 changing the local temperature calibration step

qveensi:
Based on https://github.com/Koenkk/zigbee2mqtt/issues/19462

Unfortunately, I made a mistake in my previous correction, so I want to fix it. so that no one gets into trouble because of me.


Koenkk:
Thanks!

Magkoooh (approved):

PR #8931, opened 2025-03-03
fix: Biome improvements

Koenkk:
TODO:
- [x] https://biomejs.dev/linter/rules/use-naming-convention/


Nerivec:
Might have to put `useNamingConvention` it as a warning for ZHC though, lots of variables with snake case that would break ext. converters if renamed (or lots of biome ignore required). MQTT-bound names will also create troubles.
`strictCase` option will no doubt generate a lot of refactoring (e.g. `HTTPServer` => `HttpServer`). Same for the default PascalCase for enum values (following TypeScript maintainers convention).

Koenkk:
@Nerivec I added it without making breaking changes

PR #8845, opened 2025-02-19
PA-44Z: expose the test property to ring the alarm

RemyRoux:
This change aims to enable the alarm to be triggered via home assistant.

I'm a new user so I'm not sure whether this would work. Intuitively, the alarm should be able to be triggered by home assistant. The test button rings the alarm so I'd assume that setting the property to True would ring the alarm. It's worth a try.

PR #8935, opened 2025-03-04
REVERT: #8845 PA-44Z: expose the test property to ring the alarm

RemyRoux:
Reverting as this approach is not working. 
According to the manufacturer, there should be a datapoint to trigger the alarm via Z2M but I don't have a tuya hub, and copying other devices' config didn't work.


Koenkk:
Thanks!

PR #8940, opened 2025-03-06
feat: add external switch type extension to Samotech devices and move Sunricher extend functions to a unified module

niracler:


Koenkk:
Thanks!

PR #8943, opened 2025-03-07
Add label to modern extends binary function

canardos:
The `numeric` and `enumLookup` modern extend functions both support the `label` argument. This PR adds support to `binary` also.


Koenkk:
Thanks!

PR #8944, opened 2025-03-07
Added Envilar 7848 Dimmable LED Driver

HRading:
Added Envilar 7848 Dimmable LED Driver.

Not sure if I got the format correct, as this is my first time adding a converter.

This is the external converter i have working:
```
const philips = require('zigbee-herdsman-converters/lib/philips');

const definition = {
    zigbeeModel: ['7848'],
    model: '7848',
    vendor: 'ENVILAR',
    description: 'Dimmable LED Driver',
    extend: [philips.m.light()],
    meta: {},
};

module.exports = definition;
```


Koenkk:
Thanks!

PR #8965, opened 2025-03-10
reportAttribute modernExtend

kirovilya:
reportable lowState for battery modernExtent

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -438,7 +438,9 @@ export function battery(args?: BatteryArgs): ModernExtend {
 
     if (args.percentageReporting || args.voltageReporting) {
         if (args.percentageReporting) {
-            result.configure.push(setupConfigureForReporting("genPowerCfg", "batteryPercentageRemaining", args.percentageReportingConfig, ea.STATE_GET));
+            result.configure.push(
+                setupConfigureForReporting("genPowerCfg", "batteryPercentageRemaining", args.percentageReportingConfig, ea.STATE_GET),
+            );
         }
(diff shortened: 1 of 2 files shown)


src/lib/modernExtend.ts
```diff
@@ -454,6 +455,12 @@ export function battery(args?: BatteryArgs): ModernExtend {
         result.meta = meta;
     }
 
+    if (args.lowStatusReportingConfig) {
+        const configure: Configure[] = result.configure || [];
```
Koenkk, line 459:
```js
const result: ModernExtend = {exposes, fromZigbee, toZigbee, isModernExtend: true};
result.configure = []
```

to prevent creating a configure here and under `percentageReporting`

kirovilya: done

src/lib/modernExtend.ts
```diff
@@ -2686,4 +2691,17 @@ export function bindCluster(args: {cluster: string | number; clusterType: "input
     return {configure, isModernExtend: true};
 }
 
+export function reportAttribute(args: {
```
Koenkk, line 2694: I don't think this is used? Please remove

kirovilya: currently not used, but can be used along with/like bindCluster modernExtend https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/modernExtend.ts#L2684C17

Koenkk: If it is not used than I propose to remove it for now.

Koenkk:
Thanks!

PR #8966, opened 2025-03-10
Create cwd.ts

vv4victor:
There was no vendor file for CWD.


src/devices/cwd.ts
```diff
@@ -0,0 +1,12 @@
+const m = require('zigbee-herdsman-converters/lib/modernExtend');
```
Koenkk, line 1: Please add to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts

Koenkk:
Thanks!

PR #8969, opened 2025-03-11
Add curtain motor _TZE200_hojryzzd

maxustm:
Add curtain motor _TZE200_hojryzzd produced by Roximo, model CRTZ02


Koenkk:
Thanks!

PR #8972, opened 2025-03-11
Add new thermostat OWON PCT512

Pastor02:
- Added the converter for the OWON PCT512 thermostat.
- It has been necessary to add 2 new constants for the idle and heat values: 32768: ‘idle’, 32769: ‘heat’,


Koenkk:
Thanks!

PR #8978, opened 2025-03-12
Added SQM300ZC4 with 4 large buttons

shinasys:


src/devices/shinasystem.ts
```diff
@@ -593,6 +593,41 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.bind(device.getEndpoint(4), coordinatorEndpoint, ["genOnOff"]);
         },
     },
+    {
+        zigbeeModel: ["SQM300ZC4"],
+        model: "SQM300ZC4",
+        vendor: "ShinaSystem",
+        ota: true,
+        description: "SiHAS remote control 4 full button",
+        fromZigbee: [fz.sihas_action, fz.battery],
```
Koenkk, line 602:
```suggestion
        fromZigbee: [fz.sihas_action],
        extend: [m.battery()],
```

And remove `await reporting.batteryPercentageRemaining(endpoint, {min: 300, max: 43200, change: 1});` and `genPowerCfg` from bind in configure

shinasys: Thanks. Applied it.

Koenkk:
Thanks!

PR #8934, opened 2025-03-04
TS0003 TZ3000_ju82pu2b correct support

Negatifff:
https://github.com/Koenkk/zigbee2mqtt/issues/26513

PR #8980, opened 2025-03-12
TZ3000_ju82pu2b support fix

Negatifff:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/8934#issuecomment-2715643845

practice has shown that the working converter was integrated incorrectly. deleting the code resulted in the device physically working incorrectly, although it was controlled from the HA.

Please consider the possibility of returning the deleted code.


Koenkk:
Thanks!

PR #8984, opened 2025-03-12
Orvibo MixSwitch random on/off bugfix and expose power_on_behavior 

vilaemail:
This PR has 2 improvements:
First is fix for [issue](https://github.com/Koenkk/zigbee2mqtt/issues/26687) is to not configure reporting for onOff. This configuration had 2 effects:
1. Switches start malfunctioning with firmware version 6.1.7 (with 6.1.3 I can confirm they worked properly). In particular turning off and on at random points in time.
2. Switches stop reporting their state periodically. This in particular creates issue when there is power loss event, as Z2M never gets notified that switch changed to off.

Second is exposing power_on_behavior, which is implemented in non-standard way through manufacturer specific cluster 0xFF00.

### DIFF
--- a/src/devices/orvibo.ts
+++ b/src/devices/orvibo.ts
@@ -6,7 +6,7 @@ import * as exposes from "../lib/exposes";
 import {logger} from "../lib/logger";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
-import type {DefinitionWithExtend, Fz, KeyValueAny, ModernExtend, OnEvent, Option, Tz} from "../lib/types";
+import type {DefinitionWithExtend, Expose, Fz, KeyValue, KeyValueAny, ModernExtend, OnEvent, Option, Tz} from "../lib/types";
 import {addActionGroup, hasAlreadyProcessedMessage, postfixWithEndpointName} from "../lib/utils";
 
 const e = exposes.presets;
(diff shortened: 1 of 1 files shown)


src/devices/orvibo.ts
```diff
@@ -61,6 +61,58 @@ const hexToBytes = (hex: string): number[] => {
...
+            },
+        },
+        commandsResponse: {},
+    });
+};
+const clusterManuSpecificOrviboPowerOnBehavior = () => {
+    return m.deviceAddCustomCluster("manuSpecificOrvibo2", {
+        ID: 0xFF00,
+        attributes: {
+            power_on_behavior: { ID: 0x0001, type: Zcl.DataType.UINT8 }
```
vilaemail: @Koenkk is there a way for me to configure Z2M to read the state of the attribute automatically? It is not read by default which means I need to go to frontend and manually click the "refresh" button.

src/devices/orvibo.ts
```diff
@@ -481,56 +576,22 @@ export const definitions: DefinitionWithExtend[] = [
         model: "T40W3Z_v2",
         vendor: "ORVIBO",
         description: "MixSwitch 3 gangs",
-        extend: [m.deviceEndpoints({endpoints: {right: 1, center: 2, left: 3}}), m.onOff({endpointNames: ["right", "center", "left"]})],
+        extend: [
+            clusterManuSpecificOrviboPowerOnBehavior(),
+            orviboSwitchPowerOnBehavior(),
+            m.deviceEndpoints({endpoints: {right: 1, center: 2, left: 3}}),
+            m.onOff({configureReporting: false, powerOnBehavior: false, endpointNames: ["right", "center", "left"]}),
```
vilaemail, line 583: @Koenkk I have suspicion that other switches might benefit from these changes, but I do not have them, so I can't validate. What is the usual approach here? There is possibility we might break their integration, but OTOH if other devices are affected with this issue, then folks experiencing the issue might not realize it is "an easy fix we already have but just haven't applied".

Koenkk: Let's apply it, worst case it doesn't report on/off when physically changing the state which we can easily fix.

vilaemail: Agreed. FWIW I've tested on my units, and they report the state on physical switch press, even when not configured.

Nerivec:
I think we should make a note somewhere in the orvibo file that their hub appears to be using custom frames (based on the sniffs in ref issue, even Wireshark is lost...).
It's something that's good to know if weird issues arise in the future...

vilaemail:
@Nerivec I unfortunately was not able to find those custom frames in the sniff, and I doesn't feel right to note it without having an example. If you can point out a couple of packets which are not error due to my sniffer being far away from the transmitting device (and after packet 360), I can add the note including those frames data.

Nerivec:
Packet 360 is wrong, 361 too... and plenty more after using non-MAC2003, reserved addressing mode...
Thing is though, custom frames or broken frames would look the same (Wireshark stops parsing at first non-spec-valid byte).
If that's all sniffer errors, what sniffer are you using exactly? I've never seen this behavior before, at least not in that amount.
Any chance you have another Zigbee network nearby?

Note: quick & dirty Wireshark filter `_ws.expert.severity == "Warning" or _ws.expert.severity == "Error"` shows almost 3% with the hub, and falls to near 1% with z2m.

vilaemail:
I am using ESP board with CC2531 chip. I flashed this firmware https://www.zigbee2mqtt.io/advanced/zigbee/04_sniff_zigbee_traffic.html#_1-flashing-the-cc2531-adapter

There are no ZigBee networks that I am aware of, but Wi-Fi is, unfortunately, heavily congested.

If you say in Z2M trace there is no `packets using non-MAC2003, reserved addressing mode` then I am convinced there are custom frames. I am skeptical, because in my mind manufacturer would need to go out of his way to make custom frames, which I see no incentive for. Am I missing something obvious here?

Nerivec:
They could be using it for telemetry, heartbeat, that kind of stuff (others do).
If devices communicate something like a heartbeat, it wouldn't disappear by powering off the hub, which would explain why you see less, but still some, in Z2M sniff.
Although, you may be right, it could just be the sniffer. Weird thing is, I've gone through a lot of sniffs recently from Silabs & Nordic, and never seen this before. Maybe the TI one is different... Koenkk will know! 😬

_PS: If you get a chance to pick up an EFR32, you can use it as coordinator, [sniffer](https://github.com/Nerivec/ember-zli/wiki/Sniff), and [network scanner](https://github.com/Nerivec/ember-zli/wiki/Stack#scan-network) without needing to flash anything but a coordinator firmware. Pretty useful for this sort of debugging_ 😉

Koenkk:
I'm not sure what causes these malformed frames, can also be interference (if your ESP board is 2.4ghz it might interfere with Zigbee). Also distance from CC2531 to hub matters. Anyway if this is working fine, it appears these magic frames are not needed for normal functioning of the devices should this should be good to go.

PR #8999, opened 2025-03-15
Expose power, current and voltage as normal sensor data.

kad:
For the devices with haElectricalMeasurement category, current, voltage and power are usable sensor data, not really just a diagnostic values.

This supposed to bring devices like IKEA Inspelning in zigbee2mqtt closer to how it is exposed via IKEA Dirigera hub.

### DIFF
--- a/src/lib/exposes.ts
+++ b/src/lib/exposes.ts
@@ -1015,8 +1015,7 @@ export const presets = {
     cpu_temperature: () =>
         new Numeric("cpu_temperature", access.STATE).withLabel("CPU temperature").withUnit("°C").withDescription("Temperature of the CPU"),
     cube_side: (name: string) => new Numeric(name, access.STATE).withDescription("Side of the cube").withValueMin(0).withValueMax(6).withValueStep(1),
-    current: () =>
-        new Numeric("current", access.STATE).withUnit("A").withDescription("Instantaneous measured electrical current"),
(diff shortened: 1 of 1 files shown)


Koenkk:
These have been put in `diagnostic` on purpose (by one of the HA maintainers), see https://github.com/Koenkk/zigbee2mqtt/pull/25322#issuecomment-2564318073 for more info.

kad:
According to https://github.com/Koenkk/zigbee2mqtt/pull/9293, the diagnostics are supposed to be used about device itself, e.g. battery capacity:

> An entity that provides information about the device itself (e.g., battery capacity, signal, tampering), is considered diagnostic information; thus get diagnostic is set at the entity category.

The values mentioned in this PR are about environment sensors observing external entity, not the zigbee device itself.
Another reason is consistency with other integrations: Ikea integration via Dirigera hub exposes it as sensors, ZHA (my colleague tried) for similar plugs also exposes those as normal sensors, not diagnostics.

kad:
e.g. in ZHA, those are not marked as `EntityCategory.DIAGNOSTIC` : https://github.com/zigpy/zha/blob/dev/zha/application/platforms/sensor/__init__.py#L708-L790

Koenkk:
Makes sense, thanks! Could you also submit a PR to remove e.g. these? https://github.com/Koenkk/zigbee2mqtt/blob/ba337bd329aeb4ca17735c0cf09b31293c8cff06/lib/extension/homeassistant.ts#L146

kad:
Sure, will check later today after work.

PR #9001, opened 2025-03-15
feat(add): Leleka, add MindY devices file

xyzroe:


src/devices/mindy.ts
```diff
@@ -0,0 +1,366 @@
...
+                name: "night_off_time",
+                unit: "Hr",
+                valueMin: 0,
+                valueMax: 23,
+                cluster: "genBasic",
+                attribute: {ID: 0x0242, type: Zcl.DataType.UINT8},
+                description: "Night mode deactivation time",
+            }),
+            m.binary({
+                name: "CO2_control",
```
Koenkk, line 195:
```suggestion
                name: "co2_controll",
```

Use `snake_case` here (also for others)

src/devices/mindy.ts
```diff
@@ -0,0 +1,366 @@
...
+            await endpoint.read("msCO2", [0x0220]);
+            await endpoint.read("msCO2", [0x0221]);
+            await endpoint.read("msCO2", [0x0222]);
+            await endpoint.read("msCO2", [0x0223]);
+
+            await endpoint.read("msCO2", [0x0230]);
+            await endpoint.read("msCO2", [0x0231]);
+            await endpoint.read("msCO2", [0x0232]);
+
+            await endpoint.read("msIlluminanceMeasurement", [0x0220]);
```
Koenkk, line 136: Are these needed? The extend below should already do this.

src/devices/mindy.ts
```diff
@@ -0,0 +1,366 @@
...
+    } satisfies Tz.Converter,
+    wifi: {
+        key: ["wifi"],
+        convertGet: async (entity, key, meta) => {
+            await entity.read("genBasic", ["locationDesc"]);
+        },
+        convertSet: async (entity, key, value, meta) => {
+            await entity.write("genBasic", {locationDesc: value === "ON" ? "1" : "0"});
+
+            const retryRead = async (attempts: number) => {
```
Koenkk, line 28: Why is there a retry mechanism here?

src/devices/mindy.ts
xyzroe: without forced reading, these values remain undefined after a successful interview and device configuration. until the user requests them manually.

src/devices/mindy.ts
xyzroe: The device reboots after the wifi is turned on or off. Therefore, in order to get information about the changed IP address as quickly as possible (and the IP address or an empty string is transmitted through the location description).
in some networks one request is not enough, because in some cases we get an error in response.

src/devices/mindy.ts
xyzroe: I think I've got it all fixed.

src/devices/mindy.ts
xyzroe: Maybe I should put all custom attribute numbers in a separate object at the beginning of the file?

src/devices/mindy.ts
```diff
@@ -0,0 +1,366 @@
...
+        zigbeeModel: ["Leleka"],
+        model: "Leleka",
+        vendor: "MindY",
+        description: "Advanced Environmental Monitoring Device",
+        ota: true,
+        fromZigbee: [fzLocal.last_boot, fzLocal.wifi],
+        toZigbee: [tzLocal.last_boot, tzLocal.wifi],
+        options: [optionsLocal.last_boot_update()],
+        onEvent: (type, data, device, options) => {
+            const endpoint = device.getEndpoint(1);
```
Koenkk, line 95:
```suggestion
            const poll = async () => {
                        await endpoint.read("genTime", ["lastSetTime"]);
                        await endpoint.read("genBasic", ["locationDesc"]);
            };

            utils.onEventPoll(type, data, device, options, "last_boot_update", 60, poll);
```

You can use this instead

src/devices/mindy.ts
Koenkk: Yes, and use a for loop to prevent a lot of code here.

src/devices/mindy.ts
xyzroe: I thought I understood your idea correctly.

src/devices/mindy.ts
xyzroe: Yeah, it's really better that way.

Koenkk:
Thanks!

Make sure that:
- The filename is `Leleka.png`
- The size is 512x512

PR #9017, opened 2025-03-18
Yandex dimmer YNDX-00530

kirovilya:


src/devices/yandex.ts
```diff
@@ -431,4 +436,48 @@ export const definitions: DefinitionWithExtend[] = [
...
+                setCommand: "buttonMode",
+                zigbeeCommandOptions: {manufacturerCode},
+                description: "Dimmer button mode",
+                lookup: {
+                    general: 0x00,
+                    alternative: 0x01,
+                },
+                entityCategory: "config",
+            }),
+            m.reportAttribute({
```
Koenkk, line 476: Is this not already done with:

```js
m.light({
                effect: true,
                powerOnBehavior: true,
                configureReporting: true,
            }),
```

kirovilya: No. The default minimum period in the report is 10 seconds, and here I change it to the minimum.

kirovilya: If I add reporting parameters for each attribute to m.light, the code will grow. Therefore, I decided to make a universal reporting that can be used in various cases.

Koenkk: I think it is nicer to put it in `m.light`, in this was you don't need to look in multiple places for the config (and also the execution order is not guaranteed)

kirovilya: ok

src/lib/modernExtend.ts
```diff
@@ -989,6 +989,7 @@ export interface LightArgs {
     endpointNames?: string[];
     ota?: ModernExtend["ota"];
     levelConfig?: {disabledFeatures?: LevelConfigFeatures};
+    levelConfigureReporting?: ReportingConfigWithoutAttribute;
```
Koenkk, line 992:
```suggestion
    levelReportingConfig?: ReportingConfigWithoutAttribute;
```

Koenkk:
Thanks!

Issue #8994, opened 2025-03-14
Doorlock is missing converters.


los7702:
I use a dorrlock "easyCodeTouch_v1" in [onesti.ts](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/onesti.tsl)
I works but does not report confirmation on setting or clearing pin_code 
`
[2025-03-14 12:16:41] debug: 	zh:controller: Received payload: clusterID=257, address=20698, groupID=0, endpoint=11, destinationEndpoint=1, wasBroadcast=false, linkQuality=102, frame={"header":{"frameControl":{"frameType":1,"manufacturerSpecific":false,"direction":1,"disableDefaultResponse":true,"reservedBits":0},"transactionSequenceNumber":2,"commandIdentifier":5},"payload":{"status":0},"command":{"ID":5,"parameters":[{"name":"status","type":32}],"name":"setPinCodeRsp"}}
[2025-03-14 12:16:41] debug: 	zh:zstack:unpi:parser: --- parseNext []
[2025-03-14 12:16:41] debug: 	z2m: Received Zigbee message from 'Las', type 'commandSetPinCodeRsp', cluster 'closuresDoorLock', data '{"status":0}' from endpoint 11 with groupID 0
[2025-03-14 12:16:41] debug: 	z2m: No converter available for 'easyCodeTouch_v1' with cluster 'closuresDoorLock' and type 'commandSetPinCodeRsp' and data '{"status":0}'
`
and 
`
[2025-03-14 12:41:26] debug: 	zh:controller: Received payload: clusterID=257, address=20698, groupID=0, endpoint=11, destinationEndpoint=1, wasBroadcast=false, linkQuality=60, frame={"header":{"frameControl":{"frameType":1,"manufacturerSpecific":false,"direction":1,"disableDefaultResponse":true,"reservedBits":0},"transactionSequenceNumber":5,"commandIdentifier":7},"payload":{"status":0},"command":{"ID":7,"parameters":[{"name":"status","type":32}],"name":"clearPinCodeRsp"}}
[2025-03-14 12:41:26] debug: 	zh:zstack:unpi:parser: --- parseNext []
[2025-03-14 12:41:26] debug: 	z2m: Received Zigbee message from 'Las', type 'commandClearPinCodeRsp', cluster 'closuresDoorLock', data '{"status":0}' from endpoint 11 with groupID 0
[...]

PR #9018, opened 2025-03-18
Update onesti.ts

los7702:
Trying to add response from lock when pin_code is updated as of issue #8994


src/devices/onesti.ts
```diff
@@ -9,6 +9,24 @@ const e = exposes.presets;
 const ea = exposes.access;
 
 const fzLocal = {
+    nimly_lock_commandPinCodeRsp: {
```
Koenkk, line 12: This converter is generic and can be moved to `fromZigbee.ts`

src/devices/onesti.ts
```diff
@@ -9,6 +9,24 @@ const e = exposes.presets;
...
+        cluster: 'closuresDoorLock',
+        type: ['commandSetPinCodeRsp','commandClearPinCodeRsp'],
+        convert: (model, msg, publish, options, meta) => {
+            const attributes: KeyValue = {};
+            if (msg.data.status === 0) {
+                if (msg.type === 'commandSetPinCodeRsp') {
+                    attributes.Last_successful_Pincode_Save = Date.now();
```
Koenkk, line 19:
```suggestion
                    attributes.last_successful_pincode_save = Date.now();
```

Also for below and make sure to update the expose

src/devices/onesti.ts
los7702: Actually i think Nimly report 0 when successful in oppose of standard 1, i might have read wrong in the standard but...
i will chnge to lower leading letter

los7702: i was wrong it is general
image

Koenkk:
Thanks!

PR #9020, opened 2025-03-18
feat: Generate definition for unsupported Green Power devices

Nerivec:
@Koenkk not sure about the best way to display the payload? Frontend shows the array as a plain string `[1]` as it is right now.

This should also allow partially supporting the Moes ZEU2S that doesn't appear to have a fixed source ID... (otherwise used to identify an actual converter for GP devices). ref: https://github.com/Koenkk/zigbee2mqtt/issues/19405

_Tested with a PTM216Z from enocean by removing its actual converter._

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -529,7 +529,7 @@ export async function findDefinition(device: Zh.Device, generateForUnknown = fal
         }
 
         // Do not add this definition to cache, as device configuration might change.
-        const {definition} = device.type === "GreenPower" ? generateGreenPowerDefinition(device) : await generateDefinition(device);
+        const {definition} = await generateDefinition(device);
 
         return definition;
     }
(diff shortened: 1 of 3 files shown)


src/lib/generateDefinition.ts
```diff
@@ -69,6 +137,18 @@ const definition = {
 module.exports = definition;`;
 }
 
+function generateGreenPowerSource(definition: DefinitionWithExtend, ieeeAddr: string): string {
+    return `import {genericGreenPower} from 'zigbee-herdsman-converters/lib/modernExtend';
```
Nerivec, line 141: NOTE: made that one with dynamic import PR support... so it's already ESM!
re: https://github.com/Koenkk/zigbee-herdsman-converters/pull/9003

src/index.ts
```diff
@@ -529,7 +529,7 @@ export async function findDefinition(device: Zh.Device, generateForUnknown = fal
         }
 
         // Do not add this definition to cache, as device configuration might change.
-        const {definition} = await generateDefinition(device);
+        const {definition} = device.type === "GreenPower" ? generateGreenPowerDefinition(device) : await generateDefinition(device);
```
Koenkk, line 532: I think it is better to push this if statement into `generateDefinition` (no need to do it twice in this file)

Koenkk:
> @Koenkk not sure about the best way to display the payload? Frontend shows the array as a plain string [1] as it is right now.

I think that's fine

Koenkk:
Thanks!

PR #9035, opened 2025-03-19
Fix color temperature range for Namron 3802966

Eising:
The original definition for this LED bulb is not correct regarding the colorTemperature range.

The actual range is 200-454 which is also reported by the generated definition.

This PR fixes just this.
Let me know if there is anything else needed in order to get this PR merged.

Thanks!

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -533,7 +533,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "3802966",
         vendor: "Namron",
         description: "LED 4.8W CCT GU10",
-        extend: [m.light({colorTemp: {range: [153, 370]}})],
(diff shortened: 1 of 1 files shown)


Eising:
I can't find a suitable Issue category, so I hope this PR will suffice.

Here's the external_definition generated (with just the description updated):

```js
const m = require('zigbee-herdsman-converters/lib/modernExtend');

const definition = {
    zigbeeModel: ['3802966'],
    model: '3802966',
    vendor: 'Namron AS',
    description: 'Namron Zigbee LED bulb 4,8W CCT GU10',
    extend: [m.light({"colorTemp":{"range":[200,454]}})],
    meta: {},
};

module.exports = definition;```

Koenkk:
Thanks!

PR #9040, opened 2025-03-20
[MAZDA TR-M2Z] Use thermostatSystemModeAndPresetMap function to correctly map presets and use none instead of manual

Nopraz:
I now use none instead of manual because of a weird behavior when using better thermostat. 
Better thermostat sends none as the preset when using it and when this happens and none does not exist zigbee2mqtt defaults to the schedule preset which is not really what you want if you already use the scheduler integration on Home Assistant.


Koenkk:
Thanks!

PR #9046, opened 2025-03-22
fix: Generate definition when candidates present but no match found

Nerivec:
Should take care of the problem identified in https://github.com/Koenkk/zigbee2mqtt/issues/19405#issuecomment-2744667458

When a device had potential candidates, it would not generate a definition, even if no match found.


Koenkk:
Thanks!

PR #9047, opened 2025-03-22
fix: add self report to GL-C-009P

chiva:
Add self reporting to GL-C-009P based on https://github.com/Koenkk/zigbee2mqtt/issues/23661


Koenkk:
Thanks!

bernardesarthur:
Can you add this option for other models?

PR #9052, opened 2025-03-23
Update tuya.ts

FPSUsername:
Added lightType to TS110E_2gang_2 as it's also available on the TS110E_1gang_2.

Tested and the switch works:
image
image

image
image


Koenkk:
Thanks!

PR #9060, opened 2025-03-25
Update candeo.ts

candeodevelopment:
Adds support for Candeo water sensor, contact sensor, motion sensor and temperature / humidity sensor.

Temperature / humidity sensor uses modern extend.

Contact sensor and motion sensor use partial modern extend since they do not fully support IAS Cluster enrolment, so we need to remove ias_enroll and only allow / configure ias_contact* & ias_occupancy*

Water sensor also uses partial modern extend because while the device does fully support IAS Cluster enrolment, it requires a bind on the IAS Cluster before it will start sending events.  Modern extend does not do this by default unless zoneStatusReporting is set to true, but the device does not support **reporting** on that attribute so this generates errors.

Motion sensor also only partially supports Illuminance Cluster (it reports only on motion detection) so we remove / disallow reporting configuration for it.


src/devices/candeo.ts
```diff
@@ -36,6 +36,14 @@ const fzLocal = {
             return undefined;
         },
     } satisfies Fz.Converter,
+    illuminance_report: {
```
Koenkk, line 39: Can't `fz.illuminance` be used here?

Koenkk: Ah sorry, you are right, can you check with the extend instead?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/9f1e179142501fb7647b2664f984edc38c67656b/src/devices/aurora_lighting.ts#L181

candeodevelopment: 1) When the device detects motion it also reads the current illuminance value and sends an illuminance report, just before sending the IAS ZoneState (alarm true).
2) When the device detects clear it also reads the current illuminance value and sends an illuminance report, just before sending the IAS ZoneState (alarm false).
3) The device sends the illuminance reports above as long as it has the coordinator bound to the illuminance Cluster, irrespective of whether it has any reporting configuration set on the illuminance Cluster.
4) By default the device has reporting configuration set to report illuminance periodically, but this is pointless since it will keep sending whatever value was previously measured on the last motion.

So we need:

1) fz.illuminance to handle the reported illuminance when it is sent during a motion event.
2) To un-configure the reporting (remove the reporting configuration completely) for illuminance to prevent it being sent periodically by default.
3) By doing the above we also remove the default reporting line from the Z2M UI in order to make it less likely that customers will mistakenly think that they can configure the illuminance reporting.

We could not see a way to achieve the above using modern extend?

Koenkk: Did you try with `m.illuminance({reporting: false})`?

Koenkk: No

candeodevelopment: Can we use the current code then until it becomes possible to do this with modern extend?

Koenkk: > Does that remove existing configuration from the device?

Does this have to be done? I don't see the existing code doing this?

candeodevelopment: Ideally, since by default the device has reporting configuration set to report illuminance periodically, but this is pointless since it will keep sending whatever value was previously measured on the last motion.

It's currently handled in the configure for the device:

image

Koenkk: Then `m.illuminance({reporting: {min: 1, max: 65535, change: 1})` should do it.

src/devices/candeo.ts
```diff
@@ -319,4 +319,53 @@ export const definitions: DefinitionWithExtend[] = [
...
+            await reporting.bind(endpoint, coordinatorEndpoint, ["ssIasZone"]);
+            await endpoint.read("ssIasZone", ["zoneStatus"]);
+        },
+    },
+    {
+        fingerprint: [{modelID: "C-ZB-SEMO", manufacturerName: "Candeo"}],
+        model: "C-ZB-SEMO",
+        vendor: "Candeo",
+        description: "Motion sensor",
+        extend: [m.battery(), m.illuminance({reporting: {min: 1, max: 65535, change: 1}})],
```
Koenkk, line 362:
```suggestion
        extend: [m.battery(), m.illuminance({reporting: {min: 1, max: 65535, change: 1}}), m.iasZoneAlarm({zoneType: "occupancy", zoneAttributes: ["alarm_1"]})],
```

This allows you to get red of all fromZigbee/exposes and configure. Please also do it for the other devices

PR #9067, opened 2025-03-26
Update tuya.ts support for Tuya TS0043

david-alexandru-v:
Add conversion for the signals received at genOnOff cluster into the known signals Use fingerprint for precise identification

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -3517,19 +3517,33 @@ export const definitions: DefinitionWithExtend[] = [
          */
     },
     {
-        zigbeeModel: ["TS0043"],
-        model: "TS0043",
-        vendor: "Tuya",
-        description: "Wireless switch with 3 buttons",
-        whiteLabel: [
-            {vendor: "Smart9", model: "S9TSZGB"},
-            {vendor: "Lonsonho", model: "TS0043"},
-            {vendor: "LoraTap", model: "SS600ZB"},
-        ],
-        exposes: [e.battery(), e.action(["1_single", "1_double", "1_hold", "2_single", "2_double", "2_hold", "3_single", "3_double", "3_hold"])],
-        fromZigbee: [tuya.fz.on_off_action, fz.battery],
-        toZigbee: [],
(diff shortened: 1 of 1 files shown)


Koenkk:
Could you explain the problem of the current implementation?

david-alexandru-v:
I'm afraid I can't :) I'm not a TypeScript coder and this was a project that I tested my new Copilot license on. 

What I can definitely say, is that with the devices I have (and were recognized by the impl as TS0043 - link in the original [issue](https://github.com/Koenkk/zigbee2mqtt/issues/23987)) and the current definition that is in the repo, after interviewing the device it doesn't correctly recognize the various signals that I can issue from the 3 buttons. The picture of my device is also different, again, pls check the issue and the picture of the current device that is registered in the repo and docu as TS0043. Apologies if I'm just making it more difficult. 

My best guess, either the specific fingerprint identification, or the conversion of the signals has fixed the issue for the external controller that Copilot wrote.

david-alexandru-v:
I'd like to be of help, not sure of the format I should provide it in... Let me give it a shot

Before the fix, if in the Zigbee2MQTT Frontend I would click on the device and came to the screen below

image

Clicking a button would result in the Action value shortly displaying something like "(no action)". After the fix, it actually displays the "wanted" action names like "1_single", "1_double" etc

Koenkk:
This is because you are using the home assistant legacy triggers option, try removing that from your `configuration.yaml`

david-alexandru-v:
Hmm, if I recall correctly I had to add that when I upgraded to major version 2 of zigbee2mqtt because my zigbee devices suddenly stopped working. After I added it to config, it started working again.

david-alexandru-v:
Wow, I retraced my steps, now that you said it. I updated my coordinator firmware to the latest version, removed legacy attribs from the configuration, removed the external controller file, restarted everything and everything still works ok. Perhaps the coordinator firware was older or something... I think this means this PR is moot. Shall I just close?

Koenkk:
Great, I'll close it.

PR #9077, opened 2025-03-27
Update tuya.ts

hyteplus:
Support TS0726 1 gang and 4 gang switch.
Don not merge with _TZ3000_m3pafcnk(3 gang) and _TZ3000_qhyadm57(4 gang+2 scene)! Otherwise, the wrong model will be identified, resulting in redundant endpoints and an error message.


Koenkk:
Thanks!

PR #9079, opened 2025-03-27
feat(add): 1402790 (manuf: Namron)

WJKPK:
Add support for 1402790 form manufacturer Namron
Device similar to Fire Fence Stove Guard.

Picture should be identical to https://github.com/Koenkk/zigbee2mqtt.io/pull/3553.
Question for maintainer - does I require to commit same image again under different name or there exist any mechanism to logically symlink these?


Koenkk:
> Question for maintainer - does I require to commit same image again under different name or there exist any mechanism to logically symlink these?

Please submit a new image (even if it is the same)

PR #9080, opened 2025-03-27
Update sunricher.ts

chris-1243:
Add new SR-ZG9101CS

CC: @greekgremlin


chris-1243:
@Koenkk Open question as I am stuck on that. How may I override the default reporting value set when adding `configureReporting: true` ?

Koenkk:
@chris-1243 why do you want to override the defaults? (there should be a good reason to differ for this device, as we want to keep a consistent reporting across devices)

chris-1243:
In my point of view, the `min rep interval` for the brightness is a bit too long (10 seconds). I do not have the device. I am helping a user who has this wish. I am fine with default value and we may let each user modifying according to their needs as well.

https://github.com/Koenkk/zigbee2mqtt/issues/26818#issuecomment-2758010473

The main reason and to be honest: learning improvement...

Koenkk:
I'm fine with changing the default to e.g. 5, but 1 would probably get too spammy

chris-1243:
It is a good compromise. Let's go for it.

Thanks

Koenkk:
Would you mind updating the PR with this?

Koenkk:
I've fixed the test, thanks!

Make sure that:
- The filename is `SR-ZG9101CS.png`
- The size is 512x512

chris-1243:
Thanks for the test. I was not aware I should also modify this one.

The picture has been already added according to this https://github.com/Koenkk/zigbee2mqtt.io/pull/3638 and merged in zigbee2mqtt.io

PR #9086, opened 2025-03-28
fix: enable onOffCountdown for TS011F_with_threshold

petrkalos:
I have 2 `_TZ3000_cayepv1a` and countdown feature is working properly so enabling it for everyone. 
* If the state is off and set the countdown it will power on with a specific delay
* If the state is on and set the countdown it will power off with a specific delay

image

Unfortunately I don't own any of the other devices and I am not sure if they support this feature.

This is my first PR here so please let know if there are any other testing steps I should be making.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -10404,6 +10404,7 @@ export const definitions: DefinitionWithExtend[] = [
                 electricalMeasurementsFzConverter: fzLocal.TS011F_electrical_measurement,
                 powerOutageMemory: true,
                 indicatorMode: true,
+                onOffCountdown: true,
             }),
         ],
         fromZigbee: [fz.temperature, fzLocal.TS011F_threshold],


petrkalos:
@puttemanp you should be able to add the same statement `onOffCountdown: true,` [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts#L11575) and submit a pull request. 

I'd strongly suggest testing it first though. 
The way I did was the following
* Get a shell in the z2q container `docker exec -it z2q /bin/sh`
* Edit the `tuya.js` file (be careful this is a compiled js file, not the original ts so it should be around line 11316) and add the onOffCountdown parameter
* Exit the container
* Commit the change `docker commit z2q koenkk/zigbee2mqtt:2.1.3_custom` 
* Restart z2q container but now create it from your newly created tag `2.1.3_custom`
* You should now be able to see the entities and test them.

Perhaps @Koenkk knows a better way to test this.

PR #9087, opened 2025-03-28
introduce .git-blame-ignore-revs

petrkalos:
Migration commit is making very hard to track changes so it's useful if it gets removed from blame.
For more details check [here](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view)


Koenkk:
Didn't now this was possible, thanks!

PR #9095, opened 2025-03-30
fix: Schneider S520567: expose tilt

jerome-pouiller:
My Schneider S520567 is connected to outdoor Venetian blind. The device properly report the tilt value of my blind. However, there was no way to control it.

This patch has a drawback. I also use the same device reference for normal shutters. This patch expose a tilt control while the shutter does not support it. I assume I have to expose a property to enable/disable the control of the tilt. I would be glad to have some guidance about how to do that.


Koenkk:
Thanks!

PR #9098, opened 2025-03-31
Re-add brightness actions to Philips Hue Wall Switch 929003017102

marazmarci:
I originally added the `brightness_move_up`, `brightness_move_down` and `brightness_stop` actions in #7489, but I noticed that when using the still experimental `event` entities, these actions are not exposed.

My Wall Switch is set up to control a group ([instructions](https://www.zigbee2mqtt.io/devices/929003017102.html#directly-control-zigbee-groups)), and in this configuration, I can receive each `brightness_*` event in HA after applying my fix.


Koenkk:
Thanks!

PR #9099, opened 2025-03-31
updated e.action for Paulmann 501.34 fixes #9075

### LINKED ISSUE #9075: Paulmann 501.34 e.action needs to be updated
I recently added a Paulman 501.34 wall switch and discovered that the release of the buttons when in dimming mode is not properly reported.
It boiled down to the issue that:
'brightness_move_stop_1' and 'brightness_move_stop_2'
need to be changed to 
'brightness_stop_1' and 'brightness_stop_2'
in
devices/paulmann.js

I'm not sure if this is a general issue, or only applies to this switch, as I also noticed that in paulmann.js there is also the 501.37 switch which is using also the move_stop action.

mkpfalz:


Koenkk:
Thanks!

marazmarci:
FYI, another Paulmann device (501.37) also has `brightness_move_stop` specified.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/8295953f6b2b32f10704dc281cb7ca3fb889e78e/src/devices/paulmann.ts#L26-L29

Issue #8680, opened 2025-01-26
Another Gledopto GL-C-007 Fingerprint - Not detected correctly


Xyaren:
Hi there, I bought a RGBW Controller a while ago and trouble to get it going without a custom converter.

Modle IDs reported:
Box it came in says: GL-C-007 (Box 2ID checked)
Physical Device Print: GL-C-007

Endpoint 11/genBasic: GL-C-007
Endpoint 13/genBasic: GL-C-009

Database Entry in Z2M:

```json
{
  "id": 39,
  "type": "Router",
  "ieeeAddr": "0x00124b0021efffe2",
  "nwkAddr": 48256,
  "manufId": 0,
  "manufName": "GLEDOPTO",
  "powerSource": "Mains (single phase)",
  "modelId": "GL-C-009",
  "epList": [
    11,
    13,
    15
  ],
  "endpoints": {
    "11": {
      "profId": 49246,
      "epId": 11,
      "devId": 528,
      "inClusterList": [
        0,
        3,
        4,
        5,
        6,
        8,
        768
      ],
      "outClusterList": [],
      "clusters": {
        "genOnOff": {
          "attributes": {
            "onOff": 0
          }
        },
        "lightingColorCtrl": {
          "attributes": {
            "colorCapabilities": 31,
            "colorMode": 2,
            "currentX": 0,
            "currentY": 0,
            "enhancedCurrentHue": 6400,
            "currentSaturation": 251,
            "colorTemperature": 370
          }
        },
        "genBasic": {
          "attributes": {
            "modelId": "GL-C-007",
            "manufacturerName": "GLEDOPTO",
            "powerSource": 1,
            "zclVersion": 1,
            "appVersion": 1,
            "stackVersion": 2,
            "hwVersion": 1,
            "dateCode": "20180920",
            "swBuildId": "2.0.5"
          }
        },
        "genLevelCtrl": {
          "attributes": {
            "currentLevel": 254
          }
        }
      },
      "binds": [],
      "configuredReportings": [],
      "meta": {
        "onLevelSupported": false
      }
    },
    "13": {
      "profId": 49246,
      "epId": 13,
      "devId": 57694,
      "inClusterList": [
        4096
      ],
      "outClusterList": [
        4096
      ],
      "clusters": {},
      "binds": [],
      "configuredReportings": [],
      "meta": {}
    },
    "15": {
      "profId": 49246,
      "epId": 15,
      "devId": 256,
[...]

PR #9103, opened 2025-04-01
Fix GL-C-007-2ID not being detected correctly

Xyaren:
As requested in #8680


Koenkk:
Looks good, I don't expect any conflicts.

Issue #8990, opened 2025-03-13
[New Device Support] Tuya window opener


devkaiwang:
I bought two window openers from different vendors for my skylight windows and found that they are using exact same model. 

However, Z2M does not support it by default.

<img src="https://github.com/user-attachments/assets/766b17ca-307c-4b43-8a23-93429ab89007" width="300">

<img src="https://github.com/user-attachments/assets/47fe8d76-5afa-4a89-9221-6d6ea78b3de7" width="300">

**Here is the external converter code that works:**

```js
const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
const tz = require('zigbee-herdsman-converters/converters/toZigbee');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const reporting = require('zigbee-herdsman-converters/lib/reporting');
const e = exposes.presets;
const ea = exposes.access;
const tuya = require('zigbee-herdsman-converters/lib/tuya');

const definition = {
    fingerprint: tuya.fingerprint("TS0601", ["_TZE200_a8z0g46u", "_TZE204_a8z0g46u"]),
    model: "TS0601",
    vendor: "Tuya",
    description: "Automatic window opener",
    options: [exposes.options.invert_cover()],
    fromZigbee: [tuya.fz.datapoints],
    toZigbee: [tuya.tz.datapoints],
    onEvent: tuya.onEventSetTime,
    configure: tuya.configureMagicPacket,
    exposes: [
        e.cover_position().setAccess('position', ea.STATE_SET),
    ],
    meta: {
        tuyaDatapoints: [
            [2, 'state', tuya.valueConverterBasic.lookup({'OPEN': tuya.enum(0), 'STOP': tuya.enum(1), 'CLOSE': tuya.enum(2)})],
            [7, 'position', tuya.valueConverter.coverPosition],
            [8, 'position', tuya.valueConverter.coverPosition],
        ],
    },
    extend: [
    ],
};

module.exports = definition;
```

Koenkk:
Could you make a PR by clicking [here](https://github.com/Koenkk/zigbee-herdsman-converters/edit/master/src/devices/tuya.ts)?

PR #9117, opened 2025-04-04
Add support for two Tuya window openers.

devkaiwang:
Added support for two Tuya window openers - https://github.com/Koenkk/zigbee-herdsman-converters/issues/8990


src/devices/tuya.ts
```diff
@@ -4611,6 +4611,25 @@ export const definitions: DefinitionWithExtend[] = [
             tuya.whitelabel("Nova Digital", "CMR-1", "Roller Blind Motor", ["_TZE200_9vpe3fl1"]),
         ],
     },
+    {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE200_a8z0g46u", "_TZE204_a8z0g46u"]),
+        model: "TS0601",
```
Koenkk, line 4616: What is the model number written on device or product page?

devkaiwang: There are many similar products and I think they are using the Tuya Zigbee module to make such window openers. There companies are also using Xiaomi Bluetooth modules for integrating to Xiaomi Home. 

I contacted the vendor of the product, he told me they call it "YY-LT500". Is it OK to use "YY-LT500" or should we name it in a more generic way?

devkaiwang: The product has printed model on it. It is **"YY-LT500"**.  I think we should use this name. 

I also disassembled the product and it is using [ZTU-IPEX Zigbee module](https://developer.tuya.com/en/docs/Document/ztuipex-module-datasheet?id=Kamwy20nl3rr3).

devkaiwang: Changed model name, please review, thanks!

Koenkk:
Thanks!

Make sure that:
- The filename is `YY-LT500.png`
- The size is 512x512

PR #9120, opened 2025-04-04
Add battery voltage to pvvx/ZigbeeTLc devices

jpbede:
The pvvx/ZigbeeTLc firmware does also support the battery voltage.

image


Koenkk:
Thanks!

PR #9144, opened 2025-04-08
Update PMM-300Z2 and PMM-300Z3

shinasys:
- PMM-300Z2(Single-phase power meters) : Produced_energy and ct_direction is supported in version 8 and above.
- PMM-300Z3(Three-phase power meters): Produced_energy is supported in version 9 and above

### DIFF
--- a/src/devices/shinasystem.ts
+++ b/src/devices/shinasystem.ts
@@ -197,6 +197,7 @@ const tzLocal = {
         convertSet: async (entity, key, value, meta) => {
             const lookup = {Auto: 0, "Manual(Forward)": 1, "Manual(Reverse)": 2};
             await entity.write("seMetering", {"0x9001": {value: utils.getFromLookup(value, lookup), type: 0x20}});
+            return {state: {[key]: value}};
         },
         convertGet: async (entity, key, meta) => {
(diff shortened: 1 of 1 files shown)


src/devices/shinasystem.ts
```diff
@@ -640,15 +661,89 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "ShinaSystem",
         ota: true,
         description: "SiHAS energy monitor",
-        extend: [m.electricityMeter({power: {cluster: "metering"}, acFrequency: {multiplier: 1, divisor: 10}, powerFactor: true}), m.temperature()],
+        fromZigbee: [fzLocal.ct_direction],
+        toZigbee: [tzLocal.ct_direction],
+        extend: [
+            m.electricityMeter({
+                power: {cluster: "metering", min: 30},
+                current: {min: 30, max: 600, change: 1},
```
Koenkk, line 669: Let's keep the default min/max/change reporting config here, this gives a consistent experience across devices. In case people want to change it they can change it via the reporting tab.

shinasys: I understand and agree.

Koenkk:
Thanks!

PR #9149, opened 2025-04-08
fix: latest philips hue ensis black models

mizx:
Added my new light to zigbee2mqtt and it came up as unsupported. Following the changes and comments in #3403 which are directly related. Philips Hue site confirms the number on its product page: https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-ensis-pendant-light/046677588366 (`Material number (12NC)`)

LMK if there's anything else I need to do or can help out with getting this merged!

PR #9154, opened 2025-04-10
fix: latest philips hue perifo spotlight and linear light bar

mizx:
follow-up to #9149

also add support for:
- Perifo linear light bar (black) ([ref](https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-perifo-linear-light-bar/046677583392) - `929003618201`)
- Perifo cylinder spotlight (black) ([ref](https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-perifo-cylinder-spotlight/046677583309) - `929003617901`)

Note: model #s are available on both web pages provided


Koenkk:
Thanks!

PR #9160, opened 2025-04-11
Add vendor files via upload

joliam:
add vendor file for lincukoo.


joliam:
I want to know the reason why it checked fail.

PR #9163, opened 2025-04-11
Add generated definition for Philips Hue Centris 929003808501

francoisverbeek:
Added generated definition for Philips Hue Centris 929003808501


Koenkk:
Thanks!

PR #9177, opened 2025-04-14
Update sonoff.ts

chris-1243:
Add SNZB-02LD according to https://github.com/Koenkk/zigbee2mqtt/issues/27055

CC: @MiAutomations

PR #9180, opened 2025-04-14
Update sonoff.ts

chris-1243:
Slight syntax update.

I just realized this as #9177 is made from this device. I should have done this in my last PR


Koenkk:
Thanks for the quick fix!

PR #9191, opened 2025-04-17
Add Smoke Detector in shinasystems.ts

shinasys:
Model: FAM-300Z Smoke Detector
This device is a smoke detector that activates a fire alarm when a certain level of smoke is detected. It has two batteries: one to power the smoke sensor, and the other for Zigbee communication. Depending on hardware configuration, the device can also forcibly trigger a fire alarm signal remotely.


shinasys:
Further testing is required, so I will submit another pull request.

PR #9206, opened 2025-04-19
Add Ion Industries dimmer TZE200_0hb4rdnp support

aniro:
[TZE200_0hb4rdnp](https://www.elektramat.nl/ion-industries-universele-led-zigbee-dimmer-set-master-slave-0-3-200w-90-500-090/) dimmer is essentially the same as [_TZE200_ykgar0ow](https://www.elektramat.nl/ion-industries-universele-led-zigbee-hue-compatible-draaidimmer-0-3-200w-id200w-zigb/) with an additional slave module.

The only issue I encountered while testing it locally: 
Reporting of the current state back to z2m does not work in case the dimmer is used in "no-neutral" mode. With the neutral wire connected, physical interactions with the knob are reflected in z2m state correctly.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2213,6 +2213,7 @@ export const definitions: DefinitionWithExtend[] = [
             "_TZE204_vevc4c6g",
             "_TZE200_0nauxa0p",
             "_TZE200_ykgar0ow",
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #9207, opened 2025-04-19
Detect CK-BL702-AL-01(7009_Z102LG04-1) as Tuya CK-BL702-AL-01

KapJI:
Similar to https://github.com/Koenkk/zigbee-herdsman-converters/commit/29653044f43b6c54f6a6c7c6ebab5ec45c05045a

I haven't tested it as I don't know an easy way to patch converters in HAOS environment. But given it works fine without support and model is the same, I believe it should be ok.

This is GX53 variant of Tuya RGBWW light bulb.

Reference picture:
tuya-gx53

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1556,7 +1556,7 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #9219, opened 2025-04-21
Add somfy sonesse 30 DC 24v

Sytten:
This device https://www.somfysystems.com/en-us/products/1241970/sonesse-30-24v-dc-zigbee, similar to the others but hard wired


Koenkk:
Thanks!

PR #9222, opened 2025-04-21
Update Aqara WS-EUK04 power measurement

priitliivak:
Aqara exposes standard activePower attribute for their latest switches.

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -942,10 +942,10 @@ export const definitions: DefinitionWithExtend[] = [
         model: "WS-EUK04",
         vendor: "Aqara",
         description: "Smart wall switch H1 EU (with neutral, double rocker)",
-        fromZigbee: [fz.on_off, lumi.fromZigbee.lumi_power, lumi.fromZigbee.lumi_action_multistate, lumi.fromZigbee.lumi_specific],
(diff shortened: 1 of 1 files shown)


src/devices/lumi.ts
```diff
@@ -942,10 +942,10 @@ export const definitions: DefinitionWithExtend[] = [
         model: "WS-EUK04",
         vendor: "Aqara",
         description: "Smart wall switch H1 EU (with neutral, double rocker)",
-        fromZigbee: [fz.on_off, lumi.fromZigbee.lumi_power, lumi.fromZigbee.lumi_action_multistate, lumi.fromZigbee.lumi_specific],
+        fromZigbee: [fz.on_off, fz.electrical_measurement, lumi.fromZigbee.lumi_action_multistate, lumi.fromZigbee.lumi_specific],
```
Koenkk, line 945:
```suggestion
        fromZigbee: [fz.on_off, lumi.fromZigbee.lumi_action_multistate, lumi.fromZigbee.lumi_specific],
        extend: [m.electricityMeter()],
```

priitliivak: No.  
m.electricityMeter({voltage: false, current: false }) almost works but when configuring reporting for activePower the system gets SOFTWARE_ERROR

Koenkk: > So currently I have m.electricityMeter({voltage: false, current: false, configureReporting: false}) but not sure if it makes sense to use electricityMeter at all, when configureReporting:false

That's fine!

PR #9223, opened 2025-04-21
add EPTTECH tank level monitor

markxroberts:


src/devices/tuya.ts
```diff
@@ -564,6 +564,34 @@ const tzLocal = {
             }
         },
     } satisfies Tz.Converter,
+    EPTTECH_TLC2206_ZB : {
```
Koenkk, line 567: Please use `tuyaDatapoints` instead of introducing a converter

PR #9226, opened 2025-04-22
Add 3RSB02015Z

3reality-support:


src/devices/third_reality.ts
```diff
@@ -277,6 +277,29 @@ export const definitions: DefinitionWithExtend[] = [
         toZigbee: [tz.cover_state, tz.cover_position_tilt],
         exposes: [e.cover_position()],
     },
+    {
+        zigbeeModel: ["3RSB02015Z"],
+        model: "3RSB02015Z",
+        vendor: "Third Reality",
+        description: "Smart Blind Gen2",
+        extend: [
+            m.battery(),
```
Koenkk, line 286: Can you use `m.windowCovering` instead of the from/toZigbee and exposes?

Koenkk:
Thanks!

PR #9234, opened 2025-04-23
Update tuya.ts with new ClickSmart+ Devices

dodger421:
Add new ClickSmart+ Devices, recognised by zigbee2mqtt as generic Tuya devices;

CSP041 1 gang switch module without neutral wire
CMA30651 13A smart switched fused connection unit


Koenkk:
Thanks!

PR #9236, opened 2025-04-24
Fix device type for "_TZ3000_gjrubzje" to be recognized as EndDevice

omniwired:
# Fix device type for "_TZ3000_gjrubzje" to be recognized as EndDevice

## Problem
The device with manufacturer ID "_TZ3000_gjrubzje" was incorrectly being recognized as a Router device type, when it should be an EndDevice. This causes issues with proper device functionality in the network.

## Solution
Modified the configuration function for the "TS0001_switch_module" device to check if the device has the manufacturer ID "_TZ3000_gjrubzje" and if so, override the device type to "EndDevice" and save the change.

## Changes Made
- Added a condition in the `configure` function of the `TS0001_switch_module` definition to check the manufacturer name
- Set the device type to "EndDevice" when the condition is met
- Added the `device.save()` call to persist the change

## Code Changes
```js
    await reporting.bind(device.getEndpoint(1), coordinatorEndpoint, ["genOnOff"]);
    
    // Device incorrectly identifies as Router but is actually an EndDevice
    if (device.manufacturerName === "_TZ3000_gjrubzje") {
        device.type = "EndDevice";
        device.save();
    }
},
```

## Testing
Tested with a physical "_TZ3000_gjrubzje" device. The device now correctly appears as an EndDevice in the network and functions properly.

## Related Information
This fix follows the same pattern used for similar issues in the codebase, such as the fix for "QS-Zigbee-SEC01-U" device.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -4364,6 +4364,12 @@ export const definitions: DefinitionWithExtend[] = [
         configure: async (device, coordinatorEndpoint) => {
             await tuya.configureMagicPacket(device, coordinatorEndpoint);
             await reporting.bind(device.getEndpoint(1), coordinatorEndpoint, ["genOnOff"]);
+
+            // Device incorrectly identifies as Router but is actually an EndDevice
+            if (device.manufacturerName === "_TZ3000_gjrubzje") {
+                device.type = "EndDevice";
+                device.save();
+            }
         },
     },
     {


omniwired:
I've also build an extention with the following code:

```js
class DeviceTypeFixExtension {
    constructor(
        zigbee,
        mqtt,
        state,
        publishEntityState,
        eventBus,
        enableDisableExtension,
        restartCallback,
        addExtension,
        settings,
        logger,
    ) {
        this.zigbee = zigbee;
        this.mqtt = mqtt;
        this.state = state;
        this.publishEntityState = publishEntityState;
        this.eventBus = eventBus;
        this.enableDisableExtension = enableDisableExtension;
        this.restartCallback = restartCallback;
        this.addExtension = addExtension;
        this.settings = settings;
        this.logger = logger;

        this.logger.info('Loaded DeviceTypeFixExtension for _TZ3000_gjrubzje device');
        this.targetDeviceIeeeAddr = '0xa4c138ded946710d';
    }

    /**
     * Called when the extension starts
     */
    start() {
        this.logger.info('Starting DeviceTypeFixExtension');
        
        // Fix device type when the extension starts
        this.fixDeviceType();
        
        // Also listen for device join events to fix newly joined devices
        this.eventBus.onDeviceJoined(this, this.onDeviceJoined.bind(this));
    }

    /**
     * Called when the extension stops
     */
    stop() {
        this.eventBus.removeListeners(this);
    }

    /**
     * Fix the device type for our target device
     */
    fixDeviceType() {
        try {
            // Get the device by IEEE address - using the correct API
            const device = this.zigbee.resolveEntity(this.targetDeviceIeeeAddr);
            
            if (device && device.device) {
                const zigbeeDevice = device.device;
                this.logger.info(`Found target device: ${this.targetDeviceIeeeAddr}, current type: ${zigbeeDevice.type}`);
                
                // Only change if it's currently a Router
                if (zigbeeDevice.type === 'Router') {
                    // Change device type to EndDevice
                    zigbeeDevice.type = 'EndDevice';
                    zigbeeDevice.save();
                    
                    this.logger.info(`Successfully changed device type to EndDevice for ${this.targetDeviceIeeeAddr}`);
                    this.mqtt.publish('zigbee2mqtt/device_type_fix', JSON.stringify({
                        device: this.targetDeviceIeeeAddr,
                        manufacturerName: '_TZ3000_gjrubzje',
                        oldType: 'Router',
                        newType: 'EndDevice',
                        status: 'success'
                    }));
                } else {
                    this.logger.info(`Device ${this.targetDeviceIeeeAddr} is already type: ${zigbeeDevice.type}`);
                }
            } else {
                this.logger.info(`Device ${this.targetDeviceIeeeAddr} not found`);
            }
        } catch (error) {
            this.logger.error(`Error fixing device type: ${error.message}`);
        }
    }

    /**
     * When a device joins, check if it's our target and fix if needed
     */
    onDeviceJoined(data) {
        const device = data.device;
        
        if (device && device.ieeeAddr === this.targetDeviceIeeeAddr) {
            this.logger.info(`Target device joined: ${device.ieeeAddr}`);
            
            // Wait a bit for the device to fully initialize
            setTimeout(() => {
                this.fixDeviceType();
            }, 5000);
        }
    }
}

// eslint-disable-next-line no-undef
module.exports = DeviceTypeFixExtension;
```

omniwired:
its super important not to have the device set at optimistic and also to enable Availability on the settings; so it won't go to sleep, I will report that after a few days.

Koenkk:
> This causes issues with proper device functionality in the network.

This change will not affect the functionality, it is purely cosmetic

omniwired:
Hi @Koenkk, I will test for a few more days and let you know, but this morning it was disconnected from the network again.
I don't know the reason, but it seems far more stable than before, before the changes it was breaking in less than one hour.
Also, for my particular model, the power usage report is not working.

I understand this device connects to mains power and, according to Zigbee standards, it should function as a router. However, Tuya devices often exhibit quirks. The model I have utilizes a neutral wire, which should allow it to act as a router, and it does indeed function as one. The issue is that it disconnects very frequently.

omniwired:
Carefully testing this didn't accomplish anything, so I will close this for PR.

PR #9242, opened 2025-04-25
fix modern extends numeric/binary not configuring attributes (Koenkk/zigbee2mqtt#27173)

canardos:
Fix for `numeric` and `binary` functions not reading attributes during configuration See issue:

Koenkk/zigbee2mqtt#27173

Note that the `numeric` and `binary` function signatures uses:

```
    reporting?: false | ReportingConfigWithoutAttribute;
``` 

So it seems that some functions use `undefined` to reflect no reporting arg and some use `false`.

I modified `setupConfigureForReporting` to to accept `false | undefined`, but another option would be:

```
const configure: Configure[] = [setupConfigureForReporting(cluster, attribute, reporting === false ? undefined : reporting, access, endpoints)];
```
In `numeric`/`binary`.

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -189,8 +189,8 @@ export async function setupAttributes(
 export function setupConfigureForReporting(
     cluster: string | number,
     attribute: ReportingConfigAttribute,
-    config: ReportingConfigWithoutAttribute,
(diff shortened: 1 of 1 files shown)


Koenkk:
Looks good, thanks!

PR #9244, opened 2025-04-25
fix(ignore): better exposes typing

Nerivec:
Enforce proper `type` by assigning it outside of ctor.

@Koenkk can you check that amina converter? It's forcing a change of the List `item_type`, I assume this is fine (list of enum)?


Koenkk:
Looks good, thanks!

PR #9250, opened 2025-04-27
update 3RDP01072Z configure

3reality-support:


src/devices/third_reality.ts
```diff
@@ -484,6 +484,28 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Third Reality",
         description: "Zigbee / BLE dual plug with power",
         ota: true,
+        configure: async (device, coordinatorEndpoint) => {
```
Koenkk, line 487: The reporting should already be set up by default (through ` m.electricityMeter`), so there should be no need for this as this provides good default values. In case a user wants different ones, they can change it via the reporting tab.

Koenkk: > and Divisor can only read in Zigbee

With `m.electricityMeter` the divisor is read from the device (via Zigbee). In case the device has the incorrect divisor, you can override it in the configure.

As an alternative, we can extend `m.electricityMeter()` such that you can pass the divisor (e.g. extend `ReportingConfigWithoutAttribute`)

```
energy?: false | (MultiplierDivisor & Partial<ReportingConfigWithoutAttribute>);
```

3reality-support: ok

PR #9260, opened 2025-04-28
Aqara Climate Sensor W100

kirovilya:


Koenkk:
Thanks!

src/devices/lumi.ts
```diff
@@ -4562,4 +4564,166 @@ export const definitions: DefinitionWithExtend[] = [
...
+                lookup: {no: 0, threshold: 1, period: 2, threshold_period: 3},
+                cluster: "manuSpecificLumi",
+                attribute: {ID: 0x0165, type: Zcl.DataType.UINT8},
+                description: "Temperature reporting mode",
+                zigbeeCommandOptions: {manufacturerCode},
+            }),
+            m.numeric({
+                name: "temp_period",
+                valueMin: 1,
+                valueMax: 10,
```
hagih, line 4673: @kirovilya 
These values are not correct for the reporting period of the temperature via zigbee.
This means that the device would report an update via zigbee every 1 second at the minimum setting and report via zigbee every 10 seconds at the maximum setting. Both pretty unreasonable values.

The Aqara app seems to have a minimum 10s value here, not a maximum 10s value:
https://forum.aqara.com/t/w100-explanation-of-the-temperature-sampling-setting/52930

Edit: It's possible 600 seconds / 10 minutes (rather than 10 seconds) might be the correct maximum here based on [this article](https://homekitnews.com/2025/05/15/aqara-climate-monitor-w100-review/).

Discussion: [here](https://github.com/Koenkk/zigbee2mqtt/issues/28521#issuecomment-3482084668)

src/devices/lumi.ts
```diff
@@ -4562,4 +4564,166 @@ export const definitions: DefinitionWithExtend[] = [
...
+                lookup: {no: 0, threshold: 1, period: 2, threshold_period: 3},
+                cluster: "manuSpecificLumi",
+                attribute: {ID: 0x016c, type: Zcl.DataType.UINT8},
+                description: "Humidity reporting mode",
+                zigbeeCommandOptions: {manufacturerCode},
+            }),
+            m.numeric({
+                name: "humi_period",
+                valueMin: 1,
+                valueMax: 10,
```
hagih, line 4705: Same issue as above here

src/devices/lumi.ts
Koenkk: Could you submit a pr to fix?

> 
Could you make a PR by clicking [here](https://github.com/Koenkk/zigbee-herdsman-converters/edit/master/src/devices/lumi.ts)?

MoeRT09: I created a PR: #10669

PR #9264, opened 2025-04-29
Add support for soil moisture sensor Simpla Home Soil Pro

SimplaHome:
Add support for [Simpla Home Soil Pro](https://www.simpla-home.at/de/product/soil-pro/) and prepare structure for upcoming products.

[Image Pull Request ](https://github.com/Koenkk/zigbee2mqtt.io/pull/3676) and [update](https://github.com/Koenkk/zigbee2mqtt.io/pull/3752) was performed by fwmone (thanks!) and is already merged.


Koenkk:
Is this pr ready for merge? (I see it is in draft)

SimplaHome:
> Is this pr ready for merge? (I see it is in draft)
Hi Koen,
not yet.
btw.: thanks for your great work!

SimplaHome:
pr is postponed and closed for now to keep the open pr list of zigbee-herdsman-converters clean.

SimplaHome:
PR is now ready for review/merge

src/devices/simpla_home.ts
```diff
@@ -0,0 +1,93 @@
...
+                        .withUnit("s")
+                        .withValueMin(5)
+                        .withValueMax(4 * 60 * 60)
+                        .withCategory("config")
+                        .withDescription("Defines how often the device performs measurements"),
+                );
+            }
+            return dynExposes;
+        },
+        fromZigbee: [
```
Koenkk: For both from and toZigbee converters, can you use `m.numeric()` in extend instead? Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/a204ef1131fe4d06594591cff8edf866ae5b7395/src/lib/lumi.ts#L1621

SimplaHome: Hi Koen, thanks for the hint. I've adjusted the code to use m.numeric().

src/devices/simpla_home.ts
```diff
@@ -0,0 +1,102 @@
...
+            unit: resultUnit,
+            cluster: "genAnalogOutput",
+            attribute: "presentValue",
+            scale: 1,
+            valueMin: measurementIntervalMin,
+            valueMax: measurementIntervalMax,
+            description: resultDescription,
+            ...args,
+        });
+
```
Koenkk, line 26: I think this would be cleaner here:

```
const expose = result.exposes[0];
```

and then `return [expose]` in the function below.

src/devices/simpla_home.ts
```diff
@@ -0,0 +1,102 @@
...
+            },
+        ],
+        model: "Soil Pro",
+        vendor: "Simpla Home",
+        description: "Soil moisture sensor: Simpla Home Soil Pro",
+        extend: [
+            m.deviceEndpoints({endpoints: {"1": 1, z1_top: 2, z2_bottom: 3}}),
+            m.identify(),
+            m.temperature({
+                reporting: {min: measurementIntervalMin, max: measurementIntervalMax, change: 10},
```
Koenkk, line 54: What's the reason to deviate from the standard reporting intervals? Unless there is a really good reason we want to use the default for all (to give a consistent experience).

src/devices/simpla_home.ts
```diff
@@ -0,0 +1,102 @@
...
+        ];
+
+        return result;
+    },
+};
+
+export const definitions: DefinitionWithExtend[] = [
+    {
+        zigbeeModel: ["Soil Pro"],
+        fingerprint: [
```
Koenkk: `fingerprint` can be removed here (`zigbeeModel`) does the job.

src/devices/simpla_home.ts
```diff
@@ -0,0 +1,102 @@
...
+            await endpointId.read("genBasic", ["swBuildId"]);
+
+            if (Number(`0x${device?.softwareBuildID}`) > 0x01010101) {
+                const endpointAnalogOutput = device.getEndpoint(2);
+                await endpointAnalogOutput.bind("genAnalogOutput", coordinatorEndpoint);
+                await endpointAnalogOutput.read("genAnalogOutput", ["presentValue"]);
+            }
+        },
+        meta: {multiEndpoint: true},
+        ota: true,
```
Koenkk, line 84: For this to work an OTA has to be submitted to https://github.com/koenkk/zigbee-ota

src/devices/simpla_home.ts
SimplaHome: Could you please explain that in more detail? I couldn't find an example of what it should look like.

src/devices/simpla_home.ts
SimplaHome: The device supports a configurable measurement interval from 5s to 4h. The interval can be set via MQTT. The idea was to have the reporting intervals aligned. It might be confusing if the user selects a 5s interval in e.g. Home Assistant but only receives data every 10s. The 4h interval would be to save energy. Is that a good enough reason or would you prefer the default nevertheless?

src/devices/simpla_home.ts
SimplaHome: I've removed the fingerprint section

src/devices/simpla_home.ts
SimplaHome: Currently there is no OTA available, but it would be great to have this possibility in the future without changing herdsman converters. Should it be removed nevertheless?

Btw.: Is there a possibility to use a self-hosted index.json + OTA without manual intervention by the user through "OTA index override file name"?

Koenkk: Yes local OTA index is possible with that.

src/devices/simpla_home.ts
Koenkk: >  The idea was to have the reporting intervals aligned. It might be confusing if the user selects a 5s interval in e.g. Home Assistant but only receives data every 10s. 

Where does the user select this report interval?

SimplaHome: The user selects the device-internal update interval through the "measurement_interval" numeric (analog output cluster). However, the minimum reporting interval limits the visibility of the updated measurements.

Koenkk: I see, could you document that above this line? So if understand correctly the `measurement_interval` cannot be set lower than the min reporting interval configured here.

src/devices/simpla_home.ts
Koenkk: Done!

src/devices/simpla_home.ts
SimplaHome: I've added a comment. I hope this makes things clearer.

Koenkk:
Thanks!

SimplaHome:
Thank you! :)

PR #9265, opened 2025-04-30
Newly added and repaired

sdbldn:
Some equipment cannot operate in the original directory, add support status for "ZHT-002", and repair the humidity abnormality of "801 V2".


Koenkk:
Thanks!

PR #9271, opened 2025-05-02
fix(detect): Add _TZE204_r731zlxk fingerprint to TB26-6 device

akelmanson:
This PR adds the _TZE204_r731zlxk fingerprint to the TB26-6 device from Zemismart that already has the _TZE200_r731zlxk fingerprint.


Koenkk:
Thanks!

PR #9272, opened 2025-05-02
fix: trigger IKEA unfreeze on color changes too

lbschenkel:
All color changes with transition also freeze the lights, not just temperature changes.

See: https://github.com/Koenkk/zigbee2mqtt/issues/18574
See: https://github.com/Koenkk/zigbee-herdsman-converters/pull/8637

cc @Koenkk @bobrippling


lbschenkel:
I've tested this change with my IKEA lights and they all unfreeze successfully now for both temperature and color changes.

bobrippling:
LGTM - I was going to comment about other commands but that's interesting that only `color` freezes them!

lbschenkel:
Please hold merging, I want to test all `*_step` commands as well to make sure if any extra keys are needed.

PR #9278, opened 2025-05-03
update shelly gen 4 devices to enable energy metering

marce14-git:
as of shelly firmware 1.6 the energy consumption should work natively

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -21,13 +21,13 @@ export const definitions: DefinitionWithExtend[] = [
         model: "S4SW-001P8EU",
         vendor: "Shelly",
         description: "1PM Mini Gen 4",
-        extend: [m.onOff({powerOnBehavior: false}), m.electricityMeter({energy: true, producedEnergy: true, acFrequency: true})],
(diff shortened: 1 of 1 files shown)


src/devices/shelly.ts
```diff
@@ -13,21 +13,27 @@ export const definitions: DefinitionWithExtend[] = [
...
     },
     {
         zigbeeModel: ["1PM"],
         model: "S4SW-001P16EU",
         vendor: "Shelly",
         description: "1PM Gen 4",
-        extend: [m.onOff({powerOnBehavior: false}), m.electricityMeter({energy: false, producedEnergy: true, acFrequency: true})],
+        extend: [
+            m.onOff({powerOnBehavior: false}),
+            m.electricityMeter({producedEnergy: true, acFrequency: true, energy: {change: 1}, current: {change: 2}, power: {change: 100}}),
```
Koenkk, line 36:
```suggestion
            m.electricityMeter({producedEnergy: true, acFrequency: true, energy: {change: 1}, current: {change: 2}, power: {change: 100}}),
```

Can you try without the `change` for power/current and energy?

marce14-git: Yeah this should work too

Koenkk:
Thanks!

PR #9282, opened 2025-05-05
feat(devices): mark EPT TECH TLC2206 as mains-powered

k2extreme:
Adds a previously undefined powerSource fingerprint option to mark the EPT TECH TLC2206 as mains-powered (single phase), ensuring the Zigbee2MQTT UI displays “Mains” instead of a battery status.


Koenkk:
Thanks!

PR #9288, opened 2025-05-06
Convert Inovelli UI LED1-7 to API 0-6 for individualLedEffect

sprior:
This is the fix for issue #27319 which is caused because the UI numbers the LEDs from 1 - 7, but the Individual LED Effect call expects the LEDs to be numbered 0 - 6.

### DIFF
--- a/src/devices/inovelli.ts
+++ b/src/devices/inovelli.ts
@@ -1403,7 +1403,7 @@ const tzLocal = {
                 "individualLedEffect",
                 {
                     // @ts-expect-error ignore
-                    led: Math.min(Math.max(0, Number.parseInt(values.led)), 7),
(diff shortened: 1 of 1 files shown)


ReneTode:
first test something locally before a pr is made.
if that line is wrong then line 1410, 1412 and 1414 are also wrong.

sprior:
@ReneTode You aren't understanding the code, the other lines are fine as is.

rohankapoorcom:
Lines 1410, 1412, 1414 all look correct per the specs for the switch. The only issue here is that the led number is mapped incorrectly to the switch firmware. Making this change aligns with the switch firmware and matches how it is controlled on other platforms.

This PR should be good to go.

Koenkk:
Thanks!

sprior:
Thanks for accepting the PR!

PR #9293, opened 2025-05-06
Update Sunricher 4 channel remotes

chris-1243:
Detect ZG2819S-RGBW correctly and not as Iluminize 511.344 (keep it as whitelabel) and convert ZG2819S-CCT to modernExtend


chris-1243:
Is there any way to configure the reporting of the battery voltage and percentage on one endpoint only. I did not find anything in modernExtend (maybe I did not search correctly... sorry).

In my opinion, there is no need to have each endpoint to report such value

Thanks in advance

Koenkk:
> In my opinion, there is no need to have each endpoint to report such value

Agree, it is interesting that the device has multiple eps with the genPowerCfg cluster. We need to extend the modernExtend for this.

chris-1243:
By the way and after some tests, I just realized the binding of some clusters needs to be specified. 

```typescript
        const endpoint_1 = device.getEndpoint(1);
        await reporting.bind(endpoint_1, coordinatorEndpoint, ["genOnOff", "genPowerCfg"]);
        await reporting.batteryVoltage(endpoint_1);
        await reporting.batteryPercentageRemaining(endpoint_1);
        const endpoint_2 = device.getEndpoint(2);
        await reporting.bind(endpoint_2, coordinatorEndpoint, ["genOnOff"]);
        const endpoint_3 = device.getEndpoint(3);
        await reporting.bind(endpoint_3, coordinatorEndpoint, ["genOnOff"]);
        const endpoint_4 = device.getEndpoint(4);
        await reporting.bind(endpoint_4, coordinatorEndpoint, ["genOnOff"]);
    },
``` 

The "old" fashioned way is working great, I am a bit stuck with the modernExtend. Each attempt I have done is binding the main cluster (OnOff, Scenes, ....) for each endpoint. This is problematic for the `genScenes` cluster. If the binding to the coordinator is done then each scene button press sends four commands at the same time (recall_1_ep1, recall_1_ep2, recall_1_ep3 and recall_4_ep4). Any hint to provide or at least where to look at ?

chris-1243:
> Agree, it is interesting that the device has multiple eps with the genPowerCfg cluster. We need to extend the modernExtend for this.

We could. What I have conclude after tons of tries is the modernExtend is always binding the main clusters and as I wrote previously it is not the best for this device.
To go with an approach where you have to bind then to force unbind multiple cluster for each endpoint is not a good approach.

I will revert some changes and use the "old" syntax which works perfectly

Koenkk:
Thanks!

PR #9295, opened 2025-05-07
Superled 70012 does not support effect

ikke-t:
Hi,

I remove the effect from superled dimmer, as it complains it's not supported command.

Will it be removed from the docs by some script automation, or should I edit and create PR for [this](https://github.com/Koenkk/zigbee2mqtt.io/blob/master/docs/devices/70012.md) separately?


Koenkk:
Thanks!

Docs will be updated with the next release.

PR #9302, opened 2025-05-08
Hot fix: ZG2189S-RGBW and CCT

chris-1243:
It fixes errors (binding) I made after #9293. Only `genPowerCfg` on endpoint 1 should be bound to the coordinator.


Koenkk:
Thanks!

PR #9307, opened 2025-05-09
Fix Sonoff S60ZBTPF plug power meter data

ccorderor:
In the original support for the Sonoff S60ZBTPF plug, power consumption monitoring of the device is not included. I simply added that data, and it now correctly displays voltage, power, and current.


Koenkk:
Thanks!

martin-in-da-house:
Doesn't Work anymore.

PR #9308, opened 2025-05-09
Add Nedis ZBSC10WT whitelabel

MilanFIN:
Added [Nedis ZBSC10WT](https://nedis.com/en-us/product/household-and-living/climate/humidity-control/550726063/smartlife-climate-sensor-zigbee-30-battery-powered-android-ios-white?_ga=undefined) sensor.


Koenkk:
Thanks!

PR #9310, opened 2025-05-09
Fixed percent bug of Longsam M3 curtain motor.

devkaiwang:
Issue: https://github.com/Koenkk/zigbee2mqtt/issues/26660

Added an option `cover_position_percent_fix` for inverting percentage in `toZigbee`.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -564,6 +564,14 @@ const tzLocal = {
             }
         },
     } satisfies Tz.Converter,
+    invert_cover_percent_fix: {
+        key: ["state", "position"],
+        convertSet: async (entity, key, value, meta) => {
+            const shouldInvert = key === "position" && meta.options.cover_position_percent_fix;
+            const newValue = shouldInvert ? 100 - Number(value) : value;
+            return await legacy.toZigbee.tuya_cover_control.convertSet(entity, key, newValue, meta);
+        },
+    } satisfies Tz.Converter,
 };
 
 const fzLocal = {
@@ -4444,7 +4452,6 @@ export const definitions: DefinitionWithExtend[] = [
                 "_TZE200_gaj531w3",
                 "_TZE200_yia0p3tr",
                 "_TZE200_rsj5pu8y",
-                "_TZE204_xu4a5rhj",
                 "_TZE204_nladmfvf",
                 "_TZE200_2odrmqwq",
                 "_TZE204_lh3arisb",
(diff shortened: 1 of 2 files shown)


Koenkk:
- I would not make an option for this but always do it
- Can you integrate the fix directly into `legacy.toZigbee.tuya_cover_control`? Than we don't have to make a new definition for this device. In the converter you can apply the fix for just this device by using `if (meta.device.manufacturerName === "_TZE204_xu4a5rhj")`

fivemeyestore:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

fivemhostingserver:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

eyesaider:
Great implementation of this feature. I particularly like the approach to error handling.

devkaiwang:
Hi @Koenkk , thank you for the review! 

I agree on your simpler way, however, I'm not sure if, for the updated version of the motor in the future, the `manufacturerName` will be the same or a different one. If the fixed version of this motor uses the same id, we need this option for those ones with the flaw, I don't think a massive recall will be applied.

May I leave this option as it is for now and if Longsam the manufacturer use a different name, I will remove this extra option.

What do you think?

Thanks!

Koenkk:
The `manufacturerName` very likely stays the same (I've never seen an OTA changing it), so let's integrate it in the existing converter.

devkaiwang:
> The manufacturerName very likely stays the same (I've never seen an OTA changing it), so let's integrate it in the existing converter.

If the `manufactureName` stays the same, we will have two types of the devices, one is with the bug and the other is without the bug. So we need the option for the buggy ones and let the user selectively turn on the option for the flaw ones.

Koenkk:
> So we need the option for the buggy ones and let the user selectively turn on the option for them.

That we need to figure out later (probably we can detect based on the `swBuildId` or something, but we don't know what the one of the fixed one will be.

devkaiwang:
Hi @Koenkk , I contacted the after-sales service again, they told me that they will update the "board program" of motors and the issue will be fixed in the next new release. There be no ID/Name change but program update for the "board" (though I am not sure if we can detect the revision). They will give me a replace later.

I think we should keep the option for the ones with flaw since if we hard-code the fix for this model, bug will appear immediately once they updated the board program. Or, we should postpone the PR for their update, but I'm not sure if I can distinguish the board program update.

Koenkk:
If they dont change the manuf name then it will also break with the current pr. We can only figure out how to distinguish it once we have the new model.

devkaiwang:
Current PR won't be broken since fix logic is only enabled when the added option is on. The option is off by default. If users found the issue they can turn the option on. So for the fixed release of the motor, there is nothing changed, it should just work well.

Koenkk:
I see, thanks!

Make sure that:
- The filename is `M3TYW-2.0-13.png`
- The size is 512x512

kyangk:
I think this change broke my curtain positions. 

I am on version `23.69.0`

Zigbee Model
TS0601
Zigbee Manufacturer
_TZE204_xu4a5rhj
Manufacturer
[Tuya](https://www.zigbee2mqtt.io/supported-devices/#v=Tuya)
Model
[M3TYW-2.0-13](https://www.zigbee2mqtt.io/devices/M3TYW-2.0-13.html#tuya-m3tyw-2.0-13)

When I set position to 26%
```
Transmitted 2:42:17 PM
QoS: 0
Payload: position: 26
```

The position jumps to 74
```
2025-08-03 14:42:20z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Living Room Back Curtains', payload '{"linkquality":176,"options":null,"position":74,"running":false,"state":"OPEN"}'
```

The actual curtains are open 74% (approximately 3/4 open; 1/4 covered)

Is there anyway for me to revert to the previous behavior to test?

Koenkk:
Could you both provide the data/database.db entry of your device?

PR #9158, opened 2025-04-10
Extend Tuya Modern Extend and support switches with scene and backlight

devkaiwang:
I have these samples and will add 1-gang in the future. All tested.

switch samples

The feature support is now the same as in the Tuya app:

tuya_scene_switch_backlight

PR #9313, opened 2025-05-11
fix: Add missing action in TS0726_3_gang_scene_switch

anyhow-sg-smarthome:
Recent change in https://github.com/Koenkk/zigbee-herdsman-converters/pull/9158 is missing scene actions which affects my switches thus adding them. I am unfamiliar with the new `modernExtend` thus do update me if I am doing wrong


Koenkk:
Thanks!

PR #9330, opened 2025-05-13
Update zigbeetlc.ts

igo-r:
1. The maximum limit of the "Measurement interval" has been changed to 30 seconds.
In the new firmware version v0.1.2.7, the Measurement interval limits have been changed.
https://github.com/pvvx/ZigbeeTLc/releases/tag/v0.1.2.7

2. Changed configuration reporting for temperature and humidity.
Due to the high battery drain of the CR2032, due to the large number of reports being sent unnecessarily.

3. MHO-C122 and MHO-C401 are modified as independent devices.
These devices are very popular, and therefore they have been separated to correctly display the model and the correct image of the device.
https://github.com/Koenkk/zigbee2mqtt.io/pull/3776


Koenkk:
Thanks!

PR #9347, opened 2025-05-14
Fix: typo in modernExtend and fromZigbee

chris-1243:
While searching in modernExtend I found a small error. I may be wrong and it might be correct. I also corrected fromZigbee as well


Koenkk:
Thanks!

PR #9357, opened 2025-05-17
Fix: 3RTHS24BZ

chris-1243:
Fix Power source of this end device as "battery" powered according to the issue

https://github.com/Koenkk/zigbee2mqtt/issues/22914#issuecomment-2888547104


Koenkk:
Thanks!

PR #9363, opened 2025-05-19
fix: tuya/MG-ZG01W: remove endpoint l1 to simplify definition

hexchain:
The switch only has 1 input so it's probably not useful to set up an endpoint.

This should also fix the case where `state` and `state_l1` can sometimes get out-of-sync.


Koenkk:
Thanks!

PR #9366, opened 2025-05-19
Legrand 067797 instead of BTicino 4411C

chpego:
I opened [a discussion on the subject](https://github.com/Koenkk/zigbee2mqtt/discussions/27264), but I didn't find an answer; so I created the corresponding file in external converter and it seems OK, I also created a [PR for the image to be published](https://github.com/Koenkk/zigbee2mqtt.io/pull/3798)

don't hesitate to tell me if this isn't the right method, maybe it's better to use whitelabel, but I haven't succeeded.

### DIFF
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -698,4 +698,27 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.onOff(endpoint);
         },
     },
+    {
+        zigbeeModel: [" Dimmer switch with neutral\u0000\u0000\u0000\u0000"],
+        model: "067797",
+        vendor: "Legrand",
+        description: "Dimmer switch with neutral",
+        ota: true,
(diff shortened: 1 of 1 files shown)


src/devices/legrand.ts
```diff
@@ -698,4 +698,27 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.onOff(endpoint);
         },
     },
+    {
+        fingerprint: [{modelID: " Dimmer switch with neutral\u0000\u0000\u0000\u0000", ieeeAddr: /^0x0004............$/}],
```
Koenkk, line 702:
```suggestion
        fingerprint: [{modelID: " Dimmer switch with neutral\u0000\u0000\u0000\u0000", manufacturerName: "MANUF_NAME_FROM_LEGRAND"}],
```

Koenkk:
Thanks!

Issue #9369, opened 2025-05-19
Test failed adding new device


ChristianGr1974:
Hi,

today I tried to add a new device under `/src/devices/senoro.ts`
I have a fork on https://github.com/ChristianGr1974/zigbee-herdsman-converters/tree/senorowin
where I added the new device.

After executing the commands for the tests (as described in your Readme.md), I get the following errors:

```
✓ ZHC > verifies options filter 2ms
✓ ZHC > computes calibration/precision 29ms
✓ ZHC > instantiates list expose of number type 0ms
× ZHC > check if all definitions are imported in devices/index.ts 11ms
  → expected [ 'ITCommander', 'acova', …(337) ] to strictly equal [ 'ITCommander', 'acova', …(336) ]
✓ ZHC > instantiates list expose of composite type 1ms
✓ test/checks.test.ts (5 tests) 86ms

FAIL  test/index.test.ts > ZHC > check if all definitions are imported in devices/index.ts
AssertionError: expected [ 'ITCommander', 'acova', …(337) ] to strictly equal [ 'ITCommander', 'acova', …(336) ]

- Expected
+ Received

@@ -254,10 +254,11 @@
    "schneider_electric",
    "schwaiger",
    "seastar_intelligence",
    "securifi",
    "sengled",
+   "senoro",
    "sercomm",
    "shade_control",
    "shelly",
    "shenzhen_homa",
    "shinasystem",

 ❯ test/index.test.ts:599:30
    597|         const imports = Array.from(index.matchAll(importRegex)).map((r) => r[1]);
    598|         files.splice(files.indexOf("index"), 1);
    599|         expect(files.sort()).toStrictEqual(imports.sort());
       |                              ^
    600|     });
    601|
```

What am I doing wrong, or what is missing here?

kind regards
Christian

PR #9371, opened 2025-05-20
feat(add): implement Senoro Window Alarm device definitions

ChristianGr1974:
Currently test fails: https://github.com/Koenkk/zigbee-herdsman-converters/issues/9369#issuecomment-2894793910


Koenkk:
Please add it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts

ChristianGr1974:
Ok, I did this already. But I forgot to add ...senoro, to the definitions at the end of index.ts

PR #9375, opened 2025-05-21
update 3RSB02015Z

3reality-support:


src/devices/third_reality.ts
```diff
@@ -282,13 +282,13 @@ export const definitions: DefinitionWithExtend[] = [
         zigbeeModel: ["3RSB02015Z"],
         model: "3RSB02015Z",
         vendor: "Third Reality",
-        description: "Smart blind Gen2",
+        description: "Third Reality TRZB3",
```
Koenkk, line 285:
```suggestion
        description: "TRZB3",
```

Thrid reality is alrady in the `vendor`.

3reality-support: ok

PR #9377, opened 2025-05-21
feat(add): LCV002

pikeas:
Added https://www.philips-hue.com/en-us/p/bulb-st23-e26-smart-bulb/046677577520.

Zigbee model and model are both reported as LCV002, not sure if the EAN/UPC (046677577520) should be used as the model.

### DIFF
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -4039,6 +4039,13 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [philips.m.light({colorTemp: {range: [153, 500]}, color: {modes: ["xy", "hs"], enhancedHue: true}})],
     },
     {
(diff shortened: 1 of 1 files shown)


src/devices/philips.ts
```diff
@@ -4038,6 +4038,13 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Hue Lightguide E27 Edison ST72 500lm",
         extend: [philips.m.light({colorTemp: {range: [153, 500]}, color: {modes: ["xy", "hs"], enhancedHue: true}})],
     },
+    {
+        zigbeeModel: ["LCV002"],
+        model: "LCV002",
```
Koenkk, line 4043:
```suggestion
        model: "046677577520",
```

PR #9383, opened 2025-05-22
feat(senoro): update toZigbee configuration and adjust alarm datapoin…

ChristianGr1974:
Added converter for resetting alarm.

### DIFF
--- a/src/devices/senoro.ts
+++ b/src/devices/senoro.ts
@@ -12,16 +12,7 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Senoro",
         description: "Senoro window alarm",
         fromZigbee: [tuya.fz.datapoints],
-        toZigbee: [
-            {
-                key: ["alarm"],
-                convertSet: async (entity, key, value, meta) => {
-                    const boolValue = typeof value === 'boolean' ? value : value === 'true' || value === 1;
-                    await tuya.sendDataPointBool(entity, 16, boolValue);
-                    return {state: {alarm: boolValue}};
-                }
-            }
-        ],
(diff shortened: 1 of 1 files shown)


src/devices/senoro.ts
```diff
@@ -12,7 +12,16 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Senoro",
         description: "Senoro window alarm",
         fromZigbee: [tuya.fz.datapoints],
-        toZigbee: [tuya.tz.datapoints],
+        toZigbee: [
```
Koenkk, line 15: This will break other features, please revert it and fix it inside `[16, "alarm", tuya.valueConverter.raw],`

ChristianGr1974: ok, I removed that. It works as expected.

src/devices/senoro.ts
```diff
@@ -25,7 +25,7 @@ export const definitions: DefinitionWithExtend[] = [
             // All datapoints go in here
             tuyaDatapoints: [
                 [101, "opening_state", tuya.valueConverterBasic.lookup({open: 0, closed: 1, tilted: 2})],
-                [16, "alarm", tuya.valueConverter.trueFalse1],
+                [16, "alarm", tuya.valueConverter.raw],
```
Koenkk, line 28: Does it still send true/false after this change?

ChristianGr1974: Yes, this version is fine. the sensor sends true on alarm. To turn off alarm you are able to send false back. In case of an alarm - the sirene makes a loud sound - the device turns from end-device mode to normal mode, so that it is able to receive this command. With this version it'll work. THX!

PR #9391, opened 2025-05-23
Enable battery voltage reporting on some IKEA devices

vulpes2:
All recent battery powered IKEA devices support `batteryVoltage`, let's enable it for for these devices. None of the affected devices use coin cells so the impact on battery life should be negligible.


vulpes2:
I intentionally didn't enable that because it would introduce additional battery drain. Not all users are interested in the exact voltage of the battery so I think it's a good idea to leave it disabled by default.

bernardesarthur:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/d1e2d75a042b9ed7f6e6d516292686f54e569a69/src/lib/modernExtend.ts#L339

PR #9392, opened 2025-05-23
Fix detect _TZE200_vs0skpuc as a HOBEIAN ZG-227Z

bernardesarthur:
Fix detection of _TZE200_vs0skpuc as a HOBEIAN ZG-227Z


Koenkk:
thanks!

Issue #9403, opened 2025-05-25
Request: Add Enpoint choice in battery args


chris-1243:
Some remotes like:

- EVN  - ZBHS4RGBW
- Iluminize - 511.324;
- Iluminize - 511.344;
- RGB Genie - ZB-5028;
- Sunricher - ZG2819S-RGBW;
- Sunricher - ZG2819S-CCT;
- Sunricher - ZG2858A;

expose multiple endpoints with genPowerCfg cluster on each endpoint. When using moderExtend, Z2M will configure the battery reporting on all of them which does not make sense. Would it be possible to add some arguments there

https://github.com/Koenkk/zigbee-herdsman-converters/blob/d1e2d75a042b9ed7f6e6d516292686f54e569a69/src/lib/modernExtend.ts#L320

and after to be able to specify which endpoint to be used while configuring the reporting. By default, it should be set to `false` and to remain an option.

I had a look but my knowledge are too low to provide such change.

Thanks in advance

Koenkk:
Does this currently cause any issues?

Koenkk:
Done!

chris-1243:
Thanks a lot!!!!

PR #9416, opened 2025-05-26
fix: Configure battery attributes only for a single endpoint

Koenkk:
Fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/9403

CC: @chris-1243


chris-1243:
Thanks.

It should be used like below if I have understood correctly right?

`m.setupConfigureForReporting("genPowerCfg", "ReportingConfigAttribute", {endpointNames: ["1"]})`

PR #9420, opened 2025-05-27
Update tuya.ts

dannyuk1982:
Added Quoya MC11 Support


dannyuk1982:
You asked me to add it here: https://github.com/Koenkk/zigbee2mqtt/issues/27483

Unfortunately the Zemismart converter doesn't work at all for the Quoya MC11 blinds, which is why I had to add this.

dannyuk1982:
No worries. Does this then break the Zemismart? I'm completely new to this but seems like they both have identical fingerprints, but are different devices with different command sets. Not sure how this works to support both of them

Koenkk:
They have the same manufacturerName so should be ok, let's see!

PR #9436, opened 2025-05-31
Fix time on OWON PCT504

Rjevski:
The OWON PCO504 has a bug where it will always display the UTC time and provides no way of setting the timezone in its UI.

This PR changes it so it gets a custom time response pretending that the current local time is actually UTC and that the current timezone is also UTC (though technically this wouldn't be required, as it always insists on displaying UTC anyway).

I originally tried using the existing [`customTimeResponse("2000_LOCAL")`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/25a00150a482288974870749d6bc28e487d79552/src/lib/modernExtend.ts#L640) function, but that did not work. I assume that's because it only sets the `time` in the payload and sets all other attributes unset?

Either way I copied the [full time response](https://github.com/Koenkk/zigbee-herdsman/blob/56d010a1195af72cd6f241f6aef4f0ba268bd4e6/src/controller/model/device.ts#L417C25-L422C83) from the zigbee-herdsman and modified it as needed.

There could be an argument to modify the `customTimeResponse` to fake the other values too, but I didn't want to break any other devices that could potentially rely on the current behavior (it seems like only an Aqara device is currently using it - but I don't have one so can't verify against breaking changes), so in the meantime I just hardcoded it for this specific device only.

I have originally tested this using the following external converter:

```
import * as reporting from "zigbee-herdsman-converters/lib/reporting";

const e = exposes.presets;


/** @type{import('zigbee-herdsman-converters/lib/types').DefinitionWithExtend | import('zigbee-herdsman-converters/lib/types').DefinitionWithExtend[]} */
export default {
    zigbeeModel: ["PCT504", "PCT504-E"],
    model: "PCT504",
    vendor: "OWON",
    description: "HVAC fan coil with custom time 2",
    fromZigbee: [fz.fan, fz.thermostat, fz.humidity, fz.occupancy, fz.hvac_user_interface],
    toZigbee: [
        tz.fan_mode,
        tz.thermostat_occupied_heating_setpoint,
        tz.thermostat_unoccupied_heating_setpoint,
        tz.thermostat_occupied_cooling_setpoint,
        tz.thermostat_unoccupied_cooling_setpoint,
        tz.thermostat_min_heat_setpoint_limit,
        tz.thermostat_max_heat_setpoint_limit,
        tz.thermostat_min_cool_setpoint_limit,
        tz.thermostat_max_cool_setpoint_limit,
        tz.thermostat_local_temperature,
        tz.thermostat_keypad_lockout,
        tz.thermostat_system_mode,
        tz.thermostat_running_mode,
        tz.thermostat_running_state,
        tz.thermostat_programming_operation_mode,
    ],
    exposes: [
        e.humidity(),
        e.occupancy(),
        e
            .climate()
            .withSystemMode(["off", "heat", "cool", "fan_only", "sleep"])
            .withLocalTemperature()
            .withRunningMode(["off", "heat", "cool"])
            .withRunningState(["idle", "heat", "cool", "fan_only"])
            .withSetpoint("occupied_heating_setpoint", 5, 30, 0.5)
            .withSetpoint("unoccupied_heating_setpoint", 5, 30, 0.5)
            .withSetpoint("occupied_cooling_setpoint", 7, 35, 0.5)
            .withSetpoint("unoccupied_cooling_setpoint", 7, 35, 0.5),
        e.fan().withState("fan_state").withModes(["low", "medium", "high", "on", "auto"]),
        e.programming_operation_mode(["setpoint", "eco"]),
        e.keypad_lockout(),
        e.max_heat_setpoint_limit(5, 30, 0.5),
        e.min_heat_setpoint_limit(5, 30, 0.5),
        e.max_cool_setpoint_limit(7, 35, 0.5),
        e.min_cool_setpoint_limit(7, 35, 0.5),
    ],
        const binds = [
            "genBasic",
            "genIdentify",
            "genGroups",
            "hvacThermostat",
            "hvacUserInterfaceCfg",
            "hvacFanCtrl",
            "msTemperatureMeasurement",
            "msOccupancySensing",
        ];
        await reporting.bind(endpoint, coordinatorEndpoint, binds);
        await reporting.fanMode(endpoint);
        await reporting.bind(endpoint, coordinatorEndpoint, ["hvacThermostat"]);
        await reporting.thermostatOccupiedHeatingSetpoint(endpoint);
        await reporting.thermostatOccupiedCoolingSetpoint(endpoint);
        await reporting.thermostatTemperature(endpoint, {min: 60, max: 600, change: 0.1});
        await reporting.thermostatSystemMode(endpoint);
        await reporting.thermostatRunningMode(endpoint);
        await reporting.thermostatRunningState(endpoint);
        await reporting.humidity(endpoint, {min: 60, max: 600, change: 1});
        await reporting.thermostatKeypadLockMode(endpoint);

        await endpoint.read("hvacThermostat", [
            "systemMode",
            "runningMode",
            "runningState",
            "occupiedHeatingSetpoint",
            "unoccupiedHeatingSetpoint",
            "occupiedCoolingSetpoint",
            "unoccupiedCoolingSetpoint",
            "localTemp",
        ]);
        await endpoint.read("msRelativeHumidity", ["measuredValue"]);

        const endpoint2 = device.getEndpoint(2);
        await reporting.bind(endpoint2, coordinatorEndpoint, ["msOccupancySensing"]);
        await reporting.occupancy(endpoint2, {min: 1, max: 600, change: 1});
        await endpoint2.read("msOccupancySensing", ["occupancy"]);
    },
    extend: [
        {
            onEvent: [
                async (type, data, device, options, state) => {
                    if (!device.customReadResponse) {
                        device.customReadResponse = (frame, endpoint) => {
                            if (frame.isCluster("genTime")) {
                                const oneJanuary2000 = new Date("January 01, 2000 00:00:00 UTC+00:00").getTime();
                                const secondsUTC = Math.round((new Date().getTime() - oneJanuary2000) / 1000);
                                const time = secondsUTC - new Date().getTimezoneOffset() * 60;
        
                                const payload = {
                                    timeStatus: 3, // Time-master + synchronised
                                    time: time,
                                    timeZone: 0,
                                    localTime: time,
                                    lastSetTime: time,
                                    validUntilTime: time + 24 * 60 * 60, // valid for 24 hours
                                };
        
                                endpoint.readResponse("genTime", frame.header.transactionSequenceNumber, payload);
                                return true;
                            }
                            return false;
                        };
                    }
                },
            ],
            isModernExtend: true
        },
    ]
};
```

### DIFF
--- a/src/devices/owon.ts
+++ b/src/devices/owon.ts
@@ -129,6 +129,41 @@ const tzLocal = {
     } satisfies Tz.Converter,
 };
 
+const extendsLocal = {
+    localTimeAsUtc: {
+        // similar to the customTimeResponse("2000_LOCAL"), but also contains
+        // the timeStatus, timeZone, lastSetTime and validUntilTime.
+        // OWON devices seem to require this, as they ignored the customTimeResponse.
+            onEvent: [
+                async (type, data, device, options, state) => {
+                    if (!device.customReadResponse) {
+                        device.customReadResponse = (frame, endpoint) => {
+                            if (frame.isCluster("genTime")) {
+                                const oneJanuary2000 = new Date("January 01, 2000 00:00:00 UTC+00:00").getTime();
+                                const secondsUTC = Math.round((new Date().getTime() - oneJanuary2000) / 1000);
+                                const time = secondsUTC - new Date().getTimezoneOffset() * 60;
+        
+                                const payload = {
+                                    timeStatus: 3, // Time-master + synchronised
+                                    time: time,
+                                    timeZone: 0,
+                                    localTime: time,
+                                    lastSetTime: time,
+                                    validUntilTime: time + 24 * 60 * 60, // valid for 24 hours
+                                };
+        
+                                endpoint.readResponse("genTime", frame.header.transactionSequenceNumber, payload);
+                                return true;
+                            }
+                            return false;
+                        };
+                    }
+                },
+            ],
+            isModernExtend: true
+        },
+};
+
 export const definitions: DefinitionWithExtend[] = [
     {
         zigbeeModel: ["WSP402"],
@@ -385,6 +420,9 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.occupancy(endpoint2, {min: 1, max: 600, change: 1});
             await endpoint2.read("msOccupancySensing", ["occupancy"]);
         },
+        extend: [
+            extendsLocal.localTimeAsUtc,
+        ]
     },
     {
         zigbeeModel: ["PCT512"],


Koenkk:
> There could be an argument to modify the customTimeResponse to fake the other values too, but I didn't want to break any other devices that could potentially rely on the current behavior (it seems like only an Aqara device is currently using it - but I don't have one so can't verify against breaking changes)

I don't expect this to break so please add to `customTimeResponse`

Rjevski:
In the [original](https://github.com/Koenkk/zigbee-herdsman/blob/56d010a1195af72cd6f241f6aef4f0ba268bd4e6/src/controller/model/device.ts#L427) "time response" code there's this bit of code that seemingly selectively sets response attributes based on what was requested (instead of just sending everything):

```
if (frame.cluster.name in attributes) {
                const response: KeyValue = {};
                for (const entry of frame.payload) {
                    if (frame.cluster.hasAttribute(entry.attrId)) {
                        const name = frame.cluster.getAttribute(entry.attrId).name;
                        if (name in attributes[frame.cluster.name].attributes) {
                            response[name] = attributes[frame.cluster.name].attributes[name];
                        }
                    }
                }

                try {
                    await endpoint.readResponse(frame.cluster.ID, frame.header.transactionSequenceNumber, response, {
                        srcEndpoint: dataPayload.destinationEndpoint,
                    });
...
```

And then also sets `srcEndpoint`. I wonder how much of that is actually required? The [existing](https://github.com/Koenkk/zigbee-herdsman-converters/blob/25a00150a482288974870749d6bc28e487d79552/src/lib/modernExtend.ts#L659) "custom time response" does not do any of this and just sets fields directly.

Koenkk:
That would be better indeed, I guess the devices also requests all these attributes.

PR #9440, opened 2025-06-01
Added another version for ZG-2002-RF

qbaza:
My MHCOZY relay is incorrectly identified as Tuya TS0002 instead of Tuya ZG-2002-RF.


Koenkk:
Thanks!

PR #9426, opened 2025-05-29
fix: #9413 added execute_if_off for hue color lights

sjorge:
Only add to lights that support color (or colorTemp for levelConfig), as this was all I could test.

I was already using execute_if_off and that seems to work, I could not get any other levelConfig to work. I tried with a colorTemp only build and with a color build.

PR #9442, opened 2025-06-01
Revert "feat: Hue: add `execute_if_off` for colour lights (#9426)"

sjorge:
This reverts commit afdfbe4449893c129fedb9a3a23ac93ab1f305dc.

See the discussion on #9426 also [zigbee2mqtt#27568
](https://github.com/Koenkk/zigbee2mqtt/issues/27568)
tl;dr: everything i have access to WS or Color wise seems to support this, but others report getting UNSUPPORTED_ATTRIBUTE, I'm not sure how to filter which do and do not support this. The converters were already there so people can still set these using mqtt as before.


Koenkk:
Thanks!

PR #9447, opened 2025-06-02
Fix for Hive Keypad

thebigspin:
Fixes issue with Hive Keypad with null panel status, see: [https://github.com/Koenkk/zigbee2mqtt/issues/22211](https://github.com/Koenkk/zigbee2mqtt/issues/22211)

### DIFF
--- a/src/devices/zemismart.ts
+++ b/src/devices/zemismart.ts
@@ -595,7 +595,7 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint("TS0601", ["_TZE204_wwaeqnrf"]),
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE204_wwaeqnrf", "_TZE284_wwaeqnrf"]),
         model: "ZMS-206US-4",
         vendor: "Zemismart",
         description: "Smart screen switch 4 gang US",


src/devices/hive.ts
```diff
@@ -167,8 +167,10 @@ export const definitions: DefinitionWithExtend[] = [
         },
         onEvent: async (type, data, device) => {
             if (data.type === "commandGetPanelStatus" && data.cluster === "ssIasAce") {
+                const panelstatusobj = globalStore.hasValue(data.endpoint, "panelStatus") ? globalStore.getValue(data.endpoint, "panelStatus") : 0x00;
+
                 const payload = {
-                    panelstatus: globalStore.getValue(data.endpoint, "panelStatus"),
+                    panelstatus: panelstatusobj,
```
Koenkk:
```suggestion
                    panelstatus: globalStore.getValue(data.endpoint, "panelStatus", 0x00),
```

Can you check if this works?  Then this can be removed:

```js
const panelstatusobj = globalStore.hasValue(data.endpoint, "panelStatus") ? globalStore.getValue(data.endpoint, "panelStatus") : 0x00;
```

thebigspin: It seems I need to debug the solution more - both codes seem to work the same way when used as an external converter, and solve the *Parameter panelstStus is missing* error, but I'm now getting  *z2m: Invalid message 'undefined', skipping...* every time I try to arm or disarm with the keypad. 

Will try and have a better look when I have some time

PR #9448, opened 2025-06-02
Fix: fix detection WSD500A

chris-1243:
_TZ3000_bguser20 was wrongly detected as TH03

This reverts PR #9379 as it seems multiple devices use the same zigbee manufacturer but not for the same visual output.... According to https://zigbee.blakadder.com/Tuya_WSD500A.html, I made a mistake...

Sorry for the inconvenience.


Koenkk:
Thanks!

PR #9451, opened 2025-06-04
Fix TS0601 human presence sensor

poeggi:
Fixing sensor definition typos, missing underscore ("_") in lux values and redundant "_value" in motion sensitivity.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -18009,4 +18009,25 @@ export const definitions: DefinitionWithExtend[] = [
             }),
         ],
     },
+    {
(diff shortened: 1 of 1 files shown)


src/devices/neo.ts
```diff
@@ -519,7 +519,7 @@ export const definitions: DefinitionWithExtend[] = [
                 [19, "dis_current", tuya.valueConverter.raw],
                 [12, "presence_time", tuya.valueConverter.raw],
                 [13, "motion_far_detection", tuya.valueConverter.raw],
-                [15, "motion_sensitivity", tuya.valueConverter.raw],
+                [15, "motion_sensitivity_value", tuya.valueConverter.raw],
```
Koenkk: Please change the expose instead

PR #9455, opened 2025-06-04
Added Neutral Current support

Ruubah:
Added neutral current support for haElectricalMeasurement in modernExtend genericMeter.


Koenkk:
Thanks!

PR #9456, opened 2025-06-04
Detect Girier JR-ZPM01 as Tuya TS011F_plug_1

novittukyll:
Currently this plug(Girier JR-ZPM01) is detected as "Tuya TS011F_plug_3", but power monitoring is working without polling. So correct should be "Tuya TS011F_plug_1" ?

https://www.aliexpress.com/item/1005006243517488.html


Koenkk:
Thanks!

PR #9458, opened 2025-06-04
Update neo.ts

poeggi:
Fixing sensor definition typos, add missing underscore ("_") in lux values and remove redundant "_value" in motion sensitivity expose.


poeggi:
Changed patch as as requested to correct expose.

Koenkk:
Thanks!

PR #9460, opened 2025-06-05
Recognize "hold", "double" and "release" actions of Aqara Dimmer Switch H2 EU (Aqara KD-R01D)

mricsi97:
Some actions were missing from this dimmer switch. Sadly an Aqara hub is required to set it to multi-function mode. But after that these actions could be captured.


Koenkk:
Thanks!

PR #9462, opened 2025-06-05
Fix Aqara S04D threshold scale

jamesonuk:
As https://github.com/Koenkk/zigbee2mqtt/issues/27262#issuecomment-2934069964 the scale appears wrong on the threshold attribute so you are unable to set them properly which results in errors in HA.
Certainly more issues with this converter but this resolves that one.

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -4684,7 +4684,7 @@ export const definitions: DefinitionWithExtend[] = [
                 valueMin: 0.2,
                 valueMax: 3,
                 valueStep: 0.1,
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

J108560:
@jamesonuk For humi_threshold you have change valueMax key, not key scale! Is this so right?

jamesonuk:
> @jamesonuk For humi_threshold you have change valueMax key, not key scale! Is this so right?

No....  https://github.com/Koenkk/zigbee-herdsman-converters/pull/9470

PR #9463, opened 2025-06-05
Added "Multi-click" option for Aqara Dimmer Switch H2 EU (Aqara KD-R01D)

mricsi97:


Koenkk:
Thanks!

maluueu:
This may be a dumb question, but what exactly does this multi-click mode do or enable me to do?

PR #9465, opened 2025-06-05
Add Support of "On level" Setting for Aqara Ceiling light L1-350 (ZNXDD01LM)

kevintong116:
Add support of "On level" setting for Aqara Ceiling light L1-350 (ZNXDD01LM)

Working external converter:

```
import * as lumi from "zigbee-herdsman-converters/lib/lumi";

const e = exposes.presets;
const ea = exposes.access;

const {
    lumiZigbeeOTA,
    lumiLight,
} = lumi.modernExtend;

const NS = "zhc:lumi";
const {manufacturerCode} = lumi;

export default {
    zigbeeModel: ["lumi.light.acn003"],
    model: "ZNXDD01LM",
    vendor: "Aqara",
    description: "Ceiling light L1-350",
    extend: [lumiLight({colorTemp: true, powerOutageMemory: "switch"}), lumiZigbeeOTA()],
    exposes: [
        e
            .composite("level_config", "level_config", ea.ALL)
            .withFeature(
                e
                    .numeric("on_level", ea.ALL)
                    .withValueMin(1)
                    .withValueMax(254)
                    .withPreset("previous", 255, "Use previous value")
                    .withDescription("Specifies the level that shall be applied, when an on/toggle command causes the light to turn on."),
            ),
    ],
};
```

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -170,6 +170,18 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Aqara",
         description: "Ceiling light L1-350",
         extend: [lumiLight({colorTemp: true, powerOutageMemory: "switch"}), lumiZigbeeOTA()],
+        exposes: [
+            e
+                .composite("level_config", "level_config", ea.ALL)
+                .withFeature(
+                    e
+                        .numeric("on_level", ea.ALL)
+                        .withValueMin(1)
+                        .withValueMax(254)
+                        .withPreset("previous", 255, "Use previous value")
+                        .withDescription("Specifies the level that shall be applied, when an on/toggle command causes the light to turn on."),
+                ),
+        ],
     },
     {
         zigbeeModel: ["lumi.light.cwac02", "lumi.light.acn014"],


src/devices/lumi.ts
```diff
@@ -170,6 +170,18 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Aqara",
         description: "Ceiling light L1-350",
         extend: [lumiLight({colorTemp: true, powerOutageMemory: "switch"}), lumiZigbeeOTA()],
```
Koenkk: Could you add this as an option to `lumiLight` instead? It calls the `modernExtend.light` [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/c8957d1155fcb059c9519c7b538bf1548e0ab592/src/lib/lumi.ts#L1436) which allow to add this by passing `levelConfig`

kevintong116: I'm sorry that I have little to no knowledge about coding. I can only created this Pull requrest with reference to other devices. Any further changes might need contribution from others.

Koenkk: np, fixed!

Koenkk:
Thanks!

PR #9468, opened 2025-06-06
feat: Support Shelly WiFi Setup Cluster

SvenKirschbaum:
**NOTE: This PR requires accompanying changes in zigbee-herdsman, contained in [1418](https://github.com/Koenkk/zigbee-herdsman/pull/1418)**

This PR adds support for the Shelly WiFi Setup custom Cluster, making it possible to change the WiFi configuration of Shelly devices over zigbee.

The custom Cluster is documented here: https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/WiFiSetupCluster/
The documentation is however missing the fact, that any interaction with the cluster requires the APS Frame profileid to be set to 0xc001.

Please note, that I am quite new to the codebase and zigbee in general, if there are better ways to implement support for the custom cluster please leave a comment.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -1,4 +1,4 @@
-import {ZSpec, Zcl} from "zigbee-herdsman";
+import {Zcl, ZSpec} from "zigbee-herdsman";
 import * as m from "../lib/modernExtend";
 import type {DefinitionWithExtend, ModernExtend} from "../lib/types";
 import {assertString} from "../lib/utils";


src/devices/shelly.ts
```diff
@@ -1,33 +1,233 @@
...
+
+function addCustomClusterShellyRPC() {
+    return [
+        m.deviceAddCustomCluster("shellyRPCCluster", {
+            ID: 0xfc01,
+            manufacturerCode: Zcl.ManufacturerCode.SHELLY,
+            attributes: {
+                Data: {ID: 0x0000, type: Zcl.DataType.CHAR_STR},
+                TxCtl: {ID: 0x0001, type: Zcl.DataType.UINT32},
+                RxCtl: {ID: 0x0002, type: Zcl.DataType.UINT32},
```
```suggestion
                rxCtl: {ID: 0x0002, type: Zcl.DataType.UINT32},
```

(and change the casing of the other too)

src/devices/shelly.ts
```diff
@@ -1,33 +1,233 @@
+import {Zcl} from "zigbee-herdsman";
 import * as m from "../lib/modernExtend";
 import type {DefinitionWithExtend} from "../lib/types";
+import {assertString} from "../lib/utils";
+
```
Koenkk, line 5: Instead of the two functions, define a `const modernExtendLocal` here like e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/6c45dbce612a537b040ff327a0b4d3a1c647e46a/src/lib/tuya.ts#L2100 and use it in the definitions like `modernExtendLocal. shellyRPCCluster()`


And since all devices seems to support both clusters maybe combine into a single `shellyCustomClusters` modern extend?

src/devices/shelly.ts
SvenKirschbaum: Has been changed

src/devices/shelly.ts
SvenKirschbaum: I have changed the structure to now use a local variable, and merged the cluster definitions into one.

I have however stuck to returning an array of `ModernExtend`, using a spread operator to apply it to the devices. As far as my understanding goes, merging the ~13 extends into a single one would require manually merging the individual attributes, which would IMO make the code very hard to read and maintain. Please correct me if i missunderstood something here. Would it maybe make sense to implement a centralized "mergeModernExtend" function for such a usecase?

Koenkk: This is fine, thanks!

jfmlima:
@SvenKirschbaum is there any pending changes to be able to move this forward? Let me know if I can help 👍

shaysudm:
Is there any progress with this feature?

Koenkk:
@SvenKirschbaum could you fix the ci build? Then this can be merged.

src/devices/shelly.ts
```diff
@@ -1,34 +1,229 @@
...
+                    nameServer: {ID: 0x000a, type: Zcl.DataType.CHAR_STR},
+                },
+                commands: {},
+                commandsResponse: {},
+            }),
+            m.text({
+                cluster: "shellyWiFiSetupCluster",
+                access: "STATE_GET",
+                description: "Wifi Status",
+                name: "Wifi_Status",
```
Koenkk, line 43:
```suggestion
                name: "wifi_status",
```

And make sure to make all other `name` `snake_case`.

jactmp:
Hi, it would be great to merge this patch finally.  Would be great to have possibility to activate WiFi AP at least. 
Still not too experienced with z2m, but managed to fix the problem. I modified `shelly.ts`:

1/ inteface definition 
```
interface ShellyWiFi {
        attributes: {
                status: number;
                ip: number;
                action: number;
                dhcp: number;
                enable: number;
                ssid: number;
                password: number;
                staticIP: number;
                netMask: number;
                gateway: number;
                nameServer: number;
        };
        commands: {};
        commandResponses: never;
}
```

2/ extensions

`m.text<"shellyWiFiSetupCluster", ShellyWiFi>({`

It seems to work properly now. Does it even makes sense?

jactmp:
I rebuilt shelly.ts with custom logic on top of this PR. No idea whether it's OK - quite new to Z2M (and JS/TS). It would be great to have critical feedback on my input before maintainer decides to merge it into dev branch (if ever).

Koenkk:
Let's combine it into 1 pr. I cannot provide help with the Shelly specific things, do you want other Shelly users to test this?

jfmlima:
@jactmp @Koenkk when the combined PR is ready, I can review and test, let me know.

jactmp:
Sure, new PR can be created, if anybody needs it. But making the code based on my input ready for production will take some (lots of?) time anyway (seems like I'm the only person interested in contributing to this code).

Why not cherry pick fixes (first 2 commits) to this PR and merge it and close my PR for now (or leave open)? I assume that 95% of Shelly Gen4 users need access to custom clusters just to re-enable WiFi, e.g. to reconfigure input/output or update firmware. That was my problem, actually. 

The only problem I noticed is `password` (`m.text`) and `action` (`m.enumLookup`) write-only attributes being exposed in MQTT messages. I'm too panic guy to make any changes in `lib` code (as suggested) as it's used by other converters (sorry), so I created custom handler instead (and then expanded it cause I like coding). Otherwise my input is just nice-to-have.

Koenkk:
Im also fine with that, please also include the lib changes (I will review it)

jactmp:
I looked at code in `lib` more deeply. Seems like problem with write-only attributes is not trivial? I opened #10021. Maybe it's better to cover all cases? Those few guys who wait for custom cluster in Shelly Gen4 will manage to wait a bit longer (or will compile custom `zigbee-herdsmas-converter` with this or my PR applied).

PR #9470, opened 2025-06-06
Aqara S04D - Fix correct attribute

jamesonuk:
Managed to update wrong attribute in https://github.com/Koenkk/zigbee-herdsman-converters/pull/9462


Koenkk:
Thanks!

PR #9473, opened 2025-06-07
feat: support SR-ZG9032A-MW SR-ZG9098A-Win SR-ZG9098A-Light

niracler:


Koenkk:
Thanks!

wolfmar80:
We have an issue with the SR-ZG9098A-Win In Home Assistant:
State & Position is not updated. Entity stays in state "unknown".

We reset and calibrated the device multiple times.
- We can control the device (up/Down/stop)
- We can set a position (move to 50%)

BUT:
- The device state remains "Unknown"
- When controlling (up/down) the position does not update
 

CleanShot 2025-07-08 at 14 40 45@2x

niracler:
@wolfmar80 

Thanks for the feedback. I’ll look into it in the next couple of days and get back to you.

niracler:
> We have an issue with the SR-ZG9098A-Win In Home Assistant: State & Position is not updated. Entity stays in state "unknown".

I have already fix this issue by https://github.com/Koenkk/zigbee-herdsman-converters/pull/9609#issuecomment-3062229485

PR #9480, opened 2025-06-09
feat(add): Added new device support for soanalerm 

snksos:
New Device Support.


src/devices/soanalarm.ts
```diff
@@ -0,0 +1,441 @@
+import * as exposes from "../lib/exposes";
+import * as tuya from "../lib/tuya";
+import type {Definition, DefinitionWithExtend, Fz, KeyValue, KeyValueAny, Publish, Tz, Zh} from "../lib/types";
+
+const e = exposes.presets;
+const ea = exposes.access;
+
+const fzLocal = {
+    temperature_unit_handler: {
```
Koenkk: Could you refactor these converters to `tuyaDatapoints`?

src/devices/soanalarm.ts
```diff
@@ -0,0 +1,441 @@
...
+                return {temperature_f: state.temperature_f};
+            }
+
+            return {};
+        },
+    },
+};
+
+const tzLocal = {
+    temperature_unit: {
```
Koenkk: This can also be refactored to `tuyaDatapoints`.

src/devices/soanalarm.ts
snksos: Yes, like this
```
...
const result = tuya.fz.datapoints.convert(model, msg, publish, options, meta) || {};
...
```

Koenkk: No I mean deleting this converter completely and adding it to `tuyaDatapoints`

snksos: Only the device model SNT858Z is retained, and the others will be processed later when there is time. Thanks

src/devices/soanalarm.ts
snksos: Only the device model SNT858Z is retained, and the others will be processed later when there is time. Thanks

Koenkk:
Thanks!

PR #9481, opened 2025-06-09
fix(detect): Detect _TZ3210_dwzfzfjc as Moes ZB-LZD10-RCW

4val0v:
Incorrect identification:
https://github.com/Koenkk/zigbee2mqtt/issues/21424


Koenkk:
Thanks!

PR #9483, opened 2025-06-10
New model for Tuya TS0603 - Garage door opener

tavicu:
I have bought a new gate opener (QS-Zigbee-S10-C03) from AliExpress which is the same thing as previous ones but reports a different model.

https://www.aliexpress.com/item/1005009103445839.html

Please let me know if something else should be added :)

Thank you

image


Koenkk:
Thanks!

PR #9489, opened 2025-06-11
Sprut.device Bed.box Implementation

eXist-FraGGer:
Added support head and foot controls, head and foot massage intensity, massage duration, get current head and foot positions, light control.

Have tested, works well: 
image


Koenkk:
Thanks!

PR #9496, opened 2025-06-12
add BTicino F40T125A same as Legrand 412175

OliviezGit:


src/devices/legrand.ts
```diff
@@ -413,10 +413,11 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        zigbeeModel: [" 3 Phases power consumption module\u0000\u0000", " DIN 3Ph power consumption module", "3ph Smart shedder module"],
+        zigbeeModel: [" 3 Phases power consumption module\u0000\u0000"],
```
Koenkk, line 416: this will remove support for these 2 modelIds, I don't think that is the intention?

PR #9506, opened 2025-06-16
Update tuya.ts

vecapetr:
vecapetr _TZE284_tgeqdjgk

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -18050,4 +18050,83 @@ export const definitions: DefinitionWithExtend[] = [
         },
         configure: tuya.configureMagicPacket,
     },
+    {
+        const tuya = require('zigbee-herdsman-converters/lib/tuya');
+const exposes = require('zigbee-herdsman-converters/lib/exposes');
+const e = exposes.presets;
+const ea = exposes.access;
+
+// converter 0–1000 (Tuya) <-> 2700–6500 K
+const colorTempKelvinConverter = {
+    from: (val) => {
+        const kelvin = 2700 + (val / 1000) * (6500 - 2700);
+        return Math.round(kelvin);
+    },
+    to: (val) => {
+        const raw = ((val - 2700) / (6500 - 2700)) * 1000;
+        return Math.round(Math.max(0, Math.min(1000, raw)));
+    },
+};
+
+module.exports = {
+    fingerprint: [
+        {
+            modelID: 'TS0601',
+            manufacturerName: '_TZE284_tgeqdjgk',
+        },
+    ],
+    model: 'TS0601_knob_dimmer_switch',
+    vendor: 'TuYa',
+    description: 'Dimmer knob with two lights and adjustment mode (brightness/color_temp)',
+
+    fromZigbee: [
+        tuya.fz.datapoints,
+    ],
+
+    toZigbee: [
+        {
+            key: ['adjustment_mode'],
+            convertSet: async (entity, key, value, meta) => {
+                const mode = value === 'brightness' ? 0 : 1;
+                await tuya.sendDataPointEnum(entity, 105, mode);
+                return { state: { adjustment_mode: value } };
+            },
+        },
+        tuya.tz.datapoints,
+    ],
+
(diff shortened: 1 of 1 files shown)


vecapetr (commented):
begin

src/devices/tuya.ts
```diff
@@ -18049,5 +18049,82 @@ export const definitions: DefinitionWithExtend[] = [
...
+                .withValueMax(6500)
+                .withDescription('Color temperature in Kelvin (converted from 0–1000 Tuya raw)'),
+            e.enum('adjustment_mode', ea.STATE_SET, ['brightness', 'color_temp']).withDescription('Brightness or color temperature adjustment'),
+        ],
+
+        meta: {
+            tuyaDatapoints: [
+                [102, 'state', tuya.valueConverter.onOff],
+                [103, 'brightness', tuya.valueConverter.divideBy10],
+                [107, 'color_temp', colorTempKelvinConverter],
```
Koenkk: If I understand correctly this device has 2 lights, is it possible to control the color temp and brightness of each light individually? (or can you only turn them on/off individually?)

vecapetr:
state on/off start only temp and brightness
switch C/W switching between volor_temp or brightenss
two lights individualy switching on/off (color and brightness is only one for both)
Petr

Koenkk:
Thanks!

Make sure that:
- The filename is `TS0601_knob_dimmer_switch.png`
- The size is 512x512

vecapetr:
done, but just one small mistake is that the brightness value was 0-100, but the value in the max field is probably 1000, and some return run from HA to device, i am make first time, i am promise that will good student...

Koenkk:
@vecapetr so if I understand correctly you cannot set it to the max brightness now? Can you try with:

```js
[103, "brightness", tuya.valueConverterBasic.scale(0, 254, 0, 1000)],
```

vecapetr:
brightness has datapoint value 0-1000 and display show 0-100% for real must be devide 10
color temp has datapoint value 0-1000 and display show 2700-6500K
I am set 0-100 but max is 1000 (could be 100).

vecapetr:
Snímek obrazovky 2025-06-19 v 21 59 19

vecapetr:
> @vecapetr so if I understand correctly you cannot set it to the max brightness now? Can you try with:
> 
> ```js
> [103, "brightness", tuya.valueConverterBasic.scale(0, 254, 0, 1000)],
> ```

if understand me in short what mean numbers:
0-?
254-?
0-?
1000-?

i will in future on first strike be succesfull :-)

Koenkk:
For Zigbee bulbs the brightness we use is always a number between 0 (lowest) and 254 (highest). For Tuya it is different (max is 1000 here), so we map 254 to 1000 in this case. So if you set it to 254 it will send 1000 to the device.

vecapetr:
> For Zigbee bulbs the brightness we use is always a number between 0 (lowest) and 254 (highest). For Tuya it is different (max is 1000 here), so we map 254 to 1000 in this case. So if you set it to 254 it will send 1000 to the device.

understand. 
how is possible update converter? new pull request?
because 
1. color temp is not come to bulb entity
2. state l1 and l2 from HA dont come back to device

Koenkk:
What do you mean with "do not come" here? Can you explain a bit more what the problem is.

vecapetr:
> What do you mean with "do not come" here? Can you explain a bit more what the problem is.

in Z2M edge devices-exposes looks like function. 
If put dashboard:
image
1. complete state function both direction from Dashboard to device and from device to dashboard
2. adjustment function both direction from Dashboard to device and from device to dashboard
3. brightness function both direction from Dashboard to device and from device to dashboard
4. color_temp dont return back to dashboard
5. L1 or L2 from device to dashboard function, if switch on in Dashboard order dont come to device

Koenkk:
> color_temp dont return back to dashboard

Does the color temp control itself work?

> L1 or L2 from device to dashboard function, if switch on in Dashboard order dont come to device

did this work with your initial version of the converter?

PR #9524, opened 2025-06-20
Adds Silicon Labs RB-ElectricityDsp-061-3 smart plug

s-gordon:


Koenkk:
Please also add to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts and make the export an array (look at the other files for an example)

s-gordon:
Ah, thanks for pointing that out. Let me know if I've missed anything else.

Koenkk:
Thanks!

PR #9528, opened 2025-06-22
new device Manhot MB60L-ZG-ZT-TY Smart blinds motor

00svd00:
Everything works, but not perfectly sure about battery.


Koenkk:
Thanks!

PR #9536, opened 2025-06-23
Tuya LY-108 Cover. Add illuminance sensor readings

00svd00:
Seems that there is different models with same fingerprint on market. My variant have illuminance sensor


Koenkk:
Thanks!

PR #9545, opened 2025-06-25
SNZB-02LD updates once per hour, and with only 1 full degree change

wizu:
I've updated it to behave more like normal SNZB-02.


Koenkk:
This is the case for every temp sensor and would like to keep it as is for consistency purposes. If you want to change the reporting interval locally, you can change it through the z2m frontend -> device -> reporting tab.

PR #9549, opened 2025-06-26
tuya: extend TS0726 4 gang switch to support COSWELL's variant _TZ3002_phu8ygaw

imammedo:
reuse generic definition, and just add _TZ3002_phu8ygaw vendor id


Koenkk:
Thanks!

PR #9550, opened 2025-06-26
Add _TZE284_awepdiwi to supported soil sensors

palkeo:
Device works after this change.
It's https://fr.aliexpress.com/item/1005008785131043.html


Koenkk:
Thanks!

PR #9560, opened 2025-06-29
Mods for LoraTap SC500ZBv2 and for Nous B4Z (with common reusability)

weichenb:
I implemented additional features for LoraTap SC500ZBv2 and for Nous B4Z.
Generally the changes can be activated easily for other devices. Since I have others not at hand, I restricted changes no.1, no.3 and no.4 only to these 2 devices so far.
Details were discussed here: https://github.com/Koenkk/zigbee2mqtt/issues/25996

**Changes:**
1.  Dual direction calibration
For devices, which differentiate calibration for separate directions, configuration shows separate set for "close" and "open" directions.

2. Calibration quick and extended
On one hand calibration can be performed by starting "Calibration"-mode, moving shutter from one end to the other and ending "Calibration"-mode. Here resulting time is transferred automatically into "Calibration time (to open/to close)" setting(s). This was the standard procedure so far.
On the other hand from now on time(s) with 1 decimal precision (e.g. 35,1 "s") can be directly entered into field. Values are directly transferred into device registers.

3. Configuration for physically connected electrical switch
With the modification you can easily select the button mode that is required for your scenario.
There are typcially 2 kind of shutter switches (both have for each moving direction either a single button or at least a separate position):
- Button-switch: Such kind of switch only transmits a signal as long as button is actually pressed. After the button-push the button returns into neutral position.
However there are 2 ways to press the button, either with a short or longer (typically >1s) push. In dependency of the push the modul generates a different behavior.
- Flip-switch: For such kind of switch the button remains after the push in its position and does not return into neutral position.
- Sync-switch: On Nous B4Z switch this mode allows a more precise angle control of jalousies. The remaining behavior I would rate as identical with "Flip-switch"-mode.
LoraTap SC500ZBv2 seems not to show a difference compared to "Button-switch"-mode.
- Button2-switch: On LoraTap SC500ZBv2 a long push seems to show a slightly different behavior.
Nous B4Z seems not to show a difference compared to "Button-switch"-mode.
I contacted both manufactures to provide a bit more information about the supported features. I am going to add more information later in the manual.

4. Backlight and Indicator LED options
Both devices do not support that feature. Therefore I deactivated the option for both models.
5. Code homogenization and simplification in devices/tuya.ts for generic model "TS130F"

Manual and picture for LoraTap SC500ZBv2 follow separately.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2882,9 +2882,7 @@ export const definitions: DefinitionWithExtend[] = [
                 exps.push(tuya.exposes.switchTypeCurtain());
             } else if (["_TZ3000_yruungrl"].includes(device?.manufacturerName)) {
                 exps.push(
-                    e
-                        .enum("switch_type_curtain", ea.ALL, ["flip-switch", "sync-switch", "button-switch"])
-                        .withDescription("External switch type"),
+                    e.enum("switch_type_curtain", ea.ALL, ["flip-switch", "sync-switch", "button-switch"]).withDescription("External switch type"),
                 );
             }
             return exps;


weichenb:
some corrections applied - tests on docker version were successful

src/converters/toZigbee.ts
```diff
@@ -4008,33 +4008,46 @@ export const ts0216_alarm: Tz.Converter = {
...
-            case "calibration_down":
-                await meta.device.getEndpoint(2).write("closuresWindowCovering", {tuyaCalibration: calibration});
-                break;
-        }
-        return {state: {[key]: value}};
+    key: ["calibration", "calibration_to_open", "calibration_to_close", "calibration_time", "calibration_time_to_open", "calibration_time_to_close"],
+    convertSet: async (entity, key, value, meta) => {
+	if (key.startsWith( "calibration_time")) {
+		utils.assertNumber(value, key);
+		const calibration_time = parseInt(value*10);
```
Koenkk:
```suggestion
		const calibration_time = value*10;
```

src/lib/tuya.ts
```diff
@@ -413,6 +413,7 @@ const tuyaExposes = {
             .withUnit("%")
             .withDescription(`Instantaneous measured power factor (phase ${phase.toUpperCase()})`),
     switchType: () => e.enum("switch_type", ea.ALL, ["toggle", "state", "momentary"]).withDescription("Type of the switch"),
+    switchTypeCurtain: () => e.enum("switch_type_curtain", ea.ALL, ["flip-switch", "sync-switch", "button-switch","button2-switch"]).withDescription("what external switch have you applied?"),
```
Koenkk:
```suggestion
    switchTypeCurtain: () => e.enum("switch_type_curtain", ea.ALL, ["flip_switch", "sync_switch", "button_switch","button2_switch"]).withDescription("External switch type"),
```

src/lib/tuya.ts
```diff
@@ -576,6 +577,7 @@ export const valueConverter = {
     powerOnBehavior: valueConverterBasic.lookup({off: 0, on: 1, previous: 2}),
     powerOnBehaviorEnum: valueConverterBasic.lookup({off: new Enum(0), on: new Enum(1), previous: new Enum(2)}),
     switchType: valueConverterBasic.lookup({momentary: new Enum(0), toggle: new Enum(1), state: new Enum(2)}),
+    switchTypeCurtain: exports.valueConverterBasic.lookup({"flip-switch": new Enum(0), "sync-switch": new Enum(1), "button-switch": new Enum(2), "button2-switch":new Enum(3)}),
```
Koenkk:
```suggestion
    switchTypeCurtain: exports.valueConverterBasic.lookup({"flip_switch": new Enum(0), "sync_switch": new Enum(1), "button_switch": new Enum(2), "button2_switch":new Enum(3)}),
```

src/lib/tuya.ts
```diff
@@ -1563,6 +1565,17 @@ const tuyaTz = {
             await entity.read("manuSpecificTuya_3", ["switchType"]);
         },
     } satisfies Tz.Converter,
+    switch_type_curtain: {
+        key: ["switch_type_curtain"],
+        convertSet: async (entity, key, value, meta) => {
+            const switchType = utils.getFromLookup(value, { "flip-switch": 0, "sync-switch": 1, "button-switch": 2, "button2-switch": 3 });
```
Koenkk:
```suggestion
            const switchType = utils.getFromLookup(value, { "flip_switch": 0, "sync_switch": 1, "button_switch": 2, "button2_switch": 3 });
```

src/lib/tuya.ts
```diff
@@ -1833,6 +1846,16 @@ const tuyaFz = {
                 return {switch_type: lookup[msg.data.switchType]};
             }
         },
+    } satisfies Fz.Converter,
+	switch_type_curtain: {
+        cluster: "manuSpecificTuya_3",
+        type: ["attributeReport", "readResponse"],
+        convert: (model, msg, publish, options, meta) => {
+            if (msg.data.switchType !== undefined) {
+                const lookup = { 0: "flip-switch", 1: "sync-switch", 2: "button-switch", 3: "button2-switch" };
```
Koenkk:
```suggestion
                const lookup = { 0: "flip_switch", 1: "sync_switch", 2: "button_switch", 3: "button2_switch" };
```

src/lib/tuya.ts
```diff
@@ -1833,6 +1846,16 @@ const tuyaFz = {
...
+                return { switch_type_curtain: lookup[msg.data.switchType] };
```
Koenkk:
```suggestion
                utils.assertNumber(msg.data.switchType);
                return { switch_type_curtain: lookup[msg.data.switchType] };
```

src/lib/tuya.ts
```diff
@@ -2514,7 +2537,11 @@ const tuyaModernExtend = {
             toZigbee.push(tuyaTz.switch_type);
             exposes.push(tuyaExposes.switchType());
         }
-
+		if (args.switchTypeCurtain) {
```
Koenkk: This has to be added to the args

src/converters/toZigbee.ts
weichenb: Hi Koen,
Since target register is integer, I assured that resulting const is also integer. In case anybody would enter 23,45, result is 234,5.  Effectively your proposal works as well.

weichenb:
Hi Koen,
I am not that guru on github. I commited the first proposal. After that all others show "outdated". Did I do anything wrong?
A brief advice would be welcome

src/lib/tuya.ts
weichenb: fine for me - status shows outdate, I am not able to commit.

src/lib/tuya.ts
weichenb: fine for me - status shows outdate, I am not able to commit.

src/lib/tuya.ts
weichenb: fine for me - status shows outdate, I am not able to commit.

src/lib/tuya.ts
weichenb: Personally I think, since integer definition is defined in the lookups of that lib, the test on type number might be superfluous and only producing overhead in execution.
I replicated code from lines 1840-1859. In that part command "utils.assertNumber(msg.data.switchType);" is also not there. Maybe you should add it there as well?

status shows outdated, I am not able to commit.

src/lib/tuya.ts
weichenb: In this section args for tuyaLight are defined. Therefore this change should not appear in this section right? It is only relevant for "tuyaOnOff"
In "tuyaOnOff" I realized that in line 2463/2494 "switchTypeCurtain?: boolean;" must be still added.
grafik
Interesting that I haven't got that error message on my platform ;-)

src/lib/tuya.ts
weichenb: fine for me - status shows outdated, I am not able to commit.

src/lib/tuya.ts
weichenb: Changed my opinion. Does the underscore look nice in the GUI? I think, not really?
Initially I thought, underscore would be translated into whitespace as it is for "exposes".
grafik

src/lib/tuya.ts
weichenb: Changed my opinion. Does the underscore look nice in the GUI? I think, not really?
Initially I thought, underscore would be translated into whitespace as it is for "exposes".
grafik

src/lib/tuya.ts
weichenb: Changed my opinion. Does the underscore look nice in the GUI? I think, not really?
Initially I thought, underscore would be translated into whitespace as it is for "exposes".
grafik

src/lib/tuya.ts
weichenb: Changed my opinion. Does the underscore look nice in the GUI? I think, not really?
Initially I thought, underscore would be translated into whitespace as it is for "exposes".
grafik

src/lib/tuya.ts
weichenb: Update: I included update for both routines in latest commit from my side.

src/lib/tuya.ts
weichenb: Update:
"switchTypeCurtain?: boolean" addition included in latest commit

Koenkk:
Thanks!

Make sure that:
- The filename is `SC500ZB-v4.png`
- The size is 512x512

weichenb:
Thanks a lot Koen!

Yes, for sure I will upload a picture and update manuals for both devices latest next weekend.

PR #9566, opened 2025-06-30
Update moes.ts

cumal:
Added support for this [valve](https://es.aliexpress.com/item/1005008737685262.html)

By default it uses the Tuya vendor, but it doesn't works. Tested with the Moes vendor and works fine.


Koenkk:
thanks!

PR #9568, opened 2025-07-01
feat: update Sunricher 8-button switch to support multi-endpoint commands for on/off and level control

niracler:


Koenkk:
Thanks!

Issue #9569, opened 2025-07-01
Update Converters for Novato ZPV-01


NTV20244:
Hello

The Novato ZPV-01 device already exists in the Zigbee2MQTT database. However, a newer hardware or firmware version of the product has been released which introduces support for battery level reporting via Tuya datapoint 101.

I would like to request that support for this new version be added to the existing converter because manufacturer id also changed.  I've confirmed that the battery percentage is correctly reported as an integer (e.g., 88) and can be handled using tuya.valueConverter.raw.

This update will allow users of the updated ZPV-01 hardware to monitor battery levels properly within Zigbee2MQTT.

https://www.zigbee2mqtt.io/devices/ZPV-01.html
https://novato.com.tr/urun/pilli-akilli-vana/

`const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
const tz = require('zigbee-herdsman-converters/converters/toZigbee');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const reporting = require('zigbee-herdsman-converters/lib/reporting');
const modernExtend = require('zigbee-herdsman-converters/lib/modernExtend');
const e = exposes.presets;
const ea = exposes.access;
const tuya = require('zigbee-herdsman-converters/lib/tuya');

const definition = {
    fingerprint: [
        {
            modelID: 'TS0601',
            manufacturerName: '_TZE284_zm8zpwas',
        },
    ],
    model: 'ZPV-01',
    vendor: 'Novato',
    description: 'Battery Powered Smart Valve',
    fromZigbee: [tuya.fz.datapoints],
    toZigbee: [tuya.tz.datapoints],
    onEvent: tuya.onEventSetTime,
    configure: tuya.configureMagicPacket,
    exposes: [
        e.switch().setAccess('state', ea.STATE_SET),
        e.enum('valve_state', ea.STATE, ['Kapalı', 'Bilinmiyor', 'Açık']).withDescription('Valf Durumu'),
        e.battery(),  // Artık doğru çalışacak
    ],
    meta: {
        tuyaDatapoints: [
            [1, 'state', tuya.valueConverter.onOff],
            [8, 'valve_state', tuya.valueConverterBasic.lookup({
                'Bilinmiyor': tuya.enum(0), 'Açık': tuya.enum(1), 'Kapalı': tuya.enum(2),
            })],
            [101, 'battery', tuya.valueConverter.raw],  // Doğru çalışan pil okuma
        ],
    },
};

[...]

PR #9577, opened 2025-07-03
Update tuya.ts Novato ZPV-01

NTV20244:
Added battery level
https://github.com/Koenkk/zigbee-herdsman-converters/issues/9569


Koenkk:
Thanks!

PR #9578, opened 2025-07-03
fix: TS0505B_1 transition fixes

Ten0:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/27894
Helps https://github.com/Koenkk/zigbee2mqtt/issues/27498

This light has two issues:
1. If passing transition = 0, it will behave as if transition = 1.
2. If turning off with a transition, and turning on during the transition, it will turn off at the end of the first transition timer, despite order to turn on

Workaround for issue 1:
- If we're trying to turn off, we can just remove the transition: 0 key and it will just use `on_off`, which is instant
- Otherwise best we can do is set the transition to 0.1 seconds, which will be as instant as it gets (That is the same thing as is done for TS0505B_2)
This PR currently only does this for the turn_on/turn_off case, but not when just changing color.

Workaround for issue 2:
- Get the current state of the light after transition time + 0.1s.

This won't get the light to the "desired" state, but at least it will make us aware that it's off, allowing user apps to turn it on again if needed.
This could probably be improved by actually turning it on again if necessary, but it's unclear to me if I can track "necessary" with certainty, or how I could trigger another change in a way that would make z2m aware of the associated state changes.
Right now my life will be easy enough with just knowing the actual state but with guidance I could maybe adapt this to make it re-apply actually-wanted state.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -647,7 +647,7 @@ const tzLocal = {
         ...tz.light_onoff_brightness,
         convertSet: async (entity, key, value, meta) => {
             // This light has two issues:
-            // 1. If passing transition = 0, it will behave as if transition = 1.
+            // 1. If passing transition = 0, it will behave as if transition = 1s.
             // 2. If turning off with a transition, and turning on during the transition, it will turn off
             //    at the end of the first transition timer, despite order to turn on
 
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -658,15 +658,15 @@ const tzLocal = {
             if (transitionSeconds === 0) {
                 const {message} = meta;
                 const wantedState = message.state != null ? (typeof message.state === "string" ? message.state.toLowerCase() : null) : undefined;
-                newMeta = {...meta, message: {...message}}; // Clone meta.message to avoid modifying the original
+                newMeta = {...meta}; // Clone meta to avoid modifying the original
                 if (wantedState === "off") {
                     // Erase transition because that way we get actual instant turn off
-                    newMeta.message.transition = null;
+                    newMeta.message = {state: "OFF"};
                 } else {
                     // Best we can do is set the transition to 0.1 seconds
                     // That is the same thing as is done for TS0505B_2
                     transitionSeconds = 0.1;
-                    newMeta.message.transition = transitionSeconds; // Will get re-parsed by original light_onoff_brightness
+                    newMeta.message = {...message, transition: transitionSeconds}; // Will get re-parsed by original light_onoff_brightness
                 }
             }
 
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -643,7 +643,7 @@ const tzLocal = {
         },
     } satisfies Tz.Converter,
     // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
-    TS0505B_1_transitionFixes: {
+    TS0505B_1_transitionFixesOnOffBrightness: {
         ...tz.light_onoff_brightness,
         convertSet: async (entity, key, value, meta) => {
             // This light has two issues:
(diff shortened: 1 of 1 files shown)


Ten0:
I'm currently running z2m in the HA addon.
<s>It's not super clear to me how I can test this easily.
Do you have a guide somewhere of how I could do that?</s>

EDIT: Actually I built a script that enables me to patch this on the fly as the addon starts:
<details>
<summary>
bash script to run in SSH addon
</summary>

```bash
#!/usr/bin/env bash

set -euo pipefail

# This script waits for the zigbee2mqtt container to be up and then patches
# the zigbee-herdsman-converters with custom device definitions.

ZHC_OVERRIDES_PATH="/config/zigbee2mqtt/zhc_overrides"
ZHC_DEST_PATH="/app/node_modules/.pnpm/zigbee-herdsman-converters@*/node_modules/zigbee-herdsman-converters/dist/devices"

echo "Waiting for zigbee2mqtt container to become available..."

while true; do
    # We use `|| true` to prevent the script from exiting if grep finds nothing.
    CONTAINER_ID=$(docker ps | grep zigbee2mqtt | cut -d' ' -f 1 || true)

    if [ -n "$CONTAINER_ID" ]; then
        echo "Found zigbee2mqtt container with ID: $CONTAINER_ID"
        echo "Applying custom device patches..."

        # Using a shell inside the container to handle the globbing (*) correctly.
        # The command will copy all files from the overrides path to the destination.
        # Also check the exit code of the docker exec command
        if docker exec "$CONTAINER_ID" sh -c "cp ${ZHC_OVERRIDES_PATH}/* ${ZHC_DEST_PATH}/"; then
            echo "Successfully patched zigbee-herdsman-converters."
            break
        else
            # It might fail if the destination directory doesn't exist yet,
            # so we'll retry.
            echo "Failed to patch. The container might not be fully ready. Retrying in 1 second..."
            sleep 1
        fi
    else
        # Wait for a very short time before checking again.
        sleep 0.1
    fi
done

echo "Patching script finished."
```

</details>

I need more testing now...

Ten0:
Ok that works now.

src/devices/tuya.ts
```diff
@@ -642,6 +642,54 @@ const tzLocal = {
...
+            let newMeta = meta;
+            if (transitionSeconds === 0) {
+                const {message} = meta;
+                const wantedState = message.state != null ? (typeof message.state === "string" ? message.state.toLowerCase() : null) : undefined;
+                newMeta = {...meta, message: {...message}}; // Clone meta.message to avoid modifying the original
+                if (wantedState === "off") {
+                    // Erase transition because that way we get actual instant turn off
+                    newMeta.message.transition = null;
+                } else {
+                    // Best we can do is set the transition to 0.1 seconds
```
Koenkk, line 666: I don't understand this branch, I also don't see that a transition of 0.1 seconds is set here?

Ten0: Whoops :(
I lost the `transitionSeconds = 0.1` line somehow in my iterations...
Thanks for catching it!
e650a9845394381e2c7b9ef463cdb1b94757bec0

Basically if we just want to turn off then it's instant if we don't set transition, whereas if we set transition = 0 it takes 1s to turn off. I assume if users explicitly call with transition = 0 they mean "please instantly turn off" so this is the call that will lead to that behavior.

For all other cases we set transition = 0.1 which is almost instant.

Koenkk:
thanks!

ThePegass:
Hello, I found the fix was included in zigbee2mqtt release 2.6.0. I have 2.6.2. When turn the light on, is seems to be immediate, but there is still 1s transition when turn off. Can I fix it somehow? PS: I have just started with Home Assistant and Zigbee2mqttt. Thanks.

Ten0:
I cannot reproduce.
I have 2.6.2-1 and the following instantly turns off the lamp:

image

same thing happens if I uncheck "transition"

Are you sure your lamp is `TS0505B_1`?

ThePegass:
image

Haven't tried transition checkbox in the Light actions. I meant the setting in z2m. When it was 0, there was still transition. I set it to 0.1 and now there's almost no transition:
image

PR #9580, opened 2025-07-04
Update tuya.ts

vecapetr:
add withProperty for mapping state_l1 and l2


Koenkk:
Does it work when also dropping just the endpoint part?

vecapetr:
not, good. endpoint l1 and l2 cause off on dp102.

Koenkk:
Can you try after reverting your changes and with:

```js
        meta: {
            multiEndpoint: true, // <-- add this line
            tuyaDatapoints: [
```

PR #9582, opened 2025-07-05
new device of Tuya TS0049

alanmgomes:
Just adding a new model, same as the generic Tuya TS0049. This one is made by Nova Digital brand.


Koenkk:
Thanks!

PR #9586, opened 2025-07-06
add ledvance 625 panel

jasperart:
adding the following panel:
https://www.ledvance.de/de/tech/products/index?productId=4058075364561&category=190761&documentId=459


Koenkk:
Thanks!

PR #9587, opened 2025-07-06
Add Philips Hue Triangle E26 bulb (046677577490)

aelindeman:
Adds https://www.philips-hue.com/en-us/p/bulb-triangle-e26-smart-bulb/046677577490


Koenkk:
Thanks!

PR #9588, opened 2025-07-07
fix(bht002): set tempature steps for specific model

brvn0:
as reported in https://github.com/Koenkk/zigbee2mqtt/issues/26794 one device model (_TZE204_aoclfnxz) doesn't support decimal steps when setting the local_temperature_calibration

### DIFF
--- a/src/devices/moes.ts
+++ b/src/devices/moes.ts
@@ -235,7 +235,7 @@ export const definitions: DefinitionWithExtend[] = [
                     .climate()
                     .withSetpoint("current_heating_setpoint", 5, 45, heatingStepSize, ea.STATE_SET)
                     .withLocalTemperature(ea.STATE)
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

schrej-zz:
Hi,
I have the same issues with _TZE204_ltwbm23f

Koenkk:
@schrej-zz applied the same fix!

Changes will be available in the [dev branch](https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) in a few hours from now and in the next release which is every 1st of the month.

mVik85:
Is this already included in a release?
I would assume from the date, but the status is merged, and i'm not familiar with the workflow.
Anyway, if it is in the latest release, how the change this? My thermostat support 0.5 steps, but not possible in Z2M

Many thanks!

Issue #9589, opened 2025-07-07
Support for ZTH05 (TZE284_upagmta9)


pauvt:
I am new to this and I just bought new Tuya ZTH05 (https://es.aliexpress.com/item/1005007927629320.html).

This uses a slightly different fingerprint -> TZE284_upagmta9

![Image](https://github.com/user-attachments/assets/785f15f0-45d4-4b83-97df-9df1bbd134d9)

With GPT I created a external configurator which works:

````
// File: data/external_converters/ts0601_zth_lcd.js
const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
const tz = require('zigbee-herdsman-converters/converters/toZigbee');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const tuya = require('zigbee-herdsman-converters/lib/tuya');
const e = exposes.presets;

module.exports = {
    // ★ NEW: added _TZE284_upagmta9
    fingerprint: tuya.fingerprint('TS0601', [
        '_TZE200_yjjdcqsq', '_TZE200_9yapgbuv', '_TZE204_upagmta9',
        '_TZE200_cirvgep4', '_TZE200_upagmta9', '_TZE284_upagmta9',
    ]),
    model: 'TS0601_temperature_humidity_sensor_2',
    vendor: 'TuYa',
    description: 'LCD temperature & humidity sensor',
    fromZigbee: [tuya.fz.datapoints],
    toZigbee:   [tuya.tz.datapoints],
    onEvent:    tuya.onEvent({queryOnDeviceAnnounce: true}),
    exposes: [
        e.temperature(), e.humidity(),
        tuya.exposes.batteryState(),
        tuya.exposes.temperatureUnit(),
    ],
    meta: {                        // datapoint → attribute map
        tuyaDatapoints: [
            [1, 'temperature',      tuya.valueConverter.divideBy10],
            [2, 'humidity',         tuya.valueConverter.raw],
            [3, 'battery_state',    tuya.valueConverter.batteryState],
            [9, 'temperature_unit', tuya.valueConverter.temperatureUnitEnum],
        ],
    },
    whiteLabel: [
        tuya.whitelabel('TuYa', 'ZTH01',  'Temperature & humidity sensor', ['_TZE200_yjjdcqsq']),
        tuya.whitelabel('TuYa', 'ZTH02',  'Temperature & humidity sensor', ['_TZE200_9yapgbuv']),
[...]

PR #9590, opened 2025-07-07
feat: extend support for ZTH05 (_TZE284_upagmta9)

pauvt:
As reported in #9589, there is one new fingerprint for ZTH05 Tuya sensor.


Koenkk:
Thanks!

PR #9592, opened 2025-07-07
Support of new Tuya switches

pkhodak:
Adding support for one new 1 gang switch with backlight and mapping 3 switches to existing definitions: 

_TZ3000_qvmiyxuk - new 1-gang with backlight
_TZ3002_6ahhkwyh works as _TZ3002_ymv5vytn
_TZ3002_kq3kqwjt works as _TZ3000_noru9tix
_TZ3002_umdkr64x works as _TZ3002_phu8ygaw

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -17578,7 +17578,35 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint("TS0726", ["_TZ3000_ezqbvrqz", "_TZ3002_ymv5vytn"]),
+        fingerprint: tuya.fingerprint("TS0001", ["_TZ3000_qvmiyxuk"]),
+        model: "TS0001_1_gang_switch",
+        vendor: "Tuya",
+        description: "1 gang switch with backlight",
+
+        fromZigbee: [fz.ignore_basic_report],
+        extend: [
+            tuya.modernExtend.tuyaOnOff({
+                powerOnBehavior2: true,
+                backlightModeOffOn: true,
+                onOffCountdown: true,
+            }),
+        ],
+        configure: async (device, coordinatorEndpoint) => {
+            await tuya.configureMagicPacket(device, coordinatorEndpoint);
+            await reporting.bind(device.getEndpoint(1), coordinatorEndpoint, ["genOnOff"]);
+        },
+
+        meta: {
+            tuyaDatapoints: [
+                [1, "state", tuya.valueConverter.onOff], // Switch
+                [7, "countdown", tuya.valueConverter.countdown], // Countdown
+                [14, "power_on_behavior", tuya.valueConverter.powerOnBehavior], // Power on behavior
+                [16, "backlight_mode", tuya.valueConverter.onOff], // Backlight
+            ],
+        },
+    },
+    {
+        fingerprint: tuya.fingerprint("TS0726", ["_TZ3000_ezqbvrqz", "_TZ3002_ymv5vytn", "_TZ3002_6ahhkwyh"]),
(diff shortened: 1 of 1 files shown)


devkaiwang:
Hi @pkhodak 

I found that your are using different supporting code for the 1-gang type, it would miss the scene-action, LED indicator support to show the relay status. I wonder is the 1-gang type you have is a different category of the product?

I wonder if you could use the same code like 2/3/4-gang switches but with 1 button? I also received 1 gang samples here, and would test in the following days.

src/devices/tuya.ts
```diff
@@ -17578,7 +17578,35 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint("TS0726", ["_TZ3000_ezqbvrqz", "_TZ3002_ymv5vytn"]),
+        fingerprint: tuya.fingerprint("TS0001", ["_TZ3000_qvmiyxuk"]),
+        model: "TS0001_1_gang_switch",
+        vendor: "Tuya",
+        description: "1 gang switch with backlight",
+
+        fromZigbee: [fz.ignore_basic_report],
```
devkaiwang, line 17585: This supporting code is different from the 2, 3, 4 gang switches. Seems this code would miss the scene-action and relay LED indicator features. Same code with one button should be used if they are the same type of switches.

devkaiwang (changes requested):

src/devices/tuya.ts
pkhodak: The switches are from the same manufacturer and look consistent between 1-4 gang versions, however I've tested the 1-gang version and it doesn't appear that it has the scene-action support and relay LED indicator features. I don't mind adding it for consistency with other switches, however I'm not sure it'd be useful to anybody.

devkaiwang: Thanks for clarification! Then I'm OK with your current code.
I'd add the 1-gang version with scene and relay indicator after your PR got merged.

devkaiwang: I had experienced this before, the switch with scene action support has "智能情景开关" ("Smart Scene Switch") printed on the back of it,

while the other one without scene support is named "智能开关" ("Smart Switch"): https://github.com/Koenkk/zigbee-herdsman-converters/blob/4a905dc927897f70fcc6abc15b9b8f3939aac8f4/src/devices/tuya.ts#L17556-L17568.

src/devices/tuya.ts
```diff
@@ -17578,7 +17578,35 @@ export const definitions: DefinitionWithExtend[] = [
...
+                onOffCountdown: true,
+            }),
+        ],
+        configure: async (device, coordinatorEndpoint) => {
+            await tuya.configureMagicPacket(device, coordinatorEndpoint);
+            await reporting.bind(device.getEndpoint(1), coordinatorEndpoint, ["genOnOff"]);
+        },
+
+        meta: {
+            tuyaDatapoints: [
```
Koenkk: I think this can be removed (seems unused)

pkhodak: My apologies for the naive questions, I'm new to this project and not sure if I understand your comment. Are you sugesting that we don't need to specify tuyaDatapoints in this code? When I tried running custom extender it didn't work without datapoints giving the error the xx datapoint is not defined.

Koenkk: As far as I can see it is not used so please check again.

pkhodak: No problem, thank you for looking into it. I have removed it as suggested.

Koenkk:
Thanks!

Make sure that:
- The filename is `TS0001_1_gang_switch.png`
- The size is 512x512

PR #9593, opened 2025-07-07
Update tuya.ts

vecapetr:
add multiEndpoint: true,

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -18270,6 +18270,7 @@ export const definitions: DefinitionWithExtend[] = [
             e.enum("adjustment_mode", ea.STATE_SET, ["brightness", "color_temp"]).withDescription("Adjustment mode"),
         ],
         meta: {
+            multiEndpoint: true,
             tuyaDatapoints: [
                 [102, "state", tuya.valueConverter.onOff],
                 [103, "brightness", tuya.valueConverterBasic.scale(0, 254, 0, 1000)],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

vecapetr:
also not function good. If you push button on physical on device all flags transfer to z2m, but if you push in HA dashboard switch e.light_brightness_colortemp function fine, and if you push in HA dashboard e.switch() first or second: switch after few second come back and not switch device..., in z2m(edge) exposes all entities and state function fine.  
- from dashboard log: Device 'tuya_knob_1' has no endpoint 'l1' or 2
- from z2m edge log: z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/tuya_knob_1', payload '{"adjustment_mode":"brightness","brightness":254,"color_temp":370,"linkquality":148,"state":"ON","state_l1":"OFF","state_l2":"ON"}'

I will try delete all knob mqtt entities and start new complete pairing. :-)

src/devices/tuya.ts
```diff
@@ -18281,8 +18282,6 @@ export const definitions: DefinitionWithExtend[] = [
         },
         configure: tuya.configureMagicPacket,
         endpoint: (device) => ({
-            l1: 1,
```
Koenkk, line 18284: @vecapetr I think this needs to be added back, should solve your issue.

vecapetr: i will try... THX

PR #9594, opened 2025-07-07
Fix tamper and battery values for Develco (Frient) KEPZB-110

johnrvt:
This fixes the reading of the "tamper" and "battery_low" booleans, which are currently always null.

I tested it with my own Frient KEPZB-110.

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -964,7 +964,7 @@ export const ias_no_alarm: Fz.Converter = {
     cluster: "ssIasZone",
     type: ["attributeReport", "commandStatusChangeNotification"],
     convert: (model, msg, publish, options, meta) => {
-        const zoneStatus = msg.data.zoneStatus;
+        const zoneStatus = msg.data.zoneStatus ?? msg.data.zonestatus;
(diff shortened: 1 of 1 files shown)


src/converters/fromZigbee.ts
```diff
@@ -962,7 +962,18 @@ export const power_on_behavior: Fz.Converter = {
...
+            tamper: (zoneStatus & (1 << 2)) > 0,
+            battery_low: (zoneStatus & (1 << 3)) > 0,
+        };
+    },
+};
+export const ias_no_alarm_report: Fz.Converter = {
+    cluster: "ssIasZone",
+    type: "attributeReport",
     convert: (model, msg, publish, options, meta) => {
         const zoneStatus = msg.data.zoneStatus;
```
```suggestion
        const zoneStatus = msg.data.zoneStatus ?? msg.data.zonestatus;
```

Can you try with just this change?

johnrvt: It works. Actually, I wondered if it was better to make 2 functions or just add a condition.
Capture d’écran du 2025-07-08 19-55-18

Koenkk:
Thanks!

PR #9595, opened 2025-07-07
Add support for Tuya Smartica HY607W 3A thermostat (_TZE200_khah2lkr)

sigmich:
This adds support for Tuya Smartica HY607W 3A thermostat (_TZE200_khah2lkr), tested with Zigbee2MQTT 1.38.0. Includes support for target temperature, current temperature, on/off state, and mode switching to manual via DP 128. Uses custom external converter format.


Koenkk:
Could you add this to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts?

PR #9597, opened 2025-07-08
Move fanTimerMode for Inovelli devices to a common attribute

rohankapoorcom:
The attribute `fanTimerMode` is supported by all of the Inovelli devices except the VZM36. By moving it to the `COMMON_ATTRIBUTES` section, it is automatically exposed to all of them.


Koenkk:
Thanks!

PR #9601, opened 2025-07-09
fix(ignore): Add dummy device type

Koenkk:
Add a dummy device to allow for a more accurate device page to be generated, currently mainly focuses on Tuya dynamic exposes which use the `device.manufacturerName`.

Fixes https://github.com/Koenkk/zigbee2mqtt.io/pull/3910


Nerivec:
This ends up a breaking change (changes the converter typing - can no longer pass undefined to exposes fn).

It also breaks:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/886fc7f3e477700bb0a9eb85992cf38fb37cb580/src/index.ts#L481
(only tested by Z2M, not internally, probably need a better one in ZHC itself)

Koenkk:
True, fixed in https://github.com/Koenkk/zigbee-herdsman-converters/pull/9607, thanks!

PR #9609, opened 2025-07-11
fix: support ZG9098A-Win configure reporting and endpoint

niracler:
Fix these Issue

- The device state remains "Unknown"
- When controlling (up/down) the position does not update

from https://github.com/Koenkk/zigbee-herdsman-converters/pull/9473#issuecomment-3051865523


Koenkk:
thanks!

PR #9620, opened 2025-07-14
Add support for intuis radiator

marvinroger:
This PR adds support for french intuis radiators (using their so called "nativ intuis 3.0" protocol which is in fact Zigbee).

Device image added here: https://github.com/Koenkk/zigbee2mqtt.io/pull/3932

### DIFF
--- a/src/devices/intuis.ts
+++ b/src/devices/intuis.ts
@@ -1,0 +1,64 @@
+import * as fz from "../converters/fromZigbee";
+import * as tz from "../converters/toZigbee";
+import * as exposes from "../lib/exposes";
+import * as m from "../lib/modernExtend";
+import * as reporting from "../lib/reporting";
+import type {DefinitionWithExtend} from "../lib/types";
+
+export const definitions: DefinitionWithExtend[] = [
+    {
+        zigbeeModel: ["intuisradiator                 "],
+        model: "intuisradiator",
+        vendor: "intuis",
+        description: "intuis radiator with nativ and intuis 3.0",
+        extend: [
+            m.deviceEndpoints({endpoints: {1: 1, 5: 5}}),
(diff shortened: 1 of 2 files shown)


src/devices/intuis.ts
```diff
@@ -0,0 +1,64 @@
+import * as fz from "../converters/fromZigbee";
+import * as tz from "../converters/toZigbee";
+import * as exposes from "../lib/exposes";
+import * as m from "../lib/modernExtend";
+import * as reporting from "../lib/reporting";
+import type {DefinitionWithExtend} from "../lib/types";
+
+export const definitions: DefinitionWithExtend[] = [
+    {
+        zigbeeModel: ["intuisradiator                 "],
```
marvinroger, line 10: This is not a typo on my side, it's the actual payload sent by the modules

src/devices/intuis.ts
```diff
@@ -0,0 +1,64 @@
...
+        model: "intuisradiator",
```
Koenkk, line 11: Is there a proper model number written on the device or product page?

Koenkk:
Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt.io/pull/3932

src/devices/intuis.ts
marvinroger: I'm afraid It's a tricky question; all the radiators marketed [here](https://intuis.fr/en/our-products/electric-heating/electric-radiators) (named nativ) all share the same generic microcontrollers, but the actual radiators have different model names / numbers, and even different brands (white label).

There's nothing printed on the microcontroller boards, so I chose `intuisradiator` as  a fallback as it's what's reported as Zigbee model

Koenkk:
Thanks!

PR #9628, opened 2025-07-16
Update ubisys.ts for H1 voltage exposure

senna1992:
i did enhance converter for the "h1 TRV" to include the reported voltage. 

Regards to anyone involved


Koenkk:
> the converter works locally, but i cant fix the const "ubisysvoltage" code.

Whats the issue with `ubisysvoltage`?

senna1992:
> > the converter works locally, but i cant fix the const "ubisysvoltage" code.
> 
> Whats the issue with `ubisysvoltage`?

since I got rid of the constant, the converter now works fine according to the GitHub checks (locally for a few days I have no issues using it as well)

senna1992:
cleaned up my commit, i hope its good now!

Koenkk:
Thanks!

sjorge:
I don't think this will be enough as voltage is not configure to be reported. The device only self sends this once on pairing so the value is probably very out of date unless the value is manual ready as the fz.battery does read it.

Edit: A proper change would probably be:
- drop  e.battery_voltage(), e.battery().withAccess(ea.STATE_GET), from exposes
- drop  await reporting.batteryPercentageRemaining(endpoint, {min: constants.repInterval.HOUR, max: 43200, change: 1}); from configure
-  add battery modern extend, but being carful to use batteryPercentageRemaining for battery value as that one is pretty accurate, the voltage based mapping is not.

Edit2: easy change, and I was working on a different PR so quickly made one.

PR #9629, opened 2025-07-16
Update SEM-4-1-00. Add FPS-4-1-00.

AlexisPolegato:
Update SEM-4-1-00 to add Metering Attributes and change reporting values.

Add FPS-4-1-00.

### DIFF
--- a/src/devices/nodon.ts
+++ b/src/devices/nodon.ts
@@ -228,6 +228,14 @@ const nodonModernExtend = {
 
 export const definitions: DefinitionWithExtend[] = [
     {
+        zigbeeModel: ["FPS-4-1-00"],
+        model: "FPS-4-1-00",
+        vendor: "NodOn",
+        description: "Electrical heating actuator",
+        extend: [m.onOff({powerOnBehavior: true}), m.electricityMeter({cluster: "metering"}), ...nodonPilotWire(true)],
+        ota: true,
+    },
+    {
         zigbeeModel: ["IRB-4-1-00"],
         model: "IRB-4-1-00",
         vendor: "NodOn",
@@ -292,7 +300,17 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SEM-4-1-00",
         vendor: "NodOn",
         description: "Energy monitoring sensor",
-        extend: [m.electricityMeter()],
+        extend: [
+            m.electricityMeter({
+                acFrequency: true,
+                powerFactor: true,
+                configureReporting: false,
+                voltage: {min: 30, max: 3600, change: 500},
+                current: {min: 5, max: 3600, change: 100},
+                power: {min: 5, max: 3600, change: 200},
+                energy: {min: 30, max: 3600, change: 50},
+            }),
+        ],
         ota: true,
     },
     {


src/devices/nodon.ts
```diff
@@ -292,7 +300,17 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SEM-4-1-00",
         vendor: "NodOn",
         description: "Energy monitoring sensor",
-        extend: [m.electricityMeter()],
+        extend: [
+            m.electricityMeter({
+                acFrequency: true,
+                powerFactor: true,
+                configureReporting: false,
+                voltage: {min: 30, max: 3600, change: 500},
```
Koenkk: What's the reason to use non-default reporting intervals? (we prefer not to change the defaults)

AlexisPolegato: To reduce Zigbee Traffic

Koenkk: Does any particular attribute spam too much? If the default settings are to spammy we should consider changing the defaults.

Paulo-89:
Hello, 

I have two Nodon SEM-4-1-00 devices on my Zigbee network, and I had a problem with the divider/multiplier not being applied correctly.

The energy reported was 100 times higher than it should have been.

I had to add these on my converter:

```js
await reporting.readEletricalMeasurementMultiplierDivisors(endpoint, readFrequencyAttrs = true);
```

If it helps, here is my converter for my Nodon SEM-4-1-00:

```js
import * as reporting from "zigbee-herdsman-converters/lib/reporting";

const e = exposes.presets;

export default {
    zigbeeModel: ['SEM-4-1-00'],
    model: "SEM-4-1-00",
    vendor: "NodOn",
    description: "Energy monitoring sensor",
        await reporting.readEletricalMeasurementMultiplierDivisors(endpoint, readFrequencyAttrs = true);
        await reporting.bind(endpoint, coordinatorEndpoint, ["haElectricalMeasurement", "seMetering"]);
    },    
    
extend: [
        m.electricityMeter({
            producedEnergy: true,
            acFrequency: true,
            powerFactor: true,
            configureReporting: false,
            }),
        ],

exposes: [e.power_apparent()],
ota: true,
};
```

src/devices/nodon.ts
AlexisPolegato: The attributes I modified. Yes I guess we should consider changing the default settings on the device as a next step. Meanwhile it would be great to have it here

AlexisPolegato: @Koenkk  Could we merge it ?

Koenkk: >  Yes I guess we should consider changing the default settings on the device as a next step.

Why not use the defaults directly? The policy is to use the defaults unless there is a very good reason to deviate (this gives a consistent experience to users)

Koenkk:
Thanks!

Make sure that:
- The filename is `FPS-4-1-00.png`
- The size is 512x512

sbraz:
> The energy reported was 100 times higher than it should have been.

Reported that in https://github.com/Koenkk/zigbee2mqtt/issues/28898

PR #9630, opened 2025-07-16
Update owon.ts for THS317-ET

senna1992:
added exposure of voltage reading for ths317-et

### DIFF
--- a/src/devices/owon.ts
+++ b/src/devices/owon.ts
@@ -326,20 +326,27 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        zigbeeModel: ["THS317-ET"],
-        model: "THS317-ET",
-        vendor: "OWON",
-        description: "Temperature sensor",
-        fromZigbee: [fzLocal.temperature, fz.battery],
+        zigbeeModel: ['THS317-ET'],
+        model: 'THS317-ET',
+        vendor: 'OWON',
+        description: 'Temperature sensor',
+        fromZigbee: [owonTemperature, fz.battery],
         toZigbee: [],
-        exposes: [e.battery(), e.temperature()],
+        exposes: [
+            e.battery(),
+            e.voltage().withUnit('mV'),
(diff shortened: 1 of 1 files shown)


src/devices/owon.ts
```diff
@@ -330,9 +331,9 @@ export const definitions: DefinitionWithExtend[] = [
         model: "THS317-ET",
         vendor: "OWON",
         description: "Temperature sensor",
-        fromZigbee: [fzLocal.temperature, fz.battery],
+        fromZigbee: [fz.temperature, fz.battery],
         toZigbee: [],
-        exposes: [e.battery(), e.temperature()],
+        exposes: [e.battery(), e.voltage().withUnit("mV"), e.temperature()],
```
Koenkk: Neither the temperature or battery fromZigbee converters reports voltage, are you sure a voltage value is send?

senna1992: {
    "battery": 70,
    "linkquality": 188,
    "temperature": 25.2,
    "voltage": 2800
}


This is the report my 2 sensors expose to z2m

Koenkk: I see, please use `e.battery_voltage()` for that

src/devices/owon.ts
```diff
@@ -330,9 +331,9 @@ export const definitions: DefinitionWithExtend[] = [
         model: "THS317-ET",
         vendor: "OWON",
         description: "Temperature sensor",
-        fromZigbee: [fzLocal.temperature, fz.battery],
+        fromZigbee: [fz.temperature, fz.battery],
```
Koenkk: The custom temperature converter was used to fix https://github.com/Koenkk/zigbee2mqtt/issues/15173, I don't think it should be removed.

src/devices/owon.ts
senna1992: changes applied. thank you

Koenkk:
Thanks!

PR #9641, opened 2025-07-18
Fix mazda temperature calibration

Nopraz:
With localTempCalibration1 the temperature calibration value sometimes get to a weird value like -400000000. This PR fixes this weird behavior.


Koenkk:
Thanks!

PR #9642, opened 2025-07-18
Add support for SLC SmartOne Zigbee Remote control Zigbee IP20 (S57007)

Stepko2409:
PR to add "The Light Group" remote "[SLC SmartOne Zigbee Remote control Zigbee IP20 (S57007)](https://tlg.no/products/slc-smartone-zigbee-remote-control-zigbee-ip20)"

{
  zigbeeModel: ['S57007'],
        model: 'S57007',
        vendor: 'The Light Group AS',
        description: 'SLC SmartOne Zigbee Remote control Zigbee IP20',
        extend: [m.battery(), m.commandsOnOff(), m.commandsScenes()],
        meta: {"multiEndpoint":true},
}

### DIFF
--- a/src/devices/the_light_group.ts
+++ b/src/devices/the_light_group.ts
@@ -94,6 +94,11 @@ export const definitions: DefinitionWithExtend[] = [
         model: "S57007",
         vendor: "The Light Group",
         description: "SLC SmartOne Zigbee remote control Zigbee IP20",
-        extend: [m.deviceEndpoints({endpoints: {"1": 1, "2": 2}), m.battery(), m.commandsOnOff({endpointNames: ["1", "2"]}), m.commandsScenes({endpointNames: ["1", "2"]})],
+        extend: [
+            m.deviceEndpoints({endpoints: {"1": 1, "2": 2}}),
+            m.battery(),
+            m.commandsOnOff({endpointNames: ["1", "2"]}),
+            m.commandsScenes({endpointNames: ["1", "2"]}),
+        ],
     },
 ];
--- a/src/devices/the_light_group.ts
+++ b/src/devices/the_light_group.ts
@@ -94,11 +94,6 @@ export const definitions: DefinitionWithExtend[] = [
         model: "S57007",
         vendor: "The Light Group",
         description: "SLC SmartOne Zigbee remote control Zigbee IP20",
-        extend: [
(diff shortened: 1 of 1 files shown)


src/devices/the_light_group.ts
```diff
@@ -89,4 +89,12 @@ export const definitions: DefinitionWithExtend[] = [
         description: "SLC SmartOne TW led dimmable driver 24V/75W",
         extend: [m.light({colorTemp: {range: [160, 450]}})],
     },
+    {
+        zigbeeModel: ["S57007"],
+        model: "S57007",
+        vendor: "The Light Group AS",
+        description: "SLC SmartOne Zigbee Remote control Zigbee IP20",
+        extend: [m.battery(), m.commandsOnOff(), m.commandsScenes()],
+        meta: {multiEndpoint: true},
```
Koenkk: Is this needed? (can you try without?)

Stepko2409: All seem to be relevant. 

If I delete the first line with "vendor" I get an error in the log that the external converter is invalid and it is ignored.

If I delete the second line with "description" the external converter is also ignored, although I don't get anything in the logs.

In the extend part I need all three extensions in order to get the remote working with all features. And also the meta part is required. If this is missing all button presses results only in "on" or "off" instead of "on_1", "on_2"...

Stepko2409: Or did you mean the whole change? Yes, thats also needed. Without not all features are working e.g. the scene buttons won't work.

Koenkk: I updated the code a bit, can you check with the new version?

Stepko2409:
Thanks for your updates, they work! What I just changed:

1. The remote has 3 "on" and "off" buttons. So a third endpoint is needed
2. The "endpointsNames" are not required for the remote to work. However if you think they are needed or it is cleaner code, they can also be added. The remote works with and without them

Koenkk:
> The "endpointsNames" are not required for the remote to work. However if you think they are needed or it is cleaner code, they can also be added. The remote works with and without them

Please add back, these are required for the doc generator to show all actions.

Stepko2409:
> > The "endpointsNames" are not required for the remote to work. However if you think they are needed or it is cleaner code, they can also be added. The remote works with and without them
> 
> Please add back, these are required for the doc generator to show all actions.

Done :-)

Koenkk:
Thanks!

PR #9643, opened 2025-07-18
Update candeo.ts

candeodevelopment:
@Koenkk we took care of a few outstanding changes to our devices, nothing major just a few small tweaks, hope that's OK!

1. Renames a few consts to be more distinct.
2. Adds some safe (tested) initial default reporting configuration for some of our devices to match how they are configured on other platforms.
3. Configures the "switch type" on certain modules during initial pairing to match the factory default setting of those modules.


Koenkk:
Thanks!

PR #9646, opened 2025-07-18
Drop "Z-Wave" mention for Eco-DIm.07/Eco-Dim.10

hmmbob:
Noticed in HA:

image

But the device does not support Z-Wave at all... Especially not in the Zibee2Mqtt context 🤣 

Manufacturer does sell Z-Wave (and WiFi, and Matter) variants as well, but it's not a multi-protocol device - they're distinct versions: 
https://www.ecodim.nl/nl/search/Eco-Dim-dot-07/
https://www.ecodim.nl/nl/search/Eco-Dim-dot-10/

So it is safe to only report as a Zigbee dimmer 😉


Koenkk:
Thanks!

PR #9651, opened 2025-07-19
feat(devices): Add Yale Assure Lock 2 variant, YRD450-F

dflor003:
Adding YRD450-F TS model as a YRD450-BLE variant (Yale Assure Lock 2).


Koenkk:
Thanks!

PR #9655, opened 2025-07-20
Fix TS0726_1_gang_scene_switch for the scene action

devkaiwang:
Without `configureMagicPacket`, scene action cannot be triggered by pressing the button in scene mode.


Koenkk:
Thanks!

PR #9661, opened 2025-07-21
fix(detect): add E27 variant to Philips 046677577490

gpdenny:
Add E27 variant of the same bulb.
Wasn't sure if this should be a separate definition or not.


Koenkk:
Thanks!

PR #9664, opened 2025-07-21
fix: specify illuminance unit as "lx" for ZG-223Z

Foxhill67:
The `illuminance` exposure for the ZG-223Z device in `tuya.ts` had incorrect unit 'x'. This update corrects that via `.withUnit("lx")


Koenkk:
Thanks!

PR #9665, opened 2025-07-22
Add fingerprint for MOES ZB-TD5-RCW-GU10

xBelladonna:
Add manufacturer name `_TZ3210_ljoasixl` for Tuya whitelabel MOES `ZB-TD5-RCW-GU10` bulb


Koenkk:
Thanks!

PR #9669, opened 2025-07-22
feat(H1): switch to batteryModern extend

sjorge:
This is the proper change for #9628 as this will also setup reporting for voltage, but keeps using batteryPercentageRemaining for actual battery value.


Koenkk:
Thanks!

PR #9672, opened 2025-07-23
Add fingerprint for MOES SFL02-Z

sdbldn:
Add a new device：model: 'SFL02-Z',vendor: 'Moes',9dhenr94; wv9ukqca; tzyy0rtq; z3u99qxt; stvgmdjz; uenof8jd; zo0cfekv; dq8bu0pt

### DIFF
--- a/src/devices/engo.ts
+++ b/src/devices/engo.ts
@@ -312,4 +312,109 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE204_cmyc8g5i"]),
+        model: "E25-230",
+        vendor: "ENGO",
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -6447,6 +6447,266 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE200_9dhenr94", "_TZE200_dq8bu0pt"]),
+        model: "SFL02-Z-4",
```
Koenkk: Please move to `moes.ts`

sdbldn: I haven't tried to configure "moes.ts". Can I use it in the same way if I directly upload the existing configuration to the "moes.ts" file?

PR #9680, opened 2025-07-24
Update candeo.ts

candeodevelopment:
Some further improvements for Candeo devices.

For C210 dimming smart plug:

1) Adds additional fingerprints
2) Exposes Current Level Startup and Power On Behavior options

For LC20 smart LED controller:

1) Adds tested default Level reporting configuration 
2) Removes "effect" exposes, since the device doesn't support them


Koenkk:
thanks!

PR #9681, opened 2025-07-24
Add initialization command for BLE-YL01

jl1990:
BLE-YL01 seems to need this attribute set in order to trigger constant updates.
This should be tested properly before being merged, but in my experience after setting up the device with this command, I keep receiving updates after some hours (although the frequency as I mention is not ideal, it can take multiple hours before triggering an update).


Koenkk:
> This should be tested properly before being merged,

Are you going to test this more? I see little risk in merging this PR, in most cases it doesn't work anyway so there is nothing to break (and probably this won't break it even it it is working)

jl1990:
> > This should be tested properly before being merged,
> 
> Are you going to test this more? I see little risk in merging this PR, in most cases it doesn't work anyway so there is nothing to break (and probably this won't break it even it it is working)

I was expecting someone else to provide some feedback not to depend just on my device. 
But as I guess the idea is to merge it to the dev branch and currently this device is not working properly it should be fine.
Feel free to merge it.

Koenkk:
Thanks!

PR #9683, opened 2025-07-25
fix: set calibration option step at 0.1

Nerivec:
This will set frontend inputs to `step=0.1` which should fit better for most use cases (temp/hum). 
Note: a direct MQTT request allows any decimal precision. This only affects the published schema (what frontends use to display fields).

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/26455


Koenkk:
Thanks!

PR #9685, opened 2025-07-25
Update TS0601_dimmer_3 with light_type and reorganized code

bernardesarthur:
Update TS0601_dimmer_3 with light_type and reorganized code


Koenkk:
Thanks!

PR #9687, opened 2025-07-26
lixee/zlinky: fix default value for measurement_poll_chunk in description

sbraz:
Hi,
In 622631f9e4edd64590959db1479f6bee0f7f0d7b, the default changed from 2 to 4 but the description was not updated accordingly.


Koenkk:
Thanks!

PR #9692, opened 2025-07-26
chore: import vitest instead of relying on globals

Nerivec:
Same as https://github.com/Koenkk/zigbee-herdsman/pull/1464


Koenkk:
Thanks!

PR #9698, opened 2025-07-27
Support Philips Hue Play wall washer

Spirielle:
Added support for Philips Hue Play wall washer.

"LGT012" is the model found when enrolled in z2m
"046677590161" is the EAN/UPC for the black version
"046677590130" is the EAN/UPC for the white version

Added 4 new effects. Found them by trial and error.
When I select these effect in z2m, I can see the following effects are set on the device in the Hue app:
- "underwater" : Underwater-Blue
- "cosmos": Cosmos-Purple
- "sunbeam": Sunbeam-Default
- "enchant": Enchant-Default

### DIFF
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -4129,4 +4129,17 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Hue Devote Slim ceiling light",
         extend: [philips.m.light({colorTemp: {range: [153, 500]}, color: {modes: ["xy", "hs"], enhancedHue: true}})],
     },
+    {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

decowboy:
This device seems to be also named LGT009 and LGT010, where at least LGT010 is currently not recognized by Zigbee2MQTT.

Aside from that, the three lights in this device are not able to be individually controlled, as reported [here](https://github.com/koenkk/zigbee2mqtt/issues/28298).

UPDATE: I added a pull request [here](https://github.com/Koenkk/zigbee-herdsman-converters/pull/10203).

PR #9700, opened 2025-07-28
fix: Don't throw `Failed to apply calibration to` when value is an empty string

Koenkk:
When there is a `""` value for calibration or precision in the `configuration.yaml` an error is thrown, e.g.: `Failed to apply calibration to 'current': 'current_calibration' is not a number, got string ()`. It seems these values are set by the frontend in some cases, therefore I would propose to ignore these and fallback to the default calibration or precision value.

Example config:
```yaml
devices:
  '0xa4c138b3497171b4':
    current_calibration: ''
```

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/21724

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -165,7 +165,8 @@ export function calibrateAndPrecisionRoundOptionsIsPercentual(type: string) {
 export function calibrateAndPrecisionRoundOptions(number: number, options: KeyValue, type: string) {
     // Calibrate
     const calibrateKey = `${type}_calibration`;
-    let calibrationOffset = toNumber(options?.[calibrateKey] != null ? options[calibrateKey] : 0, calibrateKey);
+    const calibrateValue = options?.[calibrateKey];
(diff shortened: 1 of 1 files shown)


Nerivec:
Maybe Z2M settings validation should strip these out instead?

PS: WindFront should send "null".

Koenkk:
> Maybe Z2M settings validation should strip these out instead?

Z2M doesn't know the type of the options, so in some cases an empty string could be valid (but maybe we can introduce a full fledged JSON validation later)

> PS: WindFront should send "null".

With windfront I can also reproduce it, clicking from one option to the other I get the following in the config:

```yaml
devices:
  '0x70b3d52b600813d8':
    friendly_name: '0x70b3d52b600813d8'
    measurement_poll_interval: ''
    power_calibration: ''
```

Screenshot 2025-07-29 at 20 16 03

Nerivec:
About WindFront, should have been fixed in https://github.com/Nerivec/zigbee2mqtt-windfront/pull/137 but not released yet.

PR #9723, opened 2025-08-01
Added TS0203 to existing definition

SapozhnikUA:


src/devices/tuya.ts
```diff
@@ -1101,21 +1101,6 @@ export const definitions: DefinitionWithExtend[] = [
             exposes.options.calibration("temperature"),
         ],
     },
-    {
-        fingerprint: tuya.fingerprint("TS0203", ["_TZ3210_jowhpxop"]),
```
Koenkk: Why was this removed?

robertklep:
FWIW, I saw [this issue on Reddit](https://www.reddit.com/r/Zigbee2MQTT/comments/1nbfo4x/door_contact_sensors_suddenly_detected_as_temp/), where `TS0203/Zbeacon` is _also_ used by a contact sensor (which would make sense, given the `TS0203` identifier).

SapozhnikUA:
> FWIW, I saw [this issue on Reddit](https://www.reddit.com/r/Zigbee2MQTT/comments/1nbfo4x/door_contact_sensors_suddenly_detected_as_temp/), where `TS0203/Zbeacon` is _also_ used by a contact sensor (which would make sense, given the `TS0203` identifier).

Unfortunately, I don't have an experience to finish this request here.  It works al external converter in my HA since beginning of august

TaylorTWBrown:
My door sensor is now being detected as a humidity sensor. Is this the PR to fix it?

image

PR #9753, opened 2025-08-06
feat(add) : add electricalmeasurement cluster to some sinope devices

chdagenais:
Adding power and energy reporting for some sinope devices.


Koenkk:
Thanks!

PR #9754, opened 2025-08-06
fix: cleanup for Koenkk/zigbee-herdsman#1424

Nerivec:
Requires https://github.com/Koenkk/zigbee-herdsman/pull/1424
Arg was never much used, so removed entirely now that the only intended uses are handled in ZH.

Also some housekeeping to match with [upcoming TS 6](https://github.com/microsoft/TypeScript/issues/62198).


Koenkk:
Thanks!

PR #9756, opened 2025-08-07
Inovelli VZM31: Temporarily remove fanTimerMode

rohankapoorcom:
In https://github.com/Koenkk/zigbee-herdsman-converters/pull/9597, we made `fanTimerMode` available to all of the other Inovelli devices that were supposed to support it. VZM31 claims to support it but it's missing in the current firmware. Let's temporarily pull it out and then revert this change once they release a firmware update that adds support for it.


Koenkk:
Thanks!

PR #9757, opened 2025-08-07
Extend the  converter to allow passing in a payload object directly to the on_off converter

rohankapoorcom:
This modified converter is used so that devices like Inovelli that need to pass in a specific value `0xffff` to the `ontime` and `offwaittime` variables don't need to copy and implement their own converter.

Resolves https://github.com/Koenkk/zigbee2mqtt/issues/28023

### DIFF
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -41,7 +41,9 @@ export const on_off: Tz.Converter = {
             if (typeof offWaitTime !== "number") {
                 throw Error("The off_wait_time value must be a number!");
             }
-            const payload = meta.payload ? meta.payload : {ctrlbits: 0, ontime: Math.round(onTime * 10), offwaittime: Math.round(offWaitTime * 10)};
+            const payload = meta.converterOptions
+                ? meta.converterOptions
+                : {ctrlbits: 0, ontime: Math.round(onTime * 10), offwaittime: Math.round(offWaitTime * 10)};
             await entity.command("genOnOff", "onWithTimedOff", payload, utils.getOptions(meta.mapped, entity));
         } else {
             await entity.command("genOnOff", state, {}, utils.getOptions(meta.mapped, entity));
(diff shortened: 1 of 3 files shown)


src/converters/toZigbee.ts
```diff
@@ -41,8 +41,9 @@ export const on_off: Tz.Converter = {
             if (typeof offWaitTime !== "number") {
                 throw Error("The off_wait_time value must be a number!");
             }
-
-            const payload = {ctrlbits: 0, ontime: Math.round(onTime * 10), offwaittime: Math.round(offWaitTime * 10)};
+            const payload = meta.message.payload
```
Koenkk: Instead of altering the message, maybe we can pass it directly on the meta? (that's a bit cleaner than altering the `message`)

rohankapoorcom: Is there an explanation somewhere of the usage of meta and message and when to use each?

Koenkk: `message` is basically the received from MQTT (and meta allows for more free-form properties)

rohankapoorcom: Got it, I think moving it to the meta makes sense then.

rohankapoorcom: If we move this to the meta, then I don't think we actually need to make a wrapped version of the `light_onoff_brightness_onoff_payload` since it should be passing along the meta as is (as far as I can tell).

src/lib/types.ts
```diff
@@ -337,6 +337,7 @@ export namespace Tz {
         endpoint_name: string | undefined;
         membersState?: {[s: string]: KeyValue};
         publish: Publish;
+        payload?: KeyValue;
```
Koenkk: Maybe we can call this `converterOptions`? (since the `payload` term is quite overloaded already)

rohankapoorcom: Done.

Koenkk (approved):

rohankapoorcom:
Thanks for the review and approval. Let's hold off on merging for a few days while @InovelliUSA and I finish testing to make sure there's not some other behavior that was missed.

InovelliUSA:
From my testing, these changes look good. Thanks!

rohankapoorcom:
@Koenkk I think we're ready to merge. Thanks!

Koenkk:
Thanks for the cleanup!

PR #9766, opened 2025-08-08
Add better support for BLE-YL01

xvolte:
Trying to add better support for BLE-YL01 pool sensor.

Please have a look here : 
https://github.com/Koenkk/zigbee2mqtt/issues/23946#issuecomment-3167453962

and here:
https://community.home-assistant.io/t/pool-monitoring-device-yieryi-ble-yl01-zigbee-ph-orp-free-chlorine-salinity-etc/659545/130?u=xvolte

Note that instead of "CUSTOM" i put the date of the latest converter modification

@Koenkk : i saw that it modified some code in the  `configure: tuya.configureMagicPacket` , i don't know if it should be modified or not (i copy pasted my config file).


Koenkk:
Thanks!

Issue #9617, opened 2025-07-13
_TZ3210_bfwvfyx1 (former "TS0505B_4") are no longer controlleable


ESP4Ever:
Hello!
I couldn't find a correct thread nor when this ID was altered so I decided to open a new issue.
Currently the color control is not operational and if the brightness <48 the bulb will not turn on or will just blink once

This code was working relatively fine with my bulbs:
```

        fingerprint: tuya.fingerprint("TS0505B", ["_TZ3210_bfwvfyx1"]),
        model: "TS0505B_4",
        vendor: "Tuya",
        description: "Zigbee RGB+CCT light",
        whiteLabel: [tuya.whitelabel("Tuya", "TS0505B_4_1", "Zigbee 3.0 15W led light bulb E27 RGBCW", ["_TZ3210_bfwvfyx1"])],
        toZigbee: [tz.on_off, tzLocal.led_control, tuya.tz.do_not_disturb],
        fromZigbee: [fz.on_off, fz.tuya_led_controller, fz.brightness, fz.ignore_basic_report],
        exposes: [e.light_brightness_colortemp_colorhs([50, 1000]).removeFeature("color_temp_startup"), tuya.exposes.doNotDisturb()],
        configure: (device, coordinatorEndpoint) => {
            device.getEndpoint(1).saveClusterAttributeKeyValue("lightingColorCtrl", {
                colorCapabilities: 29,
            });
        },
```

while the new one doesn't:

```
        fingerprint: tuya.fingerprint("TS0505B", ["_TZ3210_bfwvfyx1"]),
        model: "CK-BL702-AL-01_1",
        vendor: "eWeLink",
        description: "Zigbee 3.0 15W led light bulb E27 RGBCW",
        toZigbee: [tuya.tz.do_not_disturb],
        extend: [m.light({colorTemp: {range: [50, 500], startup: false}, effect: false, powerOnBehavior: false, color: {modes: ["xy", "hs"]}})],
        exposes: [tuya.exposes.doNotDisturb()],
        configure: (device, coordinatorEndpoint) => {
            device.getEndpoint(1).saveClusterAttributeKeyValue("lightingColorCtrl", {colorCapabilities: 29});
        },
```

wenzelt:
I have the same issue. Same Zigbee Manufacturer: _TZ3210_bfwvfyx1

Zigbee Model
TS0505B
Zigbee Manufacturer
_TZ3210_bfwvfyx1

[...]

PR #9769, opened 2025-08-09
Request for review a commited fix: eWeLink CK-BL702-AL-01_1

l1berat0r:
I wonder if it's related to the firmware version, but in my case , commit [bf81743
](https://github.com/Koenkk/zigbee-herdsman-converters/commit/bf81743423c9482f67c812525e6a4405672e1142) messes things up even more. I found, that what is working for me instead is disabling "enhancedHue" mode for the light color setting.

Do you think this solves the issue in your case as well? If not, can we try to debug things a little and find why there are problems with my devices?

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1940,9 +1940,16 @@ export const definitions: DefinitionWithExtend[] = [
         model: "CK-BL702-AL-01_1",
         vendor: "eWeLink",
         description: "Zigbee 3.0 15W led light bulb E27 RGBCW",
-        toZigbee: [tz.on_off, tzLocal.led_control, tuya.tz.do_not_disturb],
-        fromZigbee: [fz.on_off, fz.tuya_led_controller, fz.brightness, fz.ignore_basic_report],
-        exposes: [e.light_brightness_colortemp_colorhs([50, 1000]).removeFeature("color_temp_startup"), tuya.exposes.doNotDisturb()],
+        toZigbee: [tuya.tz.do_not_disturb],
+        extend: [
+            m.light({
+                colorTemp: {range: [50, 500], startup: false},
+                effect: false,
+                powerOnBehavior: false,
+                color: {modes: ["xy", "hs"], enhancedHue: false},
+            }),
+        ],
+        exposes: [tuya.exposes.doNotDisturb()],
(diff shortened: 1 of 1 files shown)


Koenkk:
Let's first check in https://github.com/Koenkk/zigbee-herdsman-converters/issues/9617

l1berat0r:
After reading latest comment to [#9617](https://github.com/Koenkk/zigbee-herdsman-converters/issues/9617#issuecomment-3174381950) I tested it one more time and indeed I was able to controll the device. I'm not sure why it didn't work to me before. Maybe I was on some different branch of some other dependency...
The only issue I see now, are the errors appearing in the log, which originate from the `tzLocal.led_control` function, trying to read unsupported attributes. 
I'm new to this project, so I'm sorry if my suggestions are off, but do you think it's good enough or should we aim to clean that up?
The code of the converter, before the commit [bf81743](https://github.com/Koenkk/zigbee-herdsman-converters/commit/bf81743423c9482f67c812525e6a4405672e1142) seems much more clean to me and in my case it needed only slight tweaking to start working properly.
On the other hand `tzLocal.led_control` function is quite complex and making it aware of which attributes are supported by which devices would require some carefull refactoring.

Koenkk:
I would keep it as is, it seems this device requires specific Tuya commands.

l1berat0r:
Could you please help me understand what type of speciffic commands could it need? I spent some time trying to tune the setup for my devices and I'm stil at the position that the "modern extend" version works pretty well for me. 

I indeed found one issue, with how those bulbs handle "moveToLevelWithOnOff" command". It works fine for the levels higher than 20, but if I set anything lower it can turn off on random (more ofthen than not). Maybe this is also simmilar case as described in another thread?
I created a workaround in my branch ([2ae0db3](https://github.com/Koenkk/zigbee-herdsman-converters/commit/2ae0db33e0338b21cb0b9fecc79b80d8d5805041)) but I can't tell for sure it won't mess up some other parts of the logic.
Currently I'm more or less sattisfied with this version, because I could also enable "powerOnBehavior", which the bulbs also seem to support. I wonder if the guy for the other thread would like to test this commit and check if it solves the issues he had previously.

Koenkk:
Let's try, thanks!

PR #9779, opened 2025-08-10
fix: adjust to new typing from ZH, fix discovered issues

Nerivec:
Requires https://github.com/Koenkk/zigbee-herdsman/pull/1479

TODO:
- [x] support for custom clusters (only custom cluster types required for current code have been added)
- [x] heavy changes in `toZigbee.light_color`, in-depth review needed

and a few more in the code itself

TODO (separate PRs):
- [ ] lib/ewelink -> lots of unnecessary clusterName, writeCommand strings?
- [ ] type `validateValue`? (could remove quite a few `as "x" | "y"` like `as "toggle" | "off" | "on"`
- [ ] `globalStore.getValue` is typed as `any`, should refactor

# Discovered issues

- fix: `toZigbee.danfoss_heat_supply_request` wrong attribute name
- fix: Danfoss Icon2 wrong attribute name
- fix: bacchus using wrong payload format in `timeHHMM`
- fix: ambiguous attr ID hex format in shinasystem `fzLocal.ct_direction` & `tzLocal.ct_direction`
- fix: ambiguous attr value format in legrand `tzLegrand.calibration_mode`
- fix: invalid payload formats in namron `toZigbee.namron_edge_thermostat_vacation_date`
- fix: invalid payload formats in namron `toZigbee.namron_edge_thermostat_holiday_temp`

from commands typing:

- fix: `cover_via_brightness` using invalid value type in command
- fix: `fan_speed` using invalid value type in command
- fix: `twinguardSmokeDetector` `initiateTestMode` command missing payload (was using manuf opts in place, i.e. manuf opts were ignored)
- fix: `twinguardSetup` `pairingCompleted` command missing payload (was using manuf opts in place, i.e. manuf opts were ignored)
- fix: `manuSpecificYokisDevice` `openNetwork` using improper casing in command name
- fix: `manuSpecificYokisInput` `sendPress` using improper custom cluster name
- fix: `manuSpecificYokisInput` `sendRelease` using improper custom cluster name
- fix: `manuSpecificYokisInput` `selectInputMode` using improper custom cluster name

### DIFF
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -368,7 +368,7 @@ export const lock: Tz.Converter = {
 export const lock_auto_relock_time: Tz.Converter = {
     key: ["auto_relock_time"],
     convertSet: async (entity, key, value, meta) => {
-        await entity.write("closuresDoorLock", {autoRelockTime: value}, utils.getOptions(meta.mapped, entity));
+        await entity.write("closuresDoorLock", {autoRelockTime: value as number}, utils.getOptions(meta.mapped, entity));
         return {state: {auto_relock_time: value}};
     },
     convertGet: async (entity, key, meta) => {
@@ -566,7 +566,7 @@ export const warning: Tz.Converter = {
 export const ias_max_duration: Tz.Converter = {
     key: ["max_duration"],
     convertSet: async (entity, key, value, meta) => {
-        await entity.write("ssIasWd", {maxDuration: value});
+        await entity.write("ssIasWd", {maxDuration: value as number});
         return {state: {max_duration: value}};
     },
     convertGet: async (entity, key, meta) => {
@@ -807,7 +807,14 @@ export const level_config: Tz.Converter = {
         }
     },
     convertGet: async (entity, key, meta) => {
-        for (const attribute of ["onOffTransitionTime", "onTransitionTime", "offTransitionTime", "startUpCurrentLevel", "onLevel", "options"]) {
+        for (const attribute of [
+            "onOffTransitionTime",
+            "onTransitionTime",
+            "offTransitionTime",
+            "startUpCurrentLevel",
+            "onLevel",
+            "options",
+        ] as const) {
             try {
                 await entity.read("genLevelCtrl", [attribute]);
             } catch {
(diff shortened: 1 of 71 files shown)


Nerivec:
@Koenkk can you take a look at the few remaining errors? Some legacy stuff, and a few weird occurrences.
```
Errors  Files
     3  src/devices/danfoss.ts:896
     2  src/devices/xyzroe.ts:83
     1  src/lib/develco.ts:55
    14  src/lib/legacy.ts:5500
```

src/devices/inovelli.ts
```diff
@@ -1586,7 +1586,8 @@ const tzLocal = {
                     entityToUse = meta.device.getEndpoint(Number(keysplit[1]));
                     keyToUse = keysplit[0];
                 }
-                await entityToUse.read(cluster, [keyToUse], {
+                // XXX: far too dynamic to properly type
+                await entityToUse.read(cluster, [keyToUse] as Parameters<typeof entityToUse.read>[1], {
```
Nerivec, line 1737: @rohankapoorcom in case you have some time to look this over, don't need to include changes in this PR, but would be nice if we could type this better in the future. Seems a lot of Inovelli code is very dynamic, which is good for size, but very hard to properly type (i.e. more dangerous).
The base would be to have statically-defined `cluster name` and `attribute name` (not dynamically generated with string manipulation & co) whenever there are calls to ZH, which would allow to use the new typing introduced in ZH and used elsewhere in this PR.
_Note: a little manipulation is doable ([example](https://github.com/Nerivec/zigbee-herdsman-converters/blob/24baaa2cab75c84ac88813f4df1bfd47a77efc2a/src/lib/modernExtend.ts#L2830)) but more advanced stuff like split, etc. breaks down typing to plain `string` which we can't validate against cluster types._

src/converters/toZigbee.ts
```diff
@@ -90,62 +86,79 @@ export const light_color: Tz.Converter = {
...
-            await entity.command(
-                "genLevelCtrl",
-                "moveToLevelWithOnOff",
-                {level: Number(value.brightness), transtime: utils.getTransition(entity, key, meta).time},
-                utils.getOptions(meta.mapped, entity),
-            );
+                await entity.command("lightingColorCtrl", "moveToSaturation", {transtime, saturation}, utils.getOptions(meta.mapped, entity));
+            }
+        } else {
+            // TODO: throw?
```
Koenkk: I think it's best to throw in this case indeed.

src/converters/toZigbee.ts
```diff
@@ -90,62 +86,79 @@ export const light_color: Tz.Converter = {
 
             newState.color_mode = constants.colorModeLookup[1];
             newState.color = xy.toObject();
-            zclData.colorx = utils.mapNumberRange(xy.x, 0, 1, 0, 65535);
-            zclData.colory = utils.mapNumberRange(xy.y, 0, 1, 0, 65535);
-            command = "moveToColor";
+            const colorx = utils.mapNumberRange(xy.x, 0, 1, 0, 65535);
+            const colory = utils.mapNumberRange(xy.y, 0, 1, 0, 65535);
+
+            // XXX: should this be skipped entirely in this codepath?
```
Koenkk: I think this can be removed indeed, will be handled by the `brightness` converter. Note that `value` will never have `brightness` (because then you would have to publish `{"color":{"x":..,"y":..,"brightness":..}}`

Nerivec:
I reverted the tsconfig `module` change to `NodeNext`. It would require testing with external converters to see if it impacts importing & whatnot (it at least requires using `.js` extensions in the repos for dyn imports). _Shouldn't include it in this big PR anyway._

PR #9780, opened 2025-08-11
Updated wireless endpoint for WS-K02E

Holbert:
Updating wireless endpoint for WS-K02E from 2 to 4. With wireless set to 2 multiclick is unable to be enabled. Changing to 4 fixes this and displays the proper action such as single_wireless, double_wireless, hold_wireless, and release_wireless.

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -4505,7 +4505,7 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [
             lumiZigbeeOTA(),
             lumiPreventReset(),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

RaggedyDandy:
i need to ask, as someone who is not well-versed in this...I have the same issue of the bottom button not working. It's not clear to me if this was supposed to sort that out? If I could ask, any help/guidance you could give? Here's what i think are important pieces of information?

Thanks!

image

image

image

PR #9781, opened 2025-08-11
Correct vendor's name from MOES to Moes (as it already exists)

bernardesarthur:
Correct vendor's name from MOES to Moes (as it already exists)


Koenkk:
Thanks!

PR #9783, opened 2025-08-11
Add vesternet VES-ZB-HLD-017 fingerprint

rabb13:
One of the device i purchased came with newer version of build id - `2.9.2_r55`
This is the device I purchased - https://www.amazon.co.uk/dp/B0C1P956JN 

Adding the new  build id to fingerprint. 

Thanks


Koenkk:
Thanks!

PR #9785, opened 2025-08-11
Remove duplicated options (from TS110E_2gang_2)

bernardesarthur:
I have these devices, and the PowerOnBehavior options work the same for all endpoints, so there’s no need to have two separate options for it.  SwitchType works fine and LightType doesn't work.


Koenkk:
Thanks!

PR #9800, opened 2025-08-14
fix: ROB_200-018-0 - Added _1 postfixed actions

bramnauta:
Added actions with _1 postfix as my ROB_200-018-0 dimmer switches are only exposing those. Please let me know if this should be handled in another way!


Koenkk:
Thanks!

PR #9812, opened 2025-08-17
Fix Ph value if ph > 99 in device BLE-YL01.

xvolte:
Fix Ph value if ph > 99 in device BLE-YL01.

See comment :  https://github.com/Koenkk/zigbee2mqtt/issues/23946#issuecomment-3185265461


Koenkk:
Thanks!

PR #9821, opened 2025-08-18
Inovelli: Fix broken on/off command with recent code cleanup

InovelliUSA:
@InovelliUSA & @rohankapoorcom pushed some changes recently that are causing on / off to throw an error:

z2m: Publish 'set' 'state' to 'ZB - B - Zone 3 (L)' failed: 'RangeError [ERR_OUT_OF_RANGE]: ZCL command 0xb43a31fffe26e207/1 genLevelCtrl.moveToLevelWithOnOff({"level":null,"transtime":655350}, {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":false,"direction":0,"reservedBits":0,"writeUndiv":false}) failed (The value of "value" is out of range. It must be >= 0 and <= 65535. Received 655350)'

It seems that the default transition time that Inovelli is supposed to use when an explicit transition time is not specified, is getting multiplied by 10 somewhere. This fix is a workaround until we can find out why that is happening.

@Koenkk curious if you may be familiar with why this is happening.

### DIFF
--- a/src/devices/inovelli.ts
+++ b/src/devices/inovelli.ts
@@ -1828,7 +1828,7 @@ const tzLocal = {
                 ...meta,
                 message: {
                     ...meta.message,
-                    transition: !transition.specified ? 6553.5 : Math.round(transition.time / 10),
+                    transition: (!transition.specified ? 0xffff : Math.round(transition.time)) / 10,
                 },
                 converterOptions: {
                     ctrlbits: 0,


src/devices/inovelli.ts
```diff
@@ -1828,7 +1828,7 @@ const tzLocal = {
                 ...meta,
                 message: {
                     ...meta.message,
-                    transition: !transition.specified ? 0xffff : transition.time,
+                    transition: !transition.specified ? 6553.5 : Math.round(transition.time / 10),
```
rohankapoorcom: 6553.5 should probably stay as 0xffff for consistency.

InovelliUSA: But the value is getting multiplied by 10. Is there a hex representation for the decimal value? In other words, the value getting passed in the converter is being multiplied by 10 somewhere before it gets sent to the device. So it needs to be divided by 10 before it continues from that point. Are you recommending that I do (0xffff / 10)? I guess I could check to see if that works.

rohankapoorcom: I must have misunderstood. I thought the incoming `transition` value was being multiplied by 10. I didn't realize, it was happening on the outgoing side.

This is kind of bizarre.

> Are you recommending that I do (0xffff / 10)

I think I would move the `/10` to be outside of the ternary so that it happens regardless.

`transition: (!transition.specified ? 0xffff : Math.round(transition.time)) / 10,`

rohankapoorcom:
Adding a general comment to link the previous PR that we think introduced this issue: https://github.com/Koenkk/zigbee-herdsman-converters/pull/9757

src/devices/inovelli.ts
InovelliUSA: Ok, I modified it as suggested.

Koenkk:
Thanks!

rohankapoorcom:
@Koenkk any idea why we need to divide by 10 here? As far as I can tell, the previous code never did that.

Koenkk:
I think because the toZigbee light converters [multiply it by 10](https://github.com/Koenkk/zigbee-herdsman-converters/blob/e0dd2914981c76019a77240c5e92bd77d1d1f045/src/lib/utils.ts#L455)

rohankapoorcom:
> I think because the toZigbee light converters [multiply it by 10](https://github.com/Koenkk/zigbee-herdsman-converters/blob/e0dd2914981c76019a77240c5e92bd77d1d1f045/src/lib/utils.ts#L455)

Interesting. I see what's happening now. We're double multiplying by 10 because we call the same function in the Inovelli converter now (which multiplies by 10), then replace the value and pass along to the stock converter. Then the function is called again by the stock converter (which multiplies by 10 again). So dividing by 10 here in both cases simply reverts the second multiply by 10 that will happen afterwards.

Now I'm curious why neither Eric nor I ran into this in testing the previous PR.

PR #9835, opened 2025-08-22
Create custom modernExtend classes to use with Inovelli devices

rohankapoorcom:
I thought it might be helpful to create some `modernExtend`s classes for Inovelli devices since they have a lot of repetitive calls to `fz`/`tz`/`configure` between them and we've run into some issues where things have been added to one device and not others.

This feels like an improvement (for everything except the VZM36). Curious if there's any thoughts on how to clean this up further.

Leaving this in draft format while @InovelliUSA and I finish testing this.

### DIFF
--- a/src/devices/inovelli.ts
+++ b/src/devices/inovelli.ts
@@ -2451,7 +2451,7 @@ export const definitions: DefinitionWithExtend[] = [
             m.identify(),
             m.temperature(),
             m.humidity(),
-            m.electricityMeter(),
+            m.electricityMeter({energy: {divisor: 1000}}),
         ],
         ota: true,
     },
@@ -2509,6 +2509,7 @@ export const definitions: DefinitionWithExtend[] = [
                 // Current and voltage were removed in version 0.8 of the firmware and expected to be restored in the future
                 current: false,
                 voltage: false,
+                energy: {divisor: 1000},
             }),
             m.illuminance(),
             m.occupancy(),
--- a/src/devices/inovelli.ts
+++ b/src/devices/inovelli.ts
@@ -550,7 +550,7 @@ const LED_NOTIFICATION_TYPES: {[key: number]: string} = {
     5: "LED_6",
     6: "LED_7",
     16: "ALL_LEDS",
-    255: "CONFIG_BUTTON_DOUBLE_PRESS",
+    "-1": "CONFIG_BUTTON_DOUBLE_PRESS",
 };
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Let me know once it's ready, want to merge this before the 1 September release such that https://github.com/Koenkk/zigbee2mqtt/issues/28293 can be closed.

rohankapoorcom:
From my perspective this is ready. I've been running it all weekend and fixed a couple of issues but overall it's looking good to me.

@InovelliUSA - have you had a chance to test?

@Koenkk - Any modifications or suggestions for the code?

rohankapoorcom:
I think I've found another instance of the same problem described in https://github.com/Koenkk/zigbee2mqtt/issues/28293. When I issue a homeassistant service call to turn off a light (VZM31/VZM32), Z2M seems to be setting the brightness to 1. When I then turn it on from homeassistant, it correctly returns to the previous brightness (not 1), but when I turn it on from the switch itself, it's turning on to 1% (because that's what the switch remembers as the previous state).

Is there a way to tell Z2M to not use 1% as the off brightness here @Koenkk? If not, I'm afraid we may have to revert https://github.com/Koenkk/zigbee-herdsman-converters/pull/9757 https://github.com/Koenkk/zigbee-herdsman-converters/pull/9821 and go back to having the Inovelli converter have it's own light_onoff_brightness implementation for the next release and then revert the reverts after the release is cut so that we have more time to sort this out.

Koenkk:
@rohankapoorcom does the [`turnsOffAtBrightness1`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/5a7f2352e34d0e6ec02c3bfd03dd31cf62253efe/src/devices/istar.ts#L10) work here?

src/devices/inovelli.ts
```diff
@@ -484,13 +491,13 @@ const inovelliExtend = {
         return {
             fromZigbee,
             toZigbee: [
+                tz.on_off,
```
rohankapoorcom, line 494: @InovelliUSA found the problem. Apparently order matters when setting up the toZigbee converters. Putting `tz.on_off` first before `light_onoff_brightness` solves it.

rohankapoorcom:
I'm good to go with this if @InovelliUSA concurs.

InovelliUSA:
I've tested with the latest commits and things seem good to me.

Koenkk:
Thanks again!

PR #9848, opened 2025-08-26
Correct vendor from GiEX to GIEX

bernardesarthur:
Just to correct vendor's name from GiEX to GIEX maintaining this standard.

### DIFF
--- a/src/devices/giex.ts
+++ b/src/devices/giex.ts
@@ -80,6 +80,6 @@ export const definitions: DefinitionWithExtend[] = [
                 .withUnit("sec")
                 .withDescription("Cycle irrigation interval"),
         ],
-        whiteLabel: [tuya.whitelabel("", "GX02", "Water valve", ["_TZE204_7ytb3h8u", "_TZE284_7ytb3h8u", "_TZE200_7ytb3h8u"])],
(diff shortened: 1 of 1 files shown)


src/devices/giex.ts
```diff
@@ -80,6 +80,6 @@ export const definitions: DefinitionWithExtend[] = [
                 .withUnit("sec")
                 .withDescription("Cycle irrigation interval"),
         ],
-        whiteLabel: [tuya.whitelabel("GiEX", "GX02", "Water valve", ["_TZE204_7ytb3h8u", "_TZE284_7ytb3h8u", "_TZE200_7ytb3h8u"])],
+        whiteLabel: [tuya.whitelabel("", "GX02", "Water valve", ["_TZE204_7ytb3h8u", "_TZE284_7ytb3h8u", "_TZE200_7ytb3h8u"])],
```
Koenkk: Was this intentional? It should specify a vendor name here.

bernardesarthur: Sorry, my mistake. I want to write GIEX instead of GiEX. I'll correct that.

Koenkk:
Thanks!

PR #9852, opened 2025-08-27
Bosch `BMCT-SLZ`: Various enhancements and fixes

DerDreschner:
This pull request includes various enhancements and fixes for the Bosch `BMCT-SLZ` (Light/shutter control unit II).

Here is a list of the changes I made and the reasons behind them:
- **Decouple switch from output**: In the last firmware updates, Bosch included the possibility to decouple the switch from the output. This means that the output state doesn't change when the switch is being used. I offer the same options as the Bosch app does, which means that the `decoupled` and `only_short_press_decoupled` options are not available when using the device for shutter control.
- **Switch state report**: The `BMCT-SLZ` reports any switch changes as a command to the coordinator. Unfortunately, there is no real separation between the inputs as Bosch uses the same command for both. This results in only one action attribute with several states depending on the used switch type, the input and what happened lastly. If there are any suggestions on how I can improve that situation, I'm open for comments!
- **Automatic turn-off in light mode**: When using the `BMCT-SLZ` via the Bosch app, you can automatically turn off each light. This is based on a time schedule for regular changes (which are being stored and executed on the Bosch SHC) as well as two separate attributes on the `BMCT-SLZ` itself for more short-term changes. The latter is being supported by this pull request. To use it, it's necessary to set the desired amount of time to disable the light as well as enable the feature. If the light is already on when enabling the feature, the internal counter starts from that moment. Otherwise, it's counted from the moment the light is turned on.
- **Reset the used energy reading**: Bosch provides a custom command to reset the internal energy meter. This pull request enables the use of that command via Z2M as older firmware versions may have an error which leads to abnormal energy readings (see #9768 for example). By resetting the internal energy meter, at least the abnormal energy reading is gone.
- **Reduce the minimum change value for `seMetering` and `haElectricalMeasurement` to 1**: By default, the `electricityMeter` function uses a minimum change for the `seMetering` and `haElectricalMeasurement` clusters of 5. As lights may use very little energy, this may be a bit high. Therefore, I change the minimum change value to 1 as that fits the use case better, in my opinion (and the Bosch SHC uses it as well).
[...]

PR #9862, opened 2025-08-27
Bosch `BMCT-SLZ` (Light/shutter control unit II): Only expose attributes compatible with the selected switch type

DerDreschner:
In the last pull request #9852 , I added support for decoupling the connected switches from the outputs. As not all switch types support all modes, it's easy to configure an unsupported mode.

This change makes sure that after changing the switch type, the switch mode and child lock is being reset to the default values (to avoid a "broken" configuration) and the exposed attributes is being updated. If no switch is connected, I hide all relevant attributes to avoid "dead" attributes (`action`, `action_duration`, `switch_mode` and `child_lock` for the relevant endpoints).


Koenkk:
Thanks!

PR #9865, opened 2025-08-27
Bosch `BSP-FZ2` (Plug compact II): Support reset of energy meter and lower min value change on electricityMeter reporting

DerDreschner:
During development of #9852 , I realized that the `BSP-FZ2` uses the same command for resetting the integrated energy meter. Therefore, I support that feature on the `BSP-FZ2`, too. As I already changed the extends, I lowered the minimum value change for the `seMetering` and `haElectricalMeasurement` clusters to match the Bosch default there (as already did for the `BMCT-SLZ`).


Koenkk:
Thanks!

PR #1201, opened 2020-05-01
Support leakSMART water leak sensor

presslab-us:
Requires pull request: https://github.com/Koenkk/zigbee-herdsman/pull/169

PR #9867, opened 2025-08-27
feat: add FZ typing

Nerivec:
As before, left several TODOs in the code.

- quite a few errors are about the typing now having proper number, but the converter calling `Number.parse...`
  - remove parse or type-assert to always parse?
- a few errors are about the use of one converter, that covers several "type" but simply `if`/`else` inside
  - split them?
  - convert check against type to check against prop? e.g. `msg.type === "commandMoveWithOnOff"` => `"movemode" in msg.data` for narrowing support (_there might be a way to narrow with `type` but the typing is already quite involved..._)
- calling `xyz.convert(...)` from within a mismatching converter type
- special cases:
  - `for (const dpValue of msg.data.dpValues)` when possibly from "raw" produces `for (const dpValue of undefined)` => `TypeError: msg.data.dpValues is not iterable`

## Discovered issues

- fix: [awox] wrong payload format in `awox_color_ctrl` & `awox_level_ctrl`
- fix: [bosch] wrong check format in `ignoreDst`
- fix: [develco] `fz.metering` array-indexing on number-type?
- fix: [heiman] awkward logic on `number | Buffer` type
- fix: [inovelli] incorrect attribute casing in `breeze_mode`
- fix: [xyzroe] incorrect attribute in `fzLocal.zigusb_button_config`
- fix: [lib/develco] incorrect attribute case in `develcoModernExtend.batteryLowAA`
- fix: [lib/legacy] incorrect command names in `fromZigbee.ZMRM02`
- fix: [lib/legacy] incorrect command names in `fromZigbee.tuya_smart_vibration_sensor`
- fix: [lib/legacy] incorrect use of `msg.data.dpValues` in `fromZigbee.tuya_smart_vibration_sensor` when type could be `raw` (see above special case)
- fix: [lib/legacy] incorrect use of `msg.data.dpValues` in `fromZigbee.matsee_garage_door_opener` when type could be `raw` (see above special case)
- fix: [lib/legacy] incorrect command names in `fromZigbee.tuya_remote`
- fix: [lib/modernExtend]: incorrect prop used to get `zclTransactionSequenceNumber` in `iasGetPanelStatusResponse`


src/devices/xyzroe.ts
```diff
@@ -212,6 +212,7 @@ const fzLocal = {
         type: ["readResponse", "attributeReport"],
         convert: (model, msg, publish, options, meta) => {
             const channel = utils.getKey(model.endpoint(msg.device), msg.endpoint.ID);
+            // TODO: wrong name? switchType?
             const {buttonMode} = msg.data;
```
Koenkk: @xyzroe what should be the property here?

xyzroe: Sorry, but I don't understand the question. If I don't miss anything, this is variation of a customized PTVO switch.

Nerivec: I think this is the same that was fixed here in previous PR?
https://github.com/Koenkk/zigbee-herdsman-converters/pull/9779/commits/f2c4645da0aee073bb6a4cc440b7572c9bd2500c

Koenkk: @xyzroe `buttonMode` is currently broken because it does not exist on `msg.data`, the lines below it uses something like this:

```
            const inputLink = msg.data[0x4001];
            const bindCommand = msg.data[0x4002];
```

So should `const {buttonMode} = msg.data;` be changed to e.g. `const buttonMode = msg.data[0x4000];`?

src/converters/fromZigbee.ts
```diff
@@ -3508,17 +3522,18 @@ export const enocean_ptm216z: Fz.Converter = {
...
 };
 // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
-export const _8840100H_water_leak_alarm: Fz.Converter = {
+export const _8840100H_water_leak_alarm: Fz.Converter<"haApplianceEventsAlerts", undefined, "commandAlertsNotification"> = {
     cluster: "haApplianceEventsAlerts",
     type: "commandAlertsNotification",
     convert: (model, msg, publish, options, meta) => {
         const alertStatus = msg.data.aalert;
         return {
+            // @ts-expect-error TODO: wrong type assumed
```
Koenkk: @presslab-us since you introduced the Waxxman 8840100H in https://github.com/Koenkk/zigbee-herdsman-converters/pull/1201, can you confirm it correctly detects the water leak? It seems this converter code is wrong and I'm worried that it always returns `water_leak: false` (even when there is one)

src/devices/xyzroe.ts
Nerivec: `const {switchType} = msg.data;` would match the previous PR:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/db83dc260b8be30c848a3fb538487b32c2b2db96/src/devices/xyzroe.ts#L80-L90

That was probably a typo between using the [actual ZCL prop from ZH](https://github.com/Koenkk/zigbee-herdsman/blob/674a72daa3d1bbf8e9c89fa23dca187889b9c9b8/src/zspec/zcl/definition/cluster.ts#L499-L507) (`switchType`), and using a name that resembles the name of the lookup variable. _If not, should also fix the above code snippet._

xyzroe: Looks like @Nerivec has write. 
I used PTVO external converter to make this converter, because PTVO firmware was used. 
Original converter looks like
```
tz.ptvo_on_off_config = {
    key: ['switch_type', 'switch_actions', 'link_to_output', 'bind_command'],
        await entity.read('genOnOffSwitchCfg', ['switchType', 'switchActions', 0x4001, 0x4002]);
    },
        let payload;
        let data;
        switch (key) {
        case 'switch_type':
            data = getListValueByKey(switchTypesList, value);
            payload = {switchType: data};
            break;
        case 'switch_actions':
            data = getListValueByKey(switchActionsList, value);
            payload = {switchActions: data};
            break;
        case 'link_to_output':
            data = getListValueByKey(inputLinkList, value);
            payload = {0x4001: {value: data, type: 32 /* uint8 */}};
            break;
        case 'bind_command':
            data = getListValueByKey(bindCommandList, value);
            payload = {0x4002: {value: data, type: 32 /* uint8 */}};
            break;
        }
        await entity.write('genOnOffSwitchCfg', payload);
    },
};

fz.ptvo_on_off_config = {
    cluster: 'genOnOffSwitchCfg',
    type: ['readResponse', 'attributeReport'],
        const channel = getKey(model.endpoint(msg.device), msg.endpoint.ID);
        const {switchActions, switchType} = msg.data;
        const inputLink = msg.data[0x4001];
        const bindCommand = msg.data[0x4002];
        return {
            [`switch_type_${channel}`]: getKey(switchTypesList, switchType),
            [`switch_actions_${channel}`]: getKey(switchActionsList, switchActions),
            [`link_to_output_${channel}`]: getKey(inputLinkList, inputLink),
            [`bind_command_${channel}`]: getKey(bindCommandList, bindCommand),
        };
    },
};
```


So basically buttonMode uses switchType

src/converters/fromZigbee.ts
Koenkk: Disabled this device in https://github.com/Koenkk/zigbee-herdsman-converters/pull/9915

src/devices/xyzroe.ts
Koenkk: Fixed, thanks @xyzroe !

Koenkk:
@Nerivec ready for merge?

Nerivec:
That's some synchronized thinking right there 😄
_I updated the top post to highlight the discovered issues for future ref._

PR #9870, opened 2025-08-28
Create inovelliMMWave modern extend to relocate mmWave commands and support future expansion

rohankapoorcom:
We wanted to move the mmWave commands to the bottom of the exposes section. Easiest way to do that was to create another ModernExtend class. This has the benefit of also making it easy to add the additional commands to the mmWave class later on.


Koenkk:
Thanks!

Issue #9837, opened 2025-08-22
TZE200_ytryxh0a and TZE204_ytryxh0a


Grandseb:
I'd like to pull request for Xanlite Tuya thermostatic radiator valves butr I don't know how.
So, here the external converter I'm using

`
const fz = require('zigbee-herdsman-converters/converters/fromZigbee.js');
const tz = require('zigbee-herdsman-converters/converters/toZigbee.js');
const exposes = require('zigbee-herdsman-converters/lib/exposes.js');
const reporting = require('zigbee-herdsman-converters/lib/reporting.js');
const modernExtend = require('zigbee-herdsman-converters/lib/modernExtend.js');
const e = exposes.presets;
const ea = exposes.access;
const tuya = require('zigbee-herdsman-converters/lib/tuya.js');

const definition = {

    fingerprint: [
        {
            modelID: 'TS0601',
            manufacturerName: '_TZE200_ytryxh0a', '_TZE204_ytryxh0a'
        },
    ],
    model: 'TS0601_thermostat',
    vendor: 'Tuya',
    description: 'Radiator valve with thermostat',
	whiteLabel: [
            { vendor: 'Xanlite', model: 'KZTETHEAU' },
        ],
    fromZigbee: [tuya.fz.datapoints],
    toZigbee: [tuya.tz.datapoints],
    onEvent: tuya.onEventSetTime, // Add this if you are getting no converter for 'commandMcuSyncTime'
    configure: tuya.configureMagicPacket,
    exposes: [
            e
                .climate()
                .withPreset(['auto', 'manual', 'holiday',])
		.withLocalTemperature(ea.STATE)
		.withSetpoint('current_heating_setpoint', 5, 30, 0.5, ea.STATE_SET)
                .withLocalTemperatureCalibration(-9, 9, 0.1, ea.STATE_SET),
            e.comfort_temperature().withValueMin(5).withValueMax(30),
            e.eco_temperature().withValueMin(5).withValueMax(30),
            e.holiday_temperature().withValueMin(0).withValueMax(30),
      	    e.window_detection(),
	    e.open_window_temperature().withValueMin(5).withValueMax(25),
	    e.binary('boost_heating', ea.STATE_SET, 'ON', 'OFF'),
             ...tuya.exposes.scheduleAllDays(ea.STATE, 'HH:MM/C'),
	    e.child_lock(),
	    e.battery().withUnit('%'),
	    tuya.exposes.errorStatus(),
    ],
[...]

PR #9871, opened 2025-08-28
Add Tuya Xanlite TZE200_ytryxh0a and TZE204_ytryxh0a

Grandseb:
Hi,
For the issue #9837 I tried to add with this PR for Thermostatic valve radiator Xanlite. I hop it's the good way to do this.
I have an issue with converter for batteryVoltToPercent, it seems not work.

Thanks.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -18504,4 +18504,60 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+		fingerprint: tuya.fingerprint("TS0601", ["_TZE204_ytryxh0a", "_TZE200_ytryxh0a"]),
+		model: 'KZTETHEAU',
+		vendor: 'Xanlite',
+		description: 'Tête thermostatique connectée / Radiator valve with thermostat',
+		fromZigbee: [tuya.fz.datapoints],
+		toZigbee: [tuya.tz.datapoints],
+		onEvent: tuya.onEventSetTime, // Add this if you are getting no converter for 'commandMcuSyncTime'
+		configure: tuya.configureMagicPacket,
+		exposes: [
+            e
+                .climate()
+                .withPreset(['auto', 'manual', 'holiday',])
+				.withLocalTemperature(ea.STATE)
+				.withSetpoint('current_heating_setpoint', 5, 30, 0.5, ea.STATE_SET)
+                .withLocalTemperatureCalibration(-9, 9, 0.1, ea.STATE_SET),
+            e.comfort_temperature().withValueMin(5).withValueMax(30),
+            e.eco_temperature().withValueMin(5).withValueMax(30),
+            e.holiday_temperature().withValueMin(0).withValueMax(30),
+      	    e.window_detection(),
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -18504,4 +18504,60 @@ export const definitions: DefinitionWithExtend[] = [
...
+                [24, "local_temperature", tuya.valueConverter.divideBy10],
+                [45, "error_status", tuya.valueConverter.raw],
+                [101, "comfort_temperature", tuya.valueConverter.divideBy2],
+                [102, "eco_temperature", tuya.valueConverter.divideBy2],
+                [103, "holiday_start_stop", tuya.valueConverter.thermostatHolidayStartStop],
+                [104, "local_temperature_calibration", tuya.valueConverter.localTempCalibration1],
+                [105, "holiday_temperature", tuya.valueConverter.divideBy2],
+                [106, "boost_heating", tuya.valueConverter.onOff],
+                [107, "window_detection", tuya.valueConverter.onOff],
+                [109, "schedule_monday", tuya.valueConverter.thermostatScheduleDayMultiDP2],
```
Koenkk, line 18550:
```suggestion
                [109, "schedule_monday", tuya.valueConverter.thermostatScheduleDayMultiDP],
```

DP2 doesn't exist, can you try with this one

src/devices/tuya.ts
```diff
@@ -18504,4 +18504,60 @@ export const definitions: DefinitionWithExtend[] = [
...
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [
+                    2,
+                    "preset",
+                    tuya.valueConverterBasic.lookup({auto: tuya.enum(0), manual: tuya.enum(1), holiday: tuya.enum(2), confort: tuya.enum(3)}),
+                ],
+                [16, "current_heating_setpoint", tuya.valueConverter.divideBy2],
+                [34, "battery", tuya.valueConverter.batteryVoltToPercent],
```
Koenkk, line 18539:
```suggestion
                [34, "battery", tuya.valueConverter.raw],
```

`batteryVoltToPercent` doesn't exist, can you try with this?

src/devices/tuya.ts
Grandseb: I used the external converter found in the issue #23079 and added the modifications. And it worked with zigbee2mqtt ver 1.4. Now, since the ver 2.5 and the same custom converter, this devices still "not supported" in frontend (before it was OK with external converter). And also I don't have any expose now except lqi signal. Before I had all exposes. 
So I don't know how to try the changes menyionned by you. Could you help me ?
Thanks.

Grandseb: I tried with thermostatScheduleDayMultiDP and it's not OK. So it works with thermostatScheduleDayMultiDP2 and need to add it in tuya.js

src/devices/tuya.ts
Grandseb: batteryVoltToPercent was added in my zigbee-herdsmann-converters/dist/lib/tuya.js to convert volt to percent but it don't work. I try with raw.

PR #9878, opened 2025-08-29
Giexperience patch 2

Giexperience:
Description

This PR adds support for a new Tuya TS0601 soil moisture sensor with the manufacturer name _TZE284_wckqztdq.
It is functionally identical to the already supported GX04Z model, only a new fingerprint is added.

No changes were made to existing functionality.

Added fingerprint: _TZE284_wckqztdq.


Koenkk:
Thanks!

PR #9882, opened 2025-08-29
Remove Unsupported Attributes from CMA30651

dodger421:
Remove unsupported attributes from CMA30651. Supported by ClickSmart+ Hub and app but not by generic Tuya, likely incorrect endpoints.


dodger421:
Ignore, didn't change the powerOnBehaviour2 to false after fixing the typo.

PR #9883, opened 2025-08-29
Remove Unsupported Attributes - CMA30651

dodger421:
CMA30651 Updates
Exposure updates for CMA30651

Removed Child Lock and Backlight On/Off as unsupported features.

Reinstated Countdown and Power-on Behaviour after further testing.


dodger421:
After a lot of further testing locally I've been able to get the Power-On Behaviour to work properly. It only works with the default tuyaOnOff definition, using PowerOnBehaviour2 does not work.

src/devices/tuya.ts
```diff
@@ -2863,11 +2863,11 @@ export const definitions: DefinitionWithExtend[] = [
         model: "CMA30651",
         vendor: "ClickSmart+",
         description: "13A Smart Switched Fused Connection Unit",
+        fromZigbee: [fz.on_off, fz.ignore_basic_report, tuya.fz.on_off_countdown],
+        toZigbee: [tz.on_off, tuya.tz.power_on_behavior_1, tuya.tz.on_off_countdown],
```
Koenkk: Are these from and toZigbee converters really needed? (can you try removing them and see if it still works)

Koenkk:
Thanks!

PR #9889, opened 2025-08-30
PJ-1203A: Add a configuration option to invert reported energy flow direction.

l1berat0r:
Hello,

I'm not sure if the power flow direction reported by the converter accurately reflects the true state (Tuya app reports "Forward" and "Reverse", which can be interpreted arbitrarily, vs "consuming" and "producing" which are objective statements).
I added a simple switch (disabled by default) in the device options, to invert the power flow direction shown if someone likes more the other way.


src/devices/tuya.ts
```diff
@@ -13746,10 +13776,10 @@ export const definitions: DefinitionWithExtend[] = [
                 [102, null, convLocal.energyFlowPJ1203A("a")], // energy_flow_a or the sign of power_a
                 [104, null, convLocal.energyFlowPJ1203A("b")], // energy_flow_b or the sign of power_b
                 [115, null, convLocal.powerAbPJ1203A()],
-                [106, "energy_a", tuya.valueConverter.divideBy100],
```
Koenkk, line 14065: For z2m we always follow this:
- In case a consuming device is connected, e.g. a light, it should show a positive number
- In case a producing device is connected, e.g. solar panels, it should show a negative number.

Without any changes, is this the case?

l1berat0r: If I understand your question correctly, before any changes positive number is shown for both "produced" and "consumed" energy
Zrzut ekranu z 2025-08-31 15-56-42

Koenkk: That should not be the case, can you update the converter to provide a negative value when it is producing energy?

l1berat0r: Does it look correct now?
Zrzut ekranu z 2025-09-12 23-40-45
The commit is pushed to the PR branch.

Koenkk: Yes this looks correct, I don't think we need the option to invert this anymore?

l1berat0r: The manual for the device says, that you should orient your probe in a way, that the forward direction of the energy is towards your devices (consumed energy). This is how I got them connected, but the converter showed the values in the oposite orientation (consumed as produced and vice versa). I wouldn't want to just switch datapoints in the converter, as someone may already be using it and have the probes installed in reverse, but I also wouldn't like to have to mess with the cables again, so simple switch appeared to me as the best option (especially that I believe the original version is wrong).

Koenkk:
Thanks!

Koenkk:
This seems to cause negative values (https://github.com/Koenkk/zigbee2mqtt/issues/28635#issuecomment-3304066123), could you check? (CC: @jerryj001)

jerryj001:
> This seems to cause negative values ([Koenkk/zigbee2mqtt#28635 (comment)](https://github.com/Koenkk/zigbee2mqtt/issues/28635#issuecomment-3304066123)), could you check? (CC: @jerryj001)

Yes the production energy changes from 5000KWH to -5000KWH after the latest-dev update

l1berat0r:
Maybe I understood your first messge wrong, but I thought it should be the case. It's a direct consequence of the last [commit](https://github.com/Koenkk/zigbee-herdsman-converters/pull/9889/commits/2d5796739e118b6a51cdfa0a975f00cba7f70173). If that's wrong it may be reverted to the [previous](https://github.com/Koenkk/zigbee-herdsman-converters/pull/9889/commits/83ac3a9a29216f458d76094a093ca7d6602b4d57) one with just the switch to change directions and no changes to the sign.

jcisio:
After the update, I got the same issue. 192 kWh now becomes -193 kWh and Home Assistant doesn't like it: there is an error and it no longer records injected solar energy. I haven't played with this new option.

jcisio:
The change (bug fixed?) is that before the power could be negative, but the production was always positive.

marcelhoogantink:
So this new option is like changing the orientation of the ct-clamp on the wire...??

jerryj001:
Production changes from Positive 5000KWH to Negative 5000KWH after updating to 2.6.2. It eventually messed up my energy statistics. I am changing back to 2.6.1

Koenkk:
Changing the `invert_energy_flow_a` option in the device setttings (specific) tab should fix this.

2Vince:
Hi, it messed up my energy statistics too.
I've updated Z2m this morning (~9:00). After the update, no produced energy (by solar panels) detected, no energy consumption detected.
I've tried to change invert_energy_flow_a (and b) option to"true" (before 16:00).
But Energy flow a and b are inverted ( when i'm consumming energy, flow is indicating "producing")
And energy statistics registered a 577KWH (!) peak in energy consumption.
capture
I reverted back invert_energy_flow_a (and b) options to false.
But, i don't understand  the logic here.
How could i fix the situation ?

Koenkk:
@l1berat0r I'm a bit confused, is this expected behaviour or not?

Based on:

> I wouldn't want to just switch datapoints in the converter, as someone may already be using it and have the probes installed in reverse

I thought it was not breaking.

jcisio:
I believe the issue here is that energy_produced_(a|b) is negative. It should never be.

l1berat0r:
The switch in it's default setting does not cause any changes to the behavior of the converter. Only when you change it, it replaces produced energy with consumed energy and vice versa.
The option was added in the first [commit](https://github.com/Koenkk/zigbee-herdsman-converters/pull/9889/commits/83ac3a9a29216f458d76094a093ca7d6602b4d57) to address different orientation of a proble.

> But Energy flow a and b are inverted ( when i'm consumming energy, flow is indicating "producing")

This is expected after the change of the option.

Then there was the second [commit](https://github.com/Koenkk/zigbee-herdsman-converters/pull/9889/commits/2d5796739e118b6a51cdfa0a975f00cba7f70173) which caused the sign of the produecd energy value to be switched (and I believe this is the one that people keep complaining about). 

I see 2 confligting points here in the comments. @Koenkk , you mentioned the following:

> In case a producing device is connected, e.g. solar panels, it should show a negative number.

And @jcisio writes:

> I believe the issue here is that energy_produced_(a|b) is negative. It should never be.

I personaly don't know which rule is correct, but reverting the second [commit](https://github.com/Koenkk/zigbee-herdsman-converters/pull/9889/commits/2d5796739e118b6a51cdfa0a975f00cba7f70173) should bring things back to normal for people that are experiencing the issues.

jcisio:
When you configure Energy in Home Assistant, if "Return to Grid" has negative value, HA apparently doesn't like it.

jerryj001:
I just tried to switch invert_energy_flow_a option in the device setttings (specific) tab to True (later tried to switch to False). The production energy shows negative. And like what jcisio said HA's energy dashboard configuration doesn't like negative energy as shown below
Screenshot 2025-10-02 232408
Screenshot 2025-10-02 232657

KingOfIce77:
I confirm, same issue for me

Koenkk:
@l1berat0r 

> In case a producing device is connected, e.g. solar panels, it should show a negative number.

I underestimated the impact, my apologies, @l1berat0r do you mind submitting a pr to fix this? (so keep old behaviour + the option?)

l1berat0r:
PR submited.

PR #9899, opened 2025-09-01
Inovelli mmwave additions

rohankapoorcom:
Implements support for [most of the advanced functionality](https://help.inovelli.com/en/articles/12773613-blue-series-mmwave-presence-dimmer-switch-advanced-mmwave-configuration) for Inovelli's VZM32 Blue Series mmWave Presence Dimmer Switch.

* Allows configuring 4 detection areas
* Allows configuring 4 stay areas
* Allows configuring 4 interference areas
* Reports the configuration of all 4 configured areas for each type on initial configuration and whenever a value is changed
* Reports occupancy status of all 4 detection areas

Co-authored-by: @ccutrer

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -812,6 +812,16 @@ export const metering: Fz.Converter<"seMetering", undefined, ["attributeReport",
             const property = postfixWithEndpointName("energy_tier_2", msg, model, meta);
             payload[property] = value * (factor ?? 1);
         }
+        if (msg.data.currentTier3SummDelivered !== undefined) {
+            const value = msg.data.currentTier3SummDelivered;
+            const property = postfixWithEndpointName("energy_tier_3", msg, model, meta);
+            payload[property] = value * (factor ?? 1);
+        }
+        if (msg.data.currentTier4SummDelivered !== undefined) {
+            const value = msg.data.currentTier4SummDelivered;
+            const property = postfixWithEndpointName("energy_tier_4", msg, model, meta);
+            payload[property] = value * (factor ?? 1);
+        }
         if (msg.data.currentTier1SummReceived !== undefined) {
             const value = msg.data.currentTier1SummReceived;
             const property = postfixWithEndpointName("produced_energy_tier_1", msg, model, meta);
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -2079,6 +2079,38 @@ export const currentsummdelivered: Tz.Converter = {
         await ep.read("seMetering", ["currentSummDelivered"]);
     },
 };
+export const currenttier1summdelivered: Tz.Converter = {
+    key: ["energy_tier_1"],
+    convertGet: async (entity, key, meta) => {
+        utils.assertEndpoint(entity);
+        const ep = determineEndpoint(entity, meta, "seMetering");
+        await ep.read("seMetering", ["currentTier1SummDelivered"]);
+    },
+};
+export const currenttier2summdelivered: Tz.Converter = {
+    key: ["energy_tier_2"],
+    convertGet: async (entity, key, meta) => {
+        utils.assertEndpoint(entity);
+        const ep = determineEndpoint(entity, meta, "seMetering");
+        await ep.read("seMetering", ["currentTier2SummDelivered"]);
+    },
+};
+export const currenttier3summdelivered: Tz.Converter = {
+    key: ["energy_tier_3"],
+    convertGet: async (entity, key, meta) => {
+        utils.assertEndpoint(entity);
+        const ep = determineEndpoint(entity, meta, "seMetering");
+        await ep.read("seMetering", ["currentTier3SummDelivered"]);
+    },
+};
+export const currenttier4summdelivered: Tz.Converter = {
+    key: ["energy_tier_4"],
+    convertGet: async (entity, key, meta) => {
+        utils.assertEndpoint(entity);
+        const ep = determineEndpoint(entity, meta, "seMetering");
+        await ep.read("seMetering", ["currentTier4SummDelivered"]);
+    },
+};
 export const currentsummreceived: Tz.Converter = {
     key: ["produced_energy"],
     convertGet: async (entity, key, meta) => {
--- a/src/devices/lds.ts
+++ b/src/devices/lds.ts
@@ -9,4 +9,11 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Ynoa smart LED E27",
         extend: [m.light({colorTemp: {range: [153, 555]}, color: true})],
     },
+    {
+        zigbeeModel: ["FWBulb03UK"],
+        model: "FWBulb03UK",
+        vendor: "LDS",
+        description: "Hive Smart Light Bulb E27 Dimmable",
+        extend: [m.light()],
+    },
 ];
--- a/src/devices/slacky_diy.ts
+++ b/src/devices/slacky_diy.ts
@@ -1,10 +1,11 @@
+import type {ClusterOrRawAttributeKeys, PartialClusterOrRawWriteAttributes} from "zigbee-herdsman/dist/controller/tstype";
 import * as fz from "../converters/fromZigbee";
 import * as tz from "../converters/toZigbee";
 import * as exposes from "../lib/exposes";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
-import * as globalStore from "../lib/store";
-import type {Definition, DefinitionWithExtend, Expose, Fz, KeyValue, KeyValueAny, ModernExtend, Tz, Zh} from "../lib/types";
+//import * as globalStore from "../lib/store";
+import type {Configure, Definition, DefinitionWithExtend, Expose, Fz, KeyValue, KeyValueAny, ModernExtend, Tz, Zh} from "../lib/types";
 import * as utils from "../lib/utils";
 
 //import {logger} from "../lib/logger";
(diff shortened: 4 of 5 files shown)


rohankapoorcom:
Not stale, I just haven't had a chance to get back to this

src/devices/inovelli.ts
```diff
@@ -2422,13 +2488,88 @@ const exposeMMWaveControl = () => {
...
         .enum("notificationComplete", ea.STATE, Object.values(LED_NOTIFICATION_TYPES))
         .withDescription("Indication that a specific notification has completed.")
         .withCategory("diagnostic");
 };
 
+const exposeMMWaveAreas = () => {
+    return [
+        e
+            .binary("Area1Occupancy", ea.STATE, "true", "false")
+            .withProperty("mmwave_areas.area1")
```
rohankapoorcom: Exposing nested properties doesn't seem to work. Is there a better way to do these rather than polluting the main state object with 4 separate properties?

src/devices/inovelli.ts
```diff
@@ -2429,6 +2737,105 @@ const exposeLedEffectComplete = () => {
         .withCategory("diagnostic");
 };
 
+const exposeMMWaveAreas = () => {
+    return [
+        e
+            .binary("Area1Occupancy", ea.STATE, true, false)
```
rohankapoorcom: These 4 new properties don't visually look like the Occupancy property nor do the set the appropriate device class in Home Assistant. See the screenshots.

I believe [this](https://github.com/Nerivec/zigbee2mqtt-windfront/pull/400) will fix the frontend and [this](https://github.com/Koenkk/zigbee2mqtt/pull/30712) will fix Home Assistant.

Please let me know if I missed anything.

Image

Image

src/devices/inovelli.ts
```diff
@@ -2429,6 +2737,105 @@ const exposeLedEffectComplete = () => {
...
+            e.numeric("depth_min", ea.STATE_SET).withValueMin(0).withValueMax(600).withDescription("Defines the area in front of the switch"),
+        )
+        .withFeature(
+            e.numeric("depth_max", ea.STATE_SET).withValueMin(0).withValueMax(600).withDescription("Defines the area in front of the switch"),
+        );
+};
+
+const exposeDetectionAreas = () => {
+    return e
+        .composite("Detection Areas", "mmwave_detection_areas", ea.STATE_SET)
```
rohankapoorcom: This nested composite is causing a weird visual behavior where there is a separate apply button below each of the areas and then an overall one at the bottom of the section. (See screenshot).

This is problematic because the individual apply buttons don't actually do anything and only the one at the bottom submits the values.

How do I make the individual ones go away?

Image

rohankapoorcom:
@Koenkk This is ready for review now. I left a couple of comments with questions for you above.

@InovelliUSA please take a look when you have a chance.

ccutrer:
So am I reading this right that if you use any of the "advanced" commands to configure an area, it automatically binds the mmWave cluster?

rohankapoorcom:
> So am I reading this right that if you use any of the "advanced" commands to configure an area, it automatically binds the mmWave cluster?

It automatically binds the mmWave cluster during the configuration process or if you run a reconfigure.

src/devices/inovelli.ts
Koenkk:
```suggestion
            .binary("occupancy_area_1", ea.STATE, true, false)
```

(same for the others)

src/devices/inovelli.ts
Koenkk: I would not make a nested composite here, maybe a composite per area?

src/devices/inovelli.ts
```diff
@@ -2429,6 +2737,105 @@ const exposeLedEffectComplete = () => {
...
+        .withFeature(createAreaComposite("Area1").withProperty("Area1"))
+        .withFeature(createAreaComposite("Area2").withProperty("Area2"))
+        .withFeature(createAreaComposite("Area3").withProperty("Area3"))
+        .withFeature(createAreaComposite("Area4").withProperty("Area4"))
+        .withCategory("config");
+};
+
+const exposeStayAreas = () => {
+    return e
+        .composite("Stay Areas", "mmwave_stay_areas", ea.STATE_SET)
```
Koenkk:
```suggestion
        .composite("stay_areas", "mmwave_stay_areas", ea.STATE_SET)
```

src/devices/inovelli.ts
```diff
@@ -2429,6 +2737,105 @@ const exposeLedEffectComplete = () => {
...
+        .withFeature(createAreaComposite("Area1").withProperty("Area1"))
+        .withFeature(createAreaComposite("Area2").withProperty("Area2"))
+        .withFeature(createAreaComposite("Area3").withProperty("Area3"))
+        .withFeature(createAreaComposite("Area4").withProperty("Area4"))
+        .withCategory("config");
+};
+
+const exposeInterferenceAreas = () => {
+    return e
+        .composite("Interference Areas", "mmwave_interference_areas", ea.STATE_SET)
```
Koenkk:
```suggestion
        .composite("inference_areas", "mmwave_interference_areas", ea.STATE_SET)
```

src/devices/inovelli.ts
ccutrer: It's not "detection, interference, and stay" for area 1. It's detection areas 1-4, interferences areas 1-4 (that apply to all detection areas), etc.

src/devices/inovelli.ts
rohankapoorcom: Same comment from here: https://github.com/Koenkk/zigbee2mqtt/pull/30712#discussion_r2709718854

> All of the custom exposed in Inovelli's converters are Pascal case though. So this would mean that some are snake cased while the majority of them are Pascal cased on the same device. With the devices each providing more than 100 entities, that seems like a weird setup.

I don't want to change everything over to snake_case since that will break existing users and mixed between snake and pascal case will confuse users.

Koenkk:
Will merge once @InovelliUSA reviewed..

src/devices/inovelli.ts
rohankapoorcom: In theory, it could be turned into 3x4 composites (1-4 for each of detection, interference stay areas) but that means:
* The description text has to be copied 4 times which will also not look great
* The data has to be stored separately for each of the composites since composites are the only way to access nested data in an expose (that I've found).
* Each of those exposes will be exposed separately to Home Assistant as a diagnostic entity (which is basically useless).

Is there a way to hide the inner apply button for nested composites? That seems like a less "kludgy" approach.

rohankapoorcom: Tagging @Nerivec as well since they developed the new Windfront frontend and may have an idea of how to solve this.

Nerivec: https://github.com/Nerivec/zigbee2mqtt-windfront/blob/6a79cb78a69cadc4dd333f9a73e309d2e0815228/src/components/features/FeatureSubFeatures.tsx#L22-L37
This is the logic that's driving the "showing" of the "Apply" button (same logic also for `List`).

Changing the supported nesting depth is a bit dicey, several things are "derived" from it (like [payload building](https://github.com/Nerivec/zigbee2mqtt-windfront/blob/6a79cb78a69cadc4dd333f9a73e309d2e0815228/src/components/features/FeatureSubFeatures.tsx#L67-L73)). But if you want to tackle it, no problem, there's definitely room for improvements around that topic (it's still the behavior of the old frontend mostly...).
_I know over at romasku/tuya-zigbee-switch they wanted a better way to display nested stuff as well. Haven't had time to look into it though._

rohankapoorcom: Thanks - As far as I can tell this fixes it: https://github.com/Nerivec/zigbee2mqtt-windfront/pull/404, but I'm not a frontend expert by far.

src/devices/inovelli.ts
Nerivec: Should fix the case on these too (2 above and this one)
```diff
-        .composite("Stay Areas", "mmwave_stay_areas", ea.STATE_SET)
+        .composite("mmwave_stay_areas", "mmwave_stay_areas", ea.STATE_SET)
```
Goes from:
```json

        "name": "Stay Areas",
        "label": "Stay Areas",
        "access": 3,
        "type": "composite",
        "property": "mmwave_stay_areas",
```
to:
```json
        "name": "mmwave_stay_areas",
        "label": "Mmwave stay areas",
        "access": 3,
        "type": "composite",
        "property": "mmwave_stay_areas",
```

The `mmwave` prefix isn't strictly necessary (I don't think...), but some have it, some don't... Might be worth a full review to align behavior in a separate PR. Should be simple enough with a script to mass-check `property` case, ensure it's snake, and apply it to `name` as well.
Both `name` and `property` should try to respect the case as much as possible since they are not just used for display (can't guarantee everything will work properly with any other case than snake, really). Even using camel case breaks some logic on ZHC side (like deriving `label`).

rohankapoorcom: > Both name and property should try to respect the case as much as possible since they are not just used for display (can't guarantee everything will work properly with any other case than snake, really). Even using camel case breaks some logic on ZHC side (like deriving label)

Sounds like something that needs to be added to the docs and / or enforced by the linter. I can tell you that there's a lot more converters than Inovelli's that are using camelCase.

Nerivec: It's in the works! :wink: https://github.com/Koenkk/zigbee-herdsman-converters/pull/11370

InovelliUSA:
I have checked this out from an interface point of view and it looks good to me!

Koenkk:
Thanks!

PR #9907, opened 2025-09-02
Update tuya.ts

Giexperience:
fix(tuya): move `_TZE284_wckqztdq` signature to the correct device definition

Previously I mistakenly added this signature under the wrong device block in `tuya.ts`.
This PR relocates the signature to the intended device (same functionality as the sibling model),
no behavior change and no breaking changes.


Koenkk:
Thanks!

PR #9927, opened 2025-09-05
Fix for PR #9468: Added ShellyWiFi interface to address Type issue for properties

jactmp:
Proposed fixes for #9468 (PR on top of #9468’s changes).

Fixed compilation error  "TS2322: Type 'string' is not assignable to type '{ ID: number; type: number; }'."
- added `nterface ShellyWiFi`
- added `<"shellyWiFiSetupCluster", ShellyWiFi>` to custom entities.

Fell free to reject my input. It's first PR ever.
~

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -186,7 +186,7 @@ namespace ShellyZigbee {
                         ),
                     );
 
-                    return {state: return_attributes};
+                    return {state: {wifi: return_attributes}};
                 },
                 convertGet: async (entity, key, meta) => {
                     const queryScope = meta.message[key];


src/devices/shelly.ts
```diff
@@ -1,34 +1,247 @@
...
+                    nameServer: {ID: 0x000a, type: Zcl.DataType.CHAR_STR},
+                },
+                commands: {},
+                commandsResponse: {},
+            }),
+            m.text<"shellyWiFiSetupCluster", ShellyWiFi>({
+                cluster: "shellyWiFiSetupCluster",
+                access: "STATE_GET",
+                description: "Wifi Status",
+                name: "Wifi_Status",
```
```suggestion
                name: "wifi_status",
```

Can you make all of the `snake_case`?

src/devices/shelly.ts
```diff
@@ -1,34 +1,247 @@
...
+                entityCategory: "diagnostic",
+                zigbeeCommandOptions: {
+                    profileId: ZSpec.CUSTOM_SHELLY_PROFILE_ID,
+                },
+            }),
+            m.text<"shellyWiFiSetupCluster", ShellyWiFi>({
+                cluster: "shellyWiFiSetupCluster",
+                access: "STATE_GET",
+                description: "IP of the device",
+                name: "IP",
```
```suggestion
                name: "ip",
```

src/devices/shelly.ts
jactmp: Sure. Some suggestions needed though. 
Shelly originally uses such naming for attributes:
```
....
7	StaticIP	character string	read/write
8	NetMask	    character string	read/write
9	Gateway	    character string	read/write
10	NameServer	character string	read/write
```
Those name where used as descriptions and attribute's names (lowercased). Cool. Now, the question is what's the proper name for:
7/ `static_ip`\* or `staticip` ?
8/ `net_mask` or `netmask`\* ?
10/ `name_server` or `nameserver`\* ?

EDIT: My personal preferences (*).

jactmp:
Seems like, contrary to what i remembered from ifconfig output, use of net_mask, and name_server with underline inside is quite common.  So I updated attributes to snake_case (including net_mask, name_server) in devices/shelly.ts as requested (commit 79d5a699e) + resolved divergence with a merge commit (ec3855b). Ready for review.

jactmp:
I've just noticed a problem. "action" name cannot be used as its sort of reserved name in z2m? I see in GUI:
_Action
Triggered action (e.g. a button click)
N/A_
instead of Apply/Reset buttons. 
I replaced "action" with "action_" and problem is fixed. Of course final name should be different - something more clean and close to original name.

jactmp:
Ok, I noticed more problems. Attributes read/written using GUI or mqtt are retained and published through MQTT any time some state changes. For most attributes it's OK, but not always. 

Problems:
- `password`: this fileld should NEVER be published to MQTT in my opinion - because, well, it's password, and secondly it's part of device state that is write only according to Shelly's specs,
- `action`: similar story. It's write-only "one-time" action, not state, and should not be pusblished to mqtt or at least should not be cached in the system (sort of one only publish). 

For now I have no idea how to address such problems. Is it even possible to make z2m not sending certain attributes to mqtt and/or not retain it's states? 
I thing feature should not be merged to stable branch until problems above are solved.

EDIT: I force removed commits for now.

jactmp:
I don't get it. For definition
```
            m.text<"shellyWiFiSetupCluster", ShellyWiFi>({
                cluster: "shellyWiFiSetupCluster",
                access: "SET",
                description: "Wifi Password",
                name: "password",
                attribute: "password",
                entityCategory: "config",
                validate(value: unknown) {
                    assertString(value);
                },
                zigbeeCommandOptions: {
                    profileId: ZSpec.CUSTOM_SHELLY_PROFILE_ID,
                },
            }),
```
where `access` is `SET` (not `STATE_SET`), as soon as password is set, it is cached and send in every MQTT concerning Shelly device afterwards. No way how to address it and why it is as it is. Seems like this is  blocking point for merging WiFi feature to master branch? I can push updated PR with snake_case names ofc (I changed name for Action attribute to `actions` to address reserved name problem).

src/devices/shelly.ts
Koenkk: I agree with your personal preferences

Koenkk:
> I've just noticed a problem. "action" name cannot be used as its sort of reserved name in z2m? I see in GUI: _Action Triggered action (e.g. a button click) N/A_ instead of Apply/Reset buttons. I replaced "action" with "action_" and problem is fixed. Of course final name should be different - something more clean and close to original name.

`action` is indeed reserved for things like button presses, please use a different name.


> password: this fileld should NEVER be published to MQTT in my opinion - because, well, it's password, and secondly it's part of device state that is write only according to Shelly's specs,

Can you try changing this line:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/modernExtend.ts#L2684

to:

```js
if (access !== ea.SET) {
return {state: {[key]: value}};
}

**Comment by jactmp** [comment]
I implemented your proposal but modified (not returning anything results in error - no response). I added
```ts
if (access === ea.SET) {
                          return {state: {[key]: undefined}};
} else {
                          return {state: {[key]: value}};
}
```
to 
`export function text<...` (password attribute )
and
`export function numeric<...` (action attribute)

Some initial tests after this modification. 
- password is not published in mqtt,
- password disappears from GUI after refresh 

Looks good.

Koenkk:
> return {state: {[key]: undefined}};

This shouldn't be needed as only in your case the password was present in the `state.json` (due to earlier publishes)

jactmp:
Ok. Seems like I got sort of random error yesterday. I made some tests today again. I cleared cache (deleted `state.json`) every time before starting z2m. I added 
```ts
                          if (access !== ea.SET) {
                              return {state: {[key]: value}};
                          }
```
to `text` function. It works as expected - no password in MQTT nor in state.json (after stoping z2m). 
Actually i added conditional return to `enumLookup` function as well. Otherwise `"actions" : "Apply"` is published in MQTT message, which makes no sense it seems, as it's SET (one time write-only) attribute as well, not permanent "current-device-state" information.

So, what's next steps? I noticed another small issue, but it could be addressed later (not critical): ...

EDIT

... when accidentally entering one of text filelds (e.g. by pressing Tab) and leaving in afterwards (Tab or mouse press outside that field) z2m is trying to send empty string to device, which results in error 10 seconds later. I already added some code to at least address empty string problem.
```ts
const validationFunctions = {
    validateStringNotEmpty(value: unknown): void {
        assertString(value);
                
        if (value.length == 0) {
            throw new Error('Empty string, ignoring.');
        }       
    },          
                    
    // SSID/IP/netMask/nameServer validation functions? Are they really needed?
};
```

Seems, like similiar utility doesn't exits in `utils.ts`. Now user is informed immediately that empty string is being ignored.

Koenkk:
Good! Then I guess https://github.com/Koenkk/zigbee-herdsman-converters/pull/9927/files#r2325769878 is still left.

jactmp:
Yes. Current status:

Changes to`shelly.ts` not yet added to PR:
1/ snake_case names (with `Action` changed to `actions`, as `action` is reserved)
2/ add `validationFunctions` and apply `validateStringNotEmpty` to all text attributes.
I can push them, but ...

Problem:
3/ `actions` and `password` are SET only, but still cached and published in MQTT in every status message.
You provided solution.  It requires changes to core functionality `modernExtend`. At least to `text` and `enumLookup`, probably to all other types for consistency? Challenging stuff as it may affect device handlers that use this library. Definitely not task for me - junior contributor. 

So - what you suggest? Seems like pushing the feature to master branch without solving the problem 3/ is not quite good idea? 

EDIT

Alternative solution (in case `modernExtend` is not to be updated soon) would be to use custom handler based on `modernExtend` code. Possible but some skills needed (still learning).

shaysudm:
@jactmp 
can you please also finish the implementation for the RPC cluster?

jactmp:
> @jactmp can you please also finish the implementation for the RPC cluster?

RPC cluster exposes low-level access to Shelly's RPC subsystem. It is designed to be "covered" with some higher level logic (write payload's length to `TxCtl`, write payload to `Data`, check `RxCtl` for number > 0, read response from `Data`). Possible, but not trivial (custom handler required). Looking into this. For now you can use DevConsole to play with it.

Koenkk:
> 3/ actions and password are SET only, but still cached and published in MQTT in every status message.
You provided solution. It requires changes to core functionality modernExtend. At least to text and enumLookup, probably to all other types for consistency? Challenging stuff as it may affect device handlers that use this library. Definitely not task for me - junior contributor.

Please include the changes in this PR

jactmp:
I faced some other problems with Shelly's custom WiFi cluster and decided to write high-level support from scratch. It has some TODOs (details below). 

1/ I moved wifi to separate composite - I thought that's what composite was created for, and added custom handler
2/ MQTT payload now looks like: `{"linkquality":69,"state":"OFF","wifi":{"dhcp":true,"enable":true,"ip":"192.168.154.203","ssid":"my_ssid","status":"got ip"}}`
3/ `Action` attribute is not directly mapped to `exposes` template, I added some logic on top
4/ I disabled static stuff for now - Shelly rejects empty string written to those attributes - no idea how to clear those attributes
5/ `action_wifi` used for attribute `0x02` temporarily as `action` is reserved
6/ I tried to make read/write to many attributes in one command, but it seems like payload is too long sometimes and Shelly doesn't accept it so I split it into groups choosing split limit manually ("because now it works").
7/ fromZigbee updates GUI only after first read request, for next reads I have to refresh browser's tab to see results - bug in my code or known feature?

Im working on adding RPC support now, and investigating and trying to solve issues. I'm quite new to GitHub workflow (and JS/TS languages) so don't know whether this code even fits in this PR. 

EDIT: 
For point 4. I wonder, is it a bug in Shelly firmware or is there a special way to clear string type attributes in Zigbee?
For point 6. Question is whether there's efficient algorithms do estimate safe limit for read/write request bundles.

jactmp:
GUI (default) looks now like below:
wifi

PR #9934, opened 2025-09-06
feat: add support for TS020C_TZ3040_wc6kfjtc PIR motion sensor

iMouath:
Link to picture pull request: [here](https://github.com/Koenkk/zigbee2mqtt.io/pull/4117)

This Adds support to generic pir motion sensor with illuminance
had to write a custom converter for the illuminance since datapoint 12 didn't work even tho from tuya iot platform the exposed datapoints were

```json
{"1":"PIR state","4":"Battery level","12":"Illuminance value","101":"PIR sensitivity","102":"Someone hold time","103":"Light sensing collection interval"}
```

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -921,16 +921,13 @@ const fzLocal = {
         undefined,
         ["commandDataResponse", "commandDataReport", "commandActiveStatusReport", "commandActiveStatusReportAlt"]
     >,
-    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
-    TS020C_illuminance: {
+    ts020cIlluminance: {
         cluster: "manuSpecificTuya_2",
         type: ["attributeReport"],
         convert: (model, msg, publish, options, meta) => {
             const result: KeyValue = {};
-            if (["_TZ3040_wc6kfjtc"].includes(meta.device.manufacturerName)) {
-                if ("57345" in msg.data) {
-                    result.illuminance = msg.data["57345"];
-                }
+            if ("57345" in msg.data) {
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -704,6 +704,19 @@ const tzLocal = {
             return ret;
         },
     } satisfies Tz.Converter,
+    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
+    TS020C_illuminance: {
```
Koenkk: Illuminance seems like read only thing, why is there a write for this?

src/devices/tuya.ts
```diff
@@ -934,6 +961,63 @@ export const definitions: DefinitionWithExtend[] = [
         toZigbee: [],
         exposes: [e.gas(), e.tamper()],
     },
+    {
+        fingerprint: tuya.fingerprint("TS020C", ["_TZ3040_wc6kfjtc"]),
+        zigbeeModel: ["TS020C"],
+        model: "TS020C",
+        vendor: "Tuya",
+        description: "PIR sensor",
+        fromZigbee: [fz.ias_occupancy_alarm_1, fz.battery, tuya.fz.datapoints, fzLocal.TS020C_illuminance],
```
```suggestion
        fromZigbee: [ tuya.fz.datapoints, fzLocal.TS020C_illuminance],
```

Does it also work with this?

iMouath: sure does :), thought those were necessary from the generated external def

src/devices/tuya.ts
iMouath: You're right. will remove it.

Koenkk:
Thanks!

Issue #9900, opened 2025-09-01
Please add new fingerprint _TZ3000_afgzktgb for Moes ZM-104-M  tuya.ts


curtash:
sorry I don't how to do it myself. 

I created a converter and tested and it works. Before I could not use a momentary switch but now I can.

bernardesarthur:
Hey Curtash, I'll do that for you. Thanks a lot!

PR #9937, opened 2025-09-06
Fix detection: _TZ3000_afgzktgb as Moes ZM-104-M

bernardesarthur:
Fix detection: _TZ3000_afgzktgb as Moes ZM-104-M

Reference to https://github.com/Koenkk/zigbee-herdsman-converters/issues/9900


Koenkk:
Thanks!

andrei-lazarov:
Sorry, why not add it as a whitelabel (+ picture/doc) if it has a recognizable brand and model?

bernardesarthur:
I saw that device has a white label tag but not using a Tuya white label. I'll find the picture for these device and fix that. Thanks for you observation, I didn't realize that.

PR #9943, opened 2025-09-07
feat(Bosch): Enable OTA updates for BWA-1, BSIR-EZ and RFPR-ZB-SH-EU

DerDreschner:
Bosch released all OTA files within their Smart Home Controller in recent weeks. Some devices have already got the necessary `ota: true` flag. This pull request adds it to all devices that have an OTA update released, but haven't had the flag yet.


Koenkk:
Thanks!

PR #9944, opened 2025-09-07
Recognize _TZE284_l8xiyymq as PN6 6-way controller

gyfis:
Bought from https://www.aliexpress.com/item/1005008082245864.html


Koenkk:
Thanks!

arevatech:
I have same Controller, but it show me unsupported. How to fix that ?

PR #9952, opened 2025-09-08
Fix detection: _TZ3000_pgq7ormg as TS0001_switch_module

andrei-lazarov:
Hello! First time contributing here, so I have some questions.

I added my device ([1-gang switch module](https://github.com/romasku/tuya-zigbee-switch/issues/105#issuecomment-3262595941)) as **TS0001_switch_module**.  
It does not have a recognizable name - so I don't think a custom page/picture is necessary.

All features work: switch type, power-on behavior, countdown. 

But the device also has a blue LED that indicates the relay state (and blinks when in pairing mode).  
**I can control the LED behaviour with `indicatorMode`**, although a power-cycle is required for the setting to apply.  

I assume this is also the case for the other switches, so is there a reason why it's omitted?  
**Should I create a new converter with all 4 options?**

---

Another question: **Can I 'link' my Z2M instance to my local clone of this repo?**

I cloned both Z2M and converters repos on my PC for testing and developing.  
When adding new devices/detections: I think it would be easier to **edit the converters in place**, instead of creating custom external converters. Is this possible?  
Thanks!


bernardesarthur:
I assume this is also the case for the other switches, so is there a reason why it's omitted?
**Should I create a new converter with all 4 options?**

Maybe the guy created this doesn't tested this option but you can create a one. 

**Another question: Can I 'link' my Z2M instance to my local clone of this repo?**

I don't know if is that possible using a default installation but that is a good idea.

PR #9958, opened 2025-09-08
Added: covers mode (SBDV-00199), diagnostics (SBDV-00196,00199,00202)

avzasorin-sd:
Adding new features for devices from SDevices:

- covers mode (SBDV-00199)
- child lock (SBDV-00196, SBDV-00199)
- diagnostics and serial number (SBDV-00196, SBDV-00199, SBDV-00202)

PR with additional information about switching SBDV-00199 into covers mode: https://github.com/Koenkk/zigbee2mqtt.io/pull/4126

### DIFF
--- a/src/devices/sber.ts
+++ b/src/devices/sber.ts
@@ -2,18 +2,36 @@ import {Zcl} from "zigbee-herdsman";
 import * as fz from "../converters/fromZigbee";
 import * as tz from "../converters/toZigbee";
 import * as exposes from "../lib/exposes";
+import {logger} from "../lib/logger";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
 import * as tuya from "../lib/tuya";
-import type {DefinitionWithExtend, Fz, KeyValue, KeyValueAny, ModernExtend, Tz} from "../lib/types";
+import type {DefinitionWithExtend, Fz, KeyValue, KeyValueAny, ModernExtend, Tz, Zh} from "../lib/types";
 import * as utils from "../lib/utils";
 
 const e = exposes.presets;
 const ea = exposes.access;
 
+const NS = "zhc:sdevices";
 const {tuyaMagicPacket, tuyaOnOffActionLegacy} = tuya.modernExtend;
 const manufacturerOptions = {manufacturerCode: Zcl.ManufacturerCode.SBERDEVICES_LTD};
 const defaultResponseOptions = {disableDefaultResponse: false};
+
+function checkMetaOption(device: Zh.Device, key: string) {
+    if (device != null) {
+        const enabled = device.meta[key];
+        if (enabled === undefined) {
+            return false;
+        }
+        return !!enabled;
+    }
+    return false;
+}
+function setMetaOption(device: Zh.Device, key: string, enabled: boolean) {
+    if (device != null && key != null) {
+        device.meta[key] = enabled;
+    }
+}
 
 const sdevices = {
     fz: {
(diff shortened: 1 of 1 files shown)


src/devices/sber.ts
```diff
@@ -106,11 +124,74 @@ const sdevices = {
                 return result;
             },
         } satisfies Fz.Converter<"manuSpecificSDevices", SberDevices, ["attributeReport", "readResponse"]>,
+        cover_position_tilt: {
+            cluster: "closuresWindowCovering",
+            type: ["attributeReport", "readResponse"],
+            options: [exposes.options.invert_cover()],
+            convert: (model, msg, publish, options, meta) => {
+                const result: KeyValueAny = {};
```
Koenkk, line 117: This adds some code duplication, can you call `tz.cover_position_tilt()` and modify it's result?

src/devices/sber.ts
```diff
@@ -2,19 +2,37 @@ import {Zcl} from "zigbee-herdsman";
...
+        const enabled = device.meta[key];
+        if (enabled === undefined) {
+            return false;
+        }
+        return !!enabled;
+    }
+    return false;
+}
+function setMetaOption(device: Zh.Device, key: string, enabled: boolean) {
+    if (device != null && key != null) {
```
Koenkk: I think this check is not needed?

src/devices/sber.ts
```diff
@@ -2,19 +2,37 @@ import {Zcl} from "zigbee-herdsman";
...
+        device.meta[key] = enabled;
```
Koenkk:
```suggestion
        device.meta[key] = enabled;
        device.save();
```

src/devices/sber.ts
```diff
@@ -2,19 +2,37 @@ import {Zcl} from "zigbee-herdsman";
...
 const e = exposes.presets;
 const ea = exposes.access;
 
+const NS = "zhc:sdevices";
 const {tuyaMagicPacket, tuyaOnOffActionLegacy} = tuya.modernExtend;
 const manufacturerOptions = {manufacturerCode: Zcl.ManufacturerCode.SBERDEVICES_LTD};
 const defaultResponseOptions = {disableDefaultResponse: false};
 
+function checkMetaOption(device: Zh.Device, key: string) {
+    if (device != null) {
```
Koenkk: This check should not be needed? I think this function can be simplified to just `return !!(device.meta[key] ?? false)`

src/devices/sber.ts
avzasorin-sd: done. Also got rid of extra functions

src/devices/sber.ts
avzasorin-sd: I think you are right, this additional check was an artefact from times when dummyDevice was not implemented yet

src/devices/sber.ts
avzasorin-sd: OK

src/devices/sber.ts
avzasorin-sd: I am afraid I do not see how I can achieve this.
The goal was to implement a simplified version of `cover_position_tilt,` because the device does not support some features of `cover_mode` (`calibration`, `maintenance`, `led`), so I do not want them to appear in exposes.

Koenkk: The tz converter doesn't impact what shows up in exposes, this is controlled through `exposes:` on the device definition

avzasorin-sd: OK, but the line you are referring to is from `fz`, not `tz`.
For TZ this converter already uses default `tz.cover_position_tilt`.
But for FZ I had to use a modified copy of `fz.cover_position_tilt` indeed.

After firmware update SBDV-00199 became a very complex device, which can be paired either as a two-button switch, or a window covers controller.
To implement such a complex functionality in a single converter (because model_id stays the same), I had to rename the `state_cover` expose for 3rd endpoint, there the windows cover control is implemented, to `current_state_cover`. During my research, wnen I tried to use non-modified `state_cover` expose name for EP3, I've encountered the following error when trying to change it via frontend:

"z2m: Error: 'close' not allowed, choose between: toggle,off,on"

Possibly this was happening, because converter also has `state_switch_1` and `state_switch_2` exposes for endpoints EP1 and EP2, but those `states` are of `OnOff` type, and for the 3rd endpoint `state_cover` is the state of windows cover. So the following set of exposes:

`state_switch_1`
`state_switch_2`
`state_cover`

did not work, but the following set of exposes worked well:

`state_switch_1`
`state_switch_2`
`current_state_cover`

That is why I had to copy `fz.cover_position_tilt`, `tz.cover_mode` and `tz.cover_state` code, and make small changes to use custom names of exposes instead of default ones.

Koenkk:
Thanks!

Issue #9670, opened 2025-07-22
Conversion error of received data with sonoff swv devices


liangjia2019:
We encountered some problems when using SWV. When the value received by the data exceeds 127, errors will occur in the data conversion

Received here is correct:
[2025-07-22 19:09:26] debug: 	zh:zstack:znp: <-- AREQ: AF - incomingMsg - {"groupid":0,"clusterid":64529,"srcaddr":28490,"srcendpoint":1,"dstendpoint":1,"wasbroadcast":0,"linkquality":180,"securityuse":0,"timestamp":7791435,"transseqnumber":0,"len":17,"data":{"type":"Buffer","data":[8,52,10,9,80,66,10,1,11,0,0,0,127,0,0,0,144]}}

But here his data conversion is abnormal
attrData":"\u0001\u000b\u0000\u0000\u0000\u0000\u0000\u0000 "  ---was  abnormal
[2025-07-22 19:09:26] debug: 	zh:controller: Received payload: clusterID=64529, address=28490, groupID=0, endpoint=1, destinationEndpoint=1, wasBroadcast=false, linkQuality=180, frame={"header":{"frameControl":{"frameType":0,"manufacturerSpecific":false,"direction":1,"disableDefaultResponse":false,"reservedBits":0},"transactionSequenceNumber":52,"commandIdentifier":10},"payload":[{"attrId":20489,"dataType":66,"attrData":"\u0001\u000b\u0000\u0000\u0000\u0000\u0000\u0000 "}],"command":{"ID":10,"name":"report","parameters":[{"name":"attrId","type":33},{"name":"dataType","type":32},{"name":"attrData","type":1000}]}}

log:
[2025-07-22 19:09:26] debug: 	zh:zstack:unpi:parser: <-- [52,10,9,80,66,10,1,11,0,0,0,127,0,0,0,144,180,75,28,114]
[2025-07-22 19:09:26] debug: 	zh:zstack:unpi:parser: --- parseNext [254,37,68,129,0,0,17,252,74,111,1,1,0,180,0,75,227,118,0,0,17,8,52,10,9,80,66,10,1,11,0,0,0,127,0,0,0,144,180,75,28,114]
[...]

PR #9975, opened 2025-09-10
fix swv attr(0x5008,0x5009)data error problem

liangjia2019:
Fix the associated issue:
https://github.com/Koenkk/zigbee-herdsman-converters/issues/9670#issuecomment-3271756892


Koenkk:
Thanks!

PR #9981, opened 2025-09-10
HY607W-3A: Add correct mode_state handling for Tuya thermostat

sigmich:
Fix mode_state converters for HY607W-3A Tuya thermostat (manual, auto, temp_override). Another attempt to make it work properly. I hope it's the last one for this device.

Implements correct from/to converters for mode_state using the first numeric byte of the datapoint.

Tested on a real HY607W-3A thermostat; other functionality remains unchanged.


Koenkk:
Thanks!

PR #9982, opened 2025-09-10
Fix detection: _TZ3000_pgq7ormg as TS0001_switch_module_3

andrei-lazarov:
Related to #9952

I created a new converter **TS0001_switch_module_3** that is TS0001_switch_module + indicator LED support.  
And I moved _TZ3000_pgq7ormg 1-gang switch there.


Koenkk:
Thanks!

PR #9983, opened 2025-09-11
fix: Inovelli VZM31-SN/VZM36 ignore transition on light.turn_off

rohankapoorcom:
- Use tz.on_off.convertSet when turning off, toggling or turning on with brightness=0 and transition not specified

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/28531

### DIFF
--- a/src/devices/inovelli.ts
+++ b/src/devices/inovelli.ts
@@ -493,7 +493,6 @@ const inovelliExtend = {
         return {
             fromZigbee,
             toZigbee: [
-                tz.on_off,
                 tzLocal.light_onoff_brightness_inovelli,
                 tz.power_on_behavior,
                 tz.ignore_transition,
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

rohankapoorcom:
@Koenkk - somewhat related question for you - I've been looking at the amount of code in this single file and the readability of it and I want to break the break Inovelli converter into several different files/modules within a inovelli folder. I don't see that pattern implemented today, is that something that you'd be open to?

Koenkk:
@rohankapoorcom for that case you can create a manufacturer specific file in the [lib](https://github.com/Koenkk/zigbee-herdsman-converters/tree/master/src/lib) folder.

rohankapoorcom:
@Koenkk I was hoping to break it up into multiple different files. Can I make a manufacturer folder in the lib folder instead?

Koenkk:
@rohankapoorcom if there is a clear benefit, yes.

Issue #9963, opened 2025-09-09
Some questions about the OTA encryption method


liangjia2019:
We found some issues while testing Zigbee OTA based on Telink chips in our team 
The error log is as follows:
`[2025-09-01 18:14:16] info: 	z2m: Zigbee2MQTT started!
[2025-09-01 18:14:47] info: 	z2m: Checking if update available for '0xa4c13865830b1fda'
[2025-09-01 18:14:49] info: 	z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/0xa4c13865830b1fda', payload '{"action":"btn1_single","battery":100,"linkquality":255,"update":{"state":"idle"}}'
[2025-09-01 18:14:51] info: 	z2m: tagID:0xf000
[2025-09-01 18:14:51] info: 	z2m: position:56
[2025-09-01 18:14:51] info: 	z2m: tagID:0xc5e2
[2025-09-01 18:14:51] info: 	z2m: position:193022
[2025-09-01 18:14:51] info: 	z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/bridge/response/device/ota_update/check', payload '{"data":{},"error":"Failed to check if update available for '0xa4c13865830b1fda' (The value of \"offset\" is out of range. It must be >= 0 and <= 193020. Received 193024)","status":"error","transaction":"85ta1-1"}'
[2025-09-01 18:14:51] error: 	z2m: Failed to check if update available for '0xa4c13865830b1fda' (The value of "offset" is out of range. It must be >= 0 and <= 193020. Received 193024)`

After our analysis and communication with Telink's original factory, we found that it was caused by OTA encryption packaging. The tag composition in his packaging format is as follows

<img width="883" height="129" alt="Image" src="https://github.com/user-attachments/assets/4e876a0c-643b-4a13-85c3-9e0c362521f7" />

We have conducted some modification tests on the current code, and the results show that it can be completed normally through OTA

`function parseSubElementTelink(buffer: Buffer, position: number): Ota.ImageElement {
    const tagID = buffer.readUInt16LE(position);
    const length = buffer.readUInt32LE(position + 2);
    const tagInfo = buffer.readUInt16LE(position + 6);//tagInfo Telink
    const data = buffer.subarray(position + 8, position + 8 + length);

    return {tagID, length, data};
}
`
[...]

PR #9984, opened 2025-09-11
add ota handle code with telink ota

liangjia2019:
fix problem with issue:https://github.com/Koenkk/zigbee-herdsman-converters/issues/9963#issuecomment-3271733556


src/lib/ota.ts
```diff
@@ -213,6 +225,26 @@ export function parseImage(buffer: Buffer, suppressElementImageParseFailure = fa
         otaHeaderString: buffer.toString("utf8", 20, 52),
         totalImageSize: buffer.readUInt32LE(52),
     };
+    /* check if specialParser devices*/
+    const SPECIAL_PRODUCTS: [number, number, (buffer: Buffer, position: number) => any, number][] = [
```
Koenkk: Instead of doing this I would propose to add an extra property: `telinkEncrypted: true` to https://github.com/Koenkk/zigbee-OTA/blob/master/index.json, in this way we can use the special logic without having to change the code here.

src/lib/ota.ts
```diff
@@ -245,9 +277,22 @@ export function parseImage(buffer: Buffer, suppressElementImageParseFailure = fa
...
-            const element = parseSubElement(buffer, position);
+
+            // Use the selected parser function
+            const element = useSpecialParser ? specialParser(buffer, position) : parseSubElement(buffer, position);
+
             elements.push(element);
-            position += element.data.length + 6;
+
+            // Update position with different offset based on product type
+            if (useSpecialParser) {
```
Koenkk: Could you also add a [test case](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/test/ota.test.ts) for this logic?

src/lib/ota.ts
liangjia2019: > Instead of doing this I would propose to add an extra property: `telinkEncrypted: true` to https://github.com/Koenkk/zigbee-OTA/blob/master/index.json, in this way we can use the special logic without having to change the code here.

I have modified the processing method by judging based on the fields in the JSON file
image

src/lib/ota.ts
liangjia2019: > Could you also add a [test case](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/test/ota.test.ts) for this logic?

I add some test in there:
describe("parseImage with telinkEncrypted parameter", () => {
        it("should use standard parser and offset when telinkEncrypted is false", async () =>

Koenkk:
I did some cleanup and renamed it to `"customParseLogic": "telinkEncrypted"` instead of `telinkEncrypted` (this is more feature proof in case more logics have to be added in the future).

Thanks!

Koenkk:
Thanks!

PR #9987, opened 2025-09-11
Bosch `BSIR-EZ` (Outdoor siren): Re-implement the device from scratch

DerDreschner:
This pull request re-implements the support for the Bosch Outdoor siren (also known as `BSIR-EZ`) from scratch. The old implementation wasn't complete, in my opinion (e.g., no `convert` / `fromZigbee` method for a lot of attributes which is pretty confusing, using attribute IDs instead of proper naming, exposing a lot of unsupported options, ...) and didn't follow the usage patterns the native Bosch solution does (e.g., exposing `startWarning` from the `ssIasWd` cluster, which is only used by Bosch for demoing the siren volume during first setup and ignores some of the device settings).

I also experienced that the new implementation solves a lot of the issues people experience when adding the device to Z2M (for example: https://github.com/Koenkk/zigbee2mqtt/issues/28525, https://github.com/Koenkk/zigbee2mqtt/issues/28088, https://github.com/Koenkk/zigbee2mqtt/issues/27544 and https://github.com/Koenkk/zigbee2mqtt/issues/27029). It looks like the device doesn't like the unbinding of the `genPollCtrl` cluster in the interview stage, as I needed multiple adding/interviewing attempts before removing that code as well. I honestly have no idea why it's there in the first place, the same with the `m.quirkCheckinInterval(0)` call. To me, both of these seem to do more harm than they solve, which is why I don't bother investigating the reason behind them any further.

<details>

<summary>This is a changelog for affected users</summary>

- **Fix**:  The lookup table for the power source selection contained wrong values. This means that selecting `solar_panel` activated the `ac_power_supply` mode in the device, selecting `ac_power_supply` activated the `dc_power_supply` mode in the device and selecting `dc_power_supply` resulted in an error as it send an invalid value to the device. **Please check the selected power source after the update and correct it (if needed).** 
- **New**: `alarm_state` that shows if the device has *really* turned on the siren and/or light. **Warning**: This is different from the previous attribute with the same name, which was used to start and stop the alarm! This functionality can now be found in the `trigger_alarm` and `stop_alarm` attributes, as there is no way to tell if the alarm is already being triggered or not, which makes an off/on slider the wrong UI element for in my opinion (`alarm_state` will show `no_active_alarm` when the device waits for the delay time of the siren/light to expire).
- **New**: `current_power_source` that shows where the siren currently draws their power from.
- **New**: `power_outage` attribute that will show if the primary power source experiences an outage. This only works when using `ac_power_supply` or `dc_power_supply` as the primary power source.
- New: `solar_panel_voltage` that shows how much voltage the device currently receives from the solar panel.
- **Changed**: The battery status is now being pulled from the `genPowerCfg` cluster only to support the native battery handling by Z2M / HA. Some notes regarding this change:
1.) The remaining battery percentage is now pulled from the device instead of being calculated locally based on the voltage. In my testing, the `genPowerCfg` percentage takes a lot more into consideration than just the voltage (I experienced a drop in battery percentage, although the voltage didn't change).
2.) The `battery_low` warning is more sensitive than before (being true when the remaining battery is <= 60%, which is reasonable in my opinion, keeping the huge power draw during an alarm in mind).
3.) The battery voltage is removed, as Bosch don't use them internally, and I don't want to kill the battery by reporting more attributes than needed.
- **Changed**: The device doesn't reset the `tamper` alarm by default. It has to be reset over a Bosch-specific command in the `ssIasZone` cluster to deactivate the alarm. As this wasn't provided by the old implementation, there was no clear way to handle the situation except to "deal with it". In the native Bosch implementation, each change message from the `ssIasZone` is acknowledged with the Bosch-specific command (if `tamper` is true or `alarm_2`, which is the `power_outage` attribute in my implementation). If the problem is resolved, the device sends a second change message afterward with the alarm being deactivated. I mimic that behavior, which means that alarms regarding the `tamper` and `power_outage` attributes resolve automatically when the tamper-switch is being pressed again, respectively the power from the network comes back. **Please take that into consideration when updating your automations/scripts.**
- **Removed**: Removal of the `ac_status` and `test` attributes, as these are just dead and provide no functionality.
- **Removed**: Removal of the warning mode selection, as the outdoor siren isn't natively triggered that way and only supports the `stop` and `burglar` modes (which is just a few of the provided options) anyway.

</details>

<details>

<summary>Screenshot from the expose tab before and after the change</summary>

### Before
Screenshot from 2025-09-11 23-31-22

### After
Screenshot from 2025-09-11 23-12-57

</details

**TL;DR: All affected users should check and adapt existing automations using the Bosch Ourdoor siren after the update, as this is a breaking change. The activated device settings should be checked afterwards as well (see "Fix" in the changelog).**

### DIFF
--- a/src/lib/bosch.ts
+++ b/src/lib/bosch.ts
@@ -1451,7 +1451,7 @@ export const boschBsirExtend = {
                 )
                 .withCategory("diagnostic"),
         ];
-        
+
         const fromZigbee = [
             {
                 cluster: "ssIasZone",


Koenkk:
Thanks!

PR #9988, opened 2025-09-12
fix: improve Sunricher SR-ZG9041A-2R endpoint naming & remove powerOnBehavior false to use default behavior

niracler:
- Change endpoint names from "1", "2" to "relay1", "relay2" for better identification
- Remove powerOnBehavior: false to use default behavior
- This allows better UI labeling in Zigbee2MQTT for distinguishing between relay channels


Koenkk:
This is a breaking change, let's not change it for now but add a comment above it: `// DEPRECATED: xxx`

PR #9990, opened 2025-09-12
Update tuya.ts support _TZE204_4bjixefp model TS0601 energy meter

ccfiel:
same model, config, and image, but different manufacturer _TZE204_4bjixefp


Koenkk:
Thanks!

PR #10003, opened 2025-09-13
fix: OSRAM 4058075816459 action list

MageFroh:
Small update to the action list for 4058075816459, to reflect what happens when using its buttons with zigbee2mqtt 2.6.1 (prompted by https://github.com/Koenkk/zigbee2mqtt/issues/28003#issuecomment-3286432208).

Am I right that this list is in no particular order?


Koenkk:
> Am I right that this list is in no particular order?

yes, thanks!

PR #10011, opened 2025-09-14
feat(device): Amina S EV Charger - Add charge limit with on/off

jondavidsson:
This pull request introduces a new charge_limit_with_on_off endpoint for the Amina EV charger converter.

The existing charge_limit attribute does not support values below 6 A. The new endpoint allows setting the charge limit below 6 A, which will pause charging and also toggle the charger on/off as needed.


Koenkk:
Thanks!

VVVVVVVip:
great work

PR #10014, opened 2025-09-14
Namron Simplify Dimmer

xses79:
1-2p dimmer, Zigbee / BT.
Energy, Watt, Amp, On off, dim

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -1624,4 +1624,53 @@ export const definitions: DefinitionWithExtend[] = [
             m.temperature({reporting: undefined}),
         ],
     },
+    {
+        zigbeeModel: ['4512791'],
+        model: '4512791',
+        vendor: 'Namron AS',
+        description: 'Namron Simplify Zigbee dimmer (1/2-polet) Zigbee / BT',
+        fromZigbee: [
+          fz.on_off,
+          fz.brightness,
+          fzLocal.electrical_measurement_scaled,
+          fzLocal.metering_scaled,
+          fz.power_on_behavior,
+          fzLocal.levelctrl_vendor_attrs,
+        ],
+        toZigbee: [
+          tz.on_off,
+          tz.light_onoff_brightness,   // standard
+          tz.power_on_behavior,
+          tzLocal.start_brightness,    // standard LevelCtrl lagring
+          tzLocal.phase_type,          // HW (0xB000)
+          tzLocal.pole_mode,           // HW (0xB001)
+        ],
+        exposes: [
+          e.light_brightness(),
+          e.power(), e.current(), e.voltage(), e.energy(),
+          exposes.enum('power_on_behavior', ea.ALL, ['off','on','toggle','previous']),
+          exposes.numeric('start_brightness', ea.ALL).withValueMin(1).withValueMax(254)
+            .withDescription('Default brightness at power-on/startup'),
+          exposes.enum('phase_type', ea.ALL, ['trailing','leading'])
+            .withDescription('AC dimming phase type (hardware via 0xB000)'),
+          exposes.enum('pole_mode', ea.ALL, ['1_pole','2_pole'])
+            .withDescription('Wiring mode (hardware via 0xB001)'),
+        ],
+        endpoint: (device) => ({default: 1}),
+        configure: async (device, coordinatorEndpoint) => {
+          const ep = device.getEndpoint(1) || device.getEndpoint(0) || device.endpoints?.[0];
+          if (!ep) return;
+          await ep.bind('genOnOff', coordinatorEndpoint);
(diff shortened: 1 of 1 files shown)


src/devices/namron.ts
```diff
@@ -1624,4 +1624,43 @@ export const definitions: DefinitionWithExtend[] = [
             m.temperature({reporting: undefined}),
         ],
     },
+    {
+        zigbeeModel: ["4512791"],
+        model: "4512791",
+        vendor: "Namron AS",
+        description: "Namron Simplify Zigbee dimmer (1/2-polet / Zigbee / BT) ",
+        fromZigbee: [fz.on_off, fz.brightness, fz.power_on_behavior],
+        toZigbee: [tz.on_off, tz.light_onoff_brightness, tz.power_on_behavior],
```
Koenkk: please use modern extend instead of to/fromZigbee (search for m.light and m.electricityMeter in this codebase for examples

src/devices/namron.ts
```diff
@@ -1624,4 +1624,48 @@ export const definitions: DefinitionWithExtend[] = [
...
         ],
     },
+    {
+        zigbeeModel: ["4512791"],
+        model: "4512791",
+        vendor: "Namron AS",
+        description: "Namron Simplify Zigbee dimmer (1/2-polet / Zigbee / BT)",
+
+        // Modern extend (on/off, brightness, transition, power_on_behavior)
+        extend: [extend.light_onoff_brightness()],
```
Koenkk:
```suggestion
        extend: [m.light(), m.electricityMeter()],
```

`fromZigbee`, `exposes`, `endpoint` and `configure` can be removed.

src/devices/namron.ts
```diff
@@ -1624,4 +1624,36 @@ export const definitions: DefinitionWithExtend[] = [
             m.temperature({reporting: undefined}),
         ],
     },
+    {
+        zigbeeModel: ['4512791'],
+        model: '4512791',
+        vendor: 'Namron AS',
+        description: 'Namron Simplify Zigbee dimmer (1/2-polet / Zigbee / BT)',
+        extend: [
+          m.deviceEndpoints({endpoints: {main: 1}}),
```
Koenkk: I don't think this is needed? `main` can be removed everywhere.

src/devices/namron.ts
xses79: OKay, think i have resolved it.

src/devices/namron.ts
```diff
@@ -1624,4 +1624,29 @@ export const definitions: DefinitionWithExtend[] = [
...
+        ],
+        exposes: [
+            exposes.numeric("dimming_speed", ea.ALL).withValueMin(1).withValueMax(10).withDescription("Default ramp time in seconds"),
+
+            exposes.numeric("min_brightness", ea.ALL).withValueMin(1).withValueMax(127).withDescription("Minimum brightness (≈1–50%)"),
+            exposes.numeric("max_brightness", ea.ALL).withValueMin(127).withValueMax(254).withDescription("Maximum brightness (≈50–100%)"),
+
+            exposes.numeric("start_brightness", ea.ALL).withValueMin(1).withValueMax(254).withDescription("Default brightness at power-on/startup"),
+
+            exposes.enum("pole_mode", ea.ALL, ["1_pole", "2_pole"]).withDescription("Wiring mode (1- or 2-pole)"),
```
Koenkk: There is no fromZigbee converter to publish a value for `pole_mode`, so it will never show a value here (?)

src/devices/namron.ts
```diff
@@ -1617,18 +1617,25 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        zigbeeModel: ["4512792"],
-        model: "4512792",
-        vendor: "Namron",
-        description: "Simplify 1-2p relay (Zigbee / BT)",
+        zigbeeModel: ["4512791"],
+        model: "4512791",
+        vendor: "Namron AS",
```
Koenkk:
```suggestion
        vendor: "Namron",
```

(should be the same as all other `vendor` in this file)

src/devices/namron.ts
```diff
@@ -1617,18 +1617,25 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        zigbeeModel: ["4512792"],
```
Koenkk, line 1641: Is this expected? This make the `4512792` unsupported.

Koenkk: @xses79 i think this was not fixed?

xses79: 4512792 is to be supported.

xses79: i may have doubled up something here.

xses79:
To messy, creating new pull.

PR #10016, opened 2025-09-14
Thermostat modernExtend. Aqara W500 and W600 thermostat

kirovilya:
Aqara W500 and W600 thermostat. models UT-A01E, WT-A03E.
Thermostat modernExtend

https://github.com/Koenkk/zigbee2mqtt.io/pull/4156


Koenkk:
Pushed some improvements, for some reason the tzConverter in the lookups are `undefined`, I'll check later.

Koenkk:
Thanks!

MeisterQ:
@kirovilya Arent there exposes missing like external temperature input, valve position etc?

kirovilya:
> @kirovilya Arent there exposes missing like external temperature input, valve position etc?

Not yet. External sensor support can be implemented, but nothing else.

MeisterQ:
> > @kirovilya Arent there exposes missing like external temperature input, valve position etc?
> 
> Not yet. External sensor support can be implemented, but nothing else.

How du you know there is no more?
How can we implement the external sensor? I need this feature!

DanielMarmander:
@kirovilya Would it not be possible to add support for metering as well? It seems ZHA supports it, so it should be exposed by the device?

Is it possible to read both internal and ntc sensor at the same time? Right now, we have to switch which sensor is used by the thermostat in order to read the values.

kirovilya:
> How du you know there is no more?

I looked at what data the device has in the native Aqara Home app.
An external sensor can be used, but it is necessary to understand the structure of the ZCL-packages that are used to implement this. I don't have enough time for that yet.

kirovilya:
> Would it not be possible to add support for metering as well? It seems ZHA supports it, so it should be exposed by the device?

Yes, it's probably possible. We need to figure it out

DanielMarmander:
> > Would it not be possible to add support for metering as well? It seems ZHA supports it, so it should be exposed by the device?
> 
> Yes, it's probably possible. We need to figure it out

In case it's of any help, this is where I found that they seem to have metering support:
https://github.com/zigpy/zha-device-handlers/issues/4412

MeisterQ:
> > How du you know there is no more?
> 
> I looked at what data the device has in the native Aqara Home app. An external sensor can be used, but it is necessary to understand the structure of the ZCL-packages that are used to implement this. I don't have enough time for that yet.

So, probably we could use the same configuration as with th E1 TRV? Id give a bet, this is the same with the W600

PR #10023, opened 2025-09-15
Update candeo.ts

candeodevelopment:
Adds support for changing the motion sensitivity, motion timeout and illuminance reporting interval to the Candeo C-ZB-SEMO Motion Sensor device

(this is a hybrid IAS / Tuya device, it only uses Tuya DP's for the specific purposes of changing these settings)


Koenkk:
Thanks!

PR #10034, opened 2025-09-16
Bosch `BSEN-M` (Motion detector): Use `iasZoneStatus` to determine if test mode is active or not and harden deferred occupancy turn-off

DerDreschner:
During analysis of the code from the `BSEN-C2`, I learned that Bosch also uses some bits in the `iasZoneStatus` to push device information to the coordinator. For the `BSEN-M`, this includes whether the test mode is active or not. By using that field, the use of the `device.meta` storage is obsolete, which makes the code much easier.

This pull request implements the use of the `iasZoneStatus` bit to determine the test mode status.

Additionally, I harden the deferred occupancy turn-off. In case the server was stopped before the turn-off was executed, the occupancy status stuck at the true value until a new motion was detected. This is fixed with this pull request.

### DIFF
--- a/src/lib/bosch.ts
+++ b/src/lib/bosch.ts
@@ -1788,25 +1788,19 @@ export const boschBsenExtend = {
             ON: true,
             OFF: false,
         };
-
-        const enableNormalOperationMode = async (endpoint: Zh.Endpoint | Zh.Group) => {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #10037, opened 2025-09-17
fix: Legrand 67772: remove dimmer converters

OlivierMarty:
This device does not support dimming, configuring the device yields warning `Failed to configure (Device 0x... has no input cluster genLevelCtrl)`.

I have this device. See also https://community.jeedom.com/t/legrand-067722-avec-variateur-ou-pas/128581 for another user reporting this issue.

Despite all my efforts, I did not manage to test the change locally.


Koenkk:
Thanks!

PR #10039, opened 2025-09-17
Add new Philips A21 E26 1600 lumen white and color ambience bulb

fffe:
This is the bulb: https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-a21-e26-smart-bulb-1600/046677591298

The product claims 50-20000k and the mired range matches. Experimentally this seems correct.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4163

### DIFF
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -4258,9 +4258,8 @@ export const definitions: DefinitionWithExtend[] = [
     {
         zigbeeModel: ["LCA015"],
         model: "9290038539",
-        vendor: "Signify Netherlands B.V.",
+        vendor: "Philips",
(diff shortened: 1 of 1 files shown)


fffe:
@Koenkk the external definition generated in the dev console detects this as model LCA015, but the model number printed on the bulb is 9290038539.

Which should I have used for the filename?

Since it's also calling the vendor Signify, should this have gone in a different file?

Koenkk:
Use `9290038539` for the filename and `model`, `zigbeeModel` should stay as is.

fffe:
I've put in a new PR for the bulb image, and updated the model.

Should I leave the vendor alone?

Koenkk:
Thanks!

Issue #8926, opened 2025-03-03
Ledron YK-16


dinamik1488:
Hi, Is it possible to make the same-MiBoxer FUT089Z Controls Exposer??
https://github.com/Koenkk/zigbee2mqtt-user-extensions/blob/main/unstable/miboxer-fut089z/controls-exposer.js

for

https://www.zigbee2mqtt.io/devices/YK-16.html

Issue #6274, opened 2023-10-12
Multiple `MiBoxer FUT089Z (RGB+CCT Remote)` devices can't be used individually


T-vK:
The `MiBoxer FUT089Z (RGB+CCT Remote)` is a device that directly sends events to predefined Zigbee Groups. These Zigbee Groups are currently static (101-108) and the same for every remote.

The group names should be set dynamically using the device's ieee address. E.g. `[ieeeAddr+1, ieeeAddr+2, ...]` instead of `101,102,...`.

(This issue was created, so that I can mention it in a PR.)

PR #10049, opened 2025-09-19
MiBoxer FUT089Z remote: Converter improvements & fixes

slugzero:
I got one of these remotes recently, and while there has already been a lot of awesome work by others to get it supported in Zigbee2MQTT, it still took me a while to collect all the bits and pieces to get everything running to my satisfaction. Here's what I came up with - I believe that this will ease setting this up also for others, and I hope that you consider at least parts of this worthy to be added - lmk what you think. 

Since this device is available for several year now and probably actively used in many installations, I decided to implement everything as user-configurable additions which should not break existing installations.

### Summary
This PR improves the MiBoxer FUT089Z converter and fixes several open issues and limitations:
- Implement features of the user extension directly in the converter (fixes https://github.com/Koenkk/zigbee2mqtt-user-extensions/issues/7, https://github.com/Koenkk/zigbee2mqtt-user-extensions/issues/14 and https://github.com/Koenkk/zigbee-herdsman-converters/issues/8926)
- Proper numeric exposes for all target values, including hue and saturation (previously not supported in user extension, see  https://www.zigbee2mqtt.io/devices/FUT089Z.html)
- Enable support for multiple FUT089Z remotes on the same network (fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/6274)
- Fix ability to send messages to the device (fixes a quirk mentioned in the documentation https://www.zigbee2mqtt.io/devices/FUT089Z.html)

### Features & fixes in detail

**Implement features from the user extension in the converter**

Previously, the device required installation of a user addon (https://github.com/Koenkk/zigbee2mqtt-user-extensions/tree/main/unstable/miboxer-fut089z) for some functionalities. That user extension is not working with recent Zigbee2MQTT versions (https://github.com/Koenkk/zigbee2mqtt-user-extensions/issues/7 and https://github.com/Koenkk/zigbee2mqtt-user-extensions/issues/14). This PR implements the functionalities of the user addon directly in the converter as user-configurable options, not changing the default behavior. This removes the need to install an additional addon, and basically resolves the issues. It should also resolve https://github.com/Koenkk/zigbee-herdsman-converters/issues/8926, because the features should also work out-of-the-box with other white-label devices.

* Publish zone-specific actions: The FUT089Z has 8 on/off button pairs. All buttons are using the same endpoint ID, unlike other remote controls, which commonly use one endpoint per button pair. Therefore the converter publishes the same actions for each button, e.g. `on`, `off`, etc.. While in principle it is possible to distiguish buttons via group IDs, this is very unusual to do and differs from other similar devices. With the zone-specific actions enabled, the converter appends a suffix to each action (`_zone_1` to `_zone_8`, where "zone" is the term that is used in the user manual of the remote control), thus mimicking the behavior of other devices with different endpoints per button pair more closely. 
* Additionally, the numeric values of the action properties (hue, saturation, level, etc.) can be exposed separately by enabling "expose values". This makes the last values more easily accessible e.g. in Home Assistant for automations. 

**Enable use of multiple FUT089Z devices**
During configuration, each `zone` (on/off button) is assigned to a Zigbee group (101-108). These groups are fixes, so if multiple FUT089Z devices are on the same network, they all control the same groups and devices. 
This PR introduces configuration options (and the corresponding functionality) which let the user define a different group for any button. This fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/6274

**Fix ability to send messages to the device**
As with most battery-powered device, this remote has its Rx turned off most of the time and is only able to receive commands after it sent a command itself (e.g. after a button press). Therefore messages to the device must be queued long enough. This is ensured by the `quirkCheckinInterval(21600)` (i.e. wait for up to 6 hours until the device checks in again - the observed reporting interval of this device is 4 hours). This change is required to reliably modify the device configuration after initial configuration, e.g. to change group IDs. 

### What's still not working
The remote seems to decide internally whether it wants to send a hue/saturation change or a color temperature request. Most of the times (e.g. after pressing one of the on/off buttons), it sends hue/saturation. After the 'W' button is pressed, it sends color temperature. I have not found a way yet to force the remote to a fixed behavior for each zone.

### Next steps
- Update documentation
- ...

### DIFF
--- a/src/devices/miboxer.ts
+++ b/src/devices/miboxer.ts
@@ -46,8 +46,8 @@ function getZoneSuffixFromGroupId(groupId: number | undefined, options: KeyValue
     // Find the zone number for the given group ID
     const zoneNumber = Object.keys(zoneGroupMapping).find((zone) => zoneGroupMapping[Number.parseInt(zone, 10)] === groupId);
 
-    // If no zone number is found for this group or if legacy actions are enabled, add no suffix
-    if (!zoneNumber || options.legacy_actions === true) {
+    // If no zone number is found for this group or if zone actions are disabled, add no suffix
+    if (!zoneNumber || options.zone_actions !== true) {
         return "";
     }
 
@@ -160,16 +160,16 @@ function miboxerFut089zControls() {
         new exposes.Binary("expose_values", exposes.access.SET, true, false).withDescription(
             "Expose additional numeric values for action properties (hue, saturation, level, etc.)",
         ),
-        new exposes.Binary("legacy_actions", exposes.access.SET, true, false).withDescription(
-            "Publish legacy actions without zone IDs (in addition to zone-specific actions)",
+        new exposes.Binary("zone_actions", exposes.access.SET, true, false).withDescription(
+            "Publish zone-specific actions with zone IDs (e.g., on_zone_1, off_zone_2)",
         ),
     ];
 
     // Generate action exposes for all possible zones
-    function zoneActions(legacy: boolean): string[] {
+    function zoneActions(enableZoneActions: boolean): string[] {
         const baseActions = ["on", "off", "brightness_move_to_level", "color_temperature_move", "move_to_hue_and_saturation"];
 
-        if (legacy) {
+        if (!enableZoneActions) {
             return baseActions;
         }
 
@@ -230,7 +230,7 @@ function miboxerFut089zControls() {
                 const dynamicExposes = [];
 
                 // Add action expose with zone-aware actions
-                dynamicExposes.push(e.action(zoneActions(options.legacy_actions === true)));
+                dynamicExposes.push(e.action(zoneActions(options.zone_actions === true)));
 
                 // Add numeric sensors if enabled in options
                 if (options.expose_values === true) {
--- a/src/devices/miboxer.ts
+++ b/src/devices/miboxer.ts
@@ -1,5 +1,6 @@
 import * as fz from "../converters/fromZigbee";
 import * as exposes from "../lib/exposes";
+import * as m from "../lib/modernExtend";
 import * as tuya from "../lib/tuya";
 import type {DefinitionWithExtend} from "../lib/types";
 
@@ -33,6 +34,9 @@ export const definitions: DefinitionWithExtend[] = [
             e.battery(),
             e.battery_voltage(),
             e.action(["on", "off", "brightness_move_to_level", "color_temperature_move", "move_to_hue_and_saturation", "tuya_switch_scene"]),
+        ],
+        extend: [
+            m.quirkCheckinInterval(21600), // Device observed to report every 4h, set to 6h (21600s) for safety margin
         ],
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
--- a/src/devices/miboxer.ts
+++ b/src/devices/miboxer.ts
@@ -2,9 +2,57 @@ import * as fz from "../converters/fromZigbee";
 import * as exposes from "../lib/exposes";
 import * as m from "../lib/modernExtend";
 import * as tuya from "../lib/tuya";
-import type {DefinitionWithExtend} from "../lib/types";
+import type {Definition, DefinitionWithExtend, Zh} from "../lib/types";
 
 const e = exposes.presets;
+
+// ModernExtend function for FUT089Z remote control (basic functionality)
+function miboxerFut089zControls() {
+    const fromZigbee = [
+        fz.battery,
+        fz.command_on,
+        fz.command_off,
+        fz.command_move_to_level,
+        fz.command_move_to_color_temp,
+        fz.command_move_to_hue_and_saturation,
+        fz.tuya_switch_scene,
+    ];
+
+    const exposesList = [
+        e.battery(),
+        e.battery_voltage(),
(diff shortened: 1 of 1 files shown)


src/devices/miboxer.ts
```diff
@@ -1,10 +1,295 @@
...
+            if (zoneSuffix === null) return undefined;
+
+            // Call the base converter
+            const baseResult = baseConverter.convert(model, msg, publish, options, meta);
+            if (!baseResult) return undefined;
+
+            // Override the action with zone suffix
+            return {
+                ...baseResult,
+                action: `${actionName}${zoneSuffix}`,
```
Koenkk: >  I decided to implement everything as user-configurable additions which should not break existing installations.

Does this not change the reported actions?

slugzero: Good point. It actually does not change the actions, because the `actionNames` are set to the same values as in the base converters, and `zoneSuffix` is empty as long as `options.zone_actions` is not set (cf. https://github.com/Koenkk/zigbee-herdsman-converters/pull/10049/files#diff-4b260b5fd015375f2e9086bc4ae23698b90809745669fc924fe8188b84153cd3R51). 
However, the `actionName` parameter is superfluous (it's actually a left-over from the local converter I started with) and I should better use `baseConverter.action` here. Fix is on the way.

src/devices/miboxer.ts
```diff
@@ -1,10 +1,294 @@
...
+    const fromZigbee: Fz.Converter<string, undefined, any>[] = [
+        onConverter,
+        offConverter,
+        brightnessConverter,
+        colorTempConverter,
+        colorConverter,
+        fz.battery,
+        fz.tuya_switch_scene,
+        // Add converters for numeric sensors (always included, but they check options internally)
+        actionPropertyConverters.hue_saturation,
```
Koenkk, line 139: the on/off converters already expose `action_level`, I don't think we need these?

src/devices/miboxer.ts
```diff
@@ -1,10 +1,294 @@
...
+                .numeric(`zone_${zoneNum}_group_id`, exposes.access.SET)
+                .withDescription(`Group ID for zone ${zoneNum} (default: ${DEFAULT_GROUP_ID_BASE + zoneNum})`)
+                .withValueMin(1)
+                .withValueMax(65535);
+        }),
+        // Feature toggles
+        new exposes.Binary("expose_values", exposes.access.SET, true, false).withDescription(
+            "Expose additional numeric values for action properties (hue, saturation, level, etc.)",
+        ),
+        new exposes.Binary("zone_actions", exposes.access.SET, true, false).withDescription(
```
Koenkk, line 162: Instead of making this an option, I think it's more logical to include the `zone in the property, so e.g. in case of a zone 1 on, the following is published:

```
{
  action: on,
   action_zone: 1
}
```

src/devices/miboxer.ts
slugzero: I took them over from the user converter. Yes, this is the same information exposed in a separate property, but it is the only way I found to properly expose the hue/saturation/color temperature values to e.g. Home Assistant and I think this is also why it was done this way in the user extension.
It is required due to some very uncommon behavior of the device: After each hue/saturation or color temperature move command, the device sends another `brightness_move_to_level` action, which immediately overwrites the values in the state topic. This is the mqtt output for a single tap on the color wheel:
```
zigbee2mqtt-dev/MiboxerRemote {"action":"move_to_hue_and_saturation","action_group":101,"action_hue":47,"action_saturation":182,"action_transition_time":0,"battery":null,"linkquality":58,"voltage":null}
zigbee2mqtt-dev/MiboxerRemote/action move_to_hue_and_saturation
zigbee2mqtt-dev/MiboxerRemote {"action":"brightness_move_to_level","action_group":101,"action_level":228,"action_transition_time":0,"battery":null,"linkquality":58,"voltage":null}
zigbee2mqtt-dev/MiboxerRemote/action brightness_move_to_level
```
So when parsing the payload after triggering on `move_to_hue_and_saturation`, the state topic might (depending on the timing) already be overwritten by the state without the `action_hue` and `action_saturation` values (For example, this seems to be what Home Assistant does, the hue/saturation values are always undefined). This can be worked around on receiver side by caching the state before the action is triggered.

So either we have a cleaner converter, but with poor "out-of-the-box" behavior (admittedly due to strange behavior of the device), and subscribers need to  to implement their own message caching logic to get the color information. Or we go with the functionality the user extension had when it was still functional. I am fine with both and will remove this part if you prefer the former.

src/devices/miboxer.ts
slugzero: I'd rather stay with appending it to the action because this matches more closely what other remotes with one proper endpoint ID per on/off button send (i.e. one dedicated action topic per on/of button pair, with the endpoint ID attached to the action by `postFixWithEndpointName()`). Users then can simply subscribe to `on_zone_1` instead of having to parse the payload.
Adding the `zone` to the payload would not give the users much benefit compared to the `group` property which is already published by the base converters. Both would require parsing the message payload, and the zone ID gives little extra information compared to the group ID.
For reference, this is what's currently being published with one "on" button press with the option disabled:
```
zigbee2mqtt-dev/MiboxerRemote {"action":"on","action_group":101,"battery":null,"linkquality":65,"voltage":null}
zigbee2mqtt-dev/MiboxerRemote/action on
```

Koenkk:
Thanks!

PR #10058, opened 2025-09-21
TS0601_knob_dimmer_switch: Few improvements

l1berat0r:
Hi,

I've added few improvements to the converter for TS0601_knob_dimmer_switch
- Add support for switch_mode, scene_action and power_on_behavior
- Add configurable group_id to bind to (Device uses "miboxerSetZones" command)
- Fix ranges for brightness and color_temp

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -734,6 +734,17 @@ const tzLocal = {
             return ret;
         },
     } satisfies Tz.Converter,
+    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
+    TS0601_knob_dimmer_switch_group_id: {
+        key: ["group_id"],
+        convertSet: async (entity, key, value, meta) => {
+            // The device uses custom group command known from miboxer switches to bind to a group.
+
+            await entity.command("genGroups", "miboxerSetZones", {
+                zones: [{zoneNum: 1, groupId: Number(value)}],
+            });
+        },
+    } satisfies Tz.Converter,
 };
 
 const fzLocal = {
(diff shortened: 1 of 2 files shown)


src/devices/tuya.ts
```diff
@@ -18819,25 +18830,51 @@ export const definitions: DefinitionWithExtend[] = [
...
         exposes: [
+            e.switch(),
             e.switch().withEndpoint("l1"),
             e.switch().withEndpoint("l2"),
-            e.light_brightness_colortemp([154, 370]),
-            e.enum("adjustment_mode", ea.STATE_SET, ["brightness", "color_temp"]).withDescription("Adjustment mode"),
+            tuya.exposes.switchMode().withEndpoint("l1"),
+            tuya.exposes.switchMode().withEndpoint("l2"),
+
+            e.numeric("brightness", ea.ALL).withValueMin(1).withValueMax(100).withDescription("Brightness"),
```
Koenkk: This is part of `e.light_brightness_colortemp(`, please use that otherwise it won't be exposed as a light

l1berat0r: Ok, I'll change that, but can you please help me understand what does it mean "exposed as light"? What difference does it make to how the switch works withing the system?

Koenkk: For e.g. HA it will discover a light instead of 2 sensors (one for on/off and one for brightness)

l1berat0r:
Separate definitions for brightness and color_temp have been replaced by light_brightness_colortemp as advised.

src/devices/tuya.ts
```diff
@@ -19008,25 +19019,45 @@ export const definitions: DefinitionWithExtend[] = [
...
         exposes: [
+            e.switch(),
             e.switch().withEndpoint("l1"),
             e.switch().withEndpoint("l2"),
-            e.light_brightness_colortemp([154, 370]),
-            e.enum("adjustment_mode", ea.STATE_SET, ["brightness", "color_temp"]).withDescription("Adjustment mode"),
+            tuya.exposes.switchMode().withEndpoint("l1"),
+            tuya.exposes.switchMode().withEndpoint("l2"),
+
+            e.light_brightness_colortemp([2700, 6500]),
```
```suggestion
            e.light_brightness_colortemp([2700, 6500]),
```

This looks like the Kelvin range, not the mireds wich it should be, the old `e.light_brightness_colortemp([154, 370]),` looks more realistic

l1berat0r: The device itself displays the color temperature in Kelvins, so I thought that the converter should reflect what's shown on the display. Does it make sense? Or the mired range should be used anyway, for some other reasons?

Koenkk: The mired range should indeed but used as all devices use mired, this also has to be updated:

```
[107, "color_temp", tuya.valueConverterBasic.scale(154 , 370, 0, 1000)],
```

l1berat0r: Ok, understood. Fix pushed.

Koenkk:
Thanks!

PR #10062, opened 2025-09-22
Inovelli: reverting change that broke device configuration parsing

InovelliUSA:
@Nerivec

We noticed that the some changes in this commit: #9867

Caused some problems with the configuration value parsing for our devices. Not sure if the best solution is to change it back or if you think something else is more appropriate.


InovelliUSA:
Going to open a new PR so closing this one.

PR #10068, opened 2025-09-23
MUR36014: Add control for the led indicator

schauveau:
Add control of the led indicator for the MUR36014 (a.k.a Schneider Electrics Mureva Evlink Socket).

By default, the led is pulsating while charging, green while powered but not charging, and off while not powered. The temporary mode is similar but only during the next 10 seconds after any change of state.

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -77,6 +77,19 @@ function socketIndicatorMode() {
     });
 }
 
+function evlinkIndicatorMode() {
+    return m.enumLookup({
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

enthalpie91:
hi 
so we can remove the code added for correction now ?

schauveau:
You mean the external converter? If so then yes, you can probably remove it after upgrading Z2M.

enthalpie91:
yes the converter. everything looks good i will see after removing it

PR #10075, opened 2025-09-24
fix: correctly handle ubisys device setup

sjorge:
Our implementation was working by accident, this is now corrected and we can write larger payloads. This is done in the format Ubisys expects, however I did remove the old fallback code. It's complicated as, people should really update the firmware if they are running such and old version.

There are so many fw bug fixes missing if they don't, I think it's OK to remove this. There is now a warning logged telling people to upgrade.

### DIFF
--- a/src/devices/ubisys.ts
+++ b/src/devices/ubisys.ts
@@ -624,37 +624,6 @@ const ubisys = {
                         );
                     }
                 }
-
-                // re-read effective settings and dump them to the log
-                await ubisys.tz.configure_device_setup.convertGet(entity, key, meta);
-            },
-
-            convertGet: async (entity, key, meta) => {
-                const cluster = Zcl.Utils.getCluster("manuSpecificUbisysDeviceSetup", null, meta.device.customClusters);
-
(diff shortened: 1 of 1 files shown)


sjorge:
https://github.com/Koenkk/zigbee-herdsman-converters/issues/9926#issuecomment-3330257690

Was raised and it's a good question, but I honestly don't know how to go about this.
1. we never really do anything with the payload we read (We can't reverse decode it anyway as we only support a few pre-made configurations)
2. since we don't really know the correct size before hand, we would need to some how guess
3. based on the guess do multiple readStructure calls?

I think it's probably OK as-is, the device does seem to sort of need the read to complete the configuration and I tested the following (as they work on my C4 wiring setup)

- 2x toggle
- 4x toggle (2 last ones are ignored as note wired but config is valid)
- 2x dimmer_single

Also written but not able to verify, the device did accept it.
- 2x dimmer_double

Also big thanks to @Nerivec I couldn't figure how to swap me to the co-author spot instead, he did most of the debugging and code fixes as I was hopelessly confused by this.

Koenkk:
Thanks!

PR #10078, opened 2025-09-25
add hejhome siren _TYZB01_sbpc1zrb

llljjj7:
I have added the siren device from Hejhome. It is the Tuya TS0216 siren model, and since the volume control range was operating differently, I modified it while adding the device.
image link - https://github.com/Koenkk/zigbee2mqtt.io/pull/4185
Thank you.


Koenkk:
Thanks!

PR #10081, opened 2025-09-25
Adjust MOSZB-140 reporting configuration to fix excessive reports

gluoncompany:


src/devices/develco.ts
```diff
@@ -640,8 +640,8 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [
             develcoModernExtend.addCustomClusterManuSpecificDevelcoGenBasic(),
             develcoModernExtend.readGenBasicPrimaryVersions(),
-            develcoModernExtend.temperature(), // TODO: ep 38
-            m.illuminance(), // TODO: ep 39
+            develcoModernExtend.temperature({reporting: {min: 10, max: 3600, change: 100}}),
+            m.illuminance({reporting: {min: 10, max: 3600, change: 100}}),
```
Koenkk, line 646: How often are both values reported? The default reporting values should be good (at least that is the goal of it)

gluoncompany: The default temperature report is the same as in the code, but the brightness report value is set to 5 ("change: 5"). This causes it to report constantly and not enter deep sleep, draining the battery. I forced the temperature setting in case the manufacturer changed any parameters so it wouldn't be affected. This could be removed from my change if you want.

Koenkk:
Thanks!

PR #10085, opened 2025-09-26
Update moes.ts

Mp1988a:
Fixed indicator_status at SFL01-Z


Koenkk:
there are more `SFL02-Z-*` devices, I think the same fix also need to be applied to those?

PR #10088, opened 2025-09-26
Add Support for Shelly Flood Gen 4

kimpenhaus:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4190

Resolves: https://github.com/Koenkk/zigbee2mqtt/issues/28553

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -83,13 +83,11 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        zigbeeModel: ['Flood'],
-        model: 'S4SN-0071A',
-        vendor: 'Shelly',
-        description: 'Flood Gen 4',
-        extend: [
-            m.battery(),
-            m.iasZoneAlarm({"zoneType":"generic","zoneAttributes":["alarm_1","alarm_2","tamper","battery_low"]})],
+        zigbeeModel: ["Flood"],
+        model: "S4SN-0071A",
+        vendor: "Shelly",
+        description: "Flood Gen 4",
+        extend: [m.battery(), m.iasZoneAlarm({zoneType: "generic", zoneAttributes: ["alarm_1", "alarm_2", "tamper", "battery_low"]})],
         meta: {},
-    }
+    },
 ];
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -72,7 +72,7 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        zigbeeModel: ["Power Strip"],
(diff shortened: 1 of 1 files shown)


src/devices/shelly.ts
```diff
@@ -82,4 +82,12 @@ export const definitions: DefinitionWithExtend[] = [
             m.electricityMeter(),
         ],
     },
+    {
+        zigbeeModel: ["Flood"],
+        model: "S4SN-0071A",
+        vendor: "Shelly",
+        description: "Flood Gen 4",
+        extend: [m.battery(), m.iasZoneAlarm({zoneType: "generic", zoneAttributes: ["alarm_1", "alarm_2", "tamper", "battery_low"]})],
```
```suggestion
        extend: [m.battery(), m.iasZoneAlarm({zoneType: "water_leak", zoneAttributes: ["alarm_1",  "tamper", "battery_low"]})],
```

kimpenhaus: hey @Koenkk - thanks for the suggestions - as you figured already this is my first time trying to add a device and I am asking myself how I can provide best value. 

from the shelly control center I see two different types of alarm modes.

image

when looking at the reported status in z2m:

```json
{
    "alarm_1": false,
    "alarm_2": false,
    "battery": 100,
    "battery_low": false,
    "linkquality": 255,
    "tamper": false
}
```
I wonder if alarm_1  and alarm_2 are each ment to be for the different alarm mode?

What do you think?

Thanks,
m.

ps.: this is slightly offtopic - but in the picture you can see that battery capacity is about 89% - in the reported status its about 100 - any clue how/why it is like that?

Koenkk: > I wonder if alarm_1 and alarm_2 are each ment to be for the different alarm mode?

I'm not sure, can you trigger the different modes and see if either alarm_2 or 1 goes to `true`?

> but in the picture you can see that battery capacity is about 89% - in the reported status its about 100 - any clue how/why it is like that?

I guess the device reports a slightly different percentage than in the app, I would try putting in a more empty battery and see if it changes (can take up to 25 hours to report the new value)

kimpenhaus: > I'm not sure, can you trigger the different modes and see if either alarm_2 or 1 goes to true?

changed to rain mode - but from what I can see this doesn't bring any difference

on the web-frontend of the device I saw this explanation to the different modes:

image

so I guess the alarm states are only true once they got reported

kimpenhaus: couldn't find any more details - checked this: https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/DeviceFeatures but that doesn't make it much clear to me - so if you're fine with this it would go with your recommended changes and see if we can be more precise in the future

Koenkk:
Thanks!

PR #10090, opened 2025-09-26
fix: All endpoints have electricityMeter for S4PL-00416EU

dennisv:
All endpoints have electricityMeter for S4PL-00416EU


Koenkk:
Thanks!

PR #10096, opened 2025-09-27
Add support for TRADFRIbulbT120E26WSopal440lm (Japanese variant)

torawo:
This PR adds support for the Japanese variant of the IKEA TRÅDFRI bulb (T120 E26 White spectrum opal 440 lumen), which shares functionality with LED1937T5 but has a different zigbeeModel.

The definition uses:
- `addCustomClusterManuSpecificIkeaUnknown()` to support IKEA-specific manufacturer cluster
- `ikeaLight()` for standard light control (on/off, brightness, color temperature)
- `m.identify()` for identify cluster support

Tested locally with Zigbee2MQTT and confirmed working with on/off, brightness, and color temperature.  
Home Assistant integration also verified.  
No LED model number was found on packaging or IKEA Japan's website, so zigbeeModel was used as model identifier.


Koenkk:
Thanks!

Issue #10091, opened 2025-09-27
[External Converter] TS0601 from _TZE284_madl8ejv


infernix:
Saswell SAS936RHB-7-Z03 zigbee thermostat for desk or wall mount. https://www.saswell.com/uploadfile/downloads/SASWELL%20Smart%20Home.pdf

My variant is an ecobay stand alone:  https://ecobaysmart.com/wp-content/uploads/2024/10/SAS936RHB-7-Z03.pdf

Can be bought and used separately. Shows up as TS0601  variant

This is my external converter for `TS0601` from `_TZE284_madl8ejv`
software_build_id: `2.6.1`
date_code: ``
endpoints:
```json
{"1":{"clusters":{"input":["genGroups","genScenes","manuSpecificTuya","genBasic","zosungIRTransmit"],"output":["genOta","genTime"]}}}
```

### What works / what doesn't?

Setpoint temperature, current temperature, child lock, off/on mode, heating demand.

Schedules not implemented or tested.

### Converter

```js

const tuya = require('zigbee-herdsman-converters/lib/tuya');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const e = exposes.presets;
const ea = exposes.access;

// Custom converter for DP 3 (heating_demand)
const heatingDemandConverter = {
    from: (v) => {
        let value;
        if (Array.isArray(v)) {
            value = v[v.length - 1];
        } else {
            value = v;
        }
        return value === 1 ? 'ON' : 'OFF';
    },
    to: (v) => (v === 'ON' ? 1 : 0),
};

// Custom converter for DP 101 (system_mode as boolean)
const systemModeConverter = {
    from: (v) => (v === true ? 'heat' : 'off'),
    to:   (v) => (v === 'heat' ? true : false),
};

module.exports = {
    fingerprint: [{ modelID: 'TS0601', manufacturerName: '_TZE284_madl8ejv' }],
    model: 'SAS936RHB-7-Z03',
    vendor: 'Saswell',
    description: 'Wireless Temperature Sensor',
    fromZigbee: [tuya.fz.datapoints],
    toZigbee: [tuya.tz.datapoints],
    configure: tuya.configureMagicPacket,
    exposes: [
        exposes.climate()
            .withLocalTemperature(ea.STATE)
            .withSetpoint('current_heating_setpoint', 5, 30, 0.5, ea.STATE_SET)
            .withSystemMode(['off', 'heat'], ea.STATE_SET),
[...]

PR #10097, opened 2025-09-27
feat: add Saswell SAS936RHB-7-Z03 sensor support to tuya.ts

infernix:
Issue link: https://github.com/Koenkk/zigbee-herdsman-converters/issues/10091

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4193

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -16872,6 +16872,54 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_madl8ejv"]),
+        model: "SAS936RHB-7-Z03",
+        vendor: "Saswell",
+        description: "Wireless Temperature Sensor",
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
+        exposes: [
+            e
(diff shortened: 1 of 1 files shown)


Koenkk:
I'm wondering if the device is the same as the Nedis ZBHTR20WT, could you try to:

- Stop Z2M
- Open `data/database.db` with a text editor and change all occurrences of `_TZE284_madl8ejv` to `_TZE200_ne4pikwm`
- Start Z2M and see if the device now works

infernix:
So, it's very similar, but:

- this is not a TRV, it's a standalone/wall mountable thermostat
- Open window does nothing to the device, there is no indicator for it and it does not change its settings
- Same for frost protection and scale protection
- Error status stays null
- When applying this device, pushing the main button toggles between normal state, leave home, and schedule mode. But the schedule cannot be read or set from the device, so that is not desirable. When applying my config, the schedule mode is not selectable. 

Do you want me to try and adapt the nedis config and strip window/frost/scale and the schedule setting?

infernix:
So i tried that, but after I used the Nedis one, now the schedule is activated on the device and it is part of the toggle with the main button (`on > leave_home > scheduler` loop), and I can't "disable" the scheduler option. With no way of changing the schedule,  that's not preferable.

Still trying to find a way to undo that.

Koenkk:
Then let's keep this, thanks!

PR #10098, opened 2025-09-27
Add new fingerprint entry for Eurotronic device

gummiangler:
This adds support for datecode 20240315 eurotronic devices


Koenkk:
Thanks!

PR #10110, opened 2025-09-29
Bosch Thermostats (BTH-RA/-RM/-RM230Z): Overwork implementation

### LINKED ISSUE #9662: Bosch BTH-RM230Z supported, but missing features. Can I help?
Saw this in the Z2M supported devices list, so picked one a while ago. However, I found that there's no way to configure the sensor connection to use a cable temperature sensor, so it was no good for me to use as an underfloor heating thermostat.

Picked up the Bosch Smart Controller II with the intention of fishing out the commands via Wireshark. Set everything up, managed to figure out their funky Zigbee encryption, then realised I know nothing about adding features to this project. Ran out of time and I've been using it the "official way" ever since.

Image

Seeing as I should still be able to read the configuration values being sent when switching between these modes, could someone add the feature if I dig my sniffing laptop out and post the traffic?

DerDreschner:
This pull request overworks the implementation for the Bosch thermostats (`BTH-RA` known as Radiator Thermostat II, `BTH-RM` known as Room Thermostat II and `BTH-RM230Z` known as Room Thermostat II 230V).

Here is a short overview of the changes I made (purely technical changes aside):
- **New**: Support for the external temperature sensor on `BTH-RM` and `BTH-RM230Z`
- **New**: Support for setting the heater type on `BTH-RM230Z`
- **New**: Support for setting the valve type on `BTH-RM230Z`
- **New**: In the case of errors on the device, the error number is now visible in Z2M as well. If the shown error number doesn't match the one on the device, please open an issue so I can change the algorithm. I was only able to force one error, so it may not be bulletproof.
- **Changed**: Valve adaptation can now be triggered with an enum instead of an on/off switch on `BTH-RA` which makes it more user-friendly
- **Changed**: Open window mode was renamed to reflect the fact it's not getting automatically enabled by the thermostat (this is handled in software on the Smart Home Controller in the native Bosch implementation to provide more flexibility and settings)
- **Changed**: Boost heating now throws an error when the system mode isn't `heat` and the heating type isn't `radiator` (last case happens on `BTH-RM230Z` only)
- **Changed**: `displayOntime` was renamed to `displaySwitchOnDuration` to fix a typo and use the same wording as the native Bosch app
- **Changed**: Some descriptions were changed to use the same wording as the native Bosch app
- **Changed**: The daylight saving time requests are now handled differently to avoid possible issues during interview

I hope I didn't forgot anything important. If I did, just let me know and I'll add it here.

This fixes #9662 and https://github.com/Koenkk/zigbee2mqtt/issues/25021.


src/lib/modernExtend.ts
```diff
@@ -3086,7 +3092,8 @@ export function thermostat(args: ThermostatArgs = {}): ModernExtend {
     const repConfigChange0: ReportingConfigWithoutAttribute = {min: "MIN", max: "1_HOUR", change: 0};
     const repConfigChange10: ReportingConfigWithoutAttribute = {min: "MIN", max: "1_HOUR", change: 10};
 
-    const expose = e.climate().withLocalTemperature();
+    const localTemperatureDescription = localTemperature.description ?? undefined;
```
DerDreschner, line 3095: I hope this change is okay. I didn't wanted to duplicate a bunch of code just to have a custom description for the `localTemperature` attribute.

src/lib/exposes.ts
```diff
@@ -1312,13 +1312,6 @@ export const presets = {
     tilt: () => new Binary("tilt", access.STATE, true, false).withDescription("Indicates whether the device detected tilt"),
     voc: () => new Numeric("voc", access.STATE).withLabel("VOC").withUnit("µg/m³").withDescription("Measured VOC value"),
     voc_index: () => new Numeric("voc_index", access.STATE).withLabel("VOC index").withDescription("VOC index"),
-    pi_heating_demand: () =>
```
DerDreschner, line 1315: @Koenkk : I decided to remove this entry as you added it solely for the `BTH-RA` (https://github.com/Koenkk/zigbee2mqtt/issues/25971). With the rewrite, it's not necessary anymore. Hope that's okay.

src/devices/bosch.ts
```diff
@@ -197,270 +170,6 @@ const boschBmctDzSettings = {
...
-            toZigbee,
-            isModernExtend: true,
-        };
-    },
-    ignoreDst: (): ModernExtend => {
-        const fromZigbee = [
-            {
-                cluster: "genTime",
-                type: "read",
-                convert: async (model, msg, publish, options, meta) => {
```
DerDreschner, line 447: @Koenkk: During testing, I realized that this solution doesn't work anymore. The response send to the device is always empty:

image

and the `attributes` parameter in zigbee-herdsman (here to be exact: https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/model/endpoint.ts#L533) is always empty as well. That was the main reason I moved it to an `customReadResponse` in the first place.

Is this a bug? I just want to learn what I did wrong.

src/lib/exposes.ts
Koenkk: yes, perfect

src/devices/bosch.ts
Koenkk: Could you provide the debug log of this?

DerDreschner: Sure, it's attached. It only affects Z2M Edge.

[z2m-edge-debug.log](https://github.com/user-attachments/files/22605010/z2m-edge-debug.log)

Koenkk: It's zigbee-herdsman responds to it (before zigbee-herdsman-converters gets the chance to). We should remove it here and extend: https://github.com/Koenkk/zigbee-herdsman/blob/4385acb3e91699995ad89fcd4046649a8dabf3f8/src/controller/model/device.ts#L356

DerDreschner: Ahh, I see! Is the new `customReadResponse` solution okay for you then? I'm not quite sure what you exactly mean with "remove it here and extend".

I researched the possibility to support daylight saving time natively in zigbee-herdsman last year, but I didn't want to include new external libraries.

DerDreschner: Ahh, you mean answer the `dstStart`, `dstEnd` and `dstShift` requests in zigbee-herdsman? But with what value? I just choose `0` as the Bosch devices work well with it. But according to the Zigbee specification, it looks like the correct answer in that case would be `0xffffffff` for start/end and `0` for shift to me:

image

I create a pull request in zigbee-herdsman with the default values from this table and will remove the code here.

DerDreschner: Okay, it has to be 0 instead of 0xffffffff. Otherwise, the devices still ask the coordinator again and again... Not thrilled about that to be honest.

Koenkk: If it's manu specific, then let's go for `customReadResponse`

Koenkk:
Thanks!

Koenkk:
I reverted this PR because it breaks, this change causes the [`action_template` and `action_topic`](https://github.com/Koenkk/zigbee2mqtt/blob/d44463a330867e97129476f3eb5e5a3b6334ab33/test/extensions/homeassistant.test.ts#L10610) to be missing because it's not exposing the running state on the climate expose (`withRunningState`). Can you check if this is expected? Then let's re-merge it after the 1 October release.

DerDreschner:
Ohh, sorry, I didn't realized that there are unit tests against the BTH-RA in Z2M! Otherwise, I would have considered these before committing. Yeah, sure, I'll check that and will tell you here.

andydvsn:
Thanks for all your work on this, @DerDreschner. It is much appreciated!

DerDreschner:
Ahh, sorry, my bad! I thought it's enough to expose an enum with the name `running_state` (like I did [here](https://github.com/Koenkk/zigbee-herdsman-converters/pull/10110/files#diff-8ea88e36cf16a0aeec72ff6539ad3ac3ded189e605096fd8b4de8d0ff3187ce5R2855)). But then the associated thermostat feature is missing.

As there is an `overrideHaDiscoveryPayload` method anyway: Is it okay for you to just add the necessary payload that the [HA controller in Z2M would add](https://github.com/Koenkk/zigbee2mqtt/blob/d44463a330867e97129476f3eb5e5a3b6334ab33/lib/extension/homeassistant.ts#L633)? Otherwise, I wouldn't be able to use the `thermostat` call from modernExtend as that makes calls against the native zigbee `runningState` attribute, which isn't supported on the BTH-RA. In that case, I would manually pick the necessary converters/exposes etc. instead, like the `m.thermostat` function does internally.

Koenkk:
What if we add an option to disable those running state calls? Its cleaner than the separate sensor

Sopsy:
@DerDreschner I started implementing (or should I say copying) the changes after your research and I believe one of the values you have declared is not correct. I had to get the SHC2 and sniff the traffic myself to verify and at least on my setup the Central heating value is 0x01, not 0x03. 0x03 is not used in the official software.

https://github.com/DerDreschner/zigbee-herdsman-converters/blob/9495ccf0ed10f91537012e7700d051607a1d75c0/src/lib/bosch.ts#L2670

DerDreschner:
@Sopsy : Can you provide more information about your setup? Which device and firmware is being used? Just to see if there is any difference in software versions etc.

Sopsy:
@DerDreschner I set up a separate system with just the SHC2 and one BTH-RM230Z. Firmware version on the thermostat is 0.3.8 and SHC2 is 10.29.4850-3811. These are the newest ones available in the official software.

DerDreschner:
Interesting. I've checked my capture and it is indeed `0x03` in my case. :thinking: Now the question is what is different between your setup and mine? Did you had anything connected to the outputs for example? I tested mine "naked" without anything connected.

image

Sopsy:
Now that's very odd. I do have a valve connected to the output as I used a base from another thermostat to avoid having to do any electrical works. But then again, it is a dry contact relay as far as I'm aware, so the device should be completely oblivious to anything connected to the contacts. And the base does not contain any logic.

On my unit the 0x03 value does not even change any parameters on attributes 0x4025, 0x4041 or 0x4061. All other values do.
I just tested with thermostat firmware 0.3.3 and the same behavior is there as well.

Are you sure this capture is from the BTH-RM230Z and not the battery powered BTH-RM?
One thing I can think of is the country selection, but I wouldn't think that can matter here.

DerDreschner:
Yeah, the BTH-RM doesn't support this Zigbee attribute. Which language do you use? I use it in English, but German timezone.

One thing that's possible is maybe an old app version or something like that? They used to support "boilers" as heater type. This isn't selectable in my case, instead, there is "central heating" in this selection screen:

image

This is the case since this update: https://www.bosch-smarthome.com/uk/en/support/release-information/system-solutions/2025-04-02.html

Sopsy:
Oh okay, then it's definitely the right one. I use it in English as well, but with Finnish timezone. The app version is 10.29.4018 API_3.18 on Android. This might not be the newest one as I had to download it with Aurora store (not officially available here, so unavailable from the Play Store directly). Different app version would explain it. Mine does show central heating instead of a boiler.

I researched it a little bit more with both version 0.3.3 and 0.3.8 thermostat, and I indeed got it to write a 0x03 as well.
It looks like 0x01 is written when there is no room selected, and 0x03 when there is.

This does not really make any sense to me as is, but seeing your screenshot where the central heating used to be called boiler, what I think is happening here is that 0x01 is the old boiler mode, which is also used as "central heating without a room" or when a thermostat that does not support the feature (0.3.3) is used.

So now I'm not totally sure if the 0x01 should be selectable as well or not. I think it's deprecated but still works. However what I find strange is that at least on my thermostat nothing changes when 0x03 is selected - as if the old setting is still applied. And I did not see any room settings going over Zigbee to the thermostat.

DerDreschner:
How did you succeed to add the device without room selection? :thinking: Maybe I can replicate it on my side then. But I didn't know that's even possible.

Sopsy:
Well... To be honest I'm not totally sure. I tried to do it again now, but was not able to replicate it, so it looks like a bug in the app. With the version 0.3.3 thermostat it does not ask for the room at all, so maybe it got confused of the thermostat versions.

DerDreschner:
Hmm, I will try it later. Maybe you left the wizard before selecting a room by closing the app? I think that's possible.

But would make sense to add 0x01 = boiler in the code then.

Sopsy:
I tried that and it did not work - just threw an error when the room selection came up. My best guess is that it's firmware related as I was using the two different versions and forcefully aborted the (automatic) update of the older one a few times by unplugging the thermostat.

Pretty sure 0x01 is the old boiler setting. I wish Bosch released a bit more data what actually is different with the central heating and why it needed a new value in the firmware.

There is some info here, but it's barebones.
https://www.bosch-smarthome.com/uk/en/support/help/product-help/room-thermostat-2-230v-help/
"The heater type setting in the app adjusts the parameters and control for boilers (potential-free), underfloor heating (or other 230V application) and radiators (purely radio application, switching output is not used)."

DerDreschner:
Maybe you can tell me what 0x01 changes on the other attributes (esp. `0x4025` and `0x4061`)? Maybe we can reverse-engineer it then?

DerDreschner:
Radiator (0x02):
0x4025: 2 (0x02)
0x4061: 10 (0x0a)

Underfloor heating (0x00):
0x4025: 3 (0x03)
0x4061: 20 (0x14)

Changing the mode to central heating (0x03) doesn't report any value changes to these attributes at all.

DerDreschner:
Hmm, my best guess right now would be that 0x01 makes the thermostat automatically turn the output on/off when the temperature needs it. When using 0x03, that's not the case and it expects the SHC II to enable the relay? But that needs further investigation.

Sopsy:
Here are my test results. I tried changing a lot of other settings as well to see if they have any effect, but no, nothing affects these.
The thermostat also accepts a write into these attributes and keeps the value as long as the heater type is not changed so I have a feeling they are some heating algorithm parameters.

0x00
0x4025: 3
0x4061: 0x14
0x406c: 0

0x01
0x4025: 1
0x4061: 0x06
0x406c: (no change from previous value)

0x02
0x4025: 2
0x4061: 0x0a
0x406c: 5

0x03
0x4025: (no change from previous value)
0x4061: (no change from previous value)
0x406c: (no change from previous value)

Sopsy:
What makes testing a bit harder is that the thermostat has some rate limiting or delays for the relay actuation, so it does not always respond as you'd expect. But I think you are onto something there. It looks like the relay is indeed disconnected with Central heating selected. And that would make sense regarding the algorithm parameters - I believe in this mode the algorithm is not used and the thermostat is just reporting the temperature and setpoint to the SHC to do the switching. Thus they are not changed as it does not affect anything. And this also makes 0x406c more interesting.

In boiler mode (0x01) the relay definitely works normally, and so it does in 0x00 and 0x02 as well.

And I think the valve status led attribute (0x4033) is used with central heating as it would then be controlled, well, centrally, by the SHC. It works in all modes though.

xoxys:
Thanks for your work! Looking forward to it :) This would also fix https://github.com/Koenkk/zigbee2mqtt/issues/20057 and https://github.com/Koenkk/zigbee2mqtt/discussions/21009

zeratul021:
@DerDreschner do you know why was this reverted in b647eab5fc9e4201a60ab96a80d049cbd0d7e2dd ?

TheMagnetar:
> @DerDreschner do you know why was this reverted in [b647eab](https://github.com/Koenkk/zigbee-herdsman-converters/commit/b647eab5fc9e4201a60ab96a80d049cbd0d7e2dd) ?

https://github.com/Koenkk/zigbee-herdsman-converters/pull/10110#issuecomment-3353523678

DerDreschner:
@Sopsy : Yeah, I've checked that and indeed, `0x03` releases the control for the relay from the device itself. I've taken that into account with the follow-up pull request here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/10309 You can change the name as well, in my opinion, as the German comment in the Bosch Community Forum indicates that there is no functional difference between `0x01` and `0x03` apart from that it's being controlled by the coordinator and not the device itself.

Thanks again for the documentation of the activity LED!

Sopsy:
@DerDreschner No problem! Thanks for your investigation as well! I think what you found out makes sense, also because it required a firmware update to work. I will update the names as well.

PR #10114, opened 2025-09-30
add noTamperModel TYZB01_ttvdudvx

llljjj7:
For the Tuya device with manufacturer name _TYZB01_ttvdudvx and model ID TS0207, the device does not have a tamper sensor. Therefore, an exception handling was necessary, and the model has been added accordingly.


Koenkk:
Thanks!

PR #10119, opened 2025-09-30
M8Pro: Add weather info support.

l1berat0r:
Hi,

I have added weather info support for M8Pro switch [#23812](https://github.com/Koenkk/zigbee2mqtt/issues/23812).
This should go together with the [PR](https://github.com/Koenkk/zigbee-herdsman/pull/1516) to zigbee-herdsman.


l1berat0r:
As advised, I moved the protocol parsing logic from ZH to the converter. It works just fine on my environment.
It requires new version of ZH to be built properly, with the new Tuya command defined.

src/lib/tuya.ts
```diff
@@ -59,6 +59,39 @@ export const dataTypes = {
...
+    LightSnow: 104,
+    Snow: 105,
+    FreezingFog: 106,
+    Rainstorm: 107,
+    Shower: 108,
+    Dust: 109,
+    Spit: 112,
+    Sleet: 113,
+    Yin: 114,
+    FreezingRain: 115,
```
Koenkk:
```suggestion
    freezing_rain: 115,
```

Please make all values `snake_case`

l1berat0r: Fix pushed.

Koenkk:
Thanks!

andrei-lazarov:
~~I believe there is something wrong with this PR.~~  
The pre-commit fails on my PC since this commit. Not sure how it got through?

image

Edit: I needed to update dependencies

l1berat0r:
Have you reinstalled project dependencies? This PR requires Tuya command deffinitions which are present in ZH 6.2.0.

andrei-lazarov:
My bad.. Indeed it works now. Thanks for the quick reply

PR #10122, opened 2025-10-01
Renamed the attribute 'sensor' for Aqara W500 to match lumiExternalSensor

kirovilya:
These changes ensure the attribute name doesn't change after the release.
I'll add support for using an external sensor in the future.


Koenkk:
Thanks!

PR #10132, opened 2025-10-02
add manufacture TZ3000_tas0zemd for 2gang switch fingerprint

llljjj7:
The TZ3000_tas0zemd model supports only the on-off switch, but it provides extended backlight mode and other non-functional features, so we added it to the white model.


Koenkk:
Thanks!

PR #10163, opened 2025-10-03
Add unit "m" and power strip 5 gang

Bjk8kds:
- Add unit "m" for Distance (ZY-M100-24GV3)
- Change Description "Power strip 5 gang" (TS011F_5)


Koenkk:
Thanks!

PR #10164, opened 2025-10-03
Bosch devices: Fix unset attributes in `expose` functions due to cluster rename

DerDreschner:
Due to the rename of the custom cluster from `boschSpecific` to `boschEnergyDevice`, the devices in the BMCT line of Bosch didn't work correctly anymore. The reason is the use of cached cluster attributes from those custom clusters via `device.getEndpoint().getClusterAttributeValue()` in some `exposes` functions, which lead to important functions not being exposed.

My proposal for this issue would be this pull request, which reads all attributes in the renamed cluster once and calls `deviceExposesChanged()` to ensure all exposed functions are based on actual values in the renamed custom cluster. That way, I don't have to consider the cluster rename throughout the whole code base in the future.

This fix should only be necessary for the BMCT line of devices (all others don't use the `getClusterAttributeValue()` function), but as this issue will possibly lead to a hotfix release, I tend to be more conservative and apply it for the Bosch door/window contacts and the smart plug compact as well ([although there is a successful upgrade for a door/window contact being reported](github.com/Koenkk/zigbee-herdsman-converters/issues/8610)). The new BSP-FD isn't affected, as it used the new cluster name from the beginning.

I'm really sorry for this stupid error and the inconvenience it might bring!

This fixes https://github.com/Koenkk/zigbee2mqtt/issues/28806.

### DIFF
--- a/src/lib/bosch.ts
+++ b/src/lib/bosch.ts
@@ -70,25 +70,16 @@ export const boschGeneralExtend = {
                         logger.debug(`Attempt to read all attributes for cluster ${newClusterName} from endpoint ${endpoint.ID}`, NS);
 
                         for (const attributeToRead in newClusterDefinition.attributes) {
-                            const attributeValueExistsInDatabase =
-                                endpoint.getClusterAttributeValue(newClusterName, newClusterDefinition.attributes[attributeToRead].ID) === undefined;
-                            if (attributeValueExistsInDatabase) {
-                                logger.debug(
-                                    `Attempt to read undefined attribute ${attributeToRead} in cluster ${newClusterName} from endpoint ${endpoint.ID}`,
-                                    NS,
-                                );
+                            logger.debug(
+                                `Attempt to read attribute ${attributeToRead} in cluster ${newClusterName} from endpoint ${endpoint.ID}`,
+                                NS,
+                            );
 
-                                try {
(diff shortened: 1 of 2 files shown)


src/lib/bosch.ts
```diff
@@ -40,6 +40,84 @@ interface BoschGeneralEnergyDeviceCluster {
...
+     * available when using "getEndpoint().getClusterAttributeValue()". */
+    handleRenamedCustomCluster: (oldClusterName: string, newClusterName: string): ModernExtend => {
+        const onEvent: OnEvent.Handler[] = [
+            async (event) => {
+                if (event.type !== "start") {
+                    return;
+                }
+
+                const device = event.data.device;
+                logger.debug(
```
Koenkk: I would only log this inside the `if (!renameAlreadyApplied) {` and not log anything when it's already applied (otherwise it will create bunch of unnecessary logging in the future.

DerDreschner: Done! I removed the logging in the else case as well.

Koenkk:
Thanks for the quick fix!

mmirannalti:
Thanks for fixing this!

bruno-dogancic:
I apologise for interrupting with a different question, but is this issue related to my newly installed Bosch Room Thermostat II 220V not updating its states automatically?

image

image

Also, feel free to mark this as inappropriate, and I will gladly delete it. :sweat_smile:

PR #10176, opened 2025-10-03
add krc702z

baucicmario:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4220

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -734,6 +734,33 @@ const tzLocal = {
             return ret;
         },
     } satisfies Tz.Converter,
+    videosec_KRC702Z_on_off_action: {
+        key: ['state'],
+        convertSet: async (entity, key, value, meta) => {
+            // This converter is for multi-endpoint devices, so we need the endpoint name (e.g., 'l1', 'l2').
+            const endpointName = meta.endpoint_name;
+            if (!endpointName) {
+                // If there's no endpoint, fall back to the default Tuya converter.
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -734,6 +734,35 @@ const tzLocal = {
             return ret;
         },
     } satisfies Tz.Converter,
+    videosecKrc702zOnOffAction: {
```
Koenkk: Please create a tuya.valueConverter out of this. You can find an [example here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/4653a7f81dd3baed11c765491a0e69c44d9b900b/src/lib/tuya.ts#L615)

baucicmario: I think I did it correctly but if possible please double check it :)

src/lib/tuya.ts
```diff
@@ -986,6 +986,31 @@ export const valueConverter = {
...
+     */
+    onOffStateIgnoringAlreadyCurrentState: {
+        to: (value: string, meta: Tz.Meta) => {
+            const endpointName = meta.endpoint_name;
+            // The multi-endpoint check is important for this logic to work correctly
+            if (!endpointName) {
+                // Fallback for non-multi-endpoint devices
+                return value.toLowerCase() === "on";
+            }
+            const stateProperty = `state_${endpointName}`;
```
Koenkk, line 1001: I think this can also work for non multiendpoint devices? In that case `stateproperty = 'state';`

PR #10180, opened 2025-10-03
fix(detect): Detect HT-THERMZ3W-1 as SR-ZG9092A

chax:
This PR is an attempt to fix Koenkk/zigbee2mqtt#28661

I tested my device by manually changing record in `database.db` replacing modelId with `ZG9092` instead of `HT-THERMZ3W-1`, and that way i can normally use my device.


Koenkk:
Thanks!

PR #10191, opened 2025-10-04
Add windSpeed

EXP-Carthage:
added windSpeed function for msWindSpeed cluster

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -900,6 +900,20 @@ export function soilMoisture(args: Partial<NumericArgs<"msSoilMoisture">> = {}) 
     });
 }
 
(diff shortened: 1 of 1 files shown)


Koenkk:
What device is currently using this?

EXP-Carthage:
Currently, none on the market. Perhaps in the future.
However, I need it for my weather station project (it will surely be useful to others as well). In any case, before proceeding with a pull request, I tested it on my installation and it works perfectly, as the screenshots show:
_z2m
_z2m_converter

The screenshot shows 0 m/s because my anemometer was not running at that moment. But after several tests, the values reported corresponded well with what I expected.

Koenkk:
Thanks!

PR #10203, opened 2025-10-05
added LGT009 and LGT010 for Hue Play wall washer

decowboy:
Added support for two more Philips Hue Play wall washer models, namely LGT009 and LGT010.


Koenkk:
Thanks!

PR #10208, opened 2025-10-05
fix: Remove some ignore converters

Koenkk:
See https://github.com/Koenkk/zigbee-herdsman/pull/1517#issuecomment-3369257242


djeliedu12:
Hello, could you please reference the "CN1408" product from the manufacturer Legrand? Thank you.

Elie

PR #10228, opened 2025-10-06
Update Mec driver recognition

chris-1243:
Add 4-channels driver

As I do not have access to the modules, I use the softwareID to make a difference between the 1-channel and the 4-channel as it seems they share the same modelID... Maybe not the best option.

I also slightly modified the 1-channel description

From discussion
https://github.com/Koenkk/zigbee2mqtt/discussions/28840

and issue
https://github.com/Koenkk/zigbee2mqtt/issues/28852

https://www.ls-light.com/hubfs/LS_MEC_LITE_EN.pdf (page 22)
https://www.ls-light.com/hubfs/LS_DE_General-Catalog-en-2.pdf (page 192)

CC: @PantheraIgnis, @benzfreak

### DIFF
--- a/src/devices/ls.ts
+++ b/src/devices/ls.ts
@@ -31,9 +31,24 @@ export const definitions: DefinitionWithExtend[] = [
     },
     {
         zigbeeModel: ["Mec Driver module"],
-        model: "7061427",
+        model: "756200026",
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #9116, opened 2025-04-04
Fixed pairing of Livolo TI0001 switches.

flaterichd:
Fixes pairing issue for Livolo switches, as discussed here:
[https://github.com/Koenkk/zigbee2mqtt/issues/26648](https://github.com/Koenkk/zigbee2mqtt/issues/26648)

PR #10229, opened 2025-10-06
Fix Livolo TI0001-pir prevent disconnect payload

PaulVanSchayck:
Getting the following error in my logs when trying to re-pair the Livolo TI0001-pir motion detector:

```
Cannot mix BigInt and other types, use explicit conversions
```

In https://github.com/Koenkk/zigbee-herdsman-converters/pull/9116 this was already fixed for the Livolo TI0001 switches, but the pir motion detector was forgotten.


Koenkk:
Bedankt!

PR #10238, opened 2025-10-07
added thermostat mode functions to the aqara w100

greenspeedracer:

### DIFF
--- a/src/lib/lumi.ts
+++ b/src/lib/lumi.ts
@@ -4260,7 +4260,7 @@ export const fromZigbee = {
             const endsWith = Buffer.from([0x08, 0x00, 0x08, 0x44]);
             if (attr.slice(-4).equals(endsWith)) {
                 logger.info(`Detected PMTSD request from device ${meta.device.ieeeAddr}`, NS);
-                return {action: "W100_PMTSD_request"};
+                return {action: "data_request"};
             }
         },
     } satisfies Fz.Converter<"manuSpecificLumi", undefined, ["attributeReport", "readResponse"]>,
@@ -4327,7 +4327,7 @@ export const fromZigbee = {
             logger.info(`Decoded PMTSD: ${JSON.stringify(result)} from ${meta.device.ieeeAddr}`, NS);
             return {
                 ...result,
-                pmtsd_from_w100_data: combinedString,
+                data: combinedString,
             };
         },
     } satisfies Fz.Converter<"manuSpecificLumi", undefined, ["attributeReport", "readResponse"]>,


src/devices/lumi.ts
```diff
@@ -4579,6 +4579,17 @@ export const definitions: DefinitionWithExtend[] = [
         model: "TH-S04D",
         vendor: "Aqara",
         description: "Climate Sensor W100",
+        fromZigbee: [lumi.fromZigbee.w100_0844_req, lumi.fromZigbee.pmtsd_from_w100],
+        toZigbee: [lumi.toZigbee.pmtsd_to_w100, lumi.toZigbee.thermostat_mode],
+        exposes: [
+            e.action(["W100_PMTSD_request"]).withDescription("W100 Requesting PMTSD Data via 08000844 Request"),
+            e.text("PMTSD_from_W100_Data", ea.STATE).withDescription("Timestamp+Most Recent PMTSD Values Sent by W100"),
```
Koenkk: This and the action above is very unclear to me, can you extend the description on what it does?

src/lib/lumi.ts
```diff
@@ -5878,4 +5960,101 @@ export const toZigbee = {
...
+
+            await entity.write(64704, {65522: {value: Buffer.from(fullPayload), type: 65}}, {manufacturerCode: 4447, disableDefaultResponse: true});
+
+            logger.info(`PMTSD frame sent: ${pmtsdStr} (Counter: ${counter}, Checksum: ${checksum})`, NS);
+            return {};
+        },
+    } satisfies Tz.Converter,
+
+    thermostat_mode: {
+        key: ["Thermostat_Mode"],
```
Koenkk:
```suggestion
        key: ["mode"],
```

src/devices/lumi.ts
greenspeedracer: W100_PMTSD_Request: The W100 requests the thermostat data (power, mode, temperature, speed, swing), PMTSD, from the "hub" using a payload. This watches for the payload and creates an action so that the user's platform can send the data to the W100.

PMTSD_from_W100_Data: The W100 sends the PMTSD data in a payload. This field is the timestamp of the last update from the W100 plus the PMTSD data.  This way, the users software can watch this and update the thermostat on a change. 

Thermostat_Mode: When the W100 is put into thermostat mode, the middle line on the W100's display turns on and the W100 ceases to send actions on button presses. Instead, it begins communicating using the payloads I referenced before.

I can change the descriptions to something like "W100 Requesting Thermostat Status," "Timestamp+W100 Thermostat Status," and "Enable/Disable Thermostat Control Mode" if that is better for the end user?

src/devices/lumi.ts
```diff
@@ -4579,6 +4579,17 @@ export const definitions: DefinitionWithExtend[] = [
         model: "TH-S04D",
         vendor: "Aqara",
         description: "Climate Sensor W100",
+        fromZigbee: [lumi.fromZigbee.w100_0844_req, lumi.fromZigbee.pmtsd_from_w100],
+        toZigbee: [lumi.toZigbee.pmtsd_to_w100, lumi.toZigbee.thermostat_mode],
+        exposes: [
+            e.action(["W100_PMTSD_request"]).withDescription("W100 Requesting PMTSD Data via 08000844 Request"),
```
Koenkk:
```suggestion
            e.action(["data_request"]).withDescription("W100 Requesting PMTSD Data via 08000844 Request"),
```

src/devices/lumi.ts
Koenkk: Then I would rename it accordingly:

```suggestion
            e.text("data", ea.STATE).withDescription("Timestamp+Most Recent PMTSD Values Sent by W100"),
```

Koenkk:
Thanks!

NicoBerlo1:
Hi !
I've updated z2m, and enabled the thermostat mode through the GUI, I can see a kind of driving wheel on the aqara, but the middle line doesn't show up, and I don't seem to get any events from clicking buttons on it.
Is there some specific states that need to be enabled by code ?
How do we get the + and - events ?
Thanks ! :)

lyubomirtraykov:
Hello, I can confirm that the thermostat mode is not working as expected. It only turns on the driving wheel icon and disables the buttons. I don't know if it is related to a firmware upgrade I did 2 weeks ago. I have 5 thermostats and all of them behave this way. If you like I can do some debugging and testing.

greenspeedracer:
> Hello, I can confirm that the thermostat mode is not working as expected. It only turns on the driving wheel icon and disables the buttons. I don't know if it is related to a firmware upgrade I did 2 weeks ago. I have 5 thermostats and all of them behave this way. If you like I can do some debugging and testing.

You need to follow the instructions on https://github.com/greenspeedracer/W100-Scripts. You have to setup HA or whatever system you are using to send the appropriate responses to the W100. The W100 is NOT a thermostat as far as zigbee is concerned.

lyubomirtraykov:
Hello I am sending this payload to the thermostat but noting is changing:

```
{
  "PMTSD_to_W100": {"P": 0, "M": 1, "T": 23, "S": 0, "D": 0}
}
```

greenspeedracer:
> Hello I am sending this payload to the thermostat but noting is changing:
> 
> ```
> {
>   "PMTSD_to_W100": {"P": 0, "M": 1, "T": 23, "S": 0, "D": 0}
> }
> ```

Oh yeah, they made me change the names of things in the PR, so those names have to be adjusted. I’ll have to refactor them after work, sorry. I only use this during the summer for AC. 

EDIT: Try pmtsd_to_w100. I think they are all lowercase now.

KipK:
@greenspeedracer Wouldn't it be better for users to expose directly the separated PMTSD values like @Noyax-37 did on his converter here ? 

https://github.com/Koenkk/zigbee2mqtt/issues/27262#issuecomment-3264730170

I've tried, it's easier to script from HA by just using entities and automations

image

KipK:
I've modified the @Noyax_37 converter to fix some issues and made it more verbose.
https://gist.github.com/KipK/3b36895138681f0b3762d3f33d3cc38f

image

Also added some descriptions, renamed entities, and moved appropriated entities to Config section:
image


It needs the device to be deleted and added again to have the proper changes on HA side.
Could be a good basis for extending the current integration, what do you think ?

douglascrc-git:
> I've modified the @Noyax_37 converter to fix some issues and made it more verbose. https://gist.github.com/KipK/3b36895138681f0b3762d3f33d3cc38f
> 
> image
> Also added some descriptions, renamed entities, and moved appropriated entities to Config section: image
> 
> It needs the device to be deleted and added again to have the proper changes on HA side. Could be a good basis for extending the current integration, what do you think ?

Great work! It worked for me. 

Now I can use the device as a thermostat :). 

The only problem is that the device does not appears as a climate entity in home assistant, but I guess it can be solved easily. 

Thanks!

KipK:
I was considering modifying the converter again to let it recognized as a Climate entity. 
Could simplify a lot HA automations but would it be appropriated for non z2m bridges or used outside HA? 

This also can be easily done on HA side using a template

douglascrc-git:
> I was considering modifying the converter again to let it recognized as a Climate entity. Could simplify a lot HA automations but would it be appropriated for non z2m bridges or used outside HA?
> 
> This also can be easily done on HA side using a template

That would be great. 

My first option is to use a template, the second is maybe something like HACS dualthermostat addon. Nevertheless, the best option is to modify the converter directly, but maybe it is more complex to be done.

KipK:
@douglascrc-git 
Here is a new version with exposed climate entity compatible with HA. It's a breaking change needs to remove/readd the device 

https://gist.github.com/KipK/3b36895138681f0b3762d3f33d3cc38f

image
image


And here is the blueprint to synchronise the virtual w100 climate with a real climate entity:

https://gist.github.com/KipK/1d2657d7cc774e9aac80519b193b9457

image

Noyax-37:
Hi @KipK, great job!

PS: I'm French too ;)

douglascrc-git:
> @douglascrc-git 
> Here is a new version with exposed climate entity compatible with HA. It's a breaking change needs to remove/readd the device 
> 
> https://gist.github.com/KipK/3b36895138681f0b3762d3f33d3cc38f
> 
> image
> image
> 
> 
> And here is the blueprint to synchronise the virtual w100 climate with a real climate entity:
> 
> https://gist.github.com/KipK/1d2657d7cc774e9aac80519b193b9457
> 
> image
> 
> 

Great! Thanks! 

I'm out of home, but I will test it next week for sure.

 Is there any tutorial on how to develop such external converters?

KipK:
I've updated again, now we have battery working. 
It's was in LumiSpecific cluster , attr: 247, id 13

image


If the device has been reset ( click 5x on back button )  it reports itselfs just after first pairing with z2m.

NicoBerlo1:
Awesome !
Everything will be available then !

thanks :) (merci ^^)

biller-aivy:
Kipk, so I tried it too, it is working, So I can see the middle line and I can see the thermostat set value in the overview. So, I have a shelly 1 gen4 on my floor heating, so ON is WATER GO, off is WATER stop. So i need now a virtual thermostat to add the functionality to control the shelly right? so when I want add a virtual thermostat in HA I cant choose the w100 as thermometer?!

KipK:
I moved the project files here:

https://github.com/KipK/aqara-w100-z2m-converter

@biller-aivy last version 0.1.3 also expose a temperature sensor entity ( previous one there was only the temperature attribute on the climite entity ) . 

I've also added Device specific settings for min/max temperature setpoint. Those limit temperature range in the climate entity too.

PR #10124, opened 2025-10-01
Add SLC S32052

StuYarrow:
Link to picture pull request: Koenkk/zigbee2mqtt.io#4202

PR #10241, opened 2025-10-07
Move S32052 to preexisting manufacturer file

StuYarrow:
Fixing my own mistake in #10124 here. There was already a file for the same manufacturer under a different name.


Koenkk:
Thanks!

PR #10242, opened 2025-10-07
Change regulatorPercentage from MQTT for Namron Edge Thermostat

OsteHovel:
Fixes the issue where the regulator_percentage attribute cannot be set via MQTT because it has the wrong data type.

I found the manual after I fiddled around with /set payload to guess what datatype it might be, where it actually says INT16S.
[4512783-84-Zigbee-User-Guide.pdf](https://github.com/user-attachments/files/22742831/4512783-84-Zigbee-User-Guide.pdf)


Koenkk:
Thanks!

PR #10248, opened 2025-10-07
NodOn SEM-4-1-00: fix attribute updates

sbraz:
Hi,
Without this, the device might report incorrect energy values and fail to update its attributes when they change.

Fixes: https://github.com/Koenkk/zigbee2mqtt/issues/28898

@AlexisPolegato, please let me know if there is a reason this was initially disabled.


Koenkk:
Thanks!

PR #10251, opened 2025-10-07
Fix & improvements to FP300, FP1e addon

kpustelnik:
- Fixed FP300 fromZigbee presence detection (meant to use attribute 100 but mistakenly used 101)
- Added lumi prevent leave to FP300
- Added two remaining functions (adaptive sensitivity, AI environment learning) to FP300 and FP1e (attributes 349, 350)

Recreating due to stuck MR ( #10250 )


Koenkk:
Thanks!

PR #10257, opened 2025-10-08
CK-BL702-AL-01_1: Enable "hs" color mode for the light definition.

l1berat0r:
Enable "hs" color mode for the light definition - [28500](https://github.com/Koenkk/zigbee2mqtt/issues/28500).


Koenkk:
Thanks!

PR #9998, opened 2025-09-12
eWeLink CK-BL702-AL-01_1 - addressing reported issues.

l1berat0r:
Hi,

I tried to address the issues reported in [#28500](https://github.com/Koenkk/zigbee2mqtt/issues/28500) and [#9617](https://github.com/Koenkk/zigbee-herdsman-converters/issues/9617).
I think most of them should now be solved, however there is thing, related to the change from "old extend" to modernExtend, which I think may be an issue to some (but doesn't have to be). 
Do you think it's something worth discussing first or the commit can be merged so the people can evaluate themselves?

PR #10269, opened 2025-10-09
Restore behavior of using MoveToLevel when state is null

ams2990:
The intended behavior is to allow the user to request MoveToLevel instead of MoveToLevelWithOnOff if they pass `"state": null` in the MQTT message. This was broken by https://github.com/Koenkk/zigbee-herdsman-converters/pull/9998, which only allowed MoveToLevel if the converter sets `moveToLevelWithOnOffDisable`.


dan-danache:
I second this, it will fix by now broken "prestaging" setup.

Koenkk:
Thanks!

PR #10274, opened 2025-10-09
feat(philips): add support for Hue Surimu model "929003598102"

OsteHovel:
This model is white, so the existing image fits just fine.


Koenkk:
Thanks!

PR #10281, opened 2025-10-09
Add 929003479901 (Koenkk/zigbee2mqtt#14841)

zebernst:
Add support for Hue Gradient Signe floor lamp (Oak) - variant of existing models as discussed in Koenkk/zigbee2mqtt#14841


Koenkk:
Thanks!

PR #10286, opened 2025-10-10
Update IRB-4-1-00 and SEM-4-1-00

AlexisPolegato:
Add support of Humidity Cluster for IRB-4-1-00
Add support of produced energy for SEM-4-1-00


Koenkk:
Thanks!

PR #10295, opened 2025-10-11
Fixing window detection on Namron panel heaters

OsteHovel:
This all started with the option to enable the window detection being the wrong way around.
Then I found out that the actual window detection triggered is not even exposed/fetched.
So this changes all that so it is now working the way it is supposed to and it also shows the status.

However, this PR names the window_open_check to window_detection to harmonise with other devices, this is technically changing the property names and may break some integration for someone,  but if they integrated it in the first place it was working the wrong way around anyway.

This is the first time I ever taken something outdoor to actually test a function that rely on a rapid change of temperature (temperature decreases 5°C within 10 minutes)


Koenkk:
Thanks!

PR #10314, opened 2025-10-12
[DO NOT MERGE] fix: Paul Neuhaus NLG-TW light state reporting when controlled by bound remote

Diaoul:
## Summary
- Add periodic state polling for Paul Neuhaus NLG-TW lights to fix state synchronization issues
- Addresses issue where lights don't report state changes when controlled by bound remotes

## Problem
Paul Neuhaus NLG-TW lights don't automatically report state changes when controlled by directly bound remotes (similar to some Gledopto devices). This causes Zigbee2MQTT to show incorrect state after remote button presses.

## Solution
- Added `paulNeuhausTWLight()` function with 5-second polling (same approach as Gledopto)
- Polls `onOff`, `currentLevel`, and `colorTemperature` attributes
- Also enabled `configureReporting: true` for better coverage

## Test plan
- [x] Test with bound remote - state syncs within 5 seconds
- [x] Test direct Zigbee2MQTT control still works
- [x] Verify no errors in polling

### DIFF
--- a/src/devices/paul_neuhaus.ts
+++ b/src/devices/paul_neuhaus.ts
@@ -32,7 +32,7 @@ function paulNeuhausTWLight(args?: m.LightArgs) {
                     await endpoint.read("lightingColorCtrl", ["colorTemperature"]);
                 }
             } catch {
-                // Do nothing
+                // do nothing
             }
         },
     }).onEvent;


Diaoul:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/19592

Diaoul:
I've done more testing and it turns out that only some firmware versions are affected by this issue, not all of them

I'm digging which ones to avoid needless polling on good versions.

Diaoul:
It turns out only some firmware versions seems to be affected by this bug.

_Note that some firmware versions start with a `V` and some don't, this is not a mistake_

| Firmware version | Status |
|--------|--------|
| `1.0.6` | ❌ Confirmed NOT working |
| `1.0.9` | ✅ Confirmed working |
| `1.1` | ❌ Confirmed NOT working |
| `V1.0` | ✅ Confirmed working |
| `V1.3` | ✅ Confirmed working |

Diaoul:
Maybe there is a better way to do that only on affected firmware versions?

Koenkk:
@Diaoul can you try contacting the manufacturer and ask for an OTA? Then people can update their devices and this workaround is not needed anymore.

Diaoul:
I'm in touch with them. OTA will not be possible on those devices for hardware reasons apparently but I'm hopping for a manual firmware upgrade procedure 🤞

Diaoul:
[Anway Elec](http://www.anwayelec.com/) (the manufacturer) just stopped responding to my emails for a module physical upgrade or OTA although we had seemed off to a good start 🤷 
I'm completely ghosted.

I just replaced all the led drivers with something else. It's a shame because the form factor of their product is really convenient.

PR #10319, opened 2025-10-12
fix: Paul Neuhaus NLG-TW light state reporting by enabling configureReporting

Diaoul:
Enable configureReporting for Paul Neuhaus NLG-TW lights. Without this, the light does not report any state changes when controlled by bound remotes.

| Before | After |
|--------|--------|
| Delayed and inconsistent state update | Instant and stable state update |
| image | image |


Diaoul:
Superseeds #10314

Koenkk:
These reporting should be automatically set-up when binding a remote to the light (if not please provide the debug log when doing so). We don't want to enable configureReporting for lights by default as it increases the traffic on the network.

Diaoul:
Yes, actually I was having a false positive when testing this with a custom converter. It turns out the polling survives and a restart is needed.
I'll follow up on the other PR.

Issue #10262, opened 2025-10-08
[External Converter] TS0601 from _TZE284_6ycgarab


wajo666:
Old issue https://github.com/Koenkk/zigbee-herdsman-converters/issues/9289

### What works / what doesn't?
Mapped everything from tuya dev console

### Converter

```js
const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
const tz = require('zigbee-herdsman-converters/converters/toZigbee');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const reporting = require('zigbee-herdsman-converters/lib/reporting');
const tuya = require('zigbee-herdsman-converters/lib/tuya');
const m = require('zigbee-herdsman-converters/lib/modernExtend');
const e = exposes.presets;
const ea = exposes.access;

const definition = {
        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_6ycgarab"]),
        model: "TS0601_smoke_7",
        vendor: "Tuya",
        description: "CO + Smoke detector",
        fromZigbee: [tuya.fz.datapoints],
        extend: [tuya.modernExtend.tuyaBase({dp: true, timeStart: "2000"})],
        exposes: [
            e.enum("smoke_sensor_state", ea.STATE, ["alarm", "normal", "detecting", "unknown"]),
            e.enum("co_state", ea.STATE, ["alarm", "normal"]),			
			e.enum("alarm_volume", ea.STATE_SET, ["min","low", "middle", "high"]),
			e.binary('muffling', ea.STATE_SET, "ON", "OFF"),
			e.binary('alarm_switch', ea.STATE_SET, "ON", "OFF"),
			e.battery(),
        ],
        meta: {
            tuyaDatapoints: [
                [15, 'battery', tuya.valueConverter.raw],
                [16, 'muffling', tuya.valueConverter.onOff],
                [1, 'smoke_sensor_state', tuya.valueConverterBasic.lookup({
                        alarm: 0,
                        normal: 1,
                        detecting: 2,
						unknown: 3,
                    })],
                [17, 'alarm_switch', tuya.valueConverter.onOff],
				[18, 'co_state', tuya.valueConverterBasic.lookup({
                        alarm: 0,
                        normal: 1,
                    })],
				[5, 'alarm_volume', tuya.valueConverterBasic.lookup({
                        min: 0,
                        low: 1,
                        middle: 2,
						high: 3,
                    })],
            ],
        },
    };

[...]

PR #10329, opened 2025-10-13
Update tuya.ts

kjpg99:
Making _TZE284_6ycgarab a new, separate device to add its carbon monoxide detecting functionality - as discussed in #10262

Link to picture pull request: [TODO](https://github.com/Koenkk/zigbee2mqtt.io/pull/4275)

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -7620,8 +7620,7 @@ export const definitions: DefinitionWithExtend[] = [
         fingerprint: tuya.fingerprint("TS0601", ["_TZE284_6ycgarab"]),
         model: "TS0601_smoke_co",
         vendor: "Tuya",
-        description: "Dual Smoke CO sensor",
-        configure: tuya.configureMagicPacket,
+        description: "Dual smoke CO sensor",
         exposes: [
             e.smoke(),
             e


src/devices/tuya.ts
```diff
@@ -7616,6 +7616,61 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_6ycgarab"]),
+        model: "TS0601_smoke_co",
+        vendor: "Tuya",
+        description: "Dual Smoke CO sensor",
+        fromZigbee: [tuya.fz.datapoints],
```
Koenkk: from/to and onEvent can be removed

src/devices/tuya.ts
```diff
@@ -7616,6 +7616,61 @@ export const definitions: DefinitionWithExtend[] = [
...
+            e.enum("alarm_volume", ea.STATE_SET, ["mute", "low", "medium", "high"]).withDescription("Alarm volume"),
+            e.battery(),
+            tuya.exposes.silence(),
+            e.binary("alarm_switch", ea.STATE_SET, true, false).withDescription("Disable the alarm"),
+            e.carbon_monoxide(),
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [
+                    1,
```
Koenkk, line 7639: I would introduce a seperate smoke_state for this, a `smoke` like the old integration is more easy to use.

src/devices/tuya.ts
kjpg99: removed

src/devices/tuya.ts
kjpg99: I hope I understood it correctly and introduced the right changes. Based on the definition of [Onenuo smoke sensor](https://github.com/Koenkk/zigbee-herdsman-converters/blob/5e1418507159d5d08201c13299b777d5f7ea07ca/src/devices/onenuo.ts#L23) I modified it in such a way, that now based on the same datapoint there will be two functions – usual smoke and additional smoke_sensor.
I hope I did it correctly – this seems to me as a pretty advanced and as a non-programmer I am not sure if it's okay and if it's going to work. If not, I will have to ask you for help to implement it.
Thank you for guidance.

Koenkk: Perfect, thanks!

PR #10337, opened 2025-10-14
Update Tuya backlightModeOffOn and indicatorMode to config entities

LucasHagen:
Currently those entities are threaded as 'control' (normal) entities, but should be handled as 'config' entities to separate them in the device control page.

For more context, see https://github.com/Koenkk/zigbee2mqtt/issues/29074

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/29074


Koenkk:
Thanks!

PR #10344, opened 2025-10-14
Copy action lookup for Aqara Z1 Pro (double - triple rockers) switches from quadruple rockers

mirackle-spb:
Copy action lookup for Aqara Z1 Pro (double - triple rockers) switche…

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -3962,6 +3962,7 @@ export const definitions: DefinitionWithExtend[] = [
                 endpointNames: ["top", "bottom"],
             }),
             lumiAction({
+                actionLookup: {hold: 0, single: 1, double: 2, release: 255},
                 endpointNames: ["top", "bottom"],
                 extraActions: ["slider_single", "slider_double", "slider_hold", "slider_up", "slider_down"],
             }),
@@ -3989,6 +3990,7 @@ export const definitions: DefinitionWithExtend[] = [
                 endpointNames: ["top", "center", "bottom"],
             }),
             lumiAction({
+                actionLookup: {hold: 0, single: 1, double: 2, release: 255},
                 endpointNames: ["top", "center", "bottom"],
                 extraActions: ["slider_single", "slider_double", "slider_hold", "slider_up", "slider_down"],
             }),


Koenkk:
Thanks!

leonlegant:
Hello,

I'm excited to see this update. Thank you for all the work.

I updated and was checking my switches, but I do not see double or triple click options when adding the decide to an automation.

Maybe I am misunderstanding ?

mirackle-spb:
There is little strange that actions is not exposed at creation time. You just need to make this actions and than it will appear.

leonlegant:
Thank you for your reply.

Just to make sure I understand, you are saying I need to initiate the actions on the switch to get them to appear in Automations > Device > Trigger > Actions?

I did try that. On all of the switch rockers and slider, I single clicked, double clicked, triple clicked, and did hold.

Then I went into automations and only single click appears for rockers and all the options for the slider appeared, but I cannot see double click, triple click, or hold for rocker switches.

I spend two days troubleshooting, reconfigured the device, interviewed the device, force removed and re-paired the device, and used an automation to track all actions coming from the device. ChatGPT had me try many things and concluded the device does not send double or triple or hold actions. 

I'm still working on my end and will try pairing with aqara app to check firmware version.

leonlegant:
I'm confirming model ZNQBKG43LM 

I updated firmware from 0.0.0_0037 to 0.0.0_0040

I re-paired to zigbee2mqtt and initiated all actions for single, double, triple, and hold.

In Automations > Devices > Triggers, I only see the following:

"single_top" action
"single_bottom" action
"slider_down" action
"slider_up" action
"slider_single" action
"slider_double" action
"slider_hold" action

mirackle-spb:
Check your mqtt homeassistant -> device_automations -> device_id
new actions appear only after they will be trigered. i dont know why.
also check "Click mode" setting of your device. I use only "fast" so i dont try to make double actions. in fast mode only single is working. Will try to check tomorrow.

leonlegant:
Thank you for your reply.

I double checked the device ID and that is the correct device I'm testing.

I then checked switch settings and tested "fast" and "quick_mode", as well as "multi" and "anti-flicker".

I initiated the various actions and then checked if double, triple, hold actions appeared as automation triggers. Still nothing.

I do have single_top and single_bottom actions, those have always been there.

PR #10055, opened 2025-09-20
Bosch Door/Window Contact II: Overwork devices to match official features

DerDreschner:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4173

This pull request overworks the support for the Bosch Door/Window Contacts II (`BSEN-C2`, `BSEN-CV` and `BSEN-C2D`).

Short overview of the changes:
- Implement support to enable/disable the break functionality
- Implement support to set the automatic break timeout from 1 to 15 minutes or disable it completely
- Implement support to see the current break functionality state
- Implement support to enable/disable the vibration sensor on the `BSEN-CV`
- Implement support to set the sensitivity for the vibration sensor on the `BSEN-CV`
- Implement value changes on `BSEN-CV` to match behavior of Bosch SHC II coordinator on interview
- Move `BSEN-C2D` to it's own device as it has a different firmware and needs totally different documentation due to Matter mode as default and no Zigbee QR code on the device
- Implement support for `double_press` action on the `BSEN-C2D` (not available on the `BSEN-C2` and `BSEN-CV`)
- Don't delete `none` as button state to make it easier to use on e.g., Home Assistant

PR #10345, opened 2025-10-14
Bosch BSEN-CV (Door/window contact II plus): Fix vibration detection sensitivity overwrite on each configuration

DerDreschner:
Bosch changes the default vibration sensitivity level on the BSEN-CV from the firmware default `very_high` to `medium` during first connection with the Bosch Smart Home Contoller. This is being done in Z2M since the change in #10055 as well. Unfortunately, I forgot to make sure that the vibration sensitivity isn't being overwritten each time the device is being configured.

This pull request changes that by adding a `device.meta` value when the changed default value was applied.


Koenkk:
Thanks!

PR #10369, opened 2025-10-16
Bosch BSEN-W (Water alarm): Overwork implementation to expose all official features

DerDreschner:
This pull request overworks the implementation for the Bosch Water Alarm (BSEN-W). The device was previously marketed as "BWA-1", which isn't the current model name anymore. Therefore, I uploaded a new picture as well (see [this pull request](https://github.com/Koenkk/zigbee2mqtt.io/pull/4288)).

Changes:
- **New**: Add test mode to verify that the device is (for example) audible in the designated place
- **New**: Add feature to mute and unmute alarms remotely
- **Changed**: Implement changed `currentZoneSensitivityLevel` to avoid possible complaints.
- **Changed**: Moved some functions out of the `boschGeneralExtend` variable to make it more clear by what device type these are being consumed exclusively (energy devices / sensor devices).

### DIFF
--- a/src/lib/bosch.ts
+++ b/src/lib/bosch.ts
@@ -2458,6 +2458,7 @@ export const boschBsenExtend = {
 
         const configure: Configure[] = [
             m.setupConfigureForBinding("ssIasZone", "input"),
+            m.setupConfigureForReading("ssIasZone", ["numZoneSensitivityLevelsSupported", "currentZoneSensitivityLevel"]),
             async (device, coordinatorEndpoint, definition) => {
                 const endpoint = device.getEndpoint(1);
 
@@ -2537,6 +2538,7 @@ export const boschWaterAlarmExtend = {
     changedSensitivityLevel: (): ModernExtend => {
         const configure: Configure[] = [
             m.setupConfigureForBinding("ssIasZone", "input"),
+            m.setupConfigureForReading("ssIasZone", ["numZoneSensitivityLevelsSupported", "currentZoneSensitivityLevel"]),
             async (device, coordinatorEndpoint, definition) => {
                 const endpoint = device.getEndpoint(1);
 


Koenkk:
Thanks!

burmistrzak:
@DerDreschner You'll have to bind `genPollCtrl` as well. Otherwise battery-powered devices operate less efficient and/or are not reachable.

Also, have you verified that newly paired Bosch devices behave the same way as they did before applying your patches?

**Edit:** I'm asking because I spent quite some time fine-tuning bindings and reporting for these things. 😄

DerDreschner:
@burmistrzak : The binding is not necessary. It's being done automatically by `zigbee-herdsman` if the device support it and report it correctly (which the Bosch devices all do), see: https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/model/device.ts#L966

And yeah, I primarily test the devices by re-pair them afterwards. It's the different update-paths I often forget. :sweat_smile:

burmistrzak:
@DerDreschner Huh, that's good to know! 🤝
IIRC, I had issues with `genPollCtrl` not being bound automatically by Bosch devices. But maybe that was before this feature was introduced in Z2M?

DerDreschner:
No, that feature is there for at least 4 years now. My best bet is the missing response to the `zclVersion` read request right on the beginning of the interview on the previous implementation. That's what it looks like on Wireshark:

image

When Bosch devices don't get any response to that request, they tend to just ignore some messages on different occasions (especially during interview) which can lead to several different issues. This is being handled for the BTH-RM230Z [in this line](https://github.com/Koenkk/zigbee-herdsman-converters/blob/28f3190ff3c9c657db066387cb58644009fb5c10/src/devices/bosch.ts#L757). Important: Not all Bosch devices need/do that (e.g., as we're discussing about the BSEN-W here: that device doesn't).

Another reason could be the missing/incomplete response to the time request when we're talking about the Bosch Thermostats. This is being fixed by myself in `zigbee-herdsman` here: https://github.com/Koenkk/zigbee-herdsman/pull/1517

But my bet is the unanswered `zclVersion` read request.

burmistrzak:
@DerDreschner Great catch!
But I wouldn't rely on Bosch to use Zigbee correctly. They could have implemented all their products with standard ZCL attributes but chose to hack together some custom clusters. 🙄

DerDreschner:
@burmistrzak : Honestly, it's not that bad in my opinion.

They just make some assumptions you have to keep in mind (e.g., only being used in markets with daylight saving time, therefore reject time answers without DST information). Apart from that, it's pretty close to standard with minor exceptions that are not very significant if you only implement what is officially supported/used (having the sensitivity thing on the BSD-2 in mind here). The `zclVersion` request is up to standard, as any device must implement that. It's a flaw in `zigbee-herdsman`, not on the Bosch side.

Yes, even the situation on the thermostats is up to standard. The behavior is covered by the Zigbee Cluster Library as well, see the  `ControlSequenceOfOperation` attribute in the HVAC thermostat cluster, which defines what is possible with the current configuration. You could say: they use the Zigbee standard too much. :smile: 

I guess @Koenkk could tell you stories about a lot worse usages of the Zigbee "standard". :smile:

burmistrzak:
@DerDreschner Yea, newer devices are definitely not as bad as the Twinguard. That one is horrific. 🫣
My personal conspiracy theory is that manufacturers intentionally butcher these standards to force users into their closed systems.

PR #10370, opened 2025-10-16
(Develco MOSZB-140) - Adjust reporting parameters for temperature and illuminance

gluoncompany:
Updated reporting parameters for temperature and illuminance.
The values ​​have been better adjusted so that they do not report excessively.


Koenkk:
Thanks!

PR #10387, opened 2025-10-16
Schneider Electric EKO07259: update temperature range for thermostat

torhelge02:
Changed the temperature range according to the datasheet. 0-40°C is the maximum range. 4-30°C that it originally was is the range where the temp sensor is the most accurate.


Koenkk:
thanks!

PR #10394, opened 2025-10-17
Add Niceboy ION temperature and humidity sensor support

adampaul012:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4296

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1585,7 +1585,7 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        fingerprint: tuya.fingerprint("TS0601", ["_TZE200_s1xgth2u"]),
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE200_s1xgth2u", "_TZE200_t3xd7l44"]),
         model: "TS0601_temperature_humidity_sensor_3",
         vendor: "Tuya",
         description: "Temperature & humidity sensor",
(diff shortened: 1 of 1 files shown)


Bjk8kds:
I think I've seen similar devices, maybe just adding a fingerprint and white label to the existing code?

adampaul012:
I wanted to give it its own picture but failed to upload it... Can do of course, I've only copied the definition for the TS0601_temperature_humidity_sensor_3 anyway.

src/devices/tuya.ts
```diff
@@ -1601,6 +1601,23 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
```
Koenkk: The code is the same as `TS0601_temperature_humidity_sensor_3`, could you merge it with that? For proper detection please use `tuya.whitelabel` (search this file for examples). Use `TZE200_t3xd7l44` as the `model`

Bjk8kds: Judging from its appearance, with the clock display, is very similar to the [SZT04](https://www.zigbee2mqtt.io/devices/SZT04.html) or [SZ-T04](https://www.zigbee2mqtt.io/devices/SZ-T04.html), I didn't check the code, maybe it can be inserted into one of them?

adampaul012: I've made a new commit where I've added it to the TS0601_temperature_humidity_sensor_3 which I've tested the device with.

Koenkk:
Thanks!

PR #10403, opened 2025-10-17
Update awox.ts

markhiswords:


src/devices/awox.ts
```diff
@@ -65,6 +65,20 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Dimmable filament lamp",
         extend: [m.light()],
     },
+    {
+        zigbeeModel: ["EZMB-RGB-TW-CLB"],
+        model: "EZMB-RGB-TW-CLB",
```
Koenkk, line 70: Is there a better model number written on the device or product site?

src/devices/awox.ts
```diff
@@ -65,6 +65,20 @@ export const definitions: DefinitionWithExtend[] = [
...
+        vendor: "AwoX",
+        description: "EGLO Zigbee RGB+TW Ceiling light",
+        extend: [
+            m.deviceEndpoints({endpoints: {"1": 1, "3": 3}}),
```
Koenkk, line 74: I think this can be removed (please check if everything works fine without). Same goes for `meta: {multiEndpoint: true},`

PR #10404, opened 2025-10-17
Bosch BSD-2 (Smoke alarm II): Overwork implementation

DerDreschner:
This pull request overworks the implementation for the Bosch BSD-2 (Smoke alarm II).

Changes:
- **New**: New attribute to see if a smoke alarm is being silenced directly on the device by pushing the button for a few seconds.
- **New**: Add support to trigger a test alarm via Z2M for a variable amount of time.
- **Fixed**: Manually triggered alarms now don't automatically turn-off after a few minutes anymore. This is being done by sending the alarm in fixed intervals to the device (or broadcast it, if enabled). This means that all manually triggered alarms must now be disabled manually as well.
- **Changed**: Transmit broadcast messages twice to ensure all sleepy devices received them (mimics behavior used by Bosch).
- **Changed**: The `test` attribute is being renamed to `button_pushed`, as the previous name didn't reflect the correct meaning.
- **Removed/Changed**: The split between broadcasting alarms and enabling them on a specific device is being removed. Instead, manually triggered alarms are now automatically being broadcasted to all devices if the associated setting is being enabled.
- **Removed/Changed**: The sensitivity setting was removed. This isn't supported by Bosch, and the set values could not be verified by me. The problem is that the previous developer assumed that Bosch follows the ZCL specification (`NumberOfZoneSensitivityLevelsSupported` returns `2`, which would indicate that the available options are `0x00` - manufacturer's default -, `0x01` and `0x02` in the `CurrentZoneSensitivityLevel` attribute), which isn't the case. Therefore, I decided to force-reset the sensitivity level to the default level for safety reasons with this pull request.

<details>

<summary>Comparison screenshots</summary>

Before:
Screenshot 2025-10-16 at 10-36-40 Zigbee2MQTT Edge – Home Assistant

After:
Screenshot 2025-10-17 at 20-07-25 Zigbee2MQTT Edge – Home Assistant

</details>

This is a **breaking change**. Please check existing automations after update!

### DIFF
--- a/src/lib/bosch.ts
+++ b/src/lib/bosch.ts
@@ -398,11 +398,11 @@ export const boschGeneralSensorDeviceExtend = {
     testMode: (args: {
         testModeDescription: string;
         sensitivityLevelToUse: number;
-        supportTimeout?: boolean;
+        variableTimeoutSupported?: boolean;
         defaultTimeout?: number;
         zoneStatusBit?: number;
     }): ModernExtend => {
-        const {testModeDescription, sensitivityLevelToUse, supportTimeout = false, defaultTimeout = 0, zoneStatusBit = 8} = args;
+        const {testModeDescription, sensitivityLevelToUse, variableTimeoutSupported = false, defaultTimeout = 0, zoneStatusBit = 8} = args;
 
         const testModeLookup = {
             ON: true,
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #10414, opened 2025-10-18
add support for Salus FC600NH

mkrle:
Salus FC600NH seems to be a new version of FC600. It's sold as FC600 in Montenegro where I got it. Tested alongside an older FC600, everything that is supported by zigbee2mqtt so far works fine.


Koenkk:
Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))?

mkrle:
@Koenkk the device is externally exactly the same, so the existing FC600 image/page can be reused.

PR #10276, opened 2025-10-09
Feature: Wi-Fi Configuration via Zigbee for Shelly Gen4 Devices

dan-danache:
This PR introduces support for configuring Wi-Fi settings directly over Zigbee for all Shelly Gen4 devices. It leverages the `shellyWiFiSetupCluster`, enabling remote network setup through Zigbee2MQTT.

Supported Wi-Fi Parameters:
- Enable/disable Wi-Fi
- SSID and password
- Static IP configuration (IP, netmask, gateway, DNS)

Full Zigbee cluster specification: https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/WiFiSetupCluster/

This is a mix between https://github.com/Koenkk/zigbee-herdsman-converters/pull/9468 and https://github.com/Koenkk/zigbee-herdsman-converters/pull/9927.

I believe diagnostic data should be exposed as standalone state entries (exposes) to enable seamless integration with Home Assistant automations (for example, triggering actions when Wi-Fi is enabled or disabled on a Shelly device). In contrast, Wi-Fi setup parameters are currently exposed as a single composite entry.

image

image

---

View from Home Assistant:

image

---

### How it works

**Refresh Action (/get or UI "refresh" icon)**
When the user triggers a refresh for any Wi-Fi attribute, the converter performs the following steps:
1. Resets attributes by writing `0` to the `Action (0x02)` attribute, syncing all values with the current device configuration.
1. Reads all Wi-Fi attributes via two Zigbee **attribute read** operations, updating the internal state and UI accordingly.

**Apply Action (/set wifi_config or UI "Apply" button)**
When the user applies a new configuration, the converter executes:
1. Writes all editable Wi-Fi attributes to the Zigbee cluster via two Zigbee **attribute write** operations
1. Applies the configuration by writing `1` to the `Action (0x02)` attribute, committing the current Zigbee attribute values to the device.

**Note**: The Wi-Fi password must be provided each time the "Apply" action is used, as it is not stored in the device state.

---

I tested this code on a **Mini 1 Gen4** device with firmware **1.7.1**, using both DHCP and static configurations, as an [external converter](https://gist.github.com/dan-danache/54200d51103fe3f0145179d0a59cce24) in my production environment. For this pull request, I made minimal adjustments to ensure compatibility, primarily updating import paths and TypeScript types where necessary.

Have fun!

PR #10423, opened 2025-10-19
Shelly: Improve Wi-Fi cluster attributes handling

dan-danache:
Shelly: Improve Wi-Fi cluster attributes handling to use smaller Zigbee frame size. See discussions from https://github.com/Koenkk/zigbee-herdsman-converters/pull/10276.


Koenkk:
Thanks!

PR #10424, opened 2025-10-19
MIR - IL100 - ZB has added sensitivity…

yuanxin32323:
This update is an official version, which has added sensitivity settings. The previous MIR-IL100 was a test model and not launched on the market, so it is directly overwritten in this update.


Koenkk:
Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))?

PR #10434, opened 2025-10-20
Adjust temperature delta min and step values

Liionboy:
1. .withValueMin(0.1) - Changed from 0.5 to 0.1
2. .withValueStep(0.1) - Changed from 0.5 to 0.1

This now allows the temperature delta to be set from 0.1°C to 10°C in 0.1°C increments, giving you much finer control over when the heating should trigger based on the difference between the current temperature and the setpoint.


Koenkk:
Thanks!

pabz7:
Hi folks - any chance this could be implemented on the Moes ZHT-SR?? 

Thanks a million

PR #10385, opened 2025-10-16
Enhance device definitions for leak and CO sensors

1tigris1:
Added datapoints and constants for water leak and CO sensors.

Link to picture pull request: TODO

PR #10437, opened 2025-10-20
Add support for Nous water leak sensor features

1tigris1:
Please check if it’s done correctly — if everything is fine, I’ll add the second one.


Koenkk:
Added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/10385

PR #10442, opened 2025-10-20
Add new fingerprint to TS0601 Human Presence Sensor

bfmatei:
I just received two TS0601 Human Presence sensors. One of them was recognized automatically `_TZE284_debczeci`, while `_TZE204_debczeci` was not.

I just completely tested this and it works fine.


Koenkk:
Thanks!

PR #10471, opened 2025-10-23
Aqara: better temperature sensor for SRTS-A01

allexoll:
Aqara SRTS-A01 currently exposes `device_temperature` which I suspect is the microcontroller's internal temperature (usually a few degrees higher than the air temp, and only has 1° precision, whereas this one has 0.1°)

The device also has `local_temperature` which seems to be the air temperature, and already is the measured temperature for the climate entity. this means that `local_temperature` is not exposed as a sensor. This fixes it.

Potentially, we could keep both, but I'm not familiar enough with the dev of this project to make it clear which is which (if you keep both, you have 2 sensors exposed, and it's difficult to know which is which because they both are named temperature.)

Also, since local_temperature is only access.STATE, you have to wait until it is reported to get a value, so it might be unavailable on startup?

Open for suggestions

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -3623,6 +3623,7 @@ export const definitions: DefinitionWithExtend[] = [
             e.battery_voltage(),
             e.battery(),
             e.power_outage_count(),
+            e.device_temperature(),
             e.local_temperature().withAccess(ea.STATE),
             e.schedule(),
             e


Koenkk:
Let's keep both to prevent a breaking change.

allexoll:
done!

Koenkk:
thanks!

Koenkk:
I reverted this because it caused https://github.com/Nerivec/zigbee2mqtt-windfront/issues/311, after taking a closer look it seems it was exposed twice (also with `.withLocalTemperature(ea.STATE, "Current temperature measured by the internal or external sensor")`). Where was the local temperature not exposed?

allexoll:
It is already exposed but only as part of the climate entity, which makes it not available as a sensor in itself. 

with the additional "standalone" `local_temperature` as it was in this PR:
Screenshot 2025-11-01 at 10 22 04

without the change, it looks like this:
Screenshot 2025-11-01 at 10 30 10

without it, the sensor part is empty, and so if you want to use, let's say BetterThermostat, the only temperature sensor available is the `device_temperature` which is part of the diagnostics, and is a few degrees higher. 

Maybe my change is not the correct way to add it, so I'm open to suggestions, but I think the actual temperature should be available, and not through the climate entity, also because it is muxed behind "either this or an external sensor"

What do you think?

PR #10485, opened 2025-10-23
LiXee ZLinky: fix polling interval + improve option descriptions

sbraz:
* fix: LiXee ZLinky: make polling interval configurable again
  e30ace459d702 broke the polling interval configuration. Since then, it
was always set to 600 seconds.

* fix: LiXee ZLinky: describe actual default polling interval
  The default description ("This device does not support reporting
  electric measurements so it is polled instead. The default poll interval
  is 60 seconds, set to -1 to disable.") mentioned a default of 60 seconds
  but it is actually 600.
  It was also misleading: not all attributes are polled, several of them
  have reporting configured and are automatically updated.

* fix: LiXee ZLinky: improve description of measurement_poll_chunk

### DIFF
--- a/src/devices/lixee.ts
+++ b/src/devices/lixee.ts
@@ -1806,7 +1806,8 @@ export const definitions: DefinitionWithExtend[] = [
                 .numeric("measurement_poll_chunk", ea.SET)
                 .withValueMin(1)
                 .withDescription(
-                    "During the poll, request multiple exposes to the Zlinky at once for reducing Zigbee network overload. Too much request at once could exceed device limit. Requires Z2M restart. Default: 4",
+                    "Number of attributes requested from the ZLinky in each poll to reduce Zigbee network load. " +
+                        "Requesting too many at once may exceed the device's limit and cause read errors. Requires Z2M restart. Default: 4.",
                 ),
             e
                 .text("tic_command_whitelist", ea.SET)


src/devices/lixee.ts
```diff
@@ -1896,7 +1896,13 @@ export const definitions: DefinitionWithExtend[] = [
             m.poll({
                 key: "measurement",
                 defaultIntervalSeconds: 600,
-                option: exposes.options.measurement_poll_interval(),
+                option: exposes.options
+                    .measurement_poll_interval()
```
sbraz, line 1901: Should I use `e.numeric("measurement_poll_interval",` instead? I wasn't sure.

sbraz:
Ideally, I'd like to move `measurement_poll_interval` above `measurement_poll_chunk` but I couldn't find how to do it. Is that possible at all?

src/devices/lixee.ts
Koenkk: this is fine

src/devices/lixee.ts
```diff
@@ -1894,9 +1895,15 @@ export const definitions: DefinitionWithExtend[] = [
         ota: {manufacturerName: "LiXee"}, // TODO: not sure if it's set properly in device
         extend: [
             m.poll({
-                key: "interval",
+                key: "measurement",
```
Koenkk, line 1898: This should stay as is

sbraz: @Koenkk but it doesn't work as-is, the option is called `measurement_poll_interval` and the poll function checks `${args.key}_poll_interval`, so the key must match the option name.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/f7323eb423aa8dd2fd4dcc796e72f16cd1e10282/src/lib/modernExtend.ts#L663

Am I missing something? Before my fix commit, the poll interval was always 600 seconds, no matter the value of `measurement_poll_interval`.

sbraz:
> Ideally, I'd like to move `measurement_poll_interval` above `measurement_poll_chunk` but I couldn't find how to do it. Is that possible at all?

@Koenkk can you help me with that?

sbraz:
@Koenkk I have another issue with this converter: I believe this never raises an exception on read errors, is that possible?
https://github.com/Koenkk/zigbee-herdsman-converters/blob/f7323eb423aa8dd2fd4dcc796e72f16cd1e10282/src/devices/lixee.ts#L1918
I'm not very good with JS/TS but if I replace `await endpoint.read(…).catch` with a try/except block, I see read errors when `measurement_poll_chunk` is too high. I believe these silent failures were the reason my ZLinky was acting weird in the first place.
Can you help me find a proper fix for this? Do you want me to create a separate issue?

Koenkk:
> Ideally, I'd like to move measurement_poll_interval above measurement_poll_chunk but I couldn't find how to do it. Is that possible at all?

What do you mean with above?

> I'm not very good with JS/TS but if I replace await endpoint.read(…).catch with a try/except block, I see read errors when measurement_poll_chunk is too high

using try/catch is fine here, please update the pr

sbraz:
> What do you mean with above?


I'd like to move the poll interval option above the poll chunk option in the UI:
image

> using try/catch is fine here, please update the pr

OK I'll do that ASAP.

Koenkk:
Currently it's not possible to define the order, so I would propose to leave this as is.

sbraz:
> using try/catch is fine here, please update the pr

I actually reverted to the current version of the converter, without any change, and I do see the error logs:
```
[2025-10-25 20:33:20] warning:  zhc:lixee: Failed to read zigbee attributes: Error: ZCL command 0x......./1 liXeePrivate.read(["currentPrice","message1","message2","drawnVAMaxN1"], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"reservedBits":0,"manufacturerCode":null,"writeUndiv":false}) failed ({"target":24165,"apsFrame":{"profileId":260,"clusterId":65382,"sourceEndpoint":1,"destinationEndpoint":1,"options":4416,"groupId":0,"sequence":200},"zclSequence":14,"commandIdentifier":1} timed out after 10000ms)
```
I think the reason I didn't see them in the logs is that the poll interval was always 10 minutes, so I missed them. It was not a JavaScript problem.

I'd say this PR is good as-is if you're OK with it.

Koenkk:
Thanks!

PR #10493, opened 2025-10-24
Extend zigbeetlc.js by ZY-ZTH02-z

dbmaxpayne:
Picture can remain the same.
However, new device definitions were added by pvvx and my device is now reported as "ZY-ZTH02-z".
(TS0201 _TZ3000_v1w2k9dd)

Should be fully compatible, though.

I Tried testing this with a local external_converters js, but even though no error appears in the logs, nothing happens.


Koenkk:
Thanks!

PR #10511, opened 2025-10-25
fix: use node:zlib for crc calc

Nerivec:
https://nodejs.org/api/zlib.html#zlibcrc32data-value

Should set 20.15.0 as min for Z2M too, some earlier Node versions are already causing problems (import, type errors...), we should at least output the warning in case some users are still using them.


Koenkk:
Thanks!

Koenkk:
Done in https://github.com/Koenkk/zigbee2mqtt/pull/29284

PR #10514, opened 2025-10-25
fix: use values from parameters for pilot_wire_mode

Gouar:
It is possible to override the different values for the predefined pilot wire mode feature but the values aren't actually used.


Koenkk:
Thanks!

PR #10516, opened 2025-10-25
Add Thaleos TWC-R01 thermostat support

Gouar:
Added support for Thaleos TWC-R01 smart thermostat with pilot wire features, including various datapoints and configurations.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4330

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -19937,15 +19937,15 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Smart thermostat for electric heater with pilot wire",
         extend: [tuya.modernExtend.tuyaBase({dp: true})],
         exposes: [
-    		e
-    			.climate()
-    			.withLocalTemperature(ea.STATE)
-    			.withLocalTemperatureCalibration(-8, 8, 0.5, ea.STATE_SET)
-    			.withPreset(["comfort", "eco", "antifrost", "off", "comfort_1", "comfort_2"]),	
-    		e
-    			.enum("mode", ea.STATE, ["comfort", "eco", "antifrost", "off", "comfort_1", "comfort_2"])
-    			.withDescription("Current running mode")
-    			.withCategory("diagnostic"),
+            e
+                .climate()
+                .withLocalTemperature(ea.STATE)
+                .withLocalTemperatureCalibration(-8, 8, 0.5, ea.STATE_SET)
+                .withPreset(["comfort", "eco", "antifrost", "off", "comfort_1", "comfort_2"]),
+            e
+                .enum("mode", ea.STATE, ["comfort", "eco", "antifrost", "off", "comfort_1", "comfort_2"])
+                .withDescription("Current running mode")
+                .withCategory("diagnostic"),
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -19688,6 +19688,78 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE204_ilzkxrav"]),
+        model: "TWC-R01",
+        vendor: "Thaleos",
+        description: "Smart thermostat for electric heater with pilot wire",
+        extend: [tuya.modernExtend.tuyaBase({dp: true})],
+        exposes: [
```
Koenkk, line 19939: Could you use `e.climate()` here (similar as other thermostats in this file?) .

src/devices/tuya.ts
```diff
@@ -19688,6 +19688,78 @@ export const definitions: DefinitionWithExtend[] = [
...
+            e
+                .enum("device_mode_type", ea.STATE, ["four", "six"])
+                .withDescription("Indicates the actual pilot wire mode of the thermostat")
+                .withCategory("diagnostic"),
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [
+                    2,
+                    "pilot_wire_mode",
```
Koenkk: I think this should be named `preset` (other Tuya thermostats also do that I think)

src/devices/tuya.ts
Gouar: you're right, I gave a better look at Climate class and it's more appropriate

Koenkk:
Thanks!

Koenkk:
@Gouar I had to revert this PR because it makes a test fail, this is because it does not a setpoint (every thermostat should support it). Doesn't this thermostat allow to set a setpoint?

Gouar:
Hi @Koenkk , this one doesn't allow to set a setpoint. 
It's a cheap one only to use with pilote wire mode (french system to control "dumb" heating devices). I can't even find a similar model on aliexpress.
I can add back withSetpoint function call but it will display the "set temperature card" on HA.
tuya dps for reference;
```
{
	"2": "Mode",
	"11": "Power",
	"16": "Current temperature",
	"19": "Temperature correction",
	"20": "Fault alarm",
	"40": "ECO mode",
	"101": "Weekly Porgram",
	"102": "Work State",
	"103": "Day1",
	"104": "周程序数据2",
	"105": "周程序数据3",
	"106": "周程序数据4",
	"107": "周程序数据5",
	"108": "周程序数据6",
	"109": "周程序数据7",
	"110": "OWD Enable",
	"111": "OWD_Temp",
	"112": "OWD_Time",
	"113": "OWD_Timeout",
	"114": "FP_Mode6",
	"115": "Voltage",
	"116": "Current",
	"117": "Consumption",
	"118": "work_pwd_clear",
	"119": "Today",
	"120": "Yesterday"
}
```

Koenkk:
In that case lets get rid of the climate expose and use just numeric exposes

PR #10519, opened 2025-10-25
fix: LiXee ZLinky: handle empty TIC command list

sbraz:
If the user defines a value for tic_command_whitelist and then clears
    it, it remains in the configuration as an empty string.
    An empty string should be the same as no value at all. Before this
    change, when the string was empty, no attributes were polled.
    
   Let's also suggest a Z2M restart in the description, it's the easiest way to
    restart the poll.

### DIFF
--- a/src/devices/lixee.ts
+++ b/src/devices/lixee.ts
@@ -1746,7 +1746,7 @@ function getCurrentConfig(device: Zh.Device, options: KeyValue) {
     }
 
     // Filter exposed attributes with user whitelist
-    if (options?.tic_command_whitelist != null) {
+    if (options?.tic_command_whitelist) {
         // @ts-expect-error ignore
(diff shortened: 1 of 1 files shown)


src/devices/lixee.ts
```diff
@@ -1811,7 +1811,7 @@ export const definitions: DefinitionWithExtend[] = [
                 ),
             e
                 .text("tic_command_whitelist", ea.SET)
-                .withDescription("List of TIC commands to be exposed (separated by comma). Reconfigure device after change. Default: all"),
+                .withDescription("List of TIC commands to be exposed (separated by comma). Requires Z2M restart. Default: all"),
```
sbraz: @Koenkk I found that changing the poll interval triggers an update of the list but I think it's better to just suggest a restart.

Koenkk:
Thanks!

PR #10528, opened 2025-10-26
Create 24ch Switche

A-J88:
The picture is not the correct one. I don't now ho to change that yet


Koenkk:
could you add this to [tuya.ts](https://github.com/Koenkk/zigbee-herdsman-converters/edit/master/src/devices/tuya.ts)?

Bjk8kds (commented):
If there are changes from an existing device, please edit the existing file, 
In this case, since it's a Tuya device, edit the `‎src/devices/tuya.ts` file.

PR #10529, opened 2025-10-26
fix: improve store

Nerivec:
Should remove quite a lot of dupe calls.
Also adds better typing. Note: removing the `any` on store would require a lot of type checking/casting throughout the codebase, so for now left as `any`.

@Koenkk Left a couple of TODOs regarding the behavior of "key is present" vs "key's value is undefined" (previous). "key is present" provides a better logic (has/get/put/clear), since we expect a clear call if the key is no longer desired.

### DIFF
--- a/src/lib/store.ts
+++ b/src/lib/store.ts
@@ -24,10 +24,7 @@ export function hasValue(entity: Zh.Endpoint | Zh.Group | Zh.Device | string, ke
     const entityKey = getEntityKey(entity);
     const entry = store.get(entityKey);
 
-    // TODO: behavior change => if key exists, return true
     return entry !== undefined && key in entry;
-    // previous behavior:
-    // return entry?.[key] !== undefined;
 }
 
 // biome-ignore lint/suspicious/noExplicitAny: generic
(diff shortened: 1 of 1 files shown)


src/lib/store.ts
```diff
@@ -14,41 +15,57 @@ function getEntityKey(entity: Zh.Endpoint | Zh.Group | Zh.Device | string) {
...
 
-export function getValue(entity: Zh.Endpoint | Zh.Group | Zh.Device | string, key: string, default_: unknown = undefined) {
+// biome-ignore lint/suspicious/noExplicitAny: generic
+export function getValue(entity: Zh.Endpoint | Zh.Group | Zh.Device | string, key: string, fallback: unknown = undefined): any {
     const entityKey = getEntityKey(entity);
-    if (store.has(entityKey) && store.get(entityKey)[key] !== undefined) {
-        return store.get(entityKey)[key];
+    const entry = store.get(entityKey);
+
+    // TODO: behavior change => if key exists, return its value
```
Koenkk: I removed all `undefined` and `null` put calls, so I think that should be fine now? (TODOs can be removed)

Nerivec: I missed that `undefined` put!
The `null` would not have been considered "not present" before though (only `undefined`, it was triple eq everywhere), so, is it correct to clear them?

Koenkk: I think for those cases yet (it was mainly use to clear timers)

Koenkk:
Thanks!

PR #10550, opened 2025-10-28
Update eglo.ts

Basti545:


src/devices/eglo.ts
```diff
@@ -7,13 +7,28 @@ const e = exposes.presets;
 const ea = exposes.access;
 
 export const definitions: DefinitionWithExtend[] = [
+    {
+        zigbeeModel: ["EZMB-RGB-TW-CLB"],
+        model: "EZMB-RGB-TW-CLB",
+        vendor: "AwoX",
+        description: "Automatically generated definition",
```
Koenkk: Could you add a proper description here?

src/devices/eglo.ts
```diff
@@ -7,13 +7,28 @@ const e = exposes.presets;
...
+        extend: [
+            m.deviceEndpoints({endpoints: {"1": 1, "3": 3}}),
```
Koenkk: Can you check if it also works without this?

Koenkk:
Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `EZMB-RGB-TW-CLB.png`

src/devices/eglo.ts
Basti545: done

Basti545:
done

src/devices/eglo.ts
Basti545: Sorry,  i dont know how to check. This is my first step in Home Assistant and Zigbee :)
I followed exactly the instructions given to me by Zigbee2MQTT.

Koenkk: Did you test code through e.g. an ext converter?

Basti545: i followed the instructions and created a external definition. And after that i created the pull request like the instruction says.

Koenkk: Then in your external definition, can you check if it also works when removing `m.deviceEndpoints({endpoints: {"1": 1, "3": 3}}),`?

Basti545: I've added the external definitions code from the device in my config (without the part m.deviceEndpoints({endpoints: {"1": 1, "3": 3}}),)
and restarted zigbee2mqtt. After that i have a red sign that says "Device interview failed", but the device still shows up with all options e.g Brightness, color temp, effect, action etc.

Koenkk: I don't expect that the red sign has to do with that, to fix it try to factory reset/re-pair the device. I think the line can be removed safely.

Basti545: You're right. The problem wasn't related to that. I re-paired the device and now the red sign is gone. :)

Koenkk:
I jus noticed this device was already added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/10599, can you check if it works out of the box in z2m 2.6.3?

Basti545:
> I jus noticed this device was already added in #10599, can you check if it works out of the box in z2m 2.6.3?

No, if I delete my external support file and restart zigbee2mqtt, the device says : not supported: generated

Koenkk:
Ah true, it has only been added to the dev branch yet, can you check that?

PR #10551, opened 2025-10-28
feat: Add support for Senoro window alarm v2 with various states and …

ChristianGr1974:
…configurations


Koenkk:
This device is already supported but not with all the features, could you update: 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/36f3a2b859b6b5b25c2cfe7fcd7072702b64a6ce/src/devices/tuya.ts#L1995

PR #10561, opened 2025-10-28
Add new device definition to fix weekly schedule day mapping for _TZE200_p3dbf6qs thermostat

manu469:
This PR adds a new device definition to fix a weekday mapping issue for the Avatto MS167/TS0601 TRV
(fingerprint: _TZE200_p3dbf6qs).

Issue:
The thermostat internally shifts the weekly schedule by +2 days.
For example, the schedule configured for Sunday is actually applied on Tuesday,
Monday is applied on Wednesday, etc. This causes the Zigbee2MQTT UI to display
the wrong schedule values on the wrong days.

Fix:
A dedicated device definition has been added for this fingerprint. The datapoint-to-day
mapping for DPs 28–34 has been corrected to align with the actual behavior of the device,
so each schedule is now applied to its correct weekday.

Impact:

Only this fingerprint is affected; no changes for other TS0601 TRVs.
The core value converters remain unchanged.
The UI now reflects the correct days, and schedules are applied correctly
by the TRV.
Testing:

Verified reading existing schedules.
Verified writing new schedules.
Confirmed that the schedules match the correct weekdays after TRV reboot.
This approach avoids modifying the core logic and keeps the fix isolated to the affected device.

Thanks!

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6556,15 +6556,11 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint("TS0601", [
-            "_TZE200_p3dbf6qs" /* model: 'ME167_1', vendor: 'AVATTO' */,
-        ]),
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE200_p3dbf6qs" /* model: 'ME167_1', vendor: 'AVATTO' */]),
         model: "TS0601_thermostat_5",
         vendor: "Tuya",
         description: "Thermostatic radiator valve",
-        whiteLabel: [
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Offebaescher:
I received 6 such Thermostats, identified as Zigbee-Modell TS0601, _TZE200_hvaxb2tc (Thermostatic radiator valve)
Modell TRV06_1 AVATTO

They also have the schedule shifted by +2. 
I verified that everything i set for schedule_wednesday got executed today on Friday....
What do i have to provide or what has to be done that that device gets corrected like you did with this one (_TZE200_p3dbf6qs) from this thread?

(just for informational purposes: i also have one thermostat which is identified as TS0601, _TZE284_o3x45p96 and there the week schedule in auto mode is also somehow shifted and changing the weekdays during the day in no logic which i can understand)

PR #10578, opened 2025-10-30
Add reverse water consumption, fix flow_rate and update report period

Liionboy:
In this new version i fixed flow_rate, add some new sensor for daily, month consumption and water revers consumption.

The only problem with Flow rate is because reporting is every hour not instant like he say in documentation, so if water is open at every hour at 10 minutes , you will se that sensor give you data

Link to picture pull request: 
image


Koenkk:
Thanks!

PR #10582, opened 2025-10-30
fix: LiXee ZLinky: rename misspelt SMAXSN attributes

sbraz:
The attributes are named SMAXSN, not SMAXN, as per the [Enedis-MOP-CPT_002E](https://www.enedis.fr/media/2035/download) reference document:
image

ZLinky's firmware uses the correct names, e.g. https://github.com/fairecasoimeme/Zlinky_TIC/blob/eb7a71a72118d069f7ecf8ea4478a0fce06ed3a7/ZLinky/Source/App_Linky.c#L660

See also https://github.com/fairecasoimeme/Zlinky_TIC/pull/368 and https://github.com/fairecasoimeme/Zlinky_TIC/issues/365.

Users who were using SMAXN* in `tic_command_whitelist` will need to update the list.

### DIFF
--- a/src/devices/lixee.ts
+++ b/src/devices/lixee.ts
@@ -1163,7 +1163,7 @@ const singlePhaseData = [
         att: "activePowerMax",
         reportable: true,
         onlyProducer: false,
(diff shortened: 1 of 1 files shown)


sbraz:
@fairecasoimeme are you OK with this change?

Koenkk:
Thanks!

PR #10586, opened 2025-10-30
Add Zemismart ZMO-606-S2 as TS0003_switch_3_gang_with_backlight

bernardesarthur:
Add Zemismart ZMO-606-S2 as TS0003_switch_3_gang_with_backlight. I've tested and it works fine, no breaking change.


Koenkk:
Thanks!

PR #10587, opened 2025-10-30
Add new one ID "_TZE204_vvvtcehj" for Roximo CRTZ01 model in tuya.ts

maxustm:
Add new one ID "_TZE204_vvvtcehj" for curtain motor by Roximo,  model CRTZ01


Koenkk:
Thanks!

PR #10621, opened 2025-11-01
Add Climax power plug device configuration

srandier:
Added model "PSMP5_00.00.03.12TC"
This is branded as PSM-29ZBS by Climax. Same as PSM-29ZBSR except PSM-29ZBS is an end device whereas PSM-29ZBSR is a router.
Link to picture pull request: [image for PSM-29ZBS](https://github.com/Koenkk/zigbee2mqtt.io/pull/4376)

### DIFF
--- a/src/devices/climax.ts
+++ b/src/devices/climax.ts
@@ -73,7 +73,6 @@ export const definitions: DefinitionWithExtend[] = [
         model: "PSM-29ZBS",
         vendor: "Climax",
         description: "Power plug",
-        fromZigbee: [fz.on_off, fz.metering],
         extend: [m.onOff({powerOnBehavior: false}), m.electricityMeter({current: false, voltage: false})],
     },
     {


src/devices/climax.ts
```diff
@@ -68,6 +68,22 @@ export const definitions: DefinitionWithExtend[] = [
         whiteLabel: [{vendor: "Blaupunkt", model: "PSM-S1"}],
         exposes: [e.switch(), e.power(), e.energy()],
     },
+    {
+        zigbeeModel: ["PSMP5_00.00.03.12TC"],
+        model: "PSM-29ZBS",
+        vendor: "Climax",
+        description: "Power plug",
+        fromZigbee: [fz.on_off, fz.metering],
```
Koenkk: Can you replace all of the code below with:

```
extend: [m.onOff(), m.electricityMeter()],
```

and see if that works?

srandier: Thanks for your review! 

Initially I cloned the code which was found in climax.ts for the similar device model (PSM-29ZBSR)

With your suggested change that seems to work.
I made another commit.

I tested using the following external converter:
[climax_psm-29zbs_1.js](https://github.com/user-attachments/files/23287131/climax_psm-29zbs_1.js)
=> I am able to control the device and to read state power and energy. 

However, when trying to set the power on behavior. 
[11/1/2025, 11:47:52 PM] z2m: Publish 'set' 'power_on_behavior' to 'PP_PSM29ZBS_1d6f' failed: 'Error: Got `UNSUPPORTED_ATTRIBUTE` error, device does not support power on behaviour'

I think that model does not support defining the power on behaviour. Do we need to change something in the code to account for that?
Thanks.

Koenkk: Yes, use `extend: [m.onOff({powerOnBehavior: false}), m.electricityMeter()],`

srandier: Thanks! OK, fixed and successfully tested on my side

Also it looks like this power plug does not report current and voltage so I tweaked the extend as follows:
`extend: [m.onOff({powerOnBehavior: false}), m.electricityMeter({current: false, voltage: false})],`

In addition I can see that some entities ending with `_1` are published, like the `state_1` which does not have the right value. Is there a way to fix this?
```
[11/3/2025, 8:52:08 AM] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/PP_PSM29ZBS_1d6f', payload '{"energy":711.1,"energy_1":706.7,"linkquality":88,"power":145,"power_1":146.4,"produced_energy":0,"produced_energy_1":0,"state":"OFF","state_1":"ON"}'
[11/3/2025, 8:52:09 AM] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/PP_PSM29ZBS_1d6f', payload '{"energy":711.1,"energy_1":706.7,"linkquality":88,"power":17.8,"power_1":146.4,"produced_energy":0,"produced_energy_1":0,"state":"OFF","state_1":"ON"}'
```

Koenkk: These are old cached state values, stop z2m, remove `data/state.json` and starting z2m should fix it.

srandier: Right, that resolved the issue!
So, unless you need me to test anything, I guess this pull request is ready to merge :)

Koenkk:
Thanks!

Issue #10617, opened 2025-11-01
[External Converter] TS0601 from _TZE204_aai5grix (Wenzhi MTD285-ZB)


Bjk8kds:
**Wenzhi 24G Human Presence Sensor MTD285-ZB**

Following up on the issue here:
https://github.com/Koenkk/zigbee2mqtt/issues/29386

I found the original external converter [here](https://github.com/wzwenzhi/Wenzhi-ZigBee2mqtt/blob/main/mtd285_251022.js).
Then I adjusted it a little. corrected the uncompleted code and some typos too.
The codes also work for the `_TZE284_aai5grix` in addition to this `_TZE204_aai5grix`, so I just left it as is.
I didn't change the code further because I'm not familiar with the standard Zigbee2MQTT reference (if there is one).
Please let me know and give suggestions if there is anything I need to adjust, e.g. naming, feature descriptions, etc.
Should I pull request?
***
This is my external converter for `TS0601` from `_TZE204_aai5grix`
software_build_id: `undefined`
date_code: ``
endpoints:
```json
{"1":{"clusters":{"input":["genGroups","genScenes","manuSpecificTuya","genBasic"],"output":["genOta","genTime"]}},"242":{"clusters":{"input":[],"output":["greenPower"]}}}
```

### What works / what doesn't?
AFAIK everything works fine

### Converter

```js
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const e = exposes.presets;
const ea = exposes.access;
const tuya = require('zigbee-herdsman-converters/lib/tuya');

// Extended custom data points
const tzDatapoints = {  
    ...tuya.tz.datapoints,  
    key: Array.isArray(tuya.tz.datapoints.key)   
        ? [...tuya.tz.datapoints.key, ...Array.from({length: 11}, (_, i) => `gate_enable_${String(i+1).padStart(2, '0')}`),
            'thr_cfg_range_gate', 'motion_thr', 'presence_thr', 'nearest_target_gate',
            'target_countdown', 'target_velocity', 'debug_switch', 'led_mode', 'delay_time',
            'block_time', 'judge_alg', 'env_noise_collect', 'device_control', 'p_sensitivity',
            'm_sensitivity', 'scene_mode', 'illum_report_sw', 'motion_detec_sw', 'dist_report_sw', 'speed_report_sw'
        ] 
[...]

PR #10642, opened 2025-11-03
Add support for Wenzhi MTD285-ZB human presence sensor

Bjk8kds:
Added support for Wenzhi MTD285-ZB human presence sensor with some adjustments and corrections from the original code.
Fingerprints:
- _TZE284_aai5grix
- _TZE204_aai5grix

This is my first time commit PR for new device. Is it safe to leave comments in the code? If they're too distracting, feel free to remove them.

Original code thanks to https://github.com/wzwenzhi/Wenzhi-ZigBee2mqtt

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/29386
Fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/10617
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/28988
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/28779
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/29281

Related discussion https://github.com/Koenkk/zigbee2mqtt/discussions/29173

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4384

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -14680,9 +14680,7 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [tuya.modernExtend.tuyaBase({dp: true})],
         exposes: [
             e.presence(),
-            e
-                .enum("state", ea.STATE, ["none", "presence", "move"])
-                .withDescription("Presence state"),
+            e.enum("state", ea.STATE, ["none", "presence", "move"]).withDescription("Presence state"),
             e.illuminance().withDescription("Measured illuminance"),
             e
                 .numeric("min_distance", ea.STATE_SET)


src/devices/tuya.ts
```diff
@@ -14671,6 +14671,164 @@ export const definitions: DefinitionWithExtend[] = [
...
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [
+                    1,
+                    null,
+                    {
+                        from: (v: number) => ({
+                            occupancy: v !== 0,
+                            presence: v === 0 ? "none" : v === 1 ? "presence" : "motion",
```
```suggestion
                            presence: v !== 0,
                            presence_state: v === 0 ? "none" : v === 1 ? "presence" : "motion",
```

I think this is better?

Bjk8kds: Wouldn't it be better to use just `state` like this?
This would match the other similar devices. Let me know which you prefer.
I'll also change the exposes.

```ts
        e.presence(),
        e.enum("state", ea.STATE, ["none", "presence", "motion"]).withDescription("Presence state"),



            presence: v !== 0,
            state: v === 0 ? "none" : v === 1 ? "presence" : "motion",

```
So what about the use of the word `motion` all over this code?
I think it would be better to change it to `move` as well to match other similar devices?
And it will also match the options in the `judge_logic`.

Koenkk: Sounds good!

Bjk8kds: OK. Done

Koenkk:
Thanks!

PR #10643, opened 2025-11-03
fix: [CK-BL702-AL-01(7009_Z102LG03-1) / eWeLink] HA cannot turn on li…

atamanishche:
Fix for https://github.com/Koenkk/zigbee2mqtt/issues/29197
It may also be used to address the issue in https://github.com/Koenkk/zigbee2mqtt/issues/29184


Koenkk:
Could you integrate this into `light_onoff_brightness` and a a new `meta` option to enable this behaviour?

atamanishche:
@Koenkk logic was moved into light_onoff_brightness

src/converters/toZigbee.ts
Koenkk, line 1288: I'm noticing this option now, could you check if it works when setting `moveToLevelWithOnOffDisable` in `meta` instead of `onAfterMoveToLevel`?

atamanishche: In my case, using `moveToLevelWithOnOffDisable` causes the light bulb to turn on with an incorrect brightness level (the actual brightness matches the last level before turning off, while the status in home assistant shows the level specified in the call).

atamanishche: It might be possible to solve the problem by making changes to the `if (moveToLevelWithOnOffDisable)` block.

I’ll check.

PR #10657, opened 2025-11-04
Added the illuminance attribute to the MIR-IL100 device.

yuanxin32323:
Added the illuminance attribute to the MIR-IL100 device.


Koenkk:
Thanks!

PR #10659, opened 2025-11-04
[Feature] Basic RPC cluster code and LED control for Shelly Powerstrip

DataGhost:
This should address https://github.com/Koenkk/zigbee2mqtt/issues/28974. 

The RPC endpoint is bugged in the current firmware on this device, as confirmed by Shelly. It won't return anything useful on read, so this pull request only implements writes to the various exposed settings. Other devices that expose the RPC endpoint might, though, and a ton of other settings can be accessed through that endpoint so I left a bunch of dev code still in and tried to do some work of making it modular so the parts can be reused without much trouble on other devices. Once the firmware bug is fixed the sending/receiving code can be improved and cleaned up, but they haven't supplied any ETA.

I personally really dislike how huge the `ledColors` thing has become, there doesn't seem to be a single control that can pick RGB and I tried at least putting the separate percentage controls in a nested composite but that resulted in the values not being passed on to my code anymore. It also seems more logical to hide these settings if the `ledMode` isn't set to `switch`. Not sure if this is a problem that should be addressed or that it's fine to have a huge "exposes" page.

I have some questions about the typing though, I found the type checker to be very strict in some weird ways. E.g. lines 92-97 should do the exact same thing but one was correct and the other one was considered an error. I find the "correct" way to be needlessly verbose.

A similar thing is going on on lines 124-130. Basically, in order to get it to work properly I had to just change the type to `any` and tell biome to ignore it. That was accepted in https://github.com/Koenkk/zigbee-herdsman-converters/pull/10276 as well but that feels "unfinished". What type should I be using there? I tried using `Zh.Endpoint | Zh.Group` as works on lines 80/115 but I feel that that's too generic and that's why this isn't working.

I hope the rest of the code is okay enough, after answers to the above questions I can clean it up so it can be merged.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -329,12 +329,6 @@ const shellyModernExtend = {
             },
         ];
 
-        const configure: Configure[] = [
-            async (device, coordinatorEndpoint, definition) => {
-                // Nothing to do here
-            },
-        ];
-
         if (featureDev) {
             exposes.push(...exposesDev);
             toZigbee.push(...toZigbeeDev);
@@ -343,7 +337,7 @@ const shellyModernExtend = {
             exposes.push(...exposesPowerstripUI);
             toZigbee.push(...toZigbeePowerstripUI);
         }
-        return {exposes, fromZigbee, toZigbee, configure, isModernExtend: true};
+        return {exposes, fromZigbee, toZigbee, isModernExtend: true};
     },
     shellyWiFiSetup(): ModernExtend {
         // biome-ignore lint/suspicious/noExplicitAny: generic


src/devices/shelly.ts
```diff
@@ -45,6 +48,313 @@ const shellyModernExtend = {
...
+                .withFeature(featurePercentage("onBrightness", "Brightness (on)"))
+                .withFeature(featurePercentage("offR", "Red (off)"))
+                .withFeature(featurePercentage("offG", "Green (off)"))
+                .withFeature(featurePercentage("offB", "Blue (off)"))
+                .withFeature(featurePercentage("offBrightness", "Brightness (off)"))
+                .withLabel("LED colors in 'switch' mode")
+                .withCategory("config"),
+            featurePercentage("ledPowerBrightness", "LED brightness in 'power' mode").withCategory("config"),
+            e
+                .composite("ledNightMode", "ledNightMode", ea.ALL)
```
Koenkk: Could you make all feature names and values snake case (e.g. `led_night_mode`)

src/devices/shelly.ts
```diff
@@ -45,6 +48,313 @@ const shellyModernExtend = {
...
+            {
+                cluster: "shellyRPCCluster",
+                type: ["attributeReport", "readResponse"],
+                convert: (model, msg, publish, options, meta) => {
+                    const state: KeyValue = {};
+
+                    // Diagnostic data
+                    if (msg.data.rxCtl !== undefined) {
+                        state.rpcRxCtl = msg.data.rxCtl;
+                        state.rpcData = "";
```
```suggestion
                        state.rpc_data = "";
```

(same applies to other attributes)

src/devices/shelly.ts
```diff
@@ -45,6 +48,313 @@ const shellyModernExtend = {
...
+
+        const featureDev = features.includes("Dev");
+        const featurePowerstripUI = features.includes("PowerstripUI");
+
+        // Generic helper functions
+        const sleepMsec = async (s: number) => {
+            return await new Promise((resolve) => setTimeout(resolve, s));
+        };
+
+        const debugLog = (message: string) => {
```
Koenkk: Could you remove this function and just use `logger.debug` always?

src/devices/shelly.ts
```diff
@@ -45,6 +48,313 @@ const shellyModernExtend = {
...
     },
+    shellyRPCSetup(features: string[] = []): ModernExtend {
+        // Set helper variables
+        const shellyRPCBugFixed = false; // For firmware 20250819-150402/ga0def2d
+
+        const featureDev = features.includes("Dev");
+        const featurePowerstripUI = features.includes("PowerstripUI");
+
+        // Generic helper functions
+        const sleepMsec = async (s: number) => {
```
Koenkk: Please use this sleep function: https://github.com/Koenkk/zigbee-herdsman-converters/blob/f3aa06c9ff5eb10fd43b6c12b1baf7351c0956a5/src/lib/utils.ts#L355

src/devices/shelly.ts
```diff
@@ -45,6 +48,313 @@ const shellyModernExtend = {
...
+                                        brightness: (value as number) ?? 0,
+                                    },
+                                },
+                            },
+                        },
+                    });
+                },
+            },
+            {
+                key: ["ledNightMode"],
```
```suggestion
                key: ["led_night_mode"],
```

also here.

src/devices/shelly.ts
DataGhost: I swear I looked for it and stumbled upon https://github.com/Koenkk/zigbee-herdsman-converters/blob/24973bee3fb213e8948f772a03ba97803ad5f710/src/devices/ubisys.ts#L114 so I assumed it wasn't in the library. Changed.

src/devices/shelly.ts
DataGhost: Done. That was set for removal once it was fully complete (after the new firmware) anyway but I basically only used it because I didn't set the config appropriately.

DataGhost:
I still have questions about https://github.com/DataGhost/zigbee-herdsman-converters/blob/849deebeac4d3feb8f715b5b61e8d7e3ccfb1e6b/src/devices/shelly.ts#L81-L86 and https://github.com/DataGhost/zigbee-herdsman-converters/blob/849deebeac4d3feb8f715b5b61e8d7e3ccfb1e6b/src/devices/shelly.ts#L113-L119 though. What should I do there? This should be resolved before it goes into the codebase IMO.

Koenkk:
- To fix this you need to provide typing, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/5dc64160b0a6cc20c975127f50614b268af20681/src/devices/niko.ts#L330
- Please also remove the use of all `any` types

DataGhost:
Done, and seems to be accepted by the checkers. That does not answer why the (to me) identical statements in https://github.com/DataGhost/zigbee-herdsman-converters/blob/849deebeac4d3feb8f715b5b61e8d7e3ccfb1e6b/src/devices/shelly.ts#L81-L86 weren't handled identically in terms of typing, but I'm sure there's a good answer to that somewhere out there.

```
// Okay
const txCtl = {
    txCtl: message.length,
};
await endpoint.write("shellyRPCCluster", txCtl, SHELLY_OPTIONS);

// Somehow also okay
const wrong = {
    txCtl: "wrong type",
    invalidProperty: 1337,
};
await endpoint.write("shellyRPCCluster", wrong, SHELLY_OPTIONS);

// Not okay
await endpoint.write("shellyRPCCluster", {txCtl: message.length}, SHELLY_OPTIONS);
```
results in:
```
src/devices/shelly.ts:103:59 - error TS2353: Object literal may only specify known properties, and 'txCtl' does not exist in type 'Partial<unknown> & RawClusterAttributes'.

103                 await endpoint.write("shellyRPCCluster", {txCtl: message.length}, SHELLY_OPTIONS);
                                                              ~~~~~

Found 1 error in src/devices/shelly.ts:103

 ELIFECYCLE  Command failed with exit code 2.
```
Seems like this shouldn't happen and all three should be rejected.

PR #10669, opened 2025-11-05
Fix: use correct max values for Aqara W100 temp_period and humi_period

MoeRT09:
This PR fixes the wrong boundaries defined or temp_period and humi_period for the Aqara W100 temperature and humidity sensors. The boundaries are taken from this article: https://homekitnews.com/2025/05/15/aqara-climate-monitor-w100-review/

I also tested the values via external converter:
Here are humidity measurements for different `humi_period` settings in Home Assistant:
image

Here are temperature measurements for different `temp_period` settings:
image


Koenkk:
Thanks!

PR #10680, opened 2025-11-07
Add action lookup for Aqara Z1 (double - triple rockers) switches

mirackle-spb:
Add action lookup for Aqara Z1 (double - triple rockers) switches


Koenkk:
Thanks!

doubleukay:
@mirackle-spb does your double/triple-rocker switch send these extra actions? Mine sends nothing except single click.

PR #10685, opened 2025-11-07
some adjustments in MTD285-ZB settings

Bjk8kds:
- change `debug_switch` to toggle
- separate `environmental_noise_collect` into 2 entitties, button and status
- descriptions adjustment

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -14771,8 +14771,8 @@ export const definitions: DefinitionWithExtend[] = [
                 .withUnit("m/s")
                 .withDescription("Target speed with direction"),
             e
-                .binary("debug_switch", ea.STATE_SET, "on", "off")
-                .withDescription("Debug mode toggle"), // change to toggle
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -14786,7 +14788,10 @@ export const definitions: DefinitionWithExtend[] = [
                 .enum("judge_logic", ea.STATE_SET, ["large_move", "small_move", "custom_move"]) // same as in the app
                 .withDescription("Presence detection algorithm"),
             e
-                .enum("environmental_noise_collect", ea.STATE_SET, ["start", "ongoing", "complete"])
+                .enum("start_noise_collect", ea.SET, ["start"]) // only button to start noise collection
```
Bjk8kds, line 14791: In the original setup, the action button and indicator/status are in one place, only the start button can be pressed, the other buttons are only for status, which I find less intuitive.
So I tried to find a way to separate them into 2, one for action button and the other only for status.

Koenkk:
Thanks!

PR #10689, opened 2025-11-08
Add Tuya ME202Z submersible liquid level sensor

MohammedAlhaid:
This PR adds support for the Tuya ME202Z submersible liquid level sensor (TS0601, _TZE284_mxujdmxo).

Features:
- Reports liquid height in cm
- Reports liquid level percentage
- Configurable tank depth (mm)
- Reports power supply voltage

Tested locally and working with Zigbee2MQTT.

### DIFF
--- a/src/devices/me202z.ts
+++ b/src/devices/me202z.ts
@@ -1,0 +1,106 @@
+import * as tuya from 'zigbee-herdsman-converters/lib/tuya';
+import * as exposes from 'zigbee-herdsman-converters/lib/exposes';
+const ea = exposes.access;
+
+// Data points
+const dp = {
+    distance: 2,          // liquid height (cm)
+    cfg_threshold: 21,    // tank depth (mm)
+    power_level_v: 5,     // power voltage (raw)
+};
+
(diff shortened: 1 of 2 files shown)


src/devices/me202z.ts
```diff
@@ -0,0 +1,106 @@
+import * as tuya from 'zigbee-herdsman-converters/lib/tuya';
```
Koenkk, line 1: Could you move this to `tuya.ts`?

src/devices/me202z.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (Array.isArray(data)) return data.reduce((acc, b) => (acc << 8) + b, 0);
+        return Number(data);
+    } catch { return Number(data); }
+}
+
+// Cache last known tank depth per device
+const lastKnown: Record<string, { depthMm?: number }> = {};
+
+// FromZigbee converter
+const fzLocal = {
```
Koenkk, line 28: Please use `tuyaDatapoints` instead of this converter, you can find examples in `tuya.ts`

PR #10693, opened 2025-11-08
fix(moes): correct time base for ZHT-SR thermostat and enable automat…

AudreynHeadburn:
…ic local time sync

This PR fixes the incorrect local time offset used by the Moes ZHT-SR (TS0601 / _TZE204_lpedvtvr) smart ring thermostat.

Previously, the device used `timeStart: "2000"`, causing the internal day-of-week calculation to drift and show the wrong weekday after pairing or reboot.

Changes:
- Updated `extend` line to use the correct epoch base: from: `tuya.modernExtend.tuyaBase({ dp: true, timeStart: "2000" })`
  to:   `tuya.modernExtend.tuyaBase({ dp: true, timeStart: "1970" })`
- Added `onEvent: tuya.onEventSetLocalTime` to automatically push the current local time when the device (re)joins.

This resolves the persistent "wrong weekday" issue (e.g., device showing Wednesday when it was actually Saturday).

All other exposes and datapoints remain unchanged.

### DIFF
--- a/src/devices/moes.ts
+++ b/src/devices/moes.ts
@@ -1402,8 +1402,6 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Moes",
         description: "Smart ring thermostat",
         extend: [tuya.modernExtend.tuyaBase({dp: true, timeStart: "1970"})],
-
-        onEvent: (tuya as any).onEventSetLocalTime,
         exposes: [
             e
                 .climate()


src/devices/moes.ts
```diff
@@ -1401,7 +1401,9 @@ export const definitions: DefinitionWithExtend[] = [
         model: "ZHT-SR",
         vendor: "Moes",
         description: "Smart ring thermostat",
-        extend: [tuya.modernExtend.tuyaBase({dp: true, timeStart: "2000"})],
+        extend: [tuya.modernExtend.tuyaBase({dp: true, timeStart: "1970"})],
+
+        onEvent: (tuya as any).onEventSetLocalTime,
```
Koenkk: This shouldn't be needed, can you test without?

AudreynHeadburn: Hey Koenkk, 
I’m a little confused by your comment. The suggested change fixes the incorrect date on the MOES Thermostat. Without this change it won’t work and always syncs the date with Wednesday. 

I thought I could help others with my small fix.

Nico

Koenkk:
Thanks!

PR #10694, opened 2025-11-08
feat(add): Somfy Ysia 1

burmistrzak:
Pretty much the same as the Ysia 5, but with only a single group _channel_.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4401

### DIFF
--- a/src/devices/bega.ts
+++ b/src/devices/bega.ts
@@ -23,4 +23,11 @@ export const definitions: DefinitionWithExtend[] = [
         description: "LED lamp with adjustable LED color temperature (Tunable White - RGBW) for use in luminaires with E27 lamp base",
         extend: [m.light({colorTemp: {range: [153, 556]}, color: true})],
     },
+    {
+        zigbeeModel: ["BEGA 85000 Garden Spotlight"],
+        model: "85000",
+        vendor: "Bega",
+        description: "Wired Garden LED Spotlight color temperature (Tunable White - RGBW)",
+        extend: [m.light({colorTemp: {range: [50, 1000]}, color: {modes: ["xy", "hs"], enhancedHue: true}})],
+    },
 ];
--- a/src/devices/easyiot.ts
+++ b/src/devices/easyiot.ts
@@ -181,9 +181,7 @@ export const definitions: DefinitionWithExtend[] = [
         fingerprint: [{modelID: "ZB-IR01", manufacturerName: "easyiot"}],
         model: "ZB-IR01",
         vendor: "easyiot",
-        description:
-            "This is an infrared remote control equipped with a local code library," +
-            "supporting devices such as air conditioners, televisions, projectors, and more.",
+        description: "Infrared remote control equipped with local code library,",
         fromZigbee: [fzLocal.easyiot_ir_recv_command],
         toZigbee: [tzLocal.easyiot_ir_send_command],
         exposes: [
@@ -195,7 +193,7 @@ export const definitions: DefinitionWithExtend[] = [
         fingerprint: [{modelID: "ZB-TTS01", manufacturerName: "easyiot"}],
         model: "ZB-TTS01",
         vendor: "easyiot",
-        description: "This is a Simplified Chinese (GB2312) TTS converter that can convert GB2312 encoded text to speech",
+        description: "TTS Converter for Simplified Chinese GB2312 encoded text",
         fromZigbee: [fzLocal.easyiot_tts_recv_status],
         toZigbee: [tzLocal.easyiot_tts_send_command],
         exposes: [
--- a/src/devices/ecodim.ts
+++ b/src/devices/ecodim.ts
@@ -8,7 +8,7 @@ const e = exposes.presets;
 
 export const definitions: DefinitionWithExtend[] = [
     {
-        zigbeeModel: ["Eco-Dim.05 Zigbee"],
+        zigbeeModel: ["Eco-Dim.05 Zigbee", "eco-dim05-zigbee"],
         fingerprint: [
             {
                 type: "Router",
--- a/src/devices/innr.ts
+++ b/src/devices/innr.ts
@@ -237,6 +237,13 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [m.light({turnsOffAtBrightness1: true})],
     },
     {
+        zigbeeModel: ["BB 262"],
+        model: "BB 262",
+        vendor: "Innr",
+        description: "E27 bulb",
+        extend: [m.light({turnsOffAtBrightness1: true})],
+    },
+    {
         zigbeeModel: ["RB 265"],
         model: "RB 265",
         vendor: "Innr",
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -1866,6 +1866,13 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [philips.m.light({colorTemp: {range: [153, 454]}})],
     },
     {
+        zigbeeModel: ["929003597802"],
+        model: "929003597802",
+        vendor: "Philips",
+        description: "Hue black ambiance Aurelle square panel light",
+        extend: [philips.m.light({colorTemp: {range: [153, 454]}})],
+    },
+    {
         zigbeeModel: ["3216231P6"],
         model: "3216231P6",
         vendor: "Philips",
@@ -3260,6 +3267,13 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Philips",
         description: "Hue Play gradient lightstrip for PC 3x 24/27",
         extend: [philips.m.light({colorTemp: {range: [153, 500]}, color: true, gradient: true})],
+    },
+    {
+        zigbeeModel: ["929004608004"],
+        model: "929004608004",
+        vendor: "Philips",
+        description: "Hue Omniglow lightstrip",
+        extend: [philips.m.light({colorTemp: {range: [153, 500]}, color: true, gradient: {extraEffects: ["sparkle", "opal", "glisten"]}})],
     },
     {
         zigbeeModel: ["929003099301"],
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -352,6 +352,21 @@ const convLocal = {
             return new TextDecoder("utf-8").decode(hexToBytes);
         },
     },
+    nameTrunc: {
+        to: (v: string, meta: Tz.Meta, len = 8) => {
+            const truncated = v.slice(0, len);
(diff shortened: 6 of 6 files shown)


src/devices/somfy.ts
```diff
@@ -62,6 +62,13 @@ export const definitions: DefinitionWithExtend[] = [
             m.commandsWindowCovering({endpointNames: ["1", "2", "3", "4", "5"]}),
         ],
     },
+    {
+        zigbeeModel: ["Ysia 1 Zigbee Europe"],
+        model: "1871157",
+        vendor: "SOMFY",
+        description: "Ysia 1 channel blinds remote",
+        extend: [m.deviceEndpoints({endpoints: {"1": 1, "232": 232}}), m.battery(), m.commandsOnOff(), m.commandsWindowCovering()],
```
Koenkk:
```suggestion
        extend: [m.battery(), m.commandsOnOff(), m.commandsWindowCovering()],
```

Can you try with this?

burmistrzak: Works on my end. 👌

Koenkk:
Thanks!

iay:
@burmistrzak I've just acquired one of these so it was really nice to see that this was supported already. Unfortunately it's not working for me and I wondered if you might comment on what steps you went through to onboard the device.

I have tried holding down the reset button until the second red flash; the device then joins my network. The interview and configuration claims to have worked but pressing any button just causes the light on the device to flash three times; I don't see anything at the Z2M end which suggests that the action is being triggered. Nothing in the logs.

I realise this isn't the best way to get support, but your profile doesn't include an e-mail address...

Any assistance appreciated.

(I have a Ysia 5 Europe, with the same behaviour)

burmistrzak:
@iay Happy to take a look. What's the firmware version of your remote?
Also, do you intend to use it via MQTT or direct Zigbee binding to a shutter?

iay:
> @iay Happy to take a look. What's the firmware version of your remote? Also, do you intend to use it via MQTT or direct Zigbee binding to a shutter?

Firmware reported in Zigbee2MQTT as 000161.

I was planning on using the remote as a general purpose one via Zigbee2MQTT > MQTT > Home Assistant. I have my blinds hooked up to HA and am controlling them directly from HA at present.

If I can't get events out of the remote I'd settle for controlling the (single) set of blinds in one room using the controller as long as that didn't lock HA out as I have some automations set up. I do remember trying to bind the 5-way remote I have to the set of three blinds I have in the other room, though, with no success. Of course I have no idea what I am doing here so it might have been user error :-)

burmistrzak:
@iay Ok, same version as mine.
However, I'm binding my remotes **directly** to Zigbee shutters, so Z2M is only used to get them onto the network.
This also _seems to be_ the[ intended use-case according to Somfy](https://service.somfy.com/downloads/au_v5/ysiastandalone_instructions.pdf).

~~Usage with a coordinator/gateway _likely_ requires additional reverse engineering.
A quick network analysis shows some manufacturer-specific clusters are used, so we'll have to find someone with a **Somfy TaHoma Switch** gateway to provide a full package capture.~~
See my [edit](https://github.com/Koenkk/zigbee-herdsman-converters/pull/10694#issuecomment-3651812670)

iay:
> @iay Ok, same version as mine. However, I'm binding my remotes **directly** to Zigbee shutters, so Z2M is only used to get them onto the network. This also _seems to be_ the[ intended use-case according to Somfy](https://service.somfy.com/downloads/au_v5/ysiastandalone_instructions.pdf).

I'm aware that Somfy are looking at Zigbee as a replacement for their previous proprietary system (some kind of sub-GHz system, I believe) but I wanted Zigbee so I didn't have to have yet another hub. So, um, I don't have one of their gateways.

> Usage with a coordinator/gateway _likely_ requires additional reverse engineering. A quick network analysis shows some manufacturer-specific clusters are used, so we'll have to find someone with a **Somfy TaHoma Switch** gateway to provide a full package capture.

At present I just set the cover position via HA and voice commands plus some other automations, which is fine. It would have been nice to use the remote the vendor insisted on including for something more than just raising and lowering, but if that's not possible right now that's also fine. Minimal functionality would be better than nothing.

You say you bind your remotes directly to the covers; can you walk me through that? I seem to recall _trying_ to bind an endpoint from the 5-way control I have to one of the covers in the other room, and it didn't seem to work (and in a way which didn't really help me understand what the problem was). Did you do the binding using Z2M or is it some kind of manual process involving getting up on a ladder and pushing the magic button on the cover again?

burmistrzak:
> I'm aware that Somfy are looking at Zigbee as a replacement for their previous proprietary system (some kind of sub-GHz system, I believe) but I wanted Zigbee so I didn't have to have yet another hub. So, um, I don't have one of their gateways.

@iay Yea, their _io-homecontrol_ protocol sucks... Not a fan.

> You say you bind your remotes directly to the covers; can you walk me through that? I seem to recall _trying_ to bind an endpoint from the 5-way control I have to one of the covers in the other room, and it didn't seem to work (and in a way which didn't really help me understand what the problem was). Did you do the binding using Z2M or is it some kind of manual process involving getting up on a ladder and pushing the magic button on the cover again?

Sure thing! Here's what I did (my shutter controller is Bosch):

1. Pair the remote(s) to Z2M. Ensure battery levels are reported.
2. In the Z2M frontend, go to [YOUR REMOTE] > Bind
3. Hit **+ Add** next to **Source endpoint 1**.
4. Add a new binding with your Zigbee cover as the **destination**. Choose the appropriate **endpoint** (e.g. `1`) that hosts the `closuresWindowCovering` cluster. Check _at least_ the `closuresWindowCovering` from the provided list.
5. **Do not** click **Bind** yet!
6. Wake up your remote. (See my edit below!)
7. Immediately hit the **Bind** button.
8. Binding should now be in place.
9. (Wake your remote once more and) unbind the **coordinator** from **Source endpoint 1** by hitting the **Unbind** button for that entry.
10. Your direct binding should be working now.

**Edit:** Before modifying bindings, you'll have to short press the pairing button on the back of the remote. The LED should blink amber. If you want to use the remote with HA, unbind **only** the `closuresWindowCovering` **from the coordinator** and leave `genOnOff` in place.

iay:
I tried this just now and it seemed to work. It took a couple of goes before the binding to the coordinator was removed, but I'm now in a state where the buttons on the remote do up/stop/down through a binding from endpoint 1 to the cover using just the `closuresWindowCovering` cluster.

Endpoint 232 is still bound (`genPollCtrl` and `genPowerCfg`) to the coordinator, which I assume will get me the battery report.

I didn't have to press the pairing button on the back of the remote... was that part of your edit?

If I rebind the `genOnOff` cluster to the coordinator, what would you expect Home Assistant to see when buttons are pressed?

Thanks very much for the help. I don't think I'd have had much chance of getting this to work without assistance. I'm not even clear on what part of the system manages the bindings; is it the source device or is the coordinator relaying things?

burmistrzak:
> Endpoint 232 is still bound (`genPollCtrl` and `genPowerCfg`) to the coordinator, which I assume will get me the battery report.

@iay Correct. 😊

> I didn't have to press the pairing button on the back of the remote... was that part of your edit?

Yep. Just easier because that button wakes up the remote for a bit longer.

> If I rebind the `genOnOff` cluster to the coordinator, what would you expect Home Assistant to see when buttons are pressed?

Yes, you'll likely see `on` and `off` only. No `stop`.

> Thanks very much for the help. I don't think I'd have had much chance of getting this to work without assistance.

Happy to help!

> I'm not even clear on what part of the system manages the bindings; is it the source device or is the coordinator relaying things?

Well, in this case the coordinator simply configures the binding. Everything else is between the two (or more!) devices.
It _is possible_ for devices to bind directly without having a coordinator around, but this often requires both devices to be in _very close_ proximity.

iay:
> > Endpoint 232 is still bound (`genPollCtrl` and `genPowerCfg`) to the coordinator, which I assume will get me the battery report.
> 
> @iay Correct. 😊

Confirmed.

> > If I rebind the `genOnOff` cluster to the coordinator, what would you expect Home Assistant to see when buttons are pressed?
> 
> Yes, you'll likely see `on` and `off` only. No `stop`.

Also confirmed. The UP button sends ON and the DOWN button sends OFF. The MY button doesn't send anything. I guess that makes sense in  terms of the genOnOff spec, but it does kind of make these controllers useless for anything other than what they were designed for. Not a complaint, just an observation.

One final comment, relevant to the original commit this discussion is tacked on to.

The thing that seems to have been causing my issues is that the default binding is for endpoint 1's `closuresWindowCovering` and `genOnOff` clusters to be bound to the coordinator. That causes nothing to arrive at the coordinator, presumably because it tries something with `closuresWindowCovering` first and that fails. If you _only_ bind `genOnOff` to the coordinator, as you suggested that does result in what I suppose is reasonable behaviour and no red flashing.

I'm not sure whether this means that the `m.commandsWindowCovering()` in the converter should be there after all, if the combination of coordinator (I have a SLZB-06, FWIW) and Zigbee2MQTT don't in fact support the messages it sends to that cluster.

burmistrzak:
@iay Yes, the _my_ button is only useful when used with a cover directly. 😊

There's an entire song and dance to set a favorite position, but it works surprisingly well!
Under the hood, the remote is configuring an internal Zigbee scene automatically. Pretty cool.

But you are right, it might be a good idea to just bind the `genOnOff` cluster by default.

iay:
> There's an entire song and dance to set a favorite position, but it works surprisingly well! Under the hood, the remote is configuring an internal Zigbee scene automatically. Pretty cool.

I saw that in the manual but it didn't work for me. ~~I am now thinking that if it's scene-based it might need the scene cluster to be bound as well as the cover controlling cluster. Can I ask: what do you have bound, exactly?~~ OK, that's hilarious, binding the scenes cluster as well does in fact enable this functionality for me. I can even see the added scene in Zigbee2MQTT, which is slightly surreal.

burmistrzak:
@iay Yea, forgot to mention that you'll have to bind the scenes cluster for the _my_ button to work. Sorry!

> I can even see the added scene in Zigbee2MQTT, which is slightly surreal.

@Nerivec AFAICT, it _should_ be possible to fetch scenes from a device that are not created via Z2M, right?
Would be super neat to see internal scenes (e.g. created by other devices via direct bindings) in the frontend as well. 😊

Nerivec:
Yes, you can fetch any scene from the device with attrs/cmds.
Expanded scene features (like device<>z2m sync) are part of https://github.com/Koenkk/zigbee-herdsman/pull/1570 (still working on it though, probably for some time in January :grin:)

PR #10701, opened 2025-11-09
fix: Explicitly enable reporting for Bosch thermostats

burmistrzak:
While some attributes are reported by Bosch thermostats on their own, it's seemingly not as reliable as configured reports.
There's also the issue of firmware updates that alter the hard-coded reporting behavior in an undesired way.

This PR brings back reporting for essential attributes, without having to rely on manufacturer's defaults.


Koenkk:
@DerDreschner could you review this?

burmistrzak:
@Koenkk FWIW, this PR simply restores just the previous reporting configuration. Without, we're relying on the hard-coded manufacturer's defaults that can change (via OTA) at anytime without notice.
Having reporting explicitly configured also helps with visibility because hard-coded firmware defaults are obviously not visible in Z2M.

Koenkk:
I see, thanks!

PR #10713, opened 2025-11-10
Add support for Philips Hue Downlight LTD022 zigbee model

preetpatel:
In addition to the philips hue zigbee model LTD021, there is also a variant LTD022 thats reported in zigbee but is the exact same device


Koenkk:
Thanks!

PR #10717, opened 2025-11-11
fix: Move incorrect control entities to configuration section for Sonoff SNZB-02D

blacknell:
I've noticed that several of the entities in the Sonoff SNZB-02D are not marked as type `config` for the `entityCategory`. For example, temperature units, comfort temperature high/low etc. There are 7 in all.

This results in a lot of extra entities displaying in dashboards when they shouldn't since they appear as control entities. Moving them to the configuration section is where they should be rather than as every day controls.


chris-1243:
Could you also check the other SNZB-02x and add your suggestion if needed ?

Koenkk:
Thanks!

PR #10709, opened 2025-11-09
fix: Bosch BTH-RM: Expose all `system_mode` for compatibility

burmistrzak:
The recent re-implementation #10309 of Bosch thermostats has introduced compatibility issues with ecosystems that do not use or cannot make use of HA discovery payloads (e.g. `homebridge-z2m`), and thus entirely depend on well-known/standard attributes (e.g. `system_mode`) to work properly.

This PR aims to rectify this by (once again) exposing all available `system_mode` states to non-HA users.

The following changes from #10309 will be reverted:

> - **Changed:** Only expose settings for current system mode on BTH-RM / BTH-RM230Z.

A future PR (_no ETA_) is going to add a new Z2M-wide device option for all thermostats to filter out system modes. This approach will be compatible across ecosystems.

PR #10309, opened 2025-10-12
Bosch Thermostats: Overwork implementation for BTH-RA / BTH-RM / BTH-RM230Z

DerDreschner:
This is a follow-up to the reverted PR #10110. All changes from there are included here as well.

Additional changes to the last pull request:
- **Change**: Extend `modernExtend.thermostat` to make some parts disableable. This fixes the failed integration test that lead to the last revert of this PR. (@Koenkk : I only extended attributes I use myself, not every attribute of the thermostat. I hope that decision as well as the way of extending it is okay. If I should include it everywhere for consistency reasons or if you have any other comments, just let me know.)
- **New**: Implement automatic valve calibration when requested by BTH-RA, for example after a firmware update.
- **New**: Expose relay state on BTH-RM230Z.
- **New**: Expose setting for activity LED on BTH-RM/BTM-RM230Z (thanks to https://github.com/Sopsy for documenting it!)
- **New/Fixed**: Include `heaterType` previously known as `boiler (0x01)` on BTH-RM230Z (see discussion under the last PR). It's now being called `central_heating`, as Bosch [renamed the setting itself while retaining the old behavior](https://community.bosch-smarthome.com/t5/feedback-und-w%C3%BCnsche/raumthermostat-230-v-ii-f%C3%BCr-zentralheizungseinsatz-ert%C3%BCchtigen/m-p/89537/highlight/true#M6711). The previous `central_heating` is now called `manual_control` as it makes the relay totally independent from the thermostat settings. This allows for automatons like e.g., if you have multiple rooms with temperature sensors and like to activate the boiler if one room calls for heat.
- **Changed**: Only expose settings for current system mode on BTH-RM / BTH-RM230Z.
- **Changed**: Changed the reported attributes to match those configured with the native Bosch coordinator. Only the BTH-RA seem to need them, the BTH-RM / BTH-RM230Z report a lot of attributes without any configuration for some reason.
- **Cleanup**: Change type of `reporting` attribute in `BinaryArgs` from `ReportingConfig<Cl, Custom>` to `ReportingConfigWithoutAttribute` as the set attribute isn't being used at all.
- **Removed**: The code for `genTime` cluster read requests is being removed as this will be handled by `zigbee-herdsman` in the future (see: https://github.com/Koenkk/zigbee-herdsman/pull/1517 ).

This fixes the following issues:
- https://github.com/Koenkk/zigbee2mqtt/issues/20057
- https://github.com/Koenkk/zigbee2mqtt/discussions/21009
[...]

PR #10722, opened 2025-11-12
fix: Bosch BTH-R*: Rename key to `window_detection`

burmistrzak:
In #10309, the property key for the window detection switch was renamed from `window_detection` to `window_open_mode` which [isn't recognized](https://github.com/Koenkk/zigbee2mqtt/blob/5797b56871215b8fac37cfeb4f702f73a09c4ae0/lib/extension/homeassistant.ts#L50) by the HA extension. This PR fixes the issue.

Depends on #10709


Koenkk:
To prevent another breaking change, I would suggest to add `window_open_mode` to https://github.com/Koenkk/zigbee2mqtt/blob/5797b56871215b8fac37cfeb4f702f73a09c4ae0/lib/extension/homeassistant.ts#L50

burmistrzak:
@Koenkk Sure, but a bit more has then to be changed as well.
https://github.com/Koenkk/zigbee2mqtt/blob/5797b56871215b8fac37cfeb4f702f73a09c4ae0/lib/extension/homeassistant.ts#L581

I can certainly do that if you prefer. However, reverting a key's name seemed like the _less invasive_ option. 😊
Just let me know.

Koenkk:
hmm, then lets do the rename, could you update this pr to do just that?

burmistrzak:
@Koenkk Alright, should be good to go. 😊

Koenkk:
thanks!

PR #10723, opened 2025-11-12
Add a new temperature-controlled valve with a screen

SANSTONE:


src/devices/tuya.ts
```diff
@@ -20829,296 +20829,98 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+
     {
-        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_4cqhd2ha", "_TZE200_8ply8mjj"]),
```
Koenkk, line 20833: There are a lot of unexpected changes here, e.g. with this PR `_TZE200_8ply8mjj` becomes unsupported, could you clean it up?

PR #10724, opened 2025-11-12
Add support for motor direction reversal for `lumi.curtain.acn002`.

deiger:
Verified to be working in dev console.


Koenkk:
Thanks!

PR #10726, opened 2025-11-12
Fix the detection range label representation

kpustelnik:
Small fix for the detection range composite label representations. Sorry for having it wrong initially...


Koenkk:
thanks!

PR #10729, opened 2025-11-13
feat(TS0601): add high-precision AC frequency (DP135)

jacky202509:
This PR adds support for high-precision AC frequency (DP135) on TS0601 3-phase clamp power meters.

- Exposes `ac_frequency_high_precision` in Hz for devices with `applicationVersion >= 132`.
- Updates `tuyaDatapoints` and `exposes` accordingly.
- Maintains backward compatibility for older devices without DP135.


Koenkk:
thanks!

PR #10730, opened 2025-11-13
fix: move weekly schedule to config section for Sonoff TRVZB

blacknell:
I've noticed that the weekly schedule for the Sonoff TRVZB is not marked as type `config` for the `entityCategory`.

This results in extra entities displaying in dashboards when they shouldn't since they aren't control entities. Moving them to the configuration section is where they should be rather than as every day controls.

Note I've amended the modern extend `weeklySchedule` which is currently only referenced by the Sonoff TRV but this should make sense as the default for any future variants.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -332,6 +332,7 @@ const sonoffExtend = {
     weeklySchedule: (): ModernExtend => {
         const exposes = e
             .composite("schedule", "weekly_schedule", ea.STATE_SET)
+            .withCategory("config")
             .withDescription(
                 'The preset heating schedule to use when the system mode is set to "auto" (indicated with ⏲ on the TRV). ' +
                     "Up to 6 transitions can be defined per day, where a transition is expressed in the format 'HH:mm/temperature', each " +


Koenkk:
thanks

blacknell:
@Koenkk - I've just installed and I don't understand why it's now showing up in `diagnostics` rather than `configuration`. The same change I made on the SNZB-02 in [PR 10717](https://github.com/Koenkk/zigbee-herdsman-converters/pull/10717) has worked.

Can you advise and I'll submit a new PR

Koenkk:
Could you provide the debug log when z2m starts such that we can see what is published?

blacknell:
> Could you provide the debug log when z2m starts such that we can see what is published?

Here are the lines where I grep'd for one of my devices and `schedule`

[sch.log](https://github.com/user-attachments/files/23898537/sch.log)

Koenkk:
Seems to be because of this: https://github.com/Koenkk/zigbee2mqtt/blob/110149fa20c743bbba540207778b05c5eb575cf8/lib/extension/homeassistant.ts#L1087, so HA doesn't support entity category `config` for sensors.

blacknell:
> Seems to be because of this: https://github.com/Koenkk/zigbee2mqtt/blob/110149fa20c743bbba540207778b05c5eb575cf8/lib/extension/homeassistant.ts#L1087, so HA doesn't support entity category `config` for sensors.

So the schedule shouldn't be a sensor. I don't understand the weeklyschedule method

PR #10733, opened 2025-11-13
Added the mute function to the MIR-SM200 smoke detector.

yuanxin32323:
Added the mute function to the MIR-SM200 smoke detector.

### DIFF
--- a/src/devices/multir.ts
+++ b/src/devices/multir.ts
@@ -79,10 +79,7 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "MultIR",
         description: "Smoke sensor",
         toZigbee: [tzLocal.MIRSM200],
-        extend: [
-            m.battery(),
-            m.iasZoneAlarm({zoneType: "smoke", zoneAttributes: ["alarm_1", "tamper", "battery_low"]}),
-        ],
+        extend: [m.battery(), m.iasZoneAlarm({zoneType: "smoke", zoneAttributes: ["alarm_1", "tamper", "battery_low"]})],
         exposes: [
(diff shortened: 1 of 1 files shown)


src/devices/multir.ts
```diff
@@ -21,6 +22,22 @@ const fzLocal = {
     } satisfies Fz.Converter<"ssIasZone", undefined, "raw">,
 };
 
+const tzLocal = {
+    MIRSM200: {
+        key: ["silence"],
+        convertSet: async (entity, key, value, meta) => {
+            if (value === "enable") {
```
Koenkk: can it also be used to turn off? otherwise we should use an enum as expose with just the enable value

yuanxin32323: > can it also be used to turn off? otherwise we should use an enum as expose with just the enable value

Yes, it can be used to turn off the alarm. I have consulted the manufacturer's firmware developer, sent the test JS file to them, and the test was successful.

I will put forward suggestions to the manufacturer later. Perhaps in a new version, the mute command will be replaced.

Koenkk: But what happens if `disabled` is send? Does the alarm turn on then? I'm asking because I don't see a `entity.command` being executed in that case. If it can only silence the alarm, change `exposes.binary("silence"` to `exposes.enum("silence", ea.SET, ["ON"]).withDescription("Silence alarm"),`

yuanxin32323: > But what happens if `disabled` is send? Does the alarm turn on then? I'm asking because I don't see a `entity.command` being executed in that case. If it can only silence the alarm, change `exposes.binary("silence"` to `exposes.enum("silence", ea.SET, ["ON"]).withDescription("Silence alarm"),`

ok , i do it

Koenkk:
Thakns!

PR #10737, opened 2025-11-14
Fixed a data type error

smartboxchannel:
Fixed a data type error for the high_humidity and low_humidity attributes


Koenkk:
thanks

PR #10749, opened 2025-11-15
Configure level reporting for Candeo C202.1

kangaroomadman:
Specify `levelReportingConfig` for Candeo C202.1 device. This matches the configuration from other Candeo devices. e.g. C201/C202/C203/C204

### DIFF
--- a/src/devices/candeo.ts
+++ b/src/devices/candeo.ts
@@ -268,7 +268,7 @@ export const definitions: DefinitionWithExtend[] = [
             m.light({
                 configureReporting: true,
                 levelReportingConfig: {min: 1, max: 3600, change: 1},
-                powerOnBehavior: false
+                powerOnBehavior: false,
             }),
         ],
     },


src/devices/candeo.ts
```diff
@@ -264,7 +264,13 @@ export const definitions: DefinitionWithExtend[] = [
         model: "C202.1",
         vendor: "Candeo",
         description: "Zigbee LED smart dimmer switch",
-        extend: [m.light({configureReporting: true, powerOnBehavior: false})],
+        extend: [
+            m.light({
+                configureReporting: true,
+                levelReportingConfig: {min: 1, max: 3600, change: 1},
```
Koenkk, line 270:
```suggestion
                levelReportingConfig: true,
```

can you try with this?

kangaroomadman: Shouldn't `levelReportingConfig` be an object that overwrites the defaults using the spread operator? https://github.com/Koenkk/zigbee-herdsman-converters/blob/22f4748638b7f247661236fe1f1b1f5639b47498/src/lib/modernExtend.ts#L1284

The reporting is already configured for this device, but the default minimum of 5 seconds is quite high so the main purpose of this change is to bring this down to 1 second.

Koenkk: I would propose to change the default instead

kangaroomadman: Sure, I can do that. Just to check, that's change the `light` default in `modernExtend`? Should I also drop any overrides that are currently set to that?

Koenkk: Yes and yes

PR #10751, opened 2025-11-15
Identify C-ZB-LC20v2-RGBCCT as GL-C-008P

antaflos:
After an OTA firmware update two of my Gledopto GL-C-008P now identify as C-ZB-LC20v2-RGBCCT


Koenkk:
thanks

PR #10757, opened 2025-11-16
Update slacky_diy.ts

slacky1965:
added

TS0001_power-SlD - mini zigbee switch
TS011F_plug-SlD - zigbee Plug
QS-Zigbee-SEC01-Mod - zigbee switch 1 gang


Koenkk:
thanks

PR #10760, opened 2025-11-17
Update tuya.ts

LordRoge:
The model number of this device was incorrect and has been modified accordingly. Additionally, I have a question: I currently have devices with model numbers TOWSMRI-40A-A and TOWSMRI-20A-AC on hand. Their PIDs differ, and there are slight variations in their data points. How should I add or modify them?

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -15603,7 +15603,9 @@ export const definitions: DefinitionWithExtend[] = [
                 respondToMcuVersionResponse: true,
             }),
         ],
-        whiteLabel: [tuya.whitelabel("Tongou", "TOWSMR1-40A-AC", "Single-phase multifunction RCBO (DIN Module)", ["_TZE284_432zhuwe", "_TZE204_432zhuwe"])],
+        whiteLabel: [
+            tuya.whitelabel("Tongou", "TOWSMR1-40A-AC", "Single-phase multifunction RCBO (DIN Module)", ["_TZE284_432zhuwe", "_TZE204_432zhuwe"]),
+        ],
         exposes: [
             tuya.exposes.switch(),
             e.temperature(),


Koenkk:
> Their PIDs differ, and there are slight variations in their data points. 

Could you explain what differs?

LordRoge:
The manufacturerName for TOWSMRI-40A-A is '_TZE284_s5vuaadg', while for TOWSMRI-20A-AC it is '_TZE284_tuhfx7tf'. Additionally, TOWSMRI-40A-A does not have the DP points [117, "leakage_threshold", tuya.valueConverter.raw] and [15, "leakage_current", tuya.valueConverter.raw]. As for TOWSMRI-20A-AC, the only modification required is changing the .withValueMax(40) to .withValueMax(20) in the over_current_setting. Could you please advise on how to proceed with these adjustments?

Koenkk:
- The datapoints can be added (since it's a superset)
- Then what you have to do is to make the expose conditional (only add features which that specific device supports), example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/c7d126f5cef06728da93a9d8a0c6dba4fe099aad/src/devices/tuya.ts#L1202

LordRoge:
Is it modified as follows?
 const exps: Expose[] = [
            tuya.exposes.switch(),
            e.temperature(),
            e.current(),
            e.power(),
            e.voltage()...
];
if (
                !device ||
                device.manufacturerName === "_TZE284_432zhuwe" ||
                device.manufacturerName === "_TZE204_432zhuwe"
            ){
                exps.push(
                e.numeric("leakage_current", ea....);
            }
 else if (
                device.manufacturerName === "_TZE284_s5vuaadg"
            ){....}

Koenkk:
Yes that looks good, please update the PR

PR #10770, opened 2025-11-18
Profalux MAI-ZTM20C: added button actions and battery voltage reporting.

KipK:
Added button actions and battery voltage reporting on Profalux MAI-TMZ20C remote
Reporting works ok on endpoint 2.


src/devices/profalux.ts
```diff
@@ -162,13 +162,25 @@ export const definitions: DefinitionWithExtend[] = [
             },
         ],
         model: "MAI-ZTM20C",
+        zigbeeModel: ["MAI-ZTM20C"],
         vendor: "Profalux",
         description: "Cover remote",
-        extend: [
```
Koenkk, line 167: Could you keep this and only add `fromZigbee: [fz.command_cover_close, fz.command_cover_open, fz.command_cover_stop],`  and ` exposes: [e.action(["up", "down", "stop"])],`

I think:

```
      endpoint: (device) => {
            return {default: 2, 1: 1, 2: 2};
        },
```

can also be removed.

PR #10779, opened 2025-11-19
Add Wireless Round Thermostat in shinasystems.ts

shinasys:
Link to picture pull request: Koenkk/zigbee2mqtt.io/pull/4442

### DIFF
--- a/src/devices/shinasystem.ts
+++ b/src/devices/shinasystem.ts
@@ -294,7 +294,7 @@ const tzLocal = {
             return {state: {[key]: value}};
         },
         convertGet: async (entity, key, meta) => {
-            await entity.read("hvacThermostat", "systemMode");
+            await entity.read("hvacThermostat", ["systemMode"]);
         },
     } satisfies Tz.Converter,
     hqm_local_temperature: {


src/devices/shinasystem.ts
```diff
@@ -1197,4 +1197,69 @@ export const definitions: DefinitionWithExtend[] = [
             }),
         ],
     },
+    {
+        zigbeeModel: ["HQM-300Z"],
+        model: "HQM-300Z",
+        vendor: "ShinaSystem",
+        ota: true,
+        description: "SiHAS Zigbee Wireless Round Thermostat",
+        extend: [
```
Koenkk, line 1263: Does it make sense to expose this device as a thermostat? https://github.com/Koenkk/zigbee-herdsman-converters/blob/db1d6f2ee92b1c05ea24b01745819e785e8c3095/src/lib/modernExtend.ts#L3173

shinasys: I wanted to use a device with modernExtend as a thermostat, but I couldn't because of constraints on the device.

The constraints are as follows:

- The device uses the Temperature Measurement (0x0402) cluster instead of having the Local Temperature (0x0000) attribute in the Thermostat (0x0201) cluster.

- The device uses vendor-specific modes ("away" or "schedule") in addition to the standard system modes.

Since it seemed difficult to use in such a case, I created the code as follows.

Koenkk: Could we still use the `e.climate` expose? (and use `preset` for the vendor specific modes)

shinasys: Yes, Thanks!

Koenkk:
Thanks!

PR #10783, opened 2025-11-19
fix ZigbeeTLc measurement interval max value (#28604)

nchieffo:
ref: https://github.com/Koenkk/zigbee2mqtt/issues/28604

### DIFF
--- a/src/devices/zigbeetlc.ts
+++ b/src/devices/zigbeetlc.ts
@@ -98,7 +98,7 @@ const extend = {
         description: "Configure sensor measurement interval (default 10 seconds).",
         unit: "s",
         valueMin: 3,
-        valueMax: 30,
+        valueMax: 255,
     }),
 };
 


Koenkk:
Thanks!

mrmaximas:
@nchieffo @Koenkk changing the maximum of measurement interval interval to 255 seconds was not the best idea; longer interval - the longer you need to wait when you change any settings. Increasing this interval will increase the time between `Data Request` packets from the parent device. A maximum of 30 seconds was sufficient. You will not save battery life significantly by increasing the measurement interval to such values. This may cause inconvenience when changing device settings.
Screenshot 2025-11-22 at 15 33 28

z-master42:
If I understand correctly, this only refers to the range and not a fixed value. The default value is 10, isn't it?

mrmaximas:
@z-master42 yes, default is 10.

PR #10789, opened 2025-11-20
Update tuya.ts

LordRoge:
Update the feature points of various models of towsmr1,The images do not need to be updated. The three device models are visually identical, with the only differences at the marked locations.


Koenkk:
Thanks!

Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `TOWSMR1-20A-AC.png`

PR #10797, opened 2025-11-20
Update awox.ts

ringaction:
added EGLO RGB bulb
https://www.eglo.com/de/leuchtmittel-110273.html

Link to picture pull request: see attachment
eglo_zm_rgb_w

### DIFF
--- a/src/devices/awox.ts
+++ b/src/devices/awox.ts
@@ -307,10 +307,6 @@ export const definitions: DefinitionWithExtend[] = [
         model: "EGLO_ZM_RGB_W",
         vendor: "AwoX",
         description: "E27 LED Bulb RGB with withe ",
-        extend: [
-            m.deviceEndpoints({endpoints: {"1": 1, "3": 3}}),
-            m.light({colorTemp: {range: [333, 333]}, color: {modes: ["xy", "hs"]}}),
-            m.commandsOnOff(),
-        ],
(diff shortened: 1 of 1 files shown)


src/devices/awox.ts
```diff
@@ -302,4 +302,11 @@ export const definitions: DefinitionWithExtend[] = [
         meta: {multiEndpoint: true},
         whiteLabel: [{vendor: "EGLO", model: "900116"}],
     },
+    {
+        zigbeeModel: ["EGLO_ZM_RGB_W"],
+        model: "EGLO_ZM_RGB_W",
+        vendor: "AwoX",
+        description: "E27 LED Bulb RGB with withe ",
+        extend: [m.light({colorTemp: {range: [333, 333]}, color: {modes: ["xy", "hs"]}})],
```
Koenkk, line 310: Does the device support color temp? Then add a proper range (if not remove it)

ringaction: No, beside to RGB colours this device is able to show warm white only. But warm white is a separate LED and this is the reason why it appears beside to the colour mode.  
see attached screenshots of the AWOX for this kind of LED.
06101664-79D0-41A0-B387-0B823850FCCA_1_102_o
26CB7BC4-201F-4CD5-8DEE-3A2575A1B363_1_102_o

Koenkk: So if I understand correctly, changing the color temp via z2m makes no diff? If yes, then please remove `colorTemp: {range: [333, 333]}`

PR #10800, opened 2025-11-20
feat: Add actions

Nerivec:
See https://github.com/Koenkk/zigbee-herdsman/pull/1559

- add `ACTIONS` record for use in `bridge/definitions` & `action` API
  - Provides a way to hook into ZH Controller with as limited access as possible.
  - Each action can do its own logic, and be called simply by name with `raw` API.
  - Also allows the use of custom clusters without requiring a specific device.

### DIFF
--- a/src/converters/actions.ts
+++ b/src/converters/actions.ts
@@ -93,7 +93,7 @@ export const ACTIONS: Record<string, (controller: Controller, args: Record<strin
     },
     hue_factory_reset: async (controller, args): Promise<undefined> => {
         assert(typeof args.extended_pan_id === "string" && /^0x[a-f0-9]{16}$/.test(args.extended_pan_id));
-        assert(Array.isArray(args.serial_numbers));
+        assert(Array.isArray(args.serial_numbers) && args.serial_numbers.length > 0);
 
         const rawPayload: RawPayload = {
             clusterKey: "manuSpecificPhilipsPairing",


src/converters/actions.ts
```diff
@@ -0,0 +1,134 @@
...
+                commandKey: payload.zcl.command_key,
+                payload: payload.zcl.payload,
+            };
+        }
+
+        return await controller.sendRaw(rawPayload);
+    },
+    hue_factory_reset: async (controller, args): Promise<undefined> => {
+        assert(typeof args.extended_pan_id === "string" && /^0x[a-f0-9]{16}$/.test(args.extended_pan_id));
+        assert(Array.isArray(args.serial_numbers));
```
Mstrodl: Would it be a good idea to ensure it has non-zero length?

Nerivec: Indeed!

Mstrodl:
Just keeping you guys posted: I'm going to stand this up to test this morning.

Mstrodl:
After making the change I suggested in https://github.com/Koenkk/zigbee-herdsman/pull/1559, I was able to pair a light strip using an ember adapter. Strangely, I'm not seeing the message from the device leaving the network anymore which was something I saw in my implementation. Not sure why that would be, unless maybe the gap between sending messages is too long, but I think both our implementations have the same delay. If I have some spare time I might try my implementation again to see if I'm just misremembering the behavior I was seeing.

Nerivec:
Yes, I don't think the gap would be noticeable (even in ms), pretty much the only difference ends up the switching in `sendRaw` to pick the correct function based on params.
When you say the "message from the device leaving", you mean you paired the device to the same network you used to reset it on, so you expected the regular leave message? Not sure what the device does exactly when the reset is triggered (regular factory reset or some added behavior with the custom trigger). Maybe parent-dependent?
_Might want to confirm with a sniffer if you can, see what's happening in the background._

---

Interesting report about using `ember`. I had had some reports a while back that Touchlink (InterPAN) wasn't working properly with some then-recent firmware. What firmware version are you using?

Koenkk:
Thanks!

Mstrodl:
> Yes, I don't think the gap would be noticeable (even in ms), pretty much the only difference ends up the switching in `sendRaw` to pick the correct function based on params. When you say the "message from the device leaving", you mean you paired the device to the same network you used to reset it on, so you expected the regular leave message?

Right. I also tried with a different EPAN ID and didn't see the device leave... I'm not sure why. It'll probably be a while before I have more time to look. My guess is it's just a delay thing... By the time I finish scanning and can look at the main channel again maybe the bulb has given up.

> Not sure what the device does exactly when the reset is triggered (regular factory reset or some added behavior with the custom trigger). Maybe parent-dependent? _Might want to confirm with a sniffer if you can, see what's happening in the background._

I'm not certain. I don't have an easy way to sniff, I want to try and get that set up longer-term. I think I have the tools I need to do it, just a matter of finding the time to get it set up:)

> Interesting report about using `ember`. I had had some reports a while back that Touchlink (InterPAN) wasn't working properly with some then-recent firmware. What firmware version are you using?

I am using a slightly modified version of https://github.com/NabuCasa/silabs-firmware-builder from about May (`65aca7dfb9aa8bef24e62bb57089dadbf997fe51`). Looks like the closest tag to that is v2024.10.21

With that said, I have never been able to make touchlink work on this device (including with the `ncp-uart` example from Simplicity Studio, albeit with an SDK version from like 1-2 years ago). Our usecase is all hue lights, so it's way more convenient to be able to reset them in their normal fixtures using the serial numbers when they are stubborn about pairing.

PR #10804, opened 2025-11-20
update slacky_diy.ts

slacky1965:
fixed "Lock" -> "LOCK and "Unlock" -> "UNLOCK" for thermostat R0A
added LED control for Tuya Plug TS011F-SlD


Koenkk:
Thanks!

PR #10808, opened 2025-11-21
fix Homeetec -37022493 - _TZE200_jhkttplm

elad-bar:
this is a PR to fix the component I have added years ago and someone recently broke it by changing the fingerprint and removing my device from it although available in the whitelist


Koenkk:
Thanks!

PR #10819, opened 2025-11-22
Update avatto.ts

luxandos:
Adding the Radiator valve "_TZE200_4utwozi2"


Koenkk:
This device is already supported, can you check if it works in the latest dev branch?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/e274308d1cb1654c48ec29f250e266a78aeb5a0f/src/devices/tuya.ts#L6594

PR #10830, opened 2025-11-23
fix: Fixed inability to add PEHPL0X as an energy consumption device to Home Assistant

Stonos:
PEHPL0X's `consumed_energy` entity wouldn't show up in Home Assistant's energy configuration. Turns out that Home Assistant expects the unit to be `Wh` and not `W*h`.


Koenkk:
Thanks!

PR #10839, opened 2025-11-24
Add ENGO EONE-230W support

RobXYZ:
ENGO EONE-230W support

### DIFF
--- a/src/devices/engo.ts
+++ b/src/devices/engo.ts
@@ -552,26 +552,6 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "ENGO",
         description: "Zigbee smart thermostat",
         extend: [tuya.modernExtend.tuyaBase({dp: true, timeStart: "2000", queryOnConfigure: true})],
-        configure: async (device, coordinatorEndpoint, logger) => {
-            const endpoint = device.getEndpoint(1);
-
-            // Force Time Sync immediately
-            try {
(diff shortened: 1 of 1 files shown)


src/devices/engo.ts
```diff
@@ -546,4 +546,149 @@ export const definitions: DefinitionWithExtend[] = [
...
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE204_ca3i8m8p"]),
+        model: "EONE-230W",
+        vendor: "ENGO",
+        description: "Zigbee smart thermostat EONE-230W",
+        extend: [tuya.modernExtend.tuyaBase({dp: true, timeStart: "2000"})],
+        configure: async (device, coordinatorEndpoint, logger) => {
+            const endpoint = device.getEndpoint(1);
+
+            // Force Time Sync immediately
+            try {
```
Koenkk: Can you check if the time also syncs immediately without this? I'm wondering if it works because the payload here is different:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/1d5e8196ff48a61b1de6ef4803d01dd8c974ef62/src/lib/tuya.ts#L2216

RobXYZ: You're right, it does work without it - it just takes a few seconds. Thanks :)

Koenkk:
Thanks!

PR #10853, opened 2025-11-25
Fix LIDL HG06106 bulbs

andrei-lazarov:
I removed `doNotDisturb` and `colorPowerOnBehavior` from the Tuya TS0505A converter.
- The settings don't do anything on my TS0505A (`_TZ3000_kdpxju99`) → remove
- I confirm they are working on my TS0505B (`_TZ3210_cieijuw1`) → keep

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -4401,13 +4401,34 @@ export const definitions: DefinitionWithExtend[] = [
             }),
         ],
         whiteLabel: [
-            tuya.whitelabel("Lidl", "HG06106B", "Livarno Lux E14 candle RGB", ["_TZ3000_odygigth"]),
-            tuya.whitelabel("Lidl", "HG06106A", "Livarno Lux GU10 spot RGB", ["_TZ3000_kdpxju99"]),
-            tuya.whitelabel("Lidl", "HG06106C", "Livarno Lux E27 bulb RGB", ["_TZ3000_dbou1ap4"]),
             tuya.whitelabel("Lidl", "14148906L", "Livarno Lux mood light RGB+CCT", ["_TZ3000_9cpuaca6"]),
             tuya.whitelabel("Lidl", "14149505L/14149506L_1", "Livarno Lux light bar RGB+CCT (black/white)", ["_TZ3000_gek6snaj"]),
             tuya.whitelabel("Mycket", "MS-SP-LE27WRGB", "E27 RGBW bulb", ["_TZ3000_evag0pvn"]),
             tuya.whitelabel("Lidl", "HG06104A", "Livarno Home RGB+CCT LED light strip 2m", ["_TZ3000_riwp3k79", "_TZ3000_riwp3k79"]),
+        ],
+        configure: (device, coordinatorEndpoint) => {
+            device.getEndpoint(1).saveClusterAttributeKeyValue("lightingColorCtrl", {
+                colorCapabilities: 29,
+            });
+        },
+    },
+    {
+        fingerprint: tuya.fingerprint("TS0505A", ["_TZ3000_kdpxju99", "_TZ3000_odygigth", "_TZ3000_dbou1ap4"]),
+        model: "TS0505A_1",
+        vendor: "Tuya",
+        description: "RGB+CCT light",
+        extend: [
+            tuya.modernExtend.tuyaLight({
+                colorTemp: {range: [150, 500]},
+                color: true,
+                doNotDisturb: false,
+                colorPowerOnBehavior: false,
+            }),
+        ],
+        whiteLabel: [
+            tuya.whitelabel("Lidl", "HG06106A", "Livarno Lux GU10 spot RGB", ["_TZ3000_kdpxju99"]),
+            tuya.whitelabel("Lidl", "HG06106B", "Livarno Lux E14 candle RGB", ["_TZ3000_odygigth"]),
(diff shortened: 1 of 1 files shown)


Koenkk:
Many devices identify themselves as `TS0505A` so it might work on others, I would propose to keep it.

andrei-lazarov:
Right. In that case I will target the LIDL HG06106 lights

andrei-lazarov:
### DND and colorPoB
I can only confirm the GU10 bulb lacks these features, but I suppose the E14 and E27 form-factors perform the same: `_TZ3000_kdpxju99`, `_TZ3000_odygigth`, `_TZ3000_dbou1ap4`.

### Dull RGB colors
Meanwhile I got a Tuya gateway and I discovered: 
- The **colors are much more vibrant on the Tuya app** (especially red).
- So Z2M at 100% saturation performs like Tuya at 97-98%. 
(Even on Tuya, the saturation slider is not calibrated well for the red LEDs. At 90% it's already white. The other colors are fine.)

**Any ideas how to fix it?** 
I tried multiple combinations with `applyRedFix` and `enhancedHue`, but those didn't change anything as far as I can tell.

Generated definition:
```
export default {
    zigbeeModel: ['TS0505A'],
    model: 'TS0505A',
    vendor: '_TZ3000_kdpxju99',
    extend: [m.light({"colorTemp":{"range":[150,500]},"color":{"modes":["xy","hs"],"enhancedHue":true}})],
};
```
The HS slider doesn't work, only XY work.
Tuya data points:
```
{"1":"Switch","2":"Mode","3":"Bright","4":"Colour Temp","5":"Colour ","6":"Scene","7":"Remaining Time","8":"Music","9":"Debugger"}
```

**Can I get any info from the Tuya platform, other than the datapoints?**

PS. I have one more Moes E14 bulb, with the color more vibrant in the Tuya app: `_TZ3210_ifga63rg` (but the difference is less noticeable) - so maybe it's a general problem?

Koenkk:
Can you try sniffing the [Zigbee traffic](https://www.zigbee2mqtt.io/advanced/zigbee/04_sniff_zigbee_traffic.html) to see what the Tuya hub sends?

andrei-lazarov:
That's exactly what I'm doing right now 😄 
It will take a bit until I get used to it, I'm starting with my Z2M network.

I think I need to find the channel and key for the Tuya hub.

andrei-lazarov:
So setting `_TZ3000_kdpxju99` to full red:
- Z2M sends Move to color x0.64, y0.33
- Tuya sends Move to Hue 0 Saturation 254

andrei-lazarov:
Ok. I tested two bulbs on Z2M+CC2531 and Tuya+LIDL gateway, and sniffed with Dongle-E:
- `TS0505A` `_TZ3000_kdpxju99` - very old LIDL GU10
- `TS0505B` `_TZ3210_ifga63rg` - new Moes E14 ([AliExpress](https://www.aliexpress.com/item/1005005538033680.html))

I identified a lot of problems:
- Both Tuya bulbs **sometimes get confused if you use both XY and HS** commands (and need a reboot to recover)

- The **Z2M XY -> HS conversion is wrong** or _doesn't behave as I expect it to_
  - I interpret the first slider as saturation and the second slider as hue (for both XY and HS)
  - On XY my bulbs never reach 100 saturation. It fluctuates 60-97 depending on the color, even if I have the first slider max right.
  - On XY second slider max left is 5 hue. I expected 0, but 0 is near max right (which is 4).
  - Is XY an attempt at getting different bulbs to display the same color?

- **The LIDL bulb has wrong saturation curve for red (at firmware level)**
  - I wouldn't mind the drawback of slightly less vibrant colors if we are achieving standardized color - but this bulb doesn't display red correctly
  - Something like 90-100 is usable for red if we could remap the saturation curve
  - The other colors are ok I think

- The Moes bulb treats 0s transition as 1s transition
  - responds to custom transitions
  - Z2M converters attempt to set it to 0.1s as a workaround. But I feel it's only applied for ON. It works better if I manually set it to 0.1s.

- The LIDL bulb is hard-coded to 1s transition (regardless of what transitions I use)

- **The LIDL bulb has a custom HS command?**
  - It responds to XY commands
  - It doesn't respond to Z2M HS commands. I see the success response, but nothing happens
 ZCL Frame: `01 11 06 00 fe 00 00` 
 (`01` - enable default response, `11` - sequence number, `06` - move HS, `00` - hue, `fe` - saturation, `00 00` - transition)
  - Tuya hub disables default response (first byte `01` -> `11`) and **appends `ff` after the ignored transition**
 ZCL Frame: `11 2d 06 00 fe 00 00 ff`
  - The `11` is present for the Moes bulb too, but the `ff` is not.

I think this is all

Koenkk:
Could you try to send this command by defining a custom cluster for this device?

andrei-lazarov:
I'll have to figure that out.

But is it intended that my bulbs never reach 100 saturation on XY?

andrei-lazarov:
**It looks to be a problem with the front-end.** I created:
- https://github.com/Nerivec/zigbee2mqtt-windfront/issues/367

> - Full red on Home Assistant sends the correct XY command
> {"x":0.7006,"y":0.2993} 
> which is 0 hue, 100 saturation
> - Full red on Z2M windfront sends the wrong XY command
> {"x":0.6399998137970202,"y":0.32999978276319014}
> which is 5 hue, 97 saturation

**In this case the XY command itself is working as expected!**
I'm not sure if there is any benefit to fixing the HS command. But I'll still have a look.

PR #10857, opened 2025-11-25
fix(ignore): Hue reset make extended pan ID optional

Koenkk:
Default to the extended pan ID of the current network, this avoids having people to look it up and potentially make mistakes in it. I expect that in 99% of the cases they want to join current network.

CC: @Mstrodl

### DIFF
--- a/src/converters/actions.ts
+++ b/src/converters/actions.ts
@@ -91,7 +91,7 @@ export const ACTIONS: Record<string, (controller: Controller, args: Record<strin
 
         return await controller.sendRaw(rawPayload);
     },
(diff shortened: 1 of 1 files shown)


Nerivec:
Already being set by default in WindFront (will avoid that extra request, since we already know it), but I suppose this will provide it for MQTT.

Mstrodl:
Very helpful. Thanks :)

Koenkk:
Also renamed it to `philips_hue_factory_reset`, let me now if this is OK

Nerivec (approved):

PR #10858, opened 2025-11-25
Fixed manual thermostat name setting

slacky1965:


Koenkk:
I don't see any changes under https://github.com/Koenkk/zigbee-herdsman-converters/pull/10858/files, I think something went wrong?

PR #10871, opened 2025-11-27
update 3r dual_plug

3reality-support:
There are two types of dual plugs, one with left and right sockets, and the other with top and bottom sockets, with EP1 located below


Koenkk:
Thanks!

PR #10880, opened 2025-11-27
Update iris.ts

martynwendon:
Adds battery voltage to iris iL07_1 motion sensor.

Sets power source to battery.

(essentially re-uses code from the iris iL06_1 contact sensor as both devices are based on the same underlying hardware)

### DIFF
--- a/src/devices/iris.ts
+++ b/src/devices/iris.ts
@@ -122,18 +122,15 @@ export const definitions: DefinitionWithExtend[] = [
         model: "iL07_1",
         vendor: "Iris",
         description: "Motion Sensor",
-        fromZigbee: [fz.ias_occupancy_alarm_2, fz.temperature, fz.humidity, fz.battery],
+        fromZigbee: [fz.ias_occupancy_alarm_2, fz.temperature, fz.humidity],
         toZigbee: [],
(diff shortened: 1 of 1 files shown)


src/devices/iris.ts
```diff
@@ -122,14 +122,18 @@ export const definitions: DefinitionWithExtend[] = [
         model: "iL07_1",
         vendor: "Iris",
         description: "Motion Sensor",
-        fromZigbee: [fz.ias_occupancy_alarm_2, fz.temperature, fz.humidity],
+        fromZigbee: [fz.ias_occupancy_alarm_2, fz.temperature, fz.humidity, fz.battery],
         toZigbee: [],
```
Koenkk, line 126:
```suggestion
        toZigbee: [],
        extend: [m.battery({voltageToPercentage: "3V_2100"})],
```

Just this change should do it, can you check?

martynwendon: OK, I updated it.  It needs voltageReporting: true as well (since that's false by default).

I also added voltage: true as I guess it would be good to expose the actual voltage as well.

Koenkk:
Thanks!

PR #10882, opened 2025-11-27
Add Air quality sensor heiman HS2AQ-EF

Pastor02:
New converter for the heiman HS2AQ-EF air quality sensor


src/devices/heiman.ts
```diff
@@ -1176,4 +1177,105 @@ export const definitions: DefinitionWithExtend[] = [
...
         endpoint: (device) => ({default: 1}),
     },
+    {
+        fingerprint: [{modelID: "HS2AQ-EF-3.0", manufacturerName: "HEIMAN"}],
+        model: "HS2AQ-EF-3.0",
+        vendor: "Heiman",
+        description: "Air quality monitor",
+        extend: [
+            addCustomClusterHeimanSpecificAirQualityShort(),
+
```
Koenkk:
```suggestion
m.battery(), m.humidity(), 
```
Use this, by using this you can remove the code in fromZigbee, configure and exposes

src/devices/heiman.ts
```diff
@@ -1176,4 +1177,100 @@ export const definitions: DefinitionWithExtend[] = [
         },
         endpoint: (device) => ({default: 1}),
     },
+    {
+        fingerprint: [{modelID: "HS2AQ-EF-3.0", manufacturerName: "HEIMAN"}],
+        model: "HS2AQ-EF-3.0",
```
Koenkk, line 1182: The code looks the same as the `HS2AQ-EM`, can you merge it with that? For proper detection use `whitelabel` (search this repo for examples)

Pastor02: I tried using whitelabel, but the device does not recognise me. When I tried using the same code, the device gave me errors due to the cluster definition, as it returned unsupported attribute and other errors. That is why I had to make a new cluster definition with only compatible features, as this Heiman model has fewer characteristics.

image

image

Koenkk: I see, thanks!

PR #10898, opened 2025-11-30
Namron 4512785

strips:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4500


src/devices/namron.ts
```diff
@@ -1795,4 +1795,264 @@ export const definitions: DefinitionWithExtend[] = [
...
+                        await endpoint.read(0x04e0, [0x0000, 0x0003]);
+                    } catch (_error) {
+                        // Silently ignore polling errors
+                    }
+                },
+            }),
+        ],
+        fromZigbee: [namron.fromZigbee.namron_relay_ntc_water_sensor],
+        toZigbee: [namron.toZigbee.namron_relay_ntc_water_sensor],
+        exposes: (device, options) => {
```
Koenkk: Please use an array here instead of a function: `exposes: [....]`

src/devices/namron.ts
```diff
@@ -1795,4 +1795,264 @@ export const definitions: DefinitionWithExtend[] = [
...
+        vendor: "Namron",
+        description: "Zigbee 30A relay with NTC temperature sensors and water leak detection",
+        extend: [
+            m.poll({
+                key: "namron_4512785_poll",
+                option: e
+                    .numeric("poll_interval", ea.SET)
+                    .withValueMin(-1)
+                    .withDescription("Polling interval in seconds for temperature sensors (default: 60s, -1 to disable)"),
+                defaultIntervalSeconds: 60,
```
Koenkk: Can be less often maybe?

src/devices/namron.ts
```diff
@@ -1795,4 +1795,264 @@ export const definitions: DefinitionWithExtend[] = [
...
+                poll: async (device) => {
+                    const endpoint = device.getEndpoint(1);
+                    try {
+                        await endpoint.read("msTemperatureMeasurement", [0x0000]);
+                        await endpoint.read(0x04e0, [0x0000, 0x0003]);
```
Koenkk: `poll` already has a try/catch, I don't think another one is needed here?

src/lib/namron.ts
```diff
@@ -66,6 +108,68 @@ export const fromZigbee = {
             return result;
         },
     } satisfies Fz.Converter<"hvacThermostat", undefined, ["attributeReport", "readResponse"]>,
+    namron_relay_ntc_water_sensor: {
```
Koenkk: This can be refactored to modernExtend ([example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/5d8679fd1688273ac011f4df60c0c67870951298/src/devices/custom_devices_diy.ts#L882)), also add a custom cluster definition ([example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/5d8679fd1688273ac011f4df60c0c67870951298/src/devices/sonoff.ts#L213))

strips: Thanks, I'll try to do that. Sorry for taking so long with this but I've had some other issues and two of my relays died in December. Should we close this and I make a new one when I'm ready, or do we keep it open and i update this PR? I'm still waiting to replace the two broken ones but I have one more I might do the changes on first.

espenfjo:
@strips It seems like your force push the other day overwrote your work. I'm in need of a heater relay, and this Namron one seems like the best choice since its Zigbee. If you could push what you had, I can take a look into Koens comments and see if can get this merged.

PR #10900, opened 2025-11-30
fix(ignore): Allow users to enter serial written on device

Koenkk:
This PR automatically converters the hex string to a number. This allows users to send exactly what is written on the Hue device itself and prevents them from having to do this translation themselves.

(CC: @Mstrodl)

### DIFF
--- a/src/converters/actions.ts
+++ b/src/converters/actions.ts
@@ -96,7 +96,7 @@ export const ACTIONS: Record<string, (controller: Controller, args: Record<strin
         if (!extendedPanId) {
             extendedPanId = (await controller.getNetworkParameters()).extendedPanID;
         }
-        assert(typeof extendedPanId === "string" && /^0x[a-f0-9]{16}$/.test(extendedPanId));
+        assert(typeof extendedPanId === "string" && /^0x[a-f0-9]{16}$/i.test(extendedPanId));
         assert(
             Array.isArray(args.serial_numbers) &&
                 args.serial_numbers.length > 0 &&


Nerivec:
Does this hold against all Hue devices though? If we always parse as hex, and some devices show it in decimal form...

We run the risk of a user sending a string with decimal in any case.

Koenkk:
From the pictures I could [find](https://www.google.com/search?sca_esv=edc10655191454d0&sxsrf=AE3TifOY3pGE0QxyhjAhDxS5jCekoKaLbw:1764532153221&udm=2&fbs=AIIjpHw2KGh6wpocn18KLjPMw8n5ofjZQFej7XxDpLEwzn5636jPhZQ2SCHYNZOTHhCqX-lIANdm34OLsXpljMXvur-qOVxvA3NXiaKxlYQy-RRiasLGkzd4zqCUc8t8vI2F-h1nJA1utTYvqu2kpTmHUtQ7kaUGVvJcbPCfc9Trwsl4hdZsOuhb1DwK7z1ZfvqQTYc6nJeD&q=hue+serial+number&sa=X&ved=2ahUKEwiC2L_B0pqRAxVYgP0HHT53M3kQtKgLegQIERAB&biw=1470&bih=839&dpr=2#vhid=jKLXps2PAEmXwM&vssid=mosaic), it's always hex. And I would expect this is always the case as otherwise the user would have to select Hex or Decimal in the Hue app which [doesn't seem to be the case](https://youtu.be/Cupp6pOMU7w?si=T7GyTgv-KQRc7UL4&t=58). 

Seems we can even force this to be a 6 char string.

src/converters/actions.ts
```diff
@@ -97,7 +97,13 @@ export const ACTIONS: Record<string, (controller: Controller, args: Record<strin
             extendedPanId = (await controller.getNetworkParameters()).extendedPanID;
         }
         assert(typeof extendedPanId === "string" && /^0x[a-f0-9]{16}$/.test(extendedPanId));
-        assert(Array.isArray(args.serial_numbers) && args.serial_numbers.length > 0);
+        assert(
+            Array.isArray(args.serial_numbers) &&
+                args.serial_numbers.length > 0 &&
+                args.serial_numbers.every((sn) => typeof sn === "string" && /^[a-f0-9]{6}$/i.test(sn)),
```
Nerivec, line 103: Should add `A-Z`, same for ext pan, in case of copying verbatim (seems to be caps on device)

Koenkk: The `/i` makes it case insensitive:

```
> /^[a-f0-9]{6}$/i.test("ABCDEF")
true
```

Added that to the ext pan id now.

Nerivec (approved):

PR #10902, opened 2025-12-01
fix for manufacturerCode in Aqara lights

absent42:
fix for missing manufacturerCode in configure for Aqara lights


Koenkk:
Thanks!

Issue #10870, opened 2025-11-26
LiXee ZLinky: support reportable attribute for pricing period with recent firmware versions


sbraz:
Hi,
Now that the custom cluster has been moved to ZHC in #10704, I'd like to add a new attribute that @fairecasoimeme mentioned in https://github.com/Koenkk/zigbee-herdsman/pull/1310.
This new 0x0010 attribute provides the same information as 0x0200 (the name of the current pricing period, which may vary throughout the day). The benefit of the new attribute is that, [according to the author](https://github.com/fairecasoimeme/Zlinky_TIC/issues/348#issuecomment-3113289205), it supports reporting whereas the old attribute has to be polled. It was added in https://github.com/fairecasoimeme/Zlinky_TIC/commit/04acc113da9ab5062556f6688b4545ee7cfbbc9e (contained in the v16 firmware release).

@Koenkk, could we change the value of the attribute depending on the firmware version or should we just assume that people update their device firmwares?

This is the relevant code on the converter:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/867c2efbb793586bb42c4ea346f0ab11bfba8c0a/src/devices/lixee.ts#L143
https://github.com/Koenkk/zigbee-herdsman-converters/blob/867c2efbb793586bb42c4ea346f0ab11bfba8c0a/src/devices/lixee.ts#L996-L997

Koenkk:
> or should we just assume that people update their device firmwares?

If it's get too complicated (I think it is in the case, would need to split device code), we assume this.

sbraz:
OK I'll create a PR to just change the attribute then.

sbraz:
Not stale but I need to find someone with a Linky counter in legacy mode to test the changes from #10903.

sbraz:
Fixed by #10903 merged in fbbcc924104068269f1cfb89b2b49e1fbecc7668.

PR #10903, opened 2025-12-01
feat: LiXee ZLinky: use reportable attribute for pricing period (LTARF)

sbraz:
Use a reportable attribute (`tariff_period`) for the pricing period:
* for meters in legacy ("historique") mode, use it for PTEC instead of
  the polled attribute (`activeRegisterTierDelivered`).
* for meters in standard mode, use it for LTARF instead of the polled
  attribute (`currentPrice`).

This requires the ZLinky to be on firmware v16 or newer, otherwise the
new attribute (`0x0010`) will not be present
(https://github.com/fairecasoimeme/Zlinky_TIC/commit/04acc113da9a).

### ⚠️  The device needs to be reconfigured to enable reporting for the new attribute.

See also https://github.com/fairecasoimeme/Zlinky_TIC/issues/318 and
https://github.com/fairecasoimeme/Zlinky_TIC/issues/348#issuecomment-3113289205.

Thanks a lot to @KipK for his help.

### DIFF
--- a/src/devices/lixee.ts
+++ b/src/devices/lixee.ts
@@ -1775,6 +1775,7 @@ function getCurrentConfig(device: Zh.Device, options: KeyValue) {
     );
 
     if (device.applicationVersion > 15) {
+        // @ts-expect-error ignore
         myExpose = myExpose.map((e) => {
             // For recent firmwares, use the reportable tariffPeriod attribute
             // instead of the old polled currentPrice and


sbraz:
@Koenkk there's one problem with this: it breaks Home Assistant entity IDs because the entity becomes `sensor.zlinky_tariff_period` instead of `sensor.zlinky_current_price`.

Should I just change the attribute value and keep the names as-is? In that case, the name of the attribute (`currentPrice`) would no longer match the ZLinky's source where it's named `tariffPeriod`: https://github.com/fairecasoimeme/Zlinky_TIC/blob/v17.0/ZLinky/Source/LixeeCluster.h#L35. Instead of the current change, I'd just do:
```diff
diff --git a/src/devices/lixee.ts b/src/devices/lixee.ts
index ba8a9fd6c..9a2b6961b 100644
--- a/src/devices/lixee.ts
+++ b/src/devices/lixee.ts
@@ -46,7 +46,7 @@ const local = {
                     warnDIR2: {ID: 0x0007, type: Zcl.DataType.UINT16},
                     warnDIR3: {ID: 0x0008, type: Zcl.DataType.UINT16},
                     motDEtat: {ID: 0x0009, type: Zcl.DataType.CHAR_STR},
-                    currentPrice: {ID: 0x0200, type: Zcl.DataType.CHAR_STR},
+                    currentPrice: {ID: 0x0010, type: Zcl.DataType.CHAR_STR},
                     currentIndexTarif: {ID: 0x0201, type: Zcl.DataType.UINT8},
                     currentDate: {ID: 0x0202, type: Zcl.DataType.CHAR_STR},
                     activeEnergyOutD01: {ID: 0x0203, type: Zcl.DataType.UINT32},
@@ -210,7 +210,7 @@ const fzLocal = {
                 /* 0x0007 */ "warnDIR2",
                 /* 0x0008 */ "warnDIR3",
                 /* 0x0009 */ "motDEtat",
-                /* 0x0200 */ "currentPrice",
+                /* 0x0010 */ "currentPrice",
                 /* 0x0201 */ "currentIndexTarif",
                 /* 0x0202 */ "currentDate",
                 /* 0x0203 */ "activeEnergyOutD01",
@@ -1064,7 +1064,7 @@ const allPhaseData = [
     {
         cluster: clustersDef._0xFF66,
         att: "currentPrice",
-        reportable: false,
+        reportable: true,
         onlyProducer: false,
         exposes: e.text("LTARF", ea.STATE).withProperty("current_price").withDescription("Current supplier price label"),
     },
```

sbraz:
Or we do this to maintain consistency with the attribute names. And in a few months, we drop the deprecated expose?
```diff
diff --git a/src/devices/lixee.ts b/src/devices/lixee.ts
index ba8a9fd6c..a0618a652 100644
--- a/src/devices/lixee.ts
+++ b/src/devices/lixee.ts
@@ -47,6 +47,7 @@ const local = {
                     warnDIR3: {ID: 0x0008, type: Zcl.DataType.UINT16},
                     motDEtat: {ID: 0x0009, type: Zcl.DataType.CHAR_STR},
                     currentPrice: {ID: 0x0200, type: Zcl.DataType.CHAR_STR},
+                    tariffPeriod: {ID: 0x0010, type: Zcl.DataType.CHAR_STR},
                     currentIndexTarif: {ID: 0x0201, type: Zcl.DataType.UINT8},
                     currentDate: {ID: 0x0202, type: Zcl.DataType.CHAR_STR},
                     activeEnergyOutD01: {ID: 0x0203, type: Zcl.DataType.UINT32},
@@ -211,6 +212,7 @@ const fzLocal = {
                 /* 0x0008 */ "warnDIR3",
                 /* 0x0009 */ "motDEtat",
                 /* 0x0200 */ "currentPrice",
+                /* 0x0010 */ "tariffPeriod",
                 /* 0x0201 */ "currentIndexTarif",
                 /* 0x0202 */ "currentDate",
                 /* 0x0203 */ "activeEnergyOutD01",
@@ -1066,7 +1068,14 @@ const allPhaseData = [
         att: "currentPrice",
         reportable: false,
         onlyProducer: false,
-        exposes: e.text("LTARF", ea.STATE).withProperty("current_price").withDescription("Current supplier price label"),
+        exposes: e.text("LTARF", ea.STATE).withProperty("current_price").withDescription("Current supplier price label (using deprecated polled attribute)"),
+    },
+    {
+        cluster: clustersDef._0xFF66,
+        att: "tariffPeriod",
+        reportable: true,
+        onlyProducer: false,
+        exposes: e.text("LTARF", ea.STATE).withProperty("tariff_period").withDescription("Current supplier price label"),
     },
     {
         cluster: clustersDef._0xFF66,
```

Koenkk:
I propose to avoid any breaking changes for now.

sbraz:
@KipK, @fairecasoimeme, can you please let me know what you think of this? We keep both attributes and expose them as LTARF. I've been using this code for a while and it works just fine. The reported attribute changes really quickly:
image
image

Ideally I'd like to completely remove the polled attribute in a few months to avoid polling it for no reason. But I don't know what @Koenkk thinks of this approach.

KipK:
> @KipK, @fairecasoimeme, can you please let me know what you think of this? We keep both attributes and expose them as LTARF. I've been using this code for a while and it works just fine. The reported attribute changes really quickly:
> Ideally I'd like to completely remove the polled attribute in a few months to avoid polling it for no reason. But I don't know what @Koenkk thinks of this approach.

That's a good change. I gave a look, sounds all good to me.  👍 

@fairecasoimeme could you add a future deprecation notice on your git ? And plan to remove the polling in x months in the converter? 

@sbraz Perhaps better, could we check the fw version in the converter and then disable the polling? Then no need to remove it.
I never tried to do that yet, but should be possible. 
@Koenkk If we detect a version change, can we force a reconfigure from a converter?

fairecasoimeme:
Note that there are two Linky modes (historical and standard). LTARF is reserved for the standard mode, while PTEC is used for the historical mode. The tariff period (0xff66 - 0x0010) applies to both modes.

@KipK: OK, I'll update this information for Z2M in the Git repository. However, if possible, I'd prefer the community or @sbraz to modify the converter. That would be better for everyone ;)

src/devices/lixee.ts
```diff
@@ -220,6 +221,7 @@ const fzLocal = {
                 /* 0x0008 */ "warnDIR3",
                 /* 0x0009 */ "motDEtat",
                 /* 0x0200 */ "currentPrice",
+                /* 0x0010 */ "tariffPeriod",
```
KipK, line 223: just for our OCD ( ^^ ), can you put it one line above ?

KipK:
> LTARF is reserved for the standard mode, while PTEC is used for the historical mode

If I well understood the change, he nows exposes booth LTARF and Tarif Period  over LTARF depending of legacy or up2date fw. 
And keeps legacy PTEC.

Can you confirm @sbraz ?

src/devices/lixee.ts
sbraz: You're right. Done!

sbraz:
> The tariff period (0xff66 - 0x0010) applies to both modes.

@fairecasoimeme, in that case, should we do something like this? We would use the same attribute for both the legacy and standard counter modes. Only its description would change (displayed as PTEC in legacy mode and LTARF in standard mode):
```diff
diff --git a/src/devices/lixee.ts b/src/devices/lixee.ts
index caa741dcd..2d9fc0399 100644
--- a/src/devices/lixee.ts
+++ b/src/devices/lixee.ts
@@ -1441,13 +1441,20 @@ const threePhasesData = [
 });
 
 const legacyData = [
+    {
+        cluster: clustersDef._0xFF66,
+        att: "tariffPeriod",
+        reportable: true,
+        onlyProducer: false,
+        exposes: e.text("PTEC", ea.STATE).withProperty("tariff_period").withDescription("Current supplier price label"),
+    },
     {
         linkyPhase: linkyPhaseDef.all,
         cluster: clustersDef._0x0702,
         att: "activeRegisterTierDelivered",
         reportable: false,
         onlyProducer: false,
-        exposes: e.text("PTEC", ea.STATE).withProperty("active_register_tier_delivered").withDescription("Current pricing period"),
+        exposes: e.text("PTEC", ea.STATE).withProperty("active_register_tier_delivered").withDescription("Current pricing period (using legacy polled attribute)"),
     },
     {
         linkyPhase: linkyPhaseDef.all,
```

> I'd prefer the community or @sbraz to modify the converter. That would be better for everyone ;)

I have no problem updating the converter, I just need us to reach an agreement on the right way to handle this migration.

> @sbraz Perhaps better, could we check the fw version in the converter and then disable the polling? Then no need to remove it.

@Koenkk said it's too complex to check the FW version and I kinda agree: https://github.com/Koenkk/zigbee-herdsman-converters/issues/10870#issuecomment-3587041058

> If I well understood the change, he nows exposes booth LTARF and Tarif Period over LTARF depending of legacy or up2date fw.

Not sure that answers your question but it looks like this for now with the counter in standard mode:
image

> @Koenkk If we detect a version change, can we force a reconfigure from a converter?

@fairecasoimeme is reporting automatically enabled by the ZLinky for this 0x0010 attribute when the firmware is updated? Or do we have to configure it?

KipK:
I don't feel it's overcomplicated to check version at interview/configure step. I can help you for that if needed.
The only thing I don't know (yet), is if we can automatically reconfigure the device from the converter. If possible, then we can handle smooth migration after a fw upgrade. 

But anyway, if a user wants the new attribute after an upgrade, he still can click on the reconfigure button 

edit: @sbraz I'll submit you a PR tomorrow. I think I have something cool for you to test.

KipK:
@sbraz @fairecasoimeme So here is what I did, tell me what you think ( I can't commit yet  because of the pre commit script failing here )

1- Unified Exposure: Added tariffPeriod to legacyData, ensuring it is now exposed in both Standard and Legacy (Historique) modes (previously only in Standard).

2- Firmware Version Check: Modified getCurrentConfig to verify the device's firmware version.

If applicationVersion < 12 (older firmware), tariffPeriod is filtered out to prevent errors or incorrect reporting.
If applicationVersion >= 12 (or unknown, preserving default behavior), tariffPeriod is exposed.

3- Cleanup Obsolete Attributes: For firmware versions >= 12, the deprecated attributes currentPrice (LTARF) and activeRegisterTierDelivered (PTEC) are now filtered out. These attributes required polling and are now superseded by the unified and reportable tariffPeriod attribute.

We ensure that users with newer firmware can benefit from the unified tariff_period attribute regardless of their Linky mode, while maintaining backward compatibility for older devices.

edit: mmh I still can't commit from my dev env. Lots of type errors : 
`Object literal may only specify known properties, and 'write' does not exist in type 'Readonly<AttributeDefinition>'.`

coming from the attributes{} like presencePotential: {ID: 0x0003, type: Zcl.DataType.UINT8, write: true, max: 0xff},

I see the commit prescript spawning the same error from other converters too... what the heck?

sbraz:
Maybe you need to update the herdsman module. You can add `--no-verify` as a workaround in the meantime.

KipK:
Ah thanks, my package were not updated, was late. 
So here is it, I couldn't test yet I'm not home until end of week but should be ok:

https://github.com/sbraz/zigbee-herdsman-converters/pull/1

For now It exposed LTAF with new attribute for booth Historique and Standard mode for fw >= 15, but depending of what we finally decide, we can use same principle and expose new attribute over PTEC for Historique mode.

fairecasoimeme:
> @fairecasoimeme, in that case, should we do something like this? We would use the same attribute for both the legacy and standard counter modes. Only its description would change (displayed as PTEC in legacy mode and LTARF in standard mode):  

yes, it looks good

> I have no problem updating the converter, I just need us to reach an agreement on the right way to handle this migration.
👍

> @fairecasoimeme is reporting automatically enabled by the ZLinky for this 0x0010 attribute when the firmware is updated? Or do we have to configure it?   

No, it must be Z2M to configure the reporting

fairecasoimeme:
> So here is what I did, tell me what you think ( I can't commit yet because of the pre commit script failing here )
> 
> 1- Unified Exposure: Added tariffPeriod to legacyData, ensuring it is now exposed in both Standard and Legacy (Historique) modes (previously only in Standard).
> 
> 2- Firmware Version Check: Modified getCurrentConfig to verify the device's firmware version.
> 
> If applicationVersion < 12 (older firmware), tariffPeriod is filtered out to prevent errors or incorrect reporting. If applicationVersion >= 12 (or unknown, preserving default behavior), tariffPeriod is exposed.
> 
> 3- Cleanup Obsolete Attributes: For firmware versions >= 12, the deprecated attributes currentPrice (LTARF) and activeRegisterTierDelivered (PTEC) are now filtered out. These attributes required polling and are now superseded by the unified and reportable tariffPeriod attribute.

tariffPeriod is available since the v16 (v15 is not stable). Concerning old attributes, just cancel the polling.

KipK:
@fairecasoimeme yes I've corrected. 

I've added the reporting configure to the [PR](https://github.com/sbraz/zigbee-herdsman-converters/pull/1). I've put 15mn max, but not necessary, just change it for a better value if needed

As old attributes are filtered on new fw, the polling won't be enabled

KipK:
@fairecasoimeme  The last point we have to decide now ( If my change works ok )  is 
Do we remove PTEC for fw >15 and unify LTARFF ( now using tariffPeriod ) for booth historic and standard mode
Or do we expose tariffPerdiod over PTEC for historic mode and Ltarff for standard one as it was before, but using TariffPerdiod attr on new fw

For now it's all on Tariff but it's a light change

Note: users will have to click on reconfigure to get new behavior, or it will keep old ones 
If @Koenkk says there's a way to force a reconfigure from the converter, then we could automate that if needed.

fairecasoimeme:
No, I think PTEC or LTARF should be removed.

However, yes, it would be very useful to be able to force a reconfiguration after the update.

On the other hand, the hardest part will be informing users of the change.

KipK:
>No, I think PTEC or LTARF should be removed.

@fairecasoimeme  So we expose a new TariffPeriod and filter the old LTARF & PTEC ? 
What name do you want for this new attr  on z2m ? TariffPeriod ? or in caps ? 
>On the other hand, the hardest part will be informing users of the change.

Add it to your README, @sbraz can add a note on the PR . 
~~Also they will quickly see the new attribute on z2m after they notice the tariff is not updating.~~
outdated, there's no breaking change on new proposal

We will advise users on the french forum too. 

Is the lixee used outside france ?

KipK:
on another side, migration is not a problem. As what I did here should not change anything until users hit the reconfigure button, if they do it means they already knows about the change.

So I think it's better to not force reconfigure and let users handle this. 
@fairecasoimeme Just confirm me the attribute name we expose and I'll do the last change.

KipK:
So I've modified, now LTARF and PTEC are kept for legacy fw and non reconfigured device with new fw.
TARIFFPERIOD is exposed for fw >15 after a reconfigure ( or new pairing )  and filter out LTARF & PTEC
https://github.com/sbraz/zigbee-herdsman-converters/pull/1/changes

Looks to me the best solution. No breaking change.

@Koenkk don't merge yet, my PR is submitted to @sbraz branch and it needs to be tested first.

fairecasoimeme:
> @fairecasoimeme So we expose a new TariffPeriod and filter the old LTARF & PTEC ? What name do you want for this new attr on z2m ? TariffPeriod ? or in caps ?

TariffPeriod and the older LTARF or PTEC labels are compatible. I don't know if filtering is necessary. LTARF and PTEC are in uppercase because they are Enedis labels. TariffPeriod is not an Enedis label, but a personalized label; therefore, it should not be in uppercase.

> Is the lixee used outside france ?

No only in France

KipK:
> s. TariffPeriod is not an Enedis label, but a personalized label; therefore, it should not be in uppercase.

@fairecasoimeme , It's still not clear to us. We need to decide a final layout.

What we propose is keep LTARF and PTEC exposed, and no TarffPeriod directly exposed on z2m. 

Means, when firmware <= 15, it keeps old behavior with polled attributes. And if >15, we feed LTARF and PTEC ( depending of linky mode ) with TariffPeriod reported attribute. 

This way, there's nothing to change on user side, it's 100% transparent. 
Users will have to hit reconfigure button if they want the new behavior, but old behavior still remains until they reconfigure. 

Do you validate that ?

fairecasoimeme:
> Means, when firmware <= 15, it keeps old behavior with polled attributes. And if >15, we feed LTARF and PTEC ( depending of linky mode ) with TariffPeriod reported attribute.
> 
> This way, there's nothing to change on user side, it's 100% transparent. Users will have to hit reconfigure button if they want the new behavior, but old behavior still remains until they reconfigure.
> 
> Do you validate that ?

👍

sbraz:
@Koenkk this should be ready now. I updated the PR description to match what the code does. Please tell me what you think. The biggest caveat is that users need to manually reconfigure reporting the device after upgrading Z2M or reporting won't be enabled and the value will never get updated.

Also, during testing, someone had Z2M show Firmware ID  `4000-0016` but `device.applicationVersion` was `14` so the condition (`if (device.applicationVersion > 15) {`) did not match.
Manually reading the attribute fixed it.
image
Is it likely to occur for a lot of users? When does Z2M update the cached `appVersion` value?

Koenkk:
LGTM! I see there is still a syntax error, can you fix it?

> When does Z2M update the cached appVersion value?

Upon re-interview

sbraz:
The incompatible types one? I don't understand why it occurs. Can you please help with that?
Or maybe @KipK can as he wrote that part.

Koenkk:
Thanks!

sbraz:
> @Koenkk If we detect a version change, can we force a reconfigure from a converter?

@KipK we can since #11148, I opened #11636.

PR #10919, opened 2025-12-03
Fix detect "_TZE200_go3tvswy" as "MG-ZG03W"

krishnachytanya:
Fixes the detection of device under the correct model. This fixes the issue https://github.com/Koenkk/zigbee2mqtt/issues/29968


Koenkk:
Thanks!

PR #10920, opened 2025-12-03
Add support for ZT-304Z soil moisture sensor

MasterFeige:
Add support ZT-304Z


src/devices/tuya.ts
```diff
@@ -21774,4 +21774,47 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint("ZS-304Z", ["Arteco"]),
```
Koenkk: Code looks the same as the `ZS-300Z` (also in this file), can you merge it with that and use `tuya.whitelabel` for proper detection?

MasterFeige: I'm sorry, but I don't know how to merge it. Because the modelID of the new product is "ZS-304Z" instead of "TS0601" for the ZS-300Z product.

Koenkk: Could you also provide the data/database.db entry of this device?

MasterFeige: {"id":90,"type":"EndDevice","ieeeAddr":"0xa4c1388ccefbc6b7","nwkAddr":47989,"manufId":4417,"manufName":"Arteco","powerSource":"Battery","modelId":"ZS-304Z","epList":[1],"endpoints":{"1":{"profId":260,"epId":1,"devId":770,"inClusterList":[0,3,61184,1026,1029,1],"outClusterList":[3,61184,4096],"clusters":{"genBasic":{"attributes":{"modelId":"ZS-304Z","manufacturerName":"Arteco","powerSource":3,"zclVersion":3,"appVersion":0}},"msTemperatureMeasurement":{"attributes":{"65278":25222,"measuredValue":2350}},"genPowerCfg":{"attributes":{"batteryVoltage":30,"batteryPercentageRemaining":200}},"msRelativeHumidity":{"attributes":{"65278":25222,"measuredValue":10000}}},"binds":[],"configuredReportings":[],"meta":{}}},"appVersion":0,"zclVersion":3,"interviewCompleted":true,"interviewState":"SUCCESSFUL","meta":{"configured":-383675228},"lastSeen":1765161604934}

Koenkk:
Thanks!

PnnnG:
Hi! Thanks for adding support for the ZS-304Z to Zigbee2MQTT.

I’ve recently ordered one and was trying to learn a bit more about its illuminance sensor, but couldn’t find much technical information.

If you happen to know:

* Is the illuminance value updated on the same interval as soil_sampling, or does it use its own reporting logic?
* Is the reported illuminance an instantaneous measurement, or is it averaged/smoothed in some way?
* Do you know which light sensor IC is used inside (or whether the value comes directly from the sensor or is processed by the Tuya MCU)?

Thanks!

MasterFeige:
I’m not clear about the specific technical implementation, since every manufacturer has its own solution. One confirmed point is that the light intensity is converted from the voltage of photoresistors or photodiodes collected by the Telink MCU through ADC.

PR #10924, opened 2025-12-03
Tuya PJ1203A single zero removal option

marcelhoogantink:
Added functionality to handle single zero values for power and current readings, allowing for optional removal of these readings based on a new option.


Koenkk:
Thanks!

PR #10930, opened 2025-12-03
feat(tuya): update TRV603WZ definition and add converters

pe-pe:
### TL;DR
- Refactor/improve **TRV603WZ** definition to better match device capabilities.
- Add `divideByWithLimits` to `valueConverterBasic`
- Add `thermostatHolidayStartStopUnixTS` and `thermostatScheduleDayMultiDP_TRV603WZ` to `valueConverter`

### New Functions:
- `valueConverterBasic.divideByWithLimits`: Divides value by a factor and clamps it within min/max limits. Useful for setpoints that report invalid values when heating is disabled.
- `valueConverter.thermostatHolidayStartStopUnixTS`: Converts 8-byte big-endian Unix timestamp pairs to/from a string format "YYYY/MM/DD HH:MM | YYYY/MM/DD HH:MM".
- `valueConverter.thermostatScheduleDayMultiDP_TRV603WZ`: Handles 8-pair schedule per day for TRV603WZ, encoding time and temperature into a custom byte format.

### Fixed(new) Features:
- `holiday_mode`, which was invalid (binary), is replaced by `holiday_time` with the new `thermostatHolidayStartStopUnixTS` converter.
- `schedule_<day>` which was invalid/not-exposed is now correctly handled with the new `thermostatScheduleDayMultiDP_TRV603WZ` converter.

### Breaking Changes:
Below changes are introduced in order to better align with climate control and HomeAssistant GUI components:
- `heating_stop` is replaced by `system_mode` (auto/off).
- `mode` is replaced by `preset` (schedule/manual/holiday).

Below changes are introduced to align to other thermostats and for clarity:
- `window` is renamed to `window_open`
- `switch_scale` is renamed to `anti_scale`

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -7694,35 +7694,45 @@ export const definitions: DefinitionWithExtend[] = [
         exposes: [
             e.battery(),
             e.child_lock(),
-            e.window_detection(),
-            e.binary("window", ea.STATE, "OPEN", "CLOSE").withDescription("Window status closed or open "),
-            e.enum("mode", ea.STATE_SET, ["auto", "manual"]).withDescription("Mode"),
-            e.binary("holiday_mode", ea.STATE_SET, "ON", "OFF"),
-            e.binary("heating_stop", ea.STATE_SET, "ON", "OFF"),
+            e.window_detection_bool(),
+            e.window_open(),
             e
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #10932, opened 2025-12-03
Enable set position for GTZ10

Groax256:
Setting valve_opening (position) worked through tuya developer portal if I turned on the valve_switch (switch_type) toggle. I investigated and tested that with local external converter and that also worked. 
They either patched that in some newer firmware or someone initially was trying to set position without enabling the switch type.


Koenkk:
Thanks!

PR #10934, opened 2025-12-03
feat: Aqara W600 (WT-A03E): Added valve position attribute

GerjandeGroot:
Added the valve position attribute to the Aqara W600 (WT-A03E) Radiator thermostat.


Koenkk:
Thanks!

PR #10937, opened 2025-12-04
fix(lumi): suppress unknown key 65522 debug message

shields:
Aqara devices (bulbs, relays, ceiling lights, etc.) send telemetry data in manuSpecificLumi attribute 65522. The data contains runtime statistics and device metadata but the meaning is not fully decoded.

Add a case to ignore this attribute and prevent noisy debug logs.


Koenkk:
Thanks!

PR #10947, opened 2025-12-04
fix: ota update parsing for newer ledvance plugs

MartB:
Fixes https://github.com/Koenkk/zigbee-OTA/issues/941 with the missed override.

image


Koenkk:
Thanks!

Issue #10939, opened 2025-12-04
occupancy in modernExtend, copy-paste bugs with ultrasonicConfig and contactConfig


theorlangur:
while experimenting with a DIY zigbee presence sensor I came a cross a logic that seems to be a bug:
lib/modernExtend.ts:1024
```
    if (ultrasonicConfig) {
        if (pirConfig.includes("otu_delay")) {
            settingsExtends.push(
                numeric({
                    name: "ultrasonic_otu_delay",
                    attribute: "ultrasonicOToUDelay",
                    valueMin: 0,
                    valueMax: 65534,
                    ...settingsTemplate,
                }),
            );
            attributesForReading.push("ultrasonicOToUDelay");
        }
        if (pirConfig.includes("uto_delay")) {
...
```
surely for `ultrasonicConfig` case we should check `ultrasonicConfig.includes("otu_delay")` and not `pirConfig`..?
This seems like a copy-paste bug to me. Also similar faulty logic can be observed for `contactConfig`. I'm trying to describe a presence sensor with only ultrasonic attributes, no PIR and that fails with the outcome that my external converter is marked as 'invalid'. So at the moment I'm forced to actually have dummy PIR attributes on the device and provide also a `pirConfig` entry in a call to `occupancy` function as a workaround.

theorlangur:
if that's not intentional and actually a bug I don't mind making a small PR with a fix.
just let me know

Koenkk:
Yes, please make a PR

theorlangur:
> Yes, please make a PR

https://github.com/Koenkk/zigbee-herdsman-converters/pull/10950

PR #10950, opened 2025-12-05
fix: modernExtend: occupancy use correct configs (#10939)

theorlangur:
contactConfig and ultrasonicConfig are used instead of pirConfig where appropriate this should allow having an occupancy cluster without PIR and only with Ultrasonic or Contact types.


Koenkk:
Thanks!

Issue #10940, opened 2025-12-04
Candeo: RD1P functionality impacted by PR 9867


candeodevelopment:
@Koenkk we had some reports of our RD1P product having some issues in Z2M and have traced them to this [PR](https://github.com/Koenkk/zigbee-herdsman-converters/pull/9867)

The PR changed the "rd1p_knob_press" function and that  seems to prevent our custom OnOff Cluster command value (0x03) being passed to our converter as "raw" data.

https://github.com/Koenkk/zigbee-herdsman-converters/blob/eab460da43a7ccc2c4784bff747eb41420fc88bb/src/devices/candeo.ts#L195

Our converter is still called successfully for values 0x0, 0x01, 0x02:

```
[04/12/2025, 12:04:43] zh:ezsp:ezsp: <== Frame: 8990014500000401060002014001000001cccf5d9cffff0301470202
[04/12/2025, 12:04:43] zh:ezsp:ezsp: <== 0x45: {"_cls_":"incomingMessageHandler","_id_":69,"_isRequest_":false,"type":0,"apsFrame":{"profileId":260,"sequence":1,"clusterId":6,"sourceEndpoint":2,"destinationEndpoint":1,"groupId":0,"options":320},"lastHopLqi":204,"lastHopRssi":-49,"sender":40029,"bindingIndex":255,"addressIndex":255,"message":{"type":"Buffer","data":[1,71,2]}}
[04/12/2025, 12:04:43] zh:ezsp: processMessage: {"messageType":0,"apsFrame":{"profileId":260,"sequence":1,"clusterId":6,"sourceEndpoint":2,"destinationEndpoint":1,"groupId":0,"options":320},"lqi":204,"rssi":-49,"sender":40029,"bindingIndex":255,"addressIndex":255,"message":{"type":"Buffer","data":[1,71,2]}}
[04/12/2025, 12:04:43] zh:controller: Received payload: clusterID=6, address=40029, groupID=0, endpoint=2, destinationEndpoint=1, wasBroadcast=false, linkQuality=204, frame={"header":{"frameControl":{"frameType":1,"manufacturerSpecific":false,"direction":0,"disableDefaultResponse":false,"reservedBits":0},"transactionSequenceNumber":71,"commandIdentifier":2},"payload":{},"command":{"ID":2,"parameters":[],"name":"toggle"}}
[...]

PR #10952, opened 2025-12-05
Update candeo.ts

candeodevelopment:
Fixes missing "release" event (generated from custom genOnOff Cluster value 0x03) caused by "raw" type data no longer being passed to converter function rd1p_knob_press.

Extended genOnOff cluster using deviceAddCustomCluster to add the 0x03 value and updated type on the rd1p_knob_press function to match.

Thanks for the assist @Koenkk 

Closes https://github.com/Koenkk/zigbee-herdsman-converters/issues/10940


Koenkk:
Thanks!

PR #10956, opened 2025-12-06
Update tuya.ts

Hyperspeed00:
adding new device 
TS0601
_TZE284_4ar89tr1 
https://a.aliexpress.com/_c3o6SzgV

Link to picture pull request: https://ae-pic-a1.aliexpress-media.com/kf/S544edaf5f2444a52a489c5dc16ead1aeo.jpg_960x960q75.jpg_.avif

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -21826,28 +21826,22 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-    fingerprint: tuya.fingerprint("TS0601", ["_TZE284_4ar89tr1"]),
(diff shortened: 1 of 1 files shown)


Koenkk:
Could you check if the additional features provided by the `TS0601_dimmer_1_gang_2` also work?

Hyperspeed00:
done more testing looks like it fits in with TS0601_dimmer_1_gang_1. not 2. the things that doesnt work is the backlight mode which i guess should be hidden using the exposes function. the other thing is that changing the brightness will turn on the lights but will not flip on the state to on

PR #10959, opened 2025-12-07
fix: Bosch BTH-R*: Tweak attribute reporting config

burmistrzak:
This PR adjusts the attribute reporting configuration for Bosch thermostats so that they at least report every hour and at most every minute if the temperature changes by 0.1 degrees.

These parameters might be tuned further, based on user feedback.

Should fix https://github.com/Koenkk/zigbee2mqtt/issues/30048


Koenkk:
Thanks!

PR #10963, opened 2025-12-07
Add Philips Hue 9290012574

neyzm:
Newer revision to the`9290012573A`, though UK variant with winder colour spectrum:
```
color":{"modes":["xy","hs"],"enhancedHue":true}})],
```
Image PR: https://github.com/Koenkk/zigbee2mqtt.io/pull/4530

### DIFF
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -1542,9 +1542,9 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Philips",
         description: "Hue white and color ambiance E26/E27/E14",
         whiteLabel: [
-                {model: "9290012574", vendor: "Philips", description: "Hue White and Color Ambiance B22/E27 806lm", fingerprint: [{modelID: "LCT010"}]},
-            ],
-            extend: [philips.m.light({colorTemp: {range: [153, 500]}, color: {modes: ["xy", "hs"], enhancedHue: true}})],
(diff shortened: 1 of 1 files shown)


neyzm:
There seems to be a clash in the zigbee models, as both are declared as `LCT010`. 
Not sure how to go about this, should we use a different model definition or fingerprint?

Koenkk:
You can add it to the existing definition through a whitelabel, example:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/4bb8e3aff15a6fa701904bf1ee4836f70781a7cc/src/devices/philips.ts#L327

neyzm:
> You can add it to the existing definition through a whitelabel, example:
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/4bb8e3aff15a6fa701904bf1ee4836f70781a7cc/src/devices/philips.ts#L327

If we update the color spectrum via ‎`extend` with the changes in op, will that not apply to all `zigbeeModel`? 
I can only confirm this support for “9290012574” `model`.

Koenkk:
the settings are the same (`color: true` == `color: {modes: ["xy", "hs"], enhancedHue: true}}`)

neyzm:
> the settings are the same (`color: true` == `color: {modes: ["xy", "hs"], enhancedHue: true}}`)

Alright thanks, I made the changes. Though it was previously just ```color: true})],``` I updated it to:
```
 color: {modes: ["xy", "hs"], enhancedHue: true}})],
```

Koenkk:
Thanks!

PR #10972, opened 2025-12-08
HA no longer accepts an illuminance to have a unit of mV. Now changed…

hmchan:
… the unit to lx. Ideally, there is a function to convert mV to lx in the future.


Koenkk:
Let's discuss the proper solution in the issue first, because the unit is not `lux`

hmchan:
I am open. At the before, the unit was shown as lux (though z2m reports it as mV).  After HA applied a stricter check, the reading can no longer be accepted.  All my automations stopped working.  I think this change is no worse than before.  The correctness of the unit is the 2nd priority to me and other users.

If I had a lux meter, I would convert the field to lux. Unfortunately, I haven't.

image

Koenkk:
Closed in favour of https://github.com/Koenkk/zigbee-herdsman-converters/commit/dfcd45115d872b89e0fafe1e15f6ef69a2e1447e

PR #10975, opened 2025-12-08
Update bituo_technik.ts

mrmaximas:
Add  ota: true for SPM01-U01


mrmaximas:
@script0803 FYI

Koenkk:
Thanks!

PR #10983, opened 2025-12-09
fix: enhance ZCL specification

Nerivec:
For https://github.com/Koenkk/zigbee-herdsman/pull/1503

- fix typing for heiman stuff that was moved to ZHC some time ago
- fix improper types for couple of params
- fix compatibility with new metadata for custom clusters (set to behave "as previously")
- fix optional params for `genLevelCtrl` and `lightiningColorCtrl`, per spec, should always be present and set to 0 by default, will just be ignored if some devices don't support it (if some need to be != 0, specific changes can be made in the future, same for custom clusters metadata)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "dependencies": {
         "iconv-lite": "^0.7.0",
         "semver": "^7.7.3",
-        "zigbee-herdsman": "^7.0.4"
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #10986, opened 2025-12-09
Add indicator modes and switch actions for 545D6514

as8net:


Koenkk:
Thanks!

lajtmaN:
@as8net Have you looked into what is required to use the indicators on the wireless dimmer?
https://www.zigbee2mqtt.io/devices/545D6102.html

Setting the `ledIndication` attribute on for each switch directly works as expected. So maybe it's enough to just `extend` with: 
```
indicatorMode("s1"),
indicatorMode("s2"),
indicatorMode("s3"),
indicatorMode("s4"),
```

Screenshot from z2m:             
image

PR #10987, opened 2025-12-10
fix: Sinope TH1123ZB-G2 and TH1124ZB-G2: swap sensing and off values for backlight dimming modes Koenkk/zigbee2mqtt#29828

lemoinem:
This is a fix for Koenkk/zigbee2mqtt#29828
As mentioned (and tested on my own devices), sensing and off values were swapped around.

If that matters, all my thermostats are using Firmware ID: 1305 (202431)


Koenkk:
Thanks!

PR #10993, opened 2025-12-10
fix: biome floating promises detection

Nerivec:
Aligned conf with zh/z2m, cleans things up a bit.
Only 1 real change in `mindy`, assumed it should not be awaited.


Koenkk:
Thanks!

PR #10641, opened 2025-11-03
Eurotronics SPZB0001 TRV, make sure the current_heating_setpoint is used in Home Assistant iso occupied_heating_setpoint 

FrankBakkerNl:
Makes sure the current_heating_setpoint takes priority over occupied_heating_setpoint for Eurotronics SPZB0001 TRV

, https://github.com/Koenkk/zigbee2mqtt/issues/29372

PR #10996, opened 2025-12-10
Fix setting set point of Eurotronic SPZB0001

johannes-mueller:
Reverts 310130d08fa16d08b880f59a1952096b6bfe16c7 as it turned out to break things Koenkk/zigbee2mqtt#30005

I am aware that there has been #10641 in which @FrankBakkerNl did exactly the opposite. As it turned out that seems to break things for many people. We need to figure out what the correct solution is and understand why one way works for some and the other for others.

In my setup this change fixes things for me. Both directions work. Setting set point from within z2m/HA is followed by the device and setting the set point at the device is followed by z2m/HA. Firmware version of my devices is `App: f.f build 255Stack: f.f build 255`.

### DIFF
--- a/src/devices/eurotronic.ts
+++ b/src/devices/eurotronic.ts
@@ -39,8 +39,8 @@ export const definitions: DefinitionWithExtend[] = [
             e.child_lock(),
             e
                 .climate()
-                .withSetpoint("current_heating_setpoint", 5, 30, 0.5)
                 .withSetpoint("occupied_heating_setpoint", 5, 30, 0.5)
+                .withSetpoint("current_heating_setpoint", 5, 30, 0.5)
                 .withLocalTemperature()
                 .withSystemMode(["off", "auto", "heat"])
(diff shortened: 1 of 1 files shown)


Koenkk:
@FrankBakkerNl could you check this?

FrankBakkerNl:
This is the same as the version I tested with before official releasd and was why I did a PR to change the order. I understand that now it is broken for others and you do not want to leave anyone in the cold right now.

One of the things I think is making it confusing is that the order here implicitly influences which gets used in the template. That is what caused mine not to show in HA because the buttons control the current setpoint, while the climate entity uses the occupied setpoint.

Maybe the solution would be to make this more explicit iso depend on the order in which they are added.

FrankBakkerNl:
A quick fix could be to merge this an release a patch. I will stick to current version until we have a solution that works eigther way

Koenkk:
Thanks!

PR #10998, opened 2025-12-10
Update Schneider Electric "NHPB/DIMMER/1" 

rettelx:
Update Schneider Electric "NHPB/DIMMER/1" to use extends and fix broken dimmer mode

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -852,33 +852,23 @@ export const definitions: DefinitionWithExtend[] = [
         model: "WDE002386",
         vendor: "Schneider Electric",
         description: "Push button dimmer",
-        fromZigbee: [fz.on_off, fz.brightness, fz.level_config, fz.lighting_ballast_configuration],
-        toZigbee: [tz.light_onoff_brightness, tz.level_config, tz.ballast_config],
-        extend: [indicatorMode("smart"), m.identify()],
-        exposes: [
-            e.light_brightness().withLevelConfig(),
-            e
-                .numeric("ballast_minimum_level", ea.ALL)
-                .withValueMin(1)
-                .withValueMax(254)
-                .withDescription("Specifies the minimum light output of the ballast"),
-            e
-                .numeric("ballast_maximum_level", ea.ALL)
-                .withValueMin(1)
-                .withValueMax(254)
-                .withDescription("Specifies the maximum light output of the ballast"),
-            e
-                .enum("dimmer_mode", ea.ALL, ["auto", "rl_led"])
-                .withDescription("Sets dimming mode to autodetect or fixed RL_LED mode (max load is reduced in RL_LED)"),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

mchacher:
@Koenkk and @rettelx 

# Some devices reject ZCL commands generated by zigbee-herdsman – no way to send minimal ZCL frames

Hello,

I'm facing an issue with a Schneider Electric Zigbee dimmer which strictly follows its own ZCL implementation and rejects valid Zigbee commands generated by zigbee-herdsman.

This makes it impossible to support the device correctly using Zigbee2MQTT or external converters.

---

## Environment

- Zigbee2MQTT: 2.7.x
- zigbee-herdsman: 8.0.1
- Device: Schneider Electric Odace Zigbee Dimmer
- modelID: `NHPB/DIMMER/1`
- Manufacturer: Schneider Electric
- Firmware: `001.005.000 R`

---

## Problem description

According to the Schneider Electric device specification (I am lucky to have it), this dimmer expects **minimal ZCL frames** for standard commands such as:

- `genOnOff`
  - `Off (0x00)`
  - `On (0x01)`
- `genLevelCtrl`
  - `moveToLevelWithOnOff (0x04)`

### Expected payload (per Schneider specification)

For `moveToLevelWithOnOff`:

- `level` (UINT8)
- `transitionTime` (UINT16)

No additional fields are expected or supported.

---

## Actual behavior

zigbee-herdsman always serializes **optional ZCL parameters** when building frames, notably:

- `optionsMask`
- `optionsOverride`

Even though these parameters are optional according to the Zigbee Cluster Library specification, **this specific Schneider device rejects frames containing them**.

Observed effects:

- Dimming does not work
- On/Off commands may fail or be ignored
- Device may return `MALFORMED_COMMAND` or silently drop commands

---

## Root cause analysis

In zigbee-herdsman, the ZCL definition for `moveToLevelWithOnOff` is:

```js
moveToLevelWithOnOff: {
  ID: 0x04,
  parameters: [
    { name: "level", type: UINT8 },
    { name: "transtime", type: UINT16 },
    { name: "optionsMask", type: BITMAP8 },
    { name: "optionsOverride", type: BITMAP8 },
  ],
}
```

From the converter API, there is currently **no way to**:

- Prevent optional parameters from being serialized
- Explicitly request a minimal ZCL payload
- Send a raw ZCL frame from a `toZigbee` converter

---

## Impact

Because of this limitation:

- External converters cannot support this device correctly
- zigbee2mqtt users cannot reliably control dimming
- Even basic on/off control may fail depending on firmware behavior

---

## Feature request

Is there a supported or planned way to:

1. Send a minimal ZCL command without optional parameters?
2. Expose a low-level API usable from converters?
3. Allow converters to explicitly disable optional ZCL fields?

... or any other solution?

Thanks for your work on zigbee-herdsman.

PR #11007, opened 2025-12-11
Fix detections to _TZE204_xibaabmu and _TZE284_xibaabmu

bernardesarthur:
Fix detections to _TZE204_xibaabmu and _TZE284_xibaabmu as Zemismart ZMS-206US-4


Koenkk:
Thanks!

PR #11009, opened 2025-12-12
Refactor Tuya device fingerprint and whiteLabel entries

depuits:
moved _TZE284_cwyqwqbf to separate entry containing battery state instead of battery percentage.


Koenkk:
Please keep it with the existing one to avoid code duplication, the `exposes` can be made different for this manufacturer name through a function, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/2e5a71f5010457a77e971eceb4187f0624fd0b2b/src/devices/tuya.ts#L1850

depuits:
Updated, let me know if anything else needs to change.

Koenkk:
Thanks!

PR #10990, opened 2025-12-10
feature(add): Sinope thermostats (TH1123ZB(-G2)/TH1124ZB(-G2)) expose ecoMode

lemoinem:
Here is my first tentative to exposing ecoMode for Sinope thermostats TH1123ZB(-G2)/TH1124ZB(-G2).

My thermostats are all -G2, a mix of 23s and 24s. My tests are done on a TH1123ZB-G2.

This gives me the following error:
`z2m sinope No converter available for 'set' 'eco_mode' (-128)`

Even comparing to other properties from the same sinope.ts, I don't see what the difference is.

I've tried adding `, manuSinope` as a third parameter to `entity.read` and `entity.write` calls. but it didn't change much, except that setting the ecoMode stopped working completely.

I've tried adding `logger.debug(JSON.stringify(***), NS)` for `entity`, `value`, `msg` and pretty much every other bit of data I had available, but that only ended up with `undefined variable entity` error message (without a line number).

Even trying to set the ecoMode using the DevTools or a `mqtt.publish` action stopped working (it did before).

Whatever I write I see this appear:
`[10/12/2025, 08:56:13] zhc:tz: Wrote '{"ecoMode":null}' to 'manuSpecificSinope'`

And follow up reads always show:
`[10/12/2025, 08:56:42] zhc:tz: Read result of 'manuSpecificSinope': {"ecoMode":0}`

On the thermostat, the eco indicator is always flashing after any write, even -128, until I discard it by using the physical controls on the thermostat.

Additionally (unsure if related, feel free to ignore if not), since I've added these, (all) my thermostats fail to reconfigure upon restarting Z2M, because of `INSUFICIENT_SPACE`  with this error:
`[10/12/2025, 08:57:55] z2m: Failed to configure 'thermostat', attempt 4 (Error: ZCL command 0xf4b3b1fffeba6093/1 haElectricalMeasurement.configReport([{"minimumReportInterval":10,"maximumReportInterval":65000,"reportableChange":5,"attribute":"activePower"},{"minimumReportInterval":10,"maximumReportInterval":65000,"reportableChange":50,"attribute":"rmsCurrent"},{"minimumReportInterval":10,"maximumReportInterval":65000,"reportableChange":50,"attribute":"rmsVoltage"}], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"reservedBits":0,"writeUndiv":false}) failed (Status 'INSUFFICIENT_SPACE')`

[...]

PR #11014, opened 2025-12-14
fix(better ui): Sinope thermostats (TH1123ZB(-G2)/TH1124ZB(-G2)) exposes better ecoMode

lemoinem:
This is a follow up on #10990.
As was the case over there, I can't test this with an external converter, but since the base PR did work, I'm assuming this one will? It passes the tests!

(I am a bit frustrated I can't test my changes. I don't like blind pushing and potentially wasting maintainers' time)

This will make the UI and values for the state's `eco_mode` much more meaningful and practical to use. :crossed_fingers:

Thanks again for all your help, please let me know if there is any problem with it.


Koenkk:
Thanks!

PR #11034, opened 2025-12-17
fix(better ui): Sinope thermostats (TH1123ZB(-G2)/TH1124ZB(-G2)) exposes better ecoMode

lemoinem:
Another follow up on #10990 and #11014.

The previous PR had a bug in the converter which applied the division by 10 twice.

Hopefully this one will work fine :crossed_fingers:, sorry for the repeated PRs.

### DIFF
--- a/src/devices/easyiot.ts
+++ b/src/devices/easyiot.ts
@@ -1,4 +1,4 @@
-import * as iconv from "iconv-lite";
+import iconv from "iconv-lite";
 
 import * as exposes from "../lib/exposes";
 import {logger} from "../lib/logger";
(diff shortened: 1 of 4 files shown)


lemoinem:
Note: The CI failures comes from a file I didn't modify. I believe the issue comes from `master`.

Koenkk:
Thanks!

lemoinem:
@Koenkk FYI: I'm happy to report that, as far as I can test for now, this version seems to be working correctly.
I'll have a better confirmation once I've had an occasion to go through a real-world scenario using it. But it looks good for now (:

Thanks for your help and great reactivity on these PRs!

PR #11024, opened 2025-12-15
Refactor Philips Surimu color device definitions

chris-1243:
Updated and added some Surimu panels. Most of them have been added as a whitelabel as they share the same definition

From issue:

https://github.com/Koenkk/zigbee2mqtt/issues/30220

Link to picture pull request: Done in

https://github.com/Koenkk/zigbee2mqtt.io/pull/4560

PR #11035, opened 2025-12-17
Update philips.ts

ridiekel:
Philips 929002966501 seems to be the same as Philips 929002966502 (new version?)


Koenkk:
Thanks but this was already added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/11024

PR #11047, opened 2025-12-19
Added Aqara RGB dynamic effect and color conversion functions, added RGB dynamic effects to T2 bulbs

absent42:
Added to lib/lumi.ts:
lumiRgbToXY
lumiEncodeRgbColor
lumiEffect
lumiEffectSpeed
lumiEffectColors

Added RGB dynamic effects to T2 bulbs
Cleaned up T2 CCT/RGB bulb device model detection


Koenkk:
thanks!

PR #11053, opened 2025-12-20
Inovelli - Adding toggle option for parameter 130 (firmware 3.0+)

InovelliUSA:
Simple update to add an option for the Inovelli VZM31-SN firmware 3.0+


Koenkk:
Thanks!

PR #11054, opened 2025-12-20
Added additional device fingerprint - _TZE284_xnwxmj8z

krishnachytanya:
Added an additional Tuya device fingerprint for 1 Gang Switch with power meter


Koenkk:
Thanks!

Issue #11032, opened 2025-12-16
[External Converter]: TS0601 from _TZE284_nj7sfid2


Bjk8kds:
### Link

https://www.aliexpress.com/item/1005009964261392.html

### Database entry

{"id":7,"type":"EndDevice","ieeeAddr":"0x781c9dfffeeca246","nwkAddr":24063,"manufId":4098,"manufName":"_TZE284_nj7sfid2","powerSource":"Battery","modelId":"TS0601","epList":[1],"endpoints":{"1":{"profId":260,"epId":1,"devId":81,"inClusterList":[0,4,5,61184],"outClusterList":[25,10],"clusters":{"genBasic":{"attributes":{"65503":"\u0006��0f\u0006��0\u0012\u0006��0g\u0007��0\u0012","65506":31,"65508":0,"modelId":"TS0601","manufacturerName":"_TZE284_nj7sfid2","stackVersion":0,"dateCode":"","appVersion":77}}},"binds":[],"configuredReportings":[],"meta":{}}},"appVersion":77,"stackVersion":0,"hwVersion":1,"dateCode":"","zclVersion":3,"interviewCompleted":true,"interviewState":"SUCCESSFUL","meta":{},"lastSeen":1764262235616}

### Zigbee2MQTT version

2.7.1-dev (5db6ae0a)

### External converter

```JavaScript
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const tuya = require('zigbee-herdsman-converters/lib/tuya');

const e = exposes.presets;
const ea = exposes.access;

// Custom toZigbee converter for scene binding
const tzLocal = {
    scene_bind: {
        key: ['bind_scene_1', 'bind_scene_2', 'bind_scene_3', 'bind_scene_4'],
        convertSet: async (entity, key, value, meta) => {
            const sceneSlot = parseInt(key.replace('bind_scene_', ''));
            
            const sceneName = `Scene ${sceneSlot}`;
            const nameBuffer = Buffer.alloc(12, 0);
            nameBuffer.write(sceneName, 'utf8');
            
            // Suffix pattern from sniffing:
            // Scene 1: 01 ff ff ff, Scene 2: ff 02 ff ff
            // Scene 3: ff ff 03 ff, Scene 4: ff ff ff 04
            const suffix = Buffer.from([0xff, 0xff, 0xff, 0xff]);
            suffix[sceneSlot - 1] = sceneSlot;
            
            const payload = Buffer.concat([
[...]

PR #11061, opened 2025-12-21
Add support for TS0601_smart_scene_knob

Bjk8kds:
Add support for the Tuya TS0601_smart_scene_knob with 4 buttons and 3 modes (Scene, Light, Curtain). Adds device definition, binding helpers, and action handling for button presses, light, and curtain control.

Fix https://github.com/Koenkk/zigbee2mqtt/issues/29853
Fix https://github.com/Koenkk/zigbee-herdsman-converters/issues/11032

### Device Overview

4-button knob controller with rotation capability. Each button can be configured to one of three modes:

| Mode | LED Color | Cluster/Protocol |
|------|-----------|------------------|
| Scene | Red | Tuya DP 1-4 |
| Light | Green | genOnOff, genLevelCtrl, lightingColorCtrl |
| Curtain | Blue | Tuya DP broadcast (manuSpecificTuya) |

**Mode Switching:** Hold button 2 or button 4 for 5 seconds to cycle between modes. Only bound modes are available (e.g., if only Light mode is bound, mode switching won't work until other modes are bound).

### Group ID Pattern

Device uses Group ID broadcast for Light and Curtain modes. Group IDs are assigned during Tuya Hub pairing via command `0xF0` and **cannot be queried** - must be detected from button press.

**Pattern:** `base_group_id + (button - 1) × 20`

Example with base=11220:
- Button 1: 11220
- Button 2: 11240
- Button 3: 11260
- Button 4: 11280

### Binding (DP 102)

Payload format: `[0x01, 0x01, mode, ...name(12 bytes), ...suffix(4 bytes)]`

- Mode: `0x01`=Scene, `0x03`=Light, `0x04`=Curtain
- Suffix: slot number at position `(slot-1)`, rest `0xff`
  - Slot 1: `[0x01, 0xff, 0xff, 0xff]`
  - Slot 2: `[0xff, 0x02, 0xff, 0xff]`

### Light Mode Actions

| Action | Cluster | Trigger |
|--------|---------|---------|
| `light_X_on/off` | genOnOff | Single press |
| `light_X_brightness_up/down` | genLevelCtrl | Knob rotation |
| `light_X_colortemp_up/down` | lightingColorCtrl | Knob rotation (hold) |

Direction detection: Compare with previous value. At min/max limit, keeps previous direction.

### Curtain Mode Actions

| DP | Value | Action |
|----|-------|--------|
| 1 | 0 | `curtain_X_start` |
| 1 | 2 | `curtain_X_stop` |
| 1 | 1 | _init rotation_ (ignored) |
| 2 | 0-100 | `curtain_X_position_open/close` |

DP 2 is 4-byte big-endian integer (0-100%). Direction determined by comparing with previous position.

### Setup Flow

1. Device joins → status: `unassigned`
2. Click `bind_all_light` or `bind_all_curtain` → status: `waiting_button_1`
3. Press physical button 1 → base Group ID captured → status: `ready`
4. All 4 buttons now correctly mapped

Use `assign_button_1` to reset/reassign, or `set_base_group_id` for manual input.

PS: This device doesn't expose battery status, even in the Tuya app. I'll update if I find anything new about this.

---
_Protocol reverse-engineered via Wireshark sniffing. Converter developed with assistance from Claude AI._

---

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4595

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -407,20 +407,20 @@ const convLocal = {
 // 4 physical buttons with knob rotation, 3 modes: Scene (DP 1-4), Light (ZCL), Curtain (DP broadcast)
 // Mode switch: hold button 2 or 4 for 5 seconds (only cycles through bound modes)
 // Group ID pattern: base + (button-1) * 20, detected from first button press
-const MODE_SCENE = 0x01;
-const MODE_LIGHT = 0x03;
-const MODE_CURTAIN = 0x04;
-const GROUP_ID_OFFSET = 20;
+const modeScene = 0x01;
+const modeLight = 0x03;
+const modeCurtain = 0x04;
+const groupIdOffset = 20;
 
-const STATUS_UNASSIGNED = "unassigned";
-const STATUS_WAITING = "waiting_button_1";
-const STATUS_READY = "ready";
+const statusUnassigned = "unassigned";
+const statusWaiting = "waiting_button_1";
+const statusReady = "ready";
 
 const getButtonFromGroupId = (groupId: number, baseGroupId: number): number | null => {
     if (!baseGroupId || !groupId) return null;
     const offset = groupId - baseGroupId;
-    if (offset >= 0 && offset % GROUP_ID_OFFSET === 0) {
-        const button = Math.floor(offset / GROUP_ID_OFFSET) + 1;
+    if (offset >= 0 && offset % groupIdOffset === 0) {
+        const button = Math.floor(offset / groupIdOffset) + 1;
         if (button >= 1 && button <= 4) return button;
     }
     return null;
@@ -428,7 +428,7 @@ const getButtonFromGroupId = (groupId: number, baseGroupId: number): number | nu
 // DP 102 payload: [0x01, 0x01, mode, ...name(12 bytes), ...suffix(4 bytes)]
 // Suffix: slot number at position (slot-1), rest 0xff
 const bindSlotTS0601SmartSceneKnob = async (entity: Zh.Endpoint | Zh.Group, slot: number, mode: number): Promise<void> => {
-    const modeNames: {[key: number]: string} = {[MODE_SCENE]: "Scene", [MODE_LIGHT]: "Light", [MODE_CURTAIN]: "Curtain"};
+    const modeNames: {[key: number]: string} = {[modeScene]: "Scene", [modeLight]: "Light", [modeCurtain]: "Curtain"};
     const slotName = `${modeNames[mode] || "Scene"} ${slot}`;
 
     const nameBuffer = Buffer.alloc(12, 0);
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -822,6 +860,58 @@ const tzLocal = {
...
+                return {state: {assignment_status: STATUS_WAITING}};
+            }
+            return {};
+        },
+    } satisfies Tz.Converter,
+    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
+    TS0601_smart_scene_knob_assign_button_1: {
+        key: ["assign_button_1"],
+        convertSet: (entity, key, value, meta) => {
+            return {state: {assignment_status: STATUS_WAITING}};
```
```suggestion
            return {state: {assignment_status: status_waiting}};
```

Can you make all of these snake case?

src/devices/tuya.ts
```diff
@@ -1087,6 +1177,214 @@ const fzLocal = {
             }
         },
     } satisfies Fz.Converter<"genOnOff", undefined, ["attributeReport", "readResponse"]>,
+    // TS0601_smart_scene_knob //
+    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
+    TS0601_smart_scene_knob_light_onoff: {
+        cluster: "genOnOff",
+        type: ["commandOn", "commandOff"],
```
Koenkk, line 1184: Is is not possible to use `fz.command_on` and `fz.command_off` here?

src/devices/tuya.ts
```diff
@@ -1087,6 +1177,214 @@ const fzLocal = {
...
+                    brightness: level,
+                    brightness_direction: direction,
+                };
+            }
+
+            return {action_group: groupId, brightness: level};
+        },
+    } satisfies Fz.Converter<"genLevelCtrl", undefined, ["commandMoveToLevelWithOnOff", "commandMoveToLevel"]>,
+    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
+    TS0601_smart_scene_knob_light_colortemp: {
```
Koenkk, line 1266: Could we use `command_move_to_color_temp` here?

src/devices/tuya.ts
Bjk8kds: > Is is not possible to use `fz.command_on` and `fz.command_off` here?

I don't think the standard `fz.command_on`/`fz.command_off` handle the Group ID detection (and the logic for button assignment). This device requires custom handling to map Group IDs of the buttons. So unfortunately no.
Same applies to `command_move_to_color_temp`.
The standard converter doesn't handle the Group ID-based button detection needed here (light mode and curtain mode).

src/devices/tuya.ts
Bjk8kds: > Could we use `command_move_to_color_temp` here?

same as above

src/devices/tuya.ts
Bjk8kds: > Can you make all of these snake case?

Changed to camelCase instead.

Koenkk:
Thanks!

PR #11064, opened 2025-12-21
fix: various energy convertions for Thaleos Thermostats

Gouar:
Some  fixes with energy values sent by thaleos electric heater thermostats


Koenkk:
Thanks!

PR #11068, opened 2025-12-22
Update Philips device descriptions and models

danielpetrovic:
Add new Hue Devote models, and also updated existing mapping that was linked to the wrong version.

Link to picture pull request: You can use the 929003823501, as they look the same.


danielpetrovic:
Not sure if the autofix is correct, please check. There are two SKU's of the S version, one with 1 in the box, and one with 2 in the box. Hence the 1 and 2 at the end of the numbers. And there is one SKU of the M version that is the one I added here.

Koenkk:
Seems the `929003055701` is now twice in this file, could you fix that?

danielpetrovic:
This was probably because of the whitelabel part when I split the device into another model. Could you check again?

Koenkk:
Thanks!

PR #11070, opened 2025-12-22
Update dresden_elektronik.ts to add scene switch

fst-dresden-elektronik:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4589

### DIFF
--- a/src/devices/dresden_elektronik.ts
+++ b/src/devices/dresden_elektronik.ts
@@ -52,22 +52,8 @@ export const definitions: DefinitionWithExtend[] = [
         meta: {disableDefaultResponse: true},
     },
     {
-        zigbeeModel: ["Lighting Switch"],
-        model: "Lighting Switch",
-        vendor: "Dresden Elektronik",
-        description: "2 part zigbee powered light switch",
-        ota: true,
-        extend: [
-            m.deviceEndpoints({endpoints: {"1": 1, "2": 2}}),
-            m.commandsOnOff({endpointNames: ["1", "2"]}),
-            m.commandsLevelCtrl({endpointNames: ["1", "2"]}),
-            m.commandsColorCtrl({endpointNames: ["1", "2"]}),
-            m.battery(),
-        ],
-    },
-    {
         zigbeeModel: ["Scene Switch"],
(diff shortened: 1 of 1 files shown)


src/devices/dresden_elektronik.ts
```diff
@@ -51,4 +51,26 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [m.deviceEndpoints({endpoints: {l1: 11, l2: 12, l3: 13, l4: 14}}), m.light({endpointNames: ["l1", "l2", "l3", "l4"]})],
         meta: {disableDefaultResponse: true},
     },
+    {
+        zigbeeModel: ["Lighting Switch"],
+        model: "Lighting Switch",
```
Koenkk: Could you use the model number written on the device/product page here?

fst-dresden-elektronik: This already is the reported model number.

fst-dresden-elektronik: Somehow there are some spaces at the end of the model id: "Scene Switch   ", but when testing without the spaces the device was configured correctly

Koenkk: It doesn't have to be the reported model of the device, but the model number written on the product page or device itself

fst-dresden-elektronik: Hey there, the model number as well as the product name is "Scene Switch" here.
https://www.dresden-elektronik.de/produkt/scene-switch.html

Koenkk: Would propose to use `BN-600085` here

fst-dresden-elektronik:
You are right, Lighting Switch and Scene switch may be bit to vague, also I think ,I mixed up a part of my commits, this here should now have the correct model for the scene switch.

Koenkk:
Thanks!

Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `BN-600085.png`

PR #11071, opened 2025-12-22
Added lighting switch from dresden-elektronik

fst-dresden-elektronik:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4589

### DIFF
--- a/src/devices/dresden_elektronik.ts
+++ b/src/devices/dresden_elektronik.ts
@@ -53,7 +53,7 @@ export const definitions: DefinitionWithExtend[] = [
     },
     {
         zigbeeModel: ["Lighting Switch"],
-        model: "Lighting Switch",
+        model: "BN-600087",
         vendor: "Dresden Elektronik",
         description: "2 part zigbee powered light switch",
         ota: true,


src/devices/dresden_elektronik.ts
```diff
@@ -51,4 +51,18 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [m.deviceEndpoints({endpoints: {l1: 11, l2: 12, l3: 13, l4: 14}}), m.light({endpointNames: ["l1", "l2", "l3", "l4"]})],
         meta: {disableDefaultResponse: true},
     },
+    {
+        zigbeeModel: ["Lighting Switch"],
+        model: "Lighting Switch",
```
Koenkk: Could you use the model number written on the device/product page here?

fst-dresden-elektronik: This is also the reported model number.

fst-dresden-elektronik: When tested localy this was enought to configure the device.

Koenkk: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/11070/changes#r2644147758

fst-dresden-elektronik: Hey there, the model number as well as the product name is "Lighting Switch" here.
https://www.dresden-elektronik.de/produkt/lighting-switch.html

Koenkk: I would propose to use `BN-600087` (as per the product sheet) here (lighting switch is very vague)

fst-dresden-elektronik:
I updated the model, do the images from here https://github.com/Koenkk/zigbee2mqtt.io/pull/4589 also have to be renamed to the model? If so what happens to the "old" images?

Koenkk:
Yes please also update the images, you can leave the old ones, just add a new e.g. `BN-600087.png`

PR #11079, opened 2025-12-22
gas meter state updated to made it HA friendly

IgnacioHR:
Hi, I'm slowly moving forward with the implementation of the Zigbee Gas Meter and found some changes to contribute to the repository in order to make the code more Home Assistant friendly.

The root of the problem: 

- The Zigbee Specification allows to use m3 and m3/h as units for energy and power but Home Assistant don't.
- The very first implementation of the seMeter for "gas" was using a plain mapping of energy -> energy and power -> power when dealing with gas which was OK for my initial understanding of the surrounding technology (Zigbee)
- At this stage, I'm trying to properly map the Zigbee clusters to what Home Assistant wants to receive from a Gas Meter which is "gas" and "volume_flow_rate"

Personally I don't like how Home Assistant requires us to change values from the zigbee specification to their interpretation of the world but this is current state of the situation so this is why I'm proposing this PR

Regards
Ignacio


src/lib/modernExtend.ts
```diff
@@ -2229,12 +2229,13 @@ function genericMeter(args: MeterArgs = {}) {
         toZigbee = [tz.metering_power, tz.currentsummdelivered, tz.currentsummreceived];
         delete configureLookup.haElectricalMeasurement;
     } else if (args.cluster === "metering" && args.type === "gas") {
-        if (args.power !== false) exposes.push(e.numeric("power", ea.STATE_GET).withUnit("m³/h").withDescription("Instantaneous gas flow in m³/h"));
-        if (args.energy !== false) exposes.push(e.numeric("energy", ea.ALL).withUnit("m³").withDescription("Total gas consumption in m³"));
+        if (args.power !== false)
+            exposes.push(e.numeric("volume_flow_rate", ea.STATE_GET).withUnit("m³/h").withDescription("Instantaneous gas flow in m³/h"));
```
Koenkk, line 2233: This change is not acceptable since it makes `energy` exposed as `gas` for electricity meters, is the change in `gas_metering` not enough?

IgnacioHR: If I'm not mistaken, line 2232 is applicable only to "gas" type meters as restricted in line 2231. Note the value of the "type" argument is provided in lines 2351 (electricity) and line 2371 (gas) respectively.

IgnacioHR: An just for confirmation:
Electricity meters and gas meters are exposed correctly:
image
image
image

Koenkk: You are totally right, thanks!

PR #11081, opened 2025-12-22
fix: detection LEDVANCE 4058075729308

chris-1243:
It should solve issue

https://github.com/Koenkk/zigbee2mqtt/issues/30293

I also move some definitions together as they were really similar (whitelabel). Let me know if I need to revert this.

### DIFF
--- a/src/devices/ledvance.ts
+++ b/src/devices/ledvance.ts
@@ -46,7 +46,6 @@ export const definitions: DefinitionWithExtend[] = [
         whiteLabel: [
             {
                 model: "4099854295256",
-                vendor: "LEDVANCE",
                 description: "SMART+ Plug indoor EU with energy meter (Black)",
                 fingerprint: [{modelID: "PLUG EU EM T, black"}],
             },


src/devices/ledvance.ts
```diff
@@ -39,17 +39,18 @@ export const definitions: DefinitionWithExtend[] = [
...
-    {
-        zigbeeModel: ["PLUG EU EM T, black"],
-        model: "4099854295256",
-        vendor: "LEDVANCE",
-        description: "SMART+ indoor plug EU with energy meter - black",
+        description: "SMART+ Plug indoor EU with energy meter ",
+        whiteLabel: [
+            {
+                model: "4099854295256",
+                vendor: "LEDVANCE",
```
Koenkk:
```suggestion
```

Koenkk:
Thanks!

src/devices/ledvance.ts
chris-1243: Thanks,

Should we also remove `vendor: LEDVANCE`  here

https://github.com/Koenkk/zigbee-herdsman-converters/blob/9e2bad05eb44f63ce8b0f4f50985c99fedf7ade3/src/devices/ledvance.ts#L348

I also modified this definition in the PR submitted

PR #11083, opened 2025-12-23
Update ctm.ts, mKomfy fixes

overskeid:
Add mKomfy 2.5
Update mKomfy 2.0 to use manufacturerZoneAttributes
Move mKomfy Tak to mKomfy

### DIFF
--- a/src/devices/ctm.ts
+++ b/src/devices/ctm.ts
@@ -1061,7 +1061,7 @@ export const definitions: DefinitionWithExtend[] = [
     },
     {
         zigbeeModel: ["mKomfy 2.5"],
-        model: "6254380",
+        model: "mkomfy25",
         vendor: "CTM Lyng",
         description: "2.5 Stove guard",
         extend: [


overskeid:
Image for mkomfy25
https://github.com/Koenkk/zigbee2mqtt.io/pull/4598

I just used the mKomfy logo because this converter will apply to 20+ different stove guards.

src/devices/ctm.ts
```diff
@@ -1029,26 +1029,68 @@ export const definitions: DefinitionWithExtend[] = [
...
     {
         zigbeeModel: ["mKomfy 2.0"],
         model: "6254380",
         vendor: "CTM Lyng",
         description: "2.0 Stove guard",
         extend: [
             m.deviceTemperature(),
-            m.iasZoneAlarm({zoneType: "alarm", zoneAttributes: ["alarm_1", "alarm_2", "tamper", "battery_low"]}),
+            m.iasZoneAlarm({
+                zoneType: "generic",
```
Koenkk, line 1040:
```suggestion
                zoneType: "alarm",
```

Maybe this is better?

src/devices/ctm.ts
```diff
@@ -1029,26 +1029,68 @@ export const definitions: DefinitionWithExtend[] = [
...
+                        description: "Power to stove disconnected",
+                    },
+                ],
+                zoneAttributes: ["tamper", "battery_low"],
+            }),
+            m.electricityMeter({cluster: "metering"}),
+        ],
+    },
+    {
+        zigbeeModel: ["mKomfy 2.5"],
```
Koenkk, line 1063: Since code is the same, can be merged with `mKomfy 2.0` from above, use whitelabel for proper detection, see:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/d0ffd447849e4159eb2d2fcb55556024bf0ee996/src/devices/philips.ts#L211

src/devices/ctm.ts
overskeid: I used 'generic' because 'alarm' overrides one of the manufacturerZoneAttributes.

src/devices/ctm.ts
overskeid: '2.0' does not support onOff cluster.

Koenkk:
I see, thanks!

PR #11101, opened 2025-12-24
fix: latest Philips Hue Ensis white models

benley:
Just like #9149, there seems to be a new pair of model IDs for the latest rev of the product.


Koenkk:
Thanks!

PR #11103, opened 2025-12-25
fix: (TS0601_smart_scene_knob) 'null' at max brightness

Bjk8kds:
Device sends `0xFF` at max brightness, which zigbee-herdsman parses as `NaN` or `null`.
Clamp to 254 (max valid ZCL level per Zigbee spec).


Koenkk:
Thanks!

PR #11106, opened 2025-12-26
Add support for OWON AC221 IR controller

jacky202509:
### Summary
This PR adds support for the OWON AC221, an IR (infrared) controller device.

The AC221 is an IR-only device and does not provide real state feedback from
controlled appliances, therefore optimistic control is used.

### Changes
- Add a new device definition for the OWON AC221 IR controller
- Define the OWON private cluster with proper TypeScript typing
- Implement one-key pairing as a SET-only action
- Parse and expose one-key pairing result data from the device
- Avoid unknown (`?`) UI states in zigbee2mqtt

### Motivation
Previously, the AC221 was not supported and one-key pairing resulted in an
unknown state in the UI. Since IR devices are inherently one-way, treating
pairing as an action instead of a readable state better matches the device
behavior and improves user experience.

### Compatibility
- No breaking changes
- Existing devices and configurations are unaffected

### Tested with
- OWON AC221
- zigbee2mqtt (latest)

### DIFF
--- a/src/devices/owon.ts
+++ b/src/devices/owon.ts
@@ -559,7 +559,7 @@ export const definitions: DefinitionWithExtend[] = [
         zigbeeModel: ["AC221"],
         model: "AC221",
         vendor: "OWON",
-        description: "AC Controller / IR Blaster (AC221)",
+        description: "AC controller / IR blaster",
         extend: [m.deviceAddCustomCluster("manuSpecificOwonAc", OwonClustersDefinition.manuSpecificOwonAc)],
         fromZigbee: [fz.fan, fz.thermostat, fzLocal.owonAcOneKeyPairingResponse, fzLocal.owonAcReadPairingCodeResponse],
         toZigbee: [


src/devices/owon.ts
```diff
@@ -185,6 +279,139 @@ const fzLocal = {
             return result;
         },
     } satisfies Fz.Converter<"fallDetectionOwon", OwonFallDetection, ["attributeReport", "readResponse"]>,
+
+    cb432Metering: {
```
Koenkk: Why is this needed? `m.electricityMeter` should cover this.

src/devices/owon.ts
```diff
@@ -185,6 +279,139 @@ const fzLocal = {
...
+        },
+    } satisfies Fz.Converter<"seMetering", OwonFallDetection, ["attributeReport", "readResponse"]>,
+
+    owonCb432Alarms: {
+        cluster: "haElectricalMeasurement",
+        type: ["attributeReport", "readResponse"],
+        convert: (model, msg, publish, options, meta) => {
+            const result: KeyValue = {};
+
+            if (msg.data.acAlarmsMask !== undefined || msg.data[2048] !== undefined) {
```
Koenkk: given this is standard functionality, this can be added to https://github.com/Koenkk/zigbee-herdsman-converters/blob/5c1bd027d7f994b7c5c4621bd4c6207c999e66c3/src/lib/modernExtend.ts#L1893

src/devices/owon.ts
```diff
@@ -185,6 +279,139 @@ const fzLocal = {
...
+                result.alarm_voltage = (mask & 0x01) > 0;
+                result.alarm_current = (mask & 0x02) > 0;
+                result.alarm_power = (mask & 0x04) > 0;
+            }
+
+            return result;
+        },
+    } satisfies Fz.Converter<"haElectricalMeasurement", undefined, ["attributeReport", "readResponse"]>,
+
+    owonCb432ThresholdRead: {
```
Koenkk: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/11106/changes#r2648314362

src/devices/owon.ts
```diff
@@ -185,6 +279,139 @@ const fzLocal = {
...
+            if (data.acActivePowerOverload !== undefined) result.over_power_threshold = data.acActivePowerOverload;
+
+            return result;
+        },
+    } satisfies Fz.Converter<"haElectricalMeasurement", undefined, ["attributeReport", "readResponse"]>,
+
+    owonAcOneKeyPairingResponse: {
+        cluster: "manuSpecificOwonAc",
+        type: ["commandOneKeyPairingResponse", "commandOneKeyPairingResultUpdate"],
+        // biome-ignore lint/suspicious/noExplicitAny: required for converter compatibility
```
Koenkk: Please remove (and don't use `any` below`)

src/devices/owon.ts
```diff
@@ -185,6 +279,139 @@ const fzLocal = {
...
+                }
+
+                payload.one_key_pairing_result = JSON.stringify({
+                    count,
+                    codes_found: codes,
+                });
+            }
+
+            return payload;
+        },
```
Koenkk, line 333: Use `satisfies Fz.Converter<` with the custom cluster definition here

src/devices/owon.ts
```diff
@@ -185,6 +279,139 @@ const fzLocal = {
...
+            if (msg.data?.pairingCode !== undefined) {
+                const code = msg.data.pairingCode;
+                const displayCode = code === 0xffff ? null : code;
+
+                return {
+                    pairing_code_current: displayCode,
+                };
+            }
+            return {};
+        },
```
Koenkk, line 351: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/11106/changes#r2648315108

src/devices/owon.ts
```diff
@@ -242,6 +469,96 @@ const tzLocal = {
             );
         },
     } satisfies Tz.Converter,
+
+    owonCB432Threshold: {
+        key: ["over_voltage_threshold", "over_current_threshold", "over_power_threshold"],
+        convertSet: async (entity, key, value, meta) => {
```
Koenkk: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/11106/changes#r2648314362

src/devices/owon.ts
```diff
@@ -272,7 +641,67 @@ export const definitions: DefinitionWithExtend[] = [
         model: "CB432",
         vendor: "OWON",
         description: "32A/63A power circuit breaker",
-        extend: [m.onOff(), m.electricityMeter({cluster: "metering"})],
```
Koenkk, line 275: Please keep these extends here (same for other definitions)

src/devices/owon.ts
```diff
@@ -604,19 +1087,19 @@ export const definitions: DefinitionWithExtend[] = [
                 ID: 0xfd00,
                 manufacturerCode: Zcl.ManufacturerCode.OWON_TECHNOLOGY_INC,
                 attributes: {
-                    status: {ID: 0x0000, type: Zcl.DataType.ENUM8, write: true, max: 0xff},
```
Koenkk, line 607: Why was this removed?

src/devices/owon.ts
```diff
@@ -647,8 +1130,9 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SLC631",
         vendor: "OWON",
         description: "Smart plug with doorbell press indicator",
+        meta: {multiEndpoint: true},
         extend: [
-            m.onOff({endpointNames: ["l1", "l2", "l3"]}),
+            m.onOff({endpointNames: ["l1", "l2", "l3"], powerOnBehavior: false}),
             m.iasZoneAlarm({
```
Koenkk, line 937: Use `m.deviceEndpoints` here (remove `meta: {multiEndpoint: true},` and `endpoint: (device) => ({l1: 1, l2: 2, l3: 3}),`)

src/devices/owon.ts
jacky202509: Thanks for the suggestion. I agree this is standard Electrical Measurement
functionality. To keep this PR focused on the CB432 device, I’ll remove the
device-specific alarm handling here and consider adding a generic
modernExtend in a separate PR.

src/devices/owon.ts
jacky202509: Thanks for the suggestion. I agree this is standard Electrical Measurement
functionality. To keep this PR focused on the CB432 device, I’ll remove the
device-specific alarm handling here and consider adding a generic
modernExtend in a separate PR.

src/devices/owon.ts
jacky202509: Thanks for the suggestion. I agree this is standard Electrical Measurement
functionality. To keep this PR focused on the CB432 device, I’ll remove the
device-specific alarm handling here and consider adding a generic
modernExtend in a separate PR.

src/devices/owon.ts
jacky202509: I had remove it

src/devices/owon.ts
jacky202509: Adjusted to satisfies Fz.Converter for the manufacturer-specific cluster; command types are kept untyped as they are not part of ZCL definitions.

src/devices/owon.ts
jacky202509: Adjusted to satisfies Fz.Converter for the manufacturer-specific cluster; command types are kept untyped as they are not part of ZCL definitions.

src/devices/owon.ts
jacky202509: Thanks for the suggestion. I agree this is standard Electrical Measurement
functionality. To keep this PR focused on the CB432 device, I’ll remove the
device-specific alarm handling here and consider adding a generic
modernExtend in a separate PR.

src/devices/owon.ts
jacky202509: Thanks for the suggestion. I agree this is standard Electrical Measurement
functionality. To keep this PR focused on the CB432 device, I’ll remove the
device-specific alarm handling here and consider adding a generic
modernExtend in a separate PR.

src/devices/owon.ts
jacky202509: This attribute is now defined via deviceAddCustomCluster. In this typed definition `write` is not a supported property anymore, so it’s omitted while keeping the same functionality.

src/devices/owon.ts
jacky202509: Using m.deviceEndpoints handles multi-endpoint, so meta.multiEndpoint is not needed.
This device does not support the StartUpOnOff attribute, so powerOnBehavior is explicitly disabled.

src/devices/owon.ts
```diff
@@ -250,15 +460,67 @@ export const definitions: DefinitionWithExtend[] = [
...
+            await ep.read("genOnOff", ["onOff"]);
+        },
     },
     {
         zigbeeModel: ["WSP403-E"],
         model: "WSP403",
         vendor: "OWON",
         whiteLabel: [{vendor: "Oz Smart Things", model: "WSP403"}],
         description: "Smart plug",
-        extend: [m.onOff(), m.electricityMeter({cluster: "metering"}), m.forcePowerSource({powerSource: "Mains (single phase)"})],
```
Koenkk, line 261: Please use this instead of the `configure/tozigbee/fromzigbee/expose`

src/devices/owon.ts
```diff
@@ -250,15 +460,67 @@ export const definitions: DefinitionWithExtend[] = [
         model: "WSP402",
         vendor: "OWON",
         description: "Smart plug",
-        extend: [m.onOff(), m.electricityMeter({cluster: "metering"})],
```
Koenkk, line 253: Same as https://github.com/Koenkk/zigbee-herdsman-converters/pull/11106/changes#r2656351482

src/devices/owon.ts
jacky202509: The device is now migrated to modernExtend.
Power-on behavior is intentionally disabled as WSP403 does not reliably support startUpOnOff.

A minimal configure() is kept only to customize metering reporting parameters,
which are not configurable via modernExtend in this codebase.

src/devices/owon.ts
jacky202509: The device is now migrated to modernExtend.
Power-on behavior is intentionally disabled as WSP402 does not reliably support startUpOnOff.

A minimal configure() is kept only to customize metering reporting parameters,
which are not configurable via modernExtend in this codebase.

src/devices/owon.ts
Koenkk: > A minimal configure() is kept only to customize metering reporting parameters,

To have a consistent experience, all devices should have the same reporting parameters unless there is a really good reason not to, is there one in this case?

jacky202509: The custom reporting parameters are added because WSP403 reports
metering data very infrequently with the default configuration.

In practice this results in delayed or missing power updates,
which makes real-time power monitoring unreliable for this device.

The adjusted reporting improves usability without increasing
network traffic significantly.

Koenkk: >  WSP403 reports metering data very infrequently

I don't understand, it should report with the same frequency as all other plugs right? The default can be found here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/0adcd38c6b8da2aac7d148783b9af5404bb1fed0/src/lib/modernExtend.ts#L1961

jacky202509: You're right, thanks for pointing this out.

After reviewing modernExtend.electricityMeter(), it already configures
standard reporting for seMetering.
The custom configure() is not needed here and will be removed to keep
reporting consistent with other devices.

src/devices/owon.ts
```diff
@@ -331,6 +555,60 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.thermostatAcLouverPosition(endpoint);
         },
     },
+    {
+        zigbeeModel: ["AC221"],
+        model: "AC221",
+        vendor: "OWON",
+        description: "AC Controller / IR Blaster (AC221)",
```
Koenkk:
```suggestion
        description: "AC controller / IR blaster",
```

Koenkk:
Thanks!

Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `AC221.png`

src/devices/owon.ts
jacky202509: Thanks for the suggestion
I agree with this wording. I’ll update the description to “AC controller / IR blaster” when I submit the CB432 device PR, so we can keep them consistent.

jacky202509:
I’ve submitted the device picture for the docs, named AC221.png.

PR #10793, opened 2025-11-20
Add: new feature IKEA E2204/2206

chris-1243:
From discussion

https://github.com/Koenkk/zigbee2mqtt/discussions/29257

I hope I have done it the right way.

PR #11110, opened 2025-12-26
Ikea E2204/E2206

chris-1243:
Following request in PR #10793, `Child lock` is now enabled for software build 2.4.4 or higher. This was tested only with the E2204. I could not test this on E2206.

I have reached my typescript knowledge for the `led enable` option. If someone is able to help here and as suggested in the comment below, it would be great.

https://github.com/Koenkk/zigbee-herdsman-converters/pull/10793#issuecomment-3676229932

CC: @andrei-lazarov

### DIFF
--- a/src/lib/ikea.ts
+++ b/src/lib/ikea.ts
@@ -1046,7 +1046,7 @@ export const ikeaModernExtend = {
                     !isDummyDevice(device) &&
                     device.softwareBuildID &&
                     semverValid(device.softwareBuildID) &&
-                    semverGte(device.softwareBuildID, "2.4.25")
+                    semverGte(device.softwareBuildID, "2.4.4")
                 ) {
                     return [binary(resultName, access.ALL, "LOCK", "UNLOCK").withDescription(resultDescription).withCategory("config")];
                 }


andrei-lazarov:
Thank you! 
We can merge this, but I think I'll remove the version check for child lock entirely, as it works for all versions.
I'll figure out the LED too and make a PR once I confirm it

chris-1243:
Do you have an E2206 to test with ?

If firmware build 2.4.4 for both (E2204/E2206) is the minimum version and `Child_lock` is working, the firmware check codes should be removed.

andrei-lazarov:
No, my E2206 is already updated. Can only test 2.4.4 on E2204.
I understand now why you kept the check 🙂

andrei-lazarov (approved):

Koenkk:
Thanks!

chris-1243:
@andrei-lazarov I just had a quick check and it seems the E2206 has been released after the E2204. In my opinion, the firmware check on the `Child_lock` option may be removed. What do you think ?

andrei-lazarov:
@chris-1243 I see the initial [E2206 support request](https://github.com/Koenkk/zigbee2mqtt/issues/23961) reports firmware: `2.4.34 - 20240215`.
Date matches my **updated** E2204: `2.4.25 - 20240215`
I agree, I'll try to make the LED PR tonight and also remove the version check for child_lock.

chris-1243:
If you remove the whole `if` statement
```javascript
...
        result.exposes = [
            (device) => {
                if (
                    !isDummyDevice(device) &&
                    device.softwareBuildID &&
                    semverValid(device.softwareBuildID) &&
                    semverGte(device.softwareBuildID, "2.4.25")
                ) {
                    return [binary(resultName, access.ALL, "TRUE", "FALSE").withDescription(resultDescription).withCategory("config")];
                }
                return [];
            },
        ];

        return result;
    },
```
And you go as below, it should work.

```javascript
...
        result.exposes = [
            (device) => {
                return [binary(resultName, access.ALL, "LOCK", "UNLOCK").withDescription(resultDescription).withCategory("config")];
            },
        ];

        return result;
    },
```

Of course, there might be a cleanest way...

PR #11118, opened 2025-12-27
Add manufacturer name for ENGO `EONE-230W`

RafalWalczak:
I have nine `ENGO EONE-230W` devices, one of which I ordered later has a supported manufacturer name, the rest have different ones. I use exsternal_covnerters to support them in my local setup, but the best would be to just add manufacturer name for the existing supported configuration files.

Thanks


Koenkk:
Thanks!

PR #6850, opened 2024-01-07
adding ignore_tuya_set_time for _TZE200_2ekuz3dz TuYa X5H-GB-B

Lederhaut:
TuYa X5H-GB-B Thermostat added ignore_tuya_set_time

PR #11124, opened 2025-12-28
fix: remove  ignore_tuya_set_time for _TZE200_2ekuz3dz TuYa X5H-GB-B

Lederhaut:
I added some time ago this flag for this thermostat in [#6850](https://github.com/Koenkk/zigbee-herdsman-converters/pull/6850).
Thermostat's time is set now from time to time to very strange values like 25:34.
Better without time update


Koenkk:
Thanks!

PR #11134, opened 2025-12-29
SZLMR10: fading_time max value from 60 to 300

Bjk8kds:
Increase fading_time max value from 60 to 300 seconds.
Based on the Tuya app, fading time settings range from 5 to 300 seconds, but from device testing the shortest time is 30 seconds.
So the min value remains 30 seconds.


Koenkk:
Thanks!

PR #11135, opened 2025-12-29
Modernextend implementation of Schneider Electric thermostat

svhelge:
Added extended features for Schneider Electric thermostat including setpoints, occupancy, electricity meter, and display settings.

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -1902,7 +1902,7 @@ export const definitions: DefinitionWithExtend[] = [
                 zigbeeCommandOptions: {manufacturerCode: Zcl.ManufacturerCode.SCHNEIDER_ELECTRIC},
             }),
             m.enumLookup({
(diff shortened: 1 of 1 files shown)


svhelge:
@Koenkk, There is a bug with the energy value. For two thermostats the energy divisor is not taken into account (1000 times too high), but for the one I "tested" with it seems OK. Could it be related to the modernExtend functions [in Zigbee-Herdsman](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/modernExtend.ts#L1893).

Koenkk:
Thanks!

Perimec:
Hi, since I have updated to the latest version, it seems that there is some challange with all my thermostat. Several exposed entities are not working anymore. See following image. It seems that now a lot of the exposed entities are not supported anymore.

Perimec:
image

Perimec:
Just to mention, reverting to the previous Zigbee version solved the problem. So this must be definetely the issue

PR #11136, opened 2025-12-29
Fix IKEA KAJPLATS LED2409G6 info

andrei-lazarov:
https://github.com/Koenkk/zigbee2mqtt/issues/30211#issuecomment-3697235378

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4630

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -565,9 +565,9 @@ export const definitions: DefinitionWithExtend[] = [
                 ],
             },
         ],
-        model: "LED2409G6",
+        model: "KAJPLATS_RGB",
         vendor: "IKEA",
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
@Koenkk this converter is still not good. It will match multiple bulbs, both TW and RGB+TW.

Koenkk:
We should probably make the model and description generic as multiple models will match against this.

> It will match multiple bulbs, both TW and RGB+TW.

Can't we do that based on the deviceID? 269 = color, 268 = color temp

andrei-lazarov:
> 269 = color, 268 = color temp

Indeed, I just made a mistake.

I hoped we could uniquely identify each model. But we'll have to go with some generic converters

Koenkk:
Thanks!

PR #11144, opened 2025-12-30
Fixed Ubisys configuration substring

andrewandneil:
Fixed an incorrect substring removal in the Ubisys functions


Koenkk:
Thanks!

PR #11153, opened 2025-12-31
EKO07259 Fix configuration for energy reporting

svhelge:
The meter reporting is on endpoint 5, I struggled with energy values 1000 times too high. Introduced a configure section for ep5 and that seems to have solved the issue.

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -1850,7 +1850,7 @@ export const definitions: DefinitionWithExtend[] = [
                 cluster: "metering",
                 voltage: false,
                 current: false,
-                energy: {divisor: 1000, multiplier: 1}, 
+                energy: {divisor: 1000, multiplier: 1},
                 configureReporting: true,
             }),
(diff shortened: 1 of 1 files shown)


src/devices/schneider_electric.ts
```diff
@@ -1946,6 +1947,14 @@ export const definitions: DefinitionWithExtend[] = [
                 reporting: {min: 0, max: 3600, change: 0},
             }),
         ],
+        fromZigbee: [fz.metering],
```
Koenkk, line 1950: The changes below should not be needed, with just the `energy: {divisor: 1000, multiplier: 1},`: 
Could you provide the debug log when configuring the device via the frontend -> yellow refresh icon? Make sure to wake up the device right before doing so by clicking a button on it.

svhelge: Unfortuneately I don't have access to the devices now (thet are at a second house/cabin 200 km away), the only thing I can do is to interview or reconfigure remotely.

PR #11154, opened 2025-12-31
Fix IKEA BILRESA E2489 & E2490

andrei-lazarov:
Fixes: https://github.com/Koenkk/zigbee2mqtt/issues/30321
Fixes: https://github.com/Koenkk/zigbee2mqtt/issues/30325

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -9,7 +9,7 @@ import {
     ikeaAirPurifier,
     ikeaArrowClick,
     ikeaBattery,
-    ikeaBilresaLong,
+    ikeaBilresaDouble,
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

woytekbode:
@Koenkk I'm not sure what the best place is for feedback on this PR but right now I never get the value of 255 for the max level when using the scroll wheel. The value is null instead.

Koenkk:
Could you provide the debug log when scrolling?

andrei-lazarov:
@woytekbode Should be fixed here:
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/11244

PR #11160, opened 2025-12-31
EKO07259 - Added controltype and fixed energy reporting

svhelge:


Koenkk:
Looks like a duplicate of https://github.com/Koenkk/zigbee-herdsman-converters/pull/11153?

svhelge:
I’m sorry that this got a bit messy.
The configure part is a duplicate, but the " m.enumLookup({  name: "Control_type",) is new. 
But it might be a better solution to extend the Thermostat cluster with several Scneider specific attributes with  m.deviceAddCustomCluster, I believe these attributes are shared with several devices. (see https://github.com/user-attachments/files/16744437/RTC-6754012INT-cFMT_16A_F_Zigbee-29.pdf)

I also realised that "m.occupancy()," should be deleted from the extend part. 
Should I create a PR or can you do it as a part of teh approval?

Koenkk:
Yes, can you combine everything in one pr?

PR #11162, opened 2026-01-01
feat: Bosch BTH-RM*: Enable `auto` mode via `operating_mode` for HA

burmistrzak:
It's currently not super convenient to use schedules with these thermostats because setting `system_mode` to `0x01` a.k.a. **Auto** isn't supported _for some reason_...
To work around this, we'll have to override the HA discovery payload and use `operating_mode` instead.

This is mostly a proof of concept, so any feedback is highly appreciated. 😊

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -744,6 +744,31 @@ export const definitions: DefinitionWithExtend[] = [
         model: "BTH-RM",
         vendor: "Bosch",
         description: "Room thermostat II",
+        meta: {
+            overrideHaDiscoveryPayload: (payload) => {
+                if (payload.mode_command_topic?.endsWith("/system_mode")) {
+                    payload.mode_command_topic = payload.mode_command_topic.substring(0, payload.mode_command_topic.lastIndexOf("/system_mode"));
+                    payload.mode_command_template =
+                        "{% set values = " +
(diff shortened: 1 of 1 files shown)


burmistrzak:
@Koenkk Let's finally make more use of the HA override function. 😉

I tested these changes locally, and everything _seems_ to work perfectly. Kinda unexpected, I must admit. 😅

Also, happy new year to you! 🎆🍀

Koenkk:
Thanks!

cromelex:
> @Koenkk Let's finally make more use of the HA override function. 😉
> 
> I tested these changes locally, and everything _seems_ to work perfectly. Kinda unexpected, I must admit. 😅


I have 2 of these thermostats and just saw this in the release notes. 
I'm essentially using a custom logic to manage the thermostat (and it works fine), but I'm wondering if you actually have any documentation on how you are using this with the auto, or a link you can share. 

If I recall correctly, previously, activating auto was just using a built in schedule.

burmistrzak:
@cromelex My previous comment was about how Z2M handles switching between modes for these thermostats. Not really something of interest for regular users. But feel free to check out the code if you're curious.

You've recalled correctly, it's still the internal schedule but this time around implemented properly. 😉
Some future Z2M release will include support for customizing said schedule.

PR #11163, opened 2026-01-01
Add support for IKEA LED2004G8 and LED1937T5 Japan variants

Gamnn:
Adds support for two IKEA bulbs sold in Japan.


Koenkk:
Thanks!

PR #11167, opened 2026-01-01
TS0601_smart_scene_knob: Adjust brightness and color_temp value ranges

Bjk8kds:
Please forgive my missed mistake.
Updated the min value for brightness from 0 to 1, and adjust the color_temp range from 150-500 to 153-523, to better reflect supported device parameters.


Koenkk:
Thanks!

Issue #11175, opened 2026-01-02
Broken Ubisys Variable Writing


andrewandneil:
Ubisys have some variables that are read and write, differing from the official spec.
From what I can tell, the changes in https://github.com/Koenkk/zigbee-herdsman/pull/1503 have broken the ability to write to these values, which breaks the ability to calibrate the Ubisys J1 blind controller.

For example, putting it in calibration mode fails:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/3098f876071fcd30222c47e6157b8c9fa43be5bc/src/devices/ubisys.ts#L153-L156
And the initial write during configuration to "reset" the controller fails as it can't write to the non-ubisys specific attributes.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/3098f876071fcd30222c47e6157b8c9fa43be5bc/src/devices/ubisys.ts#L172-L185

See below the spec given by Ubisys for these variables:
<img width="1345" height="554" alt="Image" src="https://github.com/user-attachments/assets/67261111-4157-4132-91fa-d8f5f68221e5" />
<img width="1384" height="312" alt="Image" src="https://github.com/user-attachments/assets/9feab70a-9c84-42ad-8e49-868a77879f94" />

Is there a way of overriding the the read-only status of these default variables, or would we have to create a new manufacturer specific attribute for each of these, and point to them instead?
Happy to send any logs required if you can point me to where I get them.

PR #11177, opened 2026-01-02
fixed Ubisys J1 attributes

andrewandneil:
Fixes: https://github.com/Koenkk/zigbee-herdsman-converters/issues/11175
Adds Ubisys specific attributes that are writable in order to allow the calibration of Ubisys J1 blind controllers which was broken in the latest patch.


Koenkk:
Thanks!

PR #11178, opened 2026-01-02
Bugfix double trigger of state on single button click for _TZE204_wktrysab

rhakbari:
Double triggering of state on single button click on HA and Z2MQTT frontend for _TZE204_wktrysab 8 gang wall switch.

The issue was happening when the device was paired again after deleting it from HA. When the button was clicked once the state was not registering to the device and when there was any other trigger of state on any other switch or the same switch then the previous state was register with the new one. 

This issue has been resolved in the PR.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -951,6 +951,40 @@ const fzLocal = {
             return {action: event};
         },
     } satisfies Fz.Converter<"genLevelCtrl", undefined, ["commandStep"]>,
+
+    four_buttons_scene_events: {
+        cluster: "genScenes",
+        type: ["commandRecall"],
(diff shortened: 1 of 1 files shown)


rhakbari:
@Koenkk Kindly review and merge. Thanks

Koenkk:
This duplicates a lot of code, can the fix be integrated into the existing code?

rhakbari:
@Koenkk I can do it in the existing code but i dont know if people with other models with the existing code are facing the same issue as me and if they aren't then there's a chance this code might cause issue for them because this fix is specifically for the model I am using at my home.

Koenkk:
Could you provide me the debug log when the duplicate trigger happens?

PR #11181, opened 2026-01-02
Fix: Paulmann 948.47/29165  color temperature

chris-1243:
It should fix issue

https://github.com/Koenkk/zigbee2mqtt/issues/30459


Koenkk:
Thanks!

PR #11184, opened 2026-01-03
Namron Simplify Zigbee dimmer 4512791 (Update)

xses79:
Update : 
Max /Min Brightness level fix.
Fix for on_level set.
Adding Dimming speed.

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -2194,21 +2194,6 @@ export const definitions: DefinitionWithExtend[] = [
             }),
         ],
 
-        // Sørg for state/brightness uten m.light (så vi unngår effect/dobbel)
-        fromZigbee: [fz.on_off, fz.brightness],
-
-        toZigbee: [
-            tz.on_off,
-            tzLocalSimplifyDimmer4512791.brightness_clamped,
-
-            tzLocalSimplifyDimmer4512791.min_brightness,
-            tzLocalSimplifyDimmer4512791.max_brightness,
-            tzLocalSimplifyDimmer4512791.dimming_speed,
-
-            // Re-use built-in (sets genLevelCtrl.onLevel)
-            tz.level_config,
-        ],
-
         exposes: [
             e.light_brightness(),
 


src/devices/namron.ts
```diff
@@ -211,6 +211,127 @@ const tzLocal = {
...
+        convertSet: (entity: TzEntity, key: string, value: unknown, meta: TzMeta) => {
+            const s = Number(value);
+            if (!Number.isFinite(s) || s < 1 || s > 10) throw new Error("dimming_speed must be 1..10 seconds");
+            store.putValue(meta.device, "dimming_speed", s);
+            return {state: {dimming_speed: s}};
+        },
+    } satisfies Tz.Converter,
+
+    // Device-backed via genLevelCtrl.onLevel (startUpCurrentLevel unsupported on your device)
+    start_brightness: {
```
Koenkk: this converter is not needed, can re-use https://github.com/Koenkk/zigbee-herdsman-converters/blob/42536619d42df3fa9e7100e889128c5ce9d97e2f/src/converters/toZigbee.ts#L819

src/devices/namron.ts
```diff
@@ -1780,19 +1901,68 @@ export const definitions: DefinitionWithExtend[] = [
         model: "4512791",
         vendor: "Namron",
         description: "Namron Simplify Zigbee dimmer (1/2-polet / Zigbee / BT)",
+
+        // Modern extend kun for målinger (ingen effect)
         extend: [
-            m.light({}),
```
Koenkk, line 2089: I think this PR should be implemented differently, keep this and expose option such that the brightness can be clamped in https://github.com/Koenkk/zigbee-herdsman-converters/blob/42536619d42df3fa9e7100e889128c5ce9d97e2f/src/converters/toZigbee.ts#L1213

xses79 (commented):
Changed it.

src/devices/namron.ts
xses79: Made some changes.

src/devices/namron.ts
Koenkk: @xses79 I see this line is still removed and that e.g. `fromZigbee: [fz.on_off, fz.brightness],` is still present.

xses79:
Closing to re-work newnew

PR #11200, opened 2026-01-04
Update ikea.ts

phiphi060:
The actuel converter show an EU smart plug E2206 for the CH smart plug E2224 ! (wrong picture)


Koenkk:
Thanks!

PR #11204, opened 2026-01-05
fix: Bosch Twinguard: Rework air quality feature

burmistrzak:
This PR contains breaking changes, but for good reason.

After extensive research, it [became clear](https://community.bosch-sensortec.com/mems-sensors-forum-jrmujtaw/post/bme680-min-and-max-values-of-co2e-and-b-voc-u8vzMo8UtKyV63K) that a [simple conversion](https://community.bosch-sensortec.com/mems-sensors-forum-jrmujtaw/post/bme680---how-do-i-convert-the-iaq-value-to-ug-m3-WAJhoehBZ4DEJBU) from ppm to µg/m³ is not really possible for [this type](https://community.bosch-sensortec.com/mems-sensors-forum-jrmujtaw/post/explanation-on-static-iaq-breath-voc-and-co2-equivalent-oe2xLhefJakKSLE) (Bosch Sensortec BME680) of [TVOC sensor](https://community.bosch-sensortec.com/mems-sensors-forum-jrmujtaw/post/explanation-on-static-iaq-breath-voc-and-co2-equivalent-oe2xLhefJakKSLE).

Checking the Twinguard [help article](https://www.bosch-smarthome.com/uk/en/support/help/product-help/twinguard-help/), you'll also see the following:

> The Twinguard can detect values between 500 ppm and 5500 ppm. Please note that there is an error tolerance of approximately +/- 15% at these limit values.

So we can _somewhat_ safely assume that a linear conversion from raw AQI (0–500) to ppm (500–5500) is done by the SHC. However, we would need longer-term data from a SHC-paired Twinguard to verify.

After about four years, we pretty much ended up exactly where @safakaltun left off. 🙈
If anyone can come up with an accurate formula to estimate µg/m³ from ppm, _please_ let me know!

### Breaking changes

- Remove `voc` from expose
- Adjust CO2-equivalent formula

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -224,13 +224,7 @@ const boschExtend = {
         };
         const exposes: Expose[] = [
             e.binary("smoke", ea.STATE, true, false).withDescription("Indicates whether the device detected smoke"),
-            e
-                .numeric("temperature", ea.STATE)
-                .withValueMin(0)
-                .withValueMax(65)
-                .withValueStep(0.1)
-                .withUnit("°C")
-                .withDescription("Temperature"),
+            e.numeric("temperature", ea.STATE).withValueMin(0).withValueMax(65).withValueStep(0.1).withUnit("°C").withDescription("Temperature"),
             e.numeric("humidity", ea.STATE).withValueMin(0).withValueMax(100).withValueStep(0.1).withUnit("%").withDescription("Relative humidity"),
             e
                 .numeric("co2", ea.STATE)


Koenkk:
I see this more as a fix instead of just a breaking change (as the current value is not correct), let me know once this is ready for merge

burmistrzak:
@Koenkk Yea, true.
Sadly, it's simply not possible to derive TVOC density from eCO2 or AQI.

However, we _could_ convert eCO2 from ppm to ug/m3 and use it as `voc`..? Not sure how useful that would be tho...

burmistrzak:
@itavero Just as a quick heads-up, I'd suggest you treat `co2` as an air quality sensors as well.
This device is _somewhat_ popular and people might be surprised when their air quality sensors suddenly disappear in Homebridge. 😅

Koenkk:
Thanks!

burmistrzak:
@Koenkk While we're at it, should we rename`co2` to `eco2` as well?
The reasoning here being that this device actually **can't measure** CO2 levels but instead relative TVOC levels **expressed** as CO2-equivalent.

Also, looking at the Z2M device catalog, there're quite a few eCO2/TVOC devices presenting as CO2 sensors. 👀

itavero:
> The reasoning here being that this device actually **can't measure** CO2 levels but instead relative TVOC levels **expressed** as CO2-equivalent.

In my experience from working in R&D in the HVAC industry for quite some years, eCO2 and CO2 are definitely not the same (it's an approximation, but the absolute values tend to differ quite a lot depending on the environment it is used in).
However, changing this for all of them would cause a lot of backwards incompatibility and I reckon the average user does not care that it's not actually CO2.

burmistrzak:
@itavero Thanks for your valuable input! 🙌
These TVOC/eCO2 sensors are indeed only useful to roughly estimate a change in air quality, no matter how bad the initial conditions were.
Having them output a cooked up `ppm` value, using a proprietary algorithm, is somewhat misleading.
I prefer the unit-less IAQ-Index for these MEMS sensors, because it doesn't suggest any specific accuracy.

I'm just a bit worried about people relying on these sensors to alert them in potentially life-threatening or dangerous situations.

@Koenkk I think we should _at least_ encourage contributors to double-check and properly declare a sensor's actual capabilities, especially in **safety-related** applications.
For example, a _heat_ detector exposed as _smoke_ sensor could be quite dangerous. Same goes for a TVOC sensor exposed as an actual CO2 sensor. Just some food for thought. 😅

PR #11216, opened 2026-01-06
Modified the part that unconditionally brings in the first candidate during candidate definition

yjyjkim-cr:
In the "zigbee-herdsman-converters" version 25.4.0, if the TNCE device (https://www.zigbee2mqtt.io/devices/LKTMZL02-z.html, a temperature/humidity sensor) is not explicitly defined as a supported device, it should be marked as unsupported. However, the current code mistakenly matches it to the first candidate in the list, which is a light device from another vendor. This causes incorrect associations. In newer versions, the TNCE is supported and works properly, but in older versions like 25.4.0, the code erroneously binds it to unrelated devices.

Devices with the same zigbeeModel "TLSR82xx" are assigned to different sensor types (e.g., temperature/humidity, light, PIP sensor). Why does the code always prioritize the first match in the list, even if the device type mismatches? This logic seems flawed when the same model is reused for multiple device types. I'd like to suggest this as a contribution to improve device identification accuracy.
    {
        zigbeeModel: ["TLSR82xx"],
        model: "33951/33948",
        vendor: "AwoX",
        description: "LED white",
        extend: [m.light()],
        whiteLabel: [{vendor: "EGLO", model: "12229"}],
    },


Koenkk:
Could you share the data/database.db entry of this device?

yjyjkim-cr:
@Koenkk 
Is it correct to always prioritize the first candidate definition in the list solely based on an identical ModelID (e.g., zigbeeModel: "TLSR82xx"), even when the device types differ (e.g., temperature/humidity sensor vs. light or PIP sensor)? I'd like to confirm whether this logic is flawed or if the device itself has an issue. Currently, the device works properly with Fingerfinter, but previously, it was being incorrectly assigned to unrelated definitions due to this behavior

https://github.com/koenkk/zigbee2mqtt/issues/29142 

tnce device info
  "_events": {},
  "_eventsCount": 1,
  "ieeeAddress": "a4:c1:38:6a:6d:cc:e0:d7",
  "productId": "TLSR82xx",
  "manufacturerName": "TELINK",
  "__client":

Koenkk:
- First it tries to identify based on the `fingerprint`
- If no fingerprint matches, it will match on the `zigbeeModel`

I think what we should do is to replace `zigbeeModel: ["TLSR82xx"],` with a `fingerprint`. This is what I did now.

yjyjkim-cr:
@Koenkk  
Thank you for your corrections and comments.
I have one more question. 
The current fix appears to be a measure for a single device ModelID.
For example, if a new device without a definition has the same model ID and lacks a fingerprint definition in the backend, and it gets registered as a different device type (like the issue above), should we then modify the corresponding device definition each time with a fingerprint as described above?

The code below (unconditionally selecting the first device defintion from the candidate device definitions) appears to carry the risk of registering new devices with incorrect device type definitions. I would like to hear your opinion.

https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/index.ts#L566 

        // Match based on fingerprint failed, return first matching definition based on zigbeeModel
        for (const candidate of candidates) {
            if (candidate.zigbeeModel && device.modelID && candidate.zigbeeModel.includes(device.modelID)) {
                return candidate;
            }
        }

Koenkk:
> should we then modify the corresponding device definition each time with a fingerprint as described above?

Yes, if a device is incorrectly detected, it means the current `fingerprint` is not narrow enough.

> The code below (unconditionally selecting the first device defintion from the candidate device definitions) appears to carry the risk of registering new devices with incorrect device type definitions. I would like to hear your opinion.

This code only matches when:
- No fingerprint match is found
- A definition is present which has the `device.modelID` in the definition `zigbeeModel` list

So if a device uses a generic modelID (like `TS0601` or `TLSR82xx`) which is shared by many different device, the definition should never have a `zigbeeModel`, instead it should have a `fingerprint`

PR #11217, opened 2026-01-06
Add new fingerprint for Moes Dimmer Module (MS-105-M)

curtash:
Similar to one already added but my version has a "z" on the end. I've tested with an external converter and works perfectly.


Koenkk:
Thanks!

PR #11218, opened 2026-01-06
Fix datatype in Niko switchAction attribute

vdkeybus:
Set switchAction data type of the Niko 552-721X2 to 27 (0x1b), which is BITMAP32.
Set limit to 0xffff, corresponding to 2 bytes. Possibly more are used, but at least not recognized by the converter. If the limit is reached again, we review this.


Koenkk:
Thanks!

PR #11223, opened 2026-01-07
Add Clear Energy Function

script0803:


Koenkk:
Thanks!

mrmaximas:
@script0803 hi! I don't think it was the best idea to hang the energy reset function on ‘state’. Many systems will take this entity as a switch.
I think something from this list or something else would be more logical.
```
reset_consumption
reset_energy
reset_energy_total
reset_total_energy
reset_energy_meters
```

PR #11229, opened 2026-01-07
Add Sunricher SR-ZG9001T2-SW device support

dipe:
Added support for Sunricher Zigbee 2 channel AC phase-cut dimmer SR-ZG9001T2-SW.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4679

### DIFF
--- a/src/devices/sunricher.ts
+++ b/src/devices/sunricher.ts
@@ -2487,6 +2487,6 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SR-ZG9101SAC-HP2",
         vendor: "Sunricher",
         description: "Zigbee 2 channel AC phase-cut dimmer",
-        extend: [m.deviceEndpoints({endpoints: {l1: 1, l2: 2}, configureReporting: true}), m.light({endpointNames: ["l1", "l2"]})],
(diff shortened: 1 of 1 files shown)


src/devices/sunricher.ts
```diff
@@ -2482,4 +2482,11 @@ export const definitions: DefinitionWithExtend[] = [
             await endpoint.read("hvacThermostat", [0x1008], options);
         },
     },
+    {
+        fingerprint: [{modelID: "DIM", manufacturerName: "Sunricher"}],
+        model: "SR-ZG9001T2-SW",
```
Koenkk: This model is already in use: https://github.com/Koenkk/zigbee-herdsman-converters/blob/83eb1da8a55580f72034a48f443a725536539a0f/src/devices/sunricher.ts#L195, are you sure it uses the same?

dipe: Apologies, my mistake. I have corrected the model name. I will also correct the file name of the image afterwards.

When creating the pull request, I saw that a check said that configureReporting did not exist here. However, I do not know what the correct solution is and would appreciate some guidance.

Koenkk:
Thanks!

PR #11231, opened 2026-01-07
fix(detect): Detect `_TZ3000_iv4eq7eh` as AVATTO ZWSM16-3-Zigbee

andrei-lazarov:
New AVATTO 3-gang variant from Discord.


Koenkk:
Thanks!

PR #11236, opened 2026-01-08
Revert "fix: Sinope TH1123ZB-G2 and TH1124ZB-G2: swap sensing and off values for backlight dimming modes Koenkk/zigbee2mqtt#29828"

lemoinem:
As per feedback on https://github.com/Koenkk/zigbee2mqtt/issues/29828, I am reverting the changes since it breaks other devices and will try to provide a better fix later on.

Sorry.


Koenkk:
Thanks!

PR #11238, opened 2026-01-08
fix(utils): handle case-insensitive enum values in getFromLookupByValue

UrbanZan:
## Problem

Some Sonoff TH sensors (SNZB-02D, SNZB-02LD, SNZB-02WD, SNZB-02DR2) with newer firmware (v8960+) send `temperature_units` as `"Celsius"` or `"Fahrenheit"` (capitalized) instead of `"celsius"` or `"fahrenheit"` (lowercase).

This causes Home Assistant to reject the value with an error like:

`Invalid option for select.office_th_sensor_temperature_units: 'Celsius' (valid options: ['celsius', 'fahrenheit'])`

## Solution

Added a case-insensitive key matching fallback in `getFromLookupByValue()`. When the device sends a string value that doesn't match any lookup value exactly, the function now checks if the string matches any lookup key case-insensitively and returns the correctly-cased key.

- Device sends `"Celsius"` → returns `"celsius"` ✅
- Device sends `"Fahrenheit"` → returns `"fahrenheit"` ✅

## Affected Devices

- Sonoff SNZB-02D
- Sonoff SNZB-02LD  
- Sonoff SNZB-02WD
- Sonoff SNZB-02DR2

## Testing

The fix is backward compatible - existing numeric value handling remains unchanged.

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -673,6 +673,14 @@ export function getFromLookupByValue(value: unknown, lookup: {[s: string]: unkno
             return key;
         }
     }
+    if (isString(value)) {
+        const valueLower = value.toLowerCase();
+        for (const key of Object.keys(lookup)) {
+            if (key.toLowerCase() === valueLower) {
+                return key;
+            }
+        }
+    }
     if (defaultValue === undefined) {
         throw new Error(`Expected one of: ${Object.values(lookup).join(", ")}, got: '${value}'`);
     }


Koenkk:
It should always use `celsius` and never `Celsius`, this should be fixed after changing it once.

grfernandes54:
👍 This bug is still affecting users on current versions — would it be possible to reopen and rebase this PR? It looks like it was auto-closed by the stale bot rather than for any technical reason.

Confirming it's live on my setup:
- 3× SONOFF **SNZB-02D** (fw 2.3.0)
- Zigbee2MQTT 2.9.2 + Home Assistant OS

Each sensor repeatedly floods the HA log (dozens of times per cycle):
```
Invalid option for select.<name>_temperature_units: 'Celsius' (valid options: ['celsius', 'fahrenheit'])
```

Root cause is exactly as described here and in Koenkk/zigbee2mqtt#27345: since SNZB-02D fw 2.3.0 the device reports `temperature_units` capitalized (`Celsius`/`Fahrenheit`), while the converter's `enumLookup` keys are lowercase (introduced in #9177). The case-insensitive fallback in `getFromLookupByValue()` proposed here is a clean, backward-compatible fix and also covers SNZB-02LD/02WD/02DR2.

Happy to help test on real SNZB-02D hardware if useful. Thanks!

grfernandes54:
Extra context that strengthens the case for fixing this in the converter: the Home Assistant side was explicitly closed as **`wontfix`** (and locked) by the HA MQTT code owners in home-assistant/core#144089 — the MQTT `select` is correct to reject a value outside its declared options, so HA will not add case-insensitive handling. That means the converter is the agreed place to resolve this.

This PR's case-insensitive fallback in `getFromLookupByValue()` (or, alternatively, normalizing the emitted value back to the lowercase enum key) is exactly that fix, and it's backward compatible. The problem was introduced by the lowercase `lookup` change in #9177/#9180 while the SNZB-02D/02LD/02WD/02DR2 firmware reports `temperature_units` capitalized.

Related reports: Koenkk/zigbee2mqtt#27292 (7 👍) and Koenkk/zigbee2mqtt#27345. Would really appreciate reopening/rebasing this. Thanks!

PR #11239, opened 2026-01-08
add new features to Heiman HS1SA-E

Leo2442926161:
Hi Z2M team,
I am adding new features to Heiman HS1SA-E, could you please have an review on it. thanks.

### DIFF
--- a/src/devices/heiman.ts
+++ b/src/devices/heiman.ts
@@ -964,7 +964,7 @@ const fzLocal = {
             };
 
             if (data.smokeConcentrationLevel !== undefined) {
-                const raw = Number.parseFloat(data.smokeConcentrationLevel as string);
+                const raw = Number.parseFloat(data.smokeConcentrationLevel as string) / 100;
                 // Check if valid, then force exactly 2 decimal places as a STRING
                 result.smoke_level = Number.isNaN(raw) ? "0.00" : raw.toFixed(2);
             }


src/devices/heiman.ts
```diff
@@ -640,10 +638,10 @@ const heimanExtend = {
...
         const clusterName = "heimanClusterSpecial" as const;
         const muteStateBitMap = {
-            0: "mute", // bit0
-            1: "alarm_mute", // bit1
-            2: "fault_mute", // bit2
-            3: "low_battery_mute", // bit3
+            0: "muted", // bit0
+            1: "alarm muted", // bit1
+            2: "fault muted", // bit2
+            3: "low battery muted", // bit3
```
Koenkk: Please keep `_` instead of ` `, also above in `faultStateBitMap`

Leo2442926161: Hi @Koenkk ， Thanks for point out, but interestingly, when I put underline there, the entities are exposed with _ to HA, too, that looked like `alarm_muted` instead of `alarm mute`, which looks weird. is there any ways to correct it?

Koenkk: Yes that's expected (and how it's done for all devices)

Leo2442926161: > Yes that's expected (and how it's done for all devices)

thanks, fixed, but still wonder why.

Koenkk:
thanks

PR #11244, opened 2026-01-08
Add scrollwheel functionality to IKEA BILRESA E2490

wildekek:
I added entities for the scroll wheel of the new  IKEA BILRESA E2490 device, since this was missing.
Included bugfix: When the brightness is at max, it returns NaN, so I also created a workaround for this.

Contributes to https://github.com/Koenkk/zigbee2mqtt/issues/30321


Koenkk:
Given this is a standard command, I think we should not make a specific solution for just this device. Hence we should add it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/e4e37daa1e3f85f4e0ac8b02f172b7e387150868/src/converters/fromZigbee.ts#L1400. However I'm not sure if we should expose a separate `brightness` sensor, there is already a `action_level` exposed which includes the requested brightness.

SAGV:
@Koenkk how would `action_level` get to Home Assistant? Doesn't seem to be included in the event…

Koenkk:
Can you test if when an action_level sensor is exposed, it is up-to-date when the MQTT device trigger events comes in, if yes we can expose (I'm a bit worried about race conditions here).

Tomblarom:
I made a temporary workaround using a template. Add this to your `configuration.yaml`:
```
template:
  - triggers:
      - trigger: mqtt
        topic: zigbee2mqtt/IKEA E2490 BILRESA Red
        value_template: "{{ value_json.action }}"
        payload: brightness_move_to_level
    sensor:
      - name: "E2490 Red Lvl Fix"
        unique_id: e2490_red_lvl_fix
        icon: mdi:knob
        state: >
          {% set lvl = trigger.payload_json.action_level | default(none) %}
          {% if lvl is none %}
            255
          {% else %}
            {{ [[lvl | int(255), 1] | max, 255] | min }}
          {% endif %}

  - triggers:
      - trigger: mqtt
        topic: zigbee2mqtt/IKEA E2490 BILRESA Green
        ...
```

In automations, use this `{{ states('sensor.e2490_xxx_lvl_fix') | int(255) }}`. The output is always within `1 - 255` and `null` is replaced with `255`.

andrei-lazarov:
The 255 level workaround is no longer needed. I fixed it in Z2M 2.11.0

Owlbertz:
Hi, any news on the progress of the action level being exposed?

Edit: just noticed the `action_level` is being exposed despite not being listed in the UI.

Tomblarom:
@Koenkk re-open maybe? 👀

PR #11247, opened 2026-01-09
Add Moes ZM4LT1 & ZM4LT4

andrei-lazarov:
Link to picture pull request: 
- https://github.com/Koenkk/zigbee2mqtt.io/pull/4689


src/devices/moes.ts
```diff
@@ -1748,6 +1748,24 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint("TS0001", ["_TZ3000_afgzktgb"]),
```
Koenkk: I propose to add this to https://github.com/Koenkk/zigbee-herdsman-converters/blob/f255201e6594dc0234038ea7b880051f1f041712/src/devices/tuya.ts#L5634 and use the conditional features like https://github.com/Koenkk/zigbee-herdsman-converters/blob/f255201e6594dc0234038ea7b880051f1f041712/src/devices/tuya.ts#L5732 (same for TS0004)

andrei-lazarov: - I feel that defeats the purpose of `moes.ts`, `avatto.ts`, `lonsonho.ts` etc. 
Do you plan to merge them with `tuya.ts`? - I would like this

- The conditional features look messy to me 😕
I would rather make multiple converters, for all realistic feature-combos:
e.g. `TS0004_switch`, `TS0004_switch_with_backlight`, `TS0004_module` like we already have.

Koenkk: Im a bit worried that this creates a lot of duplicate code (while just a single feature is different). Do you see a way to make the logic cleaner?

andrei-lazarov: Moved them to `tuya.ts` with conditionals, as you said

Koenkk:
thanks!

PR #11250, opened 2026-01-09
fix(zigbee2mqtt#30567): ubisysRemoteTemepratureValidDuration is UNIT16

sjorge:
should fix https://github.com/koenkk/zigbee2mqtt/issues/30567

I didn't test this as I am not home but according to the [docs](https://www.ubisys.de/downloads/ubisys-h1-technical-reference.pdf) this should be unint16.


Koenkk:
Thanks!

PR #11251, opened 2026-01-09
Overlooked the external inputs to the NIko 552-721xx devices. Extending the permitted range of the switchAction attribute.

vdkeybus:


Koenkk:
Thanks!

PR #11253, opened 2026-01-09
Add: L&S Lighting 756200643

chris-1243:
It should fix
 - https://github.com/Koenkk/zigbee2mqtt/issues/28547
 - https://github.com/Koenkk/zigbee2mqtt/issues/30532

Link to picture pull request: DONE IN
 - https://github.com/Koenkk/zigbee2mqtt.io/pull/4691

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -1418,7 +1418,6 @@ export function commandsColorCtrl(args: CommandsColorCtrl = {}): ModernExtend {
 
     const fromZigbee = [
         fz.command_move_color_temperature,
-        fz.command_move_color_temperature_stop,
         fz.command_step_color_temperature,
         fz.command_enhanced_move_to_hue_and_saturation,
         fz.command_move_to_hue_and_saturation,


chris-1243:
After reading some debug logs, it will take more work to have this device to work as expected.

By the way, have you ever seen `commandStopMoveStep` in the lightingColorCtrl cluster?

Koenkk:
I see there is no converter for it yet, so no

chris-1243:
Could you check the code submitted and maybe point what I am doing wrong?

I tried to use what is already existing and I modified the code for this use case.

This [converter](https://github.com/chris-1243/Zigbee2mqtt-External-converters/blob/main/LS_remote_test.mjs) is working as it was tested by a users. I am just stuck to convert it in typescript.

src/converters/fromZigbee.ts
```diff
@@ -1515,6 +1515,16 @@ export const command_move_color_temperature: Fz.Converter<"lightingColorCtrl", u
         return payload;
     },
 };
+export const command_move_color_temperature_stop: Fz.Converter<"lightingColorCtrl", undefined, ["commandStopMoveStep"]> = {
+    cluster: "lightingColorCtrl",
+    type: "commandStopMoveStep",
+    convert: (model, msg, publish, options, meta) => {
+        if (hasAlreadyProcessedMessage(msg, model)) return;
+        const payload = {action: postfixWithEndpointName("color_temperature_move_stop", msg, model, meta)};
```
```suggestion
        const payload = {action: postfixWithEndpointName("stop_move_step", msg, model, meta)};
```

This doesn't seem to be specific to color temp (can also stop a color move), from the spec:

Image

chris-1243: I read the spec too quickly. It might be also a good idea to rename the command name like `command_stop_move_step` and to modify the modernExtend commandsColorctrl accordingly.

src/devices/ls.ts
```diff
@@ -1,9 +1,79 @@
...
+            return payload;
+        },
+    },
+};
+
+const lsModernExtend = {
+    groupIdExpose(): ModernExtend {
+        const result: ModernExtend = {
+            exposes: [
+                e.enum("action_group", ea.STATE)
```
Koenkk: Are you using this is in Home Assistant when an `action` is triggered, if yes, can you share the automation yaml? I think we should expose it for every device that exposes `action_group` but I'm a bit worried that it might be unreliable due to race conditions (e.g. `action` sensor updating before `action_group` sensor)

Koenkk:
> Could you check the code submitted and maybe point what I am doing wrong?
> 
> I tried to use what is already existing and I modified the code for this use case.
> 
> This [converter](https://github.com/chris-1243/Zigbee2mqtt-External-converters/blob/main/LS_remote_test.mjs) is working as it was tested by a users. I am just stuck to convert it in typescript.

Fixed the typing!

chris-1243:
> > Could you check the code submitted and maybe point what I am doing wrong?
> > I tried to use what is already existing and I modified the code for this use case.
> > This [converter](https://github.com/chris-1243/Zigbee2mqtt-External-converters/blob/main/LS_remote_test.mjs) is working as it was tested by a users. I am just stuck to convert it in typescript.
> 
> Fixed the typing!

Tkanks!!!!!!!!

src/devices/ls.ts
chris-1243: Well, I do not own this remote. I was just trying to help users to add it.

chris-1243:
Finally!!!

**Koenkk**, many thanks for your help and time. My apologize for all the commits

src/devices/ls.ts
Koenkk: No problem, I removed it for now, if we add it we should add it for all.

PR #11260, opened 2026-01-10
Fix Philips Hue interPAN reset assertions

PaulMcMillan:
## Summary
- Stop setting `zcl.frameType` and `zcl.direction` for the Philips Hue interPAN reset action.

## Why
- zigbee-herdsman 8.x asserts that interPAN ZCL frames must not include explicit frameType/direction; the action currently set both and triggered AssertionError on send.

## Testing
- pnpm test


PaulMcMillan:
(I also verified this change end-to-end on an actual deployment of the -dev container - this fixes the reset)

Koenkk:
Thanks!

PR #11262, opened 2026-01-11
Fix bugs in generateDefinitions()

luar123:
This PR fixes multiple bugs in generateDefinition.ts:

- set endpointNames correctly for light/onoff endpoints: fix https://github.com/Koenkk/zigbee2mqtt/issues/27925
- add endpointName(s) in binary/analog input/outputs only if we have more than 1 endpoint: fix https://github.com/Koenkk/zigbee2mqtt/issues/29888
- Fix name/label/description in binary/analog input/output: Currently devices with names that contain a space would not show up in HA

Probably the same endpoint fix could be applied to `extenderLock` and `extenderElectricityMeter` as well, but I don't have devices to test.

### DIFF
--- a/src/lib/generateDefinition.ts
+++ b/src/lib/generateDefinition.ts
@@ -418,6 +418,291 @@ async function extenderBinaryOutput(device: Zh.Device, endpoints: Zh.Endpoint[])
     return generated;
 }
 
+function getUnitfromApplicationType(applicationType: number): string | undefined {
+    // Map of applicationType to unit strings. Only take type ( Bits 16 to 23 ) into account.
+    const type = (applicationType >> 16) & 0xff;
+    const applicationTypeMap: {[key: number]: string} = {
+        0: "°C", // Temp_Degrees_C
+        1: "%", // Relative_Humidity_Percent
+        2: "Pa", // Pressure_Pascal
+        3: "L/s", // Flow_Liters_Per_Sec
+        4: "%", // Percentage
+        5: "ppm", // Parts_Per_Million
+        6: "rpm", // Rotational_Speed_RPM
+        7: "A", // Current_Amps
+        8: "Hz", // Frequency_Hz
+        9: "W", // Power_Watts
+        10: "kW", // Power_Kilo_Watts
+        11: "kWh", // Energy_Kilo_Watt_Hours
+        12: undefined, // Count
+        13: "kJ/kg", // Enthalpy_KJoules_Per_Kg
+        14: "s", // Time_Seconds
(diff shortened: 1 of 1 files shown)


luar123:
@tomaszduda23: fyi

luar123:
@Koenkk two related questions:
1. are there any dictionaries/enums available with bacnet units (-> engineering_unit attribute) or application type? I would like to map those to unit.
2. Is there a function to test if a standard cluster has a custom attribute with a specific ManufacturerCode?

Koenkk:
> are there any dictionaries/enums available with bacnet units (-> engineering_unit attribute) or application type? I would like to map those to unit.

No, feel free to add it.

> Is there a function to test if a standard cluster has a custom attribute with a specific ManufacturerCode?

You can iterate the cluster attributes and check if there is a `manufacturerCode` present (e.g. https://github.com/Koenkk/zigbee-herdsman/blob/ef9438b8d35407c0002e7eb335971fc3795dd4a5/src/zspec/zcl/definition/cluster.ts#L35). What's the goal of this?

Koenkk:
Thanks!

luar123:
Thanks for merging!

I also want to add support for device classes, but this needs a small modification in homeassistant.ts as well otherwise we loose the custom name again. Not sure when I will look into that.

> You can iterate the cluster attributes and check if there is a manufacturerCode present (e.g. https://github.com/Koenkk/zigbee-herdsman/blob/ef9438b8d35407c0002e7eb335971fc3795dd4a5/src/zspec/zcl/definition/cluster.ts#L35). What's the goal of this?

I want to explore how to add custom names to standard clusters (like it is possible for the analog/binary input clusters) and one option would be to add custom attributes with (or without) a manufacturer code. Main usage is esphome where we create diy devices with arbitrary model/manufacturer IDs and rely on the definition generator. Additionally it would be great to set the friendly_name, but I guess that would require a bigger change, as this is only read from settings right now.

Koenkk:
Maybe we can add some sort of special attribute in the device which contains the config? Are you using https://github.com/luar123/zigbee_esphome ?

luar123:
Yes, and I am trying to bring it into official esphome. Next esphome version will have limited zigbee support for nrf52.

tomaszduda23:
Preview https://deploy-preview-5898--esphome.netlify.app/components/zigbee/

Koenkk:
> Yes, and I am trying to bring it into official esphome. Next esphome version will have limited zigbee support for nrf52.

That would be really amazing!

PR #11263, opened 2026-01-11
fix: mapping of colors

Nerivec:
- fix XY conversion, use CIE 1931 per spec
- add gamut support (with extra support for Philips Hue, and Philips LivingColors)

Philips gradients are not agreeing with previous tests. I don't have any, so, can't confirm what values are right.
@chrivers can you provide any insights?

Cf https://github.com/Nerivec/zigbee2mqtt-windfront/blob/main/src/components/editors/index.ts

### DIFF
--- a/src/lib/color.ts
+++ b/src/lib/color.ts
@@ -70,8 +70,7 @@ export class ColorRGB {
      * @returns new ColoRGB object
      */
     static fromHex(hex: string): ColorRGB {
-        hex = hex.replace("#", "");
-        const bigint = Number.parseInt(hex, 16);
+        const bigint = Number.parseInt(hex.replace("#", ""), 16);
         return new ColorRGB(((bigint >> 16) & 255) / 255, ((bigint >> 8) & 255) / 255, (bigint & 255) / 255);
     }
 
@@ -136,6 +135,7 @@ export class ColorRGB {
 
     /**
      * Convert to CIE
+     * TODO: refactor, this is the Philips Hue formula, not the CIE 1931 gamut
      * @returns color in CIE space
      */
     toXY(): ColorXY {
@@ -243,6 +243,7 @@ export class ColorXY {
 
     /**
      * Converts CIE color space to RGB color space
+     * TODO: refactor, this is the Philips Hue formula, not the CIE 1931 gamut
      * From: https://github.com/usolved/cie-rgb-converter/blob/master/cie_rgb_converter.js
      */
     toRGB(): ColorRGB {
@@ -326,7 +327,7 @@ export class ColorHSV {
      */
     constructor(hue: number, saturation: number = null, value: number = null) {
         /** hue component (0..360) */
-        this.hue = hue === null ? null : hue % 360;
+        this.hue = hue === null ? null : hue === 360 ? hue : hue % 360;
         /** saturation component (0..100) */
         this.saturation = saturation;
         /** value component (0..100) */
--- a/src/lib/color.ts
+++ b/src/lib/color.ts
@@ -1,7 +1,332 @@
 import kelvinToXyLookup from "./kelvinToXy";
 import {clampColorTemp, findColorTempRange} from "./light";
-import type {KeyValue, KeyValueAny, Tz, Zh} from "./types";
+import type {Definition, KeyValue, KeyValueAny, Tz, Zh} from "./types";
 import {precisionRound} from "./utils";
+
+type Vector3 = [number, number, number];
+type Vector2 = [number, number];
+type Matrix3 = [number, number, number, number, number, number, number, number, number];
+
+export interface GammaCorrection {
+    encode: (linear: number) => number; // linear [0..1] -> non-linear
+    decode: (nonLinear: number) => number; // non-linear [0..1] -> linear
+}
+
+export interface Gamut {
+    name: string;
+    red: Vector2;
+    green: Vector2;
+    blue: Vector2;
+    white: Vector2;
+    gammaCorrection: Readonly<GammaCorrection>;
+}
+
+const clamp = (value: number, min: number, max: number): number => {
+    return value <= min ? min : value >= max ? max : value;
+};
+
+const clampMatrixValue = (value: number): number => {
+    return Number.isFinite(value) ? value : 0;
+};
+
+const xyToXyz = ([x, y]: Vector2): Vector3 => {
+    const z = 1 - x - y;
+    const Y = 1;
+
+    return [x / y, Y, z / y];
+};
+
+const invert3x3 = (m: Matrix3): Matrix3 => {
+    const [a, b, c, d, e, f, g, h, i] = m;
+    const A = e * i - f * h;
+    const B = -(d * i - f * g);
+    const C = d * h - e * g;
+    const det = a * A + b * B + c * C;
+
+    if (Math.abs(det) < 1e-9 || !Number.isFinite(det)) {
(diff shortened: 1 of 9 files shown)


Nerivec:
@3reality-support @rohankapoorcom @sjorge @burmistrzak @andrei-lazarov would you be able to test this with some of your devices?
I unfortunately fried my RGB stuff a while back, and haven't gotten around to replace it yet.
I did this from spec, though I'm no color expert, so, if someone can review the logic in details, that would be great!

You need to use MQTT commands directly (bypass any frontend logic), to ensure the payloads you send are unaltered when reaching the device converter.

We need to figure out if the new color conversions are correct, that the devices output correct colors, per given payload:
- [ ] regular devices (spec-compliant)
- [ ] Tuya devices
- [ ] Philips Hue devices
- [ ] Philips LivingColors devices
- [ ] Philips `gradient` devices

In particular should test that all red, green, blue, and white colors are as expected, and for good measure, a couple random colors as well. Should test sending `rgb` format (internally converted to `xy`) and sending `xy` format (should apply as-is).

burmistrzak:
I'll try to setup a dedicated Zigbee hardware test bench for this and ZoH.

I've just migrated my custom _production_ stack to a rather vanilla HAOS setup on a Raspberry Pi 5 with Add-ons, and that's _no good_ for development. 😅

rohankapoorcom:
I don't have any RGB hue stuff connected to Z2M currently. All of my RGB Hue stuff is connected directly to a Hue bridge for ambilight functionality and I don't want to break any of those. Let me dig around in my spare parts bin and see if I can find any Hue RGB in there that I could use for testing.

sjorge:
Nice, when I last looked at it I sort of gave up on the gamma stuff. I was getting way out of my comfort zone with all the color theory and math.

(Let's not mention mired to rgb either)

I have since standardized on hue for all my none WS stuff. I might still have some in my zigbee box somewhere with a random selection of devices. I did give a lot away too though. Will look what I have left next time I visit my moms where the box is.

burmistrzak:
Let's hope my second ZBT-2 is in the post soon and I can finally start testing. Will be the coordinator for my dev setup until ZoH is ready for prime time and I can migrate from the Sonoff stick. 😉

Got a few Hue color bulbs and a single gradient light. No LivingColors tho...

andrei-lazarov:
Finally tested some **Tuya** bulbs to get their gamut.
TLDR: **Rec2020** should do.

- Tuya gateway + app sends directly MoveToHueAndSaturation, can't get X,Y
- If I send MoveToHueAndSaturation (from Z2M dev console), it updates hue and sat attributes, but it never updates currentX and Y !!
- If I send MoveToColor, it updates currentX and Y, but not hue and sat (yet)
- Thankfully when device boots, it always updates hue and saturation 🙂 

**So I can only do X,Y + reboot = H,S**. 

Tested on [Moes E14 CWS](https://www.zigbee2mqtt.io/devices/ZB-TDC6-RCW-E14.html#moes-zb-tdc6-rcw-e14) `TS0505B` `_TZ3210_ifga63rg`
Dev Z2M, confirmed commands are sent correctly by sniffing.

Random safe colors from AI + hs reported by bulb:

```
x: 19661, y: 19661   // x=0.30 y=0.30 (neutral slightly blue white)
165, 72 (hue, saturation)
x: 22937, y: 26214   // x=0.35 y=0.40 (warm yellowish white)
59, 124
x: 16384, y: 32768   // x=0.25 y=0.50 (greenish)
85, 180
x: 29491, y: 22937   // x=0.45 y=0.35 (orange)
11, 162
x: 32768, y: 19661   // x=0.50 y=0.30 (red/orange)
249, 191
x: 13107, y: 9830    // x=0.20 y=0.15 (blue/purple)
173, 225
x: 15073, y: 26214   // x=0.23 y=0.40 (turquoise)
104, 171
x: 26214, y: 16384   // x=0.40 y=0.25 (pink/magenta)
223, 182
```

Max reds from AI + hs reported by bulb:
```
// sRGB / Rec709
x: 41942, y: 21627 // 0.64, 0.33 
4, 246
// Display-P3 / DCI-P3
x: 44564, y: 20971 // 0.68, 0.32
no effect!
// Rec2020
x: 46400, y: 19136 // 0.708 0.292
253, 254
```

Max blues from AI + hs reported by bulb:
```
// sRGB / P3 blue
x: 9830, y: 3932 // 0.15, 0.06
171, 254
// Rec2020 blue
x: 8585, y: 3015 // 0.131, 0.046
169, 254
```

Max greens from AI + hs reported by bulb:
```
// sRGB
x: 19661, y: 39321 // 0.3, 0.6
73, 237
// Display-P3
x: 17367, y: 45219 // 0.265, 0.690
75, 254
// Adobe RGB 
x: 13762, y: 46550 // 0.21, 0.71
80, 249
// Rec2020
x: 11141, y: 52232 // 0.170, 0.797
82, 254
```

Should I do the same for IKEA bulbs (I have TRADFRI and KAJPLATS)?
So I'm not sure what to test next on the colors branch?

I noticed a bug, on both dev and colors: If I send rgb (n,0,0) over mqtt, z2m treats it as rgb (255,0,0). 

Edit: I don't fully understand. **Why do we need gamuts and conversion?**
If user gives XY, send XY to bulb, no conversion.
If user gives RGB, why don't we send HS to bulb? That should be equivalent. What's the purpose of converting it in Z2M to XY?

src/lib/color.ts
```diff
@@ -1,8 +1,333 @@
...
+
+    return {encode, decode};
+};
+
+const GAMMA_CORRECTION_LINEAR: Readonly<GammaCorrection> = {
+    encode: (linear: number): number => clamp(linear, 0, 1),
+    decode: (nonLinear: number): number => clamp(nonLinear, 0, 1),
+};
+
+const GAMMA_CORRECTION_SRGB = makeGammaCorrection({threshold: 0.0031308, slope: 12.92, exponent: 2.4, offset: 0.055});
```
ams2990, line 113: sRGB is weird and the decoding function is not the same as the encoding function. The threshold is at .04045. I don't know if you want to special-case it.

src/lib/color.ts
```diff
@@ -1,8 +1,333 @@
...
+
+/** https://en.wikipedia.org/wiki/CIE_1931_color_space */
+const CIE1931: Readonly<Gamut> = {
+    name: "Zigbee / CIE 1931",
+    red: [0.7347, 0.2653],
+    green: [0.2738, 0.7174],
+    blue: [0.1666, 0.0089],
+    white: [1 / 3, 1 / 3],
```
ams2990, line 121: Why are you using illuminant E? That's an unusual choice. I encounter D65 most often because I work with monitors, but D50 might actually be better here since that's the specified white to be used in the room where the monitor is being viewed.

src/lib/color.ts
```diff
@@ -1,8 +1,333 @@
...
+const WIDE: Readonly<Gamut> = {
+    name: "Wide",
+    red: [0.7347, 0.2653],
+    green: [0.1152, 0.8264],
+    blue: [0.1566, 0.0177],
+    white: [0.3457, 0.3585],
+    gammaCorrection: GAMMA_CORRECTION_SRGB,
+};
+
+export const SUPPORTED_GAMUTS = {
```
ams2990, line 155: Do you also want the sRGB/Rec.709 color space? I haven't traced the application code, but I would think you would need this to convert to/from what monitors can actually display.

PR #11264, opened 2026-01-11
Refactor Sunricher ZG2819S-RGBW and CCT definitions to use extend

chris-1243:
Replaced fromZigbee and exposes with extend for Sunricher ZG2819S-RGBW and CCT, adding command scene `store`. genScene cluster must not be bound to the coordinator to have all scenes working correctly from all endpoints.


Koenkk:
Thanks!

PR #11265, opened 2026-01-11
Aqara WS-K08E / WS-K08D whiteLabel

andrei-lazarov:


src/devices/lumi.ts
```diff
@@ -4528,6 +4528,10 @@ export const definitions: DefinitionWithExtend[] = [
     },
     {
         zigbeeModel: ["lumi.switch.agl010"],
+        whiteLabel: [
+            {model: "WS-K08E", vendor: "Aqara"},
```
Koenkk: Not needed, its the same as the definition.model

andrei-lazarov: Fixed

Koenkk:
thanks!

PR #4710, opened 2022-09-27
Add EDM-1ZBB-EU. https://github.com/Koenkk/zigbee2mqtt.io/pull/1612

joashc:
https://www.moeshouse.com/products/new-zigbee-smart-touch-light-dimmer-switch-timer-brightness-memory-eu

PR #11266, opened 2026-01-12
Adding _TZE284_hodyryli

handfreezer:
Dsiplay sensor with temperature/humidity/time and additional external sensor temperature

Link to picture pull request: (https://github.com/Koenkk/zigbee2mqtt.io/pull/4710)

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -19978,6 +19978,88 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_cgr0rhza"]),
+        model: "TS0601_thermostat_6",
+        vendor: "Tuya",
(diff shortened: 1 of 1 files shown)


handfreezer:
Picture in pull:
Adding image for device _TZE284_hodyryli #4710
https://github.com/Koenkk/zigbee2mqtt.io/pull/4710

Koenkk:
Could you try with this code? Seems similar:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/b8ccb9d526bb79c8768d2daccf5021dd25143065/src/devices/tuya.ts#L23138

handfreezer:
I don't think so as this device exposes the time of the day, and the option is not set in the extend tuyaBase options (timeStart: "1970").

I spotted that I forgot to set dp for battery reporting, I'll check that I forgot no other dp, and update this pull request.

PR #11269, opened 2026-01-12
Update tuya.ts

parihaaraka:
add _TZE200_qasjif9e support due to https://github.com/Andrik45719/ZY-M100: 

> ignore TZE200/TZE204 mismatch is Tuya Zigbee module version only "abracadabra" make sense

 *currently works with the config being changed


Koenkk:
Thanks!

PR #11273, opened 2026-01-12
Fix "gas" not seteable

### LINKED ISSUE #11219: Gas meter energy "gas" not seteable and exposed as "number" in home assistant
Hi,

I would like to raise this as an issue because I would like to get some feedback from the community about how to implement the fix (there are several alternatives)

### Explanation of the current situation.
- The new "zigbee gas meter" has been created as an extension of the `seMeter` as stated in the Zigbee Cluster Library Specification.
- The `currentSummDelivered` attribute is exposed as follows: `e.numeric("gas", ea.ALL).withUnit("m³").withDescription("Total gas consumption in m³")` note, the access is set to `ea.ALL` because this is very convenient:
  - The zigbee counter is going to match an existing counter so it is desirable to SET the value of the counter to match the counter of the real counter.
  - The ESP32 library allows write access to the value of that attribute.
  - If i'm not mistaken, I was able to write to it in December 2025. I'm not sure since when the new validations are in place.

### Problems with the current release of zigbee-herdsman-converters and zigbee-herdsman:

There are two issues:

1. The library raises an error trying to 'write' the value of the 'currentSummDelivered' attribute. In the Zigbee Cluster Library Specification, this attribute is set to 'read' access so the library is consistent with the specification but, as indicated before, the behaviour is not desirable (I would like to set the initial counter value to match the value of the real counter).
2. Due to the reason SET is allowed, the value is exposed under 'number' category in "HA mqtt discovery" rather than under 'sensor' category. This is also problematic because a number has a minimum and maximum value (100) and this will cause warnings later when the counter exceed that value.

### Possible fix directions:

1. Change ea.ALL to ea.STATE_GET. this is fine and is an easy to do fix, but we will lose the functionality that I would like to have in order to set the initial counter value.
2. Implement deeper changes in the code in order to allow "number" to be exposed as "sensor" and, at the same time, allow to "write" to them skipping the validations in place about the zigbee attributes not been writable in the specification.
3. Neither 1 or 2, please expose your point of view

Any preference?

Regards and sorry for the long explanation.

IgnacioHR:
Fixes #11219 losing functionality but at this time it is the simplest approach.

Note, the problem with the "gas" value represented as "number" rather than a "sensor" is that the number max value is set to 100 (default value) and HA ignores any value higher than that so the gas counter is useless as it is now.

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -2231,7 +2231,7 @@ function genericMeter(args: MeterArgs = {}) {
     } else if (args.cluster === "metering" && args.type === "gas") {
         if (args.power !== false)
             exposes.push(e.numeric("volume_flow_rate", ea.STATE_GET).withUnit("m³/h").withDescription("Instantaneous gas flow in m³/h"));
-        if (args.energy !== false) exposes.push(e.numeric("gas", ea.ALL).withUnit("m³").withDescription("Total gas consumption in m³"));
+        if (args.energy !== false) exposes.push(e.numeric("gas", ea.STATE_GET).withUnit("m³").withDescription("Total gas consumption in m³"));
         fromZigbee = [args.fzMetering ?? fz.gas_metering];
         toZigbee = [
             {


Koenkk:
thanks!

PR #11281, opened 2026-01-13
fix(sunricher): prioritize SR-ZG9040A/ZG9041A-D over Vesternet OEM

niracler:
## Summary

Add fingerprint with `manufacturerName: "Sunricher"` and `priority: 1` to ensure SR-ZG9040A/ZG9041A-D devices are correctly identified as Sunricher brand when the device reports manufacturerName as "Sunricher".

## Problem

The Vesternet definition in `vesternet.ts` uses fingerprint matching (modelID + softwareBuildID) which takes precedence over Sunricher's zigbeeModel matching. This causes Sunricher original devices to be incorrectly identified as Vesternet VES-ZB-DIM-004.

## Solution

Add a fingerprint with higher priority (1 vs default 0) to the Sunricher definition, so devices with `manufacturerName: "Sunricher"` will match Sunricher's definition first.

## Related Issue

https://github.com/Koenkk/zigbee2mqtt/issues/29517


Koenkk:
Do we know what manufacturer name is used for the Vesternet devices? (VES-ZB-DIM-004). CC: @martyn-vesternet

chris-1243:
This should also be applied to `SR-ZG9101SAC-HP-SWITCH-2CH` which might be recognized as `VES-ZB-SWI-015` depending on the softwareBuild. It is duplicate code for the same device.

It seems Vesternet uses Sunricher device like some LED-Trading product as well.

Sunricher:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/sunricher.ts#L1456

Vesternet:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/vesternet.ts#L147

Koenkk:
> Do we know what manufacturer name is used for the Vesternet devices? (VES-ZB-DIM-004). CC: @martyn-vesternet

Did not get a reply so will merge this, thanks @niracler 

@chris-1243 could you submit another PR?

chris-1243:
@Koenkk done in
 - Koenkk/zigbee-herdsman-converters/pull/11400

PR #11282, opened 2026-01-13
Add Excellux vendor sensor and fixed the warning setting max value of soil fertility sensor

iotRobot:


Koenkk:
Could you undo the changes to non `tuya.ts` files?

PR #11287, opened 2026-01-13
fix: Fix local temperature calibration range for some devices

Koenkk:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/30448

### DIFF
--- a/src/lib/exposes.ts
+++ b/src/lib/exposes.ts
@@ -635,7 +635,7 @@ export class Climate extends Base {
         return this;
     }
 
-    withLocalTemperatureCalibration(min = -12.8, max = 12.7, step = 0.1, access = a.ALL) {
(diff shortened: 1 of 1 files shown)


src/devices/sonoff.ts
```diff
@@ -2172,6 +2172,7 @@ export const definitions: DefinitionWithExtend[] = [
             tz.thermostat_running_state,
         ],
         extend: [
+            m.customLocalTemperatureCalibrationRange({min: -12.7, max: 12.7}),
```
Nerivec, line 2175: Needs to be spec value (127), or change the extend to multiply by 10.

Koenkk: Multiplied it in the modernExtend now

PR #11288, opened 2026-01-13
Adding  tze284 hodyryli

handfreezer:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/11266

adding battery reporting support


src/devices/tuya.ts
```diff
@@ -23573,4 +23573,35 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: [
```
Koenkk, line 24542: Please merge with TZ-ZT01_GA4 (code is the same).

Also update the `model` to what is written on the device or product page.

handfreezer:
I don't think it is the same code, as for TZ-ZT01_GA4 sending time to device is not activated through timeStart:1970.

Koenkk:
It's fine to add timeStart: 1970 there

handfreezer:
Hello, sorry for the delay.
OK, I'm wondering that the current device in converter won't be annoyed by this decaration, but, how will it work about the picture to différenciate both model of them? (I have to admit that some time the photo of the device do not match and it is quite disturbing)

handfreezer:
Second question (more because I'm newbie on your stack...)  if I change the model string of the device _TZE284_8se38w3c, the current declaration model name (TZ-ZT01_GA4) will be lost?

Koenkk:
I guess you refer to the detection of the device, for that you can use `tuya.whitelabel`, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/e678bc14e6f9d3048f73402624cc73da102aa33f/src/devices/tuya.ts#L3442

handfreezer:
OK, I'll have some job to do so.

PR #11292, opened 2026-01-14
Sonoff MINI-ZBRBS: add calibration support

fbrozovic:


src/devices/sonoff.ts
```diff
@@ -2929,6 +2931,20 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [
             sonoffExtend.addCustomClusterEwelink(),
             m.windowCovering({controls: ["lift"], coverInverted: false}),
+            m.enumLookup<"customClusterEwelink", SonoffEwelink>({
+                name: "motor_travel_calibration_action",
+                lookup: {
+                    "Start Automatic": 2,
```
Koenkk: please make all entries snake_case here

fbrozovic: done, still a bit unsure about the naming...

Koenkk:
Thanks!

Identity-labs:
Hi, 

When i run an automatic calibration

The module do Forward - Stop - Reverse - Stop - Forward - Stop (each step every 1 second) and the shutter don't move

Any idea why ?

PR #11293, opened 2026-01-14
Implement per-channel LED control on Niko devices

vdkeybus:
This commit implements the control of the indicator LEDs on the Niko 552-721X1 and Niko 552-721X2 devices.

LED control earlier was implemented using attribute 0x100, but it appears that this controls all LEDs on a device, and likely is intended for 'ambient light' or 'find in the dark' purposes. Especially on the 721X2 2-channel device, you could not control the LEDs in detached state when they needed to reflect the state of a light managed by another switch. This is now possible.

There appears to be an attribute 0x105 (BITMAP8), which controls both LEDs directly on bits 0 and 1. The single-channel 552-721X1 also supports it, and likely other devices as well.

To control LEDs directly, led_enable must be 'true' and the corresponding led_sync_mode set to 'Off'.

In this patch, led_state is renamed to led_color instead, so be aware of any automations using those. As long as you're using "ON" and "OFF", the behaviour is the same for the single-channel devices.

Tested with both devices, but no others. Notably, the dimmer now also has a led_color state instead of the original led_state. It is very likely that it also supports attribute 0x105 to control its LED,

### DIFF
--- a/src/devices/niko.ts
+++ b/src/devices/niko.ts
@@ -47,12 +47,13 @@ const local = {
                 //       1: Decoupled
                 //       2: Controlled relay
                 //       3+: (invalid value reported)
-                //     NOTE: affects all channels
+                //     NOTES: - affects all channels
+                //            - see warning below on data type
                 //
                 // 0x0100  34 (0x22): UINT24     "outletLedColor" (after reset: 0)
                 // Probably intended for ambient light purposes
                 //
-                // 0x0101  32 (0x20): UINT8    * (after reset: 1)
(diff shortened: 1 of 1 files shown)


vdkeybus:
I'm currently looking at how the switch devices are discovered by Home Assistant (HA). I'm not particularly impressed by the current process. For example, action_reporting ends up like a switch input, whereas 'led_enable' actually correctly ends up in the configuration section of a device.

So, expect some additional work on this, but the patch makes the LED control already available.

Koenkk:
Thanks!

PR #11296, opened 2026-01-14
fix: Aqara W600 position attribute data type

### LINKED ISSUE #11284: Aqara W600 wrong valve position readings
After updating the w600 with the latest converters, the valve position is faulty by factor 0.1

Currently "10" is displayed when the valve is fully opened.

I guess they changed the decimal number of the valve position with this update.

GerjandeGroot:
Fixes #11284 

After an update the Aqara W600 uses a different data type for the valve position attribute.
I don’t know how to maintain backward compatibility with the old firmware.


Koenkk:
Thanks!

PR #11298, opened 2026-01-15
Add zunzunbee slate switch

toaf2022:
Adds support for the Zunzunbee Slate Switch (SSWZ8T), an 8-button battery-powered Zigbee controller.
Decodes short/long button presses from IAS Zone commandStatusChangeNotification (zonestatus)
Exposes 16 actions:

button_1_short_press … button_8_short_press
button_1_long_press  … button_8_long_press

Exposes battery and temperature
Includes best-effort reporting configuration (sleepy device)
Tested with:
1. Zigbee2MQTT 2.x
2.SMLIGHT SLZB-07 (EFR32MG21 / Ember)

Device was previously validated via an external converter; this PR adds native support.
Build and checks pass locally.
This PR is submitted by the device manufacturer.
Device is Zigbee 3.0 compliant and uses standard clusters only.

### DIFF
--- a/src/devices/zunzunbee.ts
+++ b/src/devices/zunzunbee.ts
@@ -15,6 +15,8 @@ const fzLocal = {
                 zoneStatus = msg.data.zoneStatus;
             } else if ("zonestatus" in msg.data) {
                 zoneStatus = msg.data.zonestatus;
+            } else {
+                return;
             }
 
             // Bit0 encodes press type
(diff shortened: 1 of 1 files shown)


Koenkk:
Made some improvements, can you check if all is ok?

note to self: `feat(add): SSWZ8T`

toaf2022:
> Made some improvements, can you check if all is ok?
> 
> note to self: `feat(add): SSWZ8T`

Thanks for the refactor.

I’ve reviewed the updated logic and tested against equivalent converter. Button action decoding works fine.

There are just two small robustness gaps I’d recommend addressing to fully match the earlier tested behavior and avoid edge-case issues:

Guard missing IAS payloads
In the current converter, if neither zoneStatus nor zonestatus exists in msg.data, zoneStatus remains undefined but is still used.
Suggest an early return when the value is not present.

Avoid throwing on unmapped zoneStatus values
utils.getFromLookup(masked, map) may throw if an unexpected masked value is received (e.g. enrollment or non-action IAS frames).
The earlier implementation safely ignored unmapped values. A simple guard would avoid log spam or runtime errors.

Functionally, something like:

const zoneStatus =
    "zoneStatus" in msg.data ? msg.data.zoneStatus :
    "zonestatus" in msg.data ? msg.data.zonestatus :
    undefined;
if (zoneStatus === undefined) return;

const masked = zoneStatus & 0x01fe;
const button = map[masked as keyof typeof map];
if (!button) return;

Happy to push this change myself if you prefer, otherwise I can leave it to you.

Koenkk:
Done, thanks!

PR #11303, opened 2026-01-15
Fixed error with BOX ERC2201-Z: Device has no endpoint 'l1' (or 'l2')

Cykooz:
When a relay "BOX ERC2201-Z" is used from Home Assistant, Z2M notifies about errors:
`Device has no endpoint 'l1'`
or
`Device has no endpoint 'l2'`


Koenkk:
Thanks!

PR #11305, opened 2026-01-16
Add ikea lights scene support

mallardduck:
This PR extends support for scenes with Ikea light devices. It adds a few new sections to the Exposes tab:
Screenshot 2026-01-15 224943

The goal was to mimic more of the functionality that Ikea Home provides when pairing a TRADFRI light with a STYRBAR controller. The default behavior from IKEA allows the left and right buttons to change scenes.

---

This was originally created as an external convertor that seems to work well enough from my testing (for a month or so now). I figured it would be a neat experiment to try and rebuild the convertor into the upstream for all Ikea owners benefit.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://biomejs.dev/schemas/2.3.8/schema.json",
+    "$schema": "https://biomejs.dev/schemas/2.3.11/schema.json",
     "vcs": {
         "enabled": true,
         "clientKind": "git",
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -17,6 +17,7 @@ import {
     ikeaLight,
     ikeaMediaCommands,
     ikeaModernExtend,
+    ikeaScenes,
     ikeaVoc,
     styrbarCommandOn,
     tradfriCommandsLevelCtrl,
@@ -62,14 +63,14 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LED2109G6",
         vendor: "IKEA",
         description: "TRADFRI bulb E26/E27, color/white spectrum, globe, opal, 800/806/810 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, color: {modes: ["xy", "hs"]}}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true, color: {modes: ["xy", "hs"]}}), ikeaScenes(), m.identify()],
     },
     {
         zigbeeModel: ["TRADFRI bulb E27 WS clear 950lm", "TRADFRI bulb E26 WS clear 950lm", "TRADFRI bulb E27 WS\uFFFDclear 950lm"],
         model: "LED1546G12",
         vendor: "IKEA",
         description: "TRADFRI bulb E26/E27, white spectrum, globe, clear, 950 lm",
-        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), m.identify()],
+        extend: [addCustomClusterManuSpecificIkeaUnknown(), ikeaLight({colorTemp: true}), ikeaScenes(), m.identify()],
     },
     {
         zigbeeModel: ["TRADFRI bulb E27 opal 1000lm", "TRADFRI bulb E27 W opal 1000lm"],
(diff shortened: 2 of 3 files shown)


Koenkk:
Thanks for the PR, since this logic is not Ikea specific, I think it should be moved into https://github.com/Koenkk/zigbee-herdsman/pull/1570

andrei-lazarov:
> The goal was to mimic more of the functionality that Ikea Home provides when pairing a TRADFRI light with a STYRBAR controller. The default behavior from IKEA allows the left and right buttons to change scenes.

Hi, in case you didn't know, you can already do that by binding the genScenes cluster STYRBAR -> TRDAFRI.

This looks good though, thanks for adding! I'll test it too.

Are IKEA scenes different than normal scenes? 
I mean can we integrate this into the Scenes tab somehow?

mallardduck:
> in case you didn't know, you can already do that by binding the genScenes cluster STYRBAR -> TRDAFRI.

🤔  I did try that but something didn't work right - even after I populated scenes myself. I will turn off the external convertor I setup and maybe then I can collect some more debug info to help understand what I was seeing and why it wasn't working OOTB.

andrei-lazarov:
It depends on the devices' fw version.
I have the latest STYRBAR+TRADFRI bundle, that you may still find in store. It works with 3 scenes.
Remote comes with 2.4.17 and bulb received an update from Z2M I believe.

If you have an old remote you can update it via Z2M to 2.4.16 and interview to get the genScenes cluster.
Then you can bind it to 2023+ fw bulbs.
I believe the 2021 fw bulbs won't work.

mallardduck:
@andrei-lazarov - all my bulbs seem to be on `1.0.021`, I suspect that is the old version based on a quick search. I can see the Scenes on the bulb in the Scenes tab, but bindingn `genScenes` from the remote to the bulb doesn't cause changes.

```
[1/16/2026, 3:54:51 PM] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Moon Lamp Control', payload '{"action":"arrow_right_click","battery":100,"identify":null,"linkquality":216,"update":{"installed_version":33816598,"latest_version":33816598,"state":"idle"}}'
[1/16/2026, 3:54:51 PM] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Moon Lamp Control/action', payload 'arrow_right_click'
[1/16/2026, 3:54:51 PM] z2m: Publish 'set' 'scene_recall' to 'Moon Lamp' failed: 'Error: Value is not a number, got string (2)'
```

🤔  So in hindsight (as a relative noob to Z2M/HA) it seems like the heavy lifting of my external converter was `parseInt`.

andrei-lazarov:
Ok, I did some testing:

- STYRBAR left/right sends custom IKEA commands to genScenes.
  - IKEA Bulbs should interpret them as: 
  Cycle between scenes with groupID 65289
  If there are no scenes with that groupID, create them: 0 neutral, 1 cold, 2 warm
  - Other manufacturers don't recognize them
  - It's likely old IKEA bulbs with 1.0.x fw don't recognize them either
 
- I have 4 modern bulbs with 2024 (3.0.x) firmware that work like this. 
  - Bind genScenes and it works automatically with the standard scenes
  - I can add more scenes and edit the standard ones with dev console. 
(because the current Scenes tab UI doesn't have a group field)
  - The bulbs must be added to group 65289, otherwise 'store scene' fails.

- I have one older bulb with 2019 firmware (2.0.022), but it has update available on Z2M
  - It doesn't create the standard scenes automatically, so no initial effect
  - It works after I create the scenes
  
Here are more data points: 
- https://github.com/Koenkk/zigbee2mqtt/issues/29456

So back to the PR, sorry, I don't understand:
How were you planning to use the scenes, if your bulbs don't respond to the remote? Automations?
Then why bring them to Exposes tab, instead of using the Scene tab normally? 🤔

mallardduck:
🤦 Yeah this is what I get for building something that "worked" and not documenting it. 

> How were you planning to use the scenes, if your bulbs don't respond to the remote? Automations?

~Pulling it all apart - I found that I do have a HA automation (a blueprint) setup that handles responding to arrow clicks. And that a bug in that automation was the source of the flaw that required the converter. Which explains this error:~

> [1/16/2026, 3:54:51 PM] z2m: Publish 'set' 'scene_recall' to 'Moon Lamp' failed: 'Error: Value is not a number, got string (2)'

~And why my initial converter that this PR was based on was needed at all. My bad again for the confusion here - I completely forgot about the HA level changes I made when I was setting this up since I tried a bunch of different methods.~

OK I tried everything again with only HA automation setup and it was showing the bug above (got string not number). I fixed that in the automation and the left and right was only able to swap between 2 scenes. After adding back the external-converter all the scenes were able to rotate through.

Here is the external converter I am using to make that work which the PR was attempting to replicate (potentially poorly): https://gist.github.com/mallardduck/a48d8a0245d7e74f651a0599aeb8f8bb The gist of the goal being that if Z2M exports sensors for both `current_scene` and `scene_count` then the HA automation can send correct update commands to rotate thru scenes.

andrei-lazarov:
I see. Having `current_scene` and `scene_count` sounds useful.
Indeed this applies to all devices, not just IKEA blubs.

Currently you can read these values and push the output to a custom property in State, with an MQTT command (see dev console -> read attributes).
Then HA should be able to read that state.
But exposing them with your converter is definitely easier to use.

Ask Nerivec what he thinks, he is reworking scenes.

PR #11313, opened 2026-01-17
fix: add adaptation_run_status lost_in_progress

swnf:
The Danfoss Ally seems to report the adaption run status 5 if a new run was triggered after the characteristics have been lost:

```
[17.1.2026, 14:19:07] zhc:tz: Read result of 'hvacThermostat': {"danfossAdaptionRunStatus":5}
```
I've added `lost_in_progress` as a new possible value. It seems like the field is a bit mask, see slide 10 [here](https://assets.danfoss.com/documents/latest/367874/AU417130778872en-000102.pdf).


Koenkk:
Thanks

PR #11317, opened 2026-01-17
Fix Lutron Z3-1BRL (Aurora) actions

pikeas:
Switch now sends `on_press | off_press | rotate`:

```
{"action":"rotate","battery":100,"brightness":205,"linkquality":91,"update": ...}
{"action":"off_press","action_group":8093,"battery":100,"brightness":197,"linkquality":91,"update": ...}
{"action":"on_press","action_group":8093,"battery":100,"brightness":197,"linkquality":91,"update": ...}
```

We could drop `brightness` from the press payload if that's more consistent with other devices.

Separately, we can distinguish off/on based on what the switch reports, but...should we? Or should we combine them into a stateless `press`? Eg, if a light changes state because of an outside service call, it becomes out of sync with this switch. This can easily be handled at the automation level, but perhaps it should be handled here.

Related: https://github.com/Koenkk/zigbee2mqtt/issues/28450, https://github.com/Koenkk/zigbee2mqtt/issues/25538, https://github.com/Koenkk/zigbee2mqtt/issues/12291

### DIFF
--- a/src/devices/lutron.ts
+++ b/src/devices/lutron.ts
@@ -28,7 +28,7 @@ const fzLocal = {
             addActionGroup(payload, msg, model);
             return payload;
         },
-    } satisfies Fz.Converter,
+    } satisfies Fz.Converter<"genLevelCtrl", undefined, ["commandMoveToLevel", "commandMoveToLevelWithOnOff"]>,
 };
 
 export const definitions: DefinitionWithExtend[] = [
--- a/src/devices/lutron.ts
+++ b/src/devices/lutron.ts
@@ -2,23 +2,33 @@ import * as fz from "../converters/fromZigbee";
 import * as exposes from "../lib/exposes";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
-import type {DefinitionWithExtend, Fz} from "../lib/types";
+import type {DefinitionWithExtend, Fz, KeyValue} from "../lib/types";
+import {addActionGroup, hasAlreadyProcessedMessage, postfixWithEndpointName} from "../lib/utils";
 
 const e = exposes.presets;
-const ea = exposes.access;
 
 const fzLocal = {
-    custom_command_move_to_level: {
+    aurora_dimmer: {
         cluster: "genLevelCtrl",
         type: ["commandMoveToLevel", "commandMoveToLevelWithOnOff"],
-        options: [exposes.options.simulated_brightness()],
         convert: (model, msg, publish, options, meta) => {
-            // Map null to 255
-            // https://github.com/Koenkk/zigbee2mqtt/issues/28450
-            const newMsg = {...msg, data: {...msg.data, level: Number.isNaN(msg.data.level) ? 255 : msg.data.level}};
-            return fz.command_move_to_level.convert(model, newMsg, publish, options, meta);
+            if (hasAlreadyProcessedMessage(msg, model)) return;
+
+            // device reports 255 -> herdsman treats as invalid -> converter sees NaN -> map to 254
+            const level = Number.isNaN(msg.data.level) ? 254 : msg.data.level;
+
+            const payload: KeyValue = {};
+            if (msg.data.transtime === 7) {
+                const actionName = level === 0 ? "off_press" : "on_press";
+                payload.action = postfixWithEndpointName(actionName, msg, model, meta);
+            } else {
+                payload.action = postfixWithEndpointName("rotate", msg, model, meta);
+                payload.brightness = level; // note: device min level is 2, so the range is 2-254
+            }
+            addActionGroup(payload, msg, model);
+            return payload;
         },
-    } satisfies Fz.Converter<"genLevelCtrl", undefined, ["commandMoveToLevel", "commandMoveToLevelWithOnOff"]>,
+    } satisfies Fz.Converter,
 };
 
 export const definitions: DefinitionWithExtend[] = [
@@ -27,7 +37,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "LZL4BWHL01",
         vendor: "Lutron",
         description: "Connected bulb remote control",
-        fromZigbee: [fz.command_step, fz.command_step, fzLocal.custom_command_move_to_level, fz.command_stop],
+        fromZigbee: [fz.command_step, fz.command_step, fz.command_move_to_level, fz.command_stop],
         toZigbee: [],
         exposes: [e.action(["brightness_step_down", "brightness_step_up", "brightness_stop", "brightness_move_to_level"])],
     },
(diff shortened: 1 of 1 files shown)


pikeas:
After making this PR, I realized that for my own use, I prefer the actions to be `press | rotate_left | rotate_right`, because switch state isn't coupled to light state. This aligns with how other switches behave but would be very different from how the current implementation works.

Koenkk:
This is a breaking change, so will merge this with z2m 3.0

MangoScango:
First, thank you for this. These dimmers have been driving me crazy and this is a good step to having them behave normally. However there are other strange behaviors these dimmers have that this still leaves unaddressed.

The first I think is an easy fix. I've observed that if the internal state of the dimmer has the target brightness low enough, pressing the button does not send the transition time of 7 like it normally does, and instead sets the level to 0 with the transition time of a rotation. This means with this converter, if you rotate the dial to the left far enough, then press the button, the button press will be interpreted as a rotation event with brightness 0. Not a huge issue as most implementations will probably use this event to turn off a light either way, but I think we should fire off_press anytime the level is being set to 0, since the dial rotation never goes below 2.

The second is trickier. I've noticed that if these dimmers are idle for long enough, they will not fire the commandMoveToLevelWithOnOff event at all unless you interact with them twice. Instead they will try to read the state of their bound device, presumably to make sure its properly in sync before firing the event. This means that if you are trying to use them without binding, you will occasionally have to press the button twice, or rotate it multiple times before the event we are looking at fires. Take a look at these logs for these two dimmers that I interacted with for the first time today.

Here I press the button, but no commandMoveToLevelWithOnOff event fires. Other routine messages come through fine.
```
[2026-03-08 09:25:09] debug:    z2m: Received Zigbee message from 'Kitchen Aurora - Family Room', type 'read', cluster 'genOnOff', data '["onOff"]' from endpoint 1 with groupID 0
[2026-03-08 09:25:12] debug:    z2m: Received Zigbee message from 'Kitchen Aurora - Family Room', type 'commandQueryNextImageRequest', cluster 'genOta', data '{"fieldControl":0,"fileVersion":3080,"imageType":0,"manufacturerCode":4420}' from endpoint 1 with groupID 0
[2026-03-08 09:25:18] debug:    z2m: Received Zigbee message from 'Kitchen Aurora - Family Room', type 'attributeReport', cluster 'genPowerCfg', data '{"batteryPercentageRemaining":200}' from endpoint 1 with groupID 0
```

Here we see the same behavior with rotation. This time I go back to the dimmer and rotate it a second time, and the event then fires. 
```
[2026-03-08 09:28:15] debug:    z2m: Received Zigbee message from 'Living Room Aurora', type 'read', cluster 'genLevelCtrl', data '["currentLevel"]' from endpoint 1 with groupID 0
[2026-03-08 09:28:18] debug:    z2m: Received Zigbee message from 'Living Room Aurora', type 'commandQueryNextImageRequest', cluster 'genOta', data '{"fieldControl":0,"fileVersion":3090,"imageType":0,"manufacturerCode":4420}' from endpoint 1 with groupID 0
[2026-03-08 09:28:24] debug:    z2m: Received Zigbee message from 'Living Room Aurora', type 'attributeReport', cluster 'genPowerCfg', data '{"batteryPercentageRemaining":200}' from endpoint 1 with groupID 0
[2026-03-08 09:29:35] debug:    z2m: Received Zigbee message from 'Living Room Aurora', type 'read', cluster 'genLevelCtrl', data '["currentLevel"]' from endpoint 1 with groupID 0
[2026-03-08 09:29:35] debug:    z2m: Received Zigbee message from 'Living Room Aurora', type 'commandMoveToLevelWithOnOff', cluster 'genLevelCtrl', data '{"level":168,"transtime":2}' from endpoint 1 with groupID 0
```

My thinking is maybe we can fire button press events when the dimmer attempts to read onOff, but only if it's been idle for a long time. I have no idea how long that timer needs to be though, and there is no indication of whether or not this is intended to be on or off event. You can do a similar thing with firing rotation events when the device attempts to read currentLevel, but again we can't determine which direction the rotation was. For on/off events maybe we just collapse them into a single press event, but there isn't a lot of use for rotation events if there is no information on the direction.

Another thought, and I just don't know enough about zigbee to know if this even makes sense, but is it possible for us to give it a different answer when it tries to read onOff/currentLevel? I'm thinking maybe it's not getting the answer it expects and then failing to send the commandMoveToLevelWithOnOff message as a result. Maybe if we respond differently to those reads it will fire the commandMoveToLevelWithOnOff event in these circumstances.

MangoScango:
After spending some more time today debugging that second issue, it seems like the Lutron is actually sending a second message after the read, it's just getting discarded before z2m can parse it. For whatever reason, despite being bound to the coordinator, theses things always want to communicate via a totally random multicast group. The first time the device does this, the coordinator doesn't know what to do with the message, and drops it. It then adds that group to the multicast table, so subsequent messages are passed on correctly. The group it sends the messages to seems to be stable, but random per device. I've been trying to figure out some way to prime these additions to the multicast table on startup so that the first messages aren't lost, but nothing seems to work. I thought manually making a group with the the lutron was talking to and assigning the coordinator to it would work, but that seems to get ignored.

Here are logs if anyone has any ideas. I press the button for the first time at 2026-03-08 13:32:23, then for the second time at 2026-03-08 13:32:28.
[local_logs.log](https://github.com/user-attachments/files/25826821/local_logs.log)

MangoScango:
Following up on this after doing some more research.

I was able to come up with a hack for the "multicast priming" idea. By creating groups with the id the Lutron is expecting, and adding a real device to that group (preferably not something that listens to endpoint 11 so that it will ignore the multicast messages from the Lutron). Then using a HomeAssistant automation to get the state of that group will force the coordinator to register the group in the multicast table and start listening to messages sent there. The biggest problem with this approach though is that this can double (or worse) the number of entries in the multicast table. I was testing the official sonoff firmware 8.0.2 for the Dongle E and I just ran out of entries in the table. The community firmware builds all seem to have settled on 26 entries to the table, which would be enough for my use case at present, but its clearly not a solution that scales well.

So I went back to the idea of listening to the read messages, and came up with this external converter [lutron_fix.js](https://github.com/user-attachments/files/25884507/lutron_fix.js). I found that whether due to network conditions or some other strange behavior with the Lutron, rarely a press would not result a genOnOff read, though it almost always did. Therefore, I'm listening to both reads to genOnOff and commandMoveToLevel to send a single 'press' event, waiting 300ms between press events for debouncing. This of course cant solve for the _initial_ rotation command which is sent to the multicast group in some cases. However subsequent commands are sent unicast to the controller, so this should mostly still work even if the multicast table fills up.

Additionally, I've now tested firmware 3.4, 3.8, and 3.9. Every version seems to have the same strange hardcoded group behavior. Doesn't seem like there is a way to avoid dealing with it.

PR #11326, opened 2026-01-18
fix(detect): Detect `_TZ3210_tqlv4ug4` as Tuya TS0001_switch_module

andrei-lazarov:
Picture is correct

- Fixes https://github.com/Koenkk/zigbee2mqtt/issues/30299


Koenkk:
Thanks!

PR #11335, opened 2026-01-19
Excellux/patch 2026011901

iotRobot:
update model name of Excellux sensors


Koenkk:
Thanks, please also submit a pr to rename the images in https://github.com/Koenkk/zigbee2mqtt.io/

Issue #1617, opened 2020-10-01
Adding newly developed device to Z2M?


qvntra:
Hello;

Big supporter (sponsor @Koenkk) and endorser of the Z2M project.

We have developed a new type of sensor, ballistocardiography, that leverages ZigBee to transmit payloads via Meter Identification cluster in ZCL. We are very close to having our sensor communicate with Z2M coordinator but experience the following error:

**log.txt**
```
> zigbee2mqtt@1.14.4 start /opt/zigbee2mqtt
> node index.js

Zigbee2MQTT:info  2020-10-01 00:45:15: Logging to console and directory: '/opt/zigbee2mqtt/data/log/2020-10-01.00-45-14' filename: log.txt
Zigbee2MQTT:debug 2020-10-01 00:45:15: Removing old log directory '/opt/zigbee2mqtt/data/log/2020-10-01.00-04-44'
Zigbee2MQTT:debug 2020-10-01 00:45:15: Loaded state from file /opt/zigbee2mqtt/data/state.json
Zigbee2MQTT:info  2020-10-01 00:45:15: Starting Zigbee2MQTT version 1.14.4 (commit #0cd4607)
Zigbee2MQTT:info  2020-10-01 00:45:15: Starting zigbee-herdsman...
Zigbee2MQTT:debug 2020-10-01 00:45:15: Using zigbee-herdsman with settings: '{"adapter":{"concurrent":null},"backupPath":"/opt/zigbee2mqtt/data/coordinator_backup.json","databaseBackupPath":"/opt/zigbee2mqtt/data/database.db.backup","databasePath":"/opt/zigbee2mqtt/data/database.db","network":{"channelList":[11],"extendedPanID":[221,221,221,221,221,221,221,221],"networkKey":"HIDDEN","panID":6754},"serialPort":{"path":"/dev/ttyACM0"}}'
Zigbee2MQTT:info  2020-10-01 00:45:20: zigbee-herdsman started
Zigbee2MQTT:info  2020-10-01 00:45:20: Coordinator firmware version: '{"meta":{"maintrel":1,"majorrel":2,"minorrel":7,"product":1,"revision":20200417,"transportrev":2},"type":"zStack3x0"}'
Zigbee2MQTT:debug 2020-10-01 00:45:20: Zigbee network parameters: {"channel":11,"extendedPanID":"0xdddddddddddddddd","panID":6754}
[...]

PR #11337, opened 2026-01-19
Use standard light converter for Schneider WDE002386 dimmer

mchacher:
The Schneider Electric WDE002386 dimmer (NHPB/DIMMER/1) implements strict ZCL compliance and previously required a custom converter. With the zigbee-herdsman fix that treats optionsMask=0 and optionsOverride=0 as parameter omission, the standard m.light() converter now works correctly.

Changes:
- Removed custom tzLocal.light_brightness_move_compat converter
- Updated WDE002386 device definition to use m.light() with standard options
- Added comment explaining dependency on zigbee-herdsman PR #1617
- All tests passing (326 tests)

The device now uses standard converters for full compatibility while maintaining strict ZCL compliance through the herdsman layer.

Depends-on: https://github.com/Koenkk/zigbee-herdsman/pull/1617
Fixes: https://github.com/Koenkk/zigbee2mqtt/issues/30701


mchacher:
## Update: Custom Converter for Strict ZCL Compliance

I've updated this PR with a proper device-specific solution as suggested in the feedback.

### What Changed

**Previous approach:**
- Used standard `m.light()` converter
- Relied on global changes in zigbee-herdsman

**New approach:**
- **Custom converter** `schneiderLevelControl` specifically for this device
- Sends **only** required parameters: `level` and `transtime`
- Omits optional parameters: `optionsMask` and `optionsOverride`

### Implementation

Added custom converter in `src/devices/schneider_electric.ts`:

```typescript
const schneiderLevelControl: Tz.Converter = {
    key: ["brightness", "brightness_percent", "level"],
        // Convert input to level (0-254 range)
        let level;
        if (key === "brightness_percent") {
            level = Math.round((Number(value) * 254) / 100);
        } else {
            level = Number(value);
        }
        
        // Clamp to valid Zigbee range (0-254, not 255)
        level = Math.min(254, Math.max(0, level));
        
        const transitionTime = utils.getTransition(entity, key, meta).time;
        
        // Send ONLY required parameters - no optionsMask or optionsOverride
        const payload = {
            level: level,
            transtime: transitionTime,
        };
        
        await entity.command("genLevelCtrl", "moveToLevelWithOnOff", payload, 
                            utils.getOptions(meta.mapped, entity));
        return {state: {brightness: level}};
    },
};
```

Updated device definition to use this custom converter instead of `m.light()`.

### Why This Device Needs Special Handling

The Schneider Electric NHPB/DIMMER/1 implements **strict ZCL compliance**:
- It validates that incoming commands contain ONLY parameters it supports
- Rejects commands with unsupported optional parameters with `MALFORMED_COMMAND` (0x80)
- Most Zigbee devices are tolerant and ignore unknown parameters, but Schneider is not

### Architecture: Two-Layer Approach

This PR implements the **device-specific layer** of a clean two-layer solution:

1. **This PR (zigbee-herdsman-converters)**: Device-specific logic
   - Custom converter controls what parameters to send
   - Keeps device quirks isolated in the converter
   
2. **Companion PR (zigbee-herdsman #1617)**: Generic infrastructure
   - Allows converters to omit conditional parameters
   - Benefits ANY strict ZCL device

### Additional Fix

Also added brightness clamping to valid Zigbee range (0-254):
- Prevents accidentally sending brightness=255 (invalid per spec)
- Ensures commands always succeed

### Testing

Tested with Schneider Electric NHPB/DIMMER/1 (firmware 001.005.000 R):

**Commands tested:**
```bash
# Brightness value (0-254)
mosquitto_pub -t "zigbee2mqtt/WiserOdaceDimmer_00/set" -m '{"brightness": 127}'

# Brightness percentage (0-100)
mosquitto_pub -t "zigbee2mqtt/WiserOdaceDimmer_00/set" -m '{"brightness_percent": 50}'

# Edge cases
mosquitto_pub -t "zigbee2mqtt/WiserOdaceDimmer_00/set" -m '{"brightness": 254}'  # Max valid
mosquitto_pub -t "zigbee2mqtt/WiserOdaceDimmer_00/set" -m '{"brightness": 255}'  # Clamped to 254
```

**Results:**
- ✅ All brightness levels work correctly (0-254)
- ✅ Brightness correctly clamped when needed (255→254)
- ✅ No MALFORMED_COMMAND errors
- ✅ Device responds with statusCode 0 (success)
- ✅ Device reports correct currentLevel

### Benefits

✅ Device-specific logic isolated in converter (clean architecture)  
✅ No configuration needed in zigbee2mqtt  
✅ Works out-of-the-box for users  
✅ Future-proof: infrastructure supports other strict ZCL devices  

### Companion PR

Works together with: Koenkk/zigbee-herdsman#1617

PR #11322, opened 2026-01-18
Add Nolia desk lamp to Müller Licht devices

JMyrng:
Add model 404117

Link to picture pull request: [pull request](https://github.com/Koenkk/zigbee2mqtt.io/pull/4736)

PR #11343, opened 2026-01-20
Müller Licht: 404135, 404136, 404137 - adding model numbers to the same figerprint.

DennisGaida:
So this is just an update to an existing PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/11322. My new lamp (https://www.mueller-licht.de/produktinformationen/artikel/?q=404136) has the same fingerprint as this old PR.

I believe it is the same lamp, just the new version. The old lamp was "Nolia", the new lamp is "Dalia". You can't find any reference on the manufacturers website of Nolia and I can only find articles about it from 2023/2024.

Anyways the new lamp identifies as the same thing: "Desk lamp". For the three available colors they have three model IDs: 404135, 404136, 404137.

Since the currently autogenerated definition for this lamp works and looks exactly the same as the previous PR I am just adding the model IDs for future reference.

The image for **this** device (404136, the white version) is here: https://github.com/Koenkk/zigbee2mqtt.io/pull/4749


src/devices/muller_licht.ts
```diff
@@ -310,7 +310,7 @@ export const definitions: DefinitionWithExtend[] = [
     },
     {
         fingerprint: [{manufacturerName: "MLI", modelID: "Desk lamp"}],
-        model: "404117",
+        model: "404117/404135/404136/404137",
```
andrei-lazarov: Follow-up from Discord, something like this.
You can also add descriptions if you know the color/design for each model.
Then also rename the picture to `4041xx.png`

```suggestion
        model: "4041xx",
        whiteLabel: [
            {model: "404117", vendor: "Müller Licht"},
            {model: "404135", vendor: "Müller Licht"},
            {model: "404136", vendor: "Müller Licht"},
            {model: "404137", vendor: "Müller Licht"},
        ]
```

andrei-lazarov: Actually you can add pics for all. 
Z2M will show the generic 4041xx picture, but on Supported devices you will see all of them: [example](https://www.zigbee2mqtt.io/supported-devices/#s=Type%20-%20)

DennisGaida: This is awesome! Will add all of that.

DennisGaida:
Added whitelabel & descriptions with colors to all known models

All new images here: https://github.com/Koenkk/zigbee2mqtt.io/pull/4751

Koenkk:
Thanks!

Issue #11344, opened 2026-01-20
ModernExtend does not respect limited color mode capabilites for light endpoints


xLerate:
If a device only supports x/y color mode and not h/s color mode and i want to exclude this in a external converter via `color`-attribute (using `lumiLight` just as example its extend modernExtends `light`):
```
lumiModernExtend.lumiLight({
            colorTemp: true,
            deviceTemperature: false,
            powerOutageCount: false,
            color: {modes: ["xy"]},
            endpointNames: ["rgb"],
}),
```
the associated `meta`-attribute `supportsHueAndSaturation` is not set to `false` explicitly. (It is only set to `true` if `hs` is found in the color modes array (see lib/modernExtend.ts line 1233 ff.).
This leads the "light_color" converter (convertes/toZigbee.ts line 63 ff.) suppose that color mode h/s is supported and tries to send this command, which ends up in a "UNSUPPORTED" message from the device instead of throwing an exception.

There are two options to improve this:
* setting those attributes explicitly (see #11347) in modernExtend.ts
* handle undefinded values in `light_color`-converter

xLerate:
i just reviewed this (after noticing all the failed tests). i think the root cause is, that `supportsHueAndSaturation` defaults to `true` in the light_color converter (toZigbee.ts, line 70), but should be `false` by default (according to the current implemention). Otherwise there would be no way to set this value to `false`.

xLerate:
i created a PR #11364 to fix this. i would be happy to discuss this.

xLerate:
@Koenkk  Issue is still there. May be with some support, we can get this done. PR is still alive #11364. Please reopen.

xLerate:
Associated PR #11364 was integrated into #11840 by @andrei-lazarov  (thx) and finally merged into master. So this issue is done.

PR #11347, opened 2026-01-20
#11344: setting meta attributes for light endpoints explicitly

xLerate:


Koenkk:
> handle undefinded values in light_color-converter

From https://github.com/Koenkk/zigbee-herdsman-converters/issues/11344, I'm wondering if we should go for this option (because if it's not defined it should fallback to a default value)

PR #11350, opened 2026-01-21
feat: Add Gradient effects 929004276602 and improve definition Sunricher ZG2858A

chris-1243:
Add Gradient effects 929004276602
 - improve: Koenkk/zigbee2mqtt#30691)

Add unimplemented function for Sunricher ZG2858A (Home button was not sending any commands). I mainly based the change from this existing code to add the missing scene commands. As it is specific to this remote, I did not modify fromZigbee.ts.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/4e2589d6ef13abfff3bc4c90a34019b8135795fc/src/converters/fromZigbee.ts#L3346-L3353

### DIFF
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -4890,6 +4890,12 @@ export const definitions: DefinitionWithExtend[] = [
                 fingerprint: [{modelID: "929004276802"}],
             },
         ],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11355, opened 2026-01-22
Update Novato ZİS-01P

bilgi-source:
bug fixes and chip changes

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -23582,7 +23582,7 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
 
-       {
+    {
         fingerprint: tuya.fingerprint("TS0601", ["_TZE284_vceqncho", "_TZE284_who1jxwd"]),
         model: "ZIS-01P",
         vendor: "Novato",


src/devices/tuya.ts
```diff
@@ -23583,38 +23583,45 @@ export const definitions: DefinitionWithExtend[] = [
...
-        description: "Dual-tech presence sensor (PIR + radar)",
-        extend: [tuya.modernExtend.tuyaBase({dp: true})],
+        fingerprint: [{modelID: "TS0601", manufacturerName: "_TZE284_who1jxwd"}],
+        model: "TS0601",
+        vendor: "Tuya",
+        description: "Dual-tech presence sensor (PIR + mmWave Radar)",
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [tuya.tz.datapoints],
+        configure: tuya.configureMagicPacket,
         exposes: [
```
Koenkk, line 23591: Please keep everything above this line as is.

Koenkk:
Thanks!

bilgi-source:
I noticed a significant discrepancy where the datapoints in official version start at 101 (presence_distance), whereas in my code they start at 102, causing a shift that makes all subsequent datapoints incorrect as well. 

Additionally, the presence_distance exposes raw values (1, 2, 3) with the unit 'm', which is misleading since these values represent sensitivity levels (e.g., 1 = ~2m) rather than literal meters; this should ideally be handled as an enum or mapped value to avoid showing incorrect units in the UI.

resim
resim

PR #11360, opened 2026-01-23
fix: ADEO lights turn off at brightness 1

borreeero:
I have several ADEO/ENKI/LEXMAN lights, models ZBEK-3 and ZBEK-12, and they turn off at brightness 1. I imagine all their ZBEK/LXEK lights will behave in the same way, so I'm enabling the workaround for them.

I had to modify modernExtend `light({color: true})` test to use another device. I've chosen AruduroSmart AD-E14RGBW3001 at it looked similar to ZBEK-1, I don't know if it's the best one.


Koenkk:
Thanks!

PR #11362, opened 2026-01-23
Improve: Philips Omniglow Lighstrip definition

chris-1243:
Updated the 3m and 5m version to use the 10m definition which is more complete. Also updated descriptions and move all versions as whitelabel to reduce code.


Koenkk:
Thanks!

PR #11364, opened 2026-01-23
#11344: fixes issue with xy-color mode only devices

xLerate:
* changed default for hsv-meta attributes to false inside light-color converter (adresses #11344)
* added hsv->xy color conversion to support devices with xy color mode only, when sent hsv color

The PR looks bigger than it actually is. I had to swap some code blocks.  

FYI: converting colors here is useful when Zigbee2MQTT is connected to [Matterbridge-Zigbee2MQTT](https://github.com/Luligu/matterbridge-zigbee2mqtt) for example, because most Matter controllers (Alexa, Google Home) primarily send HSV colors.


Koenkk:
Could you add a test case for this to https://github.com/Koenkk/zigbee-herdsman-converters/blob/3045cfb0fa9d32303066d112df4053688c180dba/test/toZigbee.test.ts#L368?

xLerate:
added two tests for hsv color sent to xy color only device 
* xy color mode forced by `color:  { mode: ["xy"] }`
* xy color mode given by `color: true` (which also defaults to xy color only)
for both tests the incoming hsv color has to be converted to xy color and sent to the device via `moveToColor` command.

tests are added in own block (-> "light_color") inside `toZigbee.test.ts`

Koenkk:
I checked this against the z2m tests cases but it makes https://github.com/Koenkk/zigbee2mqtt/blob/3a49c95786c2fba749e7696aab4cc38e467d2c4c/test/extensions/publish.test.ts#L1322 fail:

Screenshot 2026-01-27 at 20 40 33

This test runs against the Philips 7146060PH which uses the default args so it should support `ehancedHue`: https://github.com/Koenkk/zigbee-herdsman-converters/blob/e91d0913aca84f1a4595326cc031a0c073fff40f/src/lib/modernExtend.ts#L1190

Could you check?

xLerate:
Are `supportsHueAndSaturation` and `supportsEnhancedHue`  independent of each other. Or does `supportsEnhancedHue` implies `supportsHueAndSaturation`, meaning `supportsEnhancedHue` is as a special case of `supportsHueAndSaturation` ?

Koenkk:
> Or does supportsEnhancedHue implies supportsHueAndSaturation, meaning supportsEnhancedHue is as a special case of supportsHueAndSaturation ?

Yes, when `supportsEnhancedHue = true`, then `supportsHueAndSaturation` must also be `true`.

andrei-lazarov:
Good catch. We should get this merged. So it solves this error: 
- before: Bulb without HS support receives HS color -> unsupported error 
- after: Bulb without HS support receives HS color -> Z2M converts it to XY -> color applies

After this, we must [make enhanced HS false by default](https://github.com/Koenkk/zigbee-herdsman-converters/pull/11840). Because it overwrites normal HS, without confirmation that it's supported.

And I also wanted to make HS enabled by default. Because a lot of lights support it, but are missing it from their converter. And it's needed to get the full colors.
But it may cause more harm than good. If the bulb doesn't support it, but Z2M thinks it does, apps will fail to set color by HS.
Are there many devices that don't support HS?

Then, the next step in the code would be converting RGB to HS, if it's available. It jumps directly to XY which is not right.

andrei-lazarov:
There is also this call in `light.ts`, that (wrongly) assumes HS is enabled when not set.

https://github.com/Koenkk/zigbee-herdsman-converters/blob/831185577a8d851d07df2dcba504c6acb3ad9364/src/lib/light.ts#L39

xLerate:
@andrei-lazarov I would love to get this merged, as I'm currently running a patched version at home and would like to switch back to the official releases. The reason it hasn't been merged is that there were some failing tests in the z2m repository. I ran out of time, so I abandoned it. May be with some support we get this done.

andrei-lazarov:
Ok. I merged your PR into mine, where Z2M test passes, because I left enhancedHue enabled for Philips bulbs. 
You may close this.

xLerate:
which one? ;-)

andrei-lazarov:
- here https://github.com/Koenkk/zigbee-herdsman-converters/pull/11840 😅

PR #11371, opened 2026-01-25
Add another device id to ENGO ECB62-ZB converter

PaTara43:
I have one ENGO ECB62-ZB in my hands and it works like charm but just has a different ID so I add this to an existing converter


Koenkk:
Thanks!

PR #11372, opened 2026-01-25
fix: ZBMicro: entity category 'config' for 'RF turbo mode'

mundschenk-at:
The "RF turbo mode" expose should be in the `config` entity category (not the implied default `control`).


andrei-lazarov:
Can you add the same to inching?

mundschenk-at:
> Can you add the same to inching?

Theoretically yes, but as it's done as a composite expose, it's currently not part of the Home Assistant discovery at all. (TBH, I'm not sure why it's a composite, could be two separate binary toggles and a numeric expose.)

andrei-lazarov:
Oh.. composites don't show up in Home Assistant at all 😮 
I was looking at the Z2M dashboard

Koenkk:
Thanks!

PR #11373, opened 2026-01-25
feat: Futurehome TS0601 expose energy consumption

heiparta:
Expose energy consumption metric, ignoring outdated values which seem to happen often at least in a busier network.

Fixes feature request https://github.com/Koenkk/zigbee2mqtt/issues/23139

### DIFF
--- a/src/devices/futurehome.ts
+++ b/src/devices/futurehome.ts
@@ -10,15 +10,14 @@ const localValueConverters = {
     energyMonotonic: {
         from: (value: number, meta: Fz.Meta) => {
             const scaled = tuya.valueConverter.divideBy100.from(value);
-            const lastValue = meta.device.meta.energy;
-
-            if (typeof lastValue === "number" && scaled < lastValue && scaled !== 0) {
+            const lastValue = meta.device.meta.energy ?? 0;
+            if (scaled < lastValue && scaled !== 0) {
                 // Erraneous reading that is less than previous readings (and not a reset to 0), ignore it.
-                return {};
+                return lastValue;
             }
 
             meta.device.meta.energy = scaled;
-            return {energy: scaled};
+            return scaled;
         },
     },
 };
@@ -95,7 +94,7 @@ export const definitions: DefinitionWithExtend[] = [
                 [106, "window_detection", tuya.valueConverter.onOff],
                 [107, "max_temperature_protection", tuya.valueConverter.raw],
                 // Reported as a monotonically increasing counter while heating, using unit 0.01 kWh.
-                [123, "", localValueConverters.energyMonotonic],
+                [123, "energy", localValueConverters.energyMonotonic],
             ],
         },
     },


src/devices/futurehome.ts
```diff
@@ -1,11 +1,29 @@
...
+        from: (value: number, meta: Fz.Meta) => {
+            const scaled = tuya.valueConverter.divideBy100.from(value);
+            const lastValue = globalStore.getValue(meta.device, "energy_last");
+
+            if (typeof lastValue === "number" && scaled < lastValue && scaled !== 0) {
+                // Erraneous reading that is less than previous readings (and not a reset to 0), ignore it.
+                return {};
+            }
+
+            globalStore.putValue(meta.device, "energy_last", scaled);
```
Koenkk: This store resets on every z2m restart (so that means that energy would be lost every time). You can persist it in `_meta.device.meta.energy = xx;` instead (this is saved to the `data/database.db`)

heiparta: Thanks for the suggestion! I was initially thinking that it does not matter if the status is lost on z2m restart as they are so rare, but it definitely makes sense as persisting the data turned out to be that easy.

FYI, the thermostat is really noisy, and I think it is internally sending stale data once in a while during its transmit bursts. This is what a non-filtered heating period looked on Homeassistant side:
image (I patched the filtering into my z2m container at the vertical line)

Koenkk:
Thanks!

PapsMan-No:
Thanks a lot! This is a very welcome and huge improvement – I can now access information on total energy usage per futurehome thermostat via the Zigbee2MQTT device panel. This is awesome!

However, when I add a device such as sensor.fh_4_zb_3_meter_elec_e_import in the energy panel, nothing is updated / added. (I waited for a day to be sure.)

When I monitor current activity in the Zigbee2MQTT device panel, I see regular updates from other devices (such as my Futurehome provided smart plug and my Futurehome provided Develco HAN sensor), but nothing from the futurehome thermostats.

Is this information not available or do I need to do something to access current energy consumption?

heiparta:
@PapsMan-No Did I understand you correctly that you do see the total energy reading when you open a specific device in the z2m UI and see the tab "Exposes", but do not see anything in the Activity view?
If so, that is strange. I do see the energy readings pop up in the Activity view when the value is updated by a thermostat (and it actually changes):
image

Next thing you could verify is checking the Logs view. If the thermostats are sending the energy reading to MQTT, you should see frequent log lines like
`[03/02/2026, 18.31.50] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/tyohuone_thermostat', payload '{"child_lock":"UNLOCK","current_heating_setpoint":18,"energy":18.88,"hysteresis":2,"linkquality":76,"local_temperature":21,"local_temperature_calibration":0,"local_temperature_floor":20,"max_temperature_protection":27,"preset":"away","running_state":"idle","sensor":"max_guard","system_mode":"heat","window_detection":"ON"}'`

And regarding the Homeassistant sensor, for me the new entities appeared automatically to Homeassistant as `sensor.<thermostat friendly_name>_energy`.

PapsMan-No:
Yes, I do see the total energy reading:

IMG_0207

But I do get error messages in the individual devices section of the energy panel:
IMG_0209

When I check the logs in the Zigbee2MQTT panel, there are frequent updates for the HAN Sensor and a Zigbee Smartplug, but nothing for the Futurehome thermostats…

heiparta:
Sorry, no idea what those entities with errors in your second screenshot are. Maybe the entity names have changed for some reason, you can check the current entity names from that Device info view?
In any case, if you click the Energy sensor as shown in your Device info screenshot for a thermostat that has been heating recently, you should see a graph depicting the energy consumption
image
If the graph looks ok and shows increase when heating is on, then the problem is not on zigbee2mqtt/zigbee-herdsman-converters side but likely in your Homeassistant configuration.

PapsMan-No:
Kiitos! That helped me add the correct devices in the energy panel. No idea if there didn’t appear there before now or if I was just blind…

PR #11375, opened 2026-01-25
Corrected action for the long-press on Switch 4x EU-LIGHTIFY

miltschek:
Long press of the bottom right button generates an invalid action chain. It should be distinguished from the top right button. More info see https://github.com/Koenkk/zigbee2mqtt/issues/28628


src/converters/fromZigbee.ts
```diff
@@ -1673,7 +1673,7 @@ export const command_move_hue: Fz.Converter<"lightingColorCtrl", undefined, "com
     type: "commandMoveHue",
     convert: (model, msg, publish, options, meta) => {
         if (hasAlreadyProcessedMessage(msg, model)) return;
-        const movestop = msg.data.movemode === 1 ? "move" : "stop";
+        const movestop = msg.data.movemode === 1 ? "move" : msg.data.movemode === 3 ? "reverse" : "stop";
```
```suggestion
        const movestop = msg.data.movemode === 1 ? "move" : msg.data.movemode === 3 ? "down" : "stop";
```

To align with the spec

Image

chris-1243: Should we also correct the move mode value 0x01 to `up` instead of `move` to also align with the specs ?

Koenkk: Ideally yes but that would be a breaking change so let's not do it.

miltschek: I believe, there was already a breaking change in the last few months anyway. Current version delivers actions suffixed with a button number _1, _2, _3 and _4.

Koenkk: True but I would prefer not the keep breaking, merged it, thanks!

miltschek: Thank you!

PR #11381, opened 2026-01-26
feat(sonoff): add programmable stepping sequence support for SONOFF MINI-ZB2GS (fw 1.0.7)

CubeZ2mDeveloper:
## What
Add programmable stepping sequence support for **SONOFF MINI-ZB2GS** (firmware **v1.0.7**).
This extends the existing MINI-ZB2GS support with new functionality (no breaking changes intended).  

## Why
Firmware v1.0.7 introduces/contains the programmable stepping sequence capability, and this PR makes it configurable/usable from Zigbee2MQTT.

## Changes
- Device: SONOFF MINI-ZB2GS
- Added exposes for programmable stepping sequence (configuration + control).
- Added toZigbee converter(s) to write sequence settings and trigger execution (as supported by the device).
- Added fromZigbee parsing (when device reports/returns state), so UI/state reflect actual device config.

## Backward compatibility
- Designed to be safe for older firmware: missing attributes/unsupported commands are handled gracefully (no crashes).
- Existing MINI-ZB2GS exposes remain unchanged.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -8,7 +8,7 @@ import {logger} from "../lib/logger";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
 import * as tuya from "../lib/tuya";
-import type {DefinitionWithExtend, Expose, Fz, KeyValue, ModernExtend, Tz} from "../lib/types";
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11384, opened 2026-01-26
enhance ENGO EONE 230W with presets and floor temp limits

PaTara43:
I also got a piece of ENGO EONE 230W from local representative. The firmware is 2.1 (latest at the moment)
Previous convertor lacked datapoint for temperature calibraion, used wrong valve protection convertor and didn't implemet floor temparature limits at all

Also, presets were a bit off with no datapoint and extra modes

### DIFF
--- a/src/devices/engo.ts
+++ b/src/devices/engo.ts
@@ -739,7 +739,7 @@ export const definitions: DefinitionWithExtend[] = [
                     tuya.valueConverterBasic.lookup({
                         off: tuya.enum(0),
                         on: tuya.enum(1),
-                        anti_stop: tuya.enum(2)
+                        anti_stop: tuya.enum(2),
                     }),
                 ],
             ],


src/devices/engo.ts
```diff
@@ -592,9 +594,49 @@ export const definitions: DefinitionWithExtend[] = [
...
+                    "HIS_06",
+                    "HIS_08",
+                    "HIS_10",
+                    "HIS_20",
+                    "HIS_30",
+                    "HIS_40",
+                ])
                 .withDescription("Control algorithm"),
-            e.binary("valve_protection", ea.STATE_SET, "ON", "OFF").withDescription("Valve protection ON/OFF"),
+            e.enum("valve_protection", ea.STATE_SET, ["off", "on"]).withLabel("Valve Protection"),
```
Koenkk: Please keep this a binary

PaTara43: It doesn't work then!
Same here

https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/engo.ts#L736

Cause they may also have "antistop" option, valve protection is an enum

Koenkk: Can you check if `antistop` also works? You will also have to add it to:

```
                [
                    122,
                    "valve_protection",
                    tuya.valueConverterBasic.lookup({
                        off: tuya.enum(0),
                        on: tuya.enum(1),
                    }),
                ],
```

If it works use that, otherwise use `[122, "valve_protection", tuya.valueConverter.onOff],` and provide the debug log when using `e.binary("valve_protection", ea.STATE_SET, "ON", "OFF").withDescription("Valve protection ON/OFF"),` and trying to control it.

PaTara43: It doesn't have antistop, so I'll provide you with logs asap

PaTara43: It's here

off and on

https://gist.github.com/PaTara43/1a312d387883d217e21a5839f5a5a3b4

It didn't toggle on the device though

PaTara43: UPD
Damn, I have found Antistop there. May be a firmware update. So it's enum for sure

Koenkk:
Thanks!

PR #11390, opened 2026-01-26
Remove OTA support for Xiaomi MCCGQ12LM

torandreroland:
Remove OTA support for Xiaomi MCCGQ12LM and effectively roll back the change in #5537 (also see recent conversation in that issue)


Koenkk:
Thanks!

PR #11400, opened 2026-01-27
fix(sunricher): prioritize SR-ZG9101SAC-HP-SWITCH-2CH over Vesternet OEM

chris-1243:
Following
 - Koenkk/zigbee-herdsman-converters/pull/11281


Koenkk:
Thanks!

PR #11404, opened 2026-01-28
fix(detect): Support North American Ikea Jetström

haguenau:
I recently installed an [Ikea Jetström 100x40cm](https://www.zigbee2mqtt.io/devices/L2208.html).

It identifies as "JETSTROM 40100 NA (unsupported)" in the current version of Zigbee2MQTT, but otherwise works fine.


Koenkk:
Thanks!

PR #11408, opened 2026-01-28
Fix MOES cover switch position

Stalker-4x4:
Sometimes MOES curtain switches return an incorrect position (this always happens with mine). The value falls outside the range of 0–100 in either direction and can be either greater than 100 or less than 0. This causes problems with detecting the open/closed state, as well as with settings in Home Assistant.

### DIFF
--- a/src/lib/legacy.ts
+++ b/src/lib/legacy.ts
@@ -3100,6 +3100,7 @@ const fromZigbee = {
             let fixedValue = null;
             switch (dp) {
                 case dataPoints.coverPosition: {
+                    // https://github.com/Koenkk/zigbee-herdsman-converters/pull/11408
                     if (value >= 100 && value <= 127) {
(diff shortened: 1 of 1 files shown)


src/lib/legacy.ts
```diff
@@ -3097,10 +3097,18 @@ const fromZigbee = {
             const dp = dpValue.dp;
             const value = getDataValue(dpValue);
             let result = null;
+            let fixedValue = null;
             switch (dp) {
                 case dataPoints.coverPosition: {
+                    if (value >= 100 && value <= 127) {
```
Koenkk, line 3104:
```suggestion
                    // https://github.com/Koenkk/zigbee-herdsman-converters/pull/11408
                    if (value >= 100 && value <= 127) {
```

Koenkk:
Thanks!

PR #11410, opened 2026-01-28
fix: Bosch Twinguard 8750001213: Rename expose `co2` to `eco2`

burmistrzak:
This is a follow-up to a discussion started at https://github.com/Koenkk/zigbee-herdsman-converters/pull/11204#issuecomment-3716976392.

To avoid further confusion about capabilities, renaming `co2` to `eco2` seems to be an appropriate step. This should prevent ecosystems from misrepresenting sensor readings.

### DIFF
--- a/src/devices/bosch.ts
+++ b/src/devices/bosch.ts
@@ -234,7 +234,7 @@ const boschExtend = {
                 .withLabel("eCO₂")
                 .withUnit("ppm")
                 .withDescription("TVOC-derived CO₂-equivalent"),
-            e.numeric("iaq", ea.STATE).withValueMin(0).withValueMax(500).withValueStep(1).withLabel("IAQ").withDescription("Index for Air Quality"),
+            e.numeric("aqi", ea.STATE).withValueMin(0).withValueMax(500).withValueStep(1).withLabel("IAQ").withDescription("Index for Air Quality"),
             e.illuminance(),
             e
                 .numeric("battery", ea.STATE)


Koenkk:
Thanks!

mf19862:
I'm totally confused now. Where is the Co² data now? It is just gone. And the VOC data is now "ppm" it should be "ug/m³". This change totally messed up my setup and data

burmistrzak:
@mf19862 Everything as intended. Let me explain really quick:

1. Devices like the Twinguard **are not CO2 sensors**. The BME680 inside doesn't detect CO2. It measures TVOC and derives a CO2-equivalent (eCO2) value for presentation using a proprietary algorithm.
2. According to Bosch Sensortec, the BME680 expresses **relative change** in TVOC in **ppm**. It therefore cannot measure the absolute concentration of TVOC in ug/m³.
3. eCO2 is a TVOC value, thus the change in `device_class`

I know it takes some time to getting used to, but at least it's correct now. 😅

mf19862:
@burmistrzak **Thank you** for your detailed reply! This does clarify most of it. But there **was** an additional sensor "CO2" before, which is now removed. Where did this sensor come from?
The entity "CO2" did output a different value than "AQI" and "TVOC" but it was changing relatively.

burmistrzak:
@mf19862 Well... We made it up (kinda).

Long story short: Before, we used a non-linear formula to calculate a sensible CO2 value from raw IAQ, but switched back to the default formula (likely also used by the Bosch SHC) because the sensor's raw IAQ output was not behaving as described in the BME680 data sheet.

Bosch is _probably_ doing some sort of pre-processing in firmware before sending out the IAQ value via Zigbee.

Hope that makes sense. 🤞

src/devices/bosch.ts
```diff
@@ -227,14 +227,14 @@ const boschExtend = {
...
-                .numeric("co2", ea.STATE)
+                .numeric("eco2", ea.STATE)
                 .withValueMin(500)
                 .withValueMax(5500)
                 .withValueStep(1)
                 .withLabel("eCO₂")
                 .withUnit("ppm")
                 .withDescription("TVOC-derived CO₂-equivalent"),
-            e.numeric("aqi", ea.STATE).withValueMin(0).withValueMax(500).withValueStep(1).withLabel("AQI").withDescription("Air Quality Index"),
+            e.numeric("aqi", ea.STATE).withValueMin(0).withValueMax(500).withValueStep(1).withLabel("IAQ").withDescription("Index for Air Quality"),
```
Nerivec, line 237: Glanced at this while on the docs:
`e.numeric("aqi"` wasn't renamed, it shows aqi (Exposes in top table) and IAQ (Exposes section).
https://www.zigbee2mqtt.io/devices/8750001213.html

burmistrzak: @Nerivec Good catch!
IIRC, I left that in because IAQ isn't recognized by HA (but AQI is) and updating [`homeassistant.ts`](https://github.com/Koenkk/zigbee2mqtt/blob/85875aee27caa005946bb5a446080461a0d2fc33/lib/extension/homeassistant.ts#L123) for a single device seemed a bit overkill. 😅

But you're right, I should probably fix that.

PR #11412, opened 2026-01-29
Inovelli: Initial commit of reportTargetInfo functionality

rohankapoorcom:
I'm keeping this as a draft for now. We don't want to release it in the upcoming release on Feb 1, 2026 since there's still some tweaks that may need to be made after discussing it with the firmware engineer at Inovelli.

Also requires this change https://github.com/Nerivec/zigbee2mqtt-windfront/pull/416 so that the expose looks correct.


Koenkk (approved):

rohankapoorcom:
I'm comfortable with releasing this as is at this point. We can always tweak it later if necessary once the firmware engineer takes a look at it. What do you think @InovelliUSA ?

InovelliUSA:
@rohankapoorcom looking over this again and I don't see any issues submitting it right now.

Koenkk:
Thanks!

PR #11418, opened 2026-01-30
fix!: Bosch Twinguard 8750001213: Rename `co2` to `eco2` in fZ converter

burmistrzak:
Can't get results without renaming the converter key as well! 🙈


Koenkk:
Thanks!

Issue #11367, opened 2026-01-24
Default attribute values for generated devices


ffenix113:
Hello Koen,

For [zigbee_home](https://github.com/ffenix113/zigbee_home/issues/105) I am currently adding `Electrical Measurement` cluster (ZCL 4.9, cluster ID 0x0b04) with DC measurements by 
* Adding `MeasurementType` and setting its bit 6 (DC measurement)
* Adding `DCVoltage`, `DCCurrent`, `DCPower` (ids 0x100, 0x103, 0x106)
* Adding `DCVoltageDivisor`, `DCCurrentDivisor`, `DCPowerDivisor` (ids 0x201, 0x203, 0x205)

As other attributes are optional, and most have defaults - I assumed that it would be good to go by just providing the attributes that I needed. 

Unfortunately, while trying to pair such custom device with Z2M it results in some issues:
* Z2M tries to read & setup reporting for attributes that are not available (RMS Current/Voltage, AC Voltage/Current Multiplier/Divisor, etc..)
* While converting Electrical measurement value from such device [the factor is not being calculated "correctly"](https://github.com/Koenkk/zigbee-herdsman-converters/blob/c4ee6393608a118d3ed1b64311cd5eeb66edd41b/src/converters/fromZigbee.ts#L838C1-L840C77), as it tries to use attribute that is not present and just returns no factor (== 1).
* If I provide attributes for `DC(Voltage|Current|Power)Divisor` - they will not be used until they are explicitly read from `Developer console`. After the read - they will be used to properly modify the values of the reported attributes (if previous issue is ad-hoc fixed by `const multiplier = ... || 1` and same for divisor).

Both issues are resolved by adding attributes with "dummy" values, but I think that better approach would be to use default values if attribute is not present. And for reporting missing attributes - Z2M could know which attributes are present and set up only those.

Based on ZCL 2.3.4.2 I understand that if the optional/deprecated attribute(s) is missing - default value should be used instead:
> If the specification text of a cluster depends on the value of an optional or deprecated variable (e.g. attribute, command field, etc) of the same cluster, **then the variable SHALL have a well-defined default value that
SHALL be used in place of the missing variable**.

As I understand, the solutions could be:
[...]

PR #11422, opened 2026-01-30
fix: generated definition for dc electrical measure

ffenix113:
1. Fixes issue of meter not selecting AC/DC type correctly
2. Correctly set endpoint for energy meter
3. Fix default attribute change values for DC
4. Added DC converters in `toZigbee.ts` and a test for DC measurement cluster

Note: DC converters in `toZigbee.ts` key values collide with the AC ones (as for [`fromZigbee.ts`](https://github.com/ffenix113/zigbee-herdsman-converters/blob/78a852c31d2cc240823b96101e5322d48b4e51b7/src/converters/fromZigbee.ts#L844) as well), which could be an issue, if device would have AC and DC measurement on the same cluster (however small the chance may be).

While writing the test, I noticed that `exposes` part of the definition does not include endpoint for generated device. I am working on this for next MR.

Fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/11367

### DIFF
--- a/src/lib/generateDefinition.ts
+++ b/src/lib/generateDefinition.ts
@@ -361,17 +361,10 @@ async function extenderElectricityMeter(device: Zh.Device, endpoints: Zh.Endpoin
         const measurementType = await getClusterAttributeValue(endpoint, "haElectricalMeasurement", "measurementType", 0);
         // MeasurementType will have bit on index 6 set for DC measurement.
         const isDCMeasureType = ((measurementType >> 6) & 1) === 1;
-        // Any of the first 6 bits would mean AC measurements.
-        const isACMeasureType = (measurementType & 0b111111) !== 0;
 
+        args.electricalMeasurementType = "ac";
         if (isDCMeasureType) {
             args.electricalMeasurementType = "dc";
-        }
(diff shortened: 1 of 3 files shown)


src/lib/generateDefinition.ts
```diff
@@ -355,6 +355,30 @@ function extenderElectricityMeter(device: Zh.Device, endpoints: Zh.Endpoint[]):
...
+        const isACMeasureType = (measurementType & 0b111111) !== 0;
+
+        if (isDCMeasureType) {
+            args.electricalMeasurementType = "dc";
+        }
+        if (isACMeasureType) {
+            args.electricalMeasurementType = "ac";
+        }
+        if (isDCMeasureType && isACMeasureType) {
+            args.electricalMeasurementType = "both";
```
Koenkk: Maybe we can simplify here by not supporting `both` (so either `ac` or `dc`) as I don't expect there will be any device that will support both?

ffenix113: Sounds good. Dropped `both`, now AC takes priority. This does make it simpler.

Koenkk:
Thanks!

PR #11423, opened 2026-01-30
fix: correctly test 'exposes' values

ffenix113:
For devices with more than one endpoint
'exposes' values can have endpoints attached,
but they were not correctly tested, skipping the endpoint part.

P.S: would be nice for generated definitions to write tests for endpoint ID > 1. As in this case it will test if endpoint name are correctly applied while generating definition, and for first endpoint values should still be correct either way.

Related to https://github.com/Koenkk/zigbee-herdsman-converters/pull/11422 (merging one will require updating another before merge)


Koenkk:
Thanks!

PR #11426, opened 2026-01-30
AU-A1ZBSCRGBCX enable HS color

andrei-lazarov:
XY color doesn't work for this device:
- https://github.com/Koenkk/zigbee2mqtt/discussions/30827

Trying HS color


Koenkk:
Thanks!

PR #11433, opened 2026-01-31
Add another Philips Hue white PAR38 outdoor model

shawnheide:
As far as I can tell, this is the same as this one this other model, the picture is the same too: https://www.zigbee2mqtt.io/devices/9290018189.html. Let me know if you need anything else. Thanks!


Koenkk:
Thanks!

PR #11438, opened 2026-02-01
Update EKO07259, remove attributes not present and implement Schneider Electric custom clusters and attributes

svhelge:
Update EKO07259, remove attributes not present (occupancy and running state).
Added custom clusters and attributes for Schneider Electric devices, including user interface configuration, metering, and thermostat functionalities.


Koenkk:
Thanks!

PR #11454, opened 2026-02-03
fix: Sinope TH1123ZB-G2 and TH1124ZB-G2: swap sensing and off values for backlight dimming modes (#10987) [take 2]

lemoinem:
Here is my second try for #10987.

I had to disable the pre-commit hook, so this will probably not pass CI.
I did follow the example from your comment so not quite sure what the fix should be.


Koenkk:
Thanks!

lemoinem:
Thank you for the fix and the merge!

PR #11457, opened 2026-02-03
Fix battery division on Vesternet VES-ZB-REM-013

mik-laj:
My device reports that its battery is 200%.
Screenshot 2026-02-03 at 18 42 53
I have no way to test it on other FW versions, but I hope it will work similarly.

@martyn-vesternet Do you have any information about the difference between FW versions?


mik-laj:
Okey. I found a information. https://github.com/vesternet/zigbee2mqtt-herdsman-converters/blob/5861dd37019173285cfd902d31fe2bff324b97cd/VES-ZB-REM-013_as_brightness_controller.js#L60-L66
2.5.3_r20 don't needs division.
newer firmwares needs division.

Koenkk:
Thanks!

PR #11461, opened 2026-02-03
Add category to indicatorModeNoneRelayPos 

mik-laj:

Personally, I'd also change the category for Countdown, as I don't think many people use it in HA, but I'm not sure. What do you think?

Screenshot 2026-02-03 at 21 06 36


andrei-lazarov (approved):

andrei-lazarov:
I think Countdown is ok in Controls.
Because it's not a permanent setting (configuration). It's a single action: toggle relay in x seconds

mik-laj:
Thanks for sharing your thoughts, I appreciate the explanation and I generally agree.

That said, this isn't a field that a human would typically control directly. In practice, it would be populated by a script or automation rather than manual input.

For me, the "Controls" section is a bit problematic, as it mixes controls intended for direct human interaction with controls that are primarily automation-driven, which results in a suboptimal UX when adding controls to dashboards.

Koenkk:
thanks!

PR #11462, opened 2026-02-03
Add reporting for Moes ZWV-YC

andrei-lazarov:
- from here: https://github.com/Koenkk/zigbee2mqtt/discussions/28591#discussioncomment-15685086

### DIFF
--- a/src/devices/moes.ts
+++ b/src/devices/moes.ts
@@ -1266,7 +1266,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "ZWV-YC",
         vendor: "Moes",
         description: "Water valve",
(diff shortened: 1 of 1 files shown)


src/devices/moes.ts
```diff
@@ -1267,6 +1267,9 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Moes",
         description: "Water valve",
         extend: [m.battery(), m.onOff({powerOnBehavior: false})],
```
Koenkk:
```suggestion
        extend: [m.battery(), m.onOff({powerOnBehavior: false, configureReporting: true})],
```

Try this instead.

andrei-lazarov: Done

Koenkk:
Thanks!

PR #11463, opened 2026-02-03
Update tuya.ts for TS0601 _TZE284_noixx2uz

SaschaKP:
Added support for TRV603 (TS0601 [ _TZE284_noixx2uz]). Includes full schedule support and temperature validation, which was missing in previous attempts/requests, plus functional child lock (inverted in this device, see other comment), plus custom code for decoding/encoding daily schedule. Also, for local temperature, the device can give up to a maximum of -9, at -10 it will give E1, but I translated correctly the value for a reason: we want it to function better than the original.

image

Large image with white bg (transparency R:255 G:255 B:255):
image

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -442,7 +442,7 @@ const bindSlotTS0601SmartSceneKnob = async (entity: Zh.Endpoint | Zh.Group, slot
     await tuya.sendDataPointRaw(entity as Zh.Endpoint, 102, payload, "dataRequest", 0x10 + (slot - 1));
 };
 
-const trv603ScheduleConverter = (dayNumber: number): tuya.valueConverter => {
+const trv603ScheduleConverter = (dayNumber: number) => {
     return {
         from: (value: unknown) => {
             const buf = Buffer.isBuffer(value)


SaschaKP:
The code works perfectly in the external converter listed below:

https://github.com/Koenkk/zigbee2mqtt/issues/30906
similar to other issue:
https://github.com/Koenkk/zigbee2mqtt/issues/29450 (last comment I've made)

Missing:
- fetching schedule values (only from device to zigbee2mqtt because it needs specially formed packets as REQUEST, but it's really not needed for normal use, and would flood the zigbee network anyway)
- window mode usability and window open detection (bugged in firmware, I even tested with fast temperature drops, like 5 degrees C in 1 minute to no avail even on tuya hub+APP)

Working, but potentially needs to be addressed in the future:
child_lock is inverted, so false == true and vice versa. This seems more like a firmware bug, so it could be fixed in the future. (the child lock is also bugged on tuya hub/app, so when it's non active on DEVICE you'll see it active on the APP)

Koenkk:
Thanks!

Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `TRV603.png`

Martinius79:
Hello @SaschaKP!

The device is actually not a "TRV603", as far as I know and can see.
TRV603s are Wi-Fi-based and not from the same vendor.

This is a *new* model with a **Tuya ZigBee** module and **UART converter** (`TZE284` / `TYZS4`).
The vendor calls it **"TRV331"**.

grafik

The device is also sold as **AR331-WZ** by reseller *Matosio*:
[https://matosio.com/en/product/ar331-wz-tete-thermostatique-zigbee-pour-radiateur/](https://matosio.com/en/product/ar331-wz-tete-thermostatique-zigbee-pour-radiateur/)

Actual manufacturer: **Shenzhen Myuet Energy Saving Equipment Co., Ltd.**
Not yet listed on their homepage:
[https://www.szyuet.com/TRVwenkongfa.html](https://www.szyuet.com/TRVwenkongfa.html)

I got early versions from AliExpress with firmware `0.2.4` — *really* buggy, missing DPs/functions.
After firmware update → works much better!

Current version (`1.0.4`) works as described in your converter.
**Suggestion:** Note minimum version + pairing instructions in docs.

I figured out the same stuff (child lock, window detection, time/temp conversion)! 
Great work! 😅

**Rename suggestion:** "AR331-WZ" or "TRV331" (or add as alias)

Not sure how to do this if PR already merged (Zigbee2MQTT images too).
Advice, @Koenkk?

— Martinius

Koenkk:
@Martinius79 feel free to make a PR to rename it (also create a PR to add an image for the device)

SaschaKP:
I apologize for the confusion; it was a careless oversight caused in part by the fact that the device is virtually unbranded, making it difficult to identify at first glance. After checking the Tuya side, I can confirm you're right: the previous code was indeed written without any access to the Tuya IoT Platform or cloud services (until recently).

EDIT: Anyway, since the device is already mapped with its unique Device ID, the name shouldn't cause any issues; correcting it is simply the right thing to do for the sake of accuracy.

Martinius79:
I finally found the time to run some tests and create the PR for the renaming. Image and Markdown updates will follow.

The Tuya app shows it as AR331, so I used that name.

I whitened the image a bit so it looks more similar to the real device color.

BTW: I ordered the TRV603-WZ from Matosio — it’s a Zigbee device (so TRV603 is Wi-Fi). It’s already in the devices list and working fine but unrelated to the AR331 devices.

I also ordered the Matosio AR331-WZ and found that it’s a different device than the AR331, and it’s not listed yet. I’ll try to add support for it in the near future.

It’s actually the “big brother” of the AR331 — same manufacturer (Myuet) but with more features: holiday mode, eco and comfort modes, boost mode, and eight time slots per day. The Tuya app lists it as AR331 Pro.

PR #11466, opened 2026-02-04
feat: SONOFF TRVZB add smart thermostat features (v1.4.4)

CubeZ2mDeveloper:
**Summary**

- Add support for Smart Temperature Control for SONOFF TRVZB.
- The change is aligned with firmware trvzb_v1.4.4.

**What’s included**

- New expose for smart temperature control.
- OTA integration note: firmware trvzb_v1.4.4 has been submitted to zigbee-OTA.

**Testing**

- Tested with Zigbee2MQTT on SONOFF ZBDongle-P.

- Verified:
  - Pairing / interview
  - Read/write of thermostat setpoint and reporting
  - Smart Temperature Control related attributes
  - No regression on existing features

**Related**

zigbee-OTA PR: https://github.com/Koenkk/zigbee-OTA/pull/1056

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -2581,23 +2581,16 @@ export const definitions: DefinitionWithExtend[] = [
             }),
             sonoffExtend.weeklySchedule(),
             m.customTimeResponse("1970_UTC"),
-            m.enumLookup<"customSonoffTrvzb", SonoffTrvzb>({
+            m.binary<"customSonoffTrvzb", SonoffTrvzb>({
                 name: "smart_temperature_control",
-                // manual sends 0x00 (same as off), 0x01 is report-only
-                lookup: {off: 0x00, on: 0x02},
                 cluster: "customSonoffTrvzb",
                 attribute: "smartTempControl",
                 description:
-                    "Off: fully opens or closes the radiator valve based on a threshold temperature, " +
-                    "On: adjust valve opening based on a PID algorithm. " +
-                    'After enabling the smart mode, "Valve Opening Percentage" and "Temperature Accuracy" will be automatically disabled.',
+                    "Enable adaptive valve control using a PID algorithm. " +
+                    'When enabled, "Valve Opening Percentage" and "Temperature Accuracy" are unavailable.',
                 access: "ALL",
-                fzConvert: (model, msg, publish, options, meta) => {
-                    if ("smartTempControl" in msg.data) {
-                        const valueToMode: Record<number, string> = {0: "off", 1: "off", 2: "on"};
-                        return {smart_temperature_control: valueToMode[msg.data.smartTempControl] ?? "off"};
-                    }
-                },
+                valueOn: [true, 0x02],
+                valueOff: [false, 0x00],
             }),
         ],
         ota: true,


src/devices/sonoff.ts
```diff
@@ -98,6 +98,7 @@ interface SonoffTrvzb {
         temporaryMode: number;
         temporaryModeTime: number;
         temporaryModeTemp: number;
+        smartTemControl: number;
```
grossermax: There is a typo (the same applies below).
```suggestion
        smartTempControl: number;
```

src/devices/sonoff.ts
```diff
@@ -258,6 +259,66 @@ const sonoffExtend = {
             commandsResponse: {},
         });
     },
+    smartTemControl(): ModernExtend {
+        const expose = e
+            .enum("smart_temperature_control", ea.ALL, ["off", "precision", "smart"])
+            .withDescription("Manual mode: off or precision (same effect), Smart mode: automatic temperature control");
```
grossermax: As already described in https://github.com/Koenkk/zigbee-OTA/pull/1056#issuecomment-3853773510, I would suggest adding a bit more detail for clarity.

A very good explanation of the new **Smart Temperature Control** ("Adaptive Mode") can be found here:
https://github.com/zigpy/zigpy-ota/issues/23
> _"This release adds Adaptive Mode. It is a feature designed to minimize temperature fluctuations.
When enabled, the device dynamically adjusts the valve opening based on the temperature changes and the target temperature to maintain a more stable environment._
> 
> _Note: Enabling Adaptive Mode will automatically disable Valve Opening Percentage and Temperature Accuracy."_

If linking external references is acceptable, the official Sonoff blog announcement may also be helpful:
https://sonoff.tech/blogs/news/trvzb-upgrade-smart-temperature-control-for-more-stable-room-temperature

In particular, it would be great if the PR could explicitly include (or mention) the note that enabling **Smart Temperature Control** will *disable* the "Valve Opening Percentage" and "Temp. Accuracy" features, since this behavioural change is particularly important for users.

Thank you so much for your contribution!

src/devices/sonoff.ts
```diff
@@ -258,6 +259,66 @@ const sonoffExtend = {
...
+                },
+            },
+        ];
+
+        const fromZigbee = [
+            {
+                cluster: "customSonoffTrvzb",
+                type: ["attributeReport", "readResponse"],
+                convert: (model, msg, publish, options, meta) => {
+                    if ("smartTemControl" in msg.data) {
```
Koenkk:
```suggestion
                    if ("smartTempControl" in msg.data) {
```

fix all other typos

src/devices/sonoff.ts
```diff
@@ -258,6 +259,66 @@ const sonoffExtend = {
             commandsResponse: {},
         });
     },
+    smartTemControl(): ModernExtend {
```
Koenkk: Use `m.enumLookup` instead, example:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/d4f085c94393464810f2746c0ebfb76d59454b02/src/devices/sonoff.ts#L2245

src/devices/sonoff.ts
CubeZ2mDeveloper: Thanks for the detailed feedback and the references!
We've updated the PR to add a clearer description of Smart Temperature Control (Adaptive Mode) and explicitly documented that enabling it will disable Valve Opening Percentage and Temperature Accuracy.

About the firmware: We are official maintainers of the Sonoff/eWeLink Zigbee firmware. The attached OTA image is provided by us and originates from the official Sonoff/eWeLink firmware release pipeline (i.e., an official build intended for end users).

src/devices/sonoff.ts
CubeZ2mDeveloper: Thanks for the review. I considered using `m.enumLookup` but the mapping for this attribute is asymmetric between write and report:

**toZigbee (write):**

| Mode      | Value  |
| --------- | ------ |
| off       | `0x00` |
| precision | `0x00` |
| smart     | `0x02` |

**fromZigbee (report):**

| Value  | Mode      |
| ------ | --------- |
| `0x00` | off       |
| `0x01` | precision |
| `0x02` | smart     |

The device reports `precision` as `0x01`, but to activate it you must write `0x00` (same as `off`). `0x01` is a report-only value — writing it to the device has no effect.

`m.enumLookup` uses a single `lookup` object for both directions (`getFromLookup` for toZigbee, `getFromLookupByValue` for fromZigbee), so it cannot represent this asymmetry. If we use `{off: 0x00, precision: 0x01, smart: 0x02}`, setting `precision` would incorrectly write `0x01` to the device.

Would you prefer I:

1. Keep the custom extend as-is to preserve correct behavior, or
2. Simplify to `m.enumLookup` with `{off: 0x00, precision: 0x01, smart: 0x02}` and accept that writing `precision` sends `0x01` (which the device will ignore)?

Happy to adjust either way.

src/devices/sonoff.ts
grossermax: Nice, thanks!
Really appreciate the work you put into this. Happy to see more of your stuff down the road!

KipK:
@CubeZ2mDeveloper I 've made an external converter of your PR to test.

It seems to works ok : https://gist.github.com/KipK/448992d0f5abda3c934bd098b29e4408

Something you should add to this PR, is to force the valve position to 100% when the mode is switched to Smart or the other internal logic.

If valve is previously set to 0 when in manual mode, then heating won't never start in Smart mode.
User have to manually put it to 100% and it's confusing. 
Automating it from the converter would be better.

Fixing this on firmware side would be even better

ActuallyRuben:
I've tested the addition as an external converter as well. It appears to work correctly. I can hear the valve open and close partially when manipulating the temperature.

I am just wondering, what is the 'precision mode'? Why is setting it to precision mode mapped to writing a `0`?
Looking at the [ZHA script](https://resource-wangsu.helplook.net/docker_production/a4j0b0/article/F1vhxZrI/attachments/TRVZB_ZHA_Script.zip) provided by SONOFF, it only lists `0` and `2` as valid values for the smart temperature control mode. Is the value `1` documented anywhere?

fma965:
> I've tested the addition as an external converter as well. It appears to work correctly. I can hear the valve open and close partially when manipulating the temperature.

Same here!, seems good.

src/devices/sonoff.ts
Koenkk: I've modified `m.enumLookup` to accept a custom converter: https://github.com/Koenkk/zigbee-herdsman-converters/commit/3aac9b3e9009b61f188bce2db1a9d7819df0ed8c

Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/af672bb37653e2999616df1613de14c74b14c8de/src/devices/sonoff.ts#L2618

MeisterQ:
Wow finally. I made so many support tickets because of this feature. Then i finally sold my TRVs and bought aqaras w600. But they are so bad. They are very quiet, but the algoritm is really bad. How does it act on the sonoff? Someone got any graphs with current, target temp and valveposition?? Is there an expose for valve position now?

KipK:
> Wow finally. I made so many support tickets because of this feature. Then i finally sold my TRVs and bought aqaras w600. But they are so bad. They are very quiet, but the algoritm is really bad. How does it act on the sonoff? Someone got any graphs with current, target temp and valveposition?? Is there an expose for valve position now?

There's no valve position on sonoff trvzb unfortunately.

MeisterQ:
> > Wow finally. I made so many support tickets because of this feature. Then i finally sold my TRVs and bought aqaras w600. But they are so bad. They are very quiet, but the algoritm is really bad. How does it act on the sonoff? Someone got any graphs with current, target temp and valveposition?? Is there an expose for valve position now?
> 
> There's no valve position on sonoff trvzb unfortunately.

To bad. Would be an easy made expose when they now use PID control.
I recently had a look on the hardware and found, they are using a 4-wire motor. Should be at least a simple stepper or a dc motor with rotary encoder.

src/devices/sonoff.ts
CubeZ2mDeveloper: Thanks @Koenkk! I've refactored smartTempControl to use m.enumLookup with a custom fzConvert to handle the asymmetric mapping (device reports 0x01 for manual mode, but write requires 0x00). Also renamed precision to manual for clarity.

Thanks for adding the fzConvert parameter to enumLookup — it made this much cleaner.

CubeZ2mDeveloper:
> I've tested the addition as an external converter as well. It appears to work correctly. I can hear the valve open and close partially when manipulating the temperature.
> 
> I am just wondering, what is the 'precision mode'? Why is setting it to precision mode mapped to writing a `0`? Looking at the [ZHA script](https://resource-wangsu.helplook.net/docker_production/a4j0b0/article/F1vhxZrI/attachments/TRVZB_ZHA_Script.zip) provided by SONOFF, it only lists `0` and `2` as valid values for the smart temperature control mode. Is the value `1` documented anywhere?

Thanks for testing! @ActuallyRuben 

Regarding your questions about precision mode:

**What is precision mode?**
When the device reports `0x01`, it indicates "precision" (manual precision control) mode — meaning the TRV is operating with manual valve control rather than smart automatic adjustment. In the latest commit I've renamed it to `manual` for clarity.

**Why does setting precision/manual write `0x00` instead of `0x01`?**
`0x01` is a report-only value — the device uses it to indicate its current state, but for write commands, `0x00` (off) achieves the same effect. This is consistent with the SONOFF ZHA script which only lists `0` and `2` as valid write values. Writing `0x01` to the device has no documented effect.

**Is `0x01` documented?**
Not in the official SONOFF ZHA script. I discovered it through device attribute reports — when the TRV is in manual mode and actively controlling valve precision, it reports `0x01` instead of `0x00`. So the mapping is:

| Value | Direction | Meaning |
|-------|-----------|---------|
| `0x00` | Read/Write | Off (smart control disabled) |
| `0x01` | Read only | Manual mode (reported by device) |
| `0x02` | Read/Write | Smart mode (automatic control) |

This is why a custom `fzConvert` is needed in the `m.enumLookup` — the read and write mappings are asymmetric.

CubeZ2mDeveloper:
## Changes in this update

1.  **Refactored `smartTempControl` to use `m. enumLookup`** with custom `fzConvert` (thanks to @Koenkk for adding the `fzConvert` parameter)

2.  **Renamed `precision` to `manual`** — clearer naming that pairs naturally with `smart` (`off` / `manual` / `smart`)

3.  **Fixed typo** in previous commits: `smartTemControl` → `smartTempControl` (missing `p`)

GigiPompieru:
@CubeZ2mDeveloper thanks for support. Can you please then explain what is the difference between `Off` and `Manual` from an operation of device perspective?

MeisterQ:
> Value	Direction	Meaning
> `0x00`	Read/Write	Off (smart control disabled)
> `0x01`	Read only	Manual mode (reported by device)
> `0x02`	Read/Write	Smart mode (automatic control)
> This is why a custom `fzConvert` is needed in the `m.enumLookup` — the read and write mappings are asymmetric.

Does this mean:

1. With Off 0x00 its working like it was "0% or 100% opening"
2. With Manual mode 0x01 you can set the valve position directly like "go to 50%"?
3. With Smart mode 0x02 its working by itself with the pid controller? 
4. Is there a way to tweak the PID parameters?

CubeZ2mDeveloper:
> @CubeZ2mDeveloper thanks for support. Can you please then explain what is the difference between `Off` and `Manual` from an operation of device perspective?

Hi, @GigiPompieru
0x01 is a report-only value — the device uses it to indicate its current state, but for write commands, 0x00 (off) achieves the same effect. This is consistent with the SONOFF ZHA script which only lists 0 and 2 as valid write values. Writing 0x01 to the device has no documented effect.

GigiPompieru:
So, the TRV is in off mode by default with valve values at 100% and 0%. If the user modifies these values in any way, as we can now, the TRV will switch to ``0x01`` meaning manual?

Will the TRV remain in manual until put into Smart or will it switch back to ``Off`` if the user returns the values to 100/0?

Sorry for all the questions, just trying to better understand the new feature.

CubeZ2mDeveloper:
> So, the TRV is in off mode by default with valve values at 100% and 0%. If the user modifies these values in any way, as we can now, the TRV will switch to `0x01` meaning manual?
> 
> Will the TRV remain in manual until put into Smart or will it switch back to `Off` if the user returns the values to 100/0?
> 
> Sorry for all the questions, just trying to better understand the new feature.

No worries, great questions!

**Does modifying valve values switch the TRV to manual (`0x01`)?**
No. The valve opening percentage and temperature accuracy values take effect after you manually disable smart temperature control (i.e., set it to `off` or `manual`). Changing valve values alone does not switch the mode.

**Will it switch back to off if you restore the values to 100/0?**
This attribute is purely a toggle for smart temperature control — nothing more. `off` and `manual` are functionally identical; the only difference is that `0x00` is the write value and `0x01` is the report value for the same state. It has no relation to valve values or any other settings.

To put it simply:
- `off` / `manual` → Smart temperature control **disabled**, TRV uses your configured valve opening percentage and temperature accuracy
- `smart` → Smart temperature control **enabled**, TRV automatically adjusts valve parameters (your manual valve settings are ignored)

pszemus:
So couldn't we just leave the "off" and "on" options for smart mode, for the sake of the UX?

GigiPompieru:
> So couldn't we just leave the "off" and "on" options for smart mode, for the sake of the UX?

I think that, even if they are not displayed in the UI, the ```manual``` value of ```0x01``` still needs to be coded.

Remember that when the ```external``` temp sensor was introduced we had just ```internal``` and ```external``` and people reported log errors because the TRV would have two other states, coded after as ```external_2``` and ```external_3```, which are basically the same as ```internal``` or ```external``` but have different values which Z2M didn't know about.

CubeZ2mDeveloper:
> So couldn't we just leave the "off" and "on" options for smart mode, for the sake of the UX?

Good point from a UX perspective! However, there are two reasons to keep the three-value enum:

1. **Confirming the device actually responded.** When you send `off` (`0x00`), the device reports back `0x01` (manual). If we only exposed `off`/`on`, the reported state would never match what the user sent — the UI would show `off` but the device reports a value that maps to something else. With three values, the user sends `off`, the device reports `manual`, and the state update is clear and unambiguous. It serves as confirmation that the device received and processed the command.

2. **Future extensibility.** SONOFF may assign distinct behavior to other values in future firmware updates. Keeping them as separate values avoids a breaking change later if that happens.

MeisterQ:
When will the smart mode be available? I updated to 1.4.4, but can it already use it?

fma965:
> When will the smart mode be available? I updated to 1.4.4, but can it already use it?

With this https://github.com/Koenkk/zigbee-herdsman-converters/pull/11466#issuecomment-3861048825 yes, without it when this PR is merged, then when next Z2M release is out.

src/devices/sonoff.ts
```diff
@@ -2579,6 +2581,23 @@ export const definitions: DefinitionWithExtend[] = [
             }),
             sonoffExtend.weeklySchedule(),
             m.customTimeResponse("1970_UTC"),
+            m.enumLookup<"customSonoffTrvzb", SonoffTrvzb>({
+                name: "smart_temperature_control",
+                // manual sends 0x00 (same as off), 0x01 is report-only
+                lookup: {off: 0x00, manual: 0x00, smart: 0x02},
```
ActuallyRuben:
```suggestion
                lookup: {off: 0x00, on: 0x02},
```

src/devices/sonoff.ts
```diff
@@ -2579,6 +2581,23 @@ export const definitions: DefinitionWithExtend[] = [
...
+                cluster: "customSonoffTrvzb",
+                attribute: "smartTempControl",
+                description:
+                    "Manual mode: off or manual (same effect), Smart mode: automatic temperature control. " +
+                    'After enabling the smart mode, "Valve Opening Percentage" and "Temperature Accuracy" will be automatically disabled.',
+                access: "ALL",
+                fzConvert: (model, msg, publish, options, meta) => {
+                    if ("smartTempControl" in msg.data) {
+                        const valueToMode: Record<number, string> = {0: "off", 1: "manual", 2: "smart"};
```
ActuallyRuben:
```suggestion
                        const valueToMode: Record<number, string> = {0: "off", 1: "off", 2: "on"};
```

src/devices/sonoff.ts
```diff
@@ -2579,6 +2581,23 @@ export const definitions: DefinitionWithExtend[] = [
...
             m.customTimeResponse("1970_UTC"),
+            m.enumLookup<"customSonoffTrvzb", SonoffTrvzb>({
+                name: "smart_temperature_control",
+                // manual sends 0x00 (same as off), 0x01 is report-only
+                lookup: {off: 0x00, manual: 0x00, smart: 0x02},
+                cluster: "customSonoffTrvzb",
+                attribute: "smartTempControl",
+                description:
+                    "Manual mode: off or manual (same effect), Smart mode: automatic temperature control. " +
+                    'After enabling the smart mode, "Valve Opening Percentage" and "Temperature Accuracy" will be automatically disabled.',
```
ActuallyRuben:
```suggestion
                    'Off: fully opens or closes the radiator valve based on a threshold temperature, ' +
                    'On: adjust valve opening based on a PID algorithm. ' +
                    'After enabling the smart mode, "Valve Opening Percentage" and "Temperature Accuracy" will be automatically disabled.',
```

GigiPompieru:
> In my opinion the distinction between the 0x0 and 0x1 states is confusing. I propose presenting both values as an `off` state.

Although your statement is logical, as a 'basic' user I think the best way to implement it is as close as the vendor does it in its native app - be it flawed. This way there can be consistency of features and the terms used by users of all ecosystems.

Secondly, even if Z2M is not strictly HA, we know most of us are also using HA. But TRVZB can he integrated in HA with ZHA and with eWelink. Having Z2M expose just ```off``` while the other two will, probably, expose ```off``` and `manual` will create confusion between the users of the same hardware device.

lifo9:
For posterity, my TRVZB has been automatically upgraded to v1.4.4 and `valve_opening_degree` stopped working.
I use [Versatile Thermostat](https://github.com/jmcollin78/versatile_thermostat) to control the valve directly.

I had to send the following commands via MQTT:

1. Disable smart temp control:
   ```json
   {"custom":{"cluster":"customSonoffTrvzb","type":"set","payload":{"smartTempControl":0}}}
   ```

2. Set TRV to pass-through mode:
   ```json
   {"system_mode": "heat", "occupied_heating_setpoint": 35}
   ```

KipK:
yes that's how it's made on the fw, if smartTempControl is activated then you can't control the valve position.

lifo9:
> yes that's how it's made on the fw, if smartTempCon﻿trol is activated then you can't control the valve position.

I understand, but:
- Firmware 1.44 was pushed to me before this PR was merged.
- It seems `smartTempControl` was automatically activated after the new firmware was installed, causing the Versatile Thermostat integration to suddenly stop working.

I had to use the MQTT console to resolve the issue.
I know this isn't directly related to this PR, but in the future, perhaps you could defer firmware updates until the Zigbee2Mqtt integration is ready.

Thanks

KipK:
In fact it is related. The converter should default to Off instead of Smart. 

I haven't checked but that would need to be fixed or there will have thousands of people complaining

GigiPompieru:
> * Firmware 1.44 was pushed to me before this PR was merged.
> 
> * It seems `smartTempControl` was automatically activated after the new firmware was installed, causing the Versatile Thermostat integration to suddenly stop working.

This is quite strange. I didn't even know that you can configure Z2M to auto-update.
I have updated 2 of my 6 TRVs for testing and, as @KipK said, it defaulted to ```off``` and the valve control works as on 1.4.1

MeisterQ:
> > So, the TRV is in off mode by default with valve values at 100% and 0%. If the user modifies these values in any way, as we can now, the TRV will switch to `0x01` meaning manual?
> > Will the TRV remain in manual until put into Smart or will it switch back to `Off` if the user returns the values to 100/0?
> > Sorry for all the questions, just trying to better understand the new feature.
> 
> No worries, great questions!
> 
> **Does modifying valve values switch the TRV to manual (`0x01`)?** No. The valve opening percentage and temperature accuracy values take effect after you manually disable smart temperature control (i.e., set it to `off` or `manual`). Changing valve values alone does not switch the mode.
> 
> **Will it switch back to off if you restore the values to 100/0?** This attribute is purely a toggle for smart temperature control — nothing more. `off` and `manual` are functionally identical; the only difference is that `0x00` is the write value and `0x01` is the report value for the same state. It has no relation to valve values or any other settings.
> 
> To put it simply:
> 
>     * `off` / `manual` → Smart temperature control **disabled**, TRV uses your configured valve opening percentage and temperature accuracy
> 
>     * `smart` → Smart temperature control **enabled**, TRV automatically adjusts valve parameters (your manual valve settings are ignored)

Do you know if:

1. we can modify PID parameters?
2. Have a Valve position expose?#
3. Do i have to set the valve opening and closing degeree to 100/0?
4. how many steps / how fast will the valve act in smart mode?
5. Do i have to do any other settings except of set the smart temperature control expose to "smart"?
6. Whats the difference between off and precision?

MeisterQ:
> > > So, the TRV is in off mode by default with valve values at 100% and 0%. If the user modifies these values in any way, as we can now, the TRV will switch to `0x01` meaning manual?
> > > Will the TRV remain in manual until put into Smart or will it switch back to `Off` if the user returns the values to 100/0?
> > > Sorry for all the questions, just trying to better understand the new feature.
> > 
> > 
> > No worries, great questions!
> > **Does modifying valve values switch the TRV to manual (`0x01`)?** No. The valve opening percentage and temperature accuracy values take effect after you manually disable smart temperature control (i.e., set it to `off` or `manual`). Changing valve values alone does not switch the mode.
> > **Will it switch back to off if you restore the values to 100/0?** This attribute is purely a toggle for smart temperature control — nothing more. `off` and `manual` are functionally identical; the only difference is that `0x00` is the write value and `0x01` is the report value for the same state. It has no relation to valve values or any other settings.
> > To put it simply:
> > ```
> > * `off` / `manual` → Smart temperature control **disabled**, TRV uses your configured valve opening percentage and temperature accuracy
> > 
> > * `smart` → Smart temperature control **enabled**, TRV automatically adjusts valve parameters (your manual valve settings are ignored)
> > ```
 
Do you know if:

1. we can modify PID parameters?
2. Have a Valve position expose?#
3. Do i have to set the valve opening and closing degeree to 100/0?
4. how many steps / how fast will the valve act in smart mode?
5. Do i have to do any other settings except of set the smart temperature control expose to "smart"?
6. Whats the difference between off and precision?

Currently my TRVs are on "smart", im using the external sensor and it is already 0.4°C below setpoint and the valves are closed.
Feels like "nothing happens"

ActuallyRuben:
> > In my opinion the distinction between the 0x0 and 0x1 states is confusing. I propose presenting both values as an `off` state.
> 
> Although your statement is logical, as a 'basic' user I think the best way to implement it is as close as the vendor does it in its native app - be it flawed. This way there can be consistency of features and the terms used by users of all ecosystems.
> 
> Secondly, even if Z2M is not strictly HA, we know most of us are also using HA. But TRVZB can he integrated in HA with ZHA and with eWelink. Having Z2M expose just `off` while the other two will, probably, expose `off` and `manual` will create confusion between the users of the same hardware device.

I agree that consistency is good. I have no clue how it's presented in eWelink. However, on SONOFF's [official blog](https://sonoff.tech/blogs/news/trvzb-upgrade-smart-temperature-control-for-more-stable-room-temperature), they provided a [ZHA script](https://resource-wangsu.helplook.net/docker_production/a4j0b0/article/F1vhxZrI/attachments/TRVZB_ZHA_Script.zip). In this script they appear to implement the smart temperature control setting as a simple on/off:
```py
    .switch(
        CustomSonoffCluster.AttributeDefs.smart_temperature_control.name,
        # ControlModeType,
        CustomSonoffCluster.cluster_id,
        on_value=2,
        off_value=0,
        translation_key="adaptive_mode",
        fallback_name="Adaptive mode",
    )
```
Supporting my suggestion to show it as an on/off setting in Z2M as well.

maciejsszmigiero:
> Currently my TRVs are on "smart", im using the external sensor and it is already 0.4°C below setpoint and the valves are closed.
> Feels like "nothing happens"

For me, the adaptive mode seems to only starts opening the valve again about 1 degrees C below the setpoint, sometimes it takes almost 1.5 degrees C below the setpoint for the temperature to start rising again.

On the valve closing side, TRVZB also seems to let the temperature reach almost 1.5 degrees C above the setpoint before closing the valve again.

I'm not sure if that ± 1.5 C is the intended accuracy of the adaptive mode or are these temperature swings caused by adaptive mode being too slow to react to temperature changes and letting the room overcool or overheat.

---

Overall, that's much worse performance for me than the old ON/OFF mode with temperature accuracy set to -0.2 degree C.

MeisterQ:
> > Currently my TRVs are on "smart", im using the external sensor and it is already 0.4°C below setpoint and the valves are closed.
> > Feels like "nothing happens"
> 
> For me, the adaptive mode seems to only starts opening the valve again about 1 degrees C below the setpoint, sometimes it takes almost 1.5 degrees C below the setpoint for the temperature to start rising again.
> 
> On the valve closing side, TRVZB also seems to let the temperature reach almost 1.5 degrees C above the setpoint before closing the valve again.
> 
> I'm not sure if that ± 1.5 C is the intended accuracy of the adaptive mode or are these temperature swings caused by adaptive mode being too slow to react to temperature changes and letting the room overcool or overheat.
> 
> Overall, that's much worse performance for me than the old ON/OFF mode with temperature accuracy set to -0.2 degree C.

Exactly my thought. The Adaptive mode is really worse.

GigiPompieru:
> Exactly my thought. The Adaptive mode is really worse.

Well, this is disappointing. I didn't plan on using it, since in my mind its better suited for scenarios where the heating is always on (like central heated apartments) and you want the best level of comfort, but it sounded a step in the right direction, especially since people are doing this already with VT.

KipK:
I confirm, the reports I received on this "smart" regulation is it's not working as expected. 
This looks like a bug on fw side.

MeisterQ:
> > Exactly my thought. The Adaptive mode is really worse.
> 
> Well, this is disappointing. I didn't plan on using it, since in my mind its better suited for scenarios where the heating is always on (like central heated apartments) and you want the best level of comfort, but it sounded a step in the right direction, especially since people are doing this already with VT.

Yes. The biggest problem here is related to this:
https://github.com/jmcollin78/versatile_thermostat/discussions/1704#discussion-9402097

The Valve only need few steps to control the full heating power. And also the loss of calibration of the TRV. 
If it tries to close from small steps to full closed, it will loose steps and show 0% even if its still open abit.

https://forum.ewelink.cc/t/trvzb-force-calibration-and-force-close/207675 Give this feature a vote and a like, so we may get a forced calibration feature.

ScratMan:
> I confirm, the reports I received on this "smart" regulation is it's not working as expected. This looks like a bug on fw side.

I don't think it's a bug, but rather that the feature has been designed for permanent heat sources (district heating or condominium boilers, where a shared heat production system provides hot water in continuous).
Using this mode with individual gas or oil boilers will lead to keeping the boiler active for veeeery long periods.
There should be two options in the smart adaptive mode: valves opening modulation (for continuous heat sources) and power cycling modulation (PWM, for individual boilers).

GigiPompieru:
well, if the valve stays fully closed until the actual temp gets to 1.5 below the set one, it sounds like a bug. I would expect ```Temperature accuracy``` to still apply to the new Smart Mode.

MeisterQ:
> well, if the valve stays fully closed until the actual temp gets to 1.5 below the set one, it sounds like a bug. I would expect `Temperature accuracy` to still apply to the new Smart Mode.

If the temperature drop 0.1°C below target, the valve probably should increase the heating by 1% or more according to the PID parameters

KipK:
For the graph I've seen it looks more an hysteresis with adaptive valve than a true PID loop

Koenkk:
@CubeZ2mDeveloper please move out of draft when this pr is ready for merge.

CubeZ2mDeveloper:
> In my opinion the distinction between the 0x0 and 0x1 states is confusing. I propose presenting both values as an off state. If 0x1 is never sent to the device, and 0x0 is never reported by the device, then what difference does it make if these are encoded as the same state?
If a later firmware update adds a different meaning to these values, then we should deal with that, whenever that update arrives.

Apologies for the delayed response; our team was out during the Chinese New Year holiday.

Your assessment is correct. Distinguishing 0x00 and 0x01 as separate user-facing states is unnecessary with current firmware behavior.
We now treat both 0x00 and 0x01 as `false` (off), and 0x02 as `true` (on). Write operations remain limited to 0x00 and 0x02.

This keeps the exposed behavior clear and deterministic. If a future firmware assigns a distinct meaning to 0x01, we will update the converter accordingly.

CubeZ2mDeveloper:
> > Currently my TRVs are on "smart", im using the external sensor and it is already 0.4°C below setpoint and the valves are closed.
> > Feels like "nothing happens"
> 
> For me, the adaptive mode seems to only starts opening the valve again about 1 degrees C below the setpoint, sometimes it takes almost 1.5 degrees C below the setpoint for the temperature to start rising again.
> 
> On the valve closing side, TRVZB also seems to let the temperature reach almost 1.5 degrees C above the setpoint before closing the valve again.
> 
> I'm not sure if that ± 1.5 C is the intended accuracy of the adaptive mode or are these temperature swings caused by adaptive mode being too slow to react to temperature changes and letting the room overcool or overheat.
> 
> Overall, that's much worse performance for me than the old ON/OFF mode with temperature accuracy set to -0.2 degree C.


Thank you for the detailed observations.

We have recorded the adaptive-mode behavior you described (temperature swings around the setpoint) and forwarded it to our firmware development team for further investigation.

You can also submit feedback through the SONOFF eWeLink app to provide additional data points.

MeisterQ:
> > > Currently my TRVs are on "smart", im using the external sensor and it is already 0.4°C below setpoint and the valves are closed.
> > > Feels like "nothing happens"
> > 
> > 
> > For me, the adaptive mode seems to only starts opening the valve again about 1 degrees C below the setpoint, sometimes it takes almost 1.5 degrees C below the setpoint for the temperature to start rising again.
> > On the valve closing side, TRVZB also seems to let the temperature reach almost 1.5 degrees C above the setpoint before closing the valve again.
> > I'm not sure if that ± 1.5 C is the intended accuracy of the adaptive mode or are these temperature swings caused by adaptive mode being too slow to react to temperature changes and letting the room overcool or overheat.
> > Overall, that's much worse performance for me than the old ON/OFF mode with temperature accuracy set to -0.2 degree C.
> 
> Thank you for the detailed observations.
> 
> We have recorded the adaptive-mode behavior you described (temperature swings around the setpoint) and forwarded it to our firmware development team for further investigation.
> 
> You can also submit feedback through the SONOFF eWeLink app to provide additional data points.

If you are in contact with development team, please also think about an expose where you can force the calibration, because the smart mode causes the TRV to loose its calibration very quickly and 0% wont be "closed" within afew hours / days.

Thanks for your efford.

CubeZ2mDeveloper:
> > > > Currently my TRVs are on "smart", im using the external sensor and it is already 0.4°C below setpoint and the valves are closed.
> > > > Feels like "nothing happens"
> > > 
> > > 
> > > For me, the adaptive mode seems to only starts opening the valve again about 1 degrees C below the setpoint, sometimes it takes almost 1.5 degrees C below the setpoint for the temperature to start rising again.
> > > On the valve closing side, TRVZB also seems to let the temperature reach almost 1.5 degrees C above the setpoint before closing the valve again.
> > > I'm not sure if that ± 1.5 C is the intended accuracy of the adaptive mode or are these temperature swings caused by adaptive mode being too slow to react to temperature changes and letting the room overcool or overheat.
> > > Overall, that's much worse performance for me than the old ON/OFF mode with temperature accuracy set to -0.2 degree C.
> > 
> > 
> > Thank you for the detailed observations.
> > We have recorded the adaptive-mode behavior you described (temperature swings around the setpoint) and forwarded it to our firmware development team for further investigation.
> > You can also submit feedback through the SONOFF eWeLink app to provide additional data points.
> 
> If you are in contact with development team, please also think about an expose where you can force the calibration, because the smart mode causes the TRV to loose its calibration very quickly and 0% wont be "closed" within afew hours / days.
> 
> Thanks for your efford.

Thanks, this is a great point.

We’ll forward this to the firmware team for follow-up, and we really appreciate your continued testing and feedback.

CubeZ2mDeveloper:
Sorry for the delayed follow-up, and thank you all for your patience while we were away for the Chinese New Year holiday.
Thank you as well for all the detailed testing and reports.

In this PR, we simplified `smart_temperature_control` to remove ambiguity:
- Exposed as a boolean.
- Writes stay `false -> 0x00`, `true -> 0x02`.
- Reported `0x00` and `0x01` are both mapped to `false`; `0x02` is mapped to `true`.
- Description was updated accordingly.

We really appreciate the feedback from everyone. We will collect these adaptive-mode reports and forward them to our firmware development team.

If possible, please also report this through the SONOFF eWeLink app. Even short reports are useful, and multiple reports from different setups help us confirm patterns faster.

CubeZ2mDeveloper:
@Koenkk converter-side changes are complete from our side, and this is ready for your review.

Koenkk:
Thanks!

PR #11469, opened 2026-02-04
Added support for Zemismart signle gang dimmer white version and 2gang names display switch (Horizontal)

aryelevin:
Adds support for Zemismart 2 gang switch with names display (Horizontal version)

Both are existing devices just different variants so no need for new pictures.

I've checked both for a while for full functionality.


Koenkk:
Thanks!

PR #11475, opened 2026-02-05
Inovelli - fixing data type for internalTemperature

InovelliUSA:
Fixing data type so you can configure the reporting for this attribute.


Koenkk:
Thanks!

PR #11476, opened 2026-02-05
Change access level to ea.ALL for ctrlSeqeOfOper

svhelge:
The attribute is defined as follows:
ctrlSeqeOfOper: {ID: 0x001b, type: DataType.ENUM8, write: true, required: true, default: 4},


Koenkk:
Thanks!

PR #11477, opened 2026-02-05
Enhance EKO07259 device support

svhelge:
Added custom temperature measurement cluster and updated HVAC attributes for Schneider Electric devices. Added custom support for runningState (simulate runningState based on reported power).


Koenkk:
Thanks!

PR #11478, opened 2026-02-06
fix: Sinope TH1123ZB-G2 and TH1124ZB-G2: swap sensing and off values for backlight dimming modes Koenkk/zigbee2mqtt#29828

lemoinem:
Sorry, but when reviewing the code for the last PR (Koenkk/zigbee-herdsman-converters#11454), I noticed the name of the property for the meta check in the tzLocal converter hasn't been adjusted.

I suppose there is some missing type check...

I checked with the Z2M Edge addon/app, and indeed, it doesn't work.

Anyway, here is the fix for that. Again, sorry for the back-and-forth.

(Hopefully, final fix for Koenkk/zigbee2mqtt#29828)

### DIFF
--- a/src/devices/sinope.ts
+++ b/src/devices/sinope.ts
@@ -269,7 +269,7 @@ const tzLocal = {
     backlight_autodim: {
         key: ["backlight_auto_dim"],
         convertSet: async (entity, key, value, meta) => {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11437, opened 2026-02-01
Refactor WS90 weather station calculations and converters

danielpetrovic:
## Summary

This PR enhances the Shelly WS90 Weather Station with calculated weather values derived from the sensor data. All calculations run locally on the coordinator without requiring external services.

### Calculated Sensors
- **dew_point** - Magnus formula calculation from temperature and humidity
- **wind_chill** - Effective temperature when cold and windy (Environment Canada formula)
- **humidex** - Canadian humidity index for hot conditions
- **apparent_temperature** - Combined "feels like" temperature using wind chill or humidex as appropriate
- **heat_stress** - Percentage-based heat stress indicator (0-100%)
- **rain_rate** - Precipitation intensity in mm/hour calculated from consecutive readings
- **pressure_trend** - Barometric pressure change rate in hPa/hour
- **weather_condition** - HA-compatible weather condition string derived from all sensors

### Weather Condition Logic
Determines condition based on:
- Illuminance for day/night and cloud cover detection
- Rain status + rain rate for precipitation (requires both to avoid stale cache issues)
- Temperature for snow detection (<1°C)
- Wind speed for windy conditions (>10 m/s)
- Pressure + pressure trend for hail/storm detection

Supported conditions: `sunny`, `clear-night`, `partlycloudy`, `cloudy`, `rainy`, `pouring`, `snowy`, `hail`, `windy`, `windy-variant`

## Technical Implementation

- Device state cache per IEEE address for cross-sensor calculations
- Pressure history buffer for trend calculation
- Rain rate calculation from consecutive precipitation readings with timestamps
- Custom `fromZigbee` converters that trigger calculated value updates on every sensor reading
- Proper handling of undefined values throughout all calculations

## Home Assistant Customization

The calculated sensors work out of the box, but for optimal Home Assistant integration (proper icons, device classes, and units), add this to your Zigbee2MQTT `configuration.yaml` under your device:

```yaml
devices:
  '0xYOUR_DEVICE_IEEE':
    friendly_name: Weather Station
    homeassistant:
      dew_point:
        device_class: temperature
        state_class: measurement
        icon: mdi:water-thermometer
      wind_chill:
        device_class: temperature
        state_class: measurement
        icon: mdi:snowflake-thermometer
      humidex:
        device_class: temperature
        state_class: measurement
        icon: mdi:sun-thermometer
      heat_stress:
        state_class: measurement
        unit_of_measurement: '%'
        icon: mdi:heat-wave
      apparent_temperature:
        device_class: temperature
[...]

PR #11482, opened 2026-02-06
fix(shelly): Fix WS90 pressure, rain_rate, and wind_chill issues

danielpetrovic:
Fixes three bugs in the Shelly WS90 weather station after PR #11437:

1. **Pressure 10x too high** (9,910 hPa instead of ~991 hPa)
   - Root cause: Missing `/10` division in pressure converter
   - Fix: Added `const pressure = msg.data.measuredValue / 10;`

2. **rain_rate showing "Unknown"**
   - Root cause: rain_rate only included when not null
   - Fix: Always include rain_rate, convert null to 0

3. **wind_chill showing "Unknown"**
   - Root cause: Converter conflict between m.numeric() and ws90CalculatedValues()
   - Both converters handled the same clusters, causing m.numeric() to overwrite 
     calculated values when running after ws90CalculatedValues()
   - Fix: Changed ALL ws90CalculatedValues() converters to return ONLY calculated 
     values (dew_point, wind_chill, humidex, rain_rate, etc.), letting m.temperature(), 
     m.humidity(), m.pressure(), m.illuminance(), m.numeric(), m.binary() handle 
     base sensor values exclusively

Architecture now follows clean separation of concerns:
- m.* functions: Handle base sensor values only
- ws90CalculatedValues(): Handle calculated values only
- No overlap, no conflicts

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -62,20 +62,20 @@ function getWS90Meta(device: Zh.Device): WS90Meta {
 /**
  * Calculate dew point using Magnus formula
  */
-function calculateDewPoint(T: number | undefined, RH: number | undefined): number | null {
-    if (T === undefined || RH === undefined || RH <= 0) return null;
+function calculateDewPoint(T: number | undefined, Rh: number | undefined): number | null {
+    if (T === undefined || Rh === undefined || Rh <= 0) return null;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11484, opened 2026-02-06
Update Leedarson LED_GU10_OWDT ZM350STW1TCF

jharaque:
Correction to re-enable brightness and colour temperature controls.
Plus: transition times for switching on and off


Koenkk:
What is the actual fix here? I propose to integrate that into [light_onoff_brightness](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/converters/toZigbee.ts#L1215C14-L1215C36)

PR #11485, opened 2026-02-06
fix: Add missing attribute definitions to WS90 custom clusters

danielpetrovic:
Fixes wind_chill and other calculated values not working correctly after PR #11437.

PR #11437 accidentally removed the attribute definitions from the three custom clusters (shellyWS90Wind, shellyWS90UV, shellyWS90Rain), leaving them with empty `attributes: {}` objects. Without these definitions, zigbee-herdsman cannot map attribute IDs to names in msg.data, causing the ws90CalculatedValues() converters to fail accessing wind_speed, uv_index, and precipitation values.

This prevented wind_chill and other calculated sensors from working. This PR restores the attribute definitions.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -62,20 +62,20 @@ function getWS90Meta(device: Zh.Device): WS90Meta {
 /**
  * Calculate dew point using Magnus formula
  */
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11488, opened 2026-02-07
fix: generated definition inputs/outputs issues

Nerivec:
- fix name having duplicate endpoint ID appended
- add tests for better coverage of various inputs/outputs generated definitions

@Koenkk do we have a problem with `toZigbee` having multiple instances of the same logic under the same `key`? (On Z2M side it's a `find`, so first entry wins?)


Koenkk:
Thanks!

> @Koenkk do we have a problem with toZigbee having multiple instances of the same logic under the same key? (On Z2M side it's a find, so first entry wins?)

Yes that's indeed the behaviour, do you see any issues with it for this case?

Nerivec:
I don't.
Might want to optimize this out though, the behavior of some modern extends with the behavior of the generator are clashing a bit, especially around the multi-endpoint logic.

PR #11489, opened 2026-02-07
Add Tuya TS0502B light controller support

z-master42:
Merctor Ikkü moved from mercator.ts to tuya.ts. Incidentally, all mercator.ts devices are Tuya devices and should therefore be in tuya.ts.

Whitelabel for LUUMR Smart LED C35 matt E14 4.2 W added.

See #11490


Koenkk:
Thanks!

PR #11490, opened 2026-02-07
Remove SMCL01-ZB device entry from mercator.ts

z-master42:
Removed Mercator device entry for SMCL01-ZB because it is a Tuya device. I moved it to tuya.ts and added another whitelabel.

See #11489


Koenkk:
Thanks!

PR #11498, opened 2026-02-09
fix(WS90): Fix timing bugs in pressure trend and rain rate calculations

danielpetrovic:
## Summary
- Fixed timing bug in calculatePressureTrend() and calculateRainRate() where history timestamp was updated before the minimum interval check, causing the interval to never be reached when sensors report frequently (every 9 seconds)
- Changed pressure trend interval from 5 minutes to 1 hour for meteorologically meaningful readings
- Changed rain rate interval from 30 seconds to 1 minute for more stable readings

## What was broken
The history timestamp kept advancing on every sensor report, preventing the minimum interval from ever being reached. This caused:
- pressure_trend: always showed 0 hPa/h despite actual pressure changes
- rain_rate: unreliable readings

## Fix
Moved the history update to after the interval check in both functions, so the reference point only advances when a trend/rate is actually calculated.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -62,20 +62,20 @@ function getWS90Meta(device: Zh.Device): WS90Meta {
 /**
  * Calculate dew point using Magnus formula
  */
-function calculateDewPoint(T: number | undefined, RH: number | undefined): number | null {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11499, opened 2026-02-09
fix: Develco SIRZB-111: fix wrong siren modes and stop not working

snazzybean:
Closes:
- https://github.com/Koenkk/zigbee2mqtt/issues/30583
- https://github.com/Koenkk/zigbee2mqtt/issues/30778

## Root Cause

PR #10994 added SIRZB-111 to the reversed/ZCL bit layout list alongside SIRZB-110, assuming both devices use the same startwarninginfo byte encoding. They don't:

- SIRZB-110 (reversed): info = mode + (strobe << 4) + (level << 6)
- SIRZB-111 (generic): info = (mode << 4) + (strobe << 2) + level

With the wrong layout, every mode/level combination was garbled — e.g. a stop command with default strobe=true, level=medium produced startwarninginfo=80, which the SIRZB-111 interprets as fire_panic instead of stopping.

Additionally, the warning_simple converter (key: alarm) had no device-specific handling for either SIRZB model, causing the alarm OFF command to send startwarninginfo=192 — which neither device interprets as stop.

## Changes

warning converter:
- Remove SIRZB-111 from the reversed bit layout list (it uses the generic layout)
- Add stop-mode override for SIRZB-110/111: when mode=stop, force level=low and strobe=false to ensure an all-zero info byte, which both devices require to reliably stop

warning_simple converter:
- Add explicit branches for SIRZB-110 (reversed layout) and SIRZB-111 (generic layout) with correct ON/OFF values

## Test

- Tested functionally on a live SIRZB-111 device via Zigbee2MQTT (HA addon)
- Built the project, deployed dist/converters/toZigbee.js into the Z2M addon container, verified:
  - Warning modes (burglar, fire, emergency) produce correct sounds
  - mode=stop reliably stops the siren
  - Alarm START/OFF toggle works correctly
  - No regression of [#28847](https://github.com/Koenkk/zigbee2mqtt/issues/28847)

---
External converter for context: https://github.com/Koenkk/zigbee2mqtt/issues/30583#issuecomment-3864862318

### DIFF
--- a/src/converters/toZigbee.ts
+++ b/src/converters/toZigbee.ts
@@ -587,7 +587,14 @@ export const warning: Tz.Converter = {
         let info;
         // https://github.com/Koenkk/zigbee2mqtt/issues/8310 some devices require the info to be reversed.
         if (Array.isArray(meta.mapped)) throw new Error("Not supported for groups");
-        if (["SIRZB-110", "SIRZB-111", "SRAC-23B-ZBSR", "AV2010/29A", "AV2010/24A"].includes(meta.mapped.model)) {
+        // SIRZB-110/111 require all-zero info byte to reliably stop the siren.
+        if (values.mode === "stop" && ["SIRZB-110", "SIRZB-111"].includes(meta.mapped.model)) {
+            // @ts-expect-error ignore
+            if (value.level == null) values.level = "low";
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11500, opened 2026-02-09
M9Pro support for additional attributes not controlable with datapoints

l1berat0r:
Added:
- M9Pro support for readding attributes not exposed with datapoints: brightness, color_temp, thermostat, current_heating_setpoint, fan_mode.
- M9Pro support for binding certain endpoints to group_ids with speciffic command.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -895,6 +895,29 @@ const tzLocal = {
         },
     } satisfies Tz.Converter,
     // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
+    TS0601_smart_light_group_id: {
+        key: ["group_id"],
+        convertSet: async (entity, key, value, meta) => {
+            // The device uses custom group command known from miboxer switches to bind to a group.
+
+            const zone_map = Object.fromEntries(
+                Object.keys(meta.state)
+                    .filter((key) => key.startsWith("group_id"))
+                    .map((k) => [k.replace("group_id_l", ""), meta.state[k]]),
+            );
+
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -1434,6 +1457,77 @@ const fzLocal = {
...
+                return;
+            }
+
+            return {
+                [`color_temp_l${endppointN}`]: msg.data?.colortemp,
+            };
+        },
+    } as Fz.Converter<"lightingColorCtrl", undefined, ["commandMoveToColorTemp"]>,
+    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
+    TS0601_smart_thermostat_command: {
```
Koenkk, line 1503: Please move this to `meta.tuyaDatapoints` with a custom value converter.

l1berat0r: Hi,

maybe I got it wrong, but this device sends the thermostat state using "commandDataRequest" command instead of "commandDataResponse". Datapoints it uses for thermostat state change actions are already defined in meta.tuyaDatapoints and those actually translate to different fields.
My understanding is that meta.tuyaDatapoints support parsing data that are sent from device with "commandDataResponse" and this particular device uses the manuSpecificTuya commands in some weird that does not follow the protocol, but maybe I misunderstood how this works. Could you please point me in the right direction and let me know if my understanding is accurate?

I used the same approach as I saw in [here](https://github.com/l1berat0r/zigbee-herdsman-converters/blob/0c34919e89fd465e1ce5fcd94a085a218467064f/src/devices/tuya.ts#L1388) for a different device.

Koenkk:
Thanks!

PR #11504, opened 2026-02-10
Add per-phase energy for SPM02 and SDM01 (Only app 11+)

script0803:

### DIFF
--- a/src/devices/bituo_technik.ts
+++ b/src/devices/bituo_technik.ts
@@ -47,6 +47,36 @@ const bituo_fz = {
             return result;
         },
     } satisfies Fz.Converter<"haElectricalMeasurement", undefined, ["attributeReport", "readResponse"]>,
+    // Unlike the standard converter function, so add a custom converter
+    phase_energy: {
+        cluster: "seMetering",
+        type: ["attributeReport", "readResponse"],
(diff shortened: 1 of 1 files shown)


src/devices/bituo_technik.ts
```diff
@@ -47,6 +47,36 @@ const bituo_fz = {
             return result;
         },
     } satisfies Fz.Converter<"haElectricalMeasurement", undefined, ["attributeReport", "readResponse"]>,
+    // Unlike the standard converter function, so add a custom converter
```
Koenkk, line 50: What is not supported in the standard converter?

script0803: In the standard converter, there are only energy_tier_1, energy_tier_2, energy_tier_3, energy_tier_4, and produced_energy_tier_1, produced_energy_tier_2. We need energy phase abc, produced energy phase abc. Moreover, we use this attribute as per-phase energy instead of tiered electricity charges. Therefore, we also need to make changes to the names used for expose.

Koenkk:
Thanks!

bobo-cher:
This works great.

@script0803 Thank you very much for the incredibly quick implementation of my request.

@Koenkk Thank you for the fast review and for merging the PR.

Amazing support from both of you — really appreciated!

PR #11508, opened 2026-02-10
fix(WS90): Persist pressure_trend between updates and reduce interval to 30min

danielpetrovic:
## Summary
- Fixed pressure_trend showing gaps/spikes instead of continuous values by persisting the last calculated trend between updates
- Reduced pressure_trend update interval from 1 hour to 30 minutes for more responsive weather forecasting

## What was broken
After the previous timing bug fix, pressure_trend would calculate correctly every hour but return `null` for all sensor reports in between. This caused the sensor to show sporadic spikes on graphs instead of a stable continuous reading, making it unsuitable for weather forecasting.

## Fix
- When the 30-minute interval hasn't passed, return the previously calculated trend value from `device.meta.ws90.state.pressure_trend` instead of `null`
- Changed interval from 3600000ms (1 hour) to 1800000ms (30 minutes) for better responsiveness while maintaining meteorological accuracy

## Result
Pressure trend now shows stable, continuous values that update every 30 minutes, properly supporting weather condition forecasting and forecast generation.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -62,20 +62,20 @@ function getWS90Meta(device: Zh.Device): WS90Meta {
 /**
  * Calculate dew point using Magnus formula
  */
-function calculateDewPoint(T: number | undefined, RH: number | undefined): number | null {
-    if (T === undefined || RH === undefined || RH <= 0) return null;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11513, opened 2026-02-10
Add Moes ZM-105B-M fingerprint _TZE200_bxoo2swd for existing TS0601_dimmer_4

jonamb:
I've added the fingerprint of my unit _TZE200_bxoo2swd as another fingerprint for Moes ZM-105B-M, which is a whitelabel TS0601_dimmer_4.

I've tested it via external converter and the functions already defined for ZM-105B-B worked exactly as expected for this fingerprint.

Image for ZM-105B-M already exists, my unit only has a very slightly different arrangement of text but is obviously the same model so I'd skip the extra picture, also since the model name is the same.

Hope everything else is there as needed!

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -4193,7 +4193,7 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11516, opened 2026-02-11
Refactor Tuya water valve configuration _TZE284_vuwtqx0t

Liionboy:
Updated the Tuya water valve configuration to include new features and corrected unit measurements.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1767,7 +1767,7 @@ export const definitions: DefinitionWithExtend[] = [
             tuyaDatapoints: [
                 [1, "water_consumed", tuya.valueConverter.divideBy1000],
                 [2, "month_consumption", tuya.valueConverter.waterConsumption],
-                [3, "daily_consumption", "month_consumption", tuya.valueConverter.waterConsumption],
+                [3, "daily_consumption", tuya.valueConverter.waterConsumption],
                 [
                     4,
                     "report_period",


src/devices/tuya.ts
```diff
@@ -1746,22 +1746,28 @@ export const definitions: DefinitionWithExtend[] = [
...
-            // Flow rate sensor
-            e.numeric("flow_rate", ea.STATE).withUnit("L/h").withDescription("Instantaneous water flow rate"),
+            e.numeric("water_consumed", ea.STATE).withUnit("m³").withDescription("Total water consumption").withValueMin(0).withValueStep(0.001),
+            e.numeric("month_consumption", ea.STATE).withUnit("m³").withDescription("Month consumption").withValueMin(0).withValueStep(0.001),
+            e.numeric("daily_consumption", ea.STATE).withUnit("m³").withDescription("Daily consumption").withValueMin(0).withValueStep(0.001),
+            e.numeric("flow_rate", ea.STATE).withUnit("m³/h").withDescription("Instantaneous water flow rate").withValueMin(0).withValueStep(0.001),
             e.binary("auto_clean", ea.STATE_SET, "ON", "OFF").withDescription("Auto clean"),
             e.temperature(),
             e.battery_voltage(),
+            // Senzori adăugați de pe modelul tău:
```
Koenkk: English only please

Liionboy: Done

src/devices/tuya.ts
```diff
@@ -1746,22 +1746,26 @@ export const definitions: DefinitionWithExtend[] = [
...
-            e.numeric("water_consumed", ea.STATE).withUnit("L").withDescription("Total water consumption").withValueMin(0).withValueStep(0.001),
-            e.numeric("month_consumption", ea.STATE).withUnit("L").withDescription("month consumption").withValueMin(0).withValueStep(0.001),
-            e.numeric("daily_consumption", ea.STATE).withUnit("L").withDescription("daily consumption").withValueMin(0).withValueStep(0.001),
-            // //  I can't seem to get this to work - when i try to change the value i get an error in zigbee2mqtt, stuck on 1 Hour
-            // e
-            //     .enum("report_period", ea.STATE_SET, ["1h", "2h", "3h", "4h", "6h", "8h", "12h", "24h"])
-            //     .withDescription("Report period"),
-            // Flow rate sensor
-            e.numeric("flow_rate", ea.STATE).withUnit("L/h").withDescription("Instantaneous water flow rate"),
+            e.numeric("water_consumed", ea.STATE).withUnit("m³").withDescription("Total water consumption").withValueMin(0).withValueStep(0.001),
```
Koenkk, line 1749: The unit is changed from L to M3, can you keep it L because it's breaking, same holds for the flow rate and daily/month consumption

src/devices/tuya.ts
```diff
@@ -1746,22 +1746,26 @@ export const definitions: DefinitionWithExtend[] = [
...
+            e.numeric("month_consumption", ea.STATE).withUnit("m³").withDescription("Month consumption").withValueMin(0).withValueStep(0.001),
+            e.numeric("daily_consumption", ea.STATE).withUnit("m³").withDescription("Daily consumption").withValueMin(0).withValueStep(0.001),
+            e.numeric("flow_rate", ea.STATE).withUnit("m³/h").withDescription("Instantaneous water flow rate").withValueMin(0).withValueStep(0.001),
             e.binary("auto_clean", ea.STATE_SET, "ON", "OFF").withDescription("Auto clean"),
             e.temperature(),
             e.battery_voltage(),
+            e
+                .numeric("reverse_water_consumed", ea.STATE)
+                .withUnit("m³")
```
Koenkk, line 1758: Please also make this litres

Liionboy: But they need to be on M3 like on data sheet from who made this, they send me this 2 unit's exactly for this

Image

src/devices/tuya.ts
Liionboy: They need to be on M3 not on Liters:

Image

And is the same device like this **`TS060 _TZE284_ajlu4cud (Ultrasonic water meter)`** only have a Valve inside.

Liionboy:
so nice i break the code, because always there are some comment, i will close this and i am done with this, never do a Pull Request for support a new device

src/devices/tuya.ts
Liionboy: And the problem was with who made that support for device because he need to use M3 from begin because this is what you saw on Tuya App not L

src/devices/tuya.ts
Liionboy: If the device has been sending values in liters and you change to m³, the historical data in Home Assistant will be wrong.

Example:

Now: 1234 liters → displayed as 1234 L
After change: 1234 → displayed as 1.234 m³
The old value of 1234 liters will become 1.234 m³ - a 1000x difference!

Solutions:

Delete old data from Home Assistant (History / Recorder) and then change to m³

Keep liters - don't change anything

Create a new sensor in m³ and use that instead, without deleting old data

If you want to fix it, the easiest way is to delete the data from the Recorder for that sensor and then modify the converter.

src/devices/tuya.ts
Koenkk: But this will be a breaking change for existing users so I propose to keep it L (for new devices we can use m3)

Liionboy: Let me ask the manufacturer about this, for this they send me device, they saw is a problem in Zigbee2mqtt, tommorow i will ask then and then i will see what they want, but i think is simple for user , like at me i need to give in M3 not in liters and then to pay invoice in M3

Liionboy: If we can't use in m3 i will remove this Pull Request and tell them to send convertor to user who buy device

Liionboy: And if i will modify in Liters need to modify that function to be in Liters i think, manufacture don't answer i think in china is holiday and in march they will be back

Liionboy: > But this will be a breaking change for existing users so I propose to keep it L (for new devices we can use m3)

They answer me:

```
 [09:07, 2/22/2026] We want to be able to use m3
[09:08, 2/22/2026]  Because then the data can be consistent with the tuya app.
```

Is you decision now, if can't implement in m3 i will tell them to sell with convertor.

Thanks

Koenkk: Please keep it litres, this is not worth a breaking change (and breaking tons of automations users built on top of it)

Liionboy: Ok, thanks, i will delete pull request and tell then you didn't accept, thanks and have a nice work and day

PR #11519, opened 2026-02-11
fix(WS90): use string attribute names for custom cluster sensors

danielpetrovic:
Fix wind_speed, wind_direction, gust_speed, uv_index, rain_status, and precipitation sensors not updating on Shelly WS90 weather station.

When m.deviceAddCustomCluster() defines named attributes (e.g., `{wind_speed: {ID: 0x0000, type: UINT16}}`), zigbee-herdsman maps attribute IDs to string names in msg.data. However, m.numeric() with `attribute: {ID: 0x0000}` looks up the numeric ID (0) as the key, which doesn't match the string key ("wind_speed") in msg.data.

Changing attribute from `{ID: 0x0000, type: ...}` to `"wind_speed"` (string) makes m.numeric() look up the correct string key, consistent with how standard cluster sensors (temperature, humidity, etc.) use string attribute names.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -89,20 +89,20 @@ function getWS90Meta(device: Zh.Device): WS90Meta {
 /**
  * Calculate dew point using Magnus formula
  */
-function calculateDewPoint(T: number | undefined, RH: number | undefined): number | null {
-    if (T === undefined || RH === undefined || RH <= 0) return null;
+function calculateDewPoint(T: number | undefined, Rh: number | undefined): number | null {
+    if (T === undefined || Rh === undefined || Rh <= 0) return null;
     const a = 17.27;
     const b = 237.7;
-    const alpha = (a * T) / (b + T) + Math.log(RH / 100);
-    return Math.round((b * alpha) / (a - alpha) * 10) / 10;
+    const alpha = (a * T) / (b + T) + Math.log(Rh / 100);
+    return Math.round(((b * alpha) / (a - alpha)) * 10) / 10;
 }
 
 /**
  * Calculate humidex (Canadian heat index)
  */
-function calculateHumidex(T: number | undefined, RH: number | undefined): number | null {
-    if (T === undefined || RH === undefined) return null;
-    const dewPoint = calculateDewPoint(T, RH);
+function calculateHumidex(T: number | undefined, Rh: number | undefined): number | null {
+    if (T === undefined || Rh === undefined) return null;
+    const dewPoint = calculateDewPoint(T, Rh);
     if (dewPoint === null) return null;
     const ee = 6.11 * Math.exp(5417.753 * (1 / 273.15 - 1 / (273.15 + dewPoint)));
     return Math.round((T + 0.5555 * (ee - 10)) * 10) / 10;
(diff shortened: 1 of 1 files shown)


Koenkk:
> However, m.numeric() with attribute: {ID: 0x0000} looks up the numeric ID (0) as the key, which doesn't match the string key ("wind_speed") in msg.data.

I don't understand why this doesn't work, could you provide the debug log when this attribute is reported with the current code?

danielpetrovic:
Debug log with current code (named attributes in cluster definition). Showing the sequence for a wind_speed attribute report:

1. Raw ZCL frame arrives with numeric attribute ID 0:
```
zh:controller: Received payload: clusterID=64513, address=32086, ...
frame={"header":...,"payload":[{"attrId":0,"dataType":33,"attrData":5}],...}

```
2. zigbee-herdsman maps attrId 0 → "wind_speed" because of the named cluster definition, so msg.data has a string key:
```
z2m: Received Zigbee message from 'Roof Terrace Sensor', type 'attributeReport',
cluster 'shellyWS90Wind', data '{"wind_speed":5}' from endpoint 1

```
3. m.numeric() does attributeKey = attribute.ID → 0, then checks 0 in {"wind_speed": 5} → false, so it returns nothing. MQTT publishes with stale cached value (2.7 instead of 0.5):
```
z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Roof Terrace Sensor',
payload '{"wind_speed":2.7,...}'

```
The wind_speed value stays at 2.7 (from startup cache) despite receiving a new report with value 5 (= 0.5 m/s after scale). The calculated converter (ws90CalculatedValues) does find data.wind_speed by string key, so wind_chill/apparent_temperature still update correctly - only the base values handled by m.numeric() are broken.

Koenkk:
It only does that when `attribute` is not a string: `const attributeKey = isString(attribute) ? attribute : attribute.ID;`, can you pass a string instead? It should be `wind_speed`

danielpetrovic:
I tried attribute: "wind_speed" first, but it fails TypeScript compilation:

`src/devices/shelly.ts(1092,17): error TS2322: Type 'string' is not assignable to type '{ ID: number; type: number; }'.`

It seems the type for custom cluster attributes only allows {ID, type} objects, not string names. Should we also keep the named attribute definitions in the cluster (attributes: {wind_speed: {ID: 0x0000, type: ...}}) so the generic type resolves the string? Or is there another way to make the types accept string attribute names for custom clusters?

Koenkk:
It should work, here is a working example from SONOFF:
- https://github.com/Koenkk/zigbee-herdsman-converters/blob/e410903af3516701e48d4a851cda7488f70f928d/src/devices/sonoff.ts#L1643
- https://github.com/Koenkk/zigbee-herdsman-converters/blob/e410903af3516701e48d4a851cda7488f70f928d/src/devices/sonoff.ts#L1661

danielpetrovic:
Thanks for the pointer! You're right, I was missing the custom cluster interface definitions and generic type parameters on m.numeric()/m.binary() following the same pattern as SONOFF (e.g. Sonoff SNZB02P).

Updated the PR to:

Add interfaces ShellyWS90Wind, ShellyWS90UV, ShellyWS90Rain
Use generic type params: m.numeric<"shellyWS90Wind", ShellyWS90Wind>({attribute: "windSpeed", ...})
Rename cluster attributes to camelCase to satisfy biome naming conventions
Keep name as snake_case (unchanged MQTT/entity names)
Also updated the ws90CalculatedValues() converters to read from the camelCase msg.data keys accordingly.

Koenkk:
Thanks!

PR #11524, opened 2026-02-11
Model SR-ZG2868EK7-CCT - FIX - Add commandsScenes() requirement

ctournayre:
Two buttons use genScenes/commandRecall (sceneid 1 and 2), requiring commandsScenes()


ctournayre:
Sorry for the mess, 1st time adding a new device. I didn't see that 2 buttons where missing, here is the fix.

Koenkk:
Thanks!

PR #11526, opened 2026-02-12
Add Nova Digital SA-4W (_TZ3210_iymfxdis)

alanmgomes:
- This specific manufacturer (`_TZ3210_iymfxdis`) is a bit of a "special snowflake". 
- While it's a TS0004, it rejects the standard Zigbee `startUpOnOff` attribute and the common `powerOnBehavior2` modern extension.
- I've implemented a manual `enumLookup` for `power_on_behavior` using the Tuya cluster `0x4002`, which is the only way this device agrees to remember its state after a power outage.
- Tested and working: On/Off (physical & virtual), Backlight mode, and Power-on behavior.

They say Zigbee is a standard, but Tuya manufacturers seem to treat "standards" like New Year's resolutions: they're great to talk about, but nobody actually follows them!

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4823

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -16948,7 +16948,7 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        fingerprint: [{modelID: "TS0004", manufacturerName: "_TZ3210_iymfxdis"}],
+        fingerprint: tuya.fingerprint("TS0004", ["_TZ3210_iymfxdis"]),
         model: "SA-4W",
         vendor: "Nova Digital",
         description: "Safira smart light switch - 4 gang",


src/devices/tuya.ts
```diff
@@ -16964,6 +16964,35 @@ export const definitions: DefinitionWithExtend[] = [
...
+        model: "SA-4W",
+        vendor: "Nova Digital",
+        description: "Safira smart light switch - 4 gang",
+        extend: [
+            m.deviceEndpoints({endpoints: {l1: 1, l2: 2, l3: 3, l4: 4}}),
+            tuya.modernExtend.tuyaOnOff({
+                backlightModeOffOn: true,
+                endpoints: ["l1", "l2", "l3", "l4"],
+            }),
+            m.enumLookup({
```
Koenkk: I think we can integrate this into https://github.com/Koenkk/zigbee-herdsman-converters/blob/079d933eee1afec321a5bf23871a7134fb1d9828/src/lib/tuya.ts#L2934 as `powerOnBehaviour3` and it can then be a conditional option like the others, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/079d933eee1afec321a5bf23871a7134fb1d9828/src/devices/tuya.ts#L6027

alanmgomes: I’ve updated the PR to include a new powerOnBehavior3 option within tuyaOnOff as suggested.

To keep the TypeScript compiler happy and handle multi-gang switches properly, I’m iterating through the endpoints list to pass them individually to endpointName. 
I also updated the SA-4W definition to use this new approach.

Thanks!

Koenkk:
Thanks!

PR #10389, opened 2025-10-17
Fix wrong energy value reported by Sinopé TH1123ZB-G2

AlexandreJubert:
The Sinopé TH1123ZB-G2 is reporting its energy in Wh, and not in KWh, leaading to the wrong energy value being reported.

For example here's the state of a relatively new thermostat:
```
{
    "current": 0,
    "energy": 135296.01,
    "linkquality": 200,
    "local_temperature": 19.29,
    "occupied_heating_setpoint": 17.5,
    "pi_heating_demand": 0,
    "power": 0,
    "running_state": "idle",
    "second_display_mode": "outdoor temp",
    "system_mode": "off",
    "thermostat_outdoor_temperature": 2,
    "unoccupied_heating_setpoint": 15,
    "backlight_auto_dim": null,
    "enable_outdoor_temperature": null,
    "keypad_lockout": null,
    "main_cycle_output": null,
    "outdoor_temperature_timeout": null,
    "temperature_display_mode": null,
    "thermostat_occupancy": null,
    "time_format": null,
    "voltage": null
}
```

As you can see, I've not consumed 10% of a nuclear reactor output on one heater.

I've also opened an issue on the main repository of z2m here: https://github.com/Koenkk/zigbee2mqtt/issues/26740

I've not been able to test it locally, I'd appreciate any help or guidance as to make sure this fix works properly, I've run all tests and fixes on codespace, everything is green.

PR #11527, opened 2026-02-12
fix insufisant space error on sinope TH1123ZB-G2

cloutierjo:
This INSUFFICIENT_SPACE error was happening again after a recent firmware update
```
z2m: Failed to configure 'sinope thermostat salon', attempt 4 (Error: ZCL command 0x943469fffe283c1c/1 haElectricalMeasurement.configReport([{"minimumReportInterval":10,"maximumReportInterval":65000,"reportableChange":5,"attribute":"activePower"},{"minimumReportInterval":10,"maximumReportInterval":65000,"reportableChange":50,"attribute":"rmsCurrent"},{"minimumReportInterval":10,"maximumReportInterval":65000,"reportableChange":50,"attribute":"rmsVoltage"}], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"reservedBits":0,"writeUndiv":false}) failed (Status 'INSUFFICIENT_SPACE') at Endpoint.checkStatus (/app/node_modules/.pnpm/zigbee-herdsman@8.0.1/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:461:28) at Endpoint.zclCommand (/app/node_modules/.pnpm/zigbee-herdsman@8.0.1/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:1192:26) at Endpoint.configureReporting (/app/node_modules/.pnpm/zigbee-herdsman@8.0.1/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:853:9) at setupAttributes (/app/node_modules/.pnpm/zigbee-herdsman-converters@25.98.0/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:183:17) at result.configure (/app/node_modules/.pnpm/zigbee-herdsman-converters@25.98.0/node_modules/zigbee-herdsman-converters/src/lib/modernExtend.ts:2336:29) at Object.configure (/app/node_modules/.pnpm/zigbee-herdsman-converters@25.98.0/node_modules/zigbee-herdsman-converters/src/index.ts:365:21) at Configure.configure (/app/lib/extension/configure.ts:123:13) at Configure.onMQTTMessage (/app/lib/extension/configure.ts:46:25) at EventEmitter.wrappedCallback (/app/lib/eventBus.ts:252:17))
```

I reworked the reporting, ensuring we stay within the limit of the device in its latest revision and making sure we have the electric measurement

The removed systemMode and temperature were not actually reporting anything on those devices, as contrary to the comment, those line didn't "disable" them, but instead used memory space on the device.

I also validate the proper working of the TH1124ZB and TH1123ZB which didn't require any change so left unchanged. I don't have any of the other device, in particular the TH1124ZB-G2 which has similar memory comments in the code, but as I cannot test it, I preferred to leave it untouched.

### DIFF
--- a/src/devices/sinope.ts
+++ b/src/devices/sinope.ts
@@ -974,7 +974,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "TH1124ZB-G2",
         vendor: "Sinopé",
         description: "Zigbee line volt thermostat",
-        extend: [m.electricityMeter()],
+        extend: [m.electricityMeter({energy: {divisor: 1000, multiplier: 1}})],
         meta: {sinopeAlternateBacklightAutoDim: true},
         fromZigbee: [fzLocal.thermostat, fzLocal.sinope, fz.hvac_user_interface, fz.ignore_temperature_report],
         toZigbee: [
@@ -1057,8 +1057,12 @@ export const definitions: DefinitionWithExtend[] = [
                 "hvacUserInterfaceCfg",
                 "msTemperatureMeasurement",
                 "manuSpecificSinope",
+                "seMetering",
+                "haElectricalMeasurement",
             ];
-            await reporting.bind(endpoint, coordinatorEndpoint, binds); // This G2 version has limited memory space
+            // This G2 version has limited memory space
+            // from experimenting we can configure a maximum of 8 reporting entities
+            await reporting.bind(endpoint, coordinatorEndpoint, binds);
             const thermostatDate = new Date();
             const thermostatTimeSec = thermostatDate.getTime() / 1000;
             const thermostatTimezoneOffsetSec = thermostatDate.getTimezoneOffset() * 60;
@@ -1069,9 +1073,8 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.thermostatTemperature(endpoint);
             await reporting.thermostatPIHeatingDemand(endpoint);
             await reporting.thermostatOccupiedHeatingSetpoint(endpoint);
-            await reporting.thermostatSystemMode(endpoint);
+            await reporting.thermostatRunningState(endpoint, {min: 1, max: 0xffff});
 
-            await reporting.temperature(endpoint, {min: 1, max: 0xffff}); // Disable default reporting
             await endpoint.configureReporting("msTemperatureMeasurement", [
                 {
                     attribute: "tolerance",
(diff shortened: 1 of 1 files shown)


JoRouss:
The same error is present on the TH1124ZB-G2 on the latest firmware. Maybe you can replicate your changes to that model also?

I could test the changes if you point me to how I can do that.

The TH1124ZB-G2 also report x1000 the energy value as #10389 only made the changes for the TH1123ZB-G2.

cloutierjo:
I can add it, as far as testing it, you need to be able to recompile this project following the readme and most probably do the same in the zigbeemqtt project as well replacing the actual converter dependency with your locally built one. I personally have this running on a dedicated zigbee controller and just temporarily move my thermostat to it to ensure it work. 

Unfortunately i haven't been able to use an external converter that we can just drop-in in an existing installation for sinope device. Maybe someone else would have an easier way of testing that.

JoRouss:
Alright I spent some time this morning on this PR.

I can confirm the same changes need to be done on the TH1124ZB-G2. Including the 1000 summationDelivered divider that is already applied to the TH1123ZB-G2.

I was able to test both by generating two external_converters from your work. I just converted what you've done from .ts to .js by removing some references and type casts. I had to split both devices in 2 files since the external components need to apply to a single device.

[custom_TH1123ZB-G2.js](https://github.com/user-attachments/files/25271013/custom_TH1123ZB-G2.js)
[custom_TH1124ZB-G2.js](https://github.com/user-attachments/files/25271014/custom_TH1124ZB-G2.js)

I simply added them to HA in config/zigbee2mqtt/external_converters folder. They then automatically override the built in converters upon restarting z2m.

I still had to factory reset the thermostats and re-add them to z2m to get rid of the INSUFFICIENT_SPACE error.

The only thing that did not work is the values you set for minimumReportInterval, maximumReportInterval and reportableChange. They seem to all be completely ignored and are set with default values of 10, 65000, 100.

JoRouss:
Once you edit your PR for the changes on the TH1124ZB-G2, the following issues are going to be fixed:
fixes https://github.com/Koenkk/zigbee2mqtt/issues/30365
fixes https://github.com/Koenkk/zigbee2mqtt/issues/25272

Thank you so much, I just got rid of my ZHA network and run on z2m exclusively!

cloutierjo:
I pushed the modification for the TH1124ZB-G2 thanks for validating it on your side. I'll validate the reporting value automatically setted, but I believe there were right following our discussion, I remember testing them and they reported according to the value in the code, I'll still double-check when I have access to my device.

Also as you did notice, **It is required to factory reset the thermostat** to reset the preexisting reporting, it seems the device doesn't support unbinding them. So press up and down button for 10 sec then re-pair it with z2m.

cloutierjo:
you were right, when cleaning up at the end I removed the default reporting prevention that have a bit high threshold it is now fixed:

image

This should be ready for proper review and merge if approved. Thanks @JoRouss for the feedback and testing

Koenkk:
Thanks!

PR #11529, opened 2026-02-12
Add E40-230 Zigbee thermostat support

adorobis:
Added support for E40-230 Zigbee smart thermostat with various features including state control, temperature settings, and scheduling.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4829


src/devices/engo.ts
```diff
@@ -546,6 +546,142 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE204_glk6viwg"]),
+        model: "E40-230",
```
Koenkk: It seems the code is the same as `E40`, can you merge it with that? Use `tuya.whiteLabel` for proper detection (see `tuya.ts` for examples)

adorobis: I can see there is also option to indicate two ids for the `fingerprint:` element e.g. like for another Engo device:
```
fingerprint: tuya.fingerprint("TS0601", ["_TZE204_ca3i8m8p", "_TZE200_awnadkan"]),
        model: "EONE-230W",
        vendor: "ENGO",
```
Because it is the same vendor and device (just different variant of the device)  I guess this would be ok as well?

adorobis: Updated the pull request in line with above

adorobis: I have a different question though: is it possible to have OFF hvac mode instead of a separate switch that turns the thermostat on or off? That would be more in line with HA Climate entity behaviour but I'm not sure if it is a matter of the converter implementation or a "beauty" of those Tuya devices...

Koenkk: Yes that would be nice, for that to work:

- Change to `.withSystemMode(["heat", "cool", "off"], ea.STATE_SET)`
- Write a custom value converter for ` [2, "system_mode", tuya.valueConverterBasic.lookup({heat: tuya.enum(0), cool: tuya.enum(1)})],` which can handle the off mode.

adorobis: Cool. I've found similar config for one of the other Engo thermostats but have trouble replicating it to the external converter. Is there a good way on how to test it in my system before creating a pull request?

Koenkk: External converter is the way to go, what are the issues?

adorobis: Not anymore, I've figured it out. Now the pull request is there.

PR #11329, opened 2026-01-18
Enhance Nous smart power strip with new features

1tigris1:
Hi, could you please tell me why after the update some devices stopped working correctly, namely… what could be the reason? I had to rewrite the converter.
Here is the corrected converter.

const tuya = require('zigbee-herdsman-converters/lib/tuya');
const utils = require('zigbee-herdsman-converters/lib/utils');
const exposes = require('zigbee-herdsman-converters/lib/exposes');
const e = exposes.presets;
const ea = exposes.access;

module.exports = [
    {
        fingerprint: tuya.fingerprint('TS011F', ['_TZ3210_6cmeijtd']),

        model: 'A11Z',
        vendor: 'Nous',
        description: 'Smart power strip 3 gang with energy monitoring & countdown',

        endpoint: (device) => {
            return { l1: 1, l2: 2, l3: 3 };
        },

        meta: {
            multiEndpoint: true,
            multiEndpointSkip: ['energy', 'current', 'voltage', 'power'],
        },

        extend: [
            tuya.modernExtend.tuyaOnOff({
                electricalMeasurements: true,
                powerOutageMemory: true,
                onOffCountdown: true,
                indicatorMode: true,
                childLock: true,
                endpoints: ['l1', 'l2', 'l3'],
            }),
            tuya.modernExtend.electricityMeasurementPoll({
                metering: (device) => [100, 160, 192].includes(device.applicationVersion) || ["1.0.5"].includes(device.softwareBuildID),
            }),
        ],

        configure: async (device, coordinatorEndpoint) => {
            await tuya.configureMagicPacket(device, coordinatorEndpoint);
            const endpoint = device.getEndpoint(1);
            

            endpoint.saveClusterAttributeKeyValue('haElectricalMeasurement', {
                acCurrentDivisor: 1000,
                acCurrentMultiplier: 1,
                acVoltageDivisor: 10,
                acVoltageMultiplier: 1,
                acPowerDivisor: 10,
                acPowerMultiplier: 1,
            });
            endpoint.saveClusterAttributeKeyValue('seMetering', {
                divisor: 100,
                multiplier: 1,
            });
            
            utils.attachOutputCluster(device, 'genOta');
            device.save();
        },
    },
];

PR #11530, opened 2026-02-13
Nous A11Z: 3-channel power strip with PM

andrei-lazarov:

- fixes: https://github.com/Koenkk/zigbee2mqtt/issues/30799
- closes: https://github.com/Koenkk/zigbee-herdsman-converters/pull/11329
- picture already added

No polling necessary. Reporting & everything I added works.
I verified all energy values to be correct.

This part is needed, I encountered the invalid measurement:
> Wait 5 seconds before reporting a 0 value as this could be an invalid measurement.

### DIFF
--- a/src/devices/nous.ts
+++ b/src/devices/nous.ts
@@ -267,14 +267,19 @@ export const definitions: DefinitionWithExtend[] = [
             m.deviceEndpoints({endpoints: {l1: 1, l2: 2, l3: 3}}),
             tuya.clusters.addTuyaCommonPrivateCluster(),
             tuya.modernExtend.tuyaOnOff({
-                electricalMeasurements: true,
-                electricalMeasurementsFzConverter: tuya.fz.TS011F_electrical_measurement,
                 powerOnBehavior2: true,
                 onOffCountdown: true,
                 indicatorMode: true,
                 childLock: true,
                 switchTypeButton: true,
                 endpoints: ["l1", "l2", "l3"],
+            }),
+            m.electricityMeter({
+                current: {divisor: 1000, multiplier: 1},
+                voltage: {divisor: 1, multiplier: 1},
+                power: {divisor: 1, multiplier: 1},
+                energy: {divisor: 100, multiplier: 1},
+                fzElectricalMeasurement: tuya.fz.TS011F_electrical_measurement,
             }),
             m.identify(),
         ],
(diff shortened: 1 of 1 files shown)


src/devices/nous.ts
```diff
@@ -259,15 +259,51 @@ export const definitions: DefinitionWithExtend[] = [
...
-        fingerprint: [{modelID: "TS011F", manufacturerName: "_TZ3210_6cmeijtd"}],
+        fingerprint: tuya.fingerprint("TS011F", ["_TZ3210_6cmeijtd"]),
         model: "A11Z",
         vendor: "Nous",
-        description: "Smart power strip 3 gang with power monitoring",
+        description: "3-channel power strip with total energy monitoring",
         extend: [
             m.deviceEndpoints({endpoints: {l1: 1, l2: 2, l3: 3}}),
-            m.onOff({endpointNames: ["l1", "l2", "l3"], powerOnBehavior: false}),
-            m.electricityMeter(),
```
Koenkk, line 269: Why does `m.electricityMeter(),` not work here? (should prevent the need of the `configure` code)

andrei-lazarov: `m.electricityMeter()` gives some unsupported_attribute errors during configuration and stops without configuring/binding anything.
```yaml
z2m: Failed to configure 'Nous A11Z', attempt 1 (Error: ZCL command 0xa4c138c3ac93ffd1/1 haElectricalMeasurement.read(["acPowerDivisor","acPowerMultiplier"], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"reservedBits":0,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')
```

And we need the TS011F code to prevent the occasional 0 (invalid) measurement

Koenkk: We can pass the divisor which will prevent the read: https://github.com/Koenkk/zigbee-herdsman-converters/blob/b99aacac5653cb2a67989cec97c964ae2c9617b6/src/devices/lumi.ts#L4433, for the converter we can do something similar as `electricalMeasurementsFzConverter`

andrei-lazarov: Ok, I simplified the code as you said

Koenkk:
Thanks!

PR #11532, opened 2026-02-13
Add new fingerprint _TZE204_3q3maeoo for device PO-BOCO-ELEC

Samt43:
Add Kozii by Xanlite heat wire module support (new fingerprint _TZE204_3q3maeoo on the definition of PO-BOCO-ELEC device)

After test, all functionnalities seems to work as expected.


Koenkk:
Thanks!

PR #11538, opened 2026-02-15
Namron Edge Thermostat expects LOCAL Unix time, not UTC

jingsno:
The converter for Namron Edge Thermostat (4512783, 4512784, 4566702) correctly sets the date/time on the device every 24 hours. This time is however UTC, and not local time.  

This pull request sets the Thermostat to the local time, based on the time zone of Z2M.


Koenkk:
Thanks!

PR #11539, opened 2026-02-15
Refactor threePhase handling for exposes and toZigbee

svhelge:
The original code treated threePhase as a master switch that ignored the specific power: false, voltage: false or current: false settings.

Example:  `m.electricityMeter({power: false, threePhase: true})` would expose the power attributes for phase B and C.  

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -2301,7 +2301,7 @@ function genericMeter(args: MeterArgs = {}) {
             exposes.push(e.voltage_phase_b().withAccess(ea.STATE_GET), e.voltage_phase_c().withAccess(ea.STATE_GET));
             toZigbee.push(tz.acvoltage_phase_b, tz.acvoltage_phase_c);
         }
-        if (args.current !== false) {
+        if (args.current) {
             exposes.push(e.current_phase_b().withAccess(ea.STATE_GET), e.current_phase_c().withAccess(ea.STATE_GET));
             toZigbee.push(tz.accurrent_phase_b, tz.accurrent_phase_c);
         }


src/lib/modernExtend.ts
```diff
@@ -2293,22 +2293,18 @@ function genericMeter(args: MeterArgs = {}) {
...
-        );
+        if (args.power !== false) {
+            exposes.push(e.power_phase_b().withAccess(ea.STATE_GET), e.power_phase_c().withAccess(ea.STATE_GET));
+            toZigbee.push(tz.electrical_measurement_power_phase_b, tz.electrical_measurement_power_phase_c);
+        }
+        if (args.voltage !== false) {
+            exposes.push(e.voltage_phase_b().withAccess(ea.STATE_GET), e.voltage_phase_c().withAccess(ea.STATE_GET));
+            toZigbee.push(tz.acvoltage_phase_b, tz.acvoltage_phase_c);
+        }
+        if (args.current !== false) {
```
```suggestion
        if (args.current) {
```

I think this is simpler?

svhelge:
@Koenkk , I saw that `(args.current !== false)` (and similar for power and voltage) is used elsewhere in the code. I believe it's better to be consistent.

Koenkk:
Thanks!

slacky1965:
After the last update  

sharex-18-57-54-910

with `m.electricityMeter({threePhase: true}),`

current_phase_b and current_phase_c disappeared from exposes

sharex-18-57-28-342

slacky1965:
From `modernExtend.ts`

Why are they not the same thing?

sharex-21-13-27-375

Koenkk:
@slacky1965 could you submit a PR to fix this?

slacky1965:
> @slacky1965 could you submit a PR to fix this?

Yes, but a little later.

PR #11543, opened 2026-02-16
Enhance ENGO thermostat capabilities - adding OFF mode to E40 thermostat

adorobis:
Updated the system mode options and added expose device state functionality with OFF system mode. Also adding time synchronization capability. Based on EONE model (_TZE204_djurk6p5)
Removing unsupported holiday features.

### DIFF
--- a/src/devices/engo.ts
+++ b/src/devices/engo.ts
@@ -498,7 +498,7 @@ export const definitions: DefinitionWithExtend[] = [
                             if (options?.expose_device_state === true) {
                                 return isOn ? "ON" : "OFF";
                             }
-                            meta.state.state = undefined;
+                            return undefined;
                         },
                     },
                 ],


src/devices/engo.ts
```diff
@@ -415,59 +415,119 @@ export const definitions: DefinitionWithExtend[] = [
...
+                            }
+                        },
+                        from: (value, meta, options) => {
+                            const isOn = value === true;
+                            meta.state.system_mode = isOn ? (meta.state.system_mode_device ?? "heat") : "off";
+
+                            if (options?.expose_device_state === true) {
+                                return isOn ? "ON" : "OFF";
+                            }
+                            meta.state.state = undefined;
```
Koenkk: Don't edit `meta.state.state`, instead return the new state, e.g. `return {system_mode: xx, state: xx}`

adorobis: Can't make it work to update state when changing system_mode and updating system_mode when changing state. The device reacts properly to either but the state of the other entity is not getting updated. Actually interestingly only in one case the state gets updated properly: when I change system_mode to off the state also gets updated to off. But that is the only case.
Here is the code I've tried so far:
```
    meta: {
        tuyaDatapoints: [
            [1, "state", {
                to: async (value, meta) => {
                    const isOn = value.toLowerCase() === 'on';
                    await tuya.sendDataPointBool(meta.device.getEndpoint(1), 1, isOn, "dataRequest", 1);
                    return isOn ? "ON" : "OFF";
                },
                from: (value, meta, options) => {
                    const isOn = value === true;
                    const systemMode = isOn ? (meta.state.system_mode_device ?? "heat") : "off";
                    const result = {
                        system_mode: systemMode
                    };

                    if (options?.expose_device_state === true) {
                        result.state = isOn ? "ON" : "OFF";
                    }
                    return result;
                },
            }],
            [2, "system_mode", {
                to: async (value, meta) => {
                    const endpoint = meta.device.getEndpoint(1);
                    const isOn = value !== "off";
                    await tuya.sendDataPointBool(endpoint, 1, isOn, "dataRequest", 1);
                    if (value === "heat") {
                        await tuya.sendDataPointEnum(endpoint, 2, 0, "dataRequest", 1);
                    } else if (value === "cool") {
                        await tuya.sendDataPointEnum(endpoint, 2, 1, "dataRequest", 1);
                    }
                    return value;
                },
                from: (value, meta) => {
                    const modes = ["heat", "cool"];
                    const mode = modes[value] ?? "heat";
                    return {
                        system_mode_device: mode,
                        system_mode: meta.state.state === "OFF" ? "off" : mode,
                    };
                },
            }],
```

adorobis: Oh, forget my previous comments. Updated to remove updating meta.state.state with return statement. Hope is good to go now :)

Koenkk:
Thanks!

PR #11550, opened 2026-02-17
fix: corrections for SBDV writeable custom attributes

avzasorin-sd:
Fix for custom attributes of SBDV-* devices:
properly declare writable and not-writable proprietary attributes


Koenkk:
Thanks!

PR #11554, opened 2026-02-19
Fully fix Nous E13 water sensor

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee2mqtt/issues/30673
- fixes https://github.com/Koenkk/zigbee2mqtt/issues/30411#issuecomment-3761673043
- cc: @gonzague device is working properly, buzzer will ring in the next Z2M release

It wasn't working because it needed Enum data type.

I also reworded the features for clarity. 
This device can be used as either 'water leakage sensor' or 'water shortage sensor'

before
after
<p></p>

Same converter might work on [Tuya WLS-100z](https://www.zigbee2mqtt.io/devices/WLS-100z.html#tuya-wls-100z) - the only similar device I could find

### DIFF
--- a/src/devices/nous.ts
+++ b/src/devices/nous.ts
@@ -47,8 +47,11 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [tuya.modernExtend.tuyaBase({dp: true})],
         exposes: [
             e.water(),
-            e.binary("alarm", ea.STATE, true, false).withDescription("Indicates whether the buzzer is currently ringing"),
-            e.enum("alarm_mode", ea.STATE_SET, ["water_presence", "water_absence"]).withDescription("When to sound the alarm").withCategory("config"),
+            e.water_leak().withDescription("Indicates whether the device detected a water leak and the buzzer is ringing"),
+            e
+                .enum("alarm_mode", ea.STATE_SET, ["water_presence", "water_absence"])
+                .withDescription("When to consider a water leak and sound the alarm")
+                .withCategory("config"),
             e
                 .enum("ringtone", ea.STATE_SET, ["muted", "tone_1", "tone_2", "tone_3"])
                 .withDescription("Selected buzzer ringtone for the alarm")
@@ -67,7 +70,7 @@ export const definitions: DefinitionWithExtend[] = [
                         water_absence: tuya.enum(1),
                     }),
                 ],
-                [102, "alarm", tuya.valueConverter.trueFalse1],
+                [102, "water_leak", tuya.valueConverter.trueFalse1],
                 [
                     103,
                     "ringtone",
--- a/src/lib/exposes.ts
+++ b/src/lib/exposes.ts
@@ -1339,7 +1339,9 @@ export const presets = {
             .withUnit("V")
             .withDescription("Measured electrical potential value on phase C"),
     water_leak: () => new Binary("water_leak", access.STATE, true, false).withDescription("Indicates whether the device detected a water leak"),
+    // fits one usecase  (leak = true)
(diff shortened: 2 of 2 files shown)


src/devices/nous.ts
```diff
@@ -43,22 +43,41 @@ export const definitions: DefinitionWithExtend[] = [
...
+                .withCategory("config"),
             e.battery(),
-            e.enum("working_mode", ea.ALL, ["normal", "silent", "test"]).withDescription("Operational mode of the device"),
-            e.text("status", ea.STATE).withDescription("Device status"),
-            e.enum("alarm_ringtone", ea.ALL, ["tone_1", "tone_2", "tone_3"]).withDescription("Alarm ringtone"),
         ],
         meta: {
             tuyaDatapoints: [
-                [1, "water_leak", tuya.valueConverter.trueFalse1],
+                [1, "water", tuya.valueConverter.trueFalse1],
```
Koenkk, line 63: Could we keep this `water_leak` to prevent a breaking change and to be consistent with other water leak detectors?

src/devices/nous.ts
```diff
@@ -43,22 +43,41 @@ export const definitions: DefinitionWithExtend[] = [
...
+                        "water absence": tuya.enum(1),
+                    }),
+                ],
+                [102, "alarm", tuya.valueConverter.trueFalse1],
+                [
+                    103,
+                    "ringtone",
+                    tuya.valueConverterBasic.lookup({
+                        muted: tuya.enum(0),
+                        "tone 1": tuya.enum(1),
```
Koenkk:
```suggestion
                        "tone_1": tuya.enum(1),
```

(also change others)

src/devices/nous.ts
```diff
@@ -43,22 +43,41 @@ export const definitions: DefinitionWithExtend[] = [
...
                 [4, "battery", tuya.valueConverter.raw],
-                [101, "working_mode", tuya.valueConverterBasic.lookup({normal: 0, silent: 1, test: 2})],
-                [102, "status", tuya.valueConverter.raw],
-                [103, "alarm_ringtone", tuya.valueConverterBasic.lookup({tone_1: 0, tone_2: 1, tone_3: 2})],
+                [
+                    101,
+                    "alarm_mode",
+                    tuya.valueConverterBasic.lookup({
+                        "water presence": tuya.enum(0),
+                        "water absence": tuya.enum(1),
```
Koenkk:
```suggestion
                        "water_absence": tuya.enum(1),
```

src/devices/nous.ts
andrei-lazarov: 'Water leak' is ambiguous. It depends where you place the probe. 

- Probe below water container -> water **leaking** -> water on floor -> probe **wet**
- Probe inside water container -> water **leaking** -> container empty -> probe **dry**

Additionally, it flashes red text in windfront when the probe is wet (so it only matches first usecase)

This device supports both usecases, and can sound the alarm on either wet/dry status.

I think `water_leak` should be avoided in future devices, in favor of `water` (or similar wording) for correct wording with all usecases.

Thankfully this device reports the status in 2 datapoints, so I reinterpreted them as water and water_leak, to avoid breaking changes - thanks for pointing it out!

src/devices/nous.ts
andrei-lazarov: Fixed

src/devices/nous.ts
Koenkk: Agree, thanks!

Koenkk:
Thanks!

PR #11556, opened 2026-02-19
fix: lumi: dimmer.mode payload lookup

absent42:
lumi.light.agl00x/lumi.light.acn132 are dual cct/rgb devices that use attribute 0x0509 values of 1/2 when reporting device state. 

The current dimmer.mode payload lookup only has values 1/3 which results in the error:

`z2m: Exception while calling fromZigbee converter: Value: '2' not found in: [1, 3]`

whenever a light moves from a cct (1) to rgb (2) state. 

Added 2: "rgb" to lookup to resolve the error.

### DIFF
--- a/src/lib/lumi.ts
+++ b/src/lib/lumi.ts
@@ -888,7 +888,7 @@ export const numericAttributes2Payload = async (
                 }
                 break;
             case "1289":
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11561, opened 2026-02-20
fix(ignore): do not use mistake-prone name for catch var

Nerivec:
`e` is often used for import.
Only 1 actual problem identified (was throwing the import instead of the error), but fixed all occurrences.

Also improved definition testing a bit, should make it easier to track down bugs (added example) and/or ensure a definition is explicitly tested.


Koenkk:
Thanks!

Issue #8945, opened 2025-03-08
SmartThings IM6001-MPP01 accelerator sensor report N/A


toxic0berliner:
Accelerometer sensors for SmartThings IM6001-MPP01 are all reporting N/A

I have checked and using z2m-edge on Home assistant indeed contains the "fix" from https://github.com/Koenkk/zigbee-herdsman-converters/pull/8676
But sadly it remains N/A for accelerator sensors.

I purchased the device recently and it's my only one so difficult to say if it has anything different...

Running Z2M as home assistant plugin on HAOS, the bridge is a tubeszb-cc2652-poe-2022

My config is stock except the adapter I had to specify recently:
```
port: tcp://10.0.0.20:6638
adapter: zstack
```

I am very noob to radio and RF but I am good with a shell, I can change and try things, just no idea where to start...

CD-Nitrox:
Same issue here, no motion, no x,y,z axis, no battery data. All other sensors do report correctly. 
Just updated to the version 2.1.3 hoping it would fix things but no changes even after repairing and keeping alive for 20sec.

lance-rank:
Same here, I have five of these sensors. One I purchased years ago and four I purchased last month. The one I purchased years ago I used in home assistant with ZHA for years. I migrated from ZHA to Z2M last month and four I purchased last month I use with Z2M (never used them with ZHA). I now use all five with Z2M. 

Weird thing is, the one I purchased years ago has motion, x/y/z axis and battery data. The four new ones do not.

```
{ "battery": 92.5, "battery_low": false, "contact": true, "linkquality": 220, "moving": false, "tamper": false, "temperature": 15.77, "x_axis": -17, "y_axis": 20, "z_axis": -1040 }
```

billhoz:
I have the same issue, using Z2M in HAOS on Proxmox,  SONOFF Zigbee 3.0 USB Dongle

spry-salt:
Ditto with numerous sensors of this make and model.  For example:

`Failed to configure (ZCL command
0x286xxxxxxxxx/1
genPollCtrI write({"longPollinterval":360
O},
{"timeout":10000,"disableResponse":fals e, "disableRecovery":false,"disableDefaul tResponse":true, "direction":0,"reservedB its":0,"writeUndiv":false}) failed (Status
'READ_ONLY'))`

and

[...]

PR #11565, opened 2026-02-20
fix: configure IM6001-MPP01 trying to write attr not writable

Nerivec:
Since it previously was try/catch because of the returned `READ_ONLY` status, I assume just removing the call is the correct behavior. `longPollInterval` is not writable by nature, and no other device appears to customize this behavior, I think that's a safe-enough assumption.

Fixes https://github.com/Koenkk/zigbee-herdsman/issues/1642
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/27140
Fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/8945

TODO: for the config report "edit" not passing the correct manuf code, the custom cluster needs to be moved to ZHC and tailored for each use case (I see 3 with 3 different codes), see https://github.com/Koenkk/zigbee-herdsman/issues/1622

### DIFF
--- a/src/devices/smartthings.ts
+++ b/src/devices/smartthings.ts
@@ -1,15 +1,11 @@
 import {Zcl} from "zigbee-herdsman";
-
 import * as fz from "../converters/fromZigbee";
 import * as tz from "../converters/toZigbee";
 import * as constants from "../lib/constants";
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #11568, opened 2026-02-21
fix(sinope): disable unsupported SW2500ZB metering power handling

mar-kolya:
Tested on a physical SW2500ZB running firmware 539.

Observed errors:
- seMetering configure/reporting setup fails with UNSUPPORTED_ATTRIBUTE
- direct reads of seMetering.instantaneousDemand fail with UNSUPPORTED_ATTRIBUTE

Effectively disabled for SW2500ZB:
- reporting configuration for metering attributes
- power handling via seMetering.instantaneousDemand

Kept enabled:
- delivered energy handling via seMetering.currentSummDelivered
- normal on/off switch behavior

Note: on the tested physical device, seMetering.currentSummDelivered also did not seem to produce useful data.

### DIFF
--- a/src/devices/sinope.ts
+++ b/src/devices/sinope.ts
@@ -1473,7 +1473,10 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SW2500ZB",
         vendor: "Sinopé",
         description: "Zigbee smart light switch",
-        extend: [m.onOff(), m.electricityMeter({cluster: "metering"})],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

mar-kolya:
I think this may fix https://github.com/Koenkk/zigbee2mqtt/issues/25427

raetha:
It does seem to solve the issue I originally reported in #25427. Not seeing errors anymore when looking at those devices. Thank you so much!

Issue #11173, opened 2026-01-02
(probably) Wrong/missing datapoints on SFL02-Z-1 and SFL02-Z-2 (Moes Star Feather)


maacaron:
I have at home 2 different light switches from Moes - Moes Star Feather 1 gang (SFL02-Z-1) and 4 gang (SFL02-Z-4). I found that 4 gang has option the same as wifi versions in app - Indicator Status - which can be set to off/relay/invert. In off mode backlight is off, in relay backlight lights brighter when relay is on, darker when off, and in invert mode it lights brighter when off and darker when on. I wanted to do the same in 1 gang but there is no option in exposed functions.
1 gang switch:
<img width="441" height="869" alt="Image" src="https://github.com/user-attachments/assets/524ae4e2-4ec9-4f1b-879f-084921f66a57" />
and 4 gang switch:
<img width="432" height="819" alt="Image" src="https://github.com/user-attachments/assets/3bb4a5af-1703-41af-932a-280b10193869" />
<img width="440" height="862" alt="Image" src="https://github.com/user-attachments/assets/4b619bdd-46f2-4e96-aba8-4f1bb585b7b7" />
<img width="485" height="823" alt="Image" src="https://github.com/user-attachments/assets/7e908f88-c931-46a3-a74d-f81853b0518f" />
I found in code that there is a difference between 1/2 gang and 3/4 gang.
Can someone verify it? I don't have a possibility to test datapoints by myself :(

1 gang:
no line with exposing indicator status:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/moes.ts#L404
tuyaDatapoints has `indicator_mode` not `indicator_status` and different value
https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/moes.ts#L423
`[37, "indicator_mode", tuya.valueConverterBasic.lookup({none: 0, relay: 1, pos: 2})],`

4 gang (here it works):
exposing: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/moes.ts#L634
`exposes.enum("indicator_status", ea.ALL, ["Off", "Relay", "Invert"]).withDescription("Indicator status"),`
tuyaDatapoints: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/moes.ts#L669
[...]

PR #11573, opened 2026-02-21
Fix Moes Star Feather indicators

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/11173


Koenkk:
Thanks!

PR #11579, opened 2026-02-22
Fix Battery not being exposed for TECH VNTH-T2 TRV

greenbigfrog:
Hi,
I was trying to figure out the last couple of days why the TECH TRVs I have were showing in z2m, but not ha. Tried basically everything from rediscovering, simply waiting, ...
Today I had the thought to start looking at the z2m device definitions. And lo and behold I discovered, that the data was being collected, but all other users managed to miss that the battery value is missing from the exposed section.
Hence, I've created this PR.
I hope this is correct.


Koenkk:
Thanks!

PR #11583, opened 2026-02-23
lumi: remove packet delay in segment color writes

absent42:
lumiSegmentColors: remove the 50ms sleep() delay between consecutive segment color group writes. Testing on physical devices confirms they handle back-to-back writes reliably without any packet drops.

lumiRGBEffect/lumiRGBEffectColors/lumiEffectSpeed/lumiRGBEffectSegments/lumiSegmentColors: changed to disableDefaultResponse: true - resolves after APS ACK instead of ZCL Default Response to speed up effect activation.


Koenkk:
Thanks!

PR #11587, opened 2026-02-24
Move cluster wiserDeviceInfo to ZHC

svhelge:
Moved converter wiser_device_info from fromZigbee.ts to schneider_electric.ts and updated the fromZigbee definitions for the devices using the converter. 
Added modernextend function addWiserDeviceInfoCluster that adds the custom cluster wiserDeviceInfo. 

@Koenkk , will it work to add the addWiserDeviceInfoCluster with 
  `extend: [schneiderElectricExtend.addWiserDeviceInfoCluster()],`
for the devices that use the cluster?  The actual devices are not defined with modernextend. 

I'm a bit uncertain about what will hapenn when moderenextend is mixed with traditional device definition. 

I plan to create a PR in order to remove the wiserDeviceInfo cluster from ZH.

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -779,6 +779,21 @@ const schneiderElectricExtend = {
             commands: {},
             commandsResponse: {},
         }),
+    addSchneiderLightSwitchConfigurationCluster: () =>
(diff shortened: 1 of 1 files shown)


Koenkk:
> for the devices that use the cluster? The actual devices are not defined with modernextend.

Yes that will work, thanks again!

PR #11591, opened 2026-02-24
Move unused manuSpecificDoorman cluster code from ZH to ZHC

svhelge:
Added and commented out unused manuSpecificDoorman cluster code from ZH.

### DIFF
--- a/src/devices/yale.ts
+++ b/src/devices/yale.ts
@@ -12,6 +12,46 @@ const NS = "zhc:yale";
 const e = exposes.presets;
 const ea = exposes.access;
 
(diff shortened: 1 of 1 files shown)


Koenkk:
If it's not used, I propose to not add back the commented code here.

svhelge:
See the discussion here, https://github.com/Koenkk/zigbee-herdsman/issues/1622#issuecomment-3951225848. I'm happy with either option.

Nerivec:
My fault :grimacing: 
Indeed we can just use the GitHub history to restore it later if a use case is ever needed :+1: 

@svhelge I saw the wiser info one wasn't removed from ZH, do you have the PR pending locally?

svhelge:
> My fault 😬 Indeed we can just use the GitHub history to restore it later if a use case is ever needed 👍
> 
> @svhelge I saw the wiser info one wasn't removed from ZH, do you have the PR pending locally?

PR created https://github.com/Koenkk/zigbee-herdsman/pull/1657

PR #11594, opened 2026-02-24
fix: Amina S energy reporting

mikaelboman:
In e30ace459d7028b2df697fa12ad15616b49b895a, Amina S energy reporting was refactored to use a fromZigbee coverter instead of onEvent() but the new converter was never added to the fromZigbee property so it was never used.


Koenkk:
Thanks!

PR #11597, opened 2026-02-25
Update tuya.ts

LordRoge:
The SA1 device will enter sleep mode in 30 minutes (the update speed is very slow), so I am using this method to report every 30 seconds.


Koenkk:
@voronovrs since you added this device, can you confirm this behaviour?

voronovrs:
Hi @Koenkk, this device as some other tuya can delay response for some data, especially one reported through test5, but i see that energy reported constantly on change. I see the same behavior for TO-Q-SYS-JZT which already had the suggested changes. 
So if the question is - is device stops to send any data for 30 or more minutes - no. Does it slow to send some data on changes - yes. Will this fix help - I'm not sure.
@LordRoge i would suggest some long tests, from couple hours to a day (if you already didn't do that ofc), because I saw that after powercycling or reconnecting device it will report data very often for some period of time.

LordRoge:
@voronovrs         After power-up, the device reports 3 data entries every 5 seconds. After 30 minutes, it switches to once an hour, which is a bit too infrequent.

PR #11599, opened 2026-02-25
Move elko cluster attributes (hvacThermostat) from ZH to ZHC 

svhelge:
Move Elko specific hvacThermostat attributes from ZH to ZHC.
Complete modernextend implementation for Elko thermostat.
Remove Elko converters from toZigbee and fromZigbee. 
Add actual power measurement (power), still virtual, but based on load and running_state.


Koenkk:
Thanks!

PR #11604, opened 2026-02-25
Move cluster manuSpecificSchneiderFanSwitchConfiguration from ZH to ZHC.

svhelge:
Add cluster with moderextend and update devices.


Koenkk:
Thanks!

PR #11612, opened 2026-02-26
Move manuSpecificAssaDoorLock cluster from ZH to ZHC

svhelge:
Added cluster and interface. 
Replaced reference to cluster manuSpecificUbisysDeviceSetup with manuSpecificAssaDoorLock. 
Updated the c4_assa_lock_attribute converter (using attributes instead of data[]). 
Added cluster to device "ZYA-C4-MOD-S" with extend.


Koenkk:
Thanks!

PR #11614, opened 2026-02-27
General thermostat improvements

jpaju:
Extends `thermostat` function in modernExtend.ts with three new parameters:
- `programmingOperationMode` - setpoint/schedule/schedule_with_preheat/eco modes
- `setpointChangeSource` - tracks if setpoint was changed manually, by schedule, or externally  
- `weeklySchedule` - enables weekly schedule get/set/clear

Also adds:
- Reporting utilities: `thermostatSetpointChangeSource`, `thermostatProgrammingOperationMode`
- Climate expose helpers: `withProgrammingOperationMode()`, `withSetpointChangeSource()`
- Moved `thermostatProgrammingOperationModes` lookup to constants.ts

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -3473,12 +3473,7 @@ export function thermostat(args: ThermostatArgs): ModernExtend {
         }
 
         if (!programmingOperationMode.configure?.skip) {
-            configure.push(
-                setupConfigureForReporting("hvacThermostat", "programingOperMode", {
-                    config: programmingOperationMode.configure?.reporting ?? repConfigChange0,
-                    access: programmingOperationMode.configure?.access ?? ea.ALL,
-                }),
-            );
+            configure.push(setupConfigureForReading("hvacThermostat", ["programingOperMode"], endpointNames));
         }
     }
 
--- a/src/lib/reporting.ts
+++ b/src/lib/reporting.ts
@@ -204,10 +204,6 @@ export const thermostatTemperatureSetpointHoldDuration = async (endpoint: Zh.End
     const p = payload<"hvacThermostat">("tempSetpointHoldDuration", 0, repInterval.HOUR, 10, overrides);
     await endpoint.configureReporting("hvacThermostat", p);
 };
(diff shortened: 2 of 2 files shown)


src/lib/modernExtend.ts
```diff
@@ -3456,6 +3465,51 @@ export function thermostat(args: ThermostatArgs): ModernExtend {
...
+    if (programmingOperationMode) {
+        expose.withProgrammingOperationMode(programmingOperationMode.values);
+
+        if (!programmingOperationMode.toZigbee?.skip) {
+            toZigbee.push(tz.thermostat_programming_operation_mode);
+        }
+
+        if (!programmingOperationMode.configure?.skip) {
+            configure.push(
+                setupConfigureForReporting("hvacThermostat", "programingOperMode", {
```
Koenkk: I think we should not setup reporting for these new attributes, I think it's unlikely that these are changed from any other sources than z2m. We are starting to hit limits of the amount of configures thermostats can store.

jpaju: Thanks for the review! I changed `programingOperMode` to only read instead of configure reporting. For `setpointChangeSource` the whole idea is to have the thermostat to report this attribute, which is why I left it to use reporting still

Koenkk: Can you then update `SRTS-A01` to not configure reporting for this? This thermostat can only store 5 reportings, and this would be the 6th one (leading to a configure failure)

jpaju: These new thermostat options (`setpointChangeSource`, `programmingOperationMode`, `weeklySchedule`) only affect devices that explicitly opt into them. Existing devices like SRTS-A01 won't get any new reporting configured unless someone modifies their definition to use these options.

So SRTS-A01 should continue to work exactly as before ([it doesn't](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/lumi.ts#L3652-L3701) even use `thermostat()` function. Am I missing something?

Koenkk: > Existing devices like SRTS-A01 won't get any new reporting configured unless someone modifies their definition to use these options. 

Ah yes ofcourse, apologies!

Koenkk:
Thanks!

PR #11617, opened 2026-02-28
[Update lumi.switch.b3n01] Remove lumiPower from the device list

hen6e:
This device do not have genAnalogInput cluster, lumiPower will cause error "Error: Device 0x54ef441000abe0f0 has no input cluster genAnalogInput".


Koenkk:
Thanks!

PR #11618, opened 2026-02-28
Enhance extend properties for Aqara light controllers

hen6e:
Updated the extend properties for multiple Aqara light controllers to include additional features such as dimming range and power source identification. New definition implement the function of powering on brightness transition curve curvature, Off on duration, On off duration.

image


Koenkk:
Thanks!

PR #11627, opened 2026-03-01
Correct battery voltage for _TZE200_jt50ea5d

Liionboy:

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1885,7 +1885,7 @@ export const definitions: DefinitionWithExtend[] = [
                 .withDescription("Outlet water temperature")
                 .withValueMin(0)
                 .withValueStep(0.01),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Liionboy:
@Koenkk there is a problem with this senzor from what o saw, is work only if use like this:

```
e.numeric("battery_voltage", ea.STATE).withUnit("V").withDescription("Power supply voltage").withValueMin(0).withValueStep(0.01),

[24, "battery_voltage", tuya.valueConverter.divideBy100],
```

Is ok to add like this?

Thanks

Koenkk:
What's the problem?

Liionboy:
> What's the problem?

I did 2 new PR because i see now that sensor is numeric, will work in next release

PR #11630, opened 2026-03-02
Update tuya.ts

LordRoge:
I confirmed with the technical staff of the brand (Tonogu) that resetting the update frequency requires enabling the "online status" of dp point 109. I find this operation too cumbersome, so I sincerely request adding an active query interval of 10 seconds.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -22181,7 +22181,7 @@ export const definitions: DefinitionWithExtend[] = [
             tuya.modernExtend.tuyaBase({
                 dp: true,
                 queryOnConfigure: true,
-                queryIntervalSeconds: 10,
+                queryIntervalSeconds: 60,
             }),
         ],
         exposes: [


src/devices/tuya.ts
```diff
@@ -22177,7 +22177,13 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Tongou",
         description: "Zigbee energy meter (transformer clamp)",
         whiteLabel: [tuya.whitelabel("Tongou", "TOSA1-01WXJAT2A", "Smart energy meter, two wire", ["_TZE284_4hdbt6rn"])],
-        extend: [tuya.modernExtend.tuyaBase({dp: true})],
+        extend: [
+            tuya.modernExtend.tuyaBase({
+                dp: true,
+                queryOnConfigure: true,
+                queryIntervalSeconds: 10,
```
Koenkk: This query interval is very low and might unnecessarily spam zigbee networks, how does it actually fix it and can we only do the query when it's really needed?

LordRoge: I am a partner of Tongou. The query interval in Z2M actually triggers the function that reports all datapoints from the device. In the latest test version of the device, the number of datapoints reported in bulk has been reduced to five. Additionally, the device now reports energy status only when there is a change of 0.01 kWh, and this is not included in the current function that reports all datapoints. Sending five messages every 10 seconds should not have a significant impact, but it could be changed to every 30 seconds if needed. Originally, the device reported at a frequency of three messages every five seconds for the first 30 minutes.

LordRoge:
[2026-03-02 14:26:01] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,0,4],"type":"Buffer"},"datatype":2,"dp":1}],"seq":20992}' from endpoint 1 with groupID 0
[2026-03-02 14:26:01] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,1,4],"type":"Buffer"},"datatype":2,"dp":131}],"seq":21248}' from endpoint 1 with groupID 0
[2026-03-02 14:26:01] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[9,60,0,0,74,0,0,15],"type":"Buffer"},"datatype":0,"dp":6}],"seq":21504}' from endpoint 1 with groupID 0
[2026-03-02 14:26:01] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,19,128],"type":"Buffer"},"datatype":2,"dp":32}],"seq":21760}' from endpoint 1 with groupID 0
[2026-03-02 14:26:01] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,0,85],"type":"Buffer"},"datatype":2,"dp":50}],"seq":22016}' from endpoint 1 with groupID 0
[2026-03-02 14:26:31] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,0,4],"type":"Buffer"},"datatype":2,"dp":1}],"seq":22272}' from endpoint 1 with groupID 0
[2026-03-02 14:26:31] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,1,4],"type":"Buffer"},"datatype":2,"dp":131}],"seq":22528}' from endpoint 1 with groupID 0
[2026-03-02 14:26:31] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[9,60,0,0,75,0,0,15],"type":"Buffer"},"datatype":0,"dp":6}],"seq":22784}' from endpoint 1 with groupID 0
[2026-03-02 14:26:31] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,19,130],"type":"Buffer"},"datatype":2,"dp":32}],"seq":23040}' from endpoint 1 with groupID 0
[2026-03-02 14:26:31] debug: 	z2m: Received Zigbee message from '0x60b763fffe1e6ec9', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[0,0,0,84],"type":"Buffer"},"datatype":2,"dp":50}],"seq":23296}' from endpoint 1 with groupID 0
We have indeed reduced the active reporting to five key data points: dp1: energy, dp131: temperature, dp6: current, voltage, power, dp32: AC frequency, dp50: power factor.

LordRoge:
The TOWSMR1 device will be further optimized in the future. As previously (#29865), the issue of packet duplicate reporting was caused by the function that reports all datapoints.

src/devices/tuya.ts
Koenkk: The problem is that z2m also has to send a message to the device, I propose to increase it to 60 seconds.

LordRoge: Just OK,Thanks for the reminder.

Koenkk:
Thanks!

Virusmater:
@LordRoge it seems that this change doesn't fix the issue. I tried using edge version of zigbee2mqtt - reporting slows down after 15 or so minutes, same as before the change. Only triggering alarm action helps to re-awake the device.

francouai:
new fix is there but it doesn't work for me.

PR #11635, opened 2026-03-02
BILRESA - remove voltageReporting

andrei-lazarov:
- My bad here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/11403

Voltage is supported, but "unreportable". Throws errors during configuration


Koenkk:
Thanks!

PR #11636, opened 2026-03-02
fix: LiXee ZLinky: bump definition version to trigger reconfigure

sbraz:
This ensures existing devices are automatically reconfigured on next
Z2M startup, picking up the tariffPeriod reportable attribute change
from 4741ae3 (#10903), using the reconfigure mechanism introduced in
2192bbb (#11148).

@Koenkk I'd have done it for 2.9.0 if I had known of this new feature. Will  there be another point release soon?


Koenkk:
Thanks!

Koenkk:
> Will there be another point release soon?

Unless there is a serious issue in 2.9.1, next release is 1 April.

PR #11638, opened 2026-03-02
Move manuSpecificSamsungAccelerometer from ZH to ZHC

svhelge:
Moved converter from fromZigbee.ts to smartthings,ts. 
Added cluster manuSpecificSamsungAccelerometer to smartthings.ts. 
**Had to rename the cluster attributes to camelCase (since "pnpm run check" gave failures).** 
Since the cluster is still defined in ZH this throws an error with "pnpm test".  
Updated test in checkDefinition.test.ts.

### DIFF
--- a/test/checkDefinition.test.ts
+++ b/test/checkDefinition.test.ts
@@ -36,17 +36,17 @@ describe("Check definition", () => {
                     ],
                     [
                         "manuSpecificSamsungAccelerometer",
-                        [reportingItem("x_axis", 10, repInterval.HOUR, 5)],
+                        [reportingItem("xAxis", 10, repInterval.HOUR, 5)],
                         {manufacturerCode: Zcl.ManufacturerCode.SAMJIN_CO_LTD},
                     ],
                     [
                         "manuSpecificSamsungAccelerometer",
-                        [reportingItem("y_axis", 10, repInterval.HOUR, 5)],
+                        [reportingItem("yAxis", 10, repInterval.HOUR, 5)],
                         {manufacturerCode: Zcl.ManufacturerCode.SAMJIN_CO_LTD},
                     ],
                     [
                         "manuSpecificSamsungAccelerometer",
-                        [reportingItem("z_axis", 10, repInterval.HOUR, 5)],
+                        [reportingItem("zAxis", 10, repInterval.HOUR, 5)],
                         {manufacturerCode: Zcl.ManufacturerCode.SAMJIN_CO_LTD},
                     ],
                 ],


svhelge:
@Koenkk, please advice. Revert the renaming of the cluster attributes (back to snake_case from camelCase)?

Koenkk:
Let's merge with `// @ts-expect-error ignore` added above the erroneous lines

Koenkk:
Thanks!

src/devices/smartthings.ts
```diff
@@ -92,11 +149,14 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.batteryVoltage(endpoint);
             const payloadA = reporting.payload<"manuSpecificSamsungAccelerometer">("acceleration", 10, constants.repInterval.MINUTE, 1);
             await endpoint.configureReporting("manuSpecificSamsungAccelerometer", payloadA, options);
-            const payloadX = reporting.payload<"manuSpecificSamsungAccelerometer">("x_axis", 10, constants.repInterval.MINUTE, 1);
+            // @ts-expect-error ignore
```
Nerivec, line 152: Should _almost never_ do this.
```ts
reporting.payload<"manuSpecificSamsungAccelerometer", SamsungAccelerometer>
```
Should take care of properly typing.

src/devices/smartthings.ts
Nerivec, line 369: Per https://github.com/Koenkk/zigbee-herdsman-converters/pull/11565 (see linked issues), there's a remaining problem with this custom cluster, in that it's not passing the proper `manufacturerCode` by default during manual ops (like configuring reporting). I think the custom cluster needs to be duplicated 3x for the three different manuf codes being used with it in this file.

src/devices/smartthings.ts
Koenkk: I fixed the typing in https://github.com/Koenkk/zigbee-herdsman-converters/pull/11641 (required zh bump)

test/checkDefinition.test.ts
```diff
@@ -12,7 +13,8 @@ describe("Check definition", () => {
                 endpoints: [{ID: 1, inputClusters: ["msTemperatureMeasurement", "genPowerCfg", "manuSpecificSamsungAccelerometer", "genPollCtrl"]}],
             }),
             meta: undefined,
-            fromZigbee: [fz.temperature, fz.battery, fz.ias_contact_alarm_1, fz.smartthings_acceleration],
+            extend: [smartthingsExtend.addManuSpecificSamsungAccelerometerCluster()],
```
Nerivec, line 16: @Koenkk this won't work as soon as ZH is updated to 9.0.11. This is not valid (does nothing), cluster will be missing, test will fail. We need to implement `extend` somehow in the `assertDefinition`, or some other mechanism to have custom clusters considered properly (lookup is entirely mocked in ZHC right now).
Maybe instead of mocking `Device` and `Endpoint` completely, we can rely more on the actual ZH implementations?

EDIT: see https://github.com/Koenkk/zigbee-herdsman-converters/pull/11656

PR #11646, opened 2026-03-03
Fix: TS0601_bed_presence_sensor Occupancy DP

shopsD:
Fixed mapping of occupancy flag

Corrected minor description typo (extra single quote)


Koenkk:
Thanks!

PR #11648, opened 2026-03-04
Move manuSpecificPhilips and manuSpecificPhilips2 from ZH to ZHC

svhelge:
Move converters in fromZigbee.ts with custom clusters to lib/philips.tc. 
"Integrate" cluster manuSpecificPhilips2 into philips.m.light().  
Add clusters to devices as needed (based on converters and modernextend functions). 
Added manuSpecificPhilips2 to device in atlantic.ts


Koenkk:
Thanks!

Issue #8697, opened 2025-01-29
Multiple errors and inaccuracies in Philips Hue manuSpecificPhilips2 support


chrivers:
Hello Koenkk, et. al. :)

It's been difficult to figure out the best way to report this. I'm not sure if you'd prefer many small issues, a discussion, or one big detailed issue. So here we're trying the latter option. Let me know if you'd like some other way of attacking this thing 👍 

It's @chrivers here from the [Bifrost project](https://github.com/chrivers/bifrost).

Bifrost is a Hue Bridge emulator, which uses one or more Zigbee2Mqtt servers as the backend for zigbee communication.

I've recently been working to add support for more advanced features of Hue lights, including:

 - Gradient light support
 - Effects ("Candle", "Fireplace", etc)
 - Colorered effects
 - Animation speed for effects
 - etc

Unfortunately, only a small subset of these features are currently exposed through zigbee2mqtt.

Looking at ZHC (`src/lib/philips.ts`), I was able to get a head start with the current understanding of the zigbee frame format.

After following your excellent guide to sniffing zigbee traffic, I was able to capture a large variety of sample messages, sent between a Philips Hue Bridge, and my LCX005 ("Hue Gradient strip for PC").

At that point, all I needed to do, was to spend a really intense amount of time staring at numbers... 😅 

But, good news!

I have managed to *completely* reverse engineer this format. I'm fairly confident that this is the most comprehensive, correct, and detailed open-source specification of the format available anywhere:

https://github.com/chrivers/bifrost/blob/master/doc/hue-zigbee-format.md

Given the awesome flexibility of this manufacturer-specific frame format, I can see a number of major and minor problems and inaccuracies in the current `src/lib/philips.ts` converter:

 1) Gradient colors are handled as XY-coordinates (a device-independent color space) by the device, but presented as RGB hex values in z2m (a device-*dependent* color space, with undefined gamut)

 2) The (forced) color conversion to RGB is implemented incorrectly.

For example:
  "Red" as "Color (X/Y)" Produces a nicely saturated red color
  "Red" as "Color (H/S)" Produces the same result
  "Red" as a gradient color gives a washed out pink

This is *not* because the gradient strip cannot show red in gradient mode. For example, sending the command `430101200a200000000e7f5f0e7f5f3800` will produce a deep, saturated red color in gradient mode.

 3) The transition time cannot be specified

[...]

PR #11655, opened 2026-03-04
philips: add hue native control

Mihonarium:
# philips: add hue native control

add/reweite manuSpecificPhilips2 encode/decode, fix effects and state reporting

Addresses #8697, Koenkk/zigbee2mqtt#18406, Koenkk/zigbee2mqtt#15891, Koenkk/zigbee2mqtt#24438.

## Summary

Adds a `hue_native_control` option for native control of Philips lights.

Rewrites the Philips `manuSpecificPhilips2` encoder/decoder drafted by @LaurentvdBos based on the [Bifrost spec](https://github.com/chrivers/bifrost/blob/master/doc/hue-zigbee-format.md), fixing several data corruption bugs and adding missing functionality. Also improves effect UX: effects now report state, support color+speed in a single command, and the color wheel/effect interaction is configurable.

## Opt-in

By default, standard ZCL converters handle on/off, brightness, color, and color temperature. The behavior is unchanged from before this PR. The manuSpecificPhilips2 cluster can send all of these in a single atomic command, which is required for full effect support: changing color while an effect is running, stopping effects by changing color, and setting an effect with a specific color in one command. However, native control also changes some user-facing behavior (e.g., no simulated fade-to-off transition), so it is opt-in.

Users can enable it per-device in Z2M Settings → Settings (Specific), or globally via `device_options: { hue_native_control: true }` in configuration.yaml.

## Bug fixes

(Compared to the manySpecificPhilips2 draft.)

- **Y-coordinate byte packing** — `decodeGradientColors` assembled the Y value from the wrong nibbles, corrupting every gradient color's Y component
- **Off-by-two return** — `decodeGradientColors` returned `position + nColors * 3` instead of `position + size + 1`, causing the parser to land at the wrong offset for subsequent fields
- **Wrong Y scaling constant** — `GRADIENT_COLORS_MAX_Y` was `0.8413` (CIE D65 white point), should be `0.8264` (Wide Color Gamut upper bound per Bifrost spec)
- **`knownEffects` key shift** — Effect hex keys were shifted by one byte, causing wrong effects to be triggered
- **Brightness validation** — Values 0 and 255 are invalid per spec, now clamped to 1–254
- **Effect state never reported** — `philipsTz.effect.convertSet` didn't return `{state}`, so Z2M's effect state was always `null`
- **Stale effect state** — Changing color stops the active effect on the device, but Z2M kept showing the old effect name. Now cleared to `"none"`.
- **Attribute reports only for gradient devices** — The `bind("manuSpecificPhilips2", ...)` call was inside the `if (args.gradient)` block. Moved to cover all `hueEffect` devices.

## New features

- **All Bifrost effects** — Added sunset, sparkle, opal, glisten, underwater, cosmos, sunbeam, enchant (13 total including no_effect, candle, fireplace, prism/colorloop, sunrise)
- **`effect` + `color` in one command** — `{"effect": "candle", "color": "#FF4400"}` activates the effect at that color atomically. Also supports `effect_speed` in the same payload.
- **`effect_color`** (new expose) — Sets the base color of the active effect without stopping it. Accepts hex or XY.
- **`effect_color_mode`** (new per-device option) — Controls what happens when the regular color wheel is used while an effect is active:
  - `"stop"` (default): color change stops the effect (matches Hue app)
  - `"update"`: color change re-sends the effect with the new color
- **Deferred brightness on effect re-send** — When re-sending an effect (via `effect_color` or "update" mode) with an explicit brightness, the brightness is sent as a separate command after the effect, since effects reset brightness on activation.
- **`effect` and `effect_speed` access changed to `STATE_SET`** — Current values are now visible in the Z2M frontend and HA.
- **`effect_speed`** — Settable via both the new path (`philipsLightTz`) and the effect converter.
- **`gradient_scale` / `gradient_offset`** — Settable. Fixed-point 5.3 format exposed as float.
- **`gradient_style`** — Fully settable (`linear`, `scattered`, `mirrored`). Was decoded but not writable.
- **`gradient_xy`** — Lossless XY coordinates published alongside RGB hex `gradient`.
- **`philips2_raw`** — Raw hex blob published for advanced clients (e.g. Bifrost) to decode independently.
- **HSV→XY conversion** — HSV colors are now converted correctly instead of being silently dropped.
- **`transition` → `fadeSpeed`** — Mapped to the Bifrost spec's per-message fade speed field.
- **Delayed state read after effect activation** — Reads device state 1s after activating an effect to sync actual brightness.

(Feel free to remove or ask me to remove the effect_color, effect_color_mode, and related stuff if it goes against the overall philosophy.)

## Encoder/decoder rewrite

Replaced the ad-hoc hex string parsing with structured `Encode`/`DecodeManuSpecificPhilips2` functions using `DataView` for proper binary field handling. The wire order follows the Bifrost spec (gradient colors before effect speed before gradient params — reversed relative to flag bit order). Each field type is defined as a `HueTypeDetails` struct with `encode`/`decode`/`flag`/`maxLength`, iterated in wire order.

## Tests
  
- Updated `test/philips.test.ts` — corrected expected values for the MAX_Y fix
- Updated `test/generateDefinition.test.ts` — added `manuSpecificPhilips2Fz` to expected fromZigbee and new toZigbee keys/exposes for the auto-generated definition test
- Updated `test/modernExtend.test.ts` — replaced `philips.fz.gradient` with `philips.manuSpecificPhilips2Fz` in expected fromZigbee, added new toZigbee keys/exposes for the philipsLight test
- Added `test/philips2.test.ts` (61 tests):
  - Known-payload decode of all 7 Bifrost spec examples
  - Round-trip (`Encode→Decode` and `Decode→Encode→Decode`) for every field type
  - Gradient byte packing edge cases (0x000/0xFFF, Bifrost spec example x=0x123 y=0x456)
  - Wire order verification (non-monotonic flag-to-position mapping)
  - All 13 effect types
  - 108 philips tests total (47 + 61), all passing

## Backward compatibility

All changes are backward-compatible. Existing MQTT payloads, automations, and gradient scenes continue to work. New exposes (`effect_color`, `effect_speed`, `gradient_style`, `gradient_xy`, `philips2_raw`) are additive. The `effect_color_mode` option defaults to `"stop"` (Hue app behavior). Devices may need a reconfigure after updating to establish the `manuSpecificPhilips2` binding for state reports.

`philipsFz.gradient` is superseded by `manuSpecificPhilips2Fz` and is no longer registered for any device. It could be possible that external converter authors import it, but it's not documented anywhere and people are more likely to use `philipsLight()`. Probably worth considering deprecated/removing in the future? I've not removed it myself though.

## Limitations

After activating an effect, we sync the device state via a 1-second setTimeout, as some Hue effects change light color/brightness. If the device goes offline in the window, the error is silently caught. There could be more robust approaches.

## Other

I made a very cool hass dashboard card (inspired by the Hue app, but better than it)! See https://github.com/Koenkk/zigbee-herdsman-converters/pull/11655#issuecomment-4307065598


Mihonarium:
Okay, I'm satisfied/done!

Feel free to make changes/merge.

src/lib/philips.ts
```diff
@@ -108,6 +117,63 @@ interface ManuSpecificPhilips2 {
     commandResponses: never;
 }
 
+export const manuSpecificPhilips2Fz: Fz.Converter<"manuSpecificPhilips2", ManuSpecificPhilips2, ["attributeReport", "readResponse"]> = {
+    cluster: "manuSpecificPhilips2",
+    type: ["attributeReport", "readResponse"],
+    convert: (model, msg, publish, options, meta) => {
+        const retval: KeyValueAny = {};
+
+        logger.debug(`convert() with message: ${JSON.stringify(msg)}`, NS);
```
Koenkk: z2m already logs it, no need to log it twice.

src/lib/philips.ts
```diff
@@ -108,6 +117,63 @@ interface ManuSpecificPhilips2 {
...
+
+        if (msg.data.state !== undefined) {
+            // Publish the raw, unaltered state blob so advanced clients (e.g. Bifrost)
+            // can perform their own decoding without depending on z2m's interpretation.
+            retval["philips2_raw"] = msg.data.state.toString("hex");
```
Koenkk:
```suggestion
            retval["philips_raw"] = msg.data.state.toString("hex");
```

src/lib/philips.ts
```diff
@@ -175,15 +241,245 @@ const philipsModernExtend = {
...
+
         const result = modernExtend.light(args);
-        result.toZigbee.push(philipsTz.hue_power_on_behavior, philipsTz.hue_power_on_error);
+        const toZigbee = result.toZigbee;
+        result.toZigbee = [];
+
+        const keys = toZigbee.flatMap((value, index, array) => value.key).filter((value, index, array) => array.indexOf(value) === index);
+        // Add keys for Philips2-specific features not handled by standard converters
+        keys.push("effect_speed", "gradient_scale", "gradient_offset", "gradient_style", "effect_color");
+        const philipsLightTz = {
```
Koenkk, line 257: Can you move this `const` outside this method?

src/lib/philips.ts
```diff
@@ -175,15 +241,245 @@ const philipsModernExtend = {
...
+            key: keys,
+            convertSet: async (entity, key, value, meta) => {
```
Koenkk, line 259: Since this overwrites the standard converter, are we sure that every Philips bulb support this? (even the old ones?)

Koenkk:
@chrivers could you also take a look?

src/lib/philips.ts
Mihonarium: Sorry, could you clarify which const you're referring to, the keys or the philipsLightTz? The latter would be fairly harder to move outside the method as a const (can make it a function that returns what's assigned to the const though?).

src/lib/philips.ts
Mihonarium: Added a fallback (I'd appreciate it if you could check its logic)

src/lib/philips.ts
Mihonarium: Added `const philips2Keys = ["effect_speed", "gradient_scale", "gradient_offset", "gradient_style", "effect_color"];`

src/lib/philips.ts
```diff
@@ -180,15 +244,270 @@ const philipsModernExtend = {
...
+        const keys = toZigbee.flatMap((value) => value.key).filter((value, index, array) => array.indexOf(value) === index);
+        keys.push(...philips2Keys);
+        const philipsLightTz = {
+            key: keys,
+            convertSet: async (entity, key, value, meta) => {
+                // Old bulbs may not support manuSpecificPhilips2 — fall back to standard converters.
+                // Since Z2M calls us once and marks us as used for all our keys, we must delegate
+                // ALL message keys to the appropriate standard converters in one shot,
+                // mimicking Z2M's own per-key dispatch loop.
+                if (utils.isEndpoint(entity) && !entity.supportsInputCluster("manuSpecificPhilips2")) {
```
Koenkk: I was checking the current tz converter, but this also has some special features like: https://github.com/Koenkk/zigbee-herdsman-converters/blob/fac41963fe0a3cd62b5a3a3b21ea928e5b36944f/src/converters/toZigbee.ts#L1284

I expect that with this converter the behaviour will be slightly different. To prevent issues for users, maybe wen can make this converter opt-in for now through an option? Example https://github.com/Koenkk/zigbee-herdsman-converters/blob/fac41963fe0a3cd62b5a3a3b21ea928e5b36944f/src/converters/toZigbee.ts#L67

ams2990: I would appreciate making it opt-in for now for an entirely unrelated reason -- I've built an integration around dealing with this, and it's going to take a while to updating it to work with Z2M, rather than work around it. Being able to toggle this behavior for testing (rather than having to upgrade/downgrade Z2M) would be a big timesaver.

ams2990:
With this PR, if I set color+brightness in a single MQTT message, will Z2M issue a proprietary Philips command to set both simultaneously, or will it continue to issue separate Zigbee MoveTo... commands?

Mihonarium:
@ams2990 with this PR, it's a single atomic command. `philipsLightTz` claims all light keys, Z2M calls it once, it builds a single `Philips2Data` payload with both fields (`brightness` + `colorXY`), and sends one `manuSpecificPhilips2.multiColor` command with both set in the same frame.

(Though I'll try to add opt-in, which means that by default, it'll keep issuing separate commands.)

src/lib/philips.ts
Koenkk: > maybe wen can make this converter opt-in for now through an option? 

@Mihonarium could you check this? I think it's the only thing preventing it from getting merged.

Mihonarium:
Okay, done! Please test that the special features work correctly if Philips2 native control hasn't been set to True

Mihonarium:
@Koenkk feel free to edit/change anything!

also cc @LaurentvdBos @chrivers

src/lib/philips.ts
```diff
@@ -492,15 +516,26 @@ const philipsModernExtend = {
                 }
             },
             options: [
+                new exposes.Binary("philips2_native_control", ea.SET, true, false).withDescription(
```
Koenkk: Maybe `hue_native_control` is a better name? People probably don't know what the `2` means here.

rohankapoorcom: I feel like this is going to cause confusion if it's handled as part of the expose for the light no matter how it's named.

is there a way that we could handle it either though the settings (specific) tab or globally for all hue products? In my specific case, I would want this enabled for all of my hue lights and I don't want to go turn it on manually for 20+ of them.

Mihonarium: Good thought! Updated.

Mihonarium: @rohankapoorcom it is going to be available in the interface in Settings (specific) tab.

I think if a user adds to their `configuration.yaml` this:
```
device_options:
  hue_native_control: true
  ```
  
then this should enable it on all Hue devices, without having to manually go through all of them

rohankapoorcom: Ah perfect! I must have misread where the expose was wired.

I would suggest that the change in configuration.yaml be added to the docs as well.

Thanks!

mundschenk-at: But is an `expose` really the right way to set a different backend implementation? I have my doubts. Unless I'm misunderstanding something, having a configuration option should be enough (maybe with an optional a per-device override).

Koenkk:
Thanks!

Mihonarium:
Yay!

By the way (I hope this is allowed? it is awesome and I think everyone should use it, but let me know if not, I'll remove it, it's obviously somewhat offtopic):

If you have more than three lights, I have a card that allows arranging them spatially on a 2D canvas to control arbitrary sets of lights in as few taps and as little attention as possible: https://github.com/Mihonarium/HASS-Spatial-Lights-Card/

It was inspired by the Hue app, where I could drag lots of lights around the color wheel. I wanted something at least as good; I think I made something a lot better.

This is my living room, with lights corresponding to where they physically are (and I can select one of the Hue bulbs and pick any of the effects added to presets for them in two taps):

image

src/lib/philips.ts
andrei-lazarov: Backend `options` are displayed under Settings (specific), like `transition`, not under Exposes .
Should be ok, I'll also test it

Mihonarium:
Also: I guess #8697, Koenkk/zigbee2mqtt#18406, Koenkk/zigbee2mqtt#15891, Koenkk/zigbee2mqtt#24438, Koenkk/zigbee2mqtt#31218 can be closed

src/lib/philips.ts
mundschenk-at: Ah, that should be fine then.

mundschenk-at:
Congratulations for getting this merged, I'm really looking forward to testing it!

Koenkk:
Had to revert this PR because it seems to break standard bulb control: https://github.com/Koenkk/zigbee2mqtt/actions/runs/24853454219/job/72759920745?pr=31783, e.g. [`Should publish messages to zigbee devices`](https://github.com/Koenkk/zigbee2mqtt/blob/2b485a98c5f9c879e1e9b80ffae3c7a84b0dce8d/test/extensions/publish.test.ts#L1656) fails. @Mihonarium could you check why it fails? (clone z2m, use `pnpm link ../zigbee-herdsman-converters && pnpm i` and then execute z2m tests with `pnpm test`)

Mihonarium:
(Continued in #12008)

Mihonarium:
(Fixed, needs updates to the tests in Koenkk/zigbee2mqtt#31785.)

Grandma-Betty:
This is huge! Thanks a bunch @Mihonarium

lundyfpv:
I'm noticing a few bulbs having issues when in a zigbee group. I can apply effects to them individually and they will work. If there is a group with multiple models some of them will apply while others do not. All of them accept the effect individually.

[5/3/2026, 9:11:01 PM] z2m: Publish 'set' 'effect' to 'group_bulbs_island' failed: 'Error: Status 'UNSUPPORTED_CLUSTER' manuSpecificPhilips2

Mihonarium:
Opened a draft PR to disable native control for mixed groups. idk if there's a way to send Hue-only effects to mixed groups; that would certainly be preferable if possible

Nixon506E:
I am seeing transitions in automations sometimes  being in seconds and sometimes in tenths of a second. I know the hue hub expected the latter but I am just not sure of the inconsistency.

lundyfpv:
> Opened a draft PR to disable native control for mixed groups. idk if there's a way to send Hue-only effects to mixed groups; that would certainly be preferable if possible

To be clear I'm seeing this issue with groups that contain only hue bulbs but they may be different feature sets. One group has 9290011998B (white ambiance) mixed with a 9290023351 (warm white). Individually they can be set but in a group it doesn't seem to work.

On a seperate note I have 3 9290024784 in a group together and they don't seem to accept effects as a group but do work individually. They also fail with the same error. [5/4/2026, 8:08:56 PM] z2m: Publish 'set' 'effect' to 'group_bulbs_island' failed: 'Error: Status 'UNSUPPORTED_CLUSTER' manuSpecificPhilips2'

One thing I notice is in the Z2M gui you have the option to select the effect on an individual device but not on a group under the expose tab. Would this have any effect on being able to use effects on groups?

lundyfpv:
Is there anything I can do to rectify 3 of the same hue bulb not accepting effects in a group?

PR #11656, opened 2026-03-04
fix(ignore): improve testing by using herdsman directly

Nerivec:
Mock the lower levels of ZH (Adapter, Database), and use real Device and Endpoint classes with necessary select mocks.

Also updates ZH to ensure https://github.com/Koenkk/zigbee-herdsman-converters/pull/11638#discussion_r2886038879 is fixed.


Koenkk:
Much better, thanks!

PR #11657, opened 2026-03-04
fix(DJT11LM): change sensitivity expose from numeric to enum (select)

iago-veiga:
## Problem

The Aqara Vibration Sensor DJT11LM (`lumi.vibration.aq1`) exposes `sensitivity` as a `numeric` type (range 1-21), which causes Home Assistant to register it as a `number` entity. However, the device state always contains sensitivity as a **string** (`"low"`, `"medium"`, `"high"`), causing a recurring HA warning:

```
WARNING homeassistant.components.mqtt.number
Payload '{"sensitivity":"medium",...}' is not a Number
```

## Root Cause

This asymmetry was introduced by commit [719f27f](https://github.com/Koenkk/zigbee-herdsman-converters/commit/719f27f06a52851e1909ee18d7a14bf35b32718c) (which changed from `e.enum` to `e.numeric` for more precise 1-21 control). However:

- The `fromZigbee` has **no handler** for attribute `65293` (sensitivity) — the device never reports sensitivity back via Zigbee attribute reports
- This means the state cache retains string values (from before the numeric change) and is **never updated** with a numeric value
- The expose says `number`, HA registers a `number` entity, but the published state always contains a string → warning

## Fix

Revert the expose back to `enum` with values `["low", "medium", "high"]`, consistent with what the device actually reports. Update the `lumi_vibration_sensitivity` toZigbee converter to:
- Only accept string values (`assertString`)
- Return the string value in the state (ensuring symmetry between expose type and published state)
- Still write the correct numeric value to the Zigbee attribute

The underlying Zigbee write still uses the correct numeric mapping (`low → 0x15`, `medium → 0x0b`, `high → 0x01`), so device behavior is unchanged.

## Changes

- `src/devices/lumi.ts`: `e.numeric("sensitivity", ea.STATE_SET)` → `e.enum("sensitivity", ea.STATE_SET, ["low", "medium", "high"])`
- `src/lib/lumi.ts`: `lumi_vibration_sensitivity` toZigbee now uses `assertString`, maps to numeric for Zigbee write, returns string in state

## References

- Issue: https://github.com/Koenkk/zigbee2mqtt/issues/3028
- Previous numeric change: commit `719f27f06a52851e1909ee18d7a14bf35b32718c`

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -2310,7 +2310,7 @@ export const definitions: DefinitionWithExtend[] = [
             e.vibration(),
             e.action(["vibration", "tilt", "drop"]),
             e.numeric("strength", ea.STATE),
-            e.numeric("sensitivity", ea.STATE_SET).withDescription("Sensitivity, 1 = highest, 21 = lowest").withValueMin(1).withValueMax(21),
+            e.enum("sensitivity", ea.STATE_SET, ["low", "medium", "high"]).withDescription("Sensitivity of the vibration sensor"),
             e.angle_axis("angle_x"),
             e.angle_axis("angle_y"),
             e.angle_axis("angle_z"),
--- a/src/lib/lumi.ts
+++ b/src/lib/lumi.ts
@@ -6171,13 +6171,12 @@ export const toZigbee = {
     lumi_vibration_sensitivity: {
         key: ["sensitivity"],
         convertSet: async (entity, key, value, meta) => {
-            if (isString(value)) {
-                value = getFromLookup(value, {low: 0x15, medium: 0x0b, high: 0x01});
-            }
-            assertNumber(value);
+            assertString(value, key);
+            const stringValue = value.toLowerCase() as "low" | "medium" | "high";
+            const numericValue = getFromLookup(stringValue, {low: 0x15, medium: 0x0b, high: 0x01});
             const options = {...manufacturerOptions.lumi, timeout: 35000};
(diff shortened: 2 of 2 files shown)


iago-veiga:
## Update: closing this PR

After further analysis, the root cause is simpler than initially described and does not require a code change.

### What actually happens

The `lumi_vibration_sensitivity` toZigbee converter already handles strings correctly: it maps `"medium"` → `0x0b` (11), writes the numeric value to Zigbee attribute 65293, and **returns `{state: {sensitivity: 11}}` (numeric)**. So after a SET, the state cache is correctly updated to a number.

The warning only persists for users whose `state.json` still contains `"sensitivity": "medium"` (a string) **from before commit `719f27f`** — i.e., when sensitivity was still an `e.enum`. Since the `fromZigbee` has no handler for attribute 65293 (the device doesn't actively report sensitivity back via Zigbee), that old cached string value is never overwritten with a numeric one.

### The actual fix for affected users

Simply publish to the device once:
```
zigbee2mqtt/<DEVICE_NAME>/set → {"sensitivity": "medium"}
```
This triggers the toZigbee converter, which updates the state cache to `{sensitivity: 11}`. The HA warning stops immediately.

### Why the code change is not needed

The numeric expose (1-21) is correct and intentional per the original commit. Reverting to `enum` would lose the fine-grained control that was added. The issue is purely a one-time state migration problem for users who upgraded without re-setting sensitivity.

Closing this PR. Apologies for the noise.

PR #11450, opened 2026-02-02
feat(add): Namron Panel Heater PRO 4512776/4512777

danielcherubini:
## Summary
- Adds support for Namron Panel Heater PRO devices (white 4512776 / black 4512777)
- Fixes issues from PR #11031 based on testing feedback from @corvy and @danielcherubini

## Key fixes from #11031
- **Fixed zigbeeModel** - Changed from `"Panel Heater PRO"` to `"Panel Heater"` (what device actually reports)
- **Fixed energy reporting** - Device reports energy 10x too high, added custom metering converter with corrected divisor

## Supported features
| Feature | Description |
|---------|-------------|
| Climate control | Local temperature, setpoint (5-35°C), system mode (off/heat/auto), running state |
| state | Virtual ON/OFF that maps to systemMode |
| frost_mode | Frost protection (HEAT + 7°C), restores previous mode/setpoint when disabled |
| child_lock | Lock/unlock physical buttons |
| hysteresis | Adjustable 0.5-5.0°C for on/off control |
| window_open_detection | Enable/disable window detection |
| window_open | Window open status (read-only) |
| control_method | Select between PID or hysteresis control |
| adaptive_function | Adaptive preheat on/off |
| pid_kp/ki/kd | PID controller parameters (0-1) |
| display_brightness | Read-only (1-7) |
| display_auto_off | Display auto-off setting |
| power/energy | With corrected energy divisor |

## Testing
- Implementation based on @corvy's tested external converter from https://github.com/Koenkk/zigbee2mqtt/issues/30170#issuecomment-3748568861
- Tested by @danielcherubini with real hardware

## Related
- Supersedes #11031
- Closes https://github.com/Koenkk/zigbee2mqtt/issues/30170
- Closes https://github.com/Koenkk/zigbee-herdsman-converters/pull/11031
- Device images: https://github.com/Koenkk/zigbee2mqtt.io/pull/4799

PR #11662, opened 2026-03-05
fix: Namron Panel Heater PRO power/energy reporting

danielcherubini:
## Summary

Use `m.electricityMeter` with targeted options to work around device limitations:

```typescript
extend: [m.electricityMeter({cluster: "both", energy: {divisor: 10}, power: false, voltage: false, current: false, configureReporting: false})],
```

- `configureReporting: false` — device doesn't support `configureReporting` on `seMetering` (`UNSUPPORTED_ATTRIBUTE`)
- `power: false` — device doesn't support `instantaneousDemand` on `seMetering`; power comes from `haElectricalMeasurement` (`activePower`) via `fz.electrical_measurement`
- `voltage: false, current: false` — device doesn't report these
- Energy divisor saved manually in configure: `endpoint.saveClusterAttributeKeyValue("seMetering", {divisor: 10, multiplier: 1})`
- `e.power()` added to exposes since `power: false` removes it from the extend

Fixes the issues reported in https://github.com/Koenkk/zigbee-herdsman-converters/pull/11450#issuecomment-4005613940:
1. **Energy not divided by 10** — raw values passed through (delta 9 kWh/hr instead of 0.9)
2. **Power always 0** — even when heater is actively running

## Test plan
- [x] All existing tests pass (1 pre-existing failure unrelated to this PR)
- [x] Tested on device — energy correctly divided by 10, power reports non-zero when heating


src/devices/namron.ts
```diff
@@ -1363,8 +1386,13 @@ export const definitions: DefinitionWithExtend[] = [
         model: "4512776/4512777",
         vendor: "Namron",
         description: "Zigbee thermostat for panel heater PRO (white 4512776 / black 4512777)",
-        extend: [m.electricityMeter({cluster: "metering", energy: {divisor: 10}})],
```
Koenkk, line 1366: Could you try with just this change?

```
extend: [m.electricityMeter({cluster: "metering", energy: {divisor: 10}, configureReporting: false})],
```

danielcherubini: Nope, sadly that didn't work. The issue is its still showing 0w

svhelge: try this
```
            m.electricityMeter({
                cluster: "both",
                energy: {divisor: 10},
                power: {cluster: "electrical"},
                configureReporting: false,
            }),
```

danielcherubini: Tested with power: false combined with cluster: "both":                                                                                                                                       
 ```                                                                                                                                                                                               
  extend: [m.electricityMeter({cluster: "both", energy: {divisor: 10}, power: false, voltage: false, current: false, configureReporting: false})],                                              
  fromZigbee: [fz.thermostat, fzLocal.namron_panelheater, fz.namron_hvac_user_interface, fz.electrical_measurement],                                                                            
```
  - configureReporting: false — device doesn't support configureReporting on seMetering at all
  - power: false — device doesn't support instantaneousDemand on seMetering, power comes from haElectricalMeasurement (activePower) via fz.electrical_measurement
  - voltage: false, current: false — device doesn't report these
  - Energy divisor saved manually in configure: endpoint.saveClusterAttributeKeyValue('seMetering', {divisor: 10, multiplier: 1})
  - e.power() added to exposes since power: false removes it from the extend

  Energy and power both working correctly now. Will push the updated code.

Comment (danielcherubini) [reply src/devices/namron.ts:1366]: Note: the updated commit includes a few changes beyond just the `extend` line, all necessary consequences of `configureReporting: false` and `power: false`:

- **`e.power()`** added to exposes — `power: false` removes it from the extend, but we still need it for `fz.electrical_measurement`
- **`saveClusterAttributeKeyValue("seMetering", {divisor: 10, multiplier: 1})`** in configure — with `configureReporting: false`, the extend no longer saves the divisor automatically, so `fz.metering` would get `divisor=null` and not divide
- **`"seMetering"` added to binds** — needed so the device sends energy attribute reports
- **Simplified reads** — just read `activePower` and `currentSummDelivered` for initial values, removed `reporting.activePower()` and `acPowerMultiplier`/`acPowerDivisor` reads since we're not using configureReporting

Comment (svhelge) [reply src/devices/namron.ts:1366]: did you try `power: {cluster: "electrical"},` ?
You should then be able to skip `fz.electrical_measurement` and `e.power()`, hence making the converter a bit more modern.

Comment (danielcherubini) [reply src/devices/namron.ts:1366]: Yes, I tried `power: {cluster: "electrical"}` first. Configure succeeds, but when you press the refresh button on power in the Z2M UI, the toZigbee still tries to read `instantaneousDemand` from seMetering:

```
Publish 'get' 'power' to 'Dining Room Heater' failed: 'Error: ZCL command seMetering.read(["instantaneousDemand"]) failed (Status 'UNSUPPORTED_ATTRIBUTE')'
```

So `power: false` is needed to fully suppress the seMetering power path.

svhelge: @Koenkk, I believe this is an error in modernExtend ?

Koenkk: Yes that looks like a mistake indeed, could you correct it?

danielcherubini: Sorry to ask, are you asking me to correct it @Koenkk ?

Koenkk: This is fine for now, thanks!

PR #11663, opened 2026-03-05
fix: ZBeacon TH01-z and TH01-2-z

notgood:
1. Add TH01-z
2. Move TH01-2-z from zbeacon.ts to zigbeetlc.ts

Both model ID's are used by custom ZigbeeTLc firmware by pvxx.
These devices support temperature/humidity calibrations and custom measurement interval.

https://github.com/pvvx/ZigbeeTLc/blob/master/src/board_zbeacon_th01.h#L29
https://github.com/pvvx/ZigbeeTLc/blob/master/src/board_zbeacon2_th01.h#L29

Run tested on ZBeacon TH01-2-z.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "dependencies": {
         "iconv-lite": "^0.7.2",
         "semver": "^7.7.4",
-        "zigbee-herdsman": "^9.0.10"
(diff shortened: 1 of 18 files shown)


Koenkk:
Thanks!

PR #11664, opened 2026-03-05
fix: sonoff smart plug: use STATE_SET access for overload protection properties

jakobhellermann:
The fields are exposed in `fromZigbee`, so they can be advertised as such.


Koenkk:
Thanks!

PR #11665, opened 2026-03-05
IKEA VINDSTYRKA - label, unit, remove identify

andrei-lazarov:
1. Unit and label fixes https://github.com/Koenkk/zigbee2mqtt/issues/30970

2. Identify has no visible effect

3. When the device boots, it displays and reports 5-8 degrees higher temperature than it should. Then it slowly lowers to normal.
@Koenkk can we somehow ignore the temperature readings for 3-5 minutes after the device boots?

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -1169,7 +1169,6 @@ export const definitions: DefinitionWithExtend[] = [
             m.humidity(),
             m.pm25({reporting: {min: "1_MINUTE", max: "2_MINUTES", change: 2}}),
             ikeaVoc(),
-            m.identify(),
         ],
         ota: true,
     },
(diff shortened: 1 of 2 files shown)


Koenkk:
> @Koenkk can we somehow ignore the temperature readings for 3-5 minutes after the device boots?

This would be tricky, not sure if it's really important because I don't expect you reboot this device often? If you want to ignore it, I think you can handle a `deviceAnnounce` event in the `onEvent`, store that in the `globalStore` and then in the temperature fromZigbee converter skip the `return` when the `deviceAnnounce`  was done recently.

Koenkk:
Thanks!

andrei-lazarov:
Indeed it's not very important. I'm not rebooting it. I'll add it to docs for now

PR #11671, opened 2026-03-07
fix: match new cluster-related typing from ZH

Nerivec:
Requires https://github.com/Koenkk/zigbee-herdsman/pull/1663/

First commit is from running a [script](https://gist.github.com/Nerivec/493cf40dccd645cc8e1a8eac97808fc4#file-zhc-add-device-custom-cluster-names-ts) to automatically add `name` props to all custom clusters.
Second is manual fixes:
- replace `*Definition` types
- ubisys: access custom cluster directly instead of doing lookup
- philips: fix race condition with custom clusters addition & lookup


Koenkk:
Thanks!

PR #11675, opened 2026-03-07
Move cluster manuSpecificSmartThingsArrivalSensor from ZH to ZHC

svhelge:
Moved converter fz.PGC410EU_presence from fromZigbee.ts to fzLocal. 
Added cluster manuSpecificSmartThingsArrivalSensor to smartthings.ts (modernextend) and updated device.


Koenkk:
Thanks!

PR #11679, opened 2026-03-07
Move manuSpecificTuya* clusters from ZH to ZHC

svhelge:
Moved converters using the custom clusters to lib/tuya.ts. 
Added custom clusters to lib/tuya.ts. 


src/lib/tuya.ts
```diff
@@ -35,7 +35,133 @@ interface KeyValueStringEnum {
...
+        mcuVersionRequest: {
+            seq: number;
+        };
+        sendData: {
+            seq: number;
+            dpValues: TuyaDataPointValue[];
+        };
+        mcuOtaNotify: {
+            seq: number;
+            key_hi: number;
```
svhelge, line 56: @Koenkk , convert to camelCase or 
`// biome-ignore lint/style/useNamingConvention: ignored using `--suppress`` ?

Koenkk: Lets `biome ignore` it

svhelge:
@Koenkk , what is best way to implement the custom clusters for the Tuya devices? 
In general I believe it's a good practice that it's possible to "read" which custom clusters are used in the device definition. (That's how I have done most of the other adjustments). 
But adding the cluster definitions to all the Tuya devices with precision is not straightforward, unless it's somehow possible to identify which devices that require the custom clusters definition (ie by adding `extend: [tuya.clusters.addManuSpecificTuyaCluster]`)
An other option is to keep it safe, find a way to add the three clusters to all Tuya devices?
What do you think?

Nerivec:
Koenkk will no doubt know this better, but there's a `tuyaBase` modern extend (same one that serves to configure magic packet), might help to cut the work down. The helper `configureMagicPacket` (and its few wrappers) might also help locate all "basic tuya devices".
In/out clusters during interview could help in normal circumstances to pick which need to be added, but Tuya... can't really trust anything reported by the devices :sweat_smile: Probably safer to add all by default, and then cleanup gradually (i.e. should be implemented in a configurable way)

Koenkk:
Tuya is indeed very risky, I'm wondering if we should not just keep the clusters in zigbee-herdsman, there are simply too many devices.

Nerivec:
We just need to tie it in same as where the basic tuya stuff happens, not on a per-converter basis.
_Other Tuya customs are already in ZHC, doesn't make sense to keep some in ZH._

Koenkk:
Then we at least have to make sure the definition generator adds the custom cluster, otherwise adding support for new tuya devices will become more complicated


Edit: this will also break all external converters that are not using the modernExtend tuya base, so not really sure if we should go that route.

svhelge:
Why not add the custom clusters with extend to all devices? A bit of work, but kind of "standard". One function for all clusters, and maybee additional functions for each cluster.

Koenkk:
> Why not add the custom clusters with extend to all devices?

Because there are a lot, I'm wondering if we cannot automatically add it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/index.ts#L495 (when the device has a tuya manufacturer code)

Nerivec:
Can we rely on regex-matching against manuf name `_TZ` or tuya manuf code alone?

andrei-lazarov:
There are a few outliers.. 
I know `HOBEIAN` and `Zbeacon` for manuf name, other than the `_TZ` standard, surely there's more.
And for manuf code, there is the usual `4417`, but I also have `4407`, `4107`, `4098`, `4169` in the [custom fw database](https://github.com/romasku/tuya-zigbee-switch/blob/main/device_db.yaml)

Koenkk:
I thought that maybe always adding when it has `0xef00` as an input cluster is an option, but this only works when the interview passed (which might not always be the case)

Nerivec:
I think tuya 2,3,4 are far less used than 1, might be worth making separate PRs, then we can see about 1...

svhelge:
I created PR for manuSpecificTuya2 and manuSpecificTuya3, Tuya4 is already moved (ie. not in ZH),
See https://github.com/Koenkk/zigbee-herdsman-converters/pull/11752 and https://github.com/Koenkk/zigbee-herdsman-converters/pull/11751

svhelge:
I close this PR.
manuSpecificTuya2 and manuSpecificTuya3 have been moved from ZH to ZHC in other PRs. 
It remains to find out how to move the manuSpecificTuya cluster.

PR #11678, opened 2026-03-07
Shelly SBBT-104CUS: Add support for double, triple, and long events 

Estrobeda:
The BTHome edition of the Shelly SBBT-104CUS support at least the following 

- 1_single
- 2_single
- 3_single
- 4_single
- 1_double (new)
- 2_double (new)
- 3_double (new)
- 4_double (new)
- 1_tripple (new)
- 2_tripple (new)
- 3_tripple (new)
- 4_tripple (new)
- 1_hold
- 2_hold
- 3_hold
- 4_hold
- 1_long (new)
- 2_long (new)
- 3_long (new)
- 4_long (new)

Note: This is my first attempt to add extra functionality for any zigbee device. I've used ChatGPT and the docs at https://shelly-api-docs.shelly.cloud/docs-ble/Devices/BLU_ZB/button4_ZB/?utm_source=chatgpt.com

also note that this code is not tested, I've made an external converter for my local z2m that does work, I'd be more than happy to try this PR if I knew how :)

PR #11688, opened 2026-03-09
Shelly remotes: Add toggle to 4 buttons. Add long press to 1 button

andrei-lazarov:
- https://github.com/Koenkk/zigbee2mqtt/issues/31069
- https://github.com/Koenkk/zigbee2mqtt/issues/30941
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/11678

Thanks @Estrobeda! The same scenes should work on the Wall Switch as well.
I renamed long to single_long, and I added toggle mode. Is it ok?


Estrobeda:
I prefer your updated version so it's ok from my point of view at least :)

Koenkk:
Thanks!

PR #11690, opened 2026-03-09
Add support for Avatto N-TVR16 thermostatic valve

Shamane2000:
Added support for the Avatto N-TVR16 radiator thermostat. Manufacturer ID: _TZE204_vjpaih9f

Verified functionality via external converter:
- System Mode (Off/Heat mapped to DP 2)
- Presets (Manual, Schedule, Night, Holiday, Frost, Boost mapped to DP 2)
- Target Temperature (DP 4)
- Local Temperature (DP 5)
- Battery (DP 6)
- Child Lock (DP 7)
- Running State (DP 101)

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4900

### DIFF
--- a/src/devices/avatto.ts
+++ b/src/devices/avatto.ts
@@ -615,7 +615,6 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Thermostatic radiator valve",
         fromZigbee: [tuya.fz.datapoints],
         toZigbee: [tuya.tz.datapoints],
-        onEvent: tuya.onEventSetTime,
         configure: tuya.configureMagicPacket,
         exposes: [
             e


Koenkk:
It seems this device is already supported, does it not work out of the box with the latest z2m version?

https://github.com/Koenkk/zigbee-herdsman-converters/blob/953fcb96400590678fe75593bfd4b33ecc206d3a/src/devices/tuya.ts#L23270

Shamane2000:
Hi @Koenkk ,i am so Sorry, i've had a VERY old relaes due a backup issue! :-O 
(Zigbee2MQTT-Version [1.42.0](https://github.com/Koenkk/zigbee2mqtt/releases/tag/1.42.0) so, any change from my side have to be deleted. :-O 

But a question - I have a white label vendor device (avatto) - do we maintenance all white label under the tuya.ts? Then we should leave a comment with all known fingerprints  in the avatto.ts maybe?

PR #11693, opened 2026-03-10
fix(tuya): fix double trigger of state on single button click for 8 gang switch

rhakbari:
fix for double trigger of state on single button click for 8 gang switch in Z2MQTT

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1,3 +1,5 @@
+import {Buffer} from "node:buffer";
+import {setTimeout} from "node:timers";
 import {TextDecoder, TextEncoder} from "node:util";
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -15218,6 +15220,125 @@ export const definitions: DefinitionWithExtend[] = [
         model: "TS0601_switch_8",
         vendor: "ZYXH",
         description: "8 gang switch",
+        fromZigbee: [tuya.fz.datapoints],
+        toZigbee: [
+            {
+                key: ["state"],
```
Koenkk, line 15226: I don't understand how this fixes the issue, this code is used when sending commands to the device, not when receiving messages, could you provide the debug log when receiving the duplicate event?

PR #11705, opened 2026-03-11
Allow to set gas counter value via custom attribute extension

IgnacioHR:
This pull request adds the functionality to set the counter value via a custom attribute extension to the seMetering cluster.


Koenkk:
Thanks!

PR #11720, opened 2026-03-13
feat: add Zemismart TS0726 2-gang switch (_TZ3000_icoxotza)

razshlomo:
Add support for Zemismart TS0726 2-gang switch with manufacturer name `_TZ3000_icoxotza`.

This is a plain 2-gang switch (no scene mode), distinct from the existing `TS0726_2_gang` scene switch variants.

Key notes:
- `tuyaMagicPacket()` is required for independent gang control (without it, both gangs toggle together when controlled remotely while physical buttons work fine)
- Device does not have `manuSpecificTuya` (0xEF00) cluster, only `manuSpecificTuya3` and cluster `57344` (0xE000)
- `switchType` and `backlightMode` return `UNSUPPORTED_ATTRIBUTE`
- Both gangs work independently with this definition
- Physical switches report correct per-endpoint state

Tested on Z2M Edge v2.7.2-1.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -10406,33 +10406,23 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint("TS0726", ["_TZ3000_icoxotza"]),
-        model: "TS0726_2_gang_switch",
-        vendor: "Zemismart",
-        description: "2-gang switch",
-        extend: [
-            tuya.modernExtend.tuyaMagicPacket(),
-            m.deviceEndpoints({endpoints: {l1: 1, l2: 2}}),
-            tuya.modernExtend.tuyaOnOff({
-                endpoints: ["l1", "l2"],
-                powerOnBehavior2: false,
-            }),
-        ],
-        meta: {multiEndpoint: true},
-    },
-    {
-        fingerprint: tuya.fingerprint("TS0726", ["_TZ3002_1s0vfmtv", "_TZ3002_gdwja9a7", "_TZ3002_u7d3nes3"]),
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -10405,6 +10405,21 @@ export const definitions: DefinitionWithExtend[] = [
...
+            m.deviceEndpoints({endpoints: {l1: 1, l2: 2}}),
+            tuya.modernExtend.tuyaOnOff({
+                endpoints: ["l1", "l2"],
+                powerOnBehavior2: false,
+            }),
+        ],
+        meta: {multiEndpoint: true},
+    },
     {
         fingerprint: tuya.fingerprint("TS0726", ["_TZ3002_1s0vfmtv", "_TZ3002_gdwja9a7", "_TZ3002_u7d3nes3"]),
```
Koenkk: Does it also work with this definition?

razshlomo: No for me. 
`tuya.modernExtend.tuyaMagicPacket()` must be there

Koenkk: It can be added to `TS0726_2_gang`

razshlomo: The existing TS0726_2_gang exposes switch_mode and action which this device doesn't support — it's a plain 2-gang on/off switch with no scene functionality. Merging it in would expose unsupported features. That's why I kept it as a separate definition.
`'Error: ZCL command <device_ieee>/1 genOnOff.read(["tuyaOperationMode"], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"reservedBits":0,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')`

Koenkk:
Thanks!

PR #11723, opened 2026-03-14
feat(add): SR-ZG9032A-PIR add manufacturerCode to private attributes

niracler:
Add zigbeeCommandOptions with manufacturerCode (0x1224) to all 12 manufacturer-specific attributes on the SR-ZG9032A-PIR converter.

Without manufacturerCode, most private attributes (0x8800+) on genBasic cluster return null when read. With manufacturerCode properly set, all attributes read/write correctly.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4918


Koenkk:
Thanks!

PR #11726, opened 2026-03-15
Add the missing parts to Schneider Electric WDE002497 smart thermostat

alump:
- Schneider Electric WDE002497 has all the same properties as EKO07259 thermostat has currently defined. So copying definitions from there to add display brightness, power consumption and Normal/Pi control model selections.
- I tested these with my own WDE002497 thermostat and master branch Z2M


Koenkk:
Thanks!

PR #11728, opened 2026-03-15
Add meta property to support duplicate transactions

### LINKED ISSUE #11727: Duplicate transaction detection breaks for deep-sleep Zigbee devices with reset sequence numbers
I was struggling to understand the reason behind the following two lines in the log:

```
{"level":"debug","message":"z2m: Received Zigbee message from '0xe4b323fffeb58ab0', type 'attributeReport', cluster 'seMetering', data '{\"currentSummDelivered\":9127,\"extendedStatus\":\"0\",\"status\":0}' from endpoint 1 with groupID 0","timestamp":"2026-03-15 19:13:57"}
{"level":"info","message":"z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/0xe4b323fffeb58ab0', payload '{\"battery\":80.5,\"battery_low\":false,\"extended_status\":\"0\",\"gas\":91.25,\"last_seen\":\"2026-03-15T18:13:57.993Z\",\"linkquality\":255,\"status\":0,\"update\":{\"installed_version\":16843009,\"latest_release_notes\":null,\"latest_source\":\"ota/ZigbeeGasMeter.ota\",\"latest_version\":16843009,\"state\":\"idle\"},\"voltage\":4200}'","timestamp":"2026-03-15 19:13:57"}
```

The Zigbee device is designed to run in **deep sleep** mode in order to minimize battery consumption.

The device wakes up every 15 minutes, sends a new report message, and then goes back to sleep. It can also wake up from an I/O interrupt connected to the gas counter; in that case it only increments the internal counter and goes back to sleep without sending a report.

Because the device always starts from a fresh boot after deep sleep, the ZCL transaction sequence number always starts at 1.

As a result, the code here:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/a0e60b1a64d9a1f2d178cd47bded5777511f2fdb/src/converters/fromZigbee.ts#L905

considers the transaction to have already been processed and skips it, so the conversion result is not forwarded to MQTT.

There are a couple of ways this situation could potentially be handled:
	1.	Set the `publishDuplicateTransaction` flag to true for this device so the converter is always executed.
This works well for deep-sleep devices where the transaction sequence number resets on every wake cycle.
	2.	Alternatively, the duplicate-transaction detection logic could take the **timestamp of the last processed transaction** into account instead of relying only on the transaction sequence number.
For devices that reboot frequently (like deep-sleep battery devices), this could avoid incorrectly discarding valid reports.

The device is DIY, and users can also choose to compile the firmware in **light sleep** mode instead of deep sleep. This consumes more power but preserves the transaction sequence numbers across wake cycles, allowing the existing duplicate-transaction optimization to work as intended.

The main reason for opening this issue is to document the behavior and discuss possible approaches before proposing a change.

If the maintainers agree that using a timestamp (or a similar strategy) would be a reasonable improvement to the duplicate-transaction logic, I would be happy to prepare a PR in that direction.

Thank you!

Cheers,
Ignacio

IgnacioHR:
Fixes #11727 for my gas counter

### DIFF
--- a/src/devices/custom_devices_diy.ts
+++ b/src/devices/custom_devices_diy.ts
@@ -1323,5 +1323,8 @@ export const definitions: DefinitionWithExtend[] = [
                 valueMax: 281474976710655,
             }),
         ],
+        meta: {
+            publishDuplicateTransaction: true,
+        },
     },
 ];


Koenkk:
Thanks!

PR #11729, opened 2026-03-15
Move Sinope hvacThermostat attributes from ZH to ZHC

svhelge:
Added hvacThermostat attributes with modernextend. Updated devices with attribute definitions.


Koenkk:
Thanks!

PR #11746, opened 2026-03-16
NodON FPS-4-1-00:  expose humidity endpoint

KipK:
Device has working humidity endpoint we can see with developper tools. So let expose it.


Koenkk:
Thanks!

PR #11751, opened 2026-03-19
fix: Move manuSpecificTuya3Cluster from ZH to ZHC

svhelge:
Defined cluster in lib/tuya.ts and updated devices with cluster.
90 devices updated, based on the following:

- devicess with fz/tz converters using manuSpecificTuya3
- devices using tuyaOnOff or tuyaLight (which also use manuSpecificTuya3)

EDIT:
Defined cluster in lib/tuya.ts and moved cluster "definition" to tuyaBase. 
Added tuyaBase to 90 devices.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1595,7 +1595,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "TS0111",
         vendor: "Tuya",
         description: "Socket",
-        extend: [tuya.clusters.addManuSpecificTuya3Cluster(), tuya.modernExtend.tuyaOnOff()],
+        extend: [tuya.modernExtend.tuyaBase(), tuya.modernExtend.tuyaOnOff()],
     },
     {
         zigbeeModel: ["TS0218"],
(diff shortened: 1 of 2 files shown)


src/devices/tuya.ts
```diff
@@ -1595,7 +1595,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "TS0111",
         vendor: "Tuya",
         description: "Socket",
-        extend: [tuya.modernExtend.tuyaOnOff()],
+        extend: [tuya.clusters.addManuSpecificTuya3Cluster(), tuya.modernExtend.tuyaOnOff()],
```
Koenkk: Maybe we can add this to `tuyaBase`? (https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/tuya.ts#L2601)

svhelge: Add an argument which defaults to true?

Koenkk: I would just always add it (this is basically also what happens now)

svhelge: Ok, and then adding `tuya.modernExtend.tuyaBase` to the devices (replacing `tuya.clusters.addManuSpecificTuya3Cluster`)?

Koenkk:
Thanks!

PR #11758, opened 2026-03-19
Add QA QAFZ200, QASZP, update QADZ4DIN, fix QAT44Z6 scenes

queroautomacao:
## Summary
- **QAFZ200** (TS0502B `_TZ3218_op6ztaju`): CCT light controller with custom off handler for proper turn-off behavior at 0% brightness
- **QASZP** (TS011F `_TZ3218_kwht8j5m`): Power sensor with electrical measurement polling, reactive power threshold, and Tuya DP support via `sendData` command
- **QADZ4DIN** (TS0601 `_TZE284_nthhgkd6`): Fix fingerprint, add dimming speed control (slow/middle/fast), limit switch_type to momentary/toggle only
- **QAT44Z6** (TS0601 `_TZE284_ms97nkyy`): Fix scene actions to use `e.action()` with `tuya.valueConverter.static()` so scenes appear as scene_1-6 in Home Assistant (matching QAT42Z3B behavior)

## Test plan
- [x] QAFZ200: ON/OFF, brightness, CCT, 0% brightness fully turns off
- [x] QASZP: Voltage/current/power polling, threshold settings, switch status
- [x] QADZ4DIN: 4-channel dimming, switch type, dimming speed, power on behavior
- [x] QAT44Z6: Scene actions appear as scene_1-6 in Home Assistant automations

### DIFF
--- a/src/devices/qa.ts
+++ b/src/devices/qa.ts
@@ -6,7 +6,6 @@ import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
 import * as tuya from "../lib/tuya";
 import type {DefinitionWithExtend} from "../lib/types";
-import * as utils from "../lib/utils";
 
 const e = exposes.presets;
 const ea = exposes.access;


src/devices/qa.ts
```diff
@@ -408,4 +411,75 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint("TS0502B", ["_TZ3218_op6ztaju"]),
+        model: "QAFZ200",
+        vendor: "QA",
+        description: "CCT light controller",
+        extend: [tuya.modernExtend.tuyaLight({colorTemp: {range: [153, 500]}, effect: false, doNotDisturb: false})],
+        toZigbee: [
```
Koenkk: Why is this needed? Is `tuya.modernExtend.tuyaLight` not enough?

queroautomacao: In my testing I had to disable "effect: false, doNotDisturb: false" that was generating problems.

Koenkk: Can you try to remove the entire `toZigbee` session and see if everything still works?

queroautomacao: If I remove this code, the device does not turn off when dim value = 0%.

Koenkk: That's expected behaviour, in this case z2m will use a brightness of 1: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/converters/toZigbee.ts#L1352 , so you should send `{"state": "OFF"}` instead.

queroautomacao: Done.

src/devices/qa.ts
```diff
@@ -306,7 +301,7 @@ export const definitions: DefinitionWithExtend[] = [
         exposes: [e.power_on_behavior(), tuya.exposes.switchType()],
     },
     {
-        fingerprint: tuya.fingerprint("TS0601", ["_TZE204_nthhgkd6"]),
```
Koenkk, line 309: This means support for `_TZE204_nthhgkd6`, I think that is not expected?

queroautomacao: The correct device number is: "TS0601", ["_TZE284_nthhgkd6"]

Koenkk:
Thanks!

PR #11765, opened 2026-03-20
Add additional zigbeeModel for KAJPLATS E27 bulb

tklevenz:
This is about [this device](https://www.ikea.com/nl/en/p/kajplats-led-bulb-e27-806-lumen-smart-white-spectrum-clear-globe-30611481/), showed up with a different model name for me.

### DIFF
--- a/src/devices/ikea.ts
+++ b/src/devices/ikea.ts
@@ -666,7 +666,7 @@ export const definitions: DefinitionWithExtend[] = [
     },
     {
         // https://github.com/Koenkk/zigbee2mqtt/issues/30211#issuecomment-3698674562
-        zigbeeModel: ["KAJPLATS E27 WS G95 clear 806lm"],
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
Thanks. I'm curious:
- Does power-on-behavior work for yours? 
It gives unsupported_attribute on my `KAJPLATS E27 WS G95 clear 806lm` (only in Zigbee mode)
- What's its IEEE address? to determine what chip it uses
- What fw version is it running? (You can connect to a Matter hub and update) 
Mine had 1.0.28, now 1.1.0

tklevenz:
I had it previously conmnected via matter, don't remember if I updated it... everything seems to work for me so far.

IEEE: 0x00f0fd45fff16696
FW: 1.1.0

Koenkk:
Thanks!

PR #11768, opened 2026-03-20
fix(shelly): fix seMetering FAILURE aborting configure and improve reporting intervals for Gen 4 PM devices

### LINKED ISSUE #11767: Shelly 1PM Gen 4: seMetering configureReporting FAILURE breaks configure, energy values never auto-reported
## Problem

Two issues with the Shelly 1PM Gen 4 (`S4SW-001P16EU`, firmware `20260120-145255/1.7.4`):

**1. seMetering configureReporting is rejected by the firmware**

The device returns `Status 'FAILURE'` when Z2M tries to configure reporting for the `seMetering` cluster. Since this exception is unhandled, the entire configure step is aborted — meaning `haElectricalMeasurement` never gets properly configured either, and `meta.configured` stays `0` so Z2M keeps retrying forever.

```
Failed to configure 'Device', attempt 1
ZCL command seMetering.configReport([...]) failed (Status 'FAILURE')
```

**2. Default MAX interval (65000 s / ~18 h) is too long for energy monitoring**

With a stable load like a heating power supply at 7–8 W, the `activePower ±5 W` change threshold is never crossed. Result: no values reported all day, only when pressing manual refresh.

This is a catch-22: even if you fix the intervals manually via the frontend, Z2M resets them on the next configure attempt because of issue 1.

## Fix (PR attached)

- `modernExtend.ts`: catch `configureReporting` failures per-cluster and log a warning instead of aborting. Same pattern as the existing try/catch on `endpoint.read`. Also helps with the `INVALID_DATA_TYPE` issue on `powerFactor` already noted in `shellyPowerFactorInt16Fix()`.
- `shelly.ts`: custom configure for S4SW-001P16EU and S4SW-001P8EU that skips `seMetering.configureReporting` and uses 300 s max interval for `haElectricalMeasurement` instead of MAX.

leramleram:
Fixes #11767

## Changes

**`modernExtend.ts`**

Wrap `configureReporting` per-chunk in try/catch. A single cluster rejection (e.g. `FAILURE` from `seMetering` on Shelly Gen 4, or `INVALID_DATA_TYPE` from `powerFactor`) no longer aborts configuration of all other clusters. A warning is logged. Same pattern as the existing try/catch on `endpoint.read`.

**`shelly.ts`** — three new helpers for Gen 4 PM devices (`S4SW-001P16EU`, `S4SW-001P8EU`):

- `shellyGen4ElectricityMeterConfigure`: uses `maximumReportInterval: 300 s` (5 min) instead of `MAX` (65000 s / ~18 h) for `haElectricalMeasurement`; omits `seMetering.configureReporting` since the firmware rejects it
- `shellyGen4EnergyPoll`: polls `seMetering.currentSummDelivered` and `currentSummReceived` every 300 s so energy values stay current without relying on firmware push reports (interval is user-configurable via `energy_poll_interval` device option)
- `shellyGen4EnergyZeroFilter`: filters out spurious `currentSummDelivered=0` reports from the Shelly firmware — when a valid previous energy value is known, the zero is replaced by the last good value to prevent Home Assistant energy statistics corruption

## Tested on

Shelly 1PM Gen 4 (`S4SW-001P16EU`), firmware `20260120-145255/1.7.4`, Z2M 2.4.0, Ember adapter (EFR32MG24)

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://biomejs.dev/schemas/2.4.7/schema.json",
+    "$schema": "https://biomejs.dev/schemas/2.4.8/schema.json",
     "vcs": {
         "enabled": true,
         "clientKind": "git",
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -338,6 +338,27 @@ const shellyModernExtend = {
             {attribute: "acFrequency", minimumReportInterval: 10, maximumReportInterval: 300, reportableChange: 100},
         ]);
     },
+    // Filters out spurious energy=0 reports from Shelly Gen 4 firmware.
+    // The firmware occasionally emits currentSummDelivered=0 which would corrupt
+    // Home Assistant energy statistics (treated as a meter reset). When a valid
+    // previous energy value is known, the zero is replaced by the last good value.
+    shellyGen4EnergyZeroFilter(): ModernExtend {
+        const fzFilter: Fz.Converter<"seMetering", undefined, ["attributeReport", "readResponse"]> = {
+            cluster: "seMetering",
+            type: ["attributeReport", "readResponse"],
+            convert: (model, msg, publish, options, meta) => {
+                if (msg.data.currentSummDelivered !== undefined) {
+                    const multiplier = (msg.endpoint.getClusterAttributeValue("seMetering", "multiplier") as number) || 1;
+                    const divisor = (msg.endpoint.getClusterAttributeValue("seMetering", "divisor") as number) || 1;
+                    const energy = (msg.data.currentSummDelivered * multiplier) / divisor;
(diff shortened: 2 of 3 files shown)


Koenkk:
> A single cluster rejection (e.g. FAILURE from seMetering on Shelly Gen 4, or INVALID_DATA_TYPE from powerFactor) no longer aborts configuration of all other clusters. A warning is logged. Same pattern as the existing try/catch on endpoint.read.

This is a bit dangerous as a configure failure is now suppressed (I expect that almost nobody will check the logs during the configure). Can we adapt the `m.electricityMeter()` instead?

leramleram:
I agree that this is maybe overshooting. I found that my change needs to be revised anyways because it sometimes forwards zero energy sum values. I´m off for a few days, I will apply the changes when I´m back and have time to test this thoroughly.

PR #11772, opened 2026-03-21
feat: add msh converters

SkeLLLa:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4946

Adding support for my smart house devices:

* CO2 sensor
* Temperature sensor based on bme280
* Temperature sensor based on ds18b20

### DIFF
--- a/src/devices/msh.ts
+++ b/src/devices/msh.ts
@@ -28,7 +28,7 @@ export const definitions: DefinitionWithExtend[] = [
         zigbeeModel: ["msh.AirQMon"],
         model: "msh.AirQMon",
         vendor: "MySmartHouse",
-        description: "[MSH Air quality sensor](https://www.facebook.com/my.smart.house.in.ua)",
+        description: "MSH Air quality sensor",
         fromZigbee: [fz.ptvo_switch_uart, fz.co2, fz.ptvo_switch_analog_input, fz.temperature],
         toZigbee: [tz.ptvo_switch_trigger, tz.ptvo_switch_uart, tzLocal.ptvo_senseair_calibration],
         exposes: [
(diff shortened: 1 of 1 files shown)


SkeLLLa:
Hi, @khaidakin. Adding some of your device convertors.

src/devices/msh.ts
```diff
@@ -0,0 +1,124 @@
...
+        },
+    },
+    {
+        zigbeeModel: ["msh.bme280psm"],
+        model: "msh.bme280psm",
+        vendor: "MySmartHouse",
+        description: "[MSH outdoor thermometer with BME280](https://www.facebook.com/my.smart.house.in.ua)",
+        fromZigbee: [fz.battery, fz.temperature, fzLocal.humidity, fzLocal.pressure],
+        toZigbee: [tz.ptvo_switch_trigger],
+        exposes: [
```
Koenkk:
```suggestion
        extend: [m.humidity(), m.temperature(), m.pressure(), m.battery()],
        exposes: [
```

Can you try this and remove the fromZigbee and exposes?

src/devices/msh.ts
```diff
@@ -0,0 +1,124 @@
...
+            }
+        },
+    },
+    {
+        zigbeeModel: ["msh.ds18b20psm"],
+        model: "msh.ds18b20psm",
+        vendor: "MySmartHouse",
+        description: "[MSH outdoor thermometer with DS18B20](https://www.facebook.com/my.smart.house.in.ua)",
+        fromZigbee: [fz.battery, fz.temperature],
+        toZigbee: [tz.ptvo_switch_trigger],
```
Koenkk: Are these devices using a ptvo firmware?

SkeLLLa: I'm not sure 100%, since I'm just a user. @khaidakin did the hardware and flashed the firmware. I'm just trying to modernize relatively old-style converters and add support for them natively.
The signs that it might be using ptvo is that in the original external convertor files there were variables like `ptvo_on_off`.
From what I see on my devices there are signs that they running ptvo firmware. Here's an example of recognized device in z2m:
```
availability:
  - topic: zigbee2mqtt/bridge/state
    value_template: '{{ value_json.state }}'
  - topic: zigbee2mqtt/CO2 Sensor/availability
    value_template: '{{ value_json.state }}'
availability_mode: all
command_topic: zigbee2mqtt/CO2 Sensor/l2/set/calibrate_l2
default_entity_id: select.co2_sensor_calibrate_l2
device:
  hw_version: 1
  identifiers:
    - zigbee2mqtt_xxxx
  manufacturer: MySmartHouse
  model: '[MSH Air quality sensor](https://www.facebook.com/my.smart.house.in.ua)'
  model_id: msh.AirQMon
  sw_version: '2024-06-18'
  via_device: zigbee2mqtt_bridge_xxxxxx
  name: CO2 Sensor
icon: mdi:tune
object_id: co2_sensor_calibrate_l2
options:
  - ZERO
  - SPAN
origin:
  support_url: https://www.zigbee2mqtt.io
  sw_version: 2.7.2
  name: Zigbee2MQTT
state_topic: zigbee2mqtt/CO2 Sensor
unique_id: xxxx_calibrate_l2_zigbee2mqtt
value_template: '{{ value_json.calibrate_l2 }}'
name: Calibrate l2
```

So all I can see here is `sw_version: '2024-06-18'`. It might look like ptvo, but only the devices author could tell for sure. If author's code of other zigbee converters will give you a hint, you can check https://github.com/Koenkk/zigbee-herdsman-converters/issues/10536, for instance. Or maybe if you know the way I could check it, I can try.

src/devices/msh.ts
SkeLLLa: Done. Checked with temp sensors and extend works well. And code looks much cleaner. Thanks!

Should we go with same approach with co2 sensor? It will not eliminate exposes section for it, since it has custom calibration logic.

Koenkk:
Thanks!

PR #11785, opened 2026-03-23
feat: add Tuya ZD24_Presence_Sensor support

zhongfu:
New sensor.

Physical appearance is identical to the [TS0601-PIR-Sensor](https://www.zigbee2mqtt.io/devices/TS0601-PIR-Sensor.html) and the [ZG-204ZM](https://www.zigbee2mqtt.io/devices/ZG-204ZM.html), but datapoints are quite different.

Name is from the converter provided by the seller; I made some changes to entity names and values to better match the other two sensors.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4952


Koenkk:
Thanks!

PR #11786, opened 2026-03-23
fix: custom read responses cleanup

Nerivec:
Re https://github.com/Koenkk/zigbee-herdsman/pull/1716
Mostly cleanup.
Also added warning about read response being spec-incorrect in all cases since replying to specific attrs instead of requested ones (should investigate further in case of issues with these devices).


src/devices/sonoff.ts
```diff
@@ -1517,20 +1517,22 @@ const sonoffExtend = {
...
+                        const {time, timeZone, localTime, dstShift} = getTimeClusterAttributes();
+
+                        // XXX: we're replying to specific attributes, which could be incorrect (not based on the request attrIds)
+                        endpoint
+                            .readResponse("genTime", frame.header.transactionSequenceNumber, {
+                                time,
+                                timeZone,
+                                localTime,
+                                dstStart: 0,
+                                dstEnd: 0,
```
Nerivec: @CubeZ2mDeveloper are the zero or the static payload (not based on request) mandatory here?
Else the entire `customReadResponse` could be removed and ZH would handle all this properly already.
cf https://github.com/Koenkk/zigbee-herdsman/blob/master/src/utils/timeService.ts#L99
cf https://github.com/Koenkk/zigbee-herdsman/pull/1716/changes#diff-6d00af5fc389c40b8c4dfc390d80e3eca3946d331dff84afdc1632cb5568bca2R386-R402

src/lib/bosch.ts
```diff
@@ -112,27 +111,20 @@ export const boschGeneralExtend = {
     handleZclVersionReadRequest: (): ModernExtend => {
         const onEvent: OnEvent.Handler[] = [
```
Nerivec, line 112: @burmistrzak I was reading the comment on this one. I wonder if Bosch devices would behave more appropriately if we sent the current `zclVersion` instead of 1? I don't have any to test though.
Note: the ZH PR https://github.com/Koenkk/zigbee-herdsman/pull/1716 will automatically send the default `zclVersion` (8), which means if above holds, the entire modern extend could be removed.

src/devices/sonoff.ts
CubeZ2mDeveloper: Hi Nerivec, thanks for taking a look at the Sonoff changes.

This device currently needs the custom genTime response. Its Zigbee behavior is a bit non-standard, so the default ZH handling is not sufficient here. 

For now, keeping the `customReadResponse` is the safest option to make the device work reliably. If we later verify that the default ZH implementation works as well, we can definitely simplify it.

Thanks again!

Koenkk:
Thanks!

PR #11787, opened 2026-03-23
fix: Move tradfri custom commands (genScenes) from ZH to ZHC

svhelge:
Added cluster and commands with `m.deviceAddCustomCluster`. Added the cluster definition to four devices. 

see https://github.com/Koenkk/zigbee-herdsman/pull/1714


Koenkk:
Thanks!

PR #11794, opened 2026-03-25
feat(sonoff): add SNZB-02DR2 fahrenheit temperature expose

CubeZ2mDeveloper:
## Summary

- Improve support for SONOFF SNZB-02DR2 temperature and humidity sensor.
- Add a derived `temperature_f` expose for Fahrenheit reading while keeping the existing `temperature_units` behavior unchanged.

## Changes

- Update `src/devices/sonoff.ts` to add:
  - derived `temperature_f` expose on `msTemperatureMeasurement`
  - `GET` support for `temperature_f`
  - Fahrenheit rounding to one decimal place

## Testing

Tested with Zigbee2MQTT on SONOFF ZBDongle-P.
- Verified pairing/interview.
- Verified read/write of temperature attributes.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -31,6 +31,16 @@ const defaultResponseOptions = {disableDefaultResponse: false};
 const e = exposes.presets;
 const ea = exposes.access;
 
+function deriveFahrenheitFromTemperatureOptions(celsius: number, options: KeyValue) {
+    try {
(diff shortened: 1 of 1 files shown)


Koenkk:
I think it's better to not expose derived values (just like we do for all other temp sensors), in case people want Fahrenheit they can create e.g. a sensor in HA.

PR #11802, opened 2026-03-26
Improve Tuya water sprinklers

andrei-lazarov:
I was working on the Nous water sprinkler, so I unified and fixed the converters. There are 3 types of sprinklers:
- without measurements
- with liter measurements
- with gallons measurements

The diff is a little messy to review, since I merged the liters and gallons into a single definition

### DIFF
--- a/src/devices/neo.ts
+++ b/src/devices/neo.ts
@@ -195,7 +195,7 @@ export const definitions: DefinitionWithExtend[] = [
         ],
         meta: {
             tuyaDatapoints: [
-                [1, "state", tuya.valueConverter.onOffCountdown],
+                [1, "state", tuya.valueConverter.onOff],
                 [
                     3,
                     "status",
@@ -264,7 +264,7 @@ export const definitions: DefinitionWithExtend[] = [
         ],
         meta: {
             tuyaDatapoints: [
-                [1, "state", tuya.valueConverter.onOffCountdown],
+                [1, "state", tuya.valueConverter.onOff],
                 [
                     3,
                     "status",
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -791,20 +791,6 @@ export const valueConverter = {
         from: (value: number) => (value > 200 ? value - 256 : value),
         to: (value: number) => (value < 0 ? 256 + value : value),
     },
-    onOffCountdown: {
-        to: (v: string, meta: Tz.Meta) => {
-            if (v === "ON") return true;
-
-            meta.state.countdown_left = 0;
-            return false;
-        },
-        from: (v: boolean, meta: Fz.Meta) => {
-            if (v) return "ON";
-
-            meta.state.countdown_left = 0;
-            return "OFF";
-        },
-    },
     refresh: {
         to: (v: string) => {
             return v === "refresh";


src/devices/neo.ts
```diff
@@ -139,198 +139,221 @@ export const definitions: DefinitionWithExtend[] = [
...
+                      e.water_current(),
+                      e.water_total(),
+                  ]
+                : [
+                      e.quantitative_watering().withUnit("gal"),
+                      e.single_watering_amount().withUnit("gal"),
+                      e.surplus_flow().withUnit("gal"),
+                      e.water_current().withUnit("gal/min"),
+                      e.water_total().withUnit("gal"),
+                  ]),
```
andrei-lazarov, line 185: I merged liter and galon into 1 converter, because unit is the only difference.
Let me know if this is ok, or I should keep them separate.

`z7a2jmyy` are the only ones that show liters in the Tuya dev portal

src/devices/neo.ts
```diff
@@ -139,198 +139,221 @@ export const definitions: DefinitionWithExtend[] = [
...
+            e.enum("water_total_reset", ea.STATE_SET, ["reset"]).withDescription("Reset the stored watering amount to 0").withCategory("config"),
+            // e.text("normal_timer", ea.STATE_SET).withDescription("Schedule the sprinkler: time and weekdays"),
+            // e.enum("weather_delay", ea.STATE_SET, ["24h", "48h", "72h", "cancel"]).withDescription("Unknown"),
+            // e.binary("weather_switch", ea.STATE_SET, "ON", "OFF").withDescription("Unknown"),
+            // e.binary("smart_irrigation", ea.STATE_SET, "ON", "OFF").withDescription("Unknown"),
+            // e.binary("quantitative_water_linkage", ea.STATE_SET, "ON", "OFF").withDescription("Unknown"),
```
andrei-lazarov, line 191: Normal timer allows scheduling the device, exactly like an alarm app. It's encoded into a string, but it's difficult to reverse-engineer.
The others show up in dev portal, but are not used in the Tuya app, and I don't get any response from the device.

src/devices/neo.ts
```diff
@@ -139,198 +139,221 @@ export const definitions: DefinitionWithExtend[] = [
...
+                            meta.state.countdown_left = 0;
+                            return "OFF";
+                        },
+                        to: (v: string, meta) => {
+                            if (v === "ON") return true;
+
+                            meta.state.countdown_left = 0;
+                            return false;
+                        },
+                    },
```
andrei-lazarov: The countdown resets when device is turned off, but it's not always reported. So I manually set it to 0 when device turns off

src/devices/neo.ts
```diff
@@ -139,198 +139,221 @@ export const definitions: DefinitionWithExtend[] = [
...
-            e.binary("flow_switch", ea.STATE_SET, "ON", "OFF").withDescription("Flow switch"),
-            e.binary("child_lock", ea.STATE_SET, "ON", "OFF").withDescription("Child lock"),
-            e.numeric("surplus_flow", ea.STATE).withDescription("Surplus flow"),
-            e.numeric("single_watering_duration", ea.STATE).withDescription("Single watering duration"),
-            e.numeric("single_watering_amount", ea.STATE).withDescription("Single watering amount"),
+        description: "Smart sprinkler timer with measurements",
+        whiteLabel: [
+            tuya.whitelabel("Nous", "L14", "Smart water valve", ["_TZE204_nnhwcvbk", "_TZE284_nnhwcvbk"]),
+            tuya.whitelabel("NEO", "NAS-WV05B2", "Smart sprinkler timer", ["_TZE284_rzrrjkz2", "_TZE284_uab532m0"]),
+            tuya.whitelabel("NEO", "NAS-WV05B2-L", "Smart sprinkler timer", ["_TZE284_z7a2jmyy"]),
```
andrei-lazarov, line 160: I read through old posts to match IDs to white labels, but it's not reliable.
The same ID is used by multiple brands, different designs.
I left them unchanged.

src/devices/neo.ts
```diff
@@ -139,198 +139,221 @@ export const definitions: DefinitionWithExtend[] = [
...
+                    "refresh",
+                    {
+                        from: () => {
+                            return "idle";
+                        },
+                        to: (v: string) => {
+                            return v === "refresh";
+                        },
+                    },
+                ], // (ui_refresh)
```
andrei-lazarov: This is a button that makes the device report its status. Used an enum with a single exposed value

src/devices/neo.ts
```diff
@@ -139,198 +139,221 @@ export const definitions: DefinitionWithExtend[] = [
...
+                [108, "single_watering_amount", tuya.valueConverter.raw],
+                [109, "on_with_countdown", tuya.valueConverter.raw], // (valve_opening_time)
+                [110, "quantitative_water_linkage", tuya.valueConverter.onOff],
             ],
         },
     },
     {
-        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_rzrrjkz2", "_TZE284_uab532m0", "_TZE284_nnhwcvbk"]),
-        model: "NAS-WV05B2",
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE204_a9ojznj8", "_TZE284_a9ojznj8"]),
```
andrei-lazarov, line 243: `a9ojznj8` are without measurements

src/devices/neo.ts
Koenkk: Maybe make a generic: `tuya.valueConverter.onOffCountdown`?

src/devices/neo.ts
Koenkk: That's fine, so if I understand correctly no breaking change here right? (due to unit changes)

src/devices/neo.ts
andrei-lazarov: Turns out it's not needed, I removed it 😅 Sometimes timer still goes while state is off

src/devices/neo.ts
andrei-lazarov: No breaking change. 
Unit is cosmetic, like the description. I corrected them, to match the Tuya app.
There's no conversion or anything. Values stay the same.

Koenkk:
thanks!

PR #11815, opened 2026-03-27
Move custom attribute operationalStatus from ZH to ZHC (lib/ubisys.ts)

svhelge:
Add attribute to existing cluster definition


Koenkk:
Thanks!

PR #11824, opened 2026-03-28
Move custom commands from ZH to ZHC 

svhelge:
Move converters using these attributes and commands to lib/tuya.ts (from fz and tz). 
Add custom commands and attributes to clusters and assign clusterdefinitions to tuyaBase. 
Ensure that tuya devices using these attributes and commands have tuyaBase in the definition. 
Other devices using these commands and attributes are updated by assigning the custom clusters directly (ie not using tuyaBase, but using tuya.clusters.addTuya.... ).
Move attribute definition tuyaCountDown from device/tuya.ts to the clusterdefinition in lib/tuya.


Koenkk:
Thanks!

Issue #10067, opened 2025-09-23
[Wrong device]: Window/Door sensor Zbeacon TS0203 identified as Temperature sensor


shajek:
Link 
https://www.aliexpress.com/item/1005009427896963.html

database entry:
`
{"id":23,"type":"EndDevice","ieeeAddr":"0xa4c138e13744dd19","nwkAddr":57210,"manufId":4742,"manufName":"Zbeacon","powerSource":"Battery","modelId":"TS0203","epList":[1],"endpoints":{"1":{"profId":260,"epId":1,"devId":1026,"inClusterList":[0,3,1,1280,32],"outClusterList":[25],"clusters":{"genBasic":{"attributes":{"modelId":"TS0203","manufacturerName":"Zbeacon","powerSource":3,"zclVersion":3,"appVersion":113,"stackVersion":2,"hwVersion":0,"swBuildId":"0122052017"}},"ssIasZone":{"attributes":{"61441":2,"iasCieAddr":"0x00124b002f8e6825","zoneState":1,"currentZoneSensitivityLevel":1}},"genPollCtrl":{"attributes":{"checkinInterval":6960}},"genPowerCfg":{"attributes":{"batteryPercentageRemaining":200,"batteryVoltage":30}}},"binds":[{"cluster":32,"type":"endpoint","deviceIeeeAddress":"0x00124b002f8e6825","endpointID":1}],"configuredReportings":[],"meta":{}}},"appVersion":113,"stackVersion":2,"hwVersion":0,"swBuildId":"0122052017","zclVersion":3,"interviewCompleted":true,"interviewState":"SUCCESSFUL","meta":{},"lastSeen":1758607515943,"checkinInterval":1740}
`

DP list from tuya:

```

{
  "result": {
    "active_time": 1758539565,
    "bind_space_id": "55283491",
    "category": "qt",
    "create_time": 1758539565,
    "custom_name": "",
    "icon": "smart/icon/1524817291ohyty7jfc6kyokuqeyvnp14i_0.png",
    "id": "XYZ",
    "ip": "",
    "is_online": true,
    "lat": "",
    "local_key": "",
    "lon": "",
    "model": "",
    "name": "Door Sensor",
    "product_id": "7jIGJAymiH8OsFFb",
    "product_name": "门窗磁传感器",
    "sub": true,
    "time_zone": "+02:00",
    "update_time": 1758539565,
    "uuid": "XYZ"
  },
  "success": true,
[...]

PR #11825, opened 2026-03-28
add fingerprints for Zbeacon TS0203 so the door sensor is not detecte…

DaBena:
Fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/10067

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1726,8 +1726,16 @@ export const definitions: DefinitionWithExtend[] = [
     },
     {
         fingerprint: [
-            {manufacturerName: "Zbeacon", modelID: "TS0203", endpoints: [{ID: 1, profileID: 260, deviceID: 770, inputClusters: [0, 3, 1, 32, 1026, 1029], outputClusters: [25]}]},
-            {manufacturerName: "Zbeacon", modelID: "TS0202", endpoints: [{ID: 1, profileID: 260, deviceID: 770, inputClusters: [0, 3, 1, 32, 1026, 1029], outputClusters: [25]}]},
+            {
+                manufacturerName: "Zbeacon",
+                modelID: "TS0203",
+                endpoints: [{ID: 1, profileID: 260, deviceID: 770, inputClusters: [0, 3, 1, 32, 1026, 1029], outputClusters: [25]}],
+            },
+            {
+                manufacturerName: "Zbeacon",
+                modelID: "TS0202",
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -1724,6 +1724,69 @@ export const definitions: DefinitionWithExtend[] = [
         exposes: [e.battery(), e.action(["click"])],
         toZigbee: [],
     },
+    // Zbeacon TS0203 IAS contact (devId 1026); ZC-W1 temp/humidity uses devId 770.
+    {
+        fingerprint: [
```
Koenkk, line 1728: Can't this be merged with the `TS0203` below?

src/devices/tuya.ts
```diff
@@ -1724,6 +1724,69 @@ export const definitions: DefinitionWithExtend[] = [
...
+
+            const endpoint = device.getEndpoint(1);
+            if (endpoint.binds.some((b) => b.cluster.name === "genPollCtrl")) {
+                await endpoint.unbind("genPollCtrl", coordinatorEndpoint);
+            }
+        },
+    },
     {
         fingerprint: [
             {
```
Koenkk: I would expect the new fingerprint to be added here

src/devices/tuya.ts
```diff
@@ -1743,6 +1743,27 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [m.temperature(), m.humidity(), m.battery()],
     },
     {
+        // Zbeacon TS0203 IAS contact (devId 1026); ZC-W1 temp/humidity uses devId 770.
+        fingerprint: [
+            {
+                priority: 200,
+                manufacturerName: "Zbeacon",
+                modelID: "TS0203",
```
Koenkk, line 1751: Since the modelId is `TS0203`, I expect them to be detected as a Tuya TS0203 already without any change, is that not the case?

DaBena: no, the same modelId is used for two different devices under Zbeacon https://github.com/Koenkk/zigbee2mqtt/issues/28167

Koenkk: Can you provide a screenshot of the device about page how it is detected in z2m 2.9.2?

DaBena: Honestly, with a fresh pairing on v2.9.2 I cannot reproduce the door sensor showing as a temperature/humidity sensor. It now maps correctly. I do not fully understand why I ran into this originally. Maybe when several definitions share the same modelID, it depends on which candidate is first in the candidates list. I am also fine with closing the PR.
zigbeeScreenshot

PR #11829, opened 2026-03-29
fix: lumi.vibration.agl01 manuSpecificLumi cluster

absent42:
lumi.vibration.agl01 doesn't advertise manuSpecificLumi cluster when pairing, which causes the cluster to be missing in the Z2M device.

The result of which is lumiVibration, lumiReportInterval, lumiSensitivityAdjustment, lumiBattery fail to work.

Commit adds deviceAddCustomCluster to device extend.

### DIFF
--- a/src/lib/lumi.ts
+++ b/src/lib/lumi.ts
@@ -2827,7 +2827,7 @@ export const lumiModernExtend = {
                 cluster: "manuSpecificLumi",
                 type: ["attributeReport", "readResponse"],
                 convert: (model, msg, publish, options, meta) => {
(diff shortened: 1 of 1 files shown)


src/lib/lumi.ts
svhelge, line 1756: I suggest to add attribute `movement` here instead.
And then use  `lumi.modernExtend.addManuSpecificLumiCluster()` in the device specification.

absent42: Done

svhelge, line 67: Add also `movement: number;`  here

Koenkk:
Thanks!

PR #11830, opened 2026-03-29
fix(hive): write thermostat attributes atomically for SLR/OTR receivers

GraemeF:
## Summary

Hive SLR1/SLR2/OTR1 receivers require `system_mode`, `temperature_setpoint_hold`, and `occupied_heating_setpoint` to be written in a single ZCL transaction. The generic thermostat converters write each attribute independently, which causes the device to silently ignore commands when sent via Home Assistant's climate entity (which issues separate `set_hvac_mode` and `set_temperature` service calls).

This replaces the three generic converters (`tz.thermostat_system_mode`, `tz.thermostat_occupied_heating_setpoint`, `tz.thermostat_temperature_setpoint_hold`) with a single combined converter that always includes the required companion attributes in one `entity.write()` call.

**Behaviour:**
- `set_hvac_mode("heat")` → writes `{systemMode, tempSetpointHold: 1, occupiedHeatingSetpoint}` atomically
- `set_hvac_mode("off")` → writes `{systemMode, tempSetpointHold: 0}` atomically
- `set_temperature(20)` → writes `{occupiedHeatingSetpoint, systemMode: heat, tempSetpointHold: 1}` atomically

This follows the same pattern as the Namron panel heater frost mode converter, which also writes multiple hvacThermostat attributes in a single `entity.write()`.

**Affected models:** SLR1, SLR1b, SLR1c, SLR1d, SLR2, SLR2b, SLR2c/d, OTR1

**Backward compatible:** Yes. Existing users who already send combined MQTT payloads will see no change. Users sending individual attributes now get the companion attributes included automatically.

## Motivation

The [SLR2 device page](https://www.zigbee2mqtt.io/devices/SLR2.html) documents that "the payload needs to be sent as one single command" — every operation example (heat on, heat off, boost, water on/off) is a combined payload. This requirement is also documented on the [SLR1b](https://www.zigbee2mqtt.io/devices/SLR1b.html) page.

Without this fix, controlling these receivers from Home Assistant requires either:
- Manually constructing combined MQTT payloads via `mqtt.publish` service calls
- Using the retired [HA-Hive-Local-Thermostat](https://github.com/andrew-codechimp/HA-Hive-Local-Thermostat) HACS integration (unmaintained)

With this fix, standard Home Assistant climate service calls produce valid combined payloads.

## Scope and limitations

This PR addresses the atomic write requirement — it ensures the converter always sends the required companion attributes together in a single ZCL write. This is a necessary precondition for HA climate calls to work with these devices.

However, [community reports](https://community.home-assistant.io/t/hive-heating-via-zigbee2mqtt-many-features-do-not-work/428554) indicate that some users also need to **reconfigure the device** in Z2M (the "Reconfigure" button) after initial pairing before commands are accepted at all, even with correct combined payloads. This suggests the device may silently ignore commands if the ZCL reporting/binding setup from the `configure()` step was incomplete during pairing.

Other known issues this PR does **not** address:
- Boost/emergency_heating duration resetting to 65535 on some firmware versions
- Weekly schedule writes being rejected by the device
- Dual-channel SLR2 variants exhibiting different behaviour from single-channel SLR1 models

This fix removes the most common barrier to HA integration (the split-write problem), but users migrating from the Hive cloud should expect some device-specific setup and testing.

## Related issues

- https://community.home-assistant.io/t/hive-heating-via-zigbee2mqtt-many-features-do-not-work/428554
- https://community.home-assistant.io/t/how-to-set-up-your-hive-heating-system-with-home-assistant/808765
- https://github.com/Koenkk/zigbee2mqtt/issues/7461

### DIFF
--- a/test/hive.test.ts
+++ b/test/hive.test.ts
@@ -3,6 +3,11 @@ import * as zhc from "../src";
 import type {Tz} from "../src/lib/types";
 import {mockDevice} from "./utils";
 
+// Tests for the combined hiveSLRThermostat converter, which writes
+// system_mode, temperature_setpoint_hold, and occupied_heating_setpoint
+// atomically per the SLR2 device documentation:
+// https://www.zigbee2mqtt.io/devices/SLR2.html
+
 describe("Hive SLR thermostat converter", () => {
     let device: ReturnType<typeof mockDevice>;
     let definition: Awaited<ReturnType<typeof zhc.findByDevice>>;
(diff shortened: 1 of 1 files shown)


GraemeF:
Closing for now — after reading community reports more carefully, I'm not confident that atomic writes alone are sufficient. Need to do more homework on what the Hive Local Thermostat HACS integration does beyond combining payloads, and test with real hardware before submitting. Branch preserved for when we're ready to revisit.

PR #11837, opened 2026-03-29
Single source of truth for thermostat enum values

jpaju:
### Summary
Thermostat enum values (system modes, running states, fan modes, etc.) were defined in multiple places: as constant objects in `constants.ts`, as inline string unions in `modernExtend.ts`, and as hardcoded allowed-value arrays with runtime `assert()` checks in `exposes.ts`. This PR makes the constant objects the single source of truth and derives TypeScript types from them, so invalid values are caught at compile time instead of runtime.

### Context
Discovered while working on Nobo thermostat support. Adding new thermostat devices means passing enum values to methods like `withSystemMode()` and `withRunningState()`, where typos or invalid values only surface as runtime assertion failures. Moving this validation to the type system makes it impossible to pass invalid values and removes the need for the runtime checks entirely.

### Changes
- Constant objects in `constants.ts` now use `as const satisfies` to preserve literal value types, with union types
- Methods in `exposes.ts` accept the derived union types instead of `string[]`
- Inline string union types in the `ThermostatArgs` interface reference the centralized types instead of redeclaring them
- Direct bracket access on lookup objects replaced with `utils.getFromLookup()`, which throws an error on unknown keys instead of silently returning undefined


Koenkk:
Thanks!

PR #11840, opened 2026-03-30
Disable enhanced hue by default & convert HS to XY, on XY-only bulbs

andrei-lazarov:
- supersedes and closes https://github.com/Koenkk/zigbee-herdsman-converters/pull/11364

1. Make enhancedHue disabled by default. Make it enabled by default for Philips lights
2. Convert HS command to XY, if bulb has only XY (above PR from @xLerate)
3. Leave supportsHS disabled by default (as it was). Fix errors in code, where it assumed it was enabled by default

**Enhanced hue must be disabled by default.** 
Because when enabled, it fully replaces HS color with enhanced HS color. And if the bulb doesn't support it, it silently fails and it can't fallback to non-enhanced.
Tuya bulbs already had it disabled. I left it enabled for Hue, they support it.

_When we rework color, enhanced should be a separate MQTT command / payload. You can't  benefit from the better 0-65535 resolution, if the format expects 0-360 hue._

I intially wanted to enable HS color by default. But it may cause more harm than good, so let's instead verify it when adding new devices.

cc @Nerivec


andrei-lazarov:
@Koenkk sorry for the spam! Merged the other PR into this and it's ready now!

This PR doesn't break / change anything for most users. It just fixes some rare issues. Small steps 🙂

I tested it on (TRADFRI, KAJPLATS, Hue, Tuya) x (MQTT, Windfront, HA) to make sure it doesn't break anything.
And ZHC + Z2M tests pass as well. It is safe to merge.

Koenkk:
A big thanks as usual! 😄

RubenKelevra:
@andrei-lazarov is this setting exposed in the UI somewhere? I would like to test if my lamps support this and then submit this as explicit disabled or enabled in the repo for them.

andrei-lazarov:
@RubenKelevra yes, (if you're on recent Z2M and you are using the Windfront user-interface, now default)

Go to light -> dev console.
Execute command -> ep 1 (usually) -> cluster `lightingColorCtrl`
- `moveToColor`
  - takes XY coordinates 0-65535
  - relative to color spectrum
  - bulb "rounds" to closest supported color and converts (with its own algorithm) to HS (RGB)
  - is mandatory in Zigbee spec, should be supported by all
- `moveToHueAndSaturation`
  - takes hue 0-254, sat 0-254
  - relative to bulb capabilities
  - is optional
- `enhancedMoveToHueAndSaturation` 
  - takes hue 0-65535, sat 0-254 (finer control)
  - relative to bulb capabilities
  - is optional

Also test read/write attributes:
- `lightingColorCtrl`: 
  - colorCapabilities: tells you which commands out of the above are supported - but it doesn't always match
  - options: allows changing temp/color while bulb is off (should be displayed as "execute if off" in Z2M exposes)
  - startUpColorTemperature
- `genLevelCtrl`:
  - options: allows changing brightness while bulb is off (should be displayed as "execute if off" in Z2M exposes)
  - onLevel
  - startUpCurrentLevel
  - off / on / onOffTransitionTime

More stuff
- https://github.com/Koenkk/zigbee-herdsman-converters/issues/11893

PR #11849, opened 2026-03-31
Improve IKEA KAJPLATS

andrei-lazarov:
I moved all KAJPLATS models as whiteLabels, to manage their extend and docs easier.

And I tested and exposed all supported features

- docs: https://github.com/Koenkk/zigbee2mqtt.io/pull/4971


Koenkk:
Thanks!

PR #11850, opened 2026-03-31
Nous E12 CO alarm: fix warm-up

andrei-lazarov:
Avoid some duplication and fix warm-up for Nous E12.

It does not report when warm-up ends. So I added a 2min timer


Koenkk:
thanks!

PR #11851, opened 2026-03-31
fix(detect): Detect `_TZ3210_imaccztn`as TYWB 4ch-RF

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/11387


Koenkk:
thanks!

PR #11854, opened 2026-04-01
Change _onOffTransitionTime_ unit to seconds

andrei-lazarov:
Zigbee attributes _onOffTransitionTime_, _onTransitionTimeValue_, _offTransitionTime_  need deciseconds.
That is very confusing, so let's use seconds in Zigbee2MQTT, like we usually do.

I only tested _onOffTransitionTime._ Works as expected now.

Why are these checks duplicated inside the same if ?

```ts
if (msg.data.onOffTransitionTime !== undefined && msg.data.onOffTransitionTime !== undefined)
```


Koenkk:
Thanks!

PR #11855, opened 2026-04-01
New M1-PE equipment has been added.

Chilli-liu:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4975


src/devices/heiman.ts
```diff
@@ -736,6 +776,42 @@ const heimanExtend = {
             isModernExtend: true,
         };
     },
+    heimanClusterIndicatorLightOnOff: (): ModernExtend => {
```
Koenkk, line 847: Please use `m.binary()` for this instead, examples: https://github.com/Koenkk/zigbee-herdsman-converters/blob/edbf227b62da4b5c2f23fd5d9eb8d38009df7e28/src/devices/efekta.ts#L77

src/devices/heiman.ts
```diff
@@ -944,6 +1020,486 @@ const heimanExtend = {
             isModernExtend: true,
         };
     },
+    heimanClusterSensorSensitivityLevel: (description = "sensor sensitivity level"): ModernExtend => {
```
Koenkk, line 1086: Use `m.enumLookup` here, examples: https://github.com/search?q=repo%3AKoenkk%2Fzigbee-herdsman-converters%20m.enumLookup&type=code

src/devices/heiman.ts
```diff
@@ -944,6 +1020,486 @@ const heimanExtend = {
...
+            },
+        ];
+        return {
+            exposes: exposes,
+            fromZigbee,
+            toZigbee,
+            isModernExtend: true,
+        };
+    },
+    heimanClusterOccupanySensorWorkMode: (): ModernExtend => {
```
Koenkk, line 1115: Also use `m.enumLookup` here

src/devices/heiman.ts
```diff
@@ -944,6 +1020,486 @@ const heimanExtend = {
...
+            },
+        ];
+        return {
+            exposes: exposes,
+            fromZigbee,
+            toZigbee,
+            isModernExtend: true,
+        };
+    },
+    heimanClusterOccupiedToUnoccupiedDelay: (): ModernExtend => {
```
Koenkk, line 1144: Can use `m.numeric` here

src/devices/heiman.ts
```diff
@@ -944,6 +1020,486 @@ const heimanExtend = {
...
+            },
+        ];
+        return {
+            exposes: exposes,
+            fromZigbee,
+            toZigbee,
+            isModernExtend: true,
+        };
+    },
+    heimanClusterOccupiedRepeatedReportingDuration: (): ModernExtend => {
```
Koenkk, line 1161: Can use `m.numeric` here, also replace the additions below with `m.` variants.

src/devices/heiman.ts
Chilli-liu: Yes, I made changes to this part, and then I discovered that when reading the attribute, the UI couldn't be updated synchronously. It is because m.binary uses Fz.Converter<any, undefined, ["attributeReport", "readResponse"]] by default. In my case, I have customized the Cluster. Therefore, additional processing for "fromZigbee" has been added.

src/devices/heiman.ts
Chilli-liu: OK

src/devices/heiman.ts
Chilli-liu: OK

src/devices/heiman.ts
Chilli-liu: OK

src/devices/heiman.ts
Chilli-liu: OK

src/devices/heiman.ts
Koenkk: I think if you change ` attribute: {ID: attributeId, type: Zcl.DataType.UINT8},` to ` attribute: 'indicatorLightLevelControlOf1',` it should work, if not provide the debug log when the device reports this attribute.

Chilli-liu: Thank you. I'll give it a try.

src/devices/heiman.ts
Koenkk: For this, and all other cases, you can change it to e.g. :


```
heimanClusterOccupiedToUnoccupiedDelay: (): ModernExtend => m.numeric({
            name: "unoccupied_delay",
            unit: "seconds",
            valueMin: 5,
            valueMax: 3600,
            cluster: clusterName,
            attribute: "pirOToUDelay",
            description: "occupied to unoccupied delay",
            access: "ALL",
        })
```

src/devices/heiman.ts
```diff
@@ -1008,6 +1077,383 @@ const heimanExtend = {
...
+        const valueMap = {1: "pir", 2: "radar", 3: "pir_and_radar"};
+        const mExtend = m.enumLookup({
+            name: "work_mode",
+            cluster: clusterName,
+            attribute: {ID: attributeId, type: Zcl.DataType.BITMAP8},
+            description: "occupany sensor work mode",
+            lookup: {pir: 1, radar: 2, pir_and_radar: 3},
+            access: "ALL",
+        });
+        const fromZigbee = [
```
Koenkk, line 1127: This fromZigbee converter is not needed I think

Chilli-liu: No, no, no. This is necessary. For instance, synchronizing the current mode after setting or resetting.

src/devices/heiman.ts
Chilli-liu: https://github.com/Koenkk/zigbee-herdsman-converters/pull/11855#discussion_r3031173916

As you said, initially it was modified in this way, but after the modification, it was found that the read attribute could not be synchronized to the UI. So what you see is the modification that I made and which can be displayed normally after debugging. You can give it a try, okay?

Chilli-liu: image

src/devices/heiman.ts
Koenkk: > For instance, synchronizing the current mode after setting or resetting.

This should also be the case with the `m.enumLookup`, can you provide the debug log for both with and without fz converter?

PR #11862, opened 2026-04-02
Correct battery voltage data point to Tuya device _TZE200_jt50ea5d

Liionboy:

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2024,7 +2024,7 @@ export const definitions: DefinitionWithExtend[] = [
                 .withDescription("Outlet water temperature")
                 .withValueMin(0)
                 .withValueStep(0.01),
-            e.numeric("battery_voltage", ea.STATE).withUnit("V").withDescription("Power supply voltage").withValueMin(0).withValueStep(0.01),
+            e.battery_voltage(),
         ],
         meta: {
             tuyaDatapoints: [
@@ -2113,7 +2113,7 @@ export const definitions: DefinitionWithExtend[] = [
                 // DP 22 - Outlet Water Temperature
                 [22, "outlet_water_temperature", tuya.valueConverter.divideBy100],
                 // DP 24 - Power Supply Voltage
-                [24, "battery_voltage", tuya.valueConverter.divideBy100],
+                [24, "battery_voltage", tuya.valueConverter.multiplyBy10],
             ],
         },
         options: [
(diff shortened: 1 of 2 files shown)


src/devices/tuya.ts
```diff
@@ -2024,7 +2024,7 @@ export const definitions: DefinitionWithExtend[] = [
                 .withDescription("Outlet water temperature")
                 .withValueMin(0)
                 .withValueStep(0.01),
-            e.battery_voltage(),
+            e.numeric("battery_voltage", ea.STATE).withUnit("V").withDescription("Power supply voltage").withValueMin(0).withValueStep(0.01),
```
Koenkk: This should be kept instead, I believe the issue is that `e.battery_voltage()` expects mV instead of V, so `[26, "battery_voltage", tuya.valueConverter.divideBy100],` should be changed instead

Liionboy: At my convertor is work like this

[image](https://github.com/user-attachments/assets/e4355386-c5d2-4f78-8125-66dde89209ce)

Liionboy: And please be more explicity, i need to change something? Because i don't understand anything from your comment, device send value in V not in mV, so i need to change something?

Liionboy: > This should be kept instead, I believe the issue is that `e.battery_voltage()` expects mV instead of V, so `[26, "battery_voltage", tuya.valueConverter.divideBy100],` should be changed instead

So i put  

`[24, "battery_voltage", tuya.valueConverter.raw],`
`[24, "battery_voltage", tuya.valueConverter.divideBy10],`
`[24, "battery_voltage", tuya.valueConverter.multiplyBy10],`

And only Null i got, so if you want to let how i put is ok if not i will use convertor because is very hard to explain something and don't understand.

Liionboy: > This should be kept instead, I believe the issue is that `e.battery_voltage()` expects mV instead of V, so `[26, "battery_voltage", tuya.valueConverter.divideBy100],` should be changed instead

This is what device return: 

`qtt: MQTT publish: topic 'zigbee2mqtt/Wattterrrr', payload '{"battery_voltage":**367**,"cumulative_heat":0`

So i can't use that `e.battery_voltage()` because he expect to be 3670 so is not work and i don'r see a helper with `[24, "battery_voltage", tuya.valueConverter.multiplyBy10],`

Koenkk: You can add a new `multiplyBy10` here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/de37bc4f421f3fd0b8da6a3efaf10140c9baeefd/src/lib/tuya.ts#L825

Koenkk:
Thanks!

Issue #11866, opened 2026-04-02
Lx values calculated incorrectly in Z2M


MikaFromTheRoof:
I recently wondered why all the devices in Z2M always show at least 1 lx illuminance and never 0, even if they are in pitch black surroundings.

Because of that I did some research and found out, that the Zigbee standard defines the following formular for calculating the lux value from the raw illuminance value, that the Zigbee devices report:

`lx = 10^(value/10000)-1`
This has also been mentioned in this old issue: https://github.com/Koenkk/zigbee2mqtt/issues/2837#issue-555527523

Back then the formular got corrected in the fromZigbee converter.

Now I found, that the formular used in the modernExtend.ts is wrong again: https://github.com/Koenkk/zigbee-herdsman-converters/blob/5f2f22377ca0cd966244bfa601444f39ed501bca/src/lib/modernExtend.ts#L811

The -1 is located in the wrong position, which would mathematically result in the lx value never being 0.

Is it possible to fix that globally in Z2M?

Thanks for great work!

Koenkk:
Fixed, thanks!

> Changes will be available in the [dev branch](https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) in a few hours from now and in the next release which is every 1st of the month.

andrei-lazarov:
Sorry, your "research" is just wrong. The formula was correct.

In Zigbee spec we get:
`value = 10000 * log10(lx) + 1`

Quick math:
```
value - 1 = 10000 * log10(lx)
(value - 1) / 10000 = log10(lx)
10 ^ ((value - 1) / 10000) = lx
```
Value 0 and 0xffff are special cases.
- see https://github.com/Koenkk/zigbee2mqtt/issues/31431 

It is expected that your sensor shows 1 lx. You can even see minMeasuredValue is at least 1

I'll revert this and see how to treat the special cases..

MikaFromTheRoof:
Okay, good to know but also weird that according to the Zigbee spec you mentioned it is just impossible for Zigbee devices to ever report 0 lx. 
I just wrote an external converter for one of my DIY Zigbee devices, that uses the formula I mentioned above and it works just fine. It is displayed as 0 lx and I don't have any problems or errors

jamsoft:
I have some el cheapo Zigbee Tuya Lux sensors that report 0 lx. I stumbled here as I'm looking at your Zigbee Illuminance demo at the moment as I'm building one around a VEML7700.

[...]

PR #11868, opened 2026-04-02
Correct illuminance formula and treat special cases

### LINKED ISSUE #31431: Illuminance Measurement Attributes (Value 0 and 0xffff)
### What happened?

ZigBee Cluster Library Specification.

Image

* And: MinMeasuredValue/MaxMeasuredValue: Value of 0xffff indicates that this attribute is not defined.

_No response_

_No response_

### Zigbee2MQTT version

all

### Adapter firmware version

all

### Adapter

all

### Setup

all

_No response_

### Debug log

_No response_

### Notes

_No response_

andrei-lazarov:
1. reverted https://github.com/Koenkk/zigbee-herdsman-converters/issues/11866

2. fixes: https://github.com/Koenkk/zigbee2mqtt/issues/31431

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -809,7 +809,7 @@ const luxScale: ScaleFunction = (value: number, type: "from" | "to") => {
     if (type === "from") {
         if (result === 0x0000) return 0;
 
-        if (result === 0xffff) return undefined;
+        if (Number.isNaN(result)) return undefined;
 
         result = 10 ** ((result - 1) / 10000);
     }
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -804,19 +804,19 @@ export function customTimeResponse(start: "1970_UTC" | "2000_LOCAL"): ModernExte
 
 // #region Measurement and Sensing
 
+const luxScale: ScaleFunction = (value: number, type: "from" | "to") => {
+    let result = value;
+    if (type === "from") {
+        if (result === 0x0000) return 0;
+
+        if (result === 0xffff) return undefined;
+
+        result = 10 ** ((result - 1) / 10000);
+    }
+    return result;
+};
+
 export function illuminance(args: Partial<NumericArgs<"msIlluminanceMeasurement">> = {}): ModernExtend {
-    const luxScale: ScaleFunction = (value: number, type: "from" | "to") => {
-        let result = value;
-        if (type === "from") {
-            if (result === 0) return 0;
-
-            if (result === 65535) return undefined;
-
-            result = 10 ** ((result - 1) / 10000);
-        }
-        return result;
-    };
-
     const result = numeric({
         name: "illuminance",
         cluster: "msIlluminanceMeasurement",
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -808,6 +808,10 @@ export function illuminance(args: Partial<NumericArgs<"msIlluminanceMeasurement"
     const luxScale: ScaleFunction = (value: number, type: "from" | "to") => {
         let result = value;
         if (type === "from") {
+            if (result === 0) return 0;
+
+            if (result === 65535) return undefined;
+
             result = 10 ** ((result - 1) / 10000);
         }
         return result;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks, I already reverted 1

andrei-lazarov:
@mikafromtheroof I still don't fully understand this, but I added 0 and 0xffff like in ZHA (as suggested in the linked issue)

src/lib/modernExtend.ts
```diff
@@ -808,6 +808,10 @@ export function illuminance(args: Partial<NumericArgs<"msIlluminanceMeasurement"
     const luxScale: ScaleFunction = (value: number, type: "from" | "to") => {
         let result = value;
         if (type === "from") {
+            if (result === 0) return 0;
+
+            if (result === 65535) return undefined;
```
Nerivec: Should use `0x0000` and `0xffff` (it makes it easier to identify when searching code).
Should also move the `luxScale` const outside the `illuminance` fn for good measure.

Though can't return `undefined` here; will bite the `assertNumber` in `numeric`.
_ZHC is tricky to use with anything undefined since it doesn't use TS strict (lets stuff pass through that it shouldn't, but it's a massive refactor to enforce strict)._

Nerivec:
This cluster is a little strange, usually, the attr has a range that is outside the special cases (which we handle in ZH), but for some reason, this one has full range + special cases. _And it was updated in latest version of the spec, so, for some reason, this does not appear to be a mistake (or it's just a backwards compat issue now...)..._
We could update the cluster in ZH to use max: 0xfffe, but with above, not sure what's the right call here...

The [ZH layer](https://github.com/Koenkk/zigbee-herdsman/blob/4a3d508e9de844079c664cbc85e90341d873ffc0/src/zspec/zcl/utils.ts#L232-L377) maps non-values to `Number.NaN` (and vice-versa), but for now very few converters actually detect this and handle it.
Also, in general, support for special cases is tricky with Z2M, because we don't "know" the format the other end (frontend, HA, node red, etc.) requires to signal `invalid`/`impossible`/etc..

MikaFromTheRoof:
> @MikaFromTheRoof I still don't fully understand this, but I added 0 and 0xffff like in ZHA (as suggested in the linked issue)

Would have also been my suggestion to look how ZHA does it. But I am not deep enough into the ZHC to be able to assess, if it is working for Z2M this way. 

Thanks, @Nerivec for looking into that!

I am more on the Zigbee device building side of that whole process and only did write some external converters for my DIY devices. The way I understand the approach of ZHA is, that you will only get 0 lx, if the raw illuminance reported by the device is exactly zero. As soon as it is above that, the illuminance will be interpreted as 1 lx, even if it mathematically should still be rounded down to 0 lx, right?
But that would be fine, because I could define a threshold for the raw illuminance in my devices firmware for all the values, that would resolve to 0 lx and then just always report 0, when raw illuminance falls in that range.

I am using simple photoresistors, that normally never really read the value 0, even if it is in pitch darkness. I could imagine, that it might be the same for most of the off-the-shelf devices. So even with that fix, I think most of the devices might still never reach the raw illuminance value of 0. But at least it would be possible now to report 0 lx without writing a custom external converter.

Nerivec:
ZHA has a completely different architecture, and they only have to work with HA.

The approach with that attribute isn't that of ZHA, it's that of the Zigbee specification.
illum
Thing is, in other cases, when there are special cases like this, the attribute `Range` is actually excluding them, but not here, which makes it a little weird.

The zero return if zero in this PR is correct.
As for the invalid... @Koenkk do you think we should move the max to `0xfffe` in ZH for the attribute, so it is mapped to NaN (as it should be really, just that weird table in the spec...)?
_We probably should take a closer look at how NaN is handled in ZHC in general (lots of attributes do that) and how that translates to the MQTT payload being sent. Make sure the behavior is aligned across the board._

Koenkk:
Reading the spec `0xfffe` should be indeed mapped to `NaN`, so I think `if (result === 0xffff) return undefined;` has to be changed to `if (result === Number.NaN) return undefined;`

Nerivec:
Can't return undefined in ZHC though, it will fail (see https://github.com/Koenkk/zigbee-herdsman-converters/pull/11868#discussion_r3034731317).
Also have to adjust the attribute in ZH to have `max: 0xfffe`.

andrei-lazarov:
So should we return something like `-1` to signal invalid? 🤔

Nerivec:
`-1` has proven to have weird reactions in the past. Some automation systems don't treat it properly, and it can cascade.

I think undefined (null json) is the way to go, but need to fix the whole chain to support it properly.
Problem is, `numeric` has a boatload of uses, so, the change is massive in effect (i.e. should be correct, but need to be sure).
Then the spot in the fn can just do:
```ts
if (Number.isNaN(result) {
    return undefined;
}
```

If you search for `isNaN` in ZHC, you will find places that already "somewhat" handle non-values, but I think in most cases, it just doesn't return.

andrei-lazarov:
Not stale. Still not sure what to do here 😅

PR #11889, opened 2026-04-06
Update philips.ts

Holgeretti:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/4998


src/devices/philips.ts
```diff
@@ -4980,4 +4980,12 @@ export const definitions: DefinitionWithExtend[] = [
...
         extend: [philips.m.light({colorTemp: {range: [153, 447]}, color: {modes: ["xy", "hs"], enhancedHue: true}})],
     },
+    {
+        zigbeeModel: ['929004611002'],
+        model: '929004611002',
+        vendor: 'Philips',
+        description: 'Hue Flux Outdoor strip light (6m)',
+        extend: [philips.m.light({"colorTemp":{"range":[50,1000]},"color":{"modes":["xy","hs"],"enhancedHue":true}})],
+     },
+};
```
andrei-lazarov: What is that range? Did that come up in the generated definition?
```suggestion
        extend: [philips.m.light({colorTemp: {range: [153, 447]}, color: {modes: ["xy", "hs"], enhancedHue: true}})],
    },
```

Holgeretti: Yes, this is from the automatic generated definition.

andrei-lazarov: Ok! Please remove the extra bracket at the end, and we can merge it. 
```
};
```

Koenkk:
Thanks!

PR #11896, opened 2026-04-07
Attempt to fix IKEA INSPELNING energy spikes

andrei-lazarov:
Plug has dynamic divisors and possibly multipliers.. Can't know for sure which ones are needed.
Users report energy spikes, so hopefully this will fix it.
- https://github.com/Koenkk/zigbee2mqtt/issues/25929


andrei-lazarov:
Doesn't seem to help. Affected unit has normal divisor/multiplier: https://github.com/Koenkk/zigbee2mqtt/issues/25929#issuecomment-4199845904

PR #11897, opened 2026-04-07
Fix LIDL bulbs part 1

andrei-lazarov:
There's more work needed here. Device uses some crazy custom commands, that I need to implement. 
But this is enough to _prevent_ locking the bulb into color temp mode
- https://github.com/Koenkk/zigbee-herdsman-converters/issues/11872


Koenkk:
Thanks!

kwetnico:
.The problem is still not fixed even after updating to version 2.10.0. It's impossible to control the colors. Even after removing and re-adding the LED strip, the problem persists.

@Koenkk  and @andrei-lazarov

PR #11904, opened 2026-04-08
EMIZB-132: Simplify and hopefully improve the power issue

svhelge:
Enforce that only energy is read from seMetering by implementing `fzMetering: develco.fz.metering_emizb132`. Hard code the divisor for energy.


Koenkk:
Thanks!

PR #11910, opened 2026-04-09
Add custom clusters and attributes to CTM devices. 

svhelge:
Added cluster definitions provided by CTM, thanks @soversk.  
I've added modernextend functions and updated devices with clusters. 
Removed code that wasn't used by the devices.


Koenkk:
Thanks!

PR #11911, opened 2026-04-10
fix: use correct final position for ZNCLBL01LM after stop

martinfrancois:
## Summary
- when `ZNCLBL01LM` changes from running to stopped through attr `1057`, read `closuresWindowCovering.currentPositionLiftPercentage` and use that as the final result
- keep attr `107` for movement updates, including the first update when movement starts again
- while the motor is stopped, do not use attr `107` as the final source for `position` / `state`
- add tests for stop reports, repeated stop reports, stopped-state suppression, resume behavior, and both `invert_cover` directions

## Why
For `ZNCLBL01LM`, `position` / `state` can come from two places:
- manufacturer attr `107`
- `closuresWindowCovering.currentPositionLiftPercentage`

The problem is that near the end of movement, attr `107` can stay stale after the motor has already stopped.
When that happens, it can overwrite the real final state.

I ran into this in real use.
The cover could stay in a moving state with a near-end position, instead of settling to a clean final state.
That broke automations that depend on the final cover state being correct.

A simple idea would be to normalize `1 -> 0` and `99 -> 100`.
I did think about that, but I do not think that is the right fix here.

That would only hide the symptom.
It would still leave attr `107` in charge of the final state while the motor is stopped.

It is also risky, because `1` and `99` are not always wrong.
They can be real near-end positions.

This patch fixes the real problem instead.
It keeps attr `107` for movement updates, but after stop it reads the standard window covering position and uses that as the final result.

## Notes
- only `ZNCLBL01LM` is changed
- this patch does not add `1 -> 0` or `99 -> 100` normalization

### DIFF
--- a/src/lib/lumi.ts
+++ b/src/lib/lumi.ts
@@ -44,6 +44,7 @@ import {
 const NS = "zhc:lumi";
 const e = exposes.presets;
 const ea = exposes.access;
+const ZNCLBL01LM_RUNNING_STORE_KEY = "ZNCLBL01LM_running";
 
 declare type Day = "mon" | "tue" | "wed" | "thu" | "fri" | "sat" | "sun";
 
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #11915, opened 2026-04-11
Fix Inovelli mmWave target info report.

kbuck1:
Each record in the report is 10 bytes, not 9 bytes - this incorrect stride caused incorrect deserialization of the target info report, which causes targets after the first one to be reported with bogus coordinates.

Also, remove the clamping on the `Target ID` field as it seems to be a unique identifier.


kbuck1:
BTW, here's the documentation about `0x01 - Report Target Info`: https://help.inovelli.com/en/articles/12773613-blue-series-mmwave-presence-dimmer-switch-advanced-mmwave-configuration

Each target reports 5 fields of size `int16`, so 10 bytes total for each target.

Koenkk:
thanks!

PR #11916, opened 2026-04-11
fix: Move Tuya specific attributes in lightingColorCtrl cluster from ZH to ZHC - implemented in tuyaLight

svhelge:
Tuya specific lightingColorCtrl attributes added to tuyaLight.


Koenkk:
Thanks!

PR #11922, opened 2026-04-11
Update lonsonho.ts

L0rDoCH:
Switch Lonsonho X701A detected as Tuya TS0001. Adding model to whitelabel.


L0rDoCH:
I mb have made the wrong correction  file. I think enough adding model to fingerprint, but I'm doing how it says in FAQ "Fixing Tuya device detection"

Koenkk:
Thanks!

PR #11926, opened 2026-04-12
Fix EMIZB-141 Pulse configuration

andrei-lazarov:
- I think this fixes: https://github.com/Koenkk/zigbee2mqtt/discussions/31684

Can you check @svhelge? Maybe more devices need it


svhelge:
That should fix the missing attribute `develcoPulseConfiguration`.  EMIZB-141 is the only device using `develcoModernExtend.pulseConfiguration()`.

Koenkk:
Thanks!

PR #11924, opened 2026-04-12
ci: use CODEOWNERS for external review gating

rohankapoorcom:
## Summary

- Introduces a `CODEOWNERS` file at the repo root as the single source of truth for file-to-reviewer mappings.
- Rewrites the `requires_review.yml` workflow to parse `CODEOWNERS` at runtime instead of maintaining a hardcoded associative array.
- The workflow now matches changed files using bash glob patterns (supports CODEOWNERS glob syntax).
- On first detection of owned files in a PR, the workflow **posts a comment tagging the code owners**, so they get a GitHub notification even without write access to the repo.
- Subsequent pushes to the same PR skip the duplicate comment (detected via an HTML marker).
- Adds `src/lib/inovelli.ts` and `test/inovelli.test.ts` to the Inovelli ownership alongside the existing `src/devices/inovelli.ts`.

## How it works

For example, if someone opens a PR that changes `src/lib/inovelli.ts`:

1. The workflow parses `CODEOWNERS` and finds `src/lib/inovelli.ts @InovelliUSA @rohankapoorcom`.
2. It matches the changed file against that pattern.
3. It posts a PR comment:
   > **@InovelliUSA @rohankapoorcom**:
   > - `src/lib/inovelli.ts`
   >
   > Once reviewed, add the `external-reviewer-approved` label to unblock this check.
4. The check **fails** (red status on the PR).
5. On subsequent pushes, the file still matches and the check still fails, but no duplicate comment is posted.
6. Once a maintainer adds the `external-reviewer-approved` label, the workflow re-runs via the `labeled` trigger and **exits successfully** (green status).

## Adding new owners

To gate a new file or vendor, just add a line to `CODEOWNERS` — no workflow changes needed:
```
src/devices/newvendor.ts    @vendor-github-handle
```

## Test plan

- [ ] Verify workflow parses `CODEOWNERS` correctly (blank lines, comments, whitespace)
- [ ] Verify glob matching works for exact paths
- [ ] Verify comment is posted with correct owners and file list
- [ ] Verify duplicate comments are not posted on subsequent pushes
- [ ] Verify `external-reviewer-approved` label bypasses the check
- [ ] Verify existing vendors (Vesternet, Candeo, Efekta) still trigger correctly

Made with [Cursor](https://cursor.com)

PR #11944, opened 2026-04-14
fix(ci): grant pull-requests write for requires_review workflow

rohankapoorcom:
Sorry for the regression introduced in [#11924](https://github.com/Koenkk/zigbee-herdsman-converters/pull/11924): the updated `requires_review` workflow posts a PR comment so code owners get notified, but I did not grant `pull-requests: write` on the workflow. The default `GITHUB_TOKEN` then could not call `addComment`, which surfaced as **GraphQL: Resource not accessible by integration (addComment)** and could abort the step before the intended review gate.

## Changes

- Declare `permissions: contents: read` and `pull-requests: write` so `gh pr comment` is allowed when the repository policy permits it (same idea as `ci.yml`).
- If commenting still fails (e.g. some fork PRs with a read-only token), log a workflow warning instead of failing the job on that step alone; the check still fails with `exit 1` until `external-reviewer-approved` is applied.

### DIFF
--- a/.github/workflows/requires_review.yml
+++ b/.github/workflows/requires_review.yml
@@ -9,6 +9,10 @@ on:
       - labeled
       - unlabeled
 
+permissions:
+  contents: read
+  pull-requests: write
+
 jobs:
   requires-external-review:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11946, opened 2026-04-14
fix(detect): Detect `_TZE200_vawy74yh` as Moes ZSS-HM-SSD01

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee2mqtt/issues/30945


Koenkk:
Thanks!

PR #11947, opened 2026-04-14
ZB-RS485 support more features

xsp1989:

### DIFF
--- a/src/devices/easyiot.ts
+++ b/src/devices/easyiot.ts
@@ -814,6 +814,96 @@ const tzLocal = {
             );
         },
     } satisfies Tz.Converter,
+
+    /**
+     * Configure RS485 serial port parameters using the data point protocol.
+     * Each setting is an independent data point command, concatenated and sent in one tunneling transfer.
+     *
+     * Baud rate  (0x0201): [01 02 04 23 <uint32-LE>]
+     * Parity     (0x0205): [05 02 01 30 <0=NONE|1=EVEN|2=ODD>]
+     * Stop bits  (0x0209): [09 02 01 30 <0=1bit|1=1.5bits|2=2bits>]
+     *
+     * TunnelID must be 0x0001 per the data-point tunneling protocol.
+     */
+    easyiot_rs485_set_config: {
+        key: ["rs485_config"],
+        convertSet: async (entity, key, value, meta) => {
+            assertObject(value, "rs485_config must be an object");
+            const payload = value as KeyValueAny;
+
+            const commands: Buffer[] = [];
+
+            // --- Baud rate: 0x0201, data type uint32 (0x23), little-endian ---
+            if (payload.baud_rate !== undefined) {
+                const baudRateMap: KeyValueAny = {
+                    "1200": 1200,
+                    "2400": 2400,
+                    "4800": 4800,
+                    "9600": 9600,
+                    "19200": 19200,
+                    "38400": 38400,
+                    "57600": 57600,
+                    "115200": 115200,
+                    "230400": 230400,
+                    "460800": 460800,
+                    "921600": 921600,
+                };
+                const baudRate = String(payload.baud_rate);
+                if (!(baudRate in baudRateMap)) {
+                    throw new Error(`baud_rate must be one of: ${Object.keys(baudRateMap).join(", ")}`);
+                }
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Koenkk:
@xsp1989 we are checking wether we can expose this device through a [HA IR entity](https://github.com/home-assistant/home-assistant.io/pull/45739) (https://github.com/home-assistant/core/pull/172796). To check wether it fits, could you give some examples of data provided to `send_command` and what kind of values are expected for `last_received_command`? (CC: @jbouwh)

xsp1989:
Here is the protocol documentation for the ZB-IR01: https://www.zigbee2mqtt.io/devices/ZB-IR01.html.
For the `send_command` function, the protocols are fixed (except for air conditioner-specific settings), for example:
Power [ON]: 86 01 00 00 00 87
Temperature [17°C]: 86 01 02 01 00 84
Mode [Cooling]: 86 01 01 01 00 87

Regarding `last_received_command`:
The expected received results are:
Success: 0x06, 0x89, 0x01, 0x00, 0x00, 0x8E
Failure: 0x06, 0xE0, 0xE1, 0x00, 0x00, 0x07

We can also provide a small number of samples if needed.

jbouwh:
The challenge is to find the timings. So I need the IR data for a known command (a command where we know the timings).
This is not about how to sensor and receive, but the actual data that is to be sent or received.
It does not look timings of an IR signal are actually used. More that it works with fixed codes, or that an IR signal can be stored a certain position.
HA works with a low level vendor independent coding. So I am not sure if this is going to work at all.

xsp1989:
The ZB-IR01 contains infrared codes for most commonly used devices. You only need to know how to invoke them and do not need to pay attention to the infrared timing during use.

Obtaining and transmitting low-level infrared timing data is usually difficult because each device uses different timing protocols. It is challenging to collect sufficient timing data from a large number of devices. In practice, we typically purchase this infrared code database from commercial companies.

jbouwh:
Home Assistant has a [library](https://github.com/home-assistant-libs/infrared-protocols) that solves the timings, that is how infrared entities work, but maybe we should look into a route where we use [remote](https://developers.home-assistant.io/docs/core/entity/remote/) entities to make this work.
The supported commands/codes could be part of the configuration.

xsp1989:
I don't think the focus should be solely on the low-level transmission interface. The real challenge of building an infrared remote control system has never been transmitting infrared signals; it has always been collecting and maintaining infrared codes for different brands and models.

One important point to note is that air conditioners are different from most other devices such as TVs and projectors. Air conditioner infrared commands are typically encoded as a calculated composite data packet that contains multiple parameters, such as power state, temperature, operating mode, and fan speed. The infrared signal captured from the original remote control also contains all of this information.

In contrast, for most other devices, each button corresponds to a single infrared command, making the implementation and code management much simpler.

jbouwh:
That's why this will not work with an HA Infrared Entity. A Remote Entity on the other hand could work, but it requires to configure what codes/keys are supported. I guess there could be different entities for each brand. 
Let me know if you would like to go in that direction. You can contact me at Discord if you like.

xsp1989:
As far as I know, most infrared remote control products on the market work in a similar way. They connect to the manufacturer's platform through Wi-Fi, Bluetooth, Zigbee, or other communication technologies. The manufacturer's cloud server then communicates with the infrared code database provider's server. Once a device has been successfully matched, an infrared data package is downloaded to the remote control device through an API, completing the pairing process. In this model, the device itself typically does not contain the original infrared code database.

This is the operating model used by the vast majority of infrared remote control products. Due to some special requirements, we chose to fully encode a portion of commonly used infrared codes directly into the ZB-IR01. As a result, it can operate independently of any cloud server.

There may be other products that support a similar offline operating model, but regardless of the implementation, the device interface protocol is unlikely to be universal. The reason is that the Zigbee Alliance has never defined a standard device type or protocol for infrared remote controllers. Therefore, manufacturers have generally implemented their own proprietary protocols.

I do have a Discord account, but I rarely use it. If it would be helpful, I would be happy to provide some suggestions and feedback. What would be the best way for me to contact your team?

jbouwh:
via jbouwh at Discord.

PR #11951, opened 2026-04-15
test(inovelli): phase 3 configure/OTA coverage and test hardening

rohankapoorcom:
## Summary

This pull request extends **Inovelli** automated coverage and tightens how those tests are structured so they stay reliable as the suite grows.

### Phase 3: configure and OTA

- Adds **configure** and **OTA** coverage for Inovelli models that use the shared Inovelli configure path (including **VZM35** / **VZM36** and related cases), using the same patterns as existing Phase 1/2 tests (mock device, `configure`, `read`/`write` expectations).
- Reuses small helpers (**`patchAndConfigure`**, **`runInovelliConfigure`**, etc.) so configure tests stay readable without copy-pasting large mock blocks.

### Test hardening (order and setup)

- Replaces the fragile pattern where **`definition`** was assigned inside the **first** `it(...)` of a block (which breaks under reordering, `it.only`, or future Vitest parallelism).
- Uses **`beforeAll(async () => { definition = await findByDevice(...) })`** (and similar) for **VZM31-SN**, **VZM32-SN**, **VZM35-SN**, **VZM36**, and related **fromZigbee** groups so nested tests always see a defined `definition`.

### Firmware-gated exposes

- For **VZM31** firmware buckets under **`describe("Inovelli firmware-gated exposes")`**, **`resolveExposes`** runs **once per firmware `describe`** via **`beforeAll`**, and individual tests only assert on the cached **`exposes`** list (less boilerplate, same assertions).
- **VZM30** and **VZM32** checks that **`switchType`** must **not** include **Single-Pole Full Sine Wave** on older firmware are collapsed into **`it.each`** so old vs new firmware stays explicit without duplicated tests.

### Readability: EP2 scene / button payloads

- Introduces **`rawInovelliEp2Scene(...)`** in the test file with a short pointer to **`buttonLookup` / `clickLookup`** in **`src/lib/inovelli.ts`**, so raw **`[0, 0, 0, 0, …]`** buffers are tied to the production lookup tables instead of being opaque magic arrays.

## Testing

- `pnpm run build`
- `pnpm run check`
- `pnpm test`

All pass locally after these changes.

### DIFF
--- a/test/inovelli.test.ts
+++ b/test/inovelli.test.ts
@@ -1,8 +1,13 @@
-import {describe, expect, it, vi} from "vitest";
+import {beforeAll, describe, expect, it, vi} from "vitest";
 import {definitions as inovelliDeviceDefinitions} from "../src/devices/inovelli";
 import {findByDevice} from "../src/index";
 import type {Definition, Expose, Fz, KeyValue, KeyValueAny, Tz} from "../src/lib/types";
 import {mockDevice} from "./utils";
+
+/** EP2 raw scene buffer: `data[4]` must be `0x00` for scene parsing; `data[5]` / `data[6]` index `buttonLookup` / `clickLookup` in `src/lib/inovelli.ts`. */
+function rawInovelliEp2Scene(data4: number, buttonLookupKey: number, clickLookupKey: number): number[] {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11952, opened 2026-04-15
Update tuya.ts

tiankakeji-cloud:

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -7409,7 +7409,7 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_pzm3wab5", "_TZE284_twybxdzl", "_TZE284_hgeqeyuv"]),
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -7409,7 +7409,7 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_pzm3wab5", "_TZE284_twybxdzl", "_TZE284_hgeqeyuv"]),
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_twybxdzl"]),
```
andrei-lazarov, line 7412:
```suggestion
        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_pzm3wab5", "_TZE284_twybxdzl", "_TZE284_hgeqeyuv"]),
```

Hi! This part is still needed.
Fingerprint means: "apply this converter to the following devices"
WhiteLabel means: "display these devices with a different name and picture"

tiankakeji-cloud:
Perfect! Thank you so much for the quick reply and the clear example. I'll implement this right away.

PR #11953, opened 2026-04-15
fix: replace hardcoded powerSource strings by proper ZCL type.

adriweb:
This became needed to avoid type errors, for a device integration I'm working on that uses another power source string.
I figured I might as well propose this here now.


Koenkk:
Thanks!

PR #11954, opened 2026-04-15
fix(ci): run external review gate with pull_request_target

rohankapoorcom:
## Why this change is needed

The `requires_review` workflow enforces CODEOWNERS-based external review and posts a PR comment tagging owners. On fork-based PRs, the `pull_request` event often executes with a read-only `GITHUB_TOKEN`, so `gh pr comment` can fail with:

- `GraphQL: Resource not accessible by integration (addComment)`

That leaves reviewers untagged and makes the gate harder to use in practice.

## Why previous attempts didn't fully solve it

A previous fix added workflow permissions (`pull-requests: write`) and made the comment call non-fatal. That improved resilience, but it still depended on the token granted under the `pull_request` event. For fork PRs, GitHub can still provide a read-only token despite requested permissions, so comment posting remained blocked.

Related:
* https://github.com/Koenkk/zigbee-herdsman-converters/pull/11944
* https://github.com/Koenkk/zigbee-herdsman-converters/pull/11924

## What this change does

- Switches the workflow trigger to `pull_request_target`.
- Keeps minimal required permissions (`contents: read`, `pull-requests: write`).
- Uses the PR API (`gh pr view --json files`) to get changed files instead of relying on a git diff of PR head.

This keeps the job in base-repo context for reliable commenting while avoiding execution/diff logic tied to untrusted PR head code.

Reference: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

## Test plan

- Open or update a fork-based PR that touches a CODEOWNERS-gated file.
- Verify the workflow emits the expected gate error annotation.
- Verify the workflow can post (or detect existing) gate comment with tagged owners.
- Add `external-reviewer-approved` and verify the check turns green.

### DIFF
--- a/.github/workflows/requires_review.yml
+++ b/.github/workflows/requires_review.yml
@@ -6,6 +6,7 @@ on:
       - opened
       - synchronize
       - reopened
+      - ready_for_review
       - labeled
       - unlabeled
 
--- a/.github/workflows/requires_review.yml
+++ b/.github/workflows/requires_review.yml
@@ -1,7 +1,7 @@
 name: Requires external review
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - synchronize
@@ -19,7 +19,7 @@ jobs:
     steps:
     - uses: actions/checkout@v6
       with:
-        fetch-depth: 0
+        fetch-depth: 1
 
     - name: Check for files that require external review
       env:
@@ -45,7 +45,8 @@ jobs:
           PATTERN_REVIEWERS["$pattern"]="$owners"
         done < CODEOWNERS
 
-        CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }})
+        # pull_request_target runs in base-repo context; get changed files from the PR API.
+        CHANGED_FILES=$(gh pr view "$PR_NUMBER" --json files --jq '.files[].path')
 
         MATCH=false
         declare -A MATCHED_REVIEWERS=()


Koenkk:
`pull_request_target` seems to have some security implications, I prefer not to use it.

> For fork PRs, GitHub can still provide a read-only token despite requested permissions, so comment posting remained blocked.

I don't think this workflow is relevant for fork PRs, I think it should only run when made against this repo?

rohankapoorcom:
> pull_request_target seems to have some security implications, I prefer not to use it.

I had similar concerns so came up with a decent workaround. If you look at the implementation, we don't actually run any code from the fork branch (in fact we don't even clone it), so the security implication should be extremely limited. We don't even pull the branch being tested, only the mainline branch from this repo and use the gh cli to get the list of files that the PR branch has. In fact, this usage is exactly what [Github recommends](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) in this situation. (emphasis mine)

> This event runs in the context of the default branch of the base repository, rather than in the context of the merge commit, as the pull_request event does. **This prevents execution of unsafe code from the head of the pull request** that could alter your repository or steal any secrets you use in your workflow. **This event allows your workflow to do things like label or comment on pull requests from forks**. Avoid using this event if you need to build or run code from the pull request.

Since we're not building any code, it should be completely safe.

> > For fork PRs, GitHub can still provide a read-only token despite requested permissions, so comment posting remained blocked.
> 
> I don't think this workflow is relevant for fork PRs, I think it should only run when made against this repo?

Actually every PR in this repo that is not created by a maintainer is a fork PR since it's coming from a fork to your repo. Only maintainers can create normal branches in this repo directly.

Koenkk:
> any code from the fork branch (in fact we don't even clone it)

Ah yes, ofc it's triggered by the event, not on the pr branch, thanks for explaining!

rohankapoorcom:
Thanks Koen!

PR #11956, opened 2026-04-15
Add Philips Hue Festavia globe outdoor string lights (21m, 30 bulbs)

psde:
Added support for Hue Festavia globe outdoor string lights with model 929004582001 (21m, 30 bulbs).

This is a copy of the 14m one, and we can reuse the same picture as the 14m.


Koenkk:
Thanks!

PR #11957, opened 2026-04-16
fix: Emit state from genBinaryInput fixing refresh on btcino k4001c

murarustefaan:
## Summary

The BTicino K4001C light switch on latest firmware (4408831) seems to have decoupled the `genOnOff.onOff` attribute from the actual relay state - only `genBinaryInput.presentValue` reflects reality. Result: pressing the physical button toggles the relay and emits `action` events, but `state` in z2m never updates, and the Refresh button in the UI snaps state back to the wrong value.

Also described here https://github.com/Koenkk/zigbee2mqtt/issues/29243.

Validated against a real K40001C switch.

Happy to apply any changes you deem necesary.

### DIFF
--- a/src/devices/bticino.ts
+++ b/src/devices/bticino.ts
@@ -20,6 +20,7 @@ export const definitions: DefinitionWithExtend[] = [
         toZigbee: [tzLegrand.K4003C_state, tzLegrand.led_mode, tzLegrand.identify],
         extend: [legrandExtend.addLegrandDevicesCluster()],
         exposes: [e.switch(), e.action(["identify", "on", "off"]), eLegrand.identify(), eLegrand.ledInDark(), eLegrand.ledIfOn()],
+        version: "0.0.1",
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ["genIdentify", "genOnOff", "genBinaryInput"]);
--- a/src/devices/bticino.ts
+++ b/src/devices/bticino.ts
@@ -16,13 +16,16 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "BTicino",
         description: "Light switch with neutral",
         ota: true,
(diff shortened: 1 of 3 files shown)


jetmcquack:
This is great!! Thanks for your work! I can’t wait for it to be merged in order to try it out!! @Koenkk can you help us with this merge?

src/devices/bticino.ts
```diff
@@ -16,13 +16,16 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "BTicino",
         description: "Light switch with neutral",
         ota: true,
-        fromZigbee: [fz.identify, fz.on_off, fz.K4003C_binary_input, fzLegrand.cluster_fc01],
-        toZigbee: [tz.on_off, tzLegrand.led_mode, tzLegrand.identify],
+        fromZigbee: [fz.identify, fz.K4003C_binary_input, fzLegrand.cluster_fc01],
+        toZigbee: [tzLegrand.K4003C_state, tzLegrand.led_mode, tzLegrand.identify],
         extend: [legrandExtend.addLegrandDevicesCluster()],
         exposes: [e.switch(), e.action(["identify", "on", "off"]), eLegrand.identify(), eLegrand.ledInDark(), eLegrand.ledIfOn()],
         configure: async (device, coordinatorEndpoint) => {
```
andrei-lazarov, line 24: Version bump might be needed to trigger the new reporting? Wacky device

```suggestion
        version: "0.0.1",
        configure: async (device, coordinatorEndpoint) => {
```

Koenkk:
Thanks!

Supersimo88:
After the latest update, the K4003c devices report the wrong status at random; before, they worked perfectly.
I find that the lights are marked as on even though they're actually off. After turning them off from the web interface, which obviously doesn't change the status of the actual lights, since they're already off, they remain synchronized for a short while and then incorrectly report them as on again even though they're off.

guiumb:
I have the same issue after the update.

murarustefaan:
@Koenkk any chance you could help with helping me merge the revert PR referenced above and releasing a patch version for z2m for it, so we can fix this issue for K4001C users?

PR #11959, opened 2026-04-16
fix(smarli): remove endpoint name from COV1 single-channel cover

loocd:
endpoint name is unnecessary on non-multi-channel-devices and only adds noise to naming conventions. remove to achieve null value. e.g. relevant for home assistant original_name, which is used to construct names and IDs


Koenkk:
Thanks!

PR #11961, opened 2026-04-17
lincukoo: Add _TZE284_sndkanfr fingerprint for SZLMR10

pauloromeira:
Adds support for a new Tuya manufacturer variant `_TZE284_sndkanfr` for the Lincukoo Human Motion & Presence Sensor (SZLMR10).

The device uses the same model but a different manufacturerName, causing it to not match the existing fingerprint.

No functional changes, only extends the fingerprint list.


Koenkk:
Thanks!

PR #11966, opened 2026-04-17
fix(detect): Detect `_TZE284_4fblxpma` as NEO NAS-WV03B

andrei-lazarov:
- https://github.com/Koenkk/zigbee2mqtt/issues/31724
- also added the 284 model
- verified the datapoints


Koenkk:
Thanks!

PR #11969, opened 2026-04-18
feat(add): SLZB-06MG26U router variant

Nerivec:
I also removed a couple of unknowns that apparently were added a while back. Not sure what these are, but they're not in my releases (which use custom fingerprint/features). :shrug:


Koenkk:
Thanks!

PR #11977, opened 2026-04-19
Fix: Clean up redundant tuyaBase usage caused by moving manuSpecificTuya3 from ZH to ZHC

svhelge:
When the manuSpecificTuya3 cluster was moved from ZH to ZHC (PR #11751) the cluster definition was added to tuyaBase. 
tuyaBase was added to more devices than strictly necesssary and seem to have caused problems, ref #11872 . 
I have examined the changes in PR #11781 and reverted some of them.


Koenkk:
Thanks!

PR #11968, opened 2026-04-18
test(inovelli): add VZM31-SN and VZM30-SN definition integration tests (phases 4-5)

rohankapoorcom:
## Summary

Adds end-to-end integration tests for Inovelli **VZM31-SN** (2-in-1 dimmer) and **VZM30-SN** (on/off switch) device definitions, completing phases 4 and 5 of the Inovelli comprehensive test plan. Each test exercises the full definition lookup + `configure()` pipeline and asserts structural properties of the resulting device: `meta`, `fromZigbee` fingerprints, `toZigbee` keys (both inclusion and omission), exposes, and per-endpoint `bind` / `read` / `write` / `configureReporting` calls.

Motivated by the observation that the built-in `assertDefinition` helper relies on referential equality for `fromZigbee` / `toZigbee` converter arrays — which does not work for Inovelli's factory-generated \`fzLocal\` / \`tzLocal\` converters. A new \`assertInovelliIntegration\` helper performs structural fingerprint comparisons instead, making these integration tests stable regardless of internal converter identity.

### VZM31-SN (phase 4)
- Firmware **3.07** happy path: full integration shape (8 fromZigbee converters, EP1 bind + 13 reads + 3 reporting calls, EP2 \`manuSpecificInovelli\` bind, multi-endpoint meta with \`state/power/energy/brightness\` skip list, all firmware-gated attributes present).
- Firmware **2.18** regression guard: verifies that \`dimmingAlgorithm\`, \`auxDetectionLevel\`, and \`dumbDetectionLevel\` reads are **not** issued, while common attributes (\`switchType\`, \`fanControlMode\`) remain.

### VZM30-SN (phase 5)
- Firmware **3.07** full integration shape: 10 fromZigbee converters (adds \`msTemperatureMeasurement\` + \`msRelativeHumidity\`), EP1 bind of 7 clusters + 15 reads + 5 reporting calls (including multi-attribute \`haElectricalMeasurement\` reporting for \`activePower\` / \`rmsCurrent\` / \`rmsVoltage\`), multi-endpoint meta with extended skip list including \`voltage/current/temperature/humidity\`.
- Asserts switch-specific **omissions** (\`dimmingAlgorithm\`, \`auxDetectionLevel\`, \`dumbDetectionLevel\` are explicitly absent from \`toZigbee\`).
- Asserts switch-specific **inclusions** not present on VZM31 (\`current\`, \`voltage\`, \`temperature\`, \`humidity\`).

## New helper

[...]

PR #11980, opened 2026-04-20
test(inovelli): add VZM32-SN definition integration test (phase 6)

rohankapoorcom:
## Summary

Adds an end-to-end integration test for the Inovelli **VZM32-SN** (mmWave Zigbee Dimmer) device definition, completing phase 6 of the Inovelli comprehensive test plan. Continues the hardening effort from #11951 (phases 1–3) and #11968 (phases 4–5); builds on the `assertInovelliIntegration` helper introduced in #11968, which uses structural fingerprint comparisons (not referential equality) so it works with Inovelli's factory-generated `fzLocal` / `tzLocal` converters.

VZM32-SN is the most feature-dense Inovelli dimmer — it exposes **dual custom clusters** (`manuSpecificInovelli` + `manuSpecificInovelliMMWave`) on endpoint 1 alongside standard `haElectricalMeasurement`, `seMetering`, `msIlluminanceMeasurement`, and `msOccupancySensing`, plus a `manuSpecificInovelli` binding on endpoint 2. Phase 6 locks in the exact integration shape this model produces.

### VZM32-SN integration test
- **15** `fromZigbee` converter fingerprints covering light, device, mmWave, electricity meter, illuminance, and occupancy.
- **75** `toZigbee` keys spanning light, LED effects, both Inovelli parameter sets (standard + mmWave), control commands, electricity meter, illuminance, and occupancy.
- **106** expose fingerprints (sorted, exact match), including mmWave-specific exposes (`mmwave_areaX_occupancy`, `mmwave_targets`, `mmwave_control_commands`, and the full set of mmWave detection / area parameters) alongside standard light, power, and sensor exposes.
- Per-endpoint **bind / read / write / configureReporting** assertions:
  - EP1 binds 8 clusters (`genOnOff`, `genLevelCtrl`, `manuSpecificInovelli`, `manuSpecificInovelliMMWave`, `haElectricalMeasurement`, `seMetering`, `msIlluminanceMeasurement`, `msOccupancySensing`).
  - EP1 issues **18 reads** across 6 clusters (chunked reads for both custom clusters plus the standard sensor clusters) and **0 writes**.
  - EP2 binds only `manuSpecificInovelli`; EP3 binds nothing.
  - 5 `configureReporting` calls on EP1 (`genOnOff`, `haElectricalMeasurement`, `seMetering`, `msIlluminanceMeasurement`, `msOccupancySensing`) with specific min / max / reportableChange values.
- **mmWave query_areas** assertion: the test verifies the `manuSpecificInovelliMMWave.mmWaveControl` command with `{controlID: 2}` is issued during `configure()`.

### Firmware-gating regression guard
A second test asserts that `dimmingAlgorithm`, `auxDetectionLevel`, and `dumbDetectionLevel` never appear in VZM32-SN exposes (and are never read at configure-time), regardless of firmware — these attributes have no `addedInFirmware[VZM32]` entry and are gated off by the model config. Baseline dimmer attributes (`switchType`, `dimmingSpeedUpRemote`) are still verified present.

## New helper capability

Extends `assertInovelliIntegration` with an optional **`commands`** expectation field, mapping endpoint ID → list of `{cluster, command, payload}` entries to assert via `endpoint.command(...)` call inspection. Needed here for the mmWave `query_areas` command, and reusable by future integration tests that need to assert command-side configure behavior (not just bind / read / write / reporting).

## Cleanup

- Corrected unrealistic firmware strings in the pre-existing VZM32-SN firmware-gating tests: VZM32 only ships with 1.x firmware, so `"2.18"` / `"3.05"` were replaced with `"1.0"` / `"1.15"` (the `it.each` titles updated to `"early 1.x firmware"` / `"later 1.x firmware"`). These tests still exercise the same "firmware does not affect VZM32 exposes" behavior — VZM32's own `modelValues` entries don't carry the `conditionalValues` keyed on `[3, 0]`, so any 1.x firmware yields the same result as the prior 2.x / 3.x strings.

## Test plan

- [x] `pnpm test` — full suite passes (146 tests in `inovelli.test.ts`, all green).
- [x] `pnpm run build` — succeeds, 4232 definitions processed.
- [x] `pnpm run check` — no new lint errors.
- [x] VZM32-SN configure pipeline exercised: binds, chunked reads across both custom clusters, reporting, and mmWave `query_areas` command all asserted.
- [x] Firmware-gating regression: dimmingAlgorithm / auxDetectionLevel / dumbDetectionLevel confirmed absent on any 1.x firmware.

## Related

- Follows #11951 (phases 1–3) and #11968 (phases 4–5, VZM31-SN / VZM30-SN).
- Phase 7 (VZM35-SN / VZM36 integration) will follow in a separate PR.

### DIFF
--- a/test/inovelli.test.ts
+++ b/test/inovelli.test.ts
@@ -172,6 +172,8 @@ function fzFingerprint(converter: {cluster: string | number; type: string | stri
 
 type ReportingItemExpectation = {attribute: string; min: number; max: number; change: number | null | "NaN"};
 type ReportingCallExpectation = {cluster: string; items: ReportingItemExpectation[]};
+/** Expected `endpoint.command(cluster, command, payload, ...)` call during `configure()`. */
+type CommandCallExpectation = {cluster: string; command: string; payload: Record<string, unknown>};
 
 interface IntegrationAssertion {
     model: string;
@@ -186,6 +188,8 @@ interface IntegrationAssertion {
     readClusters?: Record<number, string[]>;
     writeCount: Record<number, number>;
     configureReporting: Record<number, ReportingCallExpectation[]>;
+    /** Optional: assert that specific `endpoint.command(...)` calls were made during `configure()` (e.g. mmWave query_areas). */
+    commands?: Record<number, CommandCallExpectation[]>;
 }
 
 /** Assert a full Inovelli device definition against its configure-time side effects. */
@@ -250,6 +254,12 @@ async function assertInovelliIntegration(e: IntegrationAssertion): Promise<Defin
                 }
             });
         });
+
+        for (const want of e.commands?.[ep.ID] ?? []) {
+            const match = (ep.command as ReturnType<typeof vi.fn>).mock.calls.find((call) => call[0] === want.cluster && call[1] === want.command);
+            expect(match, `command ${want.cluster}.${want.command} (EP${ep.ID})`).toBeDefined();
+            expect(match?.[2]).toStrictEqual(want.payload);
+        }
     }
 
     return definition;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11984, opened 2026-04-20
Updated zigbeeModel for Hue Iris (Generation 2)

FoxRomeo:
Added additional zigbeeModel as some Hue Iris (Gen2) report the additional code.
Tested with my two old Hue Iris Gen 2.

(added no image as this extends an existing device)


Koenkk:
Thanks!

PR #11985, opened 2026-04-20
feat(danfoss): declare ota: true for Icon2

Maxxis20:
Enables Z2M to auto-discover for Icon2 firmware via the community OTA index.
Removes the "No official OTA support" banner.

Companion to Koenkk/zigbee-OTA#1148.


Koenkk:
Thanks!

PR #11987, opened 2026-04-20
Aqara W600: add external temperature support

nilvanis:
# Aqara W600: add external temperature support

## Summary

This PR adds external temperature sensor support to the existing Aqara W600 implementation.

Added the following items:
- `sensor`: `internal` / `external` (with custom label `Temperature source`, beacuse window sensor support will be in an upcoming PR)
- `external_temperature_input`

## Behavior

Based on sniffing outputs from an Aqara E1 Hub with W600 TRV and Aqara Temp Sensor, the thermostat expects:

1. a bind payload on `manuSpecificLumi` attribute `0xfff2`
2. the temperature source attribute `0x0280` to be set to external
3. subsequent external temperature updates to be forwarded via `0xfff2`

Those outputs also show that the payload needs a valid sensor identity, but this does not need to be a real device in the network. Due to that fact, this implementation uses a fixed private sensor identity internally rather than introducing a public `external_sensor_ieee` API. IEEE address is reused from `lumi_trv`.

### Public behavior

- Added `sensor` support for the W600 with:
  - `internal`
  - `external`
- Added `external_temperature_input`
- The `sensor` expose is labeled as `Temperature source`

### Implementation

A narrow W600-specific modern extend was added in `src/lib/lumi.ts` and wired into the existing W600 device in `src/devices/lumi.ts`.

The extend:

- reads the W600 source attribute (`0x0280`)
- writes W600 external bind / unbind payloads to `0xfff2`
- writes external temperature payloads to `0xfff2`
- keeps the helper logic private to `src/lib/lumi.ts`
- leaves unrelated W600 functionality unchanged

### `sensor: "external"`

- writes the external bind payload to `manuSpecificLumi` attribute `0xfff2`
- writes the source attribute to external
- returns state with `sensor: "external"`

### `sensor: "internal"`

- writes the source attribute to internal
- writes the unbind payload to `0xfff2`
- returns state with `sensor: "internal"`

### `external_temperature_input`

- accepts numeric temperature input
- requires external mode
- supports combined messages such as:

```json
{
  "sensor": "external",
  "external_temperature_input": 21.5
}

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -59,6 +59,7 @@ const {
     lumiMultiClick,
     lumiPreventLeave,
     lumiExternalSensor,
+    w600ExternalTempSensor,
     lumiReadPositionOnReport,
 } = lumi.modernExtend;
 
@@ -5410,6 +5411,7 @@ export const definitions: DefinitionWithExtend[] = [
                     minHeatSetpointLimit: {min: 5, max: 30, step: 0.5},
                 },
             }),
+            w600ExternalTempSensor(),
             m.enumLookup<"manuSpecificLumi", ManuSpecificLumi>({
                 name: "calibrate",
                 lookup: {start: 1},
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #11993, opened 2026-04-21
fix(defect): Energy today/yesterday/month missing label

toastedcornflakes:
Home assistant reports the same label for all three energy: today, yesterday, this month. 
 
image


Koenkk:
Thanks!

PR #11994, opened 2026-04-21
fix(detect): Detect `_TZ3210_2uollq9d` as BSEED TS011F_plug_1_2

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee2mqtt/issues/28785#issuecomment-3634152028

@TakissX will be detected correctly now!
> I also have a BSEED "TS011F _TZ3210_2uollq9d", it has been recognized as Tuya "TS011F_plug_1" and everything works as it should. The only problem is that it is not listed as BSEED and does not have its photo.


Koenkk:
Thanks again 😄

PR #11998, opened 2026-04-22
Revert "fix(detect): Detect `_TZB210_ue01a0s2` as MiBoxer E2-ZR (#11990)"

andrei-lazarov:
- revert https://github.com/Koenkk/zigbee-herdsman-converters/pull/11990

This caused a conflict. Device is part of two whiteLabels now. It's 2 lines above.

It's common that manufacturers reuse the same firmware on multiple devices. 
@mikelauskas This is only a cosmetic change. You can change the picture in your Z2M instance.


Koenkk:
Thanks!

PR #12008, opened 2026-04-23
philips: native control; fix effects and state reporting (attempt 2)

Mihonarium:
# philips: native encode/decode; fix effects and state reporting

Addresses #8697, Koenkk/zigbee2mqtt#18406, Koenkk/zigbee2mqtt#15891, Koenkk/zigbee2mqtt#24438, and Koenkk/zigbee2mqtt#31218.

## Summary

Adds a `hue_native_control` option for native control of Philips lights.

Rewrites the Philips `manuSpecificPhilips2` encoder/decoder drafted by @LaurentvdBos based on the [Bifrost spec](https://github.com/chrivers/bifrost/blob/master/doc/hue-zigbee-format.md), fixing several data corruption bugs and adding missing functionality. Also improves effect UX: effects now report state, support color+speed in a single command, and the color wheel/effect interaction is configurable.

## Opt-in

By default, standard ZCL converters handle on/off, brightness, color, and color temperature. The behavior is unchanged from before this PR. The manuSpecificPhilips2 cluster can send all of these in a single atomic command, which is required for full effect support: changing color while an effect is running, stopping effects by changing color, and setting an effect with a specific color in one command. However, native control also changes some user-facing behavior (e.g., no simulated fade-to-off transition), so it is opt-in.

Users can enable it per-device in Z2M Settings → Settings (Specific), or globally via `device_options: { hue_native_control: true }` in configuration.yaml.

## Bug fixes

- **Y-coordinate byte packing** — `decodeGradientColors` assembled the Y value from the wrong nibbles, corrupting every gradient color's Y component
- **Off-by-two return** — `decodeGradientColors` returned `position + nColors * 3` instead of `position + size + 1`, causing the parser to land at the wrong offset for subsequent fields
- **Wrong Y scaling constant** — `GRADIENT_COLORS_MAX_Y` was `0.8413` (CIE D65 white point), should be `0.8264` (Wide Color Gamut upper bound per Bifrost spec)
- **`knownEffects` key shift** — Effect hex keys were shifted by one byte, causing wrong effects to be triggered
- **Brightness validation** — Values 0 and 255 are invalid per spec, now clamped to 1–254
- **Effect state never reported** — `philipsTz.effect.convertSet` didn't return `{state}`, so Z2M's effect state was always `null`
- **Stale effect state** — Changing color stops the active effect on the device, but Z2M kept showing the old effect name. Now cleared to `"none"`.
- **Attribute reports only for gradient devices** — The `bind("manuSpecificPhilips2", ...)` call was inside the `if (args.gradient)` block. Moved to cover all `hueEffect` devices.

## New features

- **All Bifrost effects** — Added sunset, sparkle, opal, glisten, underwater, cosmos, sunbeam, enchant (13 total including no_effect, candle, fireplace, prism/colorloop, sunrise)
- **`effect` + `color` in one command** — `{"effect": "candle", "color": "#FF4400"}` activates the effect at that color atomically. Also supports `effect_speed` in the same payload.
- **`effect_color`** (new expose) — Sets the base color of the active effect without stopping it. Accepts hex or XY.
- **`effect_color_mode`** (new per-device option) — Controls what happens when the regular color wheel is used while an effect is active:
  - `"stop"` (default): color change stops the effect (matches Hue app)
  - `"update"`: color change re-sends the effect with the new color
- **Deferred brightness on effect re-send** — When re-sending an effect (via `effect_color` or "update" mode) with an explicit brightness, the brightness is sent as a separate command after the effect, since effects reset brightness on activation.
- **`effect` and `effect_speed` access changed to `STATE_SET`** — Current values are now visible in the Z2M frontend and HA.
- **`effect_speed`** — Settable via both the new path (`philipsLightTz`) and the effect converter.
- **`gradient_scale` / `gradient_offset`** — Settable. Fixed-point 5.3 format exposed as float.
- **`gradient_style`** — Fully settable (`linear`, `scattered`, `mirrored`). Was decoded but not writable.
- **`gradient_xy`** — Lossless XY coordinates published alongside RGB hex `gradient`.
- **`philips2_raw`** — Raw hex blob published for advanced clients (e.g. Bifrost) to decode independently.
- **HSV→XY conversion** — HSV colors are now converted correctly instead of being silently dropped.
- **`transition` → `fadeSpeed`** — Mapped to the Bifrost spec's per-message fade speed field.
- **Delayed state read after effect activation** — Reads device state 1s after activating an effect to sync actual brightness.

(Feel free to remove or ask me to remove the effect_color, effect_color_mode, and related stuff if it goes against the overall philosophy.)

## Encoder/decoder rewrite

Replaced the ad-hoc hex string parsing with structured `Encode`/`DecodeManuSpecificPhilips2` functions using `DataView` for proper binary field handling. The wire order follows the Bifrost spec (gradient colors before effect speed before gradient params — reversed relative to flag bit order). Each field type is defined as a `HueTypeDetails` struct with `encode`/`decode`/`flag`/`maxLength`, iterated in wire order.

## Tests
  
- Updated `test/philips.test.ts` — corrected expected values for the MAX_Y fix
- Updated `test/generateDefinition.test.ts` — added `manuSpecificPhilips2Fz` to expected fromZigbee and new toZigbee keys/exposes for the auto-generated definition test
- Updated `test/modernExtend.test.ts` — replaced `philips.fz.gradient` with `philips.manuSpecificPhilips2Fz` in expected fromZigbee, added new toZigbee keys/exposes for the philipsLight test
- Added `test/philips2.test.ts` (61 tests):
  - Known-payload decode of all 7 Bifrost spec examples
  - Round-trip (`Encode→Decode` and `Decode→Encode→Decode`) for every field type
  - Gradient byte packing edge cases (0x000/0xFFF, Bifrost spec example x=0x123 y=0x456)
  - Wire order verification (non-monotonic flag-to-position mapping)
  - All 13 effect types
  - 108 philips tests total (47 + 61), all passing

## Backward compatibility

All changes are backward-compatible. Existing MQTT payloads, automations, and gradient scenes continue to work. New exposes (`effect_color`, `effect_speed`, `gradient_style`, `gradient_xy`, `philips2_raw`) are additive. The `effect_color_mode` option defaults to `"stop"` (Hue app behavior). Devices may need a reconfigure after updating to establish the `manuSpecificPhilips2` binding for state reports.

`philipsFz.gradient` is superseded by `manuSpecificPhilips2Fz` and is no longer registered for any device. It could be possible that external converter authors import it, but it's not documented anywhere and people are more likely to use `philipsLight()`. Probably worth considering deprecated/removing in the future? I've not removed it myself though.

## Limitations

After activating an effect, we sync the device state via a 1-second setTimeout, as some Hue effects change light color/brightness. If the device goes offline in the window, the error is silently caught. There could be more robust approaches.

## Other
I made a very cool hass dashboard card (inspired by the Hue app, but better than it)! See https://github.com/Koenkk/zigbee-herdsman-converters/pull/12008#issuecomment-4346128005


Mihonarium:
@Koenkk some of the z2m failing tests look like this:

```
FAIL  test/extensions/networkMap.test.ts > Extension: NetworkMap > Should exclude disabled devices from networkmap
AssertionError: expected { data: { routes: true, …(2) }, …(1) } to strictly equal { data: { routes: true, …(2) }, …(1) }

- Expected
+ Received

@@ -110,11 +110,11 @@
          },
          {
            "definition": {
              "description": "Hue Go",
              "model": "7146060PH",
-             "supports": "light (state, brightness, color_temp, color_temp_startup, color_xy, color_hs), power_on_behavior, effect, linkquality",
+             "supports": "light (state, brightness, color_temp, color_temp_startup, color_xy, color_hs), power_on_behavior, effect, effect_speed, effect_color, linkquality",
              "vendor": "Philips",
            },
            "failed": [],
            "friendlyName": "bulb_color",
            "ieeeAddr": "0x000b57fffec6a5b3",

```

Mihonarium:
I'll submit test fixes as a PR to z2m.

Mihonarium:
Okay, should pass z2m tests with Koenkk/zigbee2mqtt#31785.

Koenkk:
Thanks!

burmistrzak:
@Mihonarium Thank you very much for implementing this! 🙌

I was investigating this cluster myself because I found it quite strange that many Hue fixtures seemingly only support one Zigbee binding. So it makes sense for them to implement an all-in-one cluster to get around this limitation. 🧐

I've now enabled native Hue control for a few of our lights and they seem to be even more responsive! Will continue my testing.

@Koenkk I've got a hunch that this cluster might be used, alongside a custom MAC-layer multicast protocol extension, for Hue Entertainment. Do you think HE support is worth perusing at this time?

rohankapoorcom:
@burmistrzak have you seen [Bifrost](https://github.com/chrivers/bifrost)? This PR is based off the reverse engineering that was done to build Bifrost (which runs on top of Zigbee2MQTT and has [full support](https://github.com/chrivers/bifrost/blob/master/doc/implementation-status.md) of Hue Entertainment).

burmistrzak:
@rohankapoorcom Thx! I know about it (even liked it), but haven't actually tied it out yet! 😅

burmistrzak:
@Mihonarium My testing indicates that sending `TOGGLE` as `state` currently does not work when _native Hue control_ is enabled.
This mainly breaks integrations that directly interact via MQTT and rely on that command.

Default implementation for `TOGGLE`:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/943b8927d70ef21b57e89ba4d5ce0f6cb963fd87/src/converters/toZigbee.ts#L1285

Mihonarium:
@burmistrzak thanks -- I guess feel free to fix it and open a PR!

The reason native Hue control is an option that's off by default is that we've previously discussed that toZigbee does lots of things that we are not replicating, which would break the logic.

(Normal light toggling via Home Assistant works and none of my automations talk directly to mqtt, so I’ve not had a chance of noticing this myself.)

As long as you've not noticed any issues when native control is *not* enabled (we really don’t want to break any automatons), I think this is fine, though in this instance might be worth fixing. Should be fairly straightforward if you want to!

burmistrzak:
@Mihonarium Got my hands quite full at the moment, but will try to cram it in before release day. 🤞

Do you know for sure that the `ON_OFF` property is just a single bit?
Maybe values beyond `0` and `1` implement stuff like `TOOGLE` natively? A whole byte of possibilities! 😉

Mihonarium:
@burmistrzak see https://github.com/chrivers/bifrost/blob/master/doc/hue-zigbee-format.md

burmistrzak:
@Mihonarium Ok, according to that doc, it's indeed just a single bit. Seems kinda wasteful, tbh.
I assume you haven't observed any values greater than `1` for `ON_OFF` yourself, right?

Mihonarium:
Also, since this is merged, I once again want to use the opportunity to tell everyone about my take on a picture-element card. Because it's awesome.

It was inspired by the Hue app, where I could drag lots of lights around the color wheel. I wanted something at least as good; I think I made something a lot better.

It's optimized for controlling lights with as little attention and time as possible. You place lights on a 2D-canvas (e.g., according to their physical locations) and then control them via clicking/dragging to select arbitrary groups of lights and setting color/brightness/color temperature in one tap, with optional preset colors/effects and colors that already exist among other lights.

(It is also possible to add binary entities, have the default light/light group that's controlled if nothing is selected, etc.)

If you have more than three lights, that's much better than a bunch of mushroom cards. If you have over a dozen lights, it is IMO the single best user experience for controlling lights. It is meaningless to compare it to huge lists of lights/light groups.

Controls for our common spaces look like this (everything here, with one exception, works on z2m):

image

image

https://github.com/Mihonarium/HASS-Spatial-Lights-Card/

andrei-lazarov:
Tested native control a bit, looks nice! 

I found a difference with `{"brightness":0}`:
- standard: light turns off
- native: light goes to minimum brightness

Not sure which is preferred... but they should behave the same.

I must add brightness control to the tracker.
With standard commands it's always _MoveToLevelWithOnOff_. No way to send _MoveToLevel_, not good..
And the 0 and 255 values might not be treated correctly

andrei-lazarov:
If I choose an unsupported effect, I see "invalid value" in debug logs, but no feedback in Z2M.
(might need a fix at ZH level?)

burmistrzak:
> If I choose an unsupported effect, I see "invalid value" in debug logs, but no feedback in Z2M. (might need a fix at ZH level?)

@andrei-lazarov Can you give a specific example so I can reproduce on my end?

andrei-lazarov:
LCA004 1.116.3 20240329 - underwater.
I have some not updated to catch cases like this.

This just needs to be caught as an error so it reaches the frontend: [invalid_effect.log](https://github.com/user-attachments/files/27220411/invalid_effect.log)

burmistrzak:
> Tested native control a bit, looks nice!

@andrei-lazarov Yeah, it's actually pretty solid in HA **and** Apple Home via HAMH. Quite a sneaky workaround from Signify and probably explains why some Hue lights are so limited.
I guess it would be worthwhile to take apart a few fixtures and document the Hue Engine revision used and how many bindings/reports are available?

> I found a difference with `{"brightness":0}`:
> 
> * standard: light turns off
> * native: light goes to minimum brightness
> 
> Not sure which is preferred... but they should behave the same.

Should be configurable IMHO, with default to standard Z2M behavior.

> I must add brightness control to the tracker. With standard commands it's always _MoveToLevelWithOnOff_. No way to send _MoveToLevel_, not good.. And the 0 and 255 values might not be treated correctly

Yep, encounter this before as well. Is there a "default" brightness control command recommended by ZCL?

burmistrzak:
> LCA004 1.116.3 20240329 - underwater. I have some not updated to catch cases like this.

@andrei-lazarov Got them as well but not underwater and (unfortunately) already up to date. 😅

> This just needs to be caught as an error so it reaches the frontend: [invalid_effect.log](https://github.com/user-attachments/files/27220411/invalid_effect.log)

Hmm, interesting that frontend doesn't show it already... 👀👀

src/lib/philips.ts
```diff
@@ -205,14 +608,85 @@ const philipsModernExtend = {
...
+                        // Custom cluster not in registry — skip
+                    }
+                    if (supported) {
                         await ep.bind("manuSpecificPhilips2", coordinatorEndpoint);
                     }
-                });
+                }
+            });
+            // All Hue-specific effects per Bifrost spec
+            effects.push("sunset", "sparkle", "opal", "glisten", "underwater", "cosmos", "sunbeam", "enchant");
```
deviantintegral, line 630: Is this missing `sunrise`?

Mihonarium: `sunrise` is exposed on line 595, only for gradient devices; we've not changed that.

Mihonarium: (I'm not sure why; I'd think it's available on normal bulbs. Please test and PR!)

deviantintegral: Yeah, I remember testing with sunrise on a single bulb back when I was hacking around with this, but I'll confirm in case later firmware updates removed it for some reason.

Mihonarium: Thanks! #12087

SisyphusMD:
Apologies if this was not in scope of this change, but I had been hopeful: Attempting to apply the hue effect to a zigbee group appears to still not work. As far as I'm aware, it does work for groups in the native Philips app/hub. Is that a limitation that can't work via Z2M?

burmistrzak:
@SisyphusMD Hmm... Seems to work for me via HA, controlling Z2M group of Hue lights with effects.

SisyphusMD:
> @SisyphusMD Hmm... Seems to work for me via HA, controlling Z2M group of Hue lights with effects.

Thanks for getting back to me. Your comment got me to try something.

The Zigbee group I was trying to send the Hue effect to consisted of 5 Hue lights plus an Inovelli Blue switch. The switch is bound as both a controller and a "bulb" so that it stays in sync with the group's intended state, meaning brightness etc. can be changed from either the app or the physical switch.

However, the group with the switch included doesn't respond to Hue effects. Creating a new group with just the 5 Hue lights (sans switch) does trigger the effect. I presume that's just a limitation of groups containing mixed Hue/non-Hue devices.

deviantintegral:
> The Zigbee group I was trying to send the Hue effect to consisted of 5 Hue lights plus an Inovelli Blue switch.

😮 and here was I thinking it was zigbee groups vs HA groups... but I also have Blue switches. Thanks for sharing this!

Mihonarium:
Opened a draft PR to disable native control for mixed groups. idk if there's a way to send Hue-only effects to mixed groups; that would certainly be preferable if possible

SisyphusMD:
Coming back to [my earlier comment](https://github.com/Koenkk/zigbee-herdsman-converters/pull/12008#issuecomment-4365112667), I finally got effects working with the Inovelli switch included in the group. It's still sending a single command to the group, which was the part I wanted to preserve.

I've been testing it locally and it seems to work well. It needed changes in both herdsman and converters, so I made two PRs:

https://github.com/Koenkk/zigbee-herdsman/pull/1878

https://github.com/Koenkk/zigbee-herdsman-converters/pull/13197

The converters one is a draft for now since it needs the herdsman change to be merged and released first.

PR #12005, opened 2026-04-23
fix(detect): Detect `_TZE284_awepdiwi` as NEO NAS-STH02B2

andrei-lazarov:
Picked from Discord
> icetail — 19:09
Hi. i need help and dont know where to begin.. 😉
i have a https://www.zigbee2mqtt.io/devices/NAS-STH02B2.html
but it gets detected as a https://www.zigbee2mqtt.io/devices/TS0601_soil_2.html
What do/can i do now?
This? _TZE284_awepdiwi

Datapoints and info from Tuya portal:
- [_TZE284_sgabhwa6-_TZE284_g2e6cpnw.md](https://github.com/user-attachments/files/27021243/_TZE284_sgabhwa6-_TZE284_g2e6cpnw.md)
- [_TZE284_awepdiwi-_TZE284_rqcuwlsa.md](https://github.com/user-attachments/files/27021244/_TZE284_awepdiwi-_TZE284_rqcuwlsa.md)

PR #12009, opened 2026-04-23
Fix Tuya soil sensor datapoints

andrei-lazarov:
- continuation of https://github.com/Koenkk/zigbee-herdsman-converters/pull/12005

Changes:

- sorted the datapoints and reused the lookups
- effectively only added batteryState for NEO sensor
- the other datapoints are not tested, so I added them as comments


Koenkk:
Thanks!

PR #12010, opened 2026-04-23
Add fingerprint 929004610403 to Hue Flux lightstrip

steaura:
Adds the UK version of the 3m hue flux lightstrip which has a slightly different model number


Koenkk:
Thanks!

PR #12017, opened 2026-04-24
Fix fz MoveToLevel 255

andrei-lazarov:
- part 1 of reviving https://github.com/Koenkk/zigbee-herdsman-converters/pull/11244

Another undefined value case 🙂 

[Bilresa scroll-wheel remote](https://www.zigbee2mqtt.io/devices/E2490.html) sends _MoveToLevel_ commands while rotating the wheel.
The level parameter ranges from 1 to 255! (max should be 254 instead)

**255 is undefined -> comes up as null in Z2M.**

image

Screenshot From 2026-04-24 21-57-28
<p></p>

Where's the correct place to fix this? Generic fz converter or device-specific?

---

Note1: My lightbulbs seem to not care. They set level 254 on receipt of 255 (with success response).

Note2: Clicking 100% brightness in WF sends 255 on mqtt @Nerivec (but down the line it gets corrected to 254)

### DIFF
--- a/src/converters/fromZigbee.ts
+++ b/src/converters/fromZigbee.ts
@@ -1430,7 +1430,7 @@ export const command_move_to_level: Fz.Converter<"genLevelCtrl", undefined, ["co
         if (hasAlreadyProcessedMessage(msg, model)) return;
         const payload: KeyValueAny = {
             action: postfixWithEndpointName("brightness_move_to_level", msg, model, meta),
-            action_level: msg.data.level,
+            action_level: msg.data.level || 254,
             action_transition_time: msg.data.transtime / 100,
         };
         addActionGroup(payload, msg, model);


Nerivec:
This need a wide PR to create a "standard" to handle non-values throughout ZHC. Currently it is only defined at the ZH level (reception of non-value results in `NaN`, sending of `NaN` results in non-value), but ZHC is not enforcing any kind of behavior to send or receive these properly, and only a few even make use of it at all.

Probably best to start with MEs support?
For the "sending `NaN`" part from the frontend's perspective, I figure, same as enum for specific values like for e.g. cold, warm, etc. for colors, so we have the slider and the enum. Although the "label" will depend on the attribute, I suppose we could use the special cases prop from ZH clusters, but for now, it's always strings for simplicity since not used, might need refactoring

@Koenkk you're more familiar with ZHC, if you see a way to design this efficiently/easily...

Koenkk:
@Nerivec I think in this case the device does not follow the spec right? Because it should send max 254 while it sends 255, I think it should be fixed through an override via custom cluster?

Nerivec:
Was mentioning the general stuff about non-values, took a closer look at the PR itself...
This is a special case, because that cluster has derived ones, so, ZH uses the base cluster for "wider coverage"/backwards compat. 
https://github.com/Koenkk/zigbee-herdsman/blob/93f702110561225d6d97fce5cdb1c749077319a7/src/zspec/zcl/definition/cluster.ts#L673-L684
https://github.com/Koenkk/zigbee-herdsman/blob/93f702110561225d6d97fce5cdb1c749077319a7/src/zspec/zcl/utils.ts#L316-L318
This device appears to be using the base cluster (level, not level for lighting), like many others.

I think the fix is to set `max: 0xff` to
https://github.com/Koenkk/zigbee-herdsman/blob/93f702110561225d6d97fce5cdb1c749077319a7/src/zspec/zcl/definition/cluster.ts#L714
to match the attribute's range.

andrei-lazarov:
This is so strange! The ZH fix makes sense, with it using the base cluster.

Can you even make the distinction between the clusters?

Nerivec:
On paper, you'd have to check the zcl & cluster revision attribute, and match it to the derived if supported for it. If supported, a device should always use the derived cluster instead of the base.
In practice, the derived cluster 0x0008 doesn't seem to be used at all (haven't seen one yet) :sweat_smile: 
I agree, this is weird, also the sole use case of derived clusters... Probably why most stacks didn't bother supporting it...

PR #12019, opened 2026-04-25
fix(tuya): configure Aubess TMZ02 on/off reporting

Komzpa:
## Problem

Aubess TMZ02 two-gang switches with manufacturer `_TZ3000_lmlsduws` were using the generic `TS0002_basic` definition. That definition exposes the two on/off endpoints, switch type, and power-outage memory controls, but it does not configure `genOnOff.onOff` reporting.

On live devices this can leave Zigbee2MQTT/Home Assistant with stale relay state after device-side changes, power/restore behavior, or commands that happen outside the normal Z2M command path. In my case the visible symptom was a relay physically on while HA/Z2M still believed the endpoint was off.

## Change

- Move `_TZ3000_lmlsduws` out of the generic `TS0002_basic` fingerprint list into a dedicated Aubess `TMZ02` definition.
- Keep the same two-gang endpoint mapping and existing exposes:
  - `l1` endpoint 1
  - `l2` endpoint 2
  - switch type
  - power outage memory
- Add `configure` for this fingerprint:
  - Tuya magic packet setup
  - bind `genOnOff` for endpoints 1 and 2
  - configure `onOff` reporting for both endpoints

This keeps the behavior scoped to the tested Aubess fingerprint instead of changing every generic `TS0002_basic` device.

## Live validation

Tested with two physical Aubess TMZ02 devices:

- model: `TS0002`
- manufacturer: `_TZ3000_lmlsduws`
- Zigbee2MQTT model: `TMZ02`
- endpoints: `l1`/1 and `l2`/2

Manual Zigbee2MQTT reporting configuration succeeded for both devices and both endpoints via `bridge/request/device/reporting/configure`:

- `cluster`: `genOnOff`
- `attribute`: `onOff`
- endpoints: 1 and 2
- result: `status: ok` for all four endpoint/device combinations

After reporting was configured, direct Z2M state reads reflected the real endpoint states and the configured `power_outage_memory` values.

One local Home Assistant startup automation was also found to be issuing an unexpected `/l1/set ON` command during HA startup; that was a separate local automation bug and is not what this converter patch is trying to fix. This PR is only for the missing device reporting configuration.

## Validation

- `git diff --check origin/master...HEAD`
- `npx biome check --error-on-warnings src/devices/tuya.ts`
- `npm run build`
- `npm test` (`491 passed`)

GitHub Actions on the PR are also green.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6566,6 +6566,7 @@ export const definitions: DefinitionWithExtend[] = [
         ]),
         model: "TS0002_basic",
         vendor: "Tuya",
+        version: "0.0.1",
         description: "2 gang switch module",
         whiteLabel: [
             {vendor: "OXT", model: "SWTZ22"},
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6592,8 +6592,11 @@ export const definitions: DefinitionWithExtend[] = [
         meta: {multiEndpoint: true},
         configure: async (device, coordinatorEndpoint) => {
             await tuya.configureMagicPacket(device, coordinatorEndpoint);
-            await reporting.bind(device.getEndpoint(1), coordinatorEndpoint, ["genOnOff"]);
-            await reporting.bind(device.getEndpoint(2), coordinatorEndpoint, ["genOnOff"]);
+
+            for (const endpoint of [device.getEndpoint(1), device.getEndpoint(2)]) {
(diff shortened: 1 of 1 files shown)


andrei-lazarov (changes requested):
No need to create a new definition.

We can add reporting to TS0002_basic.

But I have the device and for me it reports the state anyway, with binding and reporting disabled

Koenkk:
Thanks!

andrei-lazarov (commented):
Thank you

PR #12023, opened 2026-04-25
Schneider-Electric: Add local fzConverter for thermostat with running state based on piheat

svhelge:
Replace Stelpro converter with local converter


Koenkk:
thanks!

PR #12028, opened 2026-04-26
remove RGB from Paulmann 500.48

### LINKED ISSUE #12025: Paulmann 500.48 wrongly has RGB
Hi,
the [Paulmann 500.48](https://www.zigbee2mqtt.io/devices/500.48.html#paulmann-500-48) for some reason has the RGB property set: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/paulmann.ts#L253

This is not correct, this device only supports On/Off and dimming, not RGB. The sister RGB-capable device would be the Paulmann 500.49.

I don't have a 500.49 at hand, so I am hesitant to blindly create a PR for it, how should we go forward?

karlbeecken:
Fixes #12025

partly reverts 588b0353bd9246a947922d78d21b778f8a255068

### DIFF
--- a/src/devices/paulmann.ts
+++ b/src/devices/paulmann.ts
@@ -250,7 +250,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "500.48",
         vendor: "Paulmann",
         description: "SmartHome Zigbee YourLED dim/switch controller max. 60 W",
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12030, opened 2026-04-26
added ensystec waterleak protect device, update slacky_diy.ts

slacky1965:


src/devices/ensystec.ts
```diff
@@ -0,0 +1,1269 @@
...
+interface EnsystecDefenderOpenClose {
+    attributes: {
+        positionPercent: number;
+        electricValveConnection: number;
+    };
+    commands: never;
+    commandResponses: never;
+}
+
+function getEndpointsWithCluster(device: Zh.Device, cluster: string | number, type: "input" | "output") {
```
Koenkk: A method for this already exists in `modernExtend.ts`, please move that to `utils.ts`

src/devices/ensystec.ts
```diff
@@ -0,0 +1,1269 @@
...
+    if (valueMax !== undefined) expose = expose.withValueMax(valueMax);
+
+    return expose;
+}
+
+interface ElpcCurrentSummDeliveredArgs {
+    endpointNames?: string[];
+}
+
+function elpcCurrentSummDelivered(args: ElpcCurrentSummDeliveredArgs = {}): ModernExtend {
```
Koenkk: Please use `m.electricityMeter` instead of this

src/devices/ensystec.ts
```diff
@@ -0,0 +1,1269 @@
...
+}
+
+interface ElpcPowerSourceArgs {
+    devicePower?: string[];
+    reporting?: boolean;
+    reportingConfig?: m.ReportingConfigWithoutAttribute;
+    endpointNames?: string[];
+}
+
+function elpcPowerSource(args: ElpcPowerSourceArgs = {}): ModernExtend {
```
Koenkk: Can you use `m.enumLookup` for this

src/devices/ensystec.ts
slacky1965: The function from modernExtend.ts is not accessible from this converter.

src/devices/ensystec.ts
slacky1965: I'm not sure, but I'll check.

src/devices/ensystec.ts
slacky1965: Ок.

src/devices/ensystec.ts
slacky1965: Or you need to move it from `modernExtend.ts` to `utils.ts`?

src/devices/ensystec.ts
slacky1965: Sorry, but I don't understand how `m.electricityMeter` can be used for a water consumption meter?

src/devices/ensystec.ts
slacky1965: It worked.

src/devices/ensystec.ts
slacky1965: It worked.

src/devices/ensystec.ts
slacky1965: Am I correct in assuming that the water meter needs to be updated to `electricityMeter` in `modernExtend.ts`? I just won't have time to figure it out before the end of the month. Could you please leave things as is for now, and I promise to try to make the changes to `electricityMeter` by the end of May? Thank you.

slacky1965: The `electricityMeter` can't be used. A new `waterMeter` needs to be installed, similar to the `gasMeter`.

slacky1965: I'd like to leave `elpcCurrentSummDelivered` for now because it's not a quick process for me.

slacky1965: It happened faster than I thought.

src/converters/fromZigbee.ts
```diff
@@ -950,6 +950,62 @@ export const gas_metering: Fz.Converter<"seMetering", undefined, ["attributeRepo
         return payload;
     },
 };
+export const water_metering: Fz.Converter<"seMetering", undefined, ["attributeReport", "readResponse"]> = {
+    cluster: "seMetering",
+    type: ["attributeReport", "readResponse"],
+    convert: (model, msg, publish, options, meta) => {
+        if (utils.hasAlreadyProcessedMessage(msg, model)) return;
+        const payload: KeyValueAny = {};
+        const Divisor = (globalStore.getValue(msg.endpoint, "Divisor") ?? msg.endpoint.getClusterAttributeValue("seMetering", "divisor")) as number;
```
Koenkk, line 959: Should not use the global store for this, it's cleared when z2m is restarted, use this instead: https://github.com/Koenkk/zigbee-herdsman-converters/blob/f75db2baad5fe7b6d9fb26b4e4cce11a04a399af/src/lib/modernExtend.ts#L2357

src/devices/slacky_diy.ts
```diff
@@ -4463,6 +4909,26 @@ export const definitions: DefinitionWithExtend[] = [
                 valueMax: 65527,
                 description: "Group ID for scenes",
             }),
+            m.numeric<"genLevelCtrl">({
```
Koenkk, line 4912: I see it's duplicated multiple times with just different endpoint names, please create a function for it instead, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/f75db2baad5fe7b6d9fb26b4e4cce11a04a399af/src/devices/sonoff.ts#L375

src/devices/slacky_diy.ts
```diff
@@ -3937,6 +4264,28 @@ export const definitions: DefinitionWithExtend[] = [
                 commands: {},
                 commandsResponse: {},
             }),
+            m.deviceAddCustomCluster("genLevelCtrl", {
```
Koenkk, line 4267: Same applies to this, is duplicated multiple times

src/devices/ensystec.ts
```diff
@@ -0,0 +1,916 @@
...
+                reporting: {min: 0, max: 65000, change: 0},
+                access: "STATE_GET",
+                description: "Wire break control on input 6",
+            }),
+            m.binary<"EnsystecLeakProtect", EnsystecLeakSensor>({
+                name: "input_7_wire_break_control",
+                endpointName: "13",
+                cluster: "EnsystecLeakProtect",
+                attribute: "onLine",
+                valueOn: ["Online", 1],
```
Koenkk, line 553:
```suggestion
                valueOn: ["online", 1],
```

Same for all others and `offline`

src/converters/fromZigbee.ts
```diff
@@ -950,6 +950,62 @@ export const gas_metering: Fz.Converter<"seMetering", undefined, ["attributeRepo
...
+        if (msg.data.extendedStatus !== undefined) {
+            const value = msg.data.extendedStatus;
+            const property = utils.postfixWithEndpointName("extended_status", msg, model, meta);
+            payload[property] = value;
+        }
+
+        return payload;
+    },
+};
+export const water_metering_divisor: Fz.Converter<"seMetering", undefined, ["attributeReport", "readResponse"]> = {
```
Koenkk, line 991: This and the multiplier one shouldn't be needed after using `getClusterAttributeValue`

src/converters/fromZigbee.ts
```diff
@@ -950,6 +950,62 @@ export const gas_metering: Fz.Converter<"seMetering", undefined, ["attributeRepo
         return payload;
     },
 };
+export const water_metering: Fz.Converter<"seMetering", undefined, ["attributeReport", "readResponse"]> = {
```
Koenkk, line 953: I would keep this converter in the vendor file, it's not really generic (as it's not generic, abusing electrical cluster for water)

src/lib/modernExtend.ts
```diff
@@ -2293,6 +2284,73 @@ function genericMeter(args: MeterArgs = {}) {
...
+        const access = args.energyWritable ? ea.ALL : ea.STATE_GET;
+        if (args.power !== false)
+            exposes.push(e.numeric("volume_flow_rate", ea.STATE_GET).withUnit("m³/h").withDescription("Instantaneous water flow in m³/h"));
+        if (args.energy !== false)
+            exposes.push(
+                e.numeric("water", access).withUnit("m³").withDescription("Total water consumption in m³").withValueMin(0).withValueMax(99999999),
+            );
+        if (args.energyWritable) {
+            exposes.push(e.numeric("divisor", ea.ALL).withDescription("Divisor for water meter").withValueMin(1).withValueMax(16777215));
+            exposes.push(e.numeric("multiplier", ea.ALL).withDescription("Multiplier for water meter").withValueMin(1).withValueMax(16777215));
```
Koenkk, line 2297: I don't think this should be exposed

src/lib/modernExtend.ts
```diff
@@ -2293,6 +2284,73 @@ function genericMeter(args: MeterArgs = {}) {
             tz.metering_extended_status,
         ];
         delete configureLookup.haElectricalMeasurement;
+    } else if (args.cluster === "metering" && args.type === "water") {
```
Koenkk, line 2287: All this logic is quite specific for the device, I think it's better to move it in the vendor file.

slacky1965: ok, I'll return to the original version.

src/devices/slacky_diy.ts
slacky1965: Sorry, but I don't understand why you need to move `m.numeric` into a separate function. Just like all the following similar things. What will change? I want to keep my version. It makes this file easier to read and maintain.

slacky1965:
I'll redo everything and open a new PR

PR #12036, opened 2026-04-27
refactor(inovelli): clean up biome-ignore comments

rohankapoorcom:
## Summary

Cleans up the biome-ignore comments in `src/lib/inovelli.ts`. Removes 9 of the 40 ignores outright and rewrites the remaining 31 with informative justifications. No runtime behavior changes; all 491 tests (incl. 187 Inovelli tests) pass.

### Changes by category

- **`lint/performance/noAccumulatingSpread` (-2)** — Replaced the `Object.keys(msg.data).reduce((p, c) => ({...p, [key]: raw}), {})` accumulator in `fzLocal.inovelli` with a `for...of` loop mutating a single object. Same result, no quadratic allocation.

- **`lint/style/useNamingConvention` on `tzLocal.inovelli_parameters_readOnly` (-1)** — Renamed to `inovelli_parameters_read_only` (pure snake_case). Internal-only key; no MQTT or test impact.

- **`lint/style/useNamingConvention` on `Attribute.ID` (-1)** — Renamed the internal `Attribute` interface property `ID` -> `id` and updated all 122 attribute literal usages across `COMMON_ATTRIBUTES`, `VZM30_ATTRIBUTES`, etc. The unrelated `ID:` fields inside the Zcl `addCustomCluster` / `addCustomMMWaveCluster` blocks (required by zigbee-herdsman's `ClusterDefinition` type) are intentionally left untouched.

- **`lint/style/useNamingConvention` on `Model` enum members (-5)** — Converted `export const enum Model { VZM30 = ..., ... }` to `export const Model = { VZM30: ..., ... } as const` plus `export type Model = (typeof Model)[keyof typeof Model]`. Key names stay the same (`VZM30`...`VZM36`), so all 12 call sites work unchanged. Passes the existing `biome.json` `objectLiteralProperty` rule (CONSTANT_CASE allowed) without any config change.

- **`lint/suspicious/noExplicitAny` (kept, comment improved)** — The two `Fz.Converter<any, any, any>[]` declarations cannot be narrowed without losing type fidelity (this matches the same pattern in `src/lib/types.ts` `DefinitionWithExtend.fromZigbee` and `src/devices/shelly.ts`). Replaced the placeholder `generic` justification with `matches DefinitionWithExtend.fromZigbee signature`.

^ I'm not convinced this is true, but I couldn't find a way around it. Curious to see thoughts from @Koenkk or @Nerivec here.

- **`lint/style/useNamingConvention` on `VZM36_ATTRIBUTES` keys (kept, comment rewritten)** — All 29 ignores on keys like `dimmingSpeedUpRemote_1` and `outputMode_2` are kept because the `_N` endpoint suffix is the user-facing MQTT attribute-naming convention parsed by `resolveEndpointAndKey` and exercised by the tests (e.g. `findTzConverter(definition, "dimmingSpeedUpRemote_2")`). Renaming would break both the MQTT contract and the existing tests. The auto-generated `ignored using --suppress` text is replaced with `endpoint-suffixed MQTT attribute key; changing it would break existing MQTT naming conventions`.

## Test plan

- [x] `pnpm run check`
- [x] `pnpm run build`
- [x] `pnpm test` (491/491 passing, including all 187 Inovelli tests)

### DIFF
--- a/src/lib/inovelli.ts
+++ b/src/lib/inovelli.ts
@@ -625,7 +625,7 @@ const inovelliExtend = {
         supportsButtonTaps: boolean;
         splitValuesByEndpoint?: boolean;
     }) => {
-        // biome-ignore lint/suspicious/noExplicitAny: generic
+        // biome-ignore lint/suspicious/noExplicitAny: matches DefinitionWithExtend.fromZigbee signature
         const fromZigbee: Fz.Converter<any, any, any>[] = [];
         const toZigbee: Tz.Converter[] = [];
         const staticExposes: Expose[] = [];
@@ -643,7 +643,7 @@ const inovelliExtend = {
             fromZigbee.push(fzLocal.inovelli(attr.attributes, attr.clusterName, splitValuesByEndpoint, model));
             toZigbee.push(
                 tzLocal.inovelli_parameters(attr.attributes, attr.clusterName, model),
-                tzLocal.inovelli_parameters_readOnly(attr.attributes, attr.clusterName),
+                tzLocal.inovelli_parameters_read_only(attr.attributes, attr.clusterName),
             );
         }
 
@@ -719,7 +719,7 @@ const inovelliExtend = {
         } as ModernExtend;
     },
     light: ({splitValuesByEndpoint = false}: {splitValuesByEndpoint?: boolean} = {}) => {
-        // biome-ignore lint/suspicious/noExplicitAny: generic
+        // biome-ignore lint/suspicious/noExplicitAny: matches DefinitionWithExtend.fromZigbee signature
         const fromZigbee: Fz.Converter<any, any, any>[] = [fzLocal.on_off_for_endpoint(1, "state"), fzLocal.brightness];
 
         if (!splitValuesByEndpoint) {
@@ -828,18 +828,14 @@ const LED_NOTIFICATION_TYPES: {[key: number]: string} = {
 
 type FirmwareVersion = [major: number, minor: number];
 
-export const enum Model {
-    // biome-ignore lint/style/useNamingConvention: matches device model ID
-    VZM30 = "VZM30-SN",
-    // biome-ignore lint/style/useNamingConvention: matches device model ID
-    VZM31 = "VZM31-SN",
-    // biome-ignore lint/style/useNamingConvention: matches device model ID
-    VZM32 = "VZM32-SN",
-    // biome-ignore lint/style/useNamingConvention: matches device model ID
-    VZM35 = "VZM35-SN",
-    // biome-ignore lint/style/useNamingConvention: matches device model ID
-    VZM36 = "VZM36",
-}
+export const Model = {
+    VZM30: "VZM30-SN",
+    VZM31: "VZM31-SN",
+    VZM32: "VZM32-SN",
+    VZM35: "VZM35-SN",
+    VZM36: "VZM36",
+} as const;
+export type Model = (typeof Model)[keyof typeof Model];
 
 interface ModelValuesEntry {
     values?: {[s: string]: number};
@@ -847,8 +843,7 @@ interface ModelValuesEntry {
 }
 
 interface Attribute {
-    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
-    ID: number;
+    id: number;
     dataType: number;
     min?: number;
     max?: number;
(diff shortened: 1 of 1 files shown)


src/lib/inovelli.ts
```diff
@@ -625,7 +625,7 @@ const inovelliExtend = {
         supportsButtonTaps: boolean;
         splitValuesByEndpoint?: boolean;
     }) => {
-        // biome-ignore lint/suspicious/noExplicitAny: generic
+        // biome-ignore lint/suspicious/noExplicitAny: matches DefinitionWithExtend.fromZigbee signature
```
rohankapoorcom: I'm not convinced that this is actually reasonable. I would accept that we'd like to be able to type this better, but I couldn't figure out a good way to do so.

src/lib/inovelli.ts
```diff
@@ -719,7 +719,7 @@ const inovelliExtend = {
         } as ModernExtend;
     },
     light: ({splitValuesByEndpoint = false}: {splitValuesByEndpoint?: boolean} = {}) => {
-        // biome-ignore lint/suspicious/noExplicitAny: generic
+        // biome-ignore lint/suspicious/noExplicitAny: matches DefinitionWithExtend.fromZigbee signature
```
rohankapoorcom: As above.

src/lib/inovelli.ts
```diff
@@ -828,27 +828,22 @@ const LED_NOTIFICATION_TYPES: {[key: number]: string} = {
...
 
 interface ModelValuesEntry {
     values?: {[s: string]: number};
     conditionalValues?: {[valueName: string]: {addedInFirmware?: FirmwareVersion; removedInFirmware?: FirmwareVersion}};
 }
 
 interface Attribute {
-    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
-    ID: number;
+    id: number;
```
Nerivec, line 842: I assume this is either not used for custom clusters, or remapped?
ZH still uses uppercase `ID` for cluster type (too big a breaking change for now).
https://github.com/Koenkk/zigbee-herdsman/blob/93f702110561225d6d97fce5cdb1c749077319a7/src/zspec/zcl/definition/tstype.ts#L192
_There's a lot of dynamic generation in inovelli code, so, this makes it far more likely that TS would not catch such an issue at compile time._

_Alternatively, you could use the `Attribute` type from ZH instead (although will require adding `name` prop in payloads, or use `Omit<..>`). Would avoid re-declaring a type._

src/lib/inovelli.ts
```diff
@@ -719,7 +719,7 @@ const inovelliExtend = {
         } as ModernExtend;
```
Nerivec: Sidenote: this kind of type-casting can be dangerous. Usually best to let it derive properly based on the type of the root object, declare an explicit variable with the type and return it, or can use `satisfies ...` ([for compat check](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator)).

src/lib/inovelli.ts
Nerivec: This is generic fz. You could technically union-type it with every single entry, but it's unnecessary here (not using the specific types, just returning to match generic ME type).

rohankapoorcom: Sorry, point of clarification:

Are you saying we should leave it as is with the `Fz.Converter<any, any, any>[] = [];`? Or should it be changed to something like `NonNullable<ModernExtend["fromZigbee"]>` like my AI agent is suggesting. This is too far in the weeds for me with Typescript.

src/lib/inovelli.ts
rohankapoorcom: Changing to lowercase doesn't break anything here since this is a custom interface that we use for a lot of dynamic processing in the Inovelli code. Cleaning that up is going to be out of scope for this ticket.

src/lib/inovelli.ts
rohankapoorcom: Thanks, that's a good catch and fixed.

src/lib/inovelli.ts
Nerivec: Either should be fine (same resulting type).
Second would avoid the ignore comment though, so I guess cleaner.

src/lib/inovelli.ts
Nerivec: From what I can see, there's no type on the container object, so, `satisfies` ("complies") won't have the same result as `as` ("type-asserts"). e.g. `inovelliExtend.device` won't have type `ModernExtend`, it will be the exact derived const type of the return (which complies, but isn't of type).
Instead of `satisfies ModernExtend`, probably should type the return of the functions explicitly with `: ModernExtend`, less likely to run into troubles.
Example:
```ts
    mmWave: (): ModernExtend => {
```

rohankapoorcom: That's an interesting subtlety that I had missed with typescript. Thanks for the explanation. Corrected.

Koenkk:
Thanks!

PR #12046, opened 2026-04-27
For thermostats with modernExtend - add dontMapPIHeatingDemand as an option to piHeatingDemand…

svhelge:
Maybe rename the option to something more understandable?   `piHeatingDemandRawInPercentage` in order to indicate that the raw values are already in the 0-100 range?

Example usage:
```
            m.thermostat({
                systemMode: {values: ["off", "heat"]},
                piHeatingDemand: {values: true, dontMapPIHeatingDemand: true},
            }),
```


Koenkk:
Thanks!

> Maybe rename the option to something more understandable? 

You mean the pi heating demand?

svhelge:
> Thanks!
> 
> > Maybe rename the option to something more understandable?
> 
> You mean the pi heating demand?

I meant `dontMapPIHeatingDemand`. For me it wasn't obvious that `dontMapPIHeatingDemand` means that the device is reporting piHeatingDemand in percentage (0 to 100) and not the default 0 to 255. But I'm not sure if it helps to call it something else (like `piHeatingDemandRawInPercentage`). I guess it's easier to stick with the naming we have.

PR #12055, opened 2026-04-29
Add AOYAN AY-302Z soil moisture sensor

zyjsmile857:

Tested with Zigbee2MQTT.

Device info:
- Zigbee model: AY-302Z
- Manufacturer name: AOYAN  
- Uses the same Tuya datapoints as CS-201Z except humidity and humidity_calibration are not exposed.


zyjsmile857:
Add AOYAN AY-302Z soil moisture sensor

Koenkk:
> Uses the same Tuya datapoints as CS-201Z except humidity and humidity_calibration are not exposed.

We can re-use the definition by making the `expose` a function, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/b78e6eab5c9191c329fec37ee03bd49aa82a2262/src/devices/tuya.ts#L1775

zyjsmile857 (commented):
Updated as requested, re-used the exposes by moving them into a function and using it for AY-302Z without humidity and humidity_calibration.

Koenkk:
I meant moving them into the same definition, so there is 1 definition with 1 dynamic `expose` function.

PR #12060, opened 2026-04-29
Datek Idlock: Add custom attributes to clusters and move Datek specific converters to "local" converters.

svhelge:


Koenkk:
Thanks!

PR #12062, opened 2026-04-29
Support `toggle` command with native Hue control

burmistrzak:
Addresses shortcomings mentioned in https://github.com/Koenkk/zigbee-herdsman-converters/pull/12008#issuecomment-4340511184 by copying the default Z2M behavior.


burmistrzak:
@Mihonarium Feel free to take a look. ✌️

burmistrzak:
@Koenkk Should I check if the incoming state is in `["toggle", "on", "off"]` as well?

Koenkk:
I think that's good to check indeed.

burmistrzak:
@Koenkk Should be good to go.

Just as a general note, _native Hue control_ seems to fail silently when input validation fails.
Default `toZigbee` converters will throw an error. @Mihonarium Is this intended behavior?

Koenkk:
> seems to fail silently 

This shouldn't be the case indeed, it's better to throw errors, can you address it in a follow-up PR?

Koenkk:
Thanks!

Mihonarium:
Thanks!

PR #12065, opened 2026-04-30
fix(sonoff): SONOFF S60ZBTPF and S60ZBTPG prevent threshold protections from being enabled on first pairing

CubeZ2mDeveloper:
This PR fixes the default threshold protection configuration for the affected SONOFF device definition.

Previously, when the device was added for the first time in Zigbee2MQTT 2.9.2, the minimum threshold protections could be enabled by default and set to very low values:

- Min Power: 0.1W
- Min Current: 0.1A

As a result, the device could automatically turn off when no load was detected, which led to the issue reported in:
https://github.com/Koenkk/zigbee2mqtt/issues/31604

This behavior was caused by the ZHC device definition, not by the firmware itself. The firmware that was previously removed in:
https://github.com/Koenkk/zigbee-OTA/pull/1104

was therefore not the root cause of the issue.

This PR corrects the device definition to avoid enabling these threshold protections unintentionally during the initial device configuration.

For users who are still affected after upgrading, please update to Zigbee2MQTT 2.9.3 or later once available, remove the device, and pair it again.

If the issue still persists, users can manually reset the threshold configuration by:

1. Enabling all threshold protections.
2. Setting the thresholds to their respective Min/Max values:
   - Min Power: 0.1W, Max Power: 3250W
   - Min Voltage: AC165V, Max Voltage: AC277V
   - Min Current: 0.1A, Max Current: 14A
3. Saving the configuration.
4. **Disabling all threshold protections again and saving once more**.

We apologize for the inconvenience and confusion caused by this issue.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -18,7 +18,7 @@ import {
     YEAR_2000_IN_UTC,
 } from "../lib/sonoff";
 import * as tuya from "../lib/tuya";
-import type {DefinitionWithExtend, Expose, Fz, KeyValue, KeyValueAny, ModernExtend, OnEvent, Tz, Zh} from "../lib/types";
+import type {DefinitionWithExtend, Expose, Fz, KeyValue, KeyValueAny, ModernExtend, OnEvent, Tz} from "../lib/types";
 import * as utils from "../lib/utils";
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12073, opened 2026-04-30
update model to depict correct model names

loocd:
fix wrong description. should use actual model name


loocd:
did I understand this correctly that the description change is automatically carried over to zigbee2mqtt.io [according its readme](https://github.com/Koenkk/zigbee2mqtt.io/blob/master/README.md) and I would only need to make changes there in case I want to add/edit the "Notes" section?

src/devices/smarli.ts
```diff
@@ -9,14 +9,14 @@ export const definitions: DefinitionWithExtend[] = [
         zigbeeModel: ["S-ZB-PDM1-R251"],
         model: "S-ZB-PDM1-R251",
         vendor: "smarli.",
-        description: "Phase dimmer gen 1",
+        description: "smarli. phase dimmer",
```
Koenkk: Please don't include the vendor name in the description
```suggestion
        description: "Phase dimmer",
```

Koenkk:
Thanks!

PR #12075, opened 2026-04-30
Simplify the runningState support for EKO07259 and WDE002497

svhelge:
Use `m.thermostat` instead of custom function injecting the `runningState` support.


Koenkk:
Thanks!

PR #12089, opened 2026-05-02
Add Tuya device configurations for various switches

ohadvano:
Added multiple Tuya device configurations including 1 gang, 2 gang, 3 gang switches, and multifunction switches with various features such as backlight, child lock, and power monitoring.


andrei-lazarov (changes requested):
1. Please use modernExtend instead of instead of fromZigbee, toZigbee, configure:
```ts
extend: [tuya.modernExtend.tuyaBase({dp: true})],
```


2. Please check if there are similar devices already added, so we can merge the definitions. (Try to keep the code size small, re-use code)

3. Re-use the value converters. E.g. you have this twice. Add to `lib/tuya.ts` so you can make it `tuya.valueConverter.direction` (or whatever name makes sense)
```ts
 tuya.valueConverterBasic.lookup({
      FORWARD: tuya.enum(0),
      BACKWARD: tuya.enum(1),
  }),
```

4. Re-use the exposes. E.g. you have this twice.
```ts
e.enum("motor_steering", ea.STATE_SET, ["FORWARD", "BACKWARD"]).withDescription("Motor Steering"),
```
Add to lib and make it shorter like:
```ts
te=tuya.exposes;
...
te.direction()
```

ohadvano:
@andrei-lazarov I addressed 1 and 2
regarding 3, 4 - note that I followed existing pattern in the file (i.e., it's same for many other devices). I don't mind changing it, but I think that the right thing to do is refactor the usage of direction (forward/backwards) for all devices at once and not create ambiguity.
lmk what you think

andrei-lazarov:
> refactor the usage of direction (forward/backwards) for all devices

Yes, that would be great. Add it to lib and replace all occurences of: 

```ts
 tuya.valueConverterBasic.lookup({
      FORWARD: tuya.enum(0),
      BACKWARD: tuya.enum(1),
  }),
```

and `e.enum("motor_steering", ea.STATE_SET, ["FORWARD", "BACKWARD"]).withDescription("Motor Steering"),`

The ones that are different (`forward, back`) you have to leave unchanged, otherwise it would be a breaking change.

Also replace your `e.binary("child_lock", ea.STATE_SET, "ON", "OFF").withDescription("Child Lock"),`
with `e.child_lock()`

Replace `e.binary("backlight_mode", ea.STATE_SET, "ON", "OFF").withDescription("Backlight"),`
with `te.backlightModeOffOn()` (`te=tuya.exposes`)

Check all exposes of your devices, if they already exist in `exposes` or `tuya.exposes`.

Koenkk:
Thanks!

PR #12090, opened 2026-05-02
Fix native Philips transition time 

Mihonarium:
I checked upon seeing Koenkk/zigbee2mqtt#31838; transition time is off by a factor of 10. The bitfrost spec doesn’t give any details, so the assumption was that fade speed is in seconds, but apparently it’s not. (One example shows implicit conversion 4 <> 400ms, and in my testing it is indeed off by a factor of 10.)

Hopefully this fixes it.

### DIFF
--- a/src/lib/philips.ts
+++ b/src/lib/philips.ts
@@ -144,7 +144,7 @@ export const manuSpecificPhilips2Fz: Fz.Converter<"manuSpecificPhilips2", ManuSp
                 retval["color_mode"] = "xy";
             }
             if (decoded.fadeSpeed !== undefined) {
-                retval["transition"] = decoded.fadeSpeed;
+                retval["transition"] = decoded.fadeSpeed / 10;
             }
             if (decoded.effectType !== undefined) {
                 retval["effect"] = effectNames[decoded.effectType] ?? `unknown_0x${decoded.effectType.toString(16)}`;
@@ -388,7 +388,7 @@ const philipsModernExtend = {
                 // Map transition time to Philips2 fadeSpeed
                 // Bifrost spec: 0 = instant, practical range ~2..8, >0x100 = very slow
                 if (message.transition != null) {
-                    data.fadeSpeed = Math.round(Number(message.transition));
+                    data.fadeSpeed = Math.round(Number(message.transition) * 10);
                 }
 
                 // Effect speed: 0.0 = slowest, 1.0 = fastest (maps to 0..255 byte)


Mihonarium:
(Would appreciate it if people tested with various bulbs and verified that it is indeed a factor of 10.)

burmistrzak:
@Drpepper1988 Would you be able to test this PR?

Drpepper1988:
@burmistrzak

I'm happy to help test but I am very new to GitHub and all that if you can point me to some instructions on how to test I can do that and report back

burmistrzak:
@Drpepper1988 No worries! While _it is possible_ to do testing of PRs on HAOS, it's not that straightforward.

Just tell me a bit about your automation and I'll try to do the testing on my end. 😊

Drpepper1988:
@burmistrzak 
I have a lot of automations set using transition time. 
It's essentially when X happens set scene Y and it includes the transition time check box. See screenshot below 
Screenshot 2026-05-03 at 08 16 12
I tested it out with the maximum transition time, and the hypothesis about it being a factor of seems to hold true.

burmistrzak:
@Drpepper1988 Ok, can you confirm how long the transition actually takes with these settings?

Drpepper1988:
@burmistrzak Yes I can using this automation below, 100 seconds in the transition box equals 10 seconds fade time. 
Screenshot 2026-05-03 at 08 48 29

burmistrzak:
@Mihonarium Would you mind creating an additional PR to add strict input validation and error handling for Hue native control?
Because we're essentially replacing standard converters, we need to cover these things as well. 😅

Mihonarium:
I think it is a bad philosophy to have the same code replicated in multiple places. Ideally something else happens to have the logic somewhere else, available to use prior to sending zigbee messages.

See also a related discussion in the original pr.

Also tbh I got the thing I wanted — effects finally work! — and I likely won’t be able to dedicate much time for improving this stuff further, sorry. Already spent way too many hours on this!

burmistrzak:
> I think it is a bad philosophy to have the same code replicated in multiple places. Ideally something else happens to have the logic somewhere else, available to use prior to sending zigbee messages.
>
> See also a related discussion in the original pr.

@Mihonarium Yeah, see your point. However, input validation is usually only of concern when dealing with custom cluster attributes.

> Also tbh I got the thing I wanted — effects finally work! — and I likely won’t be able to dedicate much time for improving this stuff further, sorry. Already spent way too many hours on this!

That's totally fine. We appreciate your contribution! 🫶

burmistrzak:
@Drpepper1988 Alright, can confirm on my end. PR _should_ work as advertised. ☺️

Koenkk:
Thanks!

PR #12091, opened 2026-05-03
Tuya TLC2206: fix power source

andrei-lazarov:
`forcePowerSource` was added to this device without bumping the version (idk if version existed 1 year ago) - so devices already paired didn't receive the fix until manual reconfigure

- https://github.com/Koenkk/zigbee2mqtt/issues/31819


Koenkk:
Thanks!

PR #12093, opened 2026-05-03
fix: Novato ZIS-01P: correct Tuya datapoint indices

bilgi-source:
## What

Fixes the Novato ZIS-01P (PIR + mmWave radar) presence sensor converter so its settings actually work.

The previously merged converter mapped Tuya datapoints off-by-one. After a factory reset on `_TZE284_who1jxwd`, the device reports its settings on DPs 102–106 (not 101–105). With the wrong DPs, sensitivity / distance / delay writes silently failed.

## Changes

Datapoint corrections (verified via factory reset):

| Setting | Old DP | Correct DP |
|---|---|---|
| `presence_distance` | 101 | **102** |
| `presence_sensitivity` | 102 | **103** |
| `radar_switch` | 103 | **104** |
| `pir_sensitivity` | 104 | **105** |
| `delay_time` | 105 | **106** |

Other fixes:
- `delay_time` max raised from 600 s to 9600 s (matches device's actual range; factory default is 30 s).
- `presence_distance` unit `"m"` removed — it is a 1–3 step selector, not a meter value. Description clarified to show the approximate distances per step (1 ≈ 2 m, 2 ≈ 4 m, 3 ≈ 6 m).

New advanced calibration exposes (under `config` category):
- `radar_threshold` – DP 164, range 5–255, factory default 15 (manufacturer-recommended range: 10–15).
- `pir_threshold` – DP 160, range 1–250, factory default 20.

These are useful when the sensor is in an environment where the default sensitivity steps aren’t enough to suppress false alarms.

## Tested

Factory-reset on `_TZE284_who1jxwd` and observed reported DP values; verified that writing each setting now takes effect.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -25215,7 +25215,7 @@ export const definitions: DefinitionWithExtend[] = [
             e
                 .numeric("radar_threshold", ea.STATE_SET)
                 .withCategory("config")
-                .withDescription("Radar detection threshold, advanced calibration (factory default: 20)")
+                .withDescription("Radar detection threshold, advanced calibration (factory default: 15; manufacturer-recommended range: 10-15)")
                 .withValueMin(5)
                 .withValueMax(255)
                 .withValueStep(1),
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -25207,7 +25207,7 @@ export const definitions: DefinitionWithExtend[] = [
             e
                 .numeric("delay_time", ea.STATE_SET)
                 .withUnit("s")
-                .withDescription("Time delay before reporting no presence")
+                .withDescription("Time delay before reporting no presence (factory default: 30 s)")
                 .withValueMin(10)
                 .withValueMax(9600)
                 .withValueStep(1),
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
This converter applies to two devices: `_TZE284_vceqncho, _TZE284_who1jxwd`

You only tested the second one, correct?

In that case, don't merge yet - I will try to verify the datapoints for both

bilgi-source:
> This converter applies to two devices: `_TZE284_vceqncho, _TZE284_who1jxwd`
> 
> You only tested the second one, correct?
> 
> In that case, don't merge yet - I will try to verify the datapoints for both

Only tested second one, but both same as far as i know.

Koenkk:
@bilgi-source @tolgacanatan can you confirm the settings in the OP are currently **not** working?

bilgi-source:
> @bilgi-source @tolgacanatan can you confirm the settings in the OP are currently **not** working?

https://github.com/Koenkk/zigbee-herdsman-converters/pull/11355#issuecomment-3839771946

As I mentioned before, although I provided the correct data point, it appeared differently in the release. After raising that comment, I thought it had been fixed, but it seems the issue still persists. So I’ve opened a new PR.

Koenkk:
Thanks!

andrei-lazarov:
I checked and it's good now.
They have the same datapoints. There are a few more, but probably extra:
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/12103

Issue #11965, opened 2026-04-17
[BUG] EMIZB-132: power and energy not updating after #11717 (stuck on seMetering value)


skefre:
## Description
After upgrading to Z2M 2.9.2 (zigbee-herdsman-converters v26.20.0+), 
`power` and `energy` are no longer updated regularly for EMIZB-132. 
The values are stuck on the last value received from the `seMetering` 
cluster, while `current` and `voltage` from `haElectricalMeasurement` 
update every ~5 seconds as expected.

## Evidence
`current` reads 1.85A at 240V (~440W expected) while `power` is stuck 
at 1712W from a previous reading. This proves `power` is no longer 
being read from `totalActivePower` in `haElectricalMeasurement`.

## Setup
- Device: Develco EMIZB-132
- Electricity meter: Kaifa MA304H3E
- interface_mode: kaifa_and_kamstrup
- Z2M version: 2.9.2-dev (Edge addon)

## Expected behavior
`power` should update every ~5 seconds from `totalActivePower` 
in cluster `haElectricalMeasurement`, as it did before #11717.

## Related PRs
- #11717 (introduced regression)
- #11890 (partial fix)
- #11904 (did not resolve this issue)

## Log evidence
`current` varies between updates confirming the device is actively 
reporting from `haElectricalMeasurement`, but `power` and `energy` 
never change. At 1.85A × 240V the expected power is ~440W, not 1712W.

```log
[2026-04-17 22:33:43] info: z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/strommaleren', payload '{"current":1.84,"current_phase_b":1.02,"current_phase_c":1.25,"energy":52875.13,"linkquality":123,"power":1712,...}'
[2026-04-17 22:33:45] info: z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/strommaleren', payload '{"current":1.85,"current_phase_b":1.02,"current_phase_c":1.25,"energy":52875.13,"linkquality":131,"power":1712,...}'
[2026-04-17 22:33:48] info: z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/strommaleren', payload '{"current":1.85,"current_phase_b":1.02,"current_phase_c":1.25,"energy":52875.13,"linkquality":131,"power":1712,...}'
[...]

PR #12094, opened 2026-05-03
Fix missing 'producedEnergy' for EMIZB-132

svhelge:


Koenkk:
Thanks!

bdybdal:
I think produced energy got same problem as thise https://github.com/Koenkk/zigbee-herdsman-converters/issues/11965

It wont update, if i push the refresh symbol inn settings the value is jumping up

svhelge:
Could you have a look at the reporting tab. Which values do you have? 
image

bdybdal:
Skjermbilde 2026-05-11 202630


The currentSummReceived wasn't in the reporting list, maybe therfor . The sun is down now , so no production. I think it will work

PR #12095, opened 2026-05-03
fix(k4003c): Revert broken state reporting

murarustefaan:
Fixes state reporting for Legrand / BTcino K4003C switches introduced by mistake in https://github.com/Koenkk/zigbee-herdsman-converters/pull/11957


Koenkk:
thanks!

PR #12059, opened 2026-04-29
Shelly: Expose switch inputs and input mode configuration for Gen4 1/1PM/1PM Mini/2PM

SebiGo:
The external converter for the Shelly 1 Mini Gen4 has been tested by @Chicchi7393 in #12011 .

I assume the 1PM and 1PM Mini expose the same endpoint and behave identically, but this is untested as I haven't received the 2.0.0-beta1 firmware on my devices yet.

I also included an RPC-based converter for configuring non-standard switch input modes (follow, flip, detached, cycle, activation) as well as ZCL-based switch_type configuration (toggle/momentary via genOnOffSwitchCfg). Both apply to the 1 Mini, 1PM Mini, 1PM Gen4, and 2PM Gen4.

PR #12096, opened 2026-05-03
fix(shelly): support Shelly 1 Gen 4 and switch_mode, switch_type for Gen4

SebiGo:
Extends Shelly Gen4 switch input support for firmware 2.0.0-beta1:

- 1 Gen4 / 1PM Gen4: add switch input support (action events, switch_type, switch_mode) — previously missing
- switch_mode (RPC): new expose for all Gen4 devices (follow/flip/detached/cycle/activation), read on configure and via get
- Fingerprints: replace zigbeeModel with endpoint-aware fingerprints so beta firmware is recognized correctly alongside older firmware

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -3,7 +3,7 @@ import * as fz from "../converters/fromZigbee";
 import * as exposes from "../lib/exposes";
 import {logger} from "../lib/logger";
 import * as m from "../lib/modernExtend";
-import type {Configure, DefinitionWithExtend, Expose, Fz, KeyValue, ModernExtend, Tz, Zh} from "../lib/types";
+import type {Configure, DefinitionExposesFunction, DefinitionWithExtend, DummyDevice, Expose, Fz, KeyValue, ModernExtend, Tz, Zh} from "../lib/types";
 import * as utils from "../lib/utils";
 import {assertObject, determineEndpoint, sleep} from "../lib/utils";
 
@@ -451,7 +451,7 @@ const shellyModernExtend = {
             return e.binary(`switch_${id}`, ea.STATE_SET, "momentary", "detached").withLabel(`Endpoint: ${id + 1}`);
         };
 
-        const exposes: Expose[] = [];
+        const exposes: (Expose | DefinitionExposesFunction)[] = [];
         const exposesDev: Expose[] = [
             e
                 .text("rpc_tx", ea.STATE_SET)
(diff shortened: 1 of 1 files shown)


SebiGo:
Shelly 1 Gen4 tested by @KroFR and @Cemaver in #12059 .
For now, 'switch_mode' , and "switch_type" are not working except for reading switch_mode. My assumption is a partial implementation in firmware 2.0.0-beta1. As a workaround for now, use the web ui. I assume that these values usually get set once.
Maybe with a stable firmware release this will be possible with or without a new PR.

Koenkk:
> Fingerprints: replace zigbeeModel with endpoint-aware fingerprints so beta firmware is recognized correctly alongside older firmware

Instead of creating separate definitions, to reduce code duplication, it's better to use a dynamic expose function, in this way newer functions can be exposed for the newer firmwares. Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/24dcb174ca0c1b35f9dd7c25168d000da3f4ef6c/src/devices/tuya.ts#L1775

SebiGo:
Thanks for the feedback! Updated — switch_mode is now exposed via a DefinitionExposesFunction inside shellyRPCSetup that checks for the presence of the RPC endpoint (EP239), so it only appears for devices running firmware 2.0.0-beta1 or newer.

src/devices/shelly.ts
```diff
@@ -1129,7 +1180,32 @@ const tzLocal = {
 
 export const definitions: DefinitionWithExtend[] = [
     {
-        zigbeeModel: ["Mini1", "1 Mini"],
+        fingerprint: [
```
Koenkk: I think all the `fingerprint` are not needed anymore now?

SebiGo: Thanks, I removed the old modelID fingerprints. (My intention was too much on the safe side.)

src/devices/shelly.ts
```diff
@@ -1157,14 +1222,62 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        fingerprint: [{modelID: "1", manufacturerName: "Shelly"}],
+        fingerprint: [
```
Koenkk: I think these fingeprint can also stay on the simplified versions?

src/devices/shelly.ts
```diff
@@ -1157,14 +1222,62 @@ export const definitions: DefinitionWithExtend[] = [
...
+            if (ep) {
+                await ep.bind("genOnOff", coordinatorEndpoint);
+                await ep.bind("genScenes", coordinatorEndpoint);
+                await ep.read("genOnOffSwitchCfg", ["switchType"]);
+            }
+        },
     },
     {
-        zigbeeModel: ["Mini1PM", "1PM Mini"],
+        fingerprint: [
```
Koenkk: Same here

src/devices/shelly.ts
```diff
@@ -1194,17 +1307,47 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        zigbeeModel: ["1PM"],
+        fingerprint: [
```
Koenkk: Same here

SebiGo:
You're right — the extended fingerprints were a leftover from when I was thinking about firmware-version discrimination, but since the simplified entries match the same modelID anyway, the extended ones never actually trigger first. And the configure hooks already handle missing endpoints gracefully on older firmware. I went down this path because the existing 2PM definition already used extended fingerprints, so I followed the same pattern — but on closer look it has the same redundancy.
Simplified and force-pushed.

Koenkk:
Thanks!

Mosher23:
Hey,
just installed 2.10.1 and tried with both 1PM Gen4 and 1PM mini Gen4. (Both are on 2.0.0 Beta1)
Switch type (Endpoint: sw1) and Switch mode (Endpoint: sw1) do not report actual configuration of the devices. 
e.g. 1PM Gen4. Refreshing or reconfiguring the device bring no improvement.
Screenshot 2026-05-08 at 11 36 18


Action shows up as N/A
Screenshot 2026-05-08 at 11 35 18

PR #12101, opened 2026-05-04
fix: don't use manuSpecificPhilips2 path for mixed groups

Mihonarium:
With Hue native control enabled, sending commands to a group with some Hue bulbs went via manuSpecificPhilips2 regardless of group composition. For mixed groups (Hue and non-Hue bulbs), this can fail (e.g., if trying to set effects).

(See https://github.com/Koenkk/zigbee-herdsman-converters/pull/12008#issuecomment-4362609833, https://github.com/Koenkk/zigbee-herdsman-converters/pull/11655#issuecomment-4367752248, and nearby discussion.)

When `hue_native_control` is enabled, color/brightness/etc. commands go out as manuSpecificPhilips2 writes. For groups, this only works if every member supports the cluster.

Fix checks all group members and falls back to standard ZCL converters if any member doesn't support manuSpecificPhilips2.

(If you experience this issue, disable Hue native control for the groups, or create Hue-only groups to send commands to.)

### DIFF
--- a/src/lib/philips.ts
+++ b/src/lib/philips.ts
@@ -261,24 +261,31 @@ const philipsModernExtend = {
                 // Resolve control mode: explicit option wins; otherwise default to standard converters.
                 const nativeControl = (meta.options as KeyValueAny).hue_native_control === true;
 
-                // Check if device supports the manuSpecificPhilips2 cluster.
+                // Check if entity supports the manuSpecificPhilips2 cluster.
+                // For groups, all members must support it — otherwise non-Hue members
+                // would silently ignore the manufacturer-specific command.
                 // Wrapped in try-catch because supportsInputCluster may throw for
                 // custom clusters not in the cluster registry (e.g. in test mocks).
+                const endpointSupportsPhilips2 = (ep: Zh.Endpoint): boolean => {
+                    try {
+                        return ep.supportsInputCluster("manuSpecificPhilips2");
+                    } catch {
+                        return false;
+                    }
+                };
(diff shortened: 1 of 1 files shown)


Mihonarium:
This should probably fix the immediate issue, but idk if there's a way to send Hue-only effects to groups with both Hue bulbs and non-Hue entities? That would be preferable if possible. cc @Koenkk

ams2990:
> This should probably fix the immediate issue, but idk if there's a way to send Hue-only effects to groups with both Hue bulbs and non-Hue entities? That would be preferable if possible. cc @Koenkk

Yes, this is possible, and I do it today. I issue raw Hue API commands via MQTT to groups that contain a mix of Hue lights and Inovelli switches.  The switches just ignore the broadcasts they don't understand.

Koenkk:
@Mihonarium I expect that devices indeed ignore those commands.

ams2990:
Groups are definitely tricky, though, because you need to be able to send both Hue and non-Hue commands to the group. For instance, I send MoveToLevel when I want to control the switches and raw Hue messages when I want to control the lights. If we force always on or always off, we either lose the benefit of this excellent work or lose the controllability of non-Hue devices in the group (unless addressed individually).

Perhaps we could add something to the MQTT message API to allow for selectively enabling or disabling Hue command translation on a message-by-message basis?

deviantintegral:
Anyone have direction on what's next here? I'm not clear on if this PR is won't-merge as is, but if not what we should do instead.

Mihonarium:
👀

PR #12105, opened 2026-05-04
Add inching, indicator mode, fix dp in timer device

ohadvano:

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -26632,6 +26632,7 @@ export const definitions: DefinitionWithExtend[] = [
             e.switch().setAccess("state", ea.STATE_SET),
             e.power_on_behavior().withAccess(ea.STATE_SET),
             e.numeric("countdown", ea.STATE_SET).withUnit("s").withValueMin(0).withValueMax(86400).withDescription("Countdown to turn off"),
+            e.enum("indicator_mode", ea.STATE_SET, ["none", "relay", "pos"]).withDescription("Indicator status behavior"),
             e.binary("backlight_mode", ea.STATE_SET, "ON", "OFF").withDescription("Backlight"),
             e
                 .composite("inching", "inching", ea.STATE_SET)
@@ -26649,6 +26650,15 @@ export const definitions: DefinitionWithExtend[] = [
                 [1, "state", tuya.valueConverter.onOff],
                 [7, "countdown", tuya.valueConverter.countdown],
                 [14, "power_on_behavior", tuya.valueConverter.powerOnBehaviorEnum],
+                [
+                    15,
+                    "indicator_mode",
+                    tuya.valueConverterBasic.lookup({
+                        none: tuya.enum(0),
+                        relay: tuya.enum(1),
+                        pos: tuya.enum(2),
+                    }),
+                ],
                 [16, "backlight_mode", tuya.valueConverter.onOff],
                 [
                     19,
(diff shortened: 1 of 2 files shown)


src/devices/tuya.ts
```diff
@@ -26643,12 +26650,56 @@ export const definitions: DefinitionWithExtend[] = [
...
+                        relay: tuya.enum(1),
+                        pos: tuya.enum(2),
+                    }),
+                ],
                 [16, "backlight_mode", tuya.valueConverter.onOff],
+                [
+                    19,
+                    "inching",
+                    {
+                        from: (value) => {
```
andrei-lazarov: The logic for inching must be put in the lib, so other devices can re-use it.

And there already is `tuya.valueConverter.inchingSwitch` - can you check if that works?

src/devices/tuya.ts
```diff
@@ -26632,9 +26632,16 @@ export const definitions: DefinitionWithExtend[] = [
             e.switch().setAccess("state", ea.STATE_SET),
             e.power_on_behavior().withAccess(ea.STATE_SET),
             e.numeric("countdown", ea.STATE_SET).withUnit("s").withValueMin(0).withValueMax(86400).withDescription("Countdown to turn off"),
+            e.enum("indicator_mode", ea.STATE_SET, ["none", "relay", "pos"]).withDescription("Indicator status behavior"),
             e.binary("backlight_mode", ea.STATE_SET, "ON", "OFF").withDescription("Backlight"),
-            e.enum("off_color", ea.STATE_SET, ["red", "blue", "green", "white", "yellow", "magenta", "cyan"]).withDescription("OFF Color"),
+            e
```
andrei-lazarov: `te.inchingSwitch()` should work ?

andrei-lazarov (changes requested):

src/devices/tuya.ts
ohadvano: What I added is a different definition than the existing one. What I added behaves similar to how the Tuya app exposes inching, so no

src/devices/tuya.ts
ohadvano: no, see other comment

ohadvano:
thanks for the review, the definition I added is a separate one that matches how Tuya app exposes inching so current definition can't be reused

src/devices/tuya.ts
Koenkk: I think it's good to move this into `lib/tuya.ts` as `inchingSwitch2`

ohadvano: ok I adjusted

src/devices/tuya.ts
```diff
@@ -26632,9 +26632,11 @@ export const definitions: DefinitionWithExtend[] = [
             e.switch().setAccess("state", ea.STATE_SET),
             e.power_on_behavior().withAccess(ea.STATE_SET),
             e.numeric("countdown", ea.STATE_SET).withUnit("s").withValueMin(0).withValueMax(86400).withDescription("Countdown to turn off"),
+            e.enum("indicator_mode", ea.STATE_SET, ["none", "relay", "pos"]).withDescription("Indicator status behavior"),
```
andrei-lazarov:
```suggestion
            tuya.exposes.indicatorModeNoneRelayPos(),
```

ohadvano:
@Koenkk is this ready for merging?

Koenkk:
Thanks!

PR #12109, opened 2026-05-05
fix(danfoss): rename adaptionRunControl state none to idle (HA MQTT display issue)

storm1ng:
This fixes an issue where none would display as unknown on the MQTT device.
Before:
danfoss_z2m_adaption_run_control_before
danfoss_adaptation_run_control_before 

After:
danfoss_z2m_adaption_run_control_after
danfoss_adaptation_run_control_after


Koenkk:
Thanks!

PR #12115, opened 2026-05-05
Add min and max cooling/heating setpoint limits for Plugwise Emma pro 170-01

michelotten:
Add configurable min and max cooling/heating setpoint limits for Plugwise Emma Pro 170-01


Koenkk:
Thanks!

PR #11505, opened 2026-02-10
Update S60ZBTPF/TPG And Mini-ZBDIM

lcheng33775823:

PR #12116, opened 2026-05-05
feat(sonoff): add maxBrightnessThreshold

storm1ng:
This will add support for maxBrightnessThreshold based on #11505

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -6267,13 +6267,11 @@ export const definitions: DefinitionWithExtend[] = [
             }),
             m.numeric<"customClusterEwelink", SonoffEwelink>({
                 name: "level_for_calibration",
-                access: "ALL",
+                access: "STATE_GET",
                 cluster: "customClusterEwelink",
                 attribute: "levelForCalibration",
-                description:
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

CubeZ2mDeveloper:
Hi @storm1ng,

Thank you for this PR and for helping improve the SONOFF device support in Zigbee2MQTT.

We really appreciate your help.

storm1ng:
@CubeZ2mDeveloper 
Are you ok with the no-op when triggering the calibration action? There was a response which couldn't be parsed (looked like some control characters). I left clear action in the code, it was removed in the other PR, didn't quite understand why.

CubeZ2mDeveloper:
> @CubeZ2mDeveloper Are you ok with the no-op when triggering the calibration action? There was a response which couldn't be parsed (looked like some control characters). I left clear action in the code, it was removed in the other PR, didn't quite understand why.

Hi @storm1ng,

Thanks for the clarification.

The no-op looks OK to me. If the device returns an unparseable response after triggering calibration, ignoring it should be fine as long as the calibration itself is not affected.

Regarding the `clear` action, I think it is fine to remove it. Clearing the calibration does not provide much practical value, since users can simply run the calibration again to update the calibration result.

Thanks again for your work on this.

PR #12120, opened 2026-05-06

zyjsmile857:
Tested with Zigbee2MQTT.

Device info:

Zigbee model: AY-302Z
Manufacturer name: AOYAN
Uses the same Tuya datapoints as CS-201Z except humidity and humidity_calibration are not exposed.


zyjsmile857:
https://github.com/Koenkk/zigbee2mqtt.io/pull/5092

PR #12121, opened 2026-05-06

zyjsmile857:

Tested with Zigbee2MQTT.

Device info:

Zigbee model: AY-302Z
Manufacturer name: AOYAN
Uses the same Tuya datapoints as CS-201Z except humidity and humidity_calibration are not exposed.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -22356,26 +22356,33 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        zigbeeModel: ["ZG-303Z", "AY-303Z"],
+        zigbeeModel: ["ZG-303Z", "AY-303Z", "AY-302Z"],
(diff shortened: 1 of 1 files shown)


zyjsmile857 (commented):
Fix the error of "zh" reporting.

zyjsmile857:
https://github.com/Koenkk/zigbee2mqtt.io/pull/5092

Koenkk:
Thanks!

zyjsmile857:
I would like AY-302Z and AY-303Z to have their own device pages on zigbee2mqtt.io. What is the preferred way to model this? Should these be added as whiteLabel entries, or should they share the CS-201Z page?

Koenkk:
You can do so by creating a [whiteLabel entry like this](https://github.com/Koenkk/zigbee-herdsman-converters/blob/f0dcbfd65ea86bbe71923bd3d3d82bbc53dd7b0e/src/devices/philips.ts#L231), then after the next z2m release the page will be automatically generated.

PR #12123, opened 2026-05-06
Update candeo.ts

candeodevelopment:
Adds support for 2 new Candeo devices:

1) C-ZB-DM204V2 - a small form factor dimmer module 
2) C-ZB-SM30-2G - a small form factor 2 gang switch module

(We will create a PR for Z2M for the new device photos shortly)


Koenkk:
Thanks!

PR #12132, opened 2026-05-06
fix(ignore): Allow whiteLabels of whiteLabels

andrei-lazarov:
See docs PR for info ↓

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -5918,7 +5918,7 @@ export const definitions: DefinitionWithExtend[] = [
             {vendor: "Haozee", model: "ESW-OZAA-EU"},
             {vendor: "Moes", model: "ZT-SY-EU-G-4S-WH-MS"},
             {vendor: "Nedis", model: "ZBWS40WT"},
-            {vendor: "Nous", model: "C1"},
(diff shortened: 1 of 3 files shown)


src/devices/tuya.ts
```diff
@@ -5918,7 +5918,7 @@ export const definitions: DefinitionWithExtend[] = [
             {vendor: "Haozee", model: "ESW-OZAA-EU"},
             {vendor: "Moes", model: "ZT-SY-EU-G-4S-WH-MS"},
             {vendor: "Nedis", model: "ZBWS40WT"},
-            {vendor: "Nous", model: "C1"},
+            {vendor: "Nous", model: "C1", linkToModel: "TS0044_2"},
```
Koenkk: A test should be added whether the `linkToModel` exists (to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/test/index.test.ts)

Koenkk: Maybe we should name it `whiteLabelOf` instead of `linkToModel`?

andrei-lazarov:
Thanks for reviewing! I addressed the comments

Koenkk:
Thanks!

PR #12135, opened 2026-05-07
fix: add switch type for HS1RM and fix diagnostic issue

Leo2442926161:
Dear Z2M team,
fix: add switch type for HS1RM and fix diagnostic issue


Koenkk:
Thanks!

PR #12137, opened 2026-05-07
Add entityCategory to min/max heat/cool setpoint limits for Plugwise 170-01

michelotten:
Adding entityCategory: "config" for previously created min/max heat/cool setpoint limits for Plugwise 170-01


Koenkk:
Thanks!

PR #12141, opened 2026-05-07
docs(Legrand 412170): clarify device_mode description & warn of  factory default

alexhk:
This PR updates only the `device_mode` description text, clarifying that `switch` is the working mode and `auto` produces erratic behavior, without removing the `auto` value or auto-writing `switch` during configure, since either could regress installations we don't know about.

Essentially the 412170 does not actually have a 'auto' device mode, despite the firmware apparently accepting value 4 for 'auto' mode on cluster 0xfc01 without error.

Why not remove the auto value altogether? 

Because it has been exposed for 5+ years, and more importantly, being able to set device_mode back to switch is what makes the device usable in the first place. Whatever the firmware factory default actually is, it is not switch, and only after explicitly setting device_mode to switch does the device behave correctly with Legrand's documented wiring (tested on 2 new factory sealed devices).

Just to be clear, this only affects Legrand's L to C1 wiring method, Legrand's C1 to C2 "dry-contact" method still seems to work even in default device_mode.

Testing on a 412170 (firmware 104) shows:

- **`switch` mode:** momentary push-buttons wired L to C1 (per Legrand's instruction sheet) toggle the relay cleanly. Dry contact between C1/C2 also works. Zigbee on/off works.

- **`auto` mode:** erratic relay behavior with the L to C1 push-button wiring method. The C1 to C2 wiring method still works. Zigbee on/off works.

- The firmware default is **not** `switch`. New installations misbehave with L to C1 wiring until the user manually sets `device_mode` to `switch`. I experienced this on a fresh device, and a Wireshark capture of a second factory-sealed 412170 paired to the **official Legrand Gateway** confirms the gateway never reads or writes cluster `0xfc01` either, so the same trap exists on the official integration path.

- The `auto` value seems leftover firmware inherited when the 412170 entry was cloned from the 412171 contactor (which has an auto mode) in #2297 (Feb 2021).  **Legrand Home+Control app:** exposes an "Off-peak / peak time" toggle for the 412171, no equivalent for the 412170.
- **OnOff cluster behavior:** on the 412171, OnOff is disabled while in `auto`; on the 412170, OnOff keeps working in either mode - i.e. `auto` has no functional effect on relay control.
- It was always assumed that the 412171 ships in auto mode, but for the 412170 your guess is as good as mine. It's probably not 'auto', since the device does not have a auto mode, and it's not 'switch', as by default we have erratic behavior which can be eliminated by setting device_mode 'switch'.

Related: Koenkk/zigbee2mqtt#14812 (same `device_mode`/`auto` confusion on the 412171 contactor, description rewritten in #4951).

Companion docs PR: Koenkk/zigbee2mqtt.io#5102.

### DIFF
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -102,7 +102,7 @@ export const definitions: DefinitionWithExtend[] = [
             e.power().withAccess(ea.STATE_GET),
             e
                 .enum("device_mode", ea.ALL, ["switch", "auto"])
-                .withDescription("switch: allow on/off, auto will use wired action via C1/C2 on teleruptor with buttons"),
+                .withDescription("Factory default is 'auto' - set to 'switch' for predictable operation. switch: normal on/off mode, momentary push-buttons on C1/C2 toggle the relay cleanly. auto: produces erratic behavior (auto mode appears to be a leftover from reusing code from Legrand 412171)"),
         ],
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -102,7 +102,9 @@ export const definitions: DefinitionWithExtend[] = [
             e.power().withAccess(ea.STATE_GET),
             e
                 .enum("device_mode", ea.ALL, ["switch", "auto"])
-                .withDescription("Factory default is 'auto' - set to 'switch' for predictable operation. switch: normal on/off mode, momentary push-buttons on C1/C2 toggle the relay cleanly. auto: produces erratic behavior (auto mode appears to be a leftover from reusing code from Legrand 412171)"),
+                .withDescription(
+                    "Factory default is 'auto' - set to 'switch' for predictable operation. switch: normal on/off mode, momentary push-buttons on C1/C2 toggle the relay cleanly. auto: produces erratic behavior (auto mode appears to be a leftover from reusing code from Legrand 412171)",
+                ),
         ],
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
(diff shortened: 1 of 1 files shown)


alexhk:
Apologies for the dust, changed this PR to draft.

Instead of just describing erratic behavior with this PR, I decided to connect a second brand new unit to my Legrand Gateway to find out what device_mode the original Legrand integration actually sets when the devices is connected. This way we can fix the z2m pairing/configuration code for the Legrand 412170 properly and avoid the confusion of erratic behavior after adding at the first place...

Bear with me for a few days.

alexhk:
Updated PR description with final findings from bench + sniffer testing on two units (one paired to z2m, one to the official Legrand Gateway). Marking ready for review.

Turns out even with the Legrand Gateway we get the same erratic behavior on the L to C1 wiring method on factory default units - only z2m users can actually fix this (by setting `device_mode` to `switch`). Looks like a Legrand bug specific to this wiring method.

Koenkk:
Thanks!

alexhk:
Thanks for merging. 

Interestingly I unpacked a new 412171 today and looked at the installation sheet and I guess I found the culprit.

For future reference:

- The **412171** (20A contactor) is **meant to be used with sustained-state switches** (or dry contact) wired to C1/C2.
- The **412170** (16A latching relay) is **meant to be used with momentary push-buttons** wired to C1/C2.

So auto mode in the 412170's firmware isn't merely a leftover, it's expecting the wrong input shape for the wiring Legrand themselves document, which is probably why momentary buttons (method with L to C1, but also dry-contact buttons between C1 and C2 if held briefly) misbehave. 

The 412170 was never meant to be used with auto mode, but the same OTA firmware binary is used on both devices (manufacturerCode: 4129, imageType: 17 per [Koenkk/zigbee-OTA#365], which is probably why the device accepts the auto write without error. Looks like an entire family of Legrand "NLP" devices shares this firmware.

PR #12145, opened 2026-05-08
Add support for TS0726_4_gang_scene_switch fingerprint _TZ3002_tdf2m4ch

jasonj27:
It's a similar switch works with the same converter file except backlight and count down function.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -26889,5 +26889,5 @@ export const definitions: DefinitionWithExtend[] = [
                 await reporting.bind(device.getEndpoint(ep), coordinatorEndpoint, ["genOnOff"]);
             }
         },
-    }
+    },
 ];
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -21923,6 +21923,7 @@ export const definitions: DefinitionWithExtend[] = [
             "_TZ300A_vkflnsl0",
             "_TZ3002_eda6eitk",
             "_TZ3000_hurauima",
+            "_TZ3002_tdf2m4ch",
         ]),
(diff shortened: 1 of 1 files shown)


Koenkk:
> It's a similar switch works with the same converter file except backlight and count down function.

Then we can keep the current converter by making the exposes conditional, example https://github.com/Koenkk/zigbee-herdsman-converters/blob/24bb7d6131f26a2753484ea3b053274af2fe107b/src/devices/tuya.ts#L2629

jasonj27:
I am not sure how to do that because these functions is defined in extend not exposes.

```ts
        exposes: [e.action(["scene_1", "scene_2", "scene_3", "scene_4"])],
        extend: [
            tuya.modernExtend.tuyaBase(),
            tuya.modernExtend.tuyaOnOff({
                switchMode: true,
                powerOnBehavior2: true,
                indicatorModeNoneRelayPos: true,
                endpoints: ["l1", "l2", "l3", "l4"],
            }),
        ],
```

andrei-lazarov:
Like this, adjust the existing `TS0726_4_gang_scene_switch`
```ts
tuya.modernExtend.tuyaOnOff({
    switchMode: true,
    powerOnBehavior2: true,
    backlightModeOffOn: (m) => m !== "_TZ3002_tdf2m4ch",
    indicatorModeNoneRelayPos: true,
    onOffCountdown: (m) => m !== "_TZ3002_tdf2m4ch",
    endpoints: ["l1", "l2", "l3", "l4"],
}),
```

andrei-lazarov (approved):

Koenkk:
Thanks!

PR #12114, opened 2026-05-05
Modify temperature and humidity reporting settings for Plugwise Emma Pro 170-01

michelotten:
For Plugwise Emma Pro 170-01 I've updated temperature reporting change value from 0.1 to 10 as 0.1 is invalid. 
Humidity reporting is set to false as I got 'INVALID_VALUE' error messages during configuration of the device.

After these changes both the 'local temperature' and 'humidity' values are being reported almost instantly by the Emma, before these values hardly ever updated.

PR #12146, opened 2026-05-08
Fix humidity reporting for Plugwise 170-01

michelotten:
Made a mistake in https://github.com/Koenkk/zigbee-herdsman-converters/pull/12114.
This PR will fix the humidity reporting for Plugwise 170-01


Koenkk:
Thanks!

PR #11747, opened 2026-03-16
Move Danfoss custom attributes from ZH to ZHC

svhelge:

PR #12147, opened 2026-05-08
fix(danfoss): restore weekly_schedule toZigbee converter for Ally TRV

storm1ng:
- tz.thermostat_weekly_schedule was dropped when PR #11747 moved Danfoss custom attributes from zigbee-herdsman to zigbee-herdsman-converters, causing "No converter available for 'weekly_schedule'" errors.
- Add note for trigger_time limitation in Ally firmware


Koenkk:
Thanks!

PR #12151, opened 2026-05-08
fix(legrand): fix Legrand 067755 button mapping and configure

h2okopfmt:
## Problem

The Legrand 067755 wireless 4-scene remote was reporting all button presses as `recall_1_1`. The previous converter used `fz.command_recall` which cannot differentiate the 4 buttons, and the `configure` step was failing with `INVALID_EP` errors.

## Root cause

The 067755 has a single endpoint but uses **different Zigbee group IDs** per button in `genScenes commandRecall`:

| Group ID | Button |
|----------|--------|
| 0xFFED (65517) | button_1 |
| 0xFFEC (65516) | button_2 |
| 0xFFEB (65515) | button_3 |
| 0xFFEA (65514) | button_4 |

The device does **not** have `genOnOff` or `genLevelCtrl` clusters, so binding them in `configure` caused `INVALID_EP` errors.

## Changes

- Replace `fz.command_recall` with inline converter that maps group IDs to `button_1`..`button_4` (verified on physical device)
- Remove `multiEndpoint` from meta (device has only one endpoint)
- Fix `configure` to bind only `genPowerCfg` and `genPollCtrl`
- Update exposes to reflect actual button actions

## Testing

Verified on physical Legrand 067755 device — all 4 buttons correctly reported.

### DIFF
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -1,873 +1,805 @@
-  import * as fz from "../converters/fromZigbee";
-  import * as tz from "../converters/toZigbee";
-  import * as exposes from "../lib/exposes";
-  import {eLegrand, fzLegrand, legrandExtend, legrandOptions, readInitialBatteryState, tzLegrand} from
-  "../lib/legrand";
-  import * as m from "../lib/modernExtend";
-  import * as reporting from "../lib/reporting";
-  import * as tuya from "../lib/tuya";
-  import type {DefinitionWithExtend, Fz} from "../lib/types";
+import * as fz from "../converters/fromZigbee";
+import * as tz from "../converters/toZigbee";
+import * as exposes from "../lib/exposes";
+import {eLegrand, fzLegrand, legrandExtend, legrandOptions, readInitialBatteryState, tzLegrand} from "../lib/legrand";
+import * as m from "../lib/modernExtend";
+import * as reporting from "../lib/reporting";
+import * as tuya from "../lib/tuya";
+import type {DefinitionWithExtend, Fz} from "../lib/types";
 
-  const e = exposes.presets;
-  const ea = exposes.access;
+const e = exposes.presets;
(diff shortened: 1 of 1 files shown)


src/devices/legrand.ts
```diff
@@ -20,14 +20,35 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Legrand",
         description: "Wireless and batteryless 4 scenes control",
         ota: true,
-        meta: {multiEndpoint: true, battery: {voltageToPercentage: {min: 2500, max: 3000}}, publishDuplicateTransaction: true},
-        fromZigbee: [fz.identify, fz.battery, fz.command_recall],
+        meta: {battery: {voltageToPercentage: {min: 2500, max: 3000}}, publishDuplicateTransaction: true},
+        fromZigbee: [
+            fz.identify,
+            fz.battery,
+            {
```
Koenkk: Could you move this into a `fzLocal` similar to how it is done for other devices? 

https://github.com/search?q=repo%3AKoenkk%2Fzigbee-herdsman-converters%20fzlocal&type=code

Koenkk:
I see a lot of formatting changes: https://github.com/Koenkk/zigbee-herdsman-converters/pull/12151/changes, could you revert these?

Koenkk:
Thanks!

matthom3:
does this also apply to the wnacb40w1? I believe its the same issue I've encountered.

https://www.legrand.us/wiring-devices/adorne-collection/switches/adorne-wireless-smart-scene-switch-with-netatmo-white/p/wnacb40w1

alexhk:
@matthom3 
> does this also apply to the wnacb40w1? 

Is this recognized as [067755](https://www.zigbee2mqtt.io/devices/067755.html#legrand-067755) in z2m? If so, then yes.

PR #12159, opened 2026-05-09
tuya circuit-breaker support dlqFault

GrumpyMeow:
Hi Koen,
This PR provides an expose and converter named "dlqFault" for "Tuya-based circuit breakers". Those devices expose the faultcode in a standardized way. I also replaced (duplicated) code from other converters with dlqFault.

!! I have not had the code running !!
!! New dev !!

I'm hoping/planning to also implement "alarm_set_1" and "alarm_set_2" which then can also be used by the Tuya Circuit Breaker converters.

Here is the docs i've used to check the enums: https://developer.tuya.com/en/docs/iot/s?id=Kb0kikd1lya17


andrei-lazarov (commented):
This is good, thanks!

1. Please shorten `tuya.exposes.` to `te.` (add at the beginning of the file, if it isn't there already).

2. Fault is a **bitmap** afaik.
I _suppose_ the device can report multiple faults at once (but I have no way of testing that).
For example, if it reports 3, I would interpret that as 2 + 1 = `surge_alarm` + `short_circuit_alarm`.
Currently the code can't handle this scenario, right? **But it's ok to merge and fix it later.**

3. What is `dlq`? Maybe call it `circuitBreakerFault`

GrumpyMeow:
3. Tuya uses the codename `dlq` for Circuit Breakers in their documentation. That's why i chose to use that prefix as it is a specific reference to the Tuya specs. With this information, would it be ok to keep `dlq` as prefix?   
https://developer.tuya.com/en/docs/iot/f?id=Kb0kihuxmtj62

2. I do know that Fault is a bitmap and should actually be parsed differently. Most implementation were parsing it as an enum and didn't want to create too much of changes all at the same time.

1. Aha. `te.` it will be. I didn't knew the link. 

I will change

andrei-lazarov:
In that case I think dlq is gibberish. Let's do human readable 🙂

Yes let's not cause that many changes at once

src/devices/nous.ts
```diff
@@ -272,7 +272,7 @@ export const definitions: DefinitionWithExtend[] = [
             e.binary("switch_prepayment", ea.STATE_SET, true, false).withDescription("Switch prepayment mode ON/OFF"),
             e.numeric("balance_energy", ea.STATE).withUnit("kWh").withDescription("Remaining energy balance for prepayment"),
             e.numeric("charge_energy", ea.STATE).withUnit("kWh").withDescription("Last charged energy amount"),
-            e.binary("fault", ea.STATE, true, false).withDescription("General fault detected"),
+            tuya.exposes.dlqFault(),
```
Koenkk, line 275: What does `dlq` mean?

GrumpyMeow:
i've nuked my branch. (doh).
I'm gonna retry a new and improved Pull Request...

src/devices/nous.ts
GrumpyMeow: dlq is the code Tuya uses as device code for Circuit Breakers. 
I thought it would make sense to use that code. But upon suggestion i'll rename it to CircuitBreaker..

PR #12160, opened 2026-05-09
fix(ignore): Sort some Tuya datapoints

andrei-lazarov:
@Koenkk from now on, can we make sure all new devices have the datapoints sorted?
Is a test possible?

Exposes can be ordered anyhow - up to the contributor.

But the random datapoint order makes it very hard to fix detections (find the correct definition for a device).


Koenkk:
I think it will be hard to test this, would first need to fix all current violations, I think it's nicer to start using modern extend for Tuya datapoints, some initial work has already been done: https://github.com/Koenkk/zigbee-herdsman-converters/blob/3e52faff71453462f4615bda4b63b902bdadf4c6/src/lib/tuya.ts#L3483, then we can get rid of `meta.tuyaDatapoints` entirely.

PR #12162, opened 2026-05-09
feat: expose power_on_behavior for Woolley BSD29/BSD59

Xplouder:
The Woolley BSD29/BSD59 (model CK-BL702-SWP-01(7020)) smart plug supports the standard Zigbee genOnOff.startUpOnOff attribute, but it was not previously exposed in Zigbee2MQTT.

This PR adds support for power_on_behavior by:

Adding fz.power_on_behavior to fromZigbee for state reporting.
Adding tz.power_on_behavior to toZigbee for command support.
Adding exposes.power_on_behavior() to the device definition.
Verified with the following values:

off (0)
on (1)
previous (255)
The device correctly remembers and applies the behavior after a mains power restoration.

### DIFF
--- a/src/devices/woolley.ts
+++ b/src/devices/woolley.ts
@@ -36,8 +36,8 @@ export const definitions: DefinitionWithExtend[] = [
         model: "BSD29/BSD59",
         vendor: "Woolley",
         description: "Zigbee 3.0 smart plug",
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12167, opened 2026-05-10
Aqara W600: expose running_state from valve position

nilvanis:
- Adds `running_state` for the Aqara W600 with `idle` and `heat`.
- Derives it from valve position:
  - `0` -> `idle` 
  - any open valve position -> `heat`
- Keeps the existing `position` expose unchanged.
- Moves the valve-position expose out of the device entry because the value is now used in two places: as the existing `position` field and as the source for `running_state`. Keeping that W600-specific valve handling together avoids two separate copies of the same valve-position setup.
- Sets `running_state` to `idle` immediately when the thermostat is switched off.


nilvanis:
The W600 does not appear to report a standard current heating action, this is why I used the valve position.

Koenkk:
Thanks!

PR #12171, opened 2026-05-10
Add model name for Somfy Remote

geoffoxholm:
My device has a different Model. Once I added it, it showed up correctly. It works as expected.

### DIFF
--- a/src/devices/somfy.ts
+++ b/src/devices/somfy.ts
@@ -77,7 +77,7 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        zigbeeModel: ["Ysia 1 Zigbee Europe"],
+        zigbeeModel: ["Ysia 1 Zigbee Europe", "Ysia 1 HP Zigbee"],
         model: "1871157",
         vendor: "SOMFY",
         description: "Ysia 1 channel blinds remote",


Koenkk:
Thanks!

adriweb:
Unfortunately this wasn't exactly right.

`1871153` is `Ysia 1 HP Zigbee`, not an alias of the other one as per this patch.
Source: https://www.somfypro.com/products/-/e-cat/1871153

So basically, I think the correct addition is to revert the patch and add this:
```js
    {
        zigbeeModel: ["Ysia 1 HP Zigbee"],
        model: "1871153",
        vendor: "SOMFY",
        description: "Ysia 1 channel blinds remote",
    },
```

Should I make a patch?

geoffoxholm:
> Unfortunately this wasn't exactly right.

The difference is the `model` value, is that right?

I think you're right, and that fixing things will aid in discovery. It would be great to see a correction.

Sorry about my mistake. What will be needed to get the picture also?

adriweb:
Note quite sure about the image but from somfy's websites we have:
- `1871157`: https://asset.somfy.com/Image/7b9da858-ee1f-457e-b4c8-d9c4eede960c_ysia-1-zigbee-remote-control.jpg
- `1871153`: https://asset.somfy.com/Image/ced3b11f-a079-4269-bd9b-a3488e5680c6_Ysia%201c%2034.png.jpg

... which are almost the same. To the point that I'm wondering whether it's just 3D Render artifacts. Might need to check actual photos of the products.

PR #12176, opened 2026-05-10
fix: Inovelli: batch all keys from combined SET payloads

### LINKED ISSUE #11698: VZM31-SN: defaultLevelLocal and defaultLevelRemote share a toZigbee converter.   only the first key in combined SET payload is written, second is silently dropped
Running Z2M 2.9.1-1
SLZB-07 coordinator with nerivec's latest firmware

Hiya' team, can't say thanks enough for zigbee2mqtt.  First bug report.  

Description

On the Inovelli VZM31-SN (Blue Series 2-1), defaultLevelLocal (parameter 9) and defaultLevelRemote (parameter 10) appear to share the same toZigbee converter. When both keys are present in a single SET payload, Z2M processes only the first key and silently drops the second. No error or warning is produced.

**Steps to reproduce**

Note the current values of both parameters, then send a combined payload:
```
Topic:   zigbee2mqtt/<device>/set
Payload: {"defaultLevelLocal": 100, "defaultLevelRemote": 100}
```
Result: Only defaultLevelLocal is written (100). defaultLevelRemote remains unchanged.

Reverse the key order to confirm it is strictly first-key-wins:
```
Topic:   zigbee2mqtt/<device>/set
Payload: {"defaultLevelRemote": 50, "defaultLevelLocal": 50}
```
Result: Only defaultLevelRemote is written (50). defaultLevelLocal remains unchanged.

Sending each key individually works correctly:
```
Topic:   zigbee2mqtt/<device>/set
Payload: {"defaultLevelRemote": 100}

Topic:   zigbee2mqtt/<device>/set
Payload: {"defaultLevelLocal": 100}
```

**Expected behavior**

Both parameters are written when both keys appear in the same payload, or Z2M raises an error/warning indicating the conflict.

**Actual behavior**

The second key in the payload is silently dropped with no indication of failure. This is particularly problematic for automations that set both values together (e.g., keeping local and remote default levels in sync), as they will appear to succeed while only half the write occurs.  This was the case in my custom adaptive lighting scripts. 

**Workaround:** 

Per attribute sub-topics.

rohankapoorcom:
## Summary

Fixes #11698.

`inovelli_parameters` is a single shared `Tz.Converter` whose `key` array lists every writable Inovelli attribute, but its `convertSet` only wrote the one `key` it was called with. Z2M's publish loop dispatches any given converter at most once per endpoint per payload, so every Inovelli key after the first in a combined SET like `{"defaultLevelLocal": 100, "defaultLevelRemote": 100}` was silently dropped (this is exactly the symptom reported in #11698).

This PR rewrites `convertSet` to walk `meta.message` and bucket every writable key by its resolved endpoint, then emit one multi-attribute write per bucket.

- Combined payloads now write every accepted key.
- VZM36's `_1` / `_2` suffixed keys still route to the correct endpoint because bucketing uses `resolveEndpointAndKey`.
- An N-key payload on a single endpoint collapses from N OTA writes to 1.
- Read-only and unknown keys in the same payload are skipped, and they are also excluded from the returned optimistic `state` so it stays honest.
- `convertGet` is intentionally left single-key — Z2M's GET path already calls it per key, and pre-batching reads would be a behavior change.

## Test plan

- [x] `pnpm test` — all 561 tests pass (192 in `test/inovelli.test.ts`)
- [x] `pnpm run build`
- [x] Added unit tests on `VZM31-SN`:
  - Combined `{defaultLevelLocal, defaultLevelRemote}` produces a single write with both attribute IDs (the exact repro from #11698).
  - Reverse key order produces the same single batched write.
  - Mixed unknown + read-only keys are filtered out; only the writable key is sent.
  - Mixed enum + numeric keys both go through, with the enum value mapped correctly.
- [x] Added unit test on `VZM36`: combined `{defaultLevelRemote_1, defaultLevelRemote_2}` produces two writes — one to ep1, one to ep2.
- [x] Pre-existing tests for single-key writes, suffixed-key endpoint resolution, and the read-only converter still pass unchanged.

Link to picture pull request: N/A (no new device).

### DIFF
--- a/src/lib/inovelli.ts
+++ b/src/lib/inovelli.ts
@@ -2340,29 +2340,43 @@ const tzLocal = {
         ({
             key: Object.keys(attributes).filter((a) => !attributes[a].readOnly),
             convertSet: async (entity, key, value, meta) => {
-                const {entityToUse} = resolveEndpointAndKey(entity, key, meta);
+                // Walk meta.message rather than just `key`: Z2M dispatches this converter once per SET payload,
+                // so writing only `key` silently drops every other Inovelli key (issue #11698). Bucket by the
+                // resolved endpoint so VZM36 _1/_2 suffixed keys still route correctly.
+                const buckets = new Map<Zh.Endpoint | Zh.Group, {payload: {[id: number]: {value: unknown; type: number}}; state: KeyValue}>();
 
-                if (!(key in attributes)) {
+                for (const [msgKey, msgValue] of Object.entries(meta.message)) {
+                    const attribute = attributes[msgKey];
+                    if (!attribute || attribute.readOnly) {
+                        continue;
+                    }
+
+                    const {entityToUse} = resolveEndpointAndKey(entity, msgKey, meta);
+                    const valueMap = resolveValueMap(attribute, model);
+                    const writeValue = attribute.displayType === "enum" ? valueMap?.[msgValue as string] : msgValue;
+
+                    let bucket = buckets.get(entityToUse);
+                    if (!bucket) {
+                        bucket = {payload: {}, state: {}};
+                        buckets.set(entityToUse, bucket);
+                    }
+                    bucket.payload[attribute.id] = {value: writeValue, type: attribute.dataType};
+                    bucket.state[msgKey] = msgValue;
+                }
+
+                if (buckets.size === 0) {
                     return;
                 }
 
-                const valueMap = resolveValueMap(attributes[key], model);
-                const payload = {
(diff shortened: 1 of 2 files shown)


rohankapoorcom:
Docker image to test this functionality: `ghcr.io/rohankapoorcom/zigbee2mqtt-custom:rohankapoorcom.zigbee-herdsman-converters-48b363f7__koenkk.zigbee-herdsman.master-20260510-195901`

Koenkk:
thanks!

PR #12181, opened 2026-05-11
Create handshake_finland.ts

kgraafmans:
Link to picture pull request: (https://github.com/kgraafmans/zigbee2mqtt.io/pull/1)


src/devices/handshake_finland.ts
```diff
@@ -0,0 +1,15 @@
+import {deviceEndpoints, light} from "../lib/modernExtend";
+import type {Definition} from "../lib/types";
+
+const definitions: Definition[] = [
+    {
+        zigbeeModel: ["AGGE Zigbee smart push dimmer"],
+        model: "AGGE Zigbee smart push dimmer",
```
Koenkk: Is there a proper model number written on the device? If yes use that

src/devices/handshake_finland.ts
```diff
@@ -0,0 +1,15 @@
+import {deviceEndpoints, light} from "../lib/modernExtend";
```
Koenkk, line 1: Please also add this file to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts

kgraafmans: done

src/devices/handshake_finland.ts
kgraafmans: done

Koenkk:
Thanks!

Issue #2916, opened 2021-08-16
Philips Hue wall switch module OTA


uvjim:
Had a couple of errors appear in the logs (Home Assistant add-on) so I enabled debug logging to try and work out which device was causing the error.

Below are the logs... It looks the RDM001 (Philips Hue wall switch module) is having an issue with OTA.

```
debug 2021-08-16 18:39:28: Received Zigbee message from 'Kitchen - Switch', type 'commandQueryNextImageRequest', cluster 'genOta', data '{"fieldControl":0,"fileVersion":1030,"imageType":284,"manufacturerCode":4107}' from endpoint 1 with groupID null
debug 2021-08-16 18:39:28: Check if update available for '0x001788010b00092d' (RDM001)
error 2021-08-16 18:39:28: Failed to call 'OTAUpdate' 'onZigbeeEvent' (AssertionError [ERR_ASSERTION]: No image available for imageType '284'
    at getImageMeta (/app/node_modules/zigbee-herdsman-converters/lib/ota/zigbeeOTA.js:22:5)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at isNewImageAvailable (/app/node_modules/zigbee-herdsman-converters/lib/ota/common.js:265:18)
    at Object.isUpdateAvailable (/app/node_modules/zigbee-herdsman-converters/lib/ota/common.js:256:23)
    at OTAUpdate.onZigbeeEvent (/app/lib/extension/otaUpdate.js:52:31)
    at Controller.callExtensionMethod (/app/lib/controller.js:408:58)
    at Controller.onZigbeeEvent (/app/lib/controller.js:280:24))
info  2021-08-16 18:39:28: MQTT publish: topic 'zigbee2mqtt/Kitchen - Switch', payload '{"action":null,"battery":100,"device":{"applicationVersion":9,"dateCode":"20210115","friendlyName":"Kitchen - Switch","hardwareVersion":3,"ieeeAddr":"0x001788010b00092d","manufacturerID":4107,"manufacturerName":"Signify Netherlands B.V.","model":"929003017102","networkAddress":23621,"powerSource":"Battery","softwareBuildID":"1.0.3","stackVersion":2,"type":"EndDevice","zclVersion":2},"device_mode":"single_rocker","last_seen":"2021-08-16T17:39:28.360Z","linkquality":255}'
[...]

PR #12188, opened 2026-05-11
Enable OTA for Hue wall switch modules

mundschenk-at:
Was there a specific reason not to enable OTA for the Hue wall switch modules? Otherwise, I think we should allow updates (and downgrades, though there is no public image for the "good" FW version for the RDM001).

### DIFF
--- a/src/devices/philips.ts
+++ b/src/devices/philips.ts
@@ -2824,6 +2824,7 @@ export const definitions: DefinitionWithExtend[] = [
             const options = {manufacturerCode: Zcl.ManufacturerCode.SIGNIFY_NETHERLANDS_B_V, disableDefaultResponse: true};
             await endpoint.write("genBasic", {52: {value: 0, type: 48}}, options);
         },
+        ota: true,
     },
     {
         zigbeeModel: ["RWL020", "RWL021"],


Koenkk:
Thanks!

Pferdebockwurst:
There was indeed a certain reason to not enable OTA for these devices. See my comment [here](https://github.com/Koenkk/zigbee-herdsman-converters/issues/2916#issuecomment-2683447457) and the further discussion.

People updating their firmware from the older 1.x branch to the newer 2.x will indefinitely loose the function to use the switches for toggling.

@Koenkk @mundschenk-at

mundschenk-at:
@Pferdebockwurst: I am aware for RDM001, but for RDM004 that was never an option.

Pferdebockwurst:
> @Pferdebockwurst: I am aware for RDM001, but for RDM004 that was never an option.

Okay, so OTA has not been activated for the RDM001? Then it's all good.

Pferdebockwurst:
> @Pferdebockwurst: I am aware for RDM001, but for RDM004 that was never an option.

Well, I've just tested this with my updated z2m instance and for sure it offers an OTA for the old RDM001 devices. You really should deactivate this as soon as possible, otherwise you would have a lot of users breaking their wall switch modules.

@Koenkk @mundschenk-at

mundschenk-at:
Yes of course the converter applies to both HW versions. But since it's been some time since RDM001 has even been sold, don't you think everyone who is even aware of direct binding (the Hue app never does that) knows not to ever update their old wall switch modules? 

AFAIK Z2M was the only open stack that never had OTA enabled for these devices.

Pferdebockwurst:
No, I don't think so. We got to know about the regression only because some users updated their modules while they still were using a Hue bridge. Z2M never offered OTA for these decisions so there definitely have to be a lot of people, that never even had to think about the consequences of an update.

mundschenk-at:
@Koenkk I think we'd need to split the converter if we want to have different OTA behavior. If you want to go into that direction, I can prepare a PR. 

Has anyone tested the new RDM004 FW (2.85.1) for it's behavior? I have updated one of my devices, but I'm not super familiar with the direct binding process and I don't have a sniffer set up.

andrei-lazarov:
Can you guys document this behavior?
https://www.zigbee2mqtt.io/devices/929003017102.html

Z2M provides these versions:
- RDM001 (SwitchModule-EFR32MG13)
  - `33574179` (downgrade)
  - `33576197` (upgrade)
- RDM004 (SwitchModule-EFR32MG22)
  - `33574179` (downgrade)
  - `33576193` (upgrade)

In which version does the command change from Toggle to alternating On/Off?

There's actually more to it...
@burmistrzak found out: the remote asks the coordinator / the parent router for the current state, to send the appropriate On/Off command. (and defaults to alternating when doesn't receive a reply)
For the versions that ask the coordinator, we could implement a response in Z2M, to control this behavior 
- https://github.com/koenkk/zigbee2mqtt/issues/31955

Pferdebockwurst:
@andrei-lazarov I don't know if that is the case with the module too, because in your linked issue they talk about a remote. But it definitely would make sense if the module did also ask for the state.

But anyway, I can definitely say that the regression is caused by updating from 1.x to 2.x. Everything is documented [here](https://github.com/Koenkk/zigbee2mqtt/issues/26775).

mundschenk-at:
> In which version does the command change from Toggle to alternating On/Off?

AFAIK in all of them. Only the `1.0.x` firmwares for RDM001 send Toggle, and we never had update images for those versions.

> 
> There's actually more to it... @burmistrzak found out: the remote asks the coordinator / the parent router for the current state, to send the appropriate On/Off command. (and defaults to alternating when doesn't receive a reply) For the versions that ask the coordinator, we could implement a response in Z2M, to control this behavior

That's interesting. I'm not sure it actually helps for the main use case for direct binding though (having the switch work properly without the coordinator being online).

andrei-lazarov:
Ok at least I added the info to docs, so the user can decide to update or not..
I agree Toggle is usually better, but idk if we should disable OTA

jamesonuk:
> Well, I've just tested this with my updated z2m instance and for sure it offers an OTA for the old RDM001 devices. You really should deactivate this as soon as possible, otherwise you would have a lot of users breaking their wall switch modules.

Yeah came here after being offered OTA upgrade for RDM001 devices.   I did upgrade a RDM004 and that still sends on/off rather than toggle.

Is this actually a valid OTA for RDM001 and if so that raises an interesting question that if the firmware can be applied to both then in theory the 1.x firmware on RDM001 could be flashed to RDM004 ??

I am presuming no-one is going to try and apply the update to RDM001 device on the basis that if it does apply it is likely to change toggle behaviour?   Need a brave person....

mundschenk-at:
Yes, the FW updates for RDM001 are valid and used in pure Hue ecosystems, but they bring in all the "features" the RDM004 had from inception. I do believe they are different binaries though, due to the non-identical HW platoform?

Whether the RDM004 could be downgraded to 1.0.x FW for RDM001 is moot however, as we don't have installation images for those FW versions.

jamesonuk:
> Yes, the FW updates for RDM001 are valid and used in pure Hue ecosystems, but they bring in all the "features" the RDM004 had from inception.

So to be clear updating the RDM001 will change from toggle to alternate on/off?

mundschenk-at:
> So to be clear updating the RDM001 will change from toggle to alternate on/off?

Yes.

PR #12190, opened 2026-05-12
fix: changed the name of HS1CA-E PLUS and measured CO

Leo2442926161:
Dear Z2M team,
1. fix: changed the name of HS1CA-E-PLUS to HS1CA-E PLUS
2. fixed measured CO shows null
3. add bind haDiagnostic for HS1RM


Koenkk:
Thanks!

Issue #12152, opened 2026-05-08
Legrand Bticino K4027C not send state MOVING


mrgiov:
The Legrand K4027C device does not send the `MOVING` state while the cover is in motion.

I noticed that in `converters/fromZigbee.ts` there is already the following converter:

```ts
export const bticino_4027C_binary_input_moving: Fz.Converter<"genBinaryInput", undefined, ["attributeReport", "readResponse"]> = {
    cluster: "genBinaryInput",
    type: ["attributeReport", "readResponse"],
    options: [exposes.options.no_position_support()],
    convert: (model, msg, publish, options, meta) => {
        return options.no_position_support
            ? {action: msg.data.presentValue ? "stopped" : "moving", position: 50}
            : {action: msg.data.presentValue ? "stopped" : "moving"};
    },
};
```

However, it does not seem to be used in `devices/legrand.ts` around line 219.
```
fromZigbee: [fz.cover_position_tilt, fz.identify, fzLegrand.cluster_fc01, fzLegrand.calibration_mode(true), fzLegrand.command_cover],
```

Would using this converter solve the issue?

Could someone help me create an external_converter and load it into Zigbee2MQTT to run a test?

Thanks.

andrei-lazarov:
Hello. Please try: [legrand.js](https://github.com/user-attachments/files/27540684/legrand.js)
You might also need to add reporting for `genBinaryInput` -> `presentValue`

PR #12191, opened 2026-05-12
Remove converters not used by any device (bticino_4027*)

svhelge:
Remove `bticino_4027C_binary_input_moving` from fromZigbee.ts.
Remove `bticino_4027C_cover_state` and `bticino_4027C_cover_position` from toZigbee.ts.


andrei-lazarov (commented):
Wait, these might be needed
- [Legrand Bticino K4027C not send state MOVING #12152](https://github.com/Koenkk/zigbee-herdsman-converters/issues/12152)

Koenkk:
thanks!

Koenkk:
I think I have merged this to quickly, should it be reverted?

svhelge:
> I think I have merged this to quickly, should it be reverted?

My opinion is that it can be removed. The code is not in use for any device, and it can still be found here at github after removal.

PR #12192, opened 2026-05-12
Refactor Livolo converters - move to local definition

svhelge:
This PR refactors the Livolo device converters by moving device-specific converter implementations from the global converter modules to the Livolo device definition file.


Koenkk:
Thanks!

PR #12197, opened 2026-05-13
Add two unsupported zebra somfy blinds

alex-oswald:
Add two Somfy zebra shades that showed as not supported for me.

### DIFF
--- a/src/devices/somfy.ts
+++ b/src/devices/somfy.ts
@@ -51,17 +51,10 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [m.windowCovering({controls: ["lift"]}), m.battery()],
     },
     {
-        zigbeeModel: ["Sonesse 28 WF Zebra"],
+        zigbeeModel: ["Sonesse 28 WF Li-Ion Zebra", "Sonesse 28 WF Zebra"],
         model: "Sonesse 28 WF Li-Ion Zebra",
         vendor: "SOMFY",
-        description: "Somfy Zebra shades (external battery)",
-        extend: [m.windowCovering({controls: ["lift"]}), m.battery()],
-    },
-    {
-        zigbeeModel: ["Sonesse 28 WF Li-Ion Zebra"],
-        model: "Sonesse 28 WF Li-Ion Zebra",
-        vendor: "SOMFY",
-        description: "Somfy Zebra shades (internal battery)",
+        description: "Somfy Zebra shades (battery operated, Li-Ion model uses an internal battery, while the other uses an external battery)",
         extend: [m.windowCovering({controls: ["lift"]}), m.battery()],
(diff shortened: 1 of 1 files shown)


alex-oswald:
@Koenkk Do you actually want the device merged with two models even though they are actually different devices? Similar, but one has a built in battery, and one uses an external battery.

src/devices/somfy.ts
```diff
@@ -50,6 +50,20 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Blinds",
         extend: [m.windowCovering({controls: ["lift"]}), m.battery()],
     },
+    {
+        zigbeeModel: ["Sonesse 28 WF Zebra"],
+        model: "Sonesse 28 WF Li-Ion Zebra",
```
Koenkk: Is there a proper model number written on the device? If yes please use that.

Koenkk:
> Do you actually want the device merged with two models even though they are actually different devices?

I think they can be merged to avoid code duplication indeed, for separate identification you can use the whitelabel, see https://github.com/Koenkk/zigbee-herdsman-converters/blob/e5eb8443c471cacaaf041d6859e283341ebe445d/src/devices/multir.ts#L42

src/devices/somfy.ts
alex-oswald: I don't see any actual model numbers on mine.

alex-oswald:
@Koenkk Okay I combined them. Let me know if you need anything else changed. Thanks!

src/devices/somfy.ts
Koenkk: Do you have a link of the product page? Maybe we can use the ean.

alex-oswald:
My assumption is the following. But these came in Graber blinds I purchased. I didn't buy the motors by themselves.
https://www.somfysystems.com/en-us/products/1241755/sonesse-28-wirefree-zigbee-li-ion
https://www.somfypro.com/products/-/e-cat/1241754

I can update with them separate and use the ref numbers on the page, `1241755` and `1241754`. They look the match others in this file, but I can't be positive because I don't see a part number on blinds I bought.

alex-oswald:
@Koenkk Okay this should be good now. One is the same part number as one already here, but with a different model being reported by the motor. I added the part number to the other one.

src/devices/somfy.ts
alex-oswald: Done.

Koenkk:
Thanks!

Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `1241754.png`

PR #12201, opened 2026-05-13
Move Zemismart ZM25AZ01 and Trublockout TB25-DC-10/25Z to new model

pantsman0:
Moves the devices with manufacturer _TZE200_m6lwazh9 to a new cover definition that exports all data points identified in https://github.com/Koenkk/zigbee2mqtt/issues/31774

I haven't done this type before, so if there are any issues please just let me know.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -26872,17 +26872,15 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        zigbeeModel: ["lwazh9\u0000", "lwazh9"],
         fingerprint: [...tuya.fingerprint("TS0601", ["_TZE200_m6lwazh9"])],
-        model: "TS0601_cover_14",
+        model: "TB25-DC-10/25Z",
         vendor: "Tuya",
         description: "Curtain motor/roller blind motor/window pusher/tubular motor",
         whiteLabel: [
             {vendor: "Zemismart", model: "ZM25AZ01"},
             {vendor: "Trublockout", model: "TB25-DC-10/25Z"},
         ],
-        fromZigbee: [legacy.fromZigbee.tuya_cover],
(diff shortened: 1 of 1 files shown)


Koenkk:
I think we can keep it in the current definition by making the `exposes` dynamic, this will prevent some code duplication, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/3c6720d951a73ac7aea2033bb85de8711ff1737d/src/devices/tuya.ts#L1807

andrei-lazarov:
Current definition is using legacy tuya tz/fz, don't think we can combine them.. 
I didn't find a similar-enough definition for dynamic `exposes`..
I'd say new converter is fine.

But you need to fix it a little: remove `zigbeeModel`, use `tuya.fingerprint`, use `tuyaBase`.
(Start with `TS0601_cover_13` and modify it). Also test the code in an external converter

pantsman0:
I have updated the config and tested it with an external converter. I have tested each feature except favourite position, but that's only because my blinds aren't responding to the "go to favourite" command from my remotes.

Koenkk:
@pantsman0 could you also check https://github.com/Koenkk/zigbee-herdsman-converters/pull/12201#issuecomment-4444329645 ?

pantsman0:
Oh, sorry. I saw your 👍 on https://github.com/Koenkk/zigbee-herdsman-converters/pull/12201#issuecomment-4445632005 and assumed that meant you were fine without the dynamic `exposes`.

Would you like me to take it back into an  existing definition, or just change  `exposes` to a lambda in the new definition? Personally I agree with andrei-lazarov that there isn't a great candidate to merge it back into.

Koenkk:
> Would you like me to take it back into an existing definition

If possible yes as this reduces code duplication

PR #12202, opened 2026-05-13
Move Datek converters to datek.ts 

svhelge:
Move Datek specific converters to datek.ts.
Add cluster for custom attribute in `ssIasZone` cluster.
Moved function `hw_version` within `fromZigbee.ts` (from non-generic to generic).


Koenkk:
Thanks!

PR #12205, opened 2026-05-13
Move custom converter from fromZigbee.ts to meazon.tz

svhelge:
Refactor Meazon meter device support by moving the meazon_meter converter from the global converters file into the device-specific file and improving type safety.


Koenkk:
thanks!

PR #12206, opened 2026-05-13
Refactor custom converters for the ZigUP device (from global to local)

svhelge:
Refactor the ZigUP device converters by moving device-specific converter functions from the global scope (in fromZigbee.ts and toZigbee.ts) into local converters within custom_devices_diy.ts


Koenkk:
thanks!

PR #12208, opened 2026-05-13
Refactor Stelpro global converters to local scope

svhelge:
Refactor Stelpro thermostat device support by consolidating code organization and moving device-specific converters from the main toZigbee.ts file into the stelpro.ts module.


Koenkk:
thanks!

PR #12211, opened 2026-05-14
fix(danfoss): haDiagnostic bind fails if already bound, wrap in try-catch

storm1ng:
Re-configure fails if haDiagnostic is already bound, the try-catch should fix this.


Koenkk:
thanks!

PR #12214, opened 2026-05-14
Refactor diyruz: move converters, introduce custom clusters, and enforce stronger typing

svhelge:
Refactor the diyruz device converters by extracting device-specific fromZigbee and toZigbee converters from the global converter files into the device definition file itself. Adds custom clusters and proper TypeScript typing.


Koenkk:
thanks!

PR #12218, opened 2026-05-15
giex: improve GX03 2-zone valve logic and bootstrapping

PavelKrassikov:
Replaced the current rudimentory implementation. Separated timer/countdown to keep timer as a user setting and countdown as a counter, revealed last duration for statistics, reset countdown which was staying non-zero, set initial timer value to unlock this control in HA.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -20718,9 +20718,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "GX03",
         vendor: "GIEX",
         description: "GIEX 2-zone watering timer",
-        fromZigbee: [tuya.fz.datapoints, fz.ignore_tuya_set_time],
-        toZigbee: [tuya.tz.datapoints],
-        configure: tuya.queryDeviceState,
+        extend: [tuya.modernExtend.tuyaBase({dp: true, queryOnConfigure: true})],
         exposes: [
             exposes.binary("valve_1", ea.STATE_SET, "ON", "OFF").withDescription("State of the valve 1"),
             e.enum("state_1", ea.STATE, ["Manual", "Auto", "Closed"]).withDescription("State of the valve 1"),
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -20715,80 +20715,69 @@ export const definitions: DefinitionWithExtend[] = [
     },
     {
         fingerprint: tuya.fingerprint("TS0601", ["_TZE284_8zizsafo", "_TZE284_iilebqoo"]),
-        model: 'GX03',
-        vendor: 'GIEX',
-        description: 'GIEX 2-zone watering timer',
-        fromZigbee: [{
-            cluster: 'manuSpecificTuya',
-            type: ['commandDataResponse', 'commandDataReport'],
-            convert: (model, msg, publish, options, meta) => {
-                const result = tuya.fz.datapoints.convert(model, msg, publish, options, meta) || {};
-                if (meta.state.timer_1 === undefined && result.timer_1 === undefined) {
-                    result.timer_1 = result.timer_2 = 5;
-                    const endpoint = meta.device.getEndpoint(1);
-                    tuya.sendDataPointValue(endpoint, 13, result.timer_1).catch(() => {});
-                    tuya.sendDataPointValue(endpoint, 14, result.timer_2).catch(() => {});
-                    endpoint.command('manuSpecificTuya', 'dataQuery', {}).catch(() => {});
-                }
-                if ((result.state_1 === 'Closed' || meta.state.state_1 === 'Closed') && meta.state.countdown_1 !== 0) result.countdown_1 = 0;
-                if ((result.state_2 === 'Closed' || meta.state.state_2 === 'Closed') && meta.state.countdown_2 !== 0) result.countdown_2 = 0;
-                return result;
-            }
-        }, fz.ignore_tuya_set_time],
+        model: "GX03",
+        vendor: "GIEX",
+        description: "GIEX 2-zone watering timer",
+        fromZigbee: [
+            {
+                cluster: "manuSpecificTuya",
+                type: ["commandDataResponse", "commandDataReport"],
+                convert: (model, msg, publish, options, meta) => {
+                    const result = tuya.fz.datapoints.convert(model, msg, publish, options, meta) || {};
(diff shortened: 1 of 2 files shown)


src/devices/tuya.ts
```diff
@@ -20717,50 +20717,67 @@ export const definitions: DefinitionWithExtend[] = [
         fingerprint: tuya.fingerprint("TS0601", ["_TZE284_8zizsafo", "_TZE284_iilebqoo"]),
         model: "GX03",
         vendor: "GIEX",
-        description: "GIEX 2 zone watering timer",
-        extend: [tuya.modernExtend.tuyaBase({dp: true})],
+        description: "GIEX 2-zone watering timer",
+        fromZigbee: [
+            {
```
Koenkk: Please implement this as a custom value converter, see lib/tuya.ts for examples

PavelKrassikov: Done

src/devices/tuya.ts
```diff
@@ -20717,50 +20717,48 @@ export const definitions: DefinitionWithExtend[] = [
         fingerprint: tuya.fingerprint("TS0601", ["_TZE284_8zizsafo", "_TZE284_iilebqoo"]),
         model: "GX03",
         vendor: "GIEX",
-        description: "GIEX 2 zone watering timer",
-        extend: [tuya.modernExtend.tuyaBase({dp: true})],
```
Koenkk, line 20721: Please keep this and remove frmZigbee, toZigbee and configure

src/lib/tuya.ts
```diff
@@ -1069,6 +1076,32 @@ export const whitelabel = (vendor: string, model: string, description: string, m
     return {vendor, model, description, fingerprint};
 };
 
+export const giexGx03ValveState = (zoneNum: number) => {
```
Koenkk: Please move this into `export const valueConverter`

src/lib/tuya.ts
```diff
@@ -1069,6 +1076,32 @@ export const whitelabel = (vendor: string, model: string, description: string, m
...
+    return {
+        from: (value: unknown, meta: Fz.Meta, options: KeyValue, publish: Publish): string | number => {
+            // Set initial value to both timers
+            if (meta.state?.timer_1 === undefined) {
+                publish({
+                    timer_1: 5,
+                    timer_2: 5,
+                });
+                const endpoint = meta.device.getEndpoint(1);
+                sendDataPointValue(endpoint, 13, 5).catch(() => {});
```
Koenkk: Can you use await here (and below) and remove `.catch(() => {});`

src/devices/tuya.ts
PavelKrassikov: Does the modern extend do the same for me?
        fromZigbee: [tuya.fz.datapoints, fz.ignore_tuya_set_time],
        configure: tuya.queryDeviceState, // My custom function
 Thanks!

src/lib/tuya.ts
PavelKrassikov: Done

src/lib/tuya.ts
PavelKrassikov: Done

src/devices/tuya.ts
Koenkk: With `extend: [tuya.modernExtend.tuyaBase({dp: true, queryOnConfigure: true})],` it does

PavelKrassikov: Done

Koenkk:
Thanks!

jancoow:
This causes breaking changes with existing setups.  Took me a while to figure out why things didn't work anymore. 

Also, the toggle button won't go to off anymore when the timer is exceeded. So it seems the new implementation is not working properly.

PavelKrassikov:
> This causes breaking changes with existing setups. Took me a while to figure out why things didn't work anymore.
> Also, the toggle button won't go to off anymore when the timer is exceeded. So it seems the new implementation is not working properly.

@jancoow, could you please be more specific what does not work for you anymore?

As for me, the toggle button does not show switching off by itself in Z2M UI for many devices even though I see the respective property change in the log, so I have to reload the UI to see this change for devices. The Home Assistant UI shows the toggle buttons going off without a problem, as well as the related automation.

Thanks!

jancoow:
Automations fail because the toggle button stays on



Also need to subscribe to a different value for the countdown. The implementation is better now, but it was not clear that those changes where made for me

Also last duration does not work.

PavelKrassikov:
Weird values of the properties of your device(s)!
The timers are set to 8 and 45 minutes respectively. Though the respective "last durations" show 300 seconds (5 minutes) each. Also the countdowns show "0" which was not possible in the original implementation because the firmware never zeroed these countdowns, and the developers refused to fix all the firmware issues - I coded the countdowns to become "0" after the state of the respective valve is "Closed".

I would strongly advise you to reconfigure your device(s) on the "About" tab in Z2M, or even better - re-pair your valves thus resetting their properties though keeping the stats and automation.

PavelKrassikov:
I should probably update the Wiki of this device to provide more information about all my finding.

After the device reset/re-pair, the timers are preset to 5 minutes so the ON command will open the respective valve for 5 minutes. I set the timers property to 5 minutes to follow this preset configuration and unlock these controls in the UI.

Though a physical opening by the button of the device always opens this valve for 10 minutes - the firmware ignores the timer setting in this case.

The valve state has 3 states - Closed, Manual (open by command) or Auto (open by the embedded watering schedule). The current implementation only delivers the Closed and Manual states. Though I'm in a process of reverse engineering the offline schedule stored in this device allowing to schedule watering right in the device even if the connection to Zigbee network is lost - this operation will be covered by the "Auto" state.

The device firmware uses just one datapoint to both set the timer and show the countdown thus providing a weird user experience: for example, you set the timer to 3 minutes then physically open this valve and see this timer changed to 10 minutes then 9 minutes, then you close this valve and see this timer is 9 minutes so you assume the neext time you open it you will water for 9 minutes - no way - still 3 minutes as you set before! So I separated this datapoint on two controls to only set the timer and only see the countdown, and I reset the countdown after the valve state gets "Closed".

I did not change the last_duration logic.

The offline schedule is very complicated so I will add it to the "Settings (specific)" tab to keep the main UI simple.

Hope this clarifies.

PR #12228, opened 2026-05-15
Refactor CCT711119 using modernExtend - add toggle to power on behavior 

svhelge:
- Refactoring of existing device support for CCT711119 using modernExtend.
- Add `toggle` to power on behavior for the device by using default values.
- Addition of `version: "0.0.1"` to the device definition.


src/devices/schneider_electric.ts
```diff
@@ -2067,24 +2067,16 @@ export const definitions: DefinitionWithExtend[] = [
...
-        fromZigbee: [fz.on_off, fz.electrical_measurement, fz.metering, fz.power_on_behavior],
-        toZigbee: [tz.on_off, tz.power_on_behavior, tz.electrical_measurement_power],
-        exposes: [
-            e.switch(),
-            e.power().withAccess(ea.STATE_GET),
-            e.energy(),
-            e.enum("power_on_behavior", ea.ALL, ["off", "previous", "on"]).withDescription("Controls the behaviour when the device is powered on"),
+        version: "0.0.1",
+        extend: [
+            m.onOff({powerOnBehavior: {behaviors: ["off", "previous", "on"]}}),
```
Koenkk: Does this device not support the `toggle` power on behaviour?

svhelge: The old implementation wasn't configured with it, so I didn't add it. I'll test the device and see if it's support toggle.

svhelge: I tested the device and it supports `toggle` power on behaviour. I changed the code to `m.onOff()`.

src/lib/modernExtend.ts
```diff
@@ -1149,7 +1152,7 @@ export function pm25(args: Partial<NumericArgs<"pm25Measurement">> = {}): Modern
 
 export interface LightArgs {
     effect?: boolean;
-    powerOnBehavior?: boolean;
+    powerOnBehavior?: boolean | {behaviors?: ("off" | "on" | "toggle" | "previous")[]};
```
```suggestion
    powerOnBehavior?: boolean;
```

I think this change is not needed anymore? I would say remove until we see that devices that do not support all modes.

src/lib/modernExtend.ts
```diff
@@ -545,7 +545,7 @@ export function identify(args: {isSleepy: boolean} = {isSleepy: false}): ModernE
 }
 
 export interface OnOffArgs {
-    powerOnBehavior?: boolean;
+    powerOnBehavior?: boolean | {behaviors?: ("off" | "on" | "toggle" | "previous")[]};
```
Koenkk: I think all changes in this file should be reverted?

svhelge: Agree, done.

Koenkk:
Thanks!

PR #12231, opened 2026-05-16
Legrand 412170: force device_mode to 'switch' on configure, add missing 'identify' expose

alexhk:
Follow-up to #12141 (which clarified the description but left users to set the mode manually).

The 412170's factory default `device_mode` is not switch, which causes erratic behavior with the L to C1 push-button wiring method documented in Legrand's installation sheet.

- This change writes device_mode `switch` (3) on configure.
- It removes the exposed `device_mode`.
- Adds the missing 'identify' exposure.

~~This change writes device_mode `switch` (3) on first configure, setting a flag in `device.meta.deviceModeInitialized` so the write only happens once. User-chosen values are preserved across later re-configures.~~

~~The user-facing device_mode enum remains in exposes/toZigbee, so anyone with an unusual need to set auto can still do so manually. But this is not supported or documented by Legrand for this device (it was a leftover from reusing code from the contactor 412171 implementation).~~

More technical details:

The converter code for the 412170 was reused from an earlier contactor (Legrand 412171) which is a different device with different use-case and wiring method (sustained-state switches instead of momentary push buttons). The officially described use-case for 'auto' mode is to follow On-Peak/Off-Peak electricity rates via wiring of dry contacts from an electricity meter to the 412171, so a water heater will be on during the lower electricity rates. The Zigbee OnOff cluster is disabled when the 412171 is in auto mode, but on the 412170, OnOff keeps working in either mode.

Legrand's Home+Control app exposes an 'auto' mode for the 412171, but does *not* for the 412170 here. Since it's the same firmware on both devices, it accepts the mode 'auto' without complaining on the 412170, but the mode results in erratic behaviour if used with the wiring method L to C1 on the 412170. To make a long story short, the 412170 doesn't have an 'auto' mode. See #12141 for the full details.

Tested with 412170 paired from scratch, write/read confirmed via debug log, flag confirmed in database.db.

### DIFF
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -116,7 +116,7 @@ export const definitions: DefinitionWithExtend[] = [
         ota: true,
         fromZigbee: [fz.identify, fz.electrical_measurement],
         toZigbee: [tzLegrand.identify, tz.electrical_measurement_power],
-        exposes: [e.power().withAccess(ea.STATE_GET)],
+        exposes: [e.power().withAccess(ea.STATE_GET), eLegrand.identify()],
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ["genIdentify", "haElectricalMeasurement"]);
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -1,11 +1,11 @@
 import * as fz from "../converters/fromZigbee";
 import * as tz from "../converters/toZigbee";
 import * as exposes from "../lib/exposes";
-import {eLegrand, fzLegrand, legrandExtend, legrandOptions, readInitialBatteryState, tzLegrand} from "../lib/legrand";
+import {eLegrand, fzLegrand, type LegrandDevicesCluster, legrandExtend, legrandOptions, readInitialBatteryState, tzLegrand} from "../lib/legrand";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
 import * as tuya from "../lib/tuya";
-import type {DefinitionWithExtend, Fz, KeyValueAny} from "../lib/types";
+import type {DefinitionWithExtend, Fz} from "../lib/types";
 
 const e = exposes.presets;
 const ea = exposes.access;
@@ -123,8 +123,7 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
             await reporting.activePower(endpoint);
 
-            const payload: KeyValueAny = {deviceMode: 3};
-            await endpoint.write("manuSpecificLegrandDevices", payload, legrandOptions);
+            await endpoint.write<"manuSpecificLegrandDevices", LegrandDevicesCluster>("manuSpecificLegrandDevices", {deviceMode: 3}, legrandOptions);
             await endpoint.read("manuSpecificLegrandDevices", [0x0000], legrandOptions);
         },
     },
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -5,7 +5,7 @@ import {eLegrand, fzLegrand, legrandExtend, legrandOptions, readInitialBatterySt
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
 import * as tuya from "../lib/tuya";
-import type {DefinitionWithExtend, Fz} from "../lib/types";
+import type {DefinitionWithExtend, Fz, KeyValueAny} from "../lib/types";
 
 const e = exposes.presets;
 const ea = exposes.access;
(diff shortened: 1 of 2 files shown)


andrei-lazarov:
> This change writes device_mode switch (3) on first configure, setting a flag in device.meta.deviceModeInitialized so the write only happens once.
> 
> User-chosen values are preserved across later re-configures.

**I think this should happen on all configure triggers, not just the first one.**

---

But this can be a discussion question.
@Nerivec @Koenkk what do you think should happen on (re)configure?

Currently, reconfigure generally does:
- reset binds to coordinator
- reset reportings (to sensible defaults - overwrites custom user-set values)
- not delete extra binds and reportings (should it?)
- fill incomplete/wrong device-reported info (type, power source, divisors)
- mimmic manuf gateway setup
- **ocasionally change _unreasonable_ default attributes (always or only once?)**

---

- Additionally, should it set _all_ attributes to the default value, for some devices?
- Should it also reset counters, meters?

(wouldn't work with "device definition version bumps", or ota updates though.. we don't want an automatic reset with those)

I have the Nous circuit breaker which has a lot alarm settings. (over/under voltage, current etc.)
"Factory resetting" puts the device in pairing mode, but it does not reset the settings to default. There is no way to do that, even with the Tuya app.
I put the defaults in the descriptions. I'll probably also add them to reconfigure. So moving to a new network resets the values.

alexhk:
> I think this should happen on all configure triggers, not just the first one.

Is this how z2m is meant to work?

So if I set a custom mode for a device (ie. `auto` instead of `switch`) and need to reconfigure for some reason, re-configure should set it back to the default values? I am doing it on first configure only to avoid exactly that: overwriting user-chosen values (but can change it of course if that's how z2m is meant to work).

> I have the Nous circuit breaker
> "Factory resetting" puts the device in pairing mode, but it does not reset the settings to default.

Just like your Nous circuit breaker, this 412170 also only resets the Zigbee pairing info but seems to keep other settings (ie. it keeps Power-on behavior or device mode, even across pairing with different coordinators like z2m or Legrand Gateway...). I tested this a few days ago.

Nerivec:
Indeed, shouldn't gate it like this. It will break in too many scenario (post-OTA, rejoin, etc.). It should always execute the logic on configure.

alexhk:
@Nerivec 
> Indeed, shouldn't gate it like this. It will break in too many scenario (post-OTA, rejoin, etc.). It should always execute the logic on configure.

Ok, so just to make sure.
* When device is configured the first time, configure will set device mode `switch`.
* User then selects device_mode `auto` for some reason. 
* Post-OTA, rejoin, etc.: configure will run again and is supposed to set device_mode back to `switch`, despite the user having said it to `auto`? **Please clarify.**

Because exactly this overwriting of user-chosen value is avoided by the current code.

Will change if needed.

andrei-lazarov:
> 412170 doesn't have an 'auto' mode

Well, you worded it like "auto" should never be used, isn't even official for the device. So it sounds fine to always change it.

> it keeps Power-on behavior or device mode, even across pairing with different coordinators like z2m or Legrand Gateway

Tuya app has "remove" and "remove + clear data". That usually works, but not for my device. For simple attributes like power-on behavior, that user is expected to easily understand and change - I don't mind them being left over.

alexhk:
> Well, you worded it like "auto" should never be used, isn't even official for the device. So it sounds fine to always change it.

It is indeed a stupid situation with this 412170, because the device mode has been exposed for 5 years (due to reusing code from 412171) while the device doesn't support it. I have no need for it (auto) and its not supported, but then we could also just remove the device_mode being exposed at the first place. I just didn't dare to remove it... it's been there for 5 years.

But the question remains, should configure overwrite user chosen values? I can't imagine (new to z2m).

Nerivec:
It needs to always run because you don't know the state it will have after such actions (device could reset it, device could have it in RAM instead of storage, etc.).
Configure sets a known default in all cases where it could have been reverted to some unknown or undesired value. It itself can be considered a "one-time trigger", Z2M then decides when it is appropriate to trigger it.
In simplest form, current code would be like doing: `if (first) { if (first) { change_mode } }`

alexhk:
Alright, I'll remove it.
I'll also remove the non-existent device_mode exposure.

src/devices/legrand.ts
```diff
@@ -114,21 +114,18 @@ export const definitions: DefinitionWithExtend[] = [
...
+        toZigbee: [tzLegrand.identify, tz.electrical_measurement_power],
+        exposes: [e.power().withAccess(ea.STATE_GET), eLegrand.identify()],
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
             await reporting.bind(endpoint, coordinatorEndpoint, ["genIdentify", "haElectricalMeasurement"]);
             await reporting.readEletricalMeasurementMultiplierDivisors(endpoint);
             await reporting.activePower(endpoint);
+
+            const payload: KeyValueAny = {deviceMode: 3};
+            await endpoint.write("manuSpecificLegrandDevices", payload, legrandOptions);
```
Nerivec:
```suggestion
            await endpoint.write<"manuSpecificLegrandDevices", LegrandDevicesCluster>("manuSpecificLegrandDevices", {deviceMode: 3}, legrandOptions);
```

Probably need to import the type also.

alexhk: Thanks! the KeyValueAny workaround is how it was done in the existing legrand_device_mode converter in `src/lib/legrand.ts`, but this is indeed cleaner.
I also had to export the interface to make this work.

All done and tested.

alexhk:
This is tested and should be good to go now.

src/devices/legrand.ts
```diff
@@ -114,21 +114,17 @@ export const definitions: DefinitionWithExtend[] = [
...
-            e.power().withAccess(ea.STATE_GET),
-            e
-                .enum("device_mode", ea.ALL, ["switch", "auto"])
-                .withDescription(
-                    "Switch: normal on/off mode, momentary push-buttons on C1/C2 toggle the relay cleanly (recommended). auto: produces erratic behavior.",
-                ),
-        ],
+        fromZigbee: [fz.identify, fz.electrical_measurement],
+        toZigbee: [tzLegrand.identify, tz.electrical_measurement_power],
+        exposes: [e.power().withAccess(ea.STATE_GET), eLegrand.identify()],
```
Koenkk, line 119: To force a reconfigure:

```suggestion
        exposes: [e.power().withAccess(ea.STATE_GET), eLegrand.identify()],
        version: "0.0.1",
```

alexhk: Thanks!

Koenkk:
Thanks!

PR #12232, opened 2026-05-16
fix: Moes Star feather: fix network spam

andrei-lazarov:
- attempt https://github.com/Koenkk/zigbee2mqtt/discussions/32020

If I remember correctly, spamming (non-PM) Tuya device = asking for current time?

Edit: confirmed this is the problem


Koenkk:
Awesome, thanks!

PR #12240, opened 2026-05-18
Update candeo.ts

candeodevelopment:
Adds minimum brightness level setting to Candeo C203, C204, DM204 and DM204v2 dimmer modules.


Koenkk:
Thanks!

PR #12243, opened 2026-05-18
Add variant of Centris V2 ceiling light (3 spots) - 929003809001

Koseng:
Adds a variant of Centris V2 ceiling light (3 spots).

It is basically an existing device. That philips device with different model code:
https://www.zigbee2mqtt.io/devices/929003809101.html


Koenkk:
Thanks!

PR #12249, opened 2026-05-19
Update tuya.ts for _TZ3000_lmlsduws

nomadsole:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/5133/changes/6e57f8b68d49162497bb2cd7ec9d5f439424335c

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -7173,7 +7173,7 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
-        fingerprint: tuya.fingerprint("TS0004", ["_TZ3000_ltt60asa", "_TZ3000_mmkbptmx", "_TZ3000_liygxtcq"]),
+        fingerprint: tuya.fingerprint("TS0004", ["_TZ3000_ltt60asa", "_TZ3000_liygxtcq"]),
         model: "TS0004_switch_module",
         vendor: "Tuya",
         description: "4 gang switch module",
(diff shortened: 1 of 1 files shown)


andrei-lazarov (changes requested):
Thanks!
We do need to merge some definitions though, to avoid code duplication.
Even the two 1-gangs you added are identical. We should use one converter for both, like "TS0001_switch_module" and add them as `whiteLabel`.

Preferably we merge them with the existing definitions. I guess most modules support these features, but they are not exposed. Needs some refactoring

nomadsole:
> Thanks! We do need to merge some definitions though, to avoid code duplication. Even the two 1-gangs you added are identical. We should use one converter for both, like "TS0001_switch_module" and add them as `whiteLabel`.
> 
> Preferably we merge them with the existing definitions. I guess most modules support these features, but they are not exposed. Needs some refactoring

Because I am using the definitions as external converter, I turn to keep them separate.

The real issue is that there are too many of those "whitelabel" devices. And I have to deal with each type case by case. Maybe having options in the frontend allowing users to map the features of a particular model to variants from unknown manufacture is the best way to go.

Koenkk:
I agree with @andrei-lazarov that these should stay in the original definition, see https://github.com/Koenkk/zigbee-herdsman-converters/blob/d2467a085b4076aee7612f446849474bf517f96f/src/devices/tuya.ts#L6813 on how to make features conditional.

PR #12250, opened 2026-05-19
feat(add) static_state to lumi.vibration.agl002

absent42:
`manuSpecificLumi` attribute 0x01F3 (Aqara API name: `static_state`) Fires `true` whenever the device becomes still after a motion event (`movement`, `vibration`, `orientation`, `fall`, or `triple_tap`). The `false` transition is not reliably emitted by the firmware, so only the rising edge is surfaced — add as an additional `action` value named `static`, alongside the primary action codes carried on `closuresDoorLock` attribute 0x0055.

vibration.agl002 does not implement Zigbee Poll Control cluster, there is zero traffic on cluster 0x0020. The device manages its own polling via the `report_interval` setting (manuSpecificLumi 0x01EC) sending 802.15.4 Data Requests to the coordinator at exactly that interval. `quirkCheckinInterval` is a no-op for the P100, therefore removed.

Added more information to settings descriptions
Added `<"manuSpecificLumi", ManuSpecificLumi>` to extends Changed attribute types to Zcl.DataType
Changed attribute types to `Zcl.DataType`

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -2482,7 +2482,6 @@ export const definitions: DefinitionWithExtend[] = [
                 percentageAttribute: 0x18,
             }),
             lumi.lumiModernExtend.lumiZigbeeOTA(),
-            lumi.lumiModernExtend.lumiStaticStateAction(),
             m.enumLookup<"manuSpecificLumi", ManuSpecificLumi>({
                 name: "device_mode",
                 cluster: "manuSpecificLumi",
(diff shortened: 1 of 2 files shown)


src/lib/lumi.ts
```diff
@@ -2428,6 +2428,18 @@ export const lumiModernExtend = {
             attribute: "presentValue",
             ...args,
         }),
+    lumiStaticStateAction: (): ModernExtend => {
```
Koenkk, line 2431: Can you use `m.actionEnumLookup` instead?

src/devices/lumi.ts
```diff
@@ -2474,130 +2474,138 @@ export const definitions: DefinitionWithExtend[] = [
         model: "DWZTCGQ11LM",
         vendor: "Aqara",
         description: "Multi-state sensor P100",
+        configure: async (device, coordinatorEndpoint) => {
```
Koenkk: Whats the goal of these? If the goal is to read the device state on pairing, this should already be covered by e.g. `m.enumLookup`.

src/lib/lumi.ts
absent42: It doesn't fire false, so after its first action it's stuck on true. It's on a different cluster to the other actions, so is included in the main action via extraActions: ["static"] - A second actionEnumLookup would create an entity that's always stuck on true. Does actionEnumLookup allow for multiple clusters and multiple attributes in a single lookup?

The device fires 0x01f3=true as a signal a 0x0055 motion-event has completed. It does not emit false. Incorperating it as extraActions: ['static'] joins the existing event stream, where it gets overwritten by the next motion event, matching Aqara app behaviour. Declaring it as a standalone second actionEnumLookup creates an entity whose state never changes after first fire.

Koenkk:
Thanks!

PR #12255, opened 2026-05-20
fix(sunricher): remove minimum PWM debug logs

niracler:
Removes stray debug logging from the Sunricher minimum PWM converter and fixes the configure warning to reference the minimum PWM attribute.


Koenkk:
Thanks!

PR #12256, opened 2026-05-20
fix: Hue: stop unwanted state reports

### LINKED ISSUE #32050: Hue light reports intermediate brightness values when a transition is not specified with firmware 1.145.2
### What happened?

Hue Aurelle Panel White Ambiance
929003597702

I updated the firmware from 1.122.8 / 20250213 to 1.145.2 / 20260209

With the new firmware version, the lamp is acting strangely.
When you turn it on or off, it now changes the brightness level with a slight delay.
Before the update, the Lamp never changed the brightness setting when turning it on or off.
However, the behavior is very inconsistent; sometimes it works normally, so the brightness control isn't adjusted as expected.
Is there any way to downgrade the lamp's firmware?

Image

The brightness control should not change when the device is turned on or off.

Turn the lamp on and off.

### Zigbee2MQTT version

2.10.1

### Adapter firmware version

20240716

### Adapter

SLZB-MR4U ZStack3x0 

### Setup

Zigbee2MQTT is running as a Docker container on Unraid 7.3.0
Host system: ASRock J4105-ITX (Intel Celeron J4105) with 32 GiB DDR4 RAM.
Architecture: x86_64.

_No response_

### Debug log

_No response_

### Notes

I've already deleted the lamp and reconnected it, but that didn't help either.

### LINKED ISSUE #32053: Multiple MQTT messages with different brightness values for same action (tried debounce) - Philips hue lamp
### What happened?

Hi,

I ran into an issue where it seems Z2M is sending multiple messages when I turn on/off one of my lights, more specifically a Philips Hue Infuse large ceiling lamp. I have, however seen the issue on at least 1 more light which is not of the same type (but not all my lights).

The messages contain different brightness values, which makes my Home Assistant brightness UI jump up or down, but it does not seem to affect the light itself. The messages can be several seconds apart, and I've tried with a debounce of 5s to see if that could resolve the issue, but the messages can be even longer than 5s apart. In the example below, I have not set a debounce, but the time between the last 2 messages is 8s.

**Turned on via Home Assistant (directly via the light entity):**
`[5/19/2026, 9:58:37 PM] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Bedroom 1st floor ceiling light', payload '{"brightness":160,"color":{"h":35,"hue":35,"s":74,"saturation":74,"x":0.4327,"y":0.4026},"color_mode":"color_temp","color_temp":326,"color_temp_startup":366,"effect":null,"linkquality":255,"power_on_behavior":null,"state":"ON","update":{"installed_version":16786434,"latest_release_notes":"Version: 36.4","latest_source":"https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/Hue/100B-011F-01002404-ConfLightBLE-ModuLumV3-EFR32MG21.zigbee","latest_version":16786436,"state":"available"},"update_available":true}'`

**Turned off via Home Assistant (directly via the light entity):**
`[5/19/2026, 9:58:39 PM] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Bedroom 1st floor ceiling light', payload '{"brightness":160,"color":{"h":35,"hue":35,"s":74,"saturation":74,"x":0.4327,"y":0.4026},"color_mode":"color_temp","color_temp":326,"color_temp_startup":366,"effect":null,"linkquality":255,"power_on_behavior":null,"state":"OFF","update":{"installed_version":16786434,"latest_release_notes":"Version: 36.4","latest_source":"https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/Hue/100B-011F-01002404-ConfLightBLE-ModuLumV3-EFR32MG21.zigbee","latest_version":16786436,"state":"available"},"update_available":true}'`

**Turned on via Home Assistant (directly via the light entity).**
This time the message has a different brightness value, although I did not touch the brightness - causing the UI to jump):
`[5/19/2026, 9:58:41 PM] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Bedroom 1st floor ceiling light', payload '{"brightness":138,"color":{"h":35,"hue":35,"s":74,"saturation":74,"x":0.4327,"y":0.4026},"color_mode":"color_temp","color_temp":326,"color_temp_startup":366,"effect":null,"linkquality":255,"power_on_behavior":null,"state":"ON","update":{"installed_version":16786434,"latest_release_notes":"Version: 36.4","latest_source":"https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/Hue/100B-011F-01002404-ConfLightBLE-ModuLumV3-EFR32MG21.zigbee","latest_version":16786436,"state":"available"},"update_available":true}'`

**??? - Brightness returned to the original value of 160**
`[5/19/2026, 9:58:49 PM] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Bedroom 1st floor ceiling light', payload '{"brightness":160,"color":{"h":35,"hue":35,"s":74,"saturation":74,"x":0.4327,"y":0.4026},"color_mode":"color_temp","color_temp":326,"color_temp_startup":366,"effect":null,"linkquality":255,"power_on_behavior":null,"state":"ON","update":{"installed_version":16786434,"latest_release_notes":"Version: 36.4","latest_source":"https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/Hue/100B-011F-01002404-ConfLightBLE-ModuLumV3-EFR32MG21.zigbee","latest_version":16786436,"state":"available"},"update_available":true}'`

I've thought about setting the debounce higher, but I feel this could potentially filter out the message containing the correct brightness values, leading to a significant time of desync between the UI and the actual value. Although I might be wrong here.

I've double checked that this light is not part of any group (both in Z2M and Home Assistant), and is not being influenced by any automations.
I did notice the "Retain"-checkbox was checked on the device, so I removed the message using MQTT Explorer and restarted both Home Assistant and Z2M with the "Retain"-checkbox unchecked. No messages has since been retained, but the problem still persists.

To not see multiple MQTT messages for a single on/off action.

Turn a light on/off and check the MQTT logs.

### Zigbee2MQTT version

2.10.1-dev commit 52201c96bf897af9bcad30435575923a0452b6dd

### Adapter firmware version

0x26720700

### Adapter

ConBee2

### Setup

Z2M in a separate docker container in Unraid with my Home Assistant instance running as a VM on the same machine.
MSI Cubi-N Mini-PC

{"id":10,"type":"Router","ieeeAddr":"0x0017880109c1ddf1","nwkAddr":61507,"manufId":4107,"manufName":"Philips","powerSource":"Mains (single phase)","modelId":"915005997401","epList":[11,242],"endpoints":{"11":{"profId":260,"epId":11,"devId":269,"inClusterList":[0,3,4,5,6,8,4096,64515,768,64513,64516],"outClusterList":[25],"clusters":{"genBasic":{"attributes":{"modelId":"915005997401","manufacturerName":"Philips","powerSource":1,"zclVersion":8,"appVersion":2,"stackVersion":1,"hwVersion":0,"dateCode":"20240902","swBuildId":"1.122.2"}},"lightingColorCtrl":{"attributes":{"colorCapabilities":31,"colorTempPhysicalMin":153,"colorTempPhysicalMax":500,"colorMode":2,"currentX":26522,"currentY":25664,"enhancedCurrentHue":9312,"currentSaturation":51,"colorTemperature":284,"startUpColorTemperature":366}},"genOnOff":{"attributes":{"onOff":0,"startUpOnOff":1}},"genLevelCtrl":{"attributes":{"currentLevel":160}}},"binds":[{"cluster":8,"type":"endpoint","deviceIeeeAddress":"0x00212effff054bb8","endpointID":1},{"cluster":6,"type":"endpoint","deviceIeeeAddress":"0x00212effff054bb8","endpointID":1}],"configuredReportings":[{"cluster":8,"attrId":0,"minRepIntval":5,"maxRepIntval":3600,"repChange":1},{"cluster":6,"attrId":0,"minRepIntval":0,"maxRepIntval":3600,"repChange":0}],"meta":{"onLevelSupported":false}},"242":{"profId":41440,"epId":242,"devId":97,"inClusterList":[],"outClusterList":[33],"clusters":{},"binds":[],"configuredReportings":[],"meta":{}}},"appVersion":2,"stackVersion":1,"hwVersion":0,"dateCode":"20240902","swBuildId":"1.122.2","zclVersion":8,"interviewCompleted":true,"interviewState":"SUCCESSFUL","meta":{"configured":"0.0.0"},"lastSeen":1779222211171}

### Debug log

I extracted messages from the log from when I clicked "Off" on the light and included a few seconds after the last message I could see in the frontend log. If more are required, please let me know. The original logs were so unweildy due to the sheer amount of text, but I can of course include the entire log files if needed.
[Log Z2M.txt](https://github.com/user-attachments/files/28068657/Log.Z2M.txt)

### Notes

The brightness value can also be correct when turned on, jump down to a lower value and then jump back up to the original value after some seconds.

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee2mqtt/issues/32050#issuecomment-4496461658
- fixes https://github.com/Koenkk/zigbee2mqtt/issues/32053

Removes `manuSpecificPhilips2` bind from Philips hue lights.

Before fw v1.145.2, the bind without reporting has no effect afaict. I don't see why it was added?

After fw v1.145.2, Hue lights may report their state through the bind, even if reporting is not enabled.
The lights report in-between states while transitioning (even without transition specified), which cause conflicts with the optimistic Z2M / HA / etc states.

I wonder if we want an explicit unbind for the already configured lights. Or for all of them, in case the bulb binds itself automatically?

### DIFF
--- a/src/lib/philips.ts
+++ b/src/lib/philips.ts
@@ -612,23 +612,12 @@ const philipsModernExtend = {
                     ),
                 );
             }
-            // Bind manuSpecificPhilips2 and register the Fz converter for all
-            // hueEffect/gradient devices, not just gradient. This enables state
-            // reports for effects, brightness, color, etc. on non-gradient bulbs too.
+            // Register the Fz converter for all devices (if user enables state reports later)
             result.fromZigbee.push(manuSpecificPhilips2Fz);
-            result.configure.push(async (device, coordinatorEndpoint, definition) => {
-                for (const ep of device.endpoints) {
-                    let supported = false;
-                    try {
-                        supported = ep.supportsInputCluster("manuSpecificPhilips2");
-                    } catch {
-                        // Custom cluster not in registry — skip
-                    }
-                    if (supported) {
-                        await ep.bind("manuSpecificPhilips2", coordinatorEndpoint);
-                    }
-                }
-            });
+            // Don't bind or configure reporting automatically - causes unwanted effects
+            // (reports in-between states, conflicting with optimistic states)
+            // https://github.com/Koenkk/zigbee2mqtt/issues/32050#issuecomment-4496461658
+
             // All Hue-specific effects per Bifrost spec
             effects.push("sunset", "sunrise", "sparkle", "opal", "glisten", "underwater", "cosmos", "sunbeam", "enchant");
             effects.push("none", "finish_effect", "stop_effect", "stop_hue_effect");
--- a/test/modernExtend.test.ts
+++ b/test/modernExtend.test.ts
@@ -285,7 +285,7 @@ describe("ModernExtend", () => {
 
                 "power_on_behavior",
             ],
-            bind: {1: ["manuSpecificPhilips2"]},
+            bind: {},
             read: {
                 1: [
                     ["lightingColorCtrl", ["colorCapabilities"]],


burmistrzak:
What happens when you unbind and start a Hue effect? 🤔

andrei-lazarov:
In my testing it works the same. The bind is light -> coordinator, so it should only affect that direction

burmistrzak:
Also, Zigbee devices can setup _default_ attribute reporting on their own after binding was successful.
I'm not sure that the default configuration gets fetched by Z2M after pairing.

@Nerivec Do we already do that and/or would it be feasible to implement?

burmistrzak:
@andrei-lazarov Interesting... My affected Hue bulb seems to ignore a longer minimum reporting interval for `manuSpecificPhilips2`!?

Nerivec:
Not automatically (would be a whole lot of requests...), but you can fetch it manually:
https://github.com/Koenkk/zigbee2mqtt/blob/94751abaacb66989e688f21d8ac3773cf5d754a9/lib/types/api.ts#L799-L813
It will automatically adjust the cached ZH data:
https://github.com/Koenkk/zigbee-herdsman/blob/a62a4d5d7f039c3b4de4bdaf2130a8d0bfa09a01/src/controller/model/endpoint.ts#L898

We only added this a while back.
There are [buttons in the frontend](https://nerivec.github.io/zigbee2mqtt-windfront/iframe.html?id=app--single-instance&viewMode=story&sb-api-urls=localhost%3A8080%2Fapi&sb-api-names=Main#/device/0/0x00123456789abcde/reporting) to sync existing binds (& a few buttons for special values), but I guess we should add a button to read/reset without needing any reporting present. I'll take a look.

Koenkk:
Thanks!

burmistrzak:
> but I guess we should add a button to read/reset without needing any reporting present. I'll take a look.

@Nerivec Yes, that's exactly what I meant! Would be super useful to get an idea of what devices setup on their end.

andrei-lazarov:
I think the next step is to prevent mixing optimistic behavior and reporting, to avoid further conflicts like this.

- when user adds reporting for attribute `x` -> automatically add `x` to `filtered_optimistic`
- when user delets reporting for attribute `x` -> automatically remove `x` from `filtered_optimistic`

What do you think?

burmistrzak:
@andrei-lazarov Hmm, I haven't had issues with attribute reporting and optimistic, AFAIK.
Any tips for me to reproduce?

andrei-lazarov:
Well it's what I showed in the Hue issue..

You start with low brightness, and send brightness 100% with long transition 10s.
**Z2M optimistically publishes 100%, then it drops back down,** because the bulb reports during the transition: 0, 10, 20, 30 etc.

You either disable bulb reporting, or disable the optimistic publishing (if you need the reporting), to avoid the spike in brightness

burmistrzak:
@andrei-lazarov I see. But without extra long transition time, it's not an issue, right?

This is a limitation of any combined UI. Same behavior with roller shutter.
What do we show the user? The target brightness/position/etc. or the current brightness/position/etc. in progress? 🤔🤔🤔

**Edit:** All smart home ecosystems (to my knowledge) combine target/desired state with current state in one control element. It's not a bad design necessarily, but dealing with actions that are not instantaneous (e.g. transitions, blinds, etc.) can be difficult.

Nerivec:
There's a similar issue with scene transition time. Need to somehow add better support for transitions overall. Tricky with so many devices making a mess between proper support and reporting details (yes/no & if yes what steps)...

andrei-lazarov:
> But without extra long transition time, it's not an issue, right?

It is, that's what the two users reported 😅 What part is confusing?
They observed one "random" brightness value reported by the bulb, after on/off with "no" transition.

I instructed them to disable reporting and I tried to prevent reporting being active by default with this PR

burmistrzak:
> It is, that's what the two users reported 😅 What part is confusing? They observed one "random" brightness value reported by the bulb, after on/off with "no" transition.

@andrei-lazarov Ah, you mean the delayed brightness reporting?
~~I had good success with setting `Min rep change` to `150` instead of `1`. Can you try that?~~

**Edit:** I don't know what I was thinking... 🙈 `150` doesn't make much sense and the "positive" effect I was seeing was simply Z2M polling the Hue lights. I've now disabled reporting across the board and let Z2M figure it out.

burmistrzak:
@andrei-lazarov You know what?

How about an option to filter incoming `manuSpecificPhilips2` reports? We could theoretically just drop the first brightness value we extract from the Hue native control state payload.

burmistrzak:
@andrei-lazarov FYI, after removing all Bindings and with just `manuSpecificPhilips2` freshly bound, I can no longer reproduce the issue!? No jumping brightness slider to be seen. WTF?

- LTA001 running `1.145.2/20260209/16788480`
- _Hue native control_ enabled
- _Optimistic_ enabled
- Default 0.4 sec. transition


https://github.com/user-attachments/assets/cf30d869-2b2d-4bda-a466-a09b2ccb6bc0

andrei-lazarov:
> drop the first brightness value

This is too wacky. It's fine as is: no reporting + optimistic, or reporting + non optimistic.

> `manuSpecificPhilips2` freshly bound, I can no longer reproduce the issue

Yeah, the bind is not supposed to do anything on its own. It's likely the unbind action that stops the hidden reporting.

I was thinking that **not binding** is not enough to prevent that.. And we actually need to force the **unbind**.
Can you test dev branch to confirm? Pair the affected bulb, confirm Z2M did not bind the cluster, check if you still have the issue. My bulbs don't seem to have the hidden reporting by default.

burmistrzak:
> Yeah, the bind is not supposed to do anything on its own. It's likely the unbind action that stops the hidden reporting.
> 
> I was thinking that **not binding** is not enough to prevent that.. And we actually need to force the **unbind**. Can you test dev branch to confirm? Pair the affected bulb, confirm Z2M did not bind the cluster, check if you still have the issue. My bulbs don't seem to have the hidden reporting by default.

@andrei-lazarov Yep, Wireshark confirms this. No invisible _default_ reporting for `manuSpecificPhilips2` (fresh binding still active!) after completely unbinding it once.

Screenshot 2026-05-21 at 23 43 13

I did a few on/off and used a directly bound Hue Tap Dial to dim the Hue bulb.
Only after I use the Tap Dial, Z2M asks for the `currentLevel` attribute.


I'll pull the latest `dev` and run my tests again. Just to be sure.

burmistrzak:
@Nerivec What do you think about polling `manuSpecificPhilips2.state` for supported lights instead of standard attributes?
This would allow us to fetch the state of multiple attributes with just a single `read` command.

PR #12258, opened 2026-05-20
fix(ZC-LS02): Add dedicated support for Moes roller blind motor instead of whitelabel

pranjal-joshi:
Link to picture pull request: TODO (Let's update if this works)

Created an external converter for this for the issue of 'Inverted' cover controls and battery percentage issues mentioned in [#27677](https://github.com/Koenkk/zigbee2mqtt/issues/27677)

### DIFF
--- a/src/devices/moes.ts
+++ b/src/devices/moes.ts
@@ -457,6 +457,35 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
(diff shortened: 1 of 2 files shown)


santoitaliano:
Don’t know if it’s appropriate to write here - how do I try this? Do I modify the custom converter I am using or remove it until it is added to update? I’m a little rusty.

Koenkk:
Thanks!

Koenkk:

PR #12260, opened 2026-05-20
fix(eglo): improve 99099 remote controller support

fpsal:
## What works
- All short press actions: `on`, `off`, `red`, `green`, `blue`, `refresh`, `refresh_colored`, `brightness_step_up`, `brightness_step_down`, `brightness_move_to_level`, `color_temperature_step_up`, `color_temperature_step_down`, `recall_1`, `recall_2`
- All long press actions: `red_long`, `green_long`, `blue_long`, `refresh_long`, `refresh_colored_long`, `recall_1_long`, `recall_2_long`
- Group buttons 1, 2, 3 — all actions carry an `action_group` value (1, 2, or 3) that can be used in Home Assistant automations to multiplex actions across groups
- The two favorite (heart) buttons — short press sends `recall_1`/`recall_2`, long press sends `recall_1_long`/`recall_2_long`
- Additional attributes: `action_level` (for `brightness_move_to_level`) and `action_color_temperature`
- Spurious `on` events that the remote automatically sends after color, color temperature, and refresh actions are suppressed within a 500ms window, preventing unintended light-on behavior in automations

## What doesn't work
- Long press on `on` and `off` buttons — these send no Zigbee command

## Firmware requirement
This implementation requires remote firmware version **3.0.1** or higher. The firmware can be updated using the **AwoX HomeControl app** (Android/iOS) — the update is described as *"Compatibility Fixes for 3rd party devices"*. Earlier firmware versions do not expose all actions correctly over Zigbee.

## Breaking Changes
- `action_group` is now reported as an integer 1/2/3 instead of the raw Zigbee group IDs 32778/32779/32780. Users who rely on the raw values in their automations will need to update them.
- `color_temperature_step_up` and `color_temperature_step_down` are now correctly mapped according to the ZCL specification. The remote was previously sending them inverted. Users who have worked around this by swapping the actions in their automations will need to revert that workaround.
- `brightness_move_up`, `brightness_move_down` and `brightness_stop` have been removed from the exposes as the remote never sends these commands.
- `yellow` has been removed from the exposes as the remote has no yellow button.

### DIFF
--- a/src/devices/eglo.ts
+++ b/src/devices/eglo.ts
@@ -1,10 +1,168 @@
 import * as fz from "../converters/fromZigbee";
 import * as exposes from "../lib/exposes";
 import * as m from "../lib/modernExtend";
-import type {DefinitionWithExtend} from "../lib/types";
+import * as globalStore from "../lib/store";
+import type {DefinitionWithExtend, Fz, KeyValueAny} from "../lib/types";
 
 const e = exposes.presets;
 const ea = exposes.access;
+
+const setActionGroup = (result: KeyValueAny | Promise<void> | void | undefined, groupID: number) => {
+    if (result && !(result instanceof Promise)) {
+        result.action_group = groupID - 32777;
+    }
+    return result;
+};
+
+const fzLocal = {
+    eglo99099_command_on: {
+        cluster: "genOnOff",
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12262, opened 2026-05-20
fix: Tuya MG-GPO04ZSLP: fix control

### LINKED ISSUE #11735: MG-GPO04ZSLP (_TZE200_oyti2ums): set commands fail with "Cannot read properties of undefined (reading 'ID')"
## Description
The MG-GPO04ZSLP (EYZEE Smart Power Point & Light Switch, Australian market) is recognised and state reports work correctly, but all set commands fail with:

`EventBus error 'Publish/mqttMessage': Cannot read properties of undefined (reading 'ID')`

## Root cause
The converter definition has two issues:

1. `master` and `light` are both mapped to endpoint `1` in `deviceEndpoints`, which is a conflict
2. Mixing `tuyaBase({ dp: true })` with `deviceEndpoints` causes the set handler to attempt endpoint ID resolution which fails
```js
extend: [tuya.modernExtend.tuyaBase({ dp: true }), m.deviceEndpoints({ endpoints: { master: 1, light: 1, left: 2, right: 3 } })],
```

The DP table itself appears correct — incoming reports work fine for all four endpoints.

## Device info
- Model: MG-GPO04ZSLP
- Manufacturer: _TZE200_oyti2ums
- zigbee-herdsman-converters version: 26.12.0
- zigbee2mqtt version: 2.9.1

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/11735

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -5564,7 +5564,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "MG-GPO04ZSLP",
         vendor: "Tuya",
         description: "2 x socket + 1 x light with master switch and metering",
-        extend: [tuya.modernExtend.tuyaBase({dp: true}), m.deviceEndpoints({endpoints: {master: 1, light: 1, left: 2, right: 3}})],
+        extend: [tuya.modernExtend.tuyaBase({dp: true}), m.deviceEndpoints({endpoints: {master: 1, light: 1, left: 1, right: 1}})],
         exposes: [
             e.switch().withEndpoint("master").setAccess("state", ea.STATE_SET).withDescription("Master switch controlling all relays"),
             e.switch().withEndpoint("light").setAccess("state", ea.STATE_SET).withDescription("Light relay"),


src/devices/tuya.ts
```diff
@@ -5564,12 +5564,12 @@ export const definitions: DefinitionWithExtend[] = [
         model: "MG-GPO04ZSLP",
         vendor: "Tuya",
         description: "2 x socket + 1 x light with master switch and metering",
-        extend: [tuya.modernExtend.tuyaBase({dp: true}), m.deviceEndpoints({endpoints: {master: 1, light: 1, left: 2, right: 3}})],
+        extend: [tuya.modernExtend.tuyaBase({dp: true})],
```
```suggestion
        extend: [tuya.modernExtend.tuyaBase({dp: true}), m.deviceEndpoints({endpoints: {master: 1, light: 1, left: 1, right: 1}})],
```

I think this should be enough, `e.switch()` below should be kept. (will be breaking otherwise)

andrei-lazarov: I see this also works! Though can you explain what was the breaking part? Chaning `switch` to `binary`?
The states have the same name `"state_master"`, and the possible values are still `"ON"` and `"OFF"`

Koenkk: I though that binary would be exposed as binary to HA, but this isn't the case (binary with set access are also exposed as [switch](https://github.com/Koenkk/zigbee2mqtt/blob/94751abaacb66989e688f21d8ac3773cf5d754a9/lib/extension/homeassistant.ts#L1016)), so not breaking (although it's cleaner to use `switch` for switches)

Koenkk:
Thanks!

PR #12267, opened 2026-05-21
Fix system mode and off mode display for Acova Alcantara 2 and Alcantara 3

DeadlockState:
This PR permit to fix :

- Switching to Off Mode on a Thermostat card is displayed as Idle in Home Assistant (system_mode: off/hvac_action: idle)
- Switching back to Heat Mode on a Thermostat card the reset the radiator to 7°C (minimum setpoint) instead of restoring the previous BOX/Zigbee managed mode

Tested on :

- Acova Alcantara 2 (I suppose it's exactly the same for Acova Alcantara 3)

Objective :

- Reproduce correct behavior like on ZHA after my migration from ZHA to Zigbee2MQTT

### DIFF
--- a/src/devices/acova.ts
+++ b/src/devices/acova.ts
@@ -1,11 +1,31 @@
 import * as fz from "../converters/fromZigbee";
 import * as tz from "../converters/toZigbee";
+import * as constants from "../lib/constants";
 import * as exposes from "../lib/exposes";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
-import type {DefinitionWithExtend} from "../lib/types";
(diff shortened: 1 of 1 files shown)


Koenkk:
To prevent code duplication, you can call the original converter and apply the changes after that, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/273afa3f126fa6d778d89684ce683aecbfcd0a45/src/devices/develco.ts#L162

DeadlockState:
Fixed

Koenkk:
thanks, could you fix the failing CI pipeline?

DeadlockState:
Fixed

Koenkk:
Thanks!

PR #12271, opened 2026-05-21
feat: Tuya TS0601_cover_8: add _TZE28C1000000_alh14edn whiteLabel

akc-kyiv:
Adds support for Tuya roller blind motor with manufacturerName `_TZE28C1000000_alh14edn` (box label: "Smart Blinds Controller", no brand).

Per review feedback, folded into `TS0601_cover_8` rather than a new standalone definition. Exposed as a whiteLabel (`TS0601_alh14edn`) so it gets its own device page + image.

### DP test results on real hardware

| DP | Feature | Result |
|---|---|---|
| 1 | `state` (OPEN/STOP/CLOSE) | works |
| 2 / 3 | `position` (coverPositionInverted) | works, reports back at end-of-motion |
| 5 | `reverse_direction` (enum forward/back) | works — enum and boolean encode identically on the wire |
| 12 | `motor_fault` | silent — motor never reports |
| 103 | `upper_stroke_limit` (SET/RESET) | silent — no physical effect |
| 104 | `middle_stroke_limit` (SET/RESET) | silent |
| 105 | `lower_stroke_limit` (SET/RESET) | silent |
| 106 | `motor_working_mode` | write delivered, no observable difference |

The four silent features are gated off for `_TZE28C1000000_alh14edn` via the `exposes: (device) => …` conditional pattern.

Bundled 3-button remote is 433 MHz RF only, no Zigbee traffic. State reports flow back over Zigbee for any trigger source (Zigbee command, RF remote, motor body buttons) — but only once top/bottom limits are calibrated; uncalibrated motor stays silent.

Image PRs: Koenkk/zigbee2mqtt.io#5149 (original image, merged) and Koenkk/zigbee2mqtt.io#5158 (rename to match the new whiteLabel model).

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -23462,6 +23462,25 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE28C1000000_alh14edn"]),
+        model: "TS0601_cover_14",
+        vendor: "Tuya",
+        description: "Roller blind motor",
+        extend: [tuya.modernExtend.tuyaBase({dp: true})],
+        exposes: [
+            e.cover_position().setAccess("position", ea.STATE_SET),
+            e.enum("motor_direction", ea.STATE_SET, ["normal", "reversed"]).withDescription("Set the motor direction"),
+        ],
+        meta: {
+            tuyaDatapoints: [
+                [1, "state", tuya.valueConverterBasic.lookup({OPEN: tuya.enum(0), STOP: tuya.enum(1), CLOSE: tuya.enum(2)})],
+                [2, "position", tuya.valueConverter.coverPositionInverted],
+                [3, "position", tuya.valueConverter.coverPositionInverted],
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -23461,6 +23461,25 @@ export const definitions: DefinitionWithExtend[] = [
             ],
         },
     },
+    {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE28C1000000_alh14edn"]),
```
Koenkk: Could you check if it also works with the TS0601_cover_8 definition? If the additional exposed features don't work, make them conditional like this: https://github.com/Koenkk/zigbee-herdsman-converters/blob/24bb7d6131f26a2753484ea3b053274af2fe107b/src/devices/tuya.ts#L2629

akc-kyiv:
Done — refactored into `TS0601_cover_8` with conditional exposes for our manufacturerName, and a `TS0601_alh14edn` whiteLabel so the device still gets its own page + image. Test results inline in the PR body. Image-rename PR: Koenkk/zigbee2mqtt.io#5158.

Koenkk:
thanks!

PR #12276, opened 2026-05-22
Add missing reporting for hive SLR1d

si458:
https://community.home-assistant.io/t/hive-receiver-slr1d-does-not-publish-state-changes-to-z2mqtt/801090

### DIFF
--- a/src/devices/hive.ts
+++ b/src/devices/hive.ts
@@ -372,7 +372,7 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.thermostatOccupiedHeatingSetpoint(endpoint);
             await reporting.thermostatTemperatureSetpointHold(endpoint);
             await reporting.thermostatTemperatureSetpointHoldDuration(endpoint);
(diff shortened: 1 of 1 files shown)


src/devices/hive.ts
```diff
@@ -361,6 +361,18 @@ export const definitions: DefinitionWithExtend[] = [
                         " used. 0 to 360 to match the remote display",
                 ),
         ],
+        meta: {disableDefaultResponse: true},
```
Koenkk: Why is this needed?

si458: I'm not 100% sure?
I just copied what the other SLR1c, SLR1b, SLR1 have listed and it works perfectly fine here for me.
Happy to remove it if u need me too?

Koenkk: If you didn't notice that anything it's not working, then this is not needed, thanks for the PR!

si458:
Thabns for the merge! ❤️

PR #12278, opened 2026-05-22
fix: Philips 9290022169: Fix color temp range

andrei-lazarov:
Confirmed on `LTA001` latest fw `1.145.2` `20260209`.
With `undefined` it wrongly defaults to `[150,500]`


Koenkk:
thanks!

PR #12283, opened 2026-05-23
feat(moes): add countdown + battery_state + error_status to ZWV-YC water valve

FranzFelberer:
## Summary

The Moes ZWV-YC (TS0049) currently exposes only `state` (switch) + `battery_percentage` via `m.onOff()` + `m.battery()`. The same TS0049 hardware family is supported with a full datapoint mapping (countdown / battery_state / error_status) via the generic `TS0049` "Water valve" definition in `src/devices/tuya.ts`. The MOES-specific definition was missing those mappings.

This PR replaces the minimal `extend` with the full `tuya.modernExtend.tuyaBase` + `tuyaDatapoints` configuration from the generic TS0049 definition, exposing:

| Datapoint | Name | Type |
|---|---|---|
| 26 | `error_status` | raw |
| 101 | `state` | onOff |
| 111 | `countdown` (0-255 min) | raw |
| 115 | `battery_state` | batteryState |

## Why this matters

Without the `countdown` datapoint, users can only schedule irrigation via Home Assistant-side `switch.turn_on` → `delay(N)` → `switch.turn_off`. If HA crashes mid-cycle, the valve stays open and floods the planters. The hardware's native countdown auto-off (used by the generic TS0049 definition) is much safer for irrigation automations.

## Verification

Datapoint layout taken verbatim from the existing generic `TS0049` "Water valve" definition in `src/devices/tuya.ts`. Verified working on `_TZ3000_cjfmu5he` in a personal balcony irrigation setup. Other fingerprints (`_TZ3000_mq4wujmp`, `_TZ3000_5af5r192`, `_TZ3000_ogjpfoyn`) are unchanged in the fingerprint list and presumed identical hardware; please flag if anyone reports otherwise.

`npx tsc --noEmit` and `npm run build` (which runs the indexer) both pass locally.

### DIFF
--- a/src/devices/moes.ts
+++ b/src/devices/moes.ts
@@ -1376,7 +1376,22 @@ export const definitions: DefinitionWithExtend[] = [
         model: "ZWV-YC",
         vendor: "Moes",
         description: "Water valve",
-        extend: [m.battery(), m.onOff({powerOnBehavior: false, configureReporting: true})],
+        extend: [tuya.modernExtend.tuyaBase({dp: true, forceTimeUpdates: true})],
+        exposes: [
+            tuya.exposes.errorStatus(),
+            tuya.exposes.switch(),
+            tuya.exposes.batteryState(),
+            tuya.exposes.countdown().withValueMin(0).withValueMax(255).withUnit("minutes").withDescription("Max on time in minutes"),
+        ],
+        meta: {
+            tuyaSendCommand: "sendData",
+            tuyaDatapoints: [
+                [26, "error_status", tuya.valueConverter.raw],
+                [101, "state", tuya.valueConverter.onOff],
+                [111, "countdown", tuya.valueConverter.raw],
+                [115, "battery_state", tuya.valueConverter.batteryState],
+            ],
+        },
     },
     {
         fingerprint: tuya.fingerprint("TS0011", ["_TZ3000_hhiodade"]),


Koenkk:
Thanks!

Luk289:
Hi @FranzFelberer and @Koenkk,

Unfortunately, this change breaks the core functionality of the Moes ZWV-YC water valve. 

After upgrading to the version containing this PR, the valve exposes the `countdown` feature, but completely stops responding to remote ON/OFF commands from Zigbee2MQTT. The commands are published to MQTT, but the hardware ignores them. 

Rolling back to the previous version of Zigbee2MQTT (without this definition change) instantly restores full remote functionality and the valve works perfectly again. It seems that exposing these datapoints causes the valve's MCU to reject or freeze on downstream commands.

FranzFelberer:
Thanks for the detailed report @Luk289, and for confirming that a rollback restores ON/OFF — that pins it down.

Root cause: #12283 moved the ZWV-YC switch from the standard `genOnOff` path (`m.onOff()`) onto the Tuya datapoint path, mapping `state` to **DP 101**. On my unit (`_TZ3000_cjfmu5he`) that works, but your valve evidently doesn't accept `state` on DP 101, so the command is published and then ignored. I carried the other fingerprints over assuming identical hardware — that assumption was wrong, and I'm sorry for the breakage.

@Koenkk — please go ahead and **revert #12283** to unbreak affected users. I'll resubmit it gated to the verified fingerprint (`_TZ3000_cjfmu5he`) only, so the `countdown` / `battery_state` / `error_status` datapoints apply solely to confirmed hardware and all other fingerprints stay on `m.onOff()`.

@Luk289 — if you can share your valve's manufacturer fingerprint (`_TZ3000_…` from the Z2M device page → *About*), I'll make sure it's excluded from the Tuya-DP definition in the resubmission.

Luk289:
Thanks for the quick response and for requesting the revert @FranzFelberer! 

Here is the manufacturer fingerprint for my valve from the Z2M About page:

_TZ3000_ogjpfoyn

Really appreciate your help in sorting this out so quickly.

lostinhaos:
Hello,

I have the same issue and my device's fingerprint is _TZ3000_5af5r192. @FranzFelberer, please include this when addressing the issue. 

Thank you!

PR #12307, opened 2026-05-26
Move Tuya specific converters from fromZigbee.ts and toZigbee.ts

svhelge:
Moved Tuya-related converters from the generic `fromZigbee.ts` and `toZigbee.ts` files into the Tuya-specific library `(lib/tuya.ts`) and device file (`devices/tuya.ts`).
`fz.tuya_doorbell_button` renamed to `HG06668_doorbell_button` and moved to `devices/lidl.ts`.
Devices updated with new references for converters.


Koenkk:
Thanks!

PR #12310, opened 2026-05-27
fix: Lonsonho ZB-RGBCW: enable reporting by default

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee2mqtt/issues/31530


Koenkk:
Bumped the `version` to trigger a re-configure, thanks!

PR #12318, opened 2026-05-27
Add ESP32-C6 irrigation controller

mateuszsury:
Adds support for the ESP32-C6 Smart Irrigation Controller firmware.\n\nSupported exposes:\n- line_1, line_2, line_3 On/Off endpoints\n- per-line duration in seconds\n- per-line schedule enabled flag\n- per-line schedule start minute, 0..1439 min\n- per-line weekday switches\n- global interlock\n\nValidation run locally:\n- corepack pnpm exec biome check src/devices/mateusz_sury.ts src/devices/index.ts --error-on-warnings\n- corepack pnpm exec tsc --noEmit false --pretty false\n- corepack pnpm run build\n- corepack pnpm test

### DIFF
--- a/src/devices/index.ts
+++ b/src/devices/index.ts
@@ -199,6 +199,7 @@ import {definitions as lytko} from "./lytko";
 import {definitions as mElec} from "./m_elec";
 import {definitions as makegood} from "./makegood";
 import {definitions as matcallBv} from "./matcall_bv";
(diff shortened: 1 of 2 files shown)


Koenkk:
Is this is a custom DiY devices? If yes then this should be kept as an external converter (so no code merged into this repo)

PR #12321, opened 2026-05-28
Fixed param 2026052801

iotRobot:
1. Added auxiliary functions localTempCalibration4 and localTempCalibration5 for HA to issue scaled 100-fold int32 and original int32 functions; 
2. Fixed the issue of negative values when sending calibration values for temperature, humidity, and lighting.


Koenkk:
Thanks!

PR #12323, opened 2026-05-28
Update slv.ts

yphuang2009:


src/devices/slv.ts
```diff
@@ -1,6 +1,16 @@
 import * as m from "../lib/modernExtend";
 import type {DefinitionWithExtend} from "../lib/types";
 
+import * as m from 'zigbee-herdsman-converters/lib/modernExtend';
+
+export default {
```
Koenkk, line 6: Please move too `bilisity.ts` and add it to `devices/index.ts`

PR #12331, opened 2026-05-28
Remove unwanted local_temperature exposed property for Acova Alcantara 2 and Alcantara 3

DeadlockState:
This PR permit to fix :

- Remove the false Local temperature (local_temperature) exposed property (same as Occupied heating setpoint) because the radiator cannot return mesured current temperature
- Remove the current_temperature state attribute from Home Assistant and the "Current temperature" information displayed on Thermostat card

Tested on :

- Acova Alcantara 2 (I suppose it's exactly the same for Acova Alcantara 3)

Objective :

- Reproduce correct behavior like on ZHA after my migration from ZHA to Zigbee2MQTT

### DIFF
--- a/src/devices/acova.ts
+++ b/src/devices/acova.ts
@@ -24,6 +24,7 @@ const acova = {
                         result.hvac_action = "off";
                     }
                 }
+                delete result.local_temperature;
                 return result;
             },
         } satisfies Fz.Converter<"hvacThermostat", undefined, ["attributeReport", "readResponse"]>,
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Koenkk:
I reverted this because it made CI test fail: https://github.com/Koenkk/zigbee2mqtt/actions/runs/26655471358/job/78564370531?pr=32136

The `local_temperature` property is mandatory for climate exposes: https://github.com/Koenkk/zigbee2mqtt/blob/8beaf488f7fcb88471cb663b489edca8f0f38630/lib/extension/homeassistant.ts#L621, can you think of a different solution?

VincentLeDall:
Local_temperature is very usefull on Acova Alcantara3. It gives to the ambient temperature measured by the device. Is it possible to re enable it on Alcantara3 device?

PR #12333, opened 2026-05-29
Update Namron device models and description

xses79:
Updated zigbeeModel and model for Namron devices. Changed description to 'Namron Edge Dimmer'.


Koenkk:
Thanks!

PR #12335, opened 2026-05-29
Update nodon.ts: Add  power apparent exposed by 'SEM-4-1-00'

Paulo-89:
Add apparent power measurement for the Nodon ‘SEM-4-1-00’ device,  which is not currently exposed by Zigbee2MQTT.

Tested with 2  'SEM-4-1-00' devices running with firmware version 003.00.09


Koenkk:
Thanks!

PR #12339, opened 2026-05-29
Update data points _TZE284_uqfph8ah

pvela2017:
Hello,

I updated the data points for the **TS0601_TZE284_uqfph8ah** using a tuya gateway. Now "Quick calibration" and "Indicator mode" work. Although the quick calibration value is only updated when the automatic calibration routine is ran (it works the same on the tuya app, if I try to manually change the quick calibration value on the app it fails).

I dont think it will fix this issue [BSEED TS0601 _TZE284_uqfph8ah curtain/blind switch
 #29786](https://github.com/koenkk/zigbee2mqtt/issues/29786), but it is realated.

Screenshot from 2026-05-29 13-19-04

Screenshot from 2026-05-29 13-37-15

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -14885,7 +14885,7 @@ export const definitions: DefinitionWithExtend[] = [
         whiteLabel: [tuya.whitelabel("Homeetec", "37022714", "4 Gang switch with backlight", ["_TZE200_hewlydpz"])],
     },
     {
-        fingerprint: tuya.fingerprint("TS0601", ["_TZE200_p6vz3wzt", "_TZE204_p6vz3wzt", "_TZE284_uqfph8ah", "_TZE284_waa352qv"]),
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE200_p6vz3wzt", "_TZE204_p6vz3wzt", "_TZE284_waa352qv"]),
         model: "TS0601_cover_5",
         vendor: "Tuya",
         description: "Curtain/blind switch",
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -23607,6 +23604,64 @@ export const definitions: DefinitionWithExtend[] = [
...
         },
     },
+    {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_uqfph8ah"]),
+        model: "TS0601_cover_14",
+        vendor: "Tuya",
+        description: "Curtain/blind switch",
+        options: [exposes.options.invert_cover()],
+        extend: [tuya.modernExtend.tuyaBase({dp: true})],
+        exposes: [
```
Koenkk: Please keep in the existing definition and make the exposes conditional, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/24bb7d6131f26a2753484ea3b053274af2fe107b/src/devices/tuya.ts#L2629

pvela2017: > Please keep in the existing definition and make the exposes conditional, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/24bb7d6131f26a2753484ea3b053274af2fe107b/src/devices/tuya.ts#L2629

Ok, I updated the expose field to conditional.

Koenkk:
thanks!

lostinhaos:
Hello, 

I have a device with signature _TZE284_mxujdmxo which is a ME202WZwater level sensor (https://www.zigbee2mqtt.io/devices/ME202WZ.html). 

After update to version [2.11.0](https://github.com/Koenkk/zigbee2mqtt/releases/tag/2.11.0), the max liquid depth is in mm instead of m and has a range from 10-4000.
However the device interprets the input still in m for max liquid depth, which renders the % fill unusable (min is therefore 10m).

Thanks for checking!

PR #12346, opened 2026-05-31
Add GIEX GX06 soil sensor brightness level

Giexperience:
Move _TZE284_3urschql out of the generic TS0601_soil_3 definition and add a dedicated GIEX GX06 definition.

This variant has the same temperature, soil moisture, battery and temperature unit datapoints as the existing soil sensor group, but also reports DP 2 as a brightness level enum.

<!--
Thank you for your contribution!

If you are adding support for a new device, please also submit a picture for the documentation.
Tutorial: https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation
The line below can be removed if you are NOT adding support for a new device.


src/devices/tuya.ts
```diff
@@ -4311,10 +4310,42 @@ export const definitions: DefinitionWithExtend[] = [
                 [15, "battery", tuya.valueConverter.raw],
             ],
         },
-        whiteLabel: [
-            tuya.whitelabel("GIEX", "GX04", "Soil Moisture Sensor", ["_TZE284_nhgdf6qr"]),
-            tuya.whitelabel("GIEX", "GX06", "Soil Moisture Sensor", ["_TZE284_3urschql"]),
+        whiteLabel: [tuya.whitelabel("GIEX", "GX04", "Soil Moisture Sensor", ["_TZE284_nhgdf6qr"])],
+    },
+    {
+        fingerprint: tuya.fingerprint("TS0601", ["_TZE284_3urschql"]),
```
Koenkk: Please keep in the original definition by using a [conditional expose](https://github.com/Koenkk/zigbee-herdsman-converters/blob/24bb7d6131f26a2753484ea3b053274af2fe107b/src/devices/tuya.ts#L2629)

Giexperience:
Updated. I kept _TZE284_3urschql in the original TS0601_soil_3 definition and added brightness_level as a conditional expose.

Koenkk:
Thanks!

PR #12349, opened 2026-05-31
Add Tuya _TZE200_2jwrgrro curtain motor to TS0601_cover_1

xuw:
Adds the **Tuya TS0601 / `_TZE200_2jwrgrro`** curtain/roller-blind motor, currently *Not supported*.

It is a standard Tuya curtain motor using the usual `state` (open/stop/close) + `position` datapoints, so it is added to the existing generic **`TS0601_cover_1`** definition (alongside `_TZE200_swlgvdlh` etc.).

- Validated on **9 units running in production** (previously via an external converter with the same state/position datapoints).
- If position runs inverted on a given unit, the existing `cover_position_percent_fix` option handles it.
- `pnpm check` (biome), `pnpm build` (tsc) and `pnpm test` (561 tests) pass locally.


Koenkk:
Thanks!

PR #12352, opened 2026-05-31
fix: Fallback to `transition` device option with `hue_native_control`

### LINKED ISSUE #32083: Transition not working in Hue Native Mode
### What happened?

When using Hue bulbs in native mode, Z2M transition specific setting is not being honored. I'm currently on Z2M 2.10.1, and it seems there was a similar issue related to that (https://github.com/Koenkk/zigbee2mqtt/issues/31838), but it's still not working here, I've tried setting it to 0.2, 10, 5, and it seems to only use the default 0.5 from the bulbs itself. When not using the native mode, the transition work as expected.

When using Home Assistant, and explicitly setting the transition time in the automation itself, for example, it's indeed working in Hue Native Mode, but the setting under the Z2M device itself, that sets the default transition time for the bulb under `settings (specific)` is the one not taken effect.

_No response_

_No response_

### Zigbee2MQTT version

2.10.1

### Adapter firmware version

20250321

### Adapter

ZStack3x0

### Setup

Z2M with Home Assistant, but in this case I'm sending the commands directly from Z2M itself.

_No response_

### Debug log

_No response_

### Notes

_No response_

burmistrzak:
With `hue_native_control` enabled, fallback to the `transition` device option did not work.

This _should_ fix https://github.com/Koenkk/zigbee2mqtt/issues/32083

### DIFF
--- a/src/lib/philips.ts
+++ b/src/lib/philips.ts
@@ -11,7 +11,7 @@ import * as modernExtend from "./modernExtend";
 import * as globalStore from "./store";
 import type {Configure, Expose, Fz, KeyValue, KeyValueAny, ModernExtend, Tz} from "./types";
 import * as utils from "./utils";
-import {determineEndpoint, exposeEndpoints, hasAlreadyProcessedMessage, isObject, numberWithinRange} from "./utils";
+import {determineEndpoint, exposeEndpoints, hasAlreadyProcessedMessage, isObject, numberWithinRange, toNumber} from "./utils";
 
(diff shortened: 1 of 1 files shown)


burmistrzak:
@drauziooppenheimer Would you be able to test this PR on your end?

Koenkk:
Thanks!

drauziooppenheimer:
Thanks so much for all the work on this!!

PR #12011, opened 2026-04-23
Shelly 2PM Gen4 (Switch mode) - firmware 2.0.0-beta1

SebiGo:
expose switch inputs for 2PM Gen4 (Switch mode) new firmware - support for detached mode
Add fingerprint for firmware with endpoints 3/4 (switch inputs) and 5, and converters to expose input_1/input_2 on/off/toggle/hold actions.
OTA enabled, but for now there are no firmware images in the repository.

See https://github.com/Koenkk/zigbee2mqtt/issues/30637

PR #12354, opened 2026-06-01
Fix Shelly 2PM Gen4 (Cover mode) fingerprint for firmware 2.0.0-beta1

hoellen:
PR #12011 added support for the Shelly 2PM Gen4 with firmware 2.0.0-beta1. However, in cover mode fingerprint and configuration don't match the correct clusters.
This caused the device to show as "unsupported" and the `configure` step to fail with `UNSUPPORTED_CLUSTER` when trying to read `genOnOffSwitchCfg` on EP 4 (which has no input clusters in cover mode).
This PR fixes the fingerprint for cover mode which has the clusters shifted by `-1` (same as the 1PM #12059).


Koenkk:
Thanks!

PR #12360, opened 2026-06-01
Move various device specific converters from fromZigbee.ts and toZigbee.ts

svhelge:
Move devicce specific converters to local files.


Koenkk:
Thanks!

PR #12362, opened 2026-06-02
fix: add heiman private attribute for HS1RM

Leo2442926161:
dear Z2M team,
add heiman private attribute for HS1RM


Koenkk:
Thanks!

PR #12365, opened 2026-06-02
Сorrection of measurement units

smartboxchannel:
Сorrection of measurement units for the pressure offset property (PST V1, PST V1 LR,  PST POW V1)


Koenkk:
Thanks!

PR #12368, opened 2026-06-03
add _TZE2841000000_0ints6wl and _TZE2841000000_hdml1aav

iotRobot:
Due to an error on Tuya’s side during module firmware upgrade, the string "1000000" was inadvertently appended to the Zigbee Model parameter, resulting in recognition failure of newly manufactured products on the HA platform.


Koenkk:
Thanks!

PR #12374, opened 2026-06-04
fix(tuya): add _TZE284_sdvbnmj5 to Novato ZPV-01

bilgi-source:
Adds the new Tuya manufacturer name `_TZE284_sdvbnmj5` to the existing Novato ZPV-01 (battery powered smart valve) definition.

Newer production batches of the ZPV-01 ship with a new Tuya chip that reports `TS0601` / `_TZE284_sdvbnmj5` and is currently detected as unsupported. The device uses the same datapoints as the existing `_TZE204_dsagrkvg` / `_TZE284_zm8zpwas` variants (DP1 state, DP8 valve_state, DP101 battery), so only the fingerprint is added.

Not a new device, so no picture PR is needed.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -19350,7 +19350,7 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12375, opened 2026-06-04
fix(legrand): improve K4027C moving state detection

### LINKED ISSUE #12152: Legrand Bticino K4027C not send state MOVING

managerfx:
Closes #12152


src/lib/legrand.ts
```diff
@@ -302,6 +303,45 @@ export const tzLegrand = {
             return {state: {auto_mode: value}};
         },
     } satisfies Tz.Converter,
+    cover_state_with_moving: {
```
Koenkk, line 306: Instead of creating completely new converters, call the existing one and only add the `moving` attribute, example: https://github.com/search?q=repo%3AKoenkk%2Fzigbee-herdsman-converters%20tz.convertset&type=code

managerfx: I now call the existing

src/lib/legrand.ts
```diff
@@ -302,6 +303,45 @@ export const tzLegrand = {
...
+        key: ["state"],
+        convertSet: async (entity, key, value, meta) => {
+            // cover_state.convertSet returns void; call it for the side-effect only
+            await tz.cover_state.convertSet?.(entity, key, value, meta);
```
Koenkk, line 310:
```suggestion
            const result = await tz.cover_state.convertSet?.(entity, key, value, meta);
```

And then do e.g. 

```
return {state: {...result.state, moving: true}};
```

src/lib/legrand.ts
Koenkk, line 306: I don't see the converters used by any devices?

PR #12377, opened 2026-06-05
[SNZB-04PR2] Fix batteryVoltage reporting

N3rdix:
## Description
Fixes a recurring `UNSUPPORTED_ATTRIBUTE` error on the **SONOFF SNZB-04PR2** (SenseGuard DW Gen2) door contact sensor.

## Motivation
The SONOFF SNZB-04PR2 should natively support and report battery voltage, but it seems it rejects active writing of reporting intervals for `batteryVoltage`. 

Currently it results in those errors:
`[2026-06-04 19:06:14] error: 	z2m: Failed to configure 'Türkontakt Haustür', attempt 1 (Error: ZCL command 0xa4c13811cb60ffff/1 genPowerCfg.configReport([{"minimumReportInterval":3600,"maximumReportInterval":7200,"reportableChange":100,"attribute":"batteryVoltage"}], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"reservedBits":0,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')`

By setting `voltageReportingConfig: false` within `m.battery()`, we explicitly prevent Zigbee2MQTT from sending the unsupported `configReport` command while preserving the voltage data point layout and the passive reading/reporting of battery values.

## Checklist
- [x] The change fixes the recurring background configuration timeout / attribute error.
- [x] Battery percentage and millivolt data points are still correctly displayed.
- [x] Tested locally using an external converter with identical parameters.

### Result
Tested with a partially drained battery and the voltage dropped from 3200 to 3000, so it works as expected 😄 
image

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -6095,6 +6095,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SNZB-04PR2",
         vendor: "SONOFF",
         description: "Contact sensor",
+        version: "0.0.1",
         extend: [
             m.iasZoneAlarm({zoneType: "contact", zoneAttributes: ["alarm_1", "battery_low"]}),
             m.binary({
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12378, opened 2026-06-05
Add configured regions state for Aqara FP1

mayongze:
## Summary
- add a `configured_regions` state expose for the Aqara FP1 (`RTCZCGQ11LM`)
- return updated `configured_regions` after successful `region_upsert` / `region_delete` writes
- keep the value as a readable summary of regions written through Zigbee2MQTT, e.g. `1: y1=x1-4; y2=x1-4 (8 zones)`

This is intentionally a Zigbee2MQTT-side cache for display purposes. It does not try to read or decode region definitions that were configured by another gateway or before this state existed.

## Tests
- `pnpm exec biome check --write src/lib/lumi.ts src/devices/lumi.ts test/lumi.test.ts`
- `pnpm test -- test/lumi.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm run build`
- commit hook: `pnpm run build`, `pnpm run check`, `pnpm run test`

## Local verification
Hot-tested on a Zigbee2MQTT add-on with an RTCZCGQ11LM. After writing regions 1/2/3, `state.json` contains a single `configured_regions` summary and no separate `configured_regions_summary` field.

image

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -1897,6 +1897,12 @@ export const definitions: DefinitionWithExtend[] = [
                         '"enter" / "leave" events are usually triggered first, followed by "occupied" / "unoccupied" after a couple of seconds.',
                 ),
             e
+                .text("configured_regions", ea.STATE)
+                .withDescription(
+                    "Region definitions written through Zigbee2MQTT. " +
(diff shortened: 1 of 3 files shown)


Koenkk:
Thanks!

PR #12382, opened 2026-06-05
add Yoolax Top-Down Bottom-Up external converter

Sinaloit:
This external converter addresses https://github.com/Koenkk/zigbee2mqtt/issues/32148
As noted, the external converter works for almost all operations except for the top state control (open/close/stop). The position for top does not reset to 0 when changing the bottom value as expected but the position for bottom does reset to 0 when changing the top, I believe this is since dp3 is consumed by 2 different endpoints its only getting read by the first and not the second, unsure how to resolve.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -5412,6 +5412,57 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
+        fingerprint: tuya.fingerprint("TS0301", ["_TZE200_eatmkx5j"]),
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `TS0301_dual_rail_2.png`

PR #12385, opened 2026-06-06
# Add support for Namron Zigbee Edge Thermostat (4566702/4566703/4512783/4512784)

xses79:
## Devices

| Model | Description | Color |
|-------|-------------|-------|
| 4566702 | Namron Zigbee Edge Termostat | White |
| 4566703 | Namron Zigbee Edge Termostat | Black |
| 4512783 | Namron Zigbee Edge Termostat | White |
| 4512784 | Namron Zigbee Edge Termostat | Black |

4566702/03 and 4512783/84 are identical hardware (different article numbers / colors only).

## Tested firmware

- **HZC Production 1.12** (dated 20241017, OTA version 36)

## Confirmed working

- Temperature control (local temp, setpoint, system mode, running state, calibration)
- Max heat temperature limit
- Thermostat modes: manual, schedule, regulator
- Special modes: eco, frost (toggle), holiday
- Frost mode toggle (direct write to `0x8001`)
- Temperature display unit (celsius/fahrenheit)
- Regulator set point and cycle duration
- Vacation start/end dates (YYMMDD uint32 format)
- Holiday temperature
- Boost time set/remaining
- Window open detection and state
- Child lock (toggle)
- Screen on time and panel brightness
- Auto time sync toggle
- Manual time sync button
- Energy (kWh), current (A), power (W from direct `activePower` report)
- Humidity
- Firmware version and date
- OTA updates

### Time synchronization
Time is written to the device (`0x800B`) at configure, on every private attribute
message (~30 sec intervals while heating), and via `onEvent` at most once per hour
for any incoming Zigbee message. However, if the device is completely idle and sends
no messages for extended periods, the internal clock may drift. The `time_sync_value`
field in the payload reflects the last written value, not a live readback from the
device (the device does not support ZCL readback of `0x800B`).

### Vacation mode auto-activation
Vacation mode (`vacation_start` / `vacation_end`) is written as a YYMMDD date to
the device. The device is expected to activate holiday mode automatically at midnight
on the start date, based on its internal clock. This relies entirely on the time sync
described above being accurate. **Vacation auto-activation has not been fully verified**
— testing with a future start date and waiting for midnight activation was not completed
at time of submission. Manual activation via `thermostat_mode_extra: holiday` works
as expected.

### `activePower` availability
Power is reported via spontaneous ZCL reports from `haElectricalMeasurement`. The
device does not respond to direct reads of `activePower` (always returns 0). Power
is therefore only updated when the device is actively heating and sends a report.
`power` is set to 0 automatically when `running_state` transitions to `idle`.

## Notes
- Private attributes use the standard `hvacThermostat` cluster with no manufacturer code
- All private attribute addresses are in the `0x8000` range
- The device sends a time sync request flag (`0x800A = 1`) periodically; the converter
  responds automatically
- `hzc_*` fields may appear in Z2M state if the built-in Namron definition has previously
  cached state for the device. These disappear after removing and re-adding the device.


src/devices/namron.ts
```diff
@@ -446,8 +446,755 @@ const tzLocal = {
...
+        type: ["attributeReport", "readResponse"] as const,
+        convert: (model, msg): KeyValue => {
+            const result: KeyValue = {};
+            if (msg.data["swBuildId"] !== undefined) result["firmware_version"] = msg.data["swBuildId"];
+            if (msg.data["dateCode"] !== undefined) result["firmware_date"] = msg.data["dateCode"];
+            return result;
+        },
+    } satisfies Fz.Converter<"genBasic", undefined, ["attributeReport", "readResponse"]>,
+
+    thermostat_base: {
```
Koenkk: Could you use the standard `fz.thermostat` for this?

svhelge:
I believe this update could be refactored quite neatly using the ModernExtend framework. If anyone is willing to test it, I’d be happy to give it a try.

xses79:
> I believe this update could be refactored quite neatly using the ModernExtend framework. If anyone is willing to test it, I’d be happy to give it a try.

I can test if you you do, but there was alot of things that i did not get to work. m.humidity (working) and m.electricityMeter, i had some issues, but don't remeber what it was. therefore not usinge more of the modernExtend framework as of now.

src/devices/namron.ts
xses79: Yes, it as been alterd.

Koenkk:
Thanks!

PR #12391, opened 2026-06-07
add: Eglo 901463

panther7:
Any tip, how define, that Eglo 901463 is another model, but is detect as 900566?

- https://github.com/Koenkk/zigbee2mqtt.io/pull/5218

Eglo 901463:

Eglo 900566:


Koenkk:
Thanks!

PR #12392, opened 2026-06-07
fix: IKEA VINDSTYRKA: Allow reading VOC index

andrei-lazarov:
- Reading the attribute works on both 1.0.010 and 1.0.11 firmware -> STATE_GET

- The `voc_index` in `exposes.ts` seems unused


Koenkk:
Thanks!

PR #12400, opened 2026-06-09
Enhance Third Reality device attributes and limits

3reality-support:
Added xAngle, yAngle, and zAngle attributes to the Third Reality device configuration with appropriate types and limits. Updated coolDownTime attribute to allow a maximum of 7200 seconds.

### DIFF
--- a/src/devices/third_reality.ts
+++ b/src/devices/third_reality.ts
@@ -69,6 +69,9 @@ interface ThirdAcceleration {
         xAxis: number;
         yAxis: number;
         zAxis: number;
+        xAngle: number;
+        yAngle: number;
+        zAngle: number;
     };
     commands: never;
     commandResponses: never;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

andrei-lazarov:
This was a breaking change, renaming `vibration` to `alarm_1`
**Users must update their automations to use the new, correct entity.**

You can see in docs:
- [before](https://github.com/Koenkk/zigbee2mqtt.io/blob/418544c6755785747a937feb1d71964b20cc22d2/docs/devices/3RVS01031Z.md#vibration-binary): `vibration`
- [after](https://github.com/Koenkk/zigbee2mqtt.io/blob/d394d999243c8ababa84bab8b31ebc09ce20c7c6/docs/devices/3RVS01031Z.md#alarm-1-binary): `alarm_1`, no more `vibration`

@koenkk please add to changelog

eljojo:
Came here to 👍 the recommendation for changelog warning , I learned about the existence of this project because it renamed sensors in my home.

Maltz42:
I appreciate the added sensors, but what was the point of renaming "vibration" to "alarm_1"?  It seems less clear and is a breaking change.  If this was the only accessible sensor before, the change will break literally ALL existing automations, dashboards, etc in all deployments everywhere.  Ouch.

Koenkk:
I assume this was not intentional, @3reality-support could you submit a fix?

andrei-lazarov:
Maybe we can duplicate the state, in both `alarm_1` and `vibration`?
So we avoid another breaking change for the users that already adapted?

src/devices/third_reality.ts
```diff
@@ -1356,30 +1359,61 @@ export const definitions: DefinitionWithExtend[] = [
...
                 },
                 commands: {},
                 commandsResponse: {},
             }),
+            m.battery(),
+            m.iasZoneAlarm({
+                zoneType: "generic",
+                zoneAttributes: ["alarm_1"],
+                description: "Being in vibration",
+            }),
```
Toxblh, line 1385: btw, will also ask why that is renamed to unclear alarm_1 if that has the same description as the previous vibration?

iamamoose:
Hey @3reality-support please can you either add back the old name or document it as a breaking change, all my sensors broke too!

3reality-support:
Sorry for the inconvenience, our engineers are already repairing it

Maltz42:
Now that I'm looking into this device closer, I have further questions:

- In the post here, the new sensors are described as "xAngle, yAngle, and zAngle", but Zigbee2MQTT, they're listed as X/Y/Z axis acceleration.  And from looking at the reactions of those sensors as I move the device around, I assume the latter is correct - that these sensors measure acceleration, not tilt or angle?

- I notice there is a siren toggled by a DIP switch in the device.  Can this siren be controlled (and/or the DIP switch state read) via Zigbee?  This makes this "alarm_1" nomenclature even more confusing.

PR #12404, opened 2026-06-09
feat: Plugwise 170-01: OpenTherm controls, battery type, product variant

SaiPraveen22:
Extends the existing Plugwise Emma (`170-01`) definition with manufacturer-specific OpenTherm controls and a few related features. No new device — this is an update to an existing one (so no device-picture PR is needed).

## Changes

- **OpenTherm controls** (writable manufacturer-specific `hvacThermostat` attributes): `external_heat_demand`, `external_heat_demand_timeout`, `max_dhw_setpoint`, `max_boiler_setpoint`. Their descriptions note that the device requires *External Control* to be unlocked before these writes are accepted, with a link to the Plugwise External Control page.
- **`battery_type`** — config enum (`alkaline` / `nimh`) backed by a Plugwise `genPowerCfg` manufacturer attribute (`0x007F`).
- **`product_variant`** — read-only diagnostic, read from the standard `genBasic` `productCode` (`0x000A`). The configure read is wrapped so that `UNSUPPORTED_ATTRIBUTE` (current firmware does not implement it yet) does not fail the interview; it also exposes a read button and populates automatically once firmware ships the attribute. Reflects the detected backplate (e.g. `OpenTherm`, `OnOff`, `OpenTherm/OnOff`, `Wireless`).
- Fix a typo in the `keypad_lockout` description (`Keaypad` → `Keypad`).
- Bump the `170-01` definition `version` `0.0.3` → `0.0.4`.

## Test plan

Adds `test/plugwise.test.ts` (vitest) covering:
- scaled writes/reads for each OpenTherm attribute, asserting the value and the Plugwise manufacturer code on the wire;
- `battery_type` set/get;
- `product_variant` parsing (Buffer + string) and the read.

Local run of the full CI gates passes: `pnpm build`, `pnpm check`, `pnpm test` (577 passed), `pnpm bench`.

### DIFF
--- a/src/devices/plugwise.ts
+++ b/src/devices/plugwise.ts
@@ -2,13 +2,16 @@ import {Zcl} from "zigbee-herdsman";
 import * as fz from "../converters/fromZigbee";
 import * as tz from "../converters/toZigbee";
 import * as exposes from "../lib/exposes";
+import {logger} from "../lib/logger";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
-import type {DefinitionWithExtend, Fz, KeyValue, Tz} from "../lib/types";
+import type {Configure, DefinitionWithExtend, Fz, KeyValue, ModernExtend, Tz} from "../lib/types";
 import * as utils from "../lib/utils";
 
 const e = exposes.presets;
 const ea = exposes.access;
+
+const NS = "zhc:plugwise";
 
 const _manufacturerOptions = {manufacturerCode: Zcl.ManufacturerCode.PLUGWISE_B_V};
 
@@ -44,6 +47,14 @@ interface PlugwiseHvacThermostat {
     commands: {
         plugwiseCalibrateValve: Record<string, never>;
     };
+    commandResponses: never;
+}
+
+interface PlugwiseGenPowerCfg {
+    attributes: {
+        plugwiseBatteryType: number;
+    };
+    commands: never;
     commandResponses: never;
 }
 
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

src/devices/plugwise.ts
```diff
@@ -159,6 +197,72 @@ const plugwiseExtend = {
             scale: 100,
             ...args,
         }),
+    externalHeatDemand: (args?: Partial<m.NumericArgs<"hvacThermostat", PlugwiseHvacThermostat>>) =>
+        m.numeric<"hvacThermostat", PlugwiseHvacThermostat>({
+            name: "external_heat_demand",
+            cluster: "hvacThermostat",
+            attribute: "plugwiseExternalHeatDemand",
+            description:
+                "OpenTherm boiler control setpoint override. Set to 0 to disable the override and return control to the thermostat. Requires External Control to be unlocked on the device (see https://www.plugwise.com/emma-external-control).",
```
tinsukE, line 206: Is this URL supposed to be live? It returns 404 to me, and I couldn't find it on archive.org either.

SaiPraveen22: The site is not live yet and will be available soon.

tinsukE:
Do these changes allow one to directly control the boiler via `external_heat_demand` and `external_heat_demand_timeout`?

I have the impression that if I want the room with a radiator to increase temperature by, say, 2C, I could do that by setting `external_heat_demand` to the Emma's current temp + 2C, and keep updating it when the room's or Emma's temp change.

SaiPraveen22:
Not quite the way you describe it — `external_heat_demand` isn't a room-temperature target but the desired boiler control setpoint. It does let you drive the boiler directly, but it won't do the room control for you — you'd run that loop yourself and keep refreshing the value, since `external_heat_demand_timeout` is a watchdog that cancels the override if it isn't updated in time. There must be no active heat demand in the room/zone where Emma is placed, otherwise it keeps control and the override is ignored.

PR #12414, opened 2026-06-11
Update description in legrand.ts: 067755 is not batteryless

alexhk:
There is another Green Power device 067755L without battery, that's probably where the name came from:  https://www.legrand.fr/pro/catalogue/commande-sans-fils-sans-pile-self-e-celiane-with-netatmo-4-scenarios-enjoliveur-blanc

It seems these devices without battery are implemented in z2m as ZLGP:
https://www.zigbee2mqtt.io/devices/ZLGP14_ZLGP15_ZLGP16.html


Koenkk:
thanks!

PR #12415, opened 2026-06-11
fix: AVATTO TS0601_thermostat_14: fix battery low state

### LINKED ISSUE #12407: Bug: AVATTO TRV `battery_low` published as nested object instead of flat boolean
## Description

The AVATTO Thermostatic radiator valve (IEEE `0xa4c138e5569830c6`) publishes `battery_low` as a nested object in the MQTT payload instead of a flat boolean. 
This causes Home Assistant's MQTT binary sensor (created via Z2M auto-discovery) to log continuous errors because the value template `{{ value_json["battery_low"] }}` resolves to a dict rather than a matchable boolean.

## Device info

- **Manufacturer:** AVATTO
- **Model:** TS0601 (_TZE204_vjpaih9f)
- **IEEE address:** `0xa4c138e5569830c6`
- **Zigbee2MQTT version:** 2.12.0
- **zigbee-herdsman version:** 10.4.0
- **zigbee-herdsman-converters version:** 26.63.0

## Observed MQTT payload

```json
{
  "battery": 56,
  "battery_low": {"battery_low": false},
  "running_state": "idle",
  ...
}
```

The `battery_low` field is a nested object `{"battery_low": false}` rather than a flat `false`.

## Expected MQTT payload

```json
{
  "battery": 56,
  "battery_low": false,
  ...
}
```

## Impact
Home Assistant logs the following at INFO level on every TRV message (multiple times per minute):

```
No matching payload found for entity: binary_sensor.trv_rounded_battery_low with state topic: zigbee2mqtt/TRV. template output: '{'battery_low': False}' with value template 'Template<template=({{ value_json["battery_low"] }}) renders=147>'
```

The binary sensor never updates and stays in an unknown state.

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee-herdsman-converters/issues/12407

### DIFF
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -2701,10 +2701,11 @@ export const valueConverter = {
     inverse: {to: (v: boolean) => !v, from: (v: boolean) => !v},
     onOffNotStrict: {from: (v: string) => (v ? "ON" : "OFF"), to: (v: string) => v === "ON"},
     errorOrBatteryLow: {
-        from: (v: number) => {
-            if (v === 0) return {battery_low: false};
-            if (v === 1) return {battery_low: true};
-            return {error: v};
+        from: (v: number, meta: Fz.Meta, options: KeyValue, publish: Publish) => {
+            let batteryLow = false;
+            if (v === 1) batteryLow = true;
+            publish({error: v});
+            return batteryLow;
         },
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

PR #12421, opened 2026-06-12
legrand.ts ZLGP14/ZLGP15/ZLGP16: no OTA (GP device without battery)

alexhk:
There is no OTA capability in these remote switches.


Koenkk:
Thanks!

PR #12431, opened 2026-06-13
fix(tuya): TS0601_dimmer_1_gang_1 suppress stale periodic state report

EdeMeijer:
Follow-up to #12235. That PR bound genOnOff/genLevelCtrl and configured reporting with the library defaults (maximumReportInterval 3600s). On the ION Industries firmware variant (_TZE200_ykgar0ow) the periodic/timer-driven report reads a stale attribute cache: it always sends genOnOff.onOff = 1 and a frozen genLevelCtrl.currentLevel regardless of the device's real state. The result is that the correct state is overwritten with a spurious "on" once per hour.

The event-driven on-change reports are accurate (that is what #12235 unlocked), so only the periodic path is at fault. Set maximumReportInterval to 0 (change-only reporting, no heartbeat) for both attributes. This is lossless for state tracking: the periodic report was never a liveness mechanism (that is handled separately by Z2M's availability feature via active pinging of mains-powered devices), so suppressing it costs nothing.

Verified on a real _TZE200_ykgar0ow dimmer: after reconfigure, no spurious hourly reports over an overnight window with the lamp off, while knob press and rotation still report immediately.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -4666,14 +4666,22 @@ export const definitions: DefinitionWithExtend[] = [
         // physical-knob state changes (see #10272, #25203). For variants whose
         // firmware does not support standard reporting, the bind + configReport
         // requests are harmless no-ops.
-        version: "0.0.1",
+        //
+        // maximumReportInterval is set to 0 (change-only reporting, no periodic
+        // heartbeat). The library default of 3600s causes this firmware to send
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

mejalodewijks-ctrl:
Thanks. I have 7 of those devices and it drove me crazy. Already went back to zha with a few in order to check whether I had the same issue there. But this solves the problem.

PR #12433, opened 2026-06-13
Rollback remove unwanted local_temperature exposed property for Acova Alcantara 3

DeadlockState:
This PR permit to fix :

- Rollback https://github.com/Koenkk/zigbee-herdsman-converters/pull/12331 for Acova Alcantara 3 radiators that have temperature measured on device. Deactivating local_temperature exposed property only applies to Acova Alcantara 2 radiators


Koenkk:
Thanks!

PR #12437, opened 2026-06-14
Add Philips Hue Xamento GU10 (Black) device support

pieterb26:
I bought a Hue 3-pack Xamento spot. I see that it is supported by Zigbee2MQTT but it gives a not supported status.
On the support page I found this: https://www.zigbee2mqtt.io/devices/929003812601.html.

I think my model number is different? Can you add this?


Koenkk:
Thanks!

Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `929003812701.png`

PR #12438, opened 2026-06-14
wyze: fix WLCKG1 lock state reporting via cluster 64512 raw frames

dkneeland:
## Summary

The Wyze Lock v1 (`WLCKG1`, zigbeeModel `Ford`) does not emit standard ZCL lock operation events, so the existing `fz.lock_operation_event` converter never fires and lock state is never reported after pairing.

The lock reports its state via manufacturer-specific cluster 64512 (`0xFC00`) raw frames. Byte 80 of these frames encodes the current lock state in the low two bits: `0b11` (3) = locked, `0b00` (0) = unlocked.

### Changes

- Add `fzLocal.wyzeLockRaw` converter that decodes lock state from byte 80 of cluster 64512 raw frames
- Remove `fz.lock_operation_event` which has no effect on this device
- Keep `fz.lock` for ZCL command responses (lock/unlock via HA)

### Testing

Tested across three Wyze Lock v1 units. The b80 bitmask behaviour is consistent across units and firmware versions. Lock and unlock state are now correctly reported both from remote commands and physical operation of the lock.

### DIFF
--- a/src/devices/wyze.ts
+++ b/src/devices/wyze.ts
@@ -2,9 +2,25 @@ import * as fz from "../converters/fromZigbee";
 import * as tz from "../converters/toZigbee";
 import * as exposes from "../lib/exposes";
 import * as reporting from "../lib/reporting";
-import type {DefinitionWithExtend} from "../lib/types";
+import type {DefinitionWithExtend, Fz} from "../lib/types";
 
 const e = exposes.presets;
+
+// The Wyze Lock v1 reports lock state via manufacturer-specific cluster 64512 (0xFC00)
+// raw frames rather than standard ZCL lock operation events. Byte 80 of these frames
+// encodes the current state: low two bits 0b11 = locked, 0b00 = unlocked.
+const fzLocal = {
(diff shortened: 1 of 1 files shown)


dkneeland:
Updating this PR with a bug fix discovered through physical testing.

The problem with length <= 80: The original guard fires on any frame longer than 80 bytes - this includes both len=85 frames (physical lock/unlock operations) and len=123 frames (heartbeat/rejoin frames the device sends on Z2M restart). The len=123 heartbeat always has byte 80 = 0, which the decoder reads as UNLOCK bits. The result is that every Z2M restart publishes an incorrect UNLOCK state to Home Assistant, overwriting the correct cached state.

The fix: Changing the guard to length !== 85 restricts decoding to only the physical operation frames. Validated across 46+ test frames on three physical WLCKG1 units with 0 mismatches - len=85 frames correctly encode locked/unlocked at byte 80; len=123 frames are now silently skipped.

Koenkk:
Thanks!

PR #12441, opened 2026-06-15
feat: Tuya TS130F cover controllers: expose switch type

andrei-lazarov:
Every module should allow configuring the installed button type (momentary or toggle).. Option was missing..

I tested only on `TS130F _TZ3210_dwytrmda` aka [TS130F_GIRIER](https://www.zigbee2mqtt.io/devices/TS130F_GIRIER.html). It is a custom attribute in `closuresWindowCovering`. 
It seems others have it in `manuSpecificTuya3` (`_TZ3000_yruungrl`).


Koenkk:
Thanks!

PR #12447, opened 2026-06-15
fix(detect): Detect `_TZ3210_wts1g2oh` as Tuya TS0004_switch_module_2

andrei-lazarov:
- https://github.com/romasku/tuya-zigbee-switch/issues/447


Koenkk:
Thanks!

PR #12449, opened 2026-06-16
 fix(detect): Detect _TZ3210_qlmnxmac as Tuya TS0002_power 

phyzical:
Hey i received this device  _TZ3210_qlmnxmac it came in as a 1 outlout power switch but it is actually a 2

image

 https://melery.com/products/melery-zigbee-smart-wall-socket-au-electrical-plug-outlet-power-monitoring-touch-sensor-switch-wireless-remote-alexa-google-home

i think this is the correct spot to add support

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6781,7 +6781,7 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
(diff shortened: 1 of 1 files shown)


phyzical:
although maybe this is not the correct fingerprint as this device also supports countdown. but i cant seem to find a 2 gang + power with countdown?

phyzical:
reopening as i found a better one to throw it into

PR #12464, opened 2026-06-18
Namron 4512791: Write min/max brightness and dimming speed to device …

xses79:
…(off-spec read-before-write)

## Changes

Updates the Namron 4512791 Simplify Zigbee dimmer converter to actually write  min/max brightness and dimming speed to the device instead of handling them  in software only.

## Background

The device (HZC firmware) supports off-spec writing to the following  genLevelCtrl attributes:
- minLevel (0x0002)
- maxLevel (0x0003)
- defaultMoveRate (0x0014)

The device requires a **read-before-write** pattern - sending Write Attributes  without a prior Read Attributes returns NOT_AUTHORIZED, even though the write  technically succeeds with Status: Success. This was confirmed via Wireshark  sniffing of the official Namron Simplify Hub communication.

## What changed

- `min_brightness`: Now writes to device minLevel (0x0002) with read-before-write
- `max_brightness`: Now writes to device maxLevel (0x0003) with read-before-write  
- `dimming_speed`: Now writes to device defaultMoveRate (0x0014)
- Added `convertGet` for all three attributes
- Added `fromZigbee` handler to read values from device
- Updated `configure` to read current values from device on startup

## Tested on

- Firmware: 1.01, 1.09, 1.11 1.13
- Coordinator: ITEAD SONOFF Zigbee 3.0 USB Dongle Plus V2 (EFR32MG21)
- zigbee2mqtt: 2.12.0

### DIFF
--- a/src/devices/namron.ts
+++ b/src/devices/namron.ts
@@ -2847,7 +2847,7 @@ export const definitions: DefinitionWithExtend[] = [
             await endpoint.read("genLevelCtrl", ["minLevel", "maxLevel", "defaultMoveRate", "onLevel"]);
         } catch (e) {
             // Not all firmware versions support reading these
-        }
+        ],
     },
     {
         zigbeeModel: ["4512785"],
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12466, opened 2026-06-18
fix(TO-Q-SYS-JZT): map DP 108 to control_mode

### LINKED ISSUE #32310: Tongou TO-Q-SYS-JZT (_TZE284_6ocnqlhn): leakage_setting (DP 108) reports value 3 → converter throws "Value '3' is not allowed, expected one of 0,1,2"
### What happened?

The built-in definition for the Tongou **TO-Q-SYS-JZT** smart meter
(`_TZE284_6ocnqlhn`, `TS0601`) maps Tuya **DP 108 → `leakage_setting`** with
`tuya.valueConverterBasic.lookup({Ignore: tuya.enum(0), Alarm: tuya.enum(1), Trip: tuya.enum(2)})`
(in `src/devices/tuya.ts`).

My meter periodically reports **`leakage_setting = 3`**, which is not in that
lookup, so the fromZigbee converter throws roughly every 10 seconds:

    Exception while calling fromZigbee converter: Value '3' is not allowed, expected one of 0,1,2

Confirmed via debug log that DP 108 carries the raw value `3`. The other
`*_setting` datapoints (102 over_voltage, 103 under_voltage, 104 over_current,
105 over_power, 107 temperature) all report `0` and convert fine — only DP 108
(leakage) emits `3`.

`leakage_setting` should decode without raising an exception. The meter appears
to support a 4th leakage-protection mode (raw value `3`) that the lookup doesn't
define. Please add the 4th enum value — the device exposes separate
`leakage_trip` and `leakage_alarm` events, so `3` is likely a combined
"Trip + Alarm" mode (needs confirmation). Adding a `fallbackValue` to the lookup
would also harden it against any further out-of-range values.

1. Pair a Tongou TO-Q-SYS-JZT (`_TZE284_6ocnqlhn`).
2. Device reports leakage protection as the 4th mode (raw value 3) on DP 108.
3. Watch the Zigbee2MQTT log — the exception appears ~every 10s when the meter reports.

### Zigbee2MQTT version

2.12.0

### Adapter firmware version

20260310 (ZStack3x0, rev 2.7.1)

### Adapter

zstack — Texas Instruments CC2652P7 (SLZB-06P7, over TCP)

### Setup

Zigbee2MQTT on Debian 13 LXC (Proxmox), running via systemd; adapter connected over TCP

{"id": 8, "type": "Router", "ieeeAddr": "0x142d41fffe93e573", "nwkAddr": 15188, "manufId": 4098, "manufName": "_TZE284_6ocnqlhn", "powerSource": "Mains (single phase)", "modelId": "TS0601", "epList": [1], "endpoints": {"1": {"profId": 260, "epId": 1, "devId": 81, "inClusterList": [0, 4, 5, 61184, 0, 60672], "outClusterList": [25, 10], "clusters": {"genBasic": {"attributes": {"65506": 31, "65508": 0, "65534": 0, "modelId": "TS0601", "manufacturerName": "_TZE284_6ocnqlhn", "powerSource": 1, "zclVersion": 3, "appVersion": 74, "stackVersion": 0, "hwVersion": 1, "dateCode": ""}}}, "binds": [], "configuredReportings": [], "meta": {}}}, "appVersion": 74, "stackVersion": 0, "hwVersion": 1, "zclVersion": 3, "interviewCompleted": true, "interviewState": "SUCCESSFUL", "meta": {"configured": "0.0.0"}}

### Debug log

debug: z2m: Received Zigbee message from 'Energy_1', type 'commandDataReport', cluster 'manuSpecificTuya', data '{"dpValues":[{"data":{"data":[3],"type":"Buffer"},"datatype":4,"dp":108}],"seq":55296}' from endpoint 1 with groupID 0
error: z2m: Exception while calling fromZigbee converter: Value '3' is not allowed, expected one of 0,1,2

### Notes

Local workaround that resolves it: an external converter mirroring the built-in
definition with the DP 108 lookup extended to include `Trip_and_Alarm: tuya.enum(3)`
plus a fallback value. With that, `leakage_setting` decodes (shows as
`Trip_and_Alarm`) and the exception is gone.

Affected definition: `src/devices/tuya.ts`, `_TZE284_6ocnqlhn` / TO-Q-SYS-JZT, DP 108.

Romulas-12:
DP 108 on _TZE284_6ocnqlhn was mislabeled as leakage_setting with a 3-value
lookup (Ignore/Alarm/Trip). The device reports value 3, which threw
"Value '3' is not allowed, expected one of 0,1,2". Per Tuya app inspection
(issue #32310), DP 108 is the device control mode: 0=local_lock, 1=local_mode,
2=remote_mode, 3=full_control. Relabel it and expose control_mode.
Closes Koenkk/zigbee2mqtt#32310

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -18970,6 +18970,7 @@ export const definitions: DefinitionWithExtend[] = [
                     "timing_switch_off",
                 ])
                 .withDescription("Last event of the device"),
+            e.enum("control_mode", ea.STATE, ["local_lock", "local_mode", "remote_mode", "full_control"]).withDescription("Device control mode"),
             e.enum("over_current_setting", ea.STATE_SET, ["Ignore", "Alarm", "Trip"]).withDescription("Over current setting"),
             e
                 .numeric("over_current_threshold", ea.STATE_SET)
@@ -19064,11 +19065,12 @@ export const definitions: DefinitionWithExtend[] = [
                 ],
                 [
                     108,
-                    "leakage_setting",
+                    "control_mode",
                     tuya.valueConverterBasic.lookup({
-                        Ignore: tuya.enum(0),
-                        Alarm: tuya.enum(1),
-                        Trip: tuya.enum(2),
+                        local_lock: tuya.enum(0),
+                        local_mode: tuya.enum(1),
+                        remote_mode: tuya.enum(2),
+                        full_control: tuya.enum(3),
                     }),
                 ],
                 [


Koenkk:
Thanks!

PR #12471, opened 2026-06-19
fix: Tongou TO-Q-SYS-JZT: allow changing control_mode

andrei-lazarov:
@Romulas-12 forgot to mention: I believe you can select the mode.
If you figure out what it does, please update the description

- fixes https://github.com/Koenkk/zigbee-herdsman-converters/pull/12466


Koenkk:
Thanks!

PR #12472, opened 2026-06-19
fix(tuya): correct localTempCalibration1 decode for offsets above 5.5 °C

stackdiver:
fix(tuya): correct localTempCalibration1 decode for offsets above 5.5 °C

### Problem

`valueConverter.localTempCalibration1.from()` used `if (v > 55)`, treating any
positive calibration above 5.5 °C (raw > 55) as a 2's-complement negative and
subtracting `0x100000000`. An offset of `+6 °C` read back as `-429496723.6`,
which is outside the `-9..9` range and breaks the entity. Negative offsets were
unaffected (they are encoded as `>= 0x100000000 - 90`).

Reproduced on a TECH VNTH-T2_v2; affects every device using this converter
(TECH, Essentials, AVATTO, MOES).

### Fix

Use `v > 0x7fffffff`, matching the sibling converters
`localTempCalibration{3,4,5}` which already handle this correctly. The threshold
sits safely between the max positive raw (`9 °C → 90`) and the smallest
encoded-negative raw (`-0.5 °C → 0x100000000 - 5`).

### Test

Added a regression test in `test/tuya.test.ts` covering `from()` for positive
offsets above 5.5 °C, values up to 5.5 °C, negatives, and a full-range
`to()`/`from()` round-trip. It fails on master (`from(60) === -429496723.6`) and
passes with the fix. Full suite: 587 passed.

### DIFF
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -1947,7 +1947,7 @@ export const valueConverter = {
     lockUnlock: valueConverterBasic.lookup({LOCK: true, UNLOCK: false}),
     localTempCalibration1: {
         from: (v: number) => {
-            if (v > 55) v -= 0x100000000;
+            if (v > 0x7fffffff) v -= 0x100000000;
             return v / 10;
         },
         to: (v: number) => {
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks

PR #12476, opened 2026-06-19
fix(detect): Detect `_TZE28C1000000_jtbgusdc` as AVATTO ZDMS16-2

andrei-lazarov:
picked up from Discord

### DIFF
--- a/src/devices/avatto.ts
+++ b/src/devices/avatto.ts
@@ -407,7 +407,14 @@ export const definitions: DefinitionWithExtend[] = [
         },
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Off-topic: interesting to see that Tuya changes it's manuf names, e.g. `_TZE28C1000000`

andrei-lazarov:
Yes, I haven't found out what the prefix means yet (if anything)..
I think we should automatically crop that out, and match the converter based on the suffix `_jtbgusdc`.
It would prevent some of this work. What do you think @Koenkk?

Koenkk:
That sounds like an idea, maybe we can automatically populate those in index.ts? I guess same holds for the TZE204 and TZE284

andrei-lazarov:
Should work, but I don't know how to do that efficiently

PR #12482, opened 2026-06-20
fix: unref timers

Nerivec:
Avoid holding the event loop with timers.
All of these are background ops, with low impact, so, shouldn't be a problem.
Might end up solving a few "weird" shutdown-related issues.

Also switched some "sleep" usage to the dedicated util.

Note: a further refactor could be improving the store and specifically properly cleaning up timers (currently buried in random objects) with a ZHC shutdown sequence.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://biomejs.dev/schemas/2.4.11/schema.json",
+    "$schema": "https://biomejs.dev/schemas/2.5.0/schema.json",
     "vcs": {
         "enabled": true,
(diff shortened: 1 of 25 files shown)


Koenkk:
For convenience, maybe we can introduce a setTimeoutAndUnref?

Nerivec:
Actually, we can chain it (returns self), cleaner than adding a wrapper.

Koenkk:
> Actually, we can chain it (returns self), cleaner than adding a wrapper.

Even cleaner indeed, thank!

PR #12485, opened 2026-06-20
feat: add Frient WISZB-121 device

GreatWizard:
Add the config for the Frient WISZB-121 device device.
It's similar to the 131 without temp sensor.


GreatWizard:
For some reason my Frient device is not recognized but it seems it's already integrated under the Develco vendor... I'm investigating...

PR #12486, opened 2026-06-20
Fix Sylvania 74388 battery percentage

### LINKED ISSUE #12487: Sylvania 74388 reports 27% battery with new CR2 cells
The Sylvania 74388 / Zigbee model `Contact Sensor-A` currently reports low battery percentage with fresh CR2 batteries.

I have several of these Zigbee sensors, and with brand new CR2s (volt meter shows >3v), they only report `voltage: 2800`. Due to the current `voltageToPercentage: "3V_2100"`, Home Assistant shows them all at 27% with new batteries.

The device appears to cap or underreport its Zigbee `batteryVoltage` at 28 / 2800 mV, so the current 3 V curve is not appropriate for this model.

Proposed fix: #12486

klinquist:
## Summary
- adjust Sylvania 74388 battery voltage-to-percentage mapping to treat 2800 mV as full

## Root cause
The 74388 currently uses the shared `3V_2100` conversion curve, which maps a reported 2800 mV to 27%. These sensors appear to cap their Zigbee-reported `batteryVoltage` at 28 / 2800 mV even when brand new CR2 batteries measure above 3 V with a voltmeter.

Fixes #12487

### DIFF
--- a/src/devices/sylvania.ts
+++ b/src/devices/sylvania.ts
@@ -14,7 +14,7 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Smart+ contact and temperature sensor",
         fromZigbee: [fz.ias_contact_alarm_1, fz.temperature, fz.battery],
         toZigbee: [],
-        meta: {battery: {voltageToPercentage: "3V_2100"}},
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12491, opened 2026-06-21
use reported multiplier/divisor for lupus 12050 power reporting

derabbink:
This fixes a bug with Lupus 12050 smart plugs: The converter previously hardcoded the divisor and multiplier, even though the device actually reports which values should be used.  
On my own devices of this type, this meant that the reported power was off by a factor of 10.

I first tested this with a custom converter, and it works reliably & correctly.  
Hopefully this change is useful to other lupus 12050 users.


Koenkk:
Thanks!

PR #12492, opened 2026-06-21
Add Philips Hue Essential lightstrip (5m) 929004294901

Gi007i:
Adds the 5 m variant of the Hue Essential lightstrip. It's the same product as the already-supported `929004294903` (16ft) with an identical extend, so I placed `929004294901` right next to it.

Tested on the device: on/off, brightness, colour temperature (153–447 mired), full colour via xy/hs, and the Hue light effects all work. OTA comes for free through `philips.m.light`.

Fixes #32353


Koenkk:
Thanks!

Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `929004294901.png`

PR #12493, opened 2026-06-21
Improve Develco ZHEMI101 converter

derabbink:
This improves the converter for the Develco ZHEMI101 Utility Meter devices.

A brief explanation about the device's function: This device is attached to a utility meter (electricitly, gas, water), and counts pulses that the meter emits, in order to follow along with what the meter reports. Typically this device is sold with an optical probe that registers optical pulses from an electricity meter. Devices sold like that are configured to report electricity usage (kWh), and the device expects 1000 pulses per kWh.  
It is trivial to replace the optical probe with e.g. a reed contact that is closed by the magnet inside a mechanical water or gas meter. This way, the device can follow along the increasing meter position of a water or gas meter.  
When you install a ZHEMI101, the user will want to specify the following:
- The interface mode (what it is measuring): electricity, gas, water
    - the device will change the unit of measure of what it tracks; kWh/L/m³/...
    - the device _may_ change the `multiplier`/`divisor`
- The current value of their utility meter
    - when setting this, the user will just want to enter what they are reading on their metering device, without needing to do a conversion using the `multiplier`/`divisor` parameters that the ZHEMI101 picked.

This PR makes all of this more convenient.
- the user can simply pick the "Interface Mode" from a dropdown
    - this will update the units in which measurements are exported
- the user can simply enter the "Unit Summation", as they are reading it from their metering device
    - the converter takes care to figure out what scaled value must be sent to the ZHEMI101
    - there is runtime validation which checks if after scaling this number exceeds what this zigbee attribute accepts

I tested this using a custom converter first.

PS: I found a technical spec of the device's zigbee device implementation [here](https://www.manualslib.com/manual/2343264/Develco-Zhemi-101.html)

### DIFF
--- a/src/devices/develco.ts
+++ b/src/devices/develco.ts
@@ -30,6 +30,276 @@ const develcoLedControlMap = {
     255: "both",
 };
 
+const zhemi101Uint48Max = 0xffffffffffff;
+const zhemi101UnitOfMeasure = {
+    kWh: 0x00,
+    cubicMeters: 0x01,
+    cubicFeet: 0x02,
+    ccf: 0x03,
+    usGallons: 0x04,
+    imperialGallons: 0x05,
+    btu: 0x06,
+    liters: 0x07,
+} as const;
+const zhemi101MeteringDeviceType = {
+    electricity: 0x00,
+    gas: 0x01,
+    water: 0x02,
+} as const;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12496, opened 2026-06-21
Fix GIEX irrigation time exposes

Domi920:
The GIEX valve reports irrigation_start_time, irrigation_end_time and last_irrigation_duration as strings.

Examples:
- 21:52:37
- 21:40:49
- 00:03:40

These values are currently exposed as numeric, which causes them to be truncated.

Changing the exposes from numeric to text fixes the issue.

Tested on QT06_2 (_TZE200_a7sghmms).


Koenkk:
Thanks!

PR #12497, opened 2026-06-21
Make Shelly 2PM cover tilt configurable

MaxRink:
## Summary
- Make Shelly 2PM Gen4 cover tilt/slat exposure dynamic instead of always exposing tilt.
- Read `Cover.GetConfig` over the Shelly RPC cluster during configure and persist `slat.enable` into device metadata.
- Keep a `cover_tilt_enabled` device option with `auto`, `true`, and `false` for manual override when RPC config cannot be read.
- Handle `Cover.GetConfig` responses wrapped in `result`, wrapped in `params`, or returned directly.

## Scope
This PR is intentionally cover-only. The related behavior from the original combined branch was split into separate PRs:
- Wi-Fi readback/fallback fix: #12524
- Switch input endpoint cleanup: #12525

## Review feedback addressed
- `auto` is no longer equivalent to `false`: after configure it follows the device-reported `slat.enable` value, while `true` and `false` remain explicit overrides.
- Overwrites stale persisted tilt metadata when the device reports `slat.enable: false`.
- Keeps tilt controls hidden for Shelly 2PM cover variants without slat control, while still allowing a manual override when RPC config cannot be read.

## Validation
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx pnpm@10.18.3 exec vitest run test/shelly.test.ts --config ./test/vitest.config.mts` (`8` tests)
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx pnpm@10.18.3 exec biome check --error-on-warnings src/devices/shelly.ts test/shelly.test.ts`
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx pnpm@10.18.3 run build`
- `git diff --check`

## Live validation
- Verified on my Home Assistant instance that the Shelly 2PM Gen4 cover devices expose the `cover_tilt_enabled` option.
- Verified on my Home Assistant instance that cover tilt/slat controls remain hidden when Shelly slat control is disabled.
- Verified on my Home Assistant instance with a combined local override before this split; this PR now contains only the cover part of that validated behavior.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -19,6 +19,14 @@ const NS = "zhc:shelly";
 const HA_ELECTRICAL_MEASUREMENT_CLUSTER_ID = 0x0b04;
 const HA_ELECTRICAL_MEASUREMENT_POWER_FACTOR_ATTR_ID = 0x0510;
 
+const checkOption = (device: Zh.Device | DummyDevice, options: KeyValue, key: string, defaultValue = false): boolean => {
+    if (options?.[key] === "true") return true;
+    if (options?.[key] === "false") return false;
+    if (!utils.isDummyDevice(device) && device.meta[key] !== undefined) return !!device.meta[key];
+
+    return defaultValue;
+};
+
 const shellyPresenceEndpointNames = (device: Zh.Device | DummyDevice): string[] => {
     const count =
         !utils.isDummyDevice(device) && typeof device.meta.presence_zone_count === "number"
@@ -381,6 +389,25 @@ const shellyModernExtend = {
             }),
         ];
     },
+    shellyWindowCovering(): ModernExtend {
+        const result = m.windowCovering({controls: ["lift", "tilt"]});
+        const tiltOption = e
+            .enum("cover_tilt_enabled", ea.SET, ["auto", "true", "false"])
+            .withDescription("Expose tilt/slat controls for covers with Shelly slat control enabled");
+        const exposesFn: DefinitionExposesFunction = (device, options) => {
+            const cover = e.cover().withPosition();
+            if (checkOption(device, options, "cover_tilt_enabled")) {
+                cover.withTilt();
+            }
+
+            return [cover];
+        };
+
+        result.exposes = [exposesFn];
+        result.options = [...(result.options ?? []), tiltOption];
+
+        return result;
+    },
     shellyPresenceOccupancy(): ModernExtend {
         const exposesFn: DefinitionExposesFunction = (device) => {
             return shellyPresenceEndpointNames(device).map((endpointName) => e.occupancy().withAccess(ea.STATE_GET).withEndpoint(endpointName));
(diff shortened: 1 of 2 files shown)


andrei-lazarov:
This is cool. But I don't think it should be specific to Shelly, and disabled by default.
Also why include `auto` if it's the same as `false`?

MaxRink:
im currentlky still trying to find a way to read the state that is actually set via the WbUI/App on the shelly to make auto useful
image
But the general idea for this flag is to inherit whatever got confed during shelly setup

Edit: im currently looking at https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/RPCCluster and i think it should be possible

Koenkk:
> Edit: im currently looking at https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/RPCCluster and i think it should be possible

Could we go for this solution directly?

MaxRink:
Ive already implemented it, just havent updated the MR description yet

MaxRink:
Updated the implementation and PR body around this.

`auto` now reads the Shelly-side setting directly via the RPC cluster (`Cover.GetConfig`) during configure and persists `slat.enable` into device metadata. So `auto` means “follow the device config from the Shelly Web UI/app”, while `true`/`false` remain explicit Zigbee2MQTT overrides for failed RPC reads or manual testing.

I kept the option Shelly-specific because the only device-side source of truth here is Shelly RPC; the generic cover path is unchanged.

Koenkk:
thanks!

SebiGo:
Thanks a lot for adding this, @MaxRink — being able to hide tilt controls for covers without slats is a nice improvement!

Just want to flag one thing for other users upgrading: on my existing Shelly 2PM cover with real venetian-blind hardware, tilt was always exposed before, and after this update it disappeared from Home Assistant. The auto detection via Cover.GetConfig apparently didn't pick up slat.enable correctly in my case.

For anyone else running into this: you can fix it via the Z2M UI under the device's Settings (specific) tab → Cover tilt enabled, and set it to true:

Screenshot from 2026-07-04 10-02-19

To be clear, this is a one-time migration issue for existing setups, not an ongoing problem — going forward, new pairings should detect this correctly from the start.

ebeng:
> Thanks a lot for adding this, @MaxRink — being able to hide tilt controls for covers without slats is a nice improvement!
> 
> Just want to flag one thing for other users upgrading: on my existing Shelly 2PM cover with real venetian-blind hardware, tilt was always exposed before, and after this update it disappeared from Home Assistant. The auto detection via Cover.GetConfig apparently didn't pick up slat.enable correctly in my case.
> 
> For anyone else running into this: you can fix it via the Z2M UI under the device's Settings (specific) tab → Cover tilt enabled, and set it to true:
> 
> Screenshot from 2026-07-04 10-02-19
> To be clear, this is a one-time migration issue for existing setups, not an ongoing problem — going forward, new pairings should detect this correctly from the start.

Indeed! 
All of my automations for the blinds/shutters/covers stopped working. I had to one by one enable them back again :) 
Thank you! I wouldn't have found it otherwise...

PR #11451, opened 2026-02-02
Tidying up messy PRs of Excellux devices

Bjk8kds:
Tidying up messy PRs.
I only tidied up the appearance/cosmetics related stuff, like naming, etc, without changing the converter logic inside.

Original commit: ------------------------------------------- This commit:

Device name | Fingerprint | PNG name | → | Selected name |
|:--------------|:------------|:------------|:---|:-----------------|
NTCHT-01     | NTCHT01  | NTCHT01    |     | ZG-105NTH   |
                     |                  | ZG-105NTH |     |                       |
PIRIV-01       | PIRIV01     | PIRIV01        |    | ZG-104PLV     |
                     |                  | ZG-104PLV   |     |                       |
Contact-01   | CAT0001   | ZG-102MV   |     | ZG-102MV     |
VABRATE-01 | VABRATE   | ZG-103V      |    | ZG-103V         |
Scene-Switc | DSS0010    | DSS0010      |    | ZG-101K         |
                    |                    | ZG-101K      |    |                        |
FEDHT-01    | DHT0001    | DHT0001     |    |  DHT0001       |
                    | DHTA001    | DHTA001     |    | DHTA001 (whitelabel) |

Other:
- Fixed unit `S` → `s` (seconds)
- Fixed typo `illumiance` → `illuminance`
- Fixed name `presence_state` → `presence`
- Capitalized descriptions consistently
- Replaced manual battery definition with `e.battery()`
- Removed `.withUnit("times")` from `vibration_sensitivity`

Related:
- ZHC
https://github.com/Koenkk/zigbee-herdsman-converters/pull/11297
https://github.com/Koenkk/zigbee-herdsman-converters/pull/11290

- z2m.io
https://github.com/Koenkk/zigbee2mqtt.io/pull/4723
https://github.com/Koenkk/zigbee2mqtt.io/pull/4726
https://github.com/Koenkk/zigbee2mqtt.io/pull/4771

<br>

Link to picture pull request:
**NB:** For PNG images, it will follow because it has been committed by the original contributor,  just need to clean up unused images after the next update.

PR #11290, opened 2026-01-14
Add Excellux sensor

iotRobot:
Link to picture pull request: TODO

PR #12505, opened 2026-06-22
fix: Excellux ZG-104PLV: expose presence as binary occupancy

nielsbauman:
## What

Changes the `ZG-104PLV` (Excellux PIR / vibration / light sensor) `presence` from a string `enum` (`"true"`/`"false"`) to a binary `occupancy` expose.

## Why

The motion signal is currently exposed as an enum, so consumers that expect a boolean presence/motion signal don't see it as one:

- In Home Assistant it discovers as a `sensor` holding the string `"true"`/`"false"`, **not** a `binary_sensor`. That breaks anything that filters on binary sensors — e.g. the Magic Areas integration ignores it entirely for occupancy tracking, and it can't carry a `occupancy`/`motion` device class.
- It's inconsistent with the rest of the file. This is the only `enum("presence")` in `tuya.ts`; the other ~22 Tuya PIR/presence sensors using the same DP1 enum pattern all use `e.occupancy()`.
- The underlying datapoint (DP1, enum `0`/`1`) maps cleanly to a boolean, so there's no device-side reason for the enum — it's a historical artifact of the original add (#11290), and the later cosmetic cleanup (#11451) explicitly left converter logic untouched.

## Breaking change

Yes. The property/entity renames `presence` → `occupancy` (and the value type changes from the string `"true"`/`"false"` to a boolean / HA `on`/`off`). Users with automations referencing `presence` will need to update them.

Note there is precedent: #11451 already shipped a breaking rename on this same device (`presence_state` → `presence`).

## Tested

Validated on real hardware (1 unit) via an external converter before submitting: the device now discovers as `binary_sensor` `occupancy` (device_class occupancy), toggles `ON` on motion / `OFF` when clear, and all other exposes (vibration, illuminance, battery, sampling_interval, etc.) are unaffected.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -26142,7 +26142,7 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [tuya.modernExtend.tuyaBase({dp: true})],
         description: "PIR motion sensor, vibration sensor, and light sensor",
         exposes: [
-            e.enum("presence", ea.STATE, ["true", "false"]).withDescription("Presence state, true: motion detected, false: no motion"),
+            e.occupancy().withDescription("Presence state, true: motion detected, false: no motion"),
             e.binary("vibration", ea.STATE, true, false).withDescription("Vibration state, true: vibration detected, false: no vibration"),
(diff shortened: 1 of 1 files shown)


Koenkk:
Technically a breaking change, but obviously how it was implemented was not correct, so marking this as a fix, thanks!

PR #12506, opened 2026-06-22
Discover Shelly Presence zones from config

MaxRink:
## Summary
- expose Shelly Presence Gen4 occupancy endpoints dynamically from the known PresenceZone count
- read `Shelly.GetConfig` over the Shelly RPC cluster during configure and persist `presence_zone_count`
- keep the safe fallback of all 10 possible PresenceZone endpoints until device config has been read

## Notes
Shelly Presence Gen4 supports PresenceZone components. This avoids exposing unused duplicate occupancy sensors once the device config is available while still keeping first-pairing behavior safe.

## Validation
- `pnpm exec vitest run test/shelly.test.ts --config ./test/vitest.config.mts`
- `pnpm run check`
- `pnpm run build`
- pre-commit hook: `pnpm run build`, `pnpm run check`, `pnpm run test`

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -12,11 +12,21 @@ const ea = exposes.access;
 
 const SHELLY_ENDPOINT_ID = 239;
 const SHELLY_OPTIONS = {profileId: ZSpec.CUSTOM_SHELLY_PROFILE_ID};
+const SHELLY_PRESENCE_MAX_ZONES = 10;
 
 const NS = "zhc:shelly";
 
 const HA_ELECTRICAL_MEASUREMENT_CLUSTER_ID = 0x0b04;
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

teodor-hristov:
Is that feature all about "showing" only the "used" occupancy endpoints?

MaxRink:
Yes, before that all 10 zones would always be exposed, even when they are
not configured and thus won't ever have valid data

teodor-hristov:
> Yes, before that all 10 zones would always be exposed, even when they are not configured and thus won't ever have valid data
> […](#)

ah great! we didn't rly found solution to this using only the zigbee specific stuff, so this will help showing only the needed info :)

PR #12509, opened 2026-06-22
Handle named custom raw attribute responses

MaxRink:
## Summary
- allow modern extends that use raw `{ID, type}` attributes to consume both numeric and named read/attribute response payload keys
- apply the fallback to enum, numeric, binary, and text modern extends
- widen `convertGet` typing so RPC-style getters can return immediate state when they receive it synchronously

## Why
Some custom clusters are read by numeric attribute ID, but zigbee-herdsman decodes the response using the custom cluster attribute name. This left values at `unknown`/`null` even though the device replied successfully.

Observed with IKEA INSPELNING smart plug controls:
- read request: `manuSpecificIkeaSmartPlug.read([0])`
- response: `{childLock: 0}`
- read request: `manuSpecificIkeaSmartPlug.read([1])`
- response: `{ledEnable: 1}`

Before this change, the generic binary modern extend only looked for keys `0` and `1`, so `child_lock` and `led_enable` stayed unset. With this change, it also resolves the custom cluster attribute name from the registered device custom clusters.

## Review feedback addressed
- The custom cluster name lookup is now explicitly limited to raw numeric attribute IDs; string attributes stay on the direct payload-key path.

## Validation
- `./node_modules/.bin/biome check --error-on-warnings src/lib/modernExtend.ts src/lib/types.ts test/modernExtend.test.ts`
- `./node_modules/.bin/tsc`
- `node dist/indexer.js`
- `./node_modules/.bin/vitest run --config ./test/vitest.config.mts test/modernExtend.test.ts`
- post-rebase: `pnpm exec biome check --error-on-warnings src/lib/modernExtend.ts test/modernExtend.test.ts`
- post-rebase: `pnpm exec vitest run --config ./test/vitest.config.mts test/modernExtend.test.ts`

Live validated on my Home Assistant instance with an IKEA INSPELNING smart plug:
- manual get published `child_lock: "UNLOCK"` and `led_enable: "TRUE"`
- single `device/configure` read both attributes during setup and finished with `status: ok`

### DIFF
--- a/src/lib/modernExtend.ts
+++ b/src/lib/modernExtend.ts
@@ -148,6 +148,39 @@ function convertReportingConfigTime(time: ReportingConfigTime): number {
     return time;
 }
 
+function getAttributeValue<Cl extends string | number, Custom extends TCustomCluster | undefined = undefined>(
+    msg: Fz.Message<Cl, Custom, ["attributeReport", "readResponse"]>,
+    cluster: Cl,
+    attribute: ClusterWithAttribute<Cl, Custom>["attribute"],
+    manufacturerCode: number | undefined,
+    device: Zh.Device | undefined,
+): unknown {
+    const attributeKey = isString(attribute) ? attribute : attribute.ID;
+    if (attributeKey in msg.data) {
+        return msg.data[attributeKey as keyof typeof msg.data];
+    }
+
+    if (isString(attribute)) {
+        return undefined;
+    }
+
+    const attributeID = attribute.ID;
+    if (typeof attributeID !== "number") {
+        return undefined;
+    }
+
+    try {
+        const clusterDefinition = Zcl.Utils.getCluster(cluster, manufacturerCode, device?.customClusters);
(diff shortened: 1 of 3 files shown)


src/lib/modernExtend.ts
```diff
@@ -148,6 +148,44 @@ function convertReportingConfigTime(time: ReportingConfigTime): number {
...
+}
+
+function getAttributeValue<Cl extends string | number, Custom extends TCustomCluster | undefined = undefined>(
+    msg: Fz.Message<Cl, Custom, ["attributeReport", "readResponse"]>,
+    cluster: Cl,
+    attribute: ClusterWithAttribute<Cl, Custom>["attribute"],
+    manufacturerCode: number | undefined,
+    device: Zh.Device | undefined,
+): unknown {
+    for (const attributeKey of getAttributeKeys(cluster, attribute, manufacturerCode, device)) {
```
Koenkk: I think all of this looping only has to be executed when the attribute is a number?

Koenkk:
Thanks!

PR #12512, opened 2026-06-23
Fix Third Reality soil moisture fallback

MaxRink:
## Summary

- make `3RSM0147Z` soil moisture handling depend on the interviewed endpoint clusters
- keep native `msSoilMoisture` reporting/readback for units that expose it
- fall back to `msRelativeHumidity` as `soil_moisture` for older units that report the probe value there and do not expose the native soil moisture input cluster
- keep exposing `soil_moisture` instead of a normal humidity sensor for this model

## Why

My local Home Assistant/Zigbee2MQTT install has two `3RSM0147Z` variants:

- `Pflanzensensor 1`: hardware v1/application v51, endpoint exposes `msSoilMoisture`
- `Pflanzensensor 2`: hardware v0/application v38, endpoint does not expose `msSoilMoisture`, but reports the probe value through `msRelativeHumidity`

The old definition always configured `msSoilMoisture`, so the older unit repeatedly failed configure with:

```text
Device 0xffffb40e06020107 has no input cluster msSoilMoisture
```

Related public reports also describe early `3RSM0147Z` units exposing/reporting the probe moisture as humidity rather than native soil moisture:

- https://github.com/Koenkk/zigbee2mqtt/issues/26917
- https://github.com/Koenkk/zigbee2mqtt/issues/27950
- https://github.com/Koenkk/zigbee2mqtt/discussions/27484

## Validation

Local:

- `./node_modules/.bin/biome check --write src/devices/third_reality.ts test/third_reality.test.ts`
- `./node_modules/.bin/vitest run test/third_reality.test.ts --config ./test/vitest.config.mts`
- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/tsc && node dist/indexer.js`
- `git diff --check`

Live validation on my Home Assistant instance:

- deployed only `dist/devices/third_reality.js` into the Zigbee2MQTT add-on
- reconfigured `Pflanzensensor 2` through `zigbee2mqtt/bridge/request/device/configure`
- confirmed response: `{"data":{"id":"Pflanzensensor 2"},"status":"ok"}`
- confirmed `/get {"soil_moisture":""}` reads `msRelativeHumidity` and publishes `soil_moisture: 66.15`
- confirmed current HA discovery for this device includes `soil_moisture` and no humidity entity

### DIFF
--- a/src/devices/third_reality.ts
+++ b/src/devices/third_reality.ts
@@ -21,6 +21,7 @@ import type {
 import * as utils from "../lib/utils";
 
 const e = exposes.presets;
+const ea = exposes.access;
 
 function conditionalPressure(): ModernExtend {
     const base = m.pressure();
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #12513, opened 2026-06-23
fix(bosch): make RM230Z relay support optional

MaxRink:
## Summary

Some Bosch Room Thermostat II 230V (`BTH-RM230Z`) installations do not expose the `genOnOff` input cluster. The converter currently adds `relayState()` unconditionally, so configure tries to bind/configure `genOnOff` and fails on those devices.

This makes the relay support conditional on the interviewed endpoint data:

- keep the relay switch expose and reporting setup when an endpoint supports `genOnOff`
- hide the relay switch expose when no endpoint supports `genOnOff`
- skip relay reporting setup in that case instead of failing device configure

## Validation

- `node --run check`
- `node --run build`
- `./node_modules/.bin/vitest run test/bosch.test.ts --config ./test/vitest.config.mts`
- `node --run test`

## Live validation

Validated with a local Zigbee2MQTT override on my local install with five `BTH-RM230Z` devices:

- `Arbeitszimmer Thermostat`, `Wohnzimmer Thermostat Flur`, and `Wohnzimmer Thermostat Küche` do not expose `genOnOff`; manual configure now logs `Skipping Bosch thermostat relay reporting ... no genOnOff input cluster` and completes successfully.
- `Schlafzimmer Thermostat` and `Bad Thermostat` still configure successfully without taking the skip path, preserving relay-capable behavior.

Before this change, the no-relay devices failed manual configure with `Device ... has no input cluster genOnOff`.

### DIFF
--- a/src/lib/bosch.ts
+++ b/src/lib/bosch.ts
@@ -3533,7 +3533,27 @@ export const boschThermostatExtend = {
             commands: {},
             commandsResponse: {},
         }),
-    relayState: () => m.onOff({description: "The state of the relay controlling the connected heating/cooling device", powerOnBehavior: false}),
+    relayState: (): ModernExtend => {
+        const description = "The state of the relay controlling the connected heating/cooling device";
+        const relay = m.onOff({description, powerOnBehavior: false, configureReporting: false});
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks, could you fix the merge conflicts?

MaxRink:
Done

Koenkk:
thanks!

PR #12514, opened 2026-06-23
Mark battery state exposes as diagnostic

MaxRink:
## Summary

- Mark `battery_state` exposes as diagnostic metadata where they are emitted directly by device definitions.
- Mark the shared Tuya `batteryState()` helper as diagnostic so Tuya datapoint devices inherit the same Home Assistant category.
- Mark the PushOk `POK012` `m.enumLookup` battery-state expose as diagnostic via the existing `entityCategory` option.
- Add regression coverage for the shared Tuya helper, a Tuya soil sensor definition, and the modern `enumLookup` path.

## Why

Home Assistant already treats `battery`, `battery_low`, and `voltage` as diagnostic-style device health entities. `battery_state` has the same operational meaning, but some devices still exposed it as a primary sensor.

## Validation

- `node --run check`
- `node --run build`
- `node --run test` (`19` files, `615` tests)
- `git diff --check`

## Live validation

Tested on a Home Assistant / Zigbee2MQTT override with two Tuya soil sensors:

- `sensor.pflanzensensor_3_battery_state`
- `sensor.pflanzensensor_4_battery_state`

After deploying the converter override, retained MQTT discovery for both sensors published `entity_category: diagnostic`. After removing the two stale Home Assistant MQTT entity-registry rows and restarting Zigbee2MQTT, Home Assistant recreated both entities with `entity_category: diagnostic`.

### DIFF
--- a/src/devices/alecto.ts
+++ b/src/devices/alecto.ts
@@ -45,7 +45,7 @@ export const definitions: DefinitionWithExtend[] = [
         meta: {},
         exposes: [
             e.enum("smoke_state", ea.STATE, ["alarm", "normal"]),
-            e.enum("battery_state", ea.STATE, ["low", "middle", "high"]),
+            e.enum("battery_state", ea.STATE, ["low", "middle", "high"]).withCategory("diagnostic"),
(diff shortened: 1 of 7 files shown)


Koenkk:
Thanks!

PR #12523, opened 2026-06-23
feat(NodOn): modernize SDO-4-1-00 and add configureReporting for ssIasZone/zoneStatus

nthiers:
### Problem

The current `SDO-4-1-00` definition has two issues:

**1. Incomplete modernExtend migration**

The entry mixes legacy converters with `modernExtend`:

```typescript
fromZigbee: [fz.battery, fz.ias_contact_alarm_1],  // legacy
toZigbee: [],
exposes: [e.contact()],                              // legacy
extend: [m.battery({voltageReporting: true})],       // modernExtend
```

`fz.battery` is already covered by `m.battery`. `fz.ias_contact_alarm_1` and `e.contact()` are covered by `m.iasZoneAlarm`.

**2. No `configureReporting` for `ssIasZone/zoneStatus`**

Without it, the periodic heartbeat interval is never configured. The attribute does not appear in Z2M's Reporting tab and stays at whatever the device factory default is.

---

### Fix

```diff
  {
      zigbeeModel: ['SDO-4-1-00'],
      model: 'SDO-4-1-00',
      vendor: 'NodOn',
      description: 'Door & window opening sensor',
-     fromZigbee: [fz.battery, fz.ias_contact_alarm_1],
-     toZigbee: [],
-     exposes: [e.contact()],
      extend: [
          m.battery({voltageReporting: true}),
+         m.iasZoneAlarm({zoneType: 'contact', zoneAttributes: ['alarm_1']}),
      ],
+     configure: async (device, coordinatorEndpoint) => {
+         const endpoint = device.getEndpoint(1);
+         await endpoint.configureReporting('ssIasZone', [{
+             attribute: 'zoneStatus',
+             minimumReportInterval: 0,
+             maximumReportInterval: constants.repInterval.HOUR,
+             reportableChange: 1,
+         }]);
+     },
      ota: true,
  },
```

---

### Tested

- Device: **SDO-4-1-00** firmware V4.3.0
- Coordinator: SONOFF Zigbee 3.0 USB Dongle Plus-E (EFR32MG21, firmware 7.4.4 GA)
- `configureReporting` confirmed applied: `cluster ssIasZone (0x0500), attr zoneStatus (0x0002), min: 0s, max: 3600s, change: 1`
- Hourly heartbeats received consistently over 24h of monitoring
- Immediate open/close notifications confirmed on each door event

### DIFF
--- a/src/devices/nodon.ts
+++ b/src/devices/nodon.ts
@@ -291,10 +291,20 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SDO-4-1-00",
         vendor: "NodOn",
         description: "Door & window opening sensor",
-        fromZigbee: [fz.battery, fz.ias_contact_alarm_1],
-        toZigbee: [],
-        exposes: [e.contact()],
-        extend: [m.battery({voltageReporting: true})],
+        extend: [
+            m.battery({voltageReporting: true}),
+            m.iasZoneAlarm({zoneType: "contact", zoneAttributes: ["alarm_1"]}),
+        ],
+        configure: async (device, coordinatorEndpoint) => {
+            const endpoint = device.getEndpoint(1);
+            // No bind needed: coordinator is already enrolled as IAS CIE during pairing.
+            await endpoint.configureReporting("ssIasZone", [{
+                attribute: "zoneStatus",
+                minimumReportInterval: 0,
+                maximumReportInterval: constants.repInterval.HOUR,
+                reportableChange: 1,
+            }]);
+        },
         ota: true,
     },
     {
--- a/src/devices/nodon.ts
+++ b/src/devices/nodon.ts
@@ -291,19 +291,10 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SDO-4-1-00",
         vendor: "NodOn",
         description: "Door & window opening sensor",
-        extend: [m.battery({voltageReporting: true}), m.iasZoneAlarm({zoneType: "contact", zoneAttributes: ["alarm_1"]})],
-        configure: async (device, coordinatorEndpoint) => {
-            const endpoint = device.getEndpoint(1);
-            // No bind needed: coordinator is already enrolled as IAS CIE during pairing.
-            await endpoint.configureReporting("ssIasZone", [
-                {
-                    attribute: "zoneStatus",
-                    minimumReportInterval: 0,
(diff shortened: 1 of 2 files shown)


nthiers:
Hello @Koenkk , this converter upgrade is OK for you? Thanks.

src/devices/nodon.ts
```diff
@@ -292,10 +291,19 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SDO-4-1-00",
         vendor: "NodOn",
         description: "Door & window opening sensor",
-        fromZigbee: [fz.battery, fz.ias_contact_alarm_1],
-        toZigbee: [],
-        exposes: [e.contact()],
-        extend: [m.battery({voltageReporting: true})],
+        extend: [m.battery({voltageReporting: true}), m.iasZoneAlarm({zoneType: "contact", zoneAttributes: ["alarm_1"]})],
```
Koenkk: I'm wondering if we should setup reporting automatically in `m.iasZoneAlarm`, then it's applied to all devices.

nthiers: Good point indeed. `m.iasZoneAlarm` already has a `zoneStatusReporting` option that does exactly this (min: "MIN", max: "MAX", change: 0 via setupAttributes). 
One tradeoff, though: it pushes the heartbeat interval from 1h (my custom `configure`, matching NodOn's documented default reporting interval) to "MAX" (~65000s/~18h), and that option doesn't currently take a custom max. It's a meaningful drop in how quickly Z2M would notice an unresponsive device.

Given that, I'd like to keep the custom `configure` here to preserve the tested 1h heartbeat rather than switch to zoneStatusReporting: true as-is.

Koenkk: Then I would propose that we allow passing a custom reporting interval to `m.iasZoneAlarm`.

nthiers: Pushed. Added a `Partial<ReportingConfigWithoutAttribute>` override to `zoneStatusReporting` in `m.iasZoneAlarm` (same pattern as `acFrequency`/`apparentPower` on #12548) — pass `{max: ...}` to customize, boolean still gives the default MIN/MAX/change:0. SDO-4-1-00 now passes `{max: constants.repInterval.HOUR}` directly — no custom `configure` left. Retested on a physical unit, same 1h heartbeat as before.

Koenkk:
Thanks!

PR #12525, opened 2026-06-23
Fix Shelly switch input endpoint discovery

MaxRink:
## Summary
- Resolve Shelly switch input endpoints from the interviewed device instead of always exposing hardcoded `sw1` / `sw2` controls.
- Hide `switch_type` and `switch_mode` controls for Shelly 2PM cover-mode variants that do not expose switch input endpoints.
- Read and write switch mode through the dedicated Shelly RPC endpoint instead of the per-channel switch endpoint.
- Fix `switch_type` reporting so endpoint 3 maps to `sw2` for two-input devices.

## Scope
This is the switch input endpoint cleanup split out from #12497. It intentionally does not change cover tilt exposure or Wi-Fi readback/fallback behavior.

## Validation
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx pnpm@10.18.3 exec vitest run test/shelly.test.ts --config ./test/vitest.config.mts` (`6` tests)
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx pnpm@10.18.3 exec biome check --error-on-warnings src/devices/shelly.ts test/shelly.test.ts`
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx pnpm@10.18.3 run build`
- `git diff --check`

## Live validation
- Verified on my Home Assistant instance that a Shelly 2PM Gen4 cover-mode device without switch input endpoints no longer exposes stale switch type/mode controls.
- Verified on my Home Assistant instance that the stale switch input Home Assistant discovery entities can be cleaned up by the corrected endpoint-aware discovery.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -27,6 +27,23 @@ const checkOption = (device: Zh.Device | DummyDevice, options: KeyValue, key: st
     return defaultValue;
 };
 
+const shellySwitchInputEndpoints = (device: Zh.Device | DummyDevice, endpoints: Record<string, number>): Record<string, number> => {
+    if (utils.isDummyDevice(device)) return endpoints;
+
+    return Object.fromEntries(Object.entries(endpoints).filter(([, endpoint]) => device.getEndpoint(endpoint) !== undefined));
+};
+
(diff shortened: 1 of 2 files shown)


Koenkk:
LGTM, Could you fix the merge conflicts?

MaxRink:
Done

Koenkk:
thanks!

PR #12527, opened 2026-06-23
change _TZ3000_gazjngjl from tuya.ts to zemismart.ts

tl-zigbee:


Koenkk:
how is the device detected after this change?

tl-zigbee:
> how is the device detected after this change?

working well

PR #12528, opened 2026-06-23
change _TZ3000_gazjngjl from tuya.ts to zemismart.ts

tl-zigbee:
_TZ3000_gazjngjl is a 1-gang power with USB instead of 2-gang power. Also do not have power monitoring.


Koenkk:
Please merge this change with your other pr.

PR #12533, opened 2026-06-23
Improve `TS0301_dual_rail_2` (Yoolax day/night shade): real-time position and separate rail covers

msitt:
# Improve Yoolax day/night shade: real-time position and separate rail covers

Reworks the `TS0301_dual_rail_2` definition (`_TZE200_eatmkx5j`, a Yoolax top-down bottom-up / day-night shade) so it reports position in real time, keeps both rails in sync, and exposes the two rails as proper cover entities. The device is now driven entirely through the standard window-covering cluster (`closuresWindowCovering`, 258) rather than the Tuya manufacturer-specific datapoints used by the previous implementation (#12382).

## Existing Issue

The shade has two moving rails (top and bottom) which the firmware exposes as one position value. The existing DP-based definition has two problems:

1. **No real-time position.** The Tuya DP path reports position (DP3) only when the motor stops.
2. **Stale `position_top`.** DP3 was registered twice, so on an incoming report only one converter ran and `position_top` never refreshed.

## Fix

The whole device (control and live readback) can run on cluster 258. This is strictly better than using Tuya DPs because of the real time updates.

The single native position axis `N` (= `currentPositionLiftPercentage`) maps as:

| N | physical state |
|---|----------------|
| 100 | both rails up |
| 50  | both rails down |
| 0   | top up, bottom down |

`N` 51–100 moves the bottom rail (top pinned up); `N` 0–50 moves the top rail (bottom pinned down).

A single `fromZigbee` derives all three covers from one report (guarantees consistent state), and a single `toZigbee` keyed on `state`/`position` handles all three by `endpoint_name`.

## Exposed entities

- `cover` (unified): exposes a unified position cover. same as the generic `TS0301` integration used before the previous PR.
- `cover` (bottom): bottom rail
- `cover` (top): top rail
- `battery`

## New device option: `invert_top_rail`

The `invert_top_rail` option (default `false`, matching the previous convention where `position 100 = down`) lets the user flip the top cover so "open" raises the rail instead of lowering it — e.g. for an up/down unit vs. a day/night unit.

## Hardware limitation

The device exposes a single position value, and the two rails cannot be commanded to arbitrary independent positions over Zigbee. Moving one will reposition the other (e.g. lowering the bottom forces the top up).

## Testing

Verified on hardware (`_TZE200_eatmkx5j`) via an external converter:

- Live position updates while moving each rail (unified, bottom, top)
- Setting position and open/close/stop on each cover
- `invert_top_rail` toggled on/off behaves as expected
- Battery reports

## Refs

- Issue: Koenkk/zigbee2mqtt#32148
- Previous (DP-based) implementation this supersedes: #12382

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -511,9 +511,16 @@ const ar331ProHolidayTimeConverter = {
 };
 
 const tzLocal = {
+    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
     TS0301_dual_rail_2: {
         key: ["state", "position"],
-        options: [e.binary("invert_top_rail", ea.SET, true, false).withDescription("Invert the top rail so that 'open' (position 100) raises the rail instead of lowering it; default false keeps the upstream convention (position 100 = down).")],
+        options: [
+            e
+                .binary("invert_top_rail", ea.SET, true, false)
+                .withDescription(
+                    "Invert the top rail so that 'open' (position 100) raises the rail instead of lowering it; default false keeps the upstream convention (position 100 = down).",
+                ),
+        ],
         convertSet: async (entity, key, value, meta) => {
             const ep = meta.endpoint_name;
             const invertTop = meta.options?.invert_top_rail === true;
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
cc @Sinaloit

Koenkk:
Thanks!

PR #12535, opened 2026-06-24
Feat: add temperature and humidity to M1-PE sensor

Leo2442926161:
Dear Z2M team,
I am adding temperature and humidity to M1-PE sensor, could you reveiew it? thanks.


Koenkk:
Thanks!

PR #12538, opened 2026-06-24
fix: set disableDefaultResponse for several Tuya switch modules

xuw:
These Tuya switch modules — `TS0001_switch_module`, `TS0002_basic`, `TS0003_switch_module_2`, `TS0004`, `TS0011` — frequently do **not** send a ZCL Default Response to on/off commands. With the default response expected, zigbee-herdsman waits the full 10 s command timeout for every command. Because commands to a device are serialized, this stacks badly when several gangs/lights are toggled in quick succession (a scene, or a multi-button scene switch): the lights appear to "hang" for many seconds even though the switch physically actuates.

Setting `meta.disableDefaultResponse` makes on/off fire-and-forget; the new state is still confirmed by the device's own attribute reports. This matches the existing `TS0003` `_TZ3000_ouwfc1qj` variant, which already sets `meta.disableDefaultResponse`.

Verified on real `_TZ3000_*` / `_TZ3210_*` devices: the outgoing `genOnOff` command frame changes from `disableDefaultResponse:false` to `true`, the 10 s timeouts disappear, and attribute-report based state stays correct.

For the multi-endpoint definitions `multiEndpoint` is unchanged (`TS0004` keeps it via `deviceEndpoints`).

Trade-off worth noting for reviewers: for any units of these models that *do* answer reliably, z2m will no longer surface a per-command failure (it relies on the attribute report instead). Given these modules are well-known unreliable responders, fire-and-forget is the better default — happy to scope this to specific manufacturerNames if preferred.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -7036,7 +7036,7 @@ export const definitions: DefinitionWithExtend[] = [
         endpoint: (device) => {
             return {l1: 1, l2: 2};
         },
-        meta: {multiEndpoint: true},
+        meta: {multiEndpoint: true, disableDefaultResponse: true},
         configure: async (device, coordinatorEndpoint) => {
             await tuya.configureMagicPacket(device, coordinatorEndpoint);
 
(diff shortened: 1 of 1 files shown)


andrei-lazarov (commented):
> Given these modules are well-known unreliable responders, fire-and-forget is the better default — happy to scope this to specific manufacturerNames if preferred.

I haven't seen this issue before.. My modules respond correctly.
The fix should be specfic to the respective manufacturer names

But are you sure this isn't a problem with your setup? What is your coordinator model and version?

PR #12539, opened 2026-06-24
added RCS-ST16-z-SlD and LDR-GU10-Mod

slacky1965:
Added RCS-ST16-z-SlD - Socket EKF with power monitoring with custom firmware
and
LDR-GU10-Mod - Smart light Ledron LDR-GU10 with new firmware.


Koenkk:
Thanks!

PR #12540, opened 2026-06-24
fix: Prevent stale Wyze lock raw frames

dkneeland:
Follow-up to #12438 and #12462.

After running those fixes on multiple WLCKG1 locks at home for about a week, I noticed one remaining edge case: the lock can deliver manufacturer-specific raw frames out of order. In one observed sequence, an older `len=85` unlock frame arrived after newer lock frames, causing Zigbee2MQTT to publish a stale unlocked state.

This PR keeps the previous `len=85` / `b80 & 3` decoder intact and only adds sequence suppression before decoding. Byte `3` appears to act as a per-lock raw-frame sequence byte, matching the same general approach used by the ZHA community quirks to ignore duplicate/out-of-order Wyze raw frames.

Observed bad sequence:

```text
b3=89 -> UNLOCK
b3=90 -> LOCK
b3=91 -> short frame
b3=92 -> LOCK
b3=87 -> stale UNLOCK, arrived last
```

Validation:
- Running in my home for about a week without issues across multiple WLCKG1 locks.
- pnpm exec biome check --error-on-warnings src/devices/wyze.ts
- pnpm build
- pre-commit hook: pnpm check, pnpm build, pnpm test (629 tests passed)

### DIFF
--- a/src/devices/wyze.ts
+++ b/src/devices/wyze.ts
@@ -6,12 +6,28 @@ import type {DefinitionWithExtend, Fz} from "../lib/types";
 
 const e = exposes.presets;
 
+const wyzeLockRawSeq = new Map<string, number>();
+
+const isStaleWyzeLockRawFrame = (msg: Fz.Message<64512, undefined, "raw">): boolean => {
+    if (msg.data.length <= 3) return false;
+    const seq = msg.data[3];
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12545, opened 2026-06-25
fix: Tuya TS011F_plug_1 `_TZ3008_reatplte`: report correct voltage/power

### LINKED ISSUE #32404: Nous A7Z (_TZ3008_reatplte): voltage and power reported 10x too large because configure does not save acVoltageDivisor/acPowerDivisor
### What happened?

Voltage and power for the Nous A7Z (`_TZ3008_reatplte`, exposed as `TS011F_plug_1`) are published ~10× too large. Typical readings look like `voltage: 2350`, `power: 1039`, `current: 0.61` — i.e. the current is sensible (1000× divisor applied) but voltage and power are not divided.

Root cause: the device's firmware does expose the correct divisors via `haElectricalMeasurement` attributes — when read manually, it returns:

```
{"acVoltageMultiplier":1,"acVoltageDivisor":10,"acPowerDivisor":10,"acPowerMultiplier":1,"rmsVoltage":2349,"activePower":1039}
```

…but the `TS011F_plug_1` `configure()` step only saves `acCurrentDivisor` to the endpoint attribute cache (see `src/devices/tuya.ts`, around the line `const acCurrentDivisor = ["_TZ3000_typdpbpg"].includes(...)`). Because `acVoltageDivisor` / `acPowerDivisor` are never read or saved during configure, `fz.electrical_measurement` falls back to a factor of `1` (`multiplier && divisor ? multiplier/divisor : 1`) and publishes the raw register values.

After a one-time manual `read haElectricalMeasurement acVoltageDivisor acPowerDivisor` against the device (or after pasting the divisors into the attribute cache), values immediately become correct (`voltage: 235.0`, `power: 103.9`), confirming the analysis.

`voltage` published in volts (~230) and `power` published in watts (~100), as is the case for other TS011F plugs.

1. Pair a Nous A7Z whose manufacturer name is `_TZ3008_reatplte`.
2. Let Z2M run the standard configure for `TS011F_plug_1`.
3. Observe MQTT messages: `voltage` ≈ 2350, `power` ≈ 1000+ when actual load is ~100 W on 230 V mains.

### Zigbee2MQTT version

2.12.0 (commit `fd8bafc8`)

### Adapter firmware version

ZStack3x0 — `{"meta":{"maintrel":1,"majorrel":2,"minorrel":7,"product":1,"revision":20240710,"transportrev":2},"type":"ZStack3x0"}`

### Adapter

CC2652 (zStack) via UZG-01 over TCP

### Setup

Plain install via systemd (`/opt/zigbee2mqtt`, `npm start`) on Debian 12 / Linux 6.8.12 (LXC on Proxmox). Node v22.22.0.

```json
{"id":137,"type":"Router","ieeeAddr":"0xa4c138c82ca945d2","nwkAddr":24810,"manufId":4417,"manufName":"_TZ3008_reatplte","powerSource":"Mains (single phase)","modelId":"TS011F","epList":[1,242],"endpoints":{"1":{"profId":260,"epId":1,"devId":266,"inClusterList":[0,3,4,5,6,1794,2820,57345,57344],"outClusterList":[10,25],"clusters":{"genBasic":{"attributes":{"modelId":"TS011F","manufacturerName":"_TZ3008_reatplte","powerSource":1,"zclVersion":3,"appVersion":66,"stackVersion":1,"hwVersion":1,"dateCode":""}},"haElectricalMeasurement":{"attributes":{"acCurrentDivisor":1000,"acCurrentMultiplier":1,"rmsVoltage":2351,"rmsCurrent":658,"activePower":1164,"acVoltageDivisor":10,"acVoltageMultiplier":1,"acPowerDivisor":10,"acPowerMultiplier":1}},"seMetering":{"attributes":{"divisor":100,"multiplier":1,"currentSummDelivered":8}},"genOnOff":{"attributes":{"onOff":1}}},"configuredReportings":[{"cluster":1794,"attrId":0,"minRepIntval":5,"maxRepIntval":3600,"repChange":257},{"cluster":2820,"attrId":1285,"minRepIntval":5,"maxRepIntval":3600,"repChange":5},{"cluster":2820,"attrId":1288,"minRepIntval":5,"maxRepIntval":3600,"repChange":50},{"cluster":2820,"attrId":1291,"minRepIntval":5,"maxRepIntval":3600,"repChange":10}]}},"appVersion":66,"stackVersion":1,"hwVersion":1,"interviewCompleted":true,"interviewState":"SUCCESSFUL","meta":{"configured":"0.0.0"}}
```

(Note: the `acVoltageDivisor:10` / `acPowerDivisor:10` entries shown above appeared only after a manual read — they are not present on a fresh interview.)

### Notes

A minimal fix (saving the correct divisors during configure for this manufacturer) is proposed in zigbee-herdsman-converters PR — see https://github.com/Koenkk/zigbee-herdsman-converters/pull/12545.

The same problem may exist for the other `_TZ3008_*` variants listed under the A7Z whitelabel (`_TZ3008_1a8m8wd6`, `_TZ3210_rwmitwj4`), but I can only confirm `_TZ3008_reatplte`.

Nirostar:
This device reports rmsVoltage and activePower scaled by 10 (e.g. rmsVoltage=2350 means 235.0V, activePower=1039 means 103.9W) and exposes the correct acVoltageDivisor=10 and acPowerDivisor=10 when those attributes are read. The TS011F_plug_1 configure step only saves acCurrentDivisor in the haElectricalMeasurement attribute cache, so these divisors default to 1 and fz.electrical_measurement publishes values ~10x too large.

Read the voltage and power multipliers/divisors from the device during configure for `_TZ3008_reatplte` so the correct factors end up in the attribute cache. The current attributes (acCurrent*) are intentionally not read because this device returns UNSUPPORTED_ATTRIBUTE for them, like other Tuya TS011F plugs; the existing hardcoded acCurrentDivisor handling is unchanged.

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/32404

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -10037,6 +10037,15 @@ export const definitions: DefinitionWithExtend[] = [
                 acCurrentDivisor,
                 acCurrentMultiplier: 1,
             });
+            // _TZ3008_reatplte reports voltage and power scaled by 10 (e.g. rmsVoltage=2350 means 235.0V).
+            // The device exposes the correct divisors via the haElectricalMeasurement cluster, but
+            // TS011F_plug_1's configure does not read them, so the cache defaults to 1 and values appear 10x too large.
+            // Reading the acCurrent* attributes fails on this device with UNSUPPORTED_ATTRIBUTE (like other
+            // Tuya TS011F plugs, see comment about _TZ3000_0zfrhq4i above), so only the voltage and power
+            // divisors are read from the device here; current uses the hardcoded value saved above.
+            if (device.manufacturerName === "_TZ3008_reatplte") {
+                await endpoint.read("haElectricalMeasurement", ["acVoltageMultiplier", "acVoltageDivisor", "acPowerMultiplier", "acPowerDivisor"]);
+            }
             endpoint.saveClusterAttributeKeyValue("seMetering", {
                 divisor: 100,
                 multiplier: 1,
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -10037,14 +10037,24 @@ export const definitions: DefinitionWithExtend[] = [
                 acCurrentDivisor,
                 acCurrentMultiplier: 1,
             });
-            // _TZ3008_reatplte reports voltage and power scaled by 10 (e.g. rmsVoltage=2350 means 235.0V).
-            // The device exposes the correct divisors via the haElectricalMeasurement cluster, but
-            // TS011F_plug_1's configure does not read them, so the cache defaults to 1 and values appear 10x too large.
-            // Reading the acCurrent* attributes fails on this device with UNSUPPORTED_ATTRIBUTE (like other
-            // Tuya TS011F plugs, see comment about _TZ3000_0zfrhq4i above), so only the voltage and power
-            // divisors are read from the device here; current uses the hardcoded value saved above.
-            if (device.manufacturerName === "_TZ3008_reatplte") {
-                await endpoint.read("haElectricalMeasurement", ["acVoltageMultiplier", "acVoltageDivisor", "acPowerMultiplier", "acPowerDivisor"]);
+            // Some TS011F plugs (e.g. _TZ3008_reatplte) report voltage/power/current scaled by a non-default
+            // factor and expose the correct multipliers/divisors via the haElectricalMeasurement cluster.
+            // Try to read each attribute individually because some devices return UNSUPPORTED_ATTRIBUTE for a
+            // subset, and reading them in bulk would cause one failure to abort the rest. Successful reads are
+            // cached by zigbee-herdsman and override the hardcoded values saved above.
+            for (const attr of [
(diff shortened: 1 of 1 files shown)


andrei-lazarov:
Maybe also gate this by `applicationVersion`, in addition to `manufacturerName`

andrei-lazarov:
Actually would it be better to wrap this in a try-catch?
So it automatically attempts on every plug, and we skip the list of models/versions.

Nirostar:
I agree, we should query and try/catch them individually though because if one isn't supported it will fail.

andrei-lazarov:
I would guess that if one of these attributes is supported, all four are supported.
`["acVoltageMultiplier", "acVoltageDivisor", "acPowerMultiplier", "acPowerDivisor"]`

Nirostar:
It is probably safer to just query them individually to avoid errors. I updated the PR

Koenkk:
Thanks!

PR #12548, opened 2026-06-25
feat(NodOn): add SEM-4-1-00 energy reset, configureReporting and apparent power

nthiers:
### Problem

The existing entry uses `m.electricityMeter()` without a `configure` function.
As a result:
- The device falls back to firmware-default continuous reporting (~5 s interval regardless of
  load). A future firmware update will implement threshold-based reporting natively, using the
  same thresholds as defined in this PR.
- No `energy_reset` capability
- `apparentPower` and `acFrequency` missing from `configureReporting`

### Fix

- Add explicit `configure` with manual bind, multiplier reads, and batched `configureReporting`
- Add `energy_reset` via `genBasic/resetFactDefault`
- Add `m.identify()`

#### Why `configureReporting: false` + explicit configure

Without `configure`, the device uses factory-default reporting (~5 s for all attributes).
With threshold-based reporting, ×63 fewer messages were observed on a stable circuit.

Manual bind is required because `m.electricityMeter({ configureReporting: false })` skips its
own bind/configure step entirely.

Multiplier attributes (`acVoltageMultiplier`/`Divisor`, `acCurrentMultiplier`/`Divisor`,
`acPowerMultiplier`/`Divisor`, `acFrequencyMultiplier`/`Divisor`) are read before
`configureReporting` so ZHC has them in cluster cache to scale the first received reports.

`haElectricalMeasurement configureReporting` is split into batches of ≤5 attributes: a single
call with 6 attributes exceeds the ~85-byte ZCL frame limit and returns `status=FAIL`.

#### Why `energy_reset` via `genBasic/resetFactDefault`

This is the only counter-reset mechanism available in the current firmware. `resetFactDefault`
resets all cluster attributes to factory defaults — it does not affect Zigbee network membership,
bindings, or `configureReporting` (ZCL spec).

### Tested

- Device: SEM-4-1-00
- `voltage`, `current`, `power`, `power_apparent`, `ac_frequency`, `power_factor`, `energy`,
  `produced_energy` confirmed working
- `energy_reset` confirmed working
- `identify` confirmed working

### DIFF
--- a/src/devices/nodon.ts
+++ b/src/devices/nodon.ts
@@ -329,7 +329,7 @@ export const definitions: DefinitionWithExtend[] = [
                 .withCategory('config'),
             e.power_apparent(),
         ],
-        configure: async (device, coordinatorEndpoint, logger) => {
+        configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
 
             // Explicit bind — not done by m.electricityMeter because configureReporting: false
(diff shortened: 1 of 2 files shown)


src/devices/nodon.ts
```diff
@@ -304,13 +304,66 @@ export const definitions: DefinitionWithExtend[] = [
...
+            },
+        ],
+        exposes: [
+            e.enum("energy_reset", ea.SET, ["reset"]).withDescription("Reset all energy counters to 0").withCategory("config"),
+            e.power_apparent(),
+        ],
+        configure: async (device, coordinatorEndpoint) => {
+            const endpoint = device.getEndpoint(1);
+
+            // Explicit bind — not done by m.electricityMeter because configureReporting: false
```
Koenkk: Why do we not set `configureReporting: true`?

nthiers: During my tests with the external converter, I observed that a single `configureReporting` call with **6+** `haElectricalMeasurement` attributes returns` status=FAIL` on our device. Splitting into batches of ≤ 5 attributes solved that.

The custom configure also reads multiplier/divisor attributes _before_ the first `configureReporting` call, so ZHC has the correct scaling factors cached when the first reports arrive.

Koenkk: > During my tests with the external converter, I observed that a single configureReporting call with 6+ haElectricalMeasurement attributes returns status=FAIL on our device. Splitting into batches of ≤ 5 attributes solved that.

For this I propose to add an argument to `m.electricityMeter` to change this behaviour.

> so ZHC has the correct scaling factors cached when the first reports arrive.

I think we should also do this in the `m.electricityMeter` , read `divisor`/`multiplier` before doing the `configureReporting`

nthiers: You're right, `genericMeter` already does this — bind, multiplier/divisor read before `configureReporting`, and batching (chunks of 4) happen automatically as long as `configureReporting` stays `true`. Updated the PR to drop the custom `configure` for `voltage`/`current`/`power`/`acFrequency`/`energy`/`producedEnergy` — plain per-attribute thresholds passed to `m.electricityMeter()` now.

Two attributes still need something extra, `MeterArgs` has no key for either:
- `apparentPower`: not in `configureLookup`, `fz.electrical_measurement` parses it but nothing asks the device to report it. Reused `setupAttributes()` in a small `configure`.
- `powerFactor`: typed `boolean` only, no per-attribute override. Kept the generic default (10s/~18h/10%).

Retested on 3 physical units — thresholds apply automatically, `power_apparent` reports correctly now.

Koenkk: > Two attributes still need something extra, MeterArgs has no key for either:

Then I would propose to add these to `m.electricityMeter()`

nthiers: Pushed. Added `apparentPower` to `MeterArgs`/`configureLookup` (same pattern as `acFrequency`, opt-in, shares `activePower`'s divisor/multiplier) and gave `powerFactor` a proper override instead of `boolean`-only. SEM-4-1-00 now passes both directly to `m.electricityMeter()` — no custom `configure` left. Retested on 3 physical units, same reporting config as before.

Koenkk:
Thanks!

sbraz:
@nthiers shouldn't a reconfiguration be triggered on Z2M upgrade by bumping the definition version?

I have 2 of these devices and only reconfigured only one just to check what happens to the message rate: it goes from 0.2 messages per second to 0.01. Everyone should benefit from this change automatically.

I can create a PR to do this if you want.

nthiers:
Thanks @sbraz, good catch!

PR #12550, opened 2026-06-25
Fix Shelly Wi-Fi readback via RPC fallback

MaxRink:
## Summary
- read Shelly Wi-Fi configuration/status via the RPC cluster before falling back to the Wi-Fi setup cluster
- cache the full SSID from RPC reads and allow `shelly_wifi_ssid` as a manual fallback when the setup cluster only reports a shortened SSID
- keep Wi-Fi `/get` readback non-fatal when RPC and setup-cluster reads are unavailable

## Tests
- `pnpm test -- test/shelly.test.ts`
- `pnpm run check`
- `pnpm run build`
- `git diff --check`
- pre-commit hook: `pnpm run build`, `pnpm run check`, `pnpm test`


Koenkk:
Thanks!

PR #12553, opened 2026-06-25
Add flash_on_click attribute to Schneider device

lajtmaN:
The default mode from the factory is `4`. In this mode, it'll turn the LED white for a short while after clicking it.


Koenkk:
Thanks!

PR #12554, opened 2026-06-25
Energy all in 3 phase mercury

Bacchus777:
added Energy all in 3 phase mercury


src/devices/bacchus.ts
```diff
@@ -313,6 +313,17 @@ export const definitions: DefinitionWithExtend[] = [
                 precision: 3,
                 reporting: electicityReporting,
             }),
+            energy({
+                name: "energy_all",
```
Koenkk, line 317:
```suggestion
                name: "energy_combined",
```

Maybe this is a better name?

Bacchus777: Okay, lets it will be combined!

Koenkk:
Thanks!

PR #12557, opened 2026-06-26
fix: Tuya TS0601_smart_scene_knob: Allow brightness 255

andrei-lazarov:
This workaround is not needed anymore cc @Bjk8kds 
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/11103

255 brightness is actually valid for the "generic" Level control cluster.
It's only invalid in the "derived" Level control for lighting. (whatever that means)

I fixed it in ZH here:
- https://github.com/Koenkk/zigbee-herdsman/pull/1752


Koenkk:
Thanks!

PR #12566, opened 2026-06-27
feat: IKEA E2134: support on duration, trigger condition

jeslinmx:
The rear of the VALLHORN motion sensors has 2 buttons other than the link button. The top one controls how long it should maintain the on state for bound devices, the bottom one controls whether it should always trigger, or only when it is dark (seemingly around ~75 lumens measured by the onboard sensor).

This was implemented in zigpy but not on here, so I referenced that for the specific ID values, hope that's alright, I'm quite new to Z2M, Zigbee, and home automation in general.


andrei-lazarov:
This is huge, especially the custom onTime! Thanks for implementing.

I also saw the unused cluster and poked around. Never thought to check zigpy, good find!

I was preparing an implementation 3 months ago (here https://github.com/Koenkk/zigbee-herdsman-converters/commit/84e84dff1d33b08a79bcc112cb58ec90dcef28d3), but I gave up because the dark mode setting didn't actually change the behavior for my units.
I took one outside in broad daylight, and it still turned on the light. I understand it works for you? Did you test the code, or only the button?

Koenkk:
Thanks, could you change this to use `m.numeric` and `m.binary`? (search this repo for examples)

andrei-lazarov:
Copy from the commit I linked https://github.com/Koenkk/zigbee-herdsman-converters/commit/84e84dff1d33b08a79bcc112cb58ec90dcef28d3

jeslinmx:
> I also saw the unused cluster and poked around. Never thought to check zigpy, good find!

Came over from ZHA because they were missing some Tuya devices I recently bought, then realized that some VALLHORN functionality was missing, so quite naturally ported it over.

> I understand it works for you? Did you test the code, or only the button?

Yup, fully tested. Pressing the button causes the change to reflect in Z2M, changing the value in Z2M changes the behaviour on the device. Which makes me realise you could also use this as a ersatz 2-button toggle switch if you were so inclined.

> Thanks, could you change this to use m.numeric and m.binary?

Will do, tomorrow!

image

By the way, my exposes don’t have any icons. Could someone point me to the relevant docs for assigning icons if it would be appropriate?

andrei-lazarov:
Yes! You can add icons here, based on the feature names. (even change the icons based on the state)
https://github.com/Nerivec/zigbee2mqtt-windfront/blob/main/src/components/features/index.tsx
https://fontawesome.com/search?ic=free-collection

jeslinmx:
All done!

The existing implementation of the IKEA devices is somewhat mismatched with the conventions of the other devices I referenced (interfaces in `lib/` rather than `devices/`, extend definitions in lib rather than `m.whatever` directly inline in device definition), so I guessed at a reasonable compromise. Do let me know if this needs to be adjusted one way or the other.

Koenkk:
thanks!

basrieter:
Out of curiosity: would the [E1525/E1745](https://www.zigbee2mqtt.io/devices/E1525_E1745.html) also support this? It has the same buttons?

andrei-lazarov:
I will check the E1745

andrei-lazarov:
I checked the E1745, both before and after the ota update from Z2M.
It doesn't have this custom cluster, so I don't think it's possible to customize it.

And it has different buttons actually! One is brightness threshold, the other is brightness for the lights: 30 or 100%. The on time is always 3min.

image
image

The E1525 has dials for brightness threshold and on time. So I believe those already give a variable range instead of fixed numbers. I don't own it, but I don't think the attributes are accessible from Z2M.

There isn't anything similar in zigpy
https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/ikea/motionzha.py
https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/ikea/motion.py

basrieter:
Ok. That is a too bad. But thanks for checking!

PR #12234, opened 2026-05-17
feat(add): Somfy Situo 4 Zigbee 1800195

joneshf:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/5131.

Following the documentation at https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html for adding support for the Somfy Situo 4 Zigbee blinds remote.

PR #12569, opened 2026-06-28
feat(add): Somfy Situo 1 Zigbee 1800194

joneshf:
Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/5284.

Following the documentation at
https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html for adding support for the Somfy Situo 1 Zigbee blinds remote.

This is the single-channel version of the remote added in https://github.com/Koenkk/zigbee-herdsman-converters/pull/12234.

### DIFF
--- a/src/devices/somfy.ts
+++ b/src/devices/somfy.ts
@@ -145,6 +145,13 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [m.onOff(), m.electricityMeter()],
     },
     {
+        zigbeeModel: ["Situo 1 Zigbee"],
+        model: "1800194",
+        vendor: "SOMFY",
+        description: "Situo 1 channel blinds remote",
+        extend: [m.battery(), m.commandsOnOff(), m.commandsWindowCovering()],
+    },
+    {
         zigbeeModel: ["Situo 4 Zigbee"],
         model: "1800195",
         vendor: "SOMFY",


src/devices/somfy.ts
```diff
@@ -144,6 +144,13 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Connected plug E type with power monitoring",
         extend: [m.onOff(), m.electricityMeter()],
     },
+    {
+        zigbeeModel: ["Situo 1 Zigbee"],
+        model: "1800194",
+        vendor: "SOMFY",
+        description: "Situo 1 channel blinds remote",
+        extend: [m.deviceEndpoints({endpoints: {"1": 1, "232": 232}}), m.battery(), m.commandsOnOff(), m.commandsWindowCovering()],
```
Koenkk:
```suggestion
        extend: [m.battery(), m.commandsOnOff(), m.commandsWindowCovering()],
```
Could you check if this also works?

joneshf: I'm not really sure what I'm looking for, but I added an external converter without `m.deviceEndpoints` and the remote seems to function the same way as before. I don't know if that's because it was already setup, or if it's because `m.deviceEndpoints` isn't necessary. Regardless, I'm down to remove this, and if it turns out to be necessary in the future I can make a PR for that.

Koenkk:
Thanks!

joneshf:
Thanks for the review/merge!

PR #12571, opened 2026-06-28
Add additional fingerprint for additional TS0301_dual_rail device

Sinaloit:
Add _TZE210_yqwse3h5 as a fingerprint for the TS0301_dual_rail definition. This adds support for some Yoolax Top-Down Bottom-Up shades. (Apparently, they ship different motor models or changed recently as I have different fingerprints for some of my shades).

Verified functionality works as expected.


Koenkk:
Thanks!

PR #12573, opened 2026-06-28
Add fingerprint to detect _TZE28C1000000_y4jqpry8 as Zemismart ZMS-206US-4

Lukkasss:
Hello.

This PR is a follow-up of https://github.com/Koenkk/zigbee2mqtt/issues/32091

It seems that we have a new fingerprint for the Zemismart touch switch ZMS-206US-4, the datapoints remain the same so i've updated the zemismart code to add the new string

This is the switch

https://aliexpress.com/item/1005010390685707.html
https://www.zemismart.com/products/zms-206us-1c?VariantsId=17920


Koenkk:
Thanks!

PR #12577, opened 2026-06-28
fix(lumi.sensor_occupy.agl8): Adjust min/step values for absence delay timer

absent42:
Modified devices/lumi.ts

- Latest firmware for lumi.sensor_occupy.agl8 allows absence delay timer value to go down to 1 second.


Koenkk:
thanks!

PR #12582, opened 2026-06-29
feat(add): WB-MSW-ZIGBEE v.4 (Official firmware)

u236:
Adds the official Wiren Board Zigbee 3.0 firmware variant of the WB-MSW v.4 multi-sensor. Distinct firmware/ZCL stack from the existing Sprut-based `WBMSW4`, matched by a different `zigbeeModel` so the two don't overlap. Manufacturer-specific attributes and the TVOC concentration cluster `0x042e` (defined in Matter) are registered via `deviceAddCustomCluster`; reporting is firmware-driven, so clusters are only bound. Tested on hardware.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/5286

### DIFF
--- a/src/devices/wirenboard.ts
+++ b/src/devices/wirenboard.ts
@@ -769,7 +769,7 @@ const {
 export const definitions: DefinitionWithExtend[] = [
     {
         zigbeeModel: ["WB-MSW-ZIGBEE v.4"],
-        model: "WB-MSW-ZIGBEE v.4 (Official firmware)",
+        model: "WB-MSW-ZIGBEE_v.4_official",
         vendor: "Wiren Board",
         description: "Wall-mounted multi sensor with official Wiren Board firmware",
         ota: true,
--- a/src/devices/wirenboard.ts
+++ b/src/devices/wirenboard.ts
@@ -12,6 +12,44 @@ import {assertString, getFromLookup, getOptions, toNumber} from "../lib/utils";
 const e = exposes.presets;
 const ea = exposes.access;
 
+interface WbGenAnalogInput {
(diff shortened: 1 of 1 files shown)


src/devices/wirenboard.ts
```diff
@@ -729,6 +767,269 @@ const {
 } = sprutModernExtend;
 
 export const definitions: DefinitionWithExtend[] = [
+    {
+        zigbeeModel: ["WB-MSW-ZIGBEE v.4"],
+        model: "WB-MSW-ZIGBEE v.4 (Official firmware)",
```
Koenkk:
```suggestion
        model: "WB-MSW-ZIGBEE_v.4_official",
```

u236:
Thanks for the review! Renaming `model` to `WB-MSW-ZIGBEE v.4` collides with the existing definition further down (line ~1190: `zigbeeModel: ["WBMSW4"]`, `model: "WB-MSW-ZIGBEE v.4"`), which is the same hardware running the stock firmware. Since `model` must be unique, the two would clash — and we can't rename or remove that `WBMSW4` definition without breaking current users of the stock firmware.

They're genuinely two separate devices from z2m's point of view — different `zigbeeModel` (`WB-MSW-ZIGBEE v.4` for our official firmware vs `WBMSW4` for the stock one) and different clusters/features. The `(Official firmware)` suffix was there specifically to disambiguate the two while keeping both working. Would you be okay keeping a distinguishing suffix — and if so, is there a wording you'd prefer?

Koenkk:
Thanks!

PR #12583, opened 2026-06-29
OWON PIR313-E: expose battery percentage and configure reporting

gluoncompany:
Adds battery percentage exposure (e.battery()) to the OWON PIR313-E and configures attribute reporting that was previously missing: battery percentage on endpoint 1, plus temperature/humidity reporting intervals on the measurement endpoint. The existing occupancy behaviour and all other functionality are kept unchanged.

### DIFF
--- a/src/devices/owon.ts
+++ b/src/devices/owon.ts
@@ -909,19 +909,21 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Motion sensor",
         fromZigbee: [fz.battery, fz.ias_occupancy_alarm_1, fz.temperature, fz.humidity, fz.occupancy_timeout],
         toZigbee: [],
-        exposes: [e.occupancy(), e.tamper(), e.battery_low(), e.temperature(), e.humidity()],
+        exposes: [e.occupancy(), e.tamper(), e.battery_low(), e.battery(), e.temperature(), e.humidity()],
         configure: async (device, coordinatorEndpoint) => {
+            const endpoint1 = device.getEndpoint(1);
             const endpoint2 = device.getEndpoint(2);
             const endpoint3 = device.getEndpoint(3);
-            if (device.modelID === "PIR313") {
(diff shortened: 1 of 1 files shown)


src/devices/owon.ts
```diff
@@ -909,19 +909,21 @@ export const definitions: DefinitionWithExtend[] = [
...
+            await reporting.batteryPercentageRemaining(endpoint1, {min: 3600, max: 65000, change: 10});
+            const measurementEndpoint = device.modelID === "PIR313" ? endpoint3 : endpoint2;
+            await reporting.bind(measurementEndpoint, coordinatorEndpoint, ["msTemperatureMeasurement", "msRelativeHumidity"]);
+            await reporting.temperature(measurementEndpoint, {min: 60, max: 3600, change: 50});
+            await reporting.humidity(measurementEndpoint, {min: 60, max: 3600, change: 100});
             device.powerSource = "Battery";
             device.save();
         },
-        extend: [m.illuminance()],
+        extend: [m.illuminance({reporting: {min: 300, max: 3600, change: 100}})],
```
Koenkk, line 926:
```suggestion
        extend: [m.illuminance({reporting: {min: 300, max: 3600, change: 100}}), m.battery()],
```
This change alone should be enough. Also why is the reporting config for illuminance changed?

gluoncompany: Thanks for the review!

Good point on m.battery() — I'll switch to that and remove the manual battery code (e.battery(), fz.battery, the genPowerCfg bind and batteryPercentageRemaining in configure). One question: should I keep the previous reporting interval (min: 3600, max: 65000, change: 10) by passing it as an option to m.battery(), or are the defaults preferred here?

About the illuminance reporting change: it wasn't accidental. With the default config the device was waking up very frequently to report illuminance, which noticeably increased battery drain. Setting {min: 300, max: 3600, change: 100} reduced the wake-ups and improved battery life on this sensor, which is why I included it. Happy to revert it to a separate PR if you'd prefer to keep this one focused only on the battery exposure — just let me know.

Koenkk:
Thanks!

PR #12600, opened 2026-07-02
Fix SONOFF SWV configure when optional attribute is unsupported

### LINKED ISSUE #12599: SONOFF SWV configure fails when optional water shortage attribute is unsupported
### What happened?

A SONOFF SWV water valve can fail its Zigbee2MQTT configure step when the device does not support the private eWeLink attribute used for `auto_close_when_water_shortage`.

Observed log:

```text
[2026-07-01 18:22:49] info: z2m: Configuring 'Water valve'
[2026-07-01 18:23:02] error: z2m: Failed to configure 'Water valve', attempt 2 (Error: ZCL command 0xe406bffffef64c52/1 customClusterEwelink.read([20492,20497], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"reservedBits":0,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE'))
```

The decimal attributes are:

- `20492` = `0x500c` / `current_device_status`
- `20497` = `0x5011` / `lackWaterCloseValveTimeout`, exposed as `auto_close_when_water_shortage`

The SWV docs note that `auto_close_when_water_shortage` requires firmware `1.0.4` or later, so older firmware can reject `0x5011` with `UNSUPPORTED_ATTRIBUTE`.

The configure step should continue when this optional firmware-gated private attribute is unsupported. The valve's normal on/off behavior and supported status read should not be marked failed because `0x5011` is missing.

### How to reproduce it?

Configure an SWV water valve with firmware that does not support `lackWaterCloseValveTimeout` / `0x5011`. Zigbee2MQTT attempts `customClusterEwelink.read([0x500c, 0x5011])`, the device returns `UNSUPPORTED_ATTRIBUTE`, and configure is marked failed.

### Proposed fix

Read `0x500c` and `0x5011` independently and make failures non-fatal, matching the optional nature of these private attributes. This preserves configure success for older firmware while still reading the attributes when available.

### Zigbee2MQTT version

Observed with `zigbee-herdsman-converters@26.76.0` in Zigbee2MQTT.

### Adapter firmware version

No response

### Adapter

No response

### Setup

No response

### Debug log

No response

klinquist:
## Summary

Fixes #12599.

This makes the SONOFF SWV configure step tolerate missing optional private eWeLink attributes:

- reads `current_device_status` (`0x500c`) independently
- reads `lackWaterCloseValveTimeout` (`0x5011`) independently
- logs a warning instead of failing configure when either optional read is rejected

## Why

`auto_close_when_water_shortage` depends on SWV firmware `1.0.4` or later. Devices without that attribute can return `UNSUPPORTED_ATTRIBUTE` for `0x5011`, which currently makes the whole configure step fail even though the valve can otherwise operate normally.

## Tests

- `pnpm exec vitest run --config ./test/vitest.config.mts test/sonoff.test.ts`
- `pnpm run check`
- commit hook: `pnpm run build`
- commit hook: `pnpm run check`
- commit hook: `pnpm test`

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -7578,7 +7578,12 @@ export const definitions: DefinitionWithExtend[] = [
             await reporting.bind(endpoint, coordinatorEndpoint, ["genPowerCfg", "genOnOff"]);
             await reporting.bind(endpoint, coordinatorEndpoint, ["msFlowMeasurement"]);
             await reporting.onOff(endpoint, {min: 1, max: 1800, change: 0});
-            await endpoint.read("customClusterEwelink", [0x500c, 0x5011]);
+            await endpoint.read("customClusterEwelink", [0x500c]).catch((error) => {
+                logger.warning(`SWV read customClusterEwelink(current_device_status) failed: ${error}`, NS);
+            });
+            await endpoint.read("customClusterEwelink", [0x5011]).catch((error) => {
+                logger.warning(`SWV read customClusterEwelink(lackWaterCloseValveTimeout) failed: ${error}`, NS);
+            });
         },
     },
     {
--- a/test/sonoff.test.ts
+++ b/test/sonoff.test.ts
@@ -549,6 +549,32 @@ describe("Sonoff TRVZB", () => {
     });
 });
 
+describe("Sonoff SWV", () => {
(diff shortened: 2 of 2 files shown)


Koenkk:
Thanks!

PR #12405, opened 2026-06-09
feat(add): Zemismart KES-606US-LH1 (1-gang) and KES-606US-L2 (2-gang) with scene mode support

chou0706:
## Devices added

| Model | `manufacturerName` | Gangs | Scene Mode |
|---|---|---|---|
| Zemismart KES-606US-LH1 | `_TZ3000_ovbvmhiq` | 1 | ✅ |
| Zemismart KES-606US-L2  | `_TZ3000_icoxotza` | 2 | ✅ |

## Key finding: Scene mode requires Smart Life pre-initialization

Scene mode **does work**, but requires a one-time initialization via the Tuya Smart Life app before pairing to Z2M:

1. Pair the switch to Smart Life
2. Set the desired keys to **Scene Mode** in Smart Life device settings
3. Remove the device from Smart Life
4. Reset the switch and re-pair to Z2M

Without this pre-initialization, setting `switch_mode` to `scene` will not trigger action events.

## Features supported

- `switch_mode`: `switch` / `scene` (per endpoint)
- `action`: `scene_1` / `scene_2`
- `power_on_behavior`
- `indicator_mode`: `none` / `relay` / `pos` (via `genOnOff` attribute `0x8001`)
- `configureMagicPacket` required for independent per-gang control

Tested on Z2M Edge v2.7.2-1.

PR #12603, opened 2026-07-02
fix: add indicator_mode converter for Zemismart KES-606US (TS0726_1_gang/TS0726_2_gang)

chou0706:
Follow-up fix to #12405.

When the Zemismart KES-606US definitions were merged into `TS0726_1_gang` and `TS0726_2_gang`, the `indicator_mode` enum was added to `exposes`, but the corresponding toZigbee/fromZigbee converters were not included. As a result the `indicator_mode` control is shown in the UI but has no effect — setting it does nothing because there is no converter to write the attribute.

This adds:
- `tzLocal.TS0726_indicator_mode` — writes `genOnOff` attribute `0x8001` (none=0, relay=1, pos=2)
- `fzLocal.TS0726_indicator_mode` — reads back the same attribute
- Both wired into the `fromZigbee`/`toZigbee` of `TS0726_1_gang` and `TS0726_2_gang`

The `0x8001` attribute and value mapping were verified on real hardware (Zemismart KES-606US-LH1 `_TZ3000_ovbvmhiq` and KES-606US-L2 `_TZ3000_icoxotza`) — indicator mode now switches correctly between none/relay/pos.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -615,22 +615,6 @@ const tzLocal = {
             return {state: {switch_mode: value}};
         },
     } satisfies Tz.Converter,
-    // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
-    TS0726_indicator_mode: {
-        key: ["indicator_mode"],
-        convertSet: async (entity, key, value, meta) => {
-            await entity.write("genOnOff", {
-                32769: {
-                    value: utils.getFromLookup(value, {none: 0, relay: 1, pos: 2}),
-                    type: 0x30,
-                },
-            });
-            return {state: {indicator_mode: value}};
-        },
-        convertGet: async (entity, key, meta) => {
-            await entity.read("genOnOff", [0x8001]);
-        },
-    } satisfies Tz.Converter,
     led_control: {
         key: ["brightness", "color", "color_temp", "transition"],
         options: [exposes.options.color_sync()],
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -11736,8 +11725,8 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Tuya",
         description: "1 gang switch with neutral wire",
         extend: [tuya.modernExtend.tuyaBase()],
-        fromZigbee: [fz.on_off, tuya.fz.power_on_behavior_2, fzLocal.TS0726_action],
-        toZigbee: [tz.on_off, tuya.tz.power_on_behavior_2, tzLocal.TS0726_switch_mode],
+        fromZigbee: [fz.on_off, tuya.fz.power_on_behavior_2, fzLocal.TS0726_action, fzLocal.TS0726_indicator_mode],
+        toZigbee: [tz.on_off, tuya.tz.power_on_behavior_2, tzLocal.TS0726_switch_mode, tzLocal.TS0726_indicator_mode],
```
Koenkk: Please use `m.enumLookup` instead, examples: https://github.com/search?q=repo%3AKoenkk%2Fzigbee-herdsman-converters%20m.enumLookup&type=code

chou0706:
Thanks for the review! I've switched both TS0726_1_gang and TS0726_2_gang to use `m.enumLookup` for `indicator_mode` instead of the custom converters. Since enumLookup applies to the whole model, `indicator_mode` is now exposed for all variants rather than gated by manufacturerName — let me know if you'd prefer it scoped differently. Updated and CI is green.

Koenkk:
Thanks!

PR #12607, opened 2026-07-02
Enhance Tuya device configuration with electricity meter

hml96:
Updated the Tuya device configuration to include electricity meter support and refined the extend and fromZigbee/toZigbee properties.


hml96:
The Tuya socket definitions have become fragmented and inconsistent across many contributors. We are consolidating them under the main TS011F model definition, and removing the faulty fingerprint matches that rely on modelID + applicationVersion.

PR #12610, opened 2026-07-02
fix(detect): Detect `_TZ3210_lqb7lcq9` as Nova Digital SA-WK

felipemathies:
Nova Digital Safira "Full Switch" (**SA-WK**): 1 physical gang + 20A socket + USB-A + USB-C, 4x2 wall module. It has four independently switchable relays (endpoints 1–4, each with `genOnOff`) and **no** energy metering.

It reports as `TS011F` / `_TZ3210_lqb7lcq9` and currently falls back to `TS011F_plug_1` (single gang + non-existent power monitoring). It is the same 4-gang hardware already covered by `TS011F_2_gang_2_usb_wall`, which includes the Nova Digital **NTS2-W-B** via `_TZ3000_lqb7lcq9` (same base code, newer `_TZ3210_` chip revision).

This PR adds the manufacturer name to the existing definition and a `whiteLabel` entry for the SA-WK.

**Device fingerprint (from the Z2M "Clusters" tab):** endpoints 1–4 each expose `genOnOff`; endpoint 242 is Green Power; no `seMetering`/`haElectricalMeasurement` cluster.

`pnpm check`, `pnpm build` and `pnpm test` (646 tests) all pass locally.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -6565,7 +6565,14 @@ export const definitions: DefinitionWithExtend[] = [
         exposes: [e.battery(), e.temperature(), e.humidity(), e.battery_voltage()],
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12616, opened 2026-07-04
Update moes.ts

emolenev:
Adds support for the Moes BHT-002 thermostat variant with fingerprint `_TZE200_5toc8efa`. This device is commonly used for boiler temperature control and supports temperature setting in the range of 5–90 °C.


emolenev:
https://ali.click/tw7dg12

Koenkk:
Thanks!

PR #12620, opened 2026-07-04
Keep Shelly cover tilt controls visible when supported

MaxRink:
Per the message in this thread, Shelly slat/tilt discovery was hiding tilt controls unless cover_tilt_enabled was explicitly true.

This change keeps tilt controls visible by default and only hides them when the device or metadata explicitly disables tilt support.

Fixes the Shelly slat discovery report from the thread.


Koenkk:
Thanks!

PR #12621, opened 2026-07-04
Add support for OSRAM 72569 to existing Sylvania device

SiliconAvatar:
Adding alternate device name for existing Sylvania device model 72569.
The version of this device I have identifies as being manufactured by Osram but I've read there's a bit of a confusing history of early Osram/Sylvania Lightify devices.

They are the same devices as far as I can tell, and the existing device photo is already correct.


Koenkk:
Thanks!

PR #12623, opened 2026-07-05
Add EC1366

mkrle:
This is a mirror cabinet that comes with Zigbee lighting pre-installed. The controller (and probably the rest of the lighting) is manufactured by L&S Lighting. 

There are a lot of variants of the mirror cabinet (finishes, dimensions, etc.) and I can only assume they all use an identical Zigbee module, which is why I haven't used the VB part number for `model`. There can be only one photo so I will use the one from my specific variant (`B48110VJ`).

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/5304


Koenkk:
Thanks!

PR #12383, opened 2026-06-05
fix: Aqara FP300 (PS-S04D) configure reporting for presence and PIR

drauziooppenheimer:
Fixes the Aqara FP300 (`PS-S04D` / `lumi.sensor_occupy.agl8`) not updating `presence` / `pir_detection` autonomously after a firmware update, they only refreshed on a manual read. See https://github.com/Koenkk/zigbee2mqtt/issues/32153.

All credits to @ericzizzi who debugged that issue and proposed the fix.

I haven't tested this yet, we can check if other folks can test that too.

PR #12624, opened 2026-07-05
fix: Aqara PS-S04D: fix battery never reporting

hajarada:
## Problem

On firmware `0.0.0_6542` (the only firmware in the wild for the FP300, fileVersion 16682) the device never pushes the manuSpecificLumi `0x00F7` struct that carries the battery data. #12383 fixed presence/PIR reporting, but `battery`/`voltage` still stay empty forever, and nothing in the definition ever triggers the struct to be sent.

Related: https://github.com/Koenkk/zigbee2mqtt/issues/32153

## Fix

Three pieces, all scoped strictly to `PS-S04D`:

1. **Prime at configure** — `configure` now reads `0x00F7` once (the device is awake during pairing/reconfigure, same as the six reads already there). Definition `version` bumped to `0.0.2` so existing devices re-run configure.
2. **Staleness-gated poll of `0x00F7`** (`fp300BatteryPoll()` in lib/lumi.ts): a fromZigbee converter records when the struct was last received; an hourly timer (via `m.poll`, same pattern as `pollBatteryVoltage` in `profalux.ts`) reads it back only when none has arrived for 4 h. Firmware that pushes the struct autonomously keeps the poll permanently dormant; firmware that doesn't (0.0.0_6542, or any future build with the same regression) self-regulates to a read every ~4–5 h. The read uses `sendPolicy: "queue"`: the FP300 is a sleepy end device, so the read fails immediately and is queued; zigbee-herdsman's implicit check-in delivers it the next time the device sends anything. Since #12383 configures presence reporting with a 3600 s max interval, the device wakes at least hourly — the queued read is always served well before it expires.
3. **`m.quirkCheckinInterval("1_HOUR")`** — the FP300 has no genPollCtrl cluster, so without this `pendingRequestTimeout` is 0 and queued requests are dropped instead of queued (this is the documented purpose of the quirk, already used by many other lumi definitions in this file).

Additionally, this resolves the existing `//percentageAttribute: 0x0018 // TODO: Should confirm to be sure`: confirmed. `lumiBattery` now uses struct tag 24 (`0x0018`, uint8 %) — the device's own battery gauge — as the `battery` source instead of the voltage-curve estimate, with tag 23 (`0x0017`, uint16 mV) still exposed as `voltage`.

Struct layout on `0.0.0_6542`:

| tag | type | meaning |
|---|---|---|
| 23 (0x17) | uint16 | battery voltage, mV |
| 24 (0x18) | uint8 | battery percentage (device's own gauge) |

## Evidence

Running for ~1 month in production on 11 FP300s (all fw `0.0.0_6542`) as an external converter that does functionally the same thing (reading `0x00F7` whenever the device wakes, throttled to 4 h): the struct always answers while the device is awake, and tag 24 tracks discharge consistently with the millivolts in tag 23. Values update reliably across the whole fleet.

The FP310 definition has the same commented-out `percentageAttribute`, but I have no FP310 hardware to confirm its behavior, so it is intentionally left untouched.

## Testing

- Added unit tests: decoding `battery`/`voltage` from a real `0x00F7` payload, and the last-received timestamp recording that keeps the poll dormant.
- `pnpm build`, `pnpm check`, `pnpm test` all pass (653 tests).

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -5898,7 +5898,7 @@ export const definitions: DefinitionWithExtend[] = [
         fromZigbee: [lumi.fromZigbee.lumi_specific],
         toZigbee: [lumi.toZigbee.lumi_presence, lumi.toZigbee.lumi_motion_sensitivity],
         exposes: [e.power_outage_count(), e.motion_sensitivity_select(["low", "medium", "high"]).withDescription("Presence Detection Sensitivity.")],
-        version: "0.0.1",
+        version: "0.0.2",
         configure: async (device, coordinatorEndpoint) => {
             const endpoint = device.getEndpoint(1);
             await endpoint.read<"manuSpecificLumi", ManuSpecificLumi>("manuSpecificLumi", [0x00ee], {manufacturerCode: manufacturerCode}); // Read OTA data; makes the device expose more attributes related to OTA
@@ -5907,6 +5907,7 @@ export const definitions: DefinitionWithExtend[] = [
             await endpoint.read<"manuSpecificLumi", ManuSpecificLumi>("manuSpecificLumi", [0x014f], {manufacturerCode: manufacturerCode}); // Read current PIR interval
             await endpoint.read<"manuSpecificLumi", ManuSpecificLumi>("manuSpecificLumi", [0x0197], {manufacturerCode: manufacturerCode}); // Read current absence delay timer value
             await endpoint.read<"manuSpecificLumi", ManuSpecificLumi>("manuSpecificLumi", [0x019a], {manufacturerCode: manufacturerCode}); // Read detection range
+            await endpoint.read<"manuSpecificLumi", ManuSpecificLumi>("manuSpecificLumi", [0x00f7], {manufacturerCode: manufacturerCode}); // Read battery data; firmware 0.0.0_6542 does not push the 0x00F7 struct on its own
 
             // Configure reporting so presence (0x0142) and PIR detection (0x014d) update autonomously.
             await reporting.bind(endpoint, coordinatorEndpoint, ["manuSpecificLumi"]);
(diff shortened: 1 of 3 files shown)


src/devices/lumi.ts
```diff
@@ -5922,10 +5923,23 @@ export const definitions: DefinitionWithExtend[] = [
...
-                voltageToPercentage: {min: 2850, max: 3000},
-                voltageAttribute: 0x0017, // Attribute: 23
-                //percentageAttribute: 0x0018 // Attribute: 24 // TODO: Should confirm to be sure
+                voltageAttribute: 0x0017, // Attribute: 23 (battery voltage in mV)
+                percentageAttribute: 0x0018, // Attribute: 24 (battery percentage, the device's own gauge; tracks discharge consistently with attribute 23)
+            }),
+            // Firmware 0.0.0_6542 never pushes the 0x00F7 struct that carries battery data, so poll it.
+            // The device is a sleepy end device: the queued read is delivered the next time it wakes up.
+            // https://github.com/Koenkk/zigbee2mqtt/issues/32153
+            m.poll({
```
Koenkk: I think we should only poll when the device has this specific fw (based on the `device.softwareBuildID`)

hajarada: Implemented in c4062796, with one adjustment to the mechanism — instead of comparing `softwareBuildID` strings, the poll is gated on the observed behavior: a small fromZigbee converter records when a 0x00F7 struct was last received, and the (now hourly) poll timer only issues the read when none has arrived for 4 h (`fp300BatteryPoll()` in lib/lumi.ts).

Same outcome as the version check on current firmware, but it holds regardless of what any given build does: firmware that pushes the struct autonomously keeps the poll permanently dormant (devices that do not need polling are never polled), while firmware that does not — 0.0.0_6542, or a future build with the same regression — gets read at ~4 h intervals, with no version list to maintain. It also avoids depending on the `swBuildId` format: I can only verify that string on 0.0.0_6542 (all 11 of my units report exactly `0.0.0_6542`), not on other builds.

If you prefer the explicit softwareBuildID check, I will swap it back.

Koenkk: I think this is more future proof indeed, thanks!

PR #12628, opened 2026-07-05
Added battery information of TRV and firmware info TRV

henkiejan1:
Added information of the battery level of the TRV and the installed firmware version on the TRV and the local available firmware version on the E40 thermostat.


Koenkk:
Thanks!

PR #12629, opened 2026-07-05
fix(tuya): add missing endpoint mapping for ZN2S-RS02E

hesham1216:
endpoint function was lost, causing "Device has no endpoint 'l2'" when commanding per-endpoint topics from HA discovery.

ZN2S-RS02E (TS0601, _TZE284_zpvusbtv) lost its endpoint mapping, so commands published to per-endpoint topics (.../l1/set, .../l2/set) — which HA discovery uses — fail with Device 'X' has no endpoint 'l2'. State updates still work; only commands are broken.
Fix: add m.deviceEndpoints({endpoints: {l1: 1, l2: 1}}) to the extend, same pattern as the neighboring TS0601_3gang_rkbxtclc definition. Verified on hardware — publishing {"state_l2":"ON"} to the base /set topic works, confirming both gangs are on Zigbee endpoint 1 via Tuya DPs.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2721,7 +2721,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "ZN2S-RS02E",
         vendor: "Tuya",
         description: "Two gang switch with colored backlight modes",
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12630, opened 2026-07-05
Add support for Moes ZRS-USC-WH curtain switch

radinsky:
Adds support for the Moes ZRS-USC-WH V2 smart curtain switch.

Device details:
- Zigbee model: TS0601
- Manufacturer: _TZE284_b7kbnl6q
- Box model: ZRS-USC-WH V2

This moves the manufacturer fingerprint from the generic TS0601_cover_1 entry to the existing TS0601_cover_5 wall switch definition. The tested device uses the same Tuya datapoint layout as TS0601_cover_5: state, position, calibration, backlight mode, and motor steering.


Koenkk:
Thanks!

PR #12644, opened 2026-07-08
fix: smoke level to HS1SA-E Plus 

Leo2442926161:
Dear Z2M team,
1. fix: smoke level to HS1SA-E Plus 
2. fix issues on M1-PE


Koenkk:
thanks!

PR #12645, opened 2026-07-08
fix(detect): Detect _TZE204_qtnjuoae  as RTX ZVG1

molusk:
Fix: _TZE204_qtnjuoae water valve not correctly recognized


Koenkk:
thanks!

PR #12646, opened 2026-07-08
fix: Tuya: Fix negative numbers in datapoints

andrei-lazarov:

1. fromZigbee datapoint negative numbers

Numbers received with `valueConverter.raw` are always treated as int32 (signed).
I added the option to treat them as uint32 (unsigned). It's never activated, so it doesn't change anything.
We'll need to use it in the future, if a Tuya device ever uses uint32 internally and reaches 2147483648 (2^31).

2. toZigbee datapoint negative numbers

Sending a positive number is all good (ofc. in the range expected by the device).
Sending a negative number with `valueConverter.raw` is broken.

I fixed it, but I believe some device use workarounds like `valueConverter.localTempCalibration1`.
Need to check that first 😮‍💨

- https://github.com/Koenkk/zigbee2mqtt/issues/32372
- https://github.com/Koenkk/zigbee2mqtt/issues/31053


andrei-lazarov:
Ok, I deleted the localTempCalibrations.

This first line in `from` was not currently used, because the numbers were already converted to negative.
`if (v > 0x7fffffff) v -= 0x100000000;`

This line in `to` is now enabled by default in `raw`.
`if (v < 0) return v + 0x100000000;`

I kept the scaling with `divideByX`

I tested a bunch. Compared `localTempCalX` with the new `raw`. It works the same.
@iotRobot and @bilgi-source, I see you previously worked with these. Maybe you can test or review

Koenkk:
thanks!

PR #12510, opened 2026-06-22
Expose SONOFF valve settings as scalar controls

MaxRink:
## Summary
- Rework SONOFF SWV-ZFE/ZFU/SWV-ZF2 irrigation settings from broad public composites into scalar controls that Home Assistant can render as normal number/select/text/switch entities.
- Keep legacy composite MQTT payloads accepted by the converter and keep publishing legacy composite state for compatibility.
- Allow zero for inactive plan amount/interval fields so duration-mode plans do not appear as `unknown` in Home Assistant.
- Expose irrigation plan removal as scalar config `irrigation_plan_remove_plan_index` instead of a one-field composite, while still accepting the old `irrigation_plan_remove` composite command.
- Mark read-only valve fault, irrigation schedule status, and irrigation record-history exposes as diagnostics so Home Assistant does not show them as primary controls.
- Add generic expose-level Home Assistant metadata and mark the generated SONOFF valve switch exposes as `homeassistant: {type: "valve"}` for the paired Zigbee2MQTT discovery change.

## Why
Home Assistant does not provide a useful native editor for the previous large nested composites. This follows the PR #32362/#32363 discussion: use composites only for fields that really have to be written atomically, and model independent plan/default/alarm values as independent exposes.

Read-only schedule/status/history values are useful for debugging and dashboards, but they are not actions or settings. Categorizing them as diagnostics keeps the main device controls focused on the valve, manual watering defaults, alarm configuration, and plan configuration.

The Home Assistant valve discovery hint belongs to the expose that represents the actor, not to device-level metadata. That keeps the mechanism generic and lets Zigbee2MQTT consume expose metadata without knowing about SONOFF-specific model details.

## Live validation on my Home Assistant instance
Validated with a real SONOFF Zigbee smart water valve:
- `zigbee2mqtt/bridge/devices` no longer exposes `manual_default_settings`, `seasonal_watering_adjustment`, `valve_alarm_settings`, `irrigation_plan_settings`, `irrigation_plan_report`, or `irrigation_plan_remove` as public setting composites.
- Home Assistant creates scalar entities such as `manual_irrigation_duration`, `manual_irrigation_mode`, `manual_irrigation_amount`, `irrigation_plan_enabled`, `irrigation_plan_start_time`, `seasonal_watering_adjustment_january`, `irrigation_plan_remove_plan_index`, and alarm controls.
- Previously-unknown inactive fields report valid zero states: `irrigation_plan_interval_days = 0`, `irrigation_plan_amount = 0`.
[...]

PR #12495, opened 2026-06-21
Allow partial SONOFF water valve composite settings

MaxRink:
## Summary
- merge partial writes for SONOFF SWV-ZFE seasonal watering adjustment settings
- merge partial writes for valve alarm settings
- merge irrigation plan updates with the current settings or latest plan report before encoding the payload
- expand tests to cover manual defaults, seasonal adjustment, valve alarm, and irrigation plan partial updates

## Testing
- `npx pnpm@10.12.1 run check`
- `npx pnpm@10.12.1 vitest run test/sonoff.test.ts --config ./test/vitest.config.mts`
- `npx pnpm@10.12.1 vitest run test/shelly.test.ts test/sonoff.test.ts --config ./test/vitest.config.mts`

PR #12647, opened 2026-07-09
Revert #12495 and #12510: restore composite exposes for SONOFF irrigation settings

CubeZ2mDeveloper:
## Summary

This reverts #12495 and #12510 and restores the previous composite exposes for SONOFF irrigation settings.

## Reason

These settings are represented at the Zigbee protocol level as a single attribute or command carrying the complete configuration payload.

For example, an irrigation plan is written through a single Zigbee command and must include the complete plan configuration in one request, such as the plan index, enabled state, loop type, active dates, start time, irrigation mode, amount, fail-safe settings, and creation time.

This is why these settings were originally exposed as composites: the user edits the complete configuration and applies it as a single payload, matching the device protocol.

#12495 introduced support for partial updates by merging incoming values with previously cached state or reported values before rebuilding the complete payload.

#12510 then split the composite settings into separate exposes for each individual field. Changing any field now triggers a write immediately, while the device protocol requires the complete configuration to be sent in a single request.

This breaks when creating a new irrigation plan or configuring an unused plan slot, where no existing complete configuration is available. Writing an individual field results in an incomplete payload that cannot be sent to the device.

More importantly, these fields are not independently writable Zigbee attributes or commands. Reconstructing the complete configuration from partial input and cached state therefore does not match the actual device protocol.

A similar issue applies to `manual_default_settings`, which is also transferred as a complete configuration rather than a set of independently writable fields.

## Why revert both changes

These settings need to be sent to the device as one complete configuration.

#12495 added support for updating only part of the configuration by filling in the remaining fields from cached state, and #12510 then exposed those fields separately so that each field could trigger its own write.

However, the device does not support updating these fields individually. Each write still has to contain the complete configuration, which means the current behavior depends on previously cached values being available and correct.

This is especially problematic for new configurations, such as a new irrigation plan or an unused plan slot, where no complete previous state exists.

Reverting both changes restores the original composite flow: the user provides the complete configuration first, and Zigbee2MQTT then sends it to the device in a single operation, as required by the device protocol.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -3093,7 +3093,7 @@ const sonoffExtend = {
         };
     },
     rainDelayEndDatetime: (): ModernExtend => {
-        const exposes = [e.text("rain_delay_end_datetime", ea.STATE).withDescription("User triggered delay end time.").withCategory("diagnostic")];
+        const exposes = [e.text("rain_delay_end_datetime", ea.STATE).withDescription("User triggered delay end time.")];
 
         const toZigbee: Tz.Converter[] = [];
 
@@ -3149,9 +3149,7 @@ const sonoffExtend = {
             }
         }
 
-        const exposes = [
-            e.enum("valve_abnormal_state", ea.STATE, allCombinations).withDescription("Valve abnormal state").withCategory("diagnostic"),
-        ];
+        const exposes = [e.enum("valve_abnormal_state", ea.STATE, allCombinations).withDescription("Valve abnormal state")];
 
         const fromZigbee: Fz.Converter<"customClusterEwelink", SonoffSwvzn, ["attributeReport", "readResponse"]>[] = [
             {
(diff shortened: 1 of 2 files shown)


CubeZ2mDeveloper:
@Koenkk  One small request, if I may: when SONOFF-related device definitions are modified by contributors outside our team, would it be possible to give us a quick mention so we can help review the changes when appropriate?

Some SONOFF devices rely on vendor-specific protocol behavior that may not be obvious from the expose definitions or converter code alone, so we would be happy to help review and verify such changes. Hopefully, this could help catch compatibility issues like this one earlier.

Of course, we understand this may not always be practical, so please consider this only as a suggestion. Thanks!

Koenkk:
> @Koenkk One small request, if I may: when SONOFF-related device definitions are modified by contributors outside our team, would it be possible to give us a quick mention so we can help review the changes when appropriate?

Done!

Koenkk:
Thanks! (CC: @MaxRink)

MaxRink:
I only tested with existing plans :S
I guess ill come up with a different solution and rework my irrigation card 😓

MaxRink:
Follow-up on the composite-usability point raised here: I've opened Koenkk/zigbee2mqtt#32545, which makes **settable composites** (like these SONOFF irrigation settings) editable from Home Assistant while keeping the atomic single-write protocol intact.

It discovers a settable `composite` as a stock `sensor` whose attributes carry the composite **schema**, **property** and **set-topic**; a companion HACS custom card ([MaxRink/z2m-composite-card](https://github.com/MaxRink/z2m-composite-card)) renders an editor for all fields and publishes the whole object back in one `set`. No converter changes are needed and it doesn't revive the generic composite→scalar split discussed in Koenkk/zigbee2mqtt#32362 / #32363 — the composite stays atomic. CC @CubeZ2mDeveloper

PR #12651, opened 2026-07-10
fix: Tuya RB-SRAIN01: invert cleaning_reminder

andrei-lazarov:
- https://github.com/Koenkk/zigbee2mqtt/issues/25994


Koenkk:
thanks!

PR #12657, opened 2026-07-10
TS110E on_off fix add logic to use the correct state for  multi switches

### LINKED ISSUE #32246: issue with TS110E -   _TZ3210_3mpwqzuu dimmer switches
### What happened?

i first noticed this playing around with some HA blueprints around motion sensor lights.

i then started noticing lights would stop turning on, this then led me to find out in the broker side of things or rather z2m it was in fact marked as "ON" but the lights themselves do not turn on

i eventually discovered that when i disable the brightness feature of the blueprint things began working as expected

after chatting with people over in discord it was suggested to try some raw mqtt commands, and infact if i send a payload for example

```
{
  "brightness_l1": 41,
  "brightness_l2": 255,
  "last_seen": "2026-06-09T22:10:49+08:00",
  "linkquality": 84,
  "max_brightness": 255,
  "max_brightness_l1": 255,
  "max_brightness_l2": 255,
  "min_brightness": 1,
  "min_brightness_l1": 1,
  "min_brightness_l2": 1,
  "power_on_behavior": "previous",
  "power_on_behavior_l1": "previous",
  "power_on_behavior_l2": "previous",
  "state_l1": "ON",
  "state_l2": "OFF",
  "switch_type_l1": "toggle",
  "switch_type_l2": "toggle"
}
```

the actual light would not turn on if the current brightness was already set to 41

if i then dropped ` "brightness_l1": 41,`  from the payload it would indeed turn on the light

so i guess the question is, is this something that can be solved?

if i send a brightness command thats exactly the same it should turn on the light

 TS110E -   _TZ3210_3mpwqzuu

have this device send 

```
{
  "brightness_l1": 41,
  "brightness_l2": 255,
  "last_seen": "2026-06-09T22:10:49+08:00",
  "linkquality": 84,
  "max_brightness": 255,
  "max_brightness_l1": 255,
  "max_brightness_l2": 255,
  "min_brightness": 1,
  "min_brightness_l1": 1,
  "min_brightness_l2": 1,
  "power_on_behavior": "previous",
  "power_on_behavior_l1": "previous",
  "power_on_behavior_l2": "previous",
  "state_l1": "ON",
  "state_l2": "OFF",
  "switch_type_l1": "toggle",
  "switch_type_l2": "toggle"
}
```
then turn it off with

```
{
  "brightness_l1": 41,
  "brightness_l2": 255,
  "last_seen": "2026-06-09T22:10:49+08:00",
  "linkquality": 84,
  "max_brightness": 255,
  "max_brightness_l1": 255,
  "max_brightness_l2": 255,
  "min_brightness": 1,
  "min_brightness_l1": 1,
  "min_brightness_l2": 1,
  "power_on_behavior": "previous",
  "power_on_behavior_l1": "previous",
  "power_on_behavior_l2": "previous",
  "state_l1": "OFF",
  "state_l2": "OFF",
  "switch_type_l1": "toggle",
  "switch_type_l2": "toggle"
}
```

then try the first payload again, the actual light will not turn on but the switch will indicate that it is on

then the following payload will turn on the light due to changing the brightness

```
{
  "brightness_l1": 42,
  "brightness_l2": 255,
  "last_seen": "2026-06-09T22:10:49+08:00",
  "linkquality": 84,
  "max_brightness": 255,
  "max_brightness_l1": 255,
  "max_brightness_l2": 255,
  "min_brightness": 1,
  "min_brightness_l1": 1,
  "min_brightness_l2": 1,
  "power_on_behavior": "previous",
  "power_on_behavior_l1": "previous",
  "power_on_behavior_l2": "previous",
  "state_l1": "ON",
  "state_l2": "OFF",
  "switch_type_l1": "toggle",
  "switch_type_l2": "toggle"
}
```

### Zigbee2MQTT version

2.10.1-dev commit: 5458e5084def411aaad9af08e415acb6bc16d181 

### Adapter firmware version

20210708

### Adapter

_Sonoff_Zigbee_3.0

### Setup

docker

_No response_

### Debug log

_No response_

### Notes

_No response_

phyzical:
resolves https://github.com/Koenkk/zigbee2mqtt/issues/32246

Basically when you send a same brightness command `moveToLevelTuya` will only turn on the device if the brightness changes, this is due to the helper assuming the state is always state but with multi switches we need to handle `state_l1` for example

this workaround just does a lookup, if we have a `state_` blah then assume we need it to work like that

I'm unsure if other parts of the device suffer a similar fate assuming that it should be i.e `brightness` vs `brightness_l1` but it may?

i assume the reason device would still turn on when the brightness changes is a side effect of the device itself and how it handles `moveToLevelTuya`

### DIFF
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -3311,10 +3311,11 @@ const tuyaTz = {
         key: ["state", "brightness"],
         convertSet: async (entity, key, value, meta) => {
             const {message, state} = meta;
-            const brightnessKey = Object.keys(state).find((k) => k.startsWith("brightness_l")) && "ID" in entity ? `brightness_l${entity.ID}` : "brightness";
+            const brightnessKey =
+                Object.keys(state).find((k) => k.startsWith("brightness_l")) && "ID" in entity ? `brightness_l${entity.ID}` : "brightness";
             const stateKey = Object.keys(state).find((k) => k.startsWith("state_l")) && "ID" in entity ? `state_l${entity.ID}` : "state";
-            if (message.state === "OFF"  || (message.state != null && message.brightness == null)) {
-                return  await tz.on_off.convertSet(entity, key, value, meta);
+            if (message.state === "OFF" || (message.state != null && message.brightness == null)) {
+                return await tz.on_off.convertSet(entity, key, value, meta);
             }
             if (message.brightness != null) {
                 // If state includes brightness assume we need to use a custom lookup
@@ -3333,9 +3334,9 @@ const tuyaTz = {
                         "genLevelCtrl",
                         "moveToLevelTuya",
                         {level, transtime: 100},
-                        utils.getOptions(meta.mapped, entity)
+                        utils.getOptions(meta.mapped, entity),
                     );
-               }  
+                }
 
                 return {state: {state: "ON", brightness}};
             }
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -3311,8 +3311,8 @@ const tuyaTz = {
         key: ["state", "brightness"],
         convertSet: async (entity, key, value, meta) => {
             const {message, state} = meta;
-            const brightnessKey = Object.keys(state).find((k) => k.startsWith("brightness_l")) ? `brightness_l${entity.ID}` : "brightness";
-            const stateKey = Object.keys(state).find((k) => k.startsWith("state_l")) ? `state_l${entity.ID}` : "state";
+            const brightnessKey = Object.keys(state).find((k) => k.startsWith("brightness_l")) && "ID" in entity ? `brightness_l${entity.ID}` : "brightness";
+            const stateKey = Object.keys(state).find((k) => k.startsWith("state_l")) && "ID" in entity ? `state_l${entity.ID}` : "state";
             if (message.state === "OFF"  || (message.state != null && message.brightness == null)) {
                 return  await tz.on_off.convertSet(entity, key, value, meta);
             }
@@ -3328,7 +3328,7 @@ const tuyaTz = {
                     const level = utils.mapNumberRange(brightness, 0, 254, 0, 1000);
 
                     // set brightness
-                    await entity.command(
+                    await entity.command<"genLevelCtrl", "moveToLevelTuya", TuyaGenLevelCtrl>(
                         "genLevelCtrl",
                         "moveToLevelTuya",
                         {level, transtime: 100},
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -3311,24 +3311,32 @@ const tuyaTz = {
         key: ["state", "brightness"],
         convertSet: async (entity, key, value, meta) => {
             const {message, state} = meta;
-            if (message.state === "OFF" || (message.state != null && message.brightness == null)) {
-                return await tz.on_off.convertSet(entity, key, value, meta);
+            const brightnessKey = Object.keys(state).find((k) => k.startsWith("brightness_l")) ? `brightness_l${entity.ID}` : "brightness";
+            const stateKey = Object.keys(state).find((k) => k.startsWith("state_l")) ? `state_l${entity.ID}` : "state";
+            if (message.state === "OFF"  || (message.state != null && message.brightness == null)) {
+                return  await tz.on_off.convertSet(entity, key, value, meta);
             }
             if (message.brightness != null) {
-                // set brightness
-                if (state.state === "OFF") {
-                    await entity.command("genOnOff", "on", {}, utils.getOptions(meta.mapped, entity));
-                }
-
+                // If state includes brightness assume we need to use a custom lookup    
                 const brightness = utils.toNumber(message.brightness, "brightness");
-                const level = utils.mapNumberRange(brightness, 0, 254, 0, 1000);
-                await entity.command<"genLevelCtrl", "moveToLevelTuya", TuyaGenLevelCtrl>(
-                    "genLevelCtrl",
-                    "moveToLevelTuya",
-                    {level, transtime: 100},
-                    utils.getOptions(meta.mapped, entity),
-                );
-                return {state: {state: "ON", brightness}};
+                 // we allow at most 1 incase its a rounding/ float precision issue
+                const brightnessUnchanged = Math.abs(utils.mapNumberRange(brightness, 0, 254, 0, 254) - utils.toNumber(state[brightnessKey], "brightness")) <= 1;
+                // if the brightness is unchanged then we need to force it on due to weirdness with moveToLevelTuya
+                if (state[stateKey] === "OFF" && brightnessUnchanged ) {
+                    await entity.command("genOnOff", "on", {}, utils.getOptions(meta.mapped, entity));
+                } else {
+                    const level = utils.mapNumberRange(brightness, 0, 254, 0, 1000);
+
+                    // set brightness
+                    await entity.command(
+                        "genLevelCtrl",
+                        "moveToLevelTuya",
+                        {level, transtime: 100},
+                        utils.getOptions(meta.mapped, entity)
+                    );
+               }  
+
+              return {state: {state:"ON", brightness}};
             }
         },
         convertGet: async (entity, key, meta) => {
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -3313,30 +3313,26 @@ const tuyaTz = {
             const {message, state} = meta;
             const brightnessKey = Object.keys(state).find((k) => k.startsWith("brightness_l")) ? `brightness_l${entity.ID}` : "brightness";
             const stateKey = Object.keys(state).find((k) => k.startsWith("state_l")) ? `state_l${entity.ID}` : "state";
-            if (message.state === "OFF"  || (message.state != null && message.brightness == null)) {
-                return  await tz.on_off.convertSet(entity, key, value, meta);
+            if (message.state === "OFF" || (message.state != null && message.brightness == null)) {
+                return await tz.on_off.convertSet(entity, key, value, meta);
             }
             if (message.brightness != null) {
-                // If state includes brightness assume we need to use a custom lookup    
+                // If state includes brightness assume we need to use a custom lookup
                 const brightness = utils.toNumber(message.brightness, "brightness");
-                 // we allow at most 1 incase its a rounding/ float precision issue
-                const brightnessUnchanged = Math.abs(utils.mapNumberRange(brightness, 0, 254, 0, 254) - utils.toNumber(state[brightnessKey], "brightness")) <= 1;
+                // we allow at most 1 incase its a rounding/ float precision issue
+                const brightnessUnchanged =
+                    Math.abs(utils.mapNumberRange(brightness, 0, 254, 0, 254) - utils.toNumber(state[brightnessKey], "brightness")) <= 1;
                 // if the brightness is unchanged then we need to force it on due to weirdness with moveToLevelTuya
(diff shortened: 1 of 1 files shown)


phyzical:
ive also decided to add logic to only perform this when the brightness is unchanged given the `moveToLevelTuya` seems to handle things as expected

phyzical:
Hmm not sure why the CI doesnt like it i am able to confirm this works via an external converter

```
import * as modernExtend from "zigbee-herdsman-converters/lib/modernExtend";
import * as tuya from "zigbee-herdsman-converters/lib/tuya";
import * as utils from 'zigbee-herdsman-converters/lib/utils'
const e = exposes.presets;
const {tuyaBase} = tuya.modernExtend;

const TS110E_onoff_brightness_2 = {
         key: ["state", "brightness"],
            const {message, state} = meta;

            if (message.state === "OFF" || (message.state != null && message.brightness == null)) {
                return await tz.on_off.convertSet(entity, key, value, meta);
            }
            if (message.brightness != null) {
                // If state includes state_l1 assume we need to use a custom lookup
                const stateKey = Object.keys(state).find((k) => k.startsWith("state_l1")) ? "state_l" + entity.ID : "state";
                const brightnessKey = Object.keys(state).find((k) => k.startsWith("brightness_l1")) ? "brightness_l" + entity.ID : "brightness";


                const brightness = utils.toNumber(message.brightness, "brightness");
                 // we allow at most 1 incase its a rounding/ float precision issue
                const brightnessUnchanged = Math.abs(utils.mapNumberRange(brightness, 0, 254,0, 254) - state[brightnessKey]) <= 1;
console.log("state is",message, state,stateKey , brightnessKey )
                // if the brightness is unchanged then we need to force it on due to weirdness with moveToLevelTuya

                if (state[stateKey] === "OFF" && brightnessUnchanged) {
                    await entity.command("genOnOff", "on", {}, utils.getOptions(meta.mapped, entity));
                }

                const level = utils.mapNumberRange(brightness, 0, 254, 0, 1000);

                // set brightness
                await entity.command(
                    "genLevelCtrl",
                    "moveToLevelTuya",
                    {level, transtime: 100},
                    utils.getOptions(meta.mapped, entity)
                );
                return {state: {[stateKey]: "ON", brightness}};
          }
        },
            if (key === "state") await tz.on_off.convertGet(entity, key, meta);
            if (key === "brightness") await entity.read("genLevelCtrl", [61440]);
        },
    };


export default [  {
      fingerprint: tuya.fingerprint("TS110E", [ "_TZ3210_3mpwqzuu"]),
        model: "TS110E_2gang_2",
        vendor: "Tuya",
        description: "2 channel dimmer",
        whiteLabel: [tuya.whitelabel("Nedis", "ZBWD20RD", "SmartLife Triac Dimmer", ["_TZ3210_mt5xjoy6"])],
        extend: [tuyaBase()],
        fromZigbee: [tuya.fz.TS110E, tuya.fz.TS110E_switch_type, tuya.fz.power_on_behavior_1, fz.on_off],
        toZigbee: [TS110E_onoff_brightness_2, tuya.tz.TS110E_options, tuya.tz.power_on_behavior_1, tz.light_brightness_move],
        meta: {multiEndpoint: true},
        exposes: [
            e.power_on_behavior(),
            e.light_brightness().withMinBrightness().withMaxBrightness().withEndpoint("l1"),
            e.light_brightness().withMinBrightness().withMaxBrightness().withEndpoint("l2"),
            tuya.exposes.switchType().withEndpoint("l1"),
            tuya.exposes.switchType().withEndpoint("l2"),
        ],
            await reporting.bind(endpoint, coordinatorEndpoint, ["genOnOff", "genLevelCtrl"]);
        },
        endpoint: (device) => {
            return {l1: 1, l2: 2};
        },
    }];



```

phyzical:
Added some more logic to handle the clamping to reference the min max instead of a static 254, as it resulted in `brightnessUnchanged` to always be false

phyzical:
had an issue where i was still getting precision errors i.e 31 setting vs 30 in state so instead keep the old logic of hardcoded 254 and allow a variance of 1 as a `unchanged` bool to force the lights on

src/lib/tuya.ts
```diff
@@ -3300,20 +3315,28 @@ const tuyaTz = {
...
-                if (state.state === "OFF") {
+                // If state includes state_l1 assume we need to use a custom lookup
+                const stateKey = Object.keys(state).find((k) => k.startsWith("state_l1")) ? `state_l${entity.ID}` : "state";
+                const brightnessKey = Object.keys(state).find((k) => k.startsWith("brightness_l1")) ? `brightness_l${entity.ID}` : "brightness";
+                const brightness = utils.toNumber(message.brightness, "brightness");
+                // we allow at most 1 incase its a rounding/ float precision issue
+                const brightnessUnchanged = Math.abs(utils.mapNumberRange(brightness, 0, 254, 0, 254) - state[brightnessKey]) <= 1;
+
+                // if the brightness is unchanged then we need to force it on due to weirdness with moveToLevelTuya
+                if (state[stateKey] === "OFF" && brightnessUnchanged) {
```
Koenkk: Should we check if the target state is `ON` here? Because what if `{state: OFF}` is recevied?

phyzical: Do you mean in the message payload?

phyzical: I did have this thought, but I doubt anyone will send a brightness command with an off right?

Koenkk: Yes indeed, it might depends on e.g. HA automations used by people so would say it's a valid use-case.

phyzical: oh actually i just realised that will be handled by the early return above https://github.com/Koenkk/zigbee-herdsman-converters/pull/12657/changes#diff-d45f20d30a43dc3e172b16bd55872492f4867c8831c5e03a8be9be65f1cd72d5R3314

~did we instead want to handle brightness first and then perform an `on/off`~

~and avoid all the early returns?~

nah i think keep it as it is in this regard what do you think?

phyzical:
Okay so i missunderstood some of the mechanisms going on, turns out `moveToLevelTuya` will not actually send the `ON` to the state (i thought the return was handling that oops)

so we need to simply always send the on always if its off  (maybe this is what you were implying yesterday?)

So tldr the issue was simply that we did not check the correct state before telling the light to turn on which is what caused the problem in state. and the bug around same brightness not turning on the light should naturally be resolved by the fact we turn it on if its off

sadly this does result in twice as many events firing out to the network but at this time i couldnt get an all in one similar to the `moveToLevelWithOnOff` as this didnt seem to work

src/lib/tuya.ts
```diff
@@ -3300,20 +3315,25 @@ const tuyaTz = {
...
                     {level, transtime: 100},
                     utils.getOptions(meta.mapped, entity),
                 );
-                return {state: {state: "ON", brightness}};
+
+                if (state[stateKey] === "OFF") {
+                    await entity.command("genOnOff", "on", {}, utils.getOptions(meta.mapped, entity));
+                }
+
+                return {state: {[stateKey]: "ON", brightness}};
```
phyzical: Turns out there is must be magic that translates this correctly as is tarted getting state_l1_l1 getting stored

```suggestion
                return {state: {state: "ON", brightness}};
```

src/lib/tuya.ts
```diff
@@ -3300,20 +3315,25 @@ const tuyaTz = {
                 return await tz.on_off.convertSet(entity, key, value, meta);
             }
             if (message.brightness != null) {
-                // set brightness
-                if (state.state === "OFF") {
-                    await entity.command("genOnOff", "on", {}, utils.getOptions(meta.mapped, entity));
-                }
-
+                // If state includes state_l1 assume we need to use a custom lookup
```
phyzical:
```suggestion
                // turn on
                if (state[stateKey] === "OFF") {
                    await entity.command("genOnOff", "on", {}, utils.getOptions(meta.mapped, entity));
                }
                
                // If state includes state_l1 assume we need to use a custom lookup
```

src/lib/tuya.ts
```diff
@@ -3300,20 +3315,25 @@ const tuyaTz = {
...
                     "moveToLevelTuya",
                     {level, transtime: 100},
                     utils.getOptions(meta.mapped, entity),
                 );
-                return {state: {state: "ON", brightness}};
+
+                if (state[stateKey] === "OFF") {
+                    await entity.command("genOnOff", "on", {}, utils.getOptions(meta.mapped, entity));
+                }
+
```
phyzical:
```suggestion
```

src/lib/tuya.ts
```diff
@@ -3300,19 +3315,25 @@ const tuyaTz = {
                 return await tz.on_off.convertSet(entity, key, value, meta);
             }
             if (message.brightness != null) {
-                // set brightness
-                if (state.state === "OFF") {
+                // turn on
```
phyzical, line 3320:
```suggestion
               // If state includes state_l1 assume we need to use a custom lookup 
                const stateKey = Object.keys(state).find((k) => k.startsWith("state_l1")) ? `state_l${entity.ID}` : "state";
                // turn on
```

src/lib/tuya.ts
```diff
@@ -3300,19 +3315,25 @@ const tuyaTz = {
...
+                if (state[stateKey] === "OFF") {
                     await entity.command("genOnOff", "on", {}, utils.getOptions(meta.mapped, entity));
                 }
-
+                
+                // If state includes state_l1 assume we need to use a custom lookup
+                const stateKey = Object.keys(state).find((k) => k.startsWith("state_l1")) ? `state_l${entity.ID}` : "state";
```
phyzical:
```suggestion
```

phyzical:
i will reping you when i feel confident, something wasnt right i think this is it but will retest

phyzical:
okay ive been running this version for about a week and seems to fix the orginal issue and doesnt cause the other issue i had anymore, still unsure what was the cause of that may have been something unrelated in my setup.

this is the js version i tested
```
const TS110E_onoff_brightness = {
         key: ["state", "brightness"],
            const {message, state} = meta;
            const brightnessKey = Object.keys(state).find((k) => k.startsWith("brightness_l1")) ? "brightness_l" + entity.ID : "brightness";
            const stateKey = Object.keys(state).find((k) => k.startsWith("state_l1")) ? "state_l" + entity.ID : "state";
            if (message.state === "OFF"  || (message.state != null && message.brightness == null)) {
                return  await tz.on_off.convertSet(entity, key, value, meta);
            }
            if (message.brightness != null) {
                // If state includes brightness assume we need to use a custom lookup    
                const brightness = utils.toNumber(message.brightness, "brightness");
                 // we allow at most 1 incase its a rounding/ float precision issue
                const brightnessUnchanged = Math.abs(utils.mapNumberRange(brightness, 0, 254,0, 254) - state[brightnessKey]) <= 1;
                // if the brightness is unchanged then we need to force it on due to weirdness with moveToLevelTuya
                if (state[stateKey] === "OFF" && brightnessUnchanged ) {
                    await entity.command("genOnOff", "on", {}, utils.getOptions(meta.mapped, entity));
                } else {
                       const level = utils.mapNumberRange(brightness, 0, 254, 0, 1000);

                         // set brightness
                         await entity.command(
                               "genLevelCtrl",
                                "moveToLevelTuya",
                                {level, transtime: 100},
                                utils.getOptions(meta.mapped, entity)
                          );
               }  

              return {state: {state:"ON", brightness}};
          }
        },
            if (key === "state") await tz.on_off.convertGet(entity, key, meta);
            if (key === "brightness") await entity.read("genLevelCtrl", [61440]);
        },
    };
```

phyzical:
@Koenkk When you get a chance mind giving this another review? Thanks!

Koenkk:
Thanks!

PR #12659, opened 2026-07-10
Shelly WS90: Exposing solar capacitor voltage

sschwieb:
This is a small tweak for the Shelly Ecowitt WS90 Weather Station. Purpose is to expose the 'battery2 voltage', which I believe is the voltage of the super capacitor that's powered by the built-in solar panel. 

See docs here: https://shelly-api-docs.shelly.cloud/docs-ble/Devices/BLU_ZB/wstation/#zigbee

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -1961,6 +1961,17 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Weather station",
         extend: [
             m.battery({voltage: true}),
+            m.numeric({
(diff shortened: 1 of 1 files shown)


Koenkk:
Please ping me when this is ready for merge.

sschwieb:
Thanks @Koenkk - I tested it for a few days, and it seems to work well, so should be good to go.

Koenkk:
Thanks!

PR #12661, opened 2026-07-11
Fix AOYAN model matching

zyjsmile857:
Fixes AOYAN devices that report their own model IDs by adding the reported model IDs to the base definitions.\n\nChanges:\n- Add AY-101Z to the TS0203 zigbeeModel list\n- Add AY201Z to the ZG-227ZL zigbeeModel list and add a modelID fingerprint for its whiteLabel entry\n- Add AY222Z to the TS0207 water leak detector zigbeeModel list\n- Add AY301Z to the WHD02/ZG-301Z zigbeeModel list\n\nValidation:\n- biome check src/devices/tuya.ts

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -4371,7 +4371,7 @@ export const definitions: DefinitionWithExtend[] = [
         exposes: [],
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12662, opened 2026-07-11
feat: support more features on Aqara Spotlight T2 Pro

blakegong:
Every single feature added here have been manually tested to ensure they work with this particular device. The model numbers are from the paper manual inside (I can't find any PDF online, let me know if you need proof, I can attach a photo here).

Notable things that do **NOT** work (thus not added in this PR) in `lumiLight`:
- `on_off_transition_time` (`on_transition_time` and `off_transition_time` do work)
- `execute_if_off`
- `powerOutageMemory`


Koenkk:
Thanks!

PR #12665, opened 2026-07-11
Update Philips light extension with color options

mrtncode:
In the last release I added the Philips Turaco Light. 
I forgot to add color options, so currently it is not possible to control the color temp and color :(

I copied the extend: part from a similiar philips bulb with the same functionality - I hope that works..


Koenkk:
Thanks!

PR #12667, opened 2026-07-11
feat(exposes): add withDefault for composite feature defaults

MaxRink:
## Summary

Adds an optional per-feature `default` value to exposes via a new `withDefault(value)` builder on `Base`, and seeds defaults on the SONOFF cyclic timed/quantitative irrigation composites.

## Why

A `composite` must be written to the device as **one complete object** in a single message; its fields are not independently writable (this is the whole point of `composite` — see #12495 and #12510 which were reverted by #12647 because the SONOFF valve needs the full config in one write and new/empty plan slots have no cached state to merge).

A companion Home Assistant composite-editor card (and potentially other consumers) needs sensible per-field **defaults** so it can pre-fill a new/empty composite slot and build a complete payload. This adds that capability without changing any converter logic — only optional metadata.

## Changes
- `src/lib/exposes.ts`: optional `Base.default` + `withDefault()`, preserved through clone/copy (only when set, so serialized exposes are unchanged when unused).
- `src/devices/sonoff.ts`: `.withDefault(...)` on the cyclic timed/quantitative irrigation composite features (values within declared ranges).
- `test/utils.test.ts`: tests for setting/cloning and non-serialization when unset.

## Companion work
Consumed by a Zigbee2MQTT change that discovers settable composites for Home Assistant and a custom Lovelace card that edits them (linked separately).

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -2355,14 +2355,17 @@ const sonoffExtend = {
         const exposes = e
             .composite("cyclic_timed_irrigation", "cyclic_timed_irrigation", ea.ALL)
             .withDescription("Smart water valve cycle timing irrigation")
-            .withFeature(e.numeric("current_count", ea.STATE).withDescription("Number of times it has been executed").withUnit("times"))
+            .withFeature(
+                e.numeric("current_count", ea.STATE).withDescription("Number of times it has been executed").withUnit("times").withDefault(0),
+            )
             .withFeature(
                 e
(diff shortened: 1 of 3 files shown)


MaxRink:
Closing: the composite editor no longer needs converter-provided defaults. The Home Assistant card derives sensible field defaults directly from the expose schema (numeric → value_min, enum → first value, binary → value_off, text → ""), so this `withDefault` change is unnecessary. Keeping the composite atomic per the #12647 discussion.

PR #12674, opened 2026-07-12
Configure internalTemperature reporting interval for Inovelli devices

adamawolf:
Inovelli VZM31-SN/VZM30-SN switches report `internalTemperature` (attribute 0x0020) on the manufacturer-specific cluster `manuSpecificInovelli` (64561) every 30 seconds using the firmware default reporting interval. The `parameters()` configure method binds this cluster but never sets a reporting interval, so the device defaults to 30s.

This adds an `endpoint.configureReporting()` call in the `parameters()` configure method to set the minimum report interval to 1800s (30 min) and maximum to 3600s (1 hr), with a 5-degree change threshold. This reduces Zigbee traffic from one report per device every 30s to at most once every 30 minutes.

The `.catch()` silently handles devices whose firmware doesn't support configureReporting on this cluster (e.g. VZM32-SN mmWave switches, which report internalTemperature via a different mechanism).

Fixes: internal temperature reporting interval not configured for Inovelli manufacturer-specific cluster

### DIFF
--- a/src/lib/inovelli.ts
+++ b/src/lib/inovelli.ts
@@ -618,10 +618,12 @@ const inovelliExtend = {
         attrs,
         model,
         splitValuesByEndpoint = false,
+        temperatureReporting,
     }: {
         attrs: Array<{attributes: {[s: string]: Attribute}; clusterName: typeof INOVELLI_CLUSTER_NAME | typeof INOVELLI_MMWAVE_CLUSTER_NAME}>;
         model: Model;
         splitValuesByEndpoint?: boolean;
+        temperatureReporting?: {min: number; max: number; change: number};
     }): ModernExtend => {
         const fromZigbee: NonNullable<ModernExtend["fromZigbee"]> = [];
         const toZigbee: Tz.Converter[] = [];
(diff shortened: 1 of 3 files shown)


src/lib/inovelli.ts
```diff
@@ -653,6 +654,12 @@ const inovelliExtend = {
                 const endpoint = device.getEndpoint(1);
                 await reporting.bind(endpoint, coordinatorEndpoint, [INOVELLI_CLUSTER_NAME]);
 
+                await endpoint
```
rohankapoorcom: This block is located in a fairly generic device parameters section that is parameterized. I don't really like the fact that this is hard coding a specific attribute in the middle of that rather than extending the parsmeterization to include setting the reporting as well.

adamawolf: Thanks for the feedback. Is there an existing attribute to which I can refer for example that is parameterized the way you would like to see for this?

adamawolf: @rohankapoorcom took a stab. I think this is a bit closer to what you mean but please let me know.

rohankapoorcom:
Sorry, I lost track of this and haven't come back to review. I'll take a look over the weekend.

InovelliUSA:
I believe the VZM32 is supposed to support configuring this attribute in the same way as the other devices. I will check with the firmware to see if something is different. This doesn't exlude the VZM32 from this change it just catches it if it errors out right?

One thing I wonder is if the threshold gets changed to 5 degrees, does the minimum interval need to be 1800? I believe the 5C would equal about 10F and that much of a change shouldn't happen too often. I would recommend a 5 minute min int with a 5 degree min change. Curious about your thoughts on this.

PR #12675, opened 2026-07-12
fix: Change action_transition_time unit to seconds

andrei-lazarov:
I checked the spec, all level commands have transition time in deciseconds.
Any idea why it was divided by 100?

image

image

image


Koenkk:
Thanks! Should be seconds indeed

PR #12679, opened 2026-07-13
feat: add TS011F_Socket with dynamic multi-gang and metering

hml96:
Add a new TS011F_Socket definition matched by manufacturerName for multi-gang power monitoring sockets. Detects genOnOff endpoints at runtime: first endpoint stays unsuffixed (state), additional endpoints become CH2..CHN. Metering via m.electricityMeter(). Tuya-specific features (countdown per gang, power-outage memory, indicator mode, child lock, switch type) included.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -10401,69 +10401,85 @@ export const definitions: DefinitionWithExtend[] = [
             tuya.whitelabel("Haozee", "HT-SP-ZB-01", "Smart plug (with power monitoring)", ["_TZ3210_w0qqde0g"]),
         ],
         ota: true,
-        extend: [
-            tuya.modernExtend.tuyaBase(),
-            tuya.modernExtend.tuyaOnOff({
-                electricalMeasurements: true,
-                electricalMeasurementsFzConverter: tuya.fz.TS011F_electrical_measurement,
-
-                // Conditional features
-                powerOutageMemory: (manufacturerName) => manufacturerName !== "_TZ3000_cicwjqth",
-                indicatorMode: (manufacturerName) => manufacturerName !== "_TZ3000_ww6drja5",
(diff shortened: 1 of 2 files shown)


Koenkk:
I think the same comments as #12606  apply

hml96:
Yes, I’ve abandoned the plan to match based on model ID plus version number; instead, I’ve implemented matching based on specific manufacturer names, while leaving the existing polling mechanism unchanged.

hml96:
Such changes won't affect the existing equipment, will they?

Koenkk:
It modifies the `TS011F_plug_1` definition, I would prefer that a new definition is created for the multi-gang devices. There is already a `TS011F_2_gang_wall`

hml96:
`TS011F_2_gang_wall` is the definition for a 2-gang device; however, Tuya socket devices with 1 to 5 gangs all use the `TS011F` model ID. Therefore, I believe the identification logic here should be flexible.

Koenkk:
But I assume they do use a different `manufacturerName`? Can you share the `data/database.db` entries for a few of those?

hml96:
While they do use different `manufacturerName` values, the content of these strings does not reveal the number of channels the device has; this is because the `manufacturerName` follows the format `_TZ3000_xxxxxxxx` (where `xxxxxxxx` is the PID), and any individual or manufacturer can create a unique PID on the Tuya platform for production and market release. Consequently, there are likely thousands of sockets on the market with varying `manufacturerName` strings and channel counts—a number that will only grow—making management based on `manufacturerName` extremely difficult; however, they all share the common `model ID` of "TS011F".

Koenkk:
That's true, but devices also have different features, e.g. some support child lock, some support led inidicator, power outage memory etc. Therefore we are mapping every `manufacturerName ` individually.

PR #12682, opened 2026-07-13
Add Terncy WS01 wall switch variants

calvinastroboy:
# Add support for Terncy/Xiaoyan TERNCY-WS01 wall switches

Adds Zigbee2MQTT definitions for:

- `TERNCY-WS01-S1`, `TERNCY-WS01-S2`, `TERNCY-WS01-S3`, `TERNCY-WS01-S4`
- `TERNCY-WS01-D1`, `TERNCY-WS01-D2`, `TERNCY-WS01-D3`, `TERNCY-WS01-D4`

The devices use manufacturer `Xiaoyan`, manufacturer code `0x1228`, and
private cluster `0xfccc`.

## Standard relay support

The gateway node-struct corpus shows the following standard `genOnOff` relay
endpoints:

| Models | Relay exposes |
| --- | --- |
| `WS01-S1`, `WS01-D1` | `state_l1` |
| `WS01-S2`, `WS01-D2` | `state_l1`, `state_l2` |
| `WS01-S3`, `WS01-D3` | `state_l1`, `state_l2`, `state_l3` |
| `WS01-S4`, `WS01-D4` | `state_l1`, `state_l2`, `state_l3`, `state_l4` |

All models have four private endpoints, so action parsing is retained for
`l1` through `l4`, including non-relay input endpoints.

## Private features

The converter carries the existing Terncy wall-switch protocol mapping:

- operation mode through private command `0x1d` (`control_relay=0`,
  `wireless=1`);
- wireless-mode LED state through command `0x1f` (`off=0`, `on=1`);
- relay-mode LED feedback through attr `0x001f`, written from source endpoint
  `110` (`positive=0`, `negative=1`);
- raw action frames for 1-7 clicks, hold, release, and action duration.

The upstream definition intentionally excludes the community-only
`relay_enabled`, `relay_constant_power`, and `button_relay_binding` convenience
controls. Command `0x13` and attr `0x0026` are not used as LED feedback paths.

## Validation

The repository test added with this change covers all eight model IDs, each
relay endpoint boundary, all four endpoint action names, and a click frame from
a non-relay endpoint. The tests pass locally.

The endpoint layout is supported by gateway node-struct evidence. I do not have
a live WS01 packet-validation record in this change, so the private behavior is
reported as family-based inference from the verified WS07-D3 implementation and
should be confirmed on live WS01 hardware.

## Related image PR

The matching device images are prepared in [zigbee2mqtt.io PR #5322](https://github.com/Koenkk/zigbee2mqtt.io/pull/5322).


src/devices/terncy.ts
```diff
@@ -479,14 +593,41 @@ const tzLocal = {
     } satisfies Tz.Converter,
 };
 
-export const definitions: DefinitionWithExtend[] = [
-    {
-        zigbeeModel: ["TERNCY-WS01-S4"],
-        model: "TERNCY-WS01",
+function createWs01Definition(modelInfo: (typeof WS01_MODELS)[number]): DefinitionWithExtend {
```
Koenkk: Please keep the `export const definitions` (to keep it similar to other files)

calvinastroboy: Addressed in `31835bb`: the `export const definitions` structure is preserved. I have also synced the branch with the latest upstream `master` in `c1ecd1a`; all three updated checks pass.

calvinastroboy:
Hi @Koenkk, I’ve synced this PR with the latest upstream `master` in `c1ecd1a` and addressed the review comment in `31835bb` while preserving the `export const definitions` structure. All three updated checks pass, and the branch is conflict-free.

Could you please re-review when convenient? Also, could you confirm whether live WS01 packet validation is required before merge? The current implementation is based on the verified WS07-D3 behavior and WS01 node-structure evidence.

src/devices/terncy.ts
Koenkk: Thanks for the updates, but I prefer not to have a function to create definitions as this creates inconsistency with other definitions, it's better to have a dynamic expose function instead, examples: https://github.com/search?q=repo%3AKoenkk%2Fzigbee-herdsman-converters+%22exposes%3A+%28%22&type=code

Koenkk:
Thanks!

PR #12683, opened 2026-07-13
feat: Expose action parameters: group, level, rate, step size and transition time

andrei-lazarov:

This pull request exposes the following entities for remotes that use standard functions `m.commandsCluster()`:
- `action_group`
- `action_level`
- `action_rate`
- `action_step_size`
- `action_transition_time`

I covered only **OnOff** and **LevelControl** clusters.
More should be aded in the future, for **ColorControl**, **Scenes**, **WindowCovering**: `action_hue`, `action_direction`, `action_saturation`, `action_color` etc.

These parameters were **already published to MQTT,** just not exposed.
Should be no functionality changes, just easier use. It makes them show up in WF, HA and docs.

I triggered a docgen to check the results: https://github.com/andrei-lazarov/zigbee2mqtt.io/commit/6daf67e51845d7760ebe3ce281416a03651d501d

And of course I tested locally


andrei-lazarov:
Also I don't like how they are separate entities. Maybe we can rework in Z2M 3.0.0?

```json
"voltage": 2500,
"action": "off",
"action_group": 21658     
```

```json
"voltage": 2500,
"action": {
    "command": "off",
    "alias": "single_press",
    "group": 21658
}
```

Actually better move them in the event
```
'zigbee2mqtt/remote/action', payload '{"command": "off", "alias": "single_press", "group": 21658}'
```

Koenkk:
Can you check if the sensor has a value in HA when an action is published?

andrei-lazarov:
Yes. I tried to automate a bulb with `brightness_move_to_level` + `action_level` coming from the remote. Tested with both "device trigger" and "entity trigger" (`experimental_event_entities`). It's fine on my pc. The bulb always gets the latest level.

I can see in the logs, the state is updated first, then the action is published. 

```
[2026-07-15 01:30:14] info:     z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/ikea_bilresa_scroll_remote', payload '{"action":"brightness_move_to_level","action_group":21658,"action_level":71,"action_transition_time":1,"battery":42,"identify":null,"last_seen":"2026-07-15T01:05:47+03:00","linkquality":224,"voltage":2500}'
[2026-07-15 01:30:14] info:     z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/ikea_bilresa_scroll_remote/action', payload 'brightness_move_to_level'
[2026-07-15 01:30:14] info:     z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/philips_hue_cws_800_2', payload '{"brightness":71,"color":{"h":32,"hue":32,"s":82,"saturation":82,"x":0.4582,"y":0.4102},"color_mode":"color_temp","color_temp":367,"effect":null,"effect_color":null,"effect_speed":null,"last_seen":"2026-07-15T01:04:56+03:00","linkquality":255,"philips_raw":"070001166f01","power_on_behavior":null,"state":"ON","update":{"installed_version":16786434,"latest_release_notes":null,"latest_source":"https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/Hue/100B-0114-01002502-ConfLightBLE-Lamps-EFR32MG21.zigbee","latest_version":16786690,"state":"available"}}'
```

Do you think this can fail in HA? If the system is slow and the device has many state fields? Could the action come up first?

Koenkk:
> Do you think this can fail in HA? If the system is slow and the device has many state fields? Could the action come up first?

Great! I was wondering if the event handling in HA was atomic but it seems it is, thanks for confirming.

andrei-lazarov:
I tested some more, but the HA sensors are not a good approach.
They have a value when the action is published, but they can lose the value during the automation (if Z2M publishes anything)

When HA extension is enabled, Z2M publishes null for all fields that don't have a value (to clear stale HA states).
The `action_x` fields don't have a value because they are not cached (for normal MQTT / non-HA usage).

So the solution would be: include these parameters as payloads in "device trigger" and "entity trigger" + have a way to use them in HA..

burmistrzak:
> I tested some more, but the HA sensors are not a good approach. They have a value when the action is published, but they can lose the value during the automation (if Z2M publishes anything)

@andrei-lazarov Yeah, action sensors are an anti-pattern and should be deprecated in favor of events.

> When HA extension is enabled, Z2M publishes null for all fields that don't have a value (to clear stale HA states). The `action_x` fields don't have a value because they are not cached (for normal MQTT / non-HA usage).

This always annoys me because HA throws an error... 🥸

> So the solution would be: include these parameters as payloads in "device trigger" and "entity trigger" + have a way to use them in HA..

Indeed. Was about to start working on this when I saw your PR.

AFAICT, devices (e.g. Hue Tap Dial) with multiple buttons only expose a single event entity. Ideally, we would have five entities here, one for each button and one for the dimmer dial.
Whenever possible, we should also supply [predefined event types](https://developers.home-assistant.io/docs/core/entity/event#button).

andrei-lazarov:
You may continue on this topic 🙂 I didn't make any progress.

Why do we need an entity for each button?

burmistrzak:
> You may continue on this topic 🙂 I didn't make any progress.

@andrei-lazarov Alright! 👌

> Why do we need an entity for each button?

Concurrency. But this depends on device type. Zigbee Green Power switches can at most emit one event at any time. Battery-powered devices don't necessarily have this limitation and can emit multiple events simultaneously.

andrei-lazarov:
I don't think there is any concurrency issue with either "device trigger" and "event trigger".

When you press two buttons at the same time, the events reach Z2M + are published in (very) quick succession. I see HA detects all triggers "simultaneously".

image


We have always been fine with a single action field?

burmistrzak:
> I don't think there is any concurrency issue with either "device trigger" and "event trigger".

Oh, that's good to know! One less thing to worry about. 😊

> We have always been fine with a single action field?

I guess the _real_ problem is how we're exposing actions to HA. That's what I'm currently hacking on.

With a Hue Tap Dial, you can't _easily_ distinguish between different buttons because they are only passed along as a custom `button` attribute. E.g.: `{"button":"button_1", "event_type":"press_release"}`
AFAIK, best practice here would be to expose an event entity for each of the four buttons as well as the dial.

I'll probably won't touch the payload structure before v3.0, mainly because it's already a huge mess.
We should really think hard about completely revamping the payload format because at the moment, it's an _interesting_ mashup of HA, ZCL, and custom data models. Personally, I'd prefer dotdot-style payloads over MQTT instead of what we have currently. 😳

PR #12685, opened 2026-07-14
fix: Leviton DG3HL: expose and preserve dimmer configuration

PaulMcMillan:
I know this is an older device, but I wanted to add support for a couple specific behaviors and figured I'd flesh the rest of it out while I was working on it. The only question I have is whether it makes sense to expose the OTA endpoint since I couldn't find any updated firmware. I've sent a note to Leviton asking them for a copy - their certifications show newer firmware does exist. Also, not sure how you feel about the pretty meaningless tests, but they're there if you want em.

## Summary

- expose the DG3HL's tested Level Control settings: `on_transition_time`, `off_transition_time`, `on_level`, and `execute_if_off`
- expose the three useful, writable Ballast Configuration settings: minimum level, maximum level, and power-on level
- enable Identify and OTA support advertised by the device
- restore `execute_if_off` after a mains power cycle when the saved Zigbee2MQTT state says it should be enabled
- keep the standard `m.light()` command behavior, including normal `MoveToLevel` and `MoveToLevelWithOnOff` handling

Most of this change is declarative and uses standard modern extensions. The only device-specific converter handles one firmware quirk: the DG3HL clears the Level Control `ExecuteIfOff` option after losing mains power and does not send a `deviceAnnounce` when it comes back. That means the announce-based restoration used by some other lights cannot work for this device.

### Level Control

The light now exposes:

- `on_transition_time`
- `off_transition_time`
- `on_level`, including the standard `previous` value
- `execute_if_off`

A brightness command with an explicit on/off state continues to use the standard `MoveToLevelWithOnOff` path. A plain `MoveToLevel` command can update `CurrentLevel` while the load remains off when `execute_if_off` is enabled. With `on_level` set to `previous`, a later direct-bound or Zigbee On command uses that preloaded level.

Mains startup level is controlled by the device's Ballast Configuration `powerOnLevel` attribute.

### Ballast Configuration

Only the useful writable controls are exposed:

- `ballast_minimum_level`
- `ballast_maximum_level`
- `ballast_power_on_level`

The generic Zigbee cluster definition currently limits `powerOnLevel` to 254. The DG3HL was verified to accept 255 and use it as the “restore previous level” sentinel, so the cluster override raises that limit only for this model. Fixed physical limits and ballast-status diagnostics are intentionally not exposed.

## Hardware verification / Testing

Tested on a Leviton DG3HL-1BW with firmware `0x03000025` (reported decimal version `50331685`).

### Ballast levels

- read the original minimum and maximum values as 10 and 254
- wrote the minimum level to 60 and confirmed a direct read returned 60
- sent a below-minimum brightness request through the standard command path; the reported level remained clamped to 60 and the load transitioned off
- wrote `powerOnLevel` as 255, confirmed it by direct read, and observed the previous level restored after mains power returned

### On level, transitions, and off-state preload

- wrote `on_level` to 160 and confirmed a subsequent On command used level 160
- restored `on_level` to `previous` (255)
- wrote and read back separate 2-second on and off transition times and observed the transitions
- enabled `execute_if_off`, turned the load off, then sent a plain `MoveToLevel`; `CurrentLevel` changed without turning the load on
- sent a later On command with `on_level=previous`; the device turned on at the preloaded level
- confirmed commands containing an on/off state still use the normal `MoveToLevelWithOnOff` behavior

This is the behavior needed by controllers such as Adaptive Lighting: they can preload the next brightness while the dimmer is off without energizing the load, and a physical/direct-bound On command then uses that brightness.

### Power-cycle quirk

The volatile option was tested with Zigbee2MQTT debug logging and explicit attribute reads:

1. A direct `level_config` read returned `options=1` before removing mains power.
2. After a mains cycle, the DG3HL did not send `deviceAnnounce`.
3. Its startup traffic was a `genOnOff` attribute report with transaction sequence number 0, followed about three seconds later by a `genLevelCtrl.currentLevel` report with transaction sequence number 1.
4. A direct read after startup returned `options=0`, confirming the firmware had cleared the bit rather than Zigbee2MQTT merely showing cached state.
5. With the restoration converter loaded, the sequence-1 level report caused a write of `{options: 1}`.
6. A subsequent direct read returned `options=1`.

The restoration checks the saved `level_config.execute_if_off` value and writes only when it is exactly `true`; users who leave the option disabled are not changed. Zigbee transaction sequence numbers wrap, so an ordinary later report can eventually also use sequence number 1. In that case the guarded operation is an infrequent, idempotent rewrite of the already-requested value.

### Other capabilities

- triggered Identify and observed the device flash red
- confirmed the endpoint advertises the OTA cluster and that an OTA query works; no newer image was available during the test

## Automated validation

Added DG3HL-specific tests covering:

- the exact retained level and ballast exposes and omission of unsupported/diagnostic fields
- the device-scoped `powerOnLevel=255` schema and write
- restoration when `execute_if_off` is saved as true
- no restoration when it is false
- ignoring later sequence numbers and unrelated reports

Validation run on the squashed commit:

- `pnpm run check`
- `pnpm run build`
- `pnpm test` (22 test files, 651 tests passed)

The commit hook reran the build and full test suite successfully while creating the final commit.

### DIFF
--- a/src/devices/leviton.ts
+++ b/src/devices/leviton.ts
@@ -1,9 +1,11 @@
+import {Zcl} from "zigbee-herdsman";
+
 import * as fz from "../converters/fromZigbee";
 import * as tz from "../converters/toZigbee";
 import * as exposes from "../lib/exposes";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
-import type {DefinitionWithExtend, Fz} from "../lib/types";
+import type {DefinitionWithExtend, Fz, KeyValue} from "../lib/types";
 import * as utils from "../lib/utils";
 
 const e = exposes.presets;
@@ -21,6 +23,17 @@ const fzLocal = {
             }
         },
     } satisfies Fz.Converter<"genLevelCtrl", undefined, ["attributeReport", "readResponse"]>,
+    restore_execute_if_off_after_power_cycle: {
+        cluster: "genLevelCtrl",
+        type: ["attributeReport"],
+        convert: async (model, msg, publish, options, meta) => {
+            // DG3HL does not announce after a mains cycle, but its first level report always uses TSN 1.
+            const levelConfig = meta.state.level_config as KeyValue | undefined;
+            if (msg.data.currentLevel !== undefined && msg.meta.zclTransactionSequenceNumber === 1 && levelConfig?.execute_if_off === true) {
+                await msg.endpoint.write("genLevelCtrl", {options: 1});
+            }
+        },
+    } satisfies Fz.Converter<"genLevelCtrl", undefined, ["attributeReport"]>,
 };
 
 export const definitions: DefinitionWithExtend[] = [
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks!

PR #12686, opened 2026-07-14
FIX: Modify color temperature range for Innr RB 279 T

Burki24:
Technical data: https://eprel.ec.europa.eu/fiches/lightsources/Fiche_697229_EN.pdf


Koenkk:
Thanks!

PR #12541, opened 2026-06-25
fix: _TZ3290_gnl5a6a5xvql7c2a: remove battery

andrei-lazarov:
- fixes https://github.com/Koenkk/zigbee2mqtt/issues/32395
- fixes https://github.com/Koenkk/zigbee2mqtt/issues/21172

I removed the `_TZ3290_gnl5a6a5xvql7c2a` fingerprint from the Moes definition and removed the extra logic.

Now it will match to the Tuya definition without battery: [Tuya ZS06](https://www.zigbee2mqtt.io/devices/ZS06.html#tuya-zs06) (`zigbeeModel: ["TS1201"]`)

https://github.com/Koenkk/zigbee-herdsman-converters/blob/b4a5081837496bb56b15407393c38c45501568b9/src/devices/tuya.ts#L13634

PR #12689, opened 2026-07-14
fix: discovery of TS1201 based IR fails, and add expose zosung learned timings to supported devices

jbouwh:
In production discovery of my TS1201 (model _TZ3290_gnl5a6a5xvql7c2a) works, but after #12541 it is no longer discovered.
This PR fixes the discovery (tested locally).
Also it exposes the `learned_ir_timings()` from the `zosung` library and adds it the all supported devices.

After this PR the IR blaster shows like:
image


Koenkk:
Thanks!

PR #11859, opened 2026-04-01
feat(add): Mumubiz CZV20 Zigbee smart water valve

juslex:
Corrected version of PR #11827, incorporating Koenkk's feedback

PR #12690, opened 2026-07-14
feat: eWeLink CK-BL702-MSW-01(7010) / Mumubiz CZV20: expose countdown auto-off timer

juslex:
Adds a `countdown` numeric (0–6500 s) to the eWeLink `CK-BL702-MSW-01(7010)` plug that turns the device on and automatically turns it off again after the given time via `genOnOff onWithTimedOff` (setting `0` turns it off). The `onTime` attribute is decoded back into `countdown` for feedback when reported.

Primarily useful for the **Mumubiz CZV20** water valve (a white-label of this plug, added in #11859) — e.g. run irrigation / pool filling for a fixed time and auto-close. The genOnOff `on_time`/`off_wait_time` set-payload already works on this definition; this exposes it as a first-class control so it shows up in the frontend / Home Assistant.

### DIFF
--- a/src/devices/ewelink.ts
+++ b/src/devices/ewelink.ts
@@ -45,6 +45,16 @@ const fzLocal = {
             }
         },
     } satisfies Fz.Converter<"genOnOff", undefined, ["attributeReport", "readResponse"]>,
+    ewelink_countdown: {
+        cluster: "genOnOff",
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12699, opened 2026-07-16
fix: improve Aqara T3 light configs

blakegong:
The removed attributes do NOT work on this device:
- `on_off_transition_time` (`on_transition_time` and `off_transition_time` do work)
- `execute_if_off`

And color temperature (similar to many other Aqara lights) only goes down to 166.


Koenkk:
Thanks!

PR #12704, opened 2026-07-16
fix: Aqara DS-K02D/DS-K02E: disable presentValue reporting instead of max=0

glueckstein:
Sets max_report_interval to 0xFFFF (disable reporting) instead of 0 on the wireless button endpoints, as discussed in koenkk/zigbee2mqtt#32059. Button presses are firmware-generated unsolicited reports and are unaffected — tested on 7 devices (fw 4122) for 3 days: zero ghost actions, 20/20 physical presses delivered, survives Z2M restarts. Version bumped so deployed devices get reconfigured once.


Koenkk:
Thanks!

PR #12707, opened 2026-07-17
Feature/add cover open closed inversion

ThatDeltaGuy:
## Problem
For window covering devices, the state (OPEN/CLOSE) published by fz.cover_position_tilt was derived from the raw position value using only the device-definition's coverInverted meta flag — it never consulted the user-facing invert_cover option at all. This meant that if a device's percentage reporting was inverted and a user enabled invert_cover to fix it, the position value would come out correct but state would then be wrong, with no way to fix it independently.

see [`#27213 - Cover State flip`](https://github.com/Koenkk/zigbee2mqtt/issues/27213)

## Changes
- Added invert_cover_state — a new option that inverts the OPEN/CLOSE state independently of position, for both:
   - fz.cover_position_tilt: the derived state value (for both lift- and tilt-based state derivation)
   - tz.cover_state: which raw ZCL command (upOpen/downClose) gets sent for open/close
- invert_cover keeps its existing name/property (no config-breaking rename), but now has an explicit, clearer label — "Invert cover position" — to read as a clean pair alongside invert_cover_state.
- Added legacyNames/withLegacyNames(...) to the base Expose class — schema-only groundwork for a future option-rename migration path (e.g. eventually marking invert_cover_position as replacing invert_cover). Not yet used by any option, as it requires change to the zigbee2mqtt repository settingsMigration.ts.
- Added tests for fz.cover_position_tilt, tz.cover_position_tilt, and tz.cover_state covering: default (non-inverted) behavior, invert_cover affecting only position, invert_cover_state affecting only the OPEN/CLOSE label/commands independently, both combined, interaction with the device-level coverInverted meta, tilt-derived state, and cover_mode bit decoding.

## Testing
tsc --noEmit — clean
pnpm run check (biome) — clean
pnpm run test — 709/709 passing

### DIFF
--- a/src/lib/exposes.ts
+++ b/src/lib/exposes.ts
@@ -33,6 +33,7 @@ export class Base {
     features?: Feature[];
     category?: "config" | "diagnostic";
     homeassistant?: HomeAssistant;
+    legacyNames?: string[];
 
     withEndpoint(endpointName: string) {
         this.endpoint = endpointName;
@@ -86,6 +87,11 @@ export class Base {
         return this;
     }
 
+    withLegacyNames(...names: string[]) {
+        this.legacyNames = names;
+        return this;
+    }
+
     validateCategory() {
         switch (this.category) {
             case "config":
@@ -135,6 +141,7 @@ export class Base {
         }
         target.category = this.category;
         target.homeassistant = this.homeassistant ? {...this.homeassistant} : undefined;
+        target.legacyNames = this.legacyNames ? [...this.legacyNames] : undefined;
     }
 }
 
--- a/src/lib/exposes.ts
+++ b/src/lib/exposes.ts
@@ -841,9 +841,17 @@ export const options = {
             .withDescription(
                 `Number of digits after decimal point for ${name}, takes into effect on next report of device. This option can only decrease the precision, not increase it.`,
             ),
+    invert_position: () =>
+        new Binary("invert_position", access.SET, true, false).withDescription(
+            "Inverts the cover position, false: open=100,close=0, true: open=0,close=100 (default false).",
+        ),
     invert_cover: () =>
         new Binary("invert_cover", access.SET, true, false).withDescription(
-            "Inverts the cover position, false: open=100,close=0, true: open=0,close=100 (default false).",
+            "Inverts the cover position, false: open=100,close=0, true: open=0,close=100 (default false) (deprecated, use invert_position).",
+        ),
+    invert_state: () =>
+        new Binary("invert_state", access.SET, true, false).withDescription(
+            "Inverts the cover state (open/close), independent of the position (default false).",
         ),
     illuminance_raw: () => new Binary("illuminance_raw", access.SET, true, false).withDescription("Expose the raw illuminance value."),
     color_sync: () =>
--- a/src/lib/exposes.ts
+++ b/src/lib/exposes.ts
@@ -11,7 +11,7 @@ import type {
 } from "./constants";
 import {thermostatSetpointChangeSource} from "./constants";
 import type {Access, LevelConfigFeatures, Range} from "./types";
-import {formatLabel} from "./utils";
+import {getLabelFromName} from "./utils";
 
 export type Feature = Numeric | Binary | Enum | Composite | List | Text;
 export interface HomeAssistant {
@@ -199,10 +199,10 @@ export class Binary extends Base {
     // biome-ignore lint/style/useNamingConvention: ignored using `--suppress`
     value_toggle?: string;
 
-    constructor(name: string, access: number, valueOn: string | boolean, valueOff: string | boolean, label?: string) {
+    constructor(name: string, access: number, valueOn: string | boolean, valueOff: string | boolean) {
         super();
         this.name = name;
-        this.label = formatLabel(label ?? name);
+        this.label = getLabelFromName(name);
         this.property = name;
         this.access = access;
         this.value_on = valueOn;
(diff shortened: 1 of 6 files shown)


Koenkk:
> it never consulted the user-facing invert_cover option at all. 

Shouldn't we modify the existing option such that is taken into account?

ThatDeltaGuy:
It could be modified to do both, but I didn't want to assume that all models of blinds work the same way. Its a better assumption that they probably don't. 

This way the existing behaviour can be preserved if that is working for other people/blinds.

Koenkk:
I do not fully understand yet what this pr aims to fix, does that mean that with the current invert cover option the state is incorrect? I would prefer not to have 2 options doing very similar things as I think that is confusing for users

ThatDeltaGuy:
so, the current way things work, the user can flip the position, so that when the device is at 0% it states 100% and vice versa. but this doesn’t alter the open/closed position. this means that i can invert the position sending so that when the blind is open it says it's 0% or 100%. however it will say 'Closed' either way as it doesn’t affect the open closed options. Home assistant always treats open as up and closed as down. and so the open/closed buttons are inverted.

I could just have changed the inverted button to control both the position and the open/closed state but I’m conscious that this could differ for other blinds/covers, that it could differ for other home automation servers like homey, and that the inverted issue has been around so long that people could have implemented workarounds. Meaning that it has the potential to suddenly break a lot of other devices if combined.

This way it keeps the existing functionality in tact, and has the ability to allow people to tailor the state and position to their specific needs.

I would also note that state and position are already 2 distinct things in the cover 'Exposes' and that 'Invert cover' currently specifies that it only 'Inverts the cover position'

---

image
image

All the invert cover toggle does is invert the position of this slider:
image

It doesnt change the state, so the 'down' button doesn’t do anything:
image

I have to 'open' it, even though it is actually open at 0%, which would be the 'up' button it but its disabled:
image

Koenkk:
I would say this is definitely a bug and should require fixing, therefore I propose to update the existing option.

ThatDeltaGuy:
Ok, I’ve changed it to updating the existing option

src/lib/exposes.ts
```diff
@@ -86,6 +87,11 @@ export class Base {
         return this;
     }
 
+    withLegacyNames(...names: string[]) {
```
Koenkk: I think this can be removed? Doesn't seem to be used?

ThatDeltaGuy: When I was looking at adding a second toggle I noticed there was no way to rename an existing toggle id/name. There are some migration steps/methods within the zigbee2mqtt project for other things, but no way to migrate an option name. To do that, it would require the addition of this extra property, and then updating zigbee2mqtt to look for the legacy names in the configuration.yaml as well as the set name, updating it as needed. 

It felt like an obvious missing feature, so i decided to leave it, even though I was no longer needing to use it in this instance. I was going to look at adding the zigbee2mqtt half after this, but I can remove it if you'd like.

Koenkk: Let's remove it for now.

ThatDeltaGuy: I've removed it

Koenkk:
Thanks!

convicte:
Having installed `2.14.0-1` all my previously correctly working covers became inverse, and no matter if I selected `inverted` to true or false, either the position or the state would be wrong.

My covers with `inverted` set to false would OPEN when asked to CLOSE (from both Z2m and HA), and preempting the question, the motor direction is not reversed either.
When inverted Z2m would state they are closed while HA would see them as open and trying to close would be opening them, so something is certainly OFF.

image

image

image


Currently back to `2.13.0-1`:

`
{
    "calibration": "OFF",
    "calibration_time": 21.1,
    "device": {
        "applicationVersion": 67,
        "dateCode": "",
        "friendlyName": "Office shutter",
        "hardwareVersion": 1,
        "ieeeAddr": "0xa4c1387dea7aa75c",
        "manufacturerID": 4417,
        "manufacturerName": "_TZ3000_4uuaja4a",
        "model": "TS130F",
        "networkAddress": 19344,
        "powerSource": "Mains (single phase)",
        "stackVersion": 0,
        "type": "Router",
        "zclVersion": 3
    },
    "indicator_mode": "off",
    "last_seen": 1788299974431,
    "linkquality": 21,
    "motor_reversal": "OFF",
    "moving": "STOP",
    "position": 0,
    "state": "CLOSE",
    "switch_type": "momentary",
    "update": {
        "installed_version": 67,
        "latest_release_notes": null,
        "latest_source": null,
        "latest_version": 67,
        "state": "idle"
    },
    "backlight_mode": "OFF"
}
`
Please let me know what information would be required to further troubleshoot this as I have 20 such covers and all started behaving `inverted` when they shouldn't have on `2.14.0`

oepoemoepoe:
I am suffering from the same problem, all my covers became unusable with 2.14.0

VerdelothRahl:
Hello,
After this update all of my automations for covers flipped. What was set to open now closes and vice-versa.
So 'Close cover' action now opens my covers.

The state is read correctly 'Exposes' tab - eg with shutter physically open:
State: Open
Position: 100

However pressing close button with that state tries to Open the shutter, and pressing open button actually closes it.

FatihiZakaria:
Hello, 

After this update my roller shutters are no longer working correctly.
It's as if the controls were reversed, but in a way that makes no sense.

convicte:
Before the contributors lock this pull for excessive comments, please upvote the previous input if yours is identical rather than putting another comment that doesn't provide any further clarity, please and thank you.

ThatDeltaGuy:
I have the same issue, i'm getting a fix in now

ThatDeltaGuy:
I've put in the fix: #13086 

I misunderstood the function of the meta-invert. it's only supposed to be used on the receiving side. So, as I added it to the sending side, it flipped the commands, and therefore flipped twice if inverted was toggled. I've removed the meta-invert, and added more tests now that i can infer how some other blinds work.

Issue #12709, opened 2026-07-17
BTicino F20T60A actually doesn't exist


Ansuel:
Hi I'm writing this issue to gather some info on the BTicino F20T60A. After a long research (at least 2 month) it seems to me every information about this device is wrong.

I wanted this device as from https://www.zigbee2mqtt.io/devices/F20T60A.html it does offer both voltage and ampere info contrary to the Legrand 412015.
The problem is that I can't find anywhere on the web any proof that it actually could.

There is currently a converter in bticino.ts for this device but I couldn't find anybody in the web that actually made use of it (no picture, no output) So I wonder... how it was submitted?

In a big issue ticket, it was said that such device required the gateway to update and receive the firmware to enable the extra feature.

With the lovely help of Amazon where you can """"rent"""" devices, I took a gateway and added a sniffer to it. It did download some firmware for the device (I actually took 2 different F20T60A, one new and one dated 2019) but still Legrand 412015 feature. (I attached the firmware that were downloaded if anyone wants to inspect)

On top of it on some other issue ticket it was reported that those values were always 0 (similar to Legrand) so actual functionality wasn't ever tested.

I have 2 theory currently:
- In some time during the lifetime of the device, there was a specific firmware for it that had these functionality
- The device submission was wrong from the start and both the wiki page and the converter are wrong

I verified the firmware version after the gateway update and they match what we have in the OTA repository. If anyone have a clue or have any idea how to verify this I would really appreciate it.

In the current state as it gets correctly identified it's a simple clone of Legrand 412015. Voltage and ampere are exposed but always zero.

andrei-lazarov:
Yeah, it seems the converter in `bitcino.ts` is wrong and unused. I can remove it (and the generated page).
The one in `legrand.ts` is used and likely correct 

- https://github.com/Koenkk/zigbee-herdsman-converters/issues/2019

> The modelID is ` DIN power consumption module\u0000\u0000` which matches the legrand one, so the F20T60A in ~~devices.js~~ bitcino.ts is never used (because that modelID is `Bticino Din power consumption module `)

Ansuel:
@andrei-lazarov what i would like to verify and confirm is if the converter was entirely crafted or not... Cause if it i'm not sure how it worked in the first place if the guy that submit added an external converter. It shouldn't have worked anyway as the modelID was different. Am I wrong?

(removing the generated page might be wrong, just the extra entry and say that it's exactly Legrand 412015.

[...]

PR #12715, opened 2026-07-18
fix: BTicino F20T60A and FC80GCS: Remove unused converters

andrei-lazarov:
- https://github.com/Koenkk/zigbee-herdsman-converters/issues/12709

> The modelID is ` DIN power consumption module\u0000\u0000` which matches the legrand one, so the F20T60A in ~~devices.js~~ bitcino.ts is never used (because that modelID is `Bticino Din power consumption module `)

### DIFF
--- a/src/devices/legrand.ts
+++ b/src/devices/legrand.ts
@@ -514,7 +514,7 @@ export const definitions: DefinitionWithExtend[] = [
         description: "DIN power consumption module",
         whiteLabel: [
             {vendor: "Legrand", description: "DIN power consumption module", model: "412172", fingerprint: [{modelID: " Smart shedder module"}]},
-            {vendor: "BTicino", description: "DIN power consumption module", model: "FC80GCS", fingerprint: [{modelID: " Smart shedder module"}]},
+            {vendor: "BTicino", description: "DIN power consumption module", model: "FC80GCS", whiteLabelOf: "412172"},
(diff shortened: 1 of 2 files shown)


src/devices/legrand.ts
```diff
@@ -515,6 +515,7 @@ export const definitions: DefinitionWithExtend[] = [
         whiteLabel: [
             {vendor: "Legrand", description: "DIN power consumption module", model: "412172", fingerprint: [{modelID: " Smart shedder module"}]},
             {vendor: "BTicino", description: "DIN power consumption module", model: "FC80GCS", fingerprint: [{modelID: " Smart shedder module"}]},
+            {vendor: "BTicino", description: "DIN power consumption module", model: "F20T60A"},
```
Ansuel, line 518: Actually the device gets identified with the space so the ID. On my module they get directly detected as 412015. No reason to add a whiteLabel.

andrei-lazarov: This whiteLabel without a fingerprint does not affect Zigbee2MQTT. It's only for the docs.

I deleted the unused converter -> bticino page gets automatically removed on next generation.

Adding whiteLabel will:
- make BTicino F20T60A show in "supported devices", clicking it will redirect to legrand
- add a row below the picture on the legrand page "White-label: BTicino F20T60A"

I also made the bticino link redirect to the legrand page

Koenkk:
Thanks!

PR #12721, opened 2026-07-19
SMLIGHT: add SLZB-06Mg26U fingerprint to Router definition

Psykosoldier:
## Problem

When an SLZB-06Mg26 (EFR32MG26) unit is configured in **Router mode**, it reports its Zigbee `modelID` as `SLZB-06Mg26U` (with a trailing `U`), not `SLZB-06Mg26`. Since only `SLZB-06Mg26` (without the `U`) was listed in the `zigbeeModel`/`whiteLabel` arrays of the SMLIGHT Router definition, Zigbee2MQTT falls back to a generated definition and shows the device as "not supported", even though it is the same hardware/router as the already-supported `SLZB-06Mg26`.

Interview succeeds fine, and the device works correctly as a Zigbee router — this is purely a fingerprint/whiteLabel gap.

## Change

Add `SLZB-06Mg26U` to the `zigbeeModel` array and add a matching `whiteLabel` entry (same pattern as the other SMLIGHT Router variants in this file), so the device is recognized under its correct model name instead of falling back to a generated definition.

## Testing

- `pnpm run build` — OK
- `pnpm run check` — OK (no new issues; pre-existing biome.json schema-version info message only)
- `pnpm test` — 713/713 tests passing
- Verified against a real SLZB-06Mg26 unit configured in Router mode, reporting `modelID: SLZB-06Mg26U`


Koenkk:
Thanks!

PR #12723, opened 2026-07-19
Fix names of generated binary/analog clusters and use apptype to get the device class

luar123:
This PR sets the name in `extender{Analog/Binary}{In/Out}put()` to a fixed string instead of using the description attribute. For Analog clusters the apptype attribute is used to get the name.

Currently, names are generated from description attribute. This can lead to many issues, because the name is used as some kind of device class in z2m. For example a binaryInput called action_button would not work as expected, see https://github.com/Koenkk/zigbee2mqtt/issues/32546

The description attribute is still used to set the label. So in the frontend or HA it doesn't change. With HA there was one issue: the label is not used if a device class is set, so we loose custom names. ~~To re-enable them I added a `labelIsCustom` flag that is set in `expose.withLabel()`. Requires a change in z2m and fixed tests: https://github.com/Koenkk/zigbee2mqtt/pull/32607~~
Use `withHomeassistant` to set the name.

Breaking note (from [Nerivec](https://github.com/Koenkk/zigbee-herdsman-converters/pull/12723#issuecomment-5039209951)):
```
> [!IMPORTANT]
> Due to a potentially crippling naming conflict, this release introduces a breaking change for **devices relying on automatically generated definitions** (e.g. ESPHome) that use any of the following clusters: `Analog Input` (`0x000c`), `Analog Output` (`0x000d`), `Binary Input` (`0x000f`), `Binary Output` (`0x0010`).
>
> The **`name` property of the exposed feature** created by the generated definition is now using pre-defined patterns. It is based on application type for the Analog ones. It uses `binary_input` and `binary_output` respectively for the Binary ones. `_<endpoint_id>` is appended to each name for multi-endpoint devices.
> **If you rely on this property in your setup, you will have to make the appropriate changes.**
>
> NOTE: If you have an external converter for your custom device, it is not affected.
```


Nerivec:
This is a breaking change - affects all devices currently relying on generated definition that use either of these extenders.
Could be quite a lot considering the popularity of esphome devices lately...

luar123:
Hm yes, but is there an alternative? As is, it is broken and esphome user complain that their devices are not working. I guess it is mostly esphome that is using those as they were added for esphome in the first place.

The sooner it is changed the better. Official esphome Zigbee support is in a very early state. on esp32 support was added two months ago (with breaking change a week ago), on nrf52 earlier this year but those are not so popular I guess.

Koenkk:
- Is this a breaking change in the sense that it will discover new entities? (and remove the old ones?)
- I prefer to use the `withHomeAssistant` construct here, (search this repo for examples)

luar123:
> * Is this a breaking change in the sense that it will discover new entities? (and remove the old ones?)

Yes, for users that rely on generated definitions and use Analog/Binary Input/Output clusters (mainly esphome users). If they created an external definition from it, it shouldn't be an issue.

>     * I prefer to use the `withHomeAssistant` construct here, (search this repo for examples)

ok then I will revert `labelIsCustom` and add a label attribute to `interface HomeAssistant`. Won't solve the breaking change issue though.

Nerivec:
> As is, it is broken and esphome user complain that their devices are not working.

As far as I can tell, this is an edge case, not a "general problem"; user has to set the `description` attribute of the cluster same as one of the [cache patterns](https://github.com/Koenkk/zigbee2mqtt/blob/aa909a8a62f76e2dd98ace3a172bca88ee56f5fe/lib/state.ts#L11-L31).
Unless I missed another problem with it?
It needs fixing, but doing so will cause friction (& probably a few disagreements from not being able to set the name anymore). :sweat_smile: 
I wonder if it might be better to introduce a blacklist that warns if `description` is detected as reserved and remaps it if it is. As long as it is only that cache props thing that is the problem, should be a smoother result (at least until v3).

---

Notes (some won't apply depending on the direction taken re above):
- `name: name,` should be `name,`
- `name = name ?? ...` should be a wrapped `if (!name)`
- should no longer need `.replace(/\s+/g, "_").toLowerCase()` if `name` is statically declared
- we should also remove the functions that wrap object declarations. These can just be `const` at the top of the file and used where needed (with bit shifthing done at call site instead - avoids doing it twice as bonus).

PS: be sure to remove the extra logging once the PR is ready.

luar123:
> I wonder if it might be better to introduce a blacklist that warns if `description` is detected as reserved and remaps it if it is. As long as it is only that cache props thing that is the problem, should be a smoother result (at least until v3).

Initially, I thought about this as well. It would solve the main issue but has some drawbacks:

* mapping of application type is not possible, except for HA with `withHomeassistant()`
* names could map to one of [HA *_discovery_lookups](https://github.com/Koenkk/zigbee2mqtt/blob/aa909a8a62f76e2dd98ace3a172bca88ee56f5fe/lib/extension/homeassistant.ts#L62). Could have undesired side effects or be used intentionally by users (-> complain if we ever change it)
* [cache patterns](https://github.com/Koenkk/zigbee2mqtt/blob/aa909a8a62f76e2dd98ace3a172bca88ee56f5fe/lib/state.ts#L11-L31) live in z2m, so we either have to move them to zhc or duplicate.

> Notes (some won't apply depending on the direction taken re above):
> 
>     * `name: name,` should be `name,`
> 
>     * `name = name ?? ...` should be a wrapped `if (!name)`
> 
>     * should no longer need `.replace(/\s+/g, "_").toLowerCase()` if `name` is statically declared
> 
>     * we should also remove the functions that wrap object declarations. These can just be `const` at the top of the file and used where needed (with bit shifthing done at call site instead - avoids doing it twice as bonus).
>
> PS: be sure to remove the extra logging once the PR is ready.

Thanks

I will wait for your comments, but if I get you right the way to proceed is: revert name change, filter cache patterns and add everything else with `withHomeassistant()`

Nerivec:
Let's see what Koenkk says; might be worth the breaking change at that level, we can be explicit about it in the next release notes. Generated stuff is in its own category :grin: 

Also shouldn't use `undefined` for AI/AO name mapping if there is an actual value per spec (I see 5 in current PR?).

We should look into moving all these maps to ZH spec as well, have it with the rest
https://github.com/Koenkk/zigbee-herdsman/blob/master/src/zspec/zcl/definition/

Koenkk:
Then I would propose to break indeed and mention it in the release notes. ESPHome Zigbee support is in early stages so better to do it now, @luar123 can you prep something sensible which I can put there?

luar123:
Will update the PR and add a breaking change message to the description.

> Also shouldn't use undefined for AI/AO name mapping if there is an actual value per spec (I see 5 in current PR?).

good point, I just added the ones that are available in HA, but would make sense to add all.

Nerivec:
Might I suggest something like this for the release notes?
```md
> [!IMPORTANT]
> Due to a potentially crippling naming conflict, this release introduces a breaking change for **devices relying on automatically generated definitions** (e.g. ESPHome) that use any of the following clusters: `Analog Input` (`0x000c`), `Analog Output` (`0x000d`), `Binary Input` (`0x000f`), `Binary Output` (`0x0010`).
>
> The **`name` property of the exposed feature** created by the generated definition is now using pre-defined patterns. It is based on application type for the Analog ones. It uses `binary_input` and `binary_output` respectively for the Binary ones. `_<endpoint_id>` is appended to each name for multi-endpoint devices.
> **If you rely on this property in your setup, you will have to make the appropriate changes.**
>
> NOTE: If you have an external converter for your custom device, it is not affected.
```

---

@luar123 I think the currently pending logic of the PR has an issue with Analog, where it will duplicate the `name` in multi-endpoint scenario (`ID` not appended). So you could end up with e.g. two exposes named `temperature`.
_Was a problem in the before-logic as well, though due to `description`._

luar123:
Have included all suggestions.
Update:
- revert `labelIsCustom` and use `withHomeassistant` instead
- remove endpoint suffix logic. This was completely broken before (inverted logic) and is not needed, as `expose.withEndpoint` handles this.
- Always add `analog_in_` or `analog_out_` prefix before name mapped from apptype. Otherwise AnalogInput, AnalogOutput and normal clusters (e.g. Temperature) could conflict if they are on the same endpoint.
- added missing apptypes
- refactor map lookup

I like the release notes proposal from Nerivec.

src/lib/generateDefinition.ts
```diff
@@ -448,297 +443,311 @@ async function extenderBinaryOutput(device: Zh.Device, endpoints: Zh.Endpoint[])
...
-        247: "J/h",
-        248: "ft³/d",
-        249: "m³/d",
-        250: "Wh/m³",
-        251: "J/m³",
-        252: "mol%",
-        253: "Pa·s",
-        254: "MMSCFM",
-    };
+const getfromApplicationType = (applicationTypeMap: {[key: number]: string}, applicationType: number): string | undefined => {
```
Nerivec: Remove this function. This is a simple `object[key]` no need for a function.
Bit shifting logic should be in caller as mentioned before (avoids doing it twice on top).

src/lib/generateDefinition.ts
```diff
@@ -448,297 +443,311 @@ async function extenderBinaryOutput(device: Zh.Device, endpoints: Zh.Endpoint[])
...
+    14: "s", // Time_Seconds
+} as const;
+
+const APPTYPE_NAME_LOOKUP: {[key: number]: string} = {
+    0: "temperature", // Temp_Degrees_C
+    1: "humidity", // Relative_Humidity_Percent
+    2: "pressure", // Pressure_Pascal
+    3: "flow", // Flow_Liters_Per_Sec
+    4: "percentage", // Percentage
+    5: "part_per_million", // Parts_Per_Million
```
Nerivec:
```suggestion
    5: "ppm", // Parts_Per_Million
```

src/lib/generateDefinition.ts
```diff
@@ -448,297 +443,311 @@ async function extenderBinaryOutput(device: Zh.Device, endpoints: Zh.Endpoint[])
...
+    6: "rotational_speed", // Rotational_Speed_RPM
```
Nerivec:
```suggestion
    6: "rpm", // Rotational_Speed_RPM
```

src/lib/generateDefinition.ts
```diff
@@ -448,297 +443,311 @@ async function extenderBinaryOutput(device: Zh.Device, endpoints: Zh.Endpoint[])
...
+    86: "imp gal/min",
+    87: "L/s",
+    88: "L/min",
+    89: "gal/min",
+    90: "°",
+    91: "°C/h",
+    92: "°C/min",
+    93: "°F/h",
+    94: "°F/min",
+    95: undefined,
```
Nerivec:
```suggestion
    // 95: undefined,
```

src/lib/generateDefinition.ts
```diff
@@ -448,297 +443,311 @@ async function extenderBinaryOutput(device: Zh.Device, endpoints: Zh.Endpoint[])
...
+    6: "rpm", // Rotational_Speed_RPM
+    7: "A", // Current_Amps
+    8: "Hz", // Frequency_Hz
+    9: "W", // Power_Watts
+    10: "kW", // Power_Kilo_Watts
+    11: "kWh", // Energy_Kilo_Watt_Hours
+    12: undefined, // Count
+    13: "kJ/kg", // Enthalpy_KJoules_Per_Kg
+    14: "s", // Time_Seconds
+} as const;
```
Nerivec:
```suggestion
};
```
No effect since typed generically.
Remove on maps below as well.

src/lib/generateDefinition.ts
```diff
@@ -448,297 +443,311 @@ async function extenderBinaryOutput(device: Zh.Device, endpoints: Zh.Endpoint[])
...
-        254: "MMSCFM",
-    };
+const getfromApplicationType = (applicationTypeMap: {[key: number]: string}, applicationType: number): string | undefined => {
+    // Only take type ( Bits 16 to 23 ) into account.
+    return applicationTypeMap[(applicationType >> 16) & 0xff];
+};
 
-    return bacnetUnitMap[bacnetUnit];
-}
+const APPTYPE_UNIT_LOOKUP: {[key: number]: string} = {
```
Nerivec:
```suggestion
const APPTYPE_UNIT: Record<number, string> = {
```

Same for all maps below.

Nerivec:
I don't think this is what Koenkk had in mind for the HA stuff; this appears to be present/executing even when no HA.
I'll let Koenkk judge though :grimacing:

test/generateDefinition.test.ts
Nerivec: Let's change this to `input(genBinaryInput, genAnalogInput), output(genBinaryOutput, genAnalogOutput), x2 endpoints` so we can make sure both A & B behave as expected in single & multi-endpoint scenarios.
Need to change `inputClusters` & `outputClusters` below for each endpoint and fix the expected result.

Should do the same to above test as well (missing `genAnalogInput`).

src/lib/generateDefinition.ts
```diff
@@ -384,21 +384,18 @@ async function extenderBinaryInput(device: Zh.Device, endpoints: Zh.Endpoint[]):
...
         }
 
-        const description = `Binary Input ${name} on endpoint ${endpoint.ID}`;
+        const name = "binary_input";
+
+        const description = `Binary Input ${label ?? name} on endpoint ${endpoint.ID}`;
         const args: m.BinaryArgs<"genBinaryInput"> = {
-            name: name.replace(/\s+/g, "_").toLowerCase(),
+            name: name,
             label,
```
Koenkk, line 399: I saw you also mentioned it in your PRs description (just realising that now), but would using `label` not be a cleaner option here? It's already assigned to the `name` ([example](https://github.com/Koenkk/zigbee2mqtt/blob/master/lib/extension/homeassistant.ts#L671))

Koenkk:
I think the only breaking change now is that the converters key changed, so e.g. if in the past you used `{"my_description": value}` for directly controlling it via MQTT, it will now become `{"binary_output": value}`. If the user was controlling only via HA I expect no breaking change as existing sensors are updated.

luar123:
will add more tests later. -> done

HA will discover new entities because the object id changes.

src/lib/generateDefinition.ts
luar123: name from label is removed if a device class is set: https://github.com/Koenkk/zigbee2mqtt/blob/6835cbf5a73116aa4bfa097f2ad0dfb972018997/lib/extension/homeassistant.ts#L1426

Could also make it a bool, e.g. `homeassistant.keepLabel`

Koenkk: I think we should call it `preserveName`, then we can also get rid of `!NUMERIC_DISCOVERY_LOOKUP[firstExpose.name]?.preserve_name`

src/lib/generateDefinition.ts
```diff
@@ -805,35 +810,35 @@ async function extenderAnalogOutput(device: Zh.Device, endpoints: Zh.Endpoint[])
...
 
-        const description = `Analog Output ${name} on endpoint ${endpoint.ID}`;
         const applicationType = await getClusterAttributeValue(endpoint, "genAnalogOutput", "applicationType", undefined);
         let unit: string | undefined;
+        let name: string | undefined;
 
         if (applicationType !== undefined) {
-            unit = getUnitfromApplicationType(applicationType);
+            name = APPTYPE_NAME[(applicationType >> 16) & 0xff];
+            unit = APPTYPE_UNIT[(applicationType >> 16) & 0xff];
```
Nerivec:
```suggestion
            const typeBits = (applicationType >> 16) & 0xff;
            name = APPTYPE_NAME[typeBits];
            unit = APPTYPE_UNIT[typeBits];
```

src/lib/generateDefinition.ts
```diff
@@ -384,21 +384,18 @@ async function extenderBinaryInput(device: Zh.Device, endpoints: Zh.Endpoint[]):
...
-            endpointName ? "binary_input" : `binary_input_${endpoint.ID}`,
-        );
-        let label: string | undefined;
+        const label = await getClusterAttributeValue(endpoint, "genBinaryInput", "description", undefined);
 
-        if (name !== "binary_input" && name !== `binary_input_${endpoint.ID}`) {
-            label = name;
+        let labelHomeAssistant: HomeAssistant | undefined;
+        if (label) {
+            labelHomeAssistant = {preserveName: true} as const satisfies HomeAssistant;
```
Nerivec, line 391: as const/satisfies not needed here, either it matches the type of the var or something's wrong.

test/generateDefinition.test.ts
```diff
@@ -643,42 +652,120 @@ export default {
...
+            ],
+            toZigbee: [
+                "binary_input",
+                "binary_input",
+                "binary_output",
+                "binary_output",
+                "analog_in_temperature",
+                "analog_input",
+                "analog_out_temperature",
+                "analog_output",
```
Nerivec, line 710: Something was bugging me with the new logic, I think this is where it breaks (dupe names).
@Koenkk can you confirm?

Koenkk: I think we should indeed not have dupe names (it shouldn't really hurt, it will just pick the first one, but cleaner to not have them)

luar123: It seems to work fine with a real device (tested with two binary inputs and two binary outputs).

I guess a proper fix would be to extend m.binary/m.numeric/exposes to take lists of label/name/unit/etc so we have to call them only once like it is done for the other clusters. Would suggest to do this in a separate PR.

Or add the endpoint suffix to name. But this would lead to exposes with endpoint suffix added twice.

Nerivec:
Sidenote: once ready, I suggest we merge this after August release. _If only to avoid a "somewhat breaking change" on the month when lots of people are away._ :sweat_smile: 
Drawback is, we probably won't get a lot of testers either, but it will at least be a full month for possible follow-up fixes...

Koenkk:
Agree with @Nerivec , PR looks good now, will merge after the 1 August release!

luar123:
@Koenkk Could you merge this soon so we have some time before the next release?
Or is anything still missing?

Nerivec:
Remains the tz/fz duplications in the final converter, but Koenkk will know better :wink:

luar123:
I understood that this is a minor issue that could be adressed in a follow up PR.

Just let me know how to proceed.

Koenkk:
Let's do it immediately, plenty of time before the next release.

luar123:
Resolved it for binary.
Please let me know if this approach is fine, then I will apply it to analog as well.

I kept all the `string` type parameters in addition to the `string[]` parameters for compatibility. It might be worth to convert the code base and deprecate the old parameters (in a separate PR).

src/lib/modernExtend.ts
```diff
@@ -2885,33 +2888,113 @@ export interface BinaryArgs<Cl extends string | number, Custom extends TCustomCl
     name: string;
     valueOn: [string | boolean, unknown];
     valueOff: [string | boolean, unknown];
-    description: string;
+    description?: string;
```
```suggestion
    description: string | string[];
```

Same for other, I think that's simpler and prevents `description` not being set.

luar123: Yes, thought about that too. Just found the naming a bit confusing.

Should I also revert endpointNames? All but binary, enumLookup and text are using it already.

Koenkk: Yeah `endpointName: string | string[]` is also better

luar123:
Reverted additional parameters, applied changes to `numeric` and changed `homeassistant.preserveName` to `homeassistant.name = "preserve"` as discussed in Koenkk/zigbee2mqtt/pull/32752. Tested and aligned Koenkk/zigbee2mqtt/pull/32654.

I am not so happy with the `null` value in `homeassistant.name` though. Should we change it to `"remove"`, `"device"`, or `"mainFeature"`? I think this would better explain what it is and would be less error-prone for equality testing.

Koenkk:
> I am not so happy with the null value in homeassistant.name though. Should we change it to "remove", "device", or "mainFeature"? I think this would better explain what it is and would be less error-prone for equality testing.

I don't like using e.g. `remove` there, because what if you want the `name` to be `remove?

luar123:
> I don't like using e.g. `remove` there, because what if you want the `name` to be `remove`?

Then you can set expose.withLabel("remove").
I changed `homeassistant.name` to enum only allowing `null` and `"preserve"` for now, because I understood your comments that it should not duplicate the exposes `label`. Please let me know if that is not what you intended.

luar123:
@Koenkk Any chance we could move this forward?

> I don't like using e.g. remove there, because what if you want the name to be remove?

We can leave it as is (`null`) if you prefer.

Koenkk:
I think the best logic is:

- `homeassistant.name: null`: delete
- `homeassistant.name: "some-string"` set this value
- `homeassistant.name: undefined`: do nothing

luar123:
> I think the best logic is:
> 
>     * `homeassistant.name: null`: delete
> 
>     * `homeassistant.name: "some-string"` set this value
> 
>     * `homeassistant.name: undefined`: do nothing

changed it and aligned the z2m PR Koenkk/zigbee2mqtt/pull/32654

Koenkk (approved):
LGTM! Will merge after 1 September release

Koenkk:
Thanks!

PR #12724, opened 2026-07-19
fix(tuya): add _TZ3000_46vasa5h to TS011F_2_gang_wall fingerprint

arpena:
## Summary

`_TZ3000_46vasa5h` (Rylike RY-WS02Z, 2 gang socket outlet AU) was listed in the `whiteLabel` section of `TS011F_2_gang_wall` but **missing from the `fingerprint` array**. This causes the `TS011F_plug_1` catch-all fingerprint (which matches on `applicationVersion`) to claim the device first, resulting in:

- Only 1 of 2 endpoints exposed
- Power monitoring entities shown despite the device having no power monitoring hardware
- Endpoint 2 (second socket) inaccessible

## Fix

Add `_TZ3000_46vasa5h` to the `fingerprint` array of `TS011F_2_gang_wall`, matching what was already declared in its `whiteLabel`.

## Device info

- Model ID: TS011F
- Manufacturer: _TZ3000_46vasa5h
- Application version: 67
- Endpoints: 1 (genOnOff), 2 (genOnOff), 242
- No power monitoring (haElectricalMeasurement bind fails with INVALID_EP)

## Related issues

- #26559 (same class of problem: TS011F device only exposing 1 of 2 endpoints)
- Koenkk/zigbee2mqtt#29896 (TS011F fingerprint priority issues)
- Koenkk/zigbee2mqtt#13190 (device incorrectly identified)


Koenkk:
Thanks!

PR #12729, opened 2026-07-20
fix(NodOn): fix wired input, endpoint 242 and identify for SIN-4-1-20/SIN-4-1-21/SIN-4-2-20

nthiers:
Three issues on `SIN-4-1-20`, `SIN-4-1-21` and `SIN-4-2-20`:

1. Wired input doesn't control the relay by default. It sends a `genOnOff` `Toggle` command to whatever the endpoint is bound to — with no bind at all, it only reaches the coordinator and does nothing. Adding this self-bind also makes it show up in the Bind tab, so users can remove it or rebind the wired input to control a different device instead.
2. Endpoint 242 (Green Power proxy) not mapped, same issue already fixed on `FPS-4-1-00`/`SIN-4-FP-20`/`SIN-4-FP-21`.
3. `m.identify()` missing on all three, unlike every other NodOn definition here.

### DIFF
--- a/src/devices/nodon.ts
+++ b/src/devices/nodon.ts
@@ -351,7 +351,12 @@ export const definitions: DefinitionWithExtend[] = [
         model: "SIN-4-1-20",
         vendor: "NodOn",
         description: "Multifunction relay switch",
-        extend: [m.onOff(), nodonModernExtend.impulseMode(), nodonModernExtend.switchTypeOnOff()],
+        endpoint: (device) => ({default: 1}),
+        extend: [m.identify(), m.onOff(), nodonModernExtend.impulseMode(), nodonModernExtend.switchTypeOnOff()],
+        configure: async (device) => {
+            const endpoint = device.getEndpoint(1);
+            await endpoint.bind("genOnOff", endpoint);
+        },
         ota: true,
     },
     {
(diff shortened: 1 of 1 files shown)


src/devices/nodon.ts
```diff
@@ -381,11 +392,18 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "NodOn",
         description: "Lighting relay switch",
         extend: [
-            m.deviceEndpoints({endpoints: {l1: 1, l2: 2}}),
+            m.identify(),
+            m.deviceEndpoints({endpoints: {default: 1, l1: 1, l2: 2}}),
             m.onOff({endpointNames: ["l1", "l2"]}),
```
Koenkk, line 397:
```suggestion
            m.onOff({endpointNames: ["l1", "l2"], configureReporting: true}),
```

Use this instead of the `configure`, same goes for other changed definitions.

nthiers: That's already the default (`configureReporting: true`), so it wouldn't change anything here and more importantly it binds to the coordinator, not to itself. 
The `configure` in this PR isn't about reporting, it's a **self-bind** (`genOnOff` bound to its own endpoint) so the wired input can control the local relay, as described in point 1 of the PR.

I don't think there's a way to express a self-bind through `m.onOff()`'s options currently.

Koenkk:
Thanks!

zaxxon72:
This might have caused a regression : https://github.com/Koenkk/zigbee2mqtt/issues/32802
could you please be so kind to check?

PR #12732, opened 2026-07-20
Shelly Gen4: stop reading through the RPC cluster

krobipd:
The RPC cluster cannot answer a read over Zigbee. The device acknowledges a read of the `Data` attribute and then sends no `Read Attributes Response`, so everything read through that cluster stays empty — Wi-Fi state, switch mode and the cover slat detection alike.

Shelly confirmed this as a known firmware limitation ([#10276](https://github.com/Koenkk/zigbee-herdsman-converters/pull/10276#issuecomment-3467285355)). It is not fixed in firmware 2.0.0 either: that changelog lists its Zigbee changes in full — OTA, inputs, connection LED, `Zigbee.SendCommand`/`ReadAttr`/`WriteAttr`, a tracked-objects setting, and five unrelated fixes — and none of them concerns response transmission.

Verified across three Gen4 models here (Presence, Power Strip, 2PM): every RPC-read-backed value stays empty, regardless of response size — a 45-byte `Switch.GetConfig` fails just like a 500-byte `Presence.GetConfig`.

### Why leaving the attempts in place is not free

A consuming integration repeats a read for every key and every endpoint of a device, so a single device query waits minutes for answers that never come. On a Presence Gen4, with ten endpoints, it stopped making progress altogether.

Wi-Fi state suffered twice over: the dedicated setup cluster, which *does* answer, only got its turn after two RPC timeouts, so its values went stale.

### What this changes

All RPC reads now sit behind `SHELLY_RPC_CAN_READ`, in the same spirit as the existing `shellyRPCBugFixed` flag. The code implementing the documented protocol stays and returns by flipping one constant once a firmware answers again.

What remains active is what the device really serves: the dedicated Wi-Fi setup cluster, which reports the same values as plain ZCL attributes and completes the truncated SSID from the cached one.

Measured on a Presence Gen4: the device query went from never completing to 2 min 24 s, and `wifi_status` / `ip_address` are current again.

### Removed tests

Eight tests covering the RPC read paths mocked a device that answers, which no firmware currently does — they were green against code that could not work. They are removed together with the reads they covered and belong back once the flag can be flipped.

### Testing

Full suite green. The remaining Wi-Fi test asserts that the state is read through the setup cluster without trying RPC first, and that a get does not fail when neither readback path is available.


jactmp:
Checked against Shelly 1PM Gen4 Mini with 2.0.0 (Wi-Fi activation, deactivation, etc.). Works as expected now. The only problem I noticed (nothing to do with this PR ofc)- device timeouts or responds with significant delay after power-on/reboot. 1-2 minutes later configuration/state changes and properties read requests are handled immediately. Seems like CPU is too slow, and firmware is too complicated (too many features / initial tasks)?

Koenkk:
thanks!

PR #12733, opened 2026-07-20
Shelly Presence Gen4 (S4SN-0U61X): full sensor and zone configuration

krobipd:
> Builds on #12732 and contains its commit, because the configuration here depends on the `SHELLY_RPC_CAN_READ` flag introduced there. Once #12732 is merged this PR reduces to its own second commit. Reviewing the second commit alone is enough.

The Shelly Presence Gen4 arrived with occupancy, light level and zone discovery. Everything that decides how it actually detects — mounting, range, sensitivity and the fine-tuning behind it — was reachable only through the Shelly app. And the little configuration that was exposed did not work.

### The per-zone thresholds were written under names the device does not have

A zone reports `presence_thr` and `absence_thr`. The API documentation calls them `presence_delay` and `absence_delay`, and that is what the converter sent. `PresenceZone.SetConfig` discards a key it does not know **and still answers with success**, so every write was silently dropped — nothing in the logs pointed at it.

The exposes keep the documented names, which say what the value means; the wire uses what the zone actually carries.

### Everything configured here can be written, never read

This is the central limitation of the device and it deserves to be stated plainly rather than discovered.

The sensor accepts a setting and applies it, but it cannot report any of them back over Zigbee (#12732). So a configuration value shown by an integration is **the value last written from there** — not a reading off the device. Change something in the Shelly app and nothing here notices. On a device that has never been written to from Zigbee, these values are simply unset.

Because a configuration field that quietly means something other than what the user assumes is worse than no field at all, **every one of these exposes says so in its own description**, and the same sentence appears on the per-zone delays and on eco mode.

Two consequences follow:

- **The reads are removed, not just marked unavailable.** For the configuration converters there is no `convertGet` at all. Access flags alone would not do: a consuming integration walks every converter that *has* a `convertGet` and calls it regardless of the flags, so a converter that merely declares itself unreadable still pays a full timeout per key and per endpoint.
- **Zone ids come only from an RPC read**, so discovery cannot run. A device leaves the factory with a single zone and the delays now address it by its main zone id; discovered ids still take precedence where they exist. For the same reason a device with no known count exposes the main zone instead of all ten — otherwise nine zones that never report are created and vanish again once a count is known.

### What is exposed now

`Presence.SetConfig` takes one nested config object, so instead of a converter per setting each one is described by its path inside that object and handled by a single shared converter:

- **Mounting** — installation height, sensor position, flipped
- **Detection** — sensitivity, radar power, minimum and maximum range, tracked objects
- **LEDs** — brightness, night mode and its brightness
- **Fine-tuning** — detection points, velocity, SNR, maximum velocity difference and the five state thresholds

Every range and value list is the one the device enforces itself, read off its own rejection messages rather than taken from the documentation.

`sensitivity` needs care: the device accepts only `low`, `medium` and `high`, but reports `custom` as soon as the fine-tuning values differ from a preset. The enum lists `custom` so the reported state stays a value it knows, and writing it is refused with a message pointing at the individual thresholds.

### Occupancy is per tracked person, not per zone

The occupancy endpoints are the people the sensor tracks, one endpoint each. Reporting was not configured at all before, so occupancy only ever answered a read; it is now set up on all ten endpoints, and presence arrives on its own. One endpoint failing no longer aborts the rest, so an unreachable endpoint costs its own reporting rather than everyone's.

The two numberings are independent: `presence_delay_1` addresses the first zone, `occupancy_1` the first tracked person.

### Testing

- Full suite green. The write-only property is covered directly: one test walks every configuration key and asserts that none of their converters has a `convertGet` and none of their exposes announces GET; others assert that a write sends no `GetConfig` afterwards and that configure makes no RPC attempt at all while the firmware cannot answer.
- Payload shapes verified against the hardware over its HTTP RPC, which carries the same calls: the real converter builds each command, the leaf value is replaced with an invalid one, and the device must reject it naming the exact field path. All configuration fields confirmed this way without writing anything.
- Writes verified end to end over Zigbee on a Presence Gen4 (firmware 2.0.0-beta3): zone threshold, top-level and `sensor.state` paths all reach the device, and a partial write leaves its sibling fields untouched. Each value was restored afterwards and the restore verified against a snapshot taken beforehand.
- Confirmed in daily use on that device by its owner: detection and the configured delays behave as expected.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -1,5 +1,6 @@
 import {Zcl, ZSpec} from "zigbee-herdsman";
 import * as fz from "../converters/fromZigbee";
+import type {Feature} from "../lib/exposes";
 import * as exposes from "../lib/exposes";
 import {logger} from "../lib/logger";
 import * as m from "../lib/modernExtend";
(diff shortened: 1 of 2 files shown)


Koenkk:

Koenkk:
Thanks!

PR #12735, opened 2026-07-20
Shelly Gen4: stop offering settings the device cannot report

krobipd:
Two Gen4 devices show configuration that can never hold a value, in different ways.

### Power strip: settings that announce a read nobody can serve

The LED and button settings go out through the RPC cluster, which cannot answer a read over Zigbee ([#12732](https://github.com/Koenkk/zigbee-herdsman-converters/pull/12732), [#10276](https://github.com/Koenkk/zigbee-herdsman-converters/pull/10276)). Three of them were declared `ea.ALL` and so offered a refresh — but none of these converters has a `convertGet` at all, and none would gain one if the firmware were fixed, because there is nothing to re-enable. They are settable now, not readable.

The author of the original power strip PR ([#10659](https://github.com/Koenkk/zigbee-herdsman-converters/pull/10659)) put it plainly at the time: *"It won't return anything useful on read, so this pull request only implements writes."* The access flags did not follow.

### And none of them said what that means for the value on display

What a configuration value shows is **the value last written from here** — never a reading off the device. Change something in the Shelly app and nothing here notices. On a device that has never been written to from Zigbee, these values are simply unset.

Every one of them now says so in its own description, because a configuration field that quietly means something other than what the user assumes is worse than no field at all. `led_colors` and `buttons_enabled` also got the description they were missing entirely — until now the first showed up under the bare key `led_colors`.

### Single-channel Gen4: a switch input that is not there

`S4SW-001X8EU`, `S4SW-001X16EU`, `S4SW-001P8EU` and `S4SW-001P16EU` exposed `switch_type` unconditionally. The switch input endpoint, however, only exists when an input is actually wired: on a device without one, `configure` finds no endpoint, skips the bind and the read, and the state stays empty forever.

It is now exposed the way the 2PM already does it — through `shellySwitchInputExposes`, which filters on the endpoints the device actually reports. Devices with a wired input are unaffected.

### Testing

Full suite green. The new tests use the endpoint layouts real devices report rather than invented ones, since that is exactly what the change turns on: a power strip on `[1,2,3,4,239,242]`, and a 1PM Mini both with and without its input endpoint. They assert that the three settings announce no GET and have no `convertGet`, that every power strip setting says it cannot be read back, and that `switch_type` appears only when the endpoint exists.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -623,14 +623,19 @@ const shellyModernExtend = {
             e.text("rpc_rxctl", ea.STATE_GET).withLabel("RxCtl").withDescription("RX bytes available").withCategory("diagnostic"),
             e.text("rpc_data", ea.STATE_GET).withLabel("Data").withDescription("RX Data").withCategory("diagnostic"),
         ];
+        // The device applies these but cannot report any of them back over Zigbee, so what a value
+        // shows is what was last written from here - never a reading off the device. Say so on every
+        // one of them: a configuration field that quietly means something else than the user assumes
+        // is worse than no field at all.
+        const WRITE_ONLY = "The device cannot report this back over Zigbee, so it shows what was last set from here";
         const exposesPowerstripUI: Expose[] = [
             e
                 .enum("led_mode", ea.STATE_SET, ["off", "switch", "power"])
                 .withLabel("LED Mode")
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks!

PR #12736, opened 2026-07-20
Shelly 2PM Gen4: switch input mode is mapped to the wrong endpoints in switch mode

krobipd:
The two 2PM Gen4 variants place their switch inputs on different endpoints. The cover has them on 2 and 3; the switch-mode device on 3 and 4, because there endpoints 1 and 2 are its two relays.

Each definition already says exactly that, and `switch_type` follows it:

| | endpoint map | `switch_type` |
|---|---|---|
| `S4SW-002P16EU-COVER` | `{sw1: 2, sw2: 3}` | `shellySwitchInputExposes(device, {sw1: 2, sw2: 3})` |
| `S4SW-002P16EU-SWITCH` | `{l1: 1, l2: 2, sw1: 3, sw2: 4}` | `shellySwitchInputExposes(device, {sw1: 3, sw2: 4})` |

The switch input **mode** did not follow it. It sits in the shared `shellyRPCSetup` block, which had the cover's mapping hardcoded as `{sw1: 2, sw2: 3}` and applied it to both variants. On a switch-mode device that reads **endpoint 2 — the second relay — as the first switch input**.

### What it looks like on a device

On a switch-mode 2PM with no inputs wired, the device reports only `[1, 2, 239, 242]`. Endpoints 3 and 4 do not exist, so `switch_type` correctly stays away — while `switch_mode` appeared anyway, offered for an input the device does not have. Observed on such a device.

With both inputs wired the two settings end up describing the same physical input under names that point at different endpoints: `switch_type_sw1` addresses endpoint 3, `switch_mode_sw1` endpoint 2.

### The change

The mapping now comes from the variant instead of being shared, so each device uses the endpoints its own definition already declares. `2PMInputMode` becomes `2PMCoverInputMode` and `2PMSwitchInputMode`; both names are used once each, and the cover's behaviour is unchanged.

### Testing

Full suite green. The new tests build their devices from the endpoint layouts in the definitions' own fingerprints, and cover all three cases: a switch-mode 2PM without wired inputs offers no `switch_mode` at all, one with wired inputs keeps both, and the cover keeps its own endpoints. The existing cover test that reads `switch_mode_sw2` off endpoint 3 continues to pass unchanged.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -561,7 +561,14 @@ const shellyModernExtend = {
         const featureDev = features.includes("Dev");
         const featurePowerstripUI = features.includes("PowerstripUI");
         const featurePowerstripPowerOnBehavior = features.includes("PowerstripPowerOnBehavior");
-        const featureTwoPMInputMode = features.includes("2PMInputMode");
+        // The two 2PM variants put their switch inputs on different endpoints: the cover on 2 and 3,
+        // the switch-mode device on 3 and 4, because there 1 and 2 are its two relays. A single
+        // shared mapping cannot serve both - it would read the second relay as the first input.
+        const twoPMInputEndpoints = features.includes("2PMCoverInputMode")
+            ? {sw1: 2, sw2: 3}
+            : features.includes("2PMSwitchInputMode")
+              ? {sw1: 3, sw2: 4}
+              : undefined;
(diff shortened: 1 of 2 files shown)


Koenkk:

Koenkk:
Thanks!

PR #12708, opened 2026-07-17
fix: Allow schema and type override for Home Assistant infrared discovery type

jbouwh:
The PR allows to override the Home Assistant discovery type for infrared entities and set the schema.
This should allow to use `expose.withHomeAssistant()` in Zigbee2MQTT to override the type to `infrared` and the `schema`.

The initial PR only allows to override the type, this PR adds the `schema` attribute: https://github.com/Koenkk/zigbee2mqtt/pull/32549.

PR #12737, opened 2026-07-20
fix: improve infrared emitter description

jbouwh:
A description update for `ir_emitter` was missed in #12708

When publishing the value should reset, like it does for an `action`, I assume we should fix this in `Zigbee2MQTT`.


Koenkk:
thanks!

PR #12738, opened 2026-07-20
fix: Fix vendor casing to prevent duplicates

andrei-lazarov:
There were some duplicate entries in the vendor list, due to different casing:
Adeo - ADEO, Elko - ELKO etc.

Fixed those and added a test


Koenkk:
👌

PR #11547, opened 2026-02-16
Fix Shelly remotes binding to coordinator

andrei-lazarov:
So the coordinator binds were missing from my PR:
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/11409

- hopefully fixes https://github.com/Koenkk/zigbee2mqtt/issues/31069

PR #12747, opened 2026-07-22
fix: Shelly Gen4: harden switch and button input handling

krobipd:
## What users see today

- Changing `switch_type`/`switch_mode` on a single-channel Gen4 without a wired input crashes the publish with `Cannot read properties of undefined (reading 'ID')` — reported by three users in Koenkk/zigbee2mqtt#31951. The definitions name `sw1 -> endpoint 2` even when that endpoint does not exist, and the application resolves the name with a non-null assertion.
- On a cover-mode 2PM with wired inputs in detached mode, input 1 throws (`Key '2_commandOn' not found`) and input 2 is reported as `input_1_*`: the action converters used a fixed endpoint table (3/4), but the cover variant has its inputs on endpoints 2/3 — the same endpoint split #12736 fixed for `switch_type`/`switch_mode`, still present in the event converters.
- Devices paired before the bindings arrived in configure (#12011/#12059/#12096) never get their input action events, because the definitions carry no `version` — the application only configures new pairings. A user in Koenkk/zigbee2mqtt#30637 had to re-interview manually to make the clusters appear.

## Changes (one commit each)

1. **Single-channel endpoint maps are filtered** through the same helper the 2PM variants already use — absent input endpoints are no longer named, which removes the crash precondition.
2. **Input action events resolve their input number from the definition's endpoint map** instead of a fixed endpoint table, so both 2PM variants and the single-channel devices report the correct input; messages from endpoints that are no switch input are ignored instead of throwing.
3. **All button/scene converters gained the standard `hasAlreadyProcessedMessage` guard** the stock command converters use, so a retransmitted frame cannot trigger automations twice.
4. **`switch_mode` no longer announces a read it cannot deliver**: it is written through the RPC cluster, which the firmware cannot answer over Zigbee (see `SHELLY_RPC_CAN_READ`, #12732). The dormant convertGet/configure reads are gone and the expose says the shown value is the last one written — same declaration style as the other unreadable Gen4 settings (#12735).
5. **The six switch definitions bump `version` to `0.0.1`** so already paired devices are re-configured once and their bindings arrive — the rule the BLU remotes applied for their own binding fix (#11547) and the Presence Gen4 documents on its version field.

## Tests

- `pnpm run build && pnpm run check && pnpm test` green on every commit (pre-commit hook), 743 tests total.
- New coverage: cover-variant input events on endpoints 2/3, switch-variant on 3/4, non-input endpoints ignored, single-channel wired-input events unchanged, duplicate transaction published once, `switch_mode` exposes no GET and carries the write-only note, all six definitions carry the version bump.

Addresses Koenkk/zigbee2mqtt#31951, Koenkk/zigbee2mqtt#30637.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -1915,6 +1915,7 @@ const fzLocal = {
         cluster: "genOnOff",
         type: ["commandToggle"],
         convert: (model, msg, publish, options, meta) => {
+            if (utils.hasAlreadyProcessedMessage(msg, model)) return;
             const event = utils.getFromLookup(msg.endpoint.ID, {1: "single", 2: "double", 3: "triple"});
             return {action: event};
         },
@@ -1924,6 +1925,7 @@ const fzLocal = {
         cluster: "genScenes",
         type: ["commandRecall"],
         convert: (model, msg, publish, options, meta) => {
+            if (utils.hasAlreadyProcessedMessage(msg, model)) return;
             const event = utils.getFromLookup(`${msg.endpoint.ID}`, {"1": "single_long", "2": "double_long", "3": "triple_long"});
             return {action: event};
         },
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks!

PR #12748, opened 2026-07-22
fix: Shelly WS90: discard non-value markers and recover the rain rate after a counter reset

krobipd:
## What users see today

In Koenkk/zigbee2mqtt#31048 a WS90 shows a gust speed of **6553 m/s**: that is the ZCL non-value marker (`0xffff`, all bits set) the station reports when a reading is unavailable, scaled by 10 and published as a measurement. The markers also feed the calculated values (wind chill, heat stress, weather condition). And after the cumulative precipitation counter resets (battery change, restart), the rain rate sticks at 0 until the new counter overtakes the old value — for a cumulative rain counter that can take months.

## Changes (one commit each)

1. **Recover the rain rate after a counter reset**: a negative precipitation delta now rebases the stored history on the new counter value instead of leaving it pointing at the old one.
2. **Discard the non-value markers** (`0xffff` for the uint16 wind attributes, `0xff` for the uint8 UV index, `0xffffff` for the uint24 precipitation) in both places that scale raw values — the numeric converters and the calculated-value converters — so an unavailable reading contributes nothing.

3. **Persist the device meta at most once a minute**: the calculated-value converters called `device.save()` on every single report — twice on rain reports — and the station reports as often as every 10 seconds, so the whole device database was rewritten to disk continuously. The persisted histories only advance on the minute scale; a crash now loses at most a minute of history progress.

This does not touch the reported zero drops from the issue: a zero can be a real reading for these attributes, only the all-bits-set markers are provably invalid.

## Tests

- `pnpm run build && pnpm run check && pnpm test` green on every commit (pre-commit hook), 734 tests total.
- New coverage: rain rate recovers after a simulated counter reset; each marker contributes nothing while a real reading still scales as before.

Reference: Koenkk/zigbee2mqtt#31048.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -562,7 +562,13 @@ function calculateRainRate(meta: WS90Meta, precipitation: number | undefined): n
     const precipDelta = precipitation - history.value;
 
     if (timeDeltaMs < 60000) return null;
-    if (precipDelta < 0) return 0;
+    if (precipDelta < 0) {
+        // The cumulative counter was reset (battery change, restart). Rebase the history on the
+        // new counter - otherwise the delta stays negative and the rate reports 0 until the new
+        // counter overtakes the old value, which for a cumulative rain counter can take months.
+        meta.precipHistory = {value: precipitation, time: now};
+        return 0;
+    }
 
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks!

PR #12749, opened 2026-07-22
fix: Shelly Gen4: write only the given Wi-Fi fields

### LINKED ISSUE #12617: Shelly Gen4: Wi-Fi toggle writes empty credentials
Hi,

for Shelly Gen4 `wifi_config`, toggling only `enabled` currently also writes empty credentials when SSID/password are not present in the payload:

```ts
ssid: value.ssid || ""
password: value.password || ""
```

Then actionCode: 1 applies it.

This clears the stored Wi-Fi credentials when the user only wants to disable/enable Wi-Fi.

Could `enabled` be handled separately, so changing only the Wi-Fi enabled state writes only `enabled` and then the apply `actionCode`, without writing empty `ssid` / `password`?

Alternatively, could the converter only write ssid / password when they are explicitly provided?

Related: #11169

Thanks a lot in advance!

**Workaround for others finding this and do only want to use Zigbee Mode but need WiFi for Maintenance:**
<details>
<summary>Workaround Details</summary>
You can create a script inside the shelly WebUI and enable autostart. This enables WiFi at every power cycle for a set time. 

Image

**Code:**
```
// Keep Wi-Fi available for maintenance after boot, then turn it off.

//disable WIFI after WIFI_ON_MIN minutes.
let WIFI_ON_MIN = 15;

// Enable Wi-Fi station.
Shelly.call("WiFi.SetConfig", {config:{sta:{enable:true}}});

// Disable Wi-Fi station after the defined time.
Timer.set(WIFI_ON_MIN * 60 * 1000, false, function () {
  Shelly.call("WiFi.SetConfig", {config:{sta:{enable:false}}});
});
 ```

krobipd:
## What users see today

Toggling only Wi-Fi `enabled` over Zigbee **wipes the stored SSID and password** — reported and confirmed on the device in #12617. The `wifi_config` convertSet stages every attribute with empty-string fallbacks (`ssid: value.ssid || ""`, `password: value.password || ""`) before `actionCode: 1` applies the lot, so anything absent from the payload travels as its empty default. The same lines cut the other way too: a payload of only `{ssid, password}` silently writes `enabled: false`.

## Change

Stage only the fields actually given (the same only-given-values pattern the Presence sensor groups in this file already use):

- a payload of only `{enabled}` writes exactly `enabled` + apply — the stored credentials stay untouched
- a payload of only `{ssid, password}` no longer touches `enabled`
- `ssid`/`password` never travel empty — clearing credentials over Zigbee is exactly the accident this fixes
- the static network fields may still travel empty on purpose (clearing them when switching back to DHCP)
- an empty payload writes nothing, not even the apply
- the proven write grouping (small multi-attribute frames, password on its own) is kept
- optimistic state returns only the written fields; the password stays write-only

What no converter change can solve: re-enabling Wi-Fi still requires re-sending the password, because the setup cluster applies whatever is staged and caching the password client-side is the security problem already named in the issue. That limitation stays with the firmware.

## Tests

- `pnpm run build && pnpm run check && pnpm test` green (pre-commit hook), 735 tests.
- New coverage: enabled-only toggle writes no credentials, credentials-only set does not touch enabled, full set writes every group, empty payload writes nothing.

Fixes #12617

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -1649,46 +1649,49 @@ const shellyModernExtend = {
                     assertObject<KeyValue>(value);
                     const ep = determineEndpoint(entity, meta, "shellyWiFiSetupCluster");
 
-                    const attr1 = {
-                        enabled: value.enabled === true,
-                        ssid: value.ssid || "",
-                    };
-                    await ep.write("shellyWiFiSetupCluster", attr1, SHELLY_OPTIONS);
+                    // Stage only the fields actually given: every staged attribute is applied by
+                    // actionCode 1, so an absent field must not travel as its empty default - a
+                    // payload of only {enabled} used to wipe the stored credentials, and one of
+                    // only {ssid, password} silently wrote enabled=false (#12617). The ssid and
+                    // password never travel empty: clearing credentials over Zigbee is exactly
+                    // the accident this fixes. The static network fields may travel empty on
+                    // purpose (clearing them when switching back to DHCP).
+                    const staged: Partial<ShellyWiFiSetup["attributes"]> = {};
+                    if (typeof value.enabled === "boolean") staged.enabled = value.enabled;
+                    if (typeof value.ssid === "string" && value.ssid !== "") staged.ssid = value.ssid;
+                    if (typeof value.password === "string" && value.password !== "") staged.password = value.password;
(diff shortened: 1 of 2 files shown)


Koenkk:

Koenkk:
thanks!

PR #12750, opened 2026-07-22
fix: Shelly: do not serialize RPC transactions of different devices against each other

krobipd:
## What users see today

The Shelly RPC lock is a single module-global flag: a hanging transaction on one device — e.g. the write retry loop against an unreachable device (three attempts with 1.5 s pauses) — blocks RPC configuration writes to **every other Shelly on the network** for its whole duration. The multi-write transaction only needs to be atomic per device: the chunks travel to separate unicast targets with separate reassembly buffers.

## Changes

1. **`fix:` Key the RPC lock by device** (ieee address) so transactions of different devices no longer serialize against each other. Same-device atomicity is unchanged — the whole TxCtl→chunks sequence stays under one acquisition. A test proves device B completes while device A is holding its own lock.
2. **`refactor:` Zero-effect cleanup while here**, each verifiably inert today: the RPC diagnostics converter is only registered with the Dev feature (nothing else can produce `shellyRPCCluster` responses while the firmware answers no RPC read — see `SHELLY_RPC_CAN_READ` —, and it published undeclared `rpc_rxctl`/`rpc_data` state keys), a dead `published`-flag branch in the Wi-Fi refresh is gone, and the TRV alarm and `switchType` converters return nothing instead of an empty object — the TRV one used to emit an empty publish on every single thermostat report.

## Tests

- `pnpm run build && pnpm run check && pnpm test` green on both commits (pre-commit hook), 748 tests on this branch.
- New coverage: different devices do not serialize against each other; no RPC diagnostics converter is registered without the Dev feature.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -398,19 +398,21 @@ interface ShellyLightLevel {
     commandResponses: never;
 }
 
-let shellyRpcSending = false;
+// Since RPC messages require multiple writes to complete, requests to the SAME device must not
+// interleave. Different devices are independent though - a network with many Gen4 devices must
+// not queue every RPC transaction behind one global lock.
+const shellyRpcBusy = new Set<string>();
 
-const shellyRpcLock = async <T>(callback: () => Promise<T>): Promise<T> => {
-    // Since RPC messages require multiple writes to complete, we have to make sure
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks!

PR #12752, opened 2026-07-23
feat(NodOn): add support for SEM-4-3-20 three-phase energy meter

nthiers:
Adds support for the NodOn SEM-4-3-20, a DIN-rail energy meter reading three-phase power via
3 external CT clamps (Phase A/B/C).

Exposes voltage/current/active/reactive/apparent power and power factor per phase, AC frequency,
total energy consumed/produced, and per-phase energy consumed/produced. Tested on 2 physical
devices through Zigbee2MQTT 2.12.0.

A couple of firmware quirks explain the choices in the definition:

- The `seMetering` cluster has no native per-phase energy attribute — the ZCL spec doesn't define
  one. The firmware reports Phase A/B/C energy via `currentTier1/2/3SummDelivered/Received`
  instead, attributes ZHC's own `tariffs` option treats as day/night tariff tiers rather than
  phases. This definition keeps the `energy_tier_N`/`produced_energy_tier_N` property names (what
  `fz.metering` produces) but overrides the exposed label to reflect what they actually are on
  this device.
- `reactivePower`/`apparentPower`/`powerFactor` (all 3 phases) and the 3 `total*Power` attributes
  (0x0304/5/6, not parsed by `fz.electrical_measurement`) have no matching key in
  `MeterArgs`/`configureLookup`, so `configure` sets up their reporting via the exported
  `setupAttributes()` helper instead.
- Every exposed measurement is labeled with its phase, including the attributes ZHC's convention
  otherwise leaves unsuffixed for Phase A (`voltage`/`current`/`power`/`power_factor`/
  `power_reactive`/`power_apparent`) — deliberate on a dedicated 3-phase meter, where we'd rather
  make every value's phase explicit than have Phase A be implicit.

Verified locally: `tsc --noEmit`, `biome check --error-on-warnings`, and `vitest run` (full suite,
731/731 tests) all pass against `upstream/master`.

Link to picture pull request: Koenkk/zigbee2mqtt.io#5354

### DIFF
--- a/src/devices/nodon.ts
+++ b/src/devices/nodon.ts
@@ -8,8 +8,8 @@ import * as exposes from "../lib/exposes";
 import * as m from "../lib/modernExtend";
 import {nodonPilotWire} from "../lib/nodon";
 import * as reporting from "../lib/reporting";
-import type {DefinitionWithExtend, ModernExtend} from "../lib/types";
-import {isDummyDevice} from "../lib/utils";
+import type {DefinitionWithExtend, Fz, KeyValue, ModernExtend} from "../lib/types";
+import {isDummyDevice, postfixWithEndpointName} from "../lib/utils";
 
 const e = exposes.presets;
 const ea = exposes.access;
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #11783, opened 2026-03-23
Add solis_action converter for Yale digital lock

juslex:
There is still a discrepancy in the battery measurement, I couldn't solve it

Link to picture pull request: TODO

PR #12755, opened 2026-07-23
fix(yale): correct SOLIS01 operation event decoding and battery scale; drop state derivation from ymc_action

juslex:
Follow-up to #11783 (mine), after months of field use on a live SOLIS01. Three fixes, all validated against captured frames and a week of production monitoring.

## 1. SOLIS01 operation events were decoded with the wrong field layout

The raw frame the Solis emits is the standard **ZCL operation event notification (0x20)** that fails ZCL parsing and lands in the `raw` handler. Its layout follows the spec:

| byte | meaning |
|------|---------|
| `data[3]` | operation event **source** (0 keypad / 1 rf / 2 manual / 3 rfid / 4 indeterminate) |
| `data[4]` | operation event **code** (same table as `fz.lock_operation_event`) |
| `data[5..6]` | user id (uint16 LE) |

`ymc_action` decodes `data[3]` as the action code, which misreports events on this device. Live captures:

| frame | truth | published by ymc_action |
|-------|-------|------------------------|
| `[9,191,32,1,2,0,...]` | rf **unlock** (remote command) | "unlock" (coincidence: source rf=1 maps to action 1) |
| `[9,7,32,2,1,0,...]` | manual **lock** | "auto_lock" ❌ |
| `[9,16,32,2,2,0,...]` | manual **unlock** (verified: `lockState:2` attributeReport same second) | "auto_lock" ❌ |

SOLIS01 now uses a dedicated `solis_action` converter with the correct layout; action names match `fz.lock_operation_event` so the default exposes fit. YMC420-W/YDM60 keep their existing table — no behavior change for them beyond item 2.

## 2. Revert state/lock_state derivation from operation events (added by #11783)

This caused **phantom unlocks**: the Solis emits operation events whose bolt never moved (observed repeatedly, e.g. after firmware reboots), and #11783's derivation flipped the HA lock to "unlocked" for hours while the door stayed physically locked (no `lockState` attributeReport ever arrived). Since the device reliably reports `lockState` via attributeReport within the same second of any real operation, deriving state from events is redundant and unsafe. This restores the pre-#11783 behavior for all `ymc_action` users.

## 3. Battery discrepancy from #11783 solved

SOLIS01 reports `batteryPercentageRemaining` on a **0-100 scale** (not the ZCL 0-200): a captured report of raw `58` was displayed as 29%. Fixed with the existing `lockExtend({battery: {dontDividePercentage: true}})` used by other Yale locks in this file. Heads-up documented in a comment: right after a reboot the firmware sends placeholder values (200, and sometimes 0) until it re-measures — readings settle within hours.

## Testing

- `pnpm run build` and `pnpm run check` clean; full test suite 752/752 passing.
- Equivalent converter running for a week as an external converter on z2m 2.12.0 with a production SOLIS01: keypad/rf/manual events decode correctly (incl. user id), zero phantom unlocks, battery tracking plausible values.

### DIFF
--- a/src/devices/yale.ts
+++ b/src/devices/yale.ts
@@ -527,40 +527,65 @@ const fzLocal = {
                 14: "manual_unlock",
                 15: "non_access_user_operational_event",
             };
+            // NOTE: do NOT derive state/lock_state from operation events here. These
+            // devices report lockState via attributeReport on real bolt movement, and
+            // operation events can arrive without the bolt moving (or with codes that
+            // differ per firmware generation), which caused phantom unlocks in HA.
+            return {
+                action: actionLookup[msg.data[3]] ?? "unknown",
+                action_source_user: msg.data[5],
+            };
+        },
+    } satisfies Fz.Converter<"closuresDoorLock", undefined, ["raw"]>,
+    solis_action: {
+        cluster: "closuresDoorLock",
+        type: ["raw"],
+        convert: (model, msg, publish, options, meta) => {
+            // The Solis (SOLIS01) sends the standard ZCL operation event notification
+            // (command 0x20) in a frame that fails ZCL parsing, so it lands here raw.
+            // Field layout follows the ZCL spec (validated against live frames):
+            //   data[3] = operation event source (0 keypad / 1 rf / 2 manual / 3 rfid / 4 indeterminate)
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12760, opened 2026-07-24
Fix zg-302zl auto on/off lookup table

nico1080:
auto on / off converter lookup value that were mixed (`all `and `ch1 `had same value):


nico1080:
auto on / off converter lookup value that were mixed (`all `and `ch1 `had same value):

Koenkk:
Thanks!

PR #12764, opened 2026-07-25
feat: Philips Hue plugs: expose Power-on behavior

andrei-lazarov:
Not sure why Power-on behavior was missing on these plugs..
It works on [Philips 929002240401](https://www.zigbee2mqtt.io/devices/929002240401.html#philips-929002240401). It also has the Philips cluster


Koenkk:
Thanks!

PR #12771, opened 2026-07-26
Clear last published timings for Tuya IR blaster and add override option for HA `value_template`

jbouwh:
Clear last published timings for Tuya IR blaster and add override option for HA `value_template`

This PR also fixes biome.json, which was not in sync


Koenkk:
thanks!

Issue #940, opened 2020-01-25
Reduce frequency of energy reporting


djchen:
My 45853GE and Securifi Peanut smart plugs report their energy related attributes much more frequently than I'd like it to and I believe much more frequently than they need to. The side effect is that this makes my Home Assistant database much larger since the values are coming in almost every second or every other second.

@Koenkk Do you think it's ok to raise the `minimumReportInterval` to something like 30 seconds or can we make this interval configurable option per device?
https://github.com/Koenkk/zigbee-herdsman-converters/blob/8af36f039ef4a3976e825fae4709fb35faef2fcf/devices.js#L125

Just want to get your thoughts here. I'm happy to submit a PR if the right approach is found.

Thanks!

Koenkk:
Change the min report interval to 5 and reportablechange to 10 or 100 (so that it will only send values once there is a significant change)

djchen:
> Change the min report interval to 5 and reportablechange to 10 or 100 (so that it will only send values once there is a significant change)

Assuming that works though, we wouldn't necessarily want to commit that change for all devices?

Koenkk:
No, but we can easily specify different values for this device through parameters.

airdrummingfool:
I've been thinking about this recently also. I have a few energy-reporting smart plugs that I use to determine if a device is on or off (based on power draw). A 5 second delay in status updates is fine, but having to wait 30 seconds to know that a device changed state is too long in my opinion.

On the flip side, I also have a few smart plugs that report energy usage that I totally ignore. It would be awesome to be able to configure each device separately, but short of that, it's probably worth setting different `reportableChange` values for each device. E.g. I have some Iris `3210-L` plugs that report current in milliamps. A change as small as 0.001 amps triggers a new message from the sensor, but the difference is so small it doesn't even change the value reported via MQTT (which only reports down to 0.01 amps).

A thought: instead of having to provide all parameters when we want to provide custom values to `endpoint.configureReporting()`, what if the `configureReporting` convenience methods accepted an optional second object parameter that was merged with the existing parameters, so we could e.g. override only `reportableChange`?

modified `configureReporting.rmsCurrent()`:
```
    rmsCurrent: async (endpoint, overrides) => {
        const payload = [{
            attribute: 'rmsCurrent',
            minimumReportInterval: 5,
            maximumReportInterval: repInterval.MINUTES_5,
            reportableChange: 1,
        }];
[...]

PR #12772, opened 2026-07-26
fix: LEDVANCE 4099854295232/4099854293276: silence periodic divisor-constant reports

krobipd:
The two LEDVANCE energy-meter plugs (`4099854295232` indoor, `4099854293276` compact outdoor) report the six divisor/multiplier constants of `haElectricalMeasurement` (`acPowerDivisor`, `acPowerMultiplier`, `acCurrentDivisor`, `acCurrentMultiplier`, `acVoltageDivisor`, `acVoltageMultiplier`) **every five minutes**. These values never change, and `electricityMeter()` already reads them during configure. Over a 14 h device-debug capture they accounted for roughly **three quarters** of these devices' frames.

Setting their maximum reporting interval to `0xFFFF` stops the periodic reports. Verified on hardware by writing it and reading the reporting table back: all six sit at 65535 afterwards and the five-minute cycle is gone, with the device still responding normally.

### Notes
- `silenceDivisorReporting` runs last in `extend` and swallows its own errors on purpose: silencing is an optimization, the measurement reporting is the actual fix — a `Delivery failed` while silencing must not take the configure run down with it. That is what the second test guards.
- Both definitions get `version: "0.0.1"` so already-paired devices reconfigure.
- Scope reduced to the divisor-silencing only, per review — the reporting-parameter change is dropped; the frontend-less reporting default will be taken up on the ioBroker side.

### DIFF
--- a/src/devices/ledvance.ts
+++ b/src/devices/ledvance.ts
@@ -1,6 +1,41 @@
 import {ledvanceLight, ledvanceOnOff} from "../lib/ledvance";
 import * as m from "../lib/modernExtend";
-import type {DefinitionWithExtend} from "../lib/types";
+import type {DefinitionWithExtend, ModernExtend} from "../lib/types";
+
+// The plugs report the six divisor/multiplier constants of the electrical measurement cluster
+// every 5 minutes - values that never change, and which `electricityMeter()` reads actively during
+// configure anyway. Silencing them removes about three quarters of these devices' radio traffic.
+//
+// Runs after electricityMeter (last entry in `extend`) and is guarded: silencing is an
+// optimization, the measurement reporting is the fix, so this must never fail the configure run.
+// Same reasoning as setupAttributes' own read loop, and it also covers variants that lack one of
+// the attributes.
+const silenceDivisorReporting: ModernExtend = {
+    configure: [
+        async (device, coordinatorEndpoint) => {
+            try {
+                await m.setupAttributes(
+                    device,
+                    coordinatorEndpoint,
+                    "haElectricalMeasurement",
+                    [
+                        {attribute: "acPowerDivisor", min: 1, max: 0xffff, change: 1},
+                        {attribute: "acPowerMultiplier", min: 1, max: 0xffff, change: 1},
+                        {attribute: "acCurrentDivisor", min: 1, max: 0xffff, change: 1},
+                        {attribute: "acCurrentMultiplier", min: 1, max: 0xffff, change: 1},
+                        {attribute: "acVoltageDivisor", min: 1, max: 0xffff, change: 1},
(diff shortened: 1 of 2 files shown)


src/devices/ledvance.ts
```diff
@@ -1,6 +1,51 @@
 import {ledvanceLight, ledvanceOnOff} from "../lib/ledvance";
 import * as m from "../lib/modernExtend";
-import type {DefinitionWithExtend} from "../lib/types";
+import type {DefinitionWithExtend, ModernExtend} from "../lib/types";
+
+// Reporting for the plugs with an energy meter. The devices accept and keep what
+// `electricityMeter()` writes, so the defaults (max 65000 s, change 5 W) are what makes them look
+// stuck: a load that stays within 5 W only ever hits the maximum interval, i.e. once every 18 h.
```
Koenkk: This heavily depends on the device that is connected to it. With the defaults we aim to give a good balance, but depending on the connected device it might not be good for your specific use-case. I think it's better that the user modifies it's reporting setting (through the z2m frontend) instead.

krobipd: Hi Koenkk, thanks for the review — and agreed, the defaults should stay a good general balance. 🙂

The catch: `zigbee-herdsman-converters` isn't only used by zigbee2mqtt. ioBroker.zigbee and others share the same definitions but have no reporting frontend — so for them the converter default *is* the final behaviour, and the only way to change it is a hand-written custom converter (this exact change, per device). "Adjust it in the frontend" only covers z2m users.

And here the default genuinely never fires: `activePower` has `reportableChange: 5 W` with an 18 h max — a plug running a 14.6–15.5 W fan never varies by 5 W, so its power only updated every 18 h (that's what z2m #27648 / #26377 report). This also follows your own note in #940 ("specify different values for this device through parameters") and the reporting-only precedent #5265 (Inovelli). Verified on hardware: a 1 W step reports in seconds, 0.9 W stays silent, idle stays silent.

If you'd rather leave the reporting values as-is, the divisor-silencing part is independent (~¾ of these devices' frames) and can land on its own.

andrei-lazarov: We must keep the reasonable 5W default. I suggest finding a solution with the ioBroker team. Please submit only the custom divisor reporting.

krobipd:
Reduced this to the divisor-silencing only, as suggested — the reporting-parameter change is dropped and `electricityMeter()` is back to its defaults. Now the only thing silenced is the six divisor/multiplier constants that report every 5 min (~¾ of these plugs' frames). Thanks for the review. For the frontend-less reporting default (ioBroker.zigbee & co. have no reporting UI) I'll follow up on the ioBroker side rather than here.

Koenkk:
Thanks!

PR #12777, opened 2026-07-27
fix(smarli): disable unintended optimistic write of cover position

loocd:
reported cover position jumps to target, then back to actual position and then moves as intended. This change removes the jump at the beginning.


Koenkk:
Thanks!

PR #12784, opened 2026-07-28
Add SAV-DL6IN-24V-12W

Aogu181:
Add SAV-DL6IN-24V-12W

Model: SAV-DL6IN-24V-12W
Vendor: Savolar
Description: Zigbee 12W Downlight RGB+CCT with external antenna design


Koenkk:
Thanks!

PR #12787, opened 2026-07-28
feat: add minimum PWM to Sunricher-based dimmers

niracler:
Adds the missing minimum brightness setting to nine already-supported Sunricher-based dimmers.


Koenkk:
Thanks!

PR #12795, opened 2026-07-29
fix: handle Somfy Glydea spurious zero reports using command context

darkl0rd:
## Summary

Fix the Somfy Glydea Ultra Zigbee module (`9028412A`) incorrectly publishing a fully-open position when movement stops.

The module can send an additional raw `currentPositionLiftPercentage: 0` report after a movement. The standard cover converter interprets that raw value as fully open, which can overwrite the actual final position.

## Important: previous implementation superseded

This PR **completely replaces and supersedes the earlier percentage-threshold implementation**.

The previous version attempted to identify a genuine fully-open endpoint by checking whether the preceding raw position was within an arbitrary percentage of zero.

That approach is no longer relevant and should not be used as the basis for reviewing this PR.

Further hardware testing showed that the threshold-based approach was not sufficiently robust because:

- the threshold was inherently arbitrary;
- it assumed that the motor reports predictable intermediate positions near the endpoint;
- different Glydea modules report intermediate positions differently;
- some modules report only a final position;
- a valid endpoint could be rejected when the expected preceding percentage was not reported;
- a spurious zero could be accepted when the preceding position happened to fall within the threshold.

The current implementation contains **no percentage threshold** and does not depend on a particular sequence of intermediate reports.

## New approach

The converter now tracks the exact command sent to each Glydea endpoint by Zigbee2MQTT.

The tracked command context is:

- `OPEN`: expected exposed position is `100`;
- `CLOSE`: expected exposed position is `0`;
- `STOP`: no new endpoint position is expected;
- exact position command: the requested exposed position is expected.

When the device reports raw position zero, the report is first processed by the standard `fz.cover_position_tilt` converter.

The result is then compared with the known command target:

- a zero report matching the requested endpoint is accepted;
- a zero report conflicting with the requested endpoint is discarded;
- a zero report following an explicit `STOP` command is discarded;
- a zero report matching an exact requested position is accepted.

Because the comparison uses the result of the standard converter, normal cover options such as `invert_cover` have already been applied.

## Conservative behaviour without command context

When no Zigbee2MQTT command context is available, the converter preserves the normal Zigbee2MQTT behaviour.

This is intentional.

For example, when a curtain is operated using a physical control or another Zigbee controller, the received raw-zero report may not contain enough information to distinguish a genuine fully-open endpoint from the module's faulty trailing zero.

The converter therefore does not guess in that situation.

Explicit `readResponse` messages are also treated as authoritative and are never filtered.

## Implementation

- Wrap `fz.cover_position_tilt` for Glydea-specific incoming-report handling.
- Wrap `tz.cover_state` to track `OPEN`, `CLOSE`, and `STOP`.
- Wrap `tz.cover_position_tilt` to track exact position targets.
- Store command context per endpoint using Zigbee2MQTT's global store.
- Continue using the standard converters for actual Zigbee commands and position conversion.
- Preserve `invert_cover` and other standard cover behaviour.
- Remove all percentage-based heuristics.

## Tests

Tests cover:

- filtering the spurious zero after closing;
- filtering the spurious zero after an explicit stop;
- filtering the spurious zero after an intermediate position target;
- accepting a genuine fully-open endpoint;
- preserving `invert_cover`;
- preserving normal behaviour without command context;
- never filtering an explicit position read.

## Hardware testing

Tested on multiple physical Somfy Glydea Ultra curtain motors.

Confirmed:

- an intermediate stop no longer produces an incorrect fully-open position;
- a fully closed curtain no longer reopens in Zigbee2MQTT state;
- genuine fully-open and fully-closed endpoints continue to work;
- `invert_cover` continues to work;
- differences in intermediate position-reporting behaviour do not affect the command-context logic.

Fixes Koenkk/zigbee2mqtt#31490.

### DIFF
--- a/src/devices/somfy.ts
+++ b/src/devices/somfy.ts
@@ -3,9 +3,237 @@ import * as tz from "../converters/toZigbee";
 import * as exposes from "../lib/exposes";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
-import type {DefinitionWithExtend} from "../lib/types";
+import * as globalStore from "../lib/store";
+import type {DefinitionWithExtend, Fz, KeyValueAny, Tz, Zh} from "../lib/types";
+import * as utils from "../lib/utils";
 
 const e = exposes.presets;
+
+type GlydeaTarget = number | "stop";
+
+type GlydeaState = {
+    /**
+     * Expected final position in Zigbee2MQTT coordinates:
+     * 0 = closed, 100 = open.
+     *
+     * "stop" means that an explicit stop command was sent.
+     * undefined means that no Z2M command context is available.
+     */
+    target?: GlydeaTarget;
+
+    /**
+     * True after the device reports windowCoveringType=0.
+     * This allows stale command context to be discarded when a later
+     * movement starts outside Zigbee2MQTT.
+     */
+    stopped: boolean;
+};
+
+const glydeaStateKey = "glydea_position_state";
+
+function getGlydeaState(endpoint: Zh.Endpoint): GlydeaState {
+    return globalStore.getValue(endpoint, glydeaStateKey, {
+        stopped: false,
+    });
+}
+
+function saveGlydeaState(endpoint: Zh.Endpoint, state: GlydeaState): void {
+    globalStore.putValue(endpoint, glydeaStateKey, state);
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #12798, opened 2026-07-29
fix(tuya): accept over_voltage_threshold for 110-127V grids on TS011F_with_threshold

### LINKED ISSUE #32689: TS011F_with_threshold: over_voltage_threshold min of 220 V rejects valid thresholds on 110-127 V grids (Tongou TO-Q-SY2-163JZT)
### What happened?

The `TS011F_with_threshold` definition (`src/devices/tuya.ts`) hardcodes the `over_voltage_threshold` expose to a 230 V-grid range:

```js
e.numeric("over_voltage_threshold", ea.STATE_SET)
    .withValueMin(220)
    .withValueMax(265)
```

This makes the entity unusable in 110–127 V countries (Brazil, US, Japan, Mexico, ...). My grid is 127 V nominal (the device itself reports `rmsVoltage: 129`), so the correct over-voltage trip point is around 135–140 V. The device firmware happily accepts and reports `over_voltage_threshold: 137` — it is only the converter definition that forbids it:

1. Home Assistant receives `min: 220, max: 265` via MQTT discovery and then rejects every state update, flooding the HA log continuously:

```
ERROR (MainThread) [homeassistant.components.mqtt.number] Invalid value for number.bedrooms_outlets_breaker_over_voltage_threshold: 137 (range 220.0 - 265.0)
```

2. Setting a threshold below 220 V from Z2M is rejected by the expose validation, even though the device supports it.

I have 7 of these breakers (Tongou TO-Q-SY2-163JZT, `_TZ3000_cayepv1a`), all showing the same behavior.

`over_voltage_threshold` should accept values valid for 100–127 V grids as well, e.g. `withValueMin(90)` (the sibling `under_voltage_threshold` in the same definition already goes down to 75 V, so there is precedent for low-voltage-grid support in this exact expose block).

1. Pair a TS011F_with_threshold breaker (e.g. Tongou TO-Q-SY2-163JZT / `_TZ3000_cayepv1a`) on a 110–127 V grid.
2. Set `over_voltage_threshold` to a value appropriate for that grid (e.g. 137 V) — the device accepts it and reports it back.
3. Z2M republishes the state; with the HA integration enabled, HA logs `Invalid value ... (range 220.0 - 265.0)` on every publish.
4. Attempting to set any value < 220 through Z2M/HA is rejected by the expose range.

### Zigbee2MQTT version

2.12.1 (zigbee-herdsman-converters 26.76.0)

### Adapter firmware version

EmberZNet 8.0.3 [GA] build 581

### Adapter

EmberZNet-based network coordinator (`adapter: ember`, TCP connection)

### Setup

Official container image on Kubernetes (k3s), x86-64. `Linux 7.0.0-27-generic #27-Ubuntu SMP PREEMPT_DYNAMIC x86_64` (Ubuntu 26.04)

```json
{"id":29,"type":"Router","ieeeAddr":"0x6c5cb1fffeea3589","nwkAddr":42231,"manufId":4098,"manufName":"_TZ3000_cayepv1a","powerSource":"Mains (single phase)","modelId":"TS011F","epList":[1,242],"endpoints":{"1":{"profId":260,"epId":1,"devId":266,"inClusterList":[0,3,4,5,6,1794,2820,1026,57344,57345],"outClusterList":[25,10],"clusters":{"genBasic":{"attributes":{"modelId":"TS011F","manufacturerName":"_TZ3000_cayepv1a","powerSource":1,"zclVersion":3,"appVersion":69,"hwVersion":1}},"haElectricalMeasurement":{"attributes":{"rmsCurrent":0,"activePower":0,"acCurrentDivisor":1000,"acCurrentMultiplier":1,"rmsVoltage":129}}}}}}
```

### Notes

Happy to test a fix — I have 7 of these devices on a live 127 V installation. If widening the static range is undesirable, an alternative would be deriving the range from a device option (like other grid-dependent converters do), but simply lowering `withValueMin` to ~90 V seems sufficient since the trip value is user-configured anyway.

galvesribeiro:
## Summary

The `over_voltage_threshold` expose on `TS011F_with_threshold` had `withValueMin(220)` — a 230V-grid assumption. On 100–127V grids (Brazil, US, Japan, Mexico, ...) the correct over-voltage trip point is around 135–140V, which the device firmware happily accepts and reports back (verified on 7× Tongou TO-Q-SY2-163JZT / `_TZ3000_cayepv1a` on a live 127V installation, device reporting `rmsVoltage: 129`). The expose range however:

- makes Zigbee2MQTT reject any set below 220V, and
- publishes `min: 220` via Home Assistant MQTT discovery, so HA continuously logs `Invalid value for number....over_voltage_threshold: 137 (range 220.0 - 265.0)` for the device-reported state.

This PR lowers the minimum to 90V. The sibling `under_voltage_threshold` in the same expose block already supports low-voltage grids (`withValueMin(75)`), so this brings the two in line. The trip value is user-configured, so the wider range carries no risk for 230V-grid users — their existing values remain valid.

Fixes Koenkk/zigbee2mqtt#32689

## Testing

What's verified on the live 127V installation: the firmware itself accepts a 137V threshold (set out-of-band) and reports it back, so only the expose range stands in the way. I have 7 of these breakers and can test a dev build with this change if helpful.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -16495,7 +16495,9 @@ export const definitions: DefinitionWithExtend[] = [
                 e.binary("over_current_breaker", ea.STATE_SET, "ON", "OFF").withDescription("Over-current breaker"),
                 e
                     .numeric("over_voltage_threshold", ea.STATE_SET)
-                    .withValueMin(220)
+                    // Min 90 (not 220) so thresholds for 100-127V grids are accepted too;
+                    // the device firmware itself takes values in this range.
+                    .withValueMin(90)
                     .withValueMax(265)
                     .withValueStep(1)
                     .withUnit("V")


Koenkk:
Thanks!

PR #12177, opened 2026-05-10
Update tuya.ts feat: add support for Nova Digital ZBCMR-02 curtain motor

marcosvtt:
## Added support for Nova Digital ZBCMR-02 curtain motor

This PR adds support for the Nova Digital ZBCMR-02 Zigbee curtain motor.

### Fingerprint

* Model ID: `TS0105`
* Manufacturer: `_TZE600_5cu64znk`

### Supported features

* Open
* Close
* Stop
* Position control
* Position state reporting
* Home Assistant cover entity support

### Notes

The device uses Tuya datapoints:

* DP 1: state control
* DP 2: target position
* DP 3: reported position

Battery reporting is currently unsupported. The device does not expose standard Zigbee battery clusters and no reliable Tuya datapoint has been identified yet.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/5110

PR #12800, opened 2026-07-29
Add support for Nova Digital ZBCMR-02 roller blind motor

bmnascimento:
Link to picture pull request:

There seems to be two different devices by Nova Digital with the same model name "ZBCMR-02".
There is already a picture available:
https://github.com/Koenkk/zigbee2mqtt.io/blob/master/public/images/devices/ZBCMR-02.png

But the model I have on hand is slightly different:
https://github.com/bmnascimento/zigbee2mqtt.io/tree/image-ZBCMR-02/public/images/devices/ZBCMR-02.png

Also, I found there is a PR already for the other version of the device:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/12177

I have tested the device I have on hand and it worked perfectly with the already existing model "TS0601_cover_3".

Which image should I use and what device name to use in the code considering the conflict? Thank you for your time.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -9718,6 +9718,7 @@ export const definitions: DefinitionWithExtend[] = [
             "_TZE200_68nvbi09",
             "_TZE200_vexa5o82",
             "_TZE200_sfqyhvpv",
+            "_TZE284_n73badib",
         ]),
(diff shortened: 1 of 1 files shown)


Koenkk:
thanks!

PR #12801, opened 2026-07-29
fix: use correct expose key for battery voltage on TS0601_water_valve and TS0601_water_meter

Liionboy:
## Description

The `e.battery_voltage()` expose creates a sensor with key `voltage`, but the `tuyaDatapoints` mapping for DP 26 was writing to `battery_voltage`. This mismatch causes the voltage sensor to always show as **Unknown** in the Zigbee2MQTT frontend and Home Assistant.

The raw `battery_voltage` field in the MQTT state payload does contain the correct value (e.g. `3.7`), but the UI sensor bound to `voltage` remains `null` because nothing writes to that key.

## Fix

Changed DP 26 mapping key from `battery_voltage` to `voltage` so it matches the expose definition:

```diff
- [26, "battery_voltage", tuya.valueConverter.divideBy100],
+ [26, "voltage", tuya.valueConverter.divideBy100],
```

## Affected devices

| Model | Description | Fingerprints |
|-------|-------------|-------------|
| TS0601_water_valve | Ultrasonic water meter valve | _TZE284_vuwtqx0t, _TZE200_vuwtqx0t |
| TS0601_water_meter | Ultrasonic water meter | _TZE200_h367laha |

## Verification

MQTT payload before fix:
```json
{"battery_voltage": 3.7, "voltage": null}
```

After fix, `voltage` will correctly report the value from DP 26.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2506,7 +2506,7 @@ export const definitions: DefinitionWithExtend[] = [
                     },
                 ],
                 [22, "temperature", tuya.valueConverter.divideBy100],
-                [26, "battery_voltage", tuya.valueConverter.divideBy100],
+                [26, "voltage", tuya.valueConverter.divideBy100],
             ],
         },
         // Optional: Add device-specific options
@@ -2829,7 +2829,7 @@ export const definitions: DefinitionWithExtend[] = [
                     },
                 ],
                 [22, "temperature", tuya.valueConverter.divideBy100],
-                [26, "battery_voltage", tuya.valueConverter.divideBy100],
+                [26, "voltage", tuya.valueConverter.divideBy100],
             ],
         },
         options: [
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2390,7 +2390,7 @@ export const definitions: DefinitionWithExtend[] = [
             e.numeric("flow_rate", ea.STATE).withUnit("L/h").withDescription("Instantaneous water flow rate"),
             e.binary("auto_clean", ea.STATE_SET, "ON", "OFF").withDescription("Auto clean"),
             e.temperature(),
-            e.numeric("battery_voltage", ea.STATE).withUnit("V").withDescription("Battery voltage").withCategory("diagnostic"),
+            e.battery_voltage(),
             e.text("meter_id", ea.STATE).withDescription("Meter identification number"),
             e.text("faults", ea.STATE).withDescription("Active fault status"),
             e.enum("report_period", ea.STATE_SET, ["1h", "2h", "3h", "4h", "6h", "8h", "12h", "24h"]).withDescription("Report period"),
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -2506,7 +2506,7 @@ export const definitions: DefinitionWithExtend[] = [
                     },
                 ],
                 [22, "temperature", tuya.valueConverter.divideBy100],
-                [26, "battery_voltage", tuya.valueConverter.divideBy100],
+                [26, "voltage", tuya.valueConverter.divideBy100],
```
Koenkk, line 2509: The expose is still ` e.battery_voltage(),`, I think this voltage represents the voltage of the battery right?

Liionboy: Yes

Liionboy: This is what device report:

`[7/30/2026, 9:40:34 PM] z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Apometru', payload '{"auto_clean":"OFF","battery_voltage":3.7,"daily_consumption":119,"device":{"applicationVersion":80,"dateCode":"","friendlyName":"Apometru","hardwareVersion":1,"ieeeAddr":"0xa4c1386da7456ecc","manufacturerID":4417,"manufacturerName":"_TZE284_vuwtqx0t","model":"TS0601_water_valve","networkAddress":42990,"powerSource":"Battery","stackVersion":0,"type":"EndDevice","zclVersion":3},"faults":"no_alarm","flow_rate":0,"last_seen":"2026-07-30T21:40:31+03:00","linkquality":177,"meter_id":"00000025133816","month_consumption":5038,"report_period":"1h","reverse_water_consumed":0,"state":"ON","temperature":20.6,"voltage":null,"water_consumed":30151}'
[7/30/2026, 9:40:34 PM] z2m: Successfully configured 'Apometru' (definition v0.0.0)`

Koenkk: Then the expose should be corrected instead, not expose `voltage`, but `battery_voltage`

Liionboy: i did a modification now with a new commit

Koenkk: Made a mistake, turns out changing it to `voltage` is correct (as `e.battery_voltage()` uses `voltage`)

Koenkk:
Thanks!

src/devices/tuya.ts
Liionboy: Now the problem is with Unit because in Z2M shhow 3.7 mV not V and in HA is same error , 3.7 mV , i will try to do a new PR to fix the problem but need to add a custom expose

Liionboy: is Safe to make  a new PR with this?

```
e.numeric("battery_voltage", ea.STATE)
    .withUnit("V")
    .withDescription("Battery voltage in volts")
    .withCategory("diagnostic"),
```

PR #12807, opened 2026-07-30
Add Beca BAC-001 thermostat (_TZE204_hpkusvom)

dansity:
Adds support for the Beca BAC-001 heating/cooling thermostat with fan control
  (`TS0601` / `_TZE204_hpkusvom`).

  Closes https://github.com/Koenkk/zigbee2mqtt/issues/31502

  Device picture PR: Koenkk/zigbee2mqtt.io/pull/5372

  ### Datapoints

  | DP | Exposed as | Notes |
  |----|------------|-------|
  | 1  | `system_mode` | power state, mapped to `off` |
  | 2  | `system_mode` | `cool` / `heat` / `fan_only` |
  | 16 | `current_heating_setpoint` | /10 |
  | 24 | `local_temperature` | /10 |
  | 27 | `local_temperature_calibration` | negatives are 32 bit two's complement |
  | 28 | `fan_mode` | `low` / `medium` / `high` / `auto` |

  The device reports its power state and its operating mode on two separate
  datapoints, and `off` is not one of the values of DP 2. Exposing these
  separately leaves Home Assistant without a working mode control on the climate
  entity, so both are combined into a single `system_mode` the same approach
  already used by the sibling BAC-002-ALZB in this file.

  `forceTimeUpdates` with `timeStart: "1970"` is required, without it the
  schedule/time presets on the device do not work.

  ### Tested

  Verified on hardware (app version 74): on/off via `system_mode`, switching
  between cool/heat/fan_only, setpoint, local temperature, calibration
  (including negative values) and fan speed.

  ### Not implemented

  The device has further settings whose datapoints have not been identified yet:
  screen brightness, 12h/24h clock, child lock, min/max allowed temperature,
  display modes and energy saving mode. These can be added later once the DPs
  are known.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -570,6 +570,11 @@ const bindSlotTS0601SmartSceneKnob = async (entity: Zh.Endpoint | Zh.Group, slot
     await tuya.sendDataPointRaw(entity as Zh.Endpoint, 102, payload, "dataRequest", 0x10 + (slot - 1));
 };
 
+// The BAC-001 reports its power state (dp 1) and its operating mode (dp 2) on separate datapoints,
+// combine both into a single `system_mode` in which the power state maps to `off`.
+const bac001SystemMode = (meta: Fz.Meta): ThermostatSystemMode =>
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12811, opened 2026-07-30
fix: Shelly S4SN-0071A/S4SN-0071Z: merge Flood and Flood S into one definition

krobipd:
Follow-up to #12796, as suggested by @andrei-lazarov in the review there.

The Flood (`S4SN-0071A`) and Flood S (`S4SN-0071Z`) are identical on the Zigbee side (same clusters, same `extend`), so this merges them into a single definition and expresses the Flood S as a whitelabel — dropping the duplicated block.

They report distinct modelIDs (`Flood` / `Flood S`), so both are listed in `fingerprint`; the whitelabel carries its own fingerprint, so the Flood S keeps its own model number, description and image (`S4SN-0071Z`).

Verified `findByDevice` resolves both:
- modelID `Flood` → `S4SN-0071A` (Flood Gen 4)
- modelID `Flood S` → `S4SN-0071Z` (Flood S Gen 4)

No image change — both device images already exist.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -2474,21 +2474,21 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
-        fingerprint: [{modelID: "Flood", manufacturerName: "Shelly"}],
+        fingerprint: [
+            {modelID: "Flood", manufacturerName: "Shelly"},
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

andrei-lazarov (commented):
Great thanks!

PR #12812, opened 2026-07-30
fix: Shelly 1/1 Mini, 1PM/1PM Mini: merge identical variants into whitelabels

krobipd:
Follow-up to #12811 (Flood/Flood S merge). Prompted by @andrei-lazarov's whitelabel suggestion there, I went through the rest of the Shelly definitions for the same pattern and found two more pairs that are identical on the Zigbee side:

- **1** (`S4SW-001X16EU`) and **1 Mini** (`S4SW-001X8EU`)
- **1PM** (`S4SW-001P16EU`) and **1PM Mini** (`S4SW-001P8EU`)

Within each pair the definitions are byte-identical in `fromZigbee` / `toZigbee` / `exposes` / `extend` / `configure` — the only differences are the model number and the matching key (the physical difference is form factor / rating, invisible to the converter). Both pre-existing definitions already carried the same `shellyRPCSetup(["1PMInputMode"])`, so the merge changes nothing there.

Each pair is merged into one definition, expressing the non-primary variant as a whitelabel so it keeps its own model number, description and image:
- `1` matches by `fingerprint` with `manufacturerName` (the modelID `1` is too generic for a bare `zigbeeModel`) — same as the Flood.
- `1PM` matches by `zigbeeModel`, so it joins the primary `zigbeeModel` list and the whitelabel relabels it by modelID — same as the existing BLU RC Button 4 definition.

Verified `findByDevice` resolves all four (1 Mini/Mini1 → `S4SW-001X8EU`, 1 → `S4SW-001X16EU`, 1PM Mini/Mini1PM → `S4SW-001P8EU`, 1PM → `S4SW-001P16EU`). No image change — all device images already exist.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -2099,9 +2099,18 @@ const tzLocal = {
 export const definitions: DefinitionWithExtend[] = [
     {
         zigbeeModel: ["Mini1", "1 Mini"],
+        fingerprint: [{modelID: "1", manufacturerName: "Shelly"}],
         model: "S4SW-001X8EU",
         vendor: "Shelly",
         description: "1 Mini Gen 4",
+        whiteLabel: [
+            {
+                vendor: "Shelly",
+                model: "S4SW-001X16EU",
+                description: "1 Gen 4",
+                fingerprint: [{modelID: "1", manufacturerName: "Shelly"}],
+            },
+        ],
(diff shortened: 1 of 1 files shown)


andrei-lazarov (approved):

Koenkk:
thanks!

PR #12816, opened 2026-07-31
fix: SONOFF valves: fix Home Assistant control

andrei-lazarov:
Valve type isn't implemented yet:
- https://github.com/Koenkk/zigbee2mqtt/pull/32367

Home Assistant sends open/close. But Zigbee2MQTT expects on/off.

> error:    z2m: Publish 'set' 'state' to 'sonoff' failed: 'Error: 'close' not allowed, choose between: toggle,off,on'

We must remove it, or merge the PR above. Otherwise the valves won't be usable in HA after the next Z2M release.

cc @MaxRink

Thanks Xploder on Discord for noticing


Koenkk:
good catch, thanks!

PR #12817, opened 2026-07-31
fix: Tuya TS011F_plug_1: Fix current scale

andrei-lazarov:
Tuya is unpredictable 🙂 This caused some issues @Nirostar:
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/12545

Thanks yousaf465 for reporting on Discord:

> Zbeacon TS011F v1.0.5 reports acCurrentDivisor 1, but it needs 1000.

The original PR only needs custom power and voltage. So I removed the custom current. Reverted to the default 1000.

- Probably related https://github.com/Koenkk/zigbee2mqtt/discussions/32711

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -11309,6 +11309,7 @@ export const definitions: DefinitionWithExtend[] = [
         model: "TS011F_plug_1",
         description: "Smart plug (with power monitoring)",
         vendor: "Tuya",
+        version: "0.0.1",
         whiteLabel: [
             {vendor: "LELLKI", model: "TS011F_plug"},
             {vendor: "BlitzWolf", model: "BW-SHP15"},
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -11415,8 +11415,9 @@ export const definitions: DefinitionWithExtend[] = [
                 "acVoltageDivisor",
                 "acPowerMultiplier",
                 "acPowerDivisor",
-                "acCurrentMultiplier",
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -11415,8 +11416,9 @@ export const definitions: DefinitionWithExtend[] = [
                 "acVoltageDivisor",
                 "acPowerMultiplier",
                 "acPowerDivisor",
-                "acCurrentMultiplier",
-                "acCurrentDivisor",
+                // "acCurrentMultiplier",
+                // "acCurrentDivisor",
```
Koenkk, line 11420: Are we sure that all devices will work with `1000` (except the `_TZ3000_typdpbpg`)? Because now the multiplier/divisor will not be read from the device anymore.

andrei-lazarov: Yes. That's how it's always been, and there were no issues (until the PR last month which added the readings)

Koenkk:
Thanks!

src/devices/tuya.ts
Nirostar: Interesting I thought we made it read skip if it couldn't read it using try catch?

andrei-lazarov: It reads successfully. But the specific device reports the wrong value.

yousaf465:
@andrei-lazarov  issue not fixed yet
image

image

yousaf465:
Oh, reconfiguring and restarting SMHUB fixed it. 
image

PR #12821, opened 2026-08-01
Rename `motor_run_status` to `moving` for SONOFF MINI-ZBRBS

AT0myks:
As mentioned in https://github.com/Koenkk/zigbee2mqtt/pull/32702#discussion_r3692999068.

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -9718,7 +9718,7 @@ export const definitions: DefinitionWithExtend[] = [
                 access: "STATE_GET",
             }),
             m.enumLookup<"customClusterEwelink", SonoffEwelink>({
-                name: "motor_run_status",
+                name: "moving",
                 lookup: {Stop: 0, Forward: 1, Reverse: 2},
                 cluster: "customClusterEwelink",
                 attribute: "motorRunStatus",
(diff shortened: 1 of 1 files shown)


src/devices/sonoff.ts
```diff
@@ -9718,7 +9718,7 @@ export const definitions: DefinitionWithExtend[] = [
                 access: "STATE_GET",
             }),
             m.enumLookup<"customClusterEwelink", SonoffEwelink>({
-                name: "motor_run_status",
+                name: "moving",
                 lookup: {Stop: 0, Forward: 1, Reverse: 2},
```
Koenkk:
```suggestion
                lookup: {stop: 0, forward: 1, reverse: 2},
```

AT0myks: Could you explain why this change?

AT0myks: `.toLowerCase()` [here](https://github.com/Koenkk/zigbee2mqtt/blob/fcbb7ff44bdc05a16e95d3472a81d110646cc17b/lib/extension/homeassistant.ts#L963-L965) takes care of this in case that's what you mean.

Koenkk: Because all exposes should be `snake_case`, due to the lower case it's not a breaking change so good catch.

Koenkk:
Thanks!

PR #12822, opened 2026-08-01
fix: match NUL padded modelID when several candidates exist

ptondereau:
`findDefinition` strips NUL padding and whitespace from `modelID` in both `getFromIndex` and `getFromExternalDefinitionsLookup`, but the final fallback compares `candidate.zigbeeModel` against the raw `device.modelID`. A device that pads its `modelID` gets found by the normalized lookup, then matches nothing, and falls back to a generated definition.

You only hit this with two or more candidates. A single candidate returns earlier through the `candidates.length === 1` short circuit, which never reaches the comparison. Installing an external converter for a model that already has a built-in definition is enough to get there.

I ran into it on the Atlantic Galapagos radiator from #12649. It reports `modelID` as `100050060900\0`. While my external converter was the only candidate it worked fine. Once the built-in definition shipped in v26.80.0 and both were loaded, the radiator dropped to a generated definition and lost its climate entity.

I pulled the existing normalization into a `normalizeModelID` helper and used it in the comparison too. Two tests come with it: one for a padded `modelID` with a single candidate, one where an external converter adds a second. The second fails on master.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -132,6 +132,14 @@ function removeFromExternalDefinitionsLookup(zigbeeModel: string | undefined, de
     }
 }
 
+/**
+ * Some devices pad `modelID` with NUL bytes or surrounding whitespace. Strip that so lookups
+ * and comparisons use the same form as the value declared in `zigbeeModel`.
+ */
+function normalizeModelID(modelID: string): string {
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #12833, opened 2026-08-03
Update TS0601 from _TZE204_lb0fsvba

dMopp:
See https://github.com/Koenkk/zigbee2mqtt/issues/32724#issuecomment-5159221474

Changes

1. DP6 (power/current/voltage): phaseVariant2 → phaseVariant4
Root cause of the stale values — the previous variant used the wrong byte offsets when parsing the payload, so power/current stopped updating correctly. phaseVariant4 decodes it properly.

2. DP9 (faults): single enum → circuitBreakerFaults (bitmap)
Previously mapped as one enum value; now decoded as a bitmap via tvc.circuitBreakerFaults, correctly exposing multiple simultaneous fault conditions as a list.

3. DP20: new clear_fault reset control
New — allows clearing stored faults via tvc.reset (enum, value CLEAR). Didn't exist in the previous definition at all.

4. DP127 (status): raw → circuitBreakerStatus
Now properly decoded into off / consumption / production, instead of an undecoded raw value.

5. New fzSigned converter
Runs after datapoint processing, reads power, current, and status from meta.state, and derives power_signed / current_signed (positive = consumption, negative = production) — useful for PV setups where import/export direction matters.

6. extend: added queryOnConfigure: true
Ensures values are queried on configure rather than only on device-initiated reports.

7. description: extended with "with bidirectional metering"

Known limitation

After a reconfigure, over_voltage_threshold reverts to the device default (275 V) instead of retaining the previously configured value. Documented but not fixed in this PR.

Verified
Tested on real hardware

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -2011,9 +2011,7 @@ export const definitions: DefinitionWithExtend[] = [
                 .withDescription("Current with sign: positive = consumption, negative = production"),
             te
                 .leakageCurrent()
-                .withDescription(
-                    "Current measured by the external ring. Place it over BOTH live and neutral wires to detect leakage current",
-                ),
+                .withDescription("Current measured by the external ring. Place it over BOTH live and neutral wires to detect leakage current"),
             e.device_temperature(),
             te.circuitBreakerFaults(),
             exposes.enum("clear_fault", ea.STATE_SET, ["CLEAR"]).withDescription("Clear the stored faults"),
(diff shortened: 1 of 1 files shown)


src/devices/tuya.ts
```diff
@@ -1918,6 +1918,26 @@ const fzLocal = {
     } satisfies Fz.Converter<"lightingColorCtrl", undefined, "raw">,
 };
 
+const fzSigned = {
```
Koenkk, line 1921: Could you keep this as a valueConverter?

dMopp: I wish I could. I am not even sure I understand the request completely. Let me try that out later today or tomorrow.

PR #12835, opened 2026-08-04
CK-TLSR8656-SS5-01(7037)

shuang-nok:
add model for for CK-TLSR8656-SS5-01(7037)


CubeZ2mDeveloper:
Hi @shuang-nok ,

Thanks for the contribution.

Since the vendor of this device is eWelink, could you please move this definition to `ewelink.ts` instead?

Keeping eWelink devices together in the vendor-specific file would make the device definitions easier to maintain and align with the existing structure.

Thanks!

Koenkk:
Thanks!

PR #12838, opened 2026-08-04
Fix AOYAN switch module definitions

zyjsmile857:
Fixes AOYAN switch module definitions.\n\nChanges:\n- Add manufacturerName fallback to the AY301Z whiteLabel fingerprint so devices still show as AOYAN when the reported modelID contains hidden padding\n- Add deviceEndpoints mapping for AY-602ZL/AY-802ZL\n- Add deviceEndpoints mapping for AY-603ZL/AY-803ZL


Koenkk:
Thanks!

PR #12846, opened 2026-08-05
fix: NodOn SEM-4-1-00: bump definition version to trigger reconfigure

nthiers:
Devices paired before #12548 (merged) don't automatically pick up the new, much lighter `apparentPower`/`powerFactor` reporting config — this definition never had a `version` field, so Z2M's reconfigure-on-definition-change mechanism (#11148) never fires for it.


Koenkk:
Thanks!

PR #12849, opened 2026-08-05
Fix: Color Temperatur Range for 929004581801

jonas0b1011001:
The other Festavia Globe Variants already have the correct range.


Koenkk:
Thanks!

PR #12851, opened 2026-08-06
Fix formatting and update Frient device details

ChillZwix:
Was not working, fixed it now and tested it, ok.


src/devices/frient.ts
```diff
@@ -54,12 +54,12 @@ export const definitions: DefinitionWithExtend[] = [
...
         description: "Customizable siren",
         extend: [
-            m.deviceEndpoints({endpoints: {"43": 43}}),
-            m.onOff({powerOnBehavior: false}),
+            m.onOff({powerOnBehavior: false, configureReporting: false}),
             m.battery(),
             m.iasZoneAlarm({zoneType: "alarm", zoneAttributes: ["alarm_1", "alarm_2", "tamper", "battery_low"]}),
             m.iasWarning(),
         ],
+        endpoint: (device) => ({default: 43}),
```
Koenkk: please put this in the m.deviceEndpoints instead

src/devices/frient.ts
```diff
@@ -54,12 +54,12 @@ export const definitions: DefinitionWithExtend[] = [
         vendor: "Frient",
         description: "Customizable siren",
         extend: [
-            m.deviceEndpoints({endpoints: {"43": 43}}),
-            m.onOff({powerOnBehavior: false}),
+            m.onOff({powerOnBehavior: false, configureReporting: false}),
```
Koenkk: why is reporting disabled?

ChillZwix: With configureReporting enabled, Z2M tries to bind and configure reporting on genOnOff, which the device doesn't answer. On (re)join this fails all 3 attempts and blocks the whole configuration step:
Failed to configure '0x0015bc0041008e7f', attempt 1/2/3 (Error: ZCL command .../1 genOnOff.configReport(...)
The siren's on/off state is transient and doesn't need to be a reported value, so configureReporting: false avoids the failure and lets battery/IAS configuration succeed.

ChillZwix:
I also removed the `m.onOff()` extend from the SIRZB-112 definition. On the actual device the exposed `state` switch is non-functional: `genOnOff` on/off commands to ep43 are accepted (no ZCL error) but produce no sound and never set the alarm attributes (`alarm_alarm_1`/`alarm_alarm_2` stay `false`). The siren is only controllable via the `warning` expose (`ssIasWd.startWarning`), which works as expected. This unit runs the newer firmware framework (OTA `SmartSiren 2.0.4`), where the genOnOff cluster no longer drives the siren, unlike the older SIRZB-110/111. Removing onOff also drops the useless `state` switch from the UI and eliminates the genOnOff `configureReport` failures entirely, so `configureReporting: false` is no longer needed.

Koenkk:
Thanks!

PR #12852, opened 2026-08-06
fix: Philips 7299760PH (Hue Bloom): add color temp range

ChillZwix:
The Hue Bloom (7299760PH) definition only exposes `color: true`, so color temperature control is missing. Add `colorTemp: {range: [153, 500]}`. Tested on a real device (LLC012).


Koenkk:
thanks!

PR #12853, opened 2026-08-06
FIX: Implement external temperature management for Shelly BLU TRV

Burki24:
Add support for supplying an external room temperature to the Shelly BLU TRV through the Zigbee Temperature Measurement cluster.

The TRV accepts external temperatures as server-to-client attribute reports on cluster 0x0402. Since it is a sleepy end device, an individual report can be missed while its receiver is inactive. Direct writes to measuredValue are not supported and are rejected with NOT_AUTHORIZED.

Implement confirmation-driven delivery without exposing transport diagnostics:

- expose external_temperature as the requested room temperature;
- enable RemoteSensing bit 0 while preserving the other sensing-source bits;
- send the value as a Temperature Measurement measuredValue report;
- keep the requested value internally until localTemp confirms it;
- retry after incoming thermostat activity, when the TRV is likely awake;
- stop immediately once the reported local temperature matches the request;
- bound retries to 30 attempts or 60 minutes;
- restore an unconfirmed request after a Zigbee2MQTT restart;
- retry when the device announces itself;
- log confirmation and final expiry

The existing thermostat, calibration, manual-mode, valve-position, battery and identify functionality remains unchanged.

Validated on a physical BLU TRV in 12 consecutive tests, including a change from 25 °C to 1 °C. Each requested value was confirmed through local_temperature.

### DIFF
--- a/src/devices/shelly.ts
+++ b/src/devices/shelly.ts
@@ -2196,11 +2196,7 @@ function shellyTRVGetRemoteSensing(endpoint: Zh.Endpoint, state: KeyValue = {}, 
 async function shellyTRVEnableRemoteTemperature(endpoint: Zh.Endpoint, remoteSensing: number): Promise<number> {
     const enabled = remoteSensing | 1;
     if (enabled !== remoteSensing) {
-        await endpoint.write(
-            "hvacThermostat",
-            {remoteSensing: enabled},
-            {disableResponse: true, disableDefaultResponse: true},
-        );
+        await endpoint.write("hvacThermostat", {remoteSensing: enabled}, {disableResponse: true, disableDefaultResponse: true});
     }
     return enabled;
 }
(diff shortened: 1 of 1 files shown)


src/devices/shelly.ts
```diff
@@ -2092,6 +2110,249 @@ const tzLocal = {
     } satisfies Tz.Converter,
 };
 
+/**
+ * Returns the pending external temperature request for a TRV.
+ */
+function shellyTRVGetExternalTemperatureRequest(device: Zh.Device): ShellyTRVExternalTemperatureRequest | undefined {
+    return globalStore.getValue(device, SHELLY_TRV_EXTERNAL_TEMPERATURE_STORE_KEY) as ShellyTRVExternalTemperatureRequest | undefined;
```
Koenkk: This value is lost whenever z2m is restarted, is that a problem? Otherwise we have to save it in `device.meta`

Burki24: Good point. The current implementation reconstructed the pending request from the persisted Zigbee2MQTT state after a restart, but persisting the request itself is cleaner and avoids depending on that reconstruction.

I moved the pending external-temperature request to device.meta and persist it with device.save(). The request target and createdAt are therefore retained across Zigbee2MQTT restarts. Retry bookkeeping is kept in memory to avoid writing the device database on every retry; the persisted createdAt still keeps the existing 60-minute retry lifetime bounded. Confirmed or expired requests are removed from device.meta and saved immediately.

@Koenkk many thanks for the tip.

Koenkk:
Thanks!

PR #12854, opened 2026-08-06
fix(schneider): keep PowerTag empty-payload messages on the debounce …

ChillZwix:
…path

PowerTag greenPower devices (EKO01825 etc.) retransmit readings and also send frames the converter produces no data for (0xa3 multi-cluster report, acCurrentOverload-only frames). Returning an empty payload triggers Z2M's publishLastSeen fallback, which republishes the full cached state immediately — bypassing the per-device debounce and causing ~3 extra publishes per 5s reporting window. Returning meta.state for those messages routes everything through the debounce, collapsing each window into a single merged publish. With debounce: 0 the behavior is unchanged (the fallback already published the full cache). Verified on an ELKO EKO01825: 35 → 12 publishes/min.

### DIFF
--- a/src/devices/schneider_electric.ts
+++ b/src/devices/schneider_electric.ts
@@ -1390,16 +1390,16 @@ const fzLocal = {
         type: ["commandNotification", "commandCommissioningNotification"],
         convert: async (model, msg, publish, options, meta) => {
             if (msg.type !== "commandNotification") {
-                return;
+                return {...meta.state};
             }
 
             const commandID = msg.data.commandID;
 
             if (utils.hasAlreadyProcessedMessage(msg, model, msg.data.frameCounter, `${msg.device.ieeeAddr}_${commandID}`)) {
-                return;
+                return {...meta.state};
             }
 
-            if (commandID >= 0xe0) return; // Skip op commands
+            if (commandID >= 0xe0) return {...meta.state}; // Skip op commands
 
             const rxAfterTx = msg.data.options & (1 << 11);
             const ret: KeyValue = {};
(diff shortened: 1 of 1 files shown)


Koenkk:
it's not clear to me what issue is being addressed here, could you further explain the problem?

ChillZwix:
The PowerTag (EKO01825 family) sends several frames per reporting interval that the greenPower converter produces no data for:

- retransmissions of the same readings (already deduped via `frameCounter`),
- `0xa3` multi-cluster reports,
- `acCurrentOverload`-only frames,
- op commands (`commandID >= 0xe0`).

For those the converter returns nothing. In `lib/extension/receive.ts`, an empty payload makes Z2M call `utils.publishLastSeen(...)`, which republishes the **full cached state immediately** — straight through `publishEntityState`, bypassing the per-device debounce. On my PowerTag that added ~3 extra publishes per 5s reporting window, so instead of one merged publish per window it published roughly every second, around the clock.

The change makes the converter return `{...meta.state}` for those frames. The payload is then non-empty, so it flows through the normal `publish()` path and into the debounce (`publishDebounce` merges values and emits a single publish per window). Because the returned state is identical to what was last published, the result is exactly one publish per reporting interval.

With `debounce: 0` there is no debounce to bypass — `publishEntityState` is called either way — so behavior is unchanged for existing setups.

Koenkk:
This sounds like a logic flaw in zigbee2mqtt itself, I think we should fix it there

ChillZwix:
Agreed — looking at `receive.ts`, the `else { await utils.publishLastSeen(...) }` path republishes the full cached state straight through `publishEntityState`, skipping the debounce. That does look like the actual bug, and the converter workaround only routes around it.

Happy to keep this PR as-is if you'd rather fix it in zigbee2mqtt. Want me to open an issue there describing the debounce bypass?

Koenkk:
If possible, can you try submitting a PR for z2m?

ChillZwix:
Sure — I'll prepare a PR for that. The empty-payload fallback in `receive.ts` calls `utils.publishLastSeen(...)`, which goes straight to `publishEntityState` and skips the per-device debounce. I'll route it through the debounce-aware publish path instead.

ChillZwix:
PR is up: https://github.com/Koenkk/zigbee2mqtt/pull/32791

It routes the empty-payload case through the regular `publish()` path so the per-device debounce/throttle applies instead of publishing the full cached state immediately via `publishLastSeen`. All tests pass (added one that covers the debounce case).

PR #12858, opened 2026-08-07
Add another new ID for Roximo CRTZ01 in tuya.ts

kulhaker478:
The module has been changed in the new batch of motors


Koenkk:
Thanks!

PR #12859, opened 2026-08-07
feat(add): EIGHTREE Socket

z47mkp98bm-debug:
Add a new TS011F_Socket definition, Tuya-specific features (countdown per gang, power-outage memory, indicator mode, child lock, switch type) included.


Koenkk:
Could you merge this with the https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts#L11350 definition?

z47mkp98bm-debug:
> Could you merge this with the https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts#L11350 definition?

This is about fingerprint priority rather than the feature set.

These devices report `applicationVersion: 68`, which is caught by
`{modelID: "TS011F", applicationVersion: 68, priority: -1}` on `TS011F_plug_3`, so they end
up there and get measurements by 60s polling — `plug_3` never configures attribute
reporting. `TS011F_plug_1` is matched via `zigbeeModel`, which loses to any fingerprint
match, and adding them to its `whiteLabel` doesn't help either since `whiteLabel` only
changes the displayed name, not matching — e.g. `_TZ3000_okaz9tjs` is already in `plug_1`'s
whiteLabel but still resolves to `TS011F_plug_3` with `applicationVersion: 68`. Only a
`modelID` + `manufacturerName` fingerprint outranks those `priority: -1` entries.

These five do support attribute reporting, so `m.electricityMeter()` binds, configures
reporting and reads the current values during configure — measurements are available right
after joining, no polling needed.

This PR doesn't touch `TS011F_plug_3`, so polling for all existing devices is unchanged.

Koenkk:
What about adding a `fingerprint` with higher priority to `TS011F_plug_1`?

z47mkp98bm-debug:
> What about adding a `fingerprint` with higher priority to `TS011F_plug_1`?

Merging them into `TS011F_plug_1` would give wrong readings. `plug_1` hardcodes the
metering scale and never reads it from the device:
endpoint.saveClusterAttributeKeyValue("seMetering", {divisor: 100, multiplier: 1});
These devices report seMetering.divisor = 10, not 100, so currentSummDelivered would
be divided by 100 instead of 10 — energy off by a factor of 10. The same applies to
acCurrentDivisor, which plug_1 hardcodes to 1000 and, since #12817, no longer reads
from the device either.

m.electricityMeter() reads the multipliers/divisors from the device during configure, which is why this PR uses it.

data/database.db entry for one of these devices:
"manufName": "_TZ3008_iooniers",
"modelId": "TS011F",
"appVersion": 68,
"epList": [1, 242],
"haElectricalMeasurement": {
    "acVoltageMultiplier": 1, "acVoltageDivisor": 10,
    "acPowerMultiplier": 1,   "acPowerDivisor": 10
}

Koenkk:
Then we can make an if branch for this device in the configure and set a different divisor there

z47mkp98bm-debug:
> Then we can make an if branch for this device in the configure and set a different divisor there

Done. Moved the five `manufacturerName`s to a `fingerprint` on `TS011F_plug_1` and added a branch in configure.
Please help merge it into the master branch. Thanks.

Koenkk:
Thanks!

Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `EIGHTREE Socket.png`

andrei-lazarov:
Hi @z47mkp98bm-debug, you mentioned the energy scale (seMetering divisor) is also wrong. But you only fixed the current scale (acCurrentDivisor)

PR #12861, opened 2026-08-07
feat(add): Tuya TS0301_cover_2 and refactor Tuya covers a bit

andrei-lazarov:
I added another Tuya cover.. also moved into lib as much as I could, to avoid duplication 😮‍💨 Makes a long PR 😄

- https://github.com/Koenkk/zigbee2mqtt.io/pull/5387


Koenkk:
Thanks!

PR #12863, opened 2026-08-07
fix: Ubisys: write InputActions in size-bounded chunks

ovidiuch:
The device setup writes were chunked per input action template, which is not fine-grained enough for dimmers. A single `dimmer_double` template produces six records:

    attrId (2) + selector indicator (1) + one index (2) + data type (1) +
    octet string length (1) + payload

    On                          13 bytes
    Move up (carries the rate)  15 bytes
    Stop                        13 bytes
    Off                         13 bytes
    Move down (rate)            15 bytes
    Stop                        13 bytes

That is 82 bytes, and with the 3 byte ZCL header the request exceeds the ~82 byte APS payload limit of a secured unicast. The adapter rejects it with MESSAGE_TOO_LONG before anything reaches the device, so a two-button dimmer cannot be configured from Zigbee2MQTT at all - not just the combination of several templates, but a single one.

Fill chunks up to a byte budget instead, independent of how the actions are nested. Two `dimmer_double` templates now produce three writes of 57-59 bytes each.

Verified on a C4 (firmware 2.7.0) with two `dimmer_double` templates across all four inputs, which previously failed every time.

Note that reads have a matching limitation that this does not address: reading InputActions back returns INSUFFICIENT_SPACE once enough actions are configured, since the response cannot be split the way writes can.

### DIFF
--- a/src/devices/ubisys.ts
+++ b/src/devices/ubisys.ts
@@ -33,6 +33,12 @@ const manufacturerOptions = {
     ubisysNull: {manufacturerCode: null},
 };
 
+// Maximum payload of a single structured write. The adapter reports the limit it can send, but it
+// is not exposed to converters, and the usable APS payload shrinks further with source routing.
+// Use the largest payload the previous per-template chunking already produced (`dimmer_single`,
+// four records) so that no request is larger than one that already worked.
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12865, opened 2026-08-08
Device file for Ribag lights created. Converter for RIBAG Vertico Air…

rmurach:
… pendant light added.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/5392


rmurach:
This pendant light dates back to around 2016 and is no longer available with Zigbee control. I have five of them and just had to modify my external converter. Perhaps an official version would help other users as well.

rmurach:
Link to picture pull request corrected.

src/devices/ribag.ts
```diff
@@ -0,0 +1,49 @@
...
+    return extension;
+};
+
+export const definitions: DefinitionWithExtend[] = [
+    {
+        zigbeeModel: ["Ribag Air O"],
+        model: "Ribag Air O",
+        vendor: "RIBAG Licht",
+        description: "RIBAG Vertico Air Pendelleuchte",
+        extend: [ribagLight()],
```
Koenkk: Does this device not work with `m.light`?

rmurach: I believe it no longer works with `m.light`, which is why I initially opened issue https://github.com/Koenkk/zigbee2mqtt/issues/32635. However, I subsequently used AI assistance to create a new external converter that works for me. The problem is that I can't get it to pass the pull request checks, and I'm at a loss as to what to do next. Is there any documentation on how a converter should be structured?

rmurach: The Ribag Vertico Air only accepts the short ZCL form of moveToColorTemp. In particular, it does not work with the optional optionsMask/optionsOverride fields that the generic light converter currently adds. The reported OUI is from dresden elekronik.

Koenkk: I think you can use the `meta.omitOptionalLevelAndColorParams` option for that, examples: https://github.com/search?q=repo%3AKoenkk%2Fzigbee-herdsman-converters+omitOptionalLevelAndColorParams&type=code

rmurach: Thank you very much for your help. I tried it out at home and it works - of course.

However, the new converter still doesn't pass the pull request check. Sorry, but JavaScript isn't really my thing; I mainly use Python.

rmurach:
Unfortunately, I don't know a solution for the error during the check ...

src/devices/ribag.ts
```diff
@@ -0,0 +1,10 @@
+import * as m from "zigbee-herdsman-converters/lib/modernExtend";
```
Koenkk, line 1: Please also add this file to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/index.ts

rmurach: Thank you again for your support. Now I think it is resolved.

Koenkk:
Thanks!

PR #12870, opened 2026-08-09
fix(lumi): enroll SJCGQ12LM (water leak T1) as IAS Zone in configure

SebastianBienert:
Fixes the Aqara Water Leak Sensor T1 (`SJCGQ12LM` / `lumi.flood.agl02`) never reporting `water_leak`.

### Problem

The device pairs, interviews as `SUCCESSFUL`, and `last_seen` updates on a button press — but `water_leak` stays frozen at `false` forever. Shorting the probes produces no Zigbee frame at all; the only thing the device ever sends is `genBasic attributeReport {"modelId":"lumi.flood.agl02"}`.

Reading the IAS Zone attributes on real hardware shows why:

```
z2m: Received Zigbee message from 'Water sensor', type 'readResponse', cluster 'ssIasZone',
  data '{"iasCieAddr":"0x0000000000000000","zoneState":0,"zoneStatus":33,"zoneType":42}'
```

- `zoneType: 42` (0x2A) — correct water sensor type
- `zoneStatus: 33` — bit 0 (Alarm1) **is set**; the hardware detects the shorted probes correctly
- `zoneState: 0` — NOT_ENROLLED, and `iasCieAddr` is all zeros

Per the IAS Zone spec an unenrolled zone must not send `commandStatusChangeNotification`, so the alarm never leaves the device.

### Cause

`zigbee-herdsman` does perform IAS enrollment during the interview, but lumi devices match the `lumi..*` modelID quirk in `Device.interview()`:

```
Interview procedure failed but got modelID matching 'lumi..*', assuming interview succeeded
```

That shortcut marks a failed interview as successful and returns before the enrollment step ever runs. For most lumi devices this is harmless (they use the manufacturer-specific `0xFF01` attribute), but `lumi.flood.agl02` is a genuine IAS Zone device and ends up permanently unenrolled. Re-pairing does not fix it — it just hits the same quirk again.

### Fix

Enroll in `configure()`, mirroring the existing NYCE `NCZ-3011-HA` fix in `src/devices/nyce.ts` (added for the same class of problem in Koenkk/zigbee2mqtt#32480): read the current IAS state, and if it is not already enrolled to this coordinator, write `iasCieAddr` and send `enrollRsp`, then verify `zoneState` actually became 1.

### Verified on hardware

Aqara Water Leak Sensor T1, dateCode 20230825, appVersion 32, on Zigbee2MQTT 2.13.0 with a ZStack3x0 coordinator. Running the equivalent write + `enrollRsp` by hand over MQTT immediately fixed the device:

```
z2m: Received Zigbee message from 'Water sensor', type 'commandStatusChangeNotification',
  cluster 'ssIasZone', data '{"delay":0,"extendedstatus":0,"zoneID":23,"zonestatus":33}'
z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Water sensor', payload '{... "water_leak":true}'
```

`water_leak` then tracked the probes across three open/short cycles with no button press needed (`zonestatus` 33 → `true`, 32 → `false`).

### Scope

Only `SJCGQ12LM` is changed here, since that is the device I could verify. Other lumi definitions using `fz.ias_*_alarm_*` (e.g. `SJCGQ13LM` / `lumi.flood.acn001`, `JTQJ-BF-01LM/BW`) are likely affected by the same quirk and could get the same treatment — happy to extend the PR if you want it applied more broadly.

### Checks run locally

`tsc --noEmit`, `pnpm run check` (biome), and `test/index.test.ts` + `test/checkDefinition.test.ts` all pass. The full vitest suite was too slow to complete on the Raspberry Pi this was developed on, so I left it to CI.

### DIFF
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -2084,6 +2084,7 @@ export const definitions: DefinitionWithExtend[] = [
         fromZigbee: [lumi.fromZigbee.lumi_basic, fz.ias_water_leak_alarm_1, lumi.fromZigbee.lumi_specific],
         toZigbee: [],
         exposes: [e.battery(), e.water_leak(), e.battery_low(), e.tamper(), e.battery_voltage()],
+        version: "0.0.1",
         configure: async (device, coordinatorEndpoint) => {
             await ensureLumiIasEnrollment(device.getEndpoint(1), coordinatorEndpoint);
         },
--- a/src/devices/lumi.ts
+++ b/src/devices/lumi.ts
@@ -8,7 +8,7 @@ import * as lumi from "../lib/lumi";
 import * as m from "../lib/modernExtend";
 import * as reporting from "../lib/reporting";
 import type {DefinitionWithExtend, ModernExtend, Zh} from "../lib/types";
-import {assertNumber} from "../lib/utils";
+import {assertNumber, sleep} from "../lib/utils";
 
 const e = exposes.presets;
 const ea = exposes.access;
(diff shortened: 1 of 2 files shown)


SebastianBienert:
Reported with full diagnostics in Koenkk/zigbee2mqtt#32797.

Koenkk:
> zigbee-herdsman does perform IAS enrollment during the interview, but lumi devices match the lumi..* modelID quirk in Device.interview():

Would it be an option to change the regex to exclude the SJCGQ12LM?

SebastianBienert:
Hi @Koenkk 

Can do, but it would only fix new pairings, the interview isn't re-run for already-paired devices, so existing T1s stay unenrolled. Doing it in configure covers both. I pushed a version: "0.0.1" bump so already-configured devices get re-configured.

The quirk is also the fallback for lumi devices that fail the interview. Dropping this model would mean a device that fails at hasNodeDescriptor() ends up Failed instead of just unenrolled. I can't tell which way the T1 goes without unpairing mine.

For what it's worth, mine did get past the descriptor queries, its database entry has epList: [1] and inClusterList: [0, 1280, 3, 1]  so the interview most likely threw in the IAS enroll block, and the quirk in interview()'s catch marked it successful anyway. 

I can go the herdsman route if you prefer it.

Koenkk:
Thanks!

PR #12872, opened 2026-08-09
feat: Ubisys: Add `operational_status` converter

burmistrzak:
The custom attribute `operational_status` can be used to determine movement direction of window coverings.
It's basically the same as `motor_state` used by Bosch devices. 😊


Koenkk:
Thanks!

PR #12873, opened 2026-08-10
OWON DWS332-E: split from DWS312 device definition

jacky202509:
Split OWON DWS332-E from DWS312 definition.

Keep the same functionality while allowing separate model identification and device images.


Koenkk:
Please use the `whiteLabel` for this

jacky202509:
Updated to use whiteLabel as suggested. Thanks!

Koenkk:
thanks!

PR #12874, opened 2026-08-10
Fix AOYAN AY205Z white label matching

zyjsmile857:
Adds a manufacturerName fallback to the AY205Z whiteLabel fingerprint so devices that report padded model IDs still show as AOYAN/AY205Z.


Koenkk:
Thanks!

PR #12879, opened 2026-08-10
feat(add): StuckAtPrototype AirCube

krobipd:
The [AirCube](https://github.com/StuckAtPrototype/AirCube) (StuckAtPrototype, Apache-2.0) is an ESP32-H2 Zigbee air quality monitor. It ships in two hardware variants that enumerate with the same `AirCube` modelID from a single firmware image:

- **Base** (ENS210 + ENS16x): temperature, humidity, eCO2, tVOC and a TVOC-derived VOC level, plus LED brightness.
- **Pro**: additionally true CO2 (Sensirion SCD41, `msCO2`) and illuminance (VCNL4040, `msIlluminanceMeasurement`).

Both variants share one modelID and differ only in which clusters the hardware implements, so this is a single definition whose CO2 and illuminance **exposes and `configure`** are gated on cluster presence (`supportsInputCluster`), following the `third_reality.ts` `conditionalPressure()` pattern. Gating the `configure` — not just the exposes — is required so a Base unit does not fail `configure` on the absent clusters.

eCO2 / tVOC / VOC-level are read from the device's custom `0xFC01` cluster; brightness is the writable Analog Output `presentValue`. The device exposes no OTA cluster, so no `ota`.

Ported from the manufacturer's own published Zigbee2MQTT converter — credit to **StuckAtPrototype**.

**Verification:** `pnpm build`, `pnpm check` and `pnpm test` all pass; `test/stuckatprototype.test.ts` covers Base/Pro exposes discrimination and that `configure` does not throw on a Base unit. Verified against real Pro hardware (endpoint 10, input clusters `[0, 3, 13, 1024, 1026, 1029, 1037, 64513]`); the Base path follows the manufacturer's published cluster set / ZHA quirk signature and is gated by cluster presence.

Link to picture pull request: https://github.com/Koenkk/zigbee2mqtt.io/pull/5395

### DIFF
--- a/src/devices/index.ts
+++ b/src/devices/index.ts
@@ -320,6 +320,7 @@ import {definitions as spotmau} from "./spotmau";
 import {definitions as sprut} from "./sprut";
 import {definitions as stello} from "./stello";
 import {definitions as stelpro} from "./stelpro";
+import {definitions as stuckatprototype} from "./stuckatprototype";
 import {definitions as sunricher} from "./sunricher";
 import {definitions as superled} from "./superled";
 import {definitions as swann} from "./swann";
@@ -699,6 +700,7 @@ const definitions: DefinitionWithExtend[] = [
     ...sprut,
     ...stello,
     ...stelpro,
+    ...stuckatprototype,
     ...sunricher,
     ...superled,
     ...swann,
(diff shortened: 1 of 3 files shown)


Koenkk:
Thanks!

PR #12889, opened 2026-08-11
Fix AOYAN AY02SZ white label matching

zyjsmile857:
Adds a manufacturerName fallback to the AY02SZ whiteLabel fingerprint so devices that report padded model IDs still show as AOYAN/AY02SZ.


Koenkk:
thanks!

PR #12890, opened 2026-08-11
fix: PTM 215Z: match Friends of Hue source-ID range (Senic & Gira smart switch)

mrvdb:
The Senic & Gira Friends of Hue smart switch (SKU 100120) is built on the EnOcean PTM 215Z FOH-7 DB module. It emits the standard PTM 215Z Green Power command vocabulary, but its EnOcean source-ID block starts at `0x01A4xxxx`, which the current fingerprint regex (`/^0x00000000017.....$/`) doesn't cover — so Z2M falls back to the generated Green Power definition (`recall_sceneN`, no action entity).

This adds a second fingerprint entry for the `0x01A4xxxx` block plus a whiteLabel entry. Verified on real hardware (source ID `0x0000000001a41678`): all four rockers publish `press_1`/`release_1` … `press_4`/`release_4` correctly with this definition.

The new range is kept deliberately narrow (4 fixed nibbles vs 3 for the `0x017` block) since I have one confirmed device; happy to widen to `/^0x0000000001a.....$/` if you know the FoH allocation is broader.

### DIFF
--- a/src/devices/enocean.ts
+++ b/src/devices/enocean.ts
@@ -133,7 +133,10 @@ const fzLocal = {
 
 export const definitions: DefinitionWithExtend[] = [
     {
-        fingerprint: [{modelID: "GreenPower_2", ieeeAddr: /^0x00000000017.....$/}],
+        fingerprint: [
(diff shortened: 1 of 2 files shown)


Koenkk:
thanks!

PR #12898, opened 2026-08-12
Detect _TZ3210_fawk5xjv as Nova Digital NFZB-03

rodolfomenardi:
Detect the `_TZ3210_fawk5xjv` firmware variant of the Nova Digital NFZB-03 3-gang switch, matching the existing `_TZ3000_fawk5xjv` definition in `src/devices/tuya.ts`.

The `_TZ3210` prefix is the newer Tuya app framework; the identical `fawk5xjv` suffix indicates the same device profile, so the fingerprint is added to the existing definition rather than creating a duplicate (the `Model should be unique` test would fail otherwise).

Also enables `onOffCountdown`, which the newer `_TZ3210` generation supports (same option set as the Nova Digital NFZB-2, #12658).

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -8191,7 +8191,7 @@ export const definitions: DefinitionWithExtend[] = [
         ],
     },
     {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12899, opened 2026-08-12
fix: Philips LTW015 and LCB001: correct color temp range

elizabethkang8:
Both definitions used `range: undefined`, which falls back to the generic 150-500 mired default rather than the hardware limits.

Read from hardware via `lightingColorCtrl`:
LTW015 (9290011998B):
{"colorCapabilities":16,"colorTempPhysicalMax":454,"colorTempPhysicalMin":153}

LCB001 (548727):
{"colorCapabilities":31,"colorTempPhysicalMax":500,"colorTempPhysicalMin":153}


Koenkk:
Thanks!

PR #12900, opened 2026-08-12
fix: sync FUT035Z+ RF color temperature

mp33de:
The MiBoxer FUT035Z+ variant with manufacturer `_TZB210_ue01a0s2` does not update
`color_temp` in Zigbee2MQTT when the color temperature is changed using the MiBoxer
RF remote.

The device reports these RF color temperature changes as raw `lightingColorCtrl`
frames. The reported value ranges from 0 to 1000 and represents the full CCT range
in the opposite direction to the Zigbee color temperature value.

This change adds a device-specific fromZigbee converter for `_TZB210_ue01a0s2`
which converts the raw value to the existing 153–500 color temperature range.

The converter is restricted to this manufacturer variant so other TS0502B devices
using the shared definition are unaffected.

Tested with the physical device and MiBoxer RF remote.

Tested values:
- 0% -> 500
- 25% -> 413
- 50% -> 327
- 75% -> 240
- 100% -> 153

Changes made through the RF remote are now immediately reflected in Zigbee2MQTT,
while normal Zigbee color temperature control continues to work.

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -1281,6 +1281,25 @@ const tzLocal = {
 };
 
 const fzLocal = {
+    // FUT035Z+ (_TZB210_ue01a0s2) reports RF color temperature changes
+// as raw lightingColorCtrl frames without updating colorTemperature.
+    FUT035Z_rf_color_temp: {
+        cluster: "lightingColorCtrl",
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #12905, opened 2026-08-13
feat: Philips Hue: Expose identify cluster

burmistrzak:
Expose the `genIdentify` cluster of most (not all atm) Hue lights and accessories.
Speaking from first hand experience, being able to quickly identify a particular device during setup is super helpful, especially when dealing with a large number of similar devices. 😅

Most ecosystems (HA, Homebridge, etc.) have long supported this feature, but _for some reason_, the `genIdentify` cluster almost never gets exposed...

Requires https://github.com/Koenkk/zigbee-herdsman-converters/pull/12926


andrei-lazarov (commented):
I think it's cleaner to add it to philips light ?

```ts
philips.m.light({identify: true}) // default
philips.m.light({identify: false})
```

burmistrzak:
Interesting, but not everything is a light tho... 😉
It's probably better to keep `philips.m.light()` focused on lighting stuff and expose every additional clusters via `extend[]`.
Same story with the motion sensors for example.

Koenkk:
> Interesting, but not everything is a light tho

Can you give an example of a device using `philips.m.light` which does not support identify?

burmistrzak:
> Can you give an example of a device using `philips.m.light` which does not support identify?

@Koenkk None that I'm aware of. 😊
I was just thinking that keeping `philips.m.light` focused on lighting makes a bit more sense and follows the example set by e.g. Hue Motion Sensor.

But if you'd rather have `genIdentify` as a default of `philips.m.light` option, as @andrei-lazarov suggested, just let me know!

Koenkk:
Since we want to add it for every light I think it's better to do so, then we make sure we don't miss it and reduce LOC

burmistrzak:
> Since we want to add it for every light I think it's better to do so, then we make sure we don't miss it and reduce LOC

@Koenkk Ok, so I did try a basic implementation

```typescript

light: (args?: modernExtend.LightArgs & {hueEffect?: boolean; gradient?: true | {extraEffects: string[]}; identify?: boolean}) => {
    args = {hueEffect: true, turnsOffAtBrightness1: true, ota: true, identify: true, ...args};

// [...]

        if (args.identify !== false) {
          const genIdentify = modernExtend.identify();
          result.exposes = [...(result.exposes ?? []), ...genIdentify.exposes];
          result.toZigbee = [...(result.toZigbee ?? []), ...genIdentify.toZigbee];
        }

// [...]

```

but one test keeps failing (for obvious reasons).

```text

 FAIL  test/checks.test.ts > Check definitions > Exposes properties are unique
Error: Duplicate expose property found: 'identify' for '929003711201'
 ❯ test/checks.test.ts:100:27
     98|             for (const expose of exposes) {
     99|                 if (expose.property && found.includes(expose.property)) {
    100|                     throw new Error(`Duplicate expose property found: '${expose.property}' for '${def…
       |                           ^
    101|                 }
    102|

```

I don't have said Hue light (`929003711201`) on hand to check but don't multi-endpoint fixtures have `genIdentify` on each endpoint?

burmistrzak:
Looking a bit deeper into this, it seems like `m.identify()` does not handle multi-endpoint devices adequately.
Based on my understanding, only the default endpoint can currently be identified because the `endpointNames` argument is not supported and a single identifiable endpoint is assumed.

I would propose making `m.identify()` aware of multiple endpoints. @Nerivec Thoughts? 👀

Nerivec:
I won't be much help. I've never used identify, all my devices that could need it, have a sticker on them; extremely reliable alternative. :grin:
Koenkk knows the ME stuff better than me :wink:

burmistrzak:
> I won't be much help. I've never used identify, all my devices that could need it, have a sticker on them; extremely reliable alternative. 😁 Koenkk knows the ME stuff better than me 😉

An analog identify! 😲
But yeah, `genIdentify` is quite useful when setting up a lot of accessories at once or testing reachability.

burmistrzak:
I had to return to my original plan because integrating `m.identify()` into `philips.m.light()` got too convoluted for my liking.
Especially in light of https://github.com/Koenkk/zigbee-herdsman-converters/pull/12926 it's probably better to follow established conventions here.

andrei-lazarov (approved):
Whatever works 🙂

burmistrzak:
> Whatever works 🙂

@Koenkk Ready for takeoff. 🫡

Koenkk:
Thanks!

PR #12912, opened 2026-08-14
Add Philips 929004296801 (Hue wired wall switch module 2 channel)

reeett:
Adds support for the Philips Hue wired on/off wall switch module, 2 channel (EAN 8721103111753, 12NC 929004296801, Zigbee model `LOM014`).

This is one of the mains-powered in-wall relay modules Signify released in June 2026. Unlike the battery-powered wall switch module (RDM001/RDM004), which is an input device only, this one actually switches the load: two independent channels, neutral required, and it acts as a router.

Endpoint layout as reported by the device:

| Endpoint | Device ID | Purpose |
|---|---|---|
| 1 | 0x0103 | switch input (`genBasic`, `genIdentify`, `0xFC00`) |
| 11 | 0x0100 | channel 1 (`genOnOff`, `genLevelCtrl`, `0xFC03`, OTA) |
| 12 | 0x0100 | channel 2 (`genOnOff`, `genLevelCtrl`, `0xFC03`) |
| 242 | 0x0061 | Green Power |

Exposed as two switches (`l1`, `l2`) with power-on behaviour, identify and OTA.

Tested on firmware 1.157.2 (date code 20260331): both channels switch and report state correctly, including when operated from the physical wall switch; power-on behaviour and identify work on both channels.

Notes on things that are deliberately not exposed:

- The switch inputs are hard-wired to their respective relays. The device does emit On/Off commands, but it emits them for every relay change, including ones triggered over MQTT, so they cannot be used to distinguish local operation from remote control. `state_l1` / `state_l2` should be used as automation triggers instead.
- `0xFC00` (manuSpecificPhilips) is present on endpoint 1 but appears to be a leftover from the shared firmware base; no button notifications are sent. The Hue app offers no rocker/push-button selection for this device either, so the `device_mode` attribute used by the RDM001 does not apply here.
- `genLevelCtrl` is present on both output endpoints but this is the on/off variant, not the dimming one.


reeett:
Added image and docs note with https://github.com/Koenkk/zigbee2mqtt.io/pull/5410

Koenkk:
Thanks!

PR #12913, opened 2026-08-14
feat(add): PIR313-L

### LINKED ISSUE #12285: OWON PIR313-E (hw_version 8): Interview fails - cannot get active endpoints, configure crashes at owon.ts:917
Interview failed because can not get active endpoints ('0x3c6a2cfffed85011')

Interview failed because can not get active endpoints

**Device:** OWON PIR313-E / PIR313-L
**Z2M version:** 2.10.1-1
**zigbee-herdsman-converters version:** 26.46.0
**Coordinator:** EmberZNet (Home Assistant add-on)
**hw_version:** 8

## What happened?

The PIR313-E pairs successfully and is recognized as supported, but all values remain `null` (temperature, humidity, illuminance, occupancy, tamper, battery_low). Only `illuminance` occasionally reports a value spontaneously.

## Root cause identified

Z2M cannot retrieve the active endpoints from the device:

```
Interview failed because can not get active endpoints ('0x3c6a2cfffed85011')
```

This causes the configure step to crash:

```
Failed to configure 'PIR313', attempt 3 (TypeError: Cannot read properties of undefined (reading 'bind')
    at Object.bind (.../reporting.ts:48:24)
    at configure (.../owon.ts:917:33)
```

Because endpoints are undefined, `reporting.bind()` crashes at `owon.ts:917`.

## Device info

- Device reports Model ID as `PIR313` (not `PIR313-E`), but Z2M maps it to the PIR313-E definition
- `hw_version: 8` — likely a newer hardware revision with different endpoints
- `leave_count: 3` observed in bridge health logs

## MQTT payload example

```json
{"battery_low":null,"humidity":null,"illuminance":9,"occupancy":null,"tamper":null,"temperature":null,"linkquality":204}
```

## Steps to reproduce

1. Pair an OWON PIR313-E (hw_version 8, model ID `PIR313`) with Z2M 2.10.1
2. Device joins and is recognized as supported (mapped to PIR313-E definition)
3. Z2M fails to get active endpoints during interview
4. Configure crashes at `owon.ts:917` with `TypeError: Cannot read properties of undefined (reading 'bind')`
5. All exposed values remain `null` indefinitely

## Expected behavior

Interview should complete, endpoints should be retrieved, configure should succeed and all values (temperature, humidity, illuminance, occupancy) should be reported.

## Additional context

- Signal quality excellent (linkquality: 188-204)
- Battery confirmed good (new batteries tested)
- The device is a newer hardware revision (hw_version 8) that may have different endpoint structure than what owon.ts expects

grobian:
Previously detected as PIR313-E, but this device only emits illuminance, and with the E definition, this isn't visible/working.

Detect the PIR313-L (which identifies itself also as PIR313) based on its characteristics via fingerprint.  Only expose illuminance.  Do not publish battery state because even though advertised, it always seems to be 0, which is confusing and misleading.

The image for this PIR313-L should be identical to PIR313-E in zigbee2mqtt.io/public/images/devices.

For this commit, I've used ChatGPT extensively to understand how to add a new entry that would just match this device, and not break the PIR313-E support.

Fixes: https://github.com/Koenkk/zigbee-herdsman-converters/issues/12285

### DIFF
--- a/src/devices/owon.ts
+++ b/src/devices/owon.ts
@@ -903,6 +903,31 @@ export const definitions: DefinitionWithExtend[] = [
         extend: [m.onOff(), m.electricityMeter({cluster: "metering"})],
     },
     {
+        fingerprint: [
+            {
+                modelID: "PIR313",
+                manufacturerName: "OWON",
+                endpoints: [
+                    {
+                        ID: 2,
+                        profileID: 260,
+                        deviceID: 262,
+                        inputClusters: [1, 0, 3, 1024],
+                        outputClusters: [3],
+                    },
+                ],
+            },
+        ],
+        model: "PIR313-L",
+        vendor: "OWON",
+        description: "Light sensor",
+        extend: [
+            m.illuminance({
+                reporting: {min: 300, max: 3600, change: 100},
+            }),
+        ],
+    },
+    {
         zigbeeModel: ["PIR313-E", "PIR313"],
         model: "PIR313-E",
         vendor: "OWON",


src/devices/owon.ts
```diff
@@ -902,6 +902,31 @@ export const definitions: DefinitionWithExtend[] = [
         description: "32A/63A power circuit breaker",
         extend: [m.onOff(), m.electricityMeter({cluster: "metering"})],
     },
+    {
+        fingerprint: [
+            {
+                modelID: "PIR313",
+                manufacturerName: "OWON",
+                endpoints: [
```
Koenkk, line 910: No need to set `endpoints` here, the`modelID` + `manufacuterName` is already unique enough

grobian: I'm affraid not, both PIR313-E and PIR313-L have modelID PIR313 and manufacturer OWON.  See the declaration right below the added one.

Koenkk: I see, thanks!

Koenkk:
Could you also submit a picture for the docs ([guide](https://www.zigbee2mqtt.io/advanced/support-new-devices/01_support_new_devices.html#_5-add-device-picture-to-zigbee2mqtt-io-documentation))? Make sure that the file name is `PIR313-L.png`

grobian:
done https://github.com/Koenkk/zigbee2mqtt.io/pull/5463

PR #12919, opened 2026-08-15
feat: expose Power-on behavior in generated support

andrei-lazarov:
When adding a new device, it makes sense to expose as many features as possible. So the user can test everything, and determine the device capabilities.

But Power-on behavior is hidden by default. What is the reasoning here?
I believe this feature is now wrongly missing from a lot of devices.

Examples - users unaware of Power-on behavior:
- https://github.com/Koenkk/zigbee2mqtt/issues/32816
- https://github.com/Koenkk/zigbee2mqtt/issues/32505


Koenkk:
Thanks!

Issue #10030, opened 2025-09-16
SUTON Zigbee DIN RCBO energy meter support (4 poles) / similar to 3 poles


DukeIvo:
I have SUTON Zigbee DIN RCBO energy meter (Tuya) with 3 poles which is recognized as https://www.zigbee2mqtt.io/devices/STB3L-125-ZJ.html#suton-stb3l-125-zj. 
Then i have the same with 4 poles which is not recognized, and seems only difference is that it's Manufacturer: "_TZE284_wbhaespm" while 3pole one has "_TZE204_wbhaespm".

Can we add support for "_TZE284_wbhaespm" ?

To me seems we need just to update line 7661 in tuya.js to support this 

From: fingerprint: tuya.fingerprint("TS0601", ["_TZE204_wbhaespm", "_TZE200_wbhaespm"]),
To: fingerprint: tuya.fingerprint("TS0601", ["_TZE204_wbhaespm", "_TZE200_wbhaespm", "_TZE284_wbhaespm"]),
---------

Thanks in advance,

Koenkk:
Could you try to:

- Stop Z2M
- Open `data/database.db` with a text editor and change all occurrences of `_TZE204_wbhaespm` to `_TZE284_wbhaespm`
- Start Z2M and see if the device now works

DukeIvo:
Hi, i think you meant from _TZE284_wbhaespm to _TZE204_wbhaespm.

I did - starts fine, read values like V, A, W, but i cant change threshold values - gives an error: 

2025-09-17 15:27:53z2m: Publish 'set' 'over_current_threshold' to 'circuit_breaker_main' failed: 'Error: Expected string or number, got: undefined'
error 2025-09-17 15:27:53z2m: Publish 'set' 'over_voltage_threshold' to 'circuit_breaker_main' failed: 'Error: Expected string or number, got: undefined'

Koenkk:
Could you provide the debug log of that?

> See [this](https://www.zigbee2mqtt.io/guide/configuration/logging.html) on how to enable debug logging.

DukeIvo:
[log.zip](https://github.com/user-attachments/files/22419949/log.zip) , here you go....   / Thanks

Koenkk:
Tried to fix it, can you check again?

Changes will be available in the [dev branch](https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) in a few hours from now and in the next release which is every 1st of the month.

[...]

PR #12923, opened 2026-08-16
fix: correct STB3L-125-ZJ phase mapping

### LINKED ISSUE #12917: STB3L-125-ZJ (_TZE204_wbhaespm): phases A and C are swapped
### Device

- Zigbee model: `TS0601`
- Manufacturer: `_TZE204_wbhaespm`
- Zigbee2MQTT model: `STB3L-125-ZJ`
- Vendor: SUTON
- Zigbee2MQTT: 2.12.1
- Two installed units have application version 74

### Problem

The current definition maps:

- DP6 -> phase A
- DP7 -> phase B
- DP8 -> phase C

On both tested `_TZE204_wbhaespm` units, the physical phase order is instead:

- DP6 -> phase C
- DP7 -> phase B
- DP8 -> phase A

This swaps voltage, current, and power for phases A and C.

### Reproducible controlled-load test

A load was connected directly to the physical far-left output pole of the three-phase meter. The wiring and terminal order were checked, and an independent Shelly Pro 3EM identified this conductor as physical phase A.

Repeated steps produced:

| Test load | `power_a` / DP6 | `power_c` / DP8 |
|---|---:|---:|
| on | 0 W | 98 W |
| off | 0 W | 0 W |
| on | 0 W | 99 W |
| off | 0 W | 0 W |
| on | 0 W | 99 W |

An earlier larger-load comparison also gave:

- DP8 bytes `[9,130,0,8,79,0,1,227]` -> 243.4 V, 2.127 A, 483 W, matching physical phase A on the independent meter.
- DP6 bytes `[9,99,0,0,253,0,0,35]` -> 240.3 V, 0.253 A, 35 W, matching physical phase C on the independent meter.

### Tested local fix

A fingerprint-specific external converter for `_TZE204_wbhaespm`, application version 74, changes only:

```js
[6, null, tuya.valueConverter.phaseVariant2WithPhase("c")],
[7, null, tuya.valueConverter.phaseVariant2WithPhase("b")],
[8, null, tuya.valueConverter.phaseVariant2WithPhase("a")],
```

A deterministic decode using the captured payloads returns:

```json
{
  "dp6": {"voltage_c": 240.3, "current_c": 0.253, "power_c": 35},
  "dp8": {"voltage_a": 243.4, "current_a": 2.127, "power_a": 483}
}
```

The external definition is selected successfully for both units.

### Suggested scope

The built-in definition currently groups `_TZE204_wbhaespm`, `_TZE200_wbhaespm`, and `_TZE284_wbhaespm`. I only tested `_TZE204_wbhaespm` application version 74, so a global DP6/DP8 swap for all three fingerprints could regress untested variants. Splitting this fingerprint into a dedicated definition may be safer unless others confirm the same physical mapping.

analienx:
Fixes #12917

The `_TZE204_wbhaespm` variant of the SUTON `STB3L-125-ZJ` reports the three phase datapoints in C/B/A order instead of A/B/C.

Verified by loading one physical phase at a time on two identical `_TZE204_wbhaespm` devices with `applicationVersion: 74`:

- DP6 = phase C
- DP7 = phase B
- DP8 = phase A

This PR keeps the existing `STB3L-125-ZJ` definition and applies the swapped A/C mapping only for `_TZE204_wbhaespm`.

The existing A/B/C mapping is preserved for `_TZE200_wbhaespm` and `_TZE284_wbhaespm`, since their phase order has not been verified.

Validation completed successfully:

- `pnpm run build`
- `pnpm run check`
- `pnpm test`
- `pnpm run bench`

### DIFF
--- a/src/devices/tuya.ts
+++ b/src/devices/tuya.ts
@@ -200,6 +200,16 @@ const storeLocal = {
 };
 
 const convLocal = {
+    phaseVariant2WithPhaseByManufacturer: (phase: string, tze204Phase: string) => {
+        const defaultConverter = tuya.valueConverter.phaseVariant2WithPhase(phase);
+        const tze204Converter = tuya.valueConverter.phaseVariant2WithPhase(tze204Phase);
+        return {
+            from: (v: string, meta: Fz.Meta) => {
+                const converter = meta.device.manufacturerName === "_TZE204_wbhaespm" ? tze204Converter : defaultConverter;
+                return converter.from(v);
+            },
+        };
+    },
     novaDigitalToDmBrightness: {
         from: (value: unknown) => {
             const clamped = Math.max(10, Math.min(1000, Number(value) || 10));
@@ -12180,9 +12190,9 @@ export const definitions: DefinitionWithExtend[] = [
         meta: {
             tuyaDatapoints: [
                 [1, "energy", tuya.valueConverter.divideBy100],
-                [6, null, tuya.valueConverter.phaseVariant2WithPhase("a")],
+                [6, null, convLocal.phaseVariant2WithPhaseByManufacturer("a", "c")],
                 [7, null, tuya.valueConverter.phaseVariant2WithPhase("b")],
-                [8, null, tuya.valueConverter.phaseVariant2WithPhase("c")],
+                [8, null, convLocal.phaseVariant2WithPhaseByManufacturer("c", "a")],
                 [9, "faults", tuya.valueConverter.circuitBreakerFaults],
                 [16, "state", tuya.valueConverter.onOff],
                 [17, null, tuya.valueConverter.threshold_2],


andrei-lazarov (changes requested):
I've seen the report, but I'm not sure we want to do this.

**It will wreck the statistics for anyone already using the device.**

Are there labels on the device, or why should we swap the phases?
I would also guess the current Z2M mapping matches the Tuya gateway mapping.

(But you can ask for more opinions in [threads](https://github.com/Koenkk/zigbee-herdsman-converters/issues/10030) about this device)

analienx:
Hello @andrei-lazarov , that’s a fair concern, especially regarding existing statistics.

The reason for the change is that we verified the physical phase mapping directly on two "_TZE204_wbhaespm" units with applicationVersion 74, loading one phase at a time and cross-checking it against an independent 3-phase meter. The result was consistent on both devices:

- DP6 = physical phase C
- DP7 = physical phase B
- DP8 = physical phase A

So this isn’t based on an assumption about DP numbering, it’s based on the actual physical conductors being loaded.

The PR is intentionally limited to "_TZE204_wbhaespm"; the existing mapping for "_TZE200_wbhaespm" and "_TZE284_wbhaespm" is left unchanged because we haven’t physically verified those variants.

Regarding the Tuya gateway: even if its UI shows A/B/C in the expected order, it may already be remapping the raw DPs internally. That would actually be consistent with what we’re doing here.

I’m happy to ask for a few more confirmations in the device thread as well, but based on the physical tests I’m quite confident the current Z2M A/C labels are swapped for this variant.

andrei-lazarov:
Currently I think the issue can be solved by physically swapping the wires

analienx:
Thanks for raising the compatibility concern. I found an important detail since opening this PR: the two physical devices I tested are actually SMTONOFF, not SUTON, and the seller confirmed that STB3L-125-ZJ is not the model they sold.
So the measurements are still valid for these two units, but the likely problem is that _TZE204_wbhaespm is shared by multiple OEM variants and my SMTONOFF device is being matched to the SUTON definition incorrectly.
I’m withdrawing this PR rather than risk changing the mapping for genuine SUTON users. I’ll follow up with a separate fingerprint/device-identification issue.

PR #12924, opened 2026-08-16
feat(add): 929004321101

PSeros:
Adds model number 929004321101 to the existing 929004321001 definition.

Same product (Hue Play floor lamp large), different regional article number. The device paired as an automatically generated definition because the model ID did not match.


Koenkk:
thanks!

PR #12925, opened 2026-08-16
[fix] occupancy on multiendpoint devices

luar123:
Fix m.occupancy for multi endpoint devices.

There were two issues:
* exposes were not cloned but reused, so withEndpoint was applied twice for the second EP.
* On multi EP devices with occupancy only on the first EP endpointNames is undefined but `postfixWithEndpointName` still adds the EP suffix.


Koenkk:
Thanks!

PR #12926, opened 2026-08-16
fix: Support `genIdentify` on multiple endpoints

burmistrzak:
As discovered in https://github.com/Koenkk/zigbee-herdsman-converters/pull/12905, the `modernExtend` variant of `genIdentify` is currently incapable of identifying individual endpoints of multi-endpoint devices (e.g. [929003711201](https://www.zigbee2mqtt.io/devices/929003711201.html)).

This PR adds the well-known `endpointNames` argument to the existing `identify` converter.

### Usage

```typescript
extend: [
    philips.m.addManuSpecificPhilipsCluster(),
    m.deviceEndpoints({endpoints: {switch: 1, back: 11, front: 12}}),
    philips.m.light({colorTemp: {range: [153, 500]}, color: true, endpointNames: ["front"]}),
    philips.m.light({colorTemp: {range: [153, 500]}, color: true, gradient: true, endpointNames: ["back"]}),
    philips.m.twilightOnOff(),
    m.identify({endpointNames: ["front", "back"]}), // NEW
],
```


Koenkk:
thanks!

andrei-lazarov:
Nice! Do you also want to check `generateDefinition`, so it detects identify?

burmistrzak:
> Nice! Do you also want to check `generateDefinition`, so it detects identify?

@andrei-lazarov Yes, once I'm absolutely sure multi-endpoint identify works as intended, I'll update `generateDefinition` as well.
From now on, every supported and exposed endpoint should also be identifiable. It's such a basic feature, but immensely useful when you actually need it. 😅

Depending on how HA handles multiple identify buttons on a single MQTT device, we might have to add a discriminator (e.g. `Identify (2)`) to the entity name.

andrei-lazarov:
Yeah it should append the endpointName. So annoying to also verify everything in HA 😮‍💨

burmistrzak:
> Yeah it should append the endpointName. So annoying to also verify everything in HA 😮‍💨

@andrei-lazarov Ok, single `genIdentify` works perfectly in HA and via Matter-Hub in Apple Home! Even battery-powered accessories!
Unfortunately, I don't have any Hue lights with two identifiable endpoints... Do you?

Only multi-endpoint lighting devices I have around are Bosch and Ubisys in-wall modules.

andrei-lazarov:
No, I've never seen devices with multiple identifiable endpoints..
But I tested your change on a power strip and it was ok. Sent the identify command to each endpoint, even if the device didn't respond.

burmistrzak:
> No, I've never seen devices with multiple identifiable endpoints.. But I tested your change on a power strip and it was ok. Sent the identify command to each endpoint, even if the device didn't respond.

@andrei-lazarov Hmm, so an identify command was also sent to endpoints without `genIdentify`? Did you test from HA or WF?

andrei-lazarov:
Yes. I added endpointNames to try it on a random multi-endpoint device. Tested on both HA and WF

burmistrzak:
> Yes. I added endpointNames to try it on a random multi-endpoint device. Tested on both HA and WF

@andrei-lazarov I guess it makes sense because we assume that devs only add valid `endpointNames` in definitions? 😅

PR #12928, opened 2026-08-17
fix: decode 24 bit current and power in phaseVariant2WithPhase

### LINKED ISSUE #12927: `phaseVariant2WithPhase` truncates 24-bit current/power fields; `0x999A` negative-power constant is an artifact of the same bug
## What happened
 
`tuya.valueConverter.phaseVariant2WithPhase` reads only the low 2 bytes of the current and power fields, which are 3 bytes wide in the Tuya payload. Two consequences:
 
1. Current wraps above **65.536 A**.
2. The negative-power workaround added for #18603 uses `0x999A`, which is the low 16 bits of the real 24-bit offset `0x19999A`. It therefore only produces negative values in the narrow window 32768–39321 and corrupts any legitimate reading above 32767 W.
Device used for testing: `TS0601` / `_TZE284_x8diwkqb`, definition `SDM02V1-GT` (2P+N energy monitor). The same converter is used by `SPM02V1-GT` and other Tuya meters, so the fix is shared.
 
## Current code
 
https://github.com/Koenkk/zigbee-herdsman-converters/blob/d03d67e939433ad751334126638762b1ef64045a/src/lib/tuya.ts#L1693-L1711

 
`buf[2]` (current high byte) and `buf[5]` (power high byte) are never read.
 
## Evidence — current wrap
 
Two resistive loads (5.5 kW + 7.5 kW, line-to-line) switched on. Reported by z2m:
 
```json
{
  "current_l1": 3.247,
  "current_l2": 0.134,
  "power_l1": 8216,
  "power_l2": 7866,
  "power": 16082,
  "voltage_l1": 119.6,
  "voltage_l2": 120
}
```
 
`current_l1` of 3.247 A alongside `power_l1` of 8216 W at 119.6 V is impossible. Reconstructing from power: 8216 / 119.6 = **68.7 A**.
 
68783 mA = `0x010CAF`. Discarding `buf[2]` (`0x01`) leaves `0x0CAF` = 3247 → the reported 3.247 A. The power field did not wrap because 8216 fits in 16 bits.
 
After reading all three bytes, same load:
 
```json
{
  "current_l1": 69.61,
  "current_l2": 65.951,
  "power_l1": 8270,
  "power_l2": 7867,
  "power": 16137,
  "voltage_l1": 119,
  "voltage_l2": 119.5
}
```
 
Consistency restored:
 
| check | result |
|---|---|
| 119.0 × 69.61 | 8284 vs `power_l1` 8270 (0.2%, voltage rounding) |
| 119.5 × 65.951 | 7881 vs `power_l2` 7867 (0.2%) |
| `power_l1` + `power_l2` | 16137 = DP 29 exactly |
 
Note `current_l2` = 65.951 A, just 0.4 A past the wrap point — it would have been reported as 0.415 A before the fix.
 
## Evidence — the `0x999A` constant
 
From the measurements posted in #18603, raw values 39125, 39124 and 39123 correspond to −197 W, −198 W and −199 W.
 
Reading the field as 24 bits with offset `0x19999A` (1677722) reproduces all three exactly:
 
| P | 0x19999A + P | hex | low 16 bits |
|---|---|---|---|
| −197 | 1677525 | `0x1998D5` | **39125** |
| −198 | 1677524 | `0x1998D4` | **39124** |
| −199 | 1677523 | `0x1998D3` | **39123** |
 
`0x999A` is literally the low 16 bits of `0x19999A`. The same issue reports the 4-byte total power using offset `429496730` = `0x1999999A` — the same `0x1999…9A` pattern at 32-bit width. This is consistent with firmware forming the two's complement in units of 0.1 W and then doing an unsigned divide by 10, though that part is inference from two data points.
 
The offset encoding is not two's complement, so bit 23 cannot be used as a sign bit; detection has to be by threshold.
 
## Proposed fix
 
```js
const POWER_NEG_OFFSET_24 = 0x19999a; // 1,677,722
const POWER_SANITY_24 = 0x100000;     // 1,048,576 W
 
phaseVariant2WithPhase: (phase: string) => {
    return {
        from: (v: string) => {
            const buf = Buffer.from(v, "base64");
            let power = buf[7] | (buf[6] << 8) | (buf[5] << 16);
            if (power > POWER_SANITY_24) power -= POWER_NEG_OFFSET_24;
 
            return {
                [`voltage_${phase}`]: (buf[1] | (buf[0] << 8)) / 10,
                [`current_${phase}`]: (buf[4] | (buf[3] << 8) | (buf[2] << 16)) / 1000,
                [`power_${phase}`]: power,
            };
        },
    };
},
```
 
The threshold preserves current behaviour for every plausible positive reading (no single phase reaches 1 MW), while the whole plausible negative range maps correctly.
 
The 4-byte total power datapoint (DP 29 on `SDM02V1-GT` / `SPM02V1-GT`, currently `valueConverter.raw`) needs the equivalent treatment with `0x1999999A` if reverse power should be reported at all.
 
## Request for validation
 
I have no bidirectional/PV setup, so I could not test the negative branch on hardware — only reproduce the numbers reported in #18603 arithmetically. Someone with reverse power flow confirming the fix before merge would be useful.
 
## Environment
 
- Device: `TS0601` / `_TZE284_x8diwkqb` (Zemismart SDM02, 2P+N)
- Verified via external converter with the patch above

aberenguel:
The payload is 8 bytes -- voltage (2), current (3), power (3) -- the same layout already decoded by phaseVariant3 and phaseVariant4. Only the low 2 bytes of current and power were read, so current wrapped above 65.536 A: on a TS0601 / _TZE284_x8diwkqb drawing 68.783 A, current was reported as 3.247 A while power and voltage in the same payload implied the real value.

The negative power branch was built on the same truncated field. Its 0x999a constant is the low 16 bits of the actual 24 bit offset 0x19999a, so it only produced negative values between 32768 and 39321 and silently corrupted any legitimate single phase reading above 32767 W. Reading all three bytes and using the full offset reproduces the values reported in #18603 exactly (1677525 -> -197 W, 1677524 -> -198 W, 1677523 -> -199 W).

The offset encoding is not two's complement, so the sign bit cannot be used and the branch is taken on an implausibly high reading instead.

https://github.com/Koenkk/zigbee2mqtt/issues/18603

Fixes #12927

### DIFF
--- a/src/lib/tuya.ts
+++ b/src/lib/tuya.ts
@@ -1691,19 +1691,30 @@ export const valueConverter = {
         },
     },
     phaseVariant2WithPhase: (phase: string) => {
+        // Payload is 8 bytes: voltage (2), current (3), power (3), same layout as
+        // phaseVariant3/phaseVariant4. Reading only the low 2 bytes of current and
+        // power made the current wrap above 65.536 A.
+        //
+        // Support negative power readings
+        // https://github.com/Koenkk/zigbee2mqtt/issues/18603#issuecomment-2277697295
+        // Negative values are not two's complement: they are reported as
+        // NEGATIVE_POWER_OFFSET + power, so the sign bit cannot be used and the
+        // branch is taken on an implausibly high reading instead. The 0x999a
+        // constant previously used here is the low 16 bits of this offset, i.e. an
+        // artifact of the same truncation.
+        const NEGATIVE_POWER_OFFSET = 0x19999a;
+        const IMPLAUSIBLE_POWER = 0x100000; // 1048576 W on a single phase
         return {
             from: (v: string) => {
-                // Support negative power readings
-                // https://github.com/Koenkk/zigbee2mqtt/issues/18603#issuecomment-2277697295
                 const buf = Buffer.from(v, "base64");
-                let power = buf[7] | (buf[6] << 8);
-                if (power > 0x7fff) {
-                    power = (0x999a - power) * -1;
+                let power = buf[7] | (buf[6] << 8) | (buf[5] << 16);
+                if (power > IMPLAUSIBLE_POWER) {
+                    power -= NEGATIVE_POWER_OFFSET;
                 }
 
                 return {
                     [`voltage_${phase}`]: (buf[1] | (buf[0] << 8)) / 10,
-                    [`current_${phase}`]: (buf[4] | (buf[3] << 8)) / 1000,
+                    [`current_${phase}`]: (buf[4] | (buf[3] << 8) | (buf[2] << 16)) / 1000,
                     [`power_${phase}`]: power,
                 };
             },
--- a/test/tuya.test.ts
+++ b/test/tuya.test.ts
@@ -80,4 +80,71 @@ describe("lib/tuya", () => {
             expect(cluster.attributes.moesCalibrationTime).toMatchObject({ID: 0xf003, type: Zcl.DataType.UINT16});
         });
     });
+
+    describe("phaseVariant2WithPhase", () => {
+        // Regression: the payload is 8 bytes -- voltage (2), current (3), power (3) --
+        // the same layout already decoded by phaseVariant3/phaseVariant4. Only the low
+        // 2 bytes of current and power were read, so:
+        //   - current wrapped above 65.536 A (68.783 A was reported as 3.247 A)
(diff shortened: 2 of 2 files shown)


Koenkk:
Thanks!

PR #12930, opened 2026-08-17
fix: SONOFF ZBMINIR2: De-duplicate configuration

andrei-lazarov:
I noticed some duplicated commands in Wireshark.

1. Each `m.binary` and `m.numeric` already reads its respective attribute once. There's no need to read them again in `configure`

2. OnOff binding also happens twice. I removed the one in `m.commandsOnOff`

Tested on real device (v1.0.8). Everything is done only once now. Device still works as expected.


Koenkk:
thanks!

PR #12814, opened 2026-07-31
fix: SONOFF ZBM5: Fix configure failing

andrei-lazarov:
I used whiteLabels to remove the duplicated converters.

And I added a try catch, following this issue:
- https://github.com/Koenkk/zigbee2mqtt/issues/32679

Hopefully no mistakes 🙂

PR #12931, opened 2026-08-17
fix: SONOFF ZBM5 no-Neutral: fix configure still failing

andrei-lazarov:
The ZBM5 in no-Neutral mode does not advertise the output OnOff cluster, causing configuration to fail
- https://github.com/Koenkk/zigbee2mqtt/issues/32679

The previous fix was not enough, because there is one more bind in `m.commandsOnOff`
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/12814

But the bindings fail at Z2M level. So attaching the cluster should allow the bind command.
If it works, this is a cleaner fix: `utils.attachOutputCluster(device, endpoint1, "genOnOff");`

I made sure the cluster is attached before the binds. But I don't have the device to test further.

It might also help in this issue
- https://github.com/Koenkk/zigbee2mqtt/discussions/31557

### DIFF
--- a/src/devices/sonoff.ts
+++ b/src/devices/sonoff.ts
@@ -10995,7 +10995,7 @@ export const definitions: DefinitionWithExtend[] = [
         description: "Zigbee Smart one-channel wall switch (type 120)",
         ota: true,
         extend: [
-            m.commandsOnOff({commands: ["toggle"]}),
+            m.commandsOnOff({commands: ["toggle"], bind: false}),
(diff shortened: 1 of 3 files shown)


andrei-lazarov:
If you can approve this. The PR is safe, it does not risk breaking anything.

The configuration is currently failing on no-Neutral variant. This is a potential fix.

Koenkk:
Thanks!

PR #12938, opened 2026-08-19
feat: Ubisys: Expose identify cluster

burmistrzak:
Same story as in https://github.com/Koenkk/zigbee-herdsman-converters/pull/12905, exposes the `genIdentify` cluster of relevant endpoints. 😊

A `genIdentify` command causes the respective relay/TRIAC output to be toggled a few times.


Koenkk:
thanks!

PR #12940, opened 2026-08-19
fix ZD24

tiankakeji-cloud:


src/devices/tuya.ts
```diff
@@ -16006,8 +16006,14 @@ export const definitions: DefinitionWithExtend[] = [
             e.illuminance(),
             e.battery(),
             e.enum("motion_state", ea.STATE, ["none", "static", "small", "large"]).withDescription("Human body motion state"),
-            e.numeric("distance", ea.STATE).withUnit("m").withDescription("Target distance"),
             e.binary("init", ea.STATE_SET, "ON", "OFF").withDescription("Initialize"),
+            e
```
Koenkk, line 16010: I like the distance name more, why change it? it is also a breaking change

Issue #10812, opened 2025-11-21
Aqara W600 external temperature input


MeisterQ:
The device should expose the external temperature input, valve position (and all other data) like with the Aqara E1.

I cant provide the correct data, but maybe someone else can who is looking for this aswell.

mrflo97:
I also have those devices and would love this feature. I would also gladly help with the implementation, can someone point me in the right direction on how to find the correct zigbee clusters?

MeisterQ:
> I also have those devices and would love this feature. I would also gladly help with the implementation, can someone point me in the right direction on how to find the correct zigbee clusters?

I dont have a clue how to, otherwise i would help.

grossermax:
@MeisterQ and @mrflo97
I would love to have this feature too and in addition to get more exposes for the Aqara W600.

I guess to get more exposes someone needs to sniff the Zigbee traffic for the Aqara W600 following this tutorial: https://www.zigbee2mqtt.io/advanced/zigbee/04_sniff_zigbee_traffic.html

MeisterQ:
maybe @kirovilya could help? Or @GerjandeGroot 
They were doing PR for the W600 and looking like they know what they are doing.

Pel1can111:
Picked up one of these as the sensor input showed "external" and hoped to push temp sensors from other rooms. Fingers crossed this can be added

MeisterQ:
> Picked up one of these as the sensor input showed "external" and hoped to push temp sensors from other rooms. Fingers crossed this can be added

Where?

peetCreative:
Any update here?

nilvanis:
I've got this working. I'm preparing a PR, but its really huge because of multiple custom code required.

timich:
<html>
<body>
<p>I spent the weekend reverse-engineering how the Aqara M2 hub feeds external temperature readings to the W600 TRV (model <code>WT-A03E</code>, zigbeeModel <code>lumi.airrtc.aeu005</code>, firmware 2745).</p>
[...]

PR #12942, opened 2026-08-19
fix: improve Aqara W600 external temperature handling

k-lofnes:
## Summary

Improves Aqara W600 (`WT-A03E` / `lumi.airrtc.aeu005`) external-temperature handling.

- Persist the proprietary `0xFFF2` frame sequence counter in device metadata.
- Do not configure reporting for `hvacThermostat.tempSetpointHold`; W600 devices can reject it with `INSUFFICIENT_SPACE`.
- Keep normal W600 configuration reads.
- Keep endpoint 2 thermostat reports from overwriting the active endpoint 1 climate state.
- Expose the validated endpoint 2 internal sensor reading as:
  - `local_temperature_internal`

### DIFF
--- a/src/lib/lumi.ts
+++ b/src/lib/lumi.ts
@@ -4810,9 +4810,14 @@ function writeW600LumiAttribute(entity: Zh.Endpoint, attribute: string | number,
 }
 
 function getNextW600SensorBindingCounter(entity: Zh.Device | Zh.Endpoint) {
-    const storeKey = getW600DeviceStoreKey(entity);
-    const counter = globalStore.getValue(storeKey, W600_SENSOR_BINDING_COUNTER_STORE_KEY, 0x12);
-    globalStore.putValue(storeKey, W600_SENSOR_BINDING_COUNTER_STORE_KEY, (counter + 1) & 0xff);
+    const device = "ieeeAddr" in entity ? entity : entity.getDevice();
+    const counter =
+        typeof device.meta?.[W600_SENSOR_BINDING_COUNTER_STORE_KEY] === "number" ? device.meta[W600_SENSOR_BINDING_COUNTER_STORE_KEY] : 0x12;
+
+    device.meta ??= {};
+    device.meta[W600_SENSOR_BINDING_COUNTER_STORE_KEY] = (counter + 1) & 0xff;
+    device.save();
+
     return counter;
 }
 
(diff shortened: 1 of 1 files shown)


timich:
looks good 🙂 👍

src/lib/lumi.ts
```diff
@@ -5629,21 +5634,49 @@ function createW600Thermostat(): ModernExtend {
         .withDescription("Duration in minutes for the current manual override. 0 means until next schedule event, 65535 means indefinitely.");
     extend.exposes?.push(
         e.binary("override_active", ea.STATE, true, false).withLabel("Manual Override").withDescription("Temporary manual override active"),
+        e
+            .numeric("internal_temperature", ea.STATE)
+            .withUnit("°C")
+            .withLabel("Internal radiator temperature")
```
nilvanis: Not user friendly label - it's nor radiator temperature, it's TRV internal sensor reading

src/lib/lumi.ts
```diff
@@ -5629,21 +5634,49 @@ function createW600Thermostat(): ModernExtend {
...
+            .withDescription("Temperature reported by the W600 physical radiator sensor on endpoint 2")
```
nilvanis: Unfriendly description - user does not know what is enpoint 2, its not a radiator sensor

src/lib/lumi.ts
```diff
@@ -5629,21 +5634,49 @@ function createW600Thermostat(): ModernExtend {
...
+            .withCategory("diagnostic"),
+        e
+            .numeric("endpoint_2_occupied_heating_setpoint", ea.STATE)
+            .withUnit("°C")
+            .withLabel("Endpoint 2 heating setpoint")
+            .withDescription("Heating setpoint reported by W600 endpoint 2; kept separate from the active climate target")
```
nilvanis: label & description not user friendly, user would not be able to tell what  is this

src/lib/lumi.ts
```diff
@@ -5629,21 +5634,49 @@ function createW600Thermostat(): ModernExtend {
...
     extend.exposes?.push(
         e.binary("override_active", ea.STATE, true, false).withLabel("Manual Override").withDescription("Temporary manual override active"),
+        e
+            .numeric("internal_temperature", ea.STATE)
+            .withUnit("°C")
+            .withLabel("Internal radiator temperature")
+            .withDescription("Temperature reported by the W600 physical radiator sensor on endpoint 2")
+            .withCategory("diagnostic"),
+        e
+            .numeric("endpoint_2_occupied_heating_setpoint", ea.STATE)
```
nilvanis: What is the value of exposing it to the user?

src/lib/lumi.ts
```diff
@@ -5629,21 +5634,49 @@ function createW600Thermostat(): ModernExtend {
...
         type: ["attributeReport", "readResponse"],
         convert: (model, msg, publish, options, meta) => {
-            const result = fz.thermostat.convert(model, msg, publish, options, meta) as KeyValueAny | undefined;
+            const endpoint2 = msg.endpoint.ID === 2;
+            const data = endpoint2 ? {...msg.data} : msg.data;
+            const diagnostics: KeyValue = {};
 
-            if (result && msg.data.tempSetpointHold !== undefined) {
+            if (endpoint2) {
+                if (typeof data.localTemp === "number") diagnostics.internal_temperature = data.localTemp / 100;
```
nilvanis: Should we apply the same validity guard here as `fz.thermostat` uses for `localTemp`? The generic converter suppresses impossible values below -273.15 after scaling. Since this bypasses that converter path for endpoint 2, an invalid raw value like -32768 could be published as -327.68 °C.

src/lib/lumi.ts
```diff
@@ -5629,21 +5634,49 @@ function createW600Thermostat(): ModernExtend {
...
+                if (typeof data.occupiedHeatingSetpoint === "number") {
+                    diagnostics.endpoint_2_occupied_heating_setpoint = data.occupiedHeatingSetpoint / 100;
```
nilvanis: Same question for `occupiedHeatingSetpoint`: should endpoint 2 diagnostics use the same sanity guard before publishing

nilvanis (commented):
I think the endpoint 2 handling should be tightened and renamed around semantics, not Zigbee endpoint numbers.

From the discussion in #10812, endpoint 2 appears to be the W600 internal temperature path, while endpoint 1 is the active climate path. So exposing endpoint 2 `localTemp` as an internal temperature diagnostic makes sense.

For naming, I would avoid `endpoint_2_*` in both property names and labels. That leaks Zigbee transport detail to users. If we want this to align with the existing thermostat schema style, I would suggest something like `local_temperature_internal` with label `Internal sensor temperature`: it keeps the standard `local_temperature` meaning and adds a semantic endpoint suffix.

I would also avoid exposing the endpoint 2 setpoint for now. `endpoint_2_occupied_heating_setpoint`  is not user-friendly, and I do not see clear user value in exposing it. The important part is preventing endpoint 2 from overwriting the active climate `occupied_heating_setpoint`.

Implementation-wise, I would avoid manually decoding thermostat attributes here. `fz.thermostat.convert()` already handles scaling, precision, and invalid/sentinel guards for `localTemp` and `occupiedHeatingSetpoint`. The PR bypasses that for endpoint 2.

Suggested shape:

- call `fz.thermostat.convert()` first
- if `msg.endpoint.ID === 2`, remap only `local_temperature` to `local_temperature_internal` / `internal_temperature`
- return only that diagnostic field for endpoint 2
- do not allow any other endpoint 2 thermostat fields to leak into the normal climate state
- do not expose endpoint 2 setpoint unless its semantics and user value are clear

k-lofnes:
@nilvanis Thanks, applied. This was as inteded a draft, and thanks for giving valid feedback. I have now been running this the whole day/evening and it looks stable and workin on both my radiators. One running internal schedual saved on the hardware and one have one set temp. Both report stable values, and work as expected now. 

 I renamed the diagnostic to `local_temperature_internal` with the semantic “Internal sensor temperature” label/description, removed the endpoint 2 setpoint exposure, and now run the endpoint 2 report through `fz.thermostat.convert()` before remapping only its validated local temperature.

k-lofnes:
> looks good 🙂 👍

Thanks for getting valid data 🥇

Koenkk:
thanks!

